Issue #439, opened 2020-03-11
Support failure domains


CecileRobertMichon:
"An infrastructure provider may or may not implement the failure domains feature. Failure domains gives Cluster API just enough information to spread machines out reducing the risk of a target cluster failing due to a domain outage. This is particularly useful for Control Plane providers. They are now able to put control plane nodes in different domains.

An infrastructure provider can implement this by setting the InfraCluster.Status.FailureDomains field with a map of unique keys to failureDomainSpecs as well as respecting a set Machine.Spec.FailureDomain field when creating instances.

To support migration from failure domains that were previously specified through provider-specific resources, the Machine controller will support updating Machine.Spec.FailureDomain field if Spec.FailureDomain is present and defined on the provider-defined infrastructure resource.

Please see the cluster and machine infrastructure provider specifications for more detail."

https://cluster-api.sigs.k8s.io/developer/providers/v1alpha2-to-v1alpha3.html#optional-support-failure-domains

CecileRobertMichon:
Failure domains should map to Azure availability zones, eg. https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1420/files

/help

richardcase:
@CecileRobertMichon - shall i deprecate the `AvailabilityZone` field like they have in the AWS provider? Or shall we just provide a conversion function between v1alpha2(AvailabilityZone) and v1alpha3(FailureDomain)?

I think i prefer the conversion approach as it doesn't require the support of both fields and the removal of `AvailabilityZone` in the future.

CecileRobertMichon:
@detiber do you have any guidance here?

detiber:
I think things are a bit complicated by the existence of a v1alpha3 release, I don't think doing the conversion from v1alpha2 -> v1alpha3 is feasible considering it would break the v1alpha3 api that was already published.

If backward compatibility is desired, I think you can probably do the following:
- Deprecate the existing AvailabilityZone field
- Add a new FailureDomain field
- AzureMachine controller can populate FailureDomain with the value from AvailabilityZone if available
- AzureMachine controller can prioritize Machine.Spec.FailureDomain, over AzureMachine.Spec.FailureDomain, over AzureMachine.Spec.AvailabilityZone
- conversion webhook to restore AzureMachine.Spec.FailureDomain from annotation 

[...]

PR #572, opened 2020-04-28
feat: add failure domain support

richardcase:

Started to add failure domain support.

Relates To #439 

**Release note**:

```release-note
Support for failure domains has been introduced. The list of availability zones for a region will be stored in the status of the AzureCluster and Cluster API will use these to try and spread Machine's across availaibility zones. AvailabilityZone on AzureMachine is deprecated and FailureDomain should be used instead.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,9 +73,6 @@ PULL_POLICY ?= Always
 
 CLUSTER_TEMPLATE ?= cluster-template.yaml
 
-# Allow running the manager locally for debugging
-DEBUG_DEPLOY_MANAGER ?= true
-
 ## --------------------------------------
 ## Help
 ## --------------------------------------
@@ -336,7 +333,6 @@ create-management-cluster: $(KUSTOMIZE) $(ENVSUBST)
 	kubectl wait --for=condition=Ready --timeout=5m -n capi-kubeadm-bootstrap-system pod -l cluster.x-k8s.io/provider=bootstrap-kubeadm
 	kubectl wait --for=condition=Ready --timeout=5m -n capi-kubeadm-control-plane-system pod -l cluster.x-k8s.io/provider=control-plane-kubeadm
 
-ifeq ($(DEBUG_DEPLOY_MANAGER), true)
 	# Wait for CAPZ pods
 	kubectl wait --for=condition=Ready --timeout=5m -n capz-system pod -l cluster.x-k8s.io/provider=infrastructure-azure
 
@@ -359,11 +355,6 @@ create-workload-cluster: $(ENVSUBST)
 	kubectl --kubeconfig=./kubeconfig apply -f templates/addons/calico.yaml
 
 	@echo 'run "kubectl --kubeconfig=./kubeconfig ..." to work with the new target cluster'
-else
-	# Delete the deployment and webhooks as we will be running/debugging locally
-	kubectl delete deployment capz-controller-manager -n capz-system
-	kubectl delete ValidatingWebhookConfiguration capz-validating-webhook-configuration 
-endif
 
 .PHONY: create-cluster
 create-cluster: create-management-cluster create-workload-cluster ## Create a workload development Kubernetes cluster on Azure in a kind management cluster.
--- a/Makefile
+++ b/Makefile
@@ -73,6 +73,9 @@ PULL_POLICY ?= Always
 
 CLUSTER_TEMPLATE ?= cluster-template.yaml
 
+# Allow running the manager locally for debugging
+DEBUG_DEPLOY_MANAGER ?= true
+
 ## --------------------------------------
 ## Help
 ## --------------------------------------
@@ -333,6 +336,7 @@ create-management-cluster: $(KUSTOMIZE) $(ENVSUBST)
 	kubectl wait --for=condition=Ready --timeout=5m -n capi-kubeadm-bootstrap-system pod -l cluster.x-k8s.io/provider=bootstrap-kubeadm
 	kubectl wait --for=condition=Ready --timeout=5m -n capi-kubeadm-control-plane-system pod -l cluster.x-k8s.io/provider=control-plane-kubeadm
 
+ifeq ($(DEBUG_DEPLOY_MANAGER), true)
 	# Wait for CAPZ pods
 	kubectl wait --for=condition=Ready --timeout=5m -n capz-system pod -l cluster.x-k8s.io/provider=infrastructure-azure
 
@@ -355,6 +359,11 @@ create-workload-cluster: $(ENVSUBST)
 	kubectl --kubeconfig=./kubeconfig apply -f templates/addons/calico.yaml
 
 	@echo 'run "kubectl --kubeconfig=./kubeconfig ..." to work with the new target cluster'
+else
+	# Delete the deployment and webhooks as we will be running/debugging locally
+	kubectl delete deployment capz-controller-manager -n capz-system
+	kubectl delete ValidatingWebhookConfiguration capz-validating-webhook-configuration 
+endif
 
 .PHONY: create-cluster
 create-cluster: create-management-cluster create-workload-cluster ## Create a workload development Kubernetes cluster on Azure in a kind management cluster.
--- a/api/v1alpha2/azuremachine_conversion.go
+++ b/api/v1alpha2/azuremachine_conversion.go
@@ -40,6 +40,8 @@ func (src *AzureMachine) ConvertTo(dstRaw conversion.Hub) error { // nolint
 		dst.Spec.Identity = restored.Spec.Identity
 	}
 
+	dst.Spec.FailureDomain = restored.Spec.FailureDomain
+
 	return nil
 }
 
--- a/api/v1alpha3/azurecluster_types.go
+++ b/api/v1alpha3/azurecluster_types.go
@@ -50,8 +50,12 @@ type AzureClusterSpec struct {
 type AzureClusterStatus struct {
 	Network Network `json:"network,omitempty"`
 
-	// FailureDomains specifies the list of unique failure domains for the location of the cluster.
-	// This list will be used by Cluster API to try and spread the machines across thsese domains.
+	// FailureDomains specifies the list of unique failure domains for the location/region of the cluster.
+	// A FailureDomain maps to Availability Zone with an Azure Region (if the region support them). An
+	// Availability Zone is a separate data center within a region and they can be used to ensure
+	// the cluster is more resilient to failure.
+	// See: https://docs.microsoft.com/en-us/azure/availability-zones/az-overview
+	// This list will be used by Cluster API to try and spread the machines across the failure domains.
 	FailureDomains clusterv1.FailureDomains `json:"failureDomains,omitempty"`
 
 	Bastion VM `json:"bastion,omitempty"`
--- a/api/v1alpha3/azurecluster_types.go
+++ b/api/v1alpha3/azurecluster_types.go
@@ -50,6 +50,10 @@ type AzureClusterSpec struct {
 type AzureClusterStatus struct {
 	Network Network `json:"network,omitempty"`
 
+	// FailureDomains specifies the list of unique failure domains for the location of the cluster.
+	// This list will be used by Cluster API to try and spread the machines across thsese domains.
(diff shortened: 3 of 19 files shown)


api/v1alpha2/azuremachine_conversion.go
```diff
@@ -81,6 +86,13 @@ func Convert_v1alpha3_AzureMachineSpec_To_v1alpha2_AzureMachineSpec(in *infrav1a
 		return err
 	}
 
+	if in.FailureDomain != nil {
+		out.AvailabilityZone = AvailabilityZone{
```
detiber, line 90: Should this first test that it isn't overriding an existing value for out.AvailabilityZone first?

api/v1alpha3/azurecluster_types.go
```diff
@@ -50,6 +50,8 @@ type AzureClusterSpec struct {
 type AzureClusterStatus struct {
 	Network Network `json:"network,omitempty"`
 
+	FailureDomains clusterv1.FailureDomains `json:"failureDomains,omitempty"`
```
detiber, line 53: There should probably also be a conversion in place for AzureCluster.Status to avoid lossy conversions.

cloud/scope/machine.go
```diff
@@ -103,7 +103,16 @@ func (m *MachineScope) Location() string {
 
 // AvailabilityZone returns the AzureMachine Availability Zone.
 func (m *MachineScope) AvailabilityZone() string {
-	return *m.AzureMachine.Spec.AvailabilityZone.ID
+	// If an explicit AZ is on the azure machine use that
```
detiber, line 106: To best support the failure domain support for KubeadmControlPlane this should probably:
- prioritize the Machine.Spec.FailureDomain field (if set)
- fallback to AzureMachine.Spec.FailureDomain field (if set)
- fallback to AzureMachine.Spec.AvailabilityZone.ID (if set)

The controller should also likely:
- If AzureMachine.Spec.FailureDomain is not set, set it to the value of AzureMachineSpec.AvailabilityZone.ID

This will allow the Cluster API Machine controller to pick up the value and update Machine.Spec.FailureDomain accordingly.

Makefile
```diff
@@ -339,6 +339,29 @@ delete-cluster: $(CLUSTERCTL) ## Deletes the example Kubernetes Cluster "cluster
 kind-reset: ## Destroys the "clusterapi" kind cluster.
 	kind delete cluster --name=clusterapi || true
 
+.PHONY: create-debug-cluster
+create-debug-cluster: $(CLUSTERCTL) $(ENVSUBST) ## Create a development Kubernetes cluster on Azure in a KIND management cluster.
```
CecileRobertMichon, line 343: why do you need this? Can [tilt](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/docs/development.md#tilt-for-dev-in-capz) or the[ /hack/create-dev-cluster.sh script ](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/docs/development.md#creating-a-dev-cluster) methods meet your debugging needs?

richardcase: I would like a kind cluster with everything installed except the manager, so i can then run the manager locally pointing to the cluster. I'll remove the make target and  add an option to allow not deploying the manager.

api/v1alpha3/azurecluster_types.go
richardcase: I will add this conversion 👍

cloud/scope/machine.go
richardcase: Yes that makes sense. I hadn't quite worked out the precedence yet, so this really helps. Thanks.

api/v1alpha2/azuremachine_conversion.go
richardcase: I've added a change to do this

api/v1alpha3/azurecluster_types.go
richardcase: Just added that

cloud/scope/machine.go
richardcase: I've added the order define above

Makefile
richardcase: I removed this target and made the existing one configurable

api/v1alpha2/azuremachine_conversion.go
```diff
@@ -72,6 +72,14 @@ func Convert_v1alpha2_AzureMachineSpec_To_v1alpha3_AzureMachineSpec(in *AzureMac
 		return err
 	}
 
+	if in.AvailabilityZone.ID != nil {
+		// If there is a AZ specified then copy it to failure domain
+		// and then nil the AZ
+		out.FailureDomain = in.AvailabilityZone.ID
+		out.AvailabilityZone.ID = nil
+		out.AvailabilityZone.Enabled = nil
```
richardcase, line 80: @detiber @CecileRobertMichon  - On the conversion v1alpha2->v1alpha3 does this make sense? If there is an AZ specified when we convert this gets moved to FailureDomain and the value in AvailabilityZone is removed. Of should they both be filled in?

detiber: In order to facilitate the setting of Spec.FailureDomain when Spec.AvailabilityZone is set to get the data bubbled up to the Machine resource, this should probably be done in the AzureMachine controller rather than trying to handle it in the conversions. Especially since the field is being added after the initial v1alpha3 release and there will be v1alpha3 resources that will need to have the migration done.

detiber:
These changes lgtm

api/v1alpha2/azuremachine_conversion.go
richardcase: Thanks for your feedback on this, what you suggested made it simpler.

CecileRobertMichon:
@richardcase please rebase and add a release note

/ok-to-test

Makefile
```diff
@@ -309,9 +312,10 @@ create-cluster: $(CLUSTERCTL) $(ENVSUBST) ## Create a development Kubernetes clu
 	kubectl wait --for=condition=Ready --timeout=5m -n capi-kubeadm-bootstrap-system pod -l cluster.x-k8s.io/provider=bootstrap-kubeadm
 	kubectl wait --for=condition=Ready --timeout=5m -n capi-kubeadm-control-plane-system pod -l cluster.x-k8s.io/provider=control-plane-kubeadm
 
+ifeq ($(DEBUG_DEPLOY_MANAGER), true)
```
CecileRobertMichon, line 315: I'm curious how you use this and why it's better than running on tilt for debugging. If you think it might be helpful to others and we should merge this, I'd love to see a short blurb in development docs on how to use that option.

Makefile
```diff
@@ -309,9 +312,10 @@ create-cluster: $(CLUSTERCTL) $(ENVSUBST) ## Create a development Kubernetes clu
...
 	# Wait for CAPZ pods
 	kubectl wait --for=condition=Ready --timeout=5m -n capz-system pod -l cluster.x-k8s.io/provider=infrastructure-azure
-	
+
```
CecileRobertMichon, line 318:
```suggestion

```

api/v1alpha3/azurecluster_types.go
CecileRobertMichon, line 53: let's add a comment here to describe how the field is used

api/v1alpha3/azuremachine_types.go
```diff
@@ -34,7 +34,13 @@ type AzureMachineSpec struct {
...
 	ProviderID *string `json:"providerID,omitempty"`
 
-	VMSize           string           `json:"vmSize"`
+	VMSize string `json:"vmSize"`
+
+	// FailureDomain is the failure domain unique identifier this Machine should be attached to,
+	// as defined in Cluster API. This relates to an Azure Availability Zone
+	FailureDomain *string `json:"failureDomain,omitempty"`
+
+	// DEPRECATED: use FailureDomain instead
```
CecileRobertMichon, line 43: would it make sense to print a warning if `availabilityZone` is used?

api/v1alpha3/azurecluster_types.go
CecileRobertMichon: Also, can you please add a short doc in docs/topics to explain to use Failure Domains with Azure?

cloud/services/availabilityzones/availabilityzones_test.go
```diff
@@ -97,7 +114,7 @@ func TestGetAvailabilityZones(t *testing.T) {
 				Cluster: cluster,
 				AzureCluster: &infrav1.AzureCluster{
 					Spec: infrav1.AzureClusterSpec{
-						Location: "test-location",
+						Location:      "centralus",
```
CecileRobertMichon, line 117: extra indent

Makefile
richardcase: I use it mainly for debugging the manager locally (perhaps there is a better way with tilt that i'm missing?)

My debugging workflow is:
1. Run ` DEBUG_DEPLOY_MANAGER=false make create-cluster` so that i get a kind cluster with everything deployed except the capz manager and webhook.
2. Set breakpoints in the capz code somewhere 
3. Use a VSCode debug launch configuration to launch the manager
4. Apply cluster yaml

Perhaps instead of the `DEBUG_DEPLOY_MANAGER` option i could create a new kustomization.yaml that excluded the manager

Makefile
richardcase: Good spot! Removed

api/v1alpha3/azuremachine_types.go
richardcase: Good idea, i've added a warning.

cloud/services/availabilityzones/availabilityzones_test.go
richardcase: Tried to change this but `go fmt` but the change back

Makefile
devigned: Initially, I started going down the path of debugging the controller running locally too. It's so much easier to attach a debugger and iterate. My main sticking point was getting the webhooks to work since running them outside the cluster would make them unreachable for conversions and validations (perhaps, I didn't have my setup configured correctly).

Unfortunately, I can see in this PR how you have the processes configured to start, so it's not clear how everything is wired up.

I'm super interested a dev scenarios where we can develop / debug running the manager and webhooks locally.

As interested as I am in that scenario, it would probably be best to separate it from this feature. Can these changes get broken out into a new PR?

richardcase: Getting the webhooks to work has been my sticking point as well. As i was making a change to the manager i cheated and did `kubectl delete ValidatingWebhookConfiguration capz-validating-webhook-configuration` to get around it.

Having a good story around debugging would really help....especially with my sanity when trying to work out why something isn't working. 

I wondered if something like this might work:
1) For debugging change the entrypoint from /manager to using delve with remote debugging enabled `[ "dlv", "debug", "/manager", "--listen=:35000", "--headless=true", "--api-version=2", "--log" ]`
2) Patch the manager deployment to expose the debugger port
3) Start kind exposing the debugging port using something like the following:
```yaml
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
- role: control-plane
  extraPortMappings:
  - containerPort: 35000
    hostPort: 35000
    protocol: TCP
```
4) Create a VSCode launch configuration that instructs delve to debug remotely using the port used above

I will remove this change from the PR.

Shall i create an issue to discuss to dev/debugging experience? I will try the above.

devigned: I think it would be valuable to have a debugging experience like you've described. I think it can also be achieved via a running container.

Yeah, please open an issue and let's discuss how to make this better :).

richardcase:
@CecileRobertMichon  - do you have any specific docs in mind that i should add?

api/v1alpha3/azurecluster_types.go
CecileRobertMichon: @richardcase see the two comments above about docs ^

richardcase: Ah sorry @CecileRobertMichon, i missed these comments. I will make the changes.

richardcase: Docs added.

PR #583, opened 2020-04-30
Deploy clusters into multiple subscriptions from single CAPZ manager

### LINKED ISSUE #480: Deploy to Multiple Subscriptions from Single Manager

***Goals***
1. Enable deployment of clusters into multiple subscriptions

***Non-Goals/Future Work***
1. Deployment of single cluster spanning multiple subscriptions
1. Deployment of clusters into multiple Tenants
1. Deployment of clusters using unique Service Principals

**User Story**

As an operator I would like to deploy clusters for independent teams into their corresponding subscriptions so that costs associated with those clusters can be attributed to the correct team via Azure Billing.

**Detailed Description**

Large Azure customers generally deploy across a number of subscriptions to enable better financial record keeping and billing for teams that may have independent finance teams. These subscriptions are managed via Management Groups to ensure that large organizations can associate policy and governance restrictions over those teams. 

Customer Central IT Kubernetes Operations teams will want to comply with their Central IT Azure Operations teams to ensure smooth and effective collaboration. 

To support this mode of operation a single CAPZ manager should be Subscription Agnostic and should instead rely on the Cluster configuration objects to specify SubscriptionID.

Remove Subscription ID fields from all components deployed with the CAPZ controller manager.
Use Subscription ID fields from all Cluster data structures

/kind proposal

jseely:
Implements #480, allows operators to manage clusters across multiple subscriptions without the need to deploy multiple CAPZ managers.

Fixes #480 

**Release note**:

```release-note
* Adds the ability to provide `spec.subscriptionID` on `AzureCluster` to deploy cluster into a subscription other than the default set in `capz-controller-manager` via the `AZURE_SUBSCRIPTION_ID` environment variable. 
* Provides backwards compatibility in `v1alpha2` via the `azurecluster.infrastructure.cluster.x-k8s.io/subscriptionID` annotation.
```

### DIFF
--- a/api/v1alpha3/azurecluster_types.go
+++ b/api/v1alpha3/azurecluster_types.go
@@ -34,6 +34,8 @@ type AzureClusterSpec struct {
 
 	ResourceGroup string `json:"resourceGroup"`
 
+	SubscriptionID string `json:"subscriptionID,omitempty"`
+
 	Location string `json:"location"`
 
 	// ControlPlaneEndpoint represents the endpoint used to communicate with the control plane.
@@ -69,6 +71,7 @@ type AzureClusterStatus struct {
 // +kubebuilder:printcolumn:name="Cluster",type="string",JSONPath=".metadata.labels.cluster\\.x-k8s\\.io/cluster-name",description="Cluster to which this AzureCluster belongs"
 // +kubebuilder:printcolumn:name="Ready",type="boolean",JSONPath=".status.ready"
 // +kubebuilder:printcolumn:name="Resource Group",type="string",priority=1,JSONPath=".spec.resourceGroup"
+// +kubebuilder:printcolumn:name="SubscriptionID",type="string",priority=1,JSONPath=".spec.subscriptionID"
 // +kubebuilder:printcolumn:name="Location",type="string",priority=1,JSONPath=".spec.location"
 // +kubebuilder:printcolumn:name="Endpoint",type="string",priority=1,JSONPath=".spec.controlPlaneEndpoint.host",description="Control Plane Endpoint"
 // +kubebuilder:resource:path=azureclusters,scope=Namespaced,categories=cluster-api
--- a/cloud/scope/clients.go
+++ b/cloud/scope/clients.go
@@ -30,32 +30,28 @@ type AzureClients struct {
 	Authorizer     autorest.Authorizer
 }
 
-func (c *AzureClients) setCredentials() error {
-	if c.SubscriptionID == "" {
-		subID, err := getSubscriptionID()
-		if err != nil {
-			return err
-		}
-		c.SubscriptionID = subID
+func (c *AzureClients) setCredentials(subscriptionID string) error {
+	subID, err := getSubscriptionID(subscriptionID)
+	if err != nil {
+		return err
 	}
-	if c.Authorizer == nil {
-		auth, err := getAuthorizer()
-		if err != nil {
-			return err
-		}
-		c.Authorizer = auth
+	c.SubscriptionID = subID
+	settings, err := auth.GetSettingsFromEnvironment()
+	if err != nil {
+		return err
 	}
-	return nil
+	settings.Values[auth.SubscriptionID] = subscriptionID
+	c.Authorizer, err = settings.GetAuthorizer()
+	return err
 }
 
-func getSubscriptionID() (string, error) {
-	subscriptionID := os.Getenv("AZURE_SUBSCRIPTION_ID")
+func getSubscriptionID(subscriptionID string) (string, error) {
+	if subscriptionID != "" {
+		return subscriptionID, nil
+	}
+	subscriptionID = os.Getenv("AZURE_SUBSCRIPTION_ID")
 	if subscriptionID == "" {
 		return "", errors.New("error creating azure services. Environment variable AZURE_SUBSCRIPTION_ID is not set")
 	}
 	return subscriptionID, nil
-}
-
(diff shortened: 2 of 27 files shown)


jseely:
@justaugustus & @devigned, it seems some tests are failing in CI around `AZURE_SUBSCRIPTION_ID`. Is this something I broke? Is there something I need to do to fix the Mocks?

CecileRobertMichon:
@jseely that’s because the previous implementation would only get the subscription ID if it wasn’t already set on the client https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/583/files#diff-b3bc16e73337567b153a9fb4f621a745L34 and we used that in the test to set the sub ID to a fake sub id. You would have to add that back or, and that’s maybe a bit more work but I think is a better solution, change the tests to set the fake sub id on the new subscriptionID field in azureCluster you added instead of on the client.

Here’s an example of what I’m referring to:

https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/cloud/services/subnets/subnets_test.go

jseely:
@CecileRobertMichon oh that makes sense. I'll go through the failing tests and update them to set the fake subscription ID in the AzureCluster. 

Thanks for the help!

detiber:
If I'm understanding this change correctly this is allowing the use of multiple SubscriptionIDs as long as the env-provided credentials for the manager have access to the given SubscriptionIDs.

To leverage multiple SubscriptionIDs with separate sets of credentials would still require a separate manager, correct?

devigned:
@detiber, the intention of the PR is to support multiple subscriptions with a single set of credentials, though it might not be explained adequately.

A non-goal of this PR is to use multiple credentials in one instance of the provider as that would differ from how other providers handle multi-tenancy.

Each manager instance only has one set of credentials.

detiber:
@devigned main reason I bring it up is that we are currently looking at adding account multi-tenancy to the AWS provider (and also the vSphere provider). 

devigned:
Well, I though one credential for each manager instance would last longer, but figured we'd get to real multi-tenancy eventually.

Thank you for the heads up. We'll review, digest and think about how we'd support that model.

cloud/services/disks/disks_test.go
```diff
@@ -140,15 +145,16 @@ func TestDeleteDisk(t *testing.T) {
 
 			clusterScope, err := scope.NewClusterScope(scope.ClusterScopeParams{
 				AzureClients: scope.AzureClients{
-					SubscriptionID: "123",
+					SubscriptionID: subscriptionID,
```
CecileRobertMichon, line 148:
```suggestion
```
you shouldn't need to set `SubscriptionID`both in `AzureClients` and `AzureCluster.Spec`, only setting it in `AzureCluster.Spec` should be enough (and the one set on AzureClients is a no-op since we're not checking for it in setCredentials anymore). 

Same comment for other test files.

api/v1alpha3/azurecluster_types.go
```diff
@@ -61,6 +63,7 @@ type AzureClusterStatus struct {
 // +kubebuilder:printcolumn:name="Cluster",type="string",JSONPath=".metadata.labels.cluster\\.x-k8s\\.io/cluster-name",description="Cluster to which this AzureCluster belongs"
 // +kubebuilder:printcolumn:name="Ready",type="boolean",JSONPath=".status.ready"
 // +kubebuilder:printcolumn:name="Resource Group",type="string",priority=1,JSONPath=".spec.resourceGroup"
+// +kubebuilder:printcolumn:name="SubscriptionID",type="string",priority=1,JSONPath=".spec.subscriptionID"
```
CecileRobertMichon, line 66: thanks for adding this 👍

cloud/scope/clients.go
```diff
@@ -48,14 +47,53 @@ func (c *AzureClients) setCredentials() error {
...
-	return auth.NewAuthorizerFromEnvironment()
+func getAuthorizer(subscriptionID string) (autorest.Authorizer, error) {
+	settings, err := getSettingsFromEnvironment(subscriptionID)
+	if err != nil {
+		return nil, err
+	}
+	return settings.GetAuthorizer()
+}
+
+func getSettingsFromEnvironment(subscriptionID string) (s auth.EnvironmentSettings, err error) {
```
CecileRobertMichon, line 68: I just realized this is duplicating code from the Azure go-autorest https://github.com/Azure/go-autorest/blob/master/autorest/azure/auth/auth.go#L94, is there any way we can achieve what we need without copying the whole function here? We should avoid maintaining duplicated code since the code might change in the future to add new options and we wouldn't get those updates. cc @devigned 

Sorry I missed this in the first review round.

devigned: Perhaps, just call the SDK `GetSettingsFromEnvironment`, which will return an `auth.EnvironmentSettings`, and then replace the subscriptionID value.

CecileRobertMichon: > then replace the subscriptionID value.

Make sure this is only done if and only if `AzureClusterSpec.SubscriptionID != "'` as we still want to allow using the environment variable for subscription ID since the spec field is optional for now.

jseely: Good call! I knew that was kind of ugly but didn't take the time to investigate the options...

CecileRobertMichon:
@jseely looks like there's a go fmt issue, probably from the rebase

jseely:
@CecileRobertMichon or @devigned if you could throw a `/lgtm` for me, I'd really appreciate it!

PR #592, opened 2020-05-05
Update NSG only if default rules are not present, or else skip the update

### LINKED ISSUE #578: LoadBalancer service is not accessible
/kind bug

Create a LoadBalancer service, public IP is allocated, but the service is not accessible through public IP.

kube-controller-manager is adding correct inbound security rules to node-nsg with destination to be the allocated public IP. But capz securitygroup reconcile reset the rules to default and remove those rules.

**Environment:**

- cluster-api-provider-azure version: v0.4.2

snehala27:
The security groups: controlplane-nsg and node-nsg are updated during reconcilation without checking for rules thus resetting the rules added by any other source(like rules added by kube-controller-manager for LoadBalancer service).

Resolution:
Skip NSG updates if default rules are present
The flow will be:
    - Create NSG if does not exist
    - control-plane-nsg: update only if default rules are not present else skip update
    - node-nsg: skip update as no default rules are present for node-nsg
    - Updates will occur with ETAGS set in the header

**Which issue(s) this PR fixes** :
Fixes #578 

**Release note**:

```release-note
During reconcile, updates to control-plane-nsg will take place only if default rules(ssh-rule and api-server rule) is not present. The update will be skipped otherwise. 
For node-nsg, as node-nsg does not have default rules, updates will be skipped
```

### DIFF
--- a/cloud/services/securitygroups/client.go
+++ b/cloud/services/securitygroups/client.go
@@ -18,7 +18,6 @@ package securitygroups
 
 import (
 	"context"
-
 	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2019-06-01/network"
 	"github.com/Azure/go-autorest/autorest"
 	azure "sigs.k8s.io/cluster-api-provider-azure/cloud"
@@ -59,10 +58,25 @@ func (ac *AzureClient) Get(ctx context.Context, resourceGroupName, sgName string
 
 // CreateOrUpdate creates or updates a network security group in the specified resource group.
 func (ac *AzureClient) CreateOrUpdate(ctx context.Context, resourceGroupName string, sgName string, sg network.SecurityGroup) error {
-	future, err := ac.securitygroups.CreateOrUpdate(ctx, resourceGroupName, sgName, sg)
+	var etag string
+	if sg.Etag != nil {
+		etag = *sg.Etag
+	}
+	req, err := ac.securitygroups.CreateOrUpdatePreparer(ctx, resourceGroupName, sgName, sg)
 	if err != nil {
+		err = autorest.NewErrorWithError(err, "network.SecurityGroupsClient", "CreateOrUpdate", nil, "Failure preparing request")
 		return err
 	}
+	if etag != "" {
+		req.Header.Add("If-Match", etag)
+	}
+
+	future, err := ac.securitygroups.CreateOrUpdateSender(req)
+	if err != nil {
+		err = autorest.NewErrorWithError(err, "network.SecurityGroupsClient", "CreateOrUpdate", future.Response(), "Failure sending request")
+		return err
+	}
+
 	err = future.WaitForCompletionRef(ctx, ac.securitygroups.Client)
 	if err != nil {
 		return err
--- a/cloud/services/securitygroups/securitygroups.go
+++ b/cloud/services/securitygroups/securitygroups.go
@@ -19,12 +19,18 @@ package securitygroups
 import (
 	"context"
 	"strconv"
+	"strings"
 
 	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2019-06-01/network"
 	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/pkg/errors"
 	"k8s.io/klog"
 	azure "sigs.k8s.io/cluster-api-provider-azure/cloud"
+)
+
+const (
+	apiServerRule = "apiServerRule"
+	sshRule       = "sshRule"
 )
 
 // Spec specification for network security groups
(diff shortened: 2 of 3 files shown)


CecileRobertMichon:
Thanks for the PR @snehala27! Will try to review tomorrow.

cloud/services/securitygroups/client.go
```diff
@@ -71,6 +86,33 @@ func (ac *AzureClient) CreateOrUpdate(ctx context.Context, resourceGroupName str
 	return err
 }
 
+// createOrUpdatePreparer prepares the CreateOrUpdate request.
+func (ac *AzureClient) createOrUpdatePreparer(ctx context.Context, resourceGroupName string, networkSecurityGroupName string, parameters network.SecurityGroup, etag string) (*http.Request, error) {
```
devigned, line 90: This content of this func should probably be the following, not building the Preparer from scratch:
```go
// ac.CreateOrUpdatePreparer would expose ac.securitygroups.CreateOrUpdatePreparer 
req, err := ac.CreateOrUpdatePreparer(ctx, resourceGroupName, networkSecurityGroupName, parameters)
if err != nil {
    return nil, err
}

req.Header.Add("If-Match", etag)
```

CecileRobertMichon: I opened https://github.com/Azure/azure-sdk-for-go/issues/8917 to ask about plans to build this into the Go SDK.

snehala27: @devigned Removed the preparer as suggested.

cloud/services/securitygroups/client.go
```diff
@@ -58,11 +57,22 @@ func (ac *AzureClient) Get(ctx context.Context, resourceGroupName, sgName string
 }
 
 // CreateOrUpdate creates or updates a network security group in the specified resource group.
-func (ac *AzureClient) CreateOrUpdate(ctx context.Context, resourceGroupName string, sgName string, sg network.SecurityGroup) error {
-	future, err := ac.securitygroups.CreateOrUpdate(ctx, resourceGroupName, sgName, sg)
+func (ac *AzureClient) CreateOrUpdate(ctx context.Context, resourceGroupName string, sgName string, sg network.SecurityGroup, etag string) error {
```
devigned, line 60: Etag is on `network.SecurityGroup`. Perhaps, rather than adding it to the signature of the func, make a copy of the etag field before calling `ac.securitygroups.CreateOrUpdatePreparer`, then add the copy to the header.

I believe `ac.securitygroups.CreateOrUpdatePreparer` will set the `Etag *string` ptr to nil b/c if it were to be included in the body of the PUT, the service would reject the request with a 400 error.

snehala27: Updated the PR

cloud/services/securitygroups/securitygroups.go
```diff
@@ -59,52 +66,58 @@ func (s *Service) Reconcile(ctx context.Context, spec interface{}) error {
 		return errors.New("invalid security groups specification")
 	}
 
-	securityRules := &[]network.SecurityRule{}
+	securityGroup, err := s.Client.Get(ctx, s.Scope.ResourceGroup(), nsgSpec.Name)
+	if err != nil && !azure.ResourceNotFound(err) {
+		return err
```
CecileRobertMichon, line 71: let's wrap the error here

cloud/services/securitygroups/securitygroups.go
```diff
@@ -59,52 +66,58 @@ func (s *Service) Reconcile(ctx context.Context, spec interface{}) error {
...
+				klog.V(2).Infof("security group %s exists and skip update", nsgSpec.Name)
+				return nil
+			}
+		} else {
+			klog.V(2).Infof("using additional rules for control plane %s", nsgSpec.Name)
+			securityRules = append(securityRules, getRule(sshRule), getRule(apiServerRule))
 		}
+	} else if nsgExists {
+		// Skip update for node NSG as no default rules are required
+		klog.V(2).Infof("security group %s exists and skip update", nsgSpec.Name)
```
CecileRobertMichon, line 106:
```suggestion
		klog.V(2).Infof("security group %s exists and no default rules are required, skipping update", nsgSpec.Name)
```

cloud/services/securitygroups/securitygroups.go
```diff
@@ -59,52 +66,58 @@ func (s *Service) Reconcile(ctx context.Context, spec interface{}) error {
...
+			update := false
+			for _, rule := range defaultRules {
+				if !ruleExists(securityRules, rule) {
+					update = true
+					securityRules = append(securityRules, getRule(rule))
+				}
+			}
+			if !update {
+				// Skip update for control-plane NSG as the required default rules are present
+				klog.V(2).Infof("security group %s exists and skip update", nsgSpec.Name)
```
CecileRobertMichon, line 97:
```suggestion
				klog.V(2).Infof("security group %s exists and no default rules are missing, skipping update", nsgSpec.Name)
```

cloud/services/securitygroups/securitygroups.go
```diff
@@ -59,52 +66,58 @@ func (s *Service) Reconcile(ctx context.Context, spec interface{}) error {
...
+				return nil
+			}
+		} else {
+			klog.V(2).Infof("using additional rules for control plane %s", nsgSpec.Name)
```
CecileRobertMichon, line 101:
```suggestion
			klog.V(2).Infof("applying missing default rules for control plane NSG %s", nsgSpec.Name)
```

cloud/services/securitygroups/securitygroups.go
```diff
@@ -59,52 +66,58 @@ func (s *Service) Reconcile(ctx context.Context, spec interface{}) error {
...
-			},
+	sg := network.SecurityGroup{
+		Location: to.StringPtr(s.Scope.Location()),
+		SecurityGroupPropertiesFormat: &network.SecurityGroupPropertiesFormat{
+			SecurityRules: &securityRules,
 		},
-	)
+	}
+	if nsgExists {
+		sg.Etag = securityGroup.Etag
```
CecileRobertMichon, line 117:
```suggestion
                 // We append the existing NSG etag to the header to ensure we only apply the updates if the NSG has not been modified.
		sg.Etag = securityGroup.Etag
```

cloud/services/securitygroups/securitygroups.go
```diff
@@ -113,6 +126,45 @@ func (s *Service) Reconcile(ctx context.Context, spec interface{}) error {
 	return err
 }
 
+func ruleExists(rules []network.SecurityRule, rule controlPlaneDefaultRule) bool {
+	for _, existingRule := range rules {
```
CecileRobertMichon, line 130: nit: for better code performance, we could put the rules in a map[string]network.SecurityRule before passing them into the ruleExists function, that way we don't have to loop through the rules for each default rules and the cost is O(1) instead of O(n). This isn't a very big deal here because we only have 2 default rules so the cost is pretty small anyways but using a map would scale better.

PR #594, opened 2020-05-06
cleanup: remove unused VM extensions service

CecileRobertMichon:
**What this PR does / why we need it**: We don't use VM extensions in v1alpha3, removing the reconciler logic since it's dead code. 

Fixes #

**Special notes for your reviewer**: I left the Azure client as-is since it might be useful later in other services and the client code is generic enough that we wouldn't need to update it to use it. If anyone feels strongly that we should just remove the whole package I'm happy to discuss.

**Release note**:

```release-note
NONE
```


nader-ziada:
@CecileRobertMichon needs rebase, otherwise lgtm

PR #617, opened 2020-05-14
🏃extend ci-conformance.sh testing capability 

chewong:
Extending `ci-conformance.sh` testing capability to allow other tests to be run on a CAPZ cluster - a temporary solution until https://github.com/kubernetes-sigs/cluster-api/issues/2826 is fixed. 

This PR created a modified version of `ci-conformance.sh` called `ci-entrypoint.sh` instead of directly re-naming `ci-conformance.sh` so we won't break any jobs temporarily (I will open a separate PR to delete duplicated files after all test configurations in test-infra point to `ci-entrypoint.sh`). `ci-entrypoint.sh` will be the new entrypoint for all E2E tests. It can accept k/k master branch or release branches, and has the option to run upstream E2E tests or bring your own E2E tests by appending custom commands to `./scripts/ci-entrypoint.sh`.

An example of how this script will be used in test-infra to test in-tree Azure Disk driver:

```yaml
...
  spec:
    containers:
    - image: gcr.io/k8s-testimages/kubekins-e2e:v20200509-be34c64-master
      command:
      - runner.sh
      - ./scripts/ci-entrypoint.sh
      - bash
      - -c
      - >-
        kubectl apply -f templates/addons/storageclass-azure-disk.yaml &&
        cd ${GOPATH}/src/sigs.k8s.io/azuredisk-csi-driver &&
        make e2e-test
      env:
      - name: SKIP_UPSTREAM_E2E_TESTS
        value: "true"
      - name: USE_CI_ARTIFACTS
        value: "true"
...
```

/assign @CecileRobertMichon 

Fixes #

**Release note**:

```release-note
🏃extend ci-conformance.sh testing compatibility 
```

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -350,7 +350,7 @@ You can optionally set `AZURE_SSH_PUBLIC_KEY_FILE` to use your own SSH key.
 To run Conformance locally, set `AZURE_CLIENT_ID`, `AZURE_CLIENT_SECRET`, `AZURE_SUBSCRIPTION_ID`, `AZURE_TENANT_ID` and run:
 
 ```bash
-./scripts/ci-conformance.sh
+./scripts/ci-entrypoint.sh
 ```
 
 You can optionally set the following variables:
--- a/docs/development.md
+++ b/docs/development.md
@@ -140,7 +140,7 @@ To build a kind cluster and start Tilt, just run:
 ```shell
 make tilt-up
 ```
-By default, the Cluster API components deployed by Tilt have experimental features turned off. 
+By default, the Cluster API components deployed by Tilt have experimental features turned off.
 If you would like to enable these features, add `extra_args` as specified in [The Cluster API Book](https://cluster-api.sigs.k8s.io/developer/tilt.html#create-a-tilt-settingsjson-file).
 
 Once your kind management cluster is up and running, you can [deploy a workload cluster](#deploying-a-workload-cluster).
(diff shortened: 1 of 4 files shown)


CecileRobertMichon:
@chewong what about opening a test-infra PR to change the pull conformance job to point to `ci-entrypoint.sh` right now so we can test the new script in this PR? the pull conformance job optional and not run by default so it won't block other PRs

templates/addons/storageclass-azure-disk.yaml
```diff
@@ -0,0 +1,39 @@
+apiVersion: storage.k8s.io/v1beta1
```
CecileRobertMichon, line 1: I don't see this being used in the script above. If it is independent, let's split the commit in 2 commits (same PR, just two commit messages).

chewong:
/hold
Since the jobs are already pointing to `ci-entrypoint.sh`, let me directly rename `ci-conformance.sh` to `ci-entrypoint.sh`

CecileRobertMichon:
@chewong what about the periodic jobs? I think we need to merge this as-is before changing those

chewong:
> @chewong what about the periodic jobs? I think we need to merge this as-is before changing those

Ah right, we can merge this as-is and I will modify existing periodic jobs and add additional ones before removing `ci-conformance.sh`.

chewong:
Oops I forgot to set `USE_CI_ARTIFACTS` to true in https://github.com/kubernetes/test-infra/pull/17613. That's why `pull-cluster-api-provider-azure-conformance-v1alpha3-1-18` and `pull-cluster-api-provider-azure-conformance-v1alpha3-master` are testing against 1.17.4 instead master and 1.18. Let me open another PR in test-infra.

docs/development.md
```diff
@@ -355,21 +355,28 @@ To run Conformance locally, set `AZURE_CLIENT_ID`, `AZURE_CLIENT_SECRET`, `AZURE
...
 | `PARALLEL`                  | Skip serial tests and set --ginkgo-parallel.                                                                  |
-| `USE_CI_ARTIFACTS`          | Use a CI version of Kubernetes, ie. not a released version (eg. `v1.19.0-alpha.1.426+0926c9c47677e9`)          |
-| `CI_VERSION`                | Provide a custom CI version of Kubernetes. By default, the latest master commit will be used.                  |
+| `USE_CI_ARTIFACTS`          | Use a CI version of Kubernetes, ie. not a released version (eg. `v1.19.0-alpha.1.426+0926c9c47677e9`)         |
+| `CI_VERSION`                | Provide a custom CI version of Kubernetes. By default, the latest master commit will be used.                 |
 
 You can also customize the configuration of the CAPZ cluster (assuming that `SKIP_CREATE_CLUSTER` is not set). See [Customizing the cluster deployment](#customizing-the-cluster-deployment) for more details.
 
+In addition to upstream E2E, you can append custom commands to `./scripts/ci-entrypoint.sh` to run E2E from other projects against a CAPZ cluster:
```
CecileRobertMichon, line 373: you might also want to edit https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/617/files#diff-1bf07101845de6da98f1c948b82c5ce0R353 which still refers to `ci-conformance.sh`?

Issue #627, opened 2020-05-15
Refactor: improve use of scope and middle layer Spec types in services


CecileRobertMichon:
> ⚠️ Cluster API Azure maintainers can ask to turn an issue-proposal into a [CAEP](https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/proposals/YYYYMMDD-template.md) when necessary. This is to be expected for large changes that impact multiple components, breaking changes, or new large features.

Currently, each service in https://github.com/kubernetes-sigs/cluster-api-provider-azure/tree/master/cloud/services has its own spec type defined that is used as input to create the Azure resource. For example, subnets have `Spec`: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/cloud/services/subnets/subnets.go#L33

```
// Spec input specification for Get/CreateOrUpdate/Delete calls
type Spec struct {
	Name                string
	CIDR                string
	VnetName            string
	RouteTableName      string
	SecurityGroupName   string
	Role                infrav1.SubnetRole
	InternalLBIPAddress string
}
```

The values in this spec get populated by the azuremachine or azurecluster _reconciler (in this case azureclustet since subnets are part of AzureCluster) like so: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/controllers/azurecluster_reconciler.go#L154 

However, the `Reconcile(), `Get()` and `Delete()` functions already have access to the everything inside the spec from the service [subnetSvc](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/controllers/azurecluster_reconciler.go#L163) which contains the cluster scope: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/controllers/azurecluster_reconciler.go#L43.

To simplify the code and remove the need for extra types and translation layers (api -> spec -> azure), we should remove the superfluous spec objects and instead rely on the information that is already available in `scope.AzureCluster.Spec.`, like we already do in some instances: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/cloud/services/subnets/subnets.go#L49

In addition, some of the values in "Spec" are set by default in azure_reconcile.go at the moment, eg: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/controllers/azurecluster_reconciler.go

We should move those to defaulting webhooks (see #427) so that the source of truth is the AzureCluster and AzureMachine spec.

/kind proposal

[...]

Issue #416, opened 2020-02-26
Enhancement Proposal: Break out cloud resource reconciliation


devigned:
> ⚠️ Cluster API Azure maintainers can ask to turn an issue-proposal into a [CAEP](https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/proposals/YYYYMMDD-template.md) when necessary. This is to be expected for large changes that impact multiple components, breaking changes, or new large features.

***Goals***
1. Separate concerns for cloud resource reconciliation from transformation of Cluster API custom resource specs to cloud resource specs
2. Create parallel path for reconciliation in CAPZ which uses an Azure cloud resource controller / reconciler with Azure CRDs for the cloud resources required for Cluster-API resources, rather than using direct calls to Azure in CAPZ.

***Non-Goals/Future Work***
1. Future work: possibly, choose to only rely on Azure resource CRDs for reconciliation rather than parallel implementations.

**User Story**

As a CAPZ developer I would like to only need to interact with the API server and Custom Resource Definitions of Azure types rather than having to deal with the impedance of interacting with the Azure APIs. This would enable me to focus more on the core value of CAPZ, transforming Cluster-API specs to Azure resource specs.

**Detailed Description**

CAPZ currently has 2 major concerns.
1) Transforming Cluster-API types to Azure types
2) Imperatively interacting the Azure APIs to manipulate / create / delete the Azure resources

I think it would help to speed development in CAPZ if the two concerns were broken out into separate controller / reconcilers.
1) CAPZ: responsible for translating from Cluster-API CRDs to Azure CRDs
2) YTBNC (yet to be named controller): responsible for defining Azure CRDs and reconciling them with Azure

With this enhancement CAPZ could be verified to produce the correct specifications for Azure resources given an input without needing to have a dependency on the Azure API. The vast amount of verification could be done via API server and determining if a given input, Cluster API model, gets translated into the proper Azure resource specifications. At the same time, the concern on reconciling Azure resources could be focused and tested within its own component.

### YTBNC (yet to be named controller)
Should have the following characteristics:
- Define versioned CRDs which describe Azure resources across Azure API versions
- Be able to reconcile Azure resources and provide status changes / conditions
- Resources should be able to be applied in any order and converge -- goal seeking behavior
  - All of the ordering logic is currently built into CAPZ and is kind of brittle

#### Example of a ResourceGroup
Currently
```go
      group := resources.Group{
[...]

PR #327, opened 2019-11-12
✨ support the ability for public IPs on azuremachines

rsmitty:
**What this PR does / why we need it**:

This PR will add the ability to request that public IPs be allocated for
a given azuremachine/machinedeployment. The request for allocation is
simply part of the azuremachine manifest, like so:

```
spec:
  allocatePublicIP: true
```

Signed-off-by: Spencer Smith <robertspencersmith@gmail.com>

**Release note**:

```release-note
NONE
```

PR #630, opened 2020-05-19
Refactor: Move node public IP creation/deletion to NIC service

CecileRobertMichon:
**What this PR does / why we need it**: This is in preparation for #627, specifically to solve  https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/627#issuecomment-629556702. When node public IPs were added in #327, it reused the existing `Reconcile()` function for the cluster public IP. This worked as node public IPs and the cluster public IP happened to have similar properties. However, with #627, we are trying to get away from the controller -> reconcile -> service -> Azure client model and instead just have controller -> service -> Azure client (and potentially CRD in the future, see #416). In order to do that, each type of Azure resource should have it's own `Reconcile()`. The idea of this PR is to move the node public IP lifecycle to network interfaces since they are a sub resource of network interfaces in this case, similar to how network interfaces "own" NAT rules. The alternative I considered was to have a second  `Reconcile` function in the public IP service that would dedicated to node IPs but there was some friction in having the public IP service reconcile both AzureCluster resources and AzureMachine resources (similar to the NAT rule case) so I think the NIC service is more appropriate in this case. I think we can also decide to separate it out again later if that works better once we've refactored the reconcilers. This also allows for better unit testing since we can test the behavior of AllocatePublicIP actually triggers the creation of the node IP.

Relates to #627

**Release note**:

```release-note
Refactor: Move node public IP creation/deletion to NIC service
```

### DIFF
--- a/cloud/defaults.go
+++ b/cloud/defaults.go
@@ -116,6 +116,11 @@ func GeneratePublicIPName(clusterName, hash string) string {
 	return fmt.Sprintf("%s-%s", clusterName, hash)
 }
 
+// GenerateNodePublicIPName generates a node public IP name, based on the NIC name.
+func GenerateNodePublicIPName(nicName string) string {
+	return fmt.Sprintf("%s-public-ip", nicName)
+}
+
 // GenerateFQDN generates a fully qualified domain name, based on the public IP name and cluster location.
 func GenerateFQDN(publicIPName, location string) string {
 	return fmt.Sprintf("%s.%s.%s", publicIPName, location, DefaultAzureDNSZone)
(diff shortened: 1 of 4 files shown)


CecileRobertMichon:
E2E is failing because quay.io is down https://status.quay.io/ :(

devigned:
LGTM pending passing e2e

nader-ziada:
is there anything we can remove from the publicips package? I don't think the reconcile is used anymore

CecileRobertMichon:
> is there anything we can remove from the publicips package? I don't think the reconcile is used anymore
@nader-ziada it is, to reconcile the AzureCluster public IP https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/controllers/azurecluster_reconciler.go#L211

I'll do further cleanup as part of #627 but I wanted to keep this change scoped to what it is, dedup the usage of public IP reconcile()

PR #637, opened 2020-05-21
🐛 Remove required subnets webhook validation

### LINKED ISSUE #631: Skip Deletion of vnet in Unmanaged Mode
/kind bug

Specify `AZURE_VNET_NAME` to an already existing vnet and deploy the cluster. When deleting the cluster, CAPZ attempts to delete the vnet.

CAPZ should not attempt to remove vnet resources in unmanaged mode.

```text
E0519 20:36:16.478877       1 controller.go:258] controller-runtime/controller "msg"="Reconciler error" "error"="error deleting AzureCluster default/capz-mbagnara: failed to delete virtual network mbagnara-vnet for cluster capz-mbagnara: failed to delete vnet mbagnara-vnet in resource group mbagnara: network.VirtualNetworksClient#Delete: Failure sending request: StatusCode=400 -- Original Error: Code=\"InUseSubnetCannotBeDeleted\" Message=\"Subnet default is in use by /subscriptions/3b511ccd-979b-40b6-b8fe-08a6d1a75d78/resourceGroups/mbagnara/providers/Microsoft.Network/networkInterfaces/server2019208/ipConfigurations/ipconfig1 and cannot be deleted. In order to delete the subnet, delete all the resources within the subnet. See aka.ms/deletesubnet.\" Details=[]"  "controller"="azurecluster" "request"={"Namespace":"default","Name":"capz-mbagnara"}

```

**Environment:**

- cluster-api-provider-azure version: v0.4.4

CecileRobertMichon:
**What this PR does / why we need it**: The AzureCluster validation currently checks that both subnet roles. However, that is not accurate since it is valid to pass in a vnet without any subnets specified. Specifying the subnets only allows customizing them (eg. changing the subnets name or CIDR). When specifying a pre-existing vnet, it is required to create the subnets in the vnet but not to specify those in the spec. There is already validation that exists to make sure the subnets exist in the controller but that checks the actual state of the vnet, not its spec, so it makes sense for that validation to not be in the webhook.

The reason this fixes #631 is not super obvious but it is this: because we allow specifying a vnet with no subnets as long as no resource group is specified, it is possible for someone to specify the name of vnet that already exists and that is accepted by the create webhook. Then, when we try to reconcile the vnet, we `Get()` the vnet by name, see that it already exists, and skip validation. However, when we try to update the spec to match the actual state of the vnet, this fails because the update webhook sees that a resource group is specified but the required subnets are missing. This causes the spec not to be updated and therefore, even though the vnet exists, its ID is still empty. When user deletes the cluster, the vnet is detected as "managed" because it doesn't have a provider ID which causes the controller to attempt to delete the vnet it doesn't own.

Fixes #631 

**Release note**:

```release-note
fix: remove required subnets webhook validation
```

### DIFF
--- a/api/v1alpha3/azurecluster_validation.go
+++ b/api/v1alpha3/azurecluster_validation.go
@@ -56,16 +56,13 @@ func (c *AzureCluster) validateClusterSpec() field.ErrorList {
 // validateNetworkSpec validates a NetworkSpec
 func validateNetworkSpec(networkSpec NetworkSpec, fldPath *field.Path) field.ErrorList {
 	var allErrs field.ErrorList
-	// If the user specifies a resourceGroup for vnet, it means
-	// that she intends to use a pre-existing vnet. In this case,
-	// we need to verify the information she provides
 	if networkSpec.Vnet.ResourceGroup != "" {
 		if err := validateResourceGroup(networkSpec.Vnet.ResourceGroup,
 			fldPath.Child("vnet").Child("resourceGroup")); err != nil {
 			allErrs = append(allErrs, err)
 		}
-		allErrs = append(allErrs, validateSubnets(networkSpec.Subnets, fldPath.Child("subnets"))...)
 	}
+	allErrs = append(allErrs, validateSubnets(networkSpec.Subnets, fldPath.Child("subnets"))...)
 	if len(allErrs) == 0 {
 		return nil
 	}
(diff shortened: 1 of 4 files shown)


CecileRobertMichon:
/hold cancel

I'm working on a follow up to move the network defaulting to the azure cluster webhook but that can go in separately to hold this bug fix back

CecileRobertMichon:
/hold

I'm actually going to attempt fixing this with better defaulting instead of removing the validation. The bug only applies to an edge case and there is an available workaround so I don't think we need to merge this ASAP.

PR #638, opened 2020-05-21
✨ switch to clusterctl-based e2e framework

### LINKED ISSUE #607: Leverage CAPI clusterctl test framework
/kind feature

Use the new CAPI e2e tests https://github.com/kubernetes-sigs/cluster-api/blob/master/test/e2e/quick_start.go for testing Azure infrastructure.

See https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1716

**Environment:**

chewong:

This PR leverages the CAPI clusterctl test framework.

Fixes #607

**Release note**:

```release-note
✨ switch to clusterctl-based e2e framework
```


CecileRobertMichon:
@chewong will try to get to it today but if I don't will review next week.

Makefile
```diff
@@ -103,8 +109,11 @@ test-integration: ## Run integration tests
 .PHONY: test-e2e
 test-e2e: ## Run e2e tests
 	PULL_POLICY=IfNotPresent $(MAKE) docker-build
-	MANAGER_IMAGE=$(CONTROLLER_IMG)-$(ARCH):$(TAG) \
-	go test ./test/e2e -v -tags=e2e -ginkgo.v -ginkgo.trace -count=1 -timeout=90m
+	go test ./test/e2eclusterctl -v -tags=e2eclusterctl -ginkgo.v -ginkgo.trace -count=1 -timeout=90m \
```
nader-ziada, line 112: need to set MANAGER_IMAGE so the kind cluster loads the correct image

docs/development.md
```diff
@@ -343,7 +343,7 @@ To run E2E locally, set `AZURE_CLIENT_ID`, `AZURE_CLIENT_SECRET`, `AZURE_SUBSCRI
 ./scripts/ci-e2e.sh
 ```
 
-You can optionally set `AZURE_SSH_PUBLIC_KEY_FILE` to use your own SSH key.
+You can optionally set `SSH_KEY_FILE` to use your own SSH key.
```
nader-ziada, line 346: is it possible to keep the same name? it would be easier for current users to not have to set a new env var

test/e2eclusterctl/create_workload_cluster_test.go
```diff
@@ -0,0 +1,37 @@
...
+package e2eclusterctl
+
+import (
+	"context"
+
+	. "github.com/onsi/ginkgo"
+)
+
+var _ = Describe("Workoad cluster creation", func() {
+	CreateWorkloadClusterSpec(context.TODO(), func() CreateWorkloadClusterSpecInput {
```
nader-ziada, line 28: what is this test testing?

Makefile
chewong: With the new e2e framework, the MANGER_IMAGE is now defined as part of the E2E config - https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/638/files#diff-61ca65273bbe1a18ee0d5f135d8ab6e7R3-R5

test/e2eclusterctl/create_workload_cluster_test.go
chewong: This test is testing the creation and deletion of workload clusters through clusterctl, similar to what we are currently testing in test/e2e.

Makefile
nader-ziada: the error I observed was that the docker build image is different than the image its trying to load, so the load fails.

chewong: `./script/ci-e2e.sh` explicitly set the `REGISTRY` environment variable to `e2e`, and I hardcoded image URL in the E2E config to `e2e/cluster-api-azure-controller-amd64:dev`. I could set the hardcoded image URL to `${MANAGER_IMAGE}` and use envsubst to dynamically substitute the value.

test/e2eclusterctl/data/infrastructure-capz/metadata.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: clusterctl.cluster.x-k8s.io/v1alpha3
```
fabriziopandini, line 1: IMO this file can be omitted because it provides the same settings already existing in embedded metadata at  https://github.com/kubernetes-sigs/cluster-api/blob/8b29a9adeeba4677da4221b887cce61155f925a3/cmd/clusterctl/client/repository/metadata_client.go#L189

test/e2eclusterctl/create_workload_cluster_test.go
fabriziopandini: FYI, in test/e2e we are encapsulating test specs in functions in order to allow re-use of the test spec.
If you don't have such a requirement for your specs (re-usability), you can avoid this function and implement your test directly in the Describe body

test/e2eclusterctl/e2e_suite_test.go
```diff
@@ -0,0 +1,209 @@
...
+
+	By("Tearing down the management cluster")
+	if !skipCleanup {
+		tearDown(bootstrapClusterProvider, bootstrapClusterProxy)
+	}
+})
+
+func initScheme() *runtime.Scheme {
+	sc := runtime.NewScheme()
+	framework.TryAddDefaultSchemes(sc)
```
fabriziopandini, line 153: Might me you want to add Azure types to the schema. This can be useful if you want to manipulate Azure objects via the controller runtime client during tests

chewong:
@fabriziopandini thanks for reviewing. The reason why I added envsubst to `azure-dev.yaml` is that it's not used by clusterctl, but rather used by the E2E test suite itself. Currently, developers could run the E2E tests by `make test-e2e` or `./scripts/ci-e2e.sh`.

- when using `make test-e2e`, `${MANAGER_IMAGE}` depends on `${REGISTRY}`, which might be set differently depending on what environment it is executed under.

- when using `./scripts/ci-e2e.sh`, `${REGISTRY}` is set to `e2e`, so `${MANAGER_IMAGE}` is always equal to `e2e/cluster-api-azure-controller-amd64:dev`.

So instead of hardcoding the image to be loaded in `azure-dev.yaml` as `e2e/cluster-api-azure-controller-amd64:dev`, I set it to ${MANAGER_IMAGE} so both methods could work.

test/e2eclusterctl/config/azure-dev.yaml
```diff
@@ -0,0 +1,60 @@
...
+        type: url
+        replacements:
+          - old: "imagePullPolicy: Always"
+            new: "imagePullPolicy: IfNotPresent"
+          - old: "--enable-leader-election"
+            new: "--enable-leader-election=false"
+  - name: azure
+    type: InfrastructureProvider
+    versions:
+      - name: v0.4.4
```
CecileRobertMichon, line 44: How will this be maintained in the future? @fabriziopandini do we need to update the release every time after cutting a new release?

.gitignore
```diff
@@ -55,4 +55,5 @@ tilt-settings.json
 # e2e output
 test/e2e/junit.e2e_suite.*.xml
 test/e2e/logs/*
+test/e2eclusterctl/config/azure-dev-envsubst.yaml
```
CecileRobertMichon, line 58: Are you keeping both `e2eclusterctl` and `e2e` side by side to be able to switch back? should we plan to replace e2e by e2eclusterctl at some point?

chewong: I think we could directly switch to the clusterctl framework by removing `test/e2e` package and renaming `test/e2eclusterctl` to `test/e2e` since all the test cases in `test/e2e` have been converted.

CecileRobertMichon: that sounds good to me, will be easier for maintenance and less confusing. If things go wrong we can always revert. Let's make it a separate commit to delete the old framework and rename the new one so it's easier to revert without deleting the new framework.

test/e2eclusterctl/config/azure-dev.yaml
chewong: It looks like we only have to maintain the major and minor version of this field, and they have to match whatever is defined in https://github.com/kubernetes-sigs/cluster-api/blob/8b29a9adeeba4677da4221b887cce61155f925a3/cmd/clusterctl/client/repository/metadata_client.go#L189. I can set it to `v0.4.0` (following the convention of [capi](https://github.com/kubernetes-sigs/cluster-api/blob/master/test/e2e/config/docker-dev.yaml#L79 ) and [capa](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1716/files#diff-210cac49a90b04e3af86b95cc1169327R76)) so we don't have to change it unless we are cutting a new minor release.

docs/development.md
```diff
@@ -343,7 +343,17 @@ To run E2E locally, set `AZURE_CLIENT_ID`, `AZURE_CLIENT_SECRET`, `AZURE_SUBSCRI
...
+| `AZURE_CONTROL_PLANE_MACHINE_TYPE` | The machine type of the control plane.                                                                                    | `Standard_D2s_v3`                                                                               |
+| `AZURE_NODE_MACHINE_TYPE`          | The machine type of the node machine.                                                                                     | `Standard_D2s_v3`                                                                               |
+| `E2E_CONF_FILE`                    | The path of the [E2E configuration file](https://cluster-api.sigs.k8s.io/developer/e2e.html#defining-an-e2e-config-file).      | `${GOPATH}/src/sigs.k8s.io/cluster-api-provider-azure/test/e2e/config/azure-dev.yaml`            |
+| `E2E_ARTIFACTS`                    | The path of the artifacts folder.                                                                                         | `${GOPATH}/src/sigs.k8s.io/cluster-api-provider-azure/_artifacts`                               |
```
CecileRobertMichon, line 354: Isn't `ARTIFACTS` a predefined prow variable? if so we shouldn't use a different name for it so that CI jobs can pick up the prow value

docs/development.md
```diff
@@ -343,7 +343,17 @@ To run E2E locally, set `AZURE_CLIENT_ID`, `AZURE_CLIENT_SECRET`, `AZURE_SUBSCRI
...
+| `E2E_SKIP_RESOURCE_CLEANUP`        | Set to `true` if you do not want the bootstrap and workload clusters to be cleaned up after running E2E tests.            | `false`                                                                                         |
+| `E2E_USE_EXISTING_CLUSTER`         | Set to `true` if you are using your own bootstrap cluster.                                                                | `false`                                                                                         |
```
CecileRobertMichon, line 356: We should make naming between E2E and Conformance variables consistent if possible, they should both be `USE_EXISTING_CLUSTER` or `SKIP_CREATE`. Here this is about the workload cluster, not the CAPZ cluster so maybe we should also note that in the name, "USE_EXISTING_CLUSTER" sounds like this will run E2E on an existing CAPZ cluster and not provision any infrastructure (as a side note that might be something nice to be able to do in the future).

docs/development.md
```diff
@@ -343,7 +343,17 @@ To run E2E locally, set `AZURE_CLIENT_ID`, `AZURE_CLIENT_SECRET`, `AZURE_SUBSCRI
...
+|------------------------------------|---------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------|
+| `AZURE_SSH_PUBLIC_KEY_FILE`        | Use your own SSH key.                                                                                                     | `${GOPATH}/src/sigs.k8s.io/cluster-api-provider-azure/.sshkey`                                  |
+| `AZURE_CONTROL_PLANE_MACHINE_TYPE` | The machine type of the control plane.                                                                                    | `Standard_D2s_v3`                                                                               |
+| `AZURE_NODE_MACHINE_TYPE`          | The machine type of the node machine.                                                                                     | `Standard_D2s_v3`                                                                               |
```
CecileRobertMichon, line 352: isn't machine type part of the general workload cluster customization (see #customizing-the-cluster-deployment above) and not really a test config option? Can the other customization variables like `AZURE_LOCATION` and `WORKER_MACHINE_COUNT` also be overriden?

scripts/ci-e2e.sh
```diff
@@ -33,8 +33,42 @@ source "${REPO_ROOT}/hack/ensure-kind.sh"
...
 
-export REGISTRY=e2e
+# Verify the required Environment Variables are present.
+: "${AZURE_SUBSCRIPTION_ID:?Environment variable empty or not defined.}"
+: "${AZURE_TENANT_ID:?Environment variable empty or not defined.}"
+: "${AZURE_CLIENT_ID:?Environment variable empty or not defined.}"
+: "${AZURE_CLIENT_SECRET:?Environment variable empty or not defined.}"
+
+get_random_region() {
+    local REGIONS=("eastus" "eastus2" "southcentralus" "westus2" "westeurope")
```
CecileRobertMichon, line 46: Are these regions the prow sub will have quota for? If not will we override it to a specific region for the PR job?

scripts/ci-e2e.sh
```diff
@@ -33,8 +33,42 @@ source "${REPO_ROOT}/hack/ensure-kind.sh"
...
+    echo "${REGIONS[${RANDOM} % ${#REGIONS[@]}]}"
+}
+
+export REGISTRY="e2e"
+export AZURE_SUBSCRIPTION_ID_B64="$(echo -n "$AZURE_SUBSCRIPTION_ID" | base64 | tr -d '\n')"
+export AZURE_TENANT_ID_B64="$(echo -n "$AZURE_TENANT_ID" | base64 | tr -d '\n')"
+export AZURE_CLIENT_ID_B64="$(echo -n "$AZURE_CLIENT_ID" | base64 | tr -d '\n')"
+export AZURE_CLIENT_SECRET_B64="$(echo -n "$AZURE_CLIENT_SECRET" | base64 | tr -d '\n')"
+export AZURE_LOCATION="${E2E_REGION:-$(get_random_region)}"
```
CecileRobertMichon, line 55: should we use `AZURE_LOCATION` to be consistent with the already existing env vars?

CecileRobertMichon: also this is missing from the env config options table in the doc above

test/e2e/create_workload_cluster_test.go
```diff
@@ -0,0 +1,110 @@
...
+
+	Context("Create single controlplane cluster", func() {
+		It("Should create a single node cluster", func() {
+			cluster, _, _ = clusterctl.ApplyClusterTemplateAndWait(ctx, clusterctl.ApplyClusterTemplateAndWaitInput{
+				ClusterProxy: bootstrapClusterProxy,
+				ConfigCluster: clusterctl.ConfigClusterInput{
+					ClusterctlConfigPath:     clusterctlConfigPath,
+					KubeconfigPath:           bootstrapClusterProxy.GetKubeconfigPath(),
+					InfrastructureProvider:   clusterctl.DefaultInfrastructureProvider,
+					Flavor:                   clusterctl.DefaultFlavor,
```
CecileRobertMichon, line 73: Does this fetch the flavor from the `templates/` folder or from the new template you added in e2e/data/infrastructure-azure/cluster-template.yaml? If it's the latter, why do we need to store additional templates just for E2E? this might lead to overhead for maintenance: if something is changed in the default template, we now need to remember to also change the e2e template. It would be nice if this could either 1) point to the existing templates or 2) auto-generate the e2e template in e2e/data. This is probably more of a framework question though and not really specific to this PR. 
cc @fabriziopandini

CecileRobertMichon:
As discussed offline, the apidiff job is none blocking and is complaining because some exported variables and funcs got removed in the `e2e` package, which is fine since no one has dependencies on it.

docs/development.md
chewong: Yea it's a predefined variable in Prow. Let me modify this.

scripts/ci-e2e.sh
chewong: We should have enough quota since each region above has at least 200 cores available (note that this is a clone of https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/test/e2e/resource_generators.go#L462-L474)

scripts/ci-e2e.sh
chewong: I noticed that https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/test/e2e/resource_generators.go#L462-L474 is fetching `E2E_REGION` before choosing a random region so I basically converted it into bash script. I don't think `E2E_REGION` is used by anything in the codebase so let's remove this.

docs/development.md
chewong: For CAPI environment variables like `WORKER_MACHINE_COUNT`, they can't be overridden and they are set programmatically like https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/638/files#diff-3d0a31130788192989e785058215450fR78. For CAPZ environment variables like `AZURE_LOCATION`, they are set in `./scripts/ci-e2e.sh`, which can be overridden. I will make a note in the docs. Also, I removed the use of  `${AZURE_RESOURCE_GROUP}` and `${AZURE_VNET_NAME}`, and directly use `${CLUSTER_NAME}` as the resource group name and  `${CLUSTER_NAME}-vnet` as the vnet name in `e2e/data/infrastructure-azure/cluster-template.yaml`. Please let me what you think of that.

test/e2e/create_workload_cluster_test.go
chewong: I agree. I will update the PR and re-use our current default flavor at `./template/cluster-template.yaml`. In terms of using a specific flavor, you would have to define the favored YAML file in the [E2E configuration](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/28ede51fdaf6a95e878d6dff244660893232d7b4/test/e2e/config/azure-dev.yaml#L46-L47). For example, if you want to use machine pool cluster template, we would define it in the E2E configuration like:

```yaml
    files:
      - sourcePath: ../../../templates/cluster-template-machinepool.yaml
```

And in the test suite, we would have to specify the flavor by:
```go
...
Flavor: "machinepool",
...
```

fabriziopandini: +1 to what @chewong said. The template location can be changed

Issue #641, opened 2020-05-21
Templates for clusters are only partially namespaced


jsturtevant:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]
When running `make create-workload-cluster` while `kubeconfig` is pointing at a namespace other than `default` for most 'flavor' and parts of the templates will end up in default namespace and other parts will end up in current namespace.  This causes only parts of the cluster to be provisioned as CAPI components can't find the `AzureMachineTemplates` or similiar CRD's.

**What did you expect to happen:**
All components to be deployed to same namespace regardless of what namespace I am in now.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**
Kustomize 3.5.4

- cluster-api-provider-azure version: master branch
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

CecileRobertMichon:
/close

fixed by #642

PR #642, opened 2020-05-21
Added namespace override to all kustomize files

jsturtevant:
Fix for #641 when deployment templates using kustomize and on non default namespace

Fixes #
#641 

Could also set this value in ENV and override like others but this felt more native to kustomize.  I ran `make generate-flavors` to fix the generated files.

**Release note**:

```release-note
Added namespace override to kustomize files
```


nader-ziada:
@jsturtevant can you please add a release note so we can merge this in. thanks

CecileRobertMichon:
/approve

Please add a release note @jsturtevant

jsturtevant:
added release note

Issue #427, opened 2020-03-06
Refactor: use defaulting/validating webhooks for more properties


CecileRobertMichon:
Right now we do a lot of defaulting and validation in azuremachine_reconciler.go and azurecluster_reconciler.go. After #334 merges, we should apply the same default/validation method to other cluster and machine spec properties such as: subnets, vnets, NSGs, etc.

richardcase:
@CecileRobertMichon  -  I can look at this after #334 if you want?

CecileRobertMichon:
@richardcase that'd be great if you can! 

I'll assign you for now so no one else takes it

/assign richardcase

CecileRobertMichon:
@richardcase are you still interested in working on this?

richardcase:
@CecileRobertMichon  - i am and have started. But want to get the failure domains finished first. Shall i unassign for the time being?

CecileRobertMichon:
Sounds good, thanks! No need to unassign yourself since you have already started and intend on working on it later.

Issue #618, opened 2020-05-14
RFE: [v1alpha4] redesign user-facing API


CecileRobertMichon:
> ⚠️ Cluster API Azure maintainers can ask to turn an issue-proposal into a [CAEP](https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/proposals/YYYYMMDD-template.md) when necessary. This is to be expected for large changes that impact multiple components, breaking changes, or new large features.

Currently, the AzureCluster and AzureMachine specs are mostly flat. While this works as there are few configuration knobs, it is probably not the best long term as we add more features and configuration options. This is marked as [v1alpha4] because it would be a breaking change but I want to start thinking about this ahead of time so that we have a proposal ready when the time to switch to v1alpha4 comes. Ideally, we should come up with a logical way to break down the spec in AzureCluster, AzureMachine, and AzureMachinePool. For example, Storage, Compute, etc.

Open for discussion. Will bring this up at the next CAPZ office hours.

**Please add any API breaking changes you'd like to see happen in v1alpha4 as comments on this issue**

/kind proposal

CecileRobertMichon:
type Subnets should be `type Subnets [SubnetRole]*SubnetSpec` instead of `type Subnets []*SubnetSpec` and we should remove the `Role` field in `SubnetSpec`.

https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/api/v1alpha3/types.go#L78

This assumes there should be one and only one subnet per "role" (ie. ControlPlane, Node), which is the case today but we should determine if there would be a case for this not being true in the future.

CecileRobertMichon:
> This assumes there should be one and only one subnet per "role" (ie. ControlPlane, Node), which is the case today but we should determine if there would be a case for this not being true in the future.

I don't think that's always going to be true so it might actually be better to have a map where the key is `Name` :

`type Subnets [string]*SubnetSpec`

because Subnet name is and should be unique within a vnet, https://docs.microsoft.com/en-us/azure/virtual-network/virtual-network-manage-subnet#add-a-subnet, and that's an Azure infrastructure constraint, unlike role which is subject to change.

CecileRobertMichon:
OSDisk. DiskSizeGB should be optional, see https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/658#discussion_r433405993

CecileRobertMichon:
[...]

PR #643, opened 2020-05-22
🐛 Move subnets and vnet defaulting to AzureCluster mutating webhook

### LINKED ISSUE #631: Skip Deletion of vnet in Unmanaged Mode

CecileRobertMichon:
**What this PR does / why we need it**: This is a bit more involved but IMO better alternative to #637 for fixing #631.

Because we allow specifying a vnet with no subnets as long as no resource group is specified, it is possible for someone to specify the name of vnet that already exists and that is accepted by the create webhook. Then, when we try to reconcile the vnet, we Get() the vnet by name, see that it already exists, and skip validation. However, when we try to update the spec to match the actual state of the vnet, this fails because the update webhook sees that a resource group is specified but the required subnets are missing. This causes the spec not to be updated and therefore, even though the vnet exists, its ID is still empty. When user deletes the cluster, the vnet is detected as "managed" because it doesn't have a provider ID which causes the controller to attempt to delete the vnet it doesn't own.

Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged):
Fixes #631 
Partially addresses #427 

TODO:

- [x] unit tests
- [x] conversion

**Release note**:

```release-note
fix: Move subnets and vnet defaulting to AzureCluster mutating webhook
```


config/webhook/webhookcainjection_patch.yaml
```diff
@@ -5,9 +5,9 @@
 apiVersion: admissionregistration.k8s.io/v1beta1
 kind: MutatingWebhookConfiguration
 metadata:
- name: mutating-webhook-configuration
- annotations:
-   cert-manager.k8s.io/inject-ca-from: $(CERTIFICATE_NAMESPACE)/$(CERTIFICATE_NAME)
+  name: mutating-webhook-configuration
+  annotations:
+    cert-manager.io/inject-ca-from: $(CERTIFICATE_NAMESPACE)/$(CERTIFICATE_NAME)
```
CecileRobertMichon, line 10: 🐛

config/webhook/manifests.yaml
```diff
@@ -51,7 +71,6 @@ webhooks:
     operations:
     - CREATE
     - UPDATE
-    - DELETE
```
CecileRobertMichon, line 54: I don't think the validation webhook should validate deletes but please correct me if I'm wrong

api/v1alpha3/types.go
```diff
@@ -327,18 +333,46 @@ type SubnetSpec struct {
...
+	// +optional
 	InternalLBIPAddress string `json:"internalLBIPAddress,omitempty"`
 
 	// SecurityGroup defines the NSG (network security group) that should be attached to this subnet.
+	// +optional
 	SecurityGroup SecurityGroup `json:"securityGroup,omitempty"`
+
+	// RouteTable defines the route table that should be attached to this subnet.
+	// +optional
+	RouteTable RouteTable `json:"routeTable,omitempty"`
```
CecileRobertMichon, line 357: Added `RouteTable` to make it easier to track and default the associated RouteTable in the spec. Also there's no reason why it shouldn't be configurable like security group is. For now it's just `Name` and `ID` but we can more fields later as needed. This is optional therefore not a breaking change. Will add conversion (noted as TODO in PR description).

api/v1alpha2/azurecluster_conversion.go
```diff
@@ -57,6 +57,17 @@ func (src *AzureCluster) ConvertTo(dstRaw conversion.Hub) error { // nolint
 	}
 
 	dst.Status.FailureDomains = restored.Status.FailureDomains
+	//dst.Spec.NetworkSpec.Subnets = restored.Spec.NetworkSpec.Subnets
+
+	// for _, restoredSubnet := range restored.Spec.NetworkSpec.Subnets {
```
CecileRobertMichon, line 62: @alexeldeib @devigned I'm struggling to figure what is the correct way to restore here... It's slightly different from `Files.ContentFrom` because there is no mutual exclusivity with another field like `Content` and `ContentFrom`, it's just a new optional field. The above passes the fuzzer unit test but I think it's incorrect because it might overwrite existing v1alpha2 fields that have changed. The below is my attempt to only restore `RouteTable` based on `Name` (assuming `Name` doesn't change that's arguably better than by index but it's not ideal), and that is failing the test.

api/v1alpha3/types.go
CecileRobertMichon: conversion added

CecileRobertMichon:
output from apidiff job:

```
sigs.k8s.io/cluster-api-provider-azure/cloud
  Incompatible changes:
  - DefaultControlPlaneSubnetCIDR: removed
  - DefaultNodeSubnetCIDR: removed
  - DefaultVnetCIDR: removed
  - GenerateControlPlaneSecurityGroupName: removed
  - GenerateControlPlaneSubnetName: removed
  - GenerateNodeRouteTableName: removed
  - GenerateNodeSecurityGroupName: removed
  - GenerateNodeSubnetName: removed
  - GenerateVnetName: removed
  Compatible changes:
  - (*AzureCluster).Default: added
  - (*NetworkSpec).GetControlPlaneSubnet: added
  - (*NetworkSpec).GetNodeSubnet: added
  - DefaultControlPlaneSubnetCIDR: added
  - DefaultNodeSubnetCIDR: added
  - DefaultVnetCIDR: added
  - RouteTable: added
  - SubnetSpec.RouteTable: added
```

None of these are user facing breaking changes.

api/v1alpha3/types.go
```diff
@@ -327,18 +333,46 @@ type SubnetSpec struct {
...
+}
+
+// GetControlPlaneSubnet returns the cluster control plane subnet.
+func (n *NetworkSpec) GetControlPlaneSubnet() *SubnetSpec {
```
devigned, line 361: Just a comment, but it seems a bit odd and possibly error prone to use a list and expect only one of role to be present selecting the first item with that role.

CecileRobertMichon: yep, totally...

Trying to capture that in https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/618

CecileRobertMichon: I think right now the code makes the false assumption that there is one subnet per role (and will only provision one subnet per role) so I think we need to change it to allow multiple node subnets, and not store that subnets in a slice of pointers...

devigned: 💯

CecileRobertMichon: also filed https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/664 to talk about supporting multiple subnets specifically, thanks for the nudge

PR #649, opened 2020-05-27
⚠️ Support Azure client request to other Azure Clouds

### LINKED ISSUE #469: Support Azure client request to other Azure Clouds (eg. Sovereign clouds)
/kind feature

We use cluster api to create k8s cluster in China. But this project uses cloudapp.azure.com as a default DNS zone, and sends client request to management.azure.com for default.
Is it ok to add an option to support other datacenter?

https://docs.microsoft.com/en-us/azure/active-directory/develop/authentication-national-cloud

**Environment:**

nader-ziada:
- if AZURE_ENVIRONMENT is set, will get the matching settings
- all clients use baseURI from settings

Fixes #469 

**Release note**:

```release-note
Support to connect to all regions where Azure is available, not just the default public Azure.
```


cloud/scope/clients.go
```diff
@@ -40,6 +42,8 @@ func (c *AzureClients) setCredentials(subscriptionID string) error {
 	if err != nil {
 		return err
 	}
+	c.ResourceManagerEndpoint = settings.Environment.ResourceManagerEndpoint
+	c.ResourceManagerVMDNSSuffix = settings.Environment.ResourceManagerVMDNSSuffix
```
nader-ziada, line 46: @devigned I'm not sure about this, in the sdk `GetSettingsFromFile` it doesn't return `ResourceManagerVMDNSSuffix` but its used by the cluster reconciler and it has different values based on the cloud used, is there somewhere else to get it?

nader-ziada:
@devigned @CecileRobertMichon @jseely  how would you all feel about just supporting the known environment names as a start? and not read environment from an endpoint or a file?

templates/cluster-template.yaml
```diff
@@ -66,7 +66,7 @@ spec:
     files:
     - content: |
         {
-          "cloud": "AzurePublicCloud",
+          "cloud": "${AZURE_ENVIRONMENT}",
```
CecileRobertMichon, line 69: this makes `AZURE_ENVIRONMENT` a required variable, it needs to be added to the CAPI book quickstart

nader-ziada: is there a different way of doing this so it's optional?

devigned: Not right now. Checkout https://github.com/kubernetes-sigs/cluster-api/pull/3052

I hesitate to say this, but perhaps, another flavor for the time being?

nader-ziada: but its not in the templates? how would the flavor do that?

devigned: I'm not sure I understand the question.

What I'm suggesting is that `cluster-teamplate.yaml` stays as it is with `"AzurePublicCloud"`, so there is no new variable required when deploying the default template. Secondly, create a new flavor which includes the "${AZURE_ENVIRONMENT}"` var.

Does that make sense?

nader-ziada: got it, ok makes sense, will look into what needs to change

CecileRobertMichon: Going to play devil's advocate here... I really don't like the idea of having separate flavors for different variable values. I really wish we could do defaulting instead. Since that's not possible, maybe we can just document that you have to change the `cloud` value to the Azure cloud you want to use since this is part of azure.json. Also keep in mind this whole thing will be an environment variable after https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/653/files#diff-1bf07101845de6da98f1c948b82c5ce0R296.

nader-ziada: it currently defaults to AzurePublicCloud if its empty, or at least it should

nader-ziada: Are we ok in concept to add a new required variable?

devigned: Should we just wait until #653 is merged? As @CecileRobertMichon pointed out, once that is merged, there is no required variable :).

nader-ziada: that's waiting for CAPI 0.3.7 I guess

devigned: The problem with adding a required var is that it will a breaking change for anyone that will run `clusterctl config cluster --infrastructure=azure`. That will break any scripts using the default flavor.

It would be friendlier to users if we were to leave `"cloud": "AzurePublicCloud"` and provide documentation on how to change the value.

When #653 lands, that documentation is still valid, and we didn't break anyone in the mean time.

nader-ziada: don't we still have to add AzureEnvironment env var to the capz-controller which would have to come from a secret or similar? I think instructions to update the secret value or change the deployment configuration would not be very user friendly

CecileRobertMichon: For now let's keep the new var and try to make it optional later with https://github.com/kubernetes-sigs/cluster-api/issues/3189 @nader-ziada

config/default/credentials.yaml
```diff
@@ -8,4 +8,5 @@ data:
   subscription-id: ${AZURE_SUBSCRIPTION_ID_B64}
   tenant-id: ${AZURE_TENANT_ID_B64}
   client-id: ${AZURE_CLIENT_ID_B64}
-  client-secret: ${AZURE_CLIENT_SECRET_B64}
\ No newline at end of file
+  client-secret: ${AZURE_CLIENT_SECRET_B64}
+  cloud-env: ${AZURE_ENVIRONMENT_B64}
```
nader-ziada, line 12: @CecileRobertMichon this variable here?

nader-ziada:
@CecileRobertMichon @devigned  I think the PR is now ready for another look, 
 - changed the `AZURE_ENVIRONMENT` to be used in a ConfigMap so we don't have to encode it in the secret and this way we don't need another `_B64` variable 
 - updated the docs where I thought made sense 
 - added a PR  (on hold) for [the cluster-api book](https://github.com/kubernetes-sigs/cluster-api/pull/3212)

cloud/scope/clients_test.go
```diff
@@ -0,0 +1,64 @@
...
+package scope
+
+import (
+	"os"
+	"testing"
+
+	. "github.com/onsi/gomega"
+)
+
+func TestGettingEnvironment(t *testing.T) {
```
CecileRobertMichon, line 26: what happens when `AZURE_ENVIRONMENT` is an invalid cloud name?
 eg. If I set `AZURE_ENVIRONMENT` to `AzureUAGovernmentCloud`  I should get an error rather than CAPZ trying to use PublicCloud as default and failing later because my sub isn't valid in public cloud.

nader-ziada: added a test for an invalid value, good catch

cloud/scope/clients_test.go
```diff
@@ -0,0 +1,83 @@
...
+			expectedEndpoint:  "https://management.chinacloudapi.cn/",
+			expectedDNSSuffix: "cloudapp.chinacloudapi.cn",
+			expectedError:     false,
+		}, "AZURE_ENVIRONMENT is AzureGermany": {
+			azureEnv:          "AzureGermanCloud",
+			expectedEndpoint:  "https://management.microsoftazure.de/",
+			expectedDNSSuffix: "cloudapp.microsoftazure.de",
+			expectedError:     false,
+		}, "AZURE_ENVIRONMENT has an invalid value": {
+			azureEnv:             "AzureInSpace",
```
CecileRobertMichon, line 62: 😆

nader-ziada:
any concerns/comments still outstanding?

docs/getting-started.md
```diff
@@ -74,6 +74,18 @@ An Azure Service Principal is needed for populating the controller manifests. Th
   export AZURE_CLIENT_SECRET=<Password>
   export AZURE_LOCATION="eastus"
   ```
+
+  5. Set the name of the AzureCloud to be used, the default value that would be used by most users is "AzurePublicCloud", other values are:
+
+  - ChinaCloud: "AzureChinaCloud"
```
devigned, line 80: nit: looks like an extra leading space on the list item

nader-ziada: I had to do that for some reason to make the rendered page look aligned

devigned: Indeed, it looks right when rendered.

PR #652, opened 2020-05-28
✨ enable BYO IP for control plane

alexeldeib:
Enables users to bring their own IP for the control plane. This allows using a CAPI control plane with self-managed nodes by setting advertise address on KCP at create-time.

Defaults to previous behavior if unset. Updating this field is invalid.

Demo for Azure edge scenarios

**Release note**:
```release-note
Enable users to pre-provision the control plane IP so it may be used e.g. in advertise address.
```


devigned:
I think BYOR (bring your own resource) is common across almost all resources. What do folks think about allowing the ProviderID to be set by a user, and the controller would use the ProviderID being set on an unowned resources as the indication the resource should be a BYO?

MirzaSikander:
Can we get this pull request to completion? Need this functionality to move forward with our deployment.

alexeldeib:
I think at the very least, this PR needs to add a concept of ownership so that CAPZ doesn't delete IPs brought by the user and does not fail to cleanup IPs it creates.

We've discussed this as a general issue a bit, @devigned @CecileRobertMichon any thoughts on how you think it should look for this particular use case?

@MirzaSikander feel free to take over and drive this PR if you need the functionality sooner as mentioned offline

MirzaSikander:
For us, deleting the IP address is actually preferred so for us its not an issue. We can always extend it later to optionally delete it.

devigned:
Could we tag managed pubIPs with `"sigs.k8s.io_cluster-api-provider-azure_cluster_test-cluster": "owned"` (or something similar), and just check for that upon delete? That seems like this would deal with the ownership problem for this instance.

In the longer term, I don't know that we want to use only resource name as the way we look up resources. Using resource ID would be more precise. Moreover, I'm worried that if we pull this in quickly, then decide on a generic way of accepting "bring your own" resources, then this could be a painful migration in the next api version.

MirzaSikander:
Just wanted to provide our context and use case for this feature. 

We are team in Azure that is responsible for managing bare-metal infra in locations external to Azure DC aka the edge. We are in the process of layering k8s on top of our bare-metal infra to allow containerized apps to be deployed and managed. Our current topology involves the control plane deployed in Azure managing worker nodes on a private network in the edge. The master has a public IP address through which the workers are able to join the cluster. We wanted to transition over to CAPZ to avoid manually managing these clusters (Only the control plane as of now).

For things to work properly we need to specify the advertise-address on the API-server
```
kubeadmConfigSpec:
    clusterConfiguration:
      apiServer:
        timeoutForControlPlane: 20m
        extraArgs:
          advertise-address: ${AZURE_PUBLIC_IP_ADDRESS}
 ````
Currently there is no way CAPI/KCP to dynamically do this. (If there is maybe we can try that approach)

With this pull request, we can sort of get around that by pre-provisioning a static IP address and populating the cluster configuration.

**Comment by devigned** [comment]
Ok... This has been hanging around for a little while. It need to come to a conclusion.

I don't love having name required on IPAddress. What if we leave the IPAddress type as it was, and just check with ARM if the IPAddress exists if name is set? If it does exist, we use it, else create it and tag it with CAPZ managed tag. On delete, we would not delete the IPAddress if it is tagged with the CAPZ managed tag.

Thoughts?

**Comment by alexeldeib** [comment]
Yep, I think that approach makes sense. This is my next task up, I got a little distracted by a flurry of work on disks again -_- Thankfully all finished with that

File: cloud/services/publicips/publicips.go [hunk 1]
```diff
@@ -76,16 +97,46 @@ func (s *Service) Delete(ctx context.Context, spec interface{}) error {
...
+
 	if err != nil {
 		return errors.Wrapf(err, "failed to delete public ip %s in resource group %s", publicIPSpec.Name, s.Scope.ResourceGroup())
 	}
 
 	klog.V(2).Infof("deleted public ip %s", publicIPSpec.Name)
 	return err
 }
+
+func (s *Service) isManaged(ctx context.Context, name string) (bool, error) {
```
CecileRobertMichon, line 134: cc @cpanato

This is exactly the change you need for #708

CecileRobertMichon:
@alexeldeib this will need a pretty big rebase since #716 merged... sorry about that
Let me know if you want help, I'm happy to pair on applying this change to the new code organization.

alexeldeib:
@CecileRobertMichon rebase seemed surprisingly okay? let me know what you think

since I set the name in https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/f9a6515ac55766f862124f27b2804abfaaaad121/controllers/azurecluster_reconciler.go#L295 which is the first function in reconcile it seems to work ok

cloud/services/publicips/publicips.go
```diff
@@ -18,61 +18,124 @@ package publicips
...
 }
 
 // Delete deletes the public IP with the provided scope.
 func (s *Service) Delete(ctx context.Context) error {
-	for _, ip := range s.Scope.PublicIPSpecs() {
-		klog.V(2).Infof("deleting public IP %s", ip.Name)
-		err := s.Client.Delete(ctx, s.Scope.ResourceGroup(), ip.Name)
+	for _, publicIPSpec := range s.Scope.PublicIPSpecs() {
+		klog.V(2).Infof("deleting public IP %s", publicIPSpec.Name)
+		err := s.Client.Delete(ctx, s.Scope.ResourceGroup(), publicIPSpec.Name)
```
CecileRobertMichon, line 89: rebase err? seems like we're deleting the IP twice in this block

alexeldeib: whoops

cloud/services/publicips/publicips.go
```diff
@@ -18,61 +18,124 @@ package publicips
...
+			if ip.IsHTTPStatus(http.StatusNotFound) {
+				return errors.Wrapf(err, "failed to find public ip '%s' for deletion, ignoring and finishing successfully", publicIPSpec.Name)
+			}
+			return errors.Wrapf(err, "failed to fetch public ip '%s' to check if we should delete", publicIPSpec.Name)
+		}
+
+		tags := converters.MapToTags(ip.Tags)
+		isManaged := tags.HasOwned(s.Scope.ClusterName())
+		if isManaged {
+			klog.V(2).Infof("found unmanaged public ip %s, will not delete", publicIPSpec.Name)
```
CecileRobertMichon, line 101: is this a breaking change? Previously created IPs with CAPZ (prior to this version) won't have the ownership tags so they won't be deleted properly if the cluster gets upgraded to this version.... Might be worth calling out in release notes. Not sure if there is a way to make this backwards compatible, unfortunately.

api/v1alpha3/types.go
```diff
@@ -137,9 +142,14 @@ type IngressRules []*IngressRule
...
 	ID        string `json:"id,omitempty"`
-	Name      string `json:"name,omitempty"`
 	IPAddress string `json:"ipAddress,omitempty"`
 	DNSName   string `json:"dnsName,omitempty"`
+	Name      string `json:"name,omitempty"`
+}
+
+// PublicIPSpec defines the inputs to create an Azure public IP address.
+type PublicIPSpec struct {
+	Name string `json:"name,omitempty"`
```
CecileRobertMichon, line 152: do we really need this new type? Seems like it's a subset of the above `PublicIP` type and since all the fields besides Name are optional we should be able to reuse it? Is it because this is only meant to provide the name of an already existing IP?

cloud/services/publicips/publicips.go
alexeldeib: if the tag doesn't exist, HasOwned returns false. So it's only true if the tag was both set and equal to true.

api/v1alpha3/types.go
CecileRobertMichon: also what about resource group? Does this imply same RG as the rest of the cluster? BYO vnet allows a different RG for the vnet

cloud/services/publicips/publicips.go
CecileRobertMichon: right. So previously created IPs will return false to isManaged() (even though they were created by CAPZ and therefore not get deleted no? Am I missing something?

cloud/services/publicips/publicips.go
```diff
@@ -18,61 +18,116 @@ package publicips
...
+		if err != nil {
+			if ip.IsHTTPStatus(http.StatusNotFound) {
+				return errors.Wrapf(err, "failed to find public ip '%s' for deletion, ignoring and finishing successfully", publicIPSpec.Name)
+			}
+			return errors.Wrapf(err, "failed to fetch public ip '%s' to check if we should delete", publicIPSpec.Name)
+		}
+
+		tags := converters.MapToTags(ip.Tags)
+		isManaged := tags.HasOwned(s.Scope.ClusterName())
+		if isManaged {
```
CecileRobertMichon, line 100:
```suggestion
		if !isManaged {
```
?

cloud/services/publicips/publicips.go
```diff
@@ -18,61 +18,116 @@ package publicips
...
-	for _, ip := range s.Scope.PublicIPSpecs() {
-		klog.V(2).Infof("deleting public IP %s", ip.Name)
-		err := s.Client.Delete(ctx, s.Scope.ResourceGroup(), ip.Name)
-		if err != nil && azure.ResourceNotFound(err) {
+	for _, publicIPSpec := range s.Scope.PublicIPSpecs() {
+		klog.V(2).Infof("deleting public IP %s", publicIPSpec.Name)
+
+		ip, err := s.Client.Get(ctx, s.Scope.ResourceGroup(), publicIPSpec.Name)
+		if err != nil {
+			if ip.IsHTTPStatus(http.StatusNotFound) {
```
CecileRobertMichon, line 92: azure.ResourceNotFound(err) ?

api/v1alpha3/types.go
alexeldeib: existing public IP is used only for status and exposes fields like IPAddress which wouldn't actually be settable from spec. I'm fine with adding a validating webhook against populating some of those? or else modifying public IP itself to look more like a k8s object: 

```go
// PublicIP defines an Azure public IP address.
type PublicIP struct {
	Spec
        Status
}

// PublicIPSpec defines the inputs to create an Azure public IP address.
type PublicIPSpec struct {
	DNSName   string `json:"dnsName,omitempty"`
}

// PublicIPStatus defines the outputs from creating an Azure public IP address.
type PublicIPStatus struct {
	ID        string `json:"id,omitempty"`
	IPAddress string `json:"ipAddress,omitempty"`
}
```

that seems a little heavy though. Reusing the type + validating webhook seems like a nice compromise?

alexeldeib: Using a separate type has the benefit of safety on validating the struct -- when we add new fields, we would have to exclude any status-only fields in the future. That seems prone to error. So maybe splitting to spec/status is better? But I don't think DNS name would even be custom right now

cloud/services/publicips/publicips.go
alexeldeib: eep lost after refactoring my tags

cloud/services/publicips/publicips.go
alexeldeib: lol was desperately trying to remember this func and ended up using 404 😄

alexeldeib:
need to fix up the loop after rebase

cloud/services/publicips/publicips.go
alexeldeib: Nope I had it backwards 🙂 I guess I'll add a release note

alexeldeib: see release note?

cloud/services/publicips/publicips.go
cpanato: will review and follow the approach! thanks!

api/v1alpha3/types.go
alexeldeib: lmk what you think here, after fiddling with unit tests this is the main action item I see remaining

cloud/services/publicips/publicips.go
```diff
@@ -18,61 +18,107 @@ package publicips
...
-				Location: to.StringPtr(s.Scope.Location()),
-				PublicIPAddressPropertiesFormat: &network.PublicIPAddressPropertiesFormat{
-					PublicIPAddressVersion:   network.IPv4,
-					PublicIPAllocationMethod: network.Static,
-					DNSSettings: &network.PublicIPAddressDNSSettings{
-						DomainNameLabel: to.StringPtr(strings.ToLower(ip.Name)),
-						Fqdn:            to.StringPtr(ip.DNSName),
-					},
+		ip, err := s.Client.Get(ctx, s.Scope.ResourceGroup(), publicIPSpec.Name)
+		if err != nil && !azure.ResourceNotFound(err) {
```
CecileRobertMichon, line 39:
```
		if err == nil {
			klog.V(2).Infof("found existing public ip %s, returning early", publicIPSpec.Name)
			return nil
		}
                 if !azure.ResourceNotFound(err) {
			return errors.Wrapf(err, "failed to fetch public ip '%s' to check if we should create", publicIPSpec.Name)
		}
```

cloud/services/publicips/publicips.go
```diff
@@ -18,61 +18,107 @@ package publicips
...
+				ClusterName: s.Scope.ClusterName(),
+				Lifecycle:   infrav1.ResourceLifecycleOwned,
+				Name:        to.StringPtr(s.Scope.ResourceGroup()),
+				Role:        to.StringPtr(infrav1.CommonRole),
+				Additional:  s.Scope.AdditionalTags(),
+			})),
+		}
 
-		if err != nil {
+		// https://docs.microsoft.com/en-us/azure/load-balancer/load-balancer-standard-availability-zones#zone-redundant-by-default
```
CecileRobertMichon, line 69: ?

controllers/azurecluster_reconciler.go
```diff
@@ -300,7 +300,10 @@ func (r *azureClusterReconciler) deleteNSG(ctx context.Context) error {
 
 // CreateOrUpdateNetworkAPIServerIP creates or updates public ip name and dns name
 func (r *azureClusterReconciler) createOrUpdateNetworkAPIServerIP() error {
-	if r.scope.Network().APIServerIP.Name == "" {
+	// Take user provided IP address, or generate if no name exists.
+	if r.scope.AzureCluster.Spec.NetworkSpec.PublicIP != nil && r.scope.AzureCluster.Spec.NetworkSpec.PublicIP.Name != "" {
```
CecileRobertMichon, line 304: @alexeldeib one thing I'm thinking: is it necessary to pass in the unmanaged public IP to the list of public IP specs to reconcile? Since it's not mutable, it might make sense to just exclude it from the list if the condition above is true in `PublicIPSpecs()`, WDYT?

controllers/azurecluster_reconciler.go
```diff
@@ -300,7 +300,10 @@ func (r *azureClusterReconciler) deleteNSG(ctx context.Context) error {
 
 // CreateOrUpdateNetworkAPIServerIP creates or updates public ip name and dns name
 func (r *azureClusterReconciler) createOrUpdateNetworkAPIServerIP() error {
```
CecileRobertMichon, line 302: I think I missed this as part of the public IP refactor, it should probably move to PublicIPSpecs() and/or azure cluster controller to after the IP has been created since this should really be a status update (and we treat it as one in the controller)...

api/v1alpha3/types.go
CecileRobertMichon: @devigned do you have any thoughts on the above?

controllers/azurecluster_reconciler.go
alexeldeib: I think this makes sense. 

Another thing I just noticed, though -- I think it's technically possible to have a DNS name that wouldn't match what we expect if we construct it using GenerateFQDN. So we may still want to fetch the FQDN from the real ip?

alexeldeib:
@CecileRobertMichon is that what you mean with scope? mind taking a look after refreshing this?

cloud/scope/cluster.go
```diff
@@ -105,16 +107,32 @@ func (s *ClusterScope) Network() *infrav1.Network {
...
+	if s.AzureCluster.Spec.NetworkSpec.PublicIP != nil && s.AzureCluster.Spec.NetworkSpec.PublicIP.Name != "" {
+		return spec, nil
+	}
+
+	if s.Network().APIServerIP.Name == "" {
+		h := fnv.New32a()
+		if _, err := h.Write([]byte(fmt.Sprintf("%s/%s/%s", s.SubscriptionID(), s.ResourceGroup(), s.ClusterName()))); err != nil {
+			return nil, errors.Wrapf(err, "failed to write hash sum for api server ip")
+		}
+		s.Network().APIServerIP.Name = azure.GeneratePublicIPName(s.ClusterName(), fmt.Sprintf("%x", h.Sum32()))
```
CecileRobertMichon, line 127: It seems odd to be setting the value in here given that it's implied to be a Getter function, especially since this is setting status... We do need this value to be set before reconcile IPs is called. Thoughts about breaking into a separate function calling it something explicit like `SetAPIServerIP()` and calling it from the reconciler like we do with `SetFailureDomains()`? 

I realize that's a partial reversal to what you had before so I apologize if my initial comment was confusing.

cloud/scope/cluster.go
```diff
@@ -105,16 +107,32 @@ func (s *ClusterScope) Network() *infrav1.Network {
...
 			Name: azure.GenerateNodeOutboundIPName(s.ClusterName()),
 		},
-		{
-			Name:    s.Network().APIServerIP.Name,
-			DNSName: s.Network().APIServerIP.DNSName,
-		},
 	}
+
+	// If IP provided by name, expect to exist and don't precreate.
+	if s.AzureCluster.Spec.NetworkSpec.PublicIP != nil && s.AzureCluster.Spec.NetworkSpec.PublicIP.Name != "" {
```
CecileRobertMichon, line 118: Do we allow PublicIP != nil but Name == ""?

cloud/scope/cluster.go
devigned: I would opt toward a setter as it would reduce the depth of references and be specific about what the method requires from its scope interface.

cloud/scope/machine.go
```diff
@@ -87,15 +87,15 @@ type MachineScope struct {
 }
 
 // PublicIPSpec returns the public IP specs.
-func (m *MachineScope) PublicIPSpecs() []azure.PublicIPSpec {
+func (m *MachineScope) PublicIPSpecs() ([]azure.PublicIPSpec, error) {
```
devigned, line 90: Why the error return? Doesn't look like it's used.

cloud/services/publicips/service.go
```diff
@@ -26,7 +26,7 @@ import (
 type PublicIPScope interface {
 	logr.Logger
 	azure.ClusterDescriber
-	PublicIPSpecs() []azure.PublicIPSpec
+	PublicIPSpecs() ([]azure.PublicIPSpec, error)
```
devigned, line 29: I think you might be able to get rid of `error` on this signature.

api/v1alpha3/types.go
devigned: I don't feel strongly about this, but I do prefer the proposal with `PublicIP` being composed of two structures. It's a bit more verbose, but it is clear in the separation.

cloud/services/publicips/service.go
alexeldeib: it's from the hash in the name generation: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/652/files#diff-7c0af1847b5ee759aadb398ac16c4d36R124

are you thinking i'd move that elsewhere?

alexeldeib: I get it 👍

cloud/scope/cluster.go
alexeldeib: not any more 🙂

cloud/scope/cluster.go
alexeldeib: better?

api/v1alpha3/azurecluster_webhook.go
```diff
@@ -48,15 +50,51 @@ func (c *AzureCluster) Default() {
...
 	clusterlog.Info("validate create", "name", c.Name)
-
-	return c.validateCluster()
+	allErrs := c.validateCluster()
+
+	// validation is smart/lazy enough to work with the same object compared to itself
+	// it works well for update to take both objects.
+	if err := validateControlPlaneIP(
+		c.Spec.NetworkSpec.PublicIP,
+		c.Spec.NetworkSpec.PublicIP,
```
CecileRobertMichon, line 59: this seems a bit odd, left a note about it in #777 

I guess it would be hard to pass in a nil object for Create because then it's hard to tell if the user added the field during an update or if it's a new cluster create

cloud/services/publicips/publicips.go
```diff
@@ -23,56 +23,70 @@ import (
...
-	for _, ip := range s.Scope.PublicIPSpecs() {
-		s.Scope.V(2).Info("deleting public IP", "public ip", ip.Name)
-		err := s.Client.Delete(ctx, s.Scope.ResourceGroup(), ip.Name)
-		if err != nil && azure.ResourceNotFound(err) {
-			// already deleted
-			return nil
+	for _, publicIPSpec := range s.Scope.PublicIPSpecs() {
+		s.Scope.V(2).Info("deleting public IP", "public ip", publicIPSpec.Name)
+
+		err := s.Client.Delete(ctx, s.Scope.ResourceGroup(), publicIPSpec.Name)
```
CecileRobertMichon, line 77: did you change this behavior? I don't see the isManaged part anymore

CecileRobertMichon: so looks like this change isn't breaking anymore?

Based on `We expect it to be created by them and will not reconcile it.`

api/v1alpha3/azurecluster_webhook.go
alexeldeib: I could do this?

```go
func (new *PublicIPSpec) validateControlPlaneIP(old *PublicIPSpec, fldPath *field.Path) *field.Error {
	if old == nil && new != nil {
		return field.Invalid(fldPath, new, fmt.Sprintf("setting control plane endpoint after cluster creation is not allowed"))
	}
	if old != nil && new == nil {
		return field.Invalid(fldPath, new, fmt.Sprintf("removing control plane endpoint after cluster creation is not allowed"))
	}
	if old != nil && new != nil && old.Name != new.Name {
		return field.Invalid(fldPath, new, fmt.Sprintf("changing control plane endpoint after cluster creation is not allowed"))
	}
	if new != nil && new.Name == "" {
		return field.Invalid(fldPath, new, fmt.Sprintf("control plane endpoint IP name must be non-empty"))
	}
	return nil
}
```

alexeldeib: this could be similar to reconciliation -- each type exposes its own validation, we just enumerate them

cloud/services/publicips/publicips.go
alexeldeib: Ah, https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/652#discussion_r447939037 is paying dividends 🙂

CecileRobertMichon: Do we still want the tags in that case? It's probably harmless to keep them, but we're not actually using them to not delete user created IPs so it might be misleading. 

Let's remove the breaking change PR description if it's no longer a breaking change.

api/v1alpha3/azurecluster_webhook.go
CecileRobertMichon: up to you, I think either way is fine.

cloud/services/publicips/publicips.go
alexeldeib: updated

alexeldeib:
@CecileRobertMichon if it's fine with you i'll leave the validation as is. I updated the PR title and description since this implementation is no longer breaking. lmk if you have additional comments?

cloud/services/publicips/publicips.go
alexeldeib: removed tag logic.

api/v1alpha3/types.go
```diff
@@ -38,6 +38,11 @@ type Network struct {
 
 // NetworkSpec specifies what the Azure networking resources should look like.
 type NetworkSpec struct {
+	// PublicIP is the public IP to attach to the Azure load balanacer.
+	// If name is not set in the spec, one will be generated.
```
CecileRobertMichon, line 42: Is "If name is not set in the spec, one will be generated." accurate? I thought the name _had_ to be provided if the user wanted to pass in there own IP?

CecileRobertMichon: Let's make it more explicit that this is for providing your own control plane IP that already exists

cloud/services/publicips/publicips.go
```diff
@@ -28,51 +28,56 @@ import (
...
 }
 
 // Delete deletes the public IP with the provided scope.
 func (s *Service) Delete(ctx context.Context) error {
-	for _, ip := range s.Scope.PublicIPSpecs() {
-		s.Scope.V(2).Info("deleting public IP", "public ip", ip.Name)
-		err := s.Client.Delete(ctx, s.Scope.ResourceGroup(), ip.Name)
-		if err != nil && azure.ResourceNotFound(err) {
-			// already deleted
-			return nil
```
CecileRobertMichon, line 69: good catch 😬

cloud/scope/cluster.go
```diff
@@ -231,3 +243,22 @@ func (s *ClusterScope) SetFailureDomain(id string, spec clusterv1.FailureDomainS
...
+	} else if s.Network().APIServerIP.Name == "" {
+		// otherwise, generate
+		h := fnv.New32a()
+		if _, err := h.Write([]byte(fmt.Sprintf("%s/%s/%s", s.SubscriptionID(), s.ResourceGroup(), s.ClusterName()))); err != nil {
+			return errors.Wrapf(err, "failed to write hash sum for api server ip")
+		}
+		s.AzureCluster.Status.Network.APIServerIP.Name = azure.GeneratePublicIPName(s.ClusterName(), fmt.Sprintf("%x", h.Sum32()))
+	}
+
+	s.AzureCluster.Status.Network.APIServerIP.DNSName = s.GenerateFQDN()
```
CecileRobertMichon, line 261: Earlier you made a comment about how FQDN might not match expected FQDN, is that still accurate? If so, can we either document that limitation or fix it by allowing the user to provide their own?

alexeldeib: ah, yep, I thought this resulted in 400 but looks like it does work.

alexeldeib:
@MirzaSikander I know this ended up taking quite a bit longer than anticipated, would appreciate if you have a chance to take a quick look at this for your scenario.

@CecileRobertMichon force pushed but added DNS name and pass it through directly

CecileRobertMichon:
https://github.com/kubernetes-sigs/cluster-api/issues/3715

CecileRobertMichon:
/close

closing in favor of #974

PR #653, opened 2020-05-28
⚒️ eliminate plain text credentials

alexeldeib:
Avoids storing plain text credentials in text now that CAPBK supports secrets. This won't work on a tagged release until CAPI v0.3.7 (next release).

```release-note
Eliminate plain text credentials.
```

/hold

for capi v0.3.7

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -278,7 +278,31 @@ SSH_KEY_FILE=.sshkey
 rm -f "${SSH_KEY_FILE}" 2>/dev/null
 ssh-keygen -t rsa -b 2048 -f "${SSH_KEY_FILE}" -N '' 1>/dev/null
 echo "Machine SSH key generated in ${SSH_KEY_FILE}"
-export AZURE_SSH_PUBLIC_KEY=$(cat "${SSH_KEY_FILE}.pub" | base64 | tr -d '\r\n')
+export AZURE_SSH_PUBLIC_KEY=$(cat "${SSH_KEY_FILE}.pub" | base64 | tr -d '\n')
+
+# To populate secret in azure.json file.
+export AZURE_STANDARD_JSON_B64=$(echo '{
+      "cloud": "${AZURE_ENVIRONMENT}",
+      "tenantId": "${AZURE_TENANT_ID}",
+      "subscriptionId": "${AZURE_SUBSCRIPTION_ID}",
+      "aadClientId": "${AZURE_CLIENT_ID}",
+      "aadClientSecret": "${AZURE_CLIENT_SECRET}",
+      "resourceGroup": "${CLUSTER_NAME}",
+      "securityGroupName": "${CLUSTER_NAME}-node-nsg",
+      "location": "${AZURE_LOCATION}",
+      "vmType": "standard",
+      "vnetName": "${CLUSTER_NAME}-vnet",
+      "vnetResourceGroup": "${CLUSTER_NAME}",
+      "subnetName": "${CLUSTER_NAME}-node-subnet",
+      "routeTableName": "${CLUSTER_NAME}-node-routetable",
+      "loadBalancerSku": "standard",
+      "maximumLoadBalancerRuleCount": 250,
+      "useManagedIdentityExtension": false,
+      "useInstanceMetadata": true
+}' | envsubst | base64 | tr -d '\n')
+
+# VMSS requires different values
+export AZURE_VMSS_JSON_B64=$(echo "$AZURE_STANDARD_JSON_B64" | base64 -d | jq '.vmType = "vmss"' | base64 | tr -d '\n')
 ```
 
 ##### Creating the cluster
--- a/hack/create-dev-cluster.sh
+++ b/hack/create-dev-cluster.sh
@@ -54,7 +54,29 @@ if ! [ -n "$SSH_KEY_FILE" ]; then
     ssh-keygen -t rsa -b 2048 -f "${SSH_KEY_FILE}" -N '' 1>/dev/null
     echo "Machine SSH key generated in ${SSH_KEY_FILE}"
 fi
-export AZURE_SSH_PUBLIC_KEY=$(cat "${SSH_KEY_FILE}.pub" | base64 | tr -d '\r\n')
+export AZURE_SSH_PUBLIC_KEY=$(cat "${SSH_KEY_FILE}.pub" | base64 | tr -d '\n')
+
+export AZURE_STANDARD_JSON_B64=$(echo '{
+      "cloud": "${AZURE_ENVIRONMENT}",
+      "tenantId": "${AZURE_TENANT_ID}",
+      "subscriptionId": "${AZURE_SUBSCRIPTION_ID}",
+      "aadClientId": "${AZURE_CLIENT_ID}",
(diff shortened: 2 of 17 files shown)


hack/create-dev-cluster.sh
```diff
@@ -55,6 +55,28 @@ if ! [ -n "$SSH_KEY_FILE" ]; then
 fi
 export AZURE_SSH_PUBLIC_KEY=$(cat "${SSH_KEY_FILE}.pub" | base64 | tr -d '\r\n')
 
+export AZURE_STANDARD_JSON_B64=$(echo '{
```
CecileRobertMichon, line 58: Can you also update all the docs in docs/topics that have template examples?

alexeldeib: will do

CecileRobertMichon:
@alexeldeib is this ready to be rebased now that capi v0.3.7-alpha merged (#728)?

nader-ziada:
@alexeldeib I think the e2e tests failed because the variable `AZURE_VMSS_JSON_B64 ` is empty, its needs to be set in [scripts/ci-e2e.sh](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/scripts/ci-e2e.sh)

docs/development.md
```diff
@@ -279,6 +279,30 @@ rm -f "${SSH_KEY_FILE}" 2>/dev/null
 ssh-keygen -t rsa -b 2048 -f "${SSH_KEY_FILE}" -N '' 1>/dev/null
 echo "Machine SSH key generated in ${SSH_KEY_FILE}"
 export AZURE_SSH_PUBLIC_KEY=$(cat "${SSH_KEY_FILE}.pub" | base64 | tr -d '\r\n')
+
+# To populate secret in azure.json file.
+export AZURE_STANDARD_JSON_B64=$(echo '{
+      "cloud": "AzurePublicCloud",
```
CecileRobertMichon, line 285:
```suggestion
      "cloud": "${AZURE_ENVIRONMENT}",
```

hack/create-dev-cluster.sh
```diff
@@ -56,6 +56,28 @@ if ! [ -n "$SSH_KEY_FILE" ]; then
...
+      "cloud": "AzurePublicCloud",
```
CecileRobertMichon, line 60:
```suggestion
      "cloud": "${AZURE_ENVIRONMENT}",
```

nader-ziada:
sorry if this is obvious, but I'm trying to understand how this change would work. 
Does it mean that if I want to change cluster name or create a second cluster I have to regenerate and update the secret? how would that work with passing the cluster name to clusterctl?

alexeldeib:
The secret is in the template, so clusterctl will generate it for you from env vars. I can update the secret name to be something like "${CLUSTER_NAME}-azure-json" to make that more explicit.

Everything else works exactly like it does today. If you want to change cluster name, the experience would be the same for the secret as for the raw file content. 

Is there a different way you'd prefer this to be implemented?

nader-ziada:
thanks for the explanation, I was just trying to understand how this changes the cli workflow.  does clusterctl generate the `AZURE_STANDARD_JSON_B64` value or does the user have to do that and set it as an env variable

alexeldeib:
Clusterctl doesn't "know" anything about this JSON value, that's true. So users would probably have to provide the full string right now as laid out in the dev docs. I don't really know how to make that easier without teaching clusterctl about Azure specific things -- open to suggestions

nader-ziada:
I don't think there is a way currently, as far as I know, for clusterctl to generate this for you, all it does is substitute the value of a variable from the environment, so the user would have to regenerate this two new env variables before creating a new cluster. I just think it makes the ux more complicated because even though you don't need AZURE_SUBSCRIPTION_ID, AZURE_TENANT_ID, AZURE_CLIENT_ID, AZURE_CLIENT_SECRET in the template, the user still has to set them as env variables to be able to generate this new secret, and its adds two steps before creating a cluster 🤷

nader-ziada:
I'm just thinking out loud - as I was trying to actually use this feature as part of going over the PR

alexeldeib:
Aren't all four of those values already required by clusterctl to set the azure.json today? If not, where do those come from today? They are all in the existing json. 

I agree with you about the UX -- I'd very strongly prefer for this file to be generated somehow without additional user input. That might require some work on CLI (or controller?) tooling in general to make it possible.

alexeldeib:
You'd need to manually run this as a user: 

```bash
export AZURE_SSH_PUBLIC_KEY=$(cat "${SSH_KEY_FILE}.pub" | base64 | tr -d '\n')

export AZURE_STANDARD_JSON_B64=$(echo '{
This conversation was marked as resolved by alexeldeib
      "cloud": "${AZURE_ENVIRONMENT}",
      "tenantId": "${AZURE_TENANT_ID}",
      "subscriptionId": "${AZURE_SUBSCRIPTION_ID}",
      "aadClientId": "${AZURE_CLIENT_ID}",
      "aadClientSecret": "${AZURE_CLIENT_SECRET}",
      "resourceGroup": "${CLUSTER_NAME}",
      "securityGroupName": "${CLUSTER_NAME}-node-nsg",
      "location": "${AZURE_LOCATION}",
      "vmType": "standard",
      "vnetName": "${CLUSTER_NAME}-vnet",
      "vnetResourceGroup": "${CLUSTER_NAME}",
      "subnetName": "${CLUSTER_NAME}-node-subnet",
      "routeTableName": "${CLUSTER_NAME}-node-routetable",
      "loadBalancerSku": "standard",
      "maximumLoadBalancerRuleCount": 250,
      "useManagedIdentityExtension": false,
      "useInstanceMetadata": true
}' | envsubst |  base64 | tr -d '\n')

export AZURE_VMSS_JSON_B64=$(echo "$AZURE_STANDARD_JSON_B64" | base64 -d | jq '.vmType = "vmss"' | base64 | tr -d '\n')
```

but all of the variables used in that template should already be used by capz/clusterctl -- it's a copy paste of the existing file out of the yaml. 

maybe can improve docs for this?

CecileRobertMichon:
> maybe can improve docs for this?

This requires an update to the capi quickstart for sure.

CecileRobertMichon:
I can see @nader-ziada's point that this is actually making UX worse by adding an extra step. However, having plan text secrets in templates isn't an acceptable solution to better UX (and IMO is worse than having an extra variable that needs to be defined). I think we need to improve the Azure identity clusterctl experience in general, there are already too many variables that need to be set as part of "quick" start.

nader-ziada:
> I can see @nader-ziada's point that this is actually making UX worse by adding an extra step. However, having plan text secrets in templates isn't an acceptable solution to better UX (and IMO is worse than having an extra variable that needs to be defined). I think we need to improve the Azure identity clusterctl experience in general, there are already too many variables that need to be set as part of "quick" start.

ok fair enough, plain text secrets are bad!

nader-ziada:
btw @alexeldeib can you please squash the commits, thanks

CecileRobertMichon:
Also it's important to note that this change is backwards compatible, so any previously created templates that have the azure.json file as plan text will still continue to work.

nader-ziada:
ok, I'm good with this now, @CecileRobertMichon do you have any other comments?

PR #658, opened 2020-05-29
Adds validation for OS Disk spec

### LINKED ISSUE #557: Add validation for Storage Profile parameters
/kind feature

As part of the AzureMachine webhook, we should validate any spec passed in as part of `OSDisk` to make sure that it is valid and that it won't cause an error when creating the VM storage profile.

**Environment:**

srm09:
Add validation for Storage Profile parameters

Fixes #557 

n/a

**Release note**:

```release-note
Adds validation for OSDisk spec for AzureMachine
```

### DIFF
--- a/api/v1alpha3/azuremachine_validation.go
+++ b/api/v1alpha3/azuremachine_validation.go
@@ -18,7 +18,9 @@ package v1alpha3
 
 import (
 	"encoding/base64"
+	"fmt"
 
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-12-01/compute"
 	"golang.org/x/crypto/ssh"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 )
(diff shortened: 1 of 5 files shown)


srm09:
Does it also make sense to change the [OSDisk reference](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/c0d09b2beae22774c418a357d5ce2e1cb12e84d0/api/v1alpha3/types.go#L199) to a pointer to facilitate a `nil` check for the `OSDisk` field in the `VM` struct?

nader-ziada:
@srm09 I think the reason OSDisk reference in AzureMachine is not a pointer is that its required, check here https://github.com/eBay/Kubernetes/blob/master/docs/devel/api-conventions.md#optional-vs-required

api/v1alpha3/azuremachine_validation.go
```diff
@@ -51,3 +51,20 @@ func ValidateUserAssignedIdentity(identityType VMIdentity, userAssignedIdentetie
 
 	return allErrs
 }
+
+// ValidateOSDisk validates the OSDisk spec
+func ValidateOSDisk(osDisk OSDisk, fieldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	if osDisk.DiskSizeGB <= 0 {
+		allErrs = append(allErrs, field.Invalid(fieldPath.Child("DiskSizeGB"), "", "Disk Size cannot be zero or negative"))
```
CecileRobertMichon, line 60: This also needs to be smaller or equal to 2048: https://docs.microsoft.com/en-us/azure/virtual-machines/windows/managed-disks-overview#os-disk

CecileRobertMichon: We should also not require the disk size to be set, Azure APIs do default the disk size if it's empty, but that's out of scope of this PR since it would require changing the type which would be a breaking change.

https://github.com/Azure/azure-sdk-for-go/blob/master/services/compute/mgmt/2019-12-01/compute/models.go#L10794

srm09: @CecileRobertMichon Does it make sense to add a validation of the actual values of the `StorageAccountType` based on the [allowed SKU Names](https://docs.microsoft.com/en-us/powershell/module/az.compute/new-azdiskconfig?view=azps-1.8.0) and `OSType` based on the accepted values of `OsType` (Windows/Linux)?

CecileRobertMichon: CAPZ doesn't support Windows nodes at the moment so I think only `Linux` should be valid for now. If we allow `Windows` it might be confusing for users and lead them to think that it is supported, so for now I would vote for only allowing `Linux`.

I think it makes sense to validate that `StorageAccountType` is a valid SKU name.

api/v1alpha3/azuremachine_validation.go
```diff
@@ -51,3 +53,38 @@ func ValidateUserAssignedIdentity(identityType VMIdentity, userAssignedIdentetie
...
+func validateStorageAccountType(storageAccountType string, fieldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	storageAccTypeChildPath := fieldPath.Child("ManagedDisk").Child("StorageAccountType")
+
+	if storageAccountType == "" {
+		allErrs = append(allErrs, field.Required(storageAccTypeChildPath, "the Storage Account Type for Managed Disk cannot be empty"))
+		return allErrs
+	}
+
+	for _, possibleStorageAccountType := range compute.PossibleDiskStorageAccountTypesValues() {
```
CecileRobertMichon, line 83: @devigned do you foresee any issues with taking a dependency on the SDK in the CAPZ api package for k8s-infra integration?

CecileRobertMichon: maybe this sort of validation would move to k8s-infra webhooks eventually, but for now it allows failing fast if the user specifies known invalid values.

nader-ziada:
lgtm 
will let @devigned check if he has any issues

api/v1alpha3/azuremachine_validation.go
devigned: The way the infra controller works now, get request to the API for skus and related look up data isn’t implemented. It probably should be, and I’ll need to think about that.

As for quick validation, can’t argue with the effectiveness of having that check here.

PR #666, opened 2020-06-01
🏃 Add cluster-name to kubeadm controllerManager args

CecileRobertMichon:
**What this PR does / why we need it**: Without this explicitly set, kubeadm will use the default name "kubernetes": https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/apis/kubeadm/v1beta1/defaults.go#L46:1

This is used, for example, by cloud-provider to name resources created such as Load Balancers.

Fixes #

**Release note**:

```release-note
Add ClusterName to kubeadm ClusterConfiguration
```

### DIFF
--- a/templates/cluster-template-external-cloud-provider.yaml
+++ b/templates/cluster-template-external-cloud-provider.yaml
@@ -57,6 +57,7 @@ spec:
           allocate-node-cidrs: "false"
           cloud-config: /etc/kubernetes/azure.json
           cloud-provider: external
+          cluster-name: ${CLUSTER_NAME}
         extraVolumes:
         - hostPath: /etc/kubernetes/azure.json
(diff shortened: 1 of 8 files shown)


CecileRobertMichon:
/hold

need to verify this works.

CecileRobertMichon:
So it looks like this shouldn't actually be needed as CABPK already default clusterName to the `Cluster.metadata.name`, https://github.com/kubernetes-sigs/cluster-api/blob/master/bootstrap/kubeadm/README.md#L96.

There might actually be a bug in kubeadm not passing that clusterName to cloud-controller-manager.

CecileRobertMichon:
/hold cancel

Adding cluster-name as a controllerManager arg works.

PR #667, opened 2020-06-02
Add support for FailureDomains to AzureMachinePool

### LINKED ISSUE #663: Support FailureDomains for AzureMachinePools
/kind feature

https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/572 added failure domain support for `AzureMachines` and `AzureClusters`. However, `AzureMachinePools` don't have knowledge of failure domains / Azure availability zones at the moment. 

This will most likely require some changes in CAPI to bring failure domain support in `MachinePools` prior to making the change in CAPZ for `AzureMachinePools`.

For more info on VMSS + availability zones see https://docs.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-use-availability-zones.

**Environment:**

fiunchinho:
It was not possible to choose the `FailureDomains` when creating a `MachinePool` because it was using a custom `AzureMachineTemplateSpec`. This PR changes the code so that `MachinePool` uses the same `AzureMachineTemplateSpec` than other parts of the code.

Fixes #663 

**Release note**:

```release-note
Added `FailureDomains` field to `AzureMachinePoolSpec`

```


fiunchinho:
Build is failing with
```
  Incompatible changes:
  - AzureMachinePoolSpec.Template: changed from AzureMachineTemplate to sigs.k8s.io/cluster-api-provider-azure/api/v1alpha3.AzureMachineSpec
  - AzureMachineTemplate: removed
```

Not sure how to solve it or what does it mean. I need directions, please.

nader-ziada:
@fiunchinho I believe this is informational to bring attention to the fact that there are breaking changes to the api

CecileRobertMichon:
/hold

@devigned @juan-lee I remember we had a discussion about this during machine pool implementation, what was the reasoning for using a different spec?

devigned:
I believe this was the choice since AzureMachineSpec is likely to have more settings in the VM specific type. One example is in AzureMachinePool, [FailureDomains (AZs)](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/9fc12dcddbf48d5a9c6749bf6e7f31b13ae56c50/api/v1alpha3/azuremachine_types.go#L39-L41) would likely be set on the AMP rather than on the AzureMachineSpec since they are provider controlled. 

Another problem would be OSDisk. The name of the OSDisk is not honored in VirtualMachineScaleSets. If you specify a name, the PUT to VMSS fails.

I believe there were other concerns where the template would differ.

Should there be some base structure that are shared, perhaps. Though, I would prefer a little copy paste of data structure rather than more complex composition to reduce lines of code. My 2¢.

fiunchinho:
So instead of removing `AzureMachineTemplate` from `azuremachinepools_types.go`, should I add a `FailureDomain` field to it?

CecileRobertMichon:
> I think FailureDomains should be []string on the AzureMachinePoolSpec, rather than the AzureMachineTemplate since the underlying provider will decide how to balance machines across FDs.
FailureDomains on the AzureMachinePoolSpec would map to Zones: []string in the VMSS REST API.
Does this make sense to folks?

We definitely want to leverage [Scale Set AZ placement](https://docs.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-use-availability-zones) rather than defining our own logic for placing individual machines in zones, that was one of the motivations for implementing MachinePools / VMSS in the first place.  Should it be of type `clusterv1.FailureDomains` to match the existing failure domain type for AzureCluster though?

I think the default behavior needs to try to spread instances on all the available failure domains (from the `AzureCluster` status), and only use the zones defined in the `AzureMachinePoolSpec` for explicit placement, as described in https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/docs/topics/failure-domains.md#L17.

We also had [a discussion in the CAPZ office hours](https://docs.google.com/document/d/1P2FrRjuCZjGy0Yh72lwWCwmXekSEkqliUVTmJy_ETIk/edit?disco=AAAAJooakEI) with @richardcase about changing the existing logic and not needing the field in `AzureMachine`, we need to follow up on that.

devigned:
@CecileRobertMichon thank you for the correction and further explanation!

fiunchinho:
Ok, updated code just adding a new field `FailureDomains` to `AzureMachinePoolSpec`. If I understand correctly, an `AzureMachinePool` controller would create the VMSS using all the `FailureDomains` found in the status of the referenced `AzureCluster` if `AzureMachinePool.Spec.FailureDomains` is nil. Otherwise it would use `AzureMachinePool.Spec.FailureDomains`.

richardcase:
> We also had a discussion in the CAPZ office hours with @richardcase about changing the existing logic and not needing the field in AzureMachine, we need to follow up on that.
 
@CecileRobertMichon - i just raised issue #679 for that, sorry i meant to do it sooner.

fiunchinho:
Test failed, not sure how to see the errors.

exp/api/v1alpha3/azuremachinepool_types.go
```diff
@@ -51,6 +52,10 @@ type (
 
 	// AzureMachinePoolSpec defines the desired state of AzureMachinePool
 	AzureMachinePoolSpec struct {
+		// FailureDomains is the list of failure domains this MachinePool should be attached to, as defined in Cluster API.
+		// This relates to an Azure Availability Zone.
+		FailureDomains clusterv1.FailureDomains `json:"failureDomains,omitempty"`
```
CecileRobertMichon, line 57: where is this consumed by vmss?

devigned: If this were to follow the convention established by `MachineSpec.FailureDomain`, shouldn't this really be on `MachinePoolSpec`?

fiunchinho: I was preparing a different PR with changes in `vmss`. Should I include all of them in this PR?

CecileRobertMichon: yes please, 2 commits but same PR would be ideal.

CecileRobertMichon: oh also @fiunchinho, @devigned's comment above is on point. This should really go in the CAPI `MachinePool` spec. The reason it was in both `Machine` and `AzureMachine` was to allow for back compat with Availability Zones... see https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/679#issuecomment-638931845 

So actually I would expect the CAPZ change to _only_ be the `vmss` consumption.

fiunchinho: Instead of adding the field to `MachinePoolSpec` , shouldn't CAPZ use `MachinePoolSpec.Template.Spec.FailureDomain`?

fiunchinho: Oh I realized that the field is `FailureDomain *string`, and we need a list of `FailureDomains`. I'll try adding the field to `MachinePool` then.

fiunchinho:
Submitted https://github.com/kubernetes-sigs/cluster-api/pull/3157

fiunchinho:
Given that https://github.com/kubernetes-sigs/cluster-api/pull/3157 gets merged, any feedback about the implementation in this PR?

devigned:
> failureDomain (string): the string identifier of the failure domain the instance is running in for the purposes of backwards compatibility and migrating to the v1alpha3 FailureDomain support (where FailureDomain is specified in Machine.Spec.FailureDomain). This field is meant to be temporary to aid in migration of data that was previously defined on the provider type and providers will be expected to remove the field in the next version that provides breaking API changes, favoring the value defined on Machine.Spec.FailureDomain instead. If supporting conversions from previous types, the provider will need to support a conversion from the provider-specific field that was previously used to the failureDomain field to support the automated migration path.

via [Machine Infra Provider Spec](https://cluster-api.sigs.k8s.io/developer/providers/machine-infrastructure.html#data-types) 

With the above guidance in mind, I would expect the following pending https://github.com/kubernetes-sigs/cluster-api/pull/3157.

`FailureDomains []string` on `MachinePool` would be mapped into a `Zones []string` on the [VMSSSpec](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/a8da4fcf73593324542e6db738639fc8a2dde6f6/cloud/services/scalesets/vmss.go#L36), which would then be used to set `compute.VirtualMachineScaleSet.Zones`.

The `compute.VirtualMachineScaleSet.Zones` property is immutable after creation in Azure. If a change occurs to `MachinePool.Zones` after the `AzureMachinePool` has been created, then the `AzureMachinePool` and the `MachinePool` should go into a failed state with an error message indicating the reason for the failure.

@CecileRobertMichon, what do you think about having Zones on the status for `AzureMachinePool` since zones can only be specified on create?

Anyone have any other thoughts or fill in any blanks I missed?

CecileRobertMichon:
> The compute.VirtualMachineScaleSet.Zones property is immutable after creation in Azure. If a change occurs to MachinePool.Zones after the AzureMachinePool has been created, then the AzureMachinePool and the MachinePool should go into a failed state with an error message indicating the reason for the failure.

There should be a webhook validation on Update() that doesn't allow changes to FailureDomains if this applies to all providers (do we know of any cases where failure domains would be mutable?) that way it the machine pool never goes into failed state.

> @CecileRobertMichon, what do you think about having Zones on the status for AzureMachinePool since zones can only be specified on create?
That makes sense to me but we don't do this for Machines currently, we should be consistent and do it for both? Maybe as a follow up? Keep in mind that AZs aren't supported in every region so that field won't always be set.

devigned:
> That makes sense to me but we don't do this for Machines currently, we should be consistent and do it for both? Maybe as a follow up? Keep in mind that AZs aren't supported in every region so that field won't always be set.

The more I think about this, the less value I think it provides. If the `MachinePoolSpec` says it is in those zones and the infrastructure is in a succeeded state, then I think we are right to assume the zones in the spec are reconciled to the infrastructure.

> There should be a webhook validation on Update() that doesn't allow changes to FailureDomains if this applies to all providers (do we know of any cases where failure domains would be mutable?) that way it the machine pool never goes into failed state.

This would be great if we could say FailureDomains are immutable for all, but I don't think we can. For example, [AWS AutoScale groups can add zones](https://docs.aws.amazon.com/autoscaling/ec2/userguide/as-add-availability-zone.html).

CecileRobertMichon:
@fiunchinho would you like to move forward with this PR now that capi v0.3.7 is in? It looks like it's very close

fiunchinho:
Superseeded by https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1180

PR #646, opened 2020-05-22
✨ Add single stack IPv6 support

jsturtevant:
**What this PR does / why we need it**:
Adds ipv6 support. To use, set `CLUSTER_TEMPLATE` to `cluster-template-ipv6.yaml`.

```
kubectl --kubeconfig=./kubeconfig get nodes -o wide
NAME                              STATUS   ROLES    AGE   VERSION   INTERNAL-IP              EXTERNAL-IP   OS-IMAGE             KERNEL-VERSION     CONTAINER-RUNTIME
ipv6-test-2-control-plane-94sb4   Ready    master   29m   v1.18.8   2001:1234:5678:9abc::5   <none>        Ubuntu 18.04.5 LTS   5.3.0-1034-azure   containerd://1.3.4
ipv6-test-2-control-plane-c5kp6   Ready    master   25m   v1.18.8   2001:1234:5678:9abc::6   <none>        Ubuntu 18.04.5 LTS   5.3.0-1034-azure   containerd://1.3.4
ipv6-test-2-control-plane-zjxw4   Ready    master   32m   v1.18.8   2001:1234:5678:9abc::4   <none>        Ubuntu 18.04.5 LTS   5.3.0-1034-azure   containerd://1.3.4
ipv6-test-2-md-0-wgmtf            Ready    <none>   29m   v1.18.8   2001:1234:5678:9abd::5   <none>        Ubuntu 18.04.5 LTS   5.3.0-1034-azure   containerd://1.3.4
ipv6-test-2-md-0-z6lwq            Ready    <none>   29m   v1.18.8   2001:1234:5678:9abd::4   <none>        Ubuntu 18.04.5 LTS   5.3.0-1034-azure   containerd://1.3.4
```

Co-authored by @jsturtevant, @CecileRobertMichon and @aramase.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #318 

**Special notes for your reviewer**:

Before merging needs:

- [x] unit tests fixed
- [x] e2e tests
- [x] docs

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**Release note**:

```release-note
Add single stack IPv6 support
Deprecate VNet and Subnet CirdrBlock in favor of CIDRBlocks
```

PR #674, opened 2020-06-03
🐛 Fix nodes outbound connectivity with SLB

### LINKED ISSUE #648: machinepool + SLB ingress broken
/kind bug

Provision a CAPZ cluster using the machinepool cluster template.

Apply this https://github.com/alexeldeib/ocsp-demo/blob/master/apply.sh

Try to connect to the load balancer:
```
nc -vz 40.80.200.240 80
nc -vz 40.80.200.240 443
```

Resulted in connection timeouts.

This should produce a publicly trusted https site, and the load balance IP should be reachable

We have a couple problems here, they will be familiar to Azure folks who have worked with VMSS + Standard Load Balancer. 

1) SLB does not provide outbound connectivity by default, as soon as you attach VMSS to SLB you'll start having fun outbound connectivity problems. We *probably* need to precreate the SLB (when desired) and ensure it always has a backend pool + outbound rule for traffic. This configuration will be respected by cloud provider and it will not delete the load balancer, leaving outbound connectivity intact.

2) We reconcile the VMSS and remove the backend pool from the networking configuration. We need to do a get and only set the fields we specify in our own spec. We probably will run into more issues as a result of this; probably worth considering as we move forward refactoring service abstraction @CecileRobertMichon 

**Environment:**

- cluster-api-provider-azure version: 86766a9f651bfda752decded44e16afc5f067810
- Kubernetes version: (use `kubectl version`): v1.17.4

CecileRobertMichon:
**What this PR does / why we need it**: This implements outbound connectivity for all nodes as described in https://github.com/kubernetes-sigs/cloud-provider-azure/tree/master/docs/services#standard-loadbalancer. For each Azure cluster, a new node outbound public IP and public Load Balancer will be created. The Load Balancer's name is the cluster name so that cloud provider can detect the LB exists and reuse the same LB for exposing services. This LB will be created with a backend pool and each node machine (all VMs that are not control-planes) will be added to this backend pool on its creation.  The LB will also have outbound rules defined to allow UDP and TCP outbound traffic from all nodes, as described in[ Azure docs](https://docs.microsoft.com/en-us/azure/load-balancer/load-balancer-outbound-connections#proto).

See also:
https://docs.microsoft.com/en-us/azure/load-balancer/load-balancer-outbound-connections

Partially addresses #648 

**Release note**:

```release-note
Fix nodes outbound connectivity with Standard Load Balancer
```

### DIFF
--- a/cloud/defaults.go
+++ b/cloud/defaults.go
@@ -79,6 +79,11 @@ func GeneratePublicIPName(clusterName, hash string) string {
 	return fmt.Sprintf("%s-%s", clusterName, hash)
 }
 
+// GenerateNodeOutboundIPName generates a public IP name, based on the cluster name.
+func GenerateNodeOutboundIPName(clusterName string) string {
+	return fmt.Sprintf("pip-%s-node-outbound", clusterName)
+}
+
 // GenerateNodePublicIPName generates a node public IP name, based on the NIC name.
 func GenerateNodePublicIPName(nicName string) string {
 	return fmt.Sprintf("%s-public-ip", nicName)
--- a/cloud/services/publicloadbalancers/publicloadbalancers.go
+++ b/cloud/services/publicloadbalancers/publicloadbalancers.go
@@ -33,6 +33,7 @@ import (
 type Spec struct {
 	Name         string
 	PublicIPName string
+	Role         string
 }
 
 // Get provides information about a public load balancer.
@@ -56,11 +57,14 @@ func (s *Service) Reconcile(ctx context.Context, spec interface{}) error {
 	if !ok {
 		return errors.New("invalid public loadbalancer specification")
 	}
-	probeName := "tcpHTTPSProbe"
-	frontEndIPConfigName := "controlplane-lbFrontEnd"
-	backEndAddressPoolName := "controlplane-backEndPool"
-	idPrefix := fmt.Sprintf("/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Network/loadBalancers", s.Scope.SubscriptionID, s.Scope.ResourceGroup())
 	lbName := publicLBSpec.Name
+	frontEndIPConfigName := fmt.Sprintf("%s-%s", publicLBSpec.Name, "frontEnd")
+	backEndAddressPoolName := fmt.Sprintf("%s-%s", publicLBSpec.Name, "backendPool")
+	if publicLBSpec.Role == infrav1.NodeOutboundRole {
+		backEndAddressPoolName = fmt.Sprintf("%s-%s", publicLBSpec.Name, "outboundBackendPool")
+	}
+	idPrefix := fmt.Sprintf("/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Network/loadBalancers", s.Scope.SubscriptionID, s.Scope.ResourceGroup())
+
 	klog.V(2).Infof("creating public load balancer %s", lbName)
 
 	klog.V(2).Infof("getting public ip %s", publicLBSpec.PublicIPName)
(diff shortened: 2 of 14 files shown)


cloud/services/publicloadbalancers/publicloadbalancers.go
```diff
@@ -33,6 +33,7 @@ import (
 type Spec struct {
 	Name         string
 	PublicIPName string
+	Role         string
```
CecileRobertMichon, line 36: @jsturtevant please review if you can as these changes will affect #646 (positively, I hope)

jsturtevant: Looks like a nice improvement 👍

cloud/services/publicloadbalancers/publicloadbalancers.go
```diff
@@ -72,89 +76,94 @@ func (s *Service) Reconcile(ctx context.Context, spec interface{}) error {
...
-							EnableFloatingIP:     to.BoolPtr(false),
-							LoadDistribution:     network.LoadDistributionDefault,
-							FrontendIPConfiguration: &network.SubResource{
+			},
+			OutboundRules: &[]network.OutboundRule{
+				{
+					Name: to.StringPtr("OutboundNATAllProtocols"),
+					OutboundRulePropertiesFormat: &network.OutboundRulePropertiesFormat{
+						Protocol:             network.LoadBalancerOutboundRuleProtocolAll,
+						IdleTimeoutInMinutes: to.Int32Ptr(4),
```
jsturtevant, line 108: In aks engine this is set to 30 mins: 

https://github.com/Azure/aks-engine/blob/95d6127337008ed562d3723a53dbffbc99b2eee2/pkg/api/defaults.go#L421
https://github.com/Azure/aks-engine/blob/108dfabdaaf049b3062d9b6298e4ca03b9b3473e/pkg/api/const.go#L229

Maybe @jackfrancis can confirm?

jackfrancis: Correct, and it's also user-configurable. I can't speak to whether 4 or 30 is a preferable default. I do wonder if it should be user-configurable.

What does this outbound rule timeout govern?

cloud/defaults.go
```diff
@@ -79,6 +79,11 @@ func GeneratePublicIPName(clusterName, hash string) string {
 	return fmt.Sprintf("%s-%s", clusterName, hash)
 }
 
+// GenerateNodeOutboundIPName generates a public IP name, based on the cluster name.
+func GenerateNodeOutboundIPName(clusterName string) string {
+	return fmt.Sprintf("%s-node-ip-outbound", clusterName)
```
devigned, line 84: We may want to think about resource naming conventions: https://docs.microsoft.com/en-us/azure/cloud-adoption-framework/ready/azure-best-practices/naming-and-tagging.

See also https://kubernetes.slack.com/archives/CEX9HENG7/p1590166717280400

cloud/services/publicloadbalancers/publicloadbalancers.go
```diff
@@ -72,89 +76,94 @@ func (s *Service) Reconcile(ctx context.Context, spec interface{}) error {
...
+					Probe: &network.SubResource{
+						ID: to.StringPtr(fmt.Sprintf("/%s/%s/probes/%s", idPrefix, lbName, probeName)),
 					},
 				},
 			},
-		})
+		}
+	}
+
+	err = s.Client.CreateOrUpdate(ctx,
```
jsturtevant, line 163: nit: not sure what the formatting rules are but this feels like it should be on a single line

cloud/services/publicloadbalancers/publicloadbalancers.go
CecileRobertMichon: 4 is the default value https://docs.microsoft.com/en-us/azure/load-balancer/load-balancer-outbound-connections#idletimeout and is also the same value CAPZ is already using so didn't want to change the default on this PR (this block of code just moved down).
We can look at making it configurable if there is a use case for configuring that value separately.

jsturtevant: The comment above the linked line seems like it might have been a bug at some point?

>// DefaultOutboundRuleIdleTimeoutInMinutes determines the aks-engine provided default for IdleTimeoutInMinutes of the OutboundRule of the agent loadbalancer
	// This value is set greater than the default Linux idle timeout (15.4 min): https://pracucci.com/linux-tcp-rto-min-max-and-tcp-retries2.html

cloud/services/publicloadbalancers/publicloadbalancers.go
CecileRobertMichon: fixed, this was leftover from when lb was the huge struct definition

cloud/services/publicloadbalancers/publicloadbalancers.go
CecileRobertMichon: https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/676

jackfrancis: Understood. I'm not sure why the default is set so high in AKS Engine, as it would seem to have the side-effect of reducing port availability (leading to exhaustion).

In any event, having read about what this does I think the proper value of this is going to be very much app domain-specific, so: https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/677

cloud/defaults.go
CecileRobertMichon: Good point, thanks for pointing me to this resource (didn't know it existed). I will fix this name, and I opened https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/678 for broader auditing of resource naming.

CecileRobertMichon: renamed to follow best practices but I kept the cluster name in there for now because it technically allows having multiple CAPZ clusters in one RG without conflict since public IPs are scoped to the RG. In the future, we should determine whether that's a real requirement and, if not, clearly document that each cluster has to be in its own RG (so we can remove the cluster-name / resource naming dependency).

CecileRobertMichon:
/hold

I found a potential bug, investigating...

CecileRobertMichon:
ok so the bug is that cloud-provider can't find the VMSS instances to add them to the backend pool because host name != instance name, not related to this PR (it's already in master). I will work on a fix separately, this can go in as-is for now since it fixes the other issue.

/hold cancel

CecileRobertMichon:
The apidiff failure is due to:

```
sigs.k8s.io/cluster-api-provider-azure/cloud/services/scalesets
  Compatible changes:
  - Service.PublicLoadBalancersClient: added
  - Spec.PublicLoadBalancerName: added
sigs.k8s.io/cluster-api-provider-azure/cloud/services/networkinterfaces
  Compatible changes:
  - Spec.MachineRole: added
  Incompatible changes:
  - APIServerRoleTagValue: removed
  - BastionRoleTagValue: removed
  - CommonRoleTagValue: removed
  - PrivateRoleTagValue: removed
  - PublicRoleTagValue: removed
  Compatible changes:
  - APIServerRole: added
  - BastionRole: added
  - CommonRole: added
  - NodeOutboundRole: added
  - PrivateRole: added
  - PublicRole: added
sigs.k8s.io/cluster-api-provider-azure/cloud/services/publicloadbalancers
  Compatible changes:
  - Spec.Role: added
sigs.k8s.io/cluster-api-provider-azure/cloud
  Compatible changes:
  - GenerateNodeOutboundIPName: added
```

No API breaking changes.

CecileRobertMichon:
fixed the other issue in #680

PR #680, opened 2020-06-04
🐛 MachinePool template must have vmType vmss in control plane azure.json

CecileRobertMichon:
**What this PR does / why we need it**: This fixes an issue where the cloud provider is unable to find vmss instances when using MachinePool because it tries to find them as availability sets:

```
Warning  SyncLoadBalancerFailed  10m (x19 over 75m)  service-controller  Error syncing load balancer: failed to ensure load balancer: ensure(ingress-nginx/ingress-nginx-controller): backendPoolID(/subscriptions/85d99e6d-f6d6-408f-a9f1-b7a97237d5c4/resourceGroups/capz-vmss/providers/Microsoft.Network/loadBalancers/capz-vmss/backendAddressPools/capz-vmss) - failed to ensure host in pool: "instance not found"
``` 

As a follow up, I'd like to investigate if and how Azure cloud provider could support vmss and vmas in the same cluster (https://github.com/kubernetes-sigs/cloud-provider-azure/issues/338)

Partially addresses #648 

**Release note**:

```release-note
Change MachinePool template vmType to "vmss" in control plane azure.json
⚠️ Cloud provider for Azure does not currently support clusters in mixed mode (both vmss and vmas node pools), so it is not supported to have both `AzureMachinePools` and `AzureMachines` in the same workload cluster.
```

### DIFF
--- a/docs/topics/machinepools.md
+++ b/docs/topics/machinepools.md
@@ -36,6 +36,8 @@ Cluster API Provider Azure (CAPZ) has experimental support for `MachinePool` tho
 type `AzureMachinePool`. An `AzureMachinePool` corresponds to an [Azure Virtual Machine Scale Set](https://docs.microsoft.com/en-us/azure/virtual-machine-scale-sets/overview),
 which provides the cloud provider specific resource for orchestrating a group of Virtual Machines.
 
+⚠️ Cloud provider for Azure does not currently support clusters in mixed mode (both vmss and vmas node pools), so it is not supported to have both `AzureMachinePools` and `AzureMachines` in the same workload cluster.
+
 ### Using `clusterctl` to deploy
 To deploy a MachinePool / AzureMachinePool via `clusterctl config` there's a [flavor](https://cluster-api.sigs.k8s.io/clusterctl/commands/config-cluster.html#flavors) 
 for that.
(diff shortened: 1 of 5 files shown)


devigned:
That must have been a regression since machine pool would have never worked without that.

CecileRobertMichon:
@devigned it "worked" as in the cluster comes up and nodes are ready but exposing a service never worked with vmss that I know of?

devigned:
I see... This is the control plane, rather than the workload cluster. The cluster would not come up, "work", if the vmType was Standard.

This seems really weird :(.

chewong:
Getting the following error in kube-controller-manager:
```
2020-06-04T19:51:21.558078299Z stderr F E0604 19:51:21.557974       1 node_lifecycle_controller.go:171] error checking if node capz-pvmwf5-mp-0000002 exists: error splitting providerID
```

which causes

```
2020-06-04T19:54:32.464499812Z stderr F E0604 19:54:32.464421       1 azure_controller_common.go:225] error of getting data disks for node "capz-pvmwf5-mp-0000002": instance not found
```

in https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/kubernetes-e2e-capz-azure-disk-vmss-1-16/1268621763684601857

I think it's because of this issue, kube-controller-manager is trying to use the [VMAS provider ID regex](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/legacy-cloud-providers/azure/azure_standard.go#L75) to parse VMSS provider ID. This PR should fix the issue.

CecileRobertMichon:
@devigned can you please lgtm this if it looks good to you?

PR #683, opened 2020-06-05
🐛  s/vm/vmss for MachinePool's azure.json

chewong:

Follow-up PR for https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/680.

/test pull-cluster-api-provider-azure-conformance-v1alpha3
/test pull-cluster-api-provider-azure-conformance-machine-pool-v1alpha3

Fixes #

**Release note**:

```release-note
None
```


hack/log/log-dump.sh
```diff
@@ -14,7 +14,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-set -o errexit
```
CecileRobertMichon, line 17: why does this need to be removed? Instead of doing this, it might be better to change the commands that we want to run as best effort with `|| echo "failed to do xyz"`.

chewong: Added https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/683/files#diff-a9012e009906a8e18d16f7b23dfc67e1R129

PR #685, opened 2020-06-05
[WIP] :bug: only update VMSS capacity if scale set already exists

### LINKED ISSUE #648: machinepool + SLB ingress broken

CecileRobertMichon:
**What this PR does / why we need it**: Currently, we do a `CreateOrUpdate` on every reconcile loop for scale sets. However, the k8s Azure cloud provider also modifies scale sets in the cluster, for example to add them to a LB backendpool in order to expose a service. We need to make sure we don't overwrite the changes made by cloud provider. 

Fixes #648 

**Special notes for your reviewer**: Unfortunately, `VirtualMachineScaleSets` don't support `Etag` so I wasn't able to follow a similar strategy as NSG reconcile (`Get() -> apply changes -> `CreateOrUpdate()` iff unchanged). We can't do the same thing as VMs and NICs either (only create on the first loop then skip reconcile if it already exists) since we need to reconcile VMSS capacity for scaling. I decided to leverage the `Update()` approach for now and only reconcile the capacity of the scale set since the number of replicas on the machine pool should be the source of truth.

**Release note**:

```release-note
Only update VMSS capacity if scale set already exists
```


devigned:
🚀 great idea! It's not a cure, but definitely seems like a better, more conservative approach to updating the scale set.

One thing that concerns me is how to deal with changes to the model, like upgrading K8s version. What do you think about that scenario?

CecileRobertMichon:
> One thing that concerns me is how to deal with changes to the model, like upgrading K8s version. What do you think about that scenario?

@devigned that's a great point, definitely to support upgrading the k8s version. We basically need to curate 1 by 1 each property that CAPZ should be the source of truth for and "owns" fully. I can expand the update parameters to include other things.

cloud/services/scalesets/client.go
```diff
@@ -33,6 +33,7 @@ type Client interface {
 	ListInstances(context.Context, string, string) ([]compute.VirtualMachineScaleSetVM, error)
 	Get(context.Context, string, string) (compute.VirtualMachineScaleSet, error)
 	CreateOrUpdate(context.Context, string, string, compute.VirtualMachineScaleSet) error
+	Update(context.Context, string, string, compute.VirtualMachineScaleSetUpdate) error
```
awesomenix, line 36: Does it require mocks to be generated for Update?

cloud/services/scalesets/vmss.go
```diff
@@ -77,6 +77,20 @@ func (s *Service) Reconcile(ctx context.Context, spec interface{}) error {
 		return errors.New("invalid VMSS specification")
 	}
 
+	vmss, err := s.Client.Get(ctx, vmssSpec.ResourceGroup, vmssSpec.Name)
```
awesomenix, line 80: Probably you can do 

_, err = <> 

since the vmss output isnt used anywhere, just to avoid confusion

CecileRobertMichon:
I have kept the two commits on purpose to allow comparison between the current approach and the previous one during review, will squash before merge.

CecileRobertMichon:
@devigned let me know what you think of this new approach... Basically, instead of updating capacity only, we update everything _except_ network profile.

PR #686, opened 2020-06-07
📖 docs for managed cluster

### LINKED ISSUE #613: [managed cluster] documentation
/kind documentation

Users should understand how to use managed clusters in CAPZ from our documentation 🙃 We should have example templates as well as markdown around how to deploy a cluster.

alexeldeib:

Add documentation for managed clusters

Fixes #613 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/docs/topics/managedcluster.md
+++ b/docs/topics/managedcluster.md
@@ -1,0 +1,162 @@
+# Managed Clusters (AKS)
+- **Feature status:** Experimental
+- **Feature gate:** AKS=true,MachinePool=true
+
+Cluster API Provider Azure (CAPZ) experimentally supports managing Azure
+Kubernetes Service (AKS) clusters. CAPZ implements this with three
+custom resources:
+- AzureManagedControlPlane
+- AzureManagedCluster
+- AzureManagedMachinePool
+
+The combination of AzureManagedControlPlane/AzureManagedCluster
+corresponds to provisioning an AKS cluster. AzureManagedMachinePool
+corresponds one-to-one with AKS node pools. This also means that
+creating an AzureManagedControlPlane requires defining the default
+machine pool, since AKS requires at least one system pool at creation
+time.
+
+## Deploy with clusterctl
+
+A clusterctl flavor exists to deploy an AKS cluster with CAPZ. This
+flavor requires the following environment variables to be set before
+executing clusterctl.
+
+```bash
+# Kubernetes values
+export CLUSTER_NAME="foobar"
+export WORKER_MACHINE_COUNT=2
+export KUBERNETES_VERSION="v1.17.4"
+
+# Azure values
+export AZURE_LOCATION="southcentralus"
+export AZURE_RESOURCE_GROUP="${CLUSTER_NAME}"
+# set AZURE_SUBSCRIPTION_ID to the GUID of your subscription
+# this example uses an sdk authentication file and parses the subscriptionId with jq
+# this file may be created using 
+#
+# `az ad sp create-for-rbac --sdk-auth [roles] > sp.json`
+#
+# when logged in with a service principal, it's also available using
+#
+# `az account show --sdk-auth`
+#
+# Otherwise, you can set this value manually.
+#
+export AZURE_SUBSCRIPTION_ID="$(cat ~/sp.json | jq -r .subscriptionId | tr -d '\n')"
+export AZURE_NODE_MACHINE_TYPE="Standard_D2s_v3"
+```
+
+Execute clusterctl to template the resources, then apply to a management cluster:
+
+```bash
+# replace v0.4.4 with the appropriate version you're using
+clusterctl config cluster my-cluster --kubernetes-version v1.17.4 --flavor aks -i azure:v0.4.4 > cluster.yaml
+
+# assumes an existing management cluster
+kubectl apply -f cluster.yaml
+
+# check status of created resources
+kubectl get cluster-api -o wide
+```
+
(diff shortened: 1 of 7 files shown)


alexeldeib:
apidiff:
```
  Incompatible changes:
  - AzureManagedControlPlaneSpec.LoadBalancerSKU: removed
```

This is intended, it was an oversight on my part to include it. We only support SLB, and the value is hardcoded in the managed cluster client correctly.

docs/topics/managedcluster.md
```diff
@@ -0,0 +1,147 @@
...
+
+A clusterctl flavor exists to deploy an AKS cluster with CAPZ. This
+flavor requires the following environment variables to be set before
+executing clusterctl.
+
+```bash
+# Kubernetes values
+export CLUSTER_NAME="foobar"
+export WORKER_MACHINE_COUNT=2
+export KUBERNETES_VERSION="1.17.4" # NOTE: no preceding "v", unlike normal for CAPI.
```
CecileRobertMichon, line 29: why is that? is there anyway we can standardize with CAPI and trim the v when using the version? especially since we're reusing the same env var

docs/topics/managedcluster.md
```diff
@@ -0,0 +1,147 @@
...
+
+# Azure values
+export AZURE_LOCATION="southcentralus"
+export AZURE_RESOURCE_GROUP="${CLUSTER_NAME}"
+export AZURE_SUBSCRIPTION_ID="$(cat ~/sp.json | jq -r .subscriptionId | tr -d '\n')"
```
CecileRobertMichon, line 34: is `~/sp.json` from your own dev flow? not everyone will have this in their path so it might be confusing to people

CecileRobertMichon: Does it not need tenant ID, client ID and client secret as well?

docs/topics/managedcluster.md
```diff
@@ -0,0 +1,147 @@
...
+export AZURE_NODE_MACHINE_TYPE="Standard_D8s_v3"
```
CecileRobertMichon, line 35: was it intentional to specify this particular VM size? or just an example? the rest of the capz docs use `Standard_D2s_v3`.

docs/topics/managedcluster.md
```diff
@@ -0,0 +1,147 @@
...
+```
+
+Execute clusterctl to template the resources, then apply to a management cluster:
+
+```bash
+# replace v0.4.4 with the appropriate version you're using
```
CecileRobertMichon, line 41: Do we need to specify the capz version explicitly? or could this just be `--infrastructure azure` to use latest?

docs/topics/managedcluster.md
CecileRobertMichon: also the example template version below has a v prefix https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/686/files#diff-96e57bd2edb999ae76c554c70314028fR85

docs/topics/managedcluster.md
```diff
@@ -0,0 +1,147 @@
...
+apiVersion: exp.infrastructure.cluster.x-k8s.io/v1alpha3
+kind: AzureManagedControlPlane
+metadata:
+  name: my-cluster-control-plane
+spec:
+  defaultPoolRef:
+    name: agentpool0
+  location: southcentralus
+  resourceGroup: foo-bar
+  sshPublicKey: ${BASE64_SSH_PUBLIC_KEY}
```
CecileRobertMichon, line 83: possible to reuse the same env var as other templates? `AZURE_SSH_PUBLIC_KEY`

docs/topics/managedcluster.md
```diff
@@ -0,0 +1,147 @@
...
+apiVersion: exp.infrastructure.cluster.x-k8s.io/v1alpha3
+kind: AzureManagedMachinePool
+metadata:
+  name: agentpool0
+spec:
+  osDiskSizeGB: 512
+  sku: Standard_D8s_v3
+```
+
+The main features for configuration today are networkPolicy and
```
CecileRobertMichon, line 122: let's link to https://docs.microsoft.com/en-us/azure/aks/concepts-network#network-policies and https://docs.microsoft.com/en-us/azure/aks/concepts-network#azure-virtual-networks for further info

CecileRobertMichon:
@alexeldeib can you please split the docs and API change into two commits?

docs/topics/managedcluster.md
alexeldeib: because AKS, but we can change the code to trim as you suggest, sure. and whoops

docs/topics/managedcluster.md
alexeldeib: I can change this to echo a variable like the other docs, I do like documenting sdk auth files.

It doesn't need any of the other values, only msi is available right now.

docs/topics/managedcluster.md
alexeldeib: --infrastructure azure didn't work properly for me, i'll try this again maybe?

docs/topics/managedcluster.md
alexeldeib: documented instead

CecileRobertMichon:
@alexeldeib opened #699

docs/topics/managedcluster.md
rbitia: We can keep it as is for now until we get --feature-flags. :)

PR #700, opened 2020-06-11
[WIP] 💚 Allow running CAPI E2E  suite with the CAPZ config

CecileRobertMichon:
**What this PR does / why we need it**: This allows running the [CAPI E2E spec](https://github.com/kubernetes-sigs/cluster-api/tree/master/test/e2e) with the Azure infrastructure provider. The plan is to add a prow job that runs the CAPI test suite with the latest CAPZ commit. For now we still pin the CAPI components to the latest release as the goal is to test CAPZ, not CAPI. It is also more convenient to be able to control the CAPI version in the CAPZ checked in code and not requiring a change to test-infra every time we adopt a new CAPI release. In the future, we might want to use Azure infra to test CAPI changes and build a dev version of CAPI components by adding a separate config.

By default, the test-e2e script will still run only CAPZ E2E suite, with options to turn on the CAPI E2E suite.

This is still work in progress, opening to get feedback from @fabriziopandini and start testing on prow.

See also [CAPZ testing plan doc](https://docs.google.com/document/d/1a6sx3nSAcy7JGxTWhT99_0YGL7-Kqhzm02il0rKAwrc/edit?usp=sharing)

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note

```


CecileRobertMichon:
@fabriziopandini the error I'm running into right now is `failed to perform variable substitution: value for variables [AZURE_RESOURCE_GROUP, AZURE_VNET_NAME] is not set. Please set the value using os environment variables or the clusterctl config file`

The Azure vnet and resource group name should be unique per workload cluster so I can't simply set them to a static value in the config. Is there a way to have them depend on `CLUSTER_NAME` and/or set them for each test case?

Right now for the CAPZ E2E suite that uses the framework we set them within the tests https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/bb2e24bcac540b9a5e59c04a6cd25b1df38106e8/test/e2e/create_workload_cluster_test.go#L60

CecileRobertMichon:
Nevermind, found a better way to do this thanks to @sedefsavas - will open a new PR.

Issue #609, opened 2020-05-12
Refactor: Services should not return interfaces


CecileRobertMichon:
> ⚠️ Cluster API Azure maintainers can ask to turn an issue-proposal into a [CAEP](https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/proposals/YYYYMMDD-template.md) when necessary. This is to be expected for large changes that impact multiple components, breaking changes, or new large features.

**Detailed Description**

_Follow up on https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/482#discussion_r421789468_

We should be returning SDK types for Azure resources in services/ instead of interface{}.

Also, we should remove unused service mocks (we mock the Azure Clients and those are the ones used in tests):
- remove unused https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/cloud/fake_interfaces.go
- remove unused https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/cloud/mocks/service_mock.go#L31

related to #110 

**Contract changes [optional]**

[Describe contract changes between Cluster API Azure Provider controllers, if applicable.]

**Data model changes [optional]**

[Describe contract changes between Cluster API Azure Provider controllers, if applicable.]

/kind proposal

CecileRobertMichon:
@juan-lee are you still interested in helping with this?

CecileRobertMichon:
cc @alexeldeib @justaugustus @devigned

devigned:
I'd caution from returning SDK types from the services. Having dependencies on AzSDK for Go types beyond the service layer implies a tighter coupling to the SDK.

Perhaps, we should consider the services layer returns a spec + status type (maybe not a `runtime.Object`, but something with all the info needed by callers).

CecileRobertMichon:
What if the `Get()` functions were never called from outside `services`? The way I see this working is you would only have `Ensure()`/`Reconcile()` and `Delete()` be public and be called from the controllers (see #110). Reconcile and Delete should both only return `error`, the SDK type return would only be for intermediate `Get`s.

devigned:
What do you think about `Get()` being `get()`?

CecileRobertMichon:
[...]

Issue #110, opened 2019-03-04
Adopt a Reconcile/Delete pattern for exported methods


justaugustus:
**Describe the solution you'd like**
The goals are two-fold here:
1. Adopt a consistent pattern/interface for retrieving/modifying/deleting Cluster/Machine/Azure resources
2. Minimize calls to Azure by storing more information in Cluster & Machine objects

For each actuator reconciler, we currently have Reconcile and Delete methods which operate over the lower-level Azure resources. 

Instead of handling implementation details in the top-level reconciler, we should scope them to the resources that they're operating over. As an example, a `SubnetSpec` should instead be generated/manipulated by a subnet reconciler:
https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/0b7cf771058e2cef05e99b048138454afd12616e/pkg/cloud/azure/actuators/cluster/reconciler.go#L108-L113

As a concrete reference, see the [`ReconcileNetwork()`](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/c694a556e993c243326d63e38f09ee167dfdd6b8/pkg/cloud/aws/services/ec2/network.go#L23-L59) method for Cluster API Provider AWS.

A secondary problem is that in our current actuator reconciler model, we make a variety of Create/Update calls to the Azure API, in instances where a Get would suffice. This means we're attempting to modify a majority of the Azure resources we create on every round-trip of a reconciler.

Finally, we do not fully utilize the defined types in the capz API, instead accepting an [ambiguous spec](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/0b7cf771058e2cef05e99b048138454afd12616e/pkg/cloud/azure/interfaces.go#L29-L30) and returning a loosely-defined interface, instead of a concrete type. 

This means that we are required to query Azure for cluster/machine information, instead of being able to reference the `ClusterSpec`, `ClusterStatus`, `MachineSpec`, and `MachineStatus` directly, which should be our sources of truth.

---

As a solution, we can:
- Redefine the `Service` interface to present only Reconcile and Delete methods
- Add a temporary interface (`GetterService`) to represent packages that don't adhere to the Reconcile/Delete model
- Refactor packages to adhere to Reconcile/Delete model, which will convert and store Azure resource information in the relevant Spec or Status
- Redefine package-level `CreateOrUpdate`/`Get`s as unexported methods (only to be used by the package-level Reconcile/Delete)
[...]

PR #713, opened 2020-06-15
💎 refactor most services into Reconcile/Delete pattern

CecileRobertMichon:
**What this PR does / why we need it**: Remove the exported `Get()` function from every Service except availability zones, virtual machines, scale sets, and managed clusters (these will be addressed in follow-ups.)

In order to keep the size of this PR manageable and facilitate reviews, I tried to scope this refactor down to only Services that did have Get() calls within the controller/reconciler. 

** ⚠️ Important:** This is **NOT** the north star. This is closer to the end goal: 
https://github.com/CecileRobertMichon/cluster-api-provider-azure/compare/reconcile-delete...CecileRobertMichon:public-ips-scope?expand=1#diff-890b2d661da1708cb3d23d43fc548ceeR28
This PR is just an intermediary step to get there.

Future work:
- Adopt Reconcile/Delete pattern for virtual machines, scale sets and managed clusters
- remove dependency on Availability Zones Get() within the AzureMachine controller
- stop using a mix of spec + scope in services as input (WIP [here](https://github.com/CecileRobertMichon/cluster-api-provider-azure/compare/reconcile-delete...CecileRobertMichon:public-ips-scope?expand=1))

Partially addresses #110 and #609 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```


controllers/azuremachine_reconciler.go
```diff
@@ -40,14 +40,14 @@ const (
...
-// azureMachineService are list of services required by cluster actuator, easy to create a fake
+// azureMachineService is the group of services called by the AzureMachine controller
 type azureMachineService struct {
 	machineScope         *scope.MachineScope
 	clusterScope         *scope.ClusterScope
 	availabilityZonesSvc azure.GetterService
 	networkInterfacesSvc azure.Service
-	virtualMachinesSvc   azure.GetterService
-	disksSvc             azure.GetterService
+	virtualMachinesSvc   *virtualmachines.Service
```
CecileRobertMichon, line 49: this is temporary to allow removing use of GetterService. In a follow up, `virtualMachinesSvc` will also become  an `azure.Service` after we remove all Get() calls from the controller (see #110 for more info).

PR #714, opened 2020-06-17
✨ initial support for conditions for AzureCluster and AzureMachine

### LINKED ISSUE #671: Adopt Conditions for AzureCluster, AzureMachine and AzureMachinePool Status
/kind feature

The first step would be to define and get a consensus on which condition types should be defined for each CAPZ reconciled object: `AzureCluster`, `AzureMachine` and `AzureMachinePool`.

- Condition types MUST have a consistent polarity (i.e. "True = good");

- Condition types SHOULD have one of the following suffix: Ready, for resources which represent an ongoing status, like ControlplaneReady or MachineDeploymentsReady.
Succeeded, for resources which run-to-completion, e.g. CreateVnetSucceeded

Some examples could be:  `VMReady`, `CreateSecurityGroupsSucceeded`, `VMSSReady`, etc.

An aggregate Ready condition is also required to report the overall status.

Original Conditions [CAEP](https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/proposals/20200506-conditions.md)

CAPI Conditions implementation was merged in CAPI [recently](https://github.com/kubernetes-sigs/cluster-api/pulls?q=is%3Apr+add+conditions+is%3Aclosed+) and should be available in the next CAPI release (v0.3.7 or v0.3.8).

**Environment:**

nader-ziada:

Fixes #671 
- Add Conditions to AzureClusterStatus and AzureMachineStatus

Adding hold as this uses CAPI@master in order to get the conditions in progress work. should wait for the next CAPI release
/hold

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Add `Status.Conditions` field to `AzureCluster` and `AzureMachine` showing details about the current state of these objects.
```


api/v1alpha3/conditions_consts.go
```diff
@@ -0,0 +1,54 @@
...
+	// WaitingForClusterInfrastructureReason used when machine is waiting for cluster infrastructure to be ready before proceeding.
+	WaitingForClusterInfrastructureReason = "WaitingForClusterInfrastructure"
+	// WaitingForBootstrapDataReason used when machine is waiting for bootstrap data to be ready before proceeding.
+	WaitingForBootstrapDataReason = "WaitingForBootstrapData"
+)
+
+// AzureMachine Conditions and Reasons
+const (
+	// InstanceReadyCondition reports on current status of the Azure VM. Ready indicates the instance is in a Running state.
+	InstanceReadyCondition clusterv1.ConditionType = "InstanceReady"
```
CecileRobertMichon, line 41: Let's try to align these with https://docs.microsoft.com/en-us/azure/virtual-machines/linux/states-lifecycle

CecileRobertMichon: also should be "VM*" to be consistent with Azure naming. "Instance" is usually how AWS machines are described.

nader-ziada: good point 👍

nader-ziada:
@fabriziopandini would appreciate your review when you have time

api/v1alpha3/conditions_consts.go
nader-ziada: changed the name to VM, but still need to make sure its aligned with the `MS states-lifecycle`

api/v1alpha3/conditions_consts.go
```diff
@@ -0,0 +1,52 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha3
+
+import clusterv1 "sigs.k8s.io/cluster-api/api/v1alpha3"
+
+const (
+	MachineConditionCount      = 2
```
fabriziopandini, line 22: After https://github.com/kubernetes-sigs/cluster-api/pull/3152 merged it is not required anymore to define conditions Summary, you can remove those constants

api/v1alpha3/conditions_consts.go
```diff
@@ -0,0 +1,52 @@
...
+	ControlPlaneConditionCount = 3
+	ClusterConditionCount      = 2
+)
+
+// AzureCluster Conditions and Reasons
+const (
+	NetworkInfrastructureReadyCondition = "NetworkInfrastructureReady"
```
fabriziopandini, line 29: FYI in CAPI project we are commenting all the public const used for conditions.
e.g. https://github.com/kubernetes-sigs/cluster-api/blob/53e98d22ccc1510a18b2caead34f4a8014a29983/api/v1alpha3/condition_consts.go#L57-L73

api/v1alpha3/conditions_consts.go
```diff
@@ -0,0 +1,52 @@
...
+	APIServerProvisioningReason         = "APIServerProvisioning"
+	APIServerProvisioningFailedReason   = "APIServerProvisioningFailed"
+	// WaitingForClusterInfrastructureReason used when machine is waiting for cluster infrastructure to be ready before proceeding.
+	WaitingForClusterInfrastructureReason = "WaitingForClusterInfrastructure"
+	// WaitingForBootstrapDataReason used when machine is waiting for bootstrap data to be ready before proceeding.
+	WaitingForBootstrapDataReason = "WaitingForBootstrapData"
```
fabriziopandini, line 35: Am I wrong or `WaitingForClusterInfrastructureReason` and `WaitingForBootstrapDataReason` applies to AzureMachine conditions? Should we move this four lines the next conditions block?

api/v1alpha3/conditions_consts.go
```diff
@@ -0,0 +1,52 @@
...
+const (
+	MachineConditionCount      = 2
+	ControlPlaneConditionCount = 3
+	ClusterConditionCount      = 2
+)
+
+// AzureCluster Conditions and Reasons
+const (
+	NetworkInfrastructureReadyCondition = "NetworkInfrastructureReady"
+	APIServerProvisioningReason         = "APIServerProvisioning"
```
fabriziopandini, line 30: Having `APIServerProvisioning` as a step for `NetworkInfrastructureReady` confuses me a little bit...

Is this meant to provision a load balancer endpoint for the API server? Should we rename this "LoadBalancerProvisioning" or something more specific? FYI, this is the current proposal for VSphereCluster https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/c66649e1d9c8ec2597aee8fb81f0a61ca604d433/api/v1alpha3/condition_consts.go#L24-L33

api/v1alpha3/conditions_consts.go
fabriziopandini: FYI in CAPA we decided to provide user a better visibility on what's going on during the Network infrastructure provisioning because this operation might take some time and we want to give Users a perception that things are progressing. This is still WIP, but I do expect that we will document ongoing steps (VPCProvisioning, SecurityGroups provisioning, Elastic IP provisioning etc.)

I don't know if a similar consideration apply to CAPZ as well (network infrastructure provisioning takes time)

controllers/azuremachine_controller.go
```diff
@@ -149,6 +150,20 @@ func (r *AzureMachineReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, ret
 
 	// Always close the scope when exiting this function so we can persist any AzureMachine changes.
 	defer func() {
+		if machineScope.IsControlPlane() {
```
fabriziopandini, line 153: is this if statement necessary, given that if and else branch apparently are the same?

controllers/azuremachine_controller.go
```diff
@@ -235,29 +252,35 @@ func (r *AzureMachineReconciler) reconcileNormal(ctx context.Context, machineSco
 	switch vm.State {
 	case infrav1.VMStateSucceeded:
 		machineScope.V(2).Info("VM is running", "id", *machineScope.GetVMID())
+		conditions.MarkTrue(machineScope.AzureMachine, infrav1.VMReadyCondition)
 		machineScope.SetReady()
 	case infrav1.VMStateCreating:
 		machineScope.V(2).Info("VM is creating", "id", *machineScope.GetVMID())
+		conditions.MarkFalse(machineScope.AzureMachine, infrav1.VMReadyCondition, infrav1.VMNotReadyReason, clusterv1.ConditionSeverityWarning, "")
```
fabriziopandini, line 259: I'm not sure `VMStateCreating` should be a `ConditionSeverityWarning`, because I assume this is part of the normal provisioning workflow. What about using `ConditionSeverityInfo` instead?
Should we apply the same consideration to `VMStateUpdating`?

api/v1alpha3/conditions_consts.go
nader-ziada: This was the naming used buy the code so I was trying to be consistent, will check the MS documentation

controllers/azuremachine_controller.go
nader-ziada: will remove it

controllers/azuremachine_controller.go
nader-ziada: makes sense

api/v1alpha3/conditions_consts.go
nader-ziada: updated the names to match Azure style

nader-ziada:
@CecileRobertMichon @devigned any thoughts?

api/v1alpha3/azurecluster_types.go
```diff
@@ -96,6 +100,16 @@ type AzureClusterList struct {
 	Items           []AzureCluster `json:"items"`
 }
 
+// GetConditions returns the list of conditions for a AzureCluster API object.
+func (c *AzureCluster) GetConditions() clusterv1.Conditions {
+	return c.Status.Conditions
+}
+
+// SetConditions will set the given conditions on an AzureCluster object
+func (c *AzureCluster) SetConditions(conditions clusterv1.Conditions) {
```
CecileRobertMichon, line 109: Most getters/setters live in the scope package at the moment (`SetProviderID`, `SetReady`, `GetVMState`, etc.). Not sure if that's the best place and maybe that's something we need to discuss separately but I think it makes more sense to keep those together for now.

api/v1alpha3/azurecluster_types.go
```diff
@@ -96,6 +100,16 @@ type AzureClusterList struct {
 	Items           []AzureCluster `json:"items"`
 }
 
+// GetConditions returns the list of conditions for a AzureCluster API object.
```
CecileRobertMichon, line 103:
```suggestion
// GetConditions returns the list of conditions for an AzureCluster API object.
```

api/v1alpha3/azuremachine_types.go
```diff
@@ -182,6 +187,16 @@ type AzureMachineList struct {
 	Items           []AzureMachine `json:"items"`
 }
 
+// GetConditions returns the list of conditions for a AzureMachine API object.
```
CecileRobertMichon, line 190:
```suggestion
// GetConditions returns the list of conditions for an AzureMachine API object.
```

api/v1alpha3/conditions_consts.go
```diff
@@ -0,0 +1,51 @@
...
+	// APIServerProvisioningReason API Server endpoint for the loadbalancer
+	APIServerProvisioningReason = "APIServerProvisioning"
+	// APIServerProvisioningFailedReason used for failure during provisioning of API Server.
+	APIServerProvisioningFailedReason = "APIServerProvisioningFailed"
+)
+
+// AzureMachine Conditions and Reasons
+const (
+	// VMRunningCondition reports on current status of the Azure VM.
+	VMRunningCondition clusterv1.ConditionType = "VMRunning"
```
CecileRobertMichon, line 34: Should these be replacing VMState? Or are they distinct and meant to stay distinct?

controllers/azurecluster_controller.go
```diff
@@ -169,6 +180,7 @@ func (r *AzureClusterReconciler) reconcileNormal(ctx context.Context, clusterSco
 
 	// No errors, so mark us ready so the Cluster API Cluster Controller can pull it
 	azureCluster.Status.Ready = true
+	conditions.MarkTrue(azureCluster, infrav1.NetworkInfrastructureReadyCondition)
```
CecileRobertMichon, line 183: is AzureCluster ready == infrav1.NetworkInfrastructureReadyCondition true?

controllers/azuremachine_controller.go
```diff
@@ -272,29 +281,35 @@ func (r *AzureMachineReconciler) reconcileNormal(ctx context.Context, machineSco
 	switch vm.State {
 	case infrav1.VMStateSucceeded:
 		machineScope.V(2).Info("VM is running", "id", *machineScope.GetVMID())
+		conditions.MarkTrue(machineScope.AzureMachine, infrav1.VMRunningCondition)
 		machineScope.SetReady()
 	case infrav1.VMStateCreating:
 		machineScope.V(2).Info("VM is creating", "id", *machineScope.GetVMID())
+		conditions.MarkFalse(machineScope.AzureMachine, infrav1.VMRunningCondition, infrav1.VMNCreatingReason, clusterv1.ConditionSeverityInfo, "")
```
CecileRobertMichon, line 288: side note: right now we never hit these because we block on long running operations such as VM creation so that will need to change in order for conditions to really become useful

CecileRobertMichon: not feedback for this PR, more of a general comment

api/v1alpha3/conditions_consts.go
CecileRobertMichon: I think we should make these more granular as well (eg. PublicLoadBalancerReadyCondition, VirtualNetworkReadyCondition, etc.). I don't mind us doing this in a later PR though if you think it's easier to start with this for now @nader-ziada.

CecileRobertMichon:
Does this need the actual v0.3.7 release or do we have all we need in v0.3.7-alpha?

api/v1alpha3/azurecluster_types.go
nader-ziada: ok will move them

api/v1alpha3/conditions_consts.go
nader-ziada: I'll see if I can do that now

api/v1alpha3/conditions_consts.go
nader-ziada: I think they are meant to be distinct, at least for now while its a new change 

@fabriziopandini please correct me if I'm wrong

controllers/azurecluster_controller.go
nader-ziada: yes, should I rename it?

controllers/azuremachine_controller.go
nader-ziada: yes, you are correct, it will skip some of the conditions, but I thought we should add them here anyways so its there when we improve that

nader-ziada:
> Does this need the actual v0.3.7 release or do we have all we need in v0.3.7-alpha?

its in v0.3.7-alpha, but I think there is a related change to patch helper that came after alpha, but as far as I know doesn't stop it from working

api/v1alpha3/azurecluster_types.go
nader-ziada: if I move these to scope, I can't use them in the manual conversion which is required

CecileRobertMichon: gotcha, let's just leave them then. At one point we'll have to go through scope/api types and make sure all of these are where they should be and are unit tested...

api/v1alpha3/conditions_consts.go
CecileRobertMichon: I think LoadBalancerProvisioning makes more sense here since this is referring to the infrastructure status

api/v1alpha3/conditions_consts.go
nader-ziada: not sure how I can make these more granular since the code doesn't check for each step?

api/v1alpha3/conditions_consts.go
nader-ziada: thank you Cecile

nader-ziada: renamed to LoadBalancerProvisioning

nader-ziada:
/hold cancel

since I'm working on it and we have 0.3.7-alpha

api/v1alpha3/conditions_consts.go
fabriziopandini: I consider conditions are a tool to surface internal state, but then it up really up to you to decide how your internal state is managed and how many details you want to surface.
If you keep the two list separated, it can be verbose, but you get a certain degree of flexibility and decoupling

controllers/azurecluster_controller.go
fabriziopandini: FYI, the ready condition is already used for hosting the summary of all the conditions on an object

controllers/azurecluster_controller.go
```diff
@@ -157,7 +167,8 @@ func (r *AzureClusterReconciler) reconcileNormal(ctx context.Context, clusterSco
 	}
 
 	if azureCluster.Status.Network.APIServerIP.DNSName == "" {
-		clusterScope.Info("Waiting for API server endpoint to exist")
+		clusterScope.Info("Waiting for Load Balancer to exist")
```
CecileRobertMichon, line 170: so I looked into this a bit more, it's actually not accurate right now because the DNSName gets set by the azurecluster reconciler _before_ the load balancer gets created, so it not being empty doesn't indicate that the load balancer exists.

CecileRobertMichon: see https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/652#discussion_r447940198

nader-ziada: doesn't the create `createOrUpdateNetworkAPIServerIP` happen in the `Reconcile` a few lines before this? ` https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/controllers/azurecluster_controller.go#L154-L163

CecileRobertMichon: It does. But in that `Reconcile`, `createOrUpdateNetworkAPIServerIP` is the first func that gets called and `azureCluster.Status.Network.APIServerIP.DNSName` is set before the Load Balancer is created. So it could be that this gets set and later the Load Balancer create fails.

nader-ziada: I assumed if something fails the call to reconcile from the controller should exit with error and not get to that point in the code https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/controllers/azurecluster_controller.go#L156

CecileRobertMichon: Good point, I didn't think of that. In that case I think it's okay 👍

CecileRobertMichon:
lgtm

Will let @devigned review when he's back tomorrow

api/v1alpha3/azurecluster_types.go
devigned: We can always add a proxy func on the scope which sets the condition on the underlying resource.

PR #716, opened 2020-06-18
 💎 Refactor public IP service to get Spec from scope

CecileRobertMichon:
**What this PR does / why we need it**: Following up from #713, this is a proof of concept with the Public IPs service to stop using a mix of spec + scope in services as input for Reconcile() and Delete().

Before:
```
type OldService interface {
	Reconcile(ctx context.Context, spec interface{}) error
	Delete(ctx context.Context, spec interface{}) error
}
```
https://github.com/CecileRobertMichon/cluster-api-provider-azure/blob/06d0321be5fefd840ea1742b9c16a7ce72b98378/cloud/interfaces.go#L34-L37

After:
```
type Service interface {
}
```
https://github.com/CecileRobertMichon/cluster-api-provider-azure/blob/06d0321be5fefd840ea1742b9c16a7ce72b98378/cloud/interfaces.go#L27-L30

Instead of taking cluster scope or machine scope types as input, each service will have its own scope interface that explicitly defines what information getters the service needs that the passed in scope (either AzureCluster or AzureMachine) needs to implement: https://github.com/CecileRobertMichon/cluster-api-provider-azure/blob/06d0321be5fefd840ea1742b9c16a7ce72b98378/cloud/services/publicips/service.go#L22-L25

This only changes the Pubic IP service while keep all other services intact (renamed the interface to `OldService`). The idea is to facilitate review and focus on the concept rather than on each service's details. Once I get reviewer consensus, I will proceed to refactor all the other services by opening a PR for each one. This following PRs will be smaller as this one does most of the heavy lifting in machine scope and cluster scope.

/hold
/cc @devigned @alexeldeib @justaugustus @nader-ziada 
Would love to get your thoughts on this approach.

Partially addresses #110 , https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/627 and #609

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Refactor public IP service to get Spec from scope
Update mockgen to 8a3d595
```


cloud/interfaces.go
```diff
@@ -40,3 +49,18 @@ type CredentialGetter interface {
...
 }
+
+// Authorizer is an interface which can get the subscription ID and authorizer for an Azure service.
+type Authorizer interface {
+	SubscriptionID() string
+	Authorizer() autorest.Authorizer
+}
+
+// ClusterInformation is an interface which can get common Azure Cluster information
+type ClusterInformation interface {
```
CecileRobertMichon, line 60: TODO: think of better naming for this

cloud/interfaces.go
```diff
@@ -40,3 +49,18 @@ type CredentialGetter interface {
 	Service
 	GetCredentials(ctx context.Context, group string, cluster string) ([]byte, error)
 }
+
+// Authorizer is an interface which can get the subscription ID and authorizer for an Azure service.
+type Authorizer interface {
+	SubscriptionID() string
```
CecileRobertMichon, line 55: @nader-ziada FYI this is going to conflict with #649 but I can rebase on top of your PR once it lands.

cloud/interfaces.go
devigned: ClusterDescriber?

cloud/interfaces.go
nader-ziada: no worries, I can also rebase if this gets in first

controllers/azuremachine_reconciler.go
```diff
@@ -59,11 +61,17 @@ func newAzureMachineService(machineScope *scope.MachineScope, clusterScope *scop
 		networkInterfacesSvc: networkinterfaces.NewService(clusterScope, machineScope),
 		virtualMachinesSvc:   virtualmachines.NewService(clusterScope, machineScope),
 		disksSvc:             disks.NewService(clusterScope),
+		publicIPsSvc:         publicips.NewService(machineScope),
```
nader-ziada, line 64: isn't this input supposed to be PublicIPScope as input to NewService?

cloud/services/publicips/service.go
```diff
@@ -16,20 +16,30 @@ limitations under the License.
...
-	Scope *scope.ClusterScope
+	Scope PublicIPScope
 	Client
 }
 
 // NewService creates a new service.
-func NewService(scope *scope.ClusterScope) *Service {
+func NewService(scope PublicIPScope) *Service {
 	return &Service{
 		Scope:  scope,
```
nader-ziada, line 42: There are too many things called `Scope`, I'm getting confused keeping track which is which

cloud/scope/machine.go
```diff
@@ -40,65 +42,56 @@ type MachineScopeParams struct {
...
-	Cluster      *clusterv1.Cluster
+	ClusterScope azure.ClusterInformation
 	Machine      *clusterv1.Machine
-	AzureCluster *infrav1.AzureCluster
 	AzureMachine *infrav1.AzureMachine
 }
 
-// Location returns the AzureMachine location.
+// PublicIPSpec returns the public IP spec.
+func (m *MachineScope) PublicIPSpec() []publicips.Spec {
```
nader-ziada, line 61: not sure I understand why this is here

CecileRobertMichon:
@nader-ziada 

> Instead of taking cluster scope or machine scope types as input, each service will have its own scope interface that explicitly defines what information getters the service needs that the passed in scope (either AzureCluster or AzureMachine) needs to implement: https://github.com/CecileRobertMichon/cluster-api-provider-azure/blob/06d0321be5fefd840ea1742b9c16a7ce72b98378/cloud/services/publicips/service.go#L22-L25

Basically, `PublicIPScope` is a more precise, targetted scope interface, that defines what info the public IP service needs. Both ClusterScope and MachineScope (and maybe MachinePool in the future once it supports public IPs per instance) implement PublicIPScope. This is to allow reusing the same public IP reconcile service for different public IPs, eg. the Public LB public IP and each VM's IP when `allocatePublicIP` is set to true. This also makes sure that the passed in "scope" has all the required getter funcs implemented rather that making assumptions on the content of cluster scope or machine scope. Does that make sense?

nader-ziada:
@CecileRobertMichon makes sense, I get it now 👍

cloud/services/groups/service.go
```diff
@@ -30,6 +30,6 @@ type Service struct {
 func NewService(scope *scope.ClusterScope) *Service {
 	return &Service{
 		Scope:  scope,
-		Client: NewClient(scope.SubscriptionID, scope.Authorizer),
+		Client: NewClient(scope.SubscriptionID(), scope.Authorizer()),
```
devigned, line 33:
```
NewClient(scope.SubscriptionID(), scope.Authorizer())
``` 
is commonly used. Would it make sense to change refactor those methods into an interface and pass an instance of that interface, like 
```
NewClient(authorizer Authorizer)
```

nader-ziada:
sorry for all the conflicts :(

cloud/services/groups/service.go
CecileRobertMichon: I have to rebase this on top of #649 which will remove SubID and add settings, will see what I can with that after the rebase

cloud/interfaces.go
devigned: Just leaving this here to lend to the discussion: https://golang.org/doc/effective_go.html#interface-names.

CecileRobertMichon:
> sorry for all the conflicts :(

No worries!! I knew what I was getting into 😄

cloud/interfaces.go
CecileRobertMichon: > By convention, one-method interfaces are named by the method name plus an -er suffix or similar modification to construct an agent noun: Reader, Writer, Formatter, CloseNotifier etc.

Is there a convention for multiple-methods interfaces? 😕

cloud/services/groups/service.go
CecileRobertMichon: Done, let me know what you think

cloud/interfaces.go
devigned: ¯\_(ツ)_/¯

cloud/services/groups/service.go
devigned: Looks good.

Do you feel like the interface fits or should it be a struct?

nader-ziada: I like how you created a wrapper Authorizer to make the func signature simpler

cloud/interfaces.go
CecileRobertMichon: I went with ClusterDescriber for now to follow the -er suffix  thing

cloud/services/groups/service.go
CecileRobertMichon: I just pushed a new update @nader-ziada @devigned. Removed Settings() and replaced it with SubscriptionID() and BaseURI()

CecileRobertMichon: The interface fits I think because I pass in scope to NewClient (scope implements azure.Authorizer)

nader-ziada: yeah makes sense now that you have `Authroizer` as a type wrapping all the needed fields

CecileRobertMichon:
@alexeldeib would love your thoughts on this approach when you have a second. I think this achieves what we discussed in #627:

- makes sharing for e.g. machine/machinepool and cluster/managedcluster possible
- The services shouldn't care about what type they're reconciling against; just the Azure stuff involved.
- The dependencies on scope are explicit.

Makefile
```diff
@@ -166,8 +166,8 @@ $(GOLANGCI_LINT): ## Build golangci-lint from tools folder.
 $(KUSTOMIZE): ## Build kustomize from tools folder.
 	GOBIN=$(TOOLS_BIN_DIR) $(GO_INSTALL) sigs.k8s.io/kustomize/kustomize/v3@v3.5.4
 
-$(MOCKGEN): ## Build mockgen from tools folder.
-	GOBIN=$(TOOLS_BIN_DIR) $(GO_INSTALL) github.com/golang/mock/mockgen@v1.4.3
+$(MOCKGEN): ## Build mockgen from tools folder. Using unreleased version until https://github.com/golang/mock/pull/405 is part of a release.
```
CecileRobertMichon, line 169: will switch to upcoming release once it's out https://github.com/golang/mock/pull/405#issuecomment-648449558

CecileRobertMichon:
Alright, this is ready for review. I fixed up the tests and adapted the Client auth on top of #649. It's a big diff because I had to make a lot of changes in the client files for the auth but the next PRs to adapt this to other services should be a lot smaller.

nader-ziada:
I went through everything again and it all looks good to me, @devigned do you have any comments?

devigned:
@CecileRobertMichon and I have spoken quite a bit about this PR. I haven't commented much b/c I didn't want to bias the reviewers. With that said, I'm super excited to get this PR merged and I think it presents a clean pattern for future refactors. Great work!

/lgtm

CecileRobertMichon:
Going to keep the hold for lazy consensus until tomorrow, want to give @justaugustus and @alexeldeib a chance to check it out if they'd like.

Thanks for the reviews @devigned @nader-ziada

cloud/interfaces.go
alexeldeib: which one will end up in the final result 🙂 I would prefer to keep as struct instead of interface.

cloud/scope/cluster.go
```diff
@@ -83,11 +84,100 @@ type ClusterScope struct {
 	AzureCluster *infrav1.AzureCluster
 }
 
+// NewMachineScope creates a new MachineScope from the supplied parameters.
+// This is meant to be called for each reconcile iteration.
+func (s *ClusterScope) NewMachineScope(params MachineScopeParams) (*MachineScope, error) {
```
alexeldeib, line 89: nit: it makes me twitch that we have the constructor of the other object receiver like this instead of with the rest of machine scope? can we not pass cluster scope into  NewMachineScope()

cloud/interfaces.go
devigned: @alexeldeib can you expand on that a bit?

cloud/scope/machine.go
```diff
@@ -40,65 +40,61 @@ type MachineScopeParams struct {
...
+
+// Location returns the AzureCluster location.
 func (m *MachineScope) Location() string {
-	return m.AzureCluster.Spec.Location
+	return m.ClusterScope.Location()
+}
+
+// ResourceGroup returns the AzureCluster resource group.
+func (m *MachineScope) ResourceGroup() string {
+	return m.ClusterScope.ResourceGroup()
```
alexeldeib, line 77: should we embed the struct/interface?

cloud/scope/machinepool.go
```diff
@@ -56,56 +54,43 @@ type (
...
-	}
-	if params.AzureCluster == nil {
-		return nil, errors.New("azure cluster is required when creating a MachinePoolScope")
-	}
-	if params.AzureMachinePool == nil {
-		return nil, errors.New("azure machine pool is required when creating a MachinePoolScope")
-	}
+// Location returns the AzureCluster location.
+func (m *MachinePoolScope) Location() string {
+	return m.ClusterScope.Location()
```
alexeldeib, line 65: same q about embedding

cloud/scope/managedcontrolplane.go
```diff
@@ -94,6 +94,21 @@ type ManagedControlPlaneScope struct {
 	PatchTarget      runtime.Object
 }
 
+// SubscriptionID returns the Azure client Subscription ID.
+func (s *ManagedControlPlaneScope) SubscriptionID() string {
+	return s.AzureClients.SubscriptionID
```
alexeldeib, line 99: why does this need to be an interface?

alexeldeib: rather than retrieving struct data directly (ref comment at top)

cloud/scope/cluster.go
devigned: Definitely could pass in the cluster scope. It is effectively the same thing.

cloud/services/managedclusters/service.go
```diff
@@ -26,8 +26,8 @@ type Service struct {
 }
 
 // NewService creates a new service.
-func NewService(authorizer autorest.Authorizer, baseURI, subscriptionID string) *Service {
+func NewService(scope azure.Authorizer) *Service {
```
alexeldeib, line 29:
```suggestion
func NewService(auth azure.Authorizer) *Service {
```

or similar

cloud/services/agentpools/service.go
```diff
@@ -27,8 +26,8 @@ type Service struct {
 }
 
 // NewService creates a new service.
-func NewService(authorizer autorest.Authorizer, settings azure.ClientSettings) *Service {
+func NewService(scope azure.Authorizer) *Service {
```
alexeldeib, line 29:
```suggestion
func NewService(auth azure.Authorizer) *Service {
```

or similar

cloud/scope/cluster.go
CecileRobertMichon: yeah the two would be equivalent... I guess it felt cleaner to have `func (s *ClusterScope) NewMachineScope(params MachineScopeParams) (*MachineScope, error) {` than `func NewMachineScope(s *ClusterScope, params MachineScopeParams) (*MachineScope, error) {` and this was a way to be very explicit about the dependency of MachineScope on ClusterScope 

(same thing applies to MachinePool)

alexeldeib: add to param struct?

cloud/scope/machine.go
CecileRobertMichon: Main reason is that this allows re-use of Services between AzureCluster, AzureMachine, etc. If we embed the ClusterScope then we have to do `scope.ResourceGroup()` for one and `scope.ClusterScope.ResourceGroup()` for the other. See publicips.go for an example of how that works (both AzureCluster and AzureMachine call publicips `Reconcile()`).

We also want to be able to have two different implementations for some of these shared functions but have the same signature, eg. `PublicIPSpecs()` returns a different spec for AzureCluster than for AzureMachine.

CecileRobertMichon:
I realize this diff is pretty hard to read because of all the signature + auth changes... For those trying to understand the overall idea I recommend checking out https://github.com/CecileRobertMichon/cluster-api-provider-azure/compare/public-ips-scope...CecileRobertMichon:refactor-nic-scope, which is this refactor applied to network interfaces (rebased on this branch) and is a lot smaller and focused on the actual services changes.

cloud/services/agentpools/service.go
CecileRobertMichon: fixed

cloud/services/managedclusters/service.go
CecileRobertMichon: fixed

cloud/scope/cluster.go
CecileRobertMichon: yeah that's good feedback, i agree the constructor isn't really in the right place right now. Let me try that.

CecileRobertMichon: done

CecileRobertMichon:
@devigned @nader-ziada @alexeldeib any outstanding comments?

nader-ziada:
lgtm from me

devigned:
lgtm

alexeldeib:
/lgtm

definitely happy with inverting control and slimming down scope per service 👍

Issue #718, opened 2020-06-19
Does not currently handle creating seperate route tables for node and contorl plane


jsturtevant:
/kind bug

**What steps did you take and what happened:**
It is possible to set the the route table name on [control nodes](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/790ec1caa9fc79942655c6621505796f1094d137/api/v1alpha3/azurecluster_default.go#L76) and [nodes](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/790ec1caa9fc79942655c6621505796f1094d137/api/v1alpha3/azurecluster_default.go#L89) but only [the node route table ](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/790ec1caa9fc79942655c6621505796f1094d137/controllers/azurecluster_reconciler.go#L113) is created:

```golang
	rtSpec := &routetables.Spec{
		Name: r.scope.NodeSubnet().RouteTable.Name,
	}
	if err := r.routeTableSvc.Reconcile(ctx, rtSpec); err != nil {
		return errors.Wrapf(err, "failed to reconcile route table %s for cluster %s", r.scope.NodeSubnet().RouteTable.Name, r.scope.Name())
	}
```

If someone set both of these to something other than default they would run into issue on the control plane subnet reconciliation since the [subnet assumes route table is already created](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/790ec1caa9fc79942655c6621505796f1094d137/cloud/services/subnets/subnets.go#L106).

**What did you expect to happen:**
Should reconcile the both route tables if this is possible or provide appropriate error messages/defaulting if over loading the defaults

**Anything else you would like to add:**
I am not 100% sure it makes sense to have two different route tables.  Need to look into it more.

**Environment:**

- cluster-api-provider-azure version: main branch
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

CecileRobertMichon:
If we only want to allow one route table for now we could also add webhook validation that fails if route table is specified for a control plane subnet.

nader-ziada:
> I am not 100% sure it makes sense to have two different route tables. Need to look into it more.

@jsturtevant I kind of changes to address this, should I go ahead and submit a PR for this? you mentioned you wanted to check if its needed, thanks

jsturtevant:
Yes go for it. I have not had an opportunity to look into it

PR #719, opened 2020-06-19
Fix route table reconciliation

### LINKED ISSUE #682: When enabling route creation in the cloud provider routes can not be created
/kind bug

If you turn on `--configure-cloud-routes` in the [cloud provider](https://kubernetes.io/docs/reference/command-line-tools-reference/cloud-controller-manager/#options) for a CNI that requires route tables such as kubenet or when using [calico with UDRs](https://docs.projectcalico.org/reference/public-cloud/azure#azure-user-defined-routes) there will be 2 reconciliation loops running on the route table object in azure which for the cloud provider is causing the following issue when trying to update add the routes for each node:

```
E0604 17:11:05.255586       1 azure_routes.go:345] CreateRoute failed for node "capz-cluster-0-control-plane-q7mgs" with error: Retriable: false, RetryAfter: 0s, HTTPStatusCode: 412, RawError: Retriable: false, RetryAfter: 0s, HTTPStatusCode: 412, RawError: {
  "error": {
    "code": "PreconditionFailed",
    "message": "Precondition failed.",
    "details": [
      {
        "code": "PreconditionFailedEtagMismatch"
        "message": "Etag provided in if-match header W/\"8aea0627-0483e-ba38-ebd11e488\" does not match etag W/\"a96fd6fb-1bcb-4824-94306b551d5f5\" of resource /subscriptions/b9436a-0c7/resourceGroups/capz-cluster-0/providers/Microsoft.Network/routeTables/capz-cluster-0-node-routetable in NRP data store."
      }
    ]
  }
}
```
The work load cluster can update routes if required by the cni.

I tracked it down in cloud provider and turned off capz controller and things worked as expected on the cloud provider side.

NSG has this code in place to handle etags and should be able to mimic: 

https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/ff37e097b6cdad039e1df89d8887eb67436e8b32/cloud/services/securitygroups/securitygroups.go#L117

**Environment:**

- cluster-api-provider-azure version: k8s.gcr.io/kube-controller-manager:v1.18.2

jsturtevant:
When create the route table it pushes updates to the route table on every reconciliation which conflicts with the cloud controller manager.

Fixes #682 

Originally thought we might use etags, but since we don't set any other properties beyond the name of the route, once created we can avoid updating.  If in the future we want to be able to create specific routes via the spec we could update the spec to add routes and use etags to push the updates. I am not sure there is a use case (wasn't something you could do in aks-engine as far as I could tell) so will forgo for now. 

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
route table is created once and works with route creation turned on for controller manager
```

### DIFF
--- a/cloud/services/routetables/routetables.go
+++ b/cloud/services/routetables/routetables.go
@@ -18,7 +18,6 @@ package routetables
 
 import (
 	"context"
-
 	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2019-06-01/network"
 	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/pkg/errors"
@@ -41,8 +40,27 @@ func (s *Service) Reconcile(ctx context.Context, spec interface{}) error {
 	if !ok {
 		return errors.New("invalid Route Table Specification")
 	}
+
+	existingRouteTable, err := s.Get(ctx, s.Scope.ResourceGroup(), routeTableSpec.Name)
+	if !azure.ResourceNotFound(err) {
+		if err != nil {
+			return errors.Wrapf(err, "failed to get route table %s in %s", routeTableSpec.Name, s.Scope.ResourceGroup())
+		}
+
+		// route table already exists
+		// currently don't support:
+		//  1. creating separate control plane and node (#718) so update both
+		//  2. can not specify your own routes via spec
+		s.Scope.NodeSubnet().RouteTable.Name = to.String(existingRouteTable.Name)
+		s.Scope.NodeSubnet().RouteTable.ID = to.String(existingRouteTable.ID)
+		s.Scope.ControlPlaneSubnet().RouteTable.Name = to.String(existingRouteTable.Name)
+		s.Scope.ControlPlaneSubnet().RouteTable.ID = to.String(existingRouteTable.ID)
+
+		return nil
+	}
+
 	klog.V(2).Infof("creating route table %s", routeTableSpec.Name)
-	err := s.Client.CreateOrUpdate(
+	err = s.Client.CreateOrUpdate(
 		ctx,
 		s.Scope.ResourceGroup(),
 		routeTableSpec.Name,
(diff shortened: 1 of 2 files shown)


cloud/services/routetables/routetables_test.go
```diff
@@ -115,7 +116,8 @@ func TestReconcileRouteTables(t *testing.T) {
 			},
 			expectedError: "",
 			expect: func(m *mock_routetables.MockClientMockRecorder) {
-				m.CreateOrUpdate(context.TODO(), "my-rg", "my-routetable", gomock.AssignableToTypeOf(network.RouteTable{}))
```
jsturtevant, line 118: the test was missing the `defer mockCtrl.Finish()` (added below) so this was not being checked. The expectation for this test should be that the route table createorupdate is *not* called.  Also fixed this for the rest of the tests in this file and added explicit checks for number of times functions are expected to be called.

cloud/services/routetables/routetables.go
```diff
@@ -41,8 +40,27 @@ func (s *Service) Reconcile(ctx context.Context, spec interface{}) error {
...
+
+	existingRouteTable, err := s.Get(ctx, s.Scope.ResourceGroup(), routeTableSpec.Name)
+	if !azure.ResourceNotFound(err) {
+		if err != nil {
+			return errors.Wrapf(err, "failed to get route table %s in %s", routeTableSpec.Name, s.Scope.ResourceGroup())
+		}
+
+		// route table already exists
+		// currently don't support:
+		//  1. creating separate control plane and node (#718) so update both
```
jsturtevant, line 52: created #718 for this issue

cloud/services/routetables/routetables_test.go
jsturtevant: created #720 to look into other tests where this might be the case

devigned: Great catch!

cloud/services/routetables/routetables.go
```diff
@@ -41,8 +40,27 @@ func (s *Service) Reconcile(ctx context.Context, spec interface{}) error {
...
+		//  2. can not specify your own routes via spec
```
CecileRobertMichon, line 53: nit: this is a double negation
```suggestion
		//  2. specifying your own routes via spec
```

PR #724, opened 2020-06-22
fix internallb subnet addressing

### LINKED ISSUE #723: internal load-balancer address selection broken when using subnets longer than 8 chars
/kind bug

Creating a cluster using cap-azure with the non default subnetCIDR for the controlplane, e.g., 10.64.8.0 results in an error when creating the internal load-balancer that prevents the cluster from coming up. The error is that the internal load-balancer address is not on the controlplane subnet.

The default value selected for the internal load-balancer will be on the controlplane subnet.

The string slicing assumes an address matching the default which is only 8 characters in total length. The line causing the issue is https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/cloud/services/internalloadbalancers/internalloadbalancers.go#L174

**Environment:**

- cluster-api-provider-azure version: 0.4.3, but still present on main branch
- Kubernetes version: (use `kubectl version`): 1.17.4
- OS (e.g. from `/etc/os-release`): Ubuntu 18.04.4 LTS

dklyle:
This fixes assumptions in the controlplane cidr when selecting the address for the internal load balancer.

Fixes #723

**Release Note:**

```release-note
Default internal loadbalancer address selection broken for CIDR not 8 characters in length.
```

### DIFF
--- a/cloud/services/internalloadbalancers/internalloadbalancers.go
+++ b/cloud/services/internalloadbalancers/internalloadbalancers.go
@@ -19,6 +19,7 @@ package internalloadbalancers
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2019-06-01/network"
 	"github.com/Azure/go-autorest/autorest/to"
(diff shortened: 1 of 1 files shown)


cloud/services/internalloadbalancers/internalloadbalancers.go
```diff
@@ -171,7 +172,8 @@ func (s *Service) getAvailablePrivateIP(ctx context.Context, resourceGroup, vnet
 		ip = azure.DefaultInternalLBIPAddress
 		if subnetCIDR != infrav1.DefaultControlPlaneSubnetCIDR {
 			// If the user provided a custom subnet CIDR without providing a private IP, try finding an available IP in the subnet space
-			ip = subnetCIDR[0:7] + "0"
+			index := strings.LastIndex(subnetCIDR, ".")
+			ip = subnetCIDR[0:(index+1)] + "0"
```
CecileRobertMichon, line 176: Great catch @dklyle! Could you please add a unit test to validate the fix (ie. that fails without the fix and passes with the fix)?

CecileRobertMichon:
@dklyle let me know if you don't have bandwidth to work on this, happy to help push it over the finish line with some tests.

dklyle:
@CecileRobertMichon, I have not had time to get to the unit tests. Thanks for merging.

PR #729, opened 2020-06-23
:bug: respect ingressRules specified in security group spec

rsmitty:
This PR will fix a bug in which ingress rules were getting ignored if
they were specified as part of the security group spec. This will now
work:

```
    subnets:
      - name: talos-e2e-{{TAG}}-azure-controlplane-subnet
        role: control-plane
        cidrBlock: 10.0.1.0/24
        securityGroup:
          name: talos-e2e-{{TAG}}-azure-controlplane-sg
          ingressRule:
            - name: "allow_port_50000"
              description: "allow port 50000"
              priority: 103
              protocol: "*"
              destination: "*"
              destinationPorts: "50000"
              source: "*"
              sourcePorts: "*"
```

Signed-off-by: Spencer Smith <robertspencersmith@gmail.com>

Fixes #

**TODOs**:

- [X] squashed commits
- [X] includes documentation
- [X] adds unit tests

**Release note**:
```release-note
Respect IngressRules When Provided in SecurityGroup Spec
```


api/v1alpha2/types.go
```diff
@@ -181,9 +181,12 @@ var (
 
 // IngressRule defines an Azure ingress rule for security groups.
 type IngressRule struct {
-	Description string                `json:"description"`
+	Name        *string               `json:"name"`
+	Description *string               `json:"description"`
 	Protocol    SecurityGroupProtocol `json:"protocol"`
 
+	Priority *int32 `json:"priority,omitempty"`
```
CecileRobertMichon, line 188: let's keep this change scoped to v1alpha3, I don't think we need to backport the fix to v1alpha2

cloud/services/securitygroups/securitygroups.go
```diff
@@ -134,20 +175,31 @@ func ruleExists(rules []network.SecurityRule, rule network.SecurityRule) bool {
 	return false
 }
 
-func getRule(name, destinationPort string, priority int32) network.SecurityRule {
-	return network.SecurityRule{
-		Name: to.StringPtr(name),
+func getRule(ingress infrav1.IngressRule) network.SecurityRule {
```
devigned, line 178: Thoughts on changing the name from `getRule` to `newIngressSecurityRule`?

docs/topics/custom-vnet.md
```diff
@@ -81,4 +81,45 @@ spec:
 
 If no CIDR block is provided, `10.0.0.0/8` will be used by default, with default internal LB private IP `10.0.0.100`.
 
+### Ingress Rules
+
+Ingress rules can also be added as part of the subnet specification in a custom network spec.
+Note that ingress rules for the Kubernetes API Server port (default 6443) and SSH (22) are automatically added to the controlplane subnet.
```
CecileRobertMichon, line 87: Should we be explicit about these being "additional ingress rules"? These don't replace the defaults, they are appended to the defaults. 

On the other hand, would it be valuable to instead allow the user to specify _all_ rules? I could see use cases for someone to want to disable SSH, so would it makes sense to instead of appending the rules have the webhook default the "default" rules if no rules are specified, but keep the user defined rules as-is if they are provided.

Your experience as a user of CAPZ is valuable here @rsmitty. As a user, would you prefer having full control over what rules are added (but have the burden of defining the right rules for API server and risk misconfiguring the rules which would result in a nonfunctional cluster), or have CAPZ implement the right required rules for you and only be able to add additional rules specific to your use case (but not remove rules).

Additional thoughts:
- If we do allow for "additional rules", could be good to either 1) make that explicit in the spec naming/description or 2) have the defaults merged into the list of rules in the AzureCluster webhook.
- Right now I'm leaning for giving the user full control with the assumption that if the user decides to customize rules they should know what they are doing, with pitfalls clearly documented. If the user just wants CAPZ to work they can always not define rules.

api/v1alpha3/types.go
```diff
@@ -118,9 +118,12 @@ const (
 
 // IngressRule defines an Azure ingress rule for security groups.
 type IngressRule struct {
+	Name        string                `json:"name"`
 	Description string                `json:"description"`
 	Protocol    SecurityGroupProtocol `json:"protocol"`
 
+	Priority int32 `json:"priority,omitempty"`
```
CecileRobertMichon, line 125: let's add a comment/description of priority + some validation. According to Azure docs it should be:

```
A number between 100 and 4096. Rules are processed in priority order, with lower numbers processed before higher numbers, because lower numbers have higher priority. Once traffic matches a rule, processing stops. As a result, any rules that exist with lower priorities (higher numbers) that have the same attributes as rules with higher priorities are not processed.
```

Each rule should also have a unique priority.

api/v1alpha2/azuremachine_conversion.go
```diff
@@ -173,6 +173,15 @@ func Convert_v1alpha3_Image_To_v1alpha2_Image(in *infrav1alpha3.Image, out *Imag
 	return nil
 }
 
+// Convert_v1alpha3_IngressRule_To_v1alpha2_IngressRule converts IngressRule from v1alpha3 to v1alpha2
+func Convert_v1alpha3_IngressRule_To_v1alpha2_IngressRule(in *infrav1alpha3.IngressRule, out *IngressRule, s apiconversion.Scope) error { //nolint
```
CecileRobertMichon, line 177: you also need to add conversion for the newly added fields. See https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/729/files#diff-3962b452782abe8db61ec4e1f455e6b0R49 for an example.

docs/topics/custom-vnet.md
rsmitty: Yeah, I think what you mention makes sense. The defaults should be "batteries included" with easy ways to get to the apiserver and ssh b/c they are set by default. If someone wants to do anything different, they can easily specify the same defaults along with their additional ingresses if need be.

cloud/services/securitygroups/securitygroups.go
rsmitty: updated this bit in an amended commit.

api/v1alpha3/types.go
rsmitty: Fixed this bit in an amended commit.

devigned:
Well, this is interesting. I'll enumerate what I did to reproduce.
```shell
$ git fetch upstream pull/729/head:rsmitty
remote: Enumerating objects: 19, done.
remote: Counting objects: 100% (19/19), done.
remote: Compressing objects: 100% (2/2), done.
remote: Total 19 (delta 17), reused 19 (delta 17), pack-reused 0
Unpacking objects: 100% (19/19), 3.65 KiB | 93.00 KiB/s, done.
From github.com:kubernetes-sigs/cluster-api-provider-azure
 * [new ref]           refs/pull/729/head -> rsmitty

$ git co rsmitty
Switched to branch 'rsmitty'

$ make verify
...
./hack/gen-flavors.sh
generated files are out of date, run make generate
make: *** [verify-gen] Error 1

$ git status
 M config/crd/bases/exp.infrastructure.cluster.x-k8s.io_azuremachinepools.yaml
 M config/crd/bases/exp.infrastructure.cluster.x-k8s.io_azuremanagedclusters.yaml
 M config/crd/bases/exp.infrastructure.cluster.x-k8s.io_azuremanagedcontrolplanes.yaml
 M config/crd/bases/exp.infrastructure.cluster.x-k8s.io_azuremanagedmachinepools.yaml
 M config/crd/bases/infrastructure.cluster.x-k8s.io_azureclusters.yaml
 M config/crd/bases/infrastructure.cluster.x-k8s.io_azuremachines.yaml
 M config/crd/bases/infrastructure.cluster.x-k8s.io_azuremachinetemplates.yaml

$ git diff config/crd/bases/infrastructure.cluster.x-k8s.io_azuremachinetemplates.yaml
diff --git a/config/crd/bases/infrastructure.cluster.x-k8s.io_azuremachinetemplates.yaml b/config/crd/bases/infrastructure.cluster.x-k8s.io_azuremachinetemplates.yaml
index 82f7047c..cb613b6c 100644
--- a/config/crd/bases/infrastructure.cluster.x-k8s.io_azuremachinetemplates.yaml
+++ b/config/crd/bases/infrastructure.cluster.x-k8s.io_azuremachinetemplates.yaml
@@ -4,7 +4,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.2.9
+    controller-gen.kubebuilder.io/version: v0.3.0
   creationTimestamp: null
   name: azuremachinetemplates.infrastructure.cluster.x-k8s.io
 spec:
```

I feel like you might need to fetch the latest from upstream master and rebase upon it. The weird part is that I see changes in your branch (locally) which are not showing up in the PR.

rsmitty:
@devigned wonder if that was an effect of me force-pushing as you checked it out or something. I'm actively pushing to my master branch so who knows what state it was in. Just rebased and `make verify` ran successfully locally. We'll see what happens in CI here in a sec.

rsmitty:
@devigned aaaand no difference. verify is still throwing that "generated files are out of date, run make generate" error in the CI environment.

devigned:
@rsmitty just pulled down your code again, wiped out my `./hack/tools/bin`, and everything ran fine. Something must be up on the CI. I'm going to look closer at the logs. This is weird.

rsmitty:
Glad you can reproduce at least. I was feeling like I was taking crazy pills! :joy:

devigned:
@rsmitty to see what is changing, change the line to not be `--quiet` https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/04b9e8210b088eba82afc8d4967c3604e1fcc10b/Makefile#L480

api/v1alpha2/azurecluster_conversion.go
```diff
@@ -63,6 +63,18 @@ func (src *AzureCluster) ConvertTo(dstRaw conversion.Hub) error { // nolint
 			for _, dstSubnet := range dst.Spec.NetworkSpec.Subnets {
 				if dstSubnet != nil && dstSubnet.Name == restoredSubnet.Name {
 					dstSubnet.RouteTable = restoredSubnet.RouteTable
+					dstSubnet.SecurityGroup = restoredSubnet.SecurityGroup
```
CecileRobertMichon, line 66: security group isn't a new field in v1alpha3 so it shouldn't be restored (only IngressRules[i].priority and IngressRules[i].name like you did below)

api/v1alpha2/azurecluster_conversion.go
```diff
@@ -63,6 +66,17 @@ func (src *AzureCluster) ConvertTo(dstRaw conversion.Hub) error { // nolint
 			for _, dstSubnet := range dst.Spec.NetworkSpec.Subnets {
 				if dstSubnet != nil && dstSubnet.Name == restoredSubnet.Name {
 					dstSubnet.RouteTable = restoredSubnet.RouteTable
+
+					for _, restoredIngress := range restoredSubnet.SecurityGroup.IngressRules {
+						if restoredIngress != nil {
+							for _, dstIngress := range dstSubnet.SecurityGroup.IngressRules {
+								if dstIngress != nil && dstIngress.Description == restoredIngress.Description {
```
CecileRobertMichon, line 73: If Description isn't a unique field it's won't be reliable to use it to match restored ingress rules with existing ones since there might be two rules with the same description

CecileRobertMichon: Subnet name above is unique (and that's enforced both by Azure and by CAPZ in validating webhooks)

rsmitty:
Will rebase this PR once #735 is merged, so as to avoid having to implement conversion for the status fields.

docs/topics/custom-vnet.md
```diff
@@ -82,3 +82,61 @@ spec:
 If no CIDR block is provided, `10.0.0.0/8` will be used by default, with default internal LB private IP `10.0.0.100`.
 
 Whenever using custom vnet and subnet names and/or a different vnet resource group, please make sure to update the `azure.json` content part of both the nodes and control planes' `kubeadmConfigSpec` accordingly before creating the cluster.
+
+### Custom Ingress Rules
+
+Ingress rules can also be customized as part of the subnet specification in a custom network spec.
+Note that ingress rules for the Kubernetes API Server port (default 6443) and SSH (22) are automatically added to the controlplane subnet only if Ingress Rules aren't specified.
+It is the responsibility of the user to supply those rules themselves if using custom ingresses
```
CecileRobertMichon, line 90:
```suggestion
It is the responsibility of the user to supply those rules themselves if using custom ingresses.
```

api/v1alpha3/azurecluster_default.go
```diff
@@ -119,3 +124,30 @@ func generateNodeSecurityGroupName(clusterName string) string {
...
+		},
+		&IngressRule{
+			Name:             "allow_apiserver",
+			Description:      "Allow K8s API Server",
+			Priority:         101,
+			Protocol:         SecurityGroupProtocolTCP,
+			Source:           to.StringPtr("*"),
+			SourcePorts:      to.StringPtr("*"),
+			Destination:      to.StringPtr("*"),
+			DestinationPorts: to.StringPtr("6443"),
```
CecileRobertMichon, line 150: for this you'll have to determine the port based on the Cluster object as discussed on slack

api/v1alpha3/azurecluster_validation_test.go
```diff
@@ -485,6 +485,90 @@ func TestInternalLBIPAddressInvalid(t *testing.T) {
 	g.Expect(err.BadValue).To(BeEquivalentTo(internalLBIPAddress))
 }
 
+func TestIngressRuleValidPriority(t *testing.T) {
```
nader-ziada, line 488: just a style comment, I think these 3 new could be changed to a table test

api/v1alpha3/azurecluster_default.go
nader-ziada: you still going to remove the hard coding port here, right?

nader-ziada:
the tests passed for me locally

/test pull-cluster-api-provider-azure-test

api/v1alpha3/azurecluster_default.go
rsmitty: Yes, still planning on fixing that bit

api/v1alpha3/azurecluster_validation_test.go
rsmitty: fixed this in an amended commit.

api/v1alpha2/azurecluster_conversion.go
```diff
@@ -63,6 +65,10 @@ func (src *AzureCluster) ConvertTo(dstRaw conversion.Hub) error { // nolint
 			for _, dstSubnet := range dst.Spec.NetworkSpec.Subnets {
 				if dstSubnet != nil && dstSubnet.Name == restoredSubnet.Name {
 					dstSubnet.RouteTable = restoredSubnet.RouteTable
+
+					if restoredSubnet.SecurityGroup.IngressRules != nil {
+						dstSubnet.SecurityGroup.IngressRules = restoredSubnet.SecurityGroup.IngressRules
```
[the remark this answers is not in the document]
rsmitty: shouldn't I don't think. fixed in amended commit.

docs/topics/custom-vnet.md
```diff
@@ -82,3 +82,60 @@ spec:
...
+              destination: "*"
+              destinationPorts: "22"
+              source: "*"
+              sourcePorts: "*"
+            - name: "allow_apiserver"
+              description: "Allow K8s API Server"
+              priority: 101
+              protocol: "*"
+              destination: "*"
+              destinationPorts: "50000"
```
CecileRobertMichon, line 126: is this a copy paste typo? maybe should be 6443 in the example so people can copy paste it?

api/v1alpha2/azurecluster_conversion.go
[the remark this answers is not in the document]
CecileRobertMichon: should this check for dstSubnet.SecurityGroup != nil though?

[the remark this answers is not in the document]
rsmitty: Security group isn't a pointer, so best we could do would be to check if something like SecurityGroup.Name is empty string. But if we're doing that, may as well check if SecurityGroup.IngressRules is nil I suppose.

docs/topics/custom-vnet.md
rsmitty: yes, thank you. fixed in amended commit.

api/v1alpha2/azurecluster_conversion.go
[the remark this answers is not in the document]
CecileRobertMichon: oh nevermind then! I assumed it was

CecileRobertMichon:
/hold

@rsmitty please add a release note

rsmitty:
@CecileRobertMichon added!

nader-ziada:
thanks @rsmitty 👍, went through the PR again and looks like all the comments are addressed 

/lgtm

PR #735, opened 2020-06-25
⚠️ Remove SecurityGroups from AzureCluster Status

CecileRobertMichon:
**What this PR does / why we need it**: As discovered in https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/729, AzureCluster.Status.SecurityGroups is not used (and was never used). Removing it to avoid having to do conversion in #729 now that Priority and Name were added to the IngressRule type and used in AzureCluster.Spec.NetworkSpec.Subnets.

Marking this as ⚠️ because it is removing a field from AzureCluster. However that field was never used and is part of Status so it should never be set.

cc @rsmitty 

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Remove SecurityGroups from AzureCluster Status
```


CecileRobertMichon:
Before this change, this is what status looks like on an AzureCluster (notice that SecurityGroups is not set):
```
status:
    network:
      apiServerIp:
        dnsName: default-template-7dd41e16.southcentralus.cloudapp.azure.com
        name: default-template-7dd41e16
    ready: true
```

nader-ziada:
/lgtm

should be no harm in removing from Status

PR #736, opened 2020-06-26
🔧 install hack/tools when version has changed

CecileRobertMichon:
**What this PR does / why we need it**: Currently the Makefile only installs hack/tools if they are not already installed. This causes new versions to not get installed if a previous version is already installed. This checks for existing installs with the version in the filename so that the tool will be installed if the version isn't installed. It also creates a symlink to the previous location (without the version) so that the tool can still be found in the same path.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
🔧 install hack/tools when version has changed 
```


Makefile
```diff
@@ -46,19 +46,45 @@ ifneq ($(abspath $(ROOT_DIR)),$(GOPATH)/src/sigs.k8s.io/cluster-api-provider-azu
...
+
+MOCKGEN_VER := v1.4.3
+MOCKGEN_BIN := mockgen
+MOCKGEN := $(TOOLS_BIN_DIR)/$(MOCKGEN_BIN)-$(MOCKGEN_VER)
+
+RELEASE_NOTES_VER := v0.3.4
+RELEASE_NOTES_BIN := release-notes
+RELEASE_NOTES := $(TOOLS_BIN_DIR)/$(RELEASE_NOTES_BIN)-$(RELEASE_NOTES_VER)
+
+GO_APIDIFF_VER := latest
```
CecileRobertMichon, line 77: using latest since there is no available release

Makefile
```diff
@@ -154,36 +180,32 @@ manager: ## Build manager binary.
...
 
 $(MOCKGEN): ## Build mockgen from tools folder.
-	GOBIN=$(TOOLS_BIN_DIR) $(GO_INSTALL) github.com/golang/mock/mockgen@v1.4.3
+	GOBIN=$(TOOLS_BIN_DIR) $(GO_INSTALL) github.com/golang/mock/mockgen $(MOCKGEN_BIN) $(MOCKGEN_VER)
 
-$(RELEASE_NOTES): ## Build release notes
-	GOBIN=$(TOOLS_BIN_DIR) $(GO_INSTALL) sigs.k8s.io/cluster-api/hack/tools/release
-	mv $(TOOLS_BIN_DIR)/release $(RELEASE_NOTES)
+$(RELEASE_NOTES): ## Build release notes.
+	GOBIN=$(TOOLS_BIN_DIR) $(GO_INSTALL) k8s.io/release/cmd/release-notes $(RELEASE_NOTES_BIN) $(RELEASE_NOTES_VER)
```
CecileRobertMichon, line 202: changed this intentionally to the tool we actually use

Makefile
```diff
@@ -154,36 +180,32 @@ manager: ## Build manager binary.
 ## Tooling Binaries
 ## --------------------------------------
 
-$(CLUSTERCTL): ## Build clusterctl binary.
```
CecileRobertMichon, line 157: removed this, wasn't being used

PR #740, opened 2020-06-29
💚 add e2e spec for workload service and LBs

### LINKED ISSUE #702: [e2e] cluster should be able to launch a long running HTTP server and svc endpoint
An e2e test that would validate the workload cluster can run a HTTP service and expose an endpoint for it 

This test in aks-engine can provide some inspiration:  https://github.com/Azure/aks-engine/blob/master/test/e2e/kubernetes/kubernetes_test.go#L826

### LINKED ISSUE #703: [e2e] cluster should have working load balancer
An e2e test to validate a workload cluster has a working load balancer as expected 

another test from aks-engine https://github.com/Azure/aks-engine/blob/master/test/e2e/kubernetes/kubernetes_test.go#L1408

mboersma:

Adds a spec to e2e tests that provisions a deployment and internal and external load balancers, then tests that they are wired up correctly before deleting them.

Fixes #703
Fixes #702

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
💚 add e2e spec for workload service and LBs
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,6 +14,7 @@ require (
 	github.com/google/go-cmp v0.5.0
 	github.com/google/gofuzz v1.1.0
 	github.com/google/uuid v1.1.1
+	github.com/hashicorp/go-retryablehttp v0.6.6
 	github.com/onsi/ginkgo v1.14.0
 	github.com/onsi/gomega v1.10.1
 	github.com/pkg/errors v0.9.1
--- a/test/e2e/azure_lb.go
+++ b/test/e2e/azure_lb.go
@@ -1,0 +1,294 @@
+// +build e2e
+
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package e2e
+
+import (
+	"context"
+	"fmt"
+	"io/ioutil"
+	"net"
+	"regexp"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	retryablehttp "github.com/hashicorp/go-retryablehttp"
+	appsv1 "k8s.io/api/apps/v1"
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"sigs.k8s.io/cluster-api/test/framework"
+)
+
+// AzureLBSpecInput is the input for AzureLBSpec.
+type AzureLBSpecInput struct {
+	BootstrapClusterProxy framework.ClusterProxy
+	Namespace             *corev1.Namespace
+	ClusterName           string
+	SkipCleanup           bool
+}
+
+// AzureLBSpec implements a test that verifies Azure internal and external load balancers can
+// be created and work properly through a Kubernetes service.
+func AzureLBSpec(ctx context.Context, inputGetter func() AzureLBSpecInput) {
+	var (
+		specName     = "azure-lb"
+		input        AzureLBSpecInput
+		clusterProxy framework.ClusterProxy
+		clientset    *kubernetes.Clientset
+	)
+
+	input = inputGetter()
+	Expect(input.BootstrapClusterProxy).NotTo(BeNil(), "Invalid argument. input.BootstrapClusterProxy can't be nil when calling %s spec", specName)
+	Expect(input.Namespace).NotTo(BeNil(), "Invalid argument. input.Namespace can't be nil when calling %s spec", specName)
+	By("creating a Kubernetes client to the workload cluster")
+	clusterProxy = input.BootstrapClusterProxy.GetWorkloadCluster(context.TODO(), input.Namespace.Name, input.ClusterName)
+	Expect(clusterProxy).NotTo(BeNil())
+	clientset = clusterProxy.GetClientSet()
+	Expect(clientset).NotTo(BeNil())
+
+	By("creating an nginx deployment")
+	deploymentsClient := clientset.AppsV1().Deployments(corev1.NamespaceDefault)
+	var replicas int32 = 1
+	deployment := &appsv1.Deployment{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "ingress-nginx",
+			Namespace: corev1.NamespaceDefault,
+		},
+		Spec: appsv1.DeploymentSpec{
+			Replicas: &replicas,
+			Selector: &metav1.LabelSelector{
+				MatchLabels: map[string]string{
+					"app": "ingress-nginx",
+				},
+			},
+			Template: corev1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: map[string]string{
+						"app": "ingress-nginx",
+					},
+				},
+				Spec: corev1.PodSpec{
+					Containers: []corev1.Container{
+						{
+							Name:  "web",
+							Image: "nginx:1.18",
+							Ports: []corev1.ContainerPort{
+								{
+									Name:          "http",
+									Protocol:      corev1.ProtocolTCP,
+									ContainerPort: 80,
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+	_, err := deploymentsClient.Create(deployment)
+	Expect(err).NotTo(HaveOccurred())
+	deployInput := framework.WaitForDeploymentsAvailableInput{
+		Getter:     deploymentsClientAdapter{client: deploymentsClient},
+		Deployment: deployment,
+	}
+	framework.WaitForDeploymentsAvailable(context.TODO(), deployInput, e2eConfig.GetIntervals(specName, "wait-deployment")...)
+
+	By("creating an internal Load Balancer service")
+	servicesClient := clientset.CoreV1().Services(corev1.NamespaceDefault)
+	ilbService := &corev1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "ingress-nginx-ilb",
+			Namespace: corev1.NamespaceDefault,
+			Annotations: map[string]string{
+				"service.beta.kubernetes.io/azure-load-balancer-internal": "true",
+			},
+		},
+		Spec: corev1.ServiceSpec{
+			Type: corev1.ServiceTypeLoadBalancer,
+			Ports: []corev1.ServicePort{
+				{
+					Name:     "http",
+					Port:     80,
+					Protocol: corev1.ProtocolTCP,
+				},
+				{
+					Name:     "https",
+					Port:     443,
+					Protocol: corev1.ProtocolTCP,
+				},
+			},
+			Selector: map[string]string{
+				"app": "ingress-nginx",
+			},
+		},
+	}
+	_, err = servicesClient.Create(ilbService)
+	Expect(err).NotTo(HaveOccurred())
+	ilbSvcInput := WaitForServiceAvailableInput{
+		Getter:  servicesClientAdapter{client: servicesClient},
+		Service: ilbService,
+	}
+	WaitForServiceAvailable(context.TODO(), ilbSvcInput, e2eConfig.GetIntervals(specName, "wait-service")...)
+
+	By("connecting to the internal LB service from a curl pod")
+	jobsClient := clientset.BatchV1().Jobs(corev1.NamespaceDefault)
+	svc, err := servicesClient.Get("ingress-nginx-ilb", metav1.GetOptions{})
+	Expect(err).NotTo(HaveOccurred())
+	var ilbIP string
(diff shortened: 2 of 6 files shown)


test/e2e/azure_test.go
```diff
@@ -112,6 +122,272 @@ var _ = Describe("Workoad cluster creation", func() {
...
+			Expect(err).NotTo(HaveOccurred())
+			err = tmpFile.Close()
+			Expect(err).NotTo(HaveOccurred())
+
+			By("creating a Kubernetes client to the workload cluster")
+			config, err := clientcmd.BuildConfigFromFlags("", tmpFile.Name())
+			Expect(err).NotTo(HaveOccurred())
+			clientset, err := kubernetes.NewForConfig(config)
+			Expect(err).NotTo(HaveOccurred())
+
```
nader-ziada, line 150: can't we use the BootstrapClusterProxy to get access to the workload cluster without have to do all the kubeconfig steps here, there is an example in cluster-api repo https://github.com/kubernetes-sigs/cluster-api/blob/baedbb83b160ee134a065d3d9d16094e23d4e1a3/test/e2e/self_hosted.go#L114

nader-ziada: @cpanato does that apply to networking PR as well?

test/e2e/azure_test.go
```diff
@@ -112,6 +120,251 @@ var _ = Describe("Workoad cluster creation", func() {
...
+					Namespace: corev1.NamespaceDefault,
+				},
+				Spec: batchv1.JobSpec{
+					Template: corev1.PodTemplateSpec{
+						Spec: corev1.PodSpec{
+							RestartPolicy: corev1.RestartPolicyNever,
+							Containers: []corev1.Container{
+								{
+									Name:  "curl",
+									Image: "byrnedo/alpine-curl",
```
nader-ziada, line 235: Are we okay to use random images from dockerhub?

test/e2e/azure_test.go
```diff
@@ -112,6 +120,251 @@ var _ = Describe("Workoad cluster creation", func() {
...
+									Command: []string{
+										"curl",
+										ilbIP,
+									},
+								},
+							},
+							Tolerations: []corev1.Toleration{
+								{
+									Key:      "node-role.kubernetes.io/master",
+									Operator: corev1.TolerationOpEqual,
```
nader-ziada, line 245: do we need this? doesn't this cluster have nodes

test/e2e/helpers.go
```diff
@@ -0,0 +1,138 @@
...
+	return err
+}
+
+type servicesClientAdapter struct {
+	client typedcorev1.ServiceInterface
+}
+
+func (c servicesClientAdapter) Get(ctx context.Context, key client.ObjectKey, obj runtime.Object) error {
+	// TODO: honor the ctx
+	service, err := c.client.Get(key.Name, metav1.GetOptions{})
```
nader-ziada, line 59: can we handle the TODO, or remove the comment if its not needed

test/e2e/helpers.go
```diff
@@ -0,0 +1,138 @@
...
+
+	}, intervals...).Should(BeTrue(), "Service %s/%s failed to get an IP for LoadBalancer.Ingress", input.Service.GetNamespace(), input.Service.GetName())
+}
+
+type jobsClientAdapter struct {
+	client typedbatchv1.JobInterface
+}
+
+func (c jobsClientAdapter) Get(ctx context.Context, key client.ObjectKey, obj runtime.Object) error {
+	// TODO: honor the ctx
```
nader-ziada, line 102: can we handle the TODO, or remove the comment if its not needed

test/e2e/azure_test.go
cpanato: in the net pol we are using the `kubectl` command line. That's why I asked in the meeting, should we support both `clientSet` and `command line` or just one? if just one which one :)

If we say we go for the `clientSet` I will refactor the netpol tests to use that. If we support both we can keep as is for now and maybe do some refator in a follow up to refine the helpers.

@nader-ziada @CecileRobertMichon what are your thoughts?

mboersma: I'm on the fence personally, but I agree that we should pick one approach and use it consistently.

Using `kubectl` has some advantages (its `run` and `exec` commands aren't easily reproduced in Go code, and it might be easier code to read), but it also requires the test to fetch and use the correct binary version(s).

test/e2e/azure_test.go
mboersma: I copied this randomness directly from the AKS Engine test, but I'll try to find a more official container image to use.

test/e2e/azure_test.go
cpanato: Agree, so in the regards we can support both ways, depending of the test, if is just to create and check states and other things we can use the client. If need to run something inside the pod or other things we can use kubectl. In the network pol case it exec things to test the connectivity.

test/e2e/helpers.go
mboersma: I removed the comment, because when CAPZ eventually moves to a newer version of client-go, this will have to change since `Get` and friends require a context as the first argument. So it should self-correct.

The alternative is to run the current `Get` in a goroutine and cancel it if the context is done. I can go that way if we prefer.

test/e2e/azure_test.go
mboersma: `curlimages/curl` is the official Alpine-based image produced by the curl dev team. I looked for an existing CAPI or Kubernetes test image, and for something usable at mcr.microsoft.com, but didn't find anything. If there's an image we already use in these kind of test scenarios, please clue me in because that would be preferable.

test/e2e/helpers.go
nader-ziada: it fine to keep it like that for now

mboersma:
I think this spec should be refactored into a separate file, as we do in the CAPI test/e2e package. I'll do that next.

nader-ziada:
> I think this spec should be refactored into a separate file, as we do in the CAPI test/e2e package. I'll do that next.

in this pr or a followup?

Can you also please update the PR description to indicate it fixes the two issues #703 and #702

mboersma:
> in this pr or a followup?

I'm planning to do it in this PR, but I can easily do it as a followup if we want to merge this now.

> please update the PR description 

Done!

nader-ziada:
> > in this pr or a followup?
> 
> I'm planning to do it in this PR, but I can easily do it as a followup if we want to merge this now.
> 

up to you, I'm good either way

test/e2e/azure_test.go
```diff
@@ -122,6 +122,17 @@ var _ = Describe("Workoad cluster creation", func() {
 				WaitForControlPlaneIntervals: e2eConfig.GetIntervals(specName, "wait-control-plane"),
 				WaitForMachineDeployments:    e2eConfig.GetIntervals(specName, "wait-worker-nodes"),
 			})
+
+			Context("in a cluster with one or more nodes", func() {
+				AzureLBSpec(ctx, func() AzureLBSpecInput {
```
nader-ziada, line 127: I don't think a test would run with an `It`

see here https://onsi.github.io/ginkgo/

mboersma: The problem is that we're taking advantage of the workload cluster already being created to run some additional test assertions. So we can't wrap the `AzureLBSpec` in an `It` because they can't be nested (as far as I can tell), and if they're at the same level we can't guarantee the cluster creation ran before the LB test.

Maybe a better way to organize this would be to put the cluster creation in a `BeforeSuite` and then have `It`s that test functionality against the running cluster? That seemed like too much of a refactor to do here.

mboersma: Also, as far as I can tell the new test spec is being run as intended.

nader-ziada: I think that standard way i've seen before is using a `By` step

mboersma: I guess I'm misunderstanding. The `AzureLBSpec` consists only of `By` steps and does seem to run successfully:

```
INFO: Waiting for the worker machines to be provisioned
[1mSTEP[0m: waiting for the workload nodes to exist
[1mSTEP[0m: creating a Kubernetes client to the workload cluster
[1mSTEP[0m: creating an nginx deployment
[1mSTEP[0m: waiting for deployment default/ingress-nginx to be available
[1mSTEP[0m: creating an internal Load Balancer service
[1mSTEP[0m: waiting for service default/ingress-nginx-ilb to be available
[1mSTEP[0m: connecting to the internal LB service from a curl pod
[1mSTEP[0m: waiting for job default/curl-to-ilb-job to be complete
[1mSTEP[0m: creating an external Load Balancer service
[1mSTEP[0m: waiting for service default/ingress-nginx-elb to be available
[1mSTEP[0m: connecting to the external LB service from a curl pod
[1mSTEP[0m: waiting for job default/curl-to-elb-job to be complete
[1mSTEP[0m: connecting directly to the external LB service
[1mSTEP[0m: deleting the test resources
[1mSTEP[0m: Dumping all the Cluster API resources in the "create-workload-cluster-3vb4me" namespace
```

Should I replace the `Context` with a `By`? Sorry, I'm probably missing some ginkgo pattern. I had to get creative because as I said above, I wanted to re-use the existing cluster and that's already inside an `It` clause.

test/e2e/azure_test.go
```diff
@@ -122,6 +122,17 @@ var _ = Describe("Workoad cluster creation", func() {
 				WaitForControlPlaneIntervals: e2eConfig.GetIntervals(specName, "wait-control-plane"),
 				WaitForMachineDeployments:    e2eConfig.GetIntervals(specName, "wait-worker-nodes"),
 			})
+
+			Context("in a cluster with one or more nodes", func() {
```
CecileRobertMichon, line 126: Does this new spec add a significant duration to the test? if not, let's add it to the single control plane cluster as well

test/e2e/azure_lb.go
```diff
@@ -0,0 +1,294 @@
...
+type AzureLBSpecInput struct {
+	BootstrapClusterProxy framework.ClusterProxy
+	Namespace             *corev1.Namespace
+	ClusterName           string
+	SkipCleanup           bool
+}
+
+// AzureLBSpec implements a test that verifies Azure internal and external load balancers can
+// be created and work properly through a Kubernetes service.
+func AzureLBSpec(ctx context.Context, inputGetter func() AzureLBSpecInput) {
```
CecileRobertMichon, line 50: I like how you structured this and separated the spec from the actual test running it 👍

test/e2e/azure_test.go
nader-ziada: if its working as expected, you don't have to change anything

test/e2e/azure_test.go
mboersma: It doesn't add a lot of time (about two minutes max?), but even after adding the right tolerations, I wasn't able to get the test to run on the single control plane cluster. The error was something about "LB can't be created" that implied it needs worker nodes.

If that rings a bell or you have an idea what might be going wrong there, I would love to fix this up so it runs in both cluster configurations.

CecileRobertMichon: I see. We should consider adding a worker node or two to the single CP test, a cluster with a control plane and no workers isn't really a useful scenario to cover

nader-ziada: unless you plan to make this update in this PR, its lgtm

CecileRobertMichon: I'm good with doing that as a follow up PR too, this is already a good addition as-is

mboersma: I can do that change in a separate PR, so this one can proceed.

mboersma:
I squashed the commits and rebased off master.

mboersma:
It seems like this test has been flaky. I'll add some retry logic around the `connecting directly to the external LB service`.

go.mod
```diff
@@ -14,6 +14,7 @@ require (
 	github.com/google/go-cmp v0.5.0
 	github.com/google/gofuzz v1.1.0
 	github.com/google/uuid v1.1.1
+	github.com/hashicorp/go-retryablehttp v0.6.6
```
nader-ziada, line 17: seems like a cool library

test/e2e/azure_test.go
```diff
@@ -99,7 +99,7 @@ var _ = Describe("Workoad cluster creation", func() {
 		})
 	})
 
-	Context("Create multiple controlplane cluster with machine deployments", func() {
+	FContext("Create multiple controlplane cluster with machine deployments", func() {
```
nader-ziada, line 102: can you remove the Focus `F`

mboersma: Oops, yes.

go.mod
mboersma: It looked like the appropriate choice after a bit of research. Hashicorp is usually high-quality, and CAPZ already has some of their other libraries vendored in go.sum.

I like to keep vendoring as minimal as possible, so I also thought about writing our own retry loop. This can be surprisingly tricky--the retryablehttp docs convinced me I would probably miss some edge cases. But I'm happy to write the retry loop and avoid vendoring if that's a better tradeoff.

mboersma:
Eek, I noticed I left a `panic()` call in `helpers.go`. This must have been for debugging and I forgot to remove it. Also there should be doc comments on a few funcs in that file.

I've fixed both those and will force-push after this CI run finishes.

mboersma:

That's two passing runs in a row, here's hoping for a third. The retry does seem to be necessary, as the logs show the initial http `Get` fails but the second succeeds after a backoff.

test/e2e/azure_test.go
mboersma: I put this topic (client-go vs. `kubectl` for e2e) on the CAPZ office hours discussion for tomorrow.

nader-ziada:
all looks good for me, @CecileRobertMichon do you have any concerns?

PR #741, opened 2020-06-29
💎 Remove use of deprecated cluster scope Name() in favor of ClusterName()

CecileRobertMichon:
**What this PR does / why we need it**: clusterScope.Name() is deprecated since #716 as it might conflict with machine scope Name() (machine name vs cluster name). This replaces all remaining uses of Name() in the context of cluster name with ClusterName() and deletes the cluster Name() function.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Removed use of deprecated cluster scope Name() in favor of ClusterName()
```

### DIFF
--- a/cloud/scope/cluster.go
+++ b/cloud/scope/cluster.go
@@ -147,12 +147,6 @@ func (s *ClusterScope) ClusterName() string {
 	return s.Cluster.Name
 }
 
-// Name returns the cluster name.
-// DEPRECATED: use ClusterName() instead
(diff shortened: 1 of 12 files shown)


nader-ziada:
straight forward change with  no change in functionality, just cleaner code. The api-diff failure is for removing the deprecated Name() func which is acceptable

/lgtm

PR #746, opened 2020-07-01
🔧  eliminate plain text creds for additional template files

chewong:

Follow-up PR for https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/653. Also fixed CAPI e2e regression introduced by #653.

Fixes #

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
None
```

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -278,31 +278,31 @@ SSH_KEY_FILE=.sshkey
 rm -f "${SSH_KEY_FILE}" 2>/dev/null
 ssh-keygen -t rsa -b 2048 -f "${SSH_KEY_FILE}" -N '' 1>/dev/null
 echo "Machine SSH key generated in ${SSH_KEY_FILE}"
-export AZURE_SSH_PUBLIC_KEY=$(cat "${SSH_KEY_FILE}.pub" | base64 | tr -d '\n')
+export AZURE_SSH_PUBLIC_KEY=$(cat "${SSH_KEY_FILE}.pub" | base64 | tr -d '\r\n')
 
 # To populate secret in azure.json file.
 export AZURE_STANDARD_JSON_B64=$(echo '{
-      "cloud": "${AZURE_ENVIRONMENT}",
-      "tenantId": "${AZURE_TENANT_ID}",
-      "subscriptionId": "${AZURE_SUBSCRIPTION_ID}",
-      "aadClientId": "${AZURE_CLIENT_ID}",
-      "aadClientSecret": "${AZURE_CLIENT_SECRET}",
-      "resourceGroup": "${CLUSTER_NAME}",
-      "securityGroupName": "${CLUSTER_NAME}-node-nsg",
-      "location": "${AZURE_LOCATION}",
-      "vmType": "standard",
-      "vnetName": "${CLUSTER_NAME}-vnet",
-      "vnetResourceGroup": "${CLUSTER_NAME}",
-      "subnetName": "${CLUSTER_NAME}-node-subnet",
-      "routeTableName": "${CLUSTER_NAME}-node-routetable",
-      "loadBalancerSku": "standard",
-      "maximumLoadBalancerRuleCount": 250,
-      "useManagedIdentityExtension": false,
-      "useInstanceMetadata": true
-}' | envsubst | base64 | tr -d '\n')
+    "cloud": "${AZURE_ENVIRONMENT}",
+    "tenantId": "${AZURE_TENANT_ID}",
+    "subscriptionId": "${AZURE_SUBSCRIPTION_ID}",
+    "aadClientId": "${AZURE_CLIENT_ID}",
+    "aadClientSecret": "${AZURE_CLIENT_SECRET}",
+    "resourceGroup": "${CLUSTER_NAME}",
+    "securityGroupName": "${CLUSTER_NAME}-node-nsg",
+    "location": "${AZURE_LOCATION}",
+    "vmType": "standard",
+    "vnetName": "${CLUSTER_NAME}-vnet",
+    "vnetResourceGroup": "${CLUSTER_NAME}",
+    "subnetName": "${CLUSTER_NAME}-node-subnet",
+    "routeTableName": "${CLUSTER_NAME}-node-routetable",
+    "loadBalancerSku": "standard",
+    "maximumLoadBalancerRuleCount": 250,
+    "useManagedIdentityExtension": false,
+    "useInstanceMetadata": true
+}' | envsubst | base64 | tr -d '\r\n')
 
 # VMSS requires different values
-export AZURE_VMSS_JSON_B64=$(echo "$AZURE_STANDARD_JSON_B64" | base64 -d | jq '.vmType = "vmss"' | base64 | tr -d '\n')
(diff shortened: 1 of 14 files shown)


alexeldeib:
thanks for fixing this up @chewong!

hack/log/redact.sh
```diff
@@ -28,11 +28,7 @@ echo "================ REDACTING LOGS ================"
 
 log_files=( $(find "${ARTIFACTS:-${PWD}/_artifacts}" -type f) )
 redact_vars=(
-    "${AZURE_CLIENT_ID}"
```
CecileRobertMichon, line 31: we should still redact these as best effort if they do get set for any reason

maybe redirect stdout to /dev/null so it doesn't log "not found" everytime

hack/log/redact.sh
```diff
@@ -45,9 +45,9 @@ for log_file in "${log_files[@]}"; do
         # LC_CTYPE=C and LANG=C will prevent "illegal byte sequence" error from sed
         if [[ "$(uname)" == "Darwin" ]]; then
             # sed on Mac OS requires an empty string for -i flag
-            LC_CTYPE=C LANG=C sed -i "" "s|${redact_var}|===REDACTED===|g" "${log_file}" || true
+            LC_CTYPE=C LANG=C sed -i "" "s|${redact_var}|===REDACTED===|g" "${log_file}" > /dev/null || true
```
CecileRobertMichon, line 48: You might need &> /dev/null, I still see a bunch of `sed: -e expression #1, char 0: no previous regular expression` in the logs

CecileRobertMichon:
@chewong conformance is failing with `error: error validating "STDIN": error validating data: [unknown object type "nil" in Secret.data.standard, unknown object type "nil" in Secret.data.vmss]; if you choose to ignore these errors, turn validation off with --validate=false`

chewong:
> @chewong conformance is failing with `error: error validating "STDIN": error validating data: [unknown object type "nil" in Secret.data.standard, unknown object type "nil" in Secret.data.vmss]; if you choose to ignore these errors, turn validation off with --validate=false`

Hmm I wasn't able to repro this issue in my local environment. Let me dig into this

test/e2e/common.go
```diff
@@ -133,3 +138,13 @@ func getCloudProviderConfig(cluster, vmType string) (string, error) {
 	}
 	return base64.StdEncoding.EncodeToString(b), err
 }
+
+func redactLogs() {
```
CecileRobertMichon, line 142: side note: do you think it'd be easy to do the same with collecting workload cluster logs at some point?

chewong: `redactLogs()` should already be redacting workload cluster logs as long as they are stored in `${ARTIFACTS}`.

chewong:
> > @chewong conformance is failing with `error: error validating "STDIN": error validating data: [unknown object type "nil" in Secret.data.standard, unknown object type "nil" in Secret.data.vmss]; if you choose to ignore these errors, turn validation off with --validate=false`
> 
> Hmm I wasn't able to repro this issue in my local environment. Let me dig into this

Turned out kubekins image doesn't have `envsubst` binary. I pointed it to `hack/tools/bin/envsubst-v1.1.0` in `./hack/create-dev-cluster.sh`.

hack/create-dev-cluster.sh
```diff
@@ -23,6 +23,9 @@ set -o pipefail
 : "${AZURE_CLIENT_ID:?Environment variable empty or not defined.}"
 : "${AZURE_CLIENT_SECRET:?Environment variable empty or not defined.}"
 
+ENVSUBST="${PWD}/hack/tools/bin/envsubst-v1.1.0"
```
CecileRobertMichon, line 26: you should be able to use the symlink `/hack/tools/bin/envsubst` instead of having to pin to a version, and then instead of running make ${ENVSUBST} below just add `$(ENVSUBST)` as a dependency to a new makefile target called at the beginning of this script.

This is to avoid having the version hardcoded somewhere else than the makefile (which would require updating it whenever we bump the version).

test/e2e/common.go
CecileRobertMichon: sorry I meant collect logs for e2e, since right now we're only running `log-dump.sh` on conformance clusters

docs/development.md
```diff
@@ -278,31 +278,31 @@ SSH_KEY_FILE=.sshkey
...
+    "routeTableName": "${CLUSTER_NAME}-node-routetable",
+    "loadBalancerSku": "standard",
+    "maximumLoadBalancerRuleCount": 250,
+    "useManagedIdentityExtension": false,
+    "useInstanceMetadata": true
+}' | envsubst | base64 | tr -d '\r\n')
 
 # VMSS requires different values
-export AZURE_VMSS_JSON_B64=$(echo "$AZURE_STANDARD_JSON_B64" | base64 -d | jq '.vmType = "vmss"' | base64 | tr -d '\n')
+export AZURE_VMSS_JSON_B64=$(echo "$AZURE_STANDARD_JSON_B64" | base64 -d | jq '.vmType = "vmss"' | base64 | tr -d '\r\n')
```
CecileRobertMichon, line 305: I don't think we actually need both `AZURE_STANDARD_JSON_B64` and `AZURE_VMSS_JSON_B64` since we're using "vmss" vmType everywhere. We can just replace `AZURE_STANDARD_JSON_B64`with `AZURE_JSON_B64` and set the vmType to "vmss".

test/e2e/common.go
chewong: I think the CAPI test framework is already collecting a lot of relevant logs for us, similar to `log-dump.sh`, though different folder structure (e.g. https://gcsweb.k8s.io/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/746/pull-cluster-api-provider-azure-capi-e2e/1278793437226733568/artifacts/)

CecileRobertMichon: AFAIK it's collecting bootstrap cluster logs (CAPI/CAPZ managers, etc.) but not logs from the workload cluster VMs (eg. cloud-init logs)

CecileRobertMichon: maybe that's something we should add to the framework itself

docs/development.md
chewong: SGTM. Is it possible to get this PR merged first so we can have some test signals for our testgrid jobs over the long weekend? I will address this issue in a follow-up PR.

CecileRobertMichon: SGTM

test/e2e/common.go
chewong: Similar to the other issue in this PR, I can enable VM log collection in a follow-up PR. And yea this is definitely somewhere we could contribute back to CAPI.

chewong:
Looks like there is a flaky test case in pull-cluster-api-provider-azure-conformance-machine-pool-v1alpha3.
/unhold for now and I will keep an eye on https://testgrid.k8s.io/provider-azure-upstream#k8s-e2e-conformance-vmss-master-capz.

chewong:
Looks like there is a flaky test case in pull-cluster-api-provider-azure-conformance-machine-pool-v1alpha3.
/unhold
for now and I will keep an eye on https://testgrid.k8s.io/provider-azure-upstream#k8s-e2e-conformance-vmss-master-capz.

PR #750, opened 2020-07-02
💎  Embed ClusterDescriber in MachineScope to remove the need for duplicate Getters

CecileRobertMichon:
**What this PR does / why we need it**: This is an amendment to what was done in #716. What it does:
- embeds ClusterDescriber in MachineScope and MachinePoolScope (instead of having a field `ClusterScope azure.ClusterDescriber`)
- removes pass through functions such as `func (m *MachineScope) Location() string {
	return m.ClusterScope.Location()
}`
- ~changes newClusterScope, newMachineScope and newMachinePoolScope signature to not return a pointer (so that the returned scope can be used as param to build another scope without having to dereference the pointer).~
- Removes AzureClients from MachinePoolScope since it is already part of ClusterScope (already removed from MachineScope).

Now, machine/machinepool scope only need to implement functions that need to override the cluster scope one, eg. PublicIPSpecs(), instead of having to implement all the ClusterDescriber functions.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
:gem: Embed ClusterScope in MachineScope to remove the need for duplicate Cluster Getters
```

### DIFF
--- a/cloud/scope/machinepool.go
+++ b/cloud/scope/machinepool.go
@@ -19,7 +19,6 @@ package scope
 import (
 	"context"
 	"encoding/base64"
-	"github.com/Azure/go-autorest/autorest"
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
@@ -41,23 +40,21 @@ import (
 type (
 	// MachinePoolScopeParams defines the input parameters used to create a new MachinePoolScope.
 	MachinePoolScopeParams struct {
-		AzureClients
 		Client           client.Client
 		Logger           logr.Logger
 		MachinePool      *capiv1exp.MachinePool
 		AzureMachinePool *infrav1exp.AzureMachinePool
-		ClusterScope     *ClusterScope
+		azure.ClusterDescriber
 	}
 
 	// MachinePoolScope defines a scope defined around a machine pool and its cluster.
 	MachinePoolScope struct {
 		logr.Logger
-		AzureClients
 		client           client.Client
 		patchHelper      *patch.Helper
 		MachinePool      *capiv1exp.MachinePool
 		AzureMachinePool *infrav1exp.AzureMachinePool
-		ClusterScope     azure.ClusterDescriber
+		azure.ClusterDescriber
 	}
 )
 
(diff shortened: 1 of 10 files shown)


CecileRobertMichon:
/assign @alexeldeib 

Ace, let me know what you think about this change, it's inspired by your feedback in #716.

cloud/scope/machinepool.go
```diff
@@ -41,37 +39,35 @@ import (
...
 	// MachinePoolScope defines a scope defined around a machine pool and its cluster.
 	MachinePoolScope struct {
 		logr.Logger
-		AzureClients
 		client           client.Client
 		patchHelper      *patch.Helper
 		MachinePool      *capiv1exp.MachinePool
 		AzureMachinePool *infrav1exp.AzureMachinePool
-		ClusterScope     azure.ClusterDescriber
+		ClusterScope
```
alexeldeib, line 56: I might keep the interface for now, unless ClusterScope is already very small?

I thought this was a clever way to incrementally migrate us off of using the whole scope object, while starting to constrain an exposed interface. I'd assume it would merge with ClusterScope when the per-service bits get abstracted out so cluster scope is only shared vars (resourcegroup, name, location).

cloud/scope/machinepool.go
```diff
@@ -92,36 +88,6 @@ func NewMachinePoolScope(params MachinePoolScopeParams) (*MachinePoolScope, erro
 	}, nil
 }
 
-// Location returns the AzureCluster location.
```
alexeldeib, line 95: yay for removing code

cloud/scope/machinepool.go
CecileRobertMichon: Lemme know what you think of this new diff (I think it's better 😃 ) 

I kept all the commits to be able to revert if needed, will squash once we have consensus

devigned:
Pending test pass, lgtm.

CecileRobertMichon:
/hold for squash

@alexeldeib thoughts?

CecileRobertMichon:
rebased

/hold cancel

PR #764, opened 2020-07-07
💚 add test to validate accelerated networking for VMs

### LINKED ISSUE #706: [e2e] machines should have accelerated networking enabled if supported
An e2e test to validate that on a workload cluster, machine types that support accelerated networking has it enabled 

more info here: https://docs.microsoft.com/en-us/azure/virtual-network/create-vm-accelerated-networking-powershell#supported-vm-instances

mboersma:

Validates that the Azure VMs created have accelerated networking enabled if it's available for that SKU. (This is the default behavior--I didn't add a test for explicitly toggling it on or off.)

Fixes #706

I didn't add a similar test for VM scalesets because we don't utilize `MachinePool` in e2e tests (yet).

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
💚 add test to validate accelerated networking for VMs
```

### DIFF
--- a/test/e2e/azure_test.go
+++ b/test/e2e/azure_test.go
@@ -125,7 +125,7 @@ var _ = Describe("Workload cluster creation", func() {
 				})
 			})
 
-			Context("Creating a accessible load balancer", func() {
+			Context("Creating an accessible load balancer", func() {
 				AzureLBSpec(ctx, func() AzureLBSpecInput {
 					return AzureLBSpecInput{
 						BootstrapClusterProxy: bootstrapClusterProxy,
@@ -143,6 +143,14 @@ var _ = Describe("Workload cluster creation", func() {
 						Namespace:             namespace,
 						ClusterName:           clusterName,
 						SkipCleanup:           skipCleanup,
+					}
+				})
+			})
+
+			Context("Validating accelerated networking", func() {
+				AzureAcceleratedNetworkingSpec(ctx, func() AzureAcceleratedNetworkingSpecInput {
+					return AzureAcceleratedNetworkingSpecInput{
+						ClusterName: clusterName,
 					}
 				})
 			})
--- a/cloud/services/agentpools/agentpools_test.go
+++ b/cloud/services/agentpools/agentpools_test.go
@@ -229,7 +229,7 @@ func TestReconcile(t *testing.T) {
 				Name:          "my-agent-pool",
 				ResourceGroup: "my-rg",
 				Cluster:       "my-cluster",
-				SKU:           "Standard_A1",
+				SKU:           "Standard_D2s_v3",
 				Version:       to.StringPtr("9.99.9999"),
 				Replicas:      2,
 				OSDiskSizeGB:  100,
@@ -240,7 +240,7 @@ func TestReconcile(t *testing.T) {
 					ManagedClusterAgentPoolProfileProperties: &containerservice.ManagedClusterAgentPoolProfileProperties{
 						Count:               to.Int32Ptr(2),
 						OsDiskSizeGB:        to.Int32Ptr(100),
-						VMSize:              containerservice.VMSizeTypesStandardA1,
+						VMSize:              containerservice.VMSizeTypesStandardD2sV3,
 						OrchestratorVersion: to.StringPtr("9.99.9999"),
 						ProvisioningState:   to.StringPtr("Succeeded"),
 					},
--- a/test/e2e/azure_accelnet.go
+++ b/test/e2e/azure_accelnet.go
@@ -1,0 +1,96 @@
+// +build e2e
+
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package e2e
+
+import (
+	"context"
+	"fmt"
+	"path/filepath"
+
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-12-01/compute"
+	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2019-06-01/network"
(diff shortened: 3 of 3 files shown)


test/e2e/azure_test.go
```diff
@@ -33,7 +33,7 @@ import (
 	"sigs.k8s.io/cluster-api/util"
 )
 
-var _ = Describe("Workoad cluster creation", func() {
+var _ = Describe("Workload cluster creation", func() {
```
CecileRobertMichon, line 36: good catch!

needs an update to https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-sigs/cluster-api-provider-azure/cluster-api-provider-azure-presubmits.yaml#L84 as well 

I recommend doing the spelling fix in a separate PR so it can merge quickly and not impact other PRs when we merge the test-infra config fix

test/e2e/azure_accelerated_networking.go
```diff
@@ -0,0 +1,86 @@
...
+	subscriptionID := settings.GetSubscriptionID()
+	authorizer, err := settings.GetAuthorizer()
+	Expect(err).NotTo(HaveOccurred())
+	vmsClient := compute.NewVirtualMachinesClient(subscriptionID)
+	vmsClient.Authorizer = authorizer
+	nicsClient := network.NewInterfacesClient(subscriptionID)
+	nicsClient.Authorizer = authorizer
+
+	By("verifying EnableAcceleratedNetworking for the primary NIC of each VM")
+	acceleratedNetworking := map[compute.VirtualMachineSizeTypes]bool{
```
CecileRobertMichon, line 57: what is this list? A subset of SKUs that we expect to run tests with? Does that mean this test would fail if it was run with any other vmSize?

test/e2e/azure_accelerated_networking.go
```diff
@@ -0,0 +1,86 @@
...
+		compute.VirtualMachineSizeTypesStandardA1:    false,
```
CecileRobertMichon, line 58: CAPZ doesn't work with A1 sizes because kubeadm requires min 2 CPUs so it's probably misleading to have it here (see #765)

CecileRobertMichon:
So I might be misunderstanding how the test works but from a first look, I'm not sure I see what value testing this in e2e is bringing that we couldn't get with a unit test? if we're simply testing that the default is what we expect it to be and that that value is what is sent to the Azure API as part of the NIC CreateOrUpdate then I think all of that could be done through a unit test. If we're validating that the resulting VM NIC actually has accelerated networking then we would actually be testing Azure Networking itself at that point. Not saying we don't want to do that, but what's the relative (ie. added time in the e2e suite) vs. value?

test/e2e/azure_accelerated_networking.go
mboersma: Yes, exactly. Rather than (re-)implement the Azure API calls to find the accelerated networking capability, I added a static lookup table. I can populate it with all known SKUs if we like, these are just the ones that I can see referenced in CAPZ code.

mboersma:
>  I'm not sure I see what value testing this in e2e is bringing that we couldn't get with a unit test?

We do have unit tests that cover accelerated networking and ensure that it is carried properly from the cluster definition through to the API call (via mocks). But this e2e test connects to the running VMs and verifies that they actually have their accelerated networking property set as expected. 

It just tests the default case because I assumed it's too expensive to create clusters for each possibility, and because it wasn't obvious to me in the test framework how to find the acceleratedNetworking setting used from the initial cluster definition.

Maybe that doesn't add enough value to be worthwhile. We can close this if that's the case.

test/e2e/azure_test.go
```diff
@@ -96,6 +96,14 @@ var _ = Describe("Workload cluster creation", func() {
 				WaitForControlPlaneIntervals: e2eConfig.GetIntervals(specName, "wait-control-plane"),
 				WaitForMachineDeployments:    e2eConfig.GetIntervals(specName, "wait-worker-nodes"),
 			})
+
+			Context("in a cluster with a single node", func() {
```
nader-ziada, line 100: I don't think there is much value in this test with a single node in general, so I would prefer to remove this test and not add to it

nader-ziada:
@CecileRobertMichon do you prefer to not have this test increasing the time of e2e?

CecileRobertMichon:
@nader-ziada discussed with @mboersma and we concluded that there is value to adding this test but that we should look into running multiple cluster configurations in parallel (and/or clusters with multiple machine configurations) in e2e so we can test multiple scenarios instead of just the one in the future.

nader-ziada:
@mboersma I think something might have went wrong with the rebase, looks like the tests are failing for valid reasons

mboersma:
I *finally* got a clue and fixed the unit test I had broken. So this passed CI and is ready for review again. Thanks for your patience. 😄

nader-ziada:
lgtm

test/e2e/azure_accelerated_networking.go
```diff
@@ -0,0 +1,88 @@
...
+	for page.NotDone() {
+		for _, vm := range page.Values() {
+			sku := vm.HardwareProfile.VMSize
+			for _, nic := range *vm.NetworkProfile.NetworkInterfaces {
+				if nic.Primary != nil && *nic.Primary {
+					// verify that accelerated networking is enabled if the SKU is capable
+					nicInfo, err := nicsClient.Get(ctx, rgName, filepath.Base(*nic.ID), "")
+					Expect(err).NotTo(HaveOccurred())
+					capable, found := acceleratedNetworking[sku]
+					Expect(found).To(BeTrue(), "SKU %s was not found, please add to the acceleratedNetworking lookup table.")
```
CecileRobertMichon, line 79: what do you think about skipping the test instead if that sku isn't in the list? I'm concerned about people trying to run the tests locally with their own sub which might have different quotas than ours and selecting a VM size that isn't in this list causing the test to fail. If we expect the tests to only work with a small set of VM sizes that should at least be documented in developer docs, but I'd prefer we don't fail at all since that's not a real failure, just a gap in the tests.

mboersma: > what do you think about skipping the test instead if that sku isn't in the list?

I think that's a gentler way to do it--good idea. However, we can't actually use `Skip()` because this isn't isolated in an `It` clause and thus will skip the entire "3 masters 2 nodes" test. We could log this text and just return from this func instead.

mboersma:
> creating Azure clients with the workload cluster's subscription
> verifying EnableAcceleratedNetworking for the primary NIC of each VM

Looks good. I think this is ready again for a review.

Issue #757, opened 2020-07-07
[help wanted] refactor remaining services following established pattern


CecileRobertMichon:
Following the pattern established in #716, we should refactor each service to follow the same pattern. Each service should have its own refactor PR to make reviewing easier. For example, network interfaces was already done in #742.

- [x] disks (@cpanato)

- [x] groups (@cpanato)

- [x] exp/managedclusters

- [ ] exp/agentpools

- [x]  public IPs (@CecileRobertMichon)

- [x] network interfaces (@CecileRobertMichon)

- [x] load balancers  (@CecileRobertMichon)

- [x] subnets (@cpanato)

- [x] virtual networks (@cpanato)

- [x] route tables (@nader-ziada)

- [x] inbound nat rules (@CecileRobertMichon)

- [x] role assignments (@CecileRobertMichon)

- [x]  resource skus (@alexeldeib)

- [x] securitygroups (@CecileRobertMichon)

- [x] exp/scalesets (@CecileRobertMichon)

- [x] virtualmachines (@CecileRobertMichon)

Let me know if you want to help with one of these in comments and I'll add your username next to it!

/help

cpanato:
Can I take disks?

cpanato:
Can I take `groups` ?

cpanato:
@CecileRobertMichon can I take subnets?

devigned:
@cpanato, go for it! I don't think anyone is working on it right now.

cpanato:
ok, after the subnet i will take the virtual networks if that is fine

cpanato:
going to work on `virtual networks`

nader-ziada:
route tables is the last one without a name, I'll take that one

CecileRobertMichon:
@alexeldeib changed resource skus to you since you have #783 open

nader-ziada:
I can take that one after the existing "refactoring PRs" get merged

nprokopic:
I can take exp/managedclusters.

nader-ziada:
thanks @nprokopic

nprokopic:
FYI I finally got to start with AKS refactoring (managed cluster and agent pools), started working on it this week.

nprokopic:
FYI I started working on the final part of the refactoring - exp/agentpools.

nprokopic:
We can also tick off exp/managedclusters.

PR #771, opened 2020-07-08
💎 Refactor load balancer scope to interface

CecileRobertMichon:
**What this PR does / why we need it**: Follow up from #716, this PR applies the same pattern to load balancers.

Note that this also combines internal and public load balancers services into one load balancer service (to mirror the Azure load balancer client).

Partially addresses #757 

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Refactor load balancer scope to interface
```


cloud/scope/cluster.go
```diff
@@ -117,6 +117,34 @@ func (s *ClusterScope) PublicIPSpecs() []azure.PublicIPSpec {
 	}
 }
 
+// LBSpecs returns the load balancer specs.
+func (s *ClusterScope) LBSpecs() []azure.LBSpec {
+	return []azure.LBSpec{
+		{
+			// Public API Server LB
```
devigned, line 124: Is this internal or public? The role says `infrav1.InternalRole` and the comment says Public API Server.

CecileRobertMichon: that's a bug... great catch 

Should be public

CecileRobertMichon: actually just the comment is wrong

cloud/services/loadbalancers/loadbalancers.go
```diff
@@ -0,0 +1,228 @@
...
+				},
+			},
+		}
+
+		if lbSpec.Role == infrav1.APIServerRole || lbSpec.Role == infrav1.InternalRole {
+			probeName := "HTTPSProbe"
+			lb.LoadBalancerPropertiesFormat.Probes = &[]network.Probe{
+				{
+					Name: to.StringPtr(probeName),
+					ProbePropertiesFormat: &network.ProbePropertiesFormat{
```
CecileRobertMichon, line 136: this is the only thing I changed that's not strictly a refactor... The probe for the public LB (api server role) was [still using TCP ](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/cloud/services/publicloadbalancers/publicloadbalancers.go#L113), I changed it to HTTPS to match https://github.com/kubernetes-sigs/cluster-api-provider-azure/commit/316a2a694bef75158af76a99084734517bc0f457 since we're using standard LBs everywhere, and it made it simpler to not have a separate probe if/else per role. Tested functional behavior is the same and this change is backwards compatible.

Happy to revert it if reviewers think a separate PR would be more appropriate and we'd rather keep this one as a strict refactor.

cloud/services/loadbalancers/loadbalancers.go
```diff
@@ -0,0 +1,228 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
awesomenix, line 2:
```suggestion
Copyright 2020 The Kubernetes Authors.
```

cloud/services/loadbalancers/loadbalancers.go
awesomenix: Thanks for heads up, was about to comment on the if .. else part. But this comment made it clear as to this is just a copy paste rather than a complete refactor.

CecileRobertMichon:
@devigned please take another look, rebased and fixed copyright

PR #777, opened 2020-07-10
 ✨ implement ephemeral os

### LINKED ISSUE #487: Ephemeral OS Disks
/kind feature

[Ephemeral OS Disks](https://docs.microsoft.com/en-us/azure/virtual-machines/linux/ephemeral-os-disks) is a new feature in Azure that allows the OS disk to use local SSD storage, with no writes to Azure storage. If a VM is stopped or deprovisioned, its local storage is lost. 

Benefits - VMs deploy faster, and have better local storage performance. The OS disk will perform at the Max cached storage throughput for the VM size. For example with a Standard_D2s_v3 size VM using a 50 GiB OS disk - it can achieve 4000 IOPs with ephemeral disks enabled, or 240 IOPs using a Premium P6 Premium SSD at 50GiB. Apps will get faster container and emptydir performance. Container pull times are also improved.

reference: https://github.com/Azure/aks-engine/blob/master/docs/topics/features.md#ephemeral-os-disks

Requirements:

Be sure you are using a VM size that supports cache for the local disk
The OS disk size must be set to <= the VM's cache size in GiB
These are fully explained in the [Ephemeral OS Disks docs](https://docs.microsoft.com/en-us/azure/virtual-machines/linux/ephemeral-os-disks).

**Environment:**

alexeldeib:
implements ephemeral os

Fixes #487

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
add support for ephemeral os disks
```


alexeldeib:
need to add UT and validation to only use Standard_LRS SKU with ephemeral OS

docs/topics/ephemeral-os.md
```diff
@@ -0,0 +1,35 @@
+# Ephemeral OS
+
+This document describes how to use ephemeral OS for VMs provisioned in Azure. Ephemeral OS uses local VM storage for changes to the OS disk. Storage devices local to the VM host will not be bound by normal managed disk SKU limits. Instead they will always be capable of saturating the VM level limits. This can significantly performance performance on the OS disk. Ephemeral storage used for the OS will not persist between maintenance events and VM redeployments. This is ideal for stateless base OS disks, where any stateful data is kept elsewhere.
```
devigned, line 3:
```suggestion
This document describes how to use ephemeral OS for VMs provisioned in Azure. Ephemeral OS uses local VM storage for changes to the OS disk. Storage devices local to the VM host will not be bound by normal managed disk SKU limits. Instead they will always be capable of saturating the VM level limits. This can significantly increase performance of the OS disk. Ephemeral storage used for the OS will not persist between maintenance events and VM redeployments. This is ideal for stateless base OS disks, where any stateful data is kept elsewhere. 
```

alexeldeib: "This can significantly performance performance" 😄

devigned: You are unambiguous regarding the key point.

alexeldeib:
```

  Compatible changes:
  - DiffDiskSettings: added
  - ManagedDisk.DiffDiskSettings: added
  - ValidateManagedDisk: added
sigs.k8s.io/cluster-api-provider-azure/cloud/services/resourceskus
  Incompatible changes:
  - Client.HasEphemeralOSDiskSupport: added
  Compatible changes:
  - (*AzureClient).HasEphemeralOSDiskSupport: added
sigs.k8s.io/cluster-api-provider-azure/cloud/services/resourceskus/mock_resourceskus
  Compatible changes:
  - (*MockClient).HasEphemeralOSDiskSupport: added
  - (*MockClientMockRecorder).HasEphemeralOSDiskSupport: added
sigs.k8s.io/cluster-api-provider-azure/cloud/services/virtualmachines
  Compatible changes:
  - Service.ResourceSkusClient: added
```

docs/topics/ephemeral-os.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Azure Machine DiffDiskSettings
+
+Azure Machines support optionally specifying a field called `diffDiskSettings`. This mirrors the Azure Compute REST API.
+
+When `diffDiskSettings.option` is set to `Local`, ephemeral OS will be enabled. We use the API shape provided by compute directly as they expose other options, although this is the main one relevant at this time.
+
+## Example
+
+The below example shows how to enable ephemeral OS for a machine template. For control plane nodes, we strongly recommend using etcd data disks to avoid data loss.
```
CecileRobertMichon, line 32: could you link to the data disk docs here?

docs/topics/ephemeral-os.md
```diff
@@ -0,0 +1,53 @@
...
+maintenance events and VM redeployments. This is ideal for stateless
+base OS disks, where any stateful data is kept elsewhere. 
+
+There are a few kinds of local storage devices available on Azure VMs.
+Each VM size will have a different combination. For example, some sizes
+support premium storage caching, some sizes have a temp disk while
+others do not, and some sizes have local nvme devices with direct
+access. Ephemeral OS uses the cache for the VM size, if one exists.
+Otherwise it will try to use the temp disk if the VM has one. These are
+the only supported options, and we do not expose the ability to choose
```
CecileRobertMichon, line 18: the ability to choose? incomplete sentence

docs/topics/ephemeral-os.md
```diff
@@ -0,0 +1,53 @@
...
+(the default behavior is typically most desirable). This corresponds to
+the `placement` property in the Azure Compute REST API.
+
+See [the documentation](https://docs.microsoft.com/en-us/azure/virtual-machines/linux/ephemeral-os-disks) for full details.
```
CecileRobertMichon, line 22:
```suggestion
See [the Azure documentation](https://docs.microsoft.com/en-us/azure/virtual-machines/linux/ephemeral-os-disks) for full details.
```

docs/topics/ephemeral-os.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Azure Machine DiffDiskSettings
+
+Azure Machines support optionally specifying a field called `diffDiskSettings`. This mirrors the Azure Compute REST API.
+
+When `diffDiskSettings.option` is set to `Local`, ephemeral OS will be enabled. We use the API shape provided by compute directly as they expose other options, although this is the main one relevant at this time.
+
+## Example
```
CecileRobertMichon, line 30: Consider adding a `Known Limitations` section to point out potential limitations of ephemeral disks (eg. only some vm sizes are supported.)

api/v1alpha3/azuremachine_validation.go
```diff
@@ -102,6 +103,61 @@ func ValidateOSDisk(osDisk OSDisk, fieldPath *field.Path) field.ErrorList {
 
 	allErrs = append(allErrs, validateStorageAccountType(osDisk.ManagedDisk.StorageAccountType, fieldPath)...)
 
+	if errs := ValidateManagedDisk(osDisk.ManagedDisk, osDisk.ManagedDisk, fieldPath.Child("managedDisk")); len(errs) > 0 {
```
CecileRobertMichon, line 106: One more comment (noticed this from looking at #652), it seems a bit awkward to set the "old" object as itself for a create and it's different from what CAPI does, which set old to `nil` for validateCreate() https://github.com/kubernetes-sigs/cluster-api/blob/master/api/v1alpha3/machine_webhook.go#L119

api/v1alpha3/azuremachine_webhook.go
```diff
@@ -87,6 +89,10 @@ func (m *AzureMachine) ValidateUpdate(old runtime.Object) error {
 		allErrs = append(allErrs, errs...)
 	}
 
+	if errs := ValidateManagedDisk(old.Spec.OSDisk.ManagedDisk, m.Spec.OSDisk.ManagedDisk, field.NewPath("osDisk").Child("managedDisk")); len(errs) > 0 {
```
CecileRobertMichon, line 92: Can this instead call `ValidateOSDisk` which itself calls `ValidateManagedDisk`? I don't see anything in `ValidateOSDisk` that applies to AzureCluster Create but not Update

api/v1alpha3/azuremachine_validation.go
CecileRobertMichon: ValidateOSDisk should take old and new and pass it through to `ValidateManagedDisk`, this makes the assumption that it's always only used in Create validation.

api/v1alpha3/azuremachine_webhook.go
alexeldeib: diffDiskSettings can only be set on OS disks, I need to pull that up a layer. That's the difference between ValidateOSDisk and ValidateManagedDisk

alexeldeib: actually this field can't be set for managed disks which aren't os disks, so I'll also validate against that.

api/v1alpha3/azuremachine_validation.go
alexeldeib: the capi approach doesn't work for non-pointer values, I think? will try and see what I can do

CecileRobertMichon: ok, if there's a good reason for passing in a non-nil old object in create (looks like there is) then 👍 

Still think we should pass through the old object from ValidateCreate/ValidateUpdate though.

alexeldeib: yeah, actually -- I don't really see a better way to handle updates for non-nil structs and reusing the same func. We can have two different funcs?

CecileRobertMichon: I think this > two different funcs

would prefer avoiding copy paste logic

alexeldeib: I guess we can define validation funcs on the structs all the way down, and then call new.validate(old) at each layer?

api/v1alpha3/azuremachine_webhook.go
alexeldeib: *taps head* don't need to validate if it's not in the API

i'll move it up a level to osDisk struct

alexeldeib:
probably this one. when 783 goes in, I'll refactor accelerated networking and ephemeral os in addition to AZs.

alexeldeib:

@CecileRobertMichon do you see anything obvious here that would be failing the tests? it seems like it's just taking time for all the controllers to come up. I only see the cakcp and capbk controller logs: https://gcsweb.k8s.io/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/777/pull-cluster-api-provider-azure-e2e/1283489827391541248/artifacts/clusters/bootstrap/controllers/

alexeldeib:
oh, this is an issue with the capi controller, not the capz controller. So this PR shouldn't affect that at all? not sure what's going on...

CecileRobertMichon:
Can you pls squash/rebase https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/777/commits/72d628e38e5f358cb450c4d7ab1b8eac64a28e8a

alexeldeib:
squashed

Issue #765, opened 2020-07-08
Add VMSize validation to reject VM Sizes with < 2 CPU


CecileRobertMichon:
As documented at https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/#before-you-begin, kubeadm requires 2 CPUs. To improve UX, we should fail fast when the user specifies an Azure VM size that has less than 2 CPUs, that is:

- Standard_A1_v2
- Standard_B1ls
- Standard_B1s
- Standard_B1ms
- Standard_DC1s_v2
- Standard_D1_v2
- Standard_DS1_v2
- Standard_F1s
- Standard_F1

(source: https://docs.microsoft.com/en-us/azure/virtual-machines/linux/compute-benchmark-scores)

This validation should be added to azuremachine_validation.go

/help
/good-first-issue

CecileRobertMichon:
@rsmitty does the same apply to Talos?

rsmitty:
It’s not strictly required but we do mention 2 CPUs as a minimum for control plane nodes in our docs, so I’m cool with this.

alexeldeib:
It's more accurate to call the SKU API for this to avoid generating a static list, the API exposes vCPU as a capability. We could cache the known sizes at start up to avoid extra http calls inside the webhook? Maybe with some background refresh or refresh-on-miss logic?

alexeldeib:
thoughts on breaking out skus client to a standalone cache? we will probably start querying it more frequently -- it might make sense not to issue calls for every VM, every time. We can query once for a subscription and cache the result. If we see we're restricted we can use that as a cache "miss" to refresh (to avoid getting hit by shifting point in time restrictions).

mboersma:
I think a cache would be an good optimization for something like this. The results for such API calls are basically static for a given subscription. I intended to memoize the [`resourceskusclient.HasAcceleratedNetworking`](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/645/files#diff-2a40a35715cc74836c39b99b338dd4eeR77) call in #645 but didn't get there.

CecileRobertMichon:
I'm going to remove good-first-issue, given that requirements aren't well defined at the moment and that adding a cache/making calls to the sku client would require a little more work.

/remove-good-first-issue

[...]

PR #783, opened 2020-07-12
💎  refactor skus client

alexeldeib:

Refactor the skus client to query Azure once on reconcile loop per cluster/machine. Currently the cluster reconciler queries Azure for the failure domains, and the machine reconciler queries once in the nic service and once in the machine reconciler to check for failure domains and accelerated networking support. #777 adds another query for ephemeral OS, and #765 would add another to check vCPU.

We don't try to cache multiple subscriptions worth of data across reconcile runs; we just query the subscription once at the beginning of reconcile. It would be easily adaptable to do so, but the returned JSON is 9.2MB for my subscription, so that could add up quickly. A good reason to avoid multiple queries! I still need to clean that up a bit and maybe rename some things

This PR will make exposing features easier in the future -- it makes sense to use this in https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/777, but given the size of the PRs I broke it out. It also makes https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/765 and validating things like [EncryptionAtHost](https://github.com/Azure/aks-engine/pull/3041) trivial.

Eliminates the availability zones service, which is a twist on the same functionality.

none yet, but if this looks okay I can probably implement #765 on top.

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
NONE
```


alexeldeib:
```
sigs.k8s.io/cluster-api-provider-azure/cloud/services/scalesets
  Incompatible changes:
  - NewService: changed from func(sigs.k8s.io/cluster-api-provider-azure/cloud.Authorizer) *Service to func(sigs.k8s.io/cluster-api-provider-azure/cloud.Authorizer, *sigs.k8s.io/cluster-api-provider-azure/cloud/services/resourceskus.Cache) *Service
  - Service.ResourceSkusClient: removed
  Compatible changes:
  - Service.ResourceSkusCache: added
sigs.k8s.io/cluster-api-provider-azure/cloud/services/availabilityzones
  Incompatible changes:
  - AzureClient: removed
  - Client: removed
  - NewClient: removed
  - NewService: removed
  - Service: removed
  - Spec: removed
sigs.k8s.io/cluster-api-provider-azure/cloud/services/availabilityzones/mock_availabilityzones
  Incompatible changes:
  - MockClient: removed
  - MockClientMockRecorder: removed
  - NewMockClient: removed
sigs.k8s.io/cluster-api-provider-azure/cloud/services/resourceskus/mock_resourceskus
  Incompatible changes:
  - (*MockClient).HasAcceleratedNetworking: removed
  - (*MockClientMockRecorder).HasAcceleratedNetworking: removed
sigs.k8s.io/cluster-api-provider-azure/cloud/services/networkinterfaces
  Incompatible changes:
  - NewService: changed from func(NICScope) *Service to func(NICScope, *sigs.k8s.io/cluster-api-provider-azure/cloud/services/resourceskus.Cache) *Service
  - Service.ResourceSkusClient: removed
  Compatible changes:
  - Service.ResourceSkusCache: added
sigs.k8s.io/cluster-api-provider-azure/cloud/services/resourceskus
  Incompatible changes:
  - (*AzureClient).HasAcceleratedNetworking: removed
  - Client.HasAcceleratedNetworking: removed
  Compatible changes:
  - AcceleratedNetworking: added
  - Cache: added
  - CapabilitySupported: added
  - CapabilityUnupported: added
  - Disks: added
  - EphemeralOSDisk: added
  - Kind: added
  - NewCache: added
  - NewCacheFunc: added
  - NewStaticCache: added
  - NewStaticCacheFn: added
  - SKU: added
  - Supported: added
  - VirtualMachines: added
sigs.k8s.io/cluster-api-provider-azure/cloud/scope
  Incompatible changes:
  - ClusterScopeParams: old is comparable, new is not
  Compatible changes:
  - ClusterScope.SKUCache: added
  - ClusterScopeParams.NewSKUCache: added
```

cloud/services/resourceskus/cache.go
```diff
@@ -0,0 +1,242 @@
...
+	// data is the cached sku information from Azure.
+	// synchronization required if data is cached across reconcile calls, (i.e., refreshed in background as Runnable via mgr.Add(...))
+	data []compute.ResourceSku
+}
+
+// NewCacheFunc allows for mocking out the underlying client
+type NewCacheFunc func(context.Context, azure.Authorizer) (*Cache, error)
+
+// NewCache instantiates a cache and initializes its contents.
+func NewCache(ctx context.Context, auth azure.Authorizer) (*Cache, error) {
```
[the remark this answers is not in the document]
alexeldeib: I left it inside New because I don't think it makes much sense to use an uninitialized cache and it also makes mocking trivial. If we move refresh to a separate func called by users we depend on clients to correctly invoke it (compare with the usage of the staticFn). I can see why that's a bit odd here, I agree it's not a great fit. 

Could we move it inside the controller/reconciler setup and pass the cache directly into the scope params instead?

alexeldeib:
> I wonder if we should have a set of data services that outlive an individual reconcile loop

I think this makes sense, and was definitely my motivation for starting this. I originally implemented it as a runnable we could add to the manager and handle background refresh once per $interval, but then realized we now support reconciling clusters against multiple subs. When I checked the size of the response I figured we probably don't want to start caching that much data for each run. Depending on use case/acceptable resource usage, we can explore that further / iterate and see what makes sense. I'd definitely be curious where else we can take advantage of something like this, auth is one area I wanted to investigate a bit further. We probably don't need e.g. authorizer per reconcile for every reconciler (== extra http requests to AAD).

cloud/services/resourceskus/cache.go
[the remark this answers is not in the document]
alexeldeib: would probably move the NewCacheFunc into e.g. the controller and initialize it in main.go to make that work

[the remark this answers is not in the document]
devigned: The context being passed in on New definitely helps to simplify the outer interface. Also, it is nice to not have to prime the cache after initialization.

What if the cache has a default time to live before refresh, it would then check last time updated before actions on the cache, internally. If an action on the cache will act on never refreshed or expired data, the cache calls refresh. Each action on the cache would then include a `context.Context` as the first parameter of the func. Wdyt?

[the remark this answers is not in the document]
alexeldeib: The sneaky part is there's no additional refresh right now :)

After I removed it from the manager, it's really just a list produced once at the start of reconcile. The extra Cache type is useful to provide functionality on top of the existing Azure types (which is...limited, IMO). That's why I don't particularly think making refresh public makes sense -- if anything, I'd prefer to convert it to e.g. a runnable, keep refresh private, and return the error from refresh to `runnable.Start()` (or something like that).

If we do that we can move cache initialization inside the Start func which takes a done channel. Need to do a little context/done channel mapping.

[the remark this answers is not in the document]
devigned: Is that going to be more than you want to get into on this PR?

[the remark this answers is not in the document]
alexeldeib: for now, I'll probably just shuffle things a bit to make this a little nicer in the current form w.r.t context. I'll play with it and push an update once I think it's sane. Seeing the result size be ~10MB for a single subscription made me lean back a bit from caching across multiple subscriptions (which is implied by caching across reconciles), but that made it all the more valuable to only fetch + parse one time.

if we find this works well in practice we can always extend it. mostly feels like an easy/cheap way to reduce latency + extra http calls, and make feature discovery easier at the same time.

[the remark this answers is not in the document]
devigned: Dang... 10MB is quite the payload. I almost want to include this in image creation or mount it as a file for the controller.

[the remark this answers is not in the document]
alexeldeib: We might be able to take the whole list and ignore restrictions. We'd need the restriction info per sub so we'd still need to make the same call, it doesn't help much.

Alternatively the Azure API does have some pagination if I'm not mistaken, so we could try to do something more clever around that in the caching stage, since we're already looking per-request. It helps for the machine case because we can terminate early, but for cluster failure domains we'd still need to iterate a larger set.

As I was writing this I remembered we aren't using the location filter anymore -- I bet that will help cut the size down quite a bit.

cloud/services/resourceskus/cache.go
```diff
@@ -0,0 +1,242 @@
...
+
+// NewStaticCache initializes a cache with data and no ability to refresh. Used for testing.
+func NewStaticCache(data []compute.ResourceSku) *Cache {
+	return &Cache{
+		data: data,
+	}
+}
+
+func (c *Cache) refresh(ctx context.Context) error {
+	data, err := c.client.List(ctx, "")
```
devigned, line 73: Hmm... Would it be better to scope the query by location?

alexeldeib: exactly my other comment heh

I suspect this will greatly reduce the size, I forgot we stopped doing this

cloud/services/resourceskus/cache.go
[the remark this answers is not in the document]
devigned: I commented on the location and finished reading your comment at roughly the same time. Heh.

cloud/services/networkinterfaces/service.go
```diff
@@ -43,11 +43,11 @@ type Service struct {
 	InternalLoadBalancersClient internalloadbalancers.Client
 	PublicIPsClient             publicips.Client
 	InboundNATRulesClient       inboundnatrules.Client
-	ResourceSkusClient          resourceskus.Client
+	ResourceSkusCache           *resourceskus.Cache
```
CecileRobertMichon, line 46: is ResourceSkusCache needed as part of the Service struct? Isn't it already part of Scope?

alexeldeib: whoops -- good catch. I had it flipped before.

cloud/services/resourceskus/cache.go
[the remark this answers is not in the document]
alexeldeib: okay, so depending on region looks to be ~600-775 KB for one sub. That's a lot more reasonable to cache over iterations. I'll take another pass an make this a background process again I think.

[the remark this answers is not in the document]
alexeldeib: I might leave this per request right now. I'd like to turn it into a global cache of all known Azure SKUs, with subscription specific-restrictions. Then the max cache size is basically static for all subs at the original 10MB, since the rest will be a much smaller map[string]Restriction{} with details on which subs restrict some resources.

Given this is already xxl and gets some decent improvement for other features, I can iterate

cloud/services/networkinterfaces/service.go
alexeldeib: actually, because we use the constrained interface here, we can't pluck it from the scope

alexeldeib: I think that's fine though, I'd rather be explicit passing it down

alexeldeib:
re-requesting review because I think i'd leave it like this on the first pass

cloud/services/resourceskus/cache.go
```diff
@@ -0,0 +1,242 @@
...
+
+// NewStaticCache initializes a cache with data and no ability to refresh. Used for testing.
+func NewStaticCache(data []compute.ResourceSku) *Cache {
+	return &Cache{
+		data: data,
+	}
+}
+
+func (c *Cache) refresh(ctx context.Context, location string) error {
+	data, err := c.client.List(ctx, fmt.Sprintf("location eq '%s'", location))
```
devigned, line 73: 👍

devigned, line 73: 👍

cloud/services/resourceskus/cache_test.go
```diff
@@ -0,0 +1,535 @@
...
+					Name: to.StringPtr("other"),
+				},
+			},
+			err: "resource sku with name 'foo' and category 'bar' not found",
+		},
+	}
+
+	for name, tc := range cases {
+		t.Run(name, func(t *testing.T) {
+			// t.Parallel()
```
devigned, line 64: remove?

cloud/services/resourceskus/cache_test.go
```diff
@@ -0,0 +1,535 @@
...
+					},
+				},
+			},
+			want: nil,
+		},
+	}
+
+	for name, tc := range cases {
+		t.Run(name, func(t *testing.T) {
+			// t.Parallel()
```
devigned, line 199: remove?

cloud/services/resourceskus/cache_test.go
```diff
@@ -0,0 +1,535 @@
...
```
devigned, line 339: remove?

cloud/services/resourceskus/cache_test.go
```diff
@@ -0,0 +1,535 @@
...
+				},
+			},
+			want: nil,
+		},
+	}
+
+	for name, tc := range cases {
+		tc := tc
+		t.Run(name, func(t *testing.T) {
+			// t.Parallel()
```
devigned, line 485: remove?

cloud/services/resourceskus/cache_test.go
CecileRobertMichon: make sure to add `tc := tc` above if you uncomment (see #806, we were missing this in a bunch of places)

cloud/scope/cluster.go
```diff
@@ -37,12 +39,12 @@ type ClusterScopeParams struct {
...
 	Cluster      *clusterv1.Cluster
 	AzureCluster *infrav1.AzureCluster
-	Context      context.Context
+	NewSKUCache  resourceskus.NewCacheFunc
 }
 
 // NewClusterScope creates a new Scope from the supplied parameters.
 // This is meant to be called for each reconcile iteration.
-func NewClusterScope(params ClusterScopeParams) (*ClusterScope, error) {
+func NewClusterScope(ctx context.Context, params ClusterScopeParams) (*ClusterScope, error) {
```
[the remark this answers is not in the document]
CecileRobertMichon: @alexeldeib any thoughts on this one?

CecileRobertMichon:
How close do you think this PR is from making it in? I'd like to be able to do the VM refactor on top of it since it simplifies the AZ stuff. Happy to do a round of review tonight/tomorrow if it helps.

alexeldeib:
I was waiting on review for this one, I can move around the contexts and rebase.

cloud/scope/cluster.go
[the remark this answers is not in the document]
alexeldeib: fixed, honestly it's a bit uglier because background refresh isn't here yet

still no concurrent access so no mutex required even for data mutation

alexeldeib:
lmk if that solution to context looks okay. should be good pending CI

cloud/services/networkinterfaces/service.go
```diff
@@ -41,18 +41,18 @@ type Service struct {
 	LoadBalancersClient   loadbalancers.Client
 	PublicIPsClient       publicips.Client
 	InboundNATRulesClient inboundnatrules.Client
-	ResourceSkusClient    resourceskus.Client
+	ResourceSKUCache      *resourceskus.Cache
 }
 
 // NewService creates a new service.
-func NewService(scope NICScope) *Service {
+func NewService(scope NICScope, skuCache *resourceskus.Cache) *Service {
```
CecileRobertMichon, line 48: why not make SKUCache part of machine scope just like it's added to cluster scope already and add `ResourceSKUCache` to `NICScope`?

CecileRobertMichon: we can't that with VMs and VMSS yet because they haven't been refactored but eventually we'll be able to use that for all of them (also they won't have access to clusterScope anymore after the refactor)

CecileRobertMichon: another option is to add it to clusterDescriber and use the cluster scope SKUCache but I don't think I like that as much.

alexeldeib: IMO we should construct clients once and inject them into services from the reconciler to reduce duplication, instead of constructing new clients for every service.

The natural extension of what you described to solve that problem ends up with us putting every service in one of the scopes, which feels wrong. I'd rather see us invert control, instantiate all clients during reconcile, and then pass only the required ones down to the services which need them. 

I don't mind changing it back since I'm not tackling the rest right now, but I do think we should move towards passing these down rather than instantiating them everywhere.

LMK what you think

CecileRobertMichon: > https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/783/files#diff-7c0af1847b5ee759aadb398ac16c4d36R83

Completely agree.

I'm not sure how what I'm suggesting doesn't accomplish that though. I'm saying we shouldn't add a new dependency on clusterScope in the azure machine reconciler as we're moving towards getting rid of those https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/783/files#diff-fb5deef0b083d17ab538518df2570d4bR56.

What about adding a new interface to MachineScope, similar to ClusterDescriber, for the SKUCache and passing the clusterScope SKU cache to the MachineScopeParams when we create a newMachineScope? That way it's only created once and all services use the same SKU cache but we can explicitly define which service scopes need it and mock it.

alexeldeib: It's in the same vein as https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/798#discussion_r456109279 which I just ran into

I think all of the construction of the service/clients and their params can be pulled up a layer. Construct all of the clients, inject them, and then for each object to be reconciled, produce the spec struct for the desired object in the top level reconciler, and pass it to the service.

This way, each service is only a wrapper for Azure clients, and the functions themselves become a bit more "pure" and simpler to test directly. Like I said, definitely bigger change than this PR

alexeldeib: weird comment timing -- maybe i'm just overthinking it, I can try and see

alexeldeib: how's that? bit of weirdness with references to cache at different layers, but at least it removes the dep on scope everywhere

CecileRobertMichon: I think it's fine for now, might play around with it a bit with the VM refactor.

CecileRobertMichon:
I'm good with this change as-is. @devigned did you have any additional comments?

PR #786, opened 2020-07-13
🔧 Fix default tilt variables assignment

CecileRobertMichon:
**What this PR does / why we need it**: Fixes a regression from #781 where the default variables were never set in the flavor templates. 

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -204,7 +204,8 @@ def flavors():
 
     substitutions = settings.get("kustomize_substitutions", {})
     for key in keys:
-        substitutions[key[:-4]] = base64_decode(substitutions[key])
+        if key[-4:] == "_B64":
+            substitutions[key[:-4]] = base64_decode(substitutions[key])
 
     ssh_pub_key = "AZURE_SSH_PUBLIC_KEY"
     ssh_pub_key_path = "~/.ssh/id_rsa.pub"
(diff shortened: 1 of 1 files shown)


Tiltfile
```diff
@@ -204,7 +204,8 @@ def flavors():
 
     substitutions = settings.get("kustomize_substitutions", {})
     for key in keys:
-        substitutions[key[:-4]] = base64_decode(substitutions[key])
+        if key[-4] == "_B64":
```
CecileRobertMichon, line 207: AZURE_ENVIRONMENT and other not B64 vars shouldn't be base64 decoded

Tiltfile
```diff
@@ -230,6 +231,27 @@ def deploy_worker_templates(flavor, substitutions):
 
     yaml = str(read_file(yaml_file))
 
+    # azure account and ssh replacements
+    for substitution in substitutions:
```
CecileRobertMichon, line 235: we need to do this to set AZURE_SUBSCRIPTION and AZURE_SSH_KEY

Tiltfile
```diff
@@ -230,6 +231,27 @@ def deploy_worker_templates(flavor, substitutions):
...
+        value = substitutions[substitution]
+        yaml = yaml.replace("${" + substitution + "}", value)
+
+    # if metadata defined for worker-templates in tilt_settings
+    if "worker-templates" in settings:
```
CecileRobertMichon, line 240: the order here matters because otherwise `substitutions = settings.get("worker-templates").get("flavors").get(flavor, {})` overrides what was previously set from defaults with `substitutions[k] = defaults[k]` without ever replacing the defaults in the yaml template.

Tiltfile
```diff
@@ -250,22 +272,6 @@ def deploy_worker_templates(flavor, substitutions):
     az_json = azure_json(flavor, substitutions)
     substitutions["AZURE_JSON_B64"] = base64_encode(az_json)
```
CecileRobertMichon, line 273: I still need to fix this so it doesn't override a custom `AZURE_JSON_B64`

CecileRobertMichon:
Screen Shot 2020-07-13 at 12 07 45 PM

(user-assigned failed because I didn't provide an identity)

Issue #519, opened 2020-04-10
Fail fast when quota is not sufficient


sozercan:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]
When creating a cluster in a region where subscription doesn't have enough quota, it doesn't fast fail but keeps reconciling

**What did you expect to happen:**
It should fast fail after checking whether it has enough quota in the sub

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- cluster-api-provider-azure version:  
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

shanalily:
/unassign

Discussed with @ritazh and others.

CecileRobertMichon:
@shanalily @ritazh, do you have any findings to share about this issue that could be useful for someone else when they pick up the issue that you want to share?

ritazh:
The team discussed what the user experience should be for this issue. 
Some options are:
1. If creation failed due to quota issue, then create a K8s event resource to bubble up the error. Continue to reconcile in case quota issue eventually resolves. [This seems to be the behavior for other cluster api providers as well. This is the one we landed on.]
2.  If creation failed due to quota issue, then create a K8s event resource to bubble up the error and skips reconciliation.
3.  If creation failed due to quota issue, then create a K8s event resource to bubble up the error. Continue to reconcile after some duration in case quota issue eventually resolves. [Not clear what this duration should be and where to derive it from.]

devigned:
To add a bit to @ritazh's comment above, just for posterity. We also discussed using [Compute Usage API](https://docs.microsoft.com/en-us/rest/api/compute/usage/list) to see if there are enough of a VM SKU available.

Might be useful for changes to MachineDeployment / MachinePool validation.

CecileRobertMichon:
Option 1) seems like the best way to handle this to me as well.

We might also want to bubble up the error (and other errors) to AzureMachine status as well, right now AzureCluster and AzureMachine status don't show much. For future improvement, it would be nice to add a `Phase` similar to Cluster/Machine status.

CecileRobertMichon:
opened https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/524

CecileRobertMichon:
@ritazh are you working on this? I see you're still assigned.

ritazh:
[...]

PR #789, opened 2020-07-14
⚠ add variable defaults to templates

nader-ziada:
 - add variable defaults to templates
 - updates docs

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
default values added for variables in manifests and templates
action required :  creation of clusters now requires the use of clusterctl or envsubst to take advantage of default values
```

### DIFF
--- a/config/default/cloud_environment.yaml
+++ b/config/default/cloud_environment.yaml
@@ -4,4 +4,5 @@ metadata:
   name: manager-bootstrap-cloud-env
   namespace: system
 data:
-  cloud-env: ${AZURE_ENVIRONMENT}
+  cloud-env: ${AZURE_ENVIRONMENT:-"AzurePublicCloud"}
+  
--- a/templates/cluster-template-external-cloud-provider.yaml
+++ b/templates/cluster-template-external-cloud-provider.yaml
@@ -26,8 +26,8 @@ spec:
   location: ${AZURE_LOCATION}
   networkSpec:
     vnet:
-      name: ${AZURE_VNET_NAME}
-  resourceGroup: ${AZURE_RESOURCE_GROUP}
+      name: ${AZURE_VNET_NAME:=$CLUSTER_NAME-vnet}
+  resourceGroup: ${AZURE_RESOURCE_GROUP:=$CLUSTER_NAME}
   subscriptionID: ${AZURE_SUBSCRIPTION_ID}
 ---
 apiVersion: controlplane.cluster.x-k8s.io/v1alpha3
--- a/templates/flavors/aks/cluster-template.yaml
+++ b/templates/flavors/aks/cluster-template.yaml
@@ -29,7 +29,7 @@ metadata:
   name: ${CLUSTER_NAME}-control-plane
 spec:
   subscriptionID: ${AZURE_SUBSCRIPTION_ID}
-  resourceGroup: "${AZURE_RESOURCE_GROUP}"
+  resourceGroup: "${AZURE_RESOURCE_GROUP:=$CLUSTER_NAME}"
   location: "${AZURE_LOCATION}"
   defaultPoolRef:
     name: "agentpool0"
--- a/templates/flavors/base/cluster-template.yaml
+++ b/templates/flavors/base/cluster-template.yaml
@@ -21,12 +21,12 @@ kind: AzureCluster
 metadata:
   name: ${CLUSTER_NAME}
 spec:
-  resourceGroup: "${AZURE_RESOURCE_GROUP}"
+  resourceGroup: "${AZURE_RESOURCE_GROUP:=$CLUSTER_NAME}"
   location: "${AZURE_LOCATION}"
   subscriptionID: ${AZURE_SUBSCRIPTION_ID}
   networkSpec:
     vnet:
-      name: "${AZURE_VNET_NAME}"
+      name: "${AZURE_VNET_NAME:=$CLUSTER_NAME-vnet}"
 ---
 kind: KubeadmControlPlane
 apiVersion: controlplane.cluster.x-k8s.io/v1alpha3
--- a/docs/development.md
+++ b/docs/development.md
@@ -282,7 +282,7 @@ export AZURE_SSH_PUBLIC_KEY=$(cat "${SSH_KEY_FILE}.pub" | base64 | tr -d '\r\n')
 
 # To populate secret in azure.json file.
 export AZURE_JSON_B64=$(echo '{
-    "cloud": "${AZURE_ENVIRONMENT}",
+    "cloud": "${AZURE_ENVIRONMENT:="AzurePublicCloud"}",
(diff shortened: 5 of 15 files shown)


Tiltfile
```diff
@@ -9,7 +9,7 @@ settings = {
     "deploy_cert_manager": True,
     "preload_images_for_kind": True,
     "kind_cluster_name": "capz",
-    "capi_version": "v0.3.7-rc.1",
+    "capi_version": "v0.3.7",
```
nader-ziada, line 12: made the change to upgrade capi version here so I can use the new clusterctl, will rebase when its merged in

templates/cluster-template-external-cloud-provider.yaml
```diff
@@ -26,8 +26,8 @@ spec:
   location: ${AZURE_LOCATION}
   networkSpec:
     vnet:
-      name: ${AZURE_VNET_NAME}
-  resourceGroup: ${AZURE_RESOURCE_GROUP}
+      name: ${AZURE_VNET_NAME:=$CLUSTER_NAME-vnet}
+  resourceGroup: ${AZURE_RESOURCE_GROUP:=$CLUSTER_NAME-resource-group}
```
nader-ziada, line 30: I assumed this is useful, but looking for feedback

config/default/cloud_environment.yaml
```diff
@@ -4,4 +4,4 @@ metadata:
   name: manager-bootstrap-cloud-env
   namespace: system
 data:
-  cloud-env: ${AZURE_ENVIRONMENT}
\ No newline at end of file
+  cloud-env:  ${AZURE_ENVIRONMENT:-"AzurePublicCloud"}
```
CecileRobertMichon, line 7:
```suggestion
  cloud-env: ${AZURE_ENVIRONMENT:-"AzurePublicCloud"}
  
```

templates/cluster-template-external-cloud-provider.yaml
CecileRobertMichon: I think in other docs/examples we've mostly used just `$CLUSTER_NAME` as the default rg name (eg. https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/789/files#diff-1bf07101845de6da98f1c948b82c5ce0L290)

I like the idea of just requiring a `CLUSTER_NAME` for the quickstart but we need to be careful that the AZURE_JSON config matches the defaults in templates, ie. https://github.com/kubernetes-sigs/cluster-api/pull/3303/files#diff-dccb64095f5c5a09d32f119446301d12R386

nader-ziada: maybe the default value should be just the cluster name so its matches the AZURE_JSON_B64 
${AZURE_RESOURCE_GROUP:=$CLUSTER_NAME}
${AZURE_VNET_NAME:=$CLUSTER_NAME}

at the least the resource group we can do that for

nader-ziada: changed the resource group to be just the cluster name `resourceGroup: ${AZURE_RESOURCE_GROUP:=$CLUSTER_NAME}`

CecileRobertMichon:
Can you please add "action required" to the release note so it's marked as breaking in the notes?

lgtm

/assign @devigned for another pair of 👀

nader-ziada:
rebased, PTAL @CecileRobertMichon @devigned

CecileRobertMichon:
> rebased, PTAL @CecileRobertMichon @devigned

race condition 🐎

nader-ziada:
@CecileRobertMichon I added the action required, is there any special format for it?

templates/flavors/aks/cluster-template.yaml
```diff
@@ -29,7 +29,7 @@ metadata:
   name: ${CLUSTER_NAME}-control-plane
 spec:
   subscriptionID: ${AZURE_SUBSCRIPTION_ID}
-  resourceGroup: "${AZURE_RESOURCE_GROUP}"
+  resourceGroup: "${AZURE_RESOURCE_GROUP:=$CLUSTER_NAME}"
   location: "${AZURE_LOCATION}"
```
devigned, line 33: What do folks think about a location default? (applies to all the rest of the flavors too)

templates/flavors/base/cluster-template.yaml
```diff
@@ -21,12 +21,12 @@ kind: AzureCluster
 metadata:
   name: ${CLUSTER_NAME}
```
devigned, line 22: Could we default cluster name?

templates/flavors/aks/cluster-template.yaml
nader-ziada: that sounds good to me

templates/flavors/base/cluster-template.yaml
nader-ziada: don't know about that one :)

templates/flavors/aks/cluster-template.yaml
CecileRobertMichon: the reason I see for not doing that is that location does require you to have quota in that region so making the user set it is a good way for them to choose a location where they can actually deploy VMs... Otherwise it's very easy to shoot yourself in the foot as a new user and get a quota failure on your first deployment (which right now isn't handled very gracefully -- see https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/519)

templates/flavors/base/cluster-template.yaml
CecileRobertMichon: I'm split on that one... On one end it does allow a way easier "first cluster" experience. BUT each cluster name should be unique (so the Azure resources that use cluster name are also named uniquely) so your second cluster would fail, and it also requires syncing that with the cluster name you give to `clusterctl config cluster <cluster-name` in the quickstart.

templates/flavors/aks/cluster-template.yaml
devigned: I'm good with the vars that are currently defaulted. I just wanted to push a bit to how aggressive / conservative we think we should be.

templates/flavors/base/cluster-template.yaml
nader-ziada: I think its useful for the user to set the CLUSTER_NAME so when/if they are creating multiple clusters they know where the name came from, also helps them find their resources on Azure. It would feel like too much magic to me

CecileRobertMichon: How about we start soft, and increment gradually to more magic later?

templates/flavors/aks/cluster-template.yaml
alexeldeib: I'd find location defaulting somewhat annoying, personally -- I'd rather get validation failure than accidentally create a cluster in the wrong region

PR #792, opened 2020-07-15
💎 Refactor RouteTable service

nader-ziada:
following the pattern discussed in issue #757

 

- Applies the new Service interface to RouteTables

Partially addresses #757

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Refactor RouteTable service
```


cloud/scope/cluster.go
```diff
@@ -165,6 +173,11 @@ func (s *ClusterScope) NodeSubnet() *infrav1.SubnetSpec {
 	return s.AzureCluster.Spec.NetworkSpec.GetNodeSubnet()
 }
 
+// RouteTable returns the cluster node routetable.
```
CecileRobertMichon, line 176: How do you feel about also fixing https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/718 as part of this PR?

nader-ziada: I was going to fix that next, but will add it here as a 2nd commit

nader-ziada: there is a comment into the issue saying `I am not 100% sure it makes sense to have two different route tables. Need to look into it more.` 

do you think its valid to have two routetables?

CecileRobertMichon: as discussed on slack, requirements aren't clear yet so let's merge this refactor as-is and open a separate PR to fix the bug which is pre-existing

cloud/scope/cluster.go
```diff
@@ -145,6 +146,13 @@ func (s *ClusterScope) LBSpecs() []azure.LBSpec {
 	}
 }
 
+// RouteTableSpec returns the node route table
+func (s *ClusterScope) RouteTableSpec() azure.RouteTableSpec {
```
CecileRobertMichon, line 150: for consistency with the other services, and because it's likely we'll need more than one route table in the near future, let's make this 
```suggestion
func (s *ClusterScope) RouteTableSpecs() []azure.RouteTableSpec {
```

And adjust the Reconcile/Delete() functions to loop over the specs

cloud/services/routetables/routetables.go
```diff
@@ -25,78 +25,69 @@ import (
...
-	}
-	if err != nil {
-		return errors.Wrapf(err, "failed to delete route table %s in resource group %s", routeTableSpec.Name, s.Scope.ResourceGroup())
-	}
+	for _, rtSpec := range s.Scope.RouteTableSpecs() {
+		s.Scope.V(2).Info("deleting route table", "route table", rtSpec.Name)
+		err := s.Client.Delete(ctx, s.Scope.ResourceGroup(), rtSpec.Name)
+		if err != nil && azure.ResourceNotFound(err) {
+			// already deleted
+			return nil
```
CecileRobertMichon, line 84: this needs to `continue` instead so it deletes the other route tables in the list

nader-ziada: fixed

PR #798, opened 2020-07-16
🐛 fix AzureManagedCluster reconcile without a resource group

### LINKED ISSUE #791: AzureManagedCluster do not reconcile a resource group
/kind bug

When an AzureManagedCluster reconciles, it currently does not reconcile a resource group. This will cause the cluster to fail provisioning if the resource group does not already exist.

AzureManagedCluster and descendant infra should provision if the resource group does not already exist.

**Environment:**

devigned:
When deploying an `AzureManagedCluster`, the current expectation is that the resource group already exists. This is not always the case, so we need to reconcile the resource group with the `AzureManagedControlPlane`.

Along the way here, I found a handful of little issues which I've cleaned up.

Fixes #791 

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
fix reconciliation of `AzureManagedCluster` when the resource group does not exist
**Action Required:**  
- `AzureManagedControlPlaneSpec.AdditionalTags` has change to be `type Tags map[string]string`
- `AzureManagedControlPlaneSpec.SSHPublicKey` is expected to be a base64 encoded string
```


exp/api/v1alpha3/azuremanagedcontrolplane_types.go
```diff
@@ -45,7 +47,7 @@ type AzureManagedControlPlaneSpec struct {
 	// AdditionalTags is an optional set of tags to add to Azure resources managed by the Azure provider, in addition to the
 	// ones added by default.
 	// +optional
-	AdditionalTags map[string]string `json:"additionalTags,omitempty"`
+	AdditionalTags infrav1.Tags `json:"additionalTags,omitempty"`
```
devigned, line 50: This is a breaking change to the API :(.

exp/api/v1alpha3/azuremanagedcontrolplane_types.go
```diff
@@ -55,7 +57,7 @@ type AzureManagedControlPlaneSpec struct {
 	// +kubebuilder:validation:Enum=Calico;Azure
 	NetworkPolicy *string `json:"networkPolicy,omitempty"`
 
-	// SSHPublicKey is a string literal containing an ssh public key.
+	// SSHPublicKey is a string literal containing an ssh public key base64 encoded.
```
devigned, line 60: This is a breaking change to the API, but not one that is a type violation. Rather, this is a behavioral breaking change. To bring the API in line with VM and VMSS, we now ask that SSH pub keys be provided base64 encoded. Unfortunately, the APIs for VM / VMSS and AKS differ in that the former requires base64 encoding while the later take a non-encoded pub key.

cloud/services/groups/service.go
```diff
@@ -30,7 +31,11 @@ type Service struct {
 // GroupScope defines the scope interface for a disk service.
 type GroupScope interface {
 	logr.Logger
-	azure.ClusterDescriber
```
devigned, line 33: This is too broad of an interface to expect here.

exp/api/v1alpha3/azuremanagedcontrolplane_types.go
CecileRobertMichon: Let's mark these as "action required" in the release note. I think it's acceptable to have breaking changes at this point given that it's 1) an alpha api 2) an experimental feature

exp/controllers/azuremanagedmachinepool_reconciler.go
```diff
@@ -29,17 +29,38 @@ import (
...
+
+	// NodeLister is a service interface for returning generic lists.
+	NodeLister interface {
+		ListInstances(context.Context, string, string) ([]compute.VirtualMachineScaleSetVM, error)
+		List(context.Context, string) ([]compute.VirtualMachineScaleSet, error)
+	}
+)
+
+// NewAgentPoolVMSSNotFoundError creates a new AgentPoolVMSSNotFoundError
+func NewAgentPoolVMSSNotFoundError(nodeResourceGroup, poolName string) *AgentPoolVMSSNotFoundError {
```
alexeldeib, line 54: nice

exp/controllers/azuremanagedmachinepool_reconciler.go
```diff
@@ -128,3 +149,9 @@ func (r *azureManagedMachinePoolReconciler) Delete(ctx context.Context, scope *s
 
 	return nil
 }
+
+// IsAgentPoolVMSSNotFoundError returns true if the error is a AgentPoolVMSSNotFoundError
+func IsAgentPoolVMSSNotFoundError(err error) bool {
+	_, ok := err.(*AgentPoolVMSSNotFoundError)
```
alexeldeib, line 155: go1.13 errors.Is?

devigned: You are right. I think I was just being lazy.

cloud/services/groups/service.go
CecileRobertMichon: I agree. I think this applies to some of the other services as well though. Maybe we should have a more focused interface for these things that every single service needs to know about and then have a broader one that includes the Network stuff that gets passed into the machine scope.

exp/controllers/azuremanagedmachinepool_controller.go
```diff
@@ -168,6 +168,13 @@ func (r *AzureManagedMachinePoolReconciler) Reconcile(req ctrl.Request) (_ ctrl.
 		return reconcile.Result{}, errors.Errorf("failed to create scope: %+v", err)
 	}
 
+	// Always patch when exiting so we can persist changes to finalizers and status
+	defer func() {
```
devigned, line 172: We were not making sure to patch the `AzureManagedMachinePool` at the end of reconciliation, thus they would never get to ready.

cloud/services/groups/service.go
CecileRobertMichon: See https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/824

@devigned do you mind reverting this change for now so we can deal with it in all services at once with #824?

cloud/scope/managedcontrolplane.go
```diff
@@ -94,6 +97,33 @@ type ManagedControlPlaneScope struct {
...
+}
+
+func (s *ManagedControlPlaneScope) Location() string {
+	if s.ControlPlane == nil {
+		return ""
+	}
+	return s.ControlPlane.Spec.Location
+}
+
+// AdditionalTags returns AdditionalTags from the scope's AzureCluster.
```
CecileRobertMichon, line 118:
```suggestion
// AdditionalTags returns AdditionalTags from the ControlPlane spec.
```

cloud/services/groups/service.go
devigned: That means I'll have to add a bunch of superfluous funcs to the managed control plane scope. Which is better?

CecileRobertMichon: Ok in that case let's just keep the change, will work on top of it

exp/controllers/azuremangedcontrolplane_reconciler.go
```diff
@@ -18,39 +18,49 @@ package controllers
...
+	"sigs.k8s.io/cluster-api-provider-azure/cloud/services/groups"
+	"sigs.k8s.io/cluster-api-provider-azure/cloud/services/managedclusters"
+	infrav1exp "sigs.k8s.io/cluster-api-provider-azure/exp/api/v1alpha3"
 )
 
 // azureManagedControlPlaneReconciler are list of services required by cluster controller
 type azureManagedControlPlaneReconciler struct {
 	kubeclient         client.Client
 	managedClustersSvc *managedclusters.Service
+	groupsSvc          *groups.Service
```
CecileRobertMichon, line 45: why is this not azure.Service?

devigned: It should be. Great catch, ty.

exp/controllers/azuremanagedmachinepool_controller.go
alexeldeib: nice catch, I think I hit this testing #822 but didn't figure out why they weren't becoming ready.

cloud/services/groups/service.go
devigned: done and done

exp/controllers/azuremanagedmachinepool_reconciler.go
devigned: Fixed. Implemented `Is`.

cloud/scope/managedcontrolplane.go
devigned: fixed

exp/controllers/azuremangedcontrolplane_reconciler.go
devigned: fixed

devigned:
@CecileRobertMichon, @alexeldeib I think this is ready

I'll have a follow up with e2e tests.

exp/api/v1alpha3/azuremanagedcontrolplane_types.go
CecileRobertMichon: This might actually not be a breaking change from a yaml spec perspective? since infrav1.Tags is defined as `type Tags map[string]string`

exp/controllers/azuremanagedmachinepool_reconciler.go
```diff
@@ -54,23 +85,30 @@ func newAzureManagedMachinePoolReconciler(scope *scope.ManagedControlPlaneScope)
 // Reconcile reconciles all the services in pre determined order
 func (r *azureManagedMachinePoolReconciler) Reconcile(ctx context.Context, scope *scope.ManagedControlPlaneScope) error {
 	scope.Logger.Info("reconciling machine pool")
+	var normalizedVersion *string
```
CecileRobertMichon, line 88: is this the same change as https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/818/files? If so let's drop it here and get it in separately with the other PR

devigned: ok, let's pull in #818 and I'll rebase on top of it.

cloud/services/groups/service.go
alexeldeib: fwiw I think constraining the scope at site of use is more correct, network describer should only go into the top level machine/cluster scope (or lower level scopes that require all components)

exp/api/v1alpha3/azuremanagedcontrolplane_types.go
alexeldeib: Still should mark it as action required just in case; would technically be a break for anyone importing the types (likely N=0, in fairness).

alexeldeib:
looks good pending rebase now that #818 is in

exp/controllers/azuremanagedmachinepool_reconciler.go
devigned: rebased

devigned:
Rebased on master. Should be ready for another look.

Tiltfile
```diff
@@ -151,11 +153,11 @@ COPY manager .
 # Build CAPZ and add feature gates
 def capz():
     # Apply the kustomized yaml for this provider
-    yaml = str(kustomizesub("./config"))
     substitutions = settings.get("kustomize_substitutions", {})
     for substitution in substitutions:
-        value = substitutions[substitution]
-        yaml = yaml.replace("${" + substitution + "}", value)
+        os.putenv(substitution, substitutions[substitution])
```
alexeldeib, line 158: nit: https://github.com/kubernetes-sigs/cluster-api/pull/3378/files#diff-1a474bf970407efc3332970120c1a429R197

`os.environ.update(substitutions)`

exp/controllers/azuremanagedmachinepool_reconciler.go
```diff
@@ -108,9 +139,9 @@ func (r *azureManagedMachinePoolReconciler) Reconcile(ctx context.Context, scope
 	}
 
 	var providerIDs = make([]string, len(instances))
-	for _, vm := range instances {
+	for i, vm := range instances {
 		vm := vm
```
alexeldeib, line 143: pretty sure I put this here, but is it actually necessary in this instance? though this mattered more for closures/goroutines

alexeldeib: might as well use `*instances[i].ID` since we already need `i`, then can drop vm from the loop. 

just style _shrugs_

devigned: fixed

Tiltfile
devigned: 👍

exp/controllers/azuremanagedmachinepool_controller.go
```diff
@@ -188,6 +195,12 @@ func (r *AzureManagedMachinePoolReconciler) reconcileNormal(ctx context.Context,
 	}
 
 	if err := newAzureManagedMachinePoolReconciler(scope).Reconcile(ctx, scope); err != nil {
+		if IsAgentPoolVMSSNotFoundError(err) {
+			// if the underlying VMSS is not yet created, requeue for 30s in the future
+			return reconcile.Result{
+				RequeueAfter: 30 * time.Second,
```
alexeldeib, line 201: why 30s here instead of allowing exponential backoff? is this to partially account for https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/847?

alexeldeib:
/lgtm

last comment is fine either way with me

exp/controllers/azuremanagedmachinepool_controller.go
devigned: The duration is arbitrary, but the idea was to give it some time and try again to allow the AzureManagedControlPlane a chance to create the AKS cluster.

devigned:
Thx for the lgtm, @alexeldeib. Unfortunately, it had a lint error. Should be good now.

exp/controllers/azuremangedcontrolplane_reconciler.go
```diff
@@ -93,10 +108,16 @@ func (r *azureManagedControlPlaneReconciler) Delete(ctx context.Context, scope *
...
 	}
 
-	if err := r.managedClustersSvc.Delete(ctx, managedClusterSpec); err != nil {
+	scope.V(2).Info("Deleting managed cluster")
+	if err := r.managedClustersSvc.Delete(ctx, managedClusterSpec); err != nil && !azure.ResourceNotFound(err) {
 		return errors.Wrapf(err, "failed to delete managed cluster %s", scope.ControlPlane.Name)
 	}
 
+	scope.V(2).Info("Deleting managed cluster resource group")
+	if err := r.groupsSvc.Delete(ctx); err != nil {
```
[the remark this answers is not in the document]
alexeldeib: seems reasonable to handle the 404 check inside the service imo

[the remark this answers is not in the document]
devigned: Indeed, that is what the Groups service does: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/928f0c492b8dccf78703b343c7bbe543a5861210/cloud/services/groups/groups.go#L72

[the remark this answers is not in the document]
CecileRobertMichon: oh I missed that was calling the group svc, read this as groups client, all good then!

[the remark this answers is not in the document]
alexeldeib: could we do the same for MC service above 🙂

CecileRobertMichon:
/lgtm cancel

per Ace's comment

exp/controllers/azuremangedcontrolplane_reconciler.go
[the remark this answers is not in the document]
devigned: Fixed :).

Issue #356, opened 2020-01-09
Cluster with NSG in separate RG fails to expose service


CecileRobertMichon:
/kind bug

Because of https://github.com/kubernetes-sigs/cloud-provider-azure/issues/281, we are currently unable to specify a `securityGroupResourceGroup` in `azure.json` and the cloud-provider expects the NSG to be in the same RG as the cluster RG. Once the issue above is fixed, we should add `securityGroupResourceGroup` to all `azure.json` examples.

/assign

CecileRobertMichon:
Add `securityGroupResourceGroup` field in cluster template Azure json so this supported starting with 1.18+ 

https://github.com/kubernetes/kubernetes/pull/87035

/unassign
/help
/good-first-issue

CecileRobertMichon:
@aramase are you still working on this?

aramase:
@CecileRobertMichon yeah, I'll raise the PR for it soon.

CecileRobertMichon:
@aramase I'm going to unassign you for now to give to opportunity to someone else to pick this up, feel free to assign yourself again if/when you're able to get to it

/unassign @aramase

Deeps-1989:
@CecileRobertMichon I can help make this change. Do I need to make change under templates/cluster-template-*.yaml. Will it be a top level field or under which field this new field be nested to?

CecileRobertMichon:
Hi @Deeps-1989, I'm actually going to assign @alexeldeib on this one since he's already working on #773 and they are closely related. 

/assign @alexeldeib 

Feel free to pick up any of the other good first issues https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22 and comment `/assign` to assign yourself

alexeldeib:
/remove-help

Should be easy to add this on top of https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/802

CecileRobertMichon:
This would also require adding NSG RG to the AzureCluster Spec since the nsg can be in its own RG as well

dbsanfte:
I've recently run into this trying to set up a hybrid k8s deployment in Azure. 

The state of this issue is a bit confusing: the code seems to be merged, but it isn't released?

Is this feature released to the 0.4.x Azure Cloud Provider yet? Apologies if I'm asking in the wrong place.

CecileRobertMichon:
[...]

Issue #773, opened 2020-07-09
Audit cloud provider config


CecileRobertMichon:
We should go through the values in Azure.json and evaluate if:

- the default values are sane according to the cloud provider documentation: https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/docs/cloud-provider-config.md
- Do we need to set the file for both nodes and control planes (I suspect it might just be needed for control planes)
- How to simplify the UX for setting AZURE_JSON_B64

alexeldeib:
Based on my current understanding, we do need this file on all nodes, but they may be different if desired (e.g., it doesn't appear to be required to use any credentials, msi or otherwise, on worker nodes).

Kubelet seems to require the cloud provider interface to fetch node addresses (maybe other things, that's all I went looking for): https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/cloudresource/cloud_request_manager.go#L91 As vmss in a single cluster may be in different subnets, I believe the subnet field may also be different, for example.

We can disable kubelet identity on workers when using instance metadata for the same purpose: https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/docs/cloud-provider-config.md#run-kubelet-without-azure-identity

---

KubeadmConfig requires azure json, Machine and KCP require KubeadmConfig. Kubeadm config references the secret by name. If the secret does not exist, or another error occurred trying to fetch it, we requeue: https://github.com/kubernetes-sigs/cluster-api/blob/49f88d511f9584cb37bd38522a869ceaea9f2242/bootstrap/kubeadm/controllers/kubeadmconfig_controller.go#L609-L614

---

One way to simplify this is have users only provide the credentials as a secret. They never see the expanded JSON. A quick and dirty approach would be to have AzureCluster take a credentials field, and on cluster reconcile we generate e.g. `azure-json-control-plane` secret. We could check for existence to avoid overwriting a user-provided custom value. We could also generate a different file for workers, watching AzureMachineTemplate events. A lot of the json fields can be derived -- vnet comes from AzureCluster directly, loadBalanceResourceGroup too. We can detect vmType from AzureMachine/AzureMachinePool.

Once we generate the file, the rest of the controllers should fall into order

nader-ziada:
I like the idea of the cluster control generating the azure.json file and creating the secret for it

alexeldeib:
- securityGroupName is used to add NSG rules when users create services type=LoadBalancer
[...]

PR #802, opened 2020-07-17
✨ automatically generate azure.json

alexeldeib:
Automatically generates azure-json for machines/machine templates. Unfortunately 1 will be generated for every machine template (or machine not associated to a template) in the current form. This means if a user manually provides the secret to override, we would create two secrets. However with this implementation there are very few scenarios a user would really need to override: the only one I really see is cloud provider rate limiting configuration, which we don't expose at all. (future work 🙂)

Added ~Also, I haven't actually added any differentiation yet between user-provided secret not matching expected.~

related to https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/773 but doesn't close, strictly speaking

ready for reviewz ~added tests, think we just need docs~ ~Probably need to add some unit tests and/or other validation, not 100% sure this does all the right things yet. Opening up for discussion since it's already quite big.~

## FYI 
squash time ~I sorted the commits nicely for your reviewing pleasure. I'll keep them sorted as such.~

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
automatically generate azure.json for machine templates, machines
```


CecileRobertMichon:
/retitle [WIP] ✨ automatically generate azure.json

just marking this as WIP so it doesn't merge as per PR pr description

alexeldeib:
I tried doing this a few different ways, I found it worked best to break it out into a separate controller. For MachineDeployment/MachineSet/KubeadmControlPlane, all machines come from AzureMachineTemplate clones. For these machines we can share 1 secret for all machines in the group, it's wasteful to create 1 secret per machine. We do also need to allow for creating single machines which don't belong to any of those object types -- we filter with a predicate and check that the individual machine isn't a member of any of those resource types listed above which can share common azure json. We also add the associated object as an owner of the newly created secret, to get garbage collection

curious what people think!

alexeldeib:
@CecileRobertMichon thanks 👍 although this is WIP for unit tests, I do think it's quite close to functional completion. I'd appreciate reviews as-is on the approach anyway 🙂

controllers/helpers.go
```diff
@@ -113,3 +121,244 @@ func GetObjectsToRequestsByNamespaceAndClusterName(ctx context.Context, c client
...
+		}
+		nodeConfig, err = servicePrincipalNodeCloudProviderConfig(d)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	secret.Data = map[string][]byte{
+		"controlplane": controlPlaneConfig,
+		"node":         nodeConfig,
```
CecileRobertMichon, line 188: what are your thoughts on potentially needing to split up the azure.json for node into one for each machine deployment / machine pool in the future? Thinking of use cases where one of the configurations would change per node pool such as having multiple subnets

alexeldeib: that's what this does already, it just always generates both controlplane and node versions for even when the MD can only be one or the other. we differentiate by referencing the desired key in the MD kubeadm config

alexeldeib: I guess it's pretty easy to use the control plane label to tell them apart and only use one key

alexeldeib: we don't actually plumb the per-machinepool subnet through right now because we get it from the cluster spec. But if we switched that to e.g. the machine spec, we could easily adjust the interfaces/params here to make it work.

alexeldeib: sorry for stream of consciousness -- I think I remembered the reason I didn't differentiate. 

For machines, you can tell if they're control plane, machine set, or machine deployment based. If it's not any of those (i.e., a user created a standalone machine for some reason), you want the catch that and create a single-machine azure json.

For machine templates, the cluster is the owner. We would have to go back to the cluster, get the control plane ref, fetch that object, and check if the control plane uses this machine template or not. BUT there's not really a strict guarantee that the infra ref is KCP, so we can't use our knowledge of KCP schema to look up the azure machine template spec. 

The other alternative that would allow is creating azure-json per control plane machine instead of per-machine template for the control plane. That's strictly worse as we end up generating N times as many secrets as required (where N == control plane replicas), vs duplicating the data one time inside the secret data map.

alexeldeib:
need to cover machine pool

cloud/scope/clients.go
```diff
@@ -37,40 +38,68 @@ const (
 
 // AzureClients contains all the Azure clients used by the scopes.
 type AzureClients struct {
```
alexeldeib, line 40: I think this could use a rename and maybe some refactoring, the setCredentials func is a bit funky to me. out of scope for this pr

alexeldeib:
--

alexeldeib:
@CecileRobertMichon @devigned think this is ready for review

cloud/interfaces.go
```diff
@@ -70,4 +70,8 @@ type ClusterDescriber interface {
 	NodeSubnet() *infrav1.SubnetSpec
 	ControlPlaneSubnet() *infrav1.SubnetSpec
 	RouteTable() *infrav1.RouteTable
+	ClientID() string
```
CecileRobertMichon, line 73: Should this go in Authorizer above instead?

controllers/azurejson_machine_controller.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+
+	_, isControlPlaneNode := e.Meta.GetLabels()[clusterv1.MachineControlPlaneLabelName]
+	_, isMachineSetNode := e.Meta.GetLabels()[clusterv1.MachineSetLabelName]
+	_, isMachineDeploymentNode := e.Meta.GetLabels()[clusterv1.MachineDeploymentLabelName]
+
+	return !isControlPlaneNode && !isMachineSetNode && !isMachineDeploymentNode
+}
+
+// Reconcile reconciles the azure json for a specific machine not in a machine deployment
```
CecileRobertMichon, line 100: why does this need to be distinct from machine deployments? I'm sure there's a good reason but I don't think I'm completely following the logic

cloud/interfaces.go
alexeldeib: moved 👍

controllers/azurejson_machine_controller.go
alexeldeib: any machines created from KCP, machinset or machinedeployment will already have a secret created for them from the template controller. Those machines share 1 secret per template (all machines in the MD reuse the same secret). Machinepools don't use templates; they need a separate, unfiltered controller. 

if a user creates a machine one-off, without machineset or machinedeployment, we need to provision them an azure-json. This controller catches that case, while ignoring other machines.

alexeldeib: can add a comment to that effect

CecileRobertMichon:
@alexeldeib I really like where this PR is going. It is a big change though and it seems like there are a few open questions from your comments above. What would you think about doing an informal code/design review sometime this week?

alexeldeib:
sure, tomorrow or thursday would be easier for me (i'll be backup on-call friday). happy to grab some time either morning? I don't think there's too much from my perspective in design (besides the overall approach) now that I have a better understanding of what some of the fields are used for

controllers/helpers.go
```diff
@@ -113,3 +122,207 @@ func GetObjectsToRequestsByNamespaceAndClusterName(ctx context.Context, c client
...
+
+	log.Info("done updating azure json")
+
+	return nil
+}
+
+// GetOwnerMachinePool returns the MachinePool object owning the current resource.
+func GetOwnerMachinePool(ctx context.Context, c client.Client, obj metav1.ObjectMeta) (*capiv1exp.MachinePool, error) {
+	for _, ref := range obj.OwnerReferences {
+		if ref.Kind == "MachinePool" && ref.APIVersion == capiv1exp.GroupVersion.String() {
```
devigned, line 313: might want to drop the match on api version and just do Kind and Group

controllers/helpers.go
```diff
@@ -113,3 +122,207 @@ func GetObjectsToRequestsByNamespaceAndClusterName(ctx context.Context, c client
...
+	}
+
+	return secret, nil
+}
+
+func servicePrincipalCloudProviderConfig(d azure.ClusterDescriber) ([]byte, error) {
+	return json.MarshalIndent(newCloudProviderConfig(d), "", "    ")
+}
+
+func systemMSICloudProviderConfig(d azure.ClusterDescriber) ([]byte, error) {
```
devigned, line 188: I think MSI is old naming for these. What do you think about sticking with SystemAssigned and UserAssigned Managed Identities.

controllers/helpers_test.go
```diff
@@ -72,6 +77,61 @@ func TestAzureClusterToAzureMachinesMapper(t *testing.T) {
...
+	cases := map[string]struct {
+		identityType infrav1.VMIdentity
+		identityID   string
+		expect       string
+	}{
+		"serviceprincipal": {
+			identityType: infrav1.VMIdentityNone,
+			expect:       spCloudConfig,
+		},
+		"system-msi": {
```
devigned, line 92: Naming; "system-assigned"?

controllers/helpers_test.go
```diff
@@ -72,6 +77,61 @@ func TestAzureClusterToAzureMachinesMapper(t *testing.T) {
...
+			identityType: infrav1.VMIdentitySystemAssigned,
+			expect:       msiCloudConfig,
+		},
+		"user-msi": {
```
devigned, line 96: Naming; "user-assigned"?

controllers/helpers.go
CecileRobertMichon: +1, see #813

controllers/helpers.go
```diff
@@ -121,3 +129,215 @@ func GetObjectsToRequestsByNamespaceAndClusterName(ctx context.Context, c client
...
+	config.UseManagedIdentityExtension = true
+	config.UserAssignedIdentityID = identityID
+	return json.MarshalIndent(config, "", "    ")
+}
+
+func newCloudProviderConfig(d azure.ClusterDescriber) *CloudProviderConfig {
+	return &CloudProviderConfig{
+		Cloud:                        d.CloudEnvironment(),
+		AadClientID:                  d.ClientID(),
+		AadClientSecret:              d.ClientSecret(),
```
CecileRobertMichon, line 216: should we consider not putting the credentials on non-control plane nodes if they aren't used? based on https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/773#issuecomment-659082883

alexeldeib: I think we can remove almost all of the values on worker nodes, but I hadn't tested that yet.

alexeldeib: want me to try and tackle in this PR, or ok to slim down in follow up? I think it won't be too difficult, but I haven't double-checked the kubelet code for any edge cases like we hit with VMAS/VMSS/no zones

CecileRobertMichon: let's tackle in a follow up since the behavior in this PR is equivalent to current state of things

cloud/scope/clients.go
CecileRobertMichon: yeah also SubscriptionID should follow the same pattern as `tenantID` etc., ie. 
```
func (c *AzureClients) SubscriptionID() string {
	return c.subscriptionID
}
```

and remove the SubscriptionID() func in cluster.go on scope

controllers/helpers.go
```diff
@@ -121,3 +129,215 @@ func GetObjectsToRequestsByNamespaceAndClusterName(ctx context.Context, c client
...
+
+	bGV, err := schema.ParseGroupVersion(b.APIVersion)
+	if err != nil {
+		return false
+	}
+
+	return aGV.Group == bGV.Group && a.Kind == b.Kind && a.Name == b.Name
+}
+
+// GetCloudProviderSecret returns the required azure json secret for the provided parameters. It differentiates between
```
CecileRobertMichon, line 148: It differentiates between?

controllers/helpers.go
```diff
@@ -121,3 +129,215 @@ func GetObjectsToRequestsByNamespaceAndClusterName(ctx context.Context, c client
...
+	}
+
+	_, isGenerated := old.Labels[string(infrav1.ResourceLifecycleOwned)]
+
+	if !isGenerated {
+		log.Info("returning early from json reconcile, user provided secret already exists")
+		return nil
+	}
+
+	// Otherwise, check ownership and data fresheness. Update as necessary
```
CecileRobertMichon, line 283:
```suggestion
	// Otherwise, check ownership and data freshness. Update as necessary
```

templates/cluster-template-user-assigned-identity.yaml
```diff
@@ -177,7 +169,7 @@ spec:
       identity: UserAssigned
       location: ${AZURE_LOCATION}
       osDisk:
-        diskSizeGB: 30
+        diskSizeGB: 128
```
CecileRobertMichon, line 172: what happened here?

controllers/azurejson_machinetemplate_controller.go
```diff
@@ -0,0 +1,160 @@
...
+
+	apiVersion, kind := infrav1.GroupVersion.WithKind("AzureMachineTemplate").ToAPIVersionAndKind()
+	owner := metav1.OwnerReference{
+		APIVersion: apiVersion,
+		Kind:       kind,
+		Name:       azureMachineTemplate.GetName(),
+		UID:        azureMachineTemplate.GetUID(),
+	}
+
+	// Construct secret for this machine
```
CecileRobertMichon, line 137:
```suggestion
	// Construct secret for this machine template.
```

controllers/azurejson_machinepool_controller.go
```diff
@@ -0,0 +1,156 @@
...
+
+// SetupWithManager initializes this controller with a manager
+func (r *AzureJSONMachinePoolReconciler) SetupWithManager(mgr ctrl.Manager, options controller.Options) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&expv1.AzureMachinePool{}).
+		Owns(&corev1.Secret{}).
+		Complete(r)
+}
+
+// Reconcile reconciles the azure json for a specific machine not in a machine deployment
```
CecileRobertMichon, line 58: I think that doesn't apply here? should say machinepool?

controllers/azurejson_machinepool_controller.go
```diff
@@ -0,0 +1,156 @@
...
+		infrav1.VMIdentityNone,
+		"",
+	)
+
+	if err != nil {
+		return ctrl.Result{}, errors.Wrapf(err, "failed to create cloud provider config")
+	}
+
+	if err := reconcileAzureSecret(ctx, log, r.Client, owner, newSecret); err != nil {
+		return ctrl.Result{}, errors.Wrap(err, "failed to reconcile azure secret")
```
CecileRobertMichon, line 152: Could we maybe add an event on the corresponding machinepool/machine for the secret failure?

controllers/azurejson_machinetemplate_controller.go
```diff
@@ -0,0 +1,160 @@
...
+func (r *AzureJSONTemplateReconciler) SetupWithManager(mgr ctrl.Manager, options controller.Options) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		WithOptions(options).
+		For(&infrav1.AzureMachineTemplate{}).
+		Owns(&corev1.Secret{}).
+		Complete(r)
+}
+
+// Reconcile reconciles azure json secrets for azure machine templates
+func (r *AzureJSONTemplateReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, reterr error) {
```
CecileRobertMichon, line 60: A lot of this code seems duplicated between the 3 files, ideally we could reuse the same helpers for some of these in the future. That's probably related to what @devigned was talking about with the filters

controllers/helpers.go
```diff
@@ -121,3 +129,215 @@ func GetObjectsToRequestsByNamespaceAndClusterName(ctx context.Context, c client
...
+		AadClientID:                  d.ClientID(),
+		AadClientSecret:              d.ClientSecret(),
+		TenantID:                     d.TenantID(),
+		SubscriptionID:               d.SubscriptionID(),
+		ResourceGroup:                d.ResourceGroup(),
+		SecurityGroupName:            d.NodeSubnet().SecurityGroup.Name,
+		Location:                     d.Location(),
+		VMType:                       "vmss",
+		VnetName:                     d.Vnet().Name,
+		VnetResourceGroup:            d.Vnet().ResourceGroup,
```
CecileRobertMichon, line 224: Can you add the NSG group in there as well to fix https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/356?

cloud/scope/clients.go
```diff
@@ -37,40 +38,68 @@ const (
...
+
+	c.environment = settings.Values[auth.EnvironmentName]
+	if c.environment == "" {
+		c.environment = azure.PublicCloud.Name
+	}
+
 	c.ResourceManagerEndpoint = settings.Environment.ResourceManagerEndpoint
 	c.ResourceManagerVMDNSSuffix = GetAzureDNSZoneForEnvironment(settings.Environment.Name)
 	settings.Values[auth.SubscriptionID] = subscriptionID
+	settings.Values[auth.TenantID] = c.tenantID
```
CecileRobertMichon, line 97: why is this set here? Doesn't the value of c.tenantID come from `strings.TrimSuffix(settings.Values[auth.TenantID], "\n")` a few lines above?

controllers/helpers.go
```diff
@@ -121,3 +129,215 @@ func GetObjectsToRequestsByNamespaceAndClusterName(ctx context.Context, c client
...
+}
+
+// GetCloudProviderSecret returns the required azure json secret for the provided parameters. It differentiates between
+func GetCloudProviderSecret(d azure.ClusterDescriber, namespace, name string, identityType infrav1.VMIdentity, userIdentityID string) (*corev1.Secret, error) {
+	secret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: namespace,
+			Name:      fmt.Sprintf("%s-azure-json", name),
+			Labels: map[string]string{
+				string(infrav1.ResourceLifecycleOwned): "",
```
CecileRobertMichon, line 155: name of the cluster as value?

CecileRobertMichon: actually maybe it should be the other way around to match tags, `clusterName : owned`

CecileRobertMichon:
Does this work with clusterctl move already or does it need https://github.com/kubernetes-sigs/cluster-api/pull/3337 ?

alexeldeib:
> Does this work with clusterctl move already or does it need kubernetes-sigs/cluster-api#3337 ?

If i'm reading that correctly, the constraint is around instances of CRs? I'm not sure I totally understand why, I haven't looked at the clusterctl code. Sounds almost like it's enumerating a known scheme instead of using discovery, so it misses CRs but gets CRDs.

Anyway, these are just secret objects, so I think we're fine.

cloud/scope/clients.go
alexeldeib: you need to re-apply it into settings so `settings.GetAuthorizer()` works properly -___-

we should split all the env settings and credentials code up. With cluster describer we'll mostly split out subscription/credentials, should make sure we have a nice place for tenant ID (even if it's just the env for now)

controllers/helpers.go
alexeldeib: there's no way for users to override right now, correct? we're just pulling from azure cluster

CecileRobertMichon: There is, if you bring your own vnet and the NSG is in the vnet rg which is different from the cluster rg

alexeldeib: I don't think that's implied, I think all three can be different? I just went to the Azure portal and picked a subnet from a vnet in one RG, it allows me to associate with an NSG from other resource groups.

If I change this here, even if vnet + nsg is the more common scenario, we're trading the existing bug for a new one. we probably need to add nsg RG to the spec somewhere?

alexeldeib: either way happy to expose the value and set it to d.ResourceGroup(), then we can follow up and allow setting it to other things?

CecileRobertMichon: > If I change this here, even if vnet + nsg is the more common scenario, we're trading the existing bug for a new one. we probably need to add nsg RG to the spec somewhere?

💯 

let's hold then

controllers/azurejson_machinetemplate_controller.go
alexeldeib: Not totally sure how to simplify further than I have already, but agreed about this preamble. Probably need to play around with schemes and interfaces a bit to get everything to play nice. Probably not going to make it in the current PR?

controllers/azurejson_machinepool_controller.go
alexeldeib: added, sorry I accidentally rebased before letting you review the newly added bits

templates/flavors/system-assigned-identity/system-assigned-identity-md.yaml
```diff
@@ -1,76 +0,0 @@
----
-apiVersion: cluster.x-k8s.io/v1alpha3
```
CecileRobertMichon, line 2: nice, I noticed in #833 that there was a duplication of the md kubeadmConfigTemplate and was going to open a PR for it later 👍

cloud/scope/clients.go
alexeldeib: I think the flip actually, sub isn't necessary for auth. remove it here and retrieve it from the cluster/control plane (clusterdescriber) spec

alexeldeib: (still not here though)

alexeldeib: oh nvm, I see what you mean

templates/cluster-template-ephemeral.yaml
```diff
@@ -177,7 +169,7 @@ spec:
       osDisk:
         diffDiskSettings:
           option: Local
-        diskSizeGB: 30
+        diskSizeGB: 128
```
alexeldeib, line 172: re-used the common template with slightly larger disk

templates/cluster-template-external-cloud-provider.yaml
```diff
@@ -173,7 +165,7 @@ spec:
     spec:
       location: ${AZURE_LOCATION}
       osDisk:
-        diskSizeGB: 30
+        diskSizeGB: 128
```
alexeldeib, line 168: 128

hack/log/redact.sh
```diff
@@ -26,7 +26,6 @@ redact_vars=(
     "${AZURE_CLIENT_SECRET:-}"
     "${AZURE_SUBSCRIPTION_ID:-}"
     "${AZURE_TENANT_ID:-}"
-    "${AZURE_JSON_B64:-}"
```
CecileRobertMichon, line 29: We still want to attempt to redact in case one of the templates still uses the old method (but it shouldn't fail if there's nothing to redact.

alexeldeib:
haven't tested with clusterctl move, need to try that (first time!)

CecileRobertMichon:
@alexeldeib from the sounds of https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/840, clusterctl is probably broken with capz right now 😐

CecileRobertMichon:
lgtm

/assign @devigned @nader-ziada

nader-ziada:
I tried running a few scenarios and seems to work for me

/lgtm

CecileRobertMichon:
this is not the kind of pony energy I was going for ^

alexeldeib:
now _that's_ what i'm talking about

PR #803, opened 2020-07-18
🐛 fix template default variables

CecileRobertMichon:
follow up from #789 to fix Tilt + defaults in templates
- use drone/envsubst to be consistent with clusterctl 
- use ${VAR:=${FOO}} syntax as envsubst to does not work with ${VAR:=$FOO}
```
export CLUSTER_NAME=cecile
echo 'name: ${AZURE_VNET_NAME=$CLUSTER_NAME-vnet}' | ./hack/tools/bin/envsubst
name: $CLUSTER_NAME-vnet
echo 'name: ${AZURE_VNET_NAME=${CLUSTER_NAME}-vnet}' | ./hack/tools/bin/envsubst
name: cecile-vnet
```

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,9 +54,8 @@ CONVERSION_GEN_VER := v0.18.4
 CONVERSION_GEN_BIN := conversion-gen
 CONVERSION_GEN := $(TOOLS_BIN_DIR)/$(CONVERSION_GEN_BIN)-$(CONVERSION_GEN_VER)
 
-ENVSUBST_VER := v1.1.0
 ENVSUBST_BIN := envsubst
-ENVSUBST := $(TOOLS_BIN_DIR)/$(ENVSUBST_BIN)-$(ENVSUBST_VER)
+ENVSUBST := $(TOOLS_BIN_DIR)/$(ENVSUBST_BIN)
 
 GOLANGCI_LINT_VER := v1.27.0
 GOLANGCI_LINT_BIN := golangci-lint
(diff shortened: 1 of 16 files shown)


templates/flavors/base/cluster-template.yaml
```diff
@@ -21,12 +21,12 @@ kind: AzureCluster
...
   name: ${CLUSTER_NAME}
 spec:
-  resourceGroup: "${AZURE_RESOURCE_GROUP:=$CLUSTER_NAME}"
+  resourceGroup: "${AZURE_RESOURCE_GROUP:=${CLUSTER_NAME}}"
   location: "${AZURE_LOCATION}"
   subscriptionID: ${AZURE_SUBSCRIPTION_ID}
   networkSpec:
     vnet:
-      name: "${AZURE_VNET_NAME:=$CLUSTER_NAME-vnet}"
+      name: "${AZURE_VNET_NAME:=${CLUSTER_NAME}-vnet}"
```
alexeldeib, line 29: why have this in the spec at all any more with webhook defaulting? should we consider slimming down the examples?

Makefile
```diff
@@ -184,7 +183,7 @@ $(CONVERSION_GEN): ## Build conversion-gen.
 	GOBIN=$(TOOLS_BIN_DIR) $(GO_INSTALL) k8s.io/code-generator/cmd/conversion-gen $(CONVERSION_GEN_BIN) $(CONVERSION_GEN_VER)
 
 $(ENVSUBST): ## Build envsubst from tools folder.
-	GOBIN=$(TOOLS_BIN_DIR) $(GO_INSTALL) github.com/a8m/envsubst/cmd/envsubst $(ENVSUBST_BIN) $(ENVSUBST_VER)
+	cd $(TOOLS_DIR) && go build -tags=tools -o $(ENVSUBST) github.com/drone/envsubst/cmd/envsubst
```
nader-ziada, line 186: the $(TOOLS_DIR) does not exist at this point when running the e2e tests on prow

maybe `mkdir -p $(TOOLS_DIR)`

templates/flavors/base/cluster-template.yaml
CecileRobertMichon: Agreed, however I'd like to patch the release templates with minimal changes so I'd prefer merging this as is and then thinking about reducing the number of variables in the template. Also I think we shouldn't do this before #802 since the vnetName still needs to be consistent with azure.json and exposing one but not the other could lead to issues.

alexeldeib:
n.b., after this PR be sure as a dev to switch from a8m/envsubst to drone/envsubst in hack/tools/bin or you will break (make clean + make create-workload-cluster in my case):

```
  location: southcentralus
  networkSpec:
    vnet:
      name: ace-verb-vnet-vnet}    <-----
  resourceGroup: ace-verb}          <-----
```

CecileRobertMichon:
> n.b., after this PR be sure as a dev to switch from a8m/envsubst to drone/envsubst in hack/tools/bin or you will break (make clean + make create-workload-cluster in my case):

fixed in #808, should get automatically updated now

PR #804, opened 2020-07-18
💎 add resource group defaulting

alexeldeib:

default rg to cluster-name. k8s name validation is already a subset of rg restrictions (I think? happy to be corrected) so this should be safe.

https://kubernetes.io/docs/concepts/overview/working-with-objects/names/
https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules#microsoftresources

Fixes #

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
default resource group name to AzureCluster name if not specified
```

### DIFF
--- a/api/v1alpha3/azurecluster_default.go
+++ b/api/v1alpha3/azurecluster_default.go
@@ -34,8 +34,15 @@ func (c *AzureCluster) setDefaults() {
 }
 
 func (c *AzureCluster) setNetworkSpecDefaults() {
+	c.setResourceGroupDefault()
 	c.setVnetDefaults()
 	c.setSubnetDefaults()
+}
+
+func (c *AzureCluster) setResourceGroupDefault() {
+	if c.Spec.ResourceGroup == "" {
+		c.Spec.ResourceGroup = c.Name
+	}
 }
 
 func (c *AzureCluster) setVnetDefaults() {
--- a/api/v1alpha3/azurecluster_default_test.go
+++ b/api/v1alpha3/azurecluster_default_test.go
@@ -21,8 +21,64 @@ import (
 	"reflect"
 	"testing"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
+
+func TestResourceGroupDefault(t *testing.T) {
+	cases := map[string]struct {
+		cluster *AzureCluster
+		output  *AzureCluster
+	}{
+		"default empty rg": {
+			cluster: &AzureCluster{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "foo",
+				},
+				Spec: AzureClusterSpec{},
+			},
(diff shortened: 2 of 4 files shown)


api/v1alpha3/azurecluster_default.go
```diff
@@ -34,10 +34,17 @@ func (c *AzureCluster) setDefaults() {
 }
 
 func (c *AzureCluster) setNetworkSpecDefaults() {
+	c.setResourceGroupDefault()
 	c.setVnetDefaults()
 	c.setSubnetDefaults()
 }
 
+func (c *AzureCluster) setResourceGroupDefault() {
```
CecileRobertMichon, line 42: if we're going to default it, should also make the field optional? that would be a breaking change for v1alpha3 -> v1alpha2 conversion unfortunately

alexeldeib: IIRC, webhook defaulting and validating occur before API server schema validation. So we'd want the API server to see this field as required and the webhook to populate it if empty. I think this gets the behavior we want

alexeldeib: let me play around with some yaml to verify that

alexeldeib: thinking through this a bit -- switching to optional (and I assume a pointer) is typically useful when we want to differentiate between a type's zero value and that field not being sent at all. An example off the top of my head would be replicas -- sending a value of zero and not sending a value for a deployment spec result in different behavior: not sending a value gets defaulted inside the controller, sending 0 passes through as-is.

Empty string and no value would have the same behavior here. We always need this field defined from Azure perspective, so I think this is already semantically the most correct -- strictly requiring it, and simplifying the UX to autopopulate it.

CecileRobertMichon: Checked out your branch and removed the "resourceGroup" field, I'm getting `error: error validating "STDIN": error validating data: ValidationError(AzureCluster.spec): missing required field "resourceGroup" in io.x-k8s.cluster.infrastructure.v1alpha3.AzureCluster.spec; if you choose to ignore these errors, turn validation off with --validate=false`

CecileRobertMichon: I think there's a difference between `resourceGroup: ""` and no resourceGroup specified at all

CecileRobertMichon: we can mark it as `// +optional` but not make it a pointer, just like vnetSpec

alexeldeib: sad, using `--v=10` it looks like kubectl uses discovery to do client-side schema validation, even though it would work with the webhooks properly.

will do as you suggested

CecileRobertMichon:
@alexeldeib please add a release note

nader-ziada:
/lgtm

I like this change from a UX perspective, we can update the quick start guide when this gets released

PR #806, opened 2020-07-20
Fixes unit tests that didn't have parrallelism set up correctly

jsturtevant:
When adding tests for the ipv6 I noticed some strange behavior in the unit tests.  It ends up the parallel tests was missing a capture of the range variable which causes inconsistent behavior in the tests and allowed all of the tests to pass even if mocks were not called properly.

This updates the tests that were missing the capture of the range variable. Only the networking interface tests needed updating due to invalid tests.  I tried to capture the intent of the tests but made a few changes that someone else might have more context for. 

Fixes #

If you run the tests before the range variable was capture the tests will pass. You can try adding invalid values to expectations to see that the tests do pass when not captured. After disabling parallel tests or capture the range the tests with invalid values will fail.

The docs here show an example of the capture:  https://golang.org/pkg/testing/#hdr-Subtests_and_Sub_benchmarks

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/api/v1alpha2/tags_test.go
+++ b/api/v1alpha2/tags_test.go
@@ -75,6 +75,7 @@ func TestTags_Merge(t *testing.T) {
 		},
 	}
 	for _, tc := range tests {
+		tc := tc
 		t.Run(tc.name, func(t *testing.T) {
 			t.Parallel()
 			tags := Tags{
--- a/api/v1alpha3/tags_test.go
+++ b/api/v1alpha3/tags_test.go
@@ -75,6 +75,7 @@ func TestTags_Merge(t *testing.T) {
 		},
 	}
 	for _, tc := range tests {
+		tc := tc
 		t.Run(tc.name, func(t *testing.T) {
 			t.Parallel()
 			tags := Tags{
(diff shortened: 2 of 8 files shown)


jsturtevant:
I am not seeing anything in the logs for the e2e job failure that looks off. This PR only changed the unit tests.  Is there a log that I should be looking at in particular?

devigned:
Based on the output I see in the Artifacts, it looks like the infrastructure came up as expected, but ngix was not reachable.

I think we need to add more output from our test failures, so that we can better diagnose why the nginx deployment is not becoming available. The infrastructure output is great, but for each of the tests, we need something more than bool didn't equal "true".

CecileRobertMichon:
+1 to @devigned's comment above

opened https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/807

CecileRobertMichon:
Looking at some open PRs, I realized there are other test files under services/ (eg. routetables_test.go) that don't have parallelism turned on yet so we should also enable it in those, either in this PR or as a follow-up.

CecileRobertMichon:
/retest

looks like an infra flake this time :(

nader-ziada:
/approve

created a follow up PR to add t.Parallel() to services that doesn't have it

Issue #778, opened 2020-07-10
Support Client Certificate authentication for CAPZ


snehala27:
/kind feature

**Describe the solution you'd like**
Currently CAPZ supports authentication only via client secrets: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/config/default/credentials.yaml
The combination of ClientCertificate and CertPassword should also be supported

**Anything else you would like to add:**
Cloud Provider already supports all auth types: https://github.com/kubernetes-sigs/cloud-provider-azure/blob/e39d48d4ea0db59724f7c7a9c4e5fba4027b13a1/docs/cloud-provider-config.md

**Environment:**

- cluster-api-provider-azure version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

devigned:
This configuration is not documented currently, but if the certificate password and certificate file were available to the controller, I believe the Azure Golang SDK would use them. Right now, we only use CLIENT_ID and SECRET env vars.

alexeldeib:
The SDK can handle it, but we need to update our logic a bit to allow alternate sources: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/387db5a07600fdab823a56da465cbaa355deabeb/cloud/scope/clients.go#L52-L59

GetAuthorizerFromEnvironment() will try all the options IIRC, not sure if it does the correct ordering we want.

devigned:
Indeed. The SDK will check all options if using `GetAuthorizerFromEnvironment`. This logic will also need to change when we implement #586. With #586, we'll probably move an identity reference to the AzureCluster, which will be used as the AAD Identity to provision that AzureCluster.

alexeldeib:
Going to assign this to David since he's working on multitenancy which will directly affect the authorizer setup for this

/assign @devigned

nader-ziada:
fixed as part of #977 since aad-pod-identity already supports that and its mentioned in the docs https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/docs/book/src/topics/multitenancy.md

/close

Issue #481, opened 2020-04-01
AAD support


weinong:
/kind feature

**Describe the solution you'd like**
Use AAD to authenticate with k8s clusters
should be similar or equivalent to these existing products/projects
https://github.com/Azure/aks-engine/blob/master/docs/topics/aad.md
https://docs.microsoft.com/en-us/azure/aks/azure-ad-integration
https://docs.microsoft.com/en-us/azure/aks/azure-ad-v2

**Anything else you would like to add:**
In addition, when implementing AAD feature, please consider to use exec plugin (https://github.com/Azure/kubelogin out-of-tree) instead of azure (in-tree) plugin

References: 
https://github.com/kubernetes-sigs/cloud-provider-azure/issues/314
https://github.com/Azure/kubelogin

weinong:
do you have any requirements from customers? does capz have any oidc support already?

CecileRobertMichon:
/assign

Writing some docs on this as everything should already be possible today. oidc flags can be passed into apiserver via extra args.

PR #809, opened 2020-07-20
🏃add draft of multi-tenancy proposal

### LINKED ISSUE #586: Multi-tenancy within one manager instance
/kind feature

Provide multi-tenancy within the same instance of a CAPZ controller, rather than one instance per credential set.

TODOs:
- [x] discuss in upcoming CAPZ meeting (5/1/2020)
- [ ] create proposal
- [ ] break down features

devigned:

We need to provide the Cluster an identity to use for provisioning besides the default identity of the controller. This will enable the CAPZ controllers to work as a single instance with multiple tenants.

Fixes #586 

see also: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1674

Big thank you to @randomvariable and @andrewmyhre! I cribbed much of this document from the AWS provider proposal.

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/docs/proposals/20200720-single-controller-multitenancy.md
+++ b/docs/proposals/20200720-single-controller-multitenancy.md
@@ -1,0 +1,559 @@
+---
+title: Single Controller Multitenancy
+authors:
+  - "@devigned"
+reviewers:
+creation-date: 2020-07-20
+last-updated: 2020-07-20
+status: implementable
+see-also:
+- https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1674
+- https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/586
+replaces: []
+superseded-by: []
+---
+
+# Single Controller Multitenancy
+
+## Table of Contents
+
+- [Single Controller Multitenancy](#single-controller-multitenancy)
+  - [Table of Contents](#table-of-contents)
+  - [Glossary](#glossary)
+  - [Summary](#summary)
+  - [Motivation](#motivation)
+    - [Goals](#goals)
+  - [Proposal](#proposal)
+    - [User Stories](#user-stories)
+      - [Story 1](#story-1---locked-down-with-service-principal-per-subscription)
+      - [Story 2](#story-2---legacy-behavior-preserved)
+      - [Story 3](#story-3---software-as-a-service-provider)
+  - [Requirements](#requirements)
+    - [Functional](#functional)
+    - [Non-Functional](#non-functional)
+    - [Implementation Details/Notes/Constraints](#implementation-detailsnotesconstraints)
+      - [Proposed Changes](#proposed-changes)
+        - [AAD Pod Identity for Attaching Managed Identities](#aad-pod-identity-for-attaching-managed-identities)
+        - [Cluster API Provider Azure v1alpha3 types](#cluster-api-provider-azure-v1alpha3-types)
+    - [Controller Changes](#controller-changes)
+    - [Clusterctl changes](#clusterctl-changes)
+      - [Validating webhook changes](#validating-webhook-changes)
+      - [Principal Type Credential Provider Behaviour](#principal-type-credential-provider-behaviour)
+    - [Security Model](#security-model)
+      - [Roles](#roles)
+    - [RBAC](#rbac)
+        - [Write Permissions](#write-permissions)
+    - [Namespace Restrictions](#namespace-restrictions)
+    - [CAPZ Controller Requirements](#capz-controller-requirements)
+    - [Risks and Mitigations](#risks-and-mitigations)
+      - [Caching and handling refresh of credentials](#caching-and-handling-refresh-of-credentials)
+  - [Upgrade Strategy](#upgrade-strategy)
+  - [Additional Details](#additional-details)
+    - [Test Plan](#test-plan)
+    - [Graduation Criteria](#graduation-criteria)
+      - [Alpha](#alpha)
+      - [Beta](#beta)
+      - [Stable](#stable)
+  - [Implementation History](#implementation-history)
+
+## Glossary
+
+* Principal Type - One of several ways to provide a form of identity that is ultimately resolved to 
+an Azure Active Directory (AAD) Principal.
+* Authorizer - An implementation of the [Azure SDK for Golang's Authorizer Interface](https://github.com/Azure/go-autorest/blob/7ac73d3561eaa034f458f97362b2743e8b3c048e/autorest/authorization.go#L42).
+* CAPZ - An abbreviation of Cluster API Provider Azure.
+
+## Summary
+
+The CAPZ operator is able to manage Azure cloud infrastructure within the permission scope of the 
+AAD principal it is initialized with, usually through environment vars. The CAPZ operator will be 
+provided credentials via the deployment, either explicitly via environment variables or implicitly 
+via the default SDK credential provider chain, including Azure instance metadata service via System 
+Assigned or User Assigned Identities.
+
+In addition, CAPZ uses the environmentally configured identity for the lifetime of the deployment. 
+This also means that an AzureCluster could be broken if the instance of CAPZ that created it is 
+misconfigured for another set of credentials.
+
+This proposal outlines new capabilities for CAPZ to assumption a different AAD Principal, at 
+runtime, on a per-cluster basis. The proposed changes would be fully backwards compatible and maintain
+the existing behavior with no changes to user configuration required.
+
+## Motivation
+
+For large organizations, especially highly-regulated organizations, there is a need to be able to 
+perform separate duties at various levels of infrastructure - permissions, networks and accounts. 
+Azure Role Based Authorization Controls RBAC provides a model which allows admins to provide 
+identities with the least privilege to perform activities. Within this model it is appropriate for 
+tooling running within the 'management' account to manage infrastructure within the 'workload' 
+accounts. With in the AAD model, the controller can assume the identity of the cluster to perform
+cluster management activities. For CAPZ to be most useful within these organizations it will need 
+to support multi-account models.
+
+Some organizations may also delegate the management of clusters to another third-party. In that 
+case, the boundary between organizations needs to be secured. In AAD, this can be accomplished by
+providing a third party AAD principal RBAC access to the Azure resources required to manage cluster
+infrastructure.
+
+Because a single deployment of the CAPZ operator may reconcile many clusters in its lifetime, it is 
+necessary to modify the CAPZ operator to scope its Azure `Authorizer` to within the reconciliation 
+process.
+
+It follows that an organization may wish to provision control planes and worker groups in separate 
+accounts (including each worker group in a separate account). There is a desire to support this 
+configuration also.
+
+Unlike AWS, Azure doesn't provide mechanisms for assuming roles across account boundaries, but rather
+allows RBAC rights to be enabled across account boundries. This is the largest break in this proposal
+with regard to prior art of [the CAPA multitenancy proposal](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/8c0c3db8af44e3c3a1db772145d96154a3d36280/docs/proposal/20200506-single-controller-multitenancy.md)
+
+### Goals
+
+1. To enable AzureCluster resources reconciliation across AAD account boundaries
+2. To maintain backwards compatibility and cause no impact for users who don't intend to make use of 
+   this capability
+
+## Proposal
+
+### User Stories
+
+#### Story 1 - Locked down with Service Principal Per Subscription
+
+Alex is an engineer in a large organization which has a strict Azure account architecture. This 
+architecture dictates that Kubernetes clusters must be hosted in dedicated Subscriptions with AAD
+identity with RBAC rights to provision the infrastructure only in the Subscription. The workload 
+clusters must run with a System Assigned machine identity. The organization has adopted Cluster API 
+in order to manage Kubernetes infrastructure, and expects 'management' clusters running the Cluster 
+API controllers to manage 'workload' clusters in dedicated Azure Subscriptions with an AAD account 
+which only have access to that Subscription.
+
+The current configuration exists:
+* Subscriptions for each cluster
+* AAD Service Principals with Subscription Owner rights for each Subscription
+* A management Kubernetes cluster running Cluster API Provider Azure controllers
+
+Alex can provision a new workload cluster in the specified Subscription with the corresponding AAD 
+Service Principal by creating new Cluster API resources in the management cluster. Each of the
+workload cluster machines would run as the System Assigned identity described in the Cluster API
+resources. The CAPZ controller in the management cluster uses the Service Principal credentials when
+reconciling the AzureCluster so that it can create/use/destroy resources in the workload cluster.
+
+
+#### Story 2 - Legacy Behavior Preserved
+
+Dascha is an engineer in a smaller, less strict organization with a few Azure accounts intended to 
+build all infrastructure. There is a single Azure Subscription named 'dev', and Dascha wants to 
+provision a new cluster in this Subscription. An existing Kubernetes cluster is already running the 
+Cluster API operators and managing resources in the dev Subscription. Dascha can provision a new 
+cluster by creating Cluster API resources in the existing cluster, omitting the ProvisionerIdentity 
+field in the AzureCluster spec. The CAPZ operator will use the Azure credentials provided in its 
+deployment template.
+
+#### Story 3 - Software as a Service Provider
+
+ACME Industries is offering Kubernetes as a service to other organizations. ACME creates an AAD 
+Identity for each organization and each organization grants that Identity access to provision 
+infrastructure in one or multiple of their Azure Subscriptions. ACME Industries wants to minimise 
+the memory footprint of managing many clusters, and wants to move to having a single instance of 
+CAPZ to managed infrastructure across multiple organizations.
+
+## Requirements
+
+### Functional
+
+<a name="FR1">FR1.</a> CAPZ MUST support assuming an identity specified by an `AzureCluster.ProvisioningIdentity`.
+
+<a name="FR2">FR2.</a> CAPZ MUST support static credentials.
+
+<a name="FR3">FR3.</a> CAPZ MUST prevent privilege escalation allowing users to create clusters in Azure accounts they should
+  not be able to.
+
+<a name="FR4">FR4.</a> CAPZ SHOULD support credential refreshing modified principal data.
+
+<a name="FR5">FR5.</a> CAPZ SHOULD provide validation for principal data submitted by users.
+
+<a name="FR6">FR6.</a> CAPZ MUST support clusterctl move scenarios.
+
+### Non-Functional
+
+<a name="NFR1">NFR1.</a> Each instance of CAPZ SHOULD be able to support 200 clusters using role assumption.
+
+<a name="NFR2">NFR2.</a> CAPZ MUST call AAD APIs only when necessary to prevent rate limiting.
+
+<a name="NFR3">NFR3.</a> Unit tests MUST exist for all credential provider code.
+
+<a name="NFR4">NFR4.</a> e2e tests SHOULD exist for all credential provider code.
+
+<a name="NFR5">NFR5.</a> Credential provider code COULD be audited by security engineers.
+
+### Implementation Details/Notes/Constraints
+
+The current implementation of CAPZ requests new instances of Azure services per cluster and 
+sub-cluster resources. The input for these services is a ClusterScope, which provides the identity
+Azure service will operate as.
+
+```go
+type ClusterScope struct {
+	logr.Logger
+	client      client.Client
+	patchHelper *patch.Helper
+
+	AzureClients
+	Cluster      *clusterv1.Cluster
+	AzureCluster *infrav1.AzureCluster
+}
+```
+The ClusterScope contains the information needed to make an authenticated request against an
+Azure cloud. The `Authorizer` currently contains the AAD identity information loaded from the CAPZ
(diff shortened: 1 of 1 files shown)


docs/proposals/20200720-single-controller-multitenancy.md
```diff
@@ -0,0 +1,548 @@
...
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: "test-account-creds"
+  namespace: capz-system
+stringData:
+ tenantID: 6bec3eaa-cac6-11ea-87d0-0242ac130003
+ clientID: 74e870e6-cac6-11ea-87d0-0242ac130003
+ clietSecret: wJalrXUtnFEMI/K7MDENG/bPxRfiCYEXAMPLEKEY
```
CecileRobertMichon, line 305:
```suggestion
 clientSecret: wJalrXUtnFEMI/K7MDENG/bPxRfiCYEXAMPLEKEY
```

docs/proposals/20200720-single-controller-multitenancy.md
```diff
@@ -0,0 +1,548 @@
...
+* `AzureCluster`
+
+<strong><em>New Resources</strong></em>
+
+<em>Cluster scoped resources</em>
+
+* `AzureServicePrincipal` represents an AAD Service Principal. Should support both 
+   certificate and text secrets.
+* `AzureSystemAssignedIdentity` represents an Azure System Assigned Identity provided via 
+   Pod Identity
```
alexeldeib, line 245: `provided via Pod Identity`? could you clarify the intended meaning here?

docs/proposals/20200720-single-controller-multitenancy.md
```diff
@@ -0,0 +1,548 @@
...
+type Authorizer interface {
+    WithAuthorization() PrepareDecorator
+}
+
+type AzureIdentityProvider interface {
+    Authorizer
+
+    // Hash returns a unique hash of the data forming the credentials
+    // for this principal
+    Hash() (string, error)
```
alexeldeib, line 427: watch for changes on observedGeneration to the secret for same effect?

docs/proposals/20200720-single-controller-multitenancy.md
```diff
@@ -0,0 +1,548 @@
...
+The controller will maintain a cache of all the `Authorizer`s used by the clusters.
+
+In the first instance, the controller will refresh secrets on each reconciliation run, and store a 
+hash of the secret contents in the cache. If the hash changes, then the `Authorizer` will be
+recreated via the secret contents.
+
+This could be further optimised by having the controller maintain a watch on all Secrets matching 
+the principal types. Upon receiving an update event, the controller will update lookup the key in 
+the cache and update the relevant `Authorizer`. This may be implemented as its own interface. 
+Mutexes will ensure in-flight updates are completed prior to SDK calls are made. This would require
```
alexeldeib, line 444: SDK clients instantiate per-subscription anyway, what synchronization do you see being required here? I guess you can set the base URL repeatedly, but e.g. cloud environments will also differ, so IMO it's easier to build that all up per user

docs/proposals/20200720-single-controller-multitenancy.md
```diff
@@ -0,0 +1,548 @@
...
+| Infrastructure Provider      | Yes                  | Yes         | Yes     |
+| Management Cluster Operators | Yes                  | Yes         | Yes     |
+| Workload Cluster Operator    | No                   | No          | Yes     |
+
+### Namespace Restrictions
+The extra configuration options are not possible to control with RBAC. Instead,
+they will be controlled with configuration fields on GatewayClasses:
+
+* **allowedNamespaces**: This field is a selector of namespaces that
+  Gateways can use this `AzureServicePrincipal`, `AzureUserAssignedIdentity`, and
```
alexeldeib, line 486: `that Gateways can use this`?

docs/proposals/20200720-single-controller-multitenancy.md
alexeldeib: nvm, it's an interface so needs to be portable. You could treat it like a monotonic counter (because it is) but I can see why a hash is more elegant here

docs/proposals/20200720-single-controller-multitenancy.md
```diff
@@ -0,0 +1,548 @@
...
+
+In the first instance, the controller will refresh secrets on each reconciliation run, and store a 
+hash of the secret contents in the cache. If the hash changes, then the `Authorizer` will be
+recreated via the secret contents.
+
+This could be further optimised by having the controller maintain a watch on all Secrets matching 
+the principal types. Upon receiving an update event, the controller will update lookup the key in 
+the cache and update the relevant `Authorizer`. This may be implemented as its own interface. 
+Mutexes will ensure in-flight updates are completed prior to SDK calls are made. This would require 
+changes to RBAC, and maintaining a watch on secrets of a specific type will require further
```
alexeldeib, line 445: > maintaining a watch on secrets of a specific type will require further
investigation as to feasibility

this sounds fine to me, anything you're worried about? if there's any identifying metadata you can use it to write a filter

docs/proposals/20200720-single-controller-multitenancy.md
```diff
@@ -0,0 +1,548 @@
...
+type AzureIdentityProvider interface {
+    Authorizer
+
+    // Hash returns a unique hash of the data forming the credentials
+    // for this principal
+    Hash() (string, error)
+}
+```
+
+Azure client sessions are structs implementing the `Authorizer` interface. Every SDK will verify the
```
alexeldeib, line 431: Authorizers don't directly expose the tokens AFAIK, i'm a tiny bit confused if you're talking about authorizer caching or token caching. Do you plan to directly keep tokens fresh via https://github.com/Azure/go-autorest/blob/7ac73d3561eaa034f458f97362b2743e8b3c048e/autorest/adal/token.go#L765-L784 or just the authorizers?

docs/proposals/20200720-single-controller-multitenancy.md
```diff
@@ -0,0 +1,548 @@
...
+* Not implement invalid configuration. For example, if the AzureCluster references an Azure Identity
+  in an invalid namespace for it, it should indicate it through a condition or ignore.
+* Respond to changes in an Azure Identity spec change.
+
+### Risks and Mitigations
+
+#### Caching and handling refresh of credentials
+
+For handling many accounts, the number of calls to the STS service must be minimised. This is 
+currently implemented as a cache on cluster key in Cluster API Provider Azure.
```
alexeldeib, line 509: why cache on cluster instead of credential name/type? would minimize cache duplication and extra calls

docs/proposals/20200720-single-controller-multitenancy.md
```diff
@@ -0,0 +1,548 @@
...
+* If principalRef is specified, the CRD is fetched and unmarshalled into a Azure `Authorizer` for 
+  the principal type.
+* The controller will compare the hash of the credential provider against the same secret’s provider
+  in a cache ([NFR2](#NFR2)).
+* The controller will take the newer of the two and instantiate Azure clients with the selected 
+  `Authorizer`.
+* The controller will set the Identity resource as one of the OwnerReferences of the AzureCluster.
+
+### Clusterctl changes
+
```
nader-ziada, line 393: related issue in cluster-api https://github.com/kubernetes-sigs/cluster-api/issues/3354

docs/proposals/20200720-single-controller-multitenancy.md
```diff
@@ -0,0 +1,548 @@
...
+### Functional
+
+<a name="FR1">FR1.</a> CAPZ MUST support assuming an identity specified by an `AzureCluster.ProvisioningIdentity`.
+
+<a name="FR2">FR2.</a> CAPZ MUST support static credentials.
+
+<a name="FR3">FR3.</a> CAPZ MUST prevent privilege escalation allowing users to create clusters in Azure accounts they should
+  not be able to.
+
+<a name="FR4">FR4.</a> CAPZ SHOULD support credential refreshing modified principal data.
```
CecileRobertMichon, line 170: is an example of this being able to refresh an expired service principal secret to be able to keep managing an existing cluster?

docs/proposals/20200720-single-controller-multitenancy.md
```diff
@@ -0,0 +1,548 @@
...
+
+<a name="FR5">FR5.</a> CAPZ SHOULD provide validation for principal data submitted by users.
```
CecileRobertMichon, line 172: what kind of validation? Does that include validating that the credentials are valid/not expired?

docs/proposals/20200720-single-controller-multitenancy.md
```diff
@@ -0,0 +1,548 @@
...
+
+<a name="FR6">FR6.</a> CAPZ MUST support clusterctl move scenarios.
+
+### Non-Functional
+
+<a name="NFR1">NFR1.</a> Each instance of CAPZ SHOULD be able to support 200 clusters using role assumption.
```
CecileRobertMichon, line 178: why 200?

docs/proposals/20200720-single-controller-multitenancy.md
```diff
@@ -0,0 +1,548 @@
...
+
+<a name="NFR2">NFR2.</a> CAPZ MUST call STS APIs only when necessary to prevent rate limiting.
```
CecileRobertMichon, line 180: STS?

docs/proposals/20200720-single-controller-multitenancy.md
```diff
@@ -0,0 +1,548 @@
...
+	logr.Logger
+	client      client.Client
+	patchHelper *patch.Helper
+
+	AzureClients
+	Cluster      *clusterv1.Cluster
+	AzureCluster *infrav1.AzureCluster
+}
+```
+The ClusterScope contains all of the infromation needed to make an authenticated request against an
```
CecileRobertMichon, line 205:
```suggestion
The ClusterScope contains all of the information needed to make an authenticated request against an
```

docs/proposals/20200720-single-controller-multitenancy.md
```diff
@@ -0,0 +1,548 @@
...
+Azure cloud. The `Authorizer` currently contains the AAD identity information loaded from the CAPZ
+controller environment. The `Authorizer` is used to fetch and refresh tokes for all Azure clients.
+
+```go
+// AzureClients contains all the Azure clients used by the scopes.
+type AzureClients struct {
+	SubscriptionID             string
+	ResourceManagerEndpoint    string
+	ResourceManagerVMDNSSuffix string
+	Authorizer                 autorest.Authorizer
```
CecileRobertMichon, line 215: this actually changes in #802

docs/proposals/20200720-single-controller-multitenancy.md
```diff
@@ -0,0 +1,548 @@
...
+  namespace: capz-system
+stringData:
+ tenantID: 6bec3eaa-cac6-11ea-87d0-0242ac130003
+ clientID: 74e870e6-cac6-11ea-87d0-0242ac130003
+ clietSecret: wJalrXUtnFEMI/K7MDENG/bPxRfiCYEXAMPLEKEY
+```
+
+<em>AzureSystemAssignedIdentity</em>
+
+`AzureSystemAssignedIdentity` allows CAPZ to use the SystemAssigned identity provided via Pod
```
CecileRobertMichon, line 310: Is this for using Managed identity for CAPZ provisioning? I assume that's seperate from VM identity to be used by the cloud provider? (just making sure my understanding is correct)

docs/proposals/20200720-single-controller-multitenancy.md
```diff
@@ -0,0 +1,548 @@
...
+
+Today, `clusterctl move` operates by tracking objectreferences within the same namespace, since we 
+are now proposing to use cluster-scoped resources, we will need to add requisite support to 
+clusterctl's object graph to track ownerReferences pointing at cluster-scoped resources, and ensure 
+they are moved. We will naively not delete cluster-scoped resources during a move, as they maybe 
+referenced across namespaces.
+
+#### Validating webhook changes
+
+None.
```
CecileRobertMichon, line 402: Doesn't that conflict with the functional requirement above "CAPZ SHOULD provide validation for principal data submitted by user"?

docs/proposals/20200720-single-controller-multitenancy.md
CecileRobertMichon: Ah Security Token Service... Let's add that to the glossary

docs/proposals/20200720-single-controller-multitenancy.md
```diff
@@ -0,0 +1,548 @@
...
+```go
+func NewClusterScope(params ClusterScopeParams) (*ClusterScope, error) {
+  ...
+  return &ClusterScope{
+    ...
+  }, nil
+}
+```
+
+#### Proposed Changes
```
CecileRobertMichon, line 231: Does this proposal have any impact on the existing VM Identity in the AzureMachineSpec? I'm assuming the answer is no but might be worth making that distinction explicit

docs/proposals/20200720-single-controller-multitenancy.md
```diff
@@ -0,0 +1,548 @@
...
+##### Cluster API Provider Azure v1alpha3 types
+
+<strong><em>Changed Resources</strong></em>
+* `AzureCluster`
+
+<strong><em>New Resources</strong></em>
+
+<em>Cluster scoped resources</em>
+
+* `AzureServicePrincipal` represents an AAD Service Principal. Should support both
```
CecileRobertMichon, line 242: related to https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/778

docs/proposals/20200720-single-controller-multitenancy.md
```diff
@@ -0,0 +1,548 @@
...
+#### Validating webhook changes
+
+None.
+
+A validating webhook could potentially handle some cross-resource validation necessary for the [security
+model](#security-model) and provide more immediate feedback to end users. However, it would be 
+imperfect. For example, a change to an Azure Identity resource could affect the validity of 
+corresponding AzureCluster.
+
+#### Principal Type Credential Provider Behaviour
```
CecileRobertMichon, line 409: Conversion changes? What happens if you have a v1alpha3 cluster using SystemAssigned identity and you try to convert to v1alpha2?

docs/proposals/20200720-single-controller-multitenancy.md
alexeldeib: I don't think this is a huge deal, I'd rather clean up our credentials anyway. I'd consider the changes to this in #802 mostly a workaround to avoid bigger changes

docs/proposals/20200720-single-controller-multitenancy.md
devigned: I've added a short section to describe what I was thinking about for using Pod Identity.

https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/809/files#diff-8ae7a0e213d126365a14e51e003680a8R234-R242

docs/proposals/20200720-single-controller-multitenancy.md
devigned: 👍

docs/proposals/20200720-single-controller-multitenancy.md
devigned: No. The identities on the workload infrastructure will not change. The changes will only affect the control plane nodes and the CAPZ controllers.

devigned: I've added this scoping to the document. TY

docs/proposals/20200720-single-controller-multitenancy.md
devigned: This is more about someone changing the content of the secret, which should cause the `Authorizer` cache to revoke the existing `Authorizer` and rebuild it from the changed secret.

docs/proposals/20200720-single-controller-multitenancy.md
devigned: It would be nice to attempt to get a token from AAD. If we are unable to get a token from AAD, perhaps we get a 401, we should reject the secret as invalid.

docs/proposals/20200720-single-controller-multitenancy.md
devigned: Seems like a reasonable aspiration and was part of the CAPA proposal.

docs/proposals/20200720-single-controller-multitenancy.md
devigned: Changed STS to AAD.

docs/proposals/20200720-single-controller-multitenancy.md
devigned: Yes. I added an explanation and scoping of the usage of AAD Pod Identity.

docs/proposals/20200720-single-controller-multitenancy.md
devigned: Indeed. I'll update.

docs/proposals/20200720-single-controller-multitenancy.md
devigned: That's a great question. I think that fails.

We could add the identity information to the metadata of the v1alpha2 version of the AzureCluster. Thoughts?

docs/proposals/20200720-single-controller-multitenancy.md
devigned: I'd prefer to just cache the `Authorizer`, not the token directly. The `Authorizer` is responsible for refreshing an expired token.

devigned: rewrote this section for clarity

docs/proposals/20200720-single-controller-multitenancy.md
devigned: good catch... copypasta

docs/proposals/20200720-single-controller-multitenancy.md
devigned: good idea. I've updated to reflect your feedback.

docs/proposals/20200720-single-controller-multitenancy.md
devigned: rewrote this section -- There is no need for mutex afaict

docs/proposals/20200720-single-controller-multitenancy.md
devigned: I was not, but the CAPA authors were. Not sure if I missed something... Wanted to leave it in for someone smarter than me to spot the hole.

devigned:
I think I've addressed all of the feedback thus far. PTAL when you all have a min.

docs/proposals/20200720-single-controller-multitenancy.md
```diff
@@ -0,0 +1,559 @@
...
+// AzurePrincipalKind defines allowed Azure principal types
+// +kubebuilder:validation:Enum=AzureSystemAssigned;AzureUserAssigned;AzureServicePrincipal
+type AzurePrincipalKind string
+
+type AzurePrincipalRef struct {
+  Kind AzurePrincipalKind `json:"kind"`
+  Name string `json:"name"`
+}
+
+type  AzureClusterSpec  struct {
```
CecileRobertMichon, line 282: I just thought about this, what's the impact on the subscriptionID field in AzureClusterSpec? None right?

devigned: The changes should have no impact on subscriptionID. 

An AAD identity has nothing to do with a subscription. Auth only needs the tenant, client ID, a secret, and resource or scopes.

If one day in the future we decide to build multi-subscription clusters, then we'd have to make a change to subscription ID.

CecileRobertMichon: Got it. Thanks for the clarification.

CecileRobertMichon:
/hold

Starting lazy consensus for a week from now (7/31)

docs/proposals/20200720-single-controller-multitenancy.md
CecileRobertMichon: We should at least document that as breaking. Wonder how CAPA is handling conversion... we should probably align on that.

CecileRobertMichon:
another quick question: I assume this has no impact on https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/481, correct? Since this is for identity for infra provisioning

devigned:
> another quick question: I assume this has no impact on #481, correct? Since this is for identity for infra provisioning

@CecileRobertMichon, nope. I think this is more for user auth to the cluster (kubectl auth).

docs/proposals/20200720-single-controller-multitenancy.md
```diff
@@ -0,0 +1,559 @@
...
+##### Write Permissions
+|                              | AzureServicePrincipal, etc | Azure RBAC API | Cluster |
+| ---------------------------- | -------------------- | ----------- | ------- |
+| Infrastructure Provider      | Yes                  | Yes         | Yes     |
+| Management Cluster Operators | Yes                  | Yes         | Yes     |
+| Workload Cluster Operator    | No                   | No          | Yes     |
+
+### Namespace Restrictions
+The extra configuration options are not possible to control with RBAC. Instead,
+they will be controlled with configuration fields on GatewayClasses:
```
alexeldeib, line 493: GatewayClasses? 🙂

devigned: good catch. fixed

CecileRobertMichon:
Last chance for comments, will remove the hold Monday morning

nader-ziada:
lgtm

PR #811, opened 2020-07-20
💎 refactor role assignment service

CecileRobertMichon:
**What this PR does / why we need it**: Uses the new Service interface to break out role assignment creation from the virtual machine service.

Partially addresses #757 

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Refactor role assignment service
```


cloud/services/roleassignments/roleassignments_test.go
```diff
@@ -0,0 +1,140 @@
...
+)
+
+func TestReconcileVM(t *testing.T) {
+	testcases := []struct {
+		name          string
+		expect        func(s *mock_roleassignments.MockRoleAssignmentScopeMockRecorder, m *mock_roleassignments.MockClientMockRecorder, v *mock_virtualmachines.MockClientMockRecorder)
+		expectedError string
+	}{
+		//{
+		//	name:          "role assignment already exists",
```
nader-ziada, line 43: don't think that's needed

cloud/services/roleassignments/roleassignments_test.go
```diff
@@ -0,0 +1,140 @@
...
+					},
+				}, nil)
+				m.Create(context.TODO(), "/subscriptions/12345/", gomock.AssignableToTypeOf("uuid"), gomock.AssignableToTypeOf(authorization.RoleAssignmentCreateParameters{})).Return(authorization.RoleAssignment{}, autorest.NewErrorWithResponse("", "", &http.Response{StatusCode: 500}, "Internal Server Error"))
+			},
+		},
+	}
+
+	for _, tc := range testcases {
+		t.Run(tc.name, func(t *testing.T) {
+			tc := tc
```
nader-ziada, line 114: doesn't this need to be before the t.Run?

controllers/azuremachine_reconciler.go
```diff
@@ -84,6 +88,11 @@ func (s *azureMachineService) Reconcile(ctx context.Context) (*infrav1.VM, error
 		return nil, errors.Wrapf(vmErr, "failed to create VM %s ", s.machineScope.Name())
 	}
 
+	err = s.roleAssignmentsSvc.Reconcile(ctx)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to create role assignment")
+	}
+
```
nader-ziada, line 95: Having this role Assignment created in a separate step after the VM is already created can cause issues if it fails, the VM would be there with the wrong identity

right?

cloud/services/roleassignments/roleassignments_test.go
CecileRobertMichon: oops forgot to remove... I was wondering, what's the Create() behavior if an identity already exists on that VM? Does it return an error? Not really handling that case right now

cloud/services/roleassignments/roleassignments_test.go
CecileRobertMichon: yep it does

controllers/azuremachine_reconciler.go
CecileRobertMichon: yes you're right although the same thing could happen in the current code https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/811/files#diff-b2813f64759b38fd374267409eaaca65L193 since the VM gets created before the role assignment, if the role assignment fails to create then it would requeue and at the next reconcile it would get the VM, see that it exists and not reconcile.

I think the way we fix that is by removing https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/b0e2d89ea5e941c114713fca837520e64266676b/controllers/azuremachine_controller.go#L340 once we switch the vm service to not be a getter service and instead at the beginning of the VM reconcile Get to VM and return early if it exists, but still reconcile the role assignment separately (hence the question of dealing with a requeue where role assignment was already created).

nader-ziada: ok makes sense

cloud/services/roleassignments/roleassignments_test.go
CecileRobertMichon: I think part of the problem is that we use a unique UUID every time we do the create so it's not deterministic, if we run the reconcile again it will create a new role assignment with a different UUID. I might change it so that the UUID is part of the spec so it only creates one.

CecileRobertMichon:
/hold 

need to explore what happens when role assignment already exists

CecileRobertMichon:
confirmed that role assignment Create() is idempotent, the operation will essentially be a no-op if the same role assignment already exists with the same uuid.

/hold cancel

nader-ziada:
its good to merge, but the test keeps failing for unrelated reasons

CecileRobertMichon:
I filed #814 yesterday for the nginx flakes, looks like I'm hitting it again

/retest

CecileRobertMichon:
/hold 

going to let #801 merge first

nader-ziada:
@CecileRobertMichon 801 merged, let know when you rebase so I can look at it again and lgtm

CecileRobertMichon:
/hold cancel

rebased

CecileRobertMichon:
nevermind, have to rebase again 😅

nader-ziada:
didn't even have time to see the first rebase :D

nader-ziada:
looks like something went wrong with the rebase  `pull-cluster-api-provider-azure-test`  failed with 
```
GOBIN=/home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/hack/tools/bin ./scripts/go_install.sh github.com/golang/mock/mockgen mockgen 8a3d5958550701de9e6650b84b75a118771e7b49
rm: cannot remove '/home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/hack/tools/bin/mockgen*': No such file or directory
go: creating new go.mod: module fake/mod
go: finding github.com 8a3d5958550701de9e6650b84b75a118771e7b49
go: finding github.com/golang/mock/mockgen 8a3d5958550701de9e6650b84b75a118771e7b49
go: finding github.com/golang/mock 8a3d5958550701de9e6650b84b75a118771e7b49
go: finding github.com/golang 8a3d5958550701de9e6650b84b75a118771e7b49
go: downloading github.com/golang/mock v1.4.4-0.20200406172533-8a3d59585507
go: extracting github.com/golang/mock v1.4.4-0.20200406172533-8a3d59585507
go: downloading golang.org/x/tools v0.0.0-20190425150028-36563e24a262
build github.com/golang/mock/mockgen: cannot load golang.org/x/tools/go/packages: golang.org/x/tools@v0.0.0-20190425150028-36563e24a262: unexpected EOF
make[1]: *** [Makefile:200: /home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/hack/tools/bin/mockgen-8a3d5958550701de9e6650b84b75a118771e7b49] Error 1
make[1]: Leaving directory '/home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure'
make: *** [Makefile:232: generate] Error 2
```

CecileRobertMichon:
Had to go fmt

PR #813, opened 2020-07-21
Only look at Group and Kind when checking a reference

ncdc:
Only look at Group and Kind when checking a reference. Do not look at Version.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Only look at Group and Kind when checking a reference
```

### DIFF
--- a/controllers/helpers.go
+++ b/controllers/helpers.go
@@ -24,6 +24,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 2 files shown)


CecileRobertMichon:
Is this changing any user facing behaviors and/or fixing any bugs? If so can you please add a release note?

vincepri:
I can't edit the release note, but basically this is a bug if you have a cluster with an v1alpha2 owner that hasn't been updated yet. It should be a transitory issue, given that controllers update their OwnerReferences when they get reconciled, although this logic is technically more correct than looking at the version number.

There shouldn't be any user-facing changes.

CecileRobertMichon:
updated release note

/approve

PR #816, opened 2020-07-22
separate route tables for node and control-plane

### LINKED ISSUE #718: Does not currently handle creating seperate route tables for node and contorl plane

nader-ziada:
- add different route tables to for node and control-plane

Fixes #718 

- we need to confirm this is a valid requirements, there are some discussion on the issue if we need two route tables. I had done most of the code as part of the route table refactoring, so making a PR anyway

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Add separate route tables for node and control-plane
```

### DIFF
--- a/api/v1alpha3/azurecluster_default.go
+++ b/api/v1alpha3/azurecluster_default.go
@@ -88,9 +88,6 @@ func (c *AzureCluster) setSubnetDefaults() {
 	if cpSubnet.SecurityGroup.Name == "" {
 		cpSubnet.SecurityGroup.Name = generateControlPlaneSecurityGroupName(c.ObjectMeta.Name)
 	}
-	if cpSubnet.RouteTable.Name == "" {
-		cpSubnet.RouteTable.Name = generateRouteTableName(c.ObjectMeta.Name)
-	}
 
 	if nodeSubnet.Name == "" {
 		nodeSubnet.Name = generateNodeSubnetName(c.ObjectMeta.Name)
@@ -102,7 +99,7 @@ func (c *AzureCluster) setSubnetDefaults() {
 		nodeSubnet.SecurityGroup.Name = generateNodeSecurityGroupName(c.ObjectMeta.Name)
 	}
 	if nodeSubnet.RouteTable.Name == "" {
-		nodeSubnet.RouteTable.Name = generateRouteTableName(c.ObjectMeta.Name)
+		nodeSubnet.RouteTable.Name = generateNodeRouteTableName(c.ObjectMeta.Name)
 	}
 }
 
@@ -131,7 +128,7 @@ func generateNodeSecurityGroupName(clusterName string) string {
 	return fmt.Sprintf("%s-%s", clusterName, "node-nsg")
 }
 
-// generateRouteTableName generates a route table name, based on the cluster name.
-func generateRouteTableName(clusterName string) string {
+// generateNodeRouteTableName generates a node route table name, based on the cluster name.
+func generateNodeRouteTableName(clusterName string) string {
 	return fmt.Sprintf("%s-%s", clusterName, "node-routetable")
 }
--- a/api/v1alpha3/azurecluster_default_test.go
+++ b/api/v1alpha3/azurecluster_default_test.go
@@ -250,7 +250,7 @@ func TestSubnetDefaults(t *testing.T) {
 								Name:          "cluster-test-controlplane-subnet",
 								CIDRBlocks:    []string{DefaultControlPlaneSubnetCIDR},
 								SecurityGroup: SecurityGroup{Name: "cluster-test-controlplane-nsg"},
-								RouteTable:    RouteTable{Name: "cluster-test-node-routetable"},
+								RouteTable:    RouteTable{},
 							},
 							{
 								Role:          SubnetNode,
@@ -299,7 +299,7 @@ func TestSubnetDefaults(t *testing.T) {
 								Name:          "my-controlplane-subnet",
 								CIDRBlocks:    []string{"10.0.0.16/24"},
 								SecurityGroup: SecurityGroup{Name: "cluster-test-controlplane-nsg"},
-								RouteTable:    RouteTable{Name: "cluster-test-node-routetable"},
+								RouteTable:    RouteTable{},
 							},
 							{
 								Role:          SubnetNode,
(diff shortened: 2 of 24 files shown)


nader-ziada:
/hold 

for validation of the requirement

cloud/types.go
```diff
@@ -60,7 +60,8 @@ type LBSpec struct {
 
 // RouteTableSpec defines the specification for a route table
 type RouteTableSpec struct {
-	Name string
+	Name           string
+	IsControlPlane bool
```
CecileRobertMichon, line 64:
```suggestion
	MachineRole string
```

https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/816/files#diff-e0a02b9b279470c6d5411301fba575a9L33

nader-ziada: ok, similar to the Subnet I guess

nader-ziada:
@CecileRobertMichon @jsturtevant do we still think we need this feature?

CecileRobertMichon:
> @CecileRobertMichon @jsturtevant do we still think we need this feature?

I think so? it's more correct this way at least, even if we're not leveraging the route tables yet.

jsturtevant:
agreed, it is closer to the the way the public API is designed. The other alternative is to modify the API to only allow a single route table.

@aramase  Do you know of scenarios where control plane would have a different route table than the worker nodes?

nader-ziada:
ok in that case ready for review 
/hold cancel

cloud/services/routetables/routetables.go
```diff
@@ -33,40 +34,51 @@ func (s *Service) Reconcile(ctx context.Context) error {
 	}
 
 	for _, rtSpec := range s.Scope.RouteTableSpecs() {
-		existingRouteTable, err := s.Get(ctx, s.Scope.ResourceGroup(), rtSpec.Name)
-		if !azure.ResourceNotFound(err) {
+		if rtSpec.Role == azure.RouteTableControlPlane {
```
CecileRobertMichon, line 37: instead of doing this logic here, can we just have the subnet be part of the route table spec directly?

aramase:
> @aramase Do you know of scenarios where control plane would have a different route table than the worker nodes?

I don't see a scenario where we would need a different route table for control plane and worker nodes. I haven't had a chance to look at the PR completely, but when the control plane/worker node is bootstrapped is the `RouteTableName` in azure.json going to contain the control plane/worker node RT name populated appropriately?

CecileRobertMichon:
@aramase azure.json uses the node route table as "routeTableName" see https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/controllers/helpers.go#L228

BTW we should change that to get the actual `NodeRouteTable().Name` @nader-ziada, otherwise, it will break if the use uses a different name.

cloud/services/routetables/routetables.go
CecileRobertMichon: I mean instead of Role, have Subnet in https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/816/files#diff-e0a02b9b279470c6d5411301fba575a9R77

That way no need to have logic that knows about roles here

nader-ziada: got it, makes sense, will change that

nader-ziada: made the change, please take a look

nader-ziada:
> @aramase azure.json uses the node route table as "routeTableName" see https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/controllers/helpers.go#L228
> 
> BTW we should change that to get the actual `NodeRouteTable().Name` @nader-ziada, otherwise, it will break if the use uses a different name.

we only create one azure.json secret for machines, so control plane and worker machines will get the same, right?

CecileRobertMichon:
> we only create one azure.json secret for machines, so control plane and worker machines will get the same, right?

we do create one per MD/Machine so technically they could have different ones (although right now the controller sets the same thing regardless of machine role). We are going to have to differ between the two as part of  https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/921 anyways. What I'm not 100% sure about is why would you ever want you azure.json routeTable value to be the control plane route table?

cloud/services/routetables/routetables.go
CecileRobertMichon: looks good

nader-ziada:
rebased with ipv6

api/v1alpha3/azurecluster_default.go
```diff
@@ -89,7 +89,7 @@ func (c *AzureCluster) setSubnetDefaults() {
 		cpSubnet.SecurityGroup.Name = generateControlPlaneSecurityGroupName(c.ObjectMeta.Name)
 	}
 	if cpSubnet.RouteTable.Name == "" {
-		cpSubnet.RouteTable.Name = generateRouteTableName(c.ObjectMeta.Name)
+		cpSubnet.RouteTable.Name = generateControlPlaneRouteTableName(c.ObjectMeta.Name)
```
CecileRobertMichon, line 92: I think we should consider not creating a control plane route table by default and only creating it if the user specifies one since the default use case doesn't require one

CecileRobertMichon:
we also need to update https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/controllers/helpers.go#L226

api/v1alpha3/azurecluster_default.go
nader-ziada: so still use node route table unless the user specifies their own name of control plane route table?

CecileRobertMichon: yes, the route table is attached to the subnet and the cloud provider cares about the NodeSubnet so it should only care about the NodeRouteTable. This behavior is just more correct because it doesn't ignore the Spec for route table on the  control plane subnet and allows more advanced users to customize their cluster. But the default should stay the same, ie. Node subnet has a route table, control plane subnet doesn't, and the azure.json file has the node route table as routeTableName on all nodes.

api/v1alpha3/azurecluster_default.go
```diff
@@ -89,7 +89,7 @@ func (c *AzureCluster) setSubnetDefaults() {
 		cpSubnet.SecurityGroup.Name = generateControlPlaneSecurityGroupName(c.ObjectMeta.Name)
 	}
 	if cpSubnet.RouteTable.Name == "" {
-		cpSubnet.RouteTable.Name = generateRouteTableName(c.ObjectMeta.Name)
+		cpSubnet.RouteTable.Name = generateNodeRouteTableName(c.ObjectMeta.Name)
```
CecileRobertMichon, line 92: this should not be set. The node route table is only associated with the node subnet.

cloud/scope/cluster.go
```diff
@@ -152,11 +152,18 @@ func (s *ClusterScope) LBSpecs() []azure.LBSpec {
...
 
-// RouteTableSpecs returns the node route table(s)
+// RouteTableSpecs returns the node route table
 func (s *ClusterScope) RouteTableSpecs() []azure.RouteTableSpec {
-	return []azure.RouteTableSpec{{
-		Name: s.RouteTable().Name,
-	}}
+	return []azure.RouteTableSpec{
+		{
+			Name:   s.ControlPlaneRouteTable().Name,
```
CecileRobertMichon, line 159: we should only add each routeTable if its name != ""

api/v1alpha3/azurecluster_default.go
nader-ziada: I though we couldn't leave it empty?

cloud/scope/cluster.go
nader-ziada: only the control plane one? or both, we need to have the node route table, right?

api/v1alpha3/azurecluster_default.go
nader-ziada: ok, will make the change and test it out

api/v1alpha3/azurecluster_default_test.go
```diff
@@ -396,7 +446,7 @@ func TestSubnetDefaults(t *testing.T) {
 								Name:          "cluster-test-controlplane-subnet",
 								CIDRBlocks:    []string{DefaultControlPlaneSubnetCIDR},
 								SecurityGroup: SecurityGroup{Name: "cluster-test-controlplane-nsg"},
-								RouteTable:    RouteTable{Name: "cluster-test-node-routetable"},
+								RouteTable:    RouteTable{},
 							},
```
nader-ziada, line 450: @CecileRobertMichon this is what you expected? I tested it and seems to work

nader-ziada:
@CecileRobertMichon @devigned ready for another review. Thanks

api/v1alpha3/azurecluster_default_test.go
CecileRobertMichon: yep that's what I had in mind

devigned (approved):
/lgtm
/approve

Issue #615, opened 2020-05-13
[managedcluster] defaulting webhooks


alexeldeib:
/kind feature

**Describe the solution you'd like**
There are probably a fair number of properties we could default earlier in the pipeline. CNI, Load balancer SKU, networking fields, node SKU/default pool. Add a defaulting webhook and begin defaulting some of these fields.

jroden:
Hi Ace!  I'm still in the process of getting a better understanding (of managedcluster, capi & capiz, and webhooks) before beginning managedcluster work

I've got a couple questions to help validate my understanding if you don't mind :smile: 
Would the flow of creating a managedcluster look like the following?
1. CAPI controller receives request to create a new managecluster with provider azure
2. CAPI controller (may) perform transformation and/or logic based on the request
3. CAPI controller sends request to CAPZ to create a new managedcluster
4. CAPZ processes request and creates new managedcluster

And by using a webhook we can default values between steps 3 & 4 rather than if the logic were embedded in the capz api itself (step 4)?  Is this what you were referring to when you mentioned `defaulting earlier in the pipeline`?

Thanks for the guidance! :smile:

jroden:
So currently we are defaulting fields for managedCluster in the 'Process Item' step of the custom controller? And thus moving to webhook would default the fields even before the controller picks up the changes from the api server?  (snip taken from [here](https://github.com/kubernetes/sample-controller/blob/master/docs/images/client-go-controller-interaction.jpeg))
![image](https://user-images.githubusercontent.com/28738612/84599201-4c53ff00-ae3e-11ea-8b69-e38074793d95.png)

alexeldeib:
> And the webhook would default values as part of step 1? As opposed to if the logic were baked into the CAPZ controller?

Correct.

> So currently we are defaulting fields for managedCluster in the 'Process Item' step of the custom controller? And thus moving to webhook would default the fields even before the controller picks up the changes from the api server?

Also correct.

There are a few cases here that are poorly modeled even though they are technically optional fields. A good example is networkPlugin. I don't think AKS API requires this field to be set, but the API will reject attempts to update this field. We need to model that in our CRD. On the validation side, we would want to prevent the ability to persist invalid configurations (like changing the networkPlugin on update). 

There are a fairly limited number of cases i'd enable defaulting for. We should probably default networkPlugin to Azure (I think having no network policy is valid). 

[...]

Issue #612, opened 2020-05-13
[managedcluster] validating webhook for DNS IP


alexeldeib:
/kind feature

**Describe the solution you'd like**
In AKS users can provider the service CIDR but also can individually provide a DNS IP. We should ensure the DNS IP is inside the service CIDR inside a validating webhook to avoid footguns.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

CecileRobertMichon:
@jroden are you working on this one?

jroden:
@CecileRobertMichon My apologies for the delay!  Yes, I'm working on this as a joint effort with #615 .  I have the validating and mutating hooks created, just working through adding service account permissions so the amcp validating hook can extract the serviceCIDR field from the associated cluster resource.  I actually haven't worked on capz in a couple weeks (been pushing to get csapro cert hopefully this week).  Was planning on picking up the work again starting next week, but can make some time this week if you want!  Just lmk 😄

CecileRobertMichon:
@jroden no worries! Was just doing some issue triage, feel free to work on it whenever you can. Will mark you as assigned so no one else works on it at the same time.

/assign @jroden

jroden:
cleared the exam :tada: so I'm back!

Currently working on the rbac permissions for the amcp validating webhook.  Researching if I can assign a custom serviceAccount to the amcp webhooks since system:serviceaccount:capi-webhook-system:default doesn't have permissions to list cluster objects.

Once I have that sorted I should be ready to submit a PR for these!

jroden:
Hope everyone is having a good week!  I've done a little digging and am hoping to get some feedback.  A quick background on the challenge I'm working through is-- the validating webhook for amcp needs to list the owner cluster resource (to get the serviceCIDR field for validating the amcp DNSServiceIP).  The capz webhooks are served from the capz-controller-manager deployment in the capi-webhook-system namespace.  However, the default serviceAccount for that namespace doesn't have permissions to list cluster resources.  I'm wondering what the thoughts are on associating that deployment with a new capz serviceAccount that has permissions to list cluster resources?  Thanks for the help! :smile: 

@alexeldeib @CecileRobertMichon

devigned:
We have avoided using a client from webhooks up to this point. Perhaps, this deserves stronger consideration than we have provided it in the past. It appears this will be a reoccurring issue in infrastructure providers, which need information about their related CAPI resources.

[...]

PR #818, opened 2020-07-22
 🐛 normalize version in managed control plane

alexeldeib:

remove strict validation and always let prefix normalization occur if necessary on the versions. We added this in https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/745 after realizing it mismatched with Machine, but Machine and KCP themselves were mismatched which was fixed in https://github.com/kubernetes-sigs/cluster-api/pull/3147

This PR removes the stricter validation we added, and allows the prefix trimming (inside managed control plane reconciler) to always normalize the version string.

n/a

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
reduce validation strictness on `v` prefix in Kubernetes versions
```

### DIFF
--- a/exp/api/v1alpha3/azuremanagedcontrolplane_types.go
+++ b/exp/api/v1alpha3/azuremanagedcontrolplane_types.go
@@ -26,7 +26,6 @@ import (
 type AzureManagedControlPlaneSpec struct {
 	// Version defines the desired Kubernetes version.
 	// +kubebuilder:validation:MinLength:=2
-	// +kubebuilder:validation:Pattern:=^v(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)([-0-9a-zA-Z_\.+]*)?$
 	Version string `json:"version"`
 
 	// ResourceGroup is the name of the Azure resource group for this AKS Cluster.
--- a/exp/controllers/azuremanagedmachinepool_reconciler.go
+++ b/exp/controllers/azuremanagedmachinepool_reconciler.go
@@ -19,6 +19,7 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2020-06-01/compute"
 	"github.com/pkg/errors"
@@ -54,13 +55,20 @@ func newAzureManagedMachinePoolReconciler(scope *scope.ManagedControlPlaneScope)
 // Reconcile reconciles all the services in pre determined order
 func (r *azureManagedMachinePoolReconciler) Reconcile(ctx context.Context, scope *scope.ManagedControlPlaneScope) error {
 	scope.Logger.Info("reconciling machine pool")
+
+	var normalizedVersion *string
+	if scope.MachinePool.Spec.Template.Spec.Version != nil {
+		v := strings.TrimPrefix(*scope.MachinePool.Spec.Template.Spec.Version, "v")
+		normalizedVersion = &v
+	}
+
 	agentPoolSpec := &agentpools.Spec{
 		Name:          scope.InfraMachinePool.Name,
 		ResourceGroup: scope.ControlPlane.Spec.ResourceGroup,
 		Cluster:       scope.ControlPlane.Name,
 		SKU:           scope.InfraMachinePool.Spec.SKU,
 		Replicas:      1,
-		Version:       scope.MachinePool.Spec.Template.Spec.Version,
+		Version:       normalizedVersion,
 	}
 
 	if scope.InfraMachinePool.Spec.OSDiskSizeGB != nil {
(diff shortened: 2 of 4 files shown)


alexeldeib:
also fixes a panic when owner ref isn't set on managed cluster

exp/api/v1alpha3/azuremanagedcontrolplane_types.go
```diff
@@ -26,7 +26,6 @@ import (
 type AzureManagedControlPlaneSpec struct {
 	// Version defines the desired Kubernetes version.
 	// +kubebuilder:validation:MinLength:=2
-	// +kubebuilder:validation:Pattern:=^v(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)([-0-9a-zA-Z_\.+]*)?$
```
CecileRobertMichon, line 29: where is prefix normalization happening?

alexeldeib: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/6ad50d0e79546d9f3c5e5d632836412df91a6a9c/exp/controllers/azuremangedcontrolplane_reconciler.go#L59

actually should add it to https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/6ad50d0e79546d9f3c5e5d632836412df91a6a9c/exp/controllers/azuremanagedmachinepool_reconciler.go#L61

alexeldeib: added

exp/controllers/azuremanagedmachinepool_reconciler.go
```diff
@@ -54,13 +55,20 @@ func newAzureManagedMachinePoolReconciler(scope *scope.ManagedControlPlaneScope)
 // Reconcile reconciles all the services in pre determined order
 func (r *azureManagedMachinePoolReconciler) Reconcile(ctx context.Context, scope *scope.ManagedControlPlaneScope) error {
 	scope.Logger.Info("reconciling machine pool")
+
+	var normalizedVersion *string
+	if scope.MachinePool.Spec.Template.Spec.Version != nil {
```
CecileRobertMichon, line 60: can this go in a webhook?

devigned: Are you suggesting a default version?

CecileRobertMichon: yes that way the normalized version appears in the spec and it's not too hidden from the user, same as https://github.com/CecileRobertMichon/cluster-api/blob/master/api/v1alpha3/machine_webhook.go#L60

devigned: Webhooks are not currently setup for `AzureManged*`. Can we open an issue to track this and pull in these changes?

alexeldeib: https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/615
https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/612

I can add this to `AzureManagedMachinePool` in this PR.

alexeldeib: ah, actually in machine webhook it's defaulted inside the same object, this is cross object -- it's taking it from the upstream machine pool. so maybe we should fix it there, if we didn't already?

devigned: Great point.

CecileRobertMichon: That was part of the changes in https://github.com/kubernetes-sigs/cluster-api/pull/3147, the Version isn't part of MachinePool spec, it's part of Machine spec which is embedded in the MachinePool spec as part of the template https://github.com/kubernetes-sigs/cluster-api/blob/master/exp/api/v1alpha3/machinepool_types.go#L44

CecileRobertMichon: So actually do we even need this normalization here? Since the Machine webhook already takes care of it

CecileRobertMichon: For `AzureManagedControlPlane` Version _is_ part of the object https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/818/files#diff-69a80dafa12ba6b9eeaf233d609e8f5eR29 so we could do the same normalization in a webhook, for `AzureMachinePool` I believe it's already taken care of by the `Machine` webhooks

CecileRobertMichon: Ok so after some coffee and thinking through it with @devigned I think we need to: 
1) keep the AzureMachinePool normalization until we can move it into the capi MachinePool webhook
2) Add a webhook normalization for AzureManagedControlPlane, either here or in a separate PR 

Let's track whatever we don't get done here in issues so they don't get lost.

I take back what I said above, the Machine webhook doesn't do anything for MachinePool, only the schema validation is shared.

PR #822, opened 2020-07-23
[WIP]✨ integrate AzureMachine with AzureManagedControlPlane

alexeldeib:

Fixes #

This joins nodes successfully, but they show up as masters (i guess because of admin config? or maybe something aks side?) and Azure CNI doesn't work yet.

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```release-note
AzureMachines are now compatible with AzureManagedControlPlanes. This allows users to join self-managed nodes to AKS.
```


alexeldeib:
@devigned this is wildly unprepared for merge but PTAL as a use case while you consider https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/809

alexeldeib:
I think if I drop the privilege on the kubeconfig I supply, that will drop the role for node join. So I probably just need an authenticated, non-admin kubeconfig. I can create this ahead of time in the AzureManagedControlPlaneReconciler w a remote client.

controllers/azuremachine_controller.go
```diff
@@ -152,28 +157,71 @@ func (r *AzureMachineReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, ret
...
+
+	infraRefKind := cluster.Spec.InfrastructureRef.Kind
+	infraRefGV, err := schema.ParseGroupVersion(cluster.Spec.InfrastructureRef.APIVersion)
+	if err != nil {
+		return ctrl.Result{}, errors.Wrap(err, "failed to parse gv from infra ref apiversion")
+	}
+
+	// AzureManagedControlPlane only works with AzureManagedCluster
+	var isManaged bool
+	clusterDescriberRef := cluster.Spec.InfrastructureRef
```
alexeldeib, line 174: feeling the downside of making `AzureManagedCluster` a no-op right about now...(we could just use infra ref otherwise)

cloud/interfaces.go
```diff
@@ -54,21 +57,37 @@ type CredentialGetter interface {
...
-	SubscriptionID() string
 	BaseURI() string
 	Authorizer() autorest.Authorizer
 }
 
-// ClusterDescriber is an interface which can get common Azure Cluster information
+// ClusterDescriber describe a Kubernetes resource with full
+// serializability, object meta and type meta, plus can describe enough
+// data to build an Azure cluster
 type ClusterDescriber interface {
```
alexeldeib, line 67: @CecileRobertMichon any feelings about this? how do you see scope evolving long term for a use case like this?

cloud/scope/clients.go
```diff
@@ -35,56 +33,45 @@ const (
...
+	c.authorizer, err = settings.GetAuthorizer()
+	if err != nil {
+		return nil, err
 	}
-	return subscriptionID, nil
+	return c, err
 }
 
-// GetAzureDNSZoneForEnvironment returnes the DNSZone to be used with the
-// cloud environment, the default is the public cloud
```
alexeldeib, line 75: we don't need this function even as-is, the logic here is equivalent

exp/api/v1alpha3/azuremanagedcontrolplane_types.go
```diff
@@ -99,6 +106,109 @@ type AzureManagedControlPlaneList struct {
 	Items           []AzureManagedControlPlane `json:"items"`
 }
 
+func (c *AzureManagedControlPlane) ManagedResourceGroup() string {
```
alexeldeib, line 109: CRD types now implement clusterDescriber directly, rather than scope

cloud/services/virtualmachines/virtualmachines.go
```diff
@@ -126,6 +126,12 @@ func (s *Service) Reconcile(ctx context.Context, spec interface{}) error {
 			HardwareProfile: &compute.HardwareProfile{
 				VMSize: compute.VirtualMachineSizeTypes(vmSpec.Size),
 			},
+			DiagnosticsProfile: &compute.DiagnosticsProfile{
```
CecileRobertMichon, line 129: These are all great changes but it's too much for one PR, can you please split 1) vm diagnostics 2) NetworkDescriber/ClusterDescriber changes 3) AzureManagedControlPlane etc.  in multiple PRs?

alexeldeib: yeah, i'm mostly trying to see if I can get it working first // share some of the changes to patterns that have been helpful trying to make this work for different scenarios. I definitely won't try to merge this as one PR, and I found a few small bits here and there to merge as one-off fixes besides the main functional stuff.

alexeldeib: also vm diagnostics is hardcoded, that was me debugging. once I figure out which order to tackle all these properly I can assign the relevant issues.

cloud/interfaces.go
```diff
@@ -54,21 +57,37 @@ type CredentialGetter interface {
...
 	ResourceGroup() string
 	ClusterName() string
 	Location() string
+	SetFailureDomain(id string, spec clusterv1.FailureDomainSpec)
 	AdditionalTags() infrav1.Tags
+}
+
+// NetworkDescriber describes the vnet and subnet configuration for a cluster
+// abstracted because it is implemented by managed and unmanaged cluster.
+type NetworkDescriber interface {
```
CecileRobertMichon, line 81: I have the same thing on my local branch :)

PR #827, opened 2020-07-24
💎  Refactor VM service spec

CecileRobertMichon:
**What this PR does / why we need it**: Refactor the virtual machines service according to patterns from #757.

Fixes partially #757 

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Refactor VM service spec
```


cloud/defaults.go
```diff
@@ -41,27 +40,6 @@ const (
 	LatestVersion = "latest"
 )
 
-// SupportedAvailabilityZoneLocations is a slice of the locations where Availability Zones are supported.
-// This is used to validate whether a virtual machine should leverage an Availability Zone.
-// Based on the Availability Zones listed in https://docs.microsoft.com/en-us/azure/availability-zones/az-overview
-var SupportedAvailabilityZoneLocations = []string{
```
CecileRobertMichon, line 47: we don't need this anymore, we're getting the zones directly from the SKU cache

cloud/scope/machine.go
```diff
@@ -87,6 +88,39 @@ type MachineScope struct {
 	AzureMachine *infrav1.AzureMachine
 }
 
+// VMSpecs returns the VM specs.
+func (m *MachineScope) VMSpecs(ctx context.Context) ([]azure.VMSpec, error) {
```
CecileRobertMichon, line 92: I don't know how I feel about needing context and error as return here... Had to keep them to port over ssh decode/VM image parsing the version and getting the bootstrap data from the secret but not sure that's the best way to handle it yet.

controllers/azuremachine_controller.go
```diff
@@ -307,19 +283,26 @@ func (r *AzureMachineReconciler) reconcileNormal(ctx context.Context, machineSco
 		machineScope.Error(errors.New("Failed to create or update VM"), "VM is in failed state", "id", *machineScope.GetVMID())
 		r.Recorder.Eventf(machineScope.AzureMachine, corev1.EventTypeWarning, "FailedVMState", "Azure VM is in failed state")
 		machineScope.SetFailureReason(capierrors.UpdateMachineError)
-		machineScope.SetFailureMessage(errors.Errorf("Azure VM state is %s", vm.State))
+		machineScope.SetFailureMessage(errors.Errorf("Azure VM state is %s", machineScope.GetVMState()))
 		conditions.MarkFalse(machineScope.AzureMachine, infrav1.VMRunningCondition, infrav1.VMProvisionFailedReason, clusterv1.ConditionSeverityWarning, "")
+		// If VM failed provisioning, delete it so it can be recreated
```
CecileRobertMichon, line 288: not sure if we actually want to do this (maybe we should look into Machine Health Checks instead)? but this has been in the reconciler since basically v1alpha1 (moved from https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/827/files#diff-fb5deef0b083d17ab538518df2570d4bL266) so I just left it for now to not burry behavior breaking changes in a huge refactor

controllers/azuremachine_controller.go
```diff
@@ -307,19 +283,26 @@ func (r *AzureMachineReconciler) reconcileNormal(ctx context.Context, machineSco
...
-		machineScope.Info("VM state is undefined", "state", vm.State, "instance-id", *machineScope.GetVMID())
+		machineScope.Info("VM state is undefined", "state", machineScope.GetVMState(), "instance-id", *machineScope.GetVMID())
 		r.Recorder.Eventf(machineScope.AzureMachine, corev1.EventTypeWarning, "UnhandledVMState", "Azure VM state is undefined")
 		machineScope.SetFailureReason(capierrors.UpdateMachineError)
-		machineScope.SetFailureMessage(errors.Errorf("Azure VM state %q is undefined", vm.State))
+		machineScope.SetFailureMessage(errors.Errorf("Azure VM state %q is undefined", machineScope.GetVMState()))
 		conditions.MarkUnknown(machineScope.AzureMachine, infrav1.VMRunningCondition, "", "")
 	}
 
+	// TODO: move this out of the controller and use tags CreateOrUpdateAtScope instead.
```
CecileRobertMichon, line 303: blocked by https://github.com/Azure/azure-sdk-for-go/issues/11879

cloud/scope/machine.go
CecileRobertMichon: So I had to move it because bootstrap data isn't available during delete, so had to be in create only.

CecileRobertMichon:
This one's ready for review

/cc @devigned @alexeldeib

cloud/converters/vmss.go
```diff
@@ -17,7 +17,7 @@ limitations under the License.
 package converters
 
 import (
-	"github.com/Azure/azure-sdk-for-go/profiles/latest/compute/mgmt/compute"
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2020-06-01/compute"
```
devigned, line 20: 👍, lock in the version

controllers/azuremachine_controller.go
devigned: Agreed. Another PR to track the MHC work as well as an e2e test would be super cool

CecileRobertMichon:
/hold

seeing some weirdness with AZs

CecileRobertMichon:
Ok so I've confirmed that zones don't work as they should in the master branch so this isn't a regression introduced in this PR... I'm okay with merging it as-is and I can work on fixing zones separately.

What I'm seeing: 

My cluster in westus2 with VM size `Standard_D2_v3` is not setting zones on VMs even though westus2 has 3 zones. AzureCluster has no failure domains. After adding debug logs to the setFailureDomains func, I saw:

```
I0728 00:35:52.188660     108 azurecluster_reconciler.go:220] controllers/AzureCluster "msg"="zone length" "AzureCluster"="az-cluster-westus2" "cluster"="az-cluster-westus2" "namespace"="default" "zone length"=0
```

which means the resource SKU cache is returning 0 zones for that location.

I fixed a potential nil pointer reference in the last commit, will squash + cancel the hold once I get another lgtm to facilitate review on the changed bits.

alexeldeib:
I see the bug in resource cache, it's a bit silly. I'll probably move context back into the constructor for now and re-order things a bit, you'll see why

cloud/scope/machine.go
```diff
@@ -19,6 +19,7 @@ package scope
 import (
 	"context"
 	"encoding/base64"
+	"github.com/Azure/go-autorest/autorest/to"
```
alexeldeib, line 22: we're importing k8s.io/utils/pointer as well on L30, choose one perhaps?

cloud/scope/machine.go
```diff
@@ -87,6 +88,25 @@ type MachineScope struct {
 	AzureMachine *infrav1.AzureMachine
 }
 
+// VMSpecs returns the VM specs.
+func (m *MachineScope) VMSpecs() []azure.VMSpec {
```
alexeldeib, line 92: why does this return an array?

alexeldeib:
generally lgtm, two minor comments

cloud/scope/machine.go
CecileRobertMichon: I actually went back and forth on this but in the end I decided to stick with an array because 1) It's consistent with the others 2) I could see a scenario where we support a VM jumpbox / "bastion" in the future for ssh and it would require creating 2 VMs per AzureMachine.

cloud/scope/machine.go
CecileRobertMichon: true let me fix that

CecileRobertMichon: I kept autorest/to because pointers doesn't have pointer to string

controllers/azuremachine_reconciler.go
```diff
@@ -69,253 +58,62 @@ func newAzureMachineService(machineScope *scope.MachineScope, clusterScope *scop
...
-	}
-
-	klog.V(2).Infof("Found VM for AzureMachine %s", s.machineScope.Name())
-
-	return vm, nil
-}
-
-// getVirtualMachineZone gets a random availability zones from available set,
-// this will hopefully be an input from upstream machinesets so all the vms are balanced
-func (s *azureMachineService) getVirtualMachineZone(ctx context.Context) (string, error) {
```
CecileRobertMichon, line 162: So I thought it was okay to remove this and that it would automatically get taken care of by failure domains, but cluster api actually doesn't spread Machines on failure domains that are part of a machineDeployment/machineSet (only control plane machines part of KubeadmControlPlane). There's actually an open issue to potentially enable machineSets to spread machines over failure domains https://github.com/kubernetes-sigs/cluster-api/issues/3358

So essentially by removing this, it would cause worker node VMs to not be assigned a Zone until #3358 is implemented. This is potentially a regression. Reasons I'm considering accepting that temporary regression right now:
1) this doesn't actually spread VMs on zones, it picks a "random" zone (not that random it just picks the first one, see line 193).
2) The right way to spread machines part of a MachineSet on zones is with https://github.com/kubernetes-sigs/cluster-api/issues/3358, not having this hacky workaround might be a forcing function to actually do it right.
3) There is another workaround apparently which is to create one machine deployment per failure domain and scale each one independently (documented in https://github.com/kubernetes-sigs/cluster-api/issues/3358)
4) Doing this random AZ assignment at the VM level means the failureDomain on the machine spec is not set, which might cause friction with the cluster api cluster-autoscaler implementation and also means this wouldn't be compatible with a future implementation of failure domains on MachineSet (ie. existing VMs would break as we try to mutate the zone)
5) There is an in-progress MachinePool failure domains implementation which might be a good enough solution for worker nodes having AZs for now (ie. "if you need AZs, use vmss or use several deployments).

Thoughts? Happy to talk through this, it's a lot of writing :)

CecileRobertMichon: filed https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/844 to clarify docs

devigned: I vote to remove `getVirtualMachineZone`. The non-deterministic behavior of assigning zones is unexpected from the users' point of view and can cause behavioral changes amongst VMs in a machine deployment.

cloud/scope/machine.go
alexeldeib: 2 VMs per AzureMachine? That feels aggressive...I would imagine something like 1 per cluster (or perhaps 1 per pool if same subnet jumpbox is required). 

I don't see consistency as a super strong argument personally, but given the bastion scenario I could see us reusing the same service so I guess it makes sense.

cloud/scope/machine.go
```diff
@@ -299,11 +329,12 @@ func (m *MachineScope) Close(ctx context.Context) error {
 // the value from AzureMachine takes precedence.
 func (m *MachineScope) AdditionalTags() infrav1.Tags {
 	tags := make(infrav1.Tags)
-
 	// Start with the cluster-wide tags...
 	tags.Merge(m.ClusterDescriber.AdditionalTags())
 	// ... and merge in the Machine's
 	tags.Merge(m.AzureMachine.Spec.AdditionalTags)
+	// Set the cloud provider tag
+	tags[infrav1.ClusterAzureCloudProviderTagKey(m.ClusterName())] = string(infrav1.ResourceLifecycleOwned)
```
alexeldeib, line 337: we expect this to be permissive about state changes (cloud provider tag)? when?

CecileRobertMichon: what do you mean by permissive about state changes? This was previously in https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/827/files#diff-b2813f64759b38fd374267409eaaca65L109 just moved it inside the func

alexeldeib: I was just looking at https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/cf2473927944a9585de7ae7d80abb2ac3369739b/api/v1alpha3/tags.go#L85-L96

I was reading that as `ClusterAzureCloudProviderTagKey` was for things like e.g. the ingress load balancer. I wasn't sure if the cloud provider would edit the VMs here

alexeldeib: pretty minor issue, just my curiosity on the usage of these

CecileRobertMichon: yeah I'm not really sure why/how that tag is used for VMs, just preserved the existing tags

PR #828, opened 2020-07-24
Add validation on CLUSTER_NAME

### LINKED ISSUE #678: Azure resource names do not follow conventions
We should audit resource names in CAPZ to ensure that we are following [best practices](https://docs.microsoft.com/en-us/azure/cloud-adoption-framework/ready/azure-best-practices/naming-and-tagging).

Beyond following convention, not naming resources properly can lead to issues: Since `{CLUSTER_NAME}` is being used in a lot of places to name Azure resources. However, these resources have their own validation rules. So even if the K8s cluster resource can have a name up to 253 chars, we need to validate that it will be short enough to be appended to a suffix or prefix for any other resource where it is used, such as `${CLUSTER_NAME}-node-routetable`, 
`${CLUSTER_NAME}-control-plane-2`. We should enforce validation on resource names. For example, some resources cannot contain special characters such as `.`. For those, we should either 1) not rely on `$CLUSTER_NAME` or 2) enforce the same validation on `CLUSTER_NAME`.

Note that changing names of resources might be a breaking change in some cases where we look for a resource by name without storing its ID in Spec (eg. load balancers) so we need to be careful about handling backwards compatibility.

See also https://kubernetes.slack.com/archives/CEX9HENG7/p1590166717280400

nader-ziada:
validations for  cluster name to be used as prefix for naming other azure resources such as VMs, VNET, and so on
- max length of 44 characters
- no special characters allowed

Fixes #678 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Add validation on Cluster Name to limit length and not accept special characters
```

### DIFF
--- a/api/v1alpha3/azurecluster_validation.go
+++ b/api/v1alpha3/azurecluster_validation.go
@@ -26,6 +26,12 @@ import (
 )
 
 const (
+	// can't use: \/"'[]:|<>+=;,.?*@&, Can't start with underscore. Can't end with period or hyphen.
+	// not using . in the name to avoid issues when the name is part of DNS name
+	clusterNameRegex = `^[a-z][a-z0-9-]{0,44}[a-z0-9]$`
+	// max length of 44 to allow for cluster name to be used as a prefix for VMs and other resources that
+	// have limitations as outlined here https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules
+	clusterNameMaxLength = 44
 	// obtained from https://docs.microsoft.com/en-us/rest/api/resources/resourcegroups/createorupdate#uri-parameters
 	resourceGroupRegex = `^[-\w\._\(\)]+$`
 	// described in https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules
@@ -36,6 +42,7 @@ const (
 // validateCluster validates a cluster
 func (c *AzureCluster) validateCluster() error {
 	var allErrs field.ErrorList
+	allErrs = append(allErrs, c.validateClusterName()...)
 	allErrs = append(allErrs, c.validateClusterSpec()...)
 	if len(allErrs) == 0 {
 		return nil
@@ -51,6 +58,24 @@ func (c *AzureCluster) validateClusterSpec() field.ErrorList {
 	return validateNetworkSpec(
 		c.Spec.NetworkSpec,
 		field.NewPath("spec").Child("networkSpec"))
+}
+
+// validateClusterName validates ClusterName
+func (c *AzureCluster) validateClusterName() field.ErrorList {
+	var allErrs field.ErrorList
+	if len(c.Name) > clusterNameMaxLength {
+		allErrs = append(allErrs, field.Invalid(field.NewPath("metadata").Child("Name"), c.Name,
+			fmt.Sprintf("Cluster Name longer than allowed length of %d characters", clusterNameMaxLength)))
+	}
+	if success, _ := regexp.MatchString(clusterNameRegex, c.Name); !success {
+		allErrs = append(allErrs, field.Invalid(field.NewPath("metadata").Child("Name"), c.Name,
+			fmt.Sprintf("Cluster Name doesn't match regex %s, can contain only lowercase alphanumeric characters and '-', must start/end with an alphanumeric character",
+				clusterNameRegex)))
+	}
+	if len(allErrs) == 0 {
+		return nil
+	}
+	return allErrs
 }
 
 // validateNetworkSpec validates a NetworkSpec
--- a/docs/development.md
+++ b/docs/development.md
@@ -219,6 +219,8 @@ The steps below are provided in a convenient script in [hack/create-dev-cluster.
 CLUSTER_NAME=<my-capz-cluster-name> ./hack/create-dev-cluster.sh
 ```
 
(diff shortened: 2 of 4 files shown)


docs/development.md
```diff
@@ -219,6 +219,8 @@ The steps below are provided in a convenient script in [hack/create-dev-cluster.
 CLUSTER_NAME=<my-capz-cluster-name> ./hack/create-dev-cluster.sh
 ```
 
+   NOTE: `CLUSTER_NAME` can only include letters, numbers, and hyphens and can't be longer than 44 characters.
```
CecileRobertMichon, line 222: Are there any user facing docs that should also be updated? Probably the CAPI book / quickstart at least.

nader-ziada: couldn't find a better spot in docs here, but I can add a PR in cluster-api for the quick start

CecileRobertMichon:
Can you please validate that this doesn't break existing clusters and call that out in the PR description for future reference?

code lgtm

nader-ziada:
/hold

for more testing

nader-ziada:
/hold cancel

updated the regex

api/v1alpha3/azurecluster_validation.go
```diff
@@ -26,6 +26,11 @@ import (
 )
 
 const (
+	// can't use: \/"'[]:|<>+=;,?*@&, Can't start with underscore. Can't end with period or hyphen.
```
devigned, line 29: What happens with the DNS entry for the cluster if the cluster name contains a period?

nader-ziada: the regex actually doesn't allow for . for that reason, I'll update the comment

docs/development.md
CecileRobertMichon: yeah let's PR that in the quick start

nader-ziada: https://github.com/kubernetes-sigs/cluster-api/pull/3423

api/v1alpha3/azurecluster_validation.go
```diff
@@ -53,6 +60,23 @@ func (c *AzureCluster) validateClusterSpec() field.ErrorList {
...
+// validateClusterName validates ClusterName
+func (c *AzureCluster) validateClusterName() field.ErrorList {
+	var allErrs field.ErrorList
+	if len(c.Name) > clusterNameMaxLenght {
+		allErrs = append(allErrs, field.Invalid(field.NewPath("metadata").Child("Name"), c.Name,
+			fmt.Sprintf("Cluster Name longer than allowed length of %d characters", clusterNameMaxLenght)))
+	}
+	if success, _ := regexp.MatchString(clusterNameRegex, c.Name); !success {
+		allErrs = append(allErrs, field.Invalid(field.NewPath("metadata").Child("Name"), c.Name,
+			fmt.Sprintf("Cluster Name doesn't match regex %s", clusterNameRegex)))
```
CecileRobertMichon, line 72: nit: can we make this error message a little more user friendly and instead of printing the regex (or in addition to), list the constraints, ie. `Cluster name cannot contain '\/"'[]:|<>+=;,?*@&.', start with an underscore, or end with a hyphen.`

api/v1alpha3/azurecluster_validation.go
CecileRobertMichon: `Can't end with period or hyphen.` -> if the cluster name can't contain a period at all should we move `.` to the list of forbidden symbols and remove it from "end with"?

CecileRobertMichon:
sorry I said lgtm and then added more comments when taking another look 😬

api/v1alpha3/azurecluster_validation.go
nader-ziada: changed the error message

nader-ziada:
> sorry I said lgtm and then added more comments when taking another look 😬

no worries, better catch it now before merging

api/v1alpha3/azurecluster_validation.go
nader-ziada: added the `.` and there is also a comment in the following line

api/v1alpha3/azurecluster_validation.go
```diff
@@ -26,6 +26,12 @@ import (
 )
 
 const (
+	// can't use: \/"'[]:|<>+=;,.?*@&, Can't start with underscore. Can't end with period or hyphen.
+	// not using . in the name to avoid issues when the name is part of DNS name
+	clusterNameRegex = `^[a-z][a-z0-9-]{0,44}[a-z0-9]$`
+	// max length of 44 to allow for cluster name to be used as a prefix for VMs and other resources that
+	// have limitations as outlined here https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules
+	clusterNameMaxLenght = 44
```
alexeldeib, line 34:
```suggestion
	clusterNameMaxLength = 44
```

alexeldeib:
👍

out of scope for this, we should try to be stylistically consistent on field name capitalization in validation webhook errors e.g. `field.NewPath("metadata").Child("Name")` (metadata vs Name). lots of existing examples on master.

api/v1alpha3/azurecluster_validation.go
nader-ziada: fixed

PR #830, opened 2020-07-25
💎  Refactor security groups service spec

CecileRobertMichon:
**What this PR does / why we need it**: refactor security groups service according to patterns from #757.

~TODO: tests~

Fixes partially #757.

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Refactor security groups service spec
```


CecileRobertMichon:
TODO: need to check if the default ingress rules get set in the azurecluster spec.

cloud/scope/cluster.go
```diff
@@ -283,3 +299,31 @@ func (s *ClusterScope) SetFailureDomain(id string, spec clusterv1.FailureDomainS
...
+
+func (s *ClusterScope) ControlPlaneIngressRules() infrav1.IngressRules {
+	if s.ControlPlaneSubnet().SecurityGroup.IngressRules != nil {
+		return s.ControlPlaneSubnet().SecurityGroup.IngressRules
+	}
+	return infrav1.IngressRules{
+		&infrav1.IngressRule{
+			Name:             "allow_ssh",
+			Description:      "Allow SSH",
+			Priority:         100,
```
devigned, line 311: What would you think about having a higher base `Priority`. When starting with 100, there isn't much room to make rules ahead of this. Perhaps, 2000, or some other arbitrary, configurable base.

This feature probably shouldn't be in this refactor, just remembered seeing this change.

CecileRobertMichon: Not a bad idea. Not sure what the implication would be for reconciling pre-existing clusters with the old priority though so will look into this separately. These rules can be overridden by the user with a set of custom rules so there is at least that workaround in the meantime.

controllers/azurecluster_reconciler.go
```diff
@@ -71,58 +65,42 @@ func newAzureClusterReconciler(scope *scope.ClusterScope) *azureClusterReconcile
...
 	if err := r.createOrUpdateNetworkAPIServerIP(); err != nil {
 		return errors.Wrapf(err, "failed to create or update network API server IP for cluster %s in location %s", r.scope.ClusterName(), r.scope.Location())
 	}
 
 	if err := r.setFailureDomainsForLocation(ctx); err != nil {
-		return errors.Wrapf(err, "failed to get availability zones for cluster %s in location %s", r.scope.ClusterName(), r.scope.Location())
+		return errors.Wrapf(err, "failed to get availability zones")
 	}
 
+	r.scope.SetControlPlaneIngressRules()
```
CecileRobertMichon, line 76: this should probably move to a webhook but because the APIServer port is part of Cluster, not AzureCluster so requires a little more work, will refrain from changing that for now leave it as-is (just moved the below generateControlPlaneIngressRules func to cluster.go)

CecileRobertMichon:
I'll rebase once reviewers have had a chance to review the latest commit

CecileRobertMichon:
@devigned let me know when you've had a chance to take look at the new changes so I can rebase

PR #831, opened 2020-07-27
docs: add troubleshooting examples debugging missing nodes

jackfrancis:

This PR adds some supporting troubleshooting documentation to get a user started debugging why a cluster node did not come online.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note

```

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -70,9 +70,16 @@ If as a result of a new cluster create operation, or as a result of adding a new
 $ export CLUSTER_RESOURCE_GROUP=my-cluster-rg
 $ export VM_PREFIX=my-cluster-md-0-
 $ export KUBECONFIG=/Users/me/.kube/my-cluster.kubeconfig
-$ $ for vm in $(az vm list -g $CLUSTER_RESOURCE_GROUP |  jq -r --arg VM_PREFIX "${VM_PREFIX}" '.[] | select (.name | startswith($VM_PREFIX)).name'); do kubectl get node $vm 2>&1 >/dev/null && continue || echo node $vm did not join the cluster; done
+$ for vm in $(az vm list -g $CLUSTER_RESOURCE_GROUP | jq -r --arg VM_PREFIX "${VM_PREFIX}" '.[] | select (.name | startswith($VM_PREFIX)).name'); do kubectl get node $vm 2>&1 >/dev/null && continue || echo node $vm did not join the cluster; done
 Error from server (NotFound): nodes "my-cluster-md-0-8qlrg" not found
 node my-cluster-md-0-8qlrg did not join the cluster
+```
+
+The above assumes nodes as "machine" resources. If you're using "machinepool" resources:
+
+```
+$ export VMSS_NAME=$(az vmss list -g $CLUSTER_RESOURCE_GROUP | jq -r --arg VM_PREFIX "${VM_PREFIX}" '.[] | select (.name | startswith($VM_PREFIX)).name')
+$ for vm in $(az vmss list-instances -g $CLUSTER_RESOURCE_GROUP -n $VMSS_NAME | jq -r --arg VM_PREFIX "${VM_PREFIX}" '.[] | select (.name | startswith($VM_PREFIX)).name'); do kubectl get node $vm 2>&1 >/dev/null && continue || echo node $vm did not join the cluster; done
 ```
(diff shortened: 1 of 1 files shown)


docs/troubleshooting.md
```diff
@@ -60,3 +60,44 @@ As part of [CI](../scripts/ci-e2e.sh) there is a [log collection script](hack/..
...
+
+### Nodes did not come online
+
+If as a result of a new cluster create operation, or as a result of adding a new machine or machinepool resource to an existing cluster, one or more nodes did not join the cluster, you can use some of the above guidance to SSH into the VM(s) and debug what happened. First, let's find out which VMs were created but failed to join the cluster by introspecting all VMs in the cluster resource group, and comparing them to the nodes present in the cluster:
+
+```
+$ export CLUSTER_RESOURCE_GROUP=my-cluster-rg
+$ export VM_PREFIX=my-cluster-md-0-
+$ export KUBECONFIG=/Users/me/.kube/my-cluster.kubeconfig
+$ $ for vm in $(az vm list -g $CLUSTER_RESOURCE_GROUP |  jq -r --arg VM_PREFIX "${VM_PREFIX}" '.[] | select (.name | startswith($VM_PREFIX)).name'); do kubectl get node $vm 2>&1 >/dev/null && continue || echo node $vm did not join the cluster; done
```
CecileRobertMichon, line 73: this only works for VMs, not vmss (ie. not MachinePool) right?

jackfrancis: Right. I'll add an equivalent example for VMSS.

CecileRobertMichon: wouldn't `kubectl get machines` / `kubectl get machinepool`  `grep -v Ready` have the same effect without needing az cli and calling to Azure? In your case did the machine show as Ready?

CecileRobertMichon: Sorry, `grep -v Running` *

jackfrancis: I honestly didn't look at the `machine` resource at all. This was how my brain worked:

1) create cluster w/ desired node count
2) wait for nodes to come online, after a while noticed that there was one missing node
3) how many actual VMs are in my resource group? it's 20
4) O.K., so which one didn't register as a node?

FWIW

CecileRobertMichon: Right, I think both are valid, you went to the RG as a first instinct because you're familiar with Azure. For users who aren't as comfortable with Azure specific stuff it'd be nice to document how all this stuff can be done without needing to care about the underlying infrastructure... `kubectl get azuremachine` should show you the VM status while `kubectl get machine` should show you the status of the machine from a k8s perspective.

jsturtevant: I think both perspectives are valid here.  In reality will probably need to look at both CAPZ CRD and the underlying infra. Atleast that's how I have approached it so far, I look at the crd's for a quick understanding of what CAPZ thinks is going on then I look at the azure system to see what is happening.  It would be worth add a small section to see how this vms in azure related to the CRD's of capz.

jsturtevant:
Love the idea of giving use case driven debugging tips.

My thoughts are if you use the CAPZ crd's you could get rid of the  `export CLUSTER_RESOURCE_GROUP=my-cluster-rg` and `export VM_PREFIX=my-cluster-md-0-` which could then be automated into a tool similar to what is in https://github.com/kubernetes-sigs/cluster-api-provider-azure/tree/master/hack/debugging.  

I use the ssh and map tool all the time when debuggin and would find something similiar useful too. I had this problem of nodes coming online just yesterday :smile:

CecileRobertMichon:
@jackfrancis once https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/901 merges, consider changing these instructions looking at boot diagnostics from the portal or the Azure CLI (https://docs.microsoft.com/en-us/cli/azure/vm/boot-diagnostics?view=azure-cli-latest#az-vm-boot-diagnostics-get-boot-log)

CecileRobertMichon:
@jackfrancis are you still planning on getting this one in?

devigned:
@jackfrancis gentle nudge on this. We are happy to take over from current state if you would prefer.

CecileRobertMichon:
Rewrote this with new options in #1232 

/close

Issue #820, opened 2020-07-23
[e2e] Add e2e test for Availability Zones / Failure domains


CecileRobertMichon:
We should add a test to validate that Failure Domains work as expected with Azure. 

Expected behavior is:

- By default, the Cluster should spread VMs on all available zones
- If the user explicitly defines a zone for a Machine (`Machine`, _not_ `AzureMachine`) then that zone should be used for that VM.

CecileRobertMichon:
@gab-satchi thanks for picking this up! Just a heads up, I'm working on a refactor of the VM service right now as part of #757  that will remove a lot of the old AvailabilityZones code so things might change a bit in the coming week in case you want to wait for that (not a blocker strictly speaking, it shouldn't conflict with any tests and should be equivalent behavior wise).

gab-satchi:
sounds good. thanks for the heads up

CecileRobertMichon:
PR is in flight https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/827

PR #835, opened 2020-07-28
🐛 sku api should use resourceType, not kind

alexeldeib:

 

Not sure how I goofed on this, the sku API pivots on resourceType, not kind. This always was filtering out all resources. 

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cloud/services/resourceskus/cache.go
+++ b/cloud/services/resourceskus/cache.go
@@ -84,7 +84,7 @@ func (c *Cache) refresh(ctx context.Context, location string) error {
 // enhancing this function to handle restrictions (e.g. SKU not
 // supported in region), which is why it returns an error and not a
 // boolean.
-func (c *Cache) Get(ctx context.Context, name string, kind Kind) (SKU, error) {
+func (c *Cache) Get(ctx context.Context, name string, kind ResourceType) (SKU, error) {
 	if c.data == nil {
 		if err := c.refresh(ctx, c.location); err != nil {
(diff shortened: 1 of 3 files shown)


cloud/services/resourceskus/cache.go
```diff
@@ -99,22 +99,6 @@ func (c *Cache) Get(ctx context.Context, name string, kind Kind) (SKU, error) {
 	return SKU{}, fmt.Errorf("resource sku with name '%s' and category '%s' not found", name, string(kind))
 }
 
-// List returns a list of all known SKUs, useful for further filtering.
-func (c *Cache) List(ctx context.Context, name string, kind Kind) ([]SKU, error) {
```
alexeldeib, line 103: unused

devigned:
Just a quick note that the unit tests here were not super helpful to determine the validity of the feature.

CecileRobertMichon:
@devigned good point. Do have ideas around how to make them more helpful or are you saying we should really do #820?

devigned:
One way to make the tests better is record live traffic and then use that in unit tests. I've seen this done multiple times and req / res get stale, test are really brittle, etc... I like the idea of going down that path, but it is a path.

I think the only way we can honestly say a feature works, is to have robust e2e tests that replicate the user experience and are validated through user observable data. Prove it with a test.

PR #839, opened 2020-07-28
fix: load kustomize_substitutions before envsubst

devigned:

With #829 the kustomize_substitutions in the tilt settings file were not being loaded into the env or replaced in the yaml before running `envsubst`. This PR fixes this problem by loading them into the env prior to executing `envsubst` via `kustomizesub`.

/cc @CecileRobertMichon and @jsturtevant  

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -151,11 +151,11 @@ COPY manager .
 # Build CAPZ and add feature gates
 def capz():
     # Apply the kustomized yaml for this provider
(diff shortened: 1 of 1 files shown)


jsturtevant:
I confirmed it works in my set up.

/lgtm

PR #841, opened 2020-07-28
Fix issue with cluster paused map function

### LINKED ISSUE #840: AzureClusterReconciler's watch on Cluster for unpause notifications is not working
/kind bug

1. Pause a `Cluster`
1. Unpause it
1. The `AzureCluster` is not reconciled

1. The `AzureCluster` is reconciled

We are passing an empty AzureCluster's GroupVersionKind() to util.ClusterToInfrastructureMapFunc(), which means the helper will never match on AzureCluster and it always returns an empty list.

**Environment:**

- cluster-api-provider-azure version: v0.4.6

ncdc:

Switch to using the API package's GroupVersion to supply the
GroupVersionKind for the map function, as azCluster's APIVersion and
Kind fields were empty, resulting in the map function never matching on
AzureCluster.

Fixes #840

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Fixed an issue where AzureClusters were not being reprocessed when a Cluster is unpaused.
```

### DIFF
--- a/controllers/azurecluster_controller.go
+++ b/controllers/azurecluster_controller.go
@@ -51,10 +51,9 @@ type AzureClusterReconciler struct {
 }
 
 func (r *AzureClusterReconciler) SetupWithManager(mgr ctrl.Manager, options controller.Options) error {
-	azCluster := &infrav1.AzureCluster{}
 	c, err := ctrl.NewControllerManagedBy(mgr).
 		WithOptions(options).
-		For(azCluster).
+		For(&infrav1.AzureCluster{}).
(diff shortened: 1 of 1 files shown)


ncdc:
@vincepri @benmoss

CecileRobertMichon:
I wonder if there's a test we can add to validate this behavior? Right now we're only testing that pause works but not unpause

devigned:
Oh no... Thank you, @ncdc! That's a great catch.

@CecileRobertMichon, I should have added a test to create a paused cluster resource, then update it to not be paused, then verify a call to reconcile was made for the related `AzureCluster` resource. Adding an issue.

Issue #780, opened 2020-07-10
[e2e] Add more CAPI e2e test specs


CecileRobertMichon:
Currently we are only running a few of the specs in https://github.com/kubernetes-sigs/cluster-api/tree/master/test/e2e as part of the CAPZ spec (see https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/test/e2e/capi_test.go).

We should add as many of the test specs as possible and make sure they run successfully on CAPZ (some might need fixing some stuff before they can run).

note: to test in PR CI, use `/test pull-cluster-api-provider-azure-capi-e2e` and  set `GINKGO_FOCUS` to `Cluster API E2E tests` when testing locally.

CecileRobertMichon:
@nader-ziada just fyi @devigned already enabled the self-hosted one in https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/843/files#diff-cf63d4849f8ccd526cbe6e442ed9c72cR94, so you can focus on the others

nader-ziada:
> @nader-ziada just fyi @devigned already enabled the self-hosted one in https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/843/files#diff-cf63d4849f8ccd526cbe6e442ed9c72cR94, so you can focus on the others

I think its more involved than enabling it, the cluster on azure can't find the generated image and fails, but anyways will work on the other tests

devigned:
@nader-ziada, as you point out, the pivot clusters fail b/c they run into an image backoff error due to the fact we are only pushing images into kind, and not to a remote repository. I'm taking a look at that right now.

devigned:
@nader-ziada see #852

PR #843, opened 2020-07-28
🐛 fix group, version, kind not being passed into util.ClusterToObjectsMapper

### LINKED ISSUE #842: Fix and verify unpaused Cluster triggers reconcile for dependent resources
/kind feature

The way unpause notifications have been implemented is incorrect and will not capture cluster unpause events. See #841. We should correct the rest of the unpause Cluster watches and author tests to verify they work.

devigned:
The way unpause notifications have been implemented incorrectly and will not capture cluster unpause events. See #841. This PR should correct the rest of the unpause Cluster watches

Todo:
- [x] add tests to verify behavior

Fixes #842 

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
fix group, version, kind not being passed into util.ClusterToObjectsMapper to trigger reconciles when Clusters are unpaused
```

### DIFF
--- a/cloud/services/networkinterfaces/networkinterfaces_test.go
+++ b/cloud/services/networkinterfaces/networkinterfaces_test.go
@@ -24,13 +24,14 @@ import (
 
 	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1alpha3"
 	azure "sigs.k8s.io/cluster-api-provider-azure/cloud"
-	"sigs.k8s.io/cluster-api-provider-azure/internal/test/matchers"
+	gomockinternal "sigs.k8s.io/cluster-api-provider-azure/internal/test/matchers/gomock"
 
 	"github.com/Azure/go-autorest/autorest"
 	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/golang/mock/gomock"
 	"github.com/google/go-cmp/cmp"
 	. "github.com/onsi/gomega"
+
 	"sigs.k8s.io/cluster-api-provider-azure/cloud/services/loadbalancers/mock_loadbalancers"
 	"sigs.k8s.io/cluster-api-provider-azure/cloud/services/networkinterfaces/mock_networkinterfaces"
 	"sigs.k8s.io/cluster-api-provider-azure/cloud/services/publicips/mock_publicips"
@@ -140,7 +141,7 @@ func TestReconcileNetworkInterface(t *testing.T) {
 				gomock.InOrder(
 					mSubnet.Get(context.TODO(), "my-rg", "my-vnet", "my-subnet").Return(network.Subnet{}, nil),
 					mLoadBalancer.Get(context.TODO(), "my-rg", "my-public-lb").Return(getFakeNodeOutboundLoadBalancer(), nil),
-					m.CreateOrUpdate(context.TODO(), "my-rg", "my-net-interface", matchers.DiffEq(network.Interface{
+					m.CreateOrUpdate(context.TODO(), "my-rg", "my-net-interface", gomockinternal.DiffEq(network.Interface{
 						Location: to.StringPtr("fake-location"),
 						InterfacePropertiesFormat: &network.InterfacePropertiesFormat{
 							EnableAcceleratedNetworking: to.BoolPtr(true),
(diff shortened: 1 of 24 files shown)


CecileRobertMichon:
As an extra validation, on top of controller unit tests, you might want to try enabling https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/test/e2e/capi_test.go#L94

It was failing on my first attempt to enable CAPI tests and this might very well be why

CecileRobertMichon:
Have you tried creating a cluster manually and doing a clusterctl move and then seeing that reconciliation is actually unpaused? Not sure if the capi test is failing because it still doesn't work or if it's faulty

test/e2e/capi_test.go
```diff
@@ -91,16 +91,15 @@ var _ = Describe("Running the Cluster API E2E tests", func() {
...
-	// 		return capi_e2e.SelfHostedSpecInput{
-	// 			E2EConfig:             e2eConfig,
-	// 			ClusterctlConfigPath:  clusterctlConfigPath,
-	// 			BootstrapClusterProxy: bootstrapClusterProxy,
-	// 			ArtifactFolder:        artifactFolder,
-	// 			SkipCleanup:           skipCleanup,
-	// 		}
-	// 	})
-	// })
+	Context("Running the self-hosted spec", func() {
```
CecileRobertMichon, line 94: /hold

Let's make sure we don't merge this PR until we have a passing signal on the capi e2e test or undo this change otherwise it will break the periodic job

devigned: For sure! Hopefully, it passes and we have one more test enabled.

cloud/services/networkinterfaces/networkinterfaces_test.go
```diff
@@ -140,7 +141,7 @@ func TestReconcileNetworkInterface(t *testing.T) {
 				gomock.InOrder(
 					mSubnet.Get(context.TODO(), "my-rg", "my-vnet", "my-subnet").Return(network.Subnet{}, nil),
 					mLoadBalancer.Get(context.TODO(), "my-rg", "my-public-lb").Return(getFakeNodeOutboundLoadBalancer(), nil),
-					m.CreateOrUpdate(context.TODO(), "my-rg", "my-net-interface", matchers.DiffEq(network.Interface{
+					m.CreateOrUpdate(context.TODO(), "my-rg", "my-net-interface", gomockinternal.DiffEq(network.Interface{
```
CecileRobertMichon, line 144: is it better to use gomockinternal? Asking because I've been using matchers.DiffEq everywhere so wondering if I should stop doing that

devigned: The are the same thing. I moved an the `internal/matchers` package to `internal/matchers/gomock`, b/c I needed to make matchers for gomega too.

devigned:
@CecileRobertMichon looks like pull-cluster-api-provider-azure-capi-e2e didn't pass. I'm going to dig into it a bit, but if it's not initially clear, I'm leaning toward commenting and taking it on in a new issue. Thoughts?

CecileRobertMichon:
@devigned sgtm, if that's the case let's tackle it as part of #780 
cc @nader-ziada

devigned:
I'm running the test locally. I'll have a definitive answer about what's going on shortly.

devigned:
@CecileRobertMichon I've commented out the CAPI Self-Host test again and opened #852 to address the issues causing the test to fail. I think this PR is ready for final review after passing tests.

devigned:
/hold
flaky test

devigned:
I think I've resolved the flaky test issue. If `azure-test` passes without complaining about logs, I think the change set is ready to go.

devigned:
> the exp package common tests should be the same as the non-exp ones, just need to test in two places for now but we should be able to get rid of them once these features move out of exp/ correct?

Correct. We should remove the exp test and add the controller names into the non-exp tests when out of "experimental".

> why delete the tests in azurecluster_controller_test but leave the ones in azuremachine_controller_test since it looks like the common tests cover both?

Good catch. That's extra code. I'll remove it.

PR #849, opened 2020-07-31
Configurable OS disk caching

### LINKED ISSUE #584: Configurable OS disk caching
/kind feature

Currently we don't set the disk caching type for OS disks (defaults to `None`).

The azure-sdk-for-go defines three caching types:

const (
	// CachingTypesNone ...
	CachingTypesNone CachingTypes = "None"
	// CachingTypesReadOnly ...
	CachingTypesReadOnly CachingTypes = "ReadOnly"
	// CachingTypesReadWrite ...
	CachingTypesReadWrite CachingTypes = "ReadWrite"
)
Let's enable user-configurable caching types by adding a new field in `AzureMachineSpec.OSDisk` and leveraging validation and defaulting webhooks. We should also define a sane default for control plane and worker OS disks.

Reference PR in AKS Engine: https://github.com/Azure/aks-engine/pull/2863

**Environment:**

jroden:
Users can now configure [OS disk caching](https://godoc.org/github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-12-01/compute#OSDisk) for azureMachine resources.  Validating and defaulting webhooks are also included.

Fixes #584

Thanks for the help! :smile: 

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
NONE
```


CecileRobertMichon:
/ok-to-test
@jroden should we tackle to Data Disks in a separate PR or are you planning on adding it to this one?

jroden:
> /ok-to-test
> @jroden should we tackle to Data Disks in a separate PR or are you planning on adding it to this one?

Good catch, let me add dataDisk to this PR :smile:

api/v1alpha2/types.go
```diff
@@ -431,6 +431,8 @@ type OSDisk struct {
 	OSType      string      `json:"osType"`
 	DiskSizeGB  int32       `json:"diskSizeGB"`
 	ManagedDisk ManagedDisk `json:"managedDisk"`
+	// +optional
+	CachingType string `json:"cachingType"`
```
CecileRobertMichon, line 435: why add this to the v1alpha2 package? All changes to the API should be in the current API version, v1alpha3

api/v1alpha3/azuremachine_webhook.go
```diff
@@ -36,6 +36,7 @@ func (m *AzureMachine) SetupWebhookWithManager(mgr ctrl.Manager) error {
 }
 
 // +kubebuilder:webhook:verbs=create;update,path=/validate-infrastructure-cluster-x-k8s-io-v1alpha3-azuremachine,mutating=false,failurePolicy=fail,matchPolicy=Equivalent,groups=infrastructure.cluster.x-k8s.io,resources=azuremachines,versions=v1alpha3,name=validation.azuremachine.infrastructure.cluster.x-k8s.io,sideEffects=None
+// +kubebuilder:webhook:verbs=create;update,path=/mutate-infrastructure-cluster-x-k8s-io-v1alpha3-azuremachine,mutating=true,failurePolicy=fail,matchPolicy=Equivalent,groups=infrastructure.cluster.x-k8s.io,resources=azuremachines,versions=v1alpha3,name=default.azuremachine.infrastructure.cluster.x-k8s.io,sideEffects=None
```
CecileRobertMichon, line 39: good catch

CecileRobertMichon:
lgtm aside from the v1alpha2 type change

/cc @alexeldeib for another review

api/v1alpha3/types.go
```diff
@@ -311,6 +311,8 @@ type OSDisk struct {
 	DiskSizeGB       int32             `json:"diskSizeGB"`
 	ManagedDisk      ManagedDisk       `json:"managedDisk"`
 	DiffDiskSettings *DiffDiskSettings `json:"diffDiskSettings,omitempty"`
+	// +optional
+	CachingType string `json:"cachingType"`
```
alexeldeib, line 315: I suspect this should be `` `json:"cachingType,omitempty"` ``? Need to test

jroden: In this case wouldn't omitempty only have an effect between the api request and mutating webhook (since we're defaulting that field)?  Also for my understanding-- are you referring to the effect of serializing an empty field on the virtualmachine service (if an empty field exists it might cause an error as opposed to a nil field)?  Thanks for the help! :smile:

api/v1alpha2/types.go
jroden: Good catch, this was a picnic error :sweat_smile:

api/v1alpha3/types.go
alexeldeib: you're correct about behavior with the webhook, I think my memory is picking up on an old controller-gen bug where `+optional` wasn't respected so `,omitempty` was actually required for the correct behavior.

jroden:
The previous failure was due to a timeout.  After retest, this failure is from to the loadBalancer test when `connecting directly to the external LB service`.  It looks like it is able to successfully connect to the lb service from a curl pod in the previous step. Would that mean the connection that's failing is originating from an e2e manager cluster?  Any advice or direction would be much appreciated, struggling a bit to find my bearings with this one :smile:

nader-ziada:
the `connecting directly to the external LB service` has been a little flakey

PR #851, opened 2020-07-31
Asserts control planes are spread across zones when possible

gab-satchi:
This PR modifies the e2e test that creates 3 control planes to ensure that the control planes are spread across the availability zones when possible. 

The assertion is skipped if:
- The tests are run in a location that doesn't support availability zones
- The number of AZs isn't 3. The spreading logic of KCP won't be deterministic if control planes isn't a multiple of the number of AZs

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

```release-note
NONE
```


gab-satchi:
Part of the work for #820 worker machines aren't spread over AZs currently so this change just verifies KCP control plane logic is working as expected. A follow up PR is needed for validating that worker machines can be created in specific zones by creating dedicated machine deployments in each zone.

test/e2e/azure_test.go
```diff
@@ -114,6 +116,24 @@ var _ = Describe("Workload cluster creation", func() {
 				WaitForMachineDeployments:    e2eConfig.GetIntervals(specName, "wait-worker-nodes"),
 			})
 
+			// TODO: Find alternative when the number of zones is > 1 but doesn't equal to number of control plane machines.
+			// Azure tends to have 3 Availability zones or none for most regions.
+			if len(cluster.Status.FailureDomains) == 3 {
```
CecileRobertMichon, line 121: how about having a helper function that returns the expectedFailureDomains based on the Azure region? That way we can also validate that the cluster status failure domains get set as expected

gab-satchi: Yeah I was hoping to do that at first but couldn't find a simple enough way to fetch zones for a location. The only one I came across was by listing the skus for a location and going through to see the available zones. I think something similar is done [here](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/cloud/services/resourceskus/cache.go#L121). New to Azure so unsure if I missed an easier way.

CecileRobertMichon: Since part of the behavior we want to test is fetching the zones themselves, and since I expect available zones in a region to be someone static, I'd be okay with hardcoding expected number of zones per Azure region? Or maybe start with the ones that we know have 3 zones for known VM sizes... This could get tricky though as zone support is per VM size so maybe what you have right now is good first step.

@alexeldeib thoughts?

alexeldeib: probably picking a static set is fine, particularly for those with 3 already. I uploaded a list for centralus as a gist if you want a reference, I also uploaded an all-region list but be warned -- the file is something like 10MB and crashed github multiple times for me (which is why I uploaded a single region).

centralus: https://gist.github.com/alexeldeib/c27d23b71e73a294caa32ecbdfb7bb25
all: https://gist.github.com/alexeldeib/6868e6e283937bc5658981d5bd9a1648

gab-satchi: Thanks for the suggestions. I uploaded a json file that's a trimmed down version of the output from `az vm list-skus -r "virtualMachines"  -z`. Left it as a json file in the e2e package so it doesn't crash any IDEs that try opening it as a go file.

gab-satchi:
seems to have failed further down the test with LB. 

nader-ziada:
looks good

CecileRobertMichon:
lgtm

Can you please move e2e/availableZonesPerLocation.json to e2e/data/availableZonesPerLocation.json ?

nader-ziada:
will try to rebase to get this merged in before all the other changes related to e2e

nader-ziada:
/lgtm

@CecileRobertMichon Gab made the requested changes, will let you approve if you think its all good

CecileRobertMichon:
/approve

Thanks!

PR #862, opened 2020-08-06
azureManagedControlPlane webhooks

### LINKED ISSUE #612: [managedcluster] validating webhook for DNS IP

jroden:
Adds defaulting & validating webhooks for amcp objects.

Fixes #612  #615 

Thanks for the help! :smile: 

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
action required: managedclusters service spec LoadBalancerSKU, NetworkPlugin, and NetworkPolicy changed from *string to string

This release includes defaulting and validating webhooks for azureManagedControlPlane, as well as modeling DNSServiceIP and LoadBalancerSKU as fields in the amcp crd  (both are optional).  NetworkPlugin is defaulted to 'azure', LoadBalancerSKU is defaulted to 'standard', and NetworkPolicy is defaulted to 'calico'.  DNSServiceIP is validated to ensure it represents an ip address.
```


nader-ziada:
looks good, but there was some discussion in #612 about the validation of DNS IP @alexeldeib can you please take a look on this PR

cloud/services/managedclusters/managedclusters.go
```diff
@@ -70,6 +69,9 @@ type Spec struct {
 
 	// ServiceCIDR is the CIDR block for IP addresses distributed to services
 	ServiceCIDR string
+
+	// dnsServiceIP is an IP address assigned to the Kubernetes DNS service
```
alexeldeib, line 73: go nit: capitalization

cloud/services/managedclusters/managedclusters.go
```diff
@@ -123,50 +125,36 @@ func (s *Service) Reconcile(ctx context.Context, spec interface{}) error {
 			},
 			AgentPoolProfiles: &[]containerservice.ManagedClusterAgentPoolProfile{},
 			NetworkProfile: &containerservice.NetworkProfileType{
-				NetworkPlugin:   containerservice.Azure,
-				LoadBalancerSku: containerservice.Standard,
+				NetworkPlugin:   containerservice.NetworkPlugin(*managedClusterSpec.NetworkPlugin),
```
alexeldeib, line 128: deref without nil check on managedClusterSpec NetworkPlugin/LoadBalancerSKU/NetworkPolicy. It's passthrough from the control plane which is optional, albeit set in the webhook. 

Could you make these three values non-pointer in the managedClusterSpec used by the service, remove the deref here, and conditionally set them if not null on the managedClusterSpec instead?

https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/862/files#diff-78feb4e2f24becff4de1a0bc4348e52bR108-R112

exp/api/v1alpha3/azuremanagedcontrolplane_types.go
```diff
@@ -48,20 +48,31 @@ type AzureManagedControlPlaneSpec struct {
...
 
-	// NetworkPlugin used for building Kubernetes network. Possible values include: 'Azure', 'Kubenet'. Defaults to Azure.
-	// +kubebuilder:validation:Enum=Azure;Kubenet
+	// NetworkPlugin used for building Kubernetes network.
+	// +kubebuilder:validation:Enum=azure;kubenet
+	// +optional
 	NetworkPlugin *string `json:"networkPlugin,omitempty"`
 
-	// NetworkPolicy used for building Kubernetes network. Possible values include: 'Calico', 'Azure'
-	// +kubebuilder:validation:Enum=Calico;Azure
```
alexeldeib, line 56: technically I think the capitalization is breaking but I'm 👍 on it

exp/api/v1alpha3/azuremanagedcontrolplane_types.go
```diff
@@ -48,20 +48,31 @@ type AzureManagedControlPlaneSpec struct {
...
 	NetworkPolicy *string `json:"networkPolicy,omitempty"`
 
 	// SSHPublicKey is a string literal containing an ssh public key base64 encoded.
 	SSHPublicKey string `json:"sshPublicKey"`
 
 	// DefaultPoolRef is the specification for the default pool, without which an AKS cluster cannot be created.
-	// TODO(ace): consider defaulting and making optional pointer?
 	DefaultPoolRef corev1.LocalObjectReference `json:"defaultPoolRef"`
+
+	// dnsServiceIP is an IP address assigned to the Kubernetes DNS service.
```
alexeldeib, line 67: nit: capitalization

exp/api/v1alpha3/azuremanagedcontrolplane_types.go
```diff
@@ -48,20 +48,31 @@ type AzureManagedControlPlaneSpec struct {
...
+	// It must be within the Kubernetes service address range specified in serviceCidr.
+	// +optional
+	DNSServiceIP *string `json:"dnsServiceIP,omitempty"`
+
+	// loadBalancerSKU is the SKU of the loadBalancer to be provisioned.
```
alexeldeib, line 72: ditto on capitalization

main.go
```diff
@@ -67,6 +67,7 @@ func init() {
 	_ = infrav1alpha3exp.AddToScheme(scheme)
 	_ = clusterv1.AddToScheme(scheme)
 	_ = clusterv1exp.AddToScheme(scheme)
+	// _ = expinfrastructureclusterxk8siov1alpha3.AddToScheme(scheme)
```
alexeldeib, line 70: remove?

exp/controllers/azuremangedcontrolplane_reconciler.go
```diff
@@ -98,14 +100,16 @@ func (r *azureManagedControlPlaneReconciler) Reconcile(ctx context.Context, scop
...
-		SSHPublicKey:  scope.ControlPlane.Spec.SSHPublicKey,
+		Name:            scope.ControlPlane.Name,
+		ResourceGroup:   scope.ControlPlane.Spec.ResourceGroup,
+		Location:        scope.ControlPlane.Spec.Location,
+		Tags:            scope.ControlPlane.Spec.AdditionalTags,
+		Version:         strings.TrimPrefix(scope.ControlPlane.Spec.Version, "v"),
+		NetworkPlugin:   scope.ControlPlane.Spec.NetworkPlugin,
+		NetworkPolicy:   scope.ControlPlane.Spec.NetworkPolicy,
+		SSHPublicKey:    scope.ControlPlane.Spec.SSHPublicKey,
+		DNSServiceIP:    scope.ControlPlane.Spec.DNSServiceIP,
```
alexeldeib, line 111: I don't see us setting serviceCIDR? That seems like a bug from my PR 🙁

cloud/services/managedclusters/managedclusters.go
```diff
@@ -123,50 +125,36 @@ func (s *Service) Reconcile(ctx context.Context, spec interface{}) error {
...
+			// HACK: set the last octet of the IP to .10
+			// This ensures the dns IP is valid in the service cidr without forcing the user
+			// to specify it in both the Capi cluster and the Azure control plane.
+			// https://golang.org/src/net/ip.go#L48
+			ip[15] = byte(10)
+			dnsIP := ip.String()
+			properties.NetworkProfile.DNSServiceIP = &dnsIP
 		} else {
-			return fmt.Errorf("invalid network policy: '%s'. Allowed options are 'calico' and 'azure'", *managedClusterSpec.NetworkPolicy)
+			properties.NetworkProfile.DNSServiceIP = managedClusterSpec.DNSServiceIP
```
alexeldeib, line 154: if we have both serviceCIDR and DNSServiceIP defined, please validate that DNSServiceIP is in the block defined by serviceCIDR and return an error if it's not.

https://golang.org/pkg/net/#IPNet.Contains

exp/api/v1alpha3/azuremanagedcontrolplane_types.go
alexeldeib: could you add  a release note?

cloud/services/managedclusters/managedclusters.go
alexeldeib: incidentally, I think a nil deref on one of these is the cause of the test failure

jroden: good catch, that is a cardinal sin of pointers :sweat_smile:

exp/controllers/azuremangedcontrolplane_reconciler.go
jroden: I think you're good :smile: -- I believe the owner cluster object is passed via ManagedControlPlaneScope to the reconciler from the amcp controller [here](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/exp/controllers/azuremangedcontrolplane_controller.go#L112).  It looks like the reconciler is setting the serviceCIDR in the managedClusterSpec [here](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/exp/controllers/azuremangedcontrolplane_reconciler.go#L133).

cloud/services/managedclusters/managedclusters.go
jroden: Initially I wanted to include that as one of the validations, the only problem is that the serviceCIDR field resides in the owner cluster object and the capi-webhook serviceAccount doesn't have permissions to get cluster objects.  I believe we ended up deciding to just validate that the DNSServiceIp field represents a valid ip.  Thoughts?

jroden:
> Looking good overall, mostly minor comments. Thanks for your work!

Absolutely!  I really enjoy working on this project and am learning a lot.
Unfortunately I'll likely have less time to contribute once my coursework picks up this semester so I may not be around as much :cry: 
@alexeldeib

cloud/services/managedclusters/managedclusters.go
alexeldeib: we don't need to do it in the webhook, but we should still fail fast in the reconciler itself (where we have both objects, as well as RBAC rights)

jroden: Good catch, that makes sense :smile:

jroden:
@alexeldeib The apidiff test is failing after we changed the loadbalancersku, networkplugin, and networkpolicy from *string to string in the managedcluster service.  I'm wondering if we should revert those changes or update the test case?  Thanks for the help! :smile:

CecileRobertMichon:
@jroden the apidiff test is actually not required to merge but helps us catch breaking API changes. In this case, it is a breaking change to change the type but since it's in the exp/ package it is acceptable. Can you please add "action required" to the release note so it gets flagged as a breaking change?

jroden:
> @jroden the apidiff test is actually not required to merge but helps us catch breaking API changes. In this case, it is a breaking change to change the type but since it's in the exp/ package it is acceptable. Can you please add "action required" to the release note so it gets flagged as a breaking change?

Just added, thanks for the help!

CecileRobertMichon:
@alexeldeib did you have any outstanding comments?

cloud/services/managedclusters/managedclusters.go
alexeldeib: still missing this one?

alexeldeib: nvm, found it 👍 https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/862/files#diff-78feb4e2f24becff4de1a0bc4348e52bR172

alexeldeib:
don't think so! thanks again 😃 come back any time 🙂 

/lgtm

Issue #825, opened 2020-07-23
move ClusterDescriber methods to CRD types


alexeldeib:
/kind cleanup

**Describe the solution you'd like**
Move implementation of ClusterDescriber interface onto CRD types directly, e.g. AzureCluster and AzureManagedControlPlane. This allows machine scope to use either one to extract join parameters as long as it implements the required interface, which I PoC'd in #822 

**Anything else you would like to add:**
See https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/822/files#diff-890b2d661da1708cb3d23d43fc548ceeR67 and associated implementation https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/822/files#diff-db9119511124ddd279659a96a3c579cfR116

This lets us start to take a duck typing approach to machine scope parameters and allow for alternate provider implementations to reuse AzureMachine more easily.

CecileRobertMichon:
We also need to start unit testing these files, maybe we can use that as a forcing function , or at least as a stepping stone to make it easier.

mboersma:
@alexeldeib is this still in progress? Anything we can help with?

alexeldeib:
> @alexeldeib is this still in progress? Anything we can help with?

The main reason for this is to enable https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/826, which I haven't worked on much recently. I'm poking at those parts again but I don't have an explicit timeline for that. 

I think right now https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1418 and https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1376 are the top issues for managed clusters. Once we tidy that up a bit, I'll probably circle back here.

unassigning for now

/remove-lifecycle stale
/unassign

PR #864, opened 2020-08-07
 💎 refactor subscription into cluster describer

alexeldeib:
refactor subscription into cluster describe, mostly setup for 825. See https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/822/files#diff-825ec7ff34f98855f2347c7f7769438dR210-R213 for example.

also cleans up our settings/credentials a bit (should be a no-op)

Setup for https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/825

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cloud/scope/clients.go
+++ b/cloud/scope/clients.go
@@ -17,11 +17,9 @@ limitations under the License.
 package scope
 
 import (
-	"fmt"
 	"strings"
 
 	"github.com/Azure/go-autorest/autorest"
-	"github.com/Azure/go-autorest/autorest/azure"
 	"github.com/Azure/go-autorest/autorest/azure/auth"
 )
 
@@ -38,39 +36,30 @@ const (
 
 // AzureClients contains all the Azure clients used by the scopes.
 type AzureClients struct {
-	Authorizer                 autorest.Authorizer
-	environment                string
+	Authorizer autorest.Authorizer
+	auth.EnvironmentSettings
 	ResourceManagerEndpoint    string
 	ResourceManagerVMDNSSuffix string
-	subscriptionID             string
-	tenantID                   string
-	clientID                   string
-	clientSecret               string
 }
 
 // CloudEnvironment returns the Azure environment the controller runs in.
 func (c *AzureClients) CloudEnvironment() string {
-	return c.environment
-}
-
-// SubscriptionID returns the Azure subscription id from the controller environment
-func (c *AzureClients) SubscriptionID() string {
-	return c.subscriptionID
+	return c.Environment.Name
 }
 
 // TenantID returns the Azure tenant id the controller runs in.
 func (c *AzureClients) TenantID() string {
-	return c.tenantID
+	return c.Values[auth.TenantID]
 }
 
 // ClientID returns the Azure client id from the controller environment
 func (c *AzureClients) ClientID() string {
-	return c.clientID
+	return c.Values[auth.ClientID]
 }
 
 // ClientSecret returns the Azure client secret from the controller environment
 func (c *AzureClients) ClientSecret() string {
-	return c.clientSecret
+	return c.Values[auth.ClientSecret]
 }
 
 func (c *AzureClients) setCredentials(subscriptionID string) error {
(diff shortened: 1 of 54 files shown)


alexeldeib:
XL PR, but it's a net reduction 😉

alexeldeib:
/unassign @CecileRobertMichon 
/assign @devigned 

sorry ghost Cecile :)

cloud/services/disks/service.go
```diff
@@ -38,6 +38,6 @@ type Service struct {
 func NewService(scope DiskScope) *Service {
 	return &Service{
 		Scope:  scope,
-		Client: NewClient(scope),
+		Client: NewClient(scope.SubscriptionID(), scope),
```
devigned, line 41: why not just scope?

cloud/services/groups/service.go
```diff
@@ -42,6 +43,6 @@ type GroupScope interface {
 func NewService(scope GroupScope) *Service {
 	return &Service{
 		Scope:  scope,
-		Client: NewClient(scope),
+		Client: NewClient(scope.SubscriptionID(), scope),
```
devigned, line 46: Why not just scope?

cloud/services/groups/client.go
```diff
@@ -39,8 +39,8 @@ type AzureClient struct {
 var _ Client = &AzureClient{}
 
 // NewClient creates a new VM client from subscription ID.
-func NewClient(auth azure.Authorizer) *AzureClient {
-	c := newGroupsClient(auth.SubscriptionID(), auth.BaseURI(), auth.Authorizer())
+func NewClient(subscriptionID string, auth azure.Authorizer) *AzureClient {
```
devigned, line 42: What if this was a `ScopedAuthorizer interface` or something? Basically, `Authorizer` + `Subscriber`.

cloud/services/scalesets/client.go
```diff
@@ -48,11 +48,11 @@ type AzureClient struct {
 var _ Client = &AzureClient{}
 
 // NewClient creates a new VMSS client from subscription ID.
-func NewClient(auth azure.Authorizer) *AzureClient {
+func NewClient(subscriptionID string, auth azure.Authorizer) *AzureClient {
 	return &AzureClient{
-		scalesetvms: newVirtualMachineScaleSetVMsClient(auth.SubscriptionID(), auth.BaseURI(), auth.Authorizer()),
-		scalesets:   newVirtualMachineScaleSetsClient(auth.SubscriptionID(), auth.BaseURI(), auth.Authorizer()),
-		publicIPs:   newPublicIPsClient(auth.SubscriptionID(), auth.BaseURI(), auth.Authorizer()),
+		scalesetvms: newVirtualMachineScaleSetVMsClient(subscriptionID, auth.BaseURI(), auth.Authorizer()),
```
devigned, line 53: This signature is used repeatedly. What if this was a struct or an interface. It probably could be the same type that `NewClient` takes.

cloud/services/groups/client.go
alexeldeib: I think something like this is probably better, I resisted initially because that smells to me like we should abstract scope a bit more?
I also wonder if we could invert our client instantiation. Create an Azure clientset one time per reconcile. something like:
```go
type ClientSet struct {
  sub string
  auth Authorizer
}

func NewClientSet(sub string, auth Authorizer) *ClientSet { ... }

fung (c *ClientSet) Groups() resources.GroupsClient // or the mocked interface
```

we could plausibly then also group the services, which would simplify mocking for tests.

similar example from hashi:
- client builder: https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/internal/clients/client.go
- corresponding client: https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/internal/services/redis/client/client.go

devigned: `ClusterDescriber.GetClientSet() (*ClientSet, error)` or something?

alexeldeib: added `SubscriptionAuthorizer` for now

devigned: I like the `ClientSet.Groups()` idea.

alexeldeib: > ClusterDescriber.GetClientSet() (*ClientSet, error) or something?

I probably would keep it inverted -- maybe something like
```go
func NewClientSet(sub string, auth Authorizer) *ClientSet { ... }

// similar to existing services
type Service interface {
}

type ServiceSet struct {
  *ClientSet // probably interface
  Groups() Service
  Networks() Service
  LoadBalancers() Service
  // etc...
}

func NewServiceSet(d clusterDescriber, c *ClientSet // or interface for mocking) *ServiceSet { ... }

// not necessarily defined on service set, but for illustrative purposes on usage
// our reconciler loops would be calling these repeatedly like they do today for individual services
func (s *ServiceSet) ReconcileGroup(ctx context.Context) error {
  return s.Groups().Reconcile()
}

```

alexeldeib: would need to consider some differentiation for machine vs cluster, but I like grouping things and doing them once anyway 😄

alexeldeib: @devigned as I'm working through the next steps to achieve #822 I think ClusterDescriber should exclude authorizer, and we can define an interface over scope which is the union of clusterdescriber and authorizer. I can either do that here or in the next pr and use this as a stepping stone.

cloud/scope/clients.go
```diff
@@ -79,46 +68,14 @@ func (c *AzureClients) setCredentials(subscriptionID string) error {
...
+	c.Values[auth.TenantID] = strings.TrimSuffix(c.Values[auth.TenantID], "\n")
 
-	c.Authorizer, err = settings.GetAuthorizer()
+	c.Authorizer, err = c.GetAuthorizer()
 	return err
 }
-
-// GetAzureDNSZoneForEnvironment returnes the DNSZone to be used with the
-// cloud environment, the default is the public cloud
-func GetAzureDNSZoneForEnvironment(environmentName string) string {
```
CecileRobertMichon, line 110: does `ResourceManagerVMDNSSuffix` get set to the right values from the `AZURE_ENVIRONMENT` var now?

alexeldeib:
/hold

not sure I love the way this ended up working as I sliced things off, will play around a bit if I can do it better

cloud/scope/clients.go
alexeldeib: Indeed:

https://github.com/Azure/go-autorest/blob/8f84370b53ea066bda881adc6e07455c58b4b7a6/autorest/azure/environments.go#L98
https://github.com/Azure/go-autorest/blob/master/autorest/azure/environments.go#L132
https://github.com/Azure/go-autorest/blob/8f84370b53ea066bda881adc6e07455c58b4b7a6/autorest/azure/environments.go#L166
https://github.com/Azure/go-autorest/blob/8f84370b53ea066bda881adc6e07455c58b4b7a6/autorest/azure/environments.go#L200

CecileRobertMichon: nice

cloud/services/groups/client.go
devigned: Why do you say ClusterDescriber should exclude authorizer? As we move toward multitenancy, the cluster with have an identity or the default env identity.

alexeldeib: we use clusterdescriber to build out infrastructure resources to reconcile. we use the identity only for service instantiation. I want to separate these concerns, I kind of like the approach above for that purpose.

It's also why I'd potentially like to put subscription ID somewhere else, because it's part of auth, not part of reconciliation

alexeldeib:
/close

I'll part this out or reopen it later

PR #870, opened 2020-08-10
🐛 fix requeue, labels in azure json reconciler

alexeldeib:

while investigating bootstrap failures, I can see this line get hit, and then the cloud provider secret doesn't get created until 15 min later. I'm a bit surprised the cluster can't be found: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/8b2866123ccce2d63c0df52c8507d0e2996857e9/controllers/azurejson_machinetemplate_controller.go#L109-L116

because the secret doesn't exist, capbk can't successfully get to this part of the reconcile loop, where it sets config.Status.Ready = true: https://github.com/kubernetes-sigs/cluster-api/blob/f8dd8a3bdba86bae3e72abdf2c8d7cbac9573cf7/bootstrap/kubeadm/controllers/kubeadmconfig_controller.go#L798-L834

so we don't hit this switch case: https://github.com/kubernetes-sigs/cluster-api/blob/f8dd8a3bdba86bae3e72abdf2c8d7cbac9573cf7/bootstrap/kubeadm/controllers/kubeadmconfig_controller.go#L241-L264

so the token doesn't get refreshed, but the bootstrap secret still gets stored with the original token as soon as we can find the secrets to generate the data. i'm not totally sure why the jsontemplate can't find AzureCluster and how that ends up with a 15min requeue, and it's not clear to me what happens to the originally generated capbk token. I would assume ttlcleaner kills it after 15minutes, but then idk where the other token comes from that does exist in the cluster (haven't followed the logic to see how that would get created).

regardless of the final answers, this is a bug that should be fixed. we should requeue on not finding the AzureCluster

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/controllers/azurejson_machinetemplate_controller.go
+++ b/controllers/azurejson_machinetemplate_controller.go
@@ -91,10 +91,8 @@ func (r *AzureJSONTemplateReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result
 		return ctrl.Result{}, nil
 	}
 
-	_, kind := infrav1.GroupVersion.WithKind("AzureCluster").ToAPIVersionAndKind()
-
 	// only look at azure clusters
-	if cluster.Spec.InfrastructureRef.Kind != kind {
+	if cluster.Spec.InfrastructureRef.Kind != "AzureCluster" {
 		log.WithValues("kind", cluster.Spec.InfrastructureRef.Kind).Info("infra ref was not an AzureCluster")
 		return ctrl.Result{}, nil
 	}
@@ -106,12 +104,8 @@ func (r *AzureJSONTemplateReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result
 		Name:      cluster.Spec.InfrastructureRef.Name,
 	}
 
-	err = r.Get(ctx, azureClusterName, azureCluster)
-	if err != nil {
-		if apierrors.IsNotFound(err) {
-			log.Info("object was not found")
-			return reconcile.Result{}, nil
-		}
+	if err := r.Get(ctx, azureClusterName, azureCluster); err != nil {
+		log.Error(err, "failed to fetch AzureCluster")
 		return reconcile.Result{}, err
 	}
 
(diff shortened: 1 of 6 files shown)


devigned:
Can you add some tests to verify behavior?

controllers/azurejson_machinetemplate_controller.go
```diff
@@ -106,12 +104,8 @@ func (r *AzureJSONTemplateReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result
...
 	}
 
-	err = r.Get(ctx, azureClusterName, azureCluster)
-	if err != nil {
-		if apierrors.IsNotFound(err) {
-			log.Info("object was not found")
-			return reconcile.Result{}, nil
-		}
+	if err := r.Get(ctx, azureClusterName, azureCluster); err != nil {
+		log.Error(err, "failed to fetch azure cluster")
```
CecileRobertMichon, line 108:
```suggestion
		log.Error(err, "failed to fetch AzureCluster")
```

alexeldeib:
@CecileRobertMichon @devigned better?

controllers/azurejson_machinetemplate_controller.go
```diff
@@ -106,12 +104,8 @@ func (r *AzureJSONTemplateReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result
 		Name:      cluster.Spec.InfrastructureRef.Name,
 	}
 
-	err = r.Get(ctx, azureClusterName, azureCluster)
-	if err != nil {
-		if apierrors.IsNotFound(err) {
-			log.Info("object was not found")
-			return reconcile.Result{}, nil
-		}
+	if err := r.Get(ctx, azureClusterName, azureCluster); err != nil {
```
CecileRobertMichon, line 107: the same logic exists in the other azurejson_ controllers, do they also need to be fixed?

alexeldeib: yeah good catch. if it's okay since the logic is the same, i'm only going to write the test once (it'd be identical across all three).

another reason to figure out a way to clean up the preamble a bit...

alexeldeib: added

alexeldeib: meh, might as well add copies for the other reconcilers

CecileRobertMichon: yeah let's combine the tests when we combine the code.  Until then nothing is guaranteed...

PR #872, opened 2020-08-11
chore: make LOCAL_ONLY default and remove REGISTRY override

### LINKED ISSUE #865: Remove the REGISTRY override in ci-e2e
/kind bug

When #852 gets pulled, the current REGISTRY env var will be overridden in [`./scripts/ci-e2e.sh`](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/852/files#diff-20e0541def8acb609f41e6e8e278ae08R52-R55). This override should be removed after the test-infra job has been updated to use the new Azure Container Registry.

devigned:
`LOCAL_ONLY` is currently defaulted to `false`. For folks running the ci-e2e tests locally, this means by default they would need to provide a registry to push images. This is not a kind expectation (pun intended). This PR make the tests execute without defaulting to pushing images remotely.

https://github.com/kubernetes/test-infra/pull/18781 should be merged before this is merged as this will cause the CAPI e2e pivot test to fail due to current Prow configuration.

Fixes #865

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
make LOCAL_ONLY default and remove k8sprow REGISTRY override in `./scripts/ci-e2e.sh`
```


CecileRobertMichon:
/hold

let's make sure we run the capi-e2e-test after kubernetes/test-infra#18781 merges (and check that the self-hosted spec still runs successfully) before merging

docs/development.md
```diff
@@ -346,7 +346,7 @@ You can optionally set the following variables:
...
 | `SKIP_CREATE_MGMT_CLUSTER` | Skip management cluster creation.                                                                                         | `false`                                                                               |
-| `LOCAL_ONLY`               | Only use Kind local registry and run the subset of tests which don't require a remotely pushed controller image.          | `false`                                                                               |
+| `LOCAL_ONLY`               | Only use Kind local registry and run the subset of tests which don't require a remotely pushed controller image.          | `true`                                                                               |
 | `REGISTRY`                 | Registry to push the controller image. If `LOCAL_ONLY=true`, this is optional.                                            | `capzci.azurecr.io/ci-e2e`                                                            |
```
CecileRobertMichon, line 350: " If `LOCAL_ONLY=true`, this is optional"

Looks like it's always optional since we're defaulting it

devigned: good catch

PR #876, opened 2020-08-13
Allow the json secret to be created for control-plane machine

### LINKED ISSUE #874: azure.json not generated for individual AzureMachine
/kind bug

- When trying a create a control-plane machine using the `Machine`, `AzureMachine` and `KubeadmConfig` resources, the secret for azure.json is not generated and the machine gets stuck in a `Provisioning` state 

the secret to be generated and the machine to join the cluster and have a state of `Running`

I see this line in the logs
```
 azurejson_machine_controller.go:127] controllers/AzureJSONMachine "msg"="Cluster Controller has not yet set OwnerRef" "AzureJSONMachine"="cluster-hz0bzy-controlplane-0" "namespace"="kcp-adoption-yd87yd"
```

**Environment:**

- cluster-api-provider-azure version: v0.4.7

nader-ziada:
- include machines that are control plane machines in the azure json secret generation
- get cluster from metadata of AzureMachine

Fixes #874 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Fix issue with azure.json secret not generated for control plane machines
```

### DIFF
--- a/controllers/azurejson_machine_controller.go
+++ b/controllers/azurejson_machine_controller.go
@@ -94,11 +94,10 @@ func (f filterUnclonedMachinesPredicate) Generic(e event.GenericEvent) bool {
 	// outside of machinedeployments/machinesets. if a machine is part of a machineset
 	// or machinedeployment, we already created a secret for the template. All machines
 	// in the machinedeployment will share that one secret.
-	_, isControlPlaneNode := e.Meta.GetLabels()[clusterv1.MachineControlPlaneLabelName]
 	_, isMachineSetNode := e.Meta.GetLabels()[clusterv1.MachineSetLabelName]
 	_, isMachineDeploymentNode := e.Meta.GetLabels()[clusterv1.MachineDeploymentLabelName]
 
-	return !isControlPlaneNode && !isMachineSetNode && !isMachineDeploymentNode
+	return !isMachineSetNode && !isMachineDeploymentNode
 }
 
 // Reconcile reconciles the azure json for a specific machine not in a machine deployment
@@ -119,7 +118,7 @@ func (r *AzureJSONMachineReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result,
 	}
 
 	// Fetch the Cluster.
-	cluster, err := util.GetOwnerCluster(ctx, r.Client, azureMachine.ObjectMeta)
+	cluster, err := util.GetClusterFromMetadata(ctx, r.Client, azureMachine.ObjectMeta)
 	if err != nil {
 		return reconcile.Result{}, err
 	}
--- a/controllers/azurejson_machine_controller_test.go
+++ b/controllers/azurejson_machine_controller_test.go
@@ -1,0 +1,75 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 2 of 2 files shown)


controllers/azurejson_machine_controller.go
```diff
@@ -119,9 +118,9 @@ func (r *AzureJSONMachineReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result,
 	}
 
 	// Fetch the Cluster.
-	cluster, err := util.GetOwnerCluster(ctx, r.Client, azureMachine.ObjectMeta)
+	cluster, err := util.GetClusterFromMetadata(ctx, r.Client, azureMachine.ObjectMeta)
 	if err != nil {
-		return reconcile.Result{}, err
+		return reconcile.Result{}, nil
```
CecileRobertMichon, line 123: why not return the err here? that would cause it to not requeue

alexeldeib:
Ah, I didn't see your PR before sending one -- would you mind adding some tests on the predicate similar to https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/877/files#diff-03ce3b157feddf1a348de00d8560cc8d and I'll close the other PR

nader-ziada:
> Ah, I didn't see your PR before sending one -- would you mind adding some tests on the predicate similar to https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/877/files#diff-03ce3b157feddf1a348de00d8560cc8d and I'll close the other PR

sure, will add the test, sorry about that

controllers/azurejson_machine_controller.go
nader-ziada: yeah, should keep the err, (copy paste bug) 🤦‍♂️

controllers/azurejson_machine_controller.go
```diff
@@ -119,7 +118,7 @@ func (r *AzureJSONMachineReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result,
 	}
 
 	// Fetch the Cluster.
-	cluster, err := util.GetOwnerCluster(ctx, r.Client, azureMachine.ObjectMeta)
+	cluster, err := util.GetClusterFromMetadata(ctx, r.Client, azureMachine.ObjectMeta)
```
alexeldeib, line 121: why switch here? I tested locally and seems to me that using owner is correct?

alexeldeib: tested locally == created a cluster and checked object meta for owners and labels. the azure machine template doesn't get a label corresponding to the cluster but does seem to get the owner ref. Not sure whether that itself is a problem.

nader-ziada: In my test; the AzureMachine doesn’t have the cluster as an owner, but has the label. It only has the Machine as the owner.

alexeldeib: nvm, this is for machine -_-

alexeldeib: yep -- was looking at template, sorry. would be good if we can get a test for this at some point, can do it separate from this PR though. Something like the reconciler tests we have elsewhere and then we can fetch the secret at the end and validate it exists -- kinda like https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/870/files#diff-af6683d08de01aaf9c4b27359a5d3c6c

Issue #814, opened 2020-07-21
nginx test is flaky


CecileRobertMichon:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]

Seeing:

```
Timed out after 300.000s.
Deployment default/ingress-nginx failed to get status.Available = True condition
Expected
    <bool>: false
to be true
```

pretty regularly in PR e2e. Related to #807.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- cluster-api-provider-azure version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

Issue #807, opened 2020-07-20
[e2e] e2e tests should print more debug output


CecileRobertMichon:
We should add debug/logs output to capz e2e tests so it's easier to diagnose why a test failed when it does. For example, the azure LB spec should describe the nginx deployment if it fails to become available.

https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/806#issuecomment-661130287

/help

PR #860, opened 2020-08-04
💚 add debug output for e2e service and job failures

mboersma:
**What this PR does / why we need it**:

Prints a JSON dump of the service or job that failed a `WaitFor` func in e2e tests. If this seems like a reasonable approach, I'll follow with:
- a similar change to `WaitForDeploymentsAvailable` in CAPI
- printing relevant k8s events from the `DescribeFailedX` funcs

**Which issue(s) this PR fixes**:

Refs #807

**Special notes for your reviewer**:

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
💚 add debug output for e2e service and job failures
```

PR #878, opened 2020-08-13
💚 log failed e2e resources and related events

mboersma:

Logs failed deployments (or jobs or services) and related events in the load balancer e2e test, hopefully to shine a light on some failures we've seen in CI.

Refs #807
Refs #814

This copies some code from CAPI e2e helpers temporarily, rather than vendor in CAPI at an arbitrary commit.

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
 🌱 [e2e] log failed deployments and increase HTTP retries in LB test
```

### DIFF
--- a/test/e2e/azure_lb.go
+++ b/test/e2e/azure_lb.go
@@ -105,11 +105,12 @@ func AzureLBSpec(ctx context.Context, inputGetter func() AzureLBSpecInput) {
 	}
 	_, err := deploymentsClient.Create(deployment)
 	Expect(err).NotTo(HaveOccurred())
-	deployInput := framework.WaitForDeploymentsAvailableInput{
+	deployInput := WaitForDeploymentsAvailableInput{
 		Getter:     deploymentsClientAdapter{client: deploymentsClient},
 		Deployment: deployment,
+		Clientset:  clientset,
 	}
-	framework.WaitForDeploymentsAvailable(context.TODO(), deployInput, e2eConfig.GetIntervals(specName, "wait-deployment")...)
+	WaitForDeploymentsAvailable(context.TODO(), deployInput, e2eConfig.GetIntervals(specName, "wait-deployment")...)
 
 	By("creating an internal Load Balancer service")
 	servicesClient := clientset.CoreV1().Services(corev1.NamespaceDefault)
@@ -143,8 +144,9 @@ func AzureLBSpec(ctx context.Context, inputGetter func() AzureLBSpecInput) {
 	_, err = servicesClient.Create(ilbService)
 	Expect(err).NotTo(HaveOccurred())
 	ilbSvcInput := WaitForServiceAvailableInput{
-		Getter:  servicesClientAdapter{client: servicesClient},
-		Service: ilbService,
+		Getter:    servicesClientAdapter{client: servicesClient},
+		Service:   ilbService,
+		Clientset: clientset,
 	}
 	WaitForServiceAvailable(context.TODO(), ilbSvcInput, e2eConfig.GetIntervals(specName, "wait-service")...)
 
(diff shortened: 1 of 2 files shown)


test/e2e/azure_lb.go
```diff
@@ -266,11 +266,13 @@ func AzureLBSpec(ctx context.Context, inputGetter func() AzureLBSpecInput) {
...
 	url := fmt.Sprintf("http://%s", elbIP)
-	resp, err := retryablehttp.Get(url)
+	client := retryablehttp.NewClient()
+	client.RetryMax = 8 // backoff 1,2,4,8... 2 ** 7 = 128 seconds of waiting
+	resp, err := client.Get(url)
 	if resp != nil {
 		defer resp.Body.Close()
 	}
-	Expect(err).NotTo(HaveOccurred())
+	Expect(err).NotTo(HaveOccurred(), func() string { return DescribeFailedDeployment(deployInput, deployment) })
```
mboersma, line 275: Maybe this should call `DescribeFailedService` instead, after #860 merges.

CecileRobertMichon:
/lgtm
/approve
/hold

Is the same change available in CAPI v0.3.9? If so that should be out in the next couple of days

nader-ziada:
> Is the same change available in CAPI v0.3.9? If so that should be out in the next couple of days

CAPI only has the Deployments helper, but this includes also the Service and Job helpers

mboersma:
> CAPI only has the Deployments helper, but this includes also the Service and Job helpers

And this change to the Deployments helper isn't in CAPI (yet). I can PR that change against CAPI first if we'd prefer, because this does diverge a fair amount.

nader-ziada:
> > CAPI only has the Deployments helper, but this includes also the Service and Job helpers
> 
> And this change to the Deployments helper isn't in CAPI (yet). I can PR that change against CAPI first if we'd prefer, because this does diverge a fair amount.

I'm okay with diverging and not waiting now that the CAPI release is out, but lets see what @CecileRobertMichon thinks

CecileRobertMichon:
Now that #918 is open, let's just clean this up to include anything that's not in v0.3.9

mboersma:
/retest

It has failed to provision all 3 control-plane nodes twice now. Boo!

mboersma:
@CecileRobertMichon I think this is ready. It changes some CAPZ test helper functions and clones the CAPI `WaitForDeploymentsAvailable` function locally in order to make some changes to it. I'll PR those back upstream if we think this is good.

I've seen a couple e2e runs lately where this might have pointed out why an LB test failed.

test/e2e/azure_lb.go
```diff
@@ -129,11 +130,6 @@ func AzureLBSpec(ctx context.Context, inputGetter func() AzureLBSpecInput) {
 					Port:     80,
 					Protocol: corev1.ProtocolTCP,
 				},
-				{
-					Name:     "https",
```
CecileRobertMichon, line 133: why this change?

test/e2e/azure_lb.go
```diff
@@ -204,11 +202,6 @@ func AzureLBSpec(ctx context.Context, inputGetter func() AzureLBSpecInput) {
...
```
CecileRobertMichon, line 208: same question

test/e2e/azure_lb.go
mboersma: I think that's a holdover from when I was trying to fix the LB e2e spec failure, and since we don't use the HTTPS port here I thought to simplify things.

CecileRobertMichon: let's either remove it or make it a separate commit (in this PR) to make it clear there was a behavior change beyond logging

mboersma: I removed that change (restored the https:// port).

CecileRobertMichon (approved):
lgtm

let's bring these changes back into the framework as well.

@mboersma can you please fix the conflict then we'll merge it and I'll rebase #646 on top of both commits

CecileRobertMichon (approved):
/lgtm

PR #879, opened 2020-08-13
💎 simplify credential instantiation

alexeldeib:
minor cleanup, should be a no-op

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Remove public constants used to index Azure environment names, relying on the existing logic in github.com/Azure/go-autorest.
```

### DIFF
--- a/cloud/scope/clients.go
+++ b/cloud/scope/clients.go
@@ -17,60 +17,44 @@ limitations under the License.
 package scope
 
 import (
-	"fmt"
 	"strings"
 
 	"github.com/Azure/go-autorest/autorest"
-	"github.com/Azure/go-autorest/autorest/azure"
 	"github.com/Azure/go-autorest/autorest/azure/auth"
-)
-
-const (
-	// ChinaCloud is the cloud environment operated in China
-	ChinaCloud = "AzureChinaCloud"
-	// GermanCloud is the cloud environment operated in Germany
-	GermanCloud = "AzureGermanCloud"
-	// PublicCloud is the default public Azure cloud environment
-	PublicCloud = "AzurePublicCloud"
-	// USGovernmentCloud is the cloud environment for the US Government
-	USGovernmentCloud = "AzureUSGovernmentCloud"
 )
 
 // AzureClients contains all the Azure clients used by the scopes.
 type AzureClients struct {
-	Authorizer                 autorest.Authorizer
-	environment                string
+	Authorizer autorest.Authorizer
+	auth.EnvironmentSettings
 	ResourceManagerEndpoint    string
 	ResourceManagerVMDNSSuffix string
-	subscriptionID             string
-	tenantID                   string
-	clientID                   string
-	clientSecret               string
 }
 
 // CloudEnvironment returns the Azure environment the controller runs in.
 func (c *AzureClients) CloudEnvironment() string {
-	return c.environment
-}
-
-// SubscriptionID returns the Azure subscription id from the controller environment
-func (c *AzureClients) SubscriptionID() string {
-	return c.subscriptionID
+	return c.Environment.Name
 }
 
 // TenantID returns the Azure tenant id the controller runs in.
 func (c *AzureClients) TenantID() string {
-	return c.tenantID
+	return c.Values[auth.TenantID]
 }
 
 // ClientID returns the Azure client id from the controller environment
 func (c *AzureClients) ClientID() string {
-	return c.clientID
+	return c.Values[auth.ClientID]
 }
 
 // ClientSecret returns the Azure client secret from the controller environment
 func (c *AzureClients) ClientSecret() string {
-	return c.clientSecret
+	return c.Values[auth.ClientSecret]
+}
+
+// SubscriptionID returns the Azure subscription id of the cluster,
+// either specified or from the environment
+func (c *AzureClients) SubscriptionID() string {
+	return c.Values[auth.SubscriptionID]
 }
 
 func (c *AzureClients) setCredentials(subscriptionID string) error {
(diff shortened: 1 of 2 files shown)


cloud/scope/clients.go
```diff
@@ -79,46 +63,14 @@ func (c *AzureClients) setCredentials(subscriptionID string) error {
...
-	if c.environment == "" {
-		c.environment = azure.PublicCloud.Name
-	}
-
+	c.EnvironmentSettings = settings
 	c.ResourceManagerEndpoint = settings.Environment.ResourceManagerEndpoint
-	c.ResourceManagerVMDNSSuffix = GetAzureDNSZoneForEnvironment(settings.Environment.Name)
-	settings.Values[auth.SubscriptionID] = subscriptionID
-	settings.Values[auth.TenantID] = c.tenantID
+	c.ResourceManagerVMDNSSuffix = settings.Environment.ResourceManagerVMDNSSuffix
```
CecileRobertMichon, line 68: is there a unit test that can be added to make sure that the values get set properly with other clouds? unfortunately we don't have e2e testing in sovereign clouds yet so a unit test can help make sure this doesn't cause any regressions for other clouds

cloud/scope/clients.go
```diff
@@ -79,46 +63,14 @@ func (c *AzureClients) setCredentials(subscriptionID string) error {
...
+	c.Values[auth.ClientID] = strings.TrimSuffix(c.Values[auth.ClientID], "\n")
+	c.Values[auth.ClientSecret] = strings.TrimSuffix(c.Values[auth.ClientSecret], "\n")
+	c.Values[auth.SubscriptionID] = strings.TrimSuffix(subscriptionID, "\n")
+	c.Values[auth.TenantID] = strings.TrimSuffix(c.Values[auth.TenantID], "\n")
 
-	c.Authorizer, err = settings.GetAuthorizer()
+	c.Authorizer, err = c.GetAuthorizer()
```
CecileRobertMichon, line 74: How come we can't use `NewAuthorizerFromEnvironment()` directly?

devigned: We could, but by not using `NewAuthorizerFromEnvironment` we have more control over how we build the Authorizer.

cloud/scope/clients.go
alexeldeib: @CecileRobertMichon isn't this already that test? https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/cloud/scope/clients_test.go#L36

CecileRobertMichon: indeed 💯

CecileRobertMichon:
For the record:

```
 Incompatible changes:
  - AzureClients: old is comparable, new is not
  - ChinaCloud: removed
  - ClusterScope: old is comparable, new is not
  - ClusterScopeParams: old is comparable, new is not
  - GermanCloud: removed
  - GetAzureDNSZoneForEnvironment: removed
  - ManagedControlPlaneScope: old is comparable, new is not
  - ManagedControlPlaneScopeParams: old is comparable, new is not
  - PublicCloud: removed
  - USGovernmentCloud: removed
  Compatible changes:
  - AzureClients.Environment: added
  - AzureClients.EnvironmentSettings: added
  - AzureClients.Values: added
  - ClusterScope.Environment: added
  - ClusterScope.EnvironmentSettings: added
  - ClusterScope.Values: added
  - ClusterScopeParams.Environment: added
  - ClusterScopeParams.EnvironmentSettings: added
  - ClusterScopeParams.Values: added
  - EnvironmentSettings.GetAuthorizer: added
  - EnvironmentSettings.GetClientCertificate: added
  - EnvironmentSettings.GetClientCredentials: added
  - EnvironmentSettings.GetDeviceFlow: added
  - EnvironmentSettings.GetMSI: added
  - EnvironmentSettings.GetSubscriptionID: added
  - EnvironmentSettings.GetUsernamePassword: added
  - ManagedControlPlaneScope.Environment: added
  - ManagedControlPlaneScope.EnvironmentSettings: added
  - ManagedControlPlaneScope.Values: added
  - ManagedControlPlaneScopeParams.Environment: added
  - ManagedControlPlaneScopeParams.EnvironmentSettings: added
  - ManagedControlPlaneScopeParams.Values: added
```

@alexeldeib Even if this is minor, let's add a release note since this does remove/change public functions and constants

cloud/scope/clients.go
```diff
@@ -17,60 +17,44 @@ limitations under the License.
...
 
 // ClientSecret returns the Azure client secret from the controller environment
 func (c *AzureClients) ClientSecret() string {
-	return c.clientSecret
+	return c.Values[auth.ClientSecret]
+}
+
+// SubscriptionID returns the Azure subscription id of the cluster,
+// either specified or from the environment
+func (c *AzureClients) SubscriptionID() string {
```
CecileRobertMichon, line 56: Can/should we also remove the SubscriptionID() getter in cluster.go https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/879/files#diff-7c0af1847b5ee759aadb398ac16c4d36L92-L93 ?

cloud/scope/cluster.go
```diff
@@ -57,7 +59,18 @@ func NewClusterScope(params ClusterScopeParams) (*ClusterScope, error) {
...
+		subID = params.AzureCluster.Spec.SubscriptionID
+	} else {
+		subID = os.Getenv(auth.SubscriptionID)
+	}
+
+	if subID == "" {
+		return nil, fmt.Errorf("error creating azure services. subscriptionID is not set in cluster or AZURE_SUBSCRIPTION_ID env var")
+	}
+
+	err := params.AzureClients.setCredentials(subID)
```
CecileRobertMichon, line 73: not sure I 100% like this better than before, to me it makes sense to do `params.AzureClients.setCredentials(params.AzureCluster.Spec.SubscriptionID)` and override the env value with that string if it's not empty in the setCredentials function 🤔

cloud/scope/clients.go
alexeldeib: I think we specifically need that one so we can use the SubscriptionAuthorizer interface, but it should just pass through to the AzureClients (we can't let it use the embedded interface because we already have two methods embedded with that name, from AzureClients and ClusterDescriber, so it's ambigous without explicitly overriding).

cloud/scope/cluster.go
alexeldeib: yeah i'll switch this back

cloud/scope/clients.go
alexeldeib: Well, we don't need it here I guess 🙂 since cluster describer doesn't exist yet

CecileRobertMichon:
different error...

/retest

CecileRobertMichon (approved):
/lgtm
/approve

PR #893, opened 2020-08-21
🐛  don't update network interface if it already exists

CecileRobertMichon:
**What this PR does / why we need it**: This fixes a regression introduced in https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/827. We should not reconcile network interfaces that already exist as we risk overwriting changes made by Cloud Provider to update the backend pools.

NOTE: an alternative would be to first get the existing backend pools, append the missing ones, and update but we then risk entering a race with cloud provider if it is updating the pools at the same time (and unfortunately the [update](https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/vendor/k8s.io/legacy-cloud-providers/azure/azure_backoff.go#L258) doesn't have an etag).

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Don't update network interface if it already exists
```

### DIFF
--- a/cloud/services/inboundnatrules/inboundnatrules.go
+++ b/cloud/services/inboundnatrules/inboundnatrules.go
@@ -30,9 +30,6 @@ func (s *Service) Reconcile(ctx context.Context) error {
 	for _, inboundNatSpec := range s.Scope.InboundNatSpecs() {
 		s.Scope.V(2).Info("creating inbound NAT rule", "NAT rule", inboundNatSpec.Name)
 
-		var sshFrontendPort int32 = 22
-		ports := make(map[int32]struct{})
-
 		lb, err := s.LoadBalancersClient.Get(ctx, s.Scope.ResourceGroup(), inboundNatSpec.LoadBalancerName)
 		if err != nil {
 			return errors.Wrapf(err, "failed to get Load Balancer %s", inboundNatSpec.LoadBalancerName)
@@ -41,28 +38,18 @@ func (s *Service) Reconcile(ctx context.Context) error {
 		if lb.LoadBalancerPropertiesFormat == nil || lb.FrontendIPConfigurations == nil || lb.InboundNatRules == nil {
 			return errors.Errorf("Could not get existing inbound NAT rules from load balancer %s properties", to.String(lb.Name))
 		}
-		for _, v := range *lb.InboundNatRules {
-			if to.String(v.Name) == inboundNatSpec.Name {
-				// Inbound NAT Rule already exists, nothing to do here.
-				s.Scope.V(2).Info("NAT rule already exists", "NAT rule", inboundNatSpec.Name)
-				return nil
-			}
-			ports[*v.InboundNatRulePropertiesFormat.FrontendPort] = struct{}{}
+
+		ports := make(map[int32]struct{})
+		if s.natRuleExists(ports)(*lb.InboundNatRules, inboundNatSpec.Name) {
+			// Inbound NAT Rule already exists, nothing to do here.
+			continue
 		}
-		if _, ok := ports[22]; ok {
-			var i int32
-			found := false
-			for i = 2201; i < 2220; i++ {
-				if _, ok := ports[i]; !ok {
-					sshFrontendPort = i
-					found = true
-					break
-				}
-			}
-			if !found {
-				return errors.Errorf("Failed to find available SSH Frontend port for NAT Rule %s in load balancer %s", inboundNatSpec.Name, to.String(lb.Name))
-			}
+
+		sshFrontendPort, err := s.getAvailablePort(ports)
+		if err != nil {
+			return errors.Wrapf(err, "failed to find available SSH Frontend port for NAT Rule %s in load balancer %s", inboundNatSpec.Name, to.String(lb.Name))
 		}
+
 		rule := network.InboundNatRule{
 			Name: to.StringPtr(inboundNatSpec.Name),
 			InboundNatRulePropertiesFormat: &network.InboundNatRulePropertiesFormat{
@@ -100,4 +87,32 @@ func (s *Service) Delete(ctx context.Context) error {
 		s.Scope.V(2).Info("successfully deleted inbound NAT rule", "NAT rule", inboundNatSpec.Name)
 	}
 	return nil
+}
+
+func (s *Service) natRuleExists(ports map[int32]struct{}) func([]network.InboundNatRule, string) bool {
+	return func(rules []network.InboundNatRule, name string) bool {
+		for _, v := range rules {
+			if to.String(v.Name) == name {
+				s.Scope.V(2).Info("NAT rule already exists", "NAT rule", name)
+				return true
+			}
+			ports[*v.InboundNatRulePropertiesFormat.FrontendPort] = struct{}{}
+		}
+		return false
+	}
+}
+
+func (s *Service) getAvailablePort(ports map[int32]struct{}) (int32, error) {
+	var i int32 = 22
+	if _, ok := ports[22]; ok {
+		for i = 2201; i < 2220; i++ {
+			if _, ok := ports[i]; !ok {
+				s.Scope.V(2).Info("Found available port", "port", i)
+				return i, nil
+			}
+		}
+		return i, errors.Errorf("No available SSH Frontend ports")
+	}
+	s.Scope.V(2).Info("Found available port", "port", i)
+	return i, nil
 }
--- a/cloud/services/inboundnatrules/inboundnatrules_test.go
+++ b/cloud/services/inboundnatrules/inboundnatrules_test.go
@@ -172,6 +172,10 @@ func TestReconcileInboundNATRule(t *testing.T) {
 						Name:             "my-machine-nat-rule",
 						LoadBalancerName: "my-public-lb",
 					},
+					{
+						Name:             "my-other-nat-rule",
+						LoadBalancerName: "my-other-public-lb",
+					},
 				})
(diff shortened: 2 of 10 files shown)


ncdc:
@CecileRobertMichon thanks for the quick fix! LGTM so far, modulo needing to update the tests 😄. It's a shame about the lack of an ETag, though 😢.

ncdc:
New changes LGTM too 😄

cloud/services/networkinterfaces/networkinterfaces_test.go
```diff
@@ -301,7 +345,7 @@ func TestReconcileNetworkInterface(t *testing.T) {
 		tc := tc
 		t.Run(tc.name, func(t *testing.T) {
 			g := NewWithT(t)
-			//t.Parallel()
+			t.Parallel()
```
mboersma, line 348: 🚀

cloud/services/inboundnatrules/inboundnatrules.go
```diff
@@ -41,14 +41,19 @@ func (s *Service) Reconcile(ctx context.Context) error {
 		if lb.LoadBalancerPropertiesFormat == nil || lb.FrontendIPConfigurations == nil || lb.InboundNatRules == nil {
 			return errors.Errorf("Could not get existing inbound NAT rules from load balancer %s properties", to.String(lb.Name))
 		}
+		exists := false
```
alexeldeib, line 44: style: when I find myself using this pattern, I often prefer refactoring the loop logic into a function so that I can use return + continue from the inner loop:
```go
// would just be if fn() {} but verbose for illustration
if exists() {
    // now this breaks the outer loop
    continue
}

func exists() bool {
    for _, foo := range bar {
        if condition(foo) {
            return true
        }
    }
    return false
}
```

cloud/services/networkinterfaces/networkinterfaces_test.go
alexeldeib, line 348: tangential: I recently learned a bit about how these actually work -- they mean that each test in this group can be run in parallel, but not in parallel with tests at any other layer. So they actually can prevent other tests from running I think, BUT allow for controlled concurrency.

I found removing every place I used it actually made my tests run with better concurrency, since there were no weird blockers at different layers. I also started preferring to use `t.Run()` more frequently.

alexeldeib: dunno if ginkgo vs go test matters for this and if they behave differently

devigned: > tangential: I recently learned a bit about how these actually work -- they mean that each test in this group can be run in parallel, but not in parallel with tests at any other layer. So they actually can prevent other tests from running I think, but allow for controlled concurrency.

I'm going to need some proof as below doesn't seem to describe that behavior.

```
-parallel n
	    Allow parallel execution of test functions that call t.Parallel.
	    The value of this flag is the maximum number of tests to run
	    simultaneously; by default, it is set to the value of GOMAXPROCS.
	    Note that -parallel only applies within a single test binary.
	    The 'go test' command may run tests for different packages
	    in parallel as well, according to the setting of the -p flag
	    (see 'go help build').
```

alexeldeib: from the docs, emphasis mine: https://golang.org/pkg/testing/#hdr-Subtests_and_Sub_benchmarks
> Subtests can also be used to control parallelism. A parent test will only complete once all of its subtests complete. In this example, **all tests are run in parallel with each other, and only with each other, regardless of other top-level tests that may be defined**
```
func TestGroupedParallel(t *testing.T) {
    for _, tc := range tests {
        tc := tc // capture range variable
        t.Run(tc.Name, func(t *testing.T) {
            t.Parallel()
            ...
        })
    }
}
```

alexeldeib: Note: top level tests in the same package and tests in different packages are not the same, I believe (which is the phrasing difference between the two snippets).

But you can try it very easily yourself with two top level tests in a root file, and watch `go test` execute and where it pauses.

alexeldeib: also -- the flag and func are slightly different, I think?

devigned: Just read through https://blog.golang.org/subtests. The "Control of Parallelism" section was helpful. TIL

cloud/services/inboundnatrules/inboundnatrules.go
CecileRobertMichon: Usually this is better but in this case it's doing a bit more that that, we're also putting the ports in a map to be able to get an unused port later so it would mean either the exists() func has to be overloaded and return a ports map as well or we have to loop through the rules twice. I think I prefer keeping it as-is.

alexeldeib: https://diffy.org/diff/3ezc7mo6trys9n74ve2rv5z5mi would be one way to do it

Not a huge deal either way

CecileRobertMichon: how about https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/893/commits/86b9cf769e5261883b7517995d305c66bcb71627

I will squash if you like it

cloud/services/networkinterfaces/networkinterfaces_test.go
CecileRobertMichon: let's follow up on this conversation outside of this PR

cloud/services/inboundnatrules/inboundnatrules.go
```diff
@@ -101,3 +88,31 @@ func (s *Service) Delete(ctx context.Context) error {
 	}
 	return nil
 }
+
+func (s *Service) natRuleExists(ports map[int32]struct{}) func([]network.InboundNatRule, string) bool {
```
alexeldeib, line 92: as much as I appreciate my own work, no longer used :)

alexeldeib: oh jk

cloud/services/inboundnatrules/inboundnatrules.go
alexeldeib: yep, nice

alexeldeib:
whoops, thought it already squashed

CecileRobertMichon:
> whoops, thought it already squashed

It's squashed, kept the two commits fixing two different things on purpose

/approve

cloud/services/inboundnatrules/inboundnatrules.go
```diff
@@ -101,3 +88,31 @@ func (s *Service) Delete(ctx context.Context) error {
...
+	return func(rules []network.InboundNatRule, name string) bool {
+		for _, v := range rules {
+			if to.String(v.Name) == name {
+				s.Scope.V(2).Info("NAT rule already exists", "NAT rule", name)
+				return true
+			}
+			ports[*v.InboundNatRulePropertiesFormat.FrontendPort] = struct{}{}
```
alexeldeib, line 99: this is not correct, ordering matters now that we moved the return

alexeldeib: should be before the if condition

cloud/services/inboundnatrules/inboundnatrules_test.go
```diff
@@ -325,3 +341,201 @@ func TestDeleteNetworkInterface(t *testing.T) {
...
+		expectedPorts  map[int32]struct{}
+		expect         func(s *mock_inboundnatrules.MockInboundNatScopeMockRecorder,
+			m *mock_inboundnatrules.MockClientMockRecorder, mLoadBalancer *mock_loadbalancers.MockClientMockRecorder)
+	}{
+		{
+			name:     "Rule exists",
+			ruleName: "my-rule",
+			existingRules: []network.InboundNatRule{
+				{
+					InboundNatRulePropertiesFormat: &network.InboundNatRulePropertiesFormat{
```
alexeldeib, line 360: copying this test case and reversing the order should catch the other comment I just dropped

cloud/services/inboundnatrules/inboundnatrules.go
CecileRobertMichon: It doesn't because if the rule exists we don't care about existing ports since we return early from the function so that set would be incomplete anyways

cloud/services/inboundnatrules/inboundnatrules_test.go
CecileRobertMichon: it won't, we only care about the content of port if the rule doesn't already exists https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/893/files/b185781e9547bd6c9ba3b0a87ecf13d23ca6b637#diff-ecbb14ef6845369e03315e8fe557892eR438

cloud/services/inboundnatrules/inboundnatrules.go
CecileRobertMichon: I moved it down here to make it explicit, I can add a code comment too.

// if this isn't the rule we're looking for, add the port to the set of used port

alexeldeib: ah, my mistake, I see what you mean

cloud/services/inboundnatrules/inboundnatrules_test.go
alexeldeib: yep sorry

PR #896, opened 2020-08-25
💎 Update tags at scope instead of updating VMs

CecileRobertMichon:
**What this PR does / why we need it**: This fixes a TODO in the AzureMachine controller to use the new tags' SDK CreateOrUpdateAtScope (released in v0.45.1). This allows:
1. Making the update more targeted, updating tags only instead of the entire VM resource
2. Unit testing the tags reconciler
3. in the future the same createOrUpdateAtScope tags reconciler could be used for other Azure resources

Fixed a few typos on the way.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Use tags CreateOrUpdateAtScope to ensure VM tags are correct after the VM is created
```


nader-ziada:
I tried running this code and all looks good 

also 👍 for all the cleanup

api/v1alpha3/tags.go
```diff
@@ -118,6 +118,12 @@ const (
 
 	// CommonRole describes the value for the common role
 	CommonRole = "common"
+
+	// TagsLastAppliedAnnotation is the key for the machine object annotation
```
CecileRobertMichon, line 122: just noticed need to fix the comment to match the name here

nader-ziada: missed that one

PR #898, opened 2020-08-25
✨ Add support to specify subnet's Nat Gateway

### LINKED ISSUE #894: Nat gateway support in SubnetSpec
***Goals***
1. Support specifying the Nat Gateway in the `SubnetSpec` type.

***Non-Goals/Future Work***

**User Story**

As a [developer/user/operator] I would like to specify which Nat Gateway to use when describing subnets for being able to configure egress traffic in worker nodes.

**Detailed Description**

[The `SubnetSpec` type](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/92164bbf8c1ae7073c7d3f63ade6cb9c9335f971/api/v1alpha3/types.go#L357-L384) supports specifying some subnet properties like the route table, or the security group. When these subnet properties are set, [the controller](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/df22023917a8b856e2960f5f199b9281f96decb4/cloud/services/subnets/subnets.go#L76-L89) sends their value to the Azure API to create the subnet. I'd like to add a new field to the `SubnetSpec` type in CAPZ, so that users could specify the Nat Gateway to use when reconciling the subnets.

A new field in the `SubnetSpec` field would be needed. It would be backward compatible.

Controller would check the new field and use it when not empty. It would be backward compatible.

/kind proposal

fiunchinho:

Adds new field to `SubnetSpec` so that users can specify the Nat Gateway to use when creating subnets.

Fixes #894 

**TODOs**:

- [X] squashed commits
- [X] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Added new field to `SubnetSpec` so that users can specify the Nat Gateway to use when creating subnets.
```


fiunchinho:
It has been harder to add the :sparkles: emoji to the PR title than to make the actual change :smile:

api/v1alpha3/types.go
```diff
@@ -381,6 +388,10 @@ type SubnetSpec struct {
 	// RouteTable defines the route table that should be attached to this subnet.
 	// +optional
 	RouteTable RouteTable `json:"routeTable,omitempty"`
+
+	// NatGateway associated with this subnet.
+	// +optional
+	NatGateway NatGateway `json:"natGateway,omitempty"`
```
CecileRobertMichon, line 394: where is this used? you also need to update the controller to consume this. See how route table is created/set on the subnet.

fiunchinho: This is not used. I figured it would be better to first introduce the field and then make the change in the relevant controllers. Should I submit all the changes in this PR instead?

CecileRobertMichon: I personally would prefer we don't merge unused API fields. Feel free to separate the controller changes in a separate commit in the same PR.

alexeldeib:
protip for emoji: if you type it out `:gem:` in the git commit, github will show the message with the emoji and when you have the PR open in preview state. I always copy paste it from there into the title :)

fiunchinho:
I just pushed an implementation copying more or less the route tables implementation.

cloud/services/natgateways/natgateways.go
```diff
@@ -0,0 +1,93 @@
...
+
+	for _, natGatewaySpec := range s.Scope.NatGatewaySpecs() {
+		existingNatGateway, err := s.Get(ctx, s.Scope.ResourceGroup(), natGatewaySpec.Name)
+		if !azure.ResourceNotFound(err) {
+			if err != nil {
+				return errors.Wrapf(err, "failed to get nat gateway %s in %s", natGatewaySpec.Name, s.Scope.ResourceGroup())
+			}
+
+			// nat gateway already exists
+			// currently don't support:
```
alexeldeib, line 44: If you intend for this iteration so support only one nat gateway, please change the logic here and around `NatGatewaySpecs()` to return a single element, not a slice. Right now this overwrites the name/id below will be overwritten on each iteration, but because we hardcode one nat gateway spec, it happens to work.

cloud/scope/cluster.go
```diff
@@ -234,6 +243,11 @@ func (s *ClusterScope) RouteTable() *infrav1.RouteTable {
 	return &s.AzureCluster.Spec.NetworkSpec.GetNodeSubnet().RouteTable
 }
 
+// NatGateway returns the cluster node nat gateway.
+func (s *ClusterScope) NatGateway() *infrav1.NatGateway {
```
alexeldeib, line 247: If you want to return multiple nat gateways from `NatGatewaySpecs()` you can follow the example with `NodeSubnet/ControlPlaneSubnet`, or something like that.

cloud/scope/cluster.go
```diff
@@ -156,6 +157,13 @@ func (s *ClusterScope) RouteTableSpecs() []azure.RouteTableSpec {
 	}}
 }
 
+// NatGatewaySpecs returns the node nat gateway
+func (s *ClusterScope) NatGatewaySpecs() []azure.NatGatewaySpec {
+	return []azure.NatGatewaySpec{{
+		Name: s.NatGateway().Name,
```
CecileRobertMichon, line 163: what if NatGateway is nil? we should only create it if explicitly specified by the user, ie. default behavior does not change

api/v1alpha3/azurecluster_default.go
```diff
@@ -95,6 +95,9 @@ func (c *AzureCluster) setSubnetDefaults() {
 	if nodeSubnet.RouteTable.Name == "" {
 		nodeSubnet.RouteTable.Name = generateRouteTableName(c.ObjectMeta.Name)
 	}
+	if nodeSubnet.NatGateway.Name == "" {
+		nodeSubnet.NatGateway.Name = generateNatGatewayName(c.ObjectMeta.Name)
```
CecileRobertMichon, line 99: This mean we would be creating it by default. I'm not sure that's a behavior change we want, @devigned @alexeldeib thoughts?

cloud/services/natgateways/natgateways.go
```diff
@@ -0,0 +1,93 @@
...
+			//  1. creating separate control plane and node (#718) so update both
```
CecileRobertMichon, line 45: this is changing in #816 FWIW

api/v1alpha3/azurecluster_default.go
CecileRobertMichon: this is different behavior than described in the issue, ie. `Controller would check the new field and use it when not empty. It would be backward compatible.`

cloud/scope/cluster.go
fiunchinho: I want to return multiple gateways. I believe I need to add a `Role` field to the `NatGatewaySpec`, otherwise when we are reconciling NAT Gateways that have been already created and want to save the NAT Gateway ID back into the `SubnetSpec` in the CR, I wouldn't know which `NatGatewayspec` is which while iterating. Am I making sense?

api/v1alpha3/azurecluster_default.go
alexeldeib: i'd prefer the latter behavior, at least in the first iteration

cloud/scope/cluster.go
alexeldeib: since you have access to the subnets here, I might do something like:
```
func (s* clusterScope) NodeNatGateway() *infrav1.NatGateway
func (s* clusterScope)  ControlPlaneNatGateway() *infrav1.NatGateway
func (s *ClusterScope) NatGatewaySpecs() []azure.NatGatewaySpec {
	var gateways []azure.NatGatewaySpec{}
        if gateway := s.NodeNatGateway(); gateway != nil {
            gateways = append(gateways, gateway)
        }
        if gateway := s.ControlPlaneNatGateway(); gateway != nil {
            gateways = append(gateways, gateway)
        }
}
```

alexeldeib: and yeah the individual funcs could attach the role to the nat gateway, cc @CecileRobertMichon on that approach

fiunchinho:
Superseeded by https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1188

PR #320, opened 2019-11-06
🏃 update image reference to use first party image

CecileRobertMichon:
**What this PR does / why we need it**: We now have a 1st party image 🎉  no more need for users to accept legal terms before using the sku.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**Release note**:

```release-note
Update default image reference to first party image
```

PR #928, opened 2020-09-10
🌱 add thirdpartyimage flag to marketplace images to be used when creating a VM

nader-ziada:
- add an optional thirdpartyimage bool field to the VM at creation when using marketplace images 

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Add ThirdPartyImage flag to Marketplace image to indicate an image published by a third party publisher.
```

### DIFF
--- a/api/v1alpha3/types.go
+++ b/api/v1alpha3/types.go
@@ -251,6 +251,10 @@ type AzureMarketplaceImage struct {
 	// time even if a new version becomes available.
 	// +kubebuilder:validation:MinLength=1
 	Version string `json:"version"`
+	// ThirdPartyImage indicates the image is published by a third party publisher and a Plan
+	// will be generated for it.
+	// +kubebuilder:default=false
+	ThirdPartyImage bool `json:"thirdPartyImage"`
 }
 
 // AzureSharedGalleryImage defines an image in a Shared Image Gallery to use for VM creation
--- a/docs/book/src/topics/custom-images.md
+++ b/docs/book/src/topics/custom-images.md
@@ -104,7 +104,7 @@ Managed images support only 20 simultaneous deployments, so for most use cases S
 
 ### Using Azure Marketplace
 
-To use an image from [Azure Marketplace][azure-marketplace], populate the `publisher`, `offer`, `sku`, and `version` fields:
+To use an image from [Azure Marketplace][azure-marketplace], populate the `publisher`, `offer`, `sku`, and `version` fields and, if this image is published by a third party publisher, set the `thirdPartyImage` flag to `true` so an image Plan can be generated for it. In the case of a third party image, you must accept the license terms with the [Azure CLI](https://docs.microsoft.com/en-us/cli/azure/vm/image/terms?view=azure-cli-latest) before consuming it.
 
 ```yaml
 apiVersion: infrastructure.cluster.x-k8s.io/v1alpha3
@@ -119,6 +119,7 @@ spec:
         offer: "example-offer"
         sku: "k8s-1dot18dot8-ubuntu-1804"
         version: "2020-07-25"
+        thirdPartyImage: true
 ```
 
 [azure-marketplace]: https://docs.microsoft.com/azure/marketplace/marketplace-publishers-guide
--- a/api/v1alpha2/azuremachine_conversion.go
+++ b/api/v1alpha2/azuremachine_conversion.go
@@ -65,6 +65,10 @@ func restoreAzureMachineSpec(restored, dst *infrav1alpha3.AzureMachineSpec) {
 	}
 	dst.OSDisk.DiffDiskSettings = restored.OSDisk.DiffDiskSettings
 	dst.OSDisk.CachingType = restored.OSDisk.CachingType
+
+	if restored.Image != nil && restored.Image.Marketplace != nil {
+		dst.Image.Marketplace.ThirdPartyImage = restored.Image.Marketplace.ThirdPartyImage
+	}
 }
 
 // ConvertFrom converts from the Hub version (v1alpha3) to this version.
--- a/cloud/scope/machine.go
+++ b/cloud/scope/machine.go
@@ -420,7 +420,7 @@ func (m *MachineScope) GetBootstrapData(ctx context.Context) (string, error) {
 	return base64.StdEncoding.EncodeToString(value), nil
 }
 
-// Pick image from the machine configuration, or use a default one.
+// GetVMImage returns the image from the machine configuration, or a default one.
 func (m *MachineScope) GetVMImage() (*infrav1.Image, error) {
 	// Use custom Marketplace image, Image ID or a Shared Image Gallery image if provided
 	if m.AzureMachine.Spec.Image != nil {
--- a/cloud/services/virtualmachines/virtualmachines.go
+++ b/cloud/services/virtualmachines/virtualmachines.go
@@ -102,6 +102,7 @@ func (s *Service) Reconcile(ctx context.Context) error {
 			}
 
 			virtualMachine := compute.VirtualMachine{
+				Plan:     s.generateImagePlan(),
 				Location: to.StringPtr(s.Scope.Location()),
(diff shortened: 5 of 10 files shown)


api/v1alpha3/types.go
```diff
@@ -251,6 +251,29 @@ type AzureMarketplaceImage struct {
...
+	// +optional
+	Plan *ImagePlan `json:"plan,omitempty"`
+}
+
+// ImagePlan specifies information about the marketplace image used to create the virtual machine. This element is
+// only used for marketplace images. Before you can use a marketplace image from an API, you must enable the
+// image for programmatic use.  In the Azure portal, find the marketplace image that you want to use and then
+// click **Want to deploy programmatically, Get Started ->**. Enter any required information and then click
+// **Save**.
+type ImagePlan struct {
```
CecileRobertMichon, line 264: I wonder if this is a good user experience? Do we expect users to know how to fill this if they aren't the ones publishing the image? Or would this feature not be needed to be touched by end users in most cases?

CecileRobertMichon: Another way to look at this is: would a user ever need to customize these values? If not, an alternative would be to add an optional `isThirdParty` bool in the `AzureMarketplaceImage` type which defaults to false and auto-generate the plan info if it is set to true since the values should be deterministic based on the image info.

dvonthenen: > Another way to look at this is: would a user ever need to customize these values? If not, an alternative would be to add an optional `isThirdParty` bool in the `AzureMarketplaceImage` type which defaults to false and auto-generate the plan info if it is set to true since the values should be deterministic based on the image info.

That seems to be inline with my comment in our conversation. It also has the least risk for both modes (upstream marketplace image versus 3rd party marketplace image).

nader-ziada: how would we generate the Plan fields?

CecileRobertMichon: you can reuse something like this https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/320/files#diff-b2813f64759b38fd374267409eaaca65L242 

Product = offer
Name = sku
Publisher = name

I don't think we need promotion code?

nader-ziada: ok sounds good, will change that, didn't know we had the info to fill in the Plan

CecileRobertMichon:
Let's also make sure to update https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/docs/book/src/topics/custom-images.md with the changes and add conversion where needed (you won't be able to convert the new value to v1alpha2 but we can at least restore the value for round trip conversion).

nader-ziada:
@CecileRobertMichon made the changes, still has to do conversion

nader-ziada:
@dvonthenen @CecileRobertMichon I think its ready for another look now, thanks

api/v1alpha3/types.go
```diff
@@ -251,6 +251,10 @@ type AzureMarketplaceImage struct {
 	// time even if a new version becomes available.
 	// +kubebuilder:validation:MinLength=1
 	Version string `json:"version"`
+	// ThirdPartyImage indicates he image is published by a third party publisher and a Plan
```
CecileRobertMichon, line 254:
```suggestion
	// ThirdPartyImage indicates the image is published by a third party publisher and a Plan
```

docs/book/src/topics/custom-images.md
```diff
@@ -104,7 +104,8 @@ Managed images support only 20 simultaneous deployments, so for most use cases S
 
 ### Using Azure Marketplace
 
-To use an image from [Azure Marketplace][azure-marketplace], populate the `publisher`, `offer`, `sku`, and `version` fields:
+To use an image from [Azure Marketplace][azure-marketplace], populate the `publisher`, `offer`, `sku`, and `version` fields, and if this image is published by a third 
+party publisher, set the `thirdPartyImage` flag to `true` so an image plan can be generated for it.
```
CecileRobertMichon, line 108: let's also document the need to accept offer terms on the subscription before being able to use it if the image is third party

nader-ziada: updated

docs/book/src/topics/custom-images.md
```diff
@@ -104,7 +104,8 @@ Managed images support only 20 simultaneous deployments, so for most use cases S
 
 ### Using Azure Marketplace
 
-To use an image from [Azure Marketplace][azure-marketplace], populate the `publisher`, `offer`, `sku`, and `version` fields:
+To use an image from [Azure Marketplace][azure-marketplace], populate the `publisher`, `offer`, `sku`, and `version` fields, and if this image is published by a third
```
CecileRobertMichon, line 107:
```suggestion
To use an image from [Azure Marketplace][azure-marketplace], populate the `publisher`, `offer`, `sku`, and `version` fields and, if this image is published by a third 
```

(?)

docs/book/src/topics/custom-images.md
```diff
@@ -104,7 +104,8 @@ Managed images support only 20 simultaneous deployments, so for most use cases S
 
 ### Using Azure Marketplace
 
-To use an image from [Azure Marketplace][azure-marketplace], populate the `publisher`, `offer`, `sku`, and `version` fields:
+To use an image from [Azure Marketplace][azure-marketplace], populate the `publisher`, `offer`, `sku`, and `version` fields, and if this image is published by a third 
+party publisher, set the `thirdPartyImage` flag to `true` so an image plan can be generated for it. In the case of a third party image, the user has to accept the license terms before they can consume that image which can do with the [Azure cli](https://docs.microsoft.com/en-us/cli/azure/vm/image/terms?view=azure-cli-latest)
```
CecileRobertMichon, line 108:
```suggestion
party publisher, set the `thirdPartyImage` flag to `true` so an image Plan can be generated for it. In the case of a third party image, you must accept the license terms with the [Azure CLI](https://docs.microsoft.com/en-us/cli/azure/vm/image/terms?view=azure-cli-latest) before consuming it.
```

I think there's also an extra new line here

api/v1alpha3/types.go
```diff
@@ -251,6 +251,10 @@ type AzureMarketplaceImage struct {
 	// time even if a new version becomes available.
 	// +kubebuilder:validation:MinLength=1
 	Version string `json:"version"`
+	// ThirdPartyImage indicates the image is published by a third party publisher and a Plan
+	// will be generated for it
```
CecileRobertMichon, line 255:
```suggestion
	// will be generated for it.
```

CecileRobertMichon (approved):
A few minor grammar nits, otherwise lgtm

were you able to test this with a third party image to make sure it works?

CecileRobertMichon:
Can you also update the release note to reflect the user-facing change (ie. Add ThirdPartyImage option to Azure Marketplace Images to support third-party images)?

nader-ziada:
@CecileRobertMichon made the changes, in the process of testing it now

nader-ziada:
> Can you also update the release note to reflect the user-facing change (ie. Add ThirdPartyImage option to Azure Marketplace Images to support third-party images)?

do you mean action required?

CecileRobertMichon:
> do you mean action required?

no no, just like this is good - there's no breaking change / action required

Just meant update it to reflect the latest changes to the API

nader-ziada:
@CecileRobertMichon  was able to test with a third party image, not sure how to show here?

vincepri:
@nader-ziada Maybe post the configuration you used and a screenshot of the Machine / VM using the correct image?

vincepri:
Not sure if that's enough, but it might be helpful for others that want to use custom published images as well in the future :)

nader-ziada:
Here is an example of the custom image that worked

custom-image

nader-ziada:
@devigned @CecileRobertMichon anything pending in this PR? I think its ready

devigned (approved):
/lgtm
/approve

devigned:
/hold cancel

Validation achieved! Thank you for the screen shot.

Issue #826, opened 2020-07-23
Integrate AzureMachine with AzureManagedControlPlane (BYO nodes on AKS)


alexeldeib:
/kind feature

**Describe the solution you'd like**
Extend CAPZ as necessary to allow CAPZ-managed AzureMachines to join AKS clusters provisioned by AzureManagedControlPlane. A PoC exists in #822, which itself relies on #824 and #825. On top of those changes, we need to add logic in AzureMachineController to allow instantiating the appropriate cluster describer at runtime using the duck typing from #825.

**Anything else you would like to add:**

Currently there are two issues with the PoC:
- nodes register with master labels. I think this is due to using the admin kubeconfig to join nodes. we need to drop privileges and try to join in that way. We can also try doing a token join via kubeconfig with a low privilege serviceaccount.
- need to validate azure-cni functionality and/or debug any issues there

### Warning

This is 100% unsupported by AKS right now, I just think it's cool and I don't see any technical reasons it can't work. This feature would be indefinitely experimental unless AKS supported BYO node

alexeldeib:
Issues I encountered working on https://github.com/kubernetes-sigs/cluster-api-provider-azure/compare/master...alexeldeib:ace/integrate?expand=1

### Critical issues // functional problems
- Currently nodes use an admin kubeconfig to join the cluster, then patch their own role and labels to drop privileges and become agents. This is awful security practice and I feel is the most critical item to fix before anyone could use this.
- Azure CNI requires pre-allocating IP for pods to the VMs, which is not currently handled by CAPZ. This would be a critical requirement for anyone using Azure CNI, unless we can take advantage of another IPAM/IP allocation scheme.
- Joining nodes with CAPBK uses 2 preKubeadmCommands and 2 postKubeadmCommands, which is less than ideal. We should consider encapsulating this more cleanly, either in a bootstrap provider or VHD/script somehow.

### Minor issues // implementation details
- group service should not always attempt to reconcile the resource group, and it should differentiate node/control plane resource groups
- AKS vnet name is not predictable and nodes require it for provisioning. The managed cluster API does not return the vnet name, but we can extract it from the node pools API which contains vnetSubnetID as a fully qualified ARM resource
- anywhere that currently uses a cluster scope now needs to accept **either** an AzureCluster or AzureManagedControlPlane as a ClusterDescriber, increasing code bloat.
- Patching node labels/roles seems to work well on 

### Other observations // no action necessarily required
[...]

PR #929, opened 2020-09-13
[wip] 💎 add vnet to managed control plane status

alexeldeib:

prepares for #826, we need to know the AKS managed vnet name (or pre-create it) so we can join nodes to it

Fixes #

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cloud/defaults.go
+++ b/cloud/defaults.go
@@ -46,6 +46,13 @@ const (
 	LatestVersion = "latest"
 )
 
+const (
+	// AKSSubnetName is the name of the default aks managed subnet for worker nodes.
+	AKSSubnetName = "aks-subnet"
+	// UniqueNameSuffixSize is the length of the hash AKS uses as a cluster id.
+	UniqueNameSuffixSize = 8
+)
+
 // GenerateInternalLBName generates a internal load balancer name, based on the cluster name.
 func GenerateInternalLBName(clusterName string) string {
 	return fmt.Sprintf("%s-%s", clusterName, "internal-lb")
--- a/cloud/scope/managedcontrolplane.go
+++ b/cloud/scope/managedcontrolplane.go
@@ -28,6 +28,7 @@ import (
 	expv1 "sigs.k8s.io/cluster-api/exp/api/v1alpha3"
 
 	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1alpha3"
+	azure "sigs.k8s.io/cluster-api-provider-azure/cloud"
 	infrav1exp "sigs.k8s.io/cluster-api-provider-azure/exp/api/v1alpha3"
 
 	"sigs.k8s.io/cluster-api/util/patch"
@@ -142,4 +143,45 @@ func (s *ManagedControlPlaneScope) Authorizer() autorest.Authorizer {
 // PatchObject persists the cluster configuration and status.
 func (s *ManagedControlPlaneScope) PatchObject(ctx context.Context) error {
 	return s.patchHelper.Patch(ctx, s.PatchTarget)
+}
+
+// ControlPlaneSubnet is empty for managed clusters.
+func (s *ManagedControlPlaneScope) ControlPlaneSubnet() *infrav1.SubnetSpec {
+	return nil
+}
+
+// NodeSubnet is the name of the subnet nodes should join, defaults to aks-subnet.
+func (s *ManagedControlPlaneScope) NodeSubnet() *infrav1.SubnetSpec {
+	return &infrav1.SubnetSpec{
+		Role: infrav1.SubnetNode,
+		Name: azure.AKSSubnetName,
+	}
+}
+
+// IsVnetManaged returns true if the vnet is managed.
+func (s *ManagedControlPlaneScope) IsVnetManaged() bool {
+	return true
+}
+
+// RouteTable returns the cluster node routetable.
+func (s *ManagedControlPlaneScope) RouteTable() *infrav1.RouteTable {
+	return nil
+}
+
+// Vnet returns the cluster Vnet.
+func (s *ManagedControlPlaneScope) Vnet() *infrav1.VnetSpec {
+	return &infrav1.VnetSpec{
+		ResourceGroup: s.ControlPlane.Status.NodeResourceGroupName,
+		Name:          s.ControlPlane.Status.VirtualNetworkName,
+	}
+}
+
+// VNetSpecs returns the virtual network specs.
+func (s *ManagedControlPlaneScope) VNetSpecs() []azure.VNetSpec {
+	return []azure.VNetSpec{
+		{
+			ResourceGroup: s.Vnet().ResourceGroup,
+			Name:          s.Vnet().Name,
+		},
+	}
 }
--- a/exp/api/v1alpha3/azuremanagedcontrolplane_types.go
+++ b/exp/api/v1alpha3/azuremanagedcontrolplane_types.go
@@ -86,6 +86,13 @@ type AzureManagedControlPlaneStatus struct {
 	// In the AzureManagedControlPlane implementation, these are identical.
 	// +optional
 	Initialized bool `json:"initialized,omitempty"`
+
+	// VirtualNetworkName is the name of the AKS-generated VNET. We either
+	// use the AKS pre-created one or find one from an agent pool.
+	VirtualNetworkName string `json:"virtualNetworkName,omitempty"`
+
+	// NodeResourceGroupName is the name of the resource group containining cluster IaaS resources.
+	NodeResourceGroupName string `json:"nodeResourceGroupName,omitempty"`
 }
 
 // +kubebuilder:object:root=true
--- a/exp/controllers/azuremangedcontrolplane_reconciler.go
+++ b/exp/controllers/azuremangedcontrolplane_reconciler.go
@@ -92,8 +92,8 @@ func (r *azureManagedControlPlaneReconciler) Reconcile(ctx context.Context, scop
 		return errors.Wrapf(err, "failed to reconcile managed cluster")
 	}
 
-	scope.V(2).Info("Reconciling endpoint")
-	if err := r.reconcileEndpoint(ctx, scope, managedClusterSpec); err != nil {
+	scope.V(2).Info("Reconciling status")
+	if err := r.reconcileStatus(ctx, scope, managedClusterSpec); err != nil {
 		return errors.Wrapf(err, "failed to reconcile control plane endpoint")
 	}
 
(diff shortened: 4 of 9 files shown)


cloud/scope/managedcontrolplane.go
```diff
@@ -143,3 +143,45 @@ func (s *ManagedControlPlaneScope) Authorizer() autorest.Authorizer {
...
+	}
+}
+
+// VNetSpecs returns the virtual network specs.
+func (s *ManagedControlPlaneScope) VNetSpecs() []azure.VNetSpec {
+	return []azure.VNetSpec{
+		{
+			ResourceGroup: s.Vnet().ResourceGroup,
+			Name:          s.Vnet().Name,
+			CIDR:          s.Vnet().CidrBlock,
```
alexeldeib, line 184: might want to remove this, but it shouldn't be used anyway

exp/api/v1alpha3/azuremanagedcontrolplane_types.go
```diff
@@ -86,6 +88,14 @@ type AzureManagedControlPlaneStatus struct {
 	// In the AzureManagedControlPlane implementation, these are identical.
 	// +optional
 	Initialized bool `json:"initialized,omitempty"`
+
+	// VirtualNetwork is the name of the AKS-generated VNET. It's not predictable unless we pre-create it.
+	// Consider pre-creating?
```
devigned, line 93: "Consider pre-creating?" <-- Would you consider removing this and tracking it with an issue?

devigned (changes requested):
Looks pretty solid. 
- Left a request to change a comment.
- Please add tests

cloud/scope/managedcontrolplane.go
CecileRobertMichon: if it's not used it might make more sense to leave it as empty to avoid confusion

exp/controllers/azuremangedcontrolplane_reconciler.go
```diff
@@ -140,6 +148,31 @@ func (r *azureManagedControlPlaneReconciler) Delete(ctx context.Context, scope *
...
+		return fmt.Errorf("failed to find vnet in managed resource group '%s' for aks cluster", scope.ControlPlane.ManagedResourceGroup())
+	}
+
+	if vnets[0].Name == nil {
+		return errors.New("expected vnet returned by Azure to have a name, but was nil")
+	}
+
+	old := scope.ControlPlane.DeepCopyObject()
+
+	scope.ControlPlane.Status.VirtualNetwork = *vnets[0].Name
```
CecileRobertMichon, line 167: what happens if there is more than 1? the assumption might work right now but it could become fragile if changes are made in the future to avoid vnet peerings and things like that, or simply if the RG is used by multiple clusters.

exp/controllers/azuremangedcontrolplane_reconciler.go
```diff
@@ -36,13 +36,15 @@ import (
...
 	"sigs.k8s.io/cluster-api-provider-azure/cloud/services/managedclusters"
+	"sigs.k8s.io/cluster-api-provider-azure/cloud/services/virtualnetworks"
 	infrav1exp "sigs.k8s.io/cluster-api-provider-azure/exp/api/v1alpha3"
 )
 
 // azureManagedControlPlaneReconciler are list of services required by cluster controller
 type azureManagedControlPlaneReconciler struct {
 	kubeclient         client.Client
 	managedClustersSvc *managedclusters.Service
+	vnetSvc            *virtualnetworks.Service
```
CecileRobertMichon, line 47: this should be of type azure.Service

we're only using `*managedclusters.Service` as a workaround until https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/757 is done

exp/controllers/azuremangedcontrolplane_reconciler.go
alexeldeib: we can probably fetch it out of the MC object, you are right

cloud/scope/managedcontrolplane.go
```diff
@@ -143,3 +144,45 @@ func (s *ManagedControlPlaneScope) Authorizer() autorest.Authorizer {
...
+// ControlPlaneSubnet is empty for managed clusters.
+func (s *ManagedControlPlaneScope) ControlPlaneSubnet() *infrav1.SubnetSpec {
+	return nil
+}
+
+// NodeSubnet is the name of the subnet nodes should join, defaults to aks-subnet.
+func (s *ManagedControlPlaneScope) NodeSubnet() *infrav1.SubnetSpec {
+	return &infrav1.SubnetSpec{
+		Role: infrav1.SubnetNode,
+		Name: "aks-subnet",
```
CecileRobertMichon, line 157: let's move that string to a const in defaults.go

exp/api/v1alpha3/azuremanagedcontrolplane_types.go
```diff
@@ -86,6 +88,14 @@ type AzureManagedControlPlaneStatus struct {
...
+	VirtualNetwork string `json:"virtualNetwork,omitempty"`
+}
+
+func (c *AzureManagedControlPlane) ManagedResourceGroup() string {
+	return fmt.Sprintf("MC_%s_%s_%s", c.Spec.ResourceGroup, c.Name, c.Spec.Location)
```
CecileRobertMichon, line 98: defaults.go seems like a better place to put this (along with all the other Generate* functions)

exp/api/v1alpha3/azuremanagedcontrolplane_types.go
```diff
@@ -86,6 +88,14 @@ type AzureManagedControlPlaneStatus struct {
 	// In the AzureManagedControlPlane implementation, these are identical.
 	// +optional
 	Initialized bool `json:"initialized,omitempty"`
+
+	// VirtualNetwork is the name of the AKS-generated VNET. It's not predictable unless we pre-create it.
+	// Consider pre-creating?
+	VirtualNetwork string `json:"virtualNetwork,omitempty"`
```
CecileRobertMichon, line 94: Slight preference for `VirtualNetworkName` or `VirtualNetwork.Name` over `VirtualNetwork string`

exp/api/v1alpha3/azuremanagedcontrolplane_types.go
CecileRobertMichon: Another way is to make AKS allow us to specify a vnet name

exp/controllers/azuremangedcontrolplane_reconciler.go
alexeldeib: oh right, the MC object doesn’t have it, but the agent pools have vnetSubnetID. Let’s sync up offline and I can also ask RP team if there’s any historical context on this one.

exp/api/v1alpha3/azuremanagedcontrolplane_types.go
alexeldeib: Agreed this is a little annoying in the AKS API right now, I’d likely prefer to extract it from agent pools for the sake of expediency but happy to ask about/volunteer to implement this in RP if that’s the path we want to go.

CecileRobertMichon: yeah definitely meant as an alternative for that follow up issue, for now extracting the name is good

exp/controllers/azuremangedcontrolplane_reconciler.go
alexeldeib: I don't want a reconciler here? I actually want to be able to List vnets, which the vnet-specific service exposes while the reconciler interface doesn't?

exp/controllers/azuremangedcontrolplane_reconciler.go
```diff
@@ -228,15 +233,38 @@ func (r *azureManagedControlPlaneReconciler) reconcileEndpoint(ctx context.Conte
...
+
+	var vnetName string
+	if vnetSubnetID != nil && *vnetSubnetID != "" {
+		vnetName, err = extractVnetFromSubnetID(*vnetSubnetID)
+		if err != nil {
+			return err
+		}
+	} else {
+		var clusterID string
+		clusterID, err = getClusterID(*managedCluster.DNSPrefix)
```
alexeldeib, line 250: @CecileRobertMichon I know this is a hack, but it's actually just the same as clusterID from aks-engine (I looked at the code)

Happy to remove this and rely on the vnet being populated, though (I can't imagine why it wouldn't be).

exp/controllers/azuremangedcontrolplane_reconciler.go
CecileRobertMichon: Is it still being used? I don't see the vnetSvc in use anymore

exp/controllers/azuremangedcontrolplane_reconciler.go
CecileRobertMichon: If we don't expect vnetSubnetID to be empty, I would prefer we return an error if it is. Relying on the format of the clusterID which is defined in another repo and not officially documented by AKS is dangerous IMO.

exp/controllers/azuremangedcontrolplane_reconciler.go
alexeldeib: whoops yeah, I'll remove this

exp/controllers/azuremangedcontrolplane_reconciler.go
alexeldeib: removed

cloud/defaults.go
```diff
@@ -40,11 +41,28 @@ const (
...
+	UniqueNameSuffixSize = 8
+)
+
 // GenerateInternalLBName generates a internal load balancer name, based on the cluster name.
 func GenerateInternalLBName(clusterName string) string {
 	return fmt.Sprintf("%s-%s", clusterName, "internal-lb")
 }
 
+// GenerateManagedResourceGroupName returns the name of the resource group containing IaaS resources for an AKS cluster.
+func GenerateManagedResourceGroupName(resourceGroup, clusterName, location string) string {
```
CecileRobertMichon, line 57: instead of assuming the default format here, would it make sense to get the nodeResourceGroup value from ManagedClusterProperties ?

alexeldeib: good point, I forgot it doesn't technically have to be the MC group

exp/controllers/azuremangedcontrolplane_reconciler.go
alexeldeib: So...turns out this field won't be populated unless it's a user-provisioned vnet. Maybe we should forget about this dance and just pre-provision the vnet? There's no user-facing reliable way to fetch it I think (which honestly seems very odd).

alexeldeib: apparently Paul Miller is looking at fixing this up on the RP side, I'll probably re-work this PR and pre-provision the network instead 🙁

alexeldeib: Already mentioned to Cecile offline, but AKS will expose this property publicly for all node pools. We're waiting on that functionality to move forward here (~1 or 2 weeks to full release)

PR #936, opened 2020-09-15
💚 log elapsed e2e spec time and Ginkgo node number

mboersma:

Adds some logging to the main workload e2e specs, hopefully to sort out whether they are actually running in parallel.

(They are parallelized for me locally when running ./scripts/ci-e2e.sh, so maybe the CI environment is different?)

Example output:
```
INFO: "With 1 worker node" started at Tue, 15 Sep 2020 16:10:24 MDT and ran for 6m6s on node 1 of 3
```

Refs #807 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
e2e: log elapsed spec time and Ginkgo node number
```

### DIFF
--- a/test/e2e/azure_test.go
+++ b/test/e2e/azure_test.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"time"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
@@ -42,9 +43,12 @@ var _ = Describe("Workload cluster creation", func() {
 		cancelWatches context.CancelFunc
 		cluster       *clusterv1.Cluster
 		clusterName   string
+		specTimes     = map[string]time.Time{}
 	)
 
 	BeforeEach(func() {
+		logCheckpoint(specTimes)
+
 		Expect(ctx).NotTo(BeNil(), "ctx is required for %s spec", specName)
 		Expect(e2eConfig).ToNot(BeNil(), "Invalid argument. e2eConfig can't be nil when calling %s spec", specName)
 		Expect(clusterctlConfigPath).To(BeAnExistingFile(), "Invalid argument. clusterctlConfigPath must be an existing file when calling %s spec", specName)
@@ -67,6 +71,8 @@ var _ = Describe("Workload cluster creation", func() {
 		dumpSpecResourcesAndCleanup(ctx, specName, bootstrapClusterProxy, artifactFolder, namespace, cancelWatches, cluster, e2eConfig.GetIntervals, skipCleanup)
 		Expect(os.Unsetenv(AzureResourceGroup)).NotTo(HaveOccurred())
 		Expect(os.Unsetenv(AzureVNetName)).NotTo(HaveOccurred())
+
+		logCheckpoint(specTimes)
 	})
 
 	Context("Creating a single control-plane cluster", func() {
--- a/test/e2e/helpers.go
+++ b/test/e2e/helpers.go
@@ -24,11 +24,13 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net"
-	"strings"
 	"os"
 	"path/filepath"
+	"strings"
+	"time"
 
 	. "github.com/onsi/ginkgo"
+	"github.com/onsi/ginkgo/config"
 	. "github.com/onsi/gomega"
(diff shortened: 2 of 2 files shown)


CecileRobertMichon:
As a follow up, would be great if we could log timestamps / record data on how much each phase of e2e takes, because I've seen total duraton be 90 minutes when the longest spec took 45 minutes so I wonder where all that extra time is being spent and where we could make improvements.

devigned:
> I wonder where all that extra time is being spent and where we could make improvements.

My bet is tear down.

devigned (changes requested):
Rather than doing this in each test case, could we instead to this in a `AfterEach`?

nader-ziada:
> Rather than doing this in each test case, could we instead to this in a `AfterEach`?

is is it possible to kick off a process in the background to cleanup? the test suite would still wait for AfterEach

mboersma:
> INFO: "With 1 worker node" started at Tue, 15 Sep 2020 23:37:06 UTC and ran for 5m32s on node 1 of 1

There's the problem: 1 Ginkgo node in CI. We can force it to a minimum of 3 which should work well even with a single core for these type of tests.

mboersma:
> INFO: "With 1 worker node" started at Wed, 16 Sep 2020 17:05:48 UTC and ran for 6m42s on node 1 of 1

Ok...wtf.

CecileRobertMichon:
looks better without the -focus flag... if that actually fixes it let's update the PR title / description / release note to indicate this is more than logging

mboersma:
> if that actually fixes it let's update the PR title 

The `-focus` removal isn't strictly related, so I made it a separate PR for easier digestion: #939.

mboersma:
I think this logging plus the focus change in #939 (and a similar change in test-infra) will get things running in parallel.

CecileRobertMichon:
@mboersma thoughts on 

> Rather than doing this in each test case, could we instead to this in a AfterEach?
 
?

mboersma:
> could we instead to this in a `AfterEach`?

I agree that would be preferable, but it won't work with this approach of using `defer`.

It would be better not to have to remember to add that one-liner to each spec, and arguably the spec timing should include the `BeforeEach` and `AfterEach` times. Let me see if I can come up with a way to do that. (Otherwise, I don't think this current PR is too bad.)

devigned:
@mboersma another way could be to write our own `It` func which would setup the defer, then execute the actual `It` func. That would hook the way you want and would not require everyone to remember to add that line.

^ I don't love this, but it's another option.

mboersma:
> another way could be to write our own `It` func

Only if I can call it `ItsIt`. :-p

nader-ziada:
i think this is `lgtm` now, anything else pending here @mboersma

mboersma:
> anything else pending here

@devigned got me thinking about how to refactor this. I have a version which just logs the start and end through function calls in `BeforeEach` and `AfterEach`, so no changes are required to the `It`s or to any new ones added at the same level. It also accounts for the time spent in those functions, which I think is more correct.

This existing version could be considered "good enough," so I'm fine with merging it, but if the approach I mentioned above sounds better, I'll finish it up and update this PR.

nader-ziada:
> > anything else pending here
> 
> @devigned got me thinking about how to refactor this. I have a version which just logs the start and end through function calls in `BeforeEach` and `AfterEach`, so no changes are required to the `It`s or to any new ones added at the same level. It also accounts for the time spent in those functions, which I think is more correct.
> 
> This existing version could be considered "good enough," so I'm fine with merging it, but if the approach I mentioned above sounds better, I'll finish it up and update this PR.

+1 for not changing the `it`
I think its god enough to merge and then we can incrementally improve it, but will wait to see what everyone else thinks

CecileRobertMichon:
Personally I would prefer we don't merge this right away and get it in the right place without requiring to change `It`s otherwise I have to update #646

Since it's not fixing any bugs I don't think there is any rush to merge it as is.

mboersma:
Ok, I think this implementation is a bit better since it won't require new `It` clauses to do anything. Let me know what you all think.

mboersma:
Sorry--I hadn't squashed the commits. Done now.

CecileRobertMichon (approved):
/lgtm
/approve

Issue #937, opened 2020-09-15
Cleanup tilt flavor clusters on refresh


CecileRobertMichon:
/kind feature

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]
For effective development, I would like to be able to "refresh" my tilt workload clusters to deploy a new cluster after fixing an issue in my code. Currently, the `CLUSTER_NAME` defaults to `flavor + "-template"`, which means in order to create a new cluster I need to manually delete it and wait for it to be completely deleted so the new cluster resources don't collide with the existing ones. In order to be able to easily create multiple clusters of the same flavor with Tilt using the local resource UI, we would need to:

1) change the CLUSTER_NAME default to contain a random string such that 2 clusters created one after the other have a different name. This could be `flavor + "-template-" + random_6_char_string` (`default-template-2h3dag` for example).

2) (Optional) make the tilt local resource delete the existing cluster as part of a manual update (not sure if/how this is possible with Tilt but I assume we can specify some instructions that happen when the user hits the refresh icon). 

/help

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- cluster-api-provider-azure version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

nader-ziada:
didn't get a chance to work on this
/unassign

PR #938, opened 2020-09-16
🚀  Fast delete cluster resource group when entire cluster is deleted

CecileRobertMichon:
**What this PR does / why we need it**:  TLDR:
Before:
```
time kubectl delete cluster default-template 
cluster.cluster.x-k8s.io "default-template" deleted
kubectl delete cluster default-template  0.18s user 0.09s system 0% cpu 14:49.77 total
```

After:
```
time kubectl delete cluster default-template              
cluster.cluster.x-k8s.io "default-template" deleted
kubectl delete cluster default-template  0.18s user 0.16s system 0% cpu 8:21.90 total
```

Currently, every Azure resource is deleted serially when a cluster gets deleted. In most cases, it is enough to just delete the resource group. Azure will take care of deleting all resources within that resource group in the proper order.

This does PR attempts to delete the entire resource group and skip deleting individual resources like VMs when 1) the entire cluster is being deleted and 2) when that cluster owns the resource group (which means it gets deleted as part of cluster delete). If these two conditions are not met, normal delete proceeds as before, deleting each resource type by order of dependency.

REVERTED ~This also introduces a "no wait" delete for the resource group delete, which means we don't wait for actual completion. As soon as Azure returns a 200 to confirm that the delete request was received, we return success back to the user and the resource group continues deleting in the background. This is safe for resource groups and resource groups only because it is the last resource type that we delete and no other resources depend on it.~

~What happens if the resource group is still deleting in the background and the user attempts to create a new cluster in that same RG? The user will see errors similar to:~
```
controller-runtime/controller "msg"="Reconciler error" "error"="failed to reconcile cluster services: failed to reconcile virtual network: failed to create virtual network default-template-vnet: network.VirtualNetworksClient#CreateOrUpdate: Failure sending request: StatusCode=0 -- Original Error: autorest/azure: Service returned an error. Status=\u003cnil\u003e Code=\"ResourceGroupBeingDeleted\" Message=\"The resource group 'default-template' is in deprovisioning state and cannot perform this operation.\"" "controller"="azurecluster" "name"="default-template" "namespace"="default"
```
~in controller logs until the resource group deletion is completed and the new create can proceed (at which point it does).~

This PR might also help with #937 

/hold for unit tests

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Directly delete cluster resource group when entire cluster is deleted
```

### DIFF
--- a/cloud/errors.go
+++ b/cloud/errors.go
@@ -18,11 +18,12 @@ package azure
 
 import (
 	"errors"
-
 	"github.com/Azure/go-autorest/autorest"
 )
 
-// ResourceNotFound parses the error to check if it's a resource not found
+var ErrNotOwned = errors.New("resource is not managed and cannot be deleted")
+
+// ResourceNotFound parses the error to check if it's a resource not found error.
 func ResourceNotFound(err error) bool {
 	derr := autorest.DetailedError{}
 	return errors.As(err, &derr) && derr.StatusCode == 404
--- a/cloud/services/groups/groups.go
+++ b/cloud/services/groups/groups.go
@@ -56,14 +56,14 @@ func (s *Service) Reconcile(ctx context.Context) error {
 
 // Delete deletes the resource group with the provided name.
 func (s *Service) Delete(ctx context.Context) error {
-	managed, err := s.isGroupManaged(ctx)
+	managed, err := s.IsGroupManaged(ctx)
 	if err != nil {
 		return errors.Wrap(err, "could not get resource group management state")
 	}
 
 	if !managed {
-		s.Scope.V(4).Info("Skipping resource group deletion in unmanaged mode")
-		return nil
+		s.Scope.V(2).Info("Should not delete resource group in unmanaged mode")
+		return azure.ErrNotOwned
 	}
 
 	s.Scope.V(2).Info("deleting resource group", "resource group", s.Scope.ResourceGroup())
@@ -80,7 +80,9 @@ func (s *Service) Delete(ctx context.Context) error {
 	return nil
 }
 
-func (s *Service) isGroupManaged(ctx context.Context) (bool, error) {
+// IsGroupManaged returns true if the resource group has an owned tag with the cluster name as value,
+// meaning that the resource group's lifecycle is managed.
+func (s *Service) IsGroupManaged(ctx context.Context) (bool, error) {
 	group, err := s.Client.Get(ctx, s.Scope.ResourceGroup())
 	if err != nil {
 		return false, err
--- a/controllers/azurecluster_reconciler_test.go
+++ b/controllers/azurecluster_reconciler_test.go
@@ -1,0 +1,131 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import (
+	"context"
+	"errors"
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2020-06-01/compute"
+	"github.com/golang/mock/gomock"
+	. "github.com/onsi/gomega"
+	azure "sigs.k8s.io/cluster-api-provider-azure/cloud"
+	"sigs.k8s.io/cluster-api-provider-azure/cloud/mocks"
+	"sigs.k8s.io/cluster-api-provider-azure/cloud/scope"
+	"sigs.k8s.io/cluster-api-provider-azure/cloud/services/resourceskus"
+
+	"testing"
+)
+
+type expect func(grp *mocks.MockServiceMockRecorder, vnet *mocks.MockServiceMockRecorder, sg *mocks.MockServiceMockRecorder, rt *mocks.MockServiceMockRecorder, sn *mocks.MockServiceMockRecorder, pip *mocks.MockServiceMockRecorder, lb *mocks.MockServiceMockRecorder)
+
+func TestAzureClusterReconcilerDelete(t *testing.T) {
+	cases := map[string]struct {
+		expectedError string
+		expect        expect
+	}{
+		"Resource Group is deleted successfully": {
+			expectedError: "",
+			expect: func(grp *mocks.MockServiceMockRecorder, vnet *mocks.MockServiceMockRecorder, sg *mocks.MockServiceMockRecorder, rt *mocks.MockServiceMockRecorder, sn *mocks.MockServiceMockRecorder, pip *mocks.MockServiceMockRecorder, lb *mocks.MockServiceMockRecorder) {
+				gomock.InOrder(
+					grp.Delete(context.TODO()).Return(nil))
+			},
+		},
+		"Resource Group delete fails": {
+			expectedError: "failed to delete resource group: internal error",
+			expect: func(grp *mocks.MockServiceMockRecorder, vnet *mocks.MockServiceMockRecorder, sg *mocks.MockServiceMockRecorder, rt *mocks.MockServiceMockRecorder, sn *mocks.MockServiceMockRecorder, pip *mocks.MockServiceMockRecorder, lb *mocks.MockServiceMockRecorder) {
+				gomock.InOrder(
+					grp.Delete(context.TODO()).Return(errors.New("internal error")))
+			},
+		},
+		"Resource Group not owned by cluster": {
+			expectedError: "",
+			expect: func(grp *mocks.MockServiceMockRecorder, vnet *mocks.MockServiceMockRecorder, sg *mocks.MockServiceMockRecorder, rt *mocks.MockServiceMockRecorder, sn *mocks.MockServiceMockRecorder, pip *mocks.MockServiceMockRecorder, lb *mocks.MockServiceMockRecorder) {
+				gomock.InOrder(
+					grp.Delete(context.TODO()).Return(azure.ErrNotOwned),
+					lb.Delete(context.TODO()),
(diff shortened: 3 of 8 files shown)


cloud/services/groups/groups.go
```diff
@@ -56,18 +56,18 @@ func (s *Service) Reconcile(ctx context.Context) error {
...
 	if !managed {
-		s.Scope.V(4).Info("Skipping resource group deletion in unmanaged mode")
-		return nil
+		s.Scope.V(2).Info("Should not delete resource group in unmanaged mode")
+		return azure.ErrNotOwned
 	}
 
 	s.Scope.V(2).Info("deleting resource group", "resource group", s.Scope.ResourceGroup())
-	err = s.Client.Delete(ctx, s.Scope.ResourceGroup())
+	err = s.Client.DeleteNoWait(ctx, s.Scope.ResourceGroup())
```
CecileRobertMichon, line 70: I can switch this back to Delete() if we prefer to make the RG delete blocking for now

CecileRobertMichon: I measured ~15 minutes to delete the cluster before this change vs. ~7 minutes to delete the entire resource group with this method.

cloud/errors.go
```diff
@@ -18,12 +18,18 @@ package azure
...
+var ErrNotOwned = errors.New("resource is not managed and cannot be deleted")
+
+// ResourceNotFound parses the error to check if it's a resource not found error.
 func ResourceNotFound(err error) bool {
 	derr := autorest.DetailedError{}
 	return errors.As(err, &derr) && derr.StatusCode == 404
 }
+
+// ResourceNotOwned parses the error to check if it's a resource not owned error.
+func ResourceNotOwned(err error) bool {
```
devigned, line 33: Did you specifically choose to write this method rather than using `errors.Is(err, ErrNotOwned)`?

Might be nice to use them uniformly like, `errors.Is(err, ResourceNotFound)` and `errors.Is(err, ResourceNotOwned)`.

cloud/services/groups/groups.go
```diff
@@ -76,11 +76,11 @@ func (s *Service) Delete(ctx context.Context) error {
 		return errors.Wrapf(err, "failed to delete resource group %s", s.Scope.ResourceGroup())
 	}
 
-	s.Scope.V(2).Info("successfully deleted resource group", "resource group", s.Scope.ResourceGroup())
+	s.Scope.V(2).Info("Deleting resource group. It might take a while for all resources to be completely gone...", "resource group", s.Scope.ResourceGroup())
 	return nil
 }
 
-func (s *Service) isGroupManaged(ctx context.Context) (bool, error) {
+func (s *Service) IsGroupManaged(ctx context.Context) (bool, error) {
```
devigned, line 83: What is your thoughts on enforcing comments on public APIs?

devigned (changes requested):
Fast delete is a huge improvement.

We probably want to intro an e2e test that will run fast delete and then a issue an apply with the same resources to ensure the cluster ends up in a good state.

cloud/errors.go
CecileRobertMichon: no, I didn't think of using errors.Is, I guess I just tried to be consistent with what was already there for not found. I can change it, not sure if it will work for NotFound too though (?)

cloud/services/groups/groups.go
CecileRobertMichon: I generally like them :) sometimes I forget... Will add

cloud/services/groups/groups.go
CecileRobertMichon: I'm leaning more towards making the RG delete blocking for now, that way we can test this and know if there are issues deleting RGs directly... otherwise, we risk leaking resources if there is a bug and we won't know. Thoughts?

cloud/errors.go
devigned: I think it would work for not found, but we would need to create an error type in our package which would wrap the inner not found error. Then implement `func (e *CloudError) Is(target error) bool {...}` where cloud error is that wrapping type.

Maybe this should be part of an error refactoring, rather than this PR.

Just wanted to bring it up, but I'm not super passionate about it.

cloud/services/groups/groups.go
devigned: I was thinking about intro a PR to add to `make verify` to enforce public API comments. It might require some doc work to get us clean, but it would be nice to express that in our CI checks.

CecileRobertMichon: I woudn't be opposed to that but it would require a lot of cleanup prior to being able to enforce since it would definitely fail with the codebase as-is. Do you mind opening an issue to track?

controllers/helpers.go
```diff
@@ -344,3 +346,16 @@ func GetMachinePoolByName(ctx context.Context, c client.Client, namespace, name
...
 }
+
+// ShouldDeleteIndividualResources returns false if the resource group is managed and the whole cluster is being deleted
+// meaning that we can rely on a single resource group delete operation as opposed to deleting every individual VM resource.
+func ShouldDeleteIndividualResources(ctx context.Context, clusterScope *scope.ClusterScope) bool {
+	if clusterScope.Cluster.DeletionTimestamp.IsZero() {
+		return true
+	}
+	grpSvc := groups.NewService(clusterScope)
+	managed, err := grpSvc.IsGroupManaged(ctx)
```
CecileRobertMichon, line 357: I'm having a really hard time testing this... since this makes an actual API call it needs to be mocked but the mock Client we have won't work for this since it's not of type azure.Service (azure.Service doesn't have a func `IsGroupManaged`)

devigned: Is this exposed too high up the stack? Should the services understand they should skip delete if their cluster `IsGroupManaged` rather than the controller deciding to call delete on the underlying services?

CecileRobertMichon: That'd be another way of doing it... but then that would require every single service to skip the delete instead of having the logic in one central place, which might be error-prone if new services are written in the future, no? Also would mean we would call every single service no matter what instead of deleting the RG right away (not a big deal but could be a slight de-optimization).

CecileRobertMichon: The other thing is that we're not currently storing the `IsGroupManaged` value anywhere in spec (AzureCluster.Spec.ResourceGroup is a `string`) unlike the vnet for which we store the Tags, so every time we check we're reaching out to Azure right now. We shouldn't GET the RG tags for each resource during delete, so doing that would require a bigger change to start storing the RG tags. I actually had started doing that initially and was looking at either adding an annotation or a Status field but then that causes the problem of refreshing the tags when a delete is triggered to make sure we are up to date. All of this should be possible, I mostly just thinking "out loud".

CecileRobertMichon: Also - we don't want to skip a service deletion if the resource belongs to something that isn't an AzureCluster that is getting individually deleted. For example, a public IP should be deleted if the AzureMachine it belongs to is getting deleted in the context of a scale down, but it shouldn't be deleted if the entire cluster is getting deleted. So I actually think the controller is the right layer too put this logic, not the services.

CecileRobertMichon:
> We probably want to intro an e2e test that will run fast delete and then a issue an apply with the same resources to ensure the cluster ends up in a good state.

For now I reverted the DeleteNoWait changes and decided to go for the more conservative approach of blocking on the actual RG delete before returning success. This is still a big improvement because deleting the RG directly takes almost 1/2 the time compared to deleting every resource one by one. Updated the PR description to reflect this.

cloud/services/groups/groups.go
```diff
@@ -76,11 +76,13 @@ func (s *Service) Delete(ctx context.Context) error {
 		return errors.Wrapf(err, "failed to delete resource group %s", s.Scope.ResourceGroup())
 	}
 
-	s.Scope.V(2).Info("successfully deleted resource group", "resource group", s.Scope.ResourceGroup())
+	s.Scope.V(2).Info("Deleting resource group. It might take a while for all resources to be completely gone...", "resource group", s.Scope.ResourceGroup())
```
devigned, line 79: Is this comment still accurate? Is this left over from the DeleteNoWait?

devigned, line 79: Is this comment still accurate? Is this left over from the DeleteNoWait?

controllers/azurecluster_reconciler_test.go
```diff
@@ -0,0 +1,131 @@
...
+		expect        expect
+	}{
+		"Resource Group is deleted successfully": {
+			expectedError: "",
+			expect: func(grp *mocks.MockServiceMockRecorder, vnet *mocks.MockServiceMockRecorder, sg *mocks.MockServiceMockRecorder, rt *mocks.MockServiceMockRecorder, sn *mocks.MockServiceMockRecorder, pip *mocks.MockServiceMockRecorder, lb *mocks.MockServiceMockRecorder) {
+				gomock.InOrder(
+					grp.Delete(context.TODO()).Return(nil))
+			},
+		},
+		"Resource Group is deletion fails": {
```
devigned, line 47:
```suggestion
		"Resource Group delete fails": {
```

devigned (changes requested):
Couple little items

CecileRobertMichon:
`Test started today at 9:20 AM passed after 47m4s. (more info)`

We're definitely gaining a few minutes on e2e here, cc @mboersma

nader-ziada:
I think this is `lgtm` now, excited for the speed improvement this will bring

@devigned will let you decide since you did most of the review

controllers/helpers.go
devigned: Let's see how this plays for a bit. I'm not convinced the controller is where this should live. However, I do agree that each service should not need to call out to Azure to find out if the resource group is managed. 

I do think there is another way around each service calling out to Azure by making once call early in the reconcile cycle and caching the result in the cluster scope struct. I would imagine this should be added to the `ClusterDescriber` interface. Perhaps, the implementation there would do the memoization.

With that said, I don't want to hold up the PR. It's definitely an improvement. Let's just think on this one some more.

devigned (approved):
/lgtm
/approve

This is a welcomed improvement in our delete times! I left one item to think about with regard to the controller logic vs services https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/938/files#r493637447.

PR #947, opened 2020-09-18
🔧  Apply calico ClusterResourceSet to tilt and dev clusters

CecileRobertMichon:
**What this PR does / why we need it**: Use ClusterResourceSets to apply the calico addons for tilt and dev clusters. This allows using local tilt resources to bring up fully functional clusters, without needing to apply CNI manually. 

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Apply calico ClusterResourceSet to tilt and dev clusters
```

### DIFF
--- a/templates/cluster-template.yaml
+++ b/templates/cluster-template.yaml
@@ -1,6 +1,8 @@
 apiVersion: cluster.x-k8s.io/v1alpha3
 kind: Cluster
 metadata:
+  labels:
+    cni: calico
   name: ${CLUSTER_NAME}
   namespace: default
 spec:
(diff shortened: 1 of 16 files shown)


templates/cluster-template.yaml
```diff
@@ -1,6 +1,8 @@
 apiVersion: cluster.x-k8s.io/v1alpha3
 kind: Cluster
 metadata:
+  labels:
+    cni: calico
```
CecileRobertMichon, line 5: @sedefsavas can you please confirm that adding this to the released cluster template won't have any negative effect on users who use it without cluster resource sets and continue to apply the CNI addon manually?

sedefsavas: Yes, as long as there is no `ClusterResourceSet` object that has this label as selector, it will be a no-op.
`ClusterResourceSet` labels are user selected, so there is no special meaning of any of them.

CecileRobertMichon:
/hold 

until #646 merges so I can add ipv6 in here as well

CecileRobertMichon:
Added ipv6 

/hold cancel

devigned (approved):
/lgtm
/approve

Yay! The workload clusters will have networking by default!

Since @mboersma is out today and the changes aren't super contriversial, I'm just going to approve.

PR #951, opened 2020-09-21
🐛 Skips creating secrets for machines cloned from AzureMachineTemplates

gab-satchi:
When creating a cluster with KCP managed control planes, I noticed that we create an azure-json secret for the control-plane template but also for each of the control plane nodes. We only use the one created for the template so this change is to prevent creating a secret for each control plane node. 

Fixes #

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/controllers/azurejson_machine_controller.go
+++ b/controllers/azurejson_machine_controller.go
@@ -91,13 +91,13 @@ func (f filterUnclonedMachinesPredicate) Generic(e event.GenericEvent) bool {
 	}
 
 	// when watching machines, we only care about machines users created one-off
-	// outside of machinedeployments/machinesets. if a machine is part of a machineset
+	// outside of machinedeployments/machinesets and using AzureMachineTemplates. if a machine is part of a machineset
 	// or machinedeployment, we already created a secret for the template. All machines
 	// in the machinedeployment will share that one secret.
-	_, isMachineSetNode := e.Meta.GetLabels()[clusterv1.MachineSetLabelName]
-	_, isMachineDeploymentNode := e.Meta.GetLabels()[clusterv1.MachineDeploymentLabelName]
+	gvk := infrav1.GroupVersion.WithKind("AzureMachineTemplate")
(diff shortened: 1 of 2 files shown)


controllers/azurejson_machine_controller.go
```diff
@@ -91,13 +91,15 @@ func (f filterUnclonedMachinesPredicate) Generic(e event.GenericEvent) bool {
 	}
 
 	// when watching machines, we only care about machines users created one-off
-	// outside of machinedeployments/machinesets. if a machine is part of a machineset
+	// outside of machinedeployments/machinesets and using AzureMachineTemplates. if a machine is part of a machineset
 	// or machinedeployment, we already created a secret for the template. All machines
 	// in the machinedeployment will share that one secret.
 	_, isMachineSetNode := e.Meta.GetLabels()[clusterv1.MachineSetLabelName]
 	_, isMachineDeploymentNode := e.Meta.GetLabels()[clusterv1.MachineDeploymentLabelName]
```
gab-satchi, line 98: Unsure if we still need to check MachineSetLabel and MachineDeploymentLabel since machine deployments would also use templates and we'd get a cloned-from annotation.

alexeldeib: I think you're right. This controller is primarily meant to catch one off workers and control plane nodes not part of e.g. KCP. I haven't tested the logic but I think you can simplify it here. Whether a user creates MD or MS manually, they should get the cloned-from annotation.

gab-satchi: sounds good. will update

alexeldeib: I tested this and it seems to work

alexeldeib:
lgtm pending rebase and small fix in the comments above 👍

controllers/azurejson_machine_controller.go
gab-satchi: updated to only check the cloned-from annotation.

gab-satchi:
@alexeldeib ready for review again

PR #956, opened 2020-09-23
🐛 delete data disks when deleting azure machine

nader-ziada:
- when deleting a cluster in an unmanaged resource group, any data disks were not deleted with the azure machines

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Include data disks as part of disks of an Azure machine 
```

### DIFF
--- a/cloud/scope/machine.go
+++ b/cloud/scope/machine.go
@@ -118,7 +118,7 @@ func (m *MachineScope) TagsSpecs() []azure.TagsSpec {
 	}
 }
 
-// PublicIPSpec returns the public IP specs.
+// PublicIPSpecs returns the public IP specs.
 func (m *MachineScope) PublicIPSpecs() []azure.PublicIPSpec {
 	var spec []azure.PublicIPSpec
 	if m.AzureMachine.Spec.AllocatePublicIP == true {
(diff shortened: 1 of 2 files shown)


CecileRobertMichon (approved):
lgtm

@devigned I noticed we're not deleting disks for scale sets while reviewing this, are disks automatically deleted as part of a VMSS delete? if not that's another bug to fix

devigned:
> I noticed we're not deleting disks for scale sets while reviewing this, are disks automatically deleted as part of a VMSS delete?

Hmm... I'm not sure. Need to look into that. I lean toward no.

There might be another angle on this too. Are we worried about deleting data disks folks might want to keep around?

CecileRobertMichon:
> There might be another angle on this too. Are we worried about deleting data disks folks might want to keep around?

I think disks that were created by capz should also be deleted by capz when the cluster is deleted. We don't support BYO data disk currently.

nader-ziada:
> > There might be another angle on this too. Are we worried about deleting data disks folks might want to keep around?
> 
> I think disks that were created by capz should also be deleted by capz when the cluster is deleted. We don't support BYO data disk currently.

should that be included in this PR?

CecileRobertMichon:
> should that be included in this PR?

No, let's do azuremachinepool separately, this PR lgtm as-is. 

/assign @mboersma 
for an additional review

mboersma (approved):
lgtm

CecileRobertMichon (approved):
/lgtm

PR #959, opened 2020-09-23
🐛 Make role assignment name deterministic

### LINKED ISSUE #953: Errors in logs when system assigned is enabled
/kind bug

It looks like something might have changed on the Azure side. Role assignment creation used to be idempotent https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/811#issuecomment-662173226. 

Now I'm seeing a bunch of `RoleAssignmentExists` errors in controller logs when using the system assigned flavor. 

How to repro: use latest commit and deploy tilt system assigned flavor, look at capz controller logs.

**Environment:**

CecileRobertMichon:
**What this PR does / why we need it**: Using system assigned identity causes `RoleAssignmentExists` errors in logs.

According to Azure docs:
```This template is not idempotent unless the same roleNameGuid value is provided as a parameter for each deployment of the template. If no roleNameGuid is provided, by default a new GUID is generated on each deployment and subsequent deployments will fail with a Conflict: RoleAssignmentExists error.```

https://docs.microsoft.com/en-us/azure/role-based-access-control/role-assignments-template

The solution is to use a deterministic GUID, based on a hash from static information so that the generated string will be the same every time the AzureMachine is reconciled.

Fixes #953 

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Make role assignment name deterministic
```

### DIFF
--- a/api/v1alpha3/azuremachine_types.go
+++ b/api/v1alpha3/azuremachine_types.go
@@ -66,6 +66,11 @@ type AzureMachineSpec struct {
 	// +optional
 	UserAssignedIdentities []UserAssignedIdentity `json:"userAssignedIdentities,omitempty"`
 
+	// RoleAssignmentName is the name of the role assignment to create for a system assigned identity. It can be any valid GUID.
+	// If not specified, a random GUID will be generated.
+	// +optional
+	RoleAssignmentName string `json:"roleAssignmentName,omitempty"`
+
 	// OSDisk specifies the parameters for the operating system disk of the machine
 	OSDisk OSDisk `json:"osDisk"`
 
--- a/api/v1alpha2/azuremachine_conversion.go
+++ b/api/v1alpha2/azuremachine_conversion.go
@@ -53,6 +53,7 @@ func restoreAzureMachineSpec(restored, dst *infrav1alpha3.AzureMachineSpec) {
 	if len(restored.UserAssignedIdentities) > 0 {
 		dst.UserAssignedIdentities = restored.UserAssignedIdentities
 	}
+	dst.RoleAssignmentName = restored.RoleAssignmentName
 	if restored.AcceleratedNetworking != nil {
 		dst.AcceleratedNetworking = restored.AcceleratedNetworking
 	}
--- a/api/v1alpha2/zz_generated.conversion.go
+++ b/api/v1alpha2/zz_generated.conversion.go
@@ -574,6 +574,7 @@ func autoConvert_v1alpha3_AzureMachineSpec_To_v1alpha2_AzureMachineSpec(in *v1al
 	}
 	// WARNING: in.Identity requires manual conversion: does not exist in peer-type
 	// WARNING: in.UserAssignedIdentities requires manual conversion: does not exist in peer-type
+	// WARNING: in.RoleAssignmentName requires manual conversion: does not exist in peer-type
 	if err := Convert_v1alpha3_OSDisk_To_v1alpha2_OSDisk(&in.OSDisk, &out.OSDisk, s); err != nil {
 		return err
 	}
--- a/api/v1alpha3/azuremachine_default.go
+++ b/api/v1alpha3/azuremachine_default.go
@@ -18,6 +18,7 @@ package v1alpha3
 
 import (
 	"encoding/base64"
+	"k8s.io/apimachinery/pkg/util/uuid"
 
 	"golang.org/x/crypto/ssh"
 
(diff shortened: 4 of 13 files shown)


cloud/defaults.go
```diff
@@ -100,6 +102,19 @@ func GenerateDataDiskName(machineName, nameSuffix string) string {
 	return fmt.Sprintf("%s_%s", machineName, nameSuffix)
 }
 
+// GenerateRoleAssignmentName generates a valid GUID based on an input string.
+func GenerateRoleAssignmentName(s string) string {
+	h := md5.New() // #nosec
```
CecileRobertMichon, line 107: I'm not sure if this is acceptable, the issue is that the other options for hash aren't 16 bytes long so I wouldn't be able to convert them to UUID as-is. This isn't really "insecure" because this data isn't a secret but it's getting flagged by gosec. Open to suggestions for not using md5 if anyone has any.

@nader-ziada @devigned

nader-ziada: md5 is considered unsecure because it is easy to reverse, but I don't think this has any of the information used here (subscription, location, machine name) is confidential

nader-ziada: can you add a comment so its clear why we have the `#nosec`

devigned: +1 to `// #nosec` with comment.

I don't see a security issue here since it is scoped to within the sub and as @nader-ziada said, we are using common input to derive the hash.

There is a possibility that for some input we map to some same output, which would cause a hash collision. One can generate a hash collision in seconds using MD5.

What do you think about just generating a UUID and storing it in the status?

CecileRobertMichon: > What about adding the role assignment name to the spec and generate a UUID in a web hook?

That's the other option. I thought not adding to spec would be slightly better because otherwise users could potentially mess with it and it's not really an option we want exposed to the user. But given how complicated using a hash is it might make more sense, I'm not sure...

CecileRobertMichon:
/hold

@devigned @nader-ziada I pushed a commit to add the GUID to the spec and default/validate in a webhook instead. Not adding it to any templates + docs on purpose as users should mostly rely on the generated default. Let me know what you think, if we like this approach better I'll add unit tests and squash the commits. If not, I can revert the commit.

nader-ziada:
> /hold
> 
> @devigned @nader-ziada I pushed a commit to add the GUID to the spec and default/validate in a webhook instead. Not adding it to any templates + docs on purpose as users should mostly rely on the generated default. Let me know what you think, if we like this approach better I'll add unit tests and squash the commits. If not, I can revert the commit.

I like this way better, feels more straight forward

CecileRobertMichon:
squashed and rebased 

/hold cancel

api/v1alpha3/azuremachine_types.go
```diff
@@ -66,6 +66,11 @@ type AzureMachineSpec struct {
 	// +optional
 	UserAssignedIdentities []UserAssignedIdentity `json:"userAssignedIdentities,omitempty"`
 
+	// RoleAssignmentName is the name of the role assignment to create for a system assigned identity. It can be any valid GUID.
```
devigned, line 69: Did you consider allowing any string to be specified and only generating a GUID as a default?

devigned (commented):
Just a quick question. Overall, the PR looks great.

api/v1alpha3/azuremachine_types.go
CecileRobertMichon: It has to be a valid GUID, that's an Azure requirement

devigned: Oh... You are right. I was thrown off b/c ARM usually calls things that are names, as strings rather than UUIDs.

Azure SDK docs:
```go
// Create creates a role assignment.
// Parameters:
// scope - the scope of the role assignment to create. The scope can be any REST resource instance. For
// example, use '/subscriptions/{subscription-id}/' for a subscription,
// '/subscriptions/{subscription-id}/resourceGroups/{resource-group-name}' for a resource group, and
// '/subscriptions/{subscription-id}/resourceGroups/{resource-group-name}/providers/{resource-provider}/{resource-type}/{resource-name}'
// for a resource.
// roleAssignmentName - the name of the role assignment to create. It can be any valid GUID.
// parameters - parameters for the role assignment.
```
Ok. Carry on.

devigned (approved):
/lgtm
/approve

PR #962, opened 2020-09-24
🌱 add a comment for pr authors to include a PR kind 

nader-ziada:
this kind will be used by the release note tool to generate the markdown used for the github release

 

 
 

 - remind PR authors to add a kind to the PR so it can be used by the [release-note](https://github.com/kubernetes/release) tool when generating the release notes for our github releases

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
None
```

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,7 +1,21 @@
  <!-- If this is your first PR, welcome! Please make sure you read the [contributing guidelines](../CONTRIBUTING.md). -->
 
- <!-- please add an icon to the title of this PR (see ../CONTRIBUTING.md) -->
- <!-- the icon will be either ⚠️ (:warning:, major or breaking changes), ✨ (:sparkles:, feature additions), 🐛 (:bug:, patch and bugfixes), 📖 (:book:, documentation or proposals), 💚 (`:green_heart:`, testing), 💎 (`:gem:`, refactor), 🔧 (`:wrench:`, dev tooling and chores) or 🌱 (:seedling:, minor or other) -->
+ <!-- Please label this pull request according to what type of issue you are addressing (see ../CONTRIBUTING.md) -->
+**What type of PR is this?**
+
+<!--
+Add one of the following kinds:
+/kind feature
+/kind bug
+/kind api-change
+/kind cleanup
+/kind deprecation
+/kind design
+/kind documentation
+/kind failing-test
+/kind flake
+/kind other
+-->
 
 **What this PR does / why we need it**:
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -33,17 +33,20 @@ If you're new to the project and want to help, but don't know where to start, we
 2. Fork the desired repo, develop and test your code changes.
     1. See the [Development Guide](docs/development.md) for more instructions on setting up your environment and testing changes locally.
 3. Submit a pull request.
-    1. All PRs should be labeled with one of
-         - ⚠️ (`:warning:`, major or breaking changes)
-         - ✨ (`:sparkles:`, feature additions)
-         - 🐛 (`:bug:`, patch and bug fixes)
-         - 📖 (`:book:`, documentation or proposals)
-         - 💚 (`:green_heart:`, testing)
-         - 💎 (`:gem:`, refactor)
-         - 🔧 (`:wrench:`, dev tooling and chores)
-         - 🌱 (`:seedling:`, update dependencies, minor or other)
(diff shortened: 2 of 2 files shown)


.github/pull_request_template.md
```diff
@@ -3,6 +3,20 @@
...
+Add one of the following kinds:
+/kind bug
+/kind documentation
+/kind feature
+/kind testing
+/kind refactor
+/kind dev-tooling
+/kind other
+-->
+
```
nader-ziada, line 19: - are these reasonable types?
 - do we still need the icons?

devigned: This might be an unpopular point of view, but I don't particularly like the icons. They are kind of a pain. I always try to add the `:sparkles:` or other to the title and then remember I have to go find the icon somewhere else and copy/paste it into the title b/c the text doesn't transform to the icon (etc...). I prefer the labels.

One label that may be missing is `/kind breaking-change`. I think we'll want to call these out specifically.

Perhaps, `/kind breaking-change` should be more like `/label breaking-change` as it would be a modifier of a given `/kind`.

nader-ziada: I've actually heard that from a few folks, you are not the only one that doesn't like the icons :D

CecileRobertMichon: +1 for labels, they are easier to add than emojis and can be filtered on more easily.

For breaking changes we already have the "action required" keyword in the release note that marks it as breaking in the release. There is also a label "release-note-action-required" which may or may not get applied automatically when that keyword is used in the PR description release note.

.github/pull_request_template.md
```diff
@@ -3,6 +3,20 @@
  <!-- please add an icon to the title of this PR (see ../CONTRIBUTING.md) -->
```
CecileRobertMichon, line 3: if we drop the icons we should also update CONTRIBUTING.md to match instructions in here

.github/pull_request_template.md
nader-ziada: if we agree on the labels, I can make the other changes

> There is also a label "release-note-action-required" which may or may not get applied automatically when that keyword is used in the PR description release note.

yes, it does get applied by the release-notes tool

nader-ziada: updated

devigned (approved):
lgtm
/assign @CecileRobertMichon

.github/pull_request_template.md
```diff
@@ -1,7 +1,18 @@
...
+**What type of PR is this?**
+
+<!--
+Add one of the following kinds:
+/kind bug
+/kind documentation
+/kind feature
+/kind testing
+/kind refactor
+/kind dev-tooling
```
CecileRobertMichon, line 13: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/947#issuecomment-699152093

nader-ziada: looks like there is predefined list of labels in prow https://github.com/kubernetes/test-infra/blob/master/label_sync/labels.yaml 

we can either use these existing set or create a PR to add new ones (probably not super easy)
```
/kind bug
/kind cleanup
/kind documentation
/kind feature
/kind design
/kind api-change
/kind deprecation
/kind failing-test
/kind flake
/kind regression
```

CecileRobertMichon: If we add new ones those wouldn't be taken into account by release notes right? I think the only ones missing from here are kind dev-tooling and kind testing

CecileRobertMichon: (and kind other)

CecileRobertMichon: https://github.com/kubernetes/release/blob/master/pkg/notes/notes.go#L71

nader-ziada: my suggestion is to just add these ones that the release note tool can use

nader-ziada: I thought it was just reading the kind (as a text field)

nader-ziada: - testing can be either `feature`,  `failing-test` or `flake` 
 - dev-tooling can also be a `feature` or `cleanup`

CecileRobertMichon: +1 on starting with the existing ones. Maybe we can try to get dev-tooling and testing added in parallel.

nader-ziada: made the changes to use existing ones

CecileRobertMichon (approved):
/lgtm

devigned (approved):
/lgtm
/approve

PR #967, opened 2020-09-29
Remove calico apply step from create cluster make target

CecileRobertMichon:
**What type of PR is this?**

/kind failing-test

Forgot to do this as part of #947. Calico apply is no longer needed as the management cluster has calico as ClusterResourceSets. Also fixes the create-cluster make target which currently fails with `make[1]: *** No rule to make target 'make'. `.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -483,7 +483,7 @@ create-cluster: ## Create a workload development Kubernetes cluster on Azure in 
 	EXP_AKS=true \
 	EXP_MACHINE_POOL=true \
 	$(MAKE) create-management-cluster \
-	$(MAKE) create-workload-cluster
+	create-workload-cluster
 
 .PHONY: delete-workload-cluster
 delete-workload-cluster: ## Deletes the example workload Kubernetes cluster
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -453,14 +453,6 @@ create-workload-cluster: $(ENVSUBST)
 	kubectl get secrets $(CLUSTER_NAME)-kubeconfig -o json | jq -r .data.value | base64 --decode > ./kubeconfig
 	timeout --foreground 600 bash -c "while ! kubectl --kubeconfig=./kubeconfig get nodes | grep master; do sleep 1; done"
 
-	# Deploy calico
-	@if [[ "${CLUSTER_TEMPLATE}" == *ipv6* ]]; then \
-		kubectl --kubeconfig=./kubeconfig apply -f templates/addons/calico-ipv6.yaml; \
-	else \
-		kubectl --kubeconfig=./kubeconfig apply -f templates/addons/calico.yaml; \
```
CecileRobertMichon, line 460: // TODO: check if conformance templates have the CNI label

CecileRobertMichon: so the issue with this is that the conformance tests have the "wrong" CNI label, they e2e tests have one ClusterResourceSet per cluster whereas the makefile target has one common ClusterResourceSet

Issue #952, opened 2020-09-22
Internal LB not created for IPv6


CecileRobertMichon:
The internal LB is actually not being used for anything right now for ipv4 clusters (see https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/696). I plan to start using it as part of https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/486, but there is a bug for which a fix is currently being rolled out where check IP availability doesn't work when the vnet has an IPv6 address range, so I disabled the internal LB in the meantime for ipv6 until we can rationalize this stuff in #486 once the fix is rolled out completely (I already have a POC in my local branch).

_Originally posted by @CecileRobertMichon in https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/646#discussion_r488807675_

Issue #892, opened 2020-08-21
Defaulting for internal loadbalancer ip address should be moved to a webhook


jsturtevant:
/kind feature

**Describe the solution you'd like**
The internal loadbalancer ip address default is in the Service instead of a webhook where other defaults are set:

https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/92164bbf8c1ae7073c7d3f63ade6cb9c9335f971/cloud/services/loadbalancers/loadbalancers.go#L196-L204

This logic should be moved to a webhook for defaulting

**Anything else you would like to add:**
Could move the logic to https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/api/v1alpha3/azurecluster_default.go

**Environment:**

- cluster-api-provider-azure version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

CecileRobertMichon:
I think the main reason it's not is that it calls to the vnet client to check IP address availability in order to find an available IP in the subnet, which can't be done from a webhook: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/cloud/services/loadbalancers/loadbalancers.go#L205

Issue #696, opened 2020-06-10
Internal Load balancer created for public cluster


jsturtevant:
/kind feature

**Describe the solution you'd like**

I created a cluster using `make create-workload-cluster` which uses the default flavor for workload cluster.  The workload cluster was created with an internal and public load balancer for the control plane.  If I create a public cluster using aks-engine I do not get a internal load balancer.  

It makes sense to have an internal load balancer for private clusters.  Is there a reason for internal load balancer for public clusters?  Is there reason to keep it or should only create when private clusters are enabled (#486)?

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- cluster-api-provider-azure version: master branch
- Kubernetes version: (use `kubectl version`): any
- OS (e.g. from `/etc/os-release`):

CecileRobertMichon:
@jackfrancis does AKS Engine currently provision an internal LB for non-private clusters?

@justaugustus @awesomenix do you have any context on why the internal LB was originally added in capz?

jackfrancis:
@CecileRobertMichon it does *if* there are more than one VMs backing the control plane. If there's only one VM, then no.

CecileRobertMichon:
@jackfrancis even since https://github.com/Azure/aks-engine/pull/2953?

jackfrancis:
That PR didn't change the Load Balancer implementation. I'll confirm.

jackfrancis:
That PR was for control plane-originating requests to self-route. Other cluster traffic (e.g., from nodes) needs a LB to ensure a response when a single control plane VM goes offline.

jackfrancis:
Confirmed:

```
$ az network lb list -g kubernetes-westus2-17672 | jq '. | length'
3
```

From a cluster built w/ this config:

```
...
    "masterProfile": {
      "count": 3,
...
```

CecileRobertMichon:
Thanks Jack!

@jsturtevant does that answer the question?

awesomenix:
Originally the Internal LB was created for worker nodes, not sure if thats recently changed. Just as jack mentioned.

jzhoucliqr:
Went through current code and also checked a running cluster, seems the internalLB is created but not used, all kubelets (cp and worker) still talk to the public LB.  

[...]

PR #974, opened 2020-10-02
Enable configurable API Server Load Balancer

### LINKED ISSUE #486: Private clusters
/kind feature

Support deploying clusters with no public IP or external load balancer.  For reference: https://github.com/Azure/aks-engine/blob/master/docs/topics/features.md#feat-private-cluster.

Once Azure Bastion is implemented, we should document using Azure Bastion + Private clusters. related to #165

**Environment:**

CecileRobertMichon:
**What type of PR is this?**
/kind feature
/kind api-change

**What this PR does / why we need it**: This PR allows configuring the API Server load balancer to enable things like private clusters, eg:

```
  networkSpec:
    apiServerLB:
      type: Internal
```

It adds defaulting + validating webhooks for the new Load Balancer spec. It also allows BYO api server public IP via specifying a pre-existing IP.

Check out the docs to see what this exposes to the user: https://deploy-preview-974--kubernetes-sigs-cluster-api-provider-azure.netlify.app/topics/api-server-endpoint.html

Fixes #486, #696, #952, #892  

Still work in progress, opening to get early feedback.

~Changes are currently breaking.~

**TODOs**:

- [x] squash commits
- [x] documentation
- [x] unit tests
- [x] back compat for public API server
- [x] private cluster needs management cluster in the same vnet
- [x] BYO API Server IP delete
- [x] make sure that private control plane VMs have outbound access

**Release note**:

```release-note
Allow configuration of the API Server Load Balancer, including support for private API Server endpoint.
Allow BYO API Server IP
```


api/v1alpha3/types.go
```diff
@@ -145,42 +140,48 @@ type IngressRule struct {
...
+)
+
+// FrontendIP defines a load balancer frontend IP configuration.
+type FrontendIP struct {
+	// +kubebuilder:validation:MinLength=1
+	Name string `json:"name"`
+	// +optional
+	PrivateIPAddress string `json:"privateIP,omitempty"`
+	// +optional
+	PublicIP *PublicIPSpec `json:"publicIP,omitempty"`
```
CecileRobertMichon, line 177: @alexeldeib @MirzaSikander this supersedes #652, along with https://github.com/kubernetes-sigs/cluster-api/issues/3715

CecileRobertMichon:
I still need to fix unit tests and find a way to test private clusters but otherwise, most functionality should be there so this is ready for initial review.

CecileRobertMichon:
/hold for private cluster test

@devigned @alexeldeib @nader-ziada this is ready for initial review

templates/flavors/private/patches/private-lb.yaml
```diff
@@ -0,0 +1,23 @@
...
+      - name: my-subnet-node
+        role: node
+        cidrBlocks:
+          - 10.129.0.0/16
+    apiServerLB:
+      name: ${CLUSTER_NAME}-internal-lb
+      type: Internal
+      frontendIPs:
+        - name: ${CLUSTER_NAME}-internal-lb-frontend
+          privateIP: 10.128.0.100
```
devigned, line 23: newline

devigned (approved):
I really like how this reads: `.spec.controlPlaneEndpoint.host`. Nice work.

Outside of the little nit on newline, lgtm

CecileRobertMichon:
Private cluster creation fails because of the infamous hairpin routing problem with Azure internal LBs, documented [here](https://docs.microsoft.com/en-us/azure/load-balancer/load-balancer-troubleshoot#cause-4-accessing-the-internal-load-balancer-frontend-from-the-participating-load-balancer-backend-pool-vm). 

Essentially, `kubeadm init` fails because it tries to check API Server health using the internal LB private IP from the first control plane VM. Since the VM is the only one in the LB backend at that point, it routes to itself, which fails. This is exactly the same issue described in https://github.com/kubernetes/kubeadm/issues/1685.

I verified that init succeeds when manually changing the controlPlaneEndpoint value to the first control plane VM's private IP.

Potential workarounds are 1) the very hacky iptables duct tape, 2) Azure Application Gateway (?) 3)  HAProxy 3 layers infra setup 4) modify kubeadm to allow using the local endpoint for API server health check during init.

vincepri:
@CecileRobertMichon Where can I get more information about the first two solutions? I'd avoid adding another layer of load balancing if possible. The 4th option might be ok-ish, but we'll need to check with upstream and potentially backport (which mean compatibility matrix for private clusters is going to be different)

neolit123:
> This is exactly the same issue described in kubernetes/kubeadm#1685.

well at least one part of it, that issue ended up as dumping ground for a number of LB problems some of which are not the concern of the kubeadm maintainers.

> modify kubeadm to allow using the local endpoint for API server health check during init.

i'm not objecting to that. response on slack:

> in any case, if we want to change this behavior - let's see it as a change in the KEP for kubeadm HA support first:
https://github.com/kubernetes/enhancements/issues/357
tracking issue in k/kubeadm:
https://github.com/kubernetes/kubeadm/issues/2081

https://kubernetes.slack.com/archives/C2P1JHS2E/p1603291912153700?thread_ts=1603230818.143600&cid=C2P1JHS2E

CecileRobertMichon:
@vincepri the first is to simply run a preKubeadmCommand that sets up an ip table rule `iptables -t nat -A OUTPUT -p all -d "10.128.0.100" -j DNAT --to-destination 127.0.0.1` where `10.128.0.100` is the internal LB private IP.
For the second, see https://docs.microsoft.com/en-us/azure/application-gateway/overview. I don't have a working solution of that one, it's just something that was mentioned as a workaround in Azure docs. I do think it would be more involved and require extra infrastructure (infra costs ++).

Right now what I'm thinking of doing is using the iptables hack to unblock this PR and then work on an upstream proposal to achieve 4) as a long term solution.

vincepri:
> For the second, see https://docs.microsoft.com/en-us/azure/application-gateway/overview. I don't have a working solution of that one, it's just something that was mentioned as a workaround in Azure docs. I do think it would be more involved and require extra infrastructure (infra costs ++).

This seems a good use case for the pluggable load balancer initiative that was brought up today.

> Right now what I'm thinking of doing is using the iptables hack to unblock this PR and then work on an upstream proposal to achieve 4) as a long term solution.

+1

templates/flavors/private/patches/ilb-iptables-cmd.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: controlplane.cluster.x-k8s.io/v1alpha3
+kind: KubeadmControlPlane
+metadata:
+  name: "${CLUSTER_NAME}-control-plane"
+spec:
+  kubeadmConfigSpec:
+    preKubeadmCommands:
+      # this is a workaround for the hairpin routing problem with internal load balancers
+      # during kubeadm init where 10.128.0.100 is the internal LB IP.
+      - if [ -f /tmp/kubeadm.yaml ]; then iptables -t nat -A OUTPUT -p all -d 10.128.0.100 -j DNAT --to-destination 127.0.0.1; fi
```
CecileRobertMichon, line 10: `if [ -f /tmp/kubeadm.yaml ]` should be replaced by https://github.com/kubernetes-sigs/cluster-api/issues/3828 in the future

CecileRobertMichon:
This is ready pending https://github.com/kubernetes/test-infra/pull/19659 to fix CI e2e (and squash)

CecileRobertMichon:
https://github.com/kubernetes/test-infra/pull/19678 merged

/retest

cloud/scope/cluster.go
```diff
@@ -100,57 +102,78 @@ func (s *ClusterScope) Authorizer() autorest.Authorizer {
...
-			DNSName: s.Network().APIServerIP.DNSName,
-			IsIPv6:  false, // currently azure requires a ipv4 lb rule to enable ipv6
-		},
 	}
 }
 
 // LBSpecs returns the load balancer specs.
 func (s *ClusterScope) LBSpecs() []azure.LBSpec {
-	specs := []azure.LBSpec{
+	spec := []azure.LBSpec{
```
devigned, line 130:
```suggestion
	specs := []azure.LBSpec{
```

cloud/scope/cluster.go
```diff
@@ -100,57 +102,78 @@ func (s *ClusterScope) Authorizer() autorest.Authorizer {
...
-		specs = append(specs, azure.LBSpec{
-			// Internal control plane LB
-			Name:             azure.GenerateInternalLBName(s.ClusterName()),
-			SubnetName:       s.ControlPlaneSubnet().Name,
-			SubnetCidrs:      s.ControlPlaneSubnet().CIDRBlocks,
-			PrivateIPAddress: s.ControlPlaneSubnet().InternalLBIPAddress,
-			APIServerPort:    s.APIServerPort(),
-			Role:             infrav1.InternalRole,
+
+	if s.IsAPIServerPrivate() {
```
devigned, line 158: Perhaps, invert this.

``` golang
if !s.IsAPIServerPrivate() {
    return specs
}

// do private stuff
return specs
```

cloud/scope/cluster.go
```diff
@@ -367,3 +455,34 @@ func (s *ClusterScope) SetControlPlaneIngressRules() {
...
+					},
+				},
+			},
+			Type: infrav1.Public,
+		}
+		lb.DeepCopyInto(s.APIServerLB())
+	}
+	// Generate valid FQDN if not set.
+	if !s.IsAPIServerPrivate() {
+		if s.APIServerPublicIP().DNSName == "" {
```
devigned, line 484: Perhaps, the following:

```golang
if !s.IsAPIServerPrivate() && s.APIServerPublicIP().DNSName == "" {
    //...
}

**Review by devigned — CHANGES_REQUESTED** [review]
Couple nits... The PR is really big, so I think we'd probably judge it mostly on tests.

Overall it looks like great work.

**Comment by CecileRobertMichon** [comment]
MachinePool test faied but private cluster test passed 😢 

/retest

**Comment by CecileRobertMichon** [comment]
I did some squashing with what is hopefully a logical breakdown of commits. E2E test is now passing with the addition of the iptables workaround. 

@devigned and I did some thinking about how we could set up an [Azure Private DNS Zone](https://docs.microsoft.com/en-us/azure/dns/private-dns-overview) and have a hosts entry for the apiserver DNS to redirect to localhost in each control plane. I think that would be more elegant than needing the iptables rule and would solve the hairpin routing on all control plane VMs (not just for kubeadm init). Unfortunately, private dns zones are not supported in the Go SDK yet (https://github.com/Azure/azure-sdk-for-go/issues/12965) so I think we should try to merge this PR as-is with iptables hack and switch to using private DNS as soon as it's available in the SDK.

@nader-ziada @mboersma @alexeldeib please review when you can, these are extensive changes so I'd like to get as many 👀  on it as possible. Let me know if there is any part that is unclear I'm happy to pair on review and walk through code.

/hold cancel

File: config/crd/bases/infrastructure.cluster.x-k8s.io_azureclusters.yaml [hunk 7]
```diff
@@ -487,9 +527,10 @@ spec:
                             this resource.
                           type: string
                         internalLBIPAddress:
-                          description: InternalLBIPAddress is the IP address that
+                          description: 'InternalLBIPAddress is the IP address that
```
nader-ziada, line 530: seems like an extra `'`

docs/book/src/topics/api-server-endpoint.md
```diff
@@ -0,0 +1,105 @@
...
+A `Private` cluster will have an Azure internal load balancer load balancing traffic inside the VNet to the control plane nodes.
+
+For more information on Azure load balancing, see [Load Balancer documentation](https://docs.microsoft.com/en-us/azure/load-balancer/load-balancer-overview).
+
+<aside class="note warning">
+
+<h1> Warning </h1>
+
+When creating a workload cluster with `apiServerLB` type `Internal`, the management cluster needs to be in the same VNet, or a peered VNet, as the workload cluster. Otherwise, it will not be able to access the target cluster's api server and the cluster creation will fail.  
+
```
nader-ziada, line 21: maybe add a link to how to use with [bastion](https://docs.microsoft.com/en-us/azure/bastion/bastion-overview)?

templates/cluster-template-private.yaml
```diff
@@ -0,0 +1,217 @@
...
+metadata:
+  name: ${CLUSTER_NAME}
+  namespace: default
+spec:
+  location: ${AZURE_LOCATION}
+  networkSpec:
+    apiServerLB:
+      frontendIPs:
+      - name: ${CLUSTER_NAME}-internal-lb-frontend
+        privateIP: ${AZURE_INTERNAL_LB_IP}
```
nader-ziada, line 33: couldn't find the docs any mention to `AZURE_INTERNAL_LB_IP ` as an env variable

templates/cluster-template-private.yaml
```diff
@@ -0,0 +1,217 @@
+apiVersion: cluster.x-k8s.io/v1alpha3
+kind: Cluster
+metadata:
```
nader-ziada, line 3: When I tried running `make create-workload-cluster` with `CLUSTER_TEMPLATE=cluster-template-private.yaml` I kept getting the following error
```
error: error validating "STDIN": error validating data: [ValidationError(AzureCluster.spec.networkSpec.subnets[0].cidrBlocks): unknown object type "nil" in AzureCluster.spec.networkSpec.subnets[0].cidrBlocks[0], ValidationError(AzureCluster.spec.networkSpec.subnets[1].cidrBlocks): unknown object type "nil" in AzureCluster.spec.networkSpec.subnets[1].cidrBlocks[0], ValidationError(AzureCluster.spec.networkSpec.vnet): missing required field "name" in io.x-k8s.cluster.infrastructure.v1alpha3.AzureCluster.spec.networkSpec.vnet]; if you choose to ignore these errors, turn validation off with --validate=false
make: *** [create-workload-cluster] Error 1
```
even though the regular template worked fine, maybe something with the validation, but didn't find it just yet. Did you try creating the workload cluster this way?

api/v1alpha3/types.go
```diff
@@ -145,42 +140,48 @@ type IngressRule struct {
...
-	ID        string `json:"id,omitempty"`
-	Name      string `json:"name,omitempty"`
-	IPAddress string `json:"ipAddress,omitempty"`
-	DNSName   string `json:"dnsName,omitempty"`
+// LoadBalancerSpec defines an Azure load balancer.
+type LoadBalancerSpec struct {
+	ID          string       `json:"id,omitempty"`
+	Name        string       `json:"name,omitempty"`
+	SKU         SKU          `json:"sku,omitempty"`
+	FrontendIPs []FrontendIP `json:"frontendIPs,omitempty"`
```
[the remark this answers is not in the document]
CecileRobertMichon: That's a good question. From an Azure perspective, a load balancer can have more than 1 frontend IP. For the API Server LB, we only allow one frontend IP currently but my reasoning for making the type an array was that 1) we might need more than 1 in the future and 2) we might reuse the LoadBalancerSpec type for other LoadBalancer specs in the future so it's easier to make it an array from the start so we don't have to make API changes later. I agree it makes the UX a little awkward because an array implies you can specify more than one. What do you think?

config/crd/bases/infrastructure.cluster.x-k8s.io_azureclusters.yaml
CecileRobertMichon: I think that's how deprecated fields are marked? There is no `'` in the type description, but looks like cidrBlock description is also in single quotes

api/v1alpha3/types.go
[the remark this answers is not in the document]
nader-ziada: matching the azure types is a good reason, its fine to keep it an array as long as we have a good error message in the validation and document it somewhere (maybe a comment on the type).

templates/cluster-template-private.yaml
CecileRobertMichon: removed, was only using that for the test so I made it a patch on the test template instead. The privateIP field is already documented.

templates/cluster-template-private.yaml
CecileRobertMichon: That's because you were missing some env vars for subnet CIDRs. I removed all the env vars I added in the template for the test so validation should work now. Keep in mind that your management cluster needs to be in the same vnet for this to work though so it won't work with `make create-workload-cluster` if you're using kind as your management cluster.

docs/book/src/topics/api-server-endpoint.md
CecileRobertMichon: I'll hold until we have https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/165 and have tested Bastion, I don't want to guide users to use it until we've verified it works well with capz.

api/v1alpha3/azurecluster_default_test.go
```diff
@@ -523,3 +537,59 @@ func TestSubnetDefaults(t *testing.T) {
...
 	}
 }
+
+func TestAPIServerLBDefaults(t *testing.T) {
+	cases := []struct {
+		name    string
+		cluster *AzureCluster
+		output  *AzureCluster
+	}{
+		{
```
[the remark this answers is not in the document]
CecileRobertMichon: added

cloud/scope/managedcontrolplane.go
```diff
@@ -211,3 +212,19 @@ func (s *ManagedControlPlaneScope) IsIPv6Enabled() bool {
...
+
+// IsAPIServerPrivate returns true if the API Server LB is of type Internal.
+// Currently always false as managed control planes do not currently implement private clusters.
+func (s *ManagedControlPlaneScope) IsAPIServerPrivate() bool {
+	 return false
+}
+
+
+func (s *ManagedControlPlaneScope) OutboundLBName(string) string {
+	return ""
```
alexeldeib, line 229: nit: actually `"kubernetes"`

but shouldn't be provisioned by CAPZ, only used for machines (not relevant to this pr)

CecileRobertMichon:
/retest

ipv6 flake :(

api/v1alpha3/azurecluster_validation.go
```diff
@@ -184,3 +203,64 @@ func validateIngressRule(ingressRule *IngressRule, fldPath *field.Path) *field.E
 
 	return nil
 }
+
+func validateAPIServerLB(lb LoadBalancerSpec, old LoadBalancerSpec, cidrs []string, fldPath *field.Path) field.ErrorList {
+	var allErrs field.ErrorList
+	// SKU should be Standard and is immutable.
+	if lb.SKU != SKUStandard {
```
alexeldeib, line 210: if we only support standard we should remove https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/3c5cb5b975548137d1675bf685766d624bcfa48b/api/v1alpha3/types.go#L174-L175

Do we even need the field in the spec? it's hardcoded in the service right now: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/cloud/services/loadbalancers/loadbalancers.go#L78

cloud/scope/cluster.go
```diff
@@ -100,56 +102,79 @@ func (s *ClusterScope) Authorizer() autorest.Authorizer {
...
+			FrontendIPConfigs: []infrav1.FrontendIP{
+				{
+					Name: azure.GenerateFrontendIPConfigName(s.NodeOutboundLBName()),
+					PublicIP: &infrav1.PublicIPSpec{
+						Name: azure.GenerateNodeOutboundIPName(s.ClusterName()),
+					},
+				},
+			},
+			Type:            infrav1.Public,
+			BackendPoolName: azure.GenerateOutboundBackendAddressPoolName(s.NodeOutboundLBName()),
```
alexeldeib, line 153: n.b.: https://github.com/kubernetes-sigs/cluster-api-provider-azure/compare/master...alexeldeib:ace/integrate?expand=1#diff-fcb822faaca21eede09f649536848e46d7789e73a2ba9f2ad1b50ebc1d298ed7R167

https://github.com/kubernetes-sigs/cluster-api-provider-azure/compare/master...alexeldeib:ace/integrate?expand=1#diff-bfd4ab4960c565313b05443fc66a48e60a79f9866f6a39fad03044b2161ac69eR192-R194

It'll make my life a little smoother in the future if you don't use generate + LB name on the scope, here but attach the pool name directly to the scope :)

alexeldeib (commented):
I took a pass over this, looks sane to me. Couple minor comments

alexeldeib:
this is a big pr, nice work :)

api/v1alpha3/azurecluster_default.go
```diff
@@ -36,16 +38,19 @@ const (
 	DefaultControlPlaneSubnetIPv6CIDR = "2001:1234:5678:9abc::/64"
 	// DefaultNodeSubnetIPv6CIDR is the default Node Subnet CIDR
 	DefaultNodeSubnetIPv6CIDR = "2001:1234:5678:9abd::/64"
+	// DefaultInternalLBIPv6Address is the default internal load balancer ip address
+	DefaultInternalLBIPv6Address = "2001:1234:5678:9abc::100"
```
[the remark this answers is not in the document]
CecileRobertMichon: I'm not actually using it so I can remove it, I moved it from somewhere else and didn't know if I was going to need it later

api/v1alpha3/azurecluster_default_test.go
```diff
@@ -523,3 +537,59 @@ func TestSubnetDefaults(t *testing.T) {
...
+					},
+				},
+			},
+		},
+	}
+
+	for _, c := range cases {
+		tc := c
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
```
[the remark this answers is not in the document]
CecileRobertMichon: I was sure I pushed an update to this test 😕 I wonder if I lost it in the rebase

api/v1alpha3/azurecluster_validation.go
CecileRobertMichon: we technically don't, I added it as mostly informative and with the possibility that we might support more than one later in mind

good catch though, I should definitely use it if I keep it 😬

cloud/scope/cluster.go
CecileRobertMichon: I moved it to the describer but didn't take it out of scope since everything else is there right now, we can lift and shift everything as part of #825

CecileRobertMichon:
@alexeldeib I think I've addressed all your comments, will let you take a look at the added commits before I squash 🎃

test/e2e/azure_test.go
```diff
@@ -75,111 +75,136 @@ var _ = Describe("Workload cluster creation", func() {
...
-			cluster = result.Cluster
-
-			Context("Validating time synchronization", func() {
-				AzureTimeSyncSpec(ctx, func() AzureTimeSyncSpecInput {
-					return AzureTimeSyncSpecInput{
-						BootstrapClusterProxy: bootstrapClusterProxy,
-						Namespace:             namespace,
-						ClusterName:           clusterName,
-					}
+	if os.Getenv("LOCAL_ONLY") != "true" {
```
devigned, line 78: Hmm... Did you consider:

``` go
    Skip("test requires deployment to Azure")
}
```

devigned (approved):
/lgtm

/assign @alexeldeib

nader-ziada:
🎉  great work Cecile 
lgtm

alexeldeib:
is this the hairpin issue? looks like a timeout hitting the api server with 3 master setup? didn't look too deeply I admit 


```
 Unexpected error:
      <*url.Error | 0xc0004d8090>: {
          Op: "Delete",
          URL: "https://capz-e2e-b3o4h8-a7437409.westus2.cloudapp.azure.com:6443/api/v1/namespaces/default/services/httpd-ilb",
          Err: {
              Op: "read",
              Net: "tcp",
              Source: {IP: [10, 60, 5, 156], Port: 44610, Zone: ""},
              Addr: {IP: [20, 69, 81, 80], Port: 6443, Zone: ""},
              Err: {Syscall: "read", Err: 0x6e},
          },
      }
      Delete https://capz-e2e-b3o4h8-a7437409.westus2.cloudapp.azure.com:6443/api/v1/namespaces/default/services/httpd-ilb: read tcp 10.60.5.156:44610->20.69.81.80:6443: read: connection timed out
  occurred
```

CecileRobertMichon:
> is this the hairpin issue? looks like a timeout hitting the api server with 3 master setup? didn't look too deeply I admit

no it was the HA public cluster spec that failed `capz-e2e: Workload cluster creation With 3 control-plane nodes and 2 worker nodes`, looks like a flake

CecileRobertMichon:
@alexeldeib let me know when you're okay with me rebasing

CecileRobertMichon:
Rebased + squashed

@alexeldeib @devigned @nader-ziada ready for a final look

devigned (approved):
/lgtm

PR #976, opened 2020-10-03
💚 collect workload cluster logs in e2e runs

### LINKED ISSUE #761: [e2e] Collect cloud init logs for E2E workload clusters
There have been quite a few E2E flakes lately but those are hard to investigate because most of them are due to a timeout waiting for control planes to join the cluster and there are no errors in controller logs.

Right now we only collect logs from the bootstrap cluster (that is what the CAPI test framework collects). We should also collect workload cluster logs.

For conformance tests, we have https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/hack/log/log-dump.sh#L88 to get the logs directly from VMs. An easy way to fix this would be to run the same thing in E2E.

Note: the CAPI test framework might be the best place to implement this.

mboersma:
/kind other

This implements the CAPI e2e test framework [`ClusterLogCollector`](https://github.com/kubernetes-sigs/cluster-api/blob/59faa7b1c63ba301cb780836891e7b10098fce34/test/framework/cluster_proxy.go#L81) interface for Azure machines, and connects it so these logs are saved for each node to the test client when cleaning up a test cluster:

- journal.log
- kern.log
- kubelet-version.txt
- kubelet.log
- containerd.log
- cloud-init.log
- cloud-init-output.log

Fixes #761

The SSH proxy code was adapted from AKS Engine's test/e2e package, thanks 💯 to @CecileRobertMichon.

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
💚 collect workload cluster logs in e2e runs
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -30,6 +30,7 @@ require (
 	k8s.io/utils v0.0.0-20200912215256-4140de9c8800
 	sigs.k8s.io/cluster-api v0.3.10
 	sigs.k8s.io/controller-runtime v0.5.11
+	sigs.k8s.io/kind v0.7.1-0.20200303021537-981bd80d3802
 )
 
 replace github.com/Azure/go-autorest => github.com/Azure/go-autorest v14.2.0+incompatible
--- a/test/e2e/azure_logcollector.go
+++ b/test/e2e/azure_logcollector.go
@@ -1,0 +1,84 @@
+// +build e2e
+
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package e2e
+
+import (
+	"context"
+	"path/filepath"
+
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1alpha3"
+	"sigs.k8s.io/cluster-api/util"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	kinderrors "sigs.k8s.io/kind/pkg/errors"
+)
+
+// AzureLogCollector collects logs from a CAPZ workload cluster.
+type AzureLogCollector struct{}
+
+// CollectMachineLog collects logs from a machine.
+func (k AzureLogCollector) CollectMachineLog(ctx context.Context,
+	managementClusterClient client.Client, m *clusterv1.Machine, outputPath string) error {
+	cluster, err := util.GetClusterFromMetadata(ctx, managementClusterClient, m.ObjectMeta)
+	if err != nil {
+		return err
+	}
+	controlPlaneEndpoint := cluster.Spec.ControlPlaneEndpoint.Host
+	hostname := m.Spec.InfrastructureRef.Name
+	execToPathFn := func(outputFileName, command string, args ...string) func() error {
+		return func() error {
+			f, err := fileOnHost(filepath.Join(outputPath, outputFileName))
+			if err != nil {
+				return err
+			}
+			defer f.Close()
+			return execOnHost(controlPlaneEndpoint, hostname, f, command, args...)
+		}
+	}
+
+	return kinderrors.AggregateConcurrent([]func() error{
+		execToPathFn(
+			"journal.log",
+			"journalctl", "--no-pager", "--output=short-precise",
+		),
+		execToPathFn(
+			"kern.log",
+			"journalctl", "--no-pager", "--output=short-precise", "-k",
+		),
+		execToPathFn(
+			"kubelet-version.txt",
+			"kubelet", "--version",
+		),
+		execToPathFn(
+			"kubelet.log",
+			"journalctl", "--no-pager", "--output=short-precise", "-u", "kubelet.service",
+		),
+		execToPathFn(
+			"containerd.log",
+			"journalctl", "--no-pager", "--output=short-precise", "-u", "containerd.service",
+		),
+		execToPathFn(
+			"cloud-init.log",
+			"cat", "/var/log/cloud-init.log",
+		),
+		execToPathFn(
+			"cloud-init-output.log",
+			"cat", "/var/log/cloud-init-output.log",
+		),
+	})
+}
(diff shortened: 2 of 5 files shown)


mboersma:
/status in-progress

> Failed to get logs for machine capz-e2e-ri5dua-control-plane-gsbb5, cluster create-workload-cluster-aneyc9/capz-e2e-ri5dua: open ../../.sshkey: no such file or directory

This works fine locally, but I wasn't sure about the actual CI environment. I'll come up with a strategy for locating the PK reliably.

mboersma:
I'm realizing that in the prow e2e environment, this SSH approach may not work at all, since we don't have access to the private key. :-/

mboersma:
So this works fine--see the [artifacts](https://gcsweb.k8s.io/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/976/pull-cluster-api-provider-azure-e2e/1313950225408200704/artifacts/clusters/) directory for logs. My only question is whether we should just remove the `AZURE_SSH_PUBLIC_KEY_FILE` from upstream test-infra, or hack around it here as we're doing by testing for the `CI` variable.

scripts/ci-e2e.sh
```diff
@@ -77,7 +77,8 @@ export AZURE_NODE_MACHINE_TYPE="${AZURE_NODE_MACHINE_TYPE:-"Standard_D2s_v3"}"
 
 # Generate SSH key.
 AZURE_SSH_PUBLIC_KEY_FILE=${AZURE_SSH_PUBLIC_KEY_FILE:-""}
-if [ -z "${AZURE_SSH_PUBLIC_KEY_FILE}" ]; then
+CI=${CI:=""}
+if [ "${CI}" ] || [ -z "${AZURE_SSH_PUBLIC_KEY_FILE}" ]; then
```
mboersma, line 81: If we merge kubernetes/test-infra#19498, we should undo this change.

test/e2e/azure_logcollector.go
```diff
@@ -0,0 +1,192 @@
...
+
+// execOnHost runs the specified command directly on a node's host, using an SSH connection
+// proxied through a control plane host.
+func execOnHost(controlPlaneEndpoint, hostname string, f *os.File, command string,
+	args ...string) error {
+	config, err := newSSHConfig()
+	if err != nil {
+		return err
+	}
+	port := "22"
```
CecileRobertMichon, line 102: one thing I just realized, this won't work for VMSS (I think VMSS ssh port is 2201?). We don't support VMSS control planes at the moment but it's worth making this parametrizable for the future or at least leaving a note here

CecileRobertMichon: it's 50001*

CecileRobertMichon:
@mboersma interestingly enough one of the cluster failed to provision, but somehow only the two successful cluster logs were collected: https://gcsweb.k8s.io/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/976/pull-cluster-api-provider-azure-e2e/1314584065701056512/artifacts/clusters/

`INFO: Creating the workload cluster with name "capz-e2e-g8wau2" using the "(default)" template (Kubernetes v1.18.8, 824640670448 control-plane machines, 824640670456 worker machines)`

`INFO: Creating the workload cluster with name "capz-e2e-xlls2v" using the "ipv6" template (Kubernetes v1.18.8, 824641442800 control-plane machines, 824641442808 worker machines)`  --> THIS FAILED

`INFO: Creating the workload cluster with name "capz-e2e-32onum" using the "(default)" template (Kubernetes v1.18.8, 824638051568 control-plane machines, 824638051576 worker machines)`

mboersma:
@CecileRobertMichon I noticed that. I'm confused though--it looks as if `AfterEach` wasn't called? Otherwise there would be a `Failed to collect logs for ...` message.

mboersma:
I followed the upstream pattern of logging from the `dumpSpecResourcesAndCleanup()` func that's called from Ginkgo's `AfterEach`. It appears that wasn't called at all in the case where the IPv6 cluster didn't provision one of its nodes.

I'm surprised--I thought `AfterEach` was like a Python `finally:` block that would get called regardless of errors thrown. I re-read the Ginkgo docs and it doesn't make the `AfterEach` contract very clear to me.

This is really the case where we would want these logs the most. So let me try to reproduce this locally and see if I can find a better way to hook up logging that captures the failed cluster.

CecileRobertMichon:
cc @fabriziopandini, are we expecting the `dumpSpecResourcesAndCleanup()` func in `AfterEach` to get called in the case of a failed test?

fabriziopandini:
AFAIK yes. 
What I usually see in CAPI e2e in case of errors is:

1. Entry logs for the test, e.g.
2. (An error happens, but from some Ginko weirdness the error is not reported here but after cleanup)
3. Cleanup logs
4. Failure message + stack trask

e.g. https://storage.googleapis.com/kubernetes-jenkins/logs/periodic-cluster-api-e2e-main/1314587348981256192/build-log.txt

FYI, In CAPI, I decided to have a separated spec for each test to take full advantage of parallelism (which is the main reason we are still using ginkgo for)

mboersma:
If I call `Fail()` right after the IPv6 cluster is provisioned (to simulate what happened in this CI run), it fails and collects the test logs as expected.

CecileRobertMichon:
Maybe we should get this in and track flakes to see if we're able to repro the no logs issue? It's possible prow simply timed out or the AfterEach had a panic...

mboersma:
> we should get this in and track flakes

I think that's reasonable. I'll rebase this to be more consistent with #988.

CecileRobertMichon:
lgtm

/assign @nader-ziada @devigned @cpanato

cpanato (approved):
/lgtm

go.mod
```diff
@@ -30,6 +30,7 @@ require (
 	k8s.io/utils v0.0.0-20200912215256-4140de9c8800
 	sigs.k8s.io/cluster-api v0.3.10
 	sigs.k8s.io/controller-runtime v0.5.11
+	sigs.k8s.io/kind v0.7.1-0.20200303021537-981bd80d3802
```
nader-ziada, line 33: same question as the other PR, can we use v0.8?

mboersma: Apparently not without causing a vendoring conflict with client-go.

devigned (approved):
/lgtm

@nader-ziada should we update kind in a separate PR? It looks like the go.sum was not modified, so that leads me to believe we already had a transitive dep on kind v0.7.1.

nader-ziada:
> /lgtm
> 
> @nader-ziada should we update kind in a separate PR? It looks like the go.sum was not modified, so that leads me to believe we already had a transitive dep on kind v0.7.1.

yeah, changing kind can be a different PR that comes later

PR #977, opened 2020-10-05
adding multi tenancy for controller using aad-pod-identity

### LINKED ISSUE #586: Multi-tenancy within one manager instance

nader-ziada:
**What type of PR is this?**

/kind feature

- adding azure principal and identity provider types for multi-tenancy

Fixes #586 , #778   

- work in progress

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
support for controller manager multi-tenancy 
```


nader-ziada:
@devigned I thought I would push this while I'm working on it still so you can take a look and let me know what you think of the direction

cloud/scope/clients.go
```diff
@@ -82,3 +83,32 @@ func (c *AzureClients) setCredentials(subscriptionID string) error {
...
+		subscriptionID = settings.GetSubscriptionID()
+		if subscriptionID == "" {
+			return fmt.Errorf("error creating azure services. subscriptionID is not set in cluster or AZURE_SUBSCRIPTION_ID env var")
+		}
+	}
+
+	c.EnvironmentSettings = settings
+	c.ResourceManagerEndpoint = settings.Environment.ResourceManagerEndpoint
+	c.ResourceManagerVMDNSSuffix = settings.Environment.ResourceManagerVMDNSSuffix
+	c.Values[auth.ClientID] = strings.TrimSuffix(c.Values[auth.ClientID], "\n")
```
devigned, line 107: I think the only `c.Values` needed to be set here is just the SubscriptionID. The Authorizer should encapsulate the rest of this information.

api/v1alpha3/azurecluster_types.go
```diff
@@ -48,6 +59,10 @@ type AzureClusterSpec struct {
 	// ones added by default.
 	// +optional
 	AdditionalTags Tags `json:"additionalTags,omitempty"`
+
+	// PrincipalRef is a reference to a principal to be used when reconciling this cluster
+	// +optional
+	PrincipalRef *AzurePrincipalRef `json:"principalRef,omitempty"`
```
devigned, line 65: I think we can simplify this to be the name of the identity to use for this cluster. The identity must reside inside of the cluster namespace and should be owned by the cluster.

All references should be to the known type of AzureIdentity via AAD Pod Identity.

api/v1alpha3/azureserviceprincipal_types.go
```diff
@@ -0,0 +1,72 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:path=azureclusterprincipals,scope=Namespaced,categories=cluster-api
+// +kubebuilder:storageversion
+
+// AzureClusterPrincipal represents a reference to an Azure access key ID and
+// secret access key, stored in a secret.
+type AzureClusterPrincipal struct {
```
devigned, line 30: With the simplified design, I think this can be removed.

cloud/scope/principal.go
```diff
@@ -0,0 +1,167 @@
...
+		PrincipalRef:            ref,
+		ResourceManagerEndpoint: resourceManagerEndpoint,
+		TenantID:                tenantID,
+		ClientID:                clientID,
+		ClientSecret:            clientSecret,
+	}, nil
+}
+
+// GetAuthorizer returnes Azure authorizer based on the provided principal
+func (p *AzureCredentialsProvider) GetAuthorizer(ctx context.Context) (autorest.Authorizer, error) {
```
devigned, line 93: Creating the Authorizer should be trivial using AAD pod identity. We can use the MSI endpoint provided. See the demo code here for an example of creating the Authorizer.

https://github.com/Azure/aad-pod-identity/blob/fb1bdba48ab5c1fa0be64382b6a74acdbcb41089/cmd/demo/main.go#L48

config/manager/manager_image_patch.yaml
```diff
@@ -8,5 +8,5 @@ spec:
     spec:
       containers:
       # Change the value of image field below to your controller image URL
-      - image: gcr.io/k8s-staging-cluster-api-azure/cluster-api-azure-controller:latest
+      - image: gcr.io/cf-london-servces-k8s/nziada/cluster-api-azure-controller-amd64:dev
```
devigned, line 11: oops... We need to stop manipulating files in place like this. It's too easy to check these in.

cc @CecileRobertMichon

devigned (changes requested):
Per our conversation in Slack: https://kubernetes.slack.com/archives/CEX9HENG7/p1602089427042600. I believe we are able to greatly simplify the design after talking with the AAD pod identity team. This is a great start down the multi-tenancy path.

The gist of the linked conversation:
> We should run aad pod identity within the capz namespace in namespace constrained mode. We should allow users to create `AzureIdentities` within the workload cluster namespace and have the identity owned by the `AzureCluster`. The capz controller would be responsible for copying the cluster namespace identity into the controller namespace and creating a `AzureIdentityBinding` in the controller namespace to link the label of the capz controller and the capz namespaced copy of the cluster `AzureIdentity`. The capz controller would then be responsible for deleting the identities within the capz namespace when they are no longer needed.

devigned:
Probably goes without saying, but we will also need to deploy aad pod identity components with the CAPZ components.

config/manager/manager_image_patch.yaml
CecileRobertMichon: I actually tried to add this to gitignore last week and it didn't work (??) i don't know if I was doing it wrong or what kind of dark magic was going on...

nader-ziada: I always check this one in!

nader-ziada:
> Probably goes without saying, but we will also need to deploy aad pod identity components with the capz components.

yes, will add the deployment for it, was still trying to get it to work

Makefile
```diff
@@ -438,6 +439,9 @@ create-management-cluster: $(KUSTOMIZE) $(ENVSUBST)
 	sleep 10
 	@echo 'Set kubectl context to the kind management cluster by running "kubectl config set-context kind-capz"'
 
+	# aad-pod-identity deploylment for handling identities 
+	$(ENVSUBST) < $(TEMPLATES_DIR)/addons/aad-pod-identity-deployment.yaml | kubectl apply -f -
```
CecileRobertMichon, line 443: how will this work for other ways of creating Azure clusters (Tilt, clusterctl, kustomize, etc.)? should we consider adding it to infrastructure-components?

nader-ziada: Currently in the process of getting things to work, was going to add it to the tiltfile, didn’t think about infrastructure components

cloud/scope/identity.go
```diff
@@ -0,0 +1,192 @@
...
+			return nil, errors.Wrapf(err, "failed to refresh ServicePrincipalTokenFromMSI using the MSI endpoint %s", msiEndpoint)
+		}
+		token := spt.Token()
+		if token.IsZero() {
+			return nil, errors.New("zero token found using the MSI endpoint")
+		}
+	} else if p.Identity.Spec.Type == aadpodid.UserAssignedMSI {
+		spt, err = adal.NewServicePrincipalTokenFromMSIWithUserAssignedID(msiEndpoint, resourceManagerEndpoint, p.Identity.Spec.ClientID)
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to get token from user-assigned identity")
```
nader-ziada, line 156: getting this error here so far
```
E1015 15:42:31.698604       1 mic.go:945] ignoring azure identity capz-system/user-assigned-id-nz1, error: invalid resource id: "", must match /subscriptions/<subid>/resourcegroups/<resourcegroup>/providers/Microsoft.ManagedIdentity/userAssignedIdentities/<name>
```

cloud/scope/identity.go
```diff
@@ -0,0 +1,192 @@
...
+		return nil, errors.Wrap(err, "failed to get AzureIdentity")
+	}
+	crdClient, err := newRestClient(config)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get AzureIdentity")
+	}
+
+	azureIdentity := aadpodid.AzureIdentity{}
+	err = crdClient.Get().Namespace(namespace).Resource(aadpodid.AzureIDResource).Name(identityName).Do().Into(&azureIdentity)
+	if err != nil {
```
nader-ziada, line 61: This get doesn't seem to work properly

Makefile
```diff
@@ -426,6 +426,7 @@ create-management-cluster: $(KUSTOMIZE) $(ENVSUBST)
 	kubectl wait --for=condition=Available --timeout=5m -n capi-kubeadm-bootstrap-system deployment -l cluster.x-k8s.io/provider=bootstrap-kubeadm
 	kubectl wait --for=condition=Available --timeout=5m -n capi-kubeadm-control-plane-system deployment -l cluster.x-k8s.io/provider=control-plane-kubeadm
 
+	sleep 10
```
CecileRobertMichon, line 429: you shouldn't need this anymore since #995

nader-ziada:
@devigned I updated the code to support SP identity as we discussed, please take a look 

I'm still working on docs and e2e test

nader-ziada:
@CecileRobertMichon @devigned can you please take a look, it ready for an initial review. I'm still working on adding an e2e test, but creating an identity first in azure using the sdk is not straight forward

devigned:
@nader-ziada what if we pre-provision the identity and keep it around to be used in tests? Might be easier than creating one each time.

docs/book/src/topics/multitenancy.md
```diff
@@ -0,0 +1,57 @@
...
+  location: eastus
+  networkSpec:
+    vnet:
+      name: example-cluster-vnet
+  resourceGroup: example-cluster
+  subscriptionID: <AZURE_SUBSCRIPTION_ID>
+  identityName: <name-of-identity>
+```  
+
+for more details on how aad-pod-identity work, please check the guide [here](https://azure.github.io/aad-pod-identity/docs/)
```
devigned, line 53:
```suggestion
for more details on how aad-pod-identity works, please check the guide [here](https://azure.github.io/aad-pod-identity/docs/)
```

templates/flavors/multi-tenancy/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
+namespace: default
+resources:
+  - ../default
+  - azureidentity.yaml
+patchesStrategicMerge:
+  - patches/sp-multitenancy-identity.yaml
```
devigned, line 6: newline

templates/flavors/multi-tenancy/patches/sp-multitenancy-identity.yaml
```diff
@@ -0,0 +1,7 @@
+---
+apiVersion: infrastructure.cluster.x-k8s.io/v1alpha3
+kind: AzureCluster
+metadata:
+  name: ${CLUSTER_NAME}
+spec:
+  identityName: ${CLUSTER_IDENTITY_NAME}
```
devigned, line 7: newline

devigned (changes requested):
Great work, @nader-ziada. The biggest question marks I have are related to the hardcoded namespace for the AzureIdentities and Bindings, and the validation that the IdentityName is static.

Looks really close! 🚀

api/v1alpha3/azurecluster_validation.go
```diff
@@ -54,10 +55,27 @@ func (c *AzureCluster) validateCluster() error {
...
+	if len(allErrs) == 0 {
+		return nil
+	}
+	return allErrs
+}
+
+func (c *AzureCluster) validateIdentity(old *AzureCluster, fldPath *field.Path) field.ErrorList {
+	var allErrs field.ErrorList
+	if to.String(c.Spec.IdentityName) != to.String(old.Spec.IdentityName) {
+		allErrs = append(allErrs, field.Invalid(fldPath, c.Spec.IdentityName, "IdentityName should not be modified after AzureCluster creation."))
```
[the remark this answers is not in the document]
nader-ziada: if the VMs are created with a specific identity, what would happen if that changed, the new one will just replace the existing one? or will the NMI just use the new one going forward?

cloud/scope/identity.go
```diff
@@ -0,0 +1,127 @@
...
+// GetAuthorizer returnes Azure authorizer based on the provided azure identity
+func (p *AzureCredentialsProvider) GetAuthorizer(ctx context.Context, resourceManagerEndpoint string) (autorest.Authorizer, error) {
+	copiedIdentity := &aadpodv1.AzureIdentity{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "AzureIdentity",
+			APIVersion: "aadpodidentity.k8s.io/v1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      p.Identity.Name,
+			Namespace: "capz-system",
```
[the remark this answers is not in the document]
nader-ziada: its hard-coded in the deployment yamls in the config folder, but I agree we should remove the hard-coding, will fix that

api/v1alpha3/azurecluster_validation.go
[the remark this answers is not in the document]
devigned: The identity used to provision the infrastructure (`Cluster.Spec.IdentityName`) can change between each reconcile loop as long as it had the RBAC rights to mutate the infrastructure. This is the identity the CAPZ controller will assume via AAD Pod Identity to make REST requests to Azure to build / manipulate workload cluster infrastructure.

The identity used to provision the infrastructure is not the System Assigned or User Assigned Identity on provisioned workload machine.

I'd expect the new identity to be added to the controller (identity / binding moved into the controller namespace), the old IdentityBinding, possibly Identity, to be deleted from the controller namespace, and the next reconcile loop for the cluster resources to use a new Authorizer created using the new Identity ClientID.

cloud/scope/identity.go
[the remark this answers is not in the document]
nader-ziada: I think it has to be capz-system

CecileRobertMichon:
> but creating an identity first in azure using the sdk is not straight forward

you could do something like https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/974/files#diff-a18d58ea4b72ff83f83e62ae5a6500ea0de4759cdf9da9cdefa97b3b728d03b6R122

api/v1alpha3/azurecluster_types.go
```diff
@@ -48,6 +51,10 @@ type AzureClusterSpec struct {
 	// ones added by default.
 	// +optional
 	AdditionalTags Tags `json:"additionalTags,omitempty"`
+
+	// IdentityName is a reference to a AzureIdentity to be used when reconciling this cluster
+	// +optional
+	IdentityName *string `json:"identityName,omitempty"`
```
CecileRobertMichon, line 57: This API change doesn't match what was proposed to https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/docs/proposals/20200720-single-controller-multitenancy.md#cluster-api-provider-azure-v1alpha3-types 

If changes to the design had to be made, let's first update the proposal so it matches the implementation.

api/v1alpha3/azurecluster_validation.go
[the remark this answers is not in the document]
CecileRobertMichon: That makes sense to me, especially for cases where the identity needs to be rotated

cloud/scope/identity.go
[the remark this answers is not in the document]
CecileRobertMichon: Is that documented in the CAPI provider contract? If not it's probably not safe to assume

api/v1alpha3/azurecluster_types.go
nader-ziada: @devigned you mentioned you update the proposal? or I can do it if you haven't already changed it

controllers/azurecluster_reconciler.go
```diff
@@ -138,7 +141,41 @@ func (r *azureClusterReconciler) Delete(ctx context.Context) error {
...
+	if err := r.scope.Client.List(ctx, &bindings, client.InNamespace("capz-system")); err != nil {
+		return errors.Wrapf(err, "failed to list AzureIdentityBinding")
+	}
+	for _, b := range bindings.Items {
+		binding := b
+		if binding.ObjectMeta.Labels[clusterv1.ClusterLabelName] == r.scope.AzureCluster.Name &&
+			binding.ObjectMeta.Labels[infrav1.ClusterLabelNamespace] == r.scope.AzureCluster.Namespace {
+
+			identityNamesToDelete = append(identityNamesToDelete, binding.Spec.AzureIdentity)
+			if err := r.scope.Client.Delete(ctx, &binding); err != nil {
```
CecileRobertMichon, line 165: I might be misreading the code, but, are we deleting the identity that wasn't created by CAPZ? or is it created somewhere else? 

If it's applied as part of the deployment, would it make sense to have this in a separate reconciler since it's not technically part of the AzureCluster resource?

controllers/azurecluster_reconciler.go
```diff
@@ -138,7 +141,41 @@ func (r *azureClusterReconciler) Delete(ctx context.Context) error {
...
+				return errors.Wrapf(err, "failed to delete AzureIdentityBinding")
+			}
+		}
+	}
+	for _, identityName := range identityNamesToDelete {
+		azureIdentity := &aadpodv1.AzureIdentity{}
+		if err := r.scope.Client.Get(ctx, client.ObjectKey{Name: identityName, Namespace: "capz-system"}, azureIdentity); err != nil {
```
CecileRobertMichon, line 172: `capz-system` should be a const reused across the different places that have it hardcoded

nader-ziada:
> > but creating an identity first in azure using the sdk is not straight forward
> 
> you could do something like https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/974/files#diff-a18d58ea4b72ff83f83e62ae5a6500ea0de4759cdf9da9cdefa97b3b728d03b6R122

I was thinking I would create the identity in the e2e script using `az ad sp create-for-rbac`. Is that reasonable? or I can change it if you prefer, but there is no straight forward azure sdk api to create a service principal  https://github.com/Azure/azure-sdk-for-go/issues/596

CecileRobertMichon:
@nader-ziada what about https://github.com/Azure-Samples/azure-sdk-for-go-samples/blob/master/graphrbac/graph.go#L43 (from https://github.com/Azure/azure-sdk-for-go/issues/596#issuecomment-392813961) ? using CLI commands is okay as last resort but doing this programmatically would be preferable IMO

controllers/azurecluster_reconciler.go
nader-ziada: its hardcoded in the yaml, not in the code, I'm removing the hard-coding in the code, but don't know how to do that with yaml https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/f9da6a5e82da5ccd7c692f47704015be27f7ba6e/config/default/kustomization.yaml#L1

nader-ziada:
> @nader-ziada what about https://github.com/Azure-Samples/azure-sdk-for-go-samples/blob/master/graphrbac/graph.go#L43 (from [Azure/azure-sdk-for-go#596 (comment)](https://github.com/Azure/azure-sdk-for-go/issues/596#issuecomment-392813961)) ? using CLI commands is okay as last resort but doing this programmatically would be preferable IMO

ok, let me try that

controllers/azurecluster_reconciler.go
nader-ziada: so we create a copy of the azureidentity in the capz-system namespace so the NMI pod can use it, and that's the one we delete. its created as part of the initialization of the cluster scope 

do you think it needs a different reconciler to remove it if the cluster is deleted? I had that at some point, but removed it

api/v1alpha3/azurecluster_types.go
devigned: Will do. Just been short on time.

api/v1alpha3/azurecluster_validation.go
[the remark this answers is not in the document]
nader-ziada: ok, will remove this validation, but that means a new reconciler to remove the identities no longer used

nader-ziada:
still have to update e2e tests and remove the validation about updating the identity on cluster

nader-ziada:
@CecileRobertMichon @devigned I think I addressed all the comments, but had to take out the e2e test I already had for now because the ci user doesn't have permission to create an AAD identity `Insufficient privileges to complete the operation`.
`

nader-ziada:
also not sure if we need to wait for a change with clusterctl move?

test/e2e/azure_test.go
```diff
@@ -350,4 +350,5 @@ var _ = Describe("Workload cluster creation", func() {
 			})
 		})
 	})
+
```
devigned, line 353: Looks like a stray extra newline.

devigned (changes requested):
The identity reconciler looks great! 

I have one superficial change request, and one left over from last review: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/977/files#r516679923

Looks great!

controllers/azurecluster_reconciler.go
nader-ziada: we now have azureidentity_controller

cloud/scope/identity.go
```diff
@@ -0,0 +1,127 @@
...
+			Name:      fmt.Sprintf("%s-binding", copiedIdentity.Name),
+			Namespace: copiedIdentity.Namespace,
+			Labels: map[string]string{
+				clusterv1.ClusterLabelName:    p.AzureCluster.Name,
+				infrav1.ClusterLabelNamespace: p.AzureCluster.Namespace,
+			},
+		},
+		Spec: aadpodv1.AzureIdentityBindingSpec{
+			AzureIdentity: copiedIdentity.Name,
+			Selector:      "capz-controller-aadpodidentity-selector", //should be same as selector added on controller
```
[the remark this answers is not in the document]
nader-ziada: done, sorry missed this one

api/v1alpha3/azurecluster_validation.go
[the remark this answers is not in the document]
nader-ziada: removed validation and added reconciler

test/e2e/azure_test.go
nader-ziada: removed

api/v1alpha3/azurecluster_types.go
CecileRobertMichon: let's make sure we update the proposal before merging this PR 

/hold

cloud/defaults.go
```diff
@@ -39,6 +39,16 @@ const (
 	LatestVersion = "latest"
 )
 
+// GenerateInternalLBName generates a internal load balancer name, based on the cluster name.
+func GenerateInternalLBName(clusterName string) string {
```
CecileRobertMichon, line 43: why is this needed? It should have been removed since #974

cloud/defaults.go
```diff
@@ -39,6 +39,16 @@ const (
...
+	return fmt.Sprintf("%s-%s", clusterName, "internal-lb")
+}
+
+// GeneratePublicLBName generates a public load balancer name, based on the cluster name.
+func GeneratePublicLBName(clusterName string) string {
```
CecileRobertMichon, line 48: same question here

config/aadpodidentity/aad-pod-identity-deployment.yaml
```diff
@@ -0,0 +1,158 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
CecileRobertMichon, line 1: @aramase can you please take a look to make sure everything looks good from an AAD pod identity standpoint?

cloud/defaults.go
nader-ziada: must have missed something with the rebase, fixed it now

cloud/defaults.go
nader-ziada: same

api/v1alpha3/azurecluster_types.go
nader-ziada: also wondering if we need a change to clusterctl move before merging this in

CecileRobertMichon: is the identity owned by the cluster? if so it should be moved with it already

should be pretty easy to validate that assumption

config/aadpodidentity/aad-pod-identity-deployment.yaml
aramase: Thank you for the ping. I'll take a look at it today.

api/v1alpha3/azurecluster_types.go
nader-ziada: I don't think its owned by the cluster, need to double check that

cloud/scope/identity.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+// NewAzureCredentialsProvider creates a new AzureCredentialsProvider from the supplied inputs.
+func NewAzureCredentialsProvider(ctx context.Context, kubeClient client.Client, azureCluster *infrav1.AzureCluster, identityName, namespace string) (*AzureCredentialsProvider, error) {
+	if identityName == "" {
+		return nil, errors.New("failed to generate new AzureCredentialsProvider from empty identityName")
+	}
+
+	azureIdentity := &aadpodv1.AzureIdentity{}
+	err := kubeClient.Get(ctx, client.ObjectKey{Name: identityName, Namespace: namespace}, azureIdentity)
```
aramase, line 49: If the `identityName` is configured, then it's only applicable for pod identity? Can they configure managed identity without pod identity?

cloud/scope/identity.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	return &AzureCredentialsProvider{
+		Client:       kubeClient,
+		AzureCluster: azureCluster,
+		Identity:     azureIdentity,
+	}, nil
+}
+
+// GetAuthorizer returnes Azure authorizer based on the provided azure identity
```
aramase, line 64: nit: s/returnes/returns

cloud/scope/identity.go
```diff
@@ -0,0 +1,129 @@
...
+			Name:      p.Identity.Name,
+			Namespace: infrav1.ControllerNamespace,
+			Annotations: map[string]string{
+				aadpodv1.BehaviorKey: "namespaced",
+			},
+			Labels: map[string]string{
+				clusterv1.ClusterLabelName:    p.AzureCluster.Name,
+				infrav1.ClusterLabelNamespace: p.AzureCluster.Namespace,
+			},
+			OwnerReferences: p.AzureCluster.OwnerReferences,
```
aramase, line 81: +1

cloud/scope/identity.go
```diff
@@ -0,0 +1,129 @@
...
+
+// GetAuthorizer returnes Azure authorizer based on the provided azure identity
+func (p *AzureCredentialsProvider) GetAuthorizer(ctx context.Context, resourceManagerEndpoint string) (autorest.Authorizer, error) {
+	copiedIdentity := &aadpodv1.AzureIdentity{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "AzureIdentity",
+			APIVersion: "aadpodidentity.k8s.io/v1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      p.Identity.Name,
```
aramase, line 72: How do we handle `AzureIdentity` with same name in multi namespaces? When run in namespaced mode, user could configure `azid1` in `ns1` and `azid1` in `ns2` (both with different `clientIDs`).

cluster1 referring to `ns1/azid1` and cluster referring to `ns2/azid` will result in conflict while copied in controller namespace? Should we add the namespace to the copied `AzureIdentity` name?

config/aadpodidentity/aad-pod-identity-deployment.yaml
```diff
@@ -0,0 +1,158 @@
...
+kind: CustomResourceDefinition
+metadata:
+  name: azureassignedidentities.aadpodidentity.k8s.io
+spec:
+  group: aadpodidentity.k8s.io
+  version: v1
+  names:
+    kind: AzureAssignedIdentity
```
aramase, line 9: `AzureAssignedIdentity` crd isn't required for managed mode. 

The deploy yaml for managed mode: https://github.com/Azure/aad-pod-identity/blob/v1.6.3/deploy/infra/managed-mode-deployment.yaml

config/rbac/role.yaml
```diff
@@ -29,6 +29,44 @@ rules:
   - patch
   - update
   - watch
+- apiGroups:
+  - aadpodidentity.k8s.io
+  resources:
+  - azureassignedidentities
```
aramase, line 35: `azureassignedidentities` are not required for managed mode

controllers/azureidentity_controller.go
```diff
@@ -0,0 +1,167 @@
...
+		predicates.ClusterUnpaused(log),
+	); err != nil {
+		return errors.Wrapf(err, "failed adding a watch for ready clusters")
+	}
+
+	return nil
+}
+
+// +kubebuilder:rbac:groups=aadpodidentity.k8s.io,resources=azureassignedidentities,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=aadpodidentity.k8s.io,resources=azureassignedidentities/status,verbs=get;update;patch
```
aramase, line 81: Not required for managed mode

docs/book/src/topics/multitenancy.md
```diff
@@ -0,0 +1,57 @@
+# Multi-tenancy
+
+To enable single controller multi-tenancy, a different Identity can be added to the Azure Cluster that will be used as the Azure Identity when creating Azure resources related to that cluster.
+
+This is achieved using the [aad-pod-identity](https://azure.github.io/aad-pod-identity) library. 
+
+## Service Principal Identity
+
+Once a new SP Identity is created in Azure, a new resource of AzureIdentity should be created in the managment cluster, for example
```
aramase, line 9: Might be good to add a reference to this doc - https://azure.github.io/aad-pod-identity/docs/concepts/azureidentity/

main.go
```diff
@@ -73,6 +76,20 @@ func init() {
...
 	// +kubebuilder:scaffold:scheme
+
+	// Add aadpodidentity v1 to the scheme.
+	aadPodIdentityGroupVersion := schema.GroupVersion{Group: aadpodv1.CRDGroup, Version: aadpodv1.CRDVersion}
+	scheme.AddKnownTypes(aadPodIdentityGroupVersion,
+		&aadpodv1.AzureIdentity{},
+		&aadpodv1.AzureIdentityList{},
+		&aadpodv1.AzureIdentityBinding{},
+		&aadpodv1.AzureIdentityBindingList{},
+		&aadpodv1.AzureAssignedIdentity{},
```
aramase, line 87: Not required for managed mode

aramase (commented):
Added few comments.

Also, does pod identity get deployed by default with this PR or is it only deployed when there is an `AzureCluster` with `identityName`? If deployed by default, how would we handle scenarios where the cluster already has pod identity running?

cloud/scope/identity.go
nader-ziada: yes this is only used for the pod identity, and its optional, users can already configure system assigned and user managed identity before this change

cloud/scope/identity.go
nader-ziada: good idea, will add the namespace to the name

config/aadpodidentity/aad-pod-identity-deployment.yaml
nader-ziada: will remove

config/rbac/role.yaml
nader-ziada: will remove as well

cloud/scope/identity.go
nader-ziada: > does pod identity get deployed by default with this PR or is it only deployed when there is an `AzureCluster` with `identityName`? If deployed by default, how would we handle scenarios where the cluster already has pod identity running in `standard` mode?

yes, its installed on the management cluster by default in the same namespace as the manager controller, what is a use case where we would have the standard mode also running?

nader-ziada:
> does pod identity get deployed by default with this PR or is it only deployed when there is an AzureCluster with identityName? If deployed by default, how would we handle scenarios where the cluster already has pod identity running in standard mode?

@aramase  yes, its installed on the management cluster by default in the same namespace as the manager controller, what is a use case where we would have the standard mode also running?

aramase:
> @aramase yes, its installed on the management cluster by default in the same namespace as the manager controller, what is a use case where we would have the standard mode also running?

@nader-ziada It's possible the user is using pod identity in standard/managed mode as part of their workload deployments. If they want to also use the cluster as mgmt cluster, then capz controller will try to deploy pod identity again in managed mode in `capz-system` namespace. This would mean the NMI pods in `capz-system` namespace will fail to run because of host port conflicts. This will not actually affect the other functioning NMI pods the user has deployed in their own/kube-system namespace. 

What if we do provide user's the option to opt out from installing pod identity as part of the capz controller and instead just manage it on their own? By default we can still continue it install it.

nader-ziada:
> > @aramase yes, its installed on the management cluster by default in the same namespace as the manager controller, what is a use case where we would have the standard mode also running?
> 
> @nader-ziada It's possible the user is using pod identity in standard/managed mode as part of their workload deployments. If they want to also use the cluster as mgmt cluster, then capz controller will try to deploy pod identity again in managed mode in `capz-system` namespace. This would mean the NMI pods in `capz-system` namespace will fail to run because of host port conflicts. This will not actually affect the other functioning NMI pods the user has deployed in their own/kube-system namespace.
> 
> What if we do provide user's the option to opt out from installing pod identity as part of the capz controller and instead just manage it on their own? By default we can still continue it install it.

I'm not sure how we can do that with kustomize, I'll look into if we can only deploy a `deployment` based on flag 
cc @devigned

nader-ziada:
@CecileRobertMichon @devigned 
The only two things outstanding here are:

 - What do we want to do if a users doesn't wants to deploy aad-pod-identity as part of capz
```
@aramase yes, its installed on the management cluster by default in the same namespace as the manager controller, what is a use case where we would have the standard mode also running?

@nader-ziada It's possible the user is using pod identity in standard/managed mode as part of their workload deployments. If they want to also use the cluster as mgmt cluster, then capz controller will try to deploy pod identity again in managed mode in capz-system namespace. This would mean the NMI pods in capz-system namespace will fail to run because of host port conflicts. This will not actually affect the other functioning NMI pods the user has deployed in their own/kube-system namespace.

What if we do provide user's the option to opt out from installing pod identity as part of the capz controller and instead just manage it on their own? By default we can still continue it install it.

I'm not sure how we can do that with kustomize, I'll look into if we can only deploy a deployment based on flag
cc @devigned
```
- what is supposed to happen with clusterctl move, its currently not moving the AzureIdentity and AzureIdentityBinding in the capz-system namespace even though they have the cluster as an OwnerReference and the clusterctl move label
https://github.com/kubernetes-sigs/cluster-api/issues/3933

nader-ziada:
> * what is supposed to happen with clusterctl move, its currently not moving the AzureIdentity and AzureIdentityBinding in the capz-system namespace even though they have the cluster as an OwnerReference and the clusterctl move label
>   [kubernetes-sigs/cluster-api#3933](https://github.com/kubernetes-sigs/cluster-api/issues/3933)

The issue with this is not just the move, but its because we expect the user to create an AzureIdentity with the cluster and we copy it over to the capz-system namespace so the nmi pod can use it, but we can't move that user created identity since its not in clusterapi control

nader-ziada:
@devigned @CecileRobertMichon added the custom type to azurecluster to deal with with the `move` command as we discussed, PTAL

api/v1alpha3/types.go
```diff
@@ -321,6 +323,42 @@ type UserAssignedIdentity struct {
...
+	// UserAssignedMSI or Service Principal
+	Type IdentityType `json:"type"`
+	// User assigned MSI resource id.
+	// +optional
+	ResourceID string `json:"resourceID,omitempty"`
+	// Both User Assigned MSI and SP can use this field.
+	ClientID string `json:"clientID"`
+	// Used for service principal
+	// +optional
+	ClientPassword corev1.SecretReference `json:"clientPassword,omitempty"`
```
devigned, line 357:
```suggestion
	ClientSecret corev1.SecretReference `json:"clientSecret,omitempty"`
```

These are usually referred to as "Client Secrets" in UI and doc https://docs.microsoft.com/en-us/azure/active-directory/develop/howto-create-service-principal-portal#option-2-create-a-new-application-secret. Using password, might make it seem like a only the text secret is usable rather than certificate secret as well.

api/v1alpha3/types.go
```diff
@@ -321,6 +323,42 @@ type UserAssignedIdentity struct {
...
+	// Name is the name of the identity
+	Name string `json:"name"`
+	// UserAssignedMSI or Service Principal
+	Type IdentityType `json:"type"`
+	// User assigned MSI resource id.
+	// +optional
+	ResourceID string `json:"resourceID,omitempty"`
+	// Both User Assigned MSI and SP can use this field.
+	ClientID string `json:"clientID"`
+	// Used for service principal
```
devigned, line 355:
```suggestion
	// ClientSecret is a secret reference which should contain either a Service Principal password or certificate secret.
```

cloud/scope/identity.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+// NewAzureCredentialsProvider creates a new AzureCredentialsProvider from the supplied inputs.
+func NewAzureCredentialsProvider(ctx context.Context, kubeClient client.Client, azureCluster *infrav1.AzureCluster) (*AzureCredentialsProvider, error) {
+	if azureCluster.Spec.IdentitySpec == nil {
+		return nil, errors.New("failed to generate new AzureCredentialsProvider from empty identityName")
+	}
+
+	if azureCluster.Spec.IdentitySpec.Type != infrav1.ServicePrincipal {
+		return nil, errors.New("AzureIdentity is not of type Service Principal")
```
devigned, line 50: WDYT about a web hook validation for this?

templates/flavors/multi-tenancy/patches/sp-multitenancy-identity.yaml
```diff
@@ -0,0 +1,13 @@
...
+metadata:
+  name: ${CLUSTER_NAME}
+spec:
+  identitySpec: 
+    name: ${CLUSTER_IDENTITY_NAME}
+    type: ServicePrincipal
+    tenantID: ${AZURE_TENANT_ID}
+    clientID: ${AZURE_CLUSTER_IDENTITY_CLIENT_ID}
+    clientPassword: {"name":"${AZURE_CLUSTER_IDENTITY_SECRET_NAME}","namespace":"${AZURE_CLUSTER_IDENTITY_SECRET_NAMESPACE}"}
+
```
devigned, line 13: new line

test/e2e/azure_identity.go
```diff
@@ -0,0 +1,66 @@
...
+	ClusterName           string
+	SkipCleanup           bool
+	IPv6                  bool
+}
+
+// AzureServicePrincipalIdentitySpec implements a test that verifies Azure identity can be added to
+// an AzureCluster and used as the identity for that cluster.
+func AzureServicePrincipalIdentitySpec(ctx context.Context, inputGetter func() AzureServicePrincipalIdentitySpecInput) {
+	var (
+		specName = "azure-identity"
```
devigned, line 46: const?

api/v1alpha3/types.go
```diff
@@ -321,6 +323,42 @@ type UserAssignedIdentity struct {
...
+const (
+	// UserAssignedMSI represents a user-assigned identity.
+	UserAssignedMSI IdentityType = "UserAssignedMSI"
+
+	// ServicePrincipal represents a service principal.
+	ServicePrincipal IdentityType = "ServicePrincipal"
+)
+
+// IdentitySpec specifies the parameters that are used to create an AzureIdentity
+type IdentitySpec struct {
```
devigned, line 345: We had discussed having `AllowedNamespaces` label selector on the `IdentitySpec`, similar to the design approach in CAPA: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/docs/proposal/20200506-single-controller-multitenancy.md#namespace-restrictions

This could be added in a subsequent PR.

devigned (commented):
Looks great. I've added some small feedback items.

api/v1alpha3/types.go
nader-ziada: I started adding it, will submit it with the rest of the comments

test/e2e/azure_identity.go
nader-ziada: this just the name of the spec

nader-ziada:
@devigned addressed all the comments except `allowed namespaces`, need to think about it a bit more

controllers/azurecluster_controller.go
```diff
@@ -215,6 +217,23 @@ func (r *AzureClusterReconciler) reconcileNormal(ctx context.Context, clusterSco
...
+		identity, err := GetClusterIdentityFromRef(ctx, clusterScope.Client, azureCluster.Namespace, azureCluster.Spec.IdentityRef)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		if identity.Namespace == azureCluster.Namespace {
+			patchhelper, err := patch.NewHelper(identity, r.Client)
+			if err != nil {
+				return reconcile.Result{}, errors.Wrap(err, "failed to init patch helper")
+			}
+			identity.ObjectMeta.OwnerReferences = azureCluster.GetOwnerReferences()
```
devigned, line 230: I'm wondering if this would be better placed in the cluster scope so that upon scope.Close() the patch will be ensured.

controllers/azureidentity_controller.go
```diff
@@ -0,0 +1,166 @@
...
+type AzureIdentityReconciler struct {
+	client.Client
+	Log              logr.Logger
+	Recorder         record.EventRecorder
+	ReconcileTimeout time.Duration
+}
+
+// SetupWithManager initializes this controller with a manager
+func (r *AzureIdentityReconciler) SetupWithManager(mgr ctrl.Manager, options controller.Options) error {
+	log := r.Log.WithValues("controller", "AzureCluster")
```
devigned, line 56: I think there might be some copypasta here.

templates/flavors/multi-tenancy/azure-cluster-identity.yaml
```diff
@@ -0,0 +1,10 @@
+---
+apiVersion: infrastructure.cluster.x-k8s.io/v1alpha3
+kind: AzureClusterIdentity
+metadata:
+  name: "${CLUSTER_IDENTITY_NAME}"
+spec:
+  type: ServicePrincipal
+  tenantID: "${AZURE_TENANT_ID}"
+  clientID: "${AZURE_CLUSTER_IDENTITY_CLIENT_ID}"
+  clientSecret: {"name":"${AZURE_CLUSTER_IDENTITY_SECRET_NAME}","namespace":"${AZURE_CLUSTER_IDENTITY_SECRET_NAMESPACE}"}
```
devigned, line 10: new line

devigned (changes requested):
Mostly little nits. Looking good.

api/v1alpha3/azureclusteridentity_types.go
```diff
@@ -0,0 +1,98 @@
...
+	// pointer because the nil behavior (no namespaces) is undesirable here.
+	// +optional
+	AllowedNamespaces metav1.LabelSelector `json:"allowedNamespaces"`
+}
+
+// AzureClusterIdentityStatus defines the observed state of AzureClusterIdentity
+type AzureClusterIdentityStatus struct {
+	// Ready is true when the provider resource is ready.
+	// +optional
+	Ready bool `json:"ready"`
```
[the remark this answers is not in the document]
nader-ziada: not really,

controllers/azurecluster_controller.go
nader-ziada: ok, makes sense, will move it

nader-ziada: so the patch helper takes the azurecluster when its created, so it will not work with patching the identity

nader-ziada:
@devigned ready for another look

docs/book/src/topics/multitenancy.md
```diff
@@ -0,0 +1,74 @@
+# Multi-tenancy
+
+To enable single controller multi-tenancy, a different Identity can be added to the Azure Cluster that will be used as the Azure Identity when creating Azure resources related to that cluster.
+
+This is achieved using the [aad-pod-identity](https://azure.github.io/aad-pod-identity) library. 
+
+## Service Principal Identity
+
+Once a new SP Identity is created in Azure, the coresponding values should be used to create an `AzureClusterIdentity` resource
```
CecileRobertMichon, line 9:
```suggestion
Once a new SP Identity is created in Azure, the corresponding values should be used to create an `AzureClusterIdentity` resource:
```

docs/book/src/topics/multitenancy.md
```diff
@@ -0,0 +1,74 @@
...
+kind: Secret
+metadata:
+  name: <secret-name-for-client-password>
+type: Opaque
+data:
+  certificate: CERTIFICATE
+  password: PASSWORD
+```
+
+The Identity can be added to an `AzureCluster` by using `IdentityRef` field
```
CecileRobertMichon, line 48:
```suggestion
The Identity can be added to an `AzureCluster` by using `IdentityRef` field:
```

docs/book/src/topics/multitenancy.md
```diff
@@ -0,0 +1,74 @@
...
+  resourceGroup: example-cluster
+  subscriptionID: <AZURE_SUBSCRIPTION_ID>
+  identityRef:
+    apiVersion: infrastructure.cluster.x-k8s.io/v1alpha3
+    kind: AzureClusterIdentity
+    name: <name-of-identity>
+    namespace: <namespace-of-identity>
+```
+
+for more details on how aad-pod-identity works, please check the guide [here](https://azure.github.io/aad-pod-identity/docs/)
```
CecileRobertMichon, line 70:
```suggestion
For more details on how aad-pod-identity works, please check the guide [here](https://azure.github.io/aad-pod-identity/docs/).
```

controllers/helpers.go
```diff
@@ -387,3 +387,20 @@ func ShouldDeleteIndividualResources(ctx context.Context, clusterScope *scope.Cl
 	// Instead, take the long way and delete all resources one by one.
 	return err != nil || !managed
 }
+
+// GetClusterIdentityFromRef returns the AzureClusterIdentity referenced by the AzureCluster
```
CecileRobertMichon, line 391:
```suggestion
// GetClusterIdentityFromRef returns the AzureClusterIdentity referenced by the AzureCluster.
```

test/e2e/azure_test.go
```diff
@@ -350,4 +351,60 @@ var _ = Describe("Workload cluster creation", func() {
...
+	Context("Creating a cluster using a different SP identity", func() {
+		BeforeEach(func() {
```
CecileRobertMichon, line 356: Same comment I wrote in the Windows PR - we need to audit all the clusters that get created in pr e2e tests and see if any of them can be safely combined to avoid growing the cost and duration of each PR run too much

docs/book/src/topics/multitenancy.md
```diff
@@ -0,0 +1,74 @@
...
+
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: <secret-name-for-client-password>
+type: Opaque
+data:
+  certificate: CERTIFICATE
+  password: PASSWORD
```
CecileRobertMichon, line 45: this is unclear to me? is it one or the other?

docs/book/src/topics/multitenancy.md
```diff
@@ -0,0 +1,74 @@
...
+
+This is achieved using the [aad-pod-identity](https://azure.github.io/aad-pod-identity) library. 
+
+## Service Principal Identity
+
+Once a new SP Identity is created in Azure, the coresponding values should be used to create an `AzureClusterIdentity` resource
+
+```yaml
+apiVersion: infrastructure.cluster.x-k8s.io/v1alpha3
+kind: AzureClusterIdentity
```
CecileRobertMichon, line 13: Should this document AllowedNamespaces?

controllers/azureidentity_controller.go
```diff
@@ -0,0 +1,166 @@
...
+
+	log = log.WithValues("azurecluster", azureCluster.Name)
+
+	// get all the bindings
+	var bindings aadpodv1.AzureIdentityBindingList
+	if err := r.List(ctx, &bindings, client.InNamespace(infrav1.ControllerNamespace)); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	bindingsToDelete := []aadpodv1.AzureIdentityBinding{}
```
CecileRobertMichon, line 119: where are the bindings created?

cloud/scope/identity.go
```diff
@@ -0,0 +1,162 @@
...
+	if err != nil {
+		return nil, err
+	}
+	copiedIdentity := &aadpodv1.AzureIdentity{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "AzureIdentity",
+			APIVersion: "aadpodidentity.k8s.io/v1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      identity.GetAzureIdentityName(p.AzureCluster.Name, p.AzureCluster.Namespace, p.Identity.Name),
```
CecileRobertMichon, line 85: if this has the cluster name in it, its lifecycle is owned by the life of the cluster, no? Why not create/delete it when the an AzureCluster is created/deleted, similarly to how we create/delete the azure.json secret?

controllers/azureidentity_controller.go
CecileRobertMichon: nevermind found it in scope. Seems a bit weird to me to not create the bindings/azureidentity in here too. Also why are we expecting unused bindings to be left around needing cleanup? Why not delete them as part of AzureCluster deletion?

test/e2e/azure_test.go
nader-ziada: I think it makes sense for this test to have its own cluster since it using a different identity which we would want to test separately, wdyt?

docs/book/src/topics/multitenancy.md
nader-ziada: both kinds are supported by aad-pod-identity https://github.com/Azure/aad-pod-identity/tree/master/deploy/demo

controllers/azureidentity_controller.go
nader-ziada: The thinking behind this is that this object lives in a different namespace and is only for the purpose of the controller with the nmi pod, technically the cluster/user should not care about it. Also the reason we needed the cleanup controller is that if the user changes the reference to a different identity, the controller will generate a new azureIdentity/Binding and the old one will be left behind

cloud/scope/identity.go
nader-ziada: I just added the cluster name to make it easier to find, but it wasn't there before, its not part of the cluster lifecycle it lives in a different namespace and is for internal use by the controller

api/v1alpha3/azureclusteridentity_types.go
```diff
@@ -0,0 +1,105 @@
...
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1alpha3"
+)
+
+// AzureClusterIdentitySpec defines the parameters that are used to create an AzureIdentity
+type AzureClusterIdentitySpec struct {
+	// UserAssignedMSI or Service Principal
+	Type IdentityType `json:"type"`
```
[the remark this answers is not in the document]
nader-ziada: added a kubebuilder validation ` // +kubebuilder:validation:Enum=None;SystemAssigned;UserAssigned`

wdyt?

CecileRobertMichon:
lgtm

Are you tracking all follow-ups (clusterctl move, pre-existing aad pod identity deploy, etc) in issues in there respective repos?

devigned (approved):
Solid work, @nader-ziada!
lgtm

devigned:
:rocket:
/lgtm
/approve

CecileRobertMichon:
Just remembered still waiting on the proposal to match the implementation

That's why there's a /hold

devigned:
@CecileRobertMichon I promise, promise, promise to update the proposal by eod tomorrow.

CecileRobertMichon:
/hold cancel 

@devigned thanks, github is witness 😉

PR #988, opened 2020-10-09
💚 cluster should have healthy time synchronization

### LINKED ISSUE #705: [e2e] cluster should have healthy time synchronization
An e2e test to validate machines in the workload cluster have healthy time synchronization

test from aks-engine: https://github.com/Azure/aks-engine/blob/master/test/e2e/kubernetes/kubernetes_test.go#L2207

also related to this PR in image-builder: https://github.com/kubernetes-sigs/image-builder/pull/240

mboersma:

Tests each workload cluster node created in e2e tests to see if it has healthy time synchronization, as defined by the output of these commands on the host:

```bash
$ systemctl is-active chronyd
$ chronyc tracking
```

Fixes #705

This reuses the SSH proxy code from #976. ~~I'll refactor that one to put those funcs in the same place in `helpers.go`, but the two PRs will probably conflict such that one will need to be rebased, just FYI.~~

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
💚 cluster should have healthy time synchronization
```


mboersma:
Looks good:

```
...
STEP: verifying EnableAcceleratedNetworking for the primary NIC of each VM
STEP: checking that time synchronization is healthy on capz-e2e-xcr8ld-control-plane-82jxd
STEP: checking that time synchronization is healthy on capz-e2e-xcr8ld-control-plane-9qnq9
STEP: checking that time synchronization is healthy on capz-e2e-xcr8ld-control-plane-dr2dm
STEP: checking that time synchronization is healthy on capz-e2e-xcr8ld-md-0-68c44c47fc-kdcsk
STEP: checking that time synchronization is healthy on capz-e2e-xcr8ld-md-0-68c44c47fc-rp97d
STEP: Dumping all the Cluster API resources in the "create-workload-cluster-k4sytl" namespace
...
```

test/e2e/azure_test.go
```diff
@@ -161,6 +161,16 @@ var _ = Describe("Workload cluster creation", func() {
 					}
 				})
 			})
+
+			Context("Validating time synchronization", func() {
```
CecileRobertMichon, line 165: should we run this on the other two specs as well?

mboersma: > should we run this on the other two specs as well?

I don't think it adds much time, so it's probably worthwhile to add to the other two specs since "healthy time sync" is basically a requirement for sane Kubernetes.

nader-ziada: doesn't it passing in one test an indication that it should pass in all tests?

go.mod
```diff
@@ -30,6 +30,7 @@ require (
 	k8s.io/utils v0.0.0-20200912215256-4140de9c8800
 	sigs.k8s.io/cluster-api v0.3.10
 	sigs.k8s.io/controller-runtime v0.5.11
+	sigs.k8s.io/kind v0.7.1-0.20200303021537-981bd80d3802
```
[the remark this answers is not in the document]
mboersma: I borrowed the `AggregateConcurrent()` code from the CAPI e2e framework, and this is what go vendored in. It's the same as CAPI's go.mod BTW.

I can try bump the version, but we're just using a helper function that looks unrelated to core `kind` functionality. (Also there is version 0.9.0 now.) And we still say "install kind 0.7.0" in development.md, so should that be bumped as well?

test/e2e/azure_test.go
mboersma: Probably, but not necessarily. If a different cluster has different creation parameters, it's possible that could break something that affects `chrony`, or create a networking issue that affects time sync. And if there's some "once in a hundred VMs" bug relating to time sync, then it would be good to test every one we create.

Having said that, I agree it could yield diminishing returns. I think it's worth adding to the other specs, but I don't feel strongly about it.

go.mod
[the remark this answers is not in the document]
mboersma: @nader-ziada I agree that just the semver release is better. But I tried vendoring in `0.8.0`, `0.8.1`, and `0.9.0` and it appears that creates an incompatibility with client-go:

```
 => [builder 7/7] RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64     go build -a -ldflags '-extldflags "-static"'     -o manage  241.8s 
 => => # # k8s.io/client-go/tools/clientcmd                                                                                        
 => => # /go/pkg/mod/k8s.io/client-go@v0.17.12/tools/clientcmd/validation.go:73:5: cannot use errConfigurationInvalid literal (typ 
 => => # e errConfigurationInvalid) as type "k8s.io/apimachinery/pkg/util/errors".Aggregate in assignment:                         
 => => # errConfigurationInvalid does not implement "k8s.io/apimachinery/pkg/util/errors".Aggregate (missing Is method)   
```

That's a bit scary, but perhaps also an argument for keeping CAPZ's vendored version the same as CAPI's.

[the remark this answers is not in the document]
nader-ziada: sgtm

mboersma:
I'll rebase this after #976 merges. I also have a change to run all the SSH sessions for the whole cluster concurrently, rather than by node, which should make this marginally faster.

test/e2e/helpers.go
```diff
@@ -292,7 +310,7 @@ func execOnHost(controlPlaneEndpoint, hostname string, f io.StringWriter, comman
 	if err != nil {
 		return err
 	}
-	port := "22" // Need to use port 50001 for VMSS when MachinePools are supported here.
+	port := e2eConfig.GetVariable("VM_SSH_PORT") // Or VMMS_SSH_PORT when MachinePools are supported
```
mboersma, line 313: Maybe I should use a constant here instead of the string itself--I see we do that elsewhere with e2eConfig items.

CecileRobertMichon, line 313:
```suggestion
	port := e2eConfig.GetVariable("VM_SSH_PORT") // Or VMSS_SSH_PORT when MachinePools are supported
```

test/e2e/helpers.go
```diff
@@ -292,7 +310,7 @@ func execOnHost(controlPlaneEndpoint, hostname string, f io.StringWriter, comman
 	if err != nil {
 		return err
 	}
-	port := "22" // Need to use port 50001 for VMSS when MachinePools are supported here.
+	port := e2eConfig.GetVariable(VMSSHPort) // Or VMSSSSHPort when MachinePools are supported
```
CecileRobertMichon, line 313: in the future, should this be passed in as a parameter?

test/e2e/azure_timesync.go
```diff
@@ -0,0 +1,92 @@
...
+				if !strings.Contains(f.String(), expected) {
+					return fmt.Errorf("expected \"%s\" in command output:\n%s", expected, f.String())
+				}
+				return nil
+			}
+		}
+
+		testfuncs = append(testfuncs,
+			execToStringFn(
+				"active",
```
CecileRobertMichon, line 81: just to make sure... is it possible that "Inactive" would return a false positive because it contains "active"?

mboersma: Yes, entirely possible. 😶  I would rather have this check the exit code than grep for a substring anyway--let me see if I can improve the commands to be more robust.

test/e2e/helpers.go
mboersma: Yes, it obviously should. I can make that change before we merge this.

mboersma: Done.

test/e2e/azure_timesync.go
mboersma: Done.

test/e2e/azure_timesync.go
```diff
@@ -0,0 +1,94 @@
...
+				if !strings.Contains(f.String(), expected) {
+					return fmt.Errorf("expected \"%s\" in command output:\n%s", expected, f.String())
+				}
+				return nil
+			}
+		}
+
+		testfuncs = append(testfuncs,
+			execToStringFn(
+				"✓ chronyd is active",
```
CecileRobertMichon, line 82: is the `✓` required / reliable? wondering if it might lead to weird behavior

test/e2e/azure_timesync.go
```diff
@@ -0,0 +1,94 @@
...
+				"systemctl", "is-active", "chronyd", "&&",
+				"echo", "✓ chronyd is active",
```
CecileRobertMichon, line 84: why the echo here?

mboersma: Since we can't check the `systemctl is-active` exit code directly, this `&& echo <success text>` should accomplish the same thing.

test/e2e/azure_timesync.go
mboersma: It's just a unicode check mark character. I put it in the "success" text to make it obviously distinct from any text that `systemctl is-active` might print.

test/e2e/azure_timesync.go
CecileRobertMichon: Ahh I see makes sense

test/e2e/helpers.go
```diff
@@ -284,15 +285,31 @@ func logCheckpoint(specTimes map[string]time.Time) {
 	}
 }
 
+// getMachinesInCluster returns a list of all machines in the given cluster.
+// This is copied from CAPI's test/framework/cluster_proxy.go.
+func getMachinesInCluster(ctx context.Context, c framework.Lister, namespace, name string) (*clusterv1.MachineList, error) {
```
CecileRobertMichon, line 290: what about machine pools? Do we want to add that here too @devigned ?

devigned: Probably. There should now be similar funcs for MachinePools.

mboersma: I added support here for MachinePools and created a helper function to collect all the SSH info for the nodes in a cluster, whether they are control plane or agent, VM or VMSS.

One thing I'm curious about is that we anticipated needing port 50001 for VMSS instances, but in practice it seems to be good old port 22 everywhere...

CecileRobertMichon: maybe port 50001 was an AKS Engine thing? Although I would expect the VMSS node pools to go through port 22 for the first control plane ssh but only control plane VMSS nodes (not supported currently) to require a different port.

mboersma:
This test passed on the VMSS cluster as well as the VM-based clusters:

```
Workload cluster creation Creating a VMSS cluster 
  with a single control plane node and an AzureMachinePool with 2 nodes
  /home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/test/e2e/azure_test.go:234
INFO: "with a single control plane node and an AzureMachinePool with 2 nodes" started at Wed, 28 Oct 2020 
...
INFO: Waiting for the machine deployments to be provisioned
INFO: Waiting for the machine pools to be provisioned
STEP: waiting for the machine pool workload nodes to exist
STEP: checking that time synchronization is healthy on capz-e2e-va72nj-control-plane-5fbtc
STEP: checking that time synchronization is healthy on capz-e2e-va72nj-mp-0000002
STEP: checking that time synchronization is healthy on capz-e2e-va72nj-mp-0000003
...
```

Issue #1001, opened 2020-10-16
Clean up ./cloud/services 


devigned:
/kind cleanup

**Describe the solution you'd like**
A while back, the code used to interact with Azure under `./cloud/services` was interfaced to allow some separation of concern between the controller / scopes and the types from the Azure SDK. With the advent of the `azure.Service` interface, Azure SDK / client code doesn't leak into the controllers. In fact, there should not longer be dependencies on the code that directly interacts with Azure. We should use this as an opportunity to crystallize this design goal of refactoring to `azure.Service` by cleaning up the service internals.

Current structure:
```
cloud/services/scalesets/
├── client.go
├── mock_scalesets
│   ├── client_mock.go
│   ├── doc.go
│   └── scalesets_mock.go
├── service.go
├── vmss.go
└── vmss_test.go
```

Proposed structure:
```
cloud/services/scalesets/
├── client.go  <-- don't export this; make it private
├── mock_scalesets
│   ├── client_mock.go
│   ├── doc.go
│   └── scaleset_mock.go
├── scaleset.go <-- includes service and scope
└── scaleset_test.go
```
The only thing exported from the new structure is the `Service` struct and the `Scope` interface for each service. This will ensure that the encapsulation will not be violated and guide future contributors.

devigned:
@shysank please don't hesitate to ping me if you have any questions. Might be best to refactor one service in a PR before refactoring all at the same time.

alexeldeib:
any thoughts on how to replace code like https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/e92bba9b95d18578e09a4eaad1c1150c0c6f5516/exp/controllers/azuremangedcontrolplane_reconciler.go#L219-L263 ?

notably, I don't see many other reconcilers which write data back into the CRD?

alexeldeib:
that blurb is long -- referencing `managedClustersSvc.Get` and `managedClustersSvc.GetCredentials` to reconcile endpoint/kubeconfig

alexeldeib:
I see: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/e92bba9b95d18578e09a4eaad1c1150c0c6f5516/cloud/services/virtualmachines/virtualmachines.go#L69

devigned:
[...]

Issue #611, opened 2020-05-13
[managedcluster] e2e tests


alexeldeib:
/kind feature

**Describe the solution you'd like**
We should add e2e tests similar to the existing tests for Azure, particularly being sure to validate the options we provide to AKS match the result we get.

alexeldeib:
couple observations from spiking on this.

The capi test framework is a bit tightly coupled. The machine portion is fairly well abstracted but the control plane tests hardcode a KubeadmControlPlane struct as part of the test context [0], [1]. Besides the fact that it uses KCP directly instead of the CAPI control plane contract, it also relies on a control plane provider using machines for the underlying implementation, which is not necessarily true. A control plane could be pod based or in the case of AKS/EKS/GKE fully managed/externalized. 

It would be nice to decouple the pieces of the capi e2e framework to allow it to run against arbitrary control plane implementations like used for AzureManagedControlPlane. That seems like it might take a good chunk of work. A shorter term approach could be a simple pipeline which does the following:
- create AzureManagedCluster with clusterctl
- wait for all objects to reach status.ready = true
- validate control plane connectivity, ensure all nodes ready
- delete crds, ensure all objects get deleted (maybe verify against azure CLI at the end or something?)

[0]: https://github.com/kubernetes-sigs/cluster-api/blob/72ee11af1bec2e9dcd3fb22cba3f718839658d89/test/framework/clusterctl/clusterctl_helpers.go#L240
[1]: https://github.com/kubernetes-sigs/cluster-api/blob/master/test/framework/controlplane_helpers.go#L233

alexeldeib:
this should be unblocked thanks to @shysank's upstream work, at least using v0.4.0-beta.0 or later of CAPI 🎉

CecileRobertMichon:
@shysank is actually working on this in https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1386

alexeldeib:
/assign @shysank 

We can definitely add more configurations on top of the work in #1386, but marking as assigned since I feel that's enough to close the general issue for adding e2e. Let's open specific issues for desired testing scenarios after that gets merged.

nader-ziada:
/close 

fixed with e2e added as part of #1386 🚀

PR #1009, opened 2020-10-22
💎 provision vnet for aks clusters

alexeldeib:
**What type of PR is this?**

/kind api-change
/kind cleanup

Fixes #

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
AKS clusters provisioned via CAPZ now have predictably named virtual networks. Prior to this, virtual networks for AKS clusters were not predictably named. This also means backfilling the corresponding cluster specs is impossible. Action required: upgrading a cluster to this version requires manually updating `spec.virtualNetwork.Name` and `spec.virtualNetwork.Subnet.Name` to the values generated by AKS. Otherwise the cluster will fail to reconcile as the network and subnet names will not match.
```


alexeldeib:
notable:
```
sigs.k8s.io/cluster-api-provider-azure/exp/api/v1alpha3
  Incompatible changes:
  - (*AzureManagedControlPlane).SetDefaultSSHPublicKey: removed
  - (*AzureManagedControlPlane).ValidateSSHKey: removed
  - AzureManagedControlPlaneSpec.ResourceGroup: removed
  Compatible changes:
  - AzureManagedControlPlaneSpec.NodeResourceGroupName: added
  - AzureManagedControlPlaneSpec.ResourceGroupName: added
  - AzureManagedControlPlaneSpec.VirtualNetwork: added
  - ManagedControlPlaneSubnet: added
  - ManagedControlPlaneVirtualNetwork: added
```

CecileRobertMichon:
> This is pretty breaking...but I'm going to vote we iterate while in exp rather than avoid breaking users (i'm skeptical this has many users yet).

+1

I haven't code reviewed in depth yet but overall these changes are sane to me

alexeldeib:
n.b.: we only deploy aks with system msi right now, afaik this avoids several permissions issues with AKS service principals // user assigned managed identities. In the future, we may need to either create appropriate rbac assignments for the cluster service principal or update docs to reflect required permissions: https://docs.microsoft.com/en-us/azure/aks/kubernetes-service-principal#networking

cloud/services/virtualnetworks/service.go
```diff
@@ -18,14 +18,15 @@ package virtualnetworks
...
+	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1alpha3"
 	azure "sigs.k8s.io/cluster-api-provider-azure/cloud"
 )
 
 // VNetScope defines the scope interface for a virtual network service.
 type VNetScope interface {
 	logr.Logger
 	azure.ClusterDescriber
-	azure.NetworkDescriber
+	Vnet() *infrav1.VnetSpec
```
CecileRobertMichon, line 29: I think this highlights a potential bug in the code... the vnet service is supposed to be unaware of the AzureCluster spec and just create / update / delete vnets from the specs it gets. The fact that it needs `Vnet()` which doesn't have any info that `VNetSpecs()` doesn't have is a little strange. Looking at the code, it uses that to write the info back into the Vnet() which makes sense but also would cause a bug if we have decided to have multiple specs returned by VnetSpecs() since they would overwrite each other. Not specific to this PR, in fact this PR makes it easy to see that bug, just reflecting 🤔

exp/controllers/azuremangedcontrolplane_reconciler.go
```diff
@@ -36,6 +36,8 @@ import (
 	"sigs.k8s.io/cluster-api-provider-azure/cloud/scope"
 	"sigs.k8s.io/cluster-api-provider-azure/cloud/services/groups"
 	"sigs.k8s.io/cluster-api-provider-azure/cloud/services/managedclusters"
+	"sigs.k8s.io/cluster-api-provider-azure/cloud/services/subnets"
+	"sigs.k8s.io/cluster-api-provider-azure/cloud/services/virtualnetworks"
```
CecileRobertMichon, line 40: let's add an e2e test for managed control plane clusters as part of moving it out of exp/ (can be optional on PRs). Is there already an issue tracking that?

alexeldeib: yeah, I should probably do this sooner rather than later

alexeldeib: https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/611

alexeldeib:
small bump for review if anyone has some cycles 🙂

CecileRobertMichon (approved):
/lgtm
/approve

cloud/services/virtualnetworks/service.go
CecileRobertMichon: this is another case where it would make more sense if vnetSpec() didn't return an array @alexeldeib

alexeldeib: https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1001#issuecomment-718980544

Issue #982, opened 2020-10-07
Disk Encryption


mjudeikis:
***Goals***
1. Enable OSDisk and ephemeral disk encryption for Azure VMS

***Non-Goals/Future Work***
1. To enable manage Service or user keys. Encryption key management is left for the user.

**User Story**

As a User, I would like to encrypt all the disk for Azure vm/vmss for security reasons.

**Detailed Description**

Azure supports OSdisk encryption via:
`OSDisk.ManagedDisk.DiskEncryptionSet.ID` and with new compute API Host encryption too
https://docs.microsoft.com/en-us/azure/virtual-machines/disks-enable-host-based-encryption-portal 

**Contract changes [optional]**

Don't know

**Data model changes [optional]**

1. Add `ManagedDisk` option for managed disk encryption:
```
type ManagedDisk struct {
	StorageAccountType string `json:"storageAccountType"`
	StorageAccountType string                       `json:"storageAccountType"`
	DiskEncryptionSet  *DiskEncryptionSetParameters `json:"diskEncryptionSet,omitempty"`
}

// DiskEncryptionSetParameters defines disk encryption options
type DiskEncryptionSetParameters struct {
	ID string `json:"id,omitempty"`
}
```

2. Add `SecurityProfile into VM main spec (to mimic SDK configuration). Security profile adding would allow extension later on with other security-related features.
```
// AzureMachineSpec defines the desired state of AzureMachine
type AzureMachineSpec struct {
	// ProviderID is the unique identifier as specified by the cloud provider.
	// +optional
	ProviderID *string `json:"providerID,omitempty"`

	VMSize           string           `json:"vmSize"`
	......

      // SecurityProfile specifies the Security profile settings for the virtual machine or virtual machine scale
      // set.
     type SecurityProfile struct {
	  // EncryptionAtHost - This property can be used by user in the request to enable or disable the Host Encryption for the virtual machine or virtual machine scale set. This will enable the encryption for all the disks including Resource/Temp disk 
 at host itself. <br><br> Default: The Encryption at host will be disabled unless this property is set to true for the resource.
	EncryptionAtHost *bool `json:"encryptionAtHost,omitempty"`
     }

}
```

/kind proposal

CecileRobertMichon:
@mjudeikis are you planning to work on the second part or should we mark this as help wanted?

mjudeikis:
[...]

PR #1012, opened 2020-10-27
Add SecurityProfile.EncryptionAtHost parameter to enable host-based VM encryption

dkorzuno:
**What type of PR is this?**

/kind feature
/kind api-change

The PR adds a parameter which enables [encryption at host](https://docs.microsoft.com/en-us/azure/virtual-machines/linux/disk-encryption#encryption-at-host---end-to-end-encryption-for-your-vm-data) for virtual machines.

This PR addresses the second part of #982 started by @mjudeikis 

I added what seemed reasonable to me, but I'm not entirely sure that I haven't missed anything. Especially I have some doubts about the conversion part. Will be happy to make whatever changes required.

Also, I added a couple of the unit test, in somewhat "copy-pasty" manner. While they seem to follow the suit and do cover the change, maybe I should clean those up a bit to exclude extra checks?

- [+] squashed commits
- [-] includes documentation
- [+] adds unit tests

**Release note**:

```release-note
1. Add SecurityProfile.EncryptionAtHost parameter to machine spec to enable host-based VM encryption.
```


dkorzuno:
Followed the CLA instructions at at https://git.k8s.io/community/CLA.md#the-contributor-license-agreement.

cloud/services/virtualmachines/virtualmachines.go
```diff
@@ -396,3 +397,13 @@ func getSpotVMOptions(spotVMOptions *infrav1.SpotVMOptions) (compute.VirtualMach
...
 	return compute.Spot, compute.Deallocate, billingProfile, nil
 }
+
+func getSecurityProfile(vmSpec azure.VMSpec) *compute.SecurityProfile {
+	if vmSpec.SecurityProfile == nil {
+		return nil
+	}
+
+	return &compute.SecurityProfile{
+		EncryptionAtHost: to.BoolPtr(*vmSpec.SecurityProfile.EncryptionAtHost),
```
alexeldeib, line 407: this isn't supported on all VM sizes, it requires a check against the capability (see accelerated networking for an example). 

we should probably fail fast with error + no requeue in the reconciler if it's not possible to provision that for the user.

cloud/services/scalesets/vmss.go
```diff
@@ -339,3 +340,13 @@ func getVMSSUpdateFromVMSS(vmss compute.VirtualMachineScaleSet) (compute.Virtual
...
 	return update, err
 }
+
+func getSecurityProfile(scaleSpec azure.ScaleSetSpec) *compute.SecurityProfile {
+	if scaleSpec.SecurityProfile == nil {
+		return nil
+	}
+
+	return &compute.SecurityProfile{
+		EncryptionAtHost: to.BoolPtr(*scaleSpec.SecurityProfile.EncryptionAtHost),
```
alexeldeib, line 350: same comment as https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1012/files#r512541890

cloud/services/virtualmachines/virtualmachines.go
dkorzuno: I added the capability check.

cloud/services/scalesets/vmss.go
dkorzuno: Added the capability check to the function.

devigned (changes requested):
I'm a little bummed we can't reject this in the webhook rather than in reconciliation.

Actually, I think returning an error on this is going to cause an endless reconcile loop. We will never get to a terminal state unless the user changes the VM SKU or disables `EncryptionAtHost`.

@alexeldeib wdyt?

alexeldeib:
> I'm a little bummed we can't reject this in the webhook rather than in reconciliation.

Agreed 🙁 

> Actually, I think returning an error on this is going to cause an endless reconcile loop.

IMO, we need a way to bubble terminal vs non-terminal errors to the top. When you write one big reconcile loop, it's super easy -- log the error and either return if non-terminal or don't return the error. We could use a good way to bubble that up from inside the reconcilers (I smell custom error types).

devigned (approved):
With #1014 logged, I'm good with merging this. 

lgtm

I'm going to leave it open for a bit to allow others to review. Thank you for the PR!

cloud/services/scalesets/vmss_test.go
```diff
@@ -899,6 +899,185 @@ func TestReconcileVMSS(t *testing.T) {
...
+				s.GetVMImage().Return(&infrav1.Image{
+					Marketplace: &infrav1.AzureMarketplaceImage{
+						Publisher: "fake-publisher",
+						Offer:     "my-offer",
+						SKU:       "sku-id",
+						Version:   "1.0",
+					},
+				}, nil)
+				s.GetBootstrapData(context.TODO()).Return("fake-bootstrap-data", nil)
+				m.CreateOrUpdate(context.TODO(), "my-rg", "my-vmss", gomockinternal.DiffEq(compute.VirtualMachineScaleSet{
```
CecileRobertMichon, line 944: instead of checking the entire struct here since we only care about testing the encryption at host functionality is enabled it would make sense to do the same thing you did in the virtualmachines test https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1012/files#diff-f2a0391546f379219695118e52694fe6a8c5739619cb672941617a78d0c63336R770

api/v1alpha3/types.go
```diff
@@ -426,3 +426,12 @@ func (n *NetworkSpec) GetNodeSubnet() *SubnetSpec {
...
 	return nil
 }
+
+// SecurityProfile specifies the Security profile settings for a
+// virtual machine or virtual machine scale set.
+type SecurityProfile struct {
+	// This field indicates whether Host Encryption should be enabled
+	// or disabled for a virtual machine or virtual machine scale
+	// set. Default is disabled.
+	EncryptionAtHost *bool `json:"encryptionAtHost,omitempty"`
```
CecileRobertMichon, line 436: Do we expect this field to be changed after initial cluster creation or is it immutable? If it's the later, it would be good to add webhook validation to make sure it doesn't get updated

dkorzuno:
> > Actually, I think returning an error on this is going to cause an endless reconcile loop.
> 
> IMO, we need a way to bubble terminal vs non-terminal errors to the top. When you write one big reconcile loop, it's super easy -- log the error and either return if non-terminal or don't return the error if it's terminal. We could use a good way to bubble that up from inside the reconcilers (I smell custom error types).

A random thought - at least for `virtualmachine` the error handling can be improved a bit: 

  - instead of sequential processing of each `vmSpec` start a goroutine which creates a machine independently
  - `Reconcile` will wait for all goroutines to finish and will log errors if any

This way an error in one VM specification won't affect others. On next `Reconcile` the successfully created machines will be skipped, but the failed ones will be retried.

Something along the lines
```
ch := make(chan error)
for _, vmSpec := range s.Scope.VMSpecs() {
   go func(spec azure.VMSpec) error {
           ch <- ReconcileOne(ctx, vmSpec)
   } (vmSpec)
}
for i :=0; i< len(s.Scope.VMSpecs()); i++ {
    if err <- ch; err != nil {
         // handle error
     }
}
```

api/v1alpha3/types.go
dkorzuno: The [doc](https://docs.microsoft.com/en-us/azure/virtual-machines/linux/disk-encryption#restrictions-1) says:
> The encryption can be enabled on existing virtual machine scale set. However, only new VMs created after enabling the encryption are automatically encrypted.

Not sure how interpret it. Maybe prohibit modification for `AzureMachine`, but allow for `AzureMachineTemplate`?

CecileRobertMichon: Ok that means the VMSS model can be updated so that's fine for AzureMachinePools 👍  machines are immutable already so you can update a machine deployment but an individual machine update won't trigger a VM update.

CecileRobertMichon:
> instead of sequential processing of each vmSpec start a goroutine which creates a machine independently

⚠️   1 vmSpec != 1 machine 

Each `AzureMachine` has its own Reconcile loop and currently each Machine == 1 VM. So it is serial (and I agree that could be an improvement in theory) but in practice, it's serial on a list of constant size 1. The reason that VMSpecs was originally made a list was to allow for future work that requires more than one VM resource to be provisioned per `AzureMachine` (for example a jumpbox / bastion VM), as well as to be consistent with the other Azure services which allow multiple resources of the same type to be reconciled per AzureMachine object (for example multiple network interfaces attached to the same Machine).

So this:
> This way an error in one VM specification won't affect others. On next Reconcile the successfully created machines will be skipped, but the failed ones will be retried.

should already be the case, and all machines are reconciled in parallel. The main issue is that right now, when reconcile fails, it always triggers a requeue, even if the error isn't one that would fix itself - in other words there is no handling of "terminal" errors.

cloud/services/scalesets/vmss.go
```diff
@@ -356,3 +362,17 @@ func getVMSSUpdateFromVMSS(vmss compute.VirtualMachineScaleSet) (compute.Virtual
 	err = update.UnmarshalJSON(json)
 	return update, err
 }
+
+func getSecurityProfile(vmssSpec azure.ScaleSetSpec, sku resourceskus.SKU) (*compute.SecurityProfile, error) {
+	if vmssSpec.SecurityProfile == nil {
```
alexeldeib, line 367: actually, I think we should default this on when possible? if a user doesn't specify it but the sku supports it, I think we should mutate the spec to default it on.

really wish we could do that in webhook rather than controller, though 🙁 wdyt @CecileRobertMichon @devigned?

alexeldeib:
>The reason that VMSpecs was originally made a list was to allow for future work that requires more than one VM resource to be provisioned per AzureMachine (for example a jumpbox / bastion VM)

This smells fishy to me, that feels like a cluster-level or even machinepool level resource, rather than per-machine resource? I'm struggling to think of a scenario where you'd regularly want multiple VMs per AzureMachine.

cloud/services/scalesets/vmss.go
CecileRobertMichon: That's an interesting idea, we could follow the precedent set by `AcceleratedNetworking`, if the user explicitly sets a value then use that, otherwise set to true if the sku supports it. I don't know if we should mutate the spec though, I don't think accelerated networking does it... but maybe it should? 

This does bring up a point though that if we don't set it by default now and later decide to set it, there won't be a way to know if the value is nil because the user didn't set it or if it wasn't set before and should actually be false... so maybe we should set a default and mutate the spec, even if that default is false.

CecileRobertMichon:
> This smells fishy to me, that feels like a cluster-level or even machinepool level resource, rather than per-machine resource? I'm struggling to think of a scenario where you'd regularly want multiple VMs per AzureMachine. I was a bit surprised looking back at the code that VMSpecs is an array

Yeah arguably it's not a good reason, at the time it seemed easier to be consistent everywhere. I think VM and vnet (and scale set but that one is still not refactored) are probably the only ones that are exceptions to the rule? I'd be open to switching those to not be arrays. It's not an API change shouldn't be breaking for users.

cloud/services/scalesets/vmss.go
alexeldeib: actually, the way we do this today is broken: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/110f084337f64c49a13b01744a2e6d03dba55d13/cloud/services/networkinterfaces/networkinterfaces.go#L93

I don't think that writes back into anything which gets persisted -- it's writing back into the value from a range array, rather than writing back into the slice returned by the function. I think because the original value passed to the struct from the scope was a pointer to AcceleratedNetworking, we could technically get away with this if we ranged references instead of values? but if we try to write any non-pointer values back, that will definitely be broken since the values will have been copied inside NICScopes()

alexeldeib: need to test, though, also definitely out of scope for this PR

CecileRobertMichon: right accelerated networking definitely does not get persisted

alexeldeib: > This does bring up a point though that if we don't set it by default now and later decide to set it, there won't be a way to know if the value is nil because the user didn't set it or if it wasn't set before and should actually be false... so maybe we should set a default and mutate the spec, even if that default is false

agreed, we should not allow persisting objects in final form with ambiguity (phrasing it that way since we would ideally not let them get past the webhook, but we live with that for now)

cloud/services/scalesets/vmss_test.go
dkorzuno: Updated the test... maybe not in the best possible way, though. I had to modify the test case interface a bit to pass through the `*WithT` pointer. Happy to change it if needed.

cloud/services/scalesets/vmss.go
mjudeikis: Im ok with defaulting to on in the long run but maybe not part of this PR? As I think this might need to the testing effort?

mjudeikis:
> I'm a little bummed we can't reject this in the webhook rather than in reconciliation.

Im might be missing something, Is there something stopping us to do this in webhook? There were multiple comments about this. Is this a request or just a rant for that this is technologically not possible? #wannalearn

devigned:
> I'm a little bummed we can't reject this in the webhook rather than in reconciliation.

@mjudeikis, it's the start of a rant. Since the webhooks don't have an authenticated context and we don't make any service calls, we are unable to check to see if the resource sku supports `EncryptionAtHost` for a given subscription and region. It's been a long standing issue that we have punted due to the amount of work to make that happen.

Great question!

cloud/services/scalesets/vmss.go
CecileRobertMichon: So in that case, should it at least get it persisted so that when we change the default later it doesn't change pre-existing machines?

dkorzuno:
Is there anything else I can do in the context of this PR?

CecileRobertMichon:
@dkorzuno to me the only actionable item for this PR is to persist the value in the controller (ie. if nil, set it to false) so that it doesn't get changed later on existing clusters if we change the default to "true if supported". @alexeldeib @devigned does that sound reasonable to you?

dkorzuno:
@CecileRobertMichon let me clarify it as I got confused a bit. 
Do you mean that the `getSecurityProfile` function should return an allocated `SecurityProfile` with `EncryptionAtHost` set to `false`, as in
``` go
func getSecurityProfile(vmssSpec azure.ScaleSetSpec, sku resourceskus.SKU) (*compute.SecurityProfile, error) {
	if vmssSpec.SecurityProfile == nil {
		return &compute.SecurityProfile{
		      EncryptionAtHost: to.BoolPtr(false),
	        }, nil
	}

	if !sku.HasCapability(resourceskus.EncryptionAtHost) {
		return nil, errors.Errorf("encryption at host is not supported for VM type %s", vmssSpec.Size)
	}

	return &compute.SecurityProfile{
		EncryptionAtHost: to.BoolPtr(*vmssSpec.SecurityProfile.EncryptionAtHost),
	}, nil
}
```

Or (which I'm leaning towards to) if `getSecurityProfile` notices that `SecurityProfile` in the spec is `nil` then it should write back the spec to Etcd with `SecurityProfile` field pre-populated?

If it's the latter, what would be the best approach there - add `UpdateMachineSpec` method to `ScaleSetScope` interface and implement it for `MachinePoolScope`?

alexeldeib:
I'm sure Cecile will comment too, but personally I think leave the field as an optional pointer, but default it to false in the webhook?

>Or (which I'm leaning towards to) if getSecurityProfile notices that SecurityProfile in the spec is nil then it should write back the spec to Etcd with SecurityProfile field pre-populated

so basically this, but in the webhook for now should be okay I think?

in the future, we would want to be able to dynamically detect whether a size supports it without forcing the user to tell us. to do that would still require what you describe, and yes it would require some kind of setter on the scope to write back into the CRD object.

dkorzuno:
> so basically this, but in the webhook for now should be okay I think?

Done.

CecileRobertMichon:
E2E is failing with 

```
E1030 04:05:18.264096       1 controller.go:257] controller-runtime/controller "msg"="Reconciler error" "error"="failed to reconcile AzureMachine: failed to create virtual machine: failed to create VM capz-e2e-oo3dfl-control-plane-p6tmv in resource group capz-e2e-oo3dfl: compute.VirtualMachinesClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code=\"InvalidParameter\" Message=\"The property 'securityProfile.encryptionAtHost' is not valid because the 'Microsoft.Compute/EncryptionAtHost' feature is not enabled for this subscription.\" Target=\"securityProfile.encryptionAtHost\"" "controller"="azuremachine" "name"="capz-e2e-oo3dfl-control-plane-p6tmv" "namespace"="create-workload-cluster-ao4au3"
```

🤨  

Is EncryptionAtHost a gated feature? Also strange that it's for returning an error message when the value is `false`...

CecileRobertMichon:
Confirmed docs mention you have to send an email to get it activated on your sub https://docs.microsoft.com/en-us/azure/virtual-machines/linux/disks-enable-host-based-encryption-cli#prerequisites so we can't do the defaulting otherwise that will break all users that haven't enabled it 😞 

@dkorzuno let's revert the defaulting, sorry for making you do that extra work for nothing...

CecileRobertMichon (approved):
/lgtm

PR #1025, opened 2020-11-05
💎 remove AzureCluster from clusterscope

alexeldeib:
**What type of PR is this?**
/kind cleanup

starts to tackle https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/825

will follow up to move a bunch more fields from ManagedControlPlane scope to CRD, but this seemed like a viable checkpoint. 

I had to move defaults.go into a separate package to avoid an import loop. That's a big reason why the diff is so big. 

I also needed to break apart authorizer + ClusterScoper into multiple interfaces to avoid ambiguous references, or else I would've needed to refactor even more to avoid using embedded fields/methods.

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```release-note
NONE
```

### DIFF
--- a/controllers/azuremachine_controller.go
+++ b/controllers/azuremachine_controller.go
@@ -180,10 +180,10 @@ func (r *AzureMachineReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, ret
 
 	// Create the cluster scope
 	clusterScope, err := scope.NewClusterScope(scope.ClusterScopeParams{
-		Client:       r.Client,
-		Logger:       logger,
-		Cluster:      cluster,
-		AzureCluster: azureCluster,
+		Client:        r.Client,
+		Logger:        logger,
+		Cluster:       cluster,
+		ClusterScoper: azureCluster,
 	})
 	if err != nil {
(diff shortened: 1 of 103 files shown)


controllers/azuremachine_controller.go
```diff
@@ -253,7 +253,7 @@ func (r *AzureMachineReconciler) reconcileNormal(ctx context.Context, machineSco
 	if machineScope.AzureMachine.Spec.AvailabilityZone.ID != nil {
 		message := "AvailabilityZone is deprecated, use FailureDomain instead"
 		machineScope.Info(message)
-		r.Recorder.Eventf(clusterScope.AzureCluster, corev1.EventTypeWarning, "DeprecatedField", message)
+		r.Recorder.Eventf(machineScope.AzureMachine, corev1.EventTypeWarning, "DeprecatedField", message)
```
alexeldeib, line 256: n.b.: this is a bug

alexeldeib:
@CecileRobertMichon I swear this diff isn't as bad as the size makes it look 🙂

alexeldeib:
/close

at this point, if I want to get this merged, i'll have to take another attempt anyway.

PR #1049, opened 2020-11-19
Remove failed provisioning VM deletion logic

### LINKED ISSUE #1037: VM deleted if it is in Failed state
/kind bug

There is a piece of [code](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/cb486c3d7b26dc84ad5156fa97773ccb97578ebe/controllers/azuremachine_controller.go#L299) in azure machine reconciliation which deletes the VM if it's in a Failed state. The expectation is that it'll create a new VM in the next reconcile run. But with FailureReason being set [before](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/cb486c3d7b26dc84ad5156fa97773ccb97578ebe/controllers/azuremachine_controller.go#L295) that, this will not happen and will return early in the next run. 

The vm should not be deleted, and should be left to MHC to remediate the machine.

[Slack Thread](https://kubernetes.slack.com/archives/CEX9HENG7/p1605300482244500)

**Environment:**

- cluster-api-provider-azure version:  master branch

Evalle:
**What type of PR is this?**
/kind cleanup

Removes the vm deletion logic. 

> There is a piece of code in azure machine reconciliation which deletes the VM if it's in a Failed state. The expectation is that it'll create a new VM in the next reconcile run. But with FailureReason being set before that, this will not happen and will return early in the next run.

Fixes #1037

**Release note**:
```release-note
Remove failed provisioning VM deletion logic
```

### DIFF
--- a/controllers/azuremachine_controller.go
+++ b/controllers/azuremachine_controller.go
@@ -309,12 +309,6 @@ func (r *AzureMachineReconciler) reconcileNormal(ctx context.Context, machineSco
 		machineScope.SetFailureMessage(errors.Errorf("Azure VM state is %s", machineScope.VMState()))
 		conditions.MarkFalse(machineScope.AzureMachine, infrav1.VMRunningCondition, infrav1.VMProvisionFailedReason, clusterv1.ConditionSeverityWarning, "")
 		machineScope.SetNotReady()
-		// If VM failed provisioning, delete it so it can be recreated
-		err := ams.DeleteVM(ctx)
-		if err != nil {
-			return reconcile.Result{}, errors.Wrapf(err, "failed to delete VM in a failed state")
-		}
-		return reconcile.Result{}, errors.Wrapf(err, "VM deleted, retry creating in next reconcile")
 	default:
 		machineScope.V(2).Info("VM state is undefined", "id", machineScope.GetVMID())
 		conditions.MarkUnknown(machineScope.AzureMachine, infrav1.VMRunningCondition, "", "")
--- a/controllers/azuremachine_reconciler.go
+++ b/controllers/azuremachine_reconciler.go
@@ -18,6 +18,7 @@ package controllers
 
 import (
 	"context"
+
 	"sigs.k8s.io/cluster-api-provider-azure/cloud/services/tags"
 	"sigs.k8s.io/cluster-api-provider-azure/util/tele"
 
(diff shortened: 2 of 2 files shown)


Evalle:
FYI @shysank @CecileRobertMichon

controllers/azuremachine_controller.go
```diff
@@ -309,12 +309,6 @@ func (r *AzureMachineReconciler) reconcileNormal(ctx context.Context, machineSco
 		machineScope.SetFailureMessage(errors.Errorf("Azure VM state is %s", machineScope.VMState()))
 		conditions.MarkFalse(machineScope.AzureMachine, infrav1.VMRunningCondition, infrav1.VMProvisionFailedReason, clusterv1.ConditionSeverityWarning, "")
 		machineScope.SetNotReady()
-		// If VM failed provisioning, delete it so it can be recreated
-		err := ams.DeleteVM(ctx)
```
CecileRobertMichon, line 313: the DeleteVM function can also be removed, it shouldn't be used anywhere else

CecileRobertMichon (changes requested):
thanks for the PR @Evalle. Can you please reword the commit to be more specific so it doesn't sound like we're removing _all_ vm deletion logic? Something like `Remove failed provisioning VM deletion logic`.

controllers/azuremachine_controller.go
Evalle: Thanks for the review! Are you talking about [this](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/controllers/azuremachine_reconciler.go#L125) function?

CecileRobertMichon: yes, that one

Evalle: Thank you, done.

CecileRobertMichon:
@Evalle please squash the commits and amend the commit message to match the PR title

Evalle:
@CecileRobertMichon OK, done.

cloud/services/virtualmachines/virtualmachines_test.go
```diff
@@ -1507,107 +1507,3 @@ func TestReconcileVM(t *testing.T) {
 		})
 	}
 }
-
-func TestDeleteVM(t *testing.T) {
```
CecileRobertMichon, line 1511: we actually want to keep this test, it's testing https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/a6d3f30ebf16891446ca0104926cc41516905b12/cloud/services/virtualmachines/virtualmachines.go#L216, not the same Delete

my bad, did not notice this when reviewing earlier

Evalle: Ah, let me return it back.

CecileRobertMichon (approved):
/lgtm
/approve

PR #1050, opened 2020-11-19
Increase wait for ingress-nginx controller timeout

CecileRobertMichon:
**What type of PR is this?**

**What this PR does / why we need it**: I've had quite a few `make tilt-up` runs fail with  `error: timed out waiting for the condition on pods/ingress-nginx-controller-xxx` recently. When I checked the kind cluster, the ingress pod always became ready a few seconds after. Increasing the timeout to 120s to try and reduce those.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```


devigned (approved):
I was really hoping that 90s would be enough. Perhaps, ingress / tracing should be optional in the future or just rely on port forwarding.

/lgtm
/approve

PR #1053, opened 2020-11-20
Refactor AzureMachineReconciler to inject AzuremachineService as dependency

### LINKED ISSUE #1047: Refactor controller to decouple service instantiaition
/kind cleanup

Right now, in controllers, the reconciliation service is instantiated directly inside the `reconcileNormal` and `reconcileDelete` methods like [this](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/5ce2cc61a6862852cd7ed7229757422c6f6b7dfb/controllers/azuremachine_controller.go#L266). This tight coupling makes it difficult to write tests since we cannot mock the behavior of this service as experienced in #1042 . So, we want to remove this coupling and inject this as a dependency in the reconciler instance. The following are a list of changes that should be done as part of this refactor:
1. Add a factory in reconciler to create service instances
2. Change instantiation logic to use the factory
3. Extract an interface out of the service instance (this will be similar to `azure.Service` interface)

The changes will look something like this:

```
type AzureMachineReconciler struct {
	client.Client
	Log              logr.Logger
	Recorder         record.EventRecorder
	ReconcileTimeout time.Duration
	azureMachineServiceFactory func(*scope.MachineScope, *scope.ClusterScope) azureMachineService
}

// inject it using the factory 
func (r *AzureMachineReconciler) ctx context.Context, machineScope *scope.MachineScope, clusterScope *scope.ClusterScope) (reconcile.Result, error) {
 
ams := r.azureMachineServiceFactory(machineScope, clusterScope)

}

type Service interface {
   func Reconcile(ctx) error
   func Delete(ctx) error

}

``` 

**Environment:**

shysank:
**What type of PR is this?**

/kind cleanup

Refactors AzureMachineReconciler to inject AzuremachineService as dependency
Fixes#1047 (partially)

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Refactor AzureMachineReconciler to inject AzuremachineService as dependency
```

### DIFF
--- a/controllers/azuremachine_controller.go
+++ b/controllers/azuremachine_controller.go
@@ -51,9 +51,26 @@ import (
 // AzureMachineReconciler reconciles a AzureMachine object
 type AzureMachineReconciler struct {
 	client.Client
-	Log              logr.Logger
-	Recorder         record.EventRecorder
-	ReconcileTimeout time.Duration
+	Log                       logr.Logger
+	Recorder                  record.EventRecorder
+	ReconcileTimeout          time.Duration
+	createAzureMachineService azureMachineServiceCreator
+}
+
+type azureMachineServiceCreator func(machineScope *scope.MachineScope, clusterScope *scope.ClusterScope) *azureMachineService
+
+// NewAzureMachineReconciler returns a new AzureMachineReconciler instance
+func NewAzureMachineReconciler(client client.Client, log logr.Logger, recorder record.EventRecorder, reconcileTimeout time.Duration) *AzureMachineReconciler {
+	amr := &AzureMachineReconciler{
+		Client:           client,
+		Log:              log,
+		Recorder:         recorder,
+		ReconcileTimeout: reconcileTimeout,
+	}
+
+	amr.createAzureMachineService = newAzureMachineService
+
+	return amr
 }
 
 // SetupWithManager initializes this controller with a manager.
@@ -263,7 +280,7 @@ func (r *AzureMachineReconciler) reconcileNormal(ctx context.Context, machineSco
 		}
 	}
 
-	ams := newAzureMachineService(machineScope, clusterScope)
+	ams := r.createAzureMachineService(machineScope, clusterScope)
 
 	err := ams.Reconcile(ctx)
 	if err != nil {
@@ -337,7 +354,7 @@ func (r *AzureMachineReconciler) reconcileDelete(ctx context.Context, machineSco
 
 	if ShouldDeleteIndividualResources(ctx, clusterScope) {
 		machineScope.Info("Deleting AzureMachine")
-		if err := newAzureMachineService(machineScope, clusterScope).Delete(ctx); err != nil {
+		if err := r.createAzureMachineService(machineScope, clusterScope).Delete(ctx); err != nil {
 			r.Recorder.Eventf(machineScope.AzureMachine, corev1.EventTypeWarning, "Error deleting AzureMachine", errors.Wrapf(err, "error deleting AzureMachine %s/%s", clusterScope.Namespace(), clusterScope.ClusterName()).Error())
 			conditions.MarkFalse(machineScope.AzureMachine, infrav1.VMRunningCondition, clusterv1.DeletionFailedReason, clusterv1.ConditionSeverityWarning, err.Error())
 			return reconcile.Result{}, errors.Wrapf(err, "error deleting AzureMachine %s/%s", clusterScope.Namespace(), clusterScope.ClusterName())
--- a/controllers/azuremachine_reconciler.go
+++ b/controllers/azuremachine_reconciler.go
@@ -46,6 +46,8 @@ type azureMachineService struct {
 	skuCache             *resourceskus.Cache
 }
 
+var _ azure.Service = (*azureMachineService)(nil)
+
 // newAzureMachineService populates all the services based on input scope.
 func newAzureMachineService(machineScope *scope.MachineScope, clusterScope *scope.ClusterScope) *azureMachineService {
 	cache := resourceskus.NewCache(clusterScope, clusterScope.Location())
--- a/controllers/azuremachine_controller_test.go
+++ b/controllers/azuremachine_controller_test.go
@@ -26,6 +26,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/record"
 	"k8s.io/klog/klogr"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1alpha3"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -35,6 +36,7 @@ import (
 	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1alpha3"
 	"sigs.k8s.io/cluster-api-provider-azure/cloud/scope"
 	"sigs.k8s.io/cluster-api-provider-azure/internal/test"
+	"sigs.k8s.io/cluster-api-provider-azure/util/reconciler"
 )
 
 var _ = Describe("AzureMachineReconciler", func() {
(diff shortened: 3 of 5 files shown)


shysank:
apidiff:

```
sigs.k8s.io/cluster-api-provider-azure/controllers
  Incompatible changes:
  - AzureMachineReconciler: old is comparable, new is not
  Compatible changes:
  - NewAzureMachineReconciler: added
  - Service: added
```

controllers/interfaces.go
```diff
@@ -0,0 +1,25 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import "context"
+
+// Service provides operations for a reconciliation service
+type Service interface {
```
devigned, line 22: What do you think about renaming this to `Reconciler`.

controllers/azuremachine_controller.go
```diff
@@ -51,9 +51,33 @@ import (
...
-	ReconcileTimeout time.Duration
+	Log                        logr.Logger
+	Recorder                   record.EventRecorder
+	ReconcileTimeout           time.Duration
+	azureMachineServiceFactory func(machineScope *scope.MachineScope, clusterScope *scope.ClusterScope) *azureMachineService
+}
+
+// NewAzureMachineReconciler returns a new AzureMachineReconciler instance
+func NewAzureMachineReconciler(client client.Client, log logr.Logger, recorder record.EventRecorder, reconcileTimeout time.Duration,
+	opts ...func(azureMachineReconciler *AzureMachineReconciler)) *AzureMachineReconciler {
```
devigned, line 62: I'm good with the variadic options, but I'd prefer to have have a type for an option rather than a raw func definition.

controllers/azuremachine_controller.go
```diff
@@ -51,9 +51,33 @@ import (
 // AzureMachineReconciler reconciles a AzureMachine object
 type AzureMachineReconciler struct {
 	client.Client
-	Log              logr.Logger
-	Recorder         record.EventRecorder
-	ReconcileTimeout time.Duration
+	Log                        logr.Logger
+	Recorder                   record.EventRecorder
+	ReconcileTimeout           time.Duration
+	azureMachineServiceFactory func(machineScope *scope.MachineScope, clusterScope *scope.ClusterScope) *azureMachineService
```
devigned, line 57: Should `azureMachineServiceFactory` be exported? If you are going to provide variadic options for injecting the service, seems like a lot of work to not expose it on the public API? If this is only to be used in testing, perhaps, it would be best to only set `azureMachineServiceFactory` via setting the field in the struct directly and forgo all the work to add variadic options.

devigned (changes requested):
I'm not clear that the `azureMachineServiceFactory` will be set anywhere besides tests. If the ability to inject an `azureMachineServiceFactory` for testing purposes, perhaps, that can be done simply by having a test within the same package as this struct and set the struct field directly.

controllers/azuremachine_controller.go
shysank: >I'm not clear that the azureMachineServiceFactory will be set anywhere besides tests. If the ability to inject an azureMachineServiceFactory for testing purposes, perhaps, that can be done simply by having a test within the same package as this struct and set the struct field directly.

Addressing this as well since it's on a similar note, I didn't want to export `azureMachineServiceFactory` for a few reasons :
1. It's also used in `main.go, and so we'll then need to export `newAzureMachineService` as well, and I thought it might make main.go cumbersome.
2. Better to have an api to initialize rather than setting the struct values directly. Provides consistency and easier to test. 
3. Easier to set defaults.

I agree that the variadic options are not needed right now, but just added it to make sure that there is a way to override defaults in case something changes in future. wdyt?

devigned: This is the only optional we've needed to add. I'm hesitant to add variadic options for this only because we might add others in the future. Since it's not on the public API for `NewAzureMachineReconciler` func, I'd vote to keep it simple. In the `NewAzureMachineReconciler`, we can set the default as you are doing, and in tests, just replace the default value set in the constructor. Everything would work the same as today, but also provide us the ability to replace the default in tests.

controllers/azuremachine_controller.go
shysank: Do you mean something like `type option func(azureMachineReconciler *AzureMachineReconciler)`?

devigned: Yes, here's an example of what I mean.

Definition:
https://github.com/devigned/pub/blob/8124492a2e15b32f2d0a55d9b6e7a71f33ed4478/pkg/partner/client.go#L40

Usage:
https://github.com/devigned/pub/blob/8124492a2e15b32f2d0a55d9b6e7a71f33ed4478/pkg/partner/client.go#L154

devigned: Pending the need to use variadic options, of course.

shysank:
@devigned I've made some small changes based on the feedback I received for a similar refactor [here](https://github.com/kubernetes-sigs/cluster-api/pull/3946#pullrequestreview-542311994). I've also removed the variadic options to keep it simple. wdyt?

/cc @CecileRobertMichon

devigned (approved):
lgtm

@CecileRobertMichon wdyt?

controllers/interfaces.go
```diff
@@ -0,0 +1,25 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import "context"
+
+// Reconciler provides operations for a reconciliation service
+type Reconciler interface {
```
CecileRobertMichon, line 22: this looks very similar to https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/cloud/interfaces.go#L28, I wonder if we should consider combining them.

shysank: Should we just make `azuremachineService` also be of type `azure.Service`? As in it still does `reconcile` and `delete` like other cloud/services except that it is kind of a composite service which orchestrates other individual services?

devigned: If I recall, @shysank made azuremachineService a azure.Service initially. I advised against that because I didn't want anyone to infer the azuremachineService was actually a cloud service, one of the services that lives `./cloud/services`.

I think the `Service` interface is really an abstract concept that overflows the bounds of `./cloud/services`. It is really a description of the reconcile behavior, reconcile to state or delete.

@CecileRobertMichon do you remember talking about how a reconciler could be a composition of reconiling services? I think this is that pattern is surfacing. `azuremachineService` is a reconciling service composed of a collection of reconciling services likely in some order of operations.

Perhaps, we just use `azure.Service` for now, and look to lift the interface from the `azure` package into a more generic package in a refactor.

Sorry if I've going into the bike shed on this one, but I just want to ensure we are making abstractions at the right levels.

CecileRobertMichon (commented):
Overall PR lgtm, I like this second iteration a lot better. Thanks @shysank!

controllers/azuremachine_reconciler.go
```diff
@@ -34,6 +34,8 @@ import (
 	"sigs.k8s.io/cluster-api-provider-azure/cloud/services/virtualmachines"
 )
 
+var _ azure.Service = (*azureMachineService)(nil)
```
CecileRobertMichon, line 37: @shysank can you please open an issue with the thread we had about service composition to track moving the azure.Service interface to a more generic place since it's not specific to cloud/

CecileRobertMichon (commented):
/lgtm

controllers/interfaces.go
shysank: @devigned @CecileRobertMichon For now, I've made `azureMachineReconciler` to use `azure.service`. Filed an [issue](https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1075) to move `azure.Service` outside of `cloud/`, PTAL.

controllers/azuremachine_reconciler.go
shysank: oh, just now saw this comment. did it already [here](https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1075) :)

shysank:
@CecileRobertMichon Just noticed that I missed to use `createAzureMachineService` in Delete method.  Sorry, lgtm got removed. PTAL, whenever you get a chance.

CecileRobertMichon (approved):
/lgtm
/approve

PR #1057, opened 2020-11-30
💚 check for cluster == nil before dumping logs and deleting

mboersma:
/kind flake

We've been seeing some e2e failures where the workload cluster doesn't provision before a timeout. In that case, no logs are dumped and the cluster isn't cleaned up. 

But hitting a `nil` pointer exception in the `AfterEach` handler silently exits, preventing some valid cleanup from happening. This logs that case instead--which we still intend to fix--so that `AfterEach` runs to completion.

Fixes #

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
💚 check for cluster == nil before dumping logs and deleting
```

### DIFF
--- a/test/e2e/common.go
+++ b/test/e2e/common.go
@@ -71,8 +71,17 @@ func setupSpecNamespace(ctx context.Context, specName string, clusterProxy frame
 }
 
 func dumpSpecResourcesAndCleanup(ctx context.Context, specName string, clusterProxy framework.ClusterProxy, artifactFolder string, namespace *corev1.Namespace, cancelWatches context.CancelFunc, cluster *clusterv1.Cluster, intervalsGetter func(spec, key string) []interface{}, skipCleanup bool) {
-	Byf("Dumping logs from the %q workload cluster", cluster.Name)
-	clusterProxy.CollectWorkloadClusterLogs(ctx, cluster.Namespace, cluster.Name, filepath.Join(artifactFolder, "clusters", cluster.Name, "machines"))
+	defer func() {
+		cancelWatches()
+		redactLogs()
+	}()
+
+	if cluster == nil {
+		By("Unable to dump workload cluster logs as the cluster is nil")
+	} else {
+		Byf("Dumping logs from the %q workload cluster", cluster.Name)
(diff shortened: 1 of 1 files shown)


test/e2e/common.go
```diff
@@ -83,20 +88,24 @@ func dumpSpecResourcesAndCleanup(ctx context.Context, specName string, clusterPr
 	})
 
 	if !skipCleanup {
```
devigned, line 90: wdyt about:

```go
	defer func() {
		cancelWatches()
		redactLogs()
	}

	if skipCleanup {
		return
	}

	if cluster == nil {
		log.Logf("Unable to delete workload cluster as cluster is nil")
	}
	
	Byf("Deleting cluster %s/%s", cluster.Namespace, cluster.Name)
	// While https://github.com/kubernetes-sigs/cluster-api/issues/2955 is addressed in future iterations, there is a chance
	// that cluster variable is not set even if the cluster exists, so we are calling DeleteAllClustersAndWait
	// instead of DeleteClusterAndWait
	framework.DeleteAllClustersAndWait(ctx, framework.DeleteAllClustersAndWaitInput{
		Client:    clusterProxy.GetClient(),
		Namespace: namespace.Name,
	}, intervalsGetter(specName, "wait-delete-cluster")...)

	Byf("Deleting namespace used for hosting the %q test spec", specName)
	framework.DeleteNamespace(ctx, framework.DeleteNamespaceInput{
		Deleter: clusterProxy.GetClient(),
		Name:    namespace.Name,
	})
```

devigned (changes requested):

test/e2e/common.go
mboersma: Nice, I think that reads more cleanly with `defer`. I'll refactor it.

nader-ziada:
the change looks good, but not sure why the coverage job is failing, haven't seen this as a flake before

mboersma:
The flakes went away, just like a Head & Shoulders shampoo ad. But they helped illuminate that this PR is working as intended--I saw it print that it couldn't collect logs, but it did clean up the cluster in that case, so those are both improvements.

nader-ziada (commented):
Thanks @mboersma, nice improvement!

/lgtm
/approve

PR #1063, opened 2020-12-02
adding securityGroupResourceGroup to generated azure.json

### LINKED ISSUE #356: Cluster with NSG in separate RG fails to expose service

nader-ziada:
**What type of PR is this?**

/kind bug

- set securityGroupResourceGroup field to custom vnet resource group (if set) in generated azure.json

Fixes #356 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
 set securityGroupResourceGroup to custom vnet resource group in generated azure.json 
```


api/v1alpha3/types.go
```diff
@@ -407,6 +407,10 @@ type SubnetSpec struct {
 	// +optional
 	SecurityGroup SecurityGroup `json:"securityGroup,omitempty"`
 
+	// The name of the resource group that the security group is deployed in
+	// +optional
+	SecurityGroupResourceGroup string `json:"securityGroupResourceGroup,omitempty"`
```
CecileRobertMichon, line 412: I don't believe we want/need to expose this level of customization... In general I'd expect the security group to be in the same RG as the vnet or the rest of the cluster (or both). #1063 was mostly about adding `SecurityGroupResourceGroup:   d.Vnet().ResourceGroup` to support BYO vnet cases where the network infra is in a separate RG as the cluster (eg https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/356#issuecomment-735685949)

CecileRobertMichon: I haven't heard of use cases from users that want their security group to be created for them in a different RG.

nader-ziada: So just set the `security group resource group` to the `vnet resource group` if it’s different from the `cluster resource group`?

CecileRobertMichon: Right. You can set it to always be vnet().ResourceGroup and that value will `== cluster resource group` when no vnet RG is specified

nader-ziada: Oh ok, that easy to do, I misunderstood the requirement. Will make the changes.

nader-ziada: @CecileRobertMichon made the changes, PTAL

CecileRobertMichon (approved):
/lgtm
/approve

Issue #819, opened 2020-07-23
MachinePool K8s upgrade should provide MinAvailable, MaxSurge, with safe rolling updates


CecileRobertMichon:
/kind bug

https://github.com/kubernetes-sigs/cluster-api/issues/3160

Create a cluster with a machine pool, eg. : 

```
NAME                                   STATUS   ROLES    AGE     VERSION
capz-vmss-master-control-plane-7k9ql   Ready    master   15m     v1.17.5
capz-vmss-master-control-plane-bx7vl   Ready    master   18m     v1.17.5
capz-vmss-master-control-plane-jt4z5   Ready    master   19m     v1.17.5
capz-vmss-master-mp-0000000            Ready    <none>   4m10s   v1.17.5
capz-vmss-master-mp-0000002            Ready    <none>   4m14s   v1.17.5
``` 

Try to edit the mp resource and change `Spec.Version` to another k8s version, such as `v1.18.3`. Nothing happens. The mp stays in phase "running" and no update is made to amp.

```
NAME                    PHASE
capz-vmss-master-mp-0   Running

NAME                    REPLICAS   READY   STATE
capz-vmss-master-mp-0   2          true    Succeeded
```

The upgrade policy is currently set to "manual". In order to implement rolling upgrade, we'd have to loop over the instances and drain/update them one by one. 

In the future, we might be able to leverage vmss rolling upgrade by setting MaxUnavailable (supported) and MaxSurge (not currently supported) and make those configurable / exposed to the user. 

See also https://github.com/awesomenix/drainsafe by @awesomenix which we might be able to leverage to drain the nodes before they get deleted.

awesomenix:
Also want to add that there is a small setting change required on VMSS 

https://docs.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-terminate-notification#rest-api

to allow drainsafe to receive termination events. If we can integrate drainsafe (or similar mechanism), we can treat upgrades as any other maintenance operation which would mean that there shouldnt be any special code for upgrade (optionally just include VMSS rolling upgrade directly). If its not yet assigned to anybody i can work on this issue

CecileRobertMichon:
@awesomenix go for it!

CecileRobertMichon:
@awesomenix any updates on this? We'd like to move MachinePool out of exp/ as part of v1alpha4 which is planned for Q4 and this is a blocker for that.

[...]

PR #1067, opened 2020-12-05
async create and update for machine pools

devigned:
**What type of PR is this?**
/kind feature

Currently, all of the reconcilers call Azure and wait for the operation to complete. At times, this is the right thing to do, but most of the time, it's the equivalent of the UI freezing on an app b/c your have used the UI thread to fetch some data causing your user to wonder why and when the software will react. This PR is a WIP that will make the reaction time of MachinePools in CAPZ drastically faster and possibly, more resilient.

image

related to: #819

I've intentionally not change the version upgrade / VMSS instance update behavior. I'll follow up with another PR which will add min-available replicas, maxsurge, and drain behaviors. That should finish off #819.

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Enable asynchronous reconciliation for AzureMachinePools
```


devigned:
@alexeldeib notice half or the time in reconcile now is spent resourceSkus. We should discuss a good way to really cache that data rather than on each reconcile.

CecileRobertMichon (commented):
This looks great! I'll have to take an in depth look on Monday but excited to see where it's going. One question that came to mind, how will this affect # of API calls, especially for subscriptions running many clusters with many scale sets? Before, we'd make one call for create and wait for the operation to be done. Now, we might check in a few times before the actual operation is done so a single update might be x3-4 calls, which could add up with many scale sets. I see one improvement to not update the scale set unless something actually changed though so that's good.

CecileRobertMichon:
One more question: in the case of machinepools, there are only two resources involved: vmss and role assignments. If you were to create a machinepool with system assigned identity, does it mean the role assignment create would fail and requeue until the scale set has finished provisioning? In the case of an AzureMachine where there are a lot more services involved, would every resource that has a dependency on the previous one fail if the dependencies aren't ready? I don't see this necessarily being a problem but maybe we can be smart about it and only attempt to provision if we know that the dependent resources have finished provisioning by looking at their future/state? But then that means CAPZ needs to recreate the ARM dependency graph. Or maybe we just need to improve logging so the logs aren't full of noisy "vm not found" ARM errors.

devigned:
I added an LRU cache for the resource skus. The resulting performance is greatly increased. See the Jaeger output below. Notice there is no call to refresh SKU data.

image

devigned:
> This looks great! I'll have to take an in depth look on Monday but excited to see where it's going. One question that came to mind, how will this affect # of API calls, especially for subscriptions running many clusters with many scale sets? Before, we'd make one call for create and wait for the operation to be done. Now, we might check in a few times before the actual operation is done so a single update might be x3-4 calls, which could add up with many scale sets. I see one improvement to not update the scale set unless something actually changed though so that's good.

That is an excellent question! Though, the basis for concern is a little misplaced. When we issue a long running operation to Azure, as CAPZ is in master, we ask Azure to create a resource, this causes a long running operation to start. This means that we get back a 201 saying the request was accepted, but not completed. **We then rely on the SDK to poll the API for a result, often many times, like every 10s until a terminal provisioning status is reached.** Unfortunately, Azure has no way to call us back when it's done building the resource. The Azure SDK is not very smart about how often it polls. It is configurable, but at this point, we don't do anything with the polling configuration.

In this PR, I have it set to requeue after 15s, which is probably too quick for VMSS anyway. I think we might be making less API calls over time than using the SDK to poll the long running operation.

In the future, I think we could get really clever with this functionality and provide some data driven targets. For example, we could determine the statistical distribution of VM provision time, know that the controller is creating a VM, then setup a requeue for each P20% with the idea that we'd likely have 2 or 3 requests per VM creation. We could do this per resource and update the data set as time moves forward. This could be a huge win for CAPZ users.

WDYT, @CecileRobertMichon?

devigned:
> One more question: in the case of machinepools, there are only two resources involved: vmss and role assignments. If you were to create a machinepool with system assigned identity, does it mean the role assignment create would fail and requeue until the scale set has finished provisioning? In the case of an AzureMachine where there are a lot more services involved, would every resource that has a dependency on the previous one fail if the dependencies aren't ready? I don't see this necessarily being a problem but maybe we can be smart about it and only attempt to provision if we know that the dependent resources have finished provisioning by looking at their future/state? But then that means CAPZ needs to recreate the ARM dependency graph. Or maybe we just need to improve logging so the logs aren't full of noisy "vm not found" ARM errors.

This is interesting artifact of how I structured the async behavior. When we reach a point in reconciliation where we are going to wait on Azure, the service returns an `OperationNotDoneError`. This causes a short-circuit of the reconciliation. That is to say that when we encounter an error, we wrap and send up the stack immediately. We do not progress, we halt and return. This behavior is desirable because it stops us from progressing to resources that we may not be able to complete yet. This PR would not be possible without the uniform handling of transient error which was recently merged.

There is way too much noise in the logs from transient errors right now as we log them as actual errors. I don't think they should be errors, but rather a category of expected behavior where the operator should stay calm and reconcile on. Thus, we should remove them from the log.

controllers/azuremachine_reconciler.go
```diff
@@ -48,8 +47,11 @@ type azureMachineService struct {
 }
 
 // newAzureMachineService populates all the services based on input scope.
-func newAzureMachineService(machineScope *scope.MachineScope, clusterScope *scope.ClusterScope) *azureMachineService {
-	cache := resourceskus.NewCache(clusterScope, clusterScope.Location())
+func newAzureMachineService(machineScope *scope.MachineScope, clusterScope *scope.ClusterScope) (*azureMachineService, error) {
+	cache, err := resourceskus.GetCache(clusterScope, clusterScope.Location())
```
alexeldeib, line 51: note that this introduces a subtle edge case, where the cached sku data is not refresh. we could potentially mitigate with background refresh. also, the data inside a single sku cache currently refreshes on first use instead of instantiation which probably slows things down quite a bit (vs instantiating, kicking off refresh in goroutine, and then waiting for it to be ready before allowing the cache to beused).

alexeldeib:
Looks like great stuff so far :) 

> We then rely on the SDK to poll the API for a result, often many times, like every 10s until a terminal provisioning status is reached. Unfortunately, Azure has no way to call us back when it's done building the resource. The Azure SDK is not very smart about how often it polls. It is configurable, but at this point, we don't do anything with the polling configuration.

Wonder if it's worth extracting that ourselves into a one httpclient (using some of the autorest helpers?) and then using that as the base for all service clients.

devigned:
> Wonder if it's worth extracting that ourselves into a one httpclient (using some of the autorest helpers?) and then using that as the base for all service clients.

We could / should. I'll send you a proposal I had worked on unrelated to this project. The thing is, the behaviors will be domain specific. For example, a vanilla marketplace image VM with a given sku starts / deletes have different latency distributions than our specialized use case. The helpers would be useful if we can provide operation type, resource type and latency distribution. Without that specific input, it probably would be too generic to be helpful.

> note that this introduces a subtle edge case, where the cached sku data is not refresh. 

Great point! I'll work on evicting a SKU from the cache after a max TTL. Any thoughts on how long the TTL should be?

controllers/azuremachine_reconciler.go
devigned: I've opted to make the LRU cache a TTL/LRU cache for the time being. Your solution with a go routine would probably be more efficient, but I didn't want to get too deep into it in this PR.

cloud/services/scalesets/client.go
```diff
@@ -207,3 +334,21 @@ func (ac *AzureClient) GetPublicIPAddress(ctx context.Context, resourceGroupName
 
 	return ac.publicIPs.Get(ctx, resourceGroupName, publicIPName, "true")
 }
+
+// NewOperationNotDoneError returns a new OperationNotDoneError wrapping a Future
+func NewOperationNotDoneError(future *infrav1.Future) *OperationNotDoneError {
```
CecileRobertMichon, line 339: Should this stuff live in errors.go (in cloud pkg)? It's not scale set specific and it should be reusable by other services once we expand these changes to all reconcilers.

alexeldeib:
> Any thoughts on how long the TTL should be?

the easiest way might not be a TTL, but e.g. refreshing the cache for a specific subscription if you hit an unexpected error? but that might be too aggressive. anyway I doubt restrictions change per sub _that_ often, if it becomes an issue maybe we can revisit.

devigned:
/assign @CecileRobertMichon @alexeldeib @nader-ziada 

I think this PR is ready for review. I've kept the commits separate to help with review. I'll squash them before merge.

cloud/services/scalesets/scalesets_test.go
```diff
@@ -192,979 +200,210 @@ func TestGetExistingVMSS(t *testing.T) {
...
-				g.Expect(err).To(MatchError(tc.expectedError))
-			} else {
-				g.Expect(err).NotTo(HaveOccurred())
-				g.Expect(result).To(BeEquivalentTo(tc.result))
-			}
-		})
-	}
-}
-
-func TestReconcileVMSS(t *testing.T) {
```
CecileRobertMichon, line 207: 🙌

CecileRobertMichon (changes requested):
Did a code review, overall code organization and behavioral changes look great. I'm going to checkout the branch locally and try it out.

When you eventually squash the commits, let's keep the LRU cache stuff separate from the Async stuff.

api/v1alpha3/types.go
```diff
@@ -27,6 +27,23 @@ const (
 	Node string = "node"
 )
 
+// Future contains the data needed for an Azure long running operation to continue across reconcile loops
+type Future struct {
```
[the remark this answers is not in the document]
devigned: It is part of the `AzureMachinePoolStatus` in exp. I think we could add this to each resource that supports long running operations and make it part of a generic way to handle long running operations.

cloud/services/scalesets/client.go
```diff
@@ -34,18 +39,42 @@ type Client interface {
...
+	}
+)
+
+const (
+	// PatchFuture is a future that was derived from a PATCH request to VMSS
+	PatchFuture string = "PATCH"
+	// PutFuture is a future that was derived from a PUT request to VMSS
+	PutFuture string = "PUT"
+	// DeleteFuture is a future that was derived from a DELETE request to VMSS
+	DeleteFuture string = "Delete"
```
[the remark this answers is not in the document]
devigned: Oops. I'll fix it.

cloud/services/scalesets/scalesets.go
```diff
@@ -258,81 +249,112 @@ func (s *Service) Reconcile(ctx context.Context) error {
...
 
 	switch {
 	case err != nil && !azure.ResourceNotFound(err):
+		// There was an error and it was not an HTTP 404 not found. This is either a transient error in Azure or a bug.
 		return errors.Wrapf(err, "failed to get VMSS %s", vmssSpec.Name)
-	case err == nil:
-		// VMSS already exists
-		// update it
+	case err != nil && azure.ResourceNotFound(err):
+		// The Virtual Machine Scale Set was not found, HTTP 404. We should create the VMSS.
```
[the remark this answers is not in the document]
devigned: Will do.

cloud/services/scalesets/scalesets.go
```diff
@@ -258,81 +249,112 @@ func (s *Service) Reconcile(ctx context.Context) error {
...
+		vmss.Tags = converters.TagsToMap(tagsWithoutHash.AddSpecVersionHashTag(hash))
+		s.Scope.SetProvisioningState(infrav1.VMStateCreating)
+		future, err = s.Client.CreateOrUpdateAsync(ctx, s.Scope.ResourceGroup(), vmssSpec.Name, vmss)
+		if err != nil {
+			return errors.Wrapf(err, "cannot create VMSS")
+		}
+
+		s.Scope.V(2).Info("successfully created VMSS", "scale set", vmssSpec.Name)
+		s.Scope.SetLongRunningOperationState(future)
+		s.Scope.SetAnnotation("cluster-api-provider-azure", "true")
```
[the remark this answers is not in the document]
devigned: Uhm... I have no idea why this is here. I'm going to remove it.

cloud/services/scalesets/scalesets.go
```diff
@@ -258,81 +249,112 @@ func (s *Service) Reconcile(ctx context.Context) error {
...
+
 		if err := s.Client.UpdateInstances(ctx, s.Scope.ResourceGroup(), vmssSpec.Name, instanceIDs); err != nil {
 			return errors.Wrapf(err, "failed to update VMSS %s instances", vmssSpec.Name)
 		}
-		s.Scope.SaveK8sVersion()
 
+		s.Scope.SaveK8sVersion()
 		// get the VMSS to update status
-		existingVMSS, err = s.getExisting(ctx, vmssSpec.Name)
+		fetchedVMSS, err = s.getVirtualMachineScaleSet(ctx, vmssSpec.Name)
```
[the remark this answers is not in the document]
devigned: I'm going to optimize this as part of the max surge / rolling update work. I wanted to leave it the same in this PR.

cloud/services/scalesets/scalesets.go
```diff
@@ -258,81 +249,112 @@ func (s *Service) Reconcile(ctx context.Context) error {
...
-	s.Scope.SetAnnotation("cluster-api-provider-azure", "true")
-	s.Scope.SetProvisioningState(existingVMSS.State)
+
+	s.Scope.SetProvisioningState(fetchedVMSS.State)
 	return nil
 }
 
-// Delete deletes a scale set.
+// Delete deletes a scale set asynchronously. Delete sends a DELETE request to Azure and if accepted without error,
+// the VMSS will be considered deleted. The actual delete in Azure may take longer, but should eventually complete.
```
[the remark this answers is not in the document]
devigned: If delete fails, what would we do? At that point, I think you need human intervention.

controllers/azuremachine_reconciler.go
devigned: I'll open a second PR with the Cache and refactor the `ClusterScope` dep.

cloud/services/scalesets/scalesets.go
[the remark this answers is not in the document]
CecileRobertMichon: let's find out if it's being used by capi / part of an old provider contract before we remove it.

cloud/services/scalesets/scalesets.go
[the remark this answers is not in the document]
CecileRobertMichon: sounds good

cloud/scope/machinepool.go
```diff
@@ -138,7 +138,9 @@ func (m *MachinePoolScope) ProviderID() string {
 
 // SetProviderID sets the AzureMachinePool providerID in spec.
 func (m *MachinePoolScope) SetProviderID(v string) {
-	m.AzureMachinePool.Spec.ProviderID = v
+	if m.AzureMachinePool.Spec.ProviderID != "" {
```
CecileRobertMichon, line 141: I just realized this by looking at another issue, I think we need to keep the overwrite. I know it seems wrong since providerID for a VM should never change but since this field is part of spec it is technically modifiable by the user and so I think the idea is we should always overwrite it with the source of truth. We can't set up the webhook to reject updates to providerID either because otherwise the controller will fail when trying to set it for the first time.

It should probably be labeled as "READ-ONLY" in the CRD, too.

devigned: Makes sense. I had received mixed guidance on this when submitting a PR to CAPI. I like your reasoning. Also, with patch helper, it should only be updated if it has changed.

cloud/services/scalesets/scalesets.go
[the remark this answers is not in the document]
devigned: checked and can't find any references

cloud/services/scalesets/scalesets_test.go
```diff
@@ -1181,251 +420,39 @@ func TestReconcileVMSS(t *testing.T) {
 			},
 		},
 		{
-			name:          "scale set already exists",
-			expectedError: "",
+			name:          "should start updating when scale set already exists and not currently in a long running operation",
```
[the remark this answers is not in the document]
devigned: This test mutates the spec and expects the vmss to be mutated too if it is not already processing a long running operation. We are making sure that if Hash(originalSpec) != Hash(updatedSpec), we update the vmss.

devigned:
@CecileRobertMichon I have broken the PR into two commits. 1, with the sku cache and 2, with the machine pool async work.

CecileRobertMichon:
Purely cosmetic, but, I noticed the azuremachinepool briefly goes into "Updating" state when it's first created before it goes to "Succeeded"

```
Every 2.0s: kubectl get azuremachinepools -o wide                                                                               ceciles-mbp.lan: Thu Dec 17 12:08:29 2020

NAME                        REPLICAS   READY   STATE      CLUSTER                MACHINEPOOL                 VMSS ID
                                                                                                                    VM SIZE
machinepool-template-mp-0                      Updating   machinepool-template   machinepool-template-mp-0   azure:///subscriptions/redacted/
resourceGroups/machinepool-template/providers/Microsoft.Compute/virtualMachineScaleSets/machinepool-template-mp-0   Standard_D2s_v3
```

```
Every 2.0s: kubectl get azuremachinepools -o wide                                                                               ceciles-mbp.lan: Thu Dec 17 12:09:41 2020

NAME                        REPLICAS   READY   STATE       CLUSTER                MACHINEPOOL                 VMSS ID
                                                                                                                     VM SIZE
machinepool-template-mp-0   2          true    Succeeded   machinepool-template   machinepool-template-mp-0   azure:///subscriptions/redacted
/resourceGroups/machinepool-template/providers/Microsoft.Compute/virtualMachineScaleSets/machinepool-template-mp-0   Standard_D2s_v3
```

I'd expect the first create to go straight from creating to succeeded.

devigned:
> Purely cosmetic, but, I noticed the azuremachinepool briefly goes into "Updating" state when it's first created before it goes to "Succeeded"

The reason for this is that not all of the instances had joined the cluster. The instances were running, but the nodes were not all joined. During that period of time.

https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/06a4b3c38dc43190bf6c19bdec66dded7fa5b517/cloud/scope/machinepool.go#L206-L219

I'm intending on changing this behavior in the upcoming PR for MaxUnavailable and MaxSurge. What state should be shown, ScalingUp / ScalingDown, Upgrading?

CecileRobertMichon:
> The reason for this is that not all of the instances had joined the cluster. The instances were running, but the nodes were not all joined. During that period of time.

Makes sense. Updating is acceptable for now because it describes a generic "created but not ready" state IMO. We can tacle more specific scale/upgrade states with surge.

cloud/services/scalesets/scalesets.go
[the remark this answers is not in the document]
CecileRobertMichon: Fair. I guess it could fail because of a transient Azure failure in which case you'd want to retry?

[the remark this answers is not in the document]
devigned: @CecileRobertMichon that is a fair point. I'd hope that is a really low percentage of fails.

I have no attachment to async delete outside of the user experience of quick delete. That said, I'm totally open to mothballing it. Up to you.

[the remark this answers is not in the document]
CecileRobertMichon: I'm thinking the huge benefit of fast delete outweighs the occasional delete flake. 

That being said, how do you feel about moving async delete to its own PR? This PR is already pretty loaded and it has a lot of other goodness that I don't want to slow down for delete. That way if we decide async delete was a bad idea after all we can more easily revert.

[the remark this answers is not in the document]
devigned: Of course! sgtm.

controllers/azuremachine_reconciler.go
CecileRobertMichon: ClusterScope dep is still there right? Should I open an issue?

CecileRobertMichon (approved):
/lgtm
👏 👏 👏 

/assign @alexeldeib @nader-ziada

controllers/azuremachine_reconciler.go
devigned: Oops... fixup incoming.

devigned: :sparkles:  Clean up! :sparkles: :

devigned:
If folks have some time, PTAL.

nader-ziada:
looks great, lgtm, will let others approve since this is my first pass

devigned:
Lazy consensus alert: we are planning on merging tomorrow end of day if there are no new feedback.

cloud/services/resourceskus/cache.go
```diff
@@ -45,17 +48,51 @@ type Cache struct {
 	data []compute.ResourceSku
 }
 
+// Cacher describes the ability to get and to add items to cache
```
alexeldeib, line 51: dupe of the other cache interface you defined elsewhere, minus remove?

alexeldeib: ah no eviction

alexeldeib:
/lgtm

nice work! some big improvements

CecileRobertMichon:
different flake...

/retest

devigned:
failed "Creates a public management cluster in the same vnet"

config/manager/manager_image_patch.yaml
```diff
@@ -8,5 +8,5 @@ spec:
     spec:
       containers:
       # Change the value of image field below to your controller image URL
-      - image: gcr.io/k8s-staging-cluster-api-azure/cluster-api-azure-controller:latest
+      - image: localhost:5000/ci-e2e/cluster-api-azure-controller-amd64:20210112194504
```
CecileRobertMichon, line 11: pushed by mistake?

config/manager/manager_pull_policy.yaml
```diff
@@ -8,4 +8,4 @@ spec:
     spec:
       containers:
       - name: manager
-        imagePullPolicy: Always
+        imagePullPolicy: IfNotPresent
```
CecileRobertMichon, line 11: ditto

config/manager/manager_image_patch.yaml
devigned: yeah... just testing. I'll clean it up.

CecileRobertMichon:
@devigned how about separating the e2e changes in a separate PR? that way they won't get lost in this already big PR.

devigned:
@CecileRobertMichon I will. Just using this branch since it provides more details, faster feedback about VMSS nodes.

devigned:
@CecileRobertMichon I've rebased off of master with the VMSS test changes.

CecileRobertMichon (approved):
/lgtm

PR #1082, opened 2020-12-18
[wip] Async machine pool delete

devigned:
**What type of PR is this?**
/kind feature

This PR enables asynchronous deletes for AzureMachinePool resources. Upon delete of an AzureMachinePool resource, the reconciler will send the delete request to Azure, and will consider the resource deleted if Azure responds with a successful response. The reconciler _will not_ wait for the successful completion of the delete.

Implications:
- Not waiting for successful completion of the delete will greatly speed up delete from the K8s perspective.
- Not waiting for successful completion will leave open the possibility that some Azure infrastructure will fail to be deleted in the case when Azure accepts the delete, but fails while deleting the infrastructure. This will require human intervention to clean up the infrastructure resources. In the rare case Azure does fail to clean up infrastructure, it may be indicative of a bug in Azure and would likely need a customer support ticket opened.
- When deleting an AzureMachinePool and then creating another named the same while the first is still deleting will cause a delay in provisioning the new AzureMachinePool due to the need to wait for the previous infrastructure to be deleted before a create can be issued.

related to: #819, #1067

Will rebase after #1067 is merged.

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
enable asynchronous delete of AzureMachinePool backing Azure infrastructure greatly speeding up `kubectl delete machinepool my-pool`
```

### DIFF
--- a/cloud/services/scalesets/scalesets.go
+++ b/cloud/services/scalesets/scalesets.go
@@ -151,7 +151,7 @@ func (s *Service) Delete(ctx context.Context) error {
 
 	vmssSpec := s.Scope.ScaleSetSpec()
 	s.Scope.V(2).Info("deleting VMSS", "scale set", vmssSpec.Name)
-	err := s.Client.Delete(ctx, s.Scope.ResourceGroup(), vmssSpec.Name)
+	_, err := s.Client.DeleteAsync(ctx, s.Scope.ResourceGroup(), vmssSpec.Name)
 	if err != nil {
 		if azure.ResourceNotFound(err) {
 			// already deleted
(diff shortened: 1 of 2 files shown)


devigned:
I've updated this PR to be based on master since #1067 has merged. Still not sure how much I like this behavior, but wanted to propose it for discussion.

cc @CecileRobertMichon and @nader-ziada

CecileRobertMichon:
Could this also cause potential quota issues? If CAPZ considers the machine "deleted" and moves on with replacing it before it is completely deleted, that would mean you'd need quota for both machines to exist simultaneously. Pair that with many clusters in the same sub or even many machines in one cluster and you might reach quota limits a lot faster.

devigned:
For this https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1082#issuecomment-767782279 and a myriad of other reasons, I move that we close this PR. The changes are minute. If we choose to move in this direction in the future, it is trivial to dig this back up.

devigned:
/close
We may reevaluate async delete at a later time.

PR #1036, opened 2020-11-13
Windows support via kubeadm

jsturtevant:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #153 

Implements https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/proposals/20200804-windows-support.md

**Special notes for your reviewer**:
Uses an image from https://github.com/kubernetes-sigs/image-builder/pull/382. This PR is close to merging and then this can go in.

Similar to #1035 

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**TODOs**:

- [ ] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Add windows support 
```

PR #1097, opened 2020-12-22
Align kubectl install scripts and update version to min supported

jsturtevant:
**What type of PR is this?**
/kind cleanup

In https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1036#issuecomment-749767796, I saw we are using kubectl 1.16.x when there was a failure to pull the correct version.  

Looking into it I noticed the `ci-e2e.sh` calls `hack/ensure-kubectl.sh` and the `make` file does something slightly different.  I've updated to use the same script and updated to `1.17`

Fixes #

We might want to address all the differences between our `hack/ensure-<tool>.sh` and the `make` file way of installing, such as kustomize.  I didn't do this since the kustomize file is used in lots of places we as this kubectl isn't referenced by the `hack/tools/bin` path.

https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/980b22e2e4c9bf24355cf6ff16f7787d35b0a5f5/hack/ensure-kustomize.sh#L21-L22

https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/eac5daafbef1ad33ee02c105feea49a1cef906bf/Makefile#L232-L233

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -88,9 +88,9 @@ GINKGO_VER := v1.13.0
 GINKGO_BIN := ginkgo
 GINKGO := $(TOOLS_BIN_DIR)/$(GINKGO_BIN)-$(GINKGO_VER)
 
-KUBECTL_VER := v1.16.13
+KUBECTL_VER := v1.17.16
 KUBECTL_BIN := kubectl
-KUBECTL := $(TOOLS_BIN_DIR)/$(KUBECTL_BIN)-$(KUBECTL_VER)
+KUBECTL := $(KUBECTL_BIN)-$(KUBECTL_VER)
 
 KUBE_APISERVER=$(TOOLS_BIN_DIR)/kube-apiserver
 ETCD=$(TOOLS_BIN_DIR)/etcd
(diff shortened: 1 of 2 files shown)


devigned:
> We might want to address all the differences between our hack/ensure-<tool>.sh and the make file way of installing, such as kustomize. I didn't do this since the kustomize file is used in lots of places we as this kubectl isn't referenced by the hack/tools/bin path.

Agreed. It's one of the build system incongruities that's bugged me. It'd be nice to use all project local tools consistently across scripts.

jsturtevant:
Makes sense to me.  To confirm the preference is to use the makefile approach with `hack/tools/bin` and use those tools consistently across the project?

jsturtevant:
Discussed out office hours this morning and the consensus was to align on the `makefile` approach with `hack/tools/bin`.  Will update this PR to reflect that.

/hold

jsturtevant:
I haven't had a chance to get back to this pr and go with the decided upon aproach and opened #1192 in case someone else has opportunity.

PR #1099, opened 2020-12-23
Availability sets for control plane nodes

### LINKED ISSUE #657: HA Clusters in Non-AZ enabled Regions

***Goals***
1. Provide a mechanism to create HA clusters in regions that do not support AZs
1. Investigate VMSS Orchestration Mode (Preview Feature) and Availability Sets
1. Investigate interoperability between Availability Zones and (VMSS Orchestration Mode or Availability Sets)

***Non-Goals/Future Work***
1. Core API CRD support for VMSS Orchestration Mode

**User Story**

As a user I would like to highly available clusters in all regions including regions without AZs so my clusters can serve workloads consistently.

**Detailed Description**

We need to first investigate the edge cases around HA clusters backed by VMSS Orchestration Mode and Availability Sets to understand trade-offs in either implementation. Notably, Availability Sets pin resources to 1 of 3 fault domains which are selected on the first VM being assigned to the availability set. This means that all VMs assigned to that Availability Set must reside within the same set of hardware. For instance a single AvailSet can not support a Standard_DS_v2 VM at the same time as a Standard_DS_v3 VM as they reside on independent hardware. Also Availability Sets have scaling constraints due to max # of vms per AvailSet as well as capacity constraints related to the capacity of the three Fault Domains selected and not the full capacity of that region.

[VMSS Orchestration Mode - Docs](https://docs.microsoft.com/en-us/azure/virtual-machine-scale-sets/orchestration-modes)
[Availability Sets - Docs](https://docs.microsoft.com/en-us/azure/virtual-machines/windows/manage-availability)

There may be some changes required in Cluster API to support Failure Groups that are managed by the provider. While AZs can be manually managed by the User Fault Domains aren't exposed and are internally managed by Availability Sets. 

/kind proposal

shysank:
**What type of PR is this?**
/kind feature

Implements availability sets for control plane nodes as a fallback if the region does not have any failure domains

Fixes #657 

- VMs already created in existing clusters prior to this change will not be mapped to any available set even though a new availability set will be created during cluster reconciliation.
- New VMs created in existing clusters through upgrade or scale operations will be assigned to an availability set.
- New VMs Created in new clusters will be assigned to an availability set.
- Only control plane VMs will be assigned to an availability set.

All of the above applies only when there are no failure domains available.

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Availability sets for control plane nodes 
** Control plane VMs already created in existing cluster will not be affected by this change although a new availability set will be created if there are no failure domains.
```


shysank:
Tried it in `northcentralus` which has no az's:
image

cloud/scope/cluster.go
```diff
@@ -269,6 +269,14 @@ func (s *ClusterScope) PrivateDNSSpec() *azure.PrivateDNSSpec {
 	return spec
 }
 
+// AvailabilitySetSpecs returns the availability set specs.
+func (s *ClusterScope) AvailabilitySetSpecs() []azure.AvailabilitySetSpec {
+	return []azure.AvailabilitySetSpec{{
+		Name:                    azure.GenerateAvailabilitySetName(s.ClusterName(), azure.ControlPlaneNodeGroup),
+		DesiredFaultDomainCount: azure.DefaultFaultDomainCount,
```
CecileRobertMichon, line 276: Is the fault domain count always 3? I know you had questions about it before the break, did you answer those?

controllers/azurecluster_reconciler.go
```diff
@@ -110,6 +114,13 @@ func (r *azureClusterReconciler) Reconcile(ctx context.Context) error {
 		return errors.Wrapf(err, "failed to reconcile private dns")
 	}
 
+	// reconcile availability sets if there are no failure domains
+	if len(r.scope.FailureDomains()) == 0 {
```
CecileRobertMichon, line 118: there is no need for this check, if there are no failure domains the reconcile should be a no-op as `AvailabilitySetSpecs()` will return an empty slice

controllers/azurecluster_reconciler.go
```diff
@@ -147,6 +158,13 @@ func (r *azureClusterReconciler) Delete(ctx context.Context) error {
 			if err := r.vnetSvc.Delete(ctx); err != nil {
 				return errors.Wrapf(err, "failed to delete virtual network")
 			}
+
+			// delete availability sets if there are no failure domains
+			if len(r.scope.FailureDomains()) == 0 {
```
CecileRobertMichon, line 163: same here

cloud/services/virtualmachines/virtualmachines.go
```diff
@@ -181,9 +185,22 @@ func (s *Service) Reconcile(ctx context.Context) error {
...
+			if asName, ok := s.Scope.AvailabilitySet(); ok {
+				as, err := s.availabilitySetsClient.Get(ctx, s.Scope.ResourceGroup(), asName)
+				if err != nil {
+					return errors.Wrapf(err, "failed to get availability set %s", asName)
+				}
+
+				virtualMachine.AvailabilitySet = &compute.SubResource{ID: as.ID}
+			}
+
+		} else {
```
CecileRobertMichon, line 199: } else if vmSpec.Zone != "" {
			zones := []string{vmSpec.Zone}
			virtualMachine.Zones = &zones
		}

cloud/services/virtualmachines/virtualmachines.go
```diff
@@ -181,9 +185,22 @@ func (s *Service) Reconcile(ctx context.Context) error {
 			},
 		}
 
-		if vmSpec.Zone != "" {
-			zones := []string{vmSpec.Zone}
-			virtualMachine.Zones = &zones
+		// Set availability set if no failure domains are available
+		if len(s.Scope.FailureDomains()) == 0 {
```
CecileRobertMichon, line 189: this could be 

```
if vmSpec.Zone != "" {
 // set zone
} else {
 // set availability set
}
```

Since `len(s.Scope.FailureDomains()) != 0` <==> `vmSpec.Zone != ""`

cloud/services/virtualmachines/virtualmachines.go
```diff
@@ -181,9 +185,22 @@ func (s *Service) Reconcile(ctx context.Context) error {
...
+			if asName, ok := s.Scope.AvailabilitySet(); ok {
+				as, err := s.availabilitySetsClient.Get(ctx, s.Scope.ResourceGroup(), asName)
```
CecileRobertMichon, line 191: no need to make a GET call for the ID here (API calls are expensive). The ID is deterministic, you can generate it from the name. See `SubnetID()`  in defaults.go and how it's used in networkinterfaces.go for an example.

cloud/scope/cluster.go
CecileRobertMichon: Ah I spoke too soon... I see it's replaced later based on the sku cache return.

cloud/services/availabilitysets/availabilitysets.go
```diff
@@ -0,0 +1,132 @@
...
+
+// Delete deletes availability sets
+func (s *Service) Delete(ctx context.Context) error {
+	ctx, span := tele.Tracer().Start(ctx, "availabilitysets.Service.Delete")
+	defer span.End()
+
+	for _, asSpec := range s.Scope.AvailabilitySetSpecs() {
+		s.Scope.V(2).Info("deleting availability set", "availability set", asSpec.Name)
+		err := s.Client.Delete(ctx, s.Scope.ResourceGroup(), asSpec.Name)
+		if err != nil {
```
CecileRobertMichon, line 125: we shouldn't return an error if the error is not found (the AS is already deleted)

cloud/services/availabilitysets/availabilitysets.go
```diff
@@ -0,0 +1,132 @@
...
+			return errors.Wrapf(err, "failed to create availability set %s", asSpec.Name)
+		}
+		s.Scope.V(2).Info("successfully created availability set", "availability set", asSpec.Name)
+	}
+
+	return nil
+}
+
+// Delete deletes availability sets
```
CecileRobertMichon, line 117:
```suggestion
// Delete deletes availability sets.
```

cloud/services/availabilitysets/availabilitysets.go
```diff
@@ -0,0 +1,132 @@
...
+// New creates a new availability sets service.
+func New(scope AvailabilitySetScope, skuCache *resourceskus.Cache) *Service {
+	return &Service{
+		Scope:            scope,
+		Client:           NewClient(scope),
+		resourceSKUCache: skuCache,
+	}
+}
+
+// Reconcile creates or updates availability sets
```
CecileRobertMichon, line 60:
```suggestion
// Reconcile creates or updates availability sets.
```

cloud/services/availabilitysets/availabilitysets.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	for _, asSpec := range s.Scope.AvailabilitySetSpecs() {
+		s.Scope.V(2).Info("creating availability set", "availability set", asSpec.Name)
+
+		faultDomainCount := asSpec.DesiredFaultDomainCount
+		if asSpec.DesiredFaultDomainCount > int32(maxFaultDomainCount) {
+			s.Scope.V(2).Info("desired fault domain count is greater than max fault domain count, will use the max fault domain count",
+				"desired fault domain count", asSpec.DesiredFaultDomainCount, "max fault domain count", maxFaultDomainCount)
+			faultDomainCount = int32(maxFaultDomainCount)
```
CecileRobertMichon, line 87: What's the point of `DesiredFaultDomainCount`? Why not always use `maxFaultDomainCount` ? Do we ever want to use a lower number than max?

cloud/scope/machine.go
```diff
@@ -299,6 +299,15 @@ func (m *MachineScope) ProviderID() string {
 	return parsed.ID()
 }
 
+// AvailabilitySet returns the availability set for this machine if available
+func (m *MachineScope) AvailabilitySet() (string, bool) {
+	if len(m.ClusterScoper.FailureDomains()) == 0 && m.IsControlPlane() {
```
CecileRobertMichon, line 304: Just curious, have you thought about this approach can be expanded later to add availability sets for worker nodes part of machine deployments?

cloud/scope/cluster.go
shysank: Yeah, I checked the possible list of fault domains using `az vm list-skus --resource-type availabilitySets --query '[?name==`Aligned`].{Location:locationInfo[0].location, MaximumFaultDomainCount:capabilities[0].value}' -o Table
`, and the max was 3. So, I thought it makes sense to do a `min(fdCountOfSku, desiredFDCount)`

cloud/services/availabilitysets/availabilitysets.go
```diff
@@ -0,0 +1,132 @@
...
+				ClusterName: s.Scope.ClusterName(),
+				Lifecycle:   infrav1.ResourceLifecycleOwned,
+				Name:        to.StringPtr(asSpec.Name),
+				Role:        to.StringPtr(infrav1.CommonRole),
+				Additional:  s.Scope.AdditionalTags(),
+			})),
+			Location: pointer.StringPtr(s.Scope.Location()),
+		}
+
+		_, err := s.Client.CreateOrUpdate(ctx, s.Scope.ResourceGroup(), asSpec.Name, asParams)
```
CecileRobertMichon, line 107: Just a note: with this change, an availability set will be created for all existing clusters but the VMs won't be added to it since they already exist. This won't cause any functional issues but the AS will be unused, which might be a bit confusing to users. However, any new control plane VMs (eg. created as part of a k8s upgrade or a scale up) would added to the set. Might worth calling that out in the PR description.

controllers/azurecluster_reconciler.go
shysank: > if there are no failure domains the reconcile should be a no-op
IIUC, isn't it the opposite of this? As in, reconcile availability sets only when there are no failure domains for that location.

cloud/services/availabilitysets/availabilitysets.go
shysank: Not now, just wanted to keep `AvailabilitySetSpec` open for changes in future. I'm okay with removing it and hardcoding it to max count if it make it simple.

cloud/scope/machine.go
shysank: I did, it shouldn't be too hard, if we just have one AS for all worker node (of all machine deployments). That's why I've left availability spec as an array, so that if it'd be easier to accommodate the changes, if needed. Happy to do it in this pr, if everyone's okay with the approach.

cloud/services/virtualmachines/virtualmachines.go
shysank: I'll take a look into it, Thanks!

controllers/azurecluster_reconciler.go
CecileRobertMichon: sorry, if _there are_ failure domains the reconcile should be a no-op

cloud/scope/machine.go
CecileRobertMichon: let's do worker nodes in a separate PR to keep the changes well-scoped and easier to review

shysank:
@CecileRobertMichon I've addressed the review comments. PTAL, whenever you get a chance.

cloud/services/availabilitysets/availabilitysets.go
```diff
@@ -0,0 +1,140 @@
...
+		//noop if there are failure domains
+		return nil
+	}
+
+	for _, asSpec := range s.Scope.AvailabilitySetSpecs() {
+		s.Scope.V(2).Info("deleting availability set", "availability set", asSpec.Name)
+		err := s.Client.Delete(ctx, s.Scope.ResourceGroup(), asSpec.Name)
+		if err != nil && azure.ResourceNotFound(err) {
+			// already deleted
+			return nil
```
CecileRobertMichon, line 130:
```suggestion
			continue
```

We should finish the loop to delete any potential other availability sets

cloud/defaults.go
```diff
@@ -176,6 +187,12 @@ func NATRuleID(subscriptionID, resourceGroup, loadBalancerName, natRuleName stri
 	return fmt.Sprintf("/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Network/loadBalancers/%s/inboundNatRules/%s", subscriptionID, resourceGroup, loadBalancerName, natRuleName)
 }
 
+// AvailabilitySetID returns the azure resource ID for a given availability set.
+func AvailabilitySetID(subscriptionID, resourceGroup, availabilitySetName string) string {
+	///subscriptions/eba22861-4990-4213-ab8e-776393e3696b/resourceGroups/cl-as-ncus/providers/Microsoft.Compute/availabilitySets/cl-as-ncus_control-plane-as
```
CecileRobertMichon, line 192: is this comment intentional or did you forget to clean it up?

cloud/services/availabilitysets/availabilitysets.go
shysank: Ah yeah, my bad, sorry :(

CecileRobertMichon: let's make sure this is covered by a unit test too

cloud/defaults.go
shysank: Fixed it, Thanks!

cloud/services/availabilitysets/availabilitysets.go
shysank: Done.

cloud/services/availabilitysets/availabilitysets.go
shysank: Updated the `Special notes for your reviewer` section to add this note.

shysank:
pull-cluster-api-provider-azure-e2e

cloud/services/availabilitysets/availabilitysets.go
nader-ziada: maybe add it to the release notes?

nader-ziada:
PR looks good @shysank

CecileRobertMichon (approved):
/lgtm
/approve

cloud/services/availabilitysets/availabilitysets.go
shysank: Updated release notes as well, Thanks.

shysank:
Merged master to see if if fixes `pull-cluster-api-provider-azure-e2e-windows` test.

CecileRobertMichon:
/retest

#1123 should help with the flakes...

Issue #1102, opened 2021-01-04
Loadbalancer Probe uses "/healthz" https endpoint. Suggestion to switch to a different probe.


zawachte-msft:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]

Set anonymous-auth for the api-server to false in a kubeadmcontrolplane config.

**What did you expect to happen:**

Cluster comes up successifully.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

Today the load balancer code creates a healthcheck probe which hits the api-server /healthz endpoint.

Two issues with this.

1. The /healthz endpoint has been deprecated since kubernetes v1.16 https://kubernetes.io/docs/reference/using-api/health-checks/
2. If you have anonymous-auth for the api-server set to false, this breaks cluster-bring-up since the loadbalancer probe will fail with an auth error.

Not well versed in the azure load balancer so I don't know what the proper fix is, but I tested changing to a TCP probe and it works.

I assume there is a better solution. AWS seems to have a SSL check that doesn't require this endpoint.

https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/pkg/cloud/services/elb/loadbalancer.go#L318

PR #1103, opened 2021-01-04
Loadbalancer Probe uses "/healthz" https endpoint. Suggestion to switch to a different probe.

zawachte-msft:
**What type of PR is this?**

/kind bug

Today the load balancer code creates a healthcheck probe which hits the api-server /healthz endpoint. 

Two issues with this.

1. The /healthz endpoint has been deprecated since kubernetes v1.16 https://kubernetes.io/docs/reference/using-api/health-checks/
2. If you have anonymous-auth for the api-server set to false, this breaks cluster-bring-up since the loadbalancer probe will fail with an auth error.

Not well versed in the azure load balancer so I don't know what the proper fix is, but I tested changing to a TCP probe and it works. 

I assume there is a better solution. AWS seems to have a SSL check that doesn't require this endpoint.

Fixes #

#1102

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note

```

### DIFF
--- a/cloud/services/loadbalancers/loadbalancers.go
+++ b/cloud/services/loadbalancers/loadbalancers.go
@@ -101,13 +101,12 @@ func (s *Service) Reconcile(ctx context.Context) error {
 		}
 
 		if lbSpec.Role == infrav1.APIServerRole {
-			probeName := "HTTPSProbe"
+			probeName := "TCPProbe"
 			lb.LoadBalancerPropertiesFormat.Probes = &[]network.Probe{
 				{
 					Name: to.StringPtr(probeName),
 					ProbePropertiesFormat: &network.ProbePropertiesFormat{
-						Protocol:          network.ProbeProtocolHTTPS,
-						RequestPath:       to.StringPtr("/healthz"),
+						Protocol:          network.ProbeProtocolTCP,
 						Port:              to.Int32Ptr(lbSpec.APIServerPort),
 						IntervalInSeconds: to.Int32Ptr(15),
 						NumberOfProbes:    to.Int32Ptr(4),
--- a/cloud/services/loadbalancers/loadbalancers_test.go
+++ b/cloud/services/loadbalancers/loadbalancers_test.go
@@ -130,18 +130,17 @@ func TestReconcileLoadBalancer(t *testing.T) {
 											ID: to.StringPtr("/subscriptions/123/resourceGroups/my-rg/providers/Microsoft.Network/loadBalancers/my-publiclb/backendAddressPools/my-publiclb-backendPool"),
 										},
 										Probe: &network.SubResource{
-											ID: to.StringPtr("/subscriptions/123/resourceGroups/my-rg/providers/Microsoft.Network/loadBalancers/my-publiclb/probes/HTTPSProbe"),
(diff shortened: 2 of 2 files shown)


nader-ziada:
@zwachtel11 you also need to sign the CNCF CLA

CecileRobertMichon:
This is essentially a revert of https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/444. Trying to remember if there was any functional reason for the change (besides the security aspect).

Azure does not have an "SSL" protocol like [AWS](https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/elb-healthchecks.html#health-check-configuration) and [GCP](https://cloud.google.com/load-balancing/docs/health-checks) do. Using the [HTTPS](https://docs.microsoft.com/en-us/azure/load-balancer/load-balancer-custom-probe-overview#httpprobe) is the only way to use TLS.

> 1. The /healthz endpoint has been deprecated since kubernetes v1.16 https://kubernetes.io/docs/reference/using-api/health-checks/

could this be fixed by using one of the two endpoints that replaced it? Maybe `livez`?

> 2. If you have anonymous-auth for the api-server set to false, this breaks cluster-bring-up since the loadbalancer probe will fail with an auth error.

not sure what the expected behavior is for this one. What happens in CAPA if you set anonymous-auth to false with the SSL protocol?

zawachte-msft:
"An HTTP/HTTPS health check succeeds if the instance returns a 200 response code within the health check interval. A TCP health check succeeds if the TCP connection succeeds. An SSL health check succeeds if the SSL handshake succeeds." from docs.

"An SSL health check succeeds if the SSL handshake succeeds." I would assume that even in the anonymous-auth "on" case, the ssl handshake would fail since the api-server isn't listening with a publicly rooted cert. I probably have a misunderstanding.

worth asking a question in slack I guess, don't have access to an aws account.

CecileRobertMichon:
@zawachte-msft make sure you sign the CLA when you get a chance

zawachte-msft:
yea, sorry, it's my other account that needs to sign the cla. Annoying switching contexts in github. Doing it now!

zawachte-msft:
I signed it

zawachte-msft:
note: that ^ was per the instructions :) "If you haven't yet, post a comment ("I signed it") to the PR—this is needed for the system to recheck your contribution authorization."

zawachte-msft:
Link to slack thread...

https://kubernetes.slack.com/archives/C8TSNPY4T/p1610381211408100

CecileRobertMichon:
@zawachte-msft what was the conclusion here? Do we want this change as-is?

@devigned thoughts?

zawachte-msft:
> @zawachte-msft what was the conclusion here? Do we want this change as-is?
> 
> @devigned thoughts?

I have no strong opinions. I was just running into this when I was working on a bootstrap provider for k3s since they turn off anon auth by default. https://github.com/zawachte-msft/cluster-api-bootstrap-provider-k3s

Figured it was worth a shout, but there is certainly a tradeoff with this change.

CecileRobertMichon:
@zawachte-msft this conflicts with #1141 which merged yesterday, when you get a chance please rebase and we can get it in. Just a heads up that #1141 refactored most of the lines this PR is touching so it might make more sense to reapply the changes on a fresh branch instead of trying to resolve the conflicts.

PR #1105, opened 2021-01-06
machine pool surge, max unavailable and instance delete

### LINKED ISSUE #819: MachinePool K8s upgrade should provide MinAvailable, MaxSurge, with safe rolling updates

devigned:
**What type of PR is this?**
/kind feature

This PR introduces API changes to AzureMachinePool to enable users of CAPZ to perform safe, fast rolling upgrades building off of #1067. In this change set `MaxSurge` and `MaxUnavailable` fields are introduced to the `AzureMachinePool` spec. 
- `MaxSurge` enables machine pools to over-provision machines, increase the number of machines above the desired count, during an upgrade, which would allow faster upgrades at the cost of Azure compute core quota. 
- `MaxUnavailable` enables one to specify how many nodes must be available during a rolling upgrade.
- `Instance delete` enables an individual to delete Azure Machine Pool Machines individually and controller initiated node drain. As part of the instance delete / node drain state tracking for `AzureMachinePoolMachines`, it is advantageous to track state on these resources individually. That is why in the PR, the `AzureMachinePool.Status.Instances` array is removed in favor of `AzureMachinePoolMachine` resources.
- Node drain will be completed in a subsequent PR

:warning: Breaking out `AzureMachinePool` instances into individual `AzureMachinePoolMachine` resources will be a breaking change in the **experimental API**. 

This PR is work in progress. Please provide early feedback.

Fixes #819 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Introduce maxUnavailable and maxSurge to `AzureMachinePool` and remove `AzureMachinePool.Status.Instances` in favor of representing machine pool machines as individual resources of type `AzureMachinePoolMachine`. This is a breaking change to the experimental API.
```


Tiltfile
```diff
@@ -154,7 +154,7 @@ def capz():
     local_resource(
         "manager",
         cmd = 'mkdir -p .tiltbuild;CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -ldflags \'-extldflags "-static"\' -o .tiltbuild/manager',
-        deps = ["api", "cloud", "config", "controllers", "exp", "feature", "pkg", "go.mod", "go.sum", "main.go"]
+        deps = ["api", "azure", "config", "controllers", "exp", "feature", "pkg", "go.mod", "go.sum", "main.go"]
```
devigned, line 157: I should probably pull this out into another PR, but what's 1 line on 4k+ :disappointed:

controllers/helpers.go
```diff
@@ -404,3 +434,73 @@ func GetClusterIdentityFromRef(ctx context.Context, c client.Client, azureCluste
...
+type coalescingReconciler struct {
+	upstream reconcile.Reconciler
+	cache    *CoalescingRequestCache
+	log      logr.Logger
+}
+
+// NewCoalescingReconciler returns a reconcile wrapper that will delay new reconcile.Requests
+// after the cache expiry of the request string key.
+// A successful reconciliation is defined as as one where no error is returned
+func NewCoalescingReconciler(upstream reconcile.Reconciler, cache *CoalescingRequestCache, log logr.Logger) reconcile.Reconciler {
```
devigned, line 479: This wraps the `AzureMachinePool` and `AzureMachinePoolMachine` reconcilers so that they debounce, the reconcilers rate limit the incoming events so they only do so many within a window of time to not overwhelm Azure API limits.

There is no good way to do this in controller-runtime. I'll intro a proposal over there for middleware or incoming rate limiting.

exp/api/v1alpha3/azuremachinepool_types.go
```diff
@@ -68,6 +73,23 @@ type (
...
 
+	// AzureDefaultingImage defines information about the image to use for VMSS VM creation.
+	// There are three ways to specify an image: by ID, Marketplace Image or SharedImageGallery
+	// One of ID, SharedImage or Marketplace should be set.
+	//
+	// If no image is specified, the image will be defaulted by the controller and Defaulted set to true which will
+	// inform the controller that if the parent MachinePool.Spec.Template.Spec.Version field is updated, this image
+	// will be updated with the corresponding default image. If Defaulted is set to false, the controller will not
+	// update the image reference when the MachinePool.Spec.Template.Spec.Version changes.
+	AzureDefaultingImage struct {
```
devigned, line 84: Thought we should be explicit about image defaulting. This enables users to use default image versions for K8s versions specified on MachinePools while being safe to specify their own without updates to MachinePools overriding their image reference.

azure/scope/machinepool.go
```diff
@@ -153,61 +156,291 @@ func (m *MachinePoolScope) ProvisioningState() infrav1.VMState {
...
+	}
+	return ids
+}
+
+// selectMachinesToDelete will build a map of machines by provider ID to delete prioritizing in the following order
+// 1) failed machines
+// 2) over-provisioned machines prioritized by out of date models first
+// 3) over-provisioned ready machines
+// 4) ready machines within budget which have out of date models
+func (m *MachinePoolScope) selectMachinesToDelete(machinesByProviderID map[string]infrav1exp.AzureMachinePoolMachine) map[string]infrav1exp.AzureMachinePoolMachine {
```
devigned, line 305: This contains the logic for selecting machines to delete when over-provisioned or upgrading. The `AzureMachinePool` informs the `AzureMachinePoolMachine` to delete and the `AzureMachinePoolMachine` is expected to safely remove itself from the pool.

azure/scope/machinepool.go
```diff
@@ -153,61 +156,291 @@ func (m *MachinePoolScope) ProvisioningState() infrav1.VMState {
...
+	}
+	ampml := &infrav1exp.AzureMachinePoolMachineList{}
+	if err := m.client.List(ctx, ampml, client.InNamespace(m.AzureMachinePool.Namespace), client.MatchingLabels(labels)); err != nil {
+		return nil, errors.Wrap(err, "failed to list AzureMachinePoolMachines")
+	}
+
+	return ampml.Items, nil
+}
+
+func (m *MachinePoolScope) applyAzureMachinePoolMachines(ctx context.Context) error {
```
devigned, line 219: This contains the logic to compare the state of Azure VMSS with the state of `AzureMachinePool(Machine)s`. If a machine exists in Azure, a `AzureMachinePoolMachine` is created. If an `AzureMachinePoolMachine` exists, but doesn't have an Azure counterpart, it is asked to delete. At the end, the upgrade / over-provision state is evaluated and machines can be deleted if the state requires.

devigned:
@nader-ziada and @CecileRobertMichon I believe this PR is now ready for review. I'm so sorry of the enormity of the change set. With that in mind, I'm going to introduce `AzureMachinePoolMachine` cordon and drain functionality to a subsequent PR.

One design aspect I've vacillated on is how to represent `MaxSurge` and `MaxUnavailable`. Right now, they are on the top level spec for `AzureMachinePool`. Would it be better to use a rollout strategy structure similar to (or the same type as) https://github.com/kubernetes-sigs/cluster-api/pull/4073?

CecileRobertMichon:
@devigned I've been thinking about this and I think we need a proposal to explain what this does, why, what future work is planned and what alternatives were considered (and why they don't work). This is a pretty big PR and it's not obvious why we're doing upgrade this way, so we should document it for the record.

Also, it can serve as developer-facing documentation once the feature is merged. What do you think?

nader-ziada (commented):
I went through the changes and it all makes sense, but I have a comment about the structure of the code, the main logic that figures out what machines to create/delete is in the scope instead of in the controller, even setting the conditions, which is different that how we have done things in other places, was this by design?

devigned:
> I went through the changes and it all makes sense, but I have a comment about the structure of the code, the main logic that figures out what machines to create/delete is in the scope instead of in the controller, even setting the conditions, which is different that how we have done things in other places, was this by design?

It was a conscious decision. It was a bit of an experiment to see how it would turn out.
- Scope should be responsible for updating the K8s state upon closing the scope. 
- Controller / Reconciler is responsible for creating the Scope, calling services, responding to errors in reconciliation, and closing the Scope. 
- Services are responsible for manipulating external state and updating the Scope with the external state.

That was a perceptive review. Thank you. WDYT?

nader-ziada:
> 
> * Scope should be responsible for updating the K8s state upon closing the scope.
> * Controller / Reconciler is responsible for creating the Scope, calling services, responding to errors in reconciliation, and closing the Scope.
> * Services are responsible for manipulating external state and updating the Scope with the external state.
> 
> That was a perceptive review. Thank you. WDYT?

I don't feel strongly about it, but I expected the controller / reconciler to be responsible for also updating the status on the k8s resources, which would include setting the condition.

azure/services/scalesetvms/scalesetvms.go
```diff
@@ -0,0 +1,143 @@
...
+func NewService(scope ScaleSetVMScope) *Service {
+	return &Service{
+		Client: newClient(scope),
+		Scope:  scope,
+	}
+}
+
+// Reconcile idempotently gets, creates, and updates a scale set.
+func (s *Service) Reconcile(ctx context.Context) error {
+	ctx, span := tele.Tracer().Start(ctx, "scalesets.Service.Reconcile")
```
fiunchinho, line 62:
```suggestion
	ctx, span := tele.Tracer().Start(ctx, "scalesetvms.Service.Reconcile")
```

azure/services/scalesetvms/scalesetvms.go
```diff
@@ -0,0 +1,143 @@
...
+
+// NewService creates a new service.
+func NewService(scope ScaleSetVMScope) *Service {
+	return &Service{
+		Client: newClient(scope),
+		Scope:  scope,
+	}
+}
+
+// Reconcile idempotently gets, creates, and updates a scale set.
```
fiunchinho, line 60: This needs to be changed. Maybe something like "Reconcile fetches the latest data about the scaleset instance".

azure/services/scalesetvms/scalesetvms.go
```diff
@@ -0,0 +1,143 @@
...
+
+	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1alpha3"
+	"sigs.k8s.io/cluster-api-provider-azure/azure"
+	"sigs.k8s.io/cluster-api-provider-azure/azure/converters"
+	infrav1exp "sigs.k8s.io/cluster-api-provider-azure/exp/api/v1alpha3"
+	"sigs.k8s.io/cluster-api-provider-azure/util/tele"
+)
+
+type (
+	// ScaleSetVMScope defines the scope interface for a scale sets service.
```
fiunchinho, line 34:
```suggestion
	// ScaleSetVMScope defines the scope interface for a scaleset vms service.
```

azure/services/scalesetvms/scalesetvms.go
```diff
@@ -0,0 +1,143 @@
...
+		return errors.Wrap(err, "failed getting instance")
+	}
+
+	s.Scope.SetVMSSVM(converters.SDKToVMSSVM(instance))
+	return nil
+}
+
+// Delete deletes a scaleset instance asynchronously returning a future which encapsulates the long running operation.
+func (s *Service) Delete(ctx context.Context) error {
+	ctx, span := tele.Tracer().Start(ctx, "scalesets.Service.Delete")
```
fiunchinho, line 86:
```suggestion
	ctx, span := tele.Tracer().Start(ctx, "scalesetvms.Service.Delete")
```

exp/api/v1alpha3/azuremachinepool_types.go
fiunchinho: This sounds like a good idea. Should these `AzureDefaultingImage` changes be part of a different PR instead?

exp/controllers/helpers.go
```diff
@@ -271,3 +317,108 @@ func AzureClusterToAzureMachinePoolsFunc(kClient client.Client, log logr.Logger)
 		return result
 	}
 }
+
+// AzureMachinePoolToAzureMachinePoolMachines maps an AzureMachinePool to it's child AzureMachinePoolMachines through
```
fiunchinho, line 321:
```suggestion
// AzureMachinePoolToAzureMachinePoolMachines maps an AzureMachinePool to its child AzureMachinePoolMachines through
```

exp/api/v1alpha3/azuremachinepool_types.go
```diff
@@ -68,6 +73,23 @@ type (
...
+		*infrav1.Image `json:",omitempty"`
+
+		// Defaulted informs the controller that the image reference was defaulted so that it can be updated by changes
+		// to the MachinePool.Spec.Template.Spec.Version field by default.
+		// +optional
+		Defaulted bool `json:"defaulted,omitempty"`
```
fiunchinho, line 90: In this case `Defaulted` means not only that the value was defaulted but that the `Image` will be managed by the controller. What do you think about calling this field `Managed` instead? IMHO the word better conveys the behavior. I believe _managed_ is also how we refer to resources managed by the controllers in the code base.

exp/controllers/azuremachinepoolmachine_controller.go
```diff
@@ -0,0 +1,385 @@
...
+
+	var r reconcile.Reconciler = ampmr
+	if options.Cache != nil {
+		r = infracontroller.NewCoalescingReconciler(ampmr, options.Cache, log)
+	}
+
+	c, err := ctrl.NewControllerManagedBy(mgr).
+		WithOptions(options.Options).
+		For(&infrav1exp.AzureMachinePoolMachine{}).
+		WithEventFilter(predicates.ResourceNotPaused(log)). // don't queue reconcile if resource is paused
```
fiunchinho, line 95:
```suggestion
		WithEventFilter(predicates.ResourceNotPausedAndHasFilterLabel(ctrl.LoggerFrom(ctx), r.WatchFilterValue)).
```

exp/controllers/azuremachinepoolmachine_controller.go
```diff
@@ -0,0 +1,385 @@
...
+// NewAzureMachinePoolMachineController creates a new AzureMachinePoolMachineController to handle updates to Azure Machine Pool Machines
+func NewAzureMachinePoolMachineController(c client.Client, log logr.Logger, recorder record.EventRecorder, reconcileTimeout time.Duration) *AzureMachinePoolMachineController {
+	return &AzureMachinePoolMachineController{
+		Client:            c,
+		Log:               log,
+		Recorder:          recorder,
+		ReconcileTimeout:  reconcileTimeout,
+		reconcilerFactory: newAzureMachinePoolMachineReconciler,
+	}
+}
```
fiunchinho, line 81:
```suggestion
func NewAzureMachinePoolMachineController(c client.Client, log logr.Logger, recorder record.EventRecorder, reconcileTimeout time.Duration, watchFilterValue string) *AzureMachinePoolMachineController {
	return &AzureMachinePoolMachineController{
		Client:            c,
		Log:               log,
		Recorder:          recorder,
		ReconcileTimeout:  reconcileTimeout,
		reconcilerFactory: newAzureMachinePoolMachineReconciler,
		WatchFilterValue: watchFilterValue,
	}
}
```

exp/controllers/azuremachinepoolmachine_controller.go
```diff
@@ -0,0 +1,385 @@
...
+type (
+	azureMachinePoolMachineReconcilerFactory func(*scope.MachinePoolMachineScope) azure.Reconciler
+
+	// AzureMachinePoolMachineController handles Kubernetes change events for a AzureMachinePoolMachine resources
+	AzureMachinePoolMachineController struct {
+		client.Client
+		Log               logr.Logger
+		Scheme            *runtime.Scheme
+		Recorder          record.EventRecorder
+		ReconcileTimeout  time.Duration
```
fiunchinho, line 62:
```suggestion
		ReconcileTimeout  time.Duration
		WatchFilterValue  string
```

exp/controllers/azuremachinepoolmachine_controller.go
```diff
@@ -0,0 +1,385 @@
...
+	}
+
+	state := machineScope.ProvisioningState()
+	switch state {
+	case infrav1.VMStateFailed:
+		ampmr.Recorder.Eventf(machineScope.AzureMachinePoolMachine, corev1.EventTypeWarning, "FailedVMState", "Azure scale set VM is in failed state")
+		machineScope.SetFailureReason(capierrors.UpdateMachineError)
+		machineScope.SetFailureMessage(errors.Errorf("Azure VM state is %s", state))
+	case infrav1.VMStateDeleting:
+		// for some reason, the
```
fiunchinho, line 283: seems like the comment got truncated

config/crd/bases/exp.infrastructure.cluster.x-k8s.io_azuremachinepools.yaml
```diff
@@ -97,6 +97,30 @@ spec:
               location:
                 description: Location is the Azure region location e.g. westus2
                 type: string
+              maxSurge:
```
fiunchinho, line 100: The `maxSurge` field in the k8s `Deployment` object accepts an absolute number but also a percentage. Should we support percentage as well? I believe it improves the UX as the user doesn't have to adapt the `maxSurge` value when scaling up/down their VMSS. If we decide to support percentages, that would mean changing the type of the field.

fiunchinho: Oh actually [the proposal talks about both absolute numbers and percentages](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1191/files#diff-e67ed7281d9712d6611f6141a166992ccd15cc63bb553705678d7a6df8831b1bR203-R217).

azure/services/scalesetvms/client.go
```diff
@@ -0,0 +1,161 @@
...
+
+	vm, err := genericFuture.Result(ac.scalesetvms)
+	if err != nil {
+		return vm, errors.Wrapf(err, "failed fetching the result of operation for vmss")
+	}
+
+	return vm, nil
+}
+
+// DeleteAsync is the operation to delete a virtual machine scale set asynchronously. DeleteAsync sends a DELETE
```
fiunchinho, line 127:
```suggestion
// DeleteAsync is the operation to delete a virtual machine scale set vm asynchronously. DeleteAsync sends a DELETE
```

azure/services/scalesetvms/client.go
```diff
@@ -0,0 +1,161 @@
...
+// request to Azure and if accepted without error, the func will return a Future which can be used to track the ongoing
+// progress of the operation.
+//
+// Parameters:
+//   resourceGroupName - the name of the resource group.
+//   vmssName - the name of the VM scale set to create or update. parameters - the scale set object.
```
fiunchinho, line 133:
```suggestion
//   vmssName - the name of the VM scale set to create or update. parameters - the scale set object.
//   instanceID - the ID of the VM scale set VM.
```

azure/services/scalesetvms/client.go
```diff
@@ -0,0 +1,161 @@
...
+func (ac *azureClient) DeleteAsync(ctx context.Context, resourceGroupName, vmssName, instanceID string) (*infrav1.Future, error) {
+	ctx, span := tele.Tracer().Start(ctx, "scalesets.AzureClient.DeleteAsync")
```
fiunchinho, line 135:
```suggestion
	ctx, span := tele.Tracer().Start(ctx, "scalesetvms.AzureClient.DeleteAsync")
```

azure/services/scalesetvms/client.go
```diff
@@ -0,0 +1,161 @@
...
+	c := compute.NewVirtualMachineScaleSetVMsClientWithBaseURI(baseURI, subscriptionID)
+	c.Authorizer = authorizer
+	c.RetryAttempts = 1
+	_ = c.AddToUserAgent(azure.UserAgent()) // intentionally ignore error as it doesn't matter
+	return c
+}
+
+// Get retrieves the Virtual Machine Scale Set Virtual Machine
+func (ac *azureClient) Get(ctx context.Context, resourceGroupName, vmssName, instanceID string) (compute.VirtualMachineScaleSetVM, error) {
+	ctx, span := tele.Tracer().Start(ctx, "scalesets.AzureClient.Get")
```
fiunchinho, line 82:
```suggestion
	ctx, span := tele.Tracer().Start(ctx, "scalesetvms.AzureClient.Get")
```

devigned:
I'm going to close this PR and open a new PR with an updated branch based off or the proposal in #1191. I think it will help us to apply fresh eyes and a clean slate.

/close

PR #1120, opened 2021-01-13
check resource group exist before delete 

### LINKED ISSUE #1118: Cluster deletion stuck in continuous loop
/kind bug

copying issue from cluster-api repo since it is in the provider: https://github.com/kubernetes-sigs/cluster-api/issues/4065

I was expecting the cluster to be deleted and not shown in output of kubectl get cluster

i believe there are similar threads but for aws provider.

**Environment:**

nader-ziada:
**What type of PR is this?**

/kind bug

- check the resource group exist when reading its managed status so controller doesn't get stuck

Fixes #1118 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Extra check for resource group existence when deleting
```

### DIFF
--- a/cloud/services/groups/groups.go
+++ b/cloud/services/groups/groups.go
@@ -86,6 +86,10 @@ func (s *Service) Delete(ctx context.Context) error {
 	defer span.End()
 
 	managed, err := s.IsGroupManaged(ctx)
+	if err != nil && azure.ResourceNotFound(err) {
+		// already deleted or doesn't exist
+		return nil
+	}
 	if err != nil {
 		return errors.Wrap(err, "could not get resource group management state")
 	}
--- a/cloud/services/groups/groups_test.go
+++ b/cloud/services/groups/groups_test.go
@@ -151,6 +151,16 @@ func TestDeleteGroups(t *testing.T) {
 			},
 		},
 		{
+			name:          "resource group get returns error",
+			expectedError: "",
+			expect: func(s *mock_groups.MockGroupScopeMockRecorder, m *mock_groups.MockclientMockRecorder) {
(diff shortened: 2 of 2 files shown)


cloud/services/groups/groups.go
```diff
@@ -86,6 +86,10 @@ func (s *Service) Delete(ctx context.Context) error {
 	defer span.End()
 
 	managed, err := s.IsGroupManaged(ctx)
+	if err != nil && azure.ResourceNotFound(err) {
```
CecileRobertMichon, line 89: in #1118 we're not able to check the status of the RG at all because the sub ID is not found. The same issue would happen if the credentials had expired (GET fails). How does this change fix it? Is it because SubID not found returns a 404?

nader-ziada: yes, it was failing because the subscription name was wrong so the cluster never got created in Azure, but when you try to delete it, the controller was getting stuck on SubscriptionNotFound error in a loop

cloud/services/groups/groups_test.go
```diff
@@ -150,6 +150,16 @@ func TestDeleteGroups(t *testing.T) {
 				)
 			},
 		},
+		{
+			name:          "resource group does not exist",
```
CecileRobertMichon, line 154: how is this use case different than the one above?

nader-ziada: the error is returned from a different step, but good point about the name, will update that

CecileRobertMichon (approved):
/lgtm

devigned (approved):
lgtm

Issue #1115, opened 2021-01-11
E2E failure: LB GET failed after 5 attempts


CecileRobertMichon:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]

Seeing a lot of PR E2E flakes recently on the `with a single control plane node and an AzureMachinePool with 2 nodes [It]` spec. Test fails with:

```
/home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/test/e2e/azure_test.go:262
Unexpected error:
    <*fmt.wrapError | 0xc0000e2580>: {
        msg: "GET http://40.91.127.106 giving up after 5 attempt(s): Get http://40.91.127.106: dial tcp 40.91.127.106:80: i/o timeout",
        err: {
            Op: "Get",
            URL: "http://40.91.127.106",
            Err: {
                Op: "dial",
                Net: "tcp",
                Source: nil,
                Addr: {
                    IP: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 255, 255, 40, 91, 127, 106],
                    Port: 80,
                    Zone: "",
                },
                Err: {},
            },
        },
    }
    GET http://40.91.127.106 giving up after 5 attempt(s): Get http://40.91.127.106: dial tcp 40.91.127.106:80: i/o timeout
occurred
/home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/test/e2e/azure_lb.go:186
```

https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/directory/pull-cluster-api-provider-azure-e2e/1348697453653659648

**What did you expect to happen:**

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- cluster-api-provider-azure version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

CecileRobertMichon:
related #1123

CecileRobertMichon:
Update: we root caused this to a platform regression. The fix is being rolled out, ETA: 2 weeks from now.

PR #1131, opened 2021-01-21
temporarily disable ILB test for VMSS

devigned:
**What type of PR is this?**
/kind flake

ILB and ELB tests are failing due to a bug in the Azure resource provider that seems to be induced by adding an ILB to a VMSS node pool, deleting that ILB, and then creating an ELB. This is a temporary resolution for #1115. This should be removed after the bug is fixed in the API.

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/azure_lb.go
+++ b/test/e2e/azure_lb.go
@@ -46,6 +46,7 @@ type AzureLBSpecInput struct {
 	SkipCleanup           bool
 	IPv6                  bool
 	Windows               bool
+	IsVMSS                bool
 }
 
 // AzureLBSpec implements a test that verifies Azure internal and external load balancers can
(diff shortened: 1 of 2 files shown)


devigned:
/assign @CecileRobertMichon and @nader-ziada 

Looks like we have a pass on the the first run. I've limited the extent of the changes to focus only on ILB leaving ELB as is since I believe ELB would be the most important test to have signal.

nader-ziada:
/lgtm
but lets trigger another run to confirm, 

devigned:
@nader-ziada looks like it worked 2x :rocket:

devigned:
Related: https://github.com/kubernetes-sigs/cloud-provider-azure/pull/486

CecileRobertMichon:
Should we also re-enable the ELB test and only disable it for VMSS?

devigned:
> Should we also re-enable the ELB test and only disable it for VMSS?

I believe that is the case. ELB is always run. ILB is only run when not VMSS.

devigned:
@CecileRobertMichon ah, I see. I'll open another PR to enable the direct connect to the ELB. Great idea!

PR #1132, opened 2021-01-21
reenable direct connect to elb

devigned:
**What type of PR is this?**

/kind flake

We disabled the test that directly connected to the external load balancer due to flakes in the VMSS tests. This adds that test back.

Related to: #1131 #1115 

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Re-enable direct connection to the external load balancer test
```

### DIFF
--- a/test/e2e/azure_lb.go
+++ b/test/e2e/azure_lb.go
@@ -27,6 +27,7 @@ import (
 	. "github.com/onsi/gomega"
 	k8snet "k8s.io/utils/net"
 
+	"github.com/hashicorp/go-retryablehttp"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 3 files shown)


devigned:
/retest

Ran into IPv6 2 instead of 3 error https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/1132/pull-cluster-api-provider-azure-e2e/1352293189813276672

test/e2e/azure_lb.go
```diff
@@ -182,17 +183,17 @@ func AzureLBSpec(ctx context.Context, inputGetter func() AzureLBSpecInput) {
...
-	//	url := fmt.Sprintf("http://%s", elbIP)
-	//	resp, err := retryablehttp.Get(url)
-	//	if resp != nil {
-	//		defer resp.Body.Close()
-	//	}
-	//	Expect(err).NotTo(HaveOccurred())
-	//	Expect(resp.StatusCode).To(Equal(200))
-	//	Expect(err).NotTo(HaveOccurred())
-	//}
+	if !input.IPv6 {
```
CecileRobertMichon, line 186: so we're _not_ skipping this for VMSS, correct?

Is the current state of things:
- for VMSS tests, run ELB test only
- for non-VMSS tests, run ILB + ELB tests

?

devigned: 100% correct. My reasoning is that ELB is the most common scenario, so if we had to pick ILB or ELB, but not both for VMSS, ELB would be my pick.

PR #1137, opened 2021-01-22
Availability sets for worker nodes

### LINKED ISSUE #657: HA Clusters in Non-AZ enabled Regions

shysank:
**What type of PR is this?**
/kind feature

Continuing on the work done in #1099, this PR enables worker nodes to be assigned to availability sets if there are no failure domains available in the region.

Fixes #657 

- An availability set is created for each machine deployment
- Availability sets are now reconciled as part of `machine` reconciliation instead of `cluster` as in #1099. 
- Availability set is deleted when all the vms that belong to the as are deleted

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
enable availability sets for worker nodes
```

### DIFF
--- a/cloud/scope/cluster.go
+++ b/cloud/scope/cluster.go
@@ -269,21 +269,6 @@ func (s *ClusterScope) PrivateDNSSpec() *azure.PrivateDNSSpec {
 	return spec
 }
 
-// AvailabilitySetEnabled informs control plane machines they should be part of an Availability Set
-func (s *ClusterScope) AvailabilitySetEnabled() bool {
-	return len(s.AvailabilitySetSpecs()) > 0
-}
-
-// AvailabilitySetSpecs returns the availability set specs.
-func (s *ClusterScope) AvailabilitySetSpecs() []azure.AvailabilitySetSpec {
-	if len(s.AzureCluster.Status.FailureDomains) == 0 {
-		return []azure.AvailabilitySetSpec{{
-			Name: azure.GenerateAvailabilitySetName(s.ClusterName(), azure.ControlPlaneNodeGroup),
-		}}
-	}
-	return nil
-}
-
 // Vnet returns the cluster Vnet.
 func (s *ClusterScope) Vnet() *infrav1.VnetSpec {
 	return &s.AzureCluster.Spec.NetworkSpec.Vnet
@@ -398,6 +383,11 @@ func (s *ClusterScope) Namespace() string {
 // Location returns the cluster location.
 func (s *ClusterScope) Location() string {
 	return s.AzureCluster.Spec.Location
+}
+
+// AvailabilitySetEnabled informs machines that they should be part of an Availability Set.
+func (s *ClusterScope) AvailabilitySetEnabled() bool {
+	return len(s.AzureCluster.Status.FailureDomains) == 0
 }
 
 // GenerateFQDN generates a fully qualified domain name, based on a hash, cluster name and cluster location.
--- a/cloud/scope/machine.go
+++ b/cloud/scope/machine.go
@@ -328,8 +328,17 @@ func (m *MachineScope) ProviderID() string {
 
 // AvailabilitySet returns the availability set for this machine if available
 func (m *MachineScope) AvailabilitySet() (string, bool) {
-	if m.IsControlPlane() && m.AvailabilitySetEnabled() {
+	if !m.AvailabilitySetEnabled() {
+		return "", false
+	}
+
+	if m.IsControlPlane() {
 		return azure.GenerateAvailabilitySetName(m.ClusterName(), azure.ControlPlaneNodeGroup), true
+	}
+
+	// get machine deployment name from labels for machines that maybe part of a machine deployment.
+	if mdName, ok := m.Machine.Labels[clusterv1.MachineDeploymentLabelName]; ok {
+		return azure.GenerateAvailabilitySetName(m.ClusterName(), mdName), true
 	}
 
 	return "", false
--- a/cloud/scope/managedcontrolplane.go
+++ b/cloud/scope/managedcontrolplane.go
@@ -241,8 +241,3 @@ func (s *ManagedControlPlaneScope) OutboundLBName(_ string) string {
 func (s *ManagedControlPlaneScope) OutboundPoolName(_ string) string {
 	return "aksOutboundBackendPool" // hard-coded in aks
 }
-
-// FailureDomains returns the failure domains
-func (s *ManagedControlPlaneScope) FailureDomains() clusterv1.FailureDomains {
-	return clusterv1.FailureDomains{}
-}
--- a/cloud/services/availabilitysets/availabilitysets.go
+++ b/cloud/services/availabilitysets/availabilitysets.go
@@ -24,7 +24,6 @@ import (
 	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
-	"k8s.io/utils/pointer"
 
 	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1alpha3"
 	azure "sigs.k8s.io/cluster-api-provider-azure/cloud"
@@ -37,7 +36,7 @@ import (
 type AvailabilitySetScope interface {
 	logr.Logger
 	azure.ClusterDescriber
-	AvailabilitySetSpecs() []azure.AvailabilitySetSpec
+	AvailabilitySet() (string, bool)
 }
 
 // Service provides operations on azure resources
@@ -61,8 +60,8 @@ func (s *Service) Reconcile(ctx context.Context) error {
 	ctx, span := tele.Tracer().Start(ctx, "availabilitysets.Service.Reconcile")
 	defer span.End()
 
-	if len(s.Scope.AvailabilitySetSpecs()) == 0 {
-		//noop if there are failure domains
+	availabilitySetName, ok := s.Scope.AvailabilitySet()
+	if !ok {
 		return nil
 	}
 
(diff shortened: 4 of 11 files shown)


shysank:
image

cloud/scope/cluster.go
```diff
@@ -400,6 +388,16 @@ func (s *ClusterScope) Location() string {
 	return s.AzureCluster.Spec.Location
 }
 
+// FailureDomains returns the failure domains.
+func (s *ClusterScope) FailureDomains() clusterv1.FailureDomains {
+	return s.AzureCluster.Status.FailureDomains
+}
+
+// DeletionTimeStamp returns the deletion timestamp.
```
CecileRobertMichon, line 396:
```suggestion
// DeletionTimeStamp returns the deletion timestamp of the owner Cluster.
```

for clarity

cloud/scope/machine.go
```diff
@@ -328,10 +328,19 @@ func (m *MachineScope) ProviderID() string {
 
 // AvailabilitySet returns the availability set for this machine if available
 func (m *MachineScope) AvailabilitySet() (string, bool) {
-	if m.IsControlPlane() && m.AvailabilitySetEnabled() {
+	if len(m.FailureDomains()) > 0 {
```
CecileRobertMichon, line 331: I think I prefer the explicit `AvailabilitySetEnabled()` and hiding the details of what that means (`len(s.AzureCluster.Status.FailureDomains) == 0`) in helper func

CecileRobertMichon: at the very least we should add a comment to explain that why don't define an availability set when there are failure domains available

cloud/scope/machine.go
```diff
@@ -328,10 +328,19 @@ func (m *MachineScope) ProviderID() string {
...
+		return "", false
+	}
+
+	if m.IsControlPlane() {
 		return azure.GenerateAvailabilitySetName(m.ClusterName(), azure.ControlPlaneNodeGroup), true
 	}
 
+	// get machine deployment name from labels
+	if mdName, ok := m.Machine.Labels[clusterv1.MachineDeploymentLabelName]; ok {
```
CecileRobertMichon, line 340: wondering if it makes more sense to use the MachineSet or the MachineDeployment name here 🤔  any particular reason you chose MD?

cloud/scope/managedcontrolplane.go
```diff
@@ -132,6 +132,11 @@ func (s *ManagedControlPlaneScope) AdditionalTags() infrav1.Tags {
 	return tags
 }
 
+// FailureDomains returns the failure domains.
+func (s *ManagedControlPlaneScope) FailureDomains() clusterv1.FailureDomains {
```
CecileRobertMichon, line 136: if you end up getting rid of `AvailabilitySetEnabled` in ClusterDescriber, make sure to allow remove it from this file (line 122)

CecileRobertMichon (commented):
lgtm overall, added a few comments

cloud/scope/cluster.go
shysank: This code should have been removed. Fixed it, Thanks.

cloud/scope/machine.go
shysank: Yeah, Agreed. Fixed it.

cloud/scope/machine.go
shysank: Two reasons:
1. Better user experience since a cluster is mostly created with mds, I thought it'd be easier to map an md with an availability set.
2. I wasn't sure if there is a 1-1 mapping b/w md->ms. And it might end up with more than 1 as for the same md, again which I felt might be a bit odd.

Having said that, this will result in machines created with machine set will end up not being HA, so that's a concern. Maybe check for md, and then fallback to ms if not available?

cloud/scope/managedcontrolplane.go
shysank: Done, Thanks.

cloud/scope/machine.go
CecileRobertMichon: > Having said that, this will result in machines created with machine set will end up not being HA, so that's a concern. Maybe check for md, and then fallback to ms if not available?

I think that's okay. Let's keep it simple for now and stick with MachineDeployments only. Would you mind adding a line or two to https://capz.sigs.k8s.io/topics/failure-domains.html to describe the use of availability sets when failure domains are not available though?

CecileRobertMichon:
lgtm pending rebase and docs 

/assign @devigned @nader-ziada

cloud/scope/cluster.go
```diff
@@ -27,8 +27,10 @@ import (
 	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
+
```
devigned, line 30: Why the newline? I believe goimports will group together in order with a space between, stdlib, external imports, local imports.

cloud/scope/cluster.go
```diff
@@ -27,8 +27,10 @@ import (
...
 	"k8s.io/klog/klogr"
 	"k8s.io/utils/net"
+
```
devigned, line 33: Why the newline? I believe goimports will group together in order with a space between, stdlib, external imports, local imports.

cloud/services/availabilitysets/availabilitysets.go
```diff
@@ -81,33 +81,32 @@ func (s *Service) Reconcile(ctx context.Context) error {
...
-		_, err := s.Client.CreateOrUpdate(ctx, s.Scope.ResourceGroup(), asSpec.Name, asParams)
-		if err != nil {
-			return errors.Wrapf(err, "failed to create availability set %s", asSpec.Name)
-		}
-		s.Scope.V(2).Info("successfully created availability set", "availability set", asSpec.Name)
+	s.Scope.V(2).Info("creating availability set", "availability set", availabilitySetName)
+
+	asParams := compute.AvailabilitySet{
+		Sku: &compute.Sku{
+			Name: pointer.StringPtr(string(compute.Aligned)),
```
devigned, line 88: Can we eliminate the import of `pointer`? We already import `to`.

cloud/services/availabilitysets/availabilitysets.go
```diff
@@ -81,33 +81,32 @@ func (s *Service) Reconcile(ctx context.Context) error {
...
+		},
+		AvailabilitySetProperties: &compute.AvailabilitySetProperties{
+			PlatformFaultDomainCount: pointer.Int32Ptr(int32(faultDomainCount)),
```
devigned, line 91: Can we eliminate the import of `pointer`? We already import `to`.

cloud/services/availabilitysets/availabilitysets.go
```diff
@@ -116,19 +115,36 @@ func (s *Service) Delete(ctx context.Context) error {
...
+	if !ok {
+		return nil
 	}
 
+	as, err := s.Client.Get(ctx, s.Scope.ResourceGroup(), availabilitySetName)
+	if err != nil && azure.ResourceNotFound(err) {
+		// already deleted
+		return nil
+	}
+	if err != nil {
```
devigned, line 128: Please add a vertical space between if blocks.

cloud/services/availabilitysets/availabilitysets.go
```diff
@@ -116,19 +115,36 @@ func (s *Service) Delete(ctx context.Context) error {
...
+		return nil
+	}
+
+	s.Scope.V(2).Info("deleting availability set", "availability set", availabilitySetName)
+	err = s.Client.Delete(ctx, s.Scope.ResourceGroup(), availabilitySetName)
+	if err != nil && azure.ResourceNotFound(err) {
+		// already deleted
+		return nil
+	}
+	if err != nil {
```
devigned, line 143: Please add a vertical space between if blocks.

devigned (changes requested):
Just nits

nader-ziada:
looks like all the comments are addressed

/lgtm

devigned (approved):
/lgtm
/assign @CecileRobertMichon

devigned:
/approve

I just noticed that @CecileRobertMichon had already added an lgtm above.

PR #1152, opened 2021-02-01
fix make tilt-up when jaeger crds fail to be found

devigned:
**What type of PR is this?**
/kind bug

When running `make tilt-up` I often run into `kubernetes apply: unable to recognize "": no matches for kind "Jaeger" in version "jaegertracing.io/v1"`. This happens in Tilt when the observability components are applied to the cluster. By applying the Jaeger CRDs first, it seems to remedy the issue of the `Jaeger` kind not being found.

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -134,8 +134,11 @@ def capz():
     # Apply the kustomized yaml for this provider
     substitutions = settings.get("kustomize_substitutions", {})
     os.environ.update(substitutions)
-    yaml = str(kustomizesub("./hack/observability")) # build an observable kind deployment by default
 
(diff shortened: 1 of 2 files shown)


devigned:
/hold

I'm not sure if this is the best way to handle this. I'm going to experiment is a couple other ways.

CecileRobertMichon:
I wonder if this is related to tilt version somehow? I never ran into this before and after having updated to the latest with the new UI I suddenly started running into it

devigned:
> I wonder if this is related to tilt version somehow? 

Perhaps. Either way, I think I'm going to move this to be an extension then we won't have to think about it any longer.

PR #1170, opened 2021-02-10
Switch LB Probe to TCP

### LINKED ISSUE #1102: Loadbalancer Probe uses "/healthz" https endpoint. Suggestion to switch to a different probe.

CecileRobertMichon:
Co-Authored-By: Zach <45979377+zawachte-msft@users.noreply.github.com>

 

 
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**: Rebased #1103. Credit goes to @zawachte-msft.

> Today the load balancer code creates a healthcheck probe which hits the api-server /healthz endpoint.
> Two issues with this.
> The /healthz endpoint has been deprecated since kubernetes v1.16 https://kubernetes.io/docs/reference/using-api/health-checks/
If you have anonymous-auth for the api-server set to false, this breaks cluster-bring-up since the loadbalancer probe will fail with an auth error.
Not well versed in the azure load balancer so I don't know what the proper fix is, but I tested changing to a TCP probe and it works.
> I assume there is a better solution. AWS seems to have a SSL check that doesn't require this endpoint.

Fixes #1102 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Switch LB Probe to TCP
```

### DIFF
--- a/azure/services/loadbalancers/loadbalancers.go
+++ b/azure/services/loadbalancers/loadbalancers.go
@@ -32,7 +32,7 @@ import (
 )
 
 const (
-	httpsProbe  = "HTTPSProbe"
+	tcpProbe    = "TCPProbe"
 	lbRuleHTTPS = "LBRuleHTTPS"
 	outboundNAT = "OutboundNATAllProtocols"
 )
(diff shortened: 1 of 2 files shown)


nader-ziada:
/lgtm

I think it makes sense to make that change, @devigned any thoughts?

devigned (approved):
/lgtm
/approve

PR #1178, opened 2021-02-17
Fix handling of AzureMachine terminal errors

CecileRobertMichon:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**: We were calling Reconcile twice and returning the error on the first call so the handling of terminal error code was never reached. Also made some more errors in virtualmachines terminal (matching what was already done for scale sets). This now accurately puts your AzureMachine and thus Machine in a failed state if you pass in an incorrect VM size or a VM size that is not available in that particular region.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Fix handling of AzureMachine terminal errors
```

### DIFF
--- a/azure/services/networkinterfaces/networkinterfaces.go
+++ b/azure/services/networkinterfaces/networkinterfaces.go
@@ -114,7 +114,7 @@ func (s *Service) Reconcile(ctx context.Context) error {
 				// set accelerated networking to the capability of the VMSize
 				sku, err := s.resourceSKUCache.Get(ctx, nicSpec.VMSize, resourceskus.VirtualMachines)
 				if err != nil {
-					return errors.Wrapf(err, "failed to get find vm sku %s in compute api", nicSpec.VMSize)
+					return azure.WithTerminalError(errors.Wrapf(err, "failed to get SKU %s in compute api", nicSpec.VMSize))
 				}
 
 				accelNet := sku.HasCapability(resourceskus.AcceleratedNetworking)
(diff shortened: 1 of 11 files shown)


controllers/azuremachine_controller.go
```diff
@@ -311,6 +305,10 @@ func (r *AzureMachineReconciler) reconcileNormal(ctx context.Context, machineSco
 
 			if reconcileError.IsTerminal() {
 				machineScope.Error(err, "failed to reconcile AzureMachine", "name", machineScope.Name())
+				machineScope.SetFailureReason(capierrors.CreateMachineError)
+				machineScope.SetFailureMessage(err)
+				machineScope.SetNotReady()
+				machineScope.SetVMState(infrav1.VMStateFailed)
```
nader-ziada, line 311: should we also set a condition here?

azure/services/networkinterfaces/networkinterfaces.go
```diff
@@ -114,7 +114,7 @@ func (s *Service) Reconcile(ctx context.Context) error {
 				// set accelerated networking to the capability of the VMSize
 				sku, err := s.resourceSKUCache.Get(ctx, nicSpec.VMSize, resourceskus.VirtualMachines)
 				if err != nil {
-					return errors.Wrapf(err, "failed to get find vm sku %s in compute api", nicSpec.VMSize)
+					return azure.WithTerminalError(errors.Wrapf(err, "failed to get find SKU %s in compute api", nicSpec.VMSize))
```
nader-ziada, line 117: should the error message say `get find SKU`?

devigned (approved):
I really like the added specificity with terminal errors.

controllers/azuremachine_controller.go
CecileRobertMichon: the condition is already being set on line 304

CecileRobertMichon:
/retest

#1182

PR #1180, opened 2021-02-17
Reconcile FailureDomains in AzureMachinePool

### LINKED ISSUE #663: Support FailureDomains for AzureMachinePools

fiunchinho:
It was not possible to specify the `FailureDomains` when creating a `MachinePool`. This PR tries to reconcile the VMSS created by the `AzureMachinePool` using the selected `FailureDomains`.

Fixes #663 

**Release note**:

```release-note
Reconcile `FailureDomains` in `AzureMachinePool`

```

### DIFF
--- a/azure/services/scalesets/scalesets.go
+++ b/azure/services/scalesets/scalesets.go
@@ -27,7 +27,9 @@ import (
 	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
+
 	"sigs.k8s.io/cluster-api-provider-azure/util/generators"
+	"sigs.k8s.io/cluster-api-provider-azure/util/slice"
 
 	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1alpha3"
 	"sigs.k8s.io/cluster-api-provider-azure/azure"
@@ -306,6 +308,18 @@ func (s *Service) validateSpec(ctx context.Context) error {
 		return azure.WithTerminalError(errors.Errorf("encryption at host is not supported for VM type %s", spec.Size))
 	}
 
+	// Checking if selected availability zones are available selected VM type in location
+	azsInLocation, err := s.resourceSKUCache.GetZonesWithVMSize(ctx, spec.Size, s.Scope.Location())
+	if err != nil {
+		return errors.Wrapf(err, "failed to get zones for VM type %s in location %s", spec.Size, s.Scope.Location())
+	}
+
+	for _, az := range spec.FailureDomains {
+		if !slice.Contains(azsInLocation, az) {
+			return azure.WithTerminalError(errors.Errorf("availability zone %s is not available for VM type %s in location %s", az, spec.Size, s.Scope.Location()))
+		}
+	}
+
 	return nil
 }
 
@@ -364,6 +378,7 @@ func (s *Service) buildVMSSFromSpec(ctx context.Context, vmssSpec azure.ScaleSet
 			Tier:     to.StringPtr("Standard"),
 			Capacity: to.Int64Ptr(vmssSpec.Capacity),
 		},
+		Zones: to.StringSlicePtr(vmssSpec.FailureDomains),
 		VirtualMachineScaleSetProperties: &compute.VirtualMachineScaleSetProperties{
 			UpgradePolicy: &compute.UpgradePolicy{
 				Mode: compute.UpgradeModeManual,
--- a/azure/services/scalesets/scalesets_test.go
+++ b/azure/services/scalesets/scalesets_test.go
@@ -124,6 +124,7 @@ func TestGetExistingVMSS(t *testing.T) {
 				Identity: "",
 				Tags:     nil,
 				Capacity: int64(1),
+				Zones:    []string{"1", "3"},
 				Instances: []infrav1exp.VMSSVM{
 					{
 						ID:         "my-vm-id",
@@ -148,6 +149,7 @@ func TestGetExistingVMSS(t *testing.T) {
 					VirtualMachineScaleSetProperties: &compute.VirtualMachineScaleSetProperties{
 						ProvisioningState: to.StringPtr("Succeeded"),
 					},
+					Zones: &[]string{"1", "3"},
 				}, nil)
 				m.ListInstances(gomock.Any(), "my-rg", "my-vmss").Return([]compute.VirtualMachineScaleSetVM{
 					{
@@ -634,15 +636,16 @@ func TestDeleteVMSS(t *testing.T) {
 func getFakeSkus() []compute.ResourceSku {
 	return []compute.ResourceSku{
 		{
-			Name: to.StringPtr("VM_SIZE"),
-			Kind: to.StringPtr(string(resourceskus.VirtualMachines)),
+			Name:         to.StringPtr("VM_SIZE"),
+			ResourceType: to.StringPtr(string(resourceskus.VirtualMachines)),
+			Kind:         to.StringPtr(string(resourceskus.VirtualMachines)),
 			Locations: &[]string{
 				"test-location",
 			},
 			LocationInfo: &[]compute.ResourceSkuLocationInfo{
 				{
 					Location: to.StringPtr("test-location"),
-					Zones:    &[]string{"1"},
+					Zones:    &[]string{"1", "3"},
 				},
 			},
 			Capabilities: &[]compute.ResourceSkuCapabilities{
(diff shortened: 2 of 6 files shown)


azure/services/scalesets/scalesets.go
```diff
@@ -27,6 +27,7 @@ import (
 	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
+
 	"sigs.k8s.io/cluster-api-provider-azure/util/generators"
```
shysank, line 32: nit: is this vertical space intentional

shysank (commented):
Overall lgtm with just one small nit. 

I was also wondering if there would be any effect on existing machine pools with these changes. If it does not have any effect, should we document this somewhere?

/cc @CecileRobertMichon @nader-ziada

nader-ziada:
> Overall lgtm with just one small nit.
> 
> I was also wondering if there would be any effect on existing machine pools with these changes. If it does not have any effect, should we document this somewhere?
> 
> /cc @CecileRobertMichon @nader-ziada

I'm not sure how the FailureDomain is set on the vmss, the `VirtualMachineScaleSet` `Zones` is that already done somewhere else?

nader-ziada:
> > Overall lgtm with just one small nit.
> > I was also wondering if there would be any effect on existing machine pools with these changes. If it does not have any effect, should we document this somewhere?
> > /cc @CecileRobertMichon @nader-ziada
> 
> I'm not sure how the FailureDomain is set on the vmss, the `VirtualMachineScaleSet` `Zones` is that already done somewhere else?

ok never mind, I see it now, in scalesets.go
```go
Zones: to.StringSlicePtr(vmssSpec.FailureDomains)
```

CecileRobertMichon (commented):
Can we add an example of how to use this in https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/docs/book/src/topics/failure-domains.md?

azure/services/scalesets/scalesets.go
```diff
@@ -364,6 +365,7 @@ func (s *Service) buildVMSSFromSpec(ctx context.Context, vmssSpec azure.ScaleSet
 			Tier:     to.StringPtr("Standard"),
 			Capacity: to.Int64Ptr(vmssSpec.Capacity),
 		},
+		Zones: to.StringSlicePtr(vmssSpec.FailureDomains),
```
CecileRobertMichon, line 368: It might be nice to validate that the zones are valid in that region for that VM size. We can leverage the SKU cache for this, similar to how we check for VM SKU capabilities like ephemeral OS disk support and number of vCPUs

fiunchinho: How we decide when to validate in the webhook vs in the reconciler? Validating the zones in the webhook would be a better UX I think, because the user would get feedback right away.

fiunchinho:
Added usage example to docs.

azure/services/scalesets/scalesets.go
fiunchinho: Not really, goimports added it for me. I'll remove it.

azure/services/scalesets/scalesets.go
CecileRobertMichon: Whenever we can we do validation in the webhooks but for anything that requires calling out to Azure (and SKU validation is one of them since it varies per sub and evolves over time) we do it in the reconciler

Agree the UX is better in the webhook but if you have to wait 2 minutes for the webhook to get a response from Azure APIs it becomes pretty bad UX

fiunchinho:
> Overall lgtm with just one small nit.
> 
> I was also wondering if there would be any effect on existing machine pools with these changes. If it does not have any effect, should we document this somewhere?
> 
> /cc @CecileRobertMichon @nader-ziada

The `zones` property can't be changed after the creation of the VMSS. It can only be set when creating the VMSS. If an existing `MachinePool` has values in the `FailureDomains`  field, the reconciliation will fail because those values were ignored until now, meaning that the VMSS got created without availability zones.

docs/book/src/topics/failure-domains.md
```diff
@@ -130,6 +130,56 @@ spec:
 
 ```
 
+### Using Virtual Machine Scale Sets
+
+You can use an `AzureMachinePool` object to deploy a Virtual Machine Scale Set which automatically distributes VM instances across availability zones.
```
CecileRobertMichon, line 135: let's make it more clear that you as a user need to set the available failure domains that you want to spread to in the MachinePool spec so it's not completely "autmatically". You can also use `az vm list-skus -l <location> --zone -o table` to list all the available zones per vm size in that location.

fiunchinho:
> > Overall lgtm with just one small nit.
> > I was also wondering if there would be any effect on existing machine pools with these changes. If it does not have any effect, should we document this somewhere?
> > /cc @CecileRobertMichon @nader-ziada
> 
> The `zones` property can't be changed after the creation of the VMSS. It can only be set when creating the VMSS. If an existing `MachinePool` has values in the `FailureDomains` field, the reconciliation will fail because those values were ignored until now, meaning that the VMSS got created without availability zones.

Actually, I'm double checking the code, and I believe changes in the `MachinePool` `FailureDomains` won't be sent to the Azure API, so nothing would happen: no zones changed but also no error. I believe that because the SDK [won't take the `zones` field into consideration](https://github.com/Azure/azure-sdk-for-go/blob/master/services/compute/mgmt/2020-06-01/compute/models.go#L12465-L12522), so no patch for that field will be generated.

shysank:
Thanks for looking into it. Maybe we just add a small note in the pr description that says something like "Machine pools created before this change will not be affected"

docs/book/src/topics/failure-domains.md
fiunchinho: I changed it, please take a look.

azure/services/scalesets/scalesets.go
fiunchinho: Added AZ validation.

devigned (approved):

devigned:
```
Unexpected call to *mock_scalesets.MockScaleSetScope.Location([]) at /home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/azure/services/scalesets/mock_scalesets/scalesets_mock.go:294 because: 
            expected call at /home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/azure/services/scalesets/scalesets_test.go:1020 has already been called the max number of times 
```

Probably just need to adjust the mock a little.

azure/services/scalesets/scalesets.go
```diff
@@ -646,3 +659,13 @@ func base64EncodedHash(vmss compute.VirtualMachineScaleSet) (string, error) {
 	_, _ = hasher.Write(jsonData)
 	return base64.URLEncoding.EncodeToString(hasher.Sum(nil)), nil
 }
+
+// contains tells whether a contains x.
+func contains(a []string, x string) bool {
```
CecileRobertMichon, line 664: consider moving this helper to the common util package

fiunchinho: Done.

azure/services/scalesets/scalesets_test.go
```diff
@@ -635,14 +637,15 @@ func getFakeSkus() []compute.ResourceSku {
 	return []compute.ResourceSku{
 		{
 			Name: to.StringPtr("VM_SIZE"),
+			ResourceType: to.StringPtr(string(resourceskus.VirtualMachines)),
```
fiunchinho, line 640: Without specifying the `ResourceType`, the capability was not found in the tests.

CecileRobertMichon:
`azure/services/scalesets/scalesets_test.go:71: File is not `gofmt`-ed with `-s` (gofmt)`

test file needs to be gofmt'ed

util/slice/slice.go
```diff
@@ -0,0 +1,27 @@
+/*
+Copyright 2020 The Kubernetes Authors.
```
CecileRobertMichon, line 2:
```suggestion
Copyright 2021 The Kubernetes Authors.
```

nader-ziada:
I think everything looks good now

lgtm

CecileRobertMichon:
Thanks @fiunchinho!

/lgtm
/approve

@nader-ziada let's forward port this to alpha4 once the types merge

CecileRobertMichon:
/retest

investigating the registry issue. retest in case it was transient.

PR #1183, opened 2021-02-18
Accept cluster names that start with a number

### LINKED ISSUE #1179: Clusters names starting with a number char are not allowed by the webhook
/kind bug

Trying to create an `AzureCluster` CR named `1a2b3c` is not allowed, the webhook will prevent it, failing with 

```Cluster Name doesn't match regex ^[a-z][a-z0-9-]{0,44}[a-z0-9]$, can contain only lowercase alphanumeric characters and '-', must start/end with an alphanumeric character```

But the explanation and the regex don't match. While the explanation says names "_must start/end with an alphanumeric character_", the regex doesn't allow the cluster name to start with a number. I don't see any reason why a cluster name wouldn't start with a number, so I guess this is a bug in the regex. I can submit a PR changing the regex to `^[a-z0-9][a-z0-9-]{0,44}[a-z0-9]$` but I wanted to run it by you first.

When creating a cluster named ``1a2b3c``, cluster validation in admission controller webhook is successful and cluster is created.

**Environment:**

### PR DESCRIPTION
**Pull request by fiunchinho**
**What type of PR is this?**

/kind bug

Trying to create an AzureCluster CR named "1a2b3c" is not allowed. The webhook will prevent it failing with
```
Cluster Name doesn't match regex ^[a-z][a-z0-9-]{0,44}[a-z0-9]$, can contain only lowercase alphanumeric characters and '-', must start/end with an alphanumeric character
```

This PR makes the error message and the validation to match, allowing names that start with a number.

Fixes #1179 

**TODOs**:

- [X] squashed commits
- [X] includes documentation
- [X] adds unit tests

**Release note**:

```release-note
Names that start with a number are valid `AzureCluster` names.
```

### DIFF
--- a/api/v1alpha3/azurecluster_validation.go
+++ b/api/v1alpha3/azurecluster_validation.go
@@ -29,7 +29,7 @@ import (
 const (
 	// can't use: \/"'[]:|<>+=;,.?*@&, Can't start with underscore. Can't end with period or hyphen.
 	// not using . in the name to avoid issues when the name is part of DNS name
-	clusterNameRegex = `^[a-z][a-z0-9-]{0,44}[a-z0-9]$`
+	clusterNameRegex = `^[a-z0-9][a-z0-9-]{0,42}[a-z0-9]$`
 	// max length of 44 to allow for cluster name to be used as a prefix for VMs and other resources that
 	// have limitations as outlined here https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules
 	clusterNameMaxLength = 44
--- a/api/v1alpha3/azurecluster_validation_test.go
+++ b/api/v1alpha3/azurecluster_validation_test.go
@@ -67,6 +67,11 @@ func TestClusterNameValidation(t *testing.T) {
 			wantErr:     true,
 		},
 		{
+			name:        "cluster name starting with number",
+			clusterName: "1clustername",
+			wantErr:     false,
+		},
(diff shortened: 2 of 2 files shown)


api/v1alpha3/azurecluster_validation.go
```diff
@@ -29,7 +29,7 @@ import (
 const (
 	// can't use: \/"'[]:|<>+=;,.?*@&, Can't start with underscore. Can't end with period or hyphen.
 	// not using . in the name to avoid issues when the name is part of DNS name
-	clusterNameRegex = `^[a-z][a-z0-9-]{0,44}[a-z0-9]$`
+	clusterNameRegex = `^[a-z0-9][a-z0-9-]{0,44}[a-z0-9]$`
```
CecileRobertMichon, line 32: let's add a unit test case that validates this change

fiunchinho: done

shysank (commented):
/lgtm with a small nit

api/v1alpha3/azurecluster_validation.go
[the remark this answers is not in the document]
fiunchinho: Nice catch! Fixed

fiunchinho:
I believe tests failures are not related to the change.

devigned (approved):

CecileRobertMichon:
/retest

#1186 just merged

PR #1186, opened 2021-02-20
do not include windows password in hash

### LINKED ISSUE #1182: Windows VMSS E2E flakes
/kind bug

https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/1178/pull-cluster-api-provider-azure-e2e-windows/1362121987211661312

```
E0217 19:56:17.869588       1 azuremachinepool_controller.go:290] controllers/AzureMachinePool "msg"="failed to reconcile AzureMachinePool" "error"="failed to create scale set: failed to get VMSS win-p-win after create or update: failed to get result from future: reconcile error occurred that can be recovered. Object will be requeued after 15s The actual error is: operation type PATCH on Azure resource capz-e2e-odyqry/win-p-win is not done" "AzureCluster"="capz-e2e-odyqry" "azureMachinePool"="capz-e2e-odyqry-mp-win" "cluster"="capz-e2e-odyqry" "machinePool"="capz-e2e-odyqry-mp-win" "namespace"="create-workload-cluster-heu6os" "name"="win-p-win"
```
job history: https://prow.k8s.io/job-history/gs/kubernetes-jenkins/pr-logs/directory/pull-cluster-api-provider-azure-e2e-windows

**Environment:**

jsturtevant:
**What type of PR is this?**
/kind failing-test

Windows generates a password each time we build the vmss model from the spec which causes the hash to be different and causes the model to continuously update.

Fixes #1182

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/azure/services/scalesets/scalesets.go
+++ b/azure/services/scalesets/scalesets.go
@@ -637,6 +637,14 @@ func getSecurityProfile(vmssSpec azure.ScaleSetSpec, sku resourceskus.SKU) (*com
 
 // base64EncodedHash transforms a VMSS into json and then creates a sha256 hash of the data encoded as a base64 encoded string
 func base64EncodedHash(vmss compute.VirtualMachineScaleSet) (string, error) {
+	// Setting Admin Password is not supported but an initial password is required for Windows
+	// Don't include it in the hash since it is generated and won't be the same each the spec is created (#1182)
+	tmpPass := vmss.VirtualMachineProfile.OsProfile.AdminPassword
+	vmss.VirtualMachineProfile.OsProfile.AdminPassword = nil
+	defer func() {
+		vmss.VirtualMachineProfile.OsProfile.AdminPassword = tmpPass
+	}()
+
 	jsonData, err := vmss.MarshalJSON()
 	if err != nil {
 		return "", errors.Wrapf(err, "failed marshaling vmss")
--- a/azure/services/scalesets/scalesets_test.go
+++ b/azure/services/scalesets/scalesets_test.go
@@ -298,6 +298,23 @@ func TestReconcileVMSS(t *testing.T) {
 			},
 		},
 		{
+			name:          "Windows VMSS should not get patched",
+			expectedError: "",
+			expect: func(g *WithT, s *mock_scalesets.MockScaleSetScopeMockRecorder, m *mock_scalesets.MockClientMockRecorder) {
+				defaultSpec := newWindowsVMSSSpec()
+				s.ScaleSetSpec().Return(defaultSpec).AnyTimes()
+				createdVMSS := newDefaultWindowsVMSS()
+				instances := newDefaultInstances()
+				createdVMSS = setupDefaultVMSSInProgressOperationDoneExpectations(g, s, m, createdVMSS, instances)
+				s.SetProviderID(fmt.Sprintf("azure://%s", *createdVMSS.ID))
+				s.SetLongRunningOperationState(nil)
+				s.SetProvisioningState(infrav1.VMStateSucceeded)
+				s.NeedsK8sVersionUpdate().Return(false)
+				infraVMSS := converters.SDKToVMSS(createdVMSS, instances)
+				s.UpdateInstanceStatuses(gomockinternal.AContext(), infraVMSS.Instances).Return(nil)
+			},
+		},
+		{
 			name:          "should start creating vmss with defaulted accelerated networking when size allows",
 			expectedError: "failed to get VMSS my-vmss after create or update: failed to get result from future: operation type PUT on Azure resource my-rg/my-vmss is not done",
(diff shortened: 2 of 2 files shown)


azure/services/scalesets/scalesets.go
```diff
@@ -637,6 +637,14 @@ func getSecurityProfile(vmssSpec azure.ScaleSetSpec, sku resourceskus.SKU) (*com
 
 // base64EncodedHash transforms a VMSS into json and then creates a sha256 hash of the data encoded as a base64 encoded string
 func base64EncodedHash(vmss compute.VirtualMachineScaleSet) (string, error) {
+	// Setting Admin Password is not supported but an initial password is required for Windows
+	// Don't include it in the hash since it is generated and won't be the same each the spec is created (#1182)
+	tmpPass := vmss.VirtualMachineProfile.OsProfile.AdminPassword
+	vmss.VirtualMachineProfile.OsProfile.AdminPassword = nil
+	defer func() {
+		vmss.VirtualMachineProfile.OsProfile.AdminPassword = tmpPass
```
CecileRobertMichon, line 645: do we really need to set this value again here? the function param isn't a pointer so it shouldn't modify the actual value of the vmss outside the function no?

devigned: FYI: Hash calculation goes away in MachinePoolMachines. The reasoning is that we need to do a get and a diff each time. There is little gain in having a hash if you can just compare actual state. The downside is that we have to be able to transform Azure state to CAPZ Model state (compute.VMSS -> azure.VMSS), which we don't aways do for resources.

I might be nice to be able to have converters or enforce conversion from Azure model to CAPZ model.

devigned: I think this might be needed. `compute.VirtualMachineScaleSet` contains an embedded ptr to `*VirtualmachineScaleSetProperties`, so the ptr value is copied and points to the same memory even though it looks like the param should not modify the caller.

devigned (approved):
I don't love this solution, but it does fix the Windows VMSS issue. I'd prefer to see a password only get set on create, rather than on update too.

In the upcoming MachinePoolMachine changes, the hash is deprecated in favor of a model diff. I do fear this issue will still be a problem if the OsProfile continues to mutate.

/lgtm
/assign @CecileRobertMichon

devigned:
Giving another run to Windows e2e to verify stability.

jsturtevant:
Looks like the still getting the same error:

```
E0222 15:54:50.084730       1 azuremachinepool_controller.go:290] controllers/AzureMachinePool "msg"="failed to reconcile 
AzureMachinePool" "error"="failed to create scale set: failed to get VMSS win-p-win after create or update: failed to get result from 
future: reconcile error occurred that can be recovered. Object will be requeued after 15s The actual error is: operation type PUT 
on Azure resource capz-e2e-m0owx5/win-p-win is not done" "AzureCluster"="capz-e2e-m0owx5" "azureMachinePool"="capz-e2e-
m0owx5-mp-win" "cluster"="capz-e2e-m0owx5" "machinePool"="capz-e2e-m0owx5-mp-win" "namespace"="create-workload-
cluster-t817cm" "name"="win-p-win"
```

devigned:
> Looks like the still getting the same error:
> 
> ```
> E0222 15:54:50.084730       1 azuremachinepool_controller.go:290] controllers/AzureMachinePool "msg"="failed to reconcile 
> AzureMachinePool" "error"="failed to create scale set: failed to get VMSS win-p-win after create or update: failed to get result from 
> future: reconcile error occurred that can be recovered. Object will be requeued after 15s The actual error is: operation type PUT 
> on Azure resource capz-e2e-m0owx5/win-p-win is not done" "AzureCluster"="capz-e2e-m0owx5" "azureMachinePool"="capz-e2e-
> m0owx5-mp-win" "cluster"="capz-e2e-m0owx5" "machinePool"="capz-e2e-m0owx5-mp-win" "namespace"="create-workload-
> cluster-t817cm" "name"="win-p-win"
> ```  

Just to be clear, @jsturtevant this "error" is expected when an Azure operation is taking places. It should stop when the VMSS hits a terminal state, `Failed` or `Succeeded`. If the VMSS continues to be patched for some reason, the error will likely persist since the VMSS never reaches a terminal state.

CecileRobertMichon:
> I don't love this solution, but it does fix the Windows VMSS issue. I'd prefer to see a password only get set on create, rather than on update too.
> 
> In the upcoming MachinePoolMachine changes, the hash is deprecated in favor of a model diff. I do fear this issue will still be a problem if the OsProfile continues to mutate.

Agreed. Let's follow up on this.

PR #1188, opened 2021-02-22
✨ Add support to reconcile subnet's Nat Gateway

### LINKED ISSUE #894: Nat gateway support in SubnetSpec

fiunchinho:
**What type of PR is this?**

/kind feature

Adds new field to SubnetSpec so that users can specify the Nat Gateway to use when creating subnets. When Nat Gateway is created, it will be used for outbound traffic of the nodes, instead of a load balancer.

Fixes #894

The Nat Gateway can be used for the outbound IP of the nodes, replacing the load balancer that we were using until now. I tried to make the code backwards compatible, so unless a Nat Gatewat name is specified, everything should keep working the same way.

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Release note**:

```release-note
Added new field to `SubnetSpec` so that users can specify the Nat Gateway to use when creating subnets.

```


azure/services/subnets/subnets.go
```diff
@@ -99,6 +99,12 @@ func (s *Service) Reconcile(ctx context.Context) error {
 				}
 			}
 
+			if subnetSpec.NatGatewayName != "" {
```
fiunchinho, line 102: Nat Gateways are reconciled before subnets. When reconciling subnets, if a nat gateway was specified, then it's safe to use its Azure ID.

azure/defaults.go
```diff
@@ -80,6 +80,11 @@ func GenerateFrontendIPConfigName(lbName string) string {
 	return fmt.Sprintf("%s-%s", lbName, "frontEnd")
 }
 
+// GenerateNatGatewayIPConfigName generates a nat gateway IP config name.
+func GenerateNatGatewayIPName(clusterName, subnetName string) string {
+	return fmt.Sprintf("pip-%s-%s-natgw", clusterName, subnetName)
```
fiunchinho, line 85: Not sure if two different subnets in two different clusters can have the same name, so to be sure I'm using the cluster name.

azure/scope/cluster.go
```diff
@@ -128,12 +128,19 @@ func (s *ClusterScope) PublicIPSpecs() []azure.PublicIPSpec {
 		}
 	}
 
-	return []azure.PublicIPSpec{
-		controlPlaneOutboundIP,
-		{
-			Name: azure.GenerateNodeOutboundIPName(s.ClusterName()),
-		},
+	var publicIPSpec []azure.PublicIPSpec
```
shysank, line 131: `publicIPSpecs` instead of `publicIPSpec`

azure/services/natgateways/natgateways.go
```diff
@@ -0,0 +1,123 @@
...
+					Location: to.StringPtr(s.Scope.Location()),
+					NatGatewayPropertiesFormat: &network.NatGatewayPropertiesFormat{
+						PublicIPAddresses: &[]network.SubResource{
+							{
+								ID: to.StringPtr(natGatewaySpec.NatGatewayIP.Name),
+							},
+						},
+					},
+				},
+			)
```
shysank, line 90: Since we are not using the returned value from Get, can we just not directly do `createOrupdate` and handle the error?

controllers/azurecluster_reconciler.go
```diff
@@ -101,6 +104,10 @@ func (s *azureClusterService) Reconcile(ctx context.Context) error {
 		return errors.Wrapf(err, "failed to reconcile route table")
 	}
 
+	if err := s.natGatewaySvc.Reconcile(ctx); err != nil {
```
shysank, line 107: Shouldn't this be reconciled after public ip? With the current reconcile ordering, this will be conflicting because subnet is reconciled before public ip.

shysank: I think you forgot to add reconcile for delete.

api/v1alpha3/types.go
```diff
@@ -120,6 +120,14 @@ type RouteTable struct {
 	Name string `json:"name,omitempty"`
 }
 
+// NatGateway defines an Azure Nat Gateway.
+// NAT gateway resources are part of Vnet NAT and provide outbound Internet connectivity for subnets of a virtual network.
+type NatGateway struct {
+	ID           string       `json:"id,omitempty"`
```
shysank, line 126: Is this used anywhere?

shysank (changes requested):

azure/scope/cluster.go
fiunchinho: Replaced.

azure/services/natgateways/natgateways.go
fiunchinho: I changed the approach a bit, please take a look.

controllers/azurecluster_reconciler.go
fiunchinho: Right, I re-ordered the services and added the delete reconciler.

azure/services/natgateways/client.go
```diff
@@ -0,0 +1,96 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
shysank, line 2:
```suggestion
Copyright 2021 The Kubernetes Authors.
```

azure/services/natgateways/natgateways.go
```diff
@@ -0,0 +1,148 @@
...
+func (s *Service) getExisting(ctx context.Context, spec azure.NatGatewaySpec) (*infrav1.NatGateway, error) {
+	existingNatGateway, err := s.Get(ctx, s.Scope.ResourceGroup(), spec.Name)
+	if err != nil {
+		if azure.ResourceNotFound(err) {
+			return nil, err
+		}
+		return nil, errors.Wrapf(err, "failed to get Nat Gateway %s", spec.Name)
+	}
+
+	return &infrav1.NatGateway{
```
shysank, line 116: Can we move this logic to converters?

azure/services/natgateways/natgateways.go
```diff
@@ -0,0 +1,148 @@
...
+
+	for _, natGatewaySpec := range s.Scope.NatGatewaySpecs() {
+		existingNatGateway, err := s.getExisting(ctx, natGatewaySpec)
+
+		switch {
+		case err != nil && !azure.ResourceNotFound(err):
+			return errors.Wrapf(err, "failed to get nat gateway %s in %s", natGatewaySpec.Name, s.Scope.ResourceGroup())
+		case err == nil:
+			// nat gateway already exists, leave it alone
+			if natGatewaySpec.Subnet.Role == infrav1.SubnetControlPlane {
```
shysank, line 73: Can we have explicitly setters in `NatGatewayScope` for setting Id and public ip address so that we can keep track of things that we are mutating?

fiunchinho: Sorry, I'm not sure I understand what you mean. Do you have an example somewhere of what you mean?

azure/services/natgateways/natgateways.go
fiunchinho: I've seen converters used for common things like tags or SKU, but most scopes don't use them?

azure/services/natgateways/natgateways.go
shysank: Something like this https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/2d58abba334517ad59a3f2c7f6d8dd476bfb0f49/azure/services/virtualmachines/virtualmachines.go#L96?

shysank:
@fiunchinho can you rebase and migrate the api changes to `v1alpha4`. Please let me know if you need any help on this.

azure/scope/cluster.go
```diff
@@ -151,7 +151,18 @@ func (s *ClusterScope) PublicIPSpecs() []azure.PublicIPSpec {
...
+	var nodeNatGatewayIPSpecs []azure.PublicIPSpec
+	nodeNatGateway := s.NodeNatGateway()
+	if nodeNatGateway.Name != "" {
+		nodeNatGatewayIPSpecs = append(nodeNatGatewayIPSpecs, azure.PublicIPSpec{
+			Name:    s.GetNodeNatGatewayIPName(),
+			DNSName: nodeNatGateway.NatGatewayIP.DNSName,
+		})
+	}
+
 	publicIPSpecs = append(publicIPSpecs, nodeOutboundIPSpecs...)
```
CecileRobertMichon, line 164: move this up to line 153?

azure/scope/cluster.go
```diff
@@ -193,6 +204,8 @@ func (s *ClusterScope) LBSpecs() []azure.LBSpec {
 		})
 	}
 
+	// When cluster is not private it already has a public LB attached for the API server so it doesn't need an
+	// additional LB to get outbound traffic on those nodes.
```
CecileRobertMichon, line 208:
```suggestion
	// additional LB to get outbound traffic on control plane nodes.
```

azure/services/natgateways/natgateways.go
```diff
@@ -0,0 +1,141 @@
...
+// New creates a new service.
+func New(scope NatGatewayScope) *Service {
+	return &Service{
+		Scope:  scope,
+		client: newClient(scope),
+	}
+}
+
+// Reconcile gets/creates/updates a nat gateway.
+// Only when the Nat Gateway 'Name' property is defined we create the Nat Gateway: it's opt-in. Should we use the nat gateway ip name instead?
```
CecileRobertMichon, line 55: is this a question for reviewers?

azure/services/natgateways/natgateways.go
```diff
@@ -0,0 +1,141 @@
...
+		switch {
+		case err != nil && !azure.ResourceNotFound(err):
+			return errors.Wrapf(err, "failed to get nat gateway %s in %s", natGatewaySpec.Name, s.Scope.ResourceGroup())
+		case err == nil:
+			// nat gateway already exists
+			s.Scope.V(2).Info("nat gateway already exists", "nat gateway", natGatewaySpec.Name)
+			s.Scope.SetNodeNatGateway(*existingNatGateway)
+		default:
+			// nat gateway doesn't exist but its name was specified in the subnet, let's create it
+			s.Scope.V(2).Info("creating nat gateway", "nat gateway", natGatewaySpec.Name)
```
CecileRobertMichon, line 77: are any of the fields in the nat gateway expected to be mutable? if so we want to CreateOrUpdate on each reconcile instead of just if it doesn't exist. If we expect to create it once and never update it after that the current implementation is fine.

azure/scope/cluster.go
```diff
@@ -382,6 +425,19 @@ func (s *ClusterScope) NodeRouteTable() infrav1.RouteTable {
...
+func (s *ClusterScope) NodeNatGateway() infrav1.NatGateway {
+	subnet, _ := s.AzureCluster.Spec.NetworkSpec.GetNodeSubnet()
+	return subnet.NatGateway
+}
+
+// SetNodeNatGateway sets the node subnet into the scope with the updated nat gateway.
+func (s *ClusterScope) SetNodeNatGateway(natGateway infrav1.NatGateway) {
+	nodeSubnet := s.NodeSubnet()
+	nodeSubnet.NatGateway = natGateway
+	s.AzureCluster.Spec.NetworkSpec.UpdateNodeSubnet(nodeSubnet)
```
CecileRobertMichon, line 438: what's the difference between this and SetSubnet added in https://github.com/kubernetes-sigs/cluster-api-provider-azure/commit/4cff58a1e31902f79dedff1b7ff8332af3e67997 ? can we consolidate the two if they are the same?

azure/services/natgateways/natgateways.go
```diff
@@ -0,0 +1,141 @@
...
+	existingNatGateway, err := s.Get(ctx, s.Scope.ResourceGroup(), spec.Name)
+	if err != nil {
+		return nil, err
+	}
+
+	return &infrav1.NatGateway{
+		ID:   to.String(existingNatGateway.ID),
+		Name: to.String(existingNatGateway.Name),
+		NatGatewayIP: infrav1.PublicIPSpec{
+			Name: to.String((*existingNatGateway.PublicIPAddresses)[0].ID),
```
CecileRobertMichon, line 113: Is there any useful information that we're setting here that's not already in the spec? is it just the ID of the NAT Gateway?

CecileRobertMichon (commented):
Looks good overall. Let's also add some docs to the book and squash commits.

azure/services/natgateways/natgateways.go
fiunchinho: nope, it was for my past self. Removed on next commit.

azure/services/natgateways/natgateways.go
fiunchinho: The public IP created for the nat gateway could be modified. I will change the code to call `CreateOrUpdate` on each reconcile.

azure/scope/cluster.go
fiunchinho: I changed it to use the new function.

azure/services/natgateways/natgateways.go
fiunchinho: Yeah, only the ID of the NAT and ID of the Public IP.

fiunchinho:
I added docs about the nat gateway.

azure/services/natgateways/natgateways.go
```diff
@@ -0,0 +1,153 @@
...
+		case err != nil && !azure.ResourceNotFound(err):
+			return errors.Wrapf(err, "failed to get nat gateway %s in %s", natGatewaySpec.Name, s.Scope.ResourceGroup())
+		case err == nil:
+			// nat gateway already exists
+			s.Scope.V(4).Info("nat gateway already exists", "nat gateway", natGatewaySpec.Name)
+			s.Scope.SetNodeNatGateway(*existingNatGateway)
+
+			var update bool
+			if existingNatGateway.NatGatewayIP.Name != natGatewaySpec.NatGatewayIP.Name {
+				update = true
```
devigned, line 78: I don't see update being used elsewhere in the code block. Perhaps, move the `!update` block into here and invert the if statement.

azure/services/natgateways/natgateways.go
```diff
@@ -0,0 +1,153 @@
...
+		switch {
+		case err != nil && !azure.ResourceNotFound(err):
+			return errors.Wrapf(err, "failed to get nat gateway %s in %s", natGatewaySpec.Name, s.Scope.ResourceGroup())
+		case err == nil:
+			// nat gateway already exists
+			s.Scope.V(4).Info("nat gateway already exists", "nat gateway", natGatewaySpec.Name)
+			s.Scope.SetNodeNatGateway(*existingNatGateway)
+
+			var update bool
+			if existingNatGateway.NatGatewayIP.Name != natGatewaySpec.NatGatewayIP.Name {
```
devigned, line 77: I don't think this is possible. `existingNatGateway` is fetched from Azure by the spec `Name`. If there is no error during the request and a resource of that `Name` is returned, the `Name` must match the spec `Name`. 

Is there some case here you are guarding against?

devigned (changes requested):
Couple reconcile items.

docs/book/src/topics/node-outbound-nat-gateway.md
```diff
@@ -0,0 +1,42 @@
+# Node Outbound Load Balancer
```
CecileRobertMichon, line 1: Nat Gateway?

CecileRobertMichon: actually should we combine this doc with the node outbound LB one and have one "Node Outbount" document presenting the two options?

azure/services/natgateways/natgateways.go
fiunchinho: Notice that it's the public IP name, not the Nat Gateway name. This is here in case users want to change the Public IP used for the nat gateway. They could change the Public IP in the CR, then the Nat Gateway needs to be updated accordingly. Does that make sense?

docs/book/src/topics/node-outbound-nat-gateway.md
fiunchinho: I tried merging the two. Please take a look.

azure/services/natgateways/natgateways.go
fiunchinho: Done.

azure/services/natgateways/natgateways.go
devigned: My mistake. You are correct. Thank you for clarifying.

azure/services/natgateways/natgateways.go
```diff
@@ -0,0 +1,148 @@
...
+	for _, natGatewaySpec := range s.Scope.NatGatewaySpecs() {
+		existingNatGateway, err := s.getExisting(ctx, natGatewaySpec)
+
+		switch {
+		case err != nil && !azure.ResourceNotFound(err):
+			return errors.Wrapf(err, "failed to get nat gateway %s in %s", natGatewaySpec.Name, s.Scope.ResourceGroup())
+		case err == nil:
+			// nat gateway already exists
+			s.Scope.V(4).Info("nat gateway already exists", "nat gateway", natGatewaySpec.Name)
+			s.Scope.SetNodeNatGateway(*existingNatGateway)
```
devigned, line 74: Should this also get called in the case of the NAT Gateway does not exist and needs to be created?

fiunchinho: I don't know TBH. I saw that it's only done when it already exists for other resources. What should we do?

devigned: I believe what will happen is that the initial reconcile of `AzureCluster` will create the Nat Gateway, but when the reconciliation is finished, the `AzureCluster` resource will not have the Nat Gateway populated on the `AzureCluster` resource and will be saved in: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/27cd557e53306d15857d975c8835f9b641a13771/azure/scope/cluster.go#L531-L534

On the second reconciliation, the Nat Gateway will be fetched from Azure and populated via `SetNodeNatGateway`, then it will be persisted to the `AzureCluster` spec when `Close()` is called again.

I've been using a defer to do this elsewhere to ensure we set scope items that are later persisted in `scope.Close()`. For example, scale sets:
https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/27cd557e53306d15857d975c8835f9b641a13771/azure/services/scalesets/scalesets.go#L92-L105

azure/scope/cluster.go
```diff
@@ -193,6 +203,8 @@ func (s *ClusterScope) LBSpecs() []azure.LBSpec {
 		})
 	}
 
+	// When cluster is not private it already has a public LB attached for the API server so it doesn't need an
+	// additional LB to get outbound traffic on control plane nodes.
```
CecileRobertMichon, line 207: thank you 👍

azure/defaults.go
CecileRobertMichon: the subnet name should already have the clusterName as part of its name by default, and if it doesn't (eg. custom name) two subnets with the same name can't be in the same RG so I think it's okay to remove clusterName from here.

docs/book/src/topics/node-outbound-nat-gateway.md
CecileRobertMichon: looks good

azure/services/natgateways/natgateways.go
devigned: sgtm

fiunchinho:
Let me know if there is anything else I need to do. Thanks.

CecileRobertMichon:
@fiunchinho you need to rebase to fix the docs lint check

@devigned @shysank do you have any unresolved comments?

fiunchinho:
Rebased

azure/defaults.go
CecileRobertMichon: @fiunchinho did you want to update this?

CecileRobertMichon (commented):
/lgtm
/approve

PR #1191, opened 2021-02-23
add azure machine pool machine proposal

devigned:
**What type of PR is this?**
/kind design

The scope of work for implementing safe rolling upgrades for AzureMachinePool with MaxUnavailable and MaxSurge has grown to warrant a proposal. This proposal provides background, tradeoffs and the design strategy to explain why we want to break out AzureMachinePoolMachines from AzureMachinePools.

Related to #1105 

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Add AzureMachinePoolMachine proposal which describes the design strategy for implementing rolling upgrades for AzureMachinePools
```

### DIFF
--- a/docs/proposals/20210222-azure-machinepool-machine.md
+++ b/docs/proposals/20210222-azure-machinepool-machine.md
@@ -1,0 +1,417 @@
+---
+title: Azure Machine Pool Machines
+authors:
+    - @devigned
+reviewers:
+    - @CecileRobertMichon
+    - @nader-ziada
+creation-date: 2021-02-22
+last-updated: 2021-02-22
+status: implementable
+see-also:
+    - https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/819
+    - https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1067
+---
+
+
+# Azure Machine Pool Machines
+
+## Table of Contents
+- [Summary](#summary)
+- [Motivation](#motivation)
+    - [Goals](#goals)
+    - [Non-Goals / Future Work](#non-goals--future-work)
+        - [Notes About VMSS Terminate Notifications](#notes-about-vmss-terminate-notifications)
+- [Proposal](#proposal)
+    - [User Stories](#user-stories)
+        - [Story 1 - Upgrading the Kubernetes Version of a MachinePool](#story-1---upgrading-the-kubernetes-version-of-a-machinepool)
+        - [Story 2 - Reducing the Number of Replicas in a MachinePool](#story-2---reducing-the-number-of-replicas-in-a-machinepool)
+        - [Story 3 - Deleting an individual Azure Machine Pool Machine](#story-3---deleting-an-individual-azure-machine-pool-machine)
+    - [Requirements](#requirements)
+        - [Functional](#functional)
+        - [Non-Functional](#non-functional)
+    - [Implementation Details/Notes/Constraints](#implementation-detailsnotesconstraints)
+        - [Existing APIs for Clarity](#existing-apis-for-clarity)
+        - [Proposed API Changes](#proposed-api-changes)
+        - [Proposed Controllers Changes](#proposed-controller-changes)
+        - [Proposed Changes of Responsibily](#proposed-changes-of-responsibility)
+- [Available Options](#available-options-for-cluster-api-provider-azure)
+    - [Add Annotations to AzureMachinePool for Instance Delete Selection](#option-1-add-annotations-to-azuremachinepool-for-instance-delete-selection)
+        - [Pros](#option-1-pros)
+        - [Cons](#option-1-cons)
+    - [Separate AzureMachinePool and AzureMachinePoolMachines](#option-2-separate-azuremachinepool-and-azuremachinepoolmachines)
+        - [Pros](#option-2-pros)
+        - [Cons](#option-2-cons)
+- [Conclusions](#conclusions)
+- [Additional Details](#additional-details)
+    - [Test Plan](#test-plan)
+- [Implementation History](#implementation-history)    
+
+## Summary
+
+Azure MachinePool currently embeds the state of each of the instances in the MachinePool within the status of the Azure
+MachinePool. MachinePool instances should be their own resources to enable individual lifecycles.
+
+## Motivation
+
+By giving each AzureMachinePoolMachine an individual lifecycle, a user would be able to inform CAPZ of the specific
+instance to delete and then have the AzureMachinePoolMachine controller cordon and drain the node prior to deleting
+the underlying infrastructure.
+
+### Goals
+- Be able to delete specific AzureMachinePool instances
+- Rolling updates with max unavailable and max surge
+  - MaxUnavailable is the max number of machines that are allowed to be unavailable at any time
+  - MaxSurge is the number of machines to surge, add to the current replica count, during an upgrade of the VMSS model 
+- Safely update by cordoning and draining nodes prior to deleting the underlying infrastructure
+- Be able to take advantage of [Azure's Virtual Machine Scale Set Update Instance API](https://docs.microsoft.com/en-us/rest/api/compute/virtualmachinescalesets/updateinstances)
+  to in-place update a VMSS instance rather than delete and recreate the infrastructure, which would result in a much
+  quicker upgrade.
+
+### Non-Goals / Future Work
+- Create a CAPI Machine owner for each AzureMachinePoolMachine
+- Implementing different roll out and scale down strategies
+- Adopting individual Machine instances to be managed by the MachinePool
+- Create or use an on instance agent to cordon and drain in response to Azure Virtual Machine Scale Sets provide [terminate notifications](https://docs.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-terminate-notification)
+
+#### Notes About VMSS Terminate Notifications
+Azure Virtual Machine Scale Sets provide [terminate notifications](https://docs.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-terminate-notification).
+These terminate notifications would be helpful to inform Kubernetes when a node is going to be deleted. Unfortunately,
+terminate notifications do not provide notifications when an instance is Updated, in this case "Updated" means the
+instance is reimaged to match the updated VMSS model by using the [Update Instance API](https://docs.microsoft.com/en-us/rest/api/compute/virtualmachinescalesets/updateinstances). 
+If a VMSS instance were to be reimaged, rather than deleted and recreated the instance will not receive a notification. 
+Due to the design of terminate notifications the CAPZ controller needs to alert Kubernetes when an instance is being 
+Updated. Without some way to inform Kubernetes of the specific instance that is to be updated, the underlying 
+infrastructure may be removed before workloads can be safely migrated from the machine / node. By managing the lifecycle 
+from CAPZ, we are able to safely delete / upgrade machines / nodes.
+
+In the future, it would be useful to integrate [awesomenix/drainsafe](https://github.com/awesomenix/drainsafe) or 
+something similar to handle scenarios when Azure will delete or migrate a VMSS instance. Two scenarios come to mind.
+
+1. VMSS is configured to use [Spot instances](https://docs.microsoft.com/en-us/azure/virtual-machines/spot-vms) and 
+   Azure must evict an instance.
+2. Azure must [perform maintenance on an instance](https://docs.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-maintenance-notifications).    
+
+## Proposal
+
+### User Stories
+
+#### Story 1 - Upgrading the Kubernetes Version of a MachinePool
+Alex is an engineer in a large organization which has a MachinePool running 1.18.x and would like to upgrade the 
+MachinePool 1.19.x. It is important to Alex that the MachinePool doesn't experience downtime during the upgrade. Alex
+has set the MaxUnavailable and MaxSurge values on the AzureMachinePool to limit the number of machines that will be 
+unavailable during the upgrade, and the number of extra machines VMSS will add during upgrade. The MachinePool
+upgrades each machine in the pool by first cordoning and draining, then replacing the machine in the pool.
+
+#### Story 2 - Reducing the Number of Replicas in a MachinePool
+Alex is an engineer in a large organization which has a MachinePool running. Alex has too many nodes running on the 
+cluster and would like to reduce the replicas. It is important to Alex that the MachinePool doesn't experience downtime. 
+Alex decreases the replica count of the MachinePool by 2. The MachinePool deletes 2 machines from the pool by first 
+cordoning and draining, then deleting the underlying infrastructure.
+
+#### Story 3 - Deleting an individual Azure Machine Pool Machine
+Alex is an engineer in a large organization which has a MachinePool running with 5 replicas. Alex would like to delete a 
+specific MachinePool machine. It is important to Alex that the MachinePool doesn't experience downtime while deleting 
+the individual machine. Alex uses `kubectl` to delete the specific MachinePool machine resource. The MachinePool machine
+is cordoned and drained, then the underlying infrastructure is deleted. The MachinePool still has a replica count of 5,
+but only has 4 running replicas. The MachinePool creates a new machine to take the place of the deleted instance.
+
+
+### Requirements
+
+#### Functional
+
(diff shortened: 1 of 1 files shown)


docs/proposals/20210222-azure-machinepool-machine.md
```diff
@@ -0,0 +1,332 @@
...
+
+```go
+// AzureMachinePoolSpec defines the desired state of AzureMachinePool
+AzureMachinePoolSpec struct {
+
+    /*
+       Other fields omitted for brevity
+    */
+	
+    // MaxUnavailable is the max number of replicas which can be unavailable at any time. For example, if the
```
CecileRobertMichon, line 167: My immediate reaction when seeing these new types is "why put them in CAPZ? there is nothing Azure infra specific about it". I know there are nuances and adding these types directly in CAPI is not straight forward because it then implies different providers would need to honor these values by either a) passing them through to the infra properties if the Cloud/environment supports them or b) implement it themselves (what we're doing in CAPZ). 

I worry that adding these here now will make it hard to move them later since any refactor of the types to CAPI types would be either a breaking change. And the alternative of leaving them in here beyond exp/ is that we create diverging implementations and, more importantly, inconsistent user experience for MachinePool across providers.

If we do go down this path, I suggest detailing this in a risks & mitigations section in this doc so that we can justify this design choice down the road when someone inevitably asks "why was this not added to CAPI types?"

docs/proposals/20210222-azure-machinepool-machine.md
```diff
@@ -0,0 +1,332 @@
...
+    //
+    // Note: Azure Subscriptions have a limited quota of CPUs. This quota can be expanded upon request. Due to the
+    // nature of this feature, spikes due to surge may cause you to exceed quota.
+    //
+    // See Also: https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/azure-subscription-service-limits
+    MaxSurge *int32 `json:"maxSurge,omitempty"`
+}
+
+// AzureMachinePoolMachineSpec defines the desired state of AzureMachinePoolMachine
+AzureMachinePoolMachineSpec struct {
```
CecileRobertMichon, line 189: How do we make it clear to the users that they are not responsible for creating the AzureMachinePoolMachines?

docs/proposals/20210222-azure-machinepool-machine.md
```diff
@@ -0,0 +1,332 @@
...
+
+**AzureMachinePool Responsibilities:**
+- Create AzureMachinePoolMachine instances when a new VMSS instance is observed. The AzureMachinePoolMachine spec should
+  have the `ProviderID` field set with the observed resource ID. The AzureMachinePool should also be added to the
+  AzureMachinePoolMachine's OwnerReferences. 
+- Upgrade / replace selection of AzureMachinePoolMachine instances. When a change to the AzureMachinePool model occurs,
+  the MachinePoolScope will be responsible for coordinating the rollout of the updated model by selecting
+  AzureMachinePoolMachines to delete / replace with respect to MaxUnavailable.
+- Scale up. AzureMachinePool should increase the number of VMSS replicas if the replica count increases on MachinePool
+- Scale down. AzureMachinePool should select and delete AzureMachinePoolMachines that are overprovisioned with respect
```
CecileRobertMichon, line 277: > should select

how does selection happen? at random? oldest first? newest first?

docs/proposals/20210222-azure-machinepool-machine.md
```diff
@@ -0,0 +1,332 @@
...
+  - MaxUnavailable is the max number of machines that are allowed to be unavailable at any time
+  - MaxSurge is the number of machines to surge, add to the current replica count, during an upgrade to 
+- Safely update by cordoning and draining nodes prior to deleting the underlying infrastructure
+
+### Non-Goals / Future Work
+- Create a CAPI Machine owner for each AzureMachinePoolMachine (possible future work)
+
+## Proposal
+
+### User Stories
```
CecileRobertMichon, line 76: I'd love to see a user story for deleting a specific instance in a scale set

docs/proposals/20210222-azure-machinepool-machine.md
```diff
@@ -0,0 +1,332 @@
...
+- Create AzureMachinePoolMachine instances when a new VMSS instance is observed. The AzureMachinePoolMachine spec should
+  have the `ProviderID` field set with the observed resource ID. The AzureMachinePool should also be added to the
+  AzureMachinePoolMachine's OwnerReferences. 
+- Upgrade / replace selection of AzureMachinePoolMachine instances. When a change to the AzureMachinePool model occurs,
+  the MachinePoolScope will be responsible for coordinating the rollout of the updated model by selecting
+  AzureMachinePoolMachines to delete / replace with respect to MaxUnavailable.
+- Scale up. AzureMachinePool should increase the number of VMSS replicas if the replica count increases on MachinePool
+- Scale down. AzureMachinePool should select and delete AzureMachinePoolMachines that are overprovisioned with respect
+  to MaxUnavailable.
+- Clean up. When a AzureMachinePoolMachine is no longer in the list of instances in Azure, but a matching
```
CecileRobertMichon, line 279: This implies Azure is the source of truth and the goal state. But at the same time, we're saying you can trigger a VMSS VM delete by deleting an AzureMachinePoolMachine which kind of goes against that assumption. IMO the CRs should be the source of truth / goal state, and the controller's job should be to reconcile against that goal state and try to get the real world (Azure resources) to match. Now if something happens out of band to delete a VMSS instance, the controller should see that and then our options are either 1) recreate it (that's what we do for all resources excepts VMs right now) or, and this is probably the option we need to pick if we're going to follow the immutable machines principle, 2) mark the AzureMachinePoolMachine as "Failed" and have some external remediation mechanism (equivalent to a MachineHealthCheck) who deletes failed machines - which should cause a new instance to be created to replace it since the desired replica count will become > than the actual number of replicas.

docs/proposals/20210222-azure-machinepool-machine.md
```diff
@@ -0,0 +1,332 @@
...
+  AzureMachinePoolMachine resource exists, delete the AzureMachinePoolMachine.
+  
+**AzureMachinePoolMachine Responsibilities:**
+- Update Azure Provisioning State: when creating a new VMSS instance, the AzureMachinePoolMachine controller will poll 
+  the Azure API until the instance reaches a terminal state.
+- Cordon and Drain: when deleting the AzureMachinePoolMachine resource, the AzureMachinePoolMachine controller is
+  responsible for ensuring workload is moved from the node prior to removing the underlying Azure infrastructure.
+- NodeRef: as a VMSS instance joins the cluster, the AzureMachinePoolMachine controller is responsible for ensuring
```
CecileRobertMichon, line 287: The other behavior we need to define is here is what happens when I, a user, go and delete an AzureMachinePoolMachine? Does it get replaced by a new one (similar to MachineDeployment Machine behavior)? Or does the replica count get decreased by one (similar to VMSS delete instance behavior).

CecileRobertMichon (commented):
Thank you so much for taking the time to write this down. It helped me understand better what the goals are in #1105, and I'm sure it will help future contributors and users as well.

Overall this looks good to me but I challenged some of the assumptions / implied design choices that might not be obvious to others. My main concern is around the delete scenario and how the sync works between Azure resource deletions and k8s CR deletions.

docs/proposals/20210222-azure-machinepool-machine.md
devigned: That's a great question. Do you have any suggestions?

My first inclination is to add documentation.

There really isn't anything that stops someone from creating an AMPM. It won't hurt anything.

docs/proposals/20210222-azure-machinepool-machine.md
devigned: Great observation. I totally agree. It'd be great to remove these from the AMP spec and move them to MP.

With https://github.com/kubernetes-sigs/cluster-api/pull/4073 nearly merged, I'll open up a PR to add a `RolloutStrategy` and `DeleteStrategy` to `MachinePool`.

docs/proposals/20210222-azure-machinepool-machine.md
```diff
@@ -0,0 +1,332 @@
...
+    */
+	
+    // MaxUnavailable is the max number of replicas which can be unavailable at any time. For example, if the
+    // machine pool has a replica count of 5 and a MaxUnavailable value of 2, a rolling upgrade will never exceed
+    // more than 2 machines unavailable at a time.
+    //
+    // Note: If the MaxUnavailable value is equal to the machine pool replica value, there is
+    // the possibility of application down time.
+    // +kubebuilder:default=1
+    MaxUnavailable int32 `json:"maxUnavailable,omitempty"`
```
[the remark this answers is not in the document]
devigned: Yes, they should.

docs/proposals/20210222-azure-machinepool-machine.md
```diff
@@ -0,0 +1,332 @@
...
+Azure MachinePool currently embeds the state of each of the machines in the MachinePool within the status of the Azure
+MachinePool. MachinePool machines should be their own resources to enable individual lifecycles.
+
+## Motivation
+
+By giving each AzureMachinePoolMachine an individual lifecycle, a user would be able to inform CAPZ of the specific
+instance to delete and then have the AzureMachinePoolMachine controller cordon and drain the node prior to deleting
+the underlying infrastructure.
+
+Azure Virtual Machine Scale Set Instances do not provide the specific machine events needed to alert Kubernetes when
```
fiunchinho, line 58: Can't we leverage the [Azure scheduled events](https://docs.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-terminate-notification)?

nader-ziada:
Thank you @devigned for this doc, very helpful in understanding how all this works

docs/proposals/20210222-azure-machinepool-machine.md
devigned: Imagine when we over-provision VMs. Azure will spin up multiple, wait for the first to successfully deploy, then delete the others that didn't win the provisioning race. In this case, the loser VMs will be deleted by Azure. We must clean up the AMPMs which may have been created while the provisioning race was running.

docs/proposals/20210222-azure-machinepool-machine.md
devigned: > Terminate notifications only on ‘delete’ operations – All delete operations (manual delete or Autoscale-initiated scale-in) will generate Terminate events if your scale set has scheduledEventsProfile enabled. Other operations such as reboot, reimage, redeploy, and stop/deallocate do not generate Terminate events. 

We could if we decide to only delete and replace machines, which will cause infra to deprovision and then provision again. Another way we could do this is to use `Reimage`, which does not deprovision the infra, but only updates the model. In `Reimage`, we would not get an event from Azure letting the machine know that it should cordon and drain.

On the other hand, if we are using spot instances, we will likely need to listen to machine events to know when a node must cordon and drain.

This thread covers the problem pretty well: #819. I do think we will need some on host component like https://github.com/awesomenix/drainsafe at some point to help with spot instances or VM migration.

docs/proposals/20210222-azure-machinepool-machine.md
devigned: Added.

docs/proposals/20210222-azure-machinepool-machine.md
devigned: Added to take into account the DeletePolicy.

docs/proposals/20210222-azure-machinepool-machine.md
devigned: Right now, the behavior is that the replica count stays the same. When an VMSS instance is deleted, the next time the AMP reconciler runs, it sees Azure with 1 less capacity than the MachinePool replica count. It PATCHs the capacity of the VMSS to be the replica count of MachinePool, which increases the capacity by 1. 

I think that's not what we want though, right? I think we want the capacity to decrease in that case, right?

CecileRobertMichon: > I think that's not what we want though, right? I think we want the capacity to decrease in that case, right?

To me, it makes sense to keep the replica count the same for the use case in user story #3. It's also more consistent with the MachineDeployment behavior. My comment is mostly that we should be explicit about that design choice, whether it's that or decreasing capacity.

docs/proposals/20210222-azure-machinepool-machine.md
CecileRobertMichon: That makes sense

docs/proposals/20210222-azure-machinepool-machine.md
CecileRobertMichon: How is that action plan tracked in the proposal?

docs/proposals/20210222-azure-machinepool-machine.md
fiunchinho: Oh I see. Thanks for the clarification. Still, I'd reword it a bit because it seems VMSS do provide machine events when an instance is deleted.

docs/proposals/20210222-azure-machinepool-machine.md
fiunchinho: It looks like [`AWSMachinePool` in CAPA contains similar fields](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/17c3144c5f0060526a005eb5a2927db22fcd259a/exp/api/v1alpha3/awsmachinepool_types.go#L82-L99). Not sure if it will be possible to move them to `MachinePool` as different cloud providers use very different mechanisms (ASG vs VMSS).

CecileRobertMichon: That's a good point @fiunchinho, it might make sense to keep them in cloud providers then. Let's document this as an intentional choice in this proposal.

fiunchinho:
Looking at how [the upgrade process is implemented in KCP](https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/proposals/20191017-kubeadm-based-control-plane.md#kubeadmcontrolplane-rollout) we could try to make the two look even more similar. There are two fields that I believe could be useful here:
- `nodeDrainTimeout` introduced in KCP [here](https://github.com/kubernetes-sigs/cluster-api/pull/3662), it allows the user to specify the amount of time we want to wait for a node to be drained. If capz is taking care of draining, I believe it should allow users to specify this.
- `upgradeAfter` introduced in KCP [here](https://github.com/kubernetes-sigs/cluster-api/pull/2238), is a field to indicate an upgrade should be performed after the specified time, to allow users to schedule upgrades.

What do you think?

devigned:
> upgradeAfter introduced in KCP here, is a field to indicate an upgrade should be performed after the specified time, to allow users to schedule upgrades.

@fiunchinho I don't think that upgradeAfter makes sense for AzureMachinePool since if there are no changes to the underlying model, then there will be nothing to upgrade.

I added the `NodeDrainTimeout`. Thank you!

docs/proposals/20210222-azure-machinepool-machine.md
devigned: @fiunchinho wdyt of the termination notifications wording now?

fiunchinho:
> > upgradeAfter introduced in KCP here, is a field to indicate an upgrade should be performed after the specified time, to allow users to schedule upgrades.
> 
> @fiunchinho I don't think that upgradeAfter makes sense for AzureMachinePool since if there are no changes to the underlying model, then there will be nothing to upgrade.
> 
> I added the `NodeDrainTimeout`. Thank you!

Totally with you. I miss understood the `upgradeAfter` field. Thanks.

docs/proposals/20210222-azure-machinepool-machine.md
fiunchinho: Perfect.

docs/proposals/20210222-azure-machinepool-machine.md
```diff
@@ -0,0 +1,417 @@
...
+    // The deployment strategy to use to replace existing machines with
+    // new ones.
+    // +optional
+    Strategy MachineDeploymentStrategy `json:"strategy,omitempty"`
+
+	// NodeDrainTimeout is the total amount of time that the controller will spend on draining a node.
+	// The default value is 0, meaning that the node can be drained without any time limitations.
+	// NOTE: NodeDrainTimeout is different from `kubectl drain --timeout`
+	// +optional
+	NodeDrainTimeout *metav1.Duration `json:"nodeDrainTimeout,omitempty"`
```
CecileRobertMichon, line 264: nit: indent

devigned: fixed

CecileRobertMichon (commented):
/lgtm

Issue #1190, opened 2021-02-22
> I don't love this solution, but it does fix the Windows VMSS issue. I'd prefer to see a password only get set on create, rather than on update too.


CecileRobertMichon:
> I don't love this solution, but it does fix the Windows VMSS issue. I'd prefer to see a password only get set on create, rather than on update too.
> 
> In the upcoming MachinePoolMachine changes, the hash is deprecated in favor of a model diff. I do fear this issue will still be a problem if the OsProfile continues to mutate.

Agreed. Let's follow up on this.

_Originally posted by @CecileRobertMichon in https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1186#issuecomment-783635800_

jsturtevant:
agreed it isn't ideal.  Right now Cloudbase-init is configured to reset the password on the machine to a random password on first boot but the VM does need to be provided a temp password for creation.  This was to make it so the customers don't have to have a "magic" windows password which they have to manage out of band.

Considered a few couple other options in the PR to fix the hash generation:

- Introduce a Windows Password to the spec to store the password.  
  - Decided against this because we wanted to avoid encouraging "magic" windows password, plus it is reset via cloudbase init.
- Generate the temp password deterministically based of some properties of the template. 
  - Though only temporary and unlikely it could be used, thought it was better to have a random password for that temporary time period.

PR #1197, opened 2021-02-24
do not include customData in AzureMachinePool hash calculation

devigned:
**What type of PR is this?**
/kind bug

AzureMachinePool will occasionally enter into a cycle where the hash tag calculated from the last model update will be different than the hash calculated for a newly reconciled AzureMachinePool since the bootstrap data secret has been updated since the previous reconcile causing the customData of the VMSS to change, thus changing the hash for the entire model.

This PR removes the customData from the hash calculation and increases test controller log verbosity.

Related: #1190, #1186, #1182

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Do not include VMSS customData in the hash tag calculation to correct a reconcile cycle caused when the bootstrap token refreshes.
```

### DIFF
--- a/azure/services/scalesets/scalesets.go
+++ b/azure/services/scalesets/scalesets.go
@@ -210,9 +210,12 @@ func (s *Service) patchVMSSIfNeeded(ctx context.Context, infraVMSS *infrav1exp.V
 		//
 		// Note: if a user were to mutate the VMSS in Azure rather than through CAPZ, this hash match may match, but not
 		// reflect the state of the specification in K8s.
+		s.Scope.V(2).Info("found matching spec hash no need to PATCH")
 		return nil, nil
 	}
 
+	s.Scope.V(2).Info("hashes don't match PATCHING VMSS", "oldHash", infraVMSS.Tags[infrav1.SpecVersionHashTagKey()], "newHash", result.Hash)
+	s.Scope.V(4).Info("diff", "oldVMSS", infraVMSS, "newVMSS", result)
 	vmss := result.VMSSWithoutHash
 	vmss.Tags = converters.TagsToMap(result.Tags.AddSpecVersionHashTag(result.Hash))
 	patch, err := getVMSSUpdateFromVMSS(vmss)
@@ -655,9 +658,13 @@ func base64EncodedHash(vmss compute.VirtualMachineScaleSet) (string, error) {
 	// Setting Admin Password is not supported but an initial password is required for Windows
 	// Don't include it in the hash since it is generated and won't be the same each the spec is created (#1182)
 	tmpPass := vmss.VirtualMachineProfile.OsProfile.AdminPassword
+	// Don't include customData in the hash since it will change due to the kubeadm bootstrap token being regenerated.
+	tmpCustomData := vmss.VirtualMachineProfile.OsProfile.CustomData
 	vmss.VirtualMachineProfile.OsProfile.AdminPassword = nil
+	vmss.VirtualMachineProfile.OsProfile.CustomData = nil
 	defer func() {
 		vmss.VirtualMachineProfile.OsProfile.AdminPassword = tmpPass
+		vmss.VirtualMachineProfile.OsProfile.CustomData = tmpCustomData
 	}()
 
 	jsonData, err := vmss.MarshalJSON()
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -24,6 +24,7 @@ spec:
             - --leader-elect
             - "--metrics-bind-addr=127.0.0.1:8080"
             - "--feature-gates=MachinePool=${EXP_MACHINE_POOL:=false},AKS=${EXP_AKS:=false}"
+            - "--v=0"
           image: controller:latest
           imagePullPolicy: Always
           name: manager
--- a/test/e2e/config/azure-dev.yaml
+++ b/test/e2e/config/azure-dev.yaml
@@ -75,6 +75,9 @@ providers:
       targetName: "cluster-template-windows.yaml"
     - sourcePath: "${PWD}/templates/test/cluster-template-prow-machine-pool-windows.yaml"
       targetName: "cluster-template-machine-pool-windows.yaml"
+    replacements:
+    - old: "--v=0"
+      new: "--v=4"
 
 variables:
   KUBERNETES_VERSION: "${KUBERNETES_VERSION:-v1.19.7}"


azure/services/scalesets/scalesets.go
```diff
@@ -452,6 +453,15 @@ func (s *Service) buildVMSSFromSpec(ctx context.Context, vmssSpec azure.ScaleSet
...
 
 	vmss.Tags = converters.TagsToMap(tagsWithoutHash)
+
+	if vmss.VirtualMachineProfile.OsProfile.CustomData != nil {
+		cloneCustomData := *vmss.VirtualMachineProfile.OsProfile.CustomData
+		vmss.VirtualMachineProfile.OsProfile.CustomData = nil
+		defer func() {
+			vmss.VirtualMachineProfile.OsProfile.CustomData = to.StringPtr(cloneCustomData)
+		}()
+	}
```
jsturtevant, line 463: should this be in same place as the exclusion for the windows pass?

devigned: Yes, indeed. Moving it there. Thank you.

devigned:
Pass 1, retry

jsturtevant:
Pass 2, retry

devigned:
Pass 2 was successful

Pass 3, retry

jsturtevant:
Pass 3 was successful

Pass 4, retry

devigned:
Pass 4 was successful

Pass 5, retry

devigned:
Pass 5 was partially successful with Windows failing with an unrelated error:
```
make[1]: *** [Makefile:249: /home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/hack/tools/bin/kubectl-v1.19.2] Error 56
```

Pass 6, retry

devigned:
Pass 6 was partially successful with Windows with VMSS error. Rebased on master to get the latest from v1alpha4 and hopefully cluster data from test clusters that failed to come up.

devigned:
Pass 7 was successful

Pass 8, retry

devigned:
Pass 8 was successful

Pass 9, retry

jsturtevant:
Pass 9 failed with 

e[1]: *** [Makefile:249: /home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/hack/tools/bin/kubectl-v1.19.2] Error 56

again.

Looking that looks like it is related to curl having a network issue.  created #1202 

Pass 10, retry

devigned:
Pass 11, retry

devigned:
Pass 11 success

Pass 12, retry

devigned:
Pass 12 success

Pass 13, retry

azure/services/scalesets/scalesets.go
```diff
@@ -655,9 +655,12 @@ func base64EncodedHash(vmss compute.VirtualMachineScaleSet) (string, error) {
 	// Setting Admin Password is not supported but an initial password is required for Windows
 	// Don't include it in the hash since it is generated and won't be the same each the spec is created (#1182)
 	tmpPass := vmss.VirtualMachineProfile.OsProfile.AdminPassword
+	tmpCustomData := vmss.VirtualMachineProfile.OsProfile.CustomData
 	vmss.VirtualMachineProfile.OsProfile.AdminPassword = nil
+	vmss.VirtualMachineProfile.OsProfile.CustomData = nil
 	defer func() {
 		vmss.VirtualMachineProfile.OsProfile.AdminPassword = tmpPass
+		vmss.VirtualMachineProfile.OsProfile.CustomData = tmpCustomData
```
CecileRobertMichon, line 663: how do this affect scale up? if we're not giving the new bootstrap token to the model how do new nodes join after the initial token has expired?

devigned: - If the replica count increases on MachinePool, then the AMP will PATCH with the updated token and capacity b/c it was caused by an action other than the token changing.
- If the model (image data, identities, etc) changes on MachinePool / AMP, then the VMSS will be PATCH'd with the updated model and token.

Does this make sense? Am I missing a scenario? I was concerned about this possibility too.

To be clear, I don't love this approach.

devigned:
Pass 13 success

Pass 14, retry

devigned:
Pass 15 success

Pass 16, retry

devigned:
Pass 16 success. [Only one PUT and no PATCHs. The hashes matched each time.](https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/1197/pull-cluster-api-provider-azure-e2e-windows/1366713879182184448/artifacts/clusters/bootstrap/controllers/capz-controller-manager/capz-controller-manager-5db558fdd8-6v5zg/manager.log)

Pass 17, retry

devigned:
Pass 17 success.

Pass 18, retry

azure/services/scalesets/scalesets.go
CecileRobertMichon: Ah I see, this func only calculates the diff to determine if a PATCH is required, but it doesn't actually exclude these things from the actual content of the PATCH that is sent if there is a diff. Makes sense to me 👍 

I don't love the approach either but I think it's acceptable if it's very well commented so we don't look at this code later and wonder exactly the above.

devigned:
Pass 18 success.

Pass 19, retry

devigned:
Pass 19 Fail; IPv6

Pass 20, retry

devigned:
Pass 20 success

Pass 21, retry

devigned:
Pass 21 success

Pass 22, retry

devigned:
Pass 22 success

@nader-ziada, @CecileRobertMichon I think this is looking pretty solid. wdyt?

nader-ziada:
looks good to go! passed most of the times

/lgtm

azure/services/scalesets/scalesets.go
```diff
@@ -655,9 +658,12 @@ func base64EncodedHash(vmss compute.VirtualMachineScaleSet) (string, error) {
 	// Setting Admin Password is not supported but an initial password is required for Windows
 	// Don't include it in the hash since it is generated and won't be the same each the spec is created (#1182)
 	tmpPass := vmss.VirtualMachineProfile.OsProfile.AdminPassword
+	tmpCustomData := vmss.VirtualMachineProfile.OsProfile.CustomData
```
CecileRobertMichon, line 661: let's update the comment above to explain why we also don't include custom data

devigned: Fixed.

config/manager/manager.yaml
```diff
@@ -24,6 +24,7 @@ spec:
             - --leader-elect
             - "--metrics-bind-addr=127.0.0.1:8080"
             - "--feature-gates=MachinePool=${EXP_MACHINE_POOL:=false},AKS=${EXP_AKS:=false}"
+            - "--v=2"
```
CecileRobertMichon, line 27: was this checked in by mistake?

devigned: I increased verbosity of the test logging. To make it easier, I add an explicit verbosity to the controller.

CecileRobertMichon: isn't 1 the default though? if so this increases verbosity for all default clusterctl deployments

devigned: [Looks like the default is 0.](https://pkg.go.dev/k8s.io/klog#pkg-overview) I did not intend on increasing the default. I'll change it.

test/e2e/config/azure-dev.yaml
```diff
@@ -75,6 +75,9 @@ providers:
       targetName: "cluster-template-windows.yaml"
     - sourcePath: "${PWD}/templates/test/cluster-template-prow-machine-pool-windows.yaml"
       targetName: "cluster-template-machine-pool-windows.yaml"
+    replacements:
+    - old: "--v=2"
```
CecileRobertMichon, line 79:
```suggestion
    - old: "--v=0"
```

devigned: Great catch!! Fixed

CecileRobertMichon (approved):
/lgtm
/approve

nothing was changed in the last few commits that should affect the stability of the fix (added a code comment and changed default verbosity) so we're still good to merge this

PR #1204, opened 2021-03-01
fix set network for kind

nader-ziada:
**What type of PR is this?**

/kind bug

- put back set kind network to bridge that got removed with the upgrade to newer kind

Fixes #

- also updated capi to latest nightly 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -450,7 +450,7 @@ create-management-cluster: $(KUSTOMIZE) $(ENVSUBST)
 
 	# Deploy CAPI
 	# Using cluster-api nightly build until 0.4.x is released.
-	curl --retry $(CURL_RETRIES) -sSL https://storage.googleapis.com/artifacts.k8s-staging-cluster-api.appspot.com/components/nightly_master_20210210/cluster-api-components.yaml | $(ENVSUBST) | kubectl apply -f -
+	curl --retry $(CURL_RETRIES) -sSL https://storage.googleapis.com/artifacts.k8s-staging-cluster-api.appspot.com/components/nightly_master_20210301/cluster-api-components.yaml | $(ENVSUBST) | kubectl apply -f -
 
 	# Deploy CAPZ
 	kind load docker-image $(CONTROLLER_IMG)-$(ARCH):$(TAG) --name=capz
--- a/scripts/kind-with-registry.sh
+++ b/scripts/kind-with-registry.sh
@@ -35,11 +35,7 @@ kind_version=$(kind version)
 kind_network='kind'
 reg_name='kind-registry'
 reg_port='5000'
-case "${kind_version}" in
-  "kind v0.7."* | "kind v0.6."* | "kind v0.5."*)
-    kind_network='bridge'
-    ;;
-esac
+kind_network='bridge'
 
 # create registry container unless it already exists
 running="$(docker inspect -f '{{.State.Running}}' "${reg_name}" 2>/dev/null || true)"
--- a/Tiltfile
+++ b/Tiltfile
@@ -16,7 +16,7 @@ settings = {
     "deploy_cert_manager": True,
     "preload_images_for_kind": True,
     "kind_cluster_name": "capz",
-    "capi_version": "nightly_master_20210210",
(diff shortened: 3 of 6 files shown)


nader-ziada:
@devigned can you take a look, setting the kind network resolved my tilt issues

scripts/kind-with-registry.sh
```diff
@@ -36,7 +36,7 @@ kind_network='kind'
 reg_name='kind-registry'
 reg_port='5000'
 case "${kind_version}" in
-  "kind v0.7."* | "kind v0.6."* | "kind v0.5."*)
+  "kind v0.10."* | "kind v0.9."*)
```
nprokopic, line 39: How future-proof is this, would it need to be updated when kind v0.11.0 comes out?

nader-ziada: yeah i think we should set it without the if condition

devigned (approved):
`make tilt-up` works for me with this PR after getting past the first run where the Jaeger CRDs and Jaeger resource are in a race.

Makefile
```diff
@@ -450,7 +450,7 @@ create-management-cluster: $(KUSTOMIZE) $(ENVSUBST)
 
 	# Deploy CAPI
 	# Using cluster-api nightly build until 0.4.x is released.
-	curl --retry $(CURL_RETRIES) -sSL https://storage.googleapis.com/artifacts.k8s-staging-cluster-api.appspot.com/components/nightly_master_20210210/cluster-api-components.yaml | $(ENVSUBST) | kubectl apply -f -
+	curl --retry $(CURL_RETRIES) -sSL https://storage.googleapis.com/artifacts.k8s-staging-cluster-api.appspot.com/components/nightly_master_20210301/cluster-api-components.yaml | $(ENVSUBST) | kubectl apply -f -
```
devigned, line 453: Any way we can do this programmatically rather than hard coding?

nader-ziada: I was thinking we should have a tag called `nightly_latest` or something like that, but then you never know if some type changed and could break things

how would be do this programmatically, we don't know the last successful tag, or are you thinking we can check latest in gcr?

devigned: I wasn't sure on the how more questioning if there is a how. 

We know the format. Could we try today, then say 10 of the previous days to find the first that is successful? Fail if we don't find one.

scripts/kind-with-registry.sh
nprokopic: Found this kind PR https://github.com/kubernetes-sigs/kind/pull/1538/files. The `KIND_EXPERIMENTAL_DOCKER_NETWORK` environment variable was added in kind `v0.8.0`.

So I believe here we should check if kind version is >= `0.8.0`. Then if we got kind >= `0.8.0`, and `KIND_EXPERIMENTAL_DOCKER_NETWORK` is set, use whatever network name is set in `KIND_EXPERIMENTAL_DOCKER_NETWORK` environment variable (`bridge` here), otherwise use default network name (`kind`).

nprokopic: Or we set kind `v0.8.0` as a minimum kind version (should be fine IMHO since `v0.10.0` is the latest), and just use `KIND_EXPERIMENTAL_DOCKER_NETWORK` for network name if it is set, otherwise fallback to default network name `kind`.

Makefile
CecileRobertMichon: I think there is value in pinning to a specific version so we can deliberately update the version when we're ready to ingest new changes, especially since changes tend to be breaking these days.

devigned: Moar GitHub points!

devigned (approved):
/lgtm

Makefile
CecileRobertMichon: > Moar GitHub points!

does that mean you concur @devigned? 😄

devigned: Indeed. Seems reasonable enough to me. We'll just have to keep ourselves honest and keep moving forward as we can.

PR #1205, opened 2021-03-02
Fix AKS cluster provisioning errors

### LINKED ISSUE #847: Errors during AKS cluster provisioning
/kind bug

- run `EXP_MACHINE_POOL=true EXP_AKS=true make tilt-up`
- click on "worker-aks" in Tilt UI to deploy an AKS managed cluster
- during reconciliation, the following errors are seen repeatedly, but the infra eventually reconciles successfully
```
[manager] I0730 16:18:44.872984       9 azuremanagedmachinepool_reconciler.go:87] controllers/AzureManagedMachinePool "msg"="reconciling machine pool" "infraPool"="agentpool1" "namespace"="default" "ownerCluster"="aks-template" 
[manager] E0730 16:18:47.836734       9 controller.go:248] controller-runtime/controller "msg"="Reconciler error" "error"="error creating AzureManagedMachinePool default/agentpool1: failed to reconcile machine pool agentpool1: failed to create or update agent pool: failed to begin operation: containerservice.AgentPoolsClient#CreateOrUpdate: Failure sending request: StatusCode=0 -- Original Error: autorest/azure: Service returned an error. Status=\u003cnil\u003e \u003cnil\u003e" "controller"="azuremanagedmachinepool" "name"="agentpool1" "namespace"="default"
```

I expect in the normal case there would be no errors in the capz-controller logs when deploying an AKS managed cluster

I'm betting this is caused by the `AzureManagedMachinePool`s calling `CreateOrUpdate` to create a new agent pool prior to the successful, initial creation of the AKS cluster.

Discovered during: #798

nprokopic:
**What type of PR is this?**

/kind bug

Currently when AKS cluster is being created there are many Azure API errors, where some occur almost on every cluster creation, while other occur only every now and then (since Azure API is called from multiple controllers from their own reconciliation loops, so it depends in which order the API calls are made).

Here are some of the examples.

_1. Error when trying to create or update an additional non-system agent pool, which happens when the AKS cluster is still not created_

Error from CAPZ controller logs:
```
error creating AzureManagedMachinePool default/<agenpool name>:
  failed to reconcile machine pool <agenpool name>:
    failed to create or update agent pool:
      failed to begin operation:
        containerservice.AgentPoolsClient#CreateOrUpdate: Failure sending request: StatusCode=0
          -- Original Error: autorest/azure: Service returned an error. Status=\u003cnil\u003e \u003cnil\u003e
```

_2. Error when trying to create or update AKS cluster while agent pool is being created_

Error from Azure activity log (truncated for the sake of readability):

```
{
    "authorization": {
        "action": "Microsoft.ContainerService/managedClusters/write",
        "scope": "/subscriptions/<SUBSCRIPTION_ID>/resourceGroups/<CLUSTER NAME>/providers/Microsoft.ContainerService/managedClusters/<CLUSTER NAME>"
    },
    "channels": "Operation",
    "eventName": {
        "value": "EndRequest",
        "localizedValue": "End request"
    },
    "category": {
        "value": "Administrative",
        "localizedValue": "Administrative"
    },
    "level": "Error",
    "operationName": {
        "value": "Microsoft.ContainerService/managedClusters/write",
        "localizedValue": "Create or Update Managed Cluster"
    },
    "resourceGroupName": "<CLUSTER NAME>",
    "resourceProviderName": {
        "value": "Microsoft.ContainerService",
        "localizedValue": "Microsoft.ContainerService"
    },
    "resourceType": {
        "value": "Microsoft.ContainerService/managedClusters",
        "localizedValue": "Microsoft.ContainerService/managedClusters"
    },
    "resourceId": "/subscriptions/<SUBSCRIPTION_ID>/resourceGroups/<CLUSTER NAME>/providers/Microsoft.ContainerService/managedClusters/<CLUSTER NAME>",
    "status": {
        "value": "Failed",
        "localizedValue": "Failed"
    },
    "subStatus": {
        "value": "Conflict",
        "localizedValue": "Conflict (HTTP Status Code: 409)"
    },
    "submissionTimestamp": "2021-02-22T11:25:38.1599708Z",
    "subscriptionId": "<SUBSCRIPTION_ID>",
    "properties": {
        "statusCode": "Conflict",
        "statusMessage": "{\"code\":\"OperationNotAllowed\",\"message\":\"Operation is not allowed: Another operation (<agent pool name> - Creating) is in progress, please wait for it to finish before starting a new operation. See https://aka.ms/aks-pending-operation for more details\"}",
        "eventCategory": "Administrative",
        "entity": "/subscriptions/<SUBSCRIPTION_ID>/resourceGroups/<CLUSTER NAME>/providers/Microsoft.ContainerService/managedClusters/<CLUSTER NAME>",
        "message": "Microsoft.ContainerService/managedClusters/write",
        "hierarchy": "<SUBSCRIPTION_ID>"
    },
}
```

_3. Throttling error and getting 429 response code due to too many requests. `CreateOrUpdate` request for the managed cluster is sent on every reconciliation loop, even when there are no changes in the cluster CRs_

Error from Azure activity log (truncated for the sake of readability):
```
{
    "authorization": {
        "action": "Microsoft.ContainerService/managedClusters/write",
        "scope": "/subscriptions/<SUBSCRIPTION ID>/resourceGroups/<CLUSTER ID>/providers/Microsoft.ContainerService/managedClusters/<CLUSTER ID>"
    },
    "channels": "Operation",
    "eventName": {
        "value": "EndRequest",
        "localizedValue": "End request"
    },
    "category": {
        "value": "Administrative",
        "localizedValue": "Administrative"
    },
    "eventTimestamp": "2021-02-22T11:24:37.3869786Z",
    "level": "Error",
    "operationName": {
        "value": "Microsoft.ContainerService/managedClusters/write",
        "localizedValue": "Create or Update Managed Cluster"
    },
    "resourceGroupName": "<CLUSTER ID>",
    "resourceProviderName": {
        "value": "Microsoft.ContainerService",
        "localizedValue": "Microsoft.ContainerService"
    },
    "resourceType": {
        "value": "Microsoft.ContainerService/managedClusters",
        "localizedValue": "Microsoft.ContainerService/managedClusters"
    },
    "resourceId": "/subscriptions/<SUBSCRIPTION ID>/resourceGroups/<CLUSTER ID>/providers/Microsoft.ContainerService/managedClusters/<CLUSTER ID>",
    "status": {
        "value": "Failed",
        "localizedValue": "Failed"
    },
    "subStatus": {
        "value": "429",
        "localizedValue": "429"
    },
    "subscriptionId": "<SUBSCRIPTION ID>",
    "properties": {
        "statusCode": "429",
        "serviceRequestId": null,
        "statusMessage": "{\"error\":{\"code\":\"FailedIdentityOperation\",\"message\":\"Identity operation for resource '/subscriptions/<SUBSCRIPTION ID>/resourceGroups/<CLUSTER ID>/providers/Microsoft.ContainerService/managedClusters/<CLUSTER ID>' failed with error 'Failed to perform resource identity operation. Status: '429'. Response: '{\\\"error\\\":{\\\"code\\\":\\\"429\\\",\\\"message\\\":\\\"Request is temporarily throttled because resource /subscriptions/<SUBSCRIPTION ID>/resourcegroups/<CLUSTER ID>/providers/Microsoft.ContainerService/managedClusters/<CLUSTER ID> has issued too many requests. Retry after 20 seconds.\\\"}}'.'.\"}}",
        "eventCategory": "Administrative",
        "entity": "/subscriptions/<SUBSCRIPTION ID>/resourceGroups/<CLUSTER ID>/providers/Microsoft.ContainerService/managedClusters/<CLUSTER ID>",
        "message": "Microsoft.ContainerService/managedClusters/write",
        "hierarchy": "<SUBSCRIPTION ID>"
    },
    "relatedEvents": []
}
```

Fixes #847

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Fix AKS cluster provisioning errors and validate update of immutable fields in AzureManagedControlPlane webhook.
```

### DIFF
--- a/azure/services/managedclusters/managedclusters.go
+++ b/azure/services/managedclusters/managedclusters.go
@@ -22,6 +22,7 @@ import (
 	"net"
 
 	"github.com/Azure/azure-sdk-for-go/services/containerservice/mgmt/2020-02-01/containerservice"
+	"github.com/google/go-cmp/cmp"
 	"github.com/pkg/errors"
 	"k8s.io/klog/v2"
 
@@ -120,7 +121,7 @@ func (s *Service) Reconcile(ctx context.Context, spec interface{}) error {
 		return errors.New("expected managed cluster specification")
 	}
 
-	properties := containerservice.ManagedCluster{
+	managedCluster := containerservice.ManagedCluster{
 		Identity: &containerservice.ManagedClusterIdentity{
 			Type: containerservice.SystemAssigned,
 		},
@@ -152,12 +153,12 @@ func (s *Service) Reconcile(ctx context.Context, spec interface{}) error {
 	}
 
 	if managedClusterSpec.PodCIDR != "" {
-		properties.NetworkProfile.PodCidr = &managedClusterSpec.PodCIDR
+		managedCluster.NetworkProfile.PodCidr = &managedClusterSpec.PodCIDR
 	}
 
 	if managedClusterSpec.ServiceCIDR != "" {
 		if managedClusterSpec.DNSServiceIP == nil {
-			properties.NetworkProfile.ServiceCidr = &managedClusterSpec.ServiceCIDR
+			managedCluster.NetworkProfile.ServiceCidr = &managedClusterSpec.ServiceCIDR
 			ip, _, err := net.ParseCIDR(managedClusterSpec.ServiceCIDR)
 			if err != nil {
 				return fmt.Errorf("failed to parse service cidr: %w", err)
@@ -168,9 +169,9 @@ func (s *Service) Reconcile(ctx context.Context, spec interface{}) error {
 			// https://golang.org/src/net/ip.go#L48
 			ip[15] = byte(10)
 			dnsIP := ip.String()
-			properties.NetworkProfile.DNSServiceIP = &dnsIP
+			managedCluster.NetworkProfile.DNSServiceIP = &dnsIP
 		} else {
-			properties.NetworkProfile.DNSServiceIP = managedClusterSpec.DNSServiceIP
+			managedCluster.NetworkProfile.DNSServiceIP = managedClusterSpec.DNSServiceIP
 		}
 	}
 
@@ -183,23 +184,48 @@ func (s *Service) Reconcile(ctx context.Context, spec interface{}) error {
 			Type:         containerservice.VirtualMachineScaleSets,
 			VnetSubnetID: &managedClusterSpec.VnetSubnetID,
 		}
-		*properties.AgentPoolProfiles = append(*properties.AgentPoolProfiles, profile)
+		*managedCluster.AgentPoolProfiles = append(*managedCluster.AgentPoolProfiles, profile)
 	}
 
 	existingMC, err := s.Client.Get(ctx, managedClusterSpec.ResourceGroupName, managedClusterSpec.Name)
 	if err != nil && !azure.ResourceNotFound(err) {
 		return errors.Wrap(err, "failed to get existing managed cluster")
-	} else if !azure.ResourceNotFound(err) {
+	}
+
+	isCreate := azure.ResourceNotFound(err)
+	if isCreate {
+		err = s.Client.CreateOrUpdate(ctx, managedClusterSpec.ResourceGroupName, managedClusterSpec.Name, managedCluster)
+		if err != nil {
+			return fmt.Errorf("failed to create managed cluster, %w", err)
+		}
+	} else {
 		ps := *existingMC.ManagedClusterProperties.ProvisioningState
 		if ps != "Canceled" && ps != "Failed" && ps != "Succeeded" {
 			klog.V(2).Infof("Unable to update existing managed cluster in non terminal state.  Managed cluster must be in one of the following provisioning states: canceled, failed, or succeeded")
 			return nil
 		}
-	}
 
-	err = s.Client.CreateOrUpdate(ctx, managedClusterSpec.ResourceGroupName, managedClusterSpec.Name, properties)
-	if err != nil {
-		return fmt.Errorf("failed to create or update managed cluster, %w", err)
+		// Normalize properties for the desired (CR spec) and existing managed
+		// cluster, so that we check only those fields that were specified in
+		// the initial CreateOrUpdate request and that can be modified.
+		// Without comparing to normalized properties, we would always get a
+		// difference in desired and existing, which would result in sending
+		// unnecessary Azure API requests.
+		propertiesNormalized := &containerservice.ManagedClusterProperties{
+			KubernetesVersion: managedCluster.ManagedClusterProperties.KubernetesVersion,
+		}
+		existingMCPropertiesNormalized := &containerservice.ManagedClusterProperties{
+			KubernetesVersion: existingMC.ManagedClusterProperties.KubernetesVersion,
+		}
+
+		diff := cmp.Diff(propertiesNormalized, existingMCPropertiesNormalized)
+		if diff != "" {
+			klog.V(2).Infof("Update required (+new -old):\n%s", diff)
+			err = s.Client.CreateOrUpdate(ctx, managedClusterSpec.ResourceGroupName, managedClusterSpec.Name, managedCluster)
+			if err != nil {
+				return fmt.Errorf("failed to update managed cluster, %w", err)
+			}
+		}
 	}
 
 	return nil
--- a/exp/api/v1alpha4/azuremanagedcontrolplane_webhook.go
+++ b/exp/api/v1alpha4/azuremanagedcontrolplane_webhook.go
@@ -22,13 +22,15 @@ import (
 	"regexp"
 	"strings"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/validation/field"
-	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1alpha4"
 	ctrl "sigs.k8s.io/controller-runtime"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
+
+	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1alpha4"
 )
 
 // log is for logging in this package.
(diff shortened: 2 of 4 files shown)


azure/services/managedclusters/managedclusters.go
```diff
@@ -183,23 +183,64 @@ func (s *Service) Reconcile(ctx context.Context, spec interface{}) error {
...
+			// We will check ServiceCidr and DNSServiceIP only in the case they
+			// were set in the initial request, so basically under the same
+			// conditions like above when setting values in managedCluster.
+			if managedClusterSpec.DNSServiceIP == nil {
+				existingMCPropertiesNormalized.NetworkProfile.ServiceCidr = existingMC.ManagedClusterProperties.NetworkProfile.ServiceCidr
+			}
+			existingMCPropertiesNormalized.NetworkProfile.DNSServiceIP = existingMC.ManagedClusterProperties.NetworkProfile.DNSServiceIP
+		}
+
+		diff := cmp.Diff(managedCluster.ManagedClusterProperties, existingMCPropertiesNormalized)
```
nprokopic, line 236: This was already done for agent pool `CreateOrUpdate` requests, so I did the same thing here. One of the consequences of this approach is that any manual change on the Azure AKS resource will remain unaffected, until some part of the CR changes so the new `CreateOrUpdate` is sent.

OTOH there are not many properties that are configurable in AKS after the cluster has been created (especially the properties that are currently supported in the CAPZ managed cluster types), so this might be fine for now.

nprokopic:
About `pull-cluster-api-provider-azure-apidiff` test, is that change ok, or should I revert it?

The whole AKS implementation should be refactored I believe, so it's aligned with other controllers a d services, just didn't want to do the whole refactoring in this PR.

CecileRobertMichon:
apidiff changes for the record:

```
  Incompatible changes:
  - NewService: changed from func(sigs.k8s.io/cluster-api-provider-azure/azure.Authorizer) *Service to func(*sigs.k8s.io/cluster-api-provider-azure/azure/scope.ManagedControlPlaneScope) *Service
  Compatible changes:
  - Service.Scope: added
  ```
  
 @nprokopic those are fine, no need to revert

**Comment by CecileRobertMichon** [comment]
The unit tests are also failing though and the errors seem related to the changes, @nprokopic can take a look when you get a chance?

File: azure/services/managedclusters/managedclusters.go (hunk 1 above)
Comment (alexeldeib) [reply azure/services/managedclusters/managedclusters.go:236]: > any manual change on the Azure AKS resource will remain unaffected, until some part of the CR changes so the new CreateOrUpdate is sent.

what would the expected behavior be in your mind? that's what the normalization was intended to cover. we only support a subset of properties, so we only set those and diff against those. once CAPZ supports any given property, we should be diffing/applying it.

If we don't support a property, how would we send something other than what was done out of band? If we do support it, we should be covering the diff here, converging on 100% support.

File: azure/services/managedclusters/managedclusters.go [hunk 2]
```diff
@@ -183,23 +183,64 @@ func (s *Service) Reconcile(ctx context.Context, spec interface{}) error {
...
+		// Normalize properties for the existing managed cluster, so that we
+		// check only those fields that were specified in the initial
+		// CreateOrUpdate request. Without comparing to normalized properties,
+		// we would always get a difference in desired and existing, which
+		// would result in sending unnecessary Azure API requests.
+		existingMCPropertiesNormalized := &containerservice.ManagedClusterProperties{
+			NodeResourceGroup:       existingMC.ManagedClusterProperties.NodeResourceGroup,
+			DNSPrefix:               existingMC.ManagedClusterProperties.DNSPrefix,
+			KubernetesVersion:       existingMC.ManagedClusterProperties.KubernetesVersion,
+			LinuxProfile:            existingMC.ManagedClusterProperties.LinuxProfile,
```
alexeldeib, line 216: don't diff the ssh keys, as I found out recently they are unfortunately not updateable -_-

alexeldeib:
thanks for fixing these issues! 😃 only minor comments from me

nprokopic:
> The unit tests are also failing though and the errors seem related to the changes, @nprokopic can take a look when you get a chance?

Sure, just couldn't find time in previous days. I'll try to take a look today.

azure/services/managedclusters/managedclusters.go
nprokopic: > I found out recently they are unfortunately not updatable -_-

+1 :)

Yeah, makes sense not to check them. Probably some other fields here are also immutable and can also be skipped in the check, so I'll check that out.

nprokopic:
I looked into the failures, yeah it's my changes that caused that, looks like I forgot to run unit tests after making final changes here.

I have to take a look how mocking is working to add some new mocks, e.g. `MockManagedControlPlaneScope`, to fix this properly.

CecileRobertMichon:
@nprokopic some of the scope / mocking stuff might be a bit different for managedcontrolplanes right now because we still haven't refactored that and agentpools to match the rest of the services https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/757

azure/services/managedclusters/managedclusters.go
nprokopic: I missed this comment, sorry about that.

I was maybe overthinking it here a bit, you're right, if some property is supported by CAPZ, it should be included in the diff, otherwise no need to check it. Another limiting factor here is if Azure supports the modification of the property or not.

Out of all properties that are set here for managed cluster creation https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/02ca6c383c4c459efb79c8d3673f332f212f4046/azure/services/managedclusters/managedclusters.go#L123, it seems to me that only the Kubernetes version can be updated. All other properties are either hard-coded in CAPZ, or they are used only during creation and Azure does not allow to modify them (e.g. like Linux profile), so I don't think that there is a point in checking them.

Here are the properties that can be updated via CLI https://docs.microsoft.com/en-us/cli/azure/aks?view=azure-cli-latest#az_aks_update, now I guess that you can modify a bit more via API directly, but I'm not sure how much more, I couldn't find any docs that explain it, since Azure API for managed clusters does not have separate Create and Update requests, but only a single `CreateOrUpdate` request.

To be more concrete, out of all managed cluster properties that are set [here](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/02ca6c383c4c459efb79c8d3673f332f212f4046/azure/services/managedclusters/managedclusters.go#L123), I don't think that these can be modified:
- Location
- ManagedClusterProperties.NodeResourceGroup
- ManagedClusterProperties.DNSPrefix
- ManagedClusterProperties.LinuxProfile
- ManagedClusterProperties.NetworkProfile
  - NetworkPlugin
  - LoadBalancerSKU
  - NetworkPolicy
  - PodCidr
  - DNSServiceIP

These are hard-coded in CAPZ for now:
- Identity
- ManagedClusterProperties.ServicePrincipalProfile

AgentPoolProfiles are handled by agentpools service, so no need to check them in managed cluster update as well.

Which then leaves only ManagedClusterProperties.KubernetesVersion as the only property here which can be updated.

Am I making any sense here? :smile:

nprokopic:
> @nprokopic some of the scope / mocking stuff might be a bit different for managedcontrolplanes right now because we still haven't refactored that and agentpools to match the rest of the services #757

@CecileRobertMichon I have reverted some of the changes around logging and scope that were not that necessary, to reduce the PR size here a bit. I'd prefer to do the refactoring in a separate PR (like other refactoring PRs from https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/757), and focus this one on fixing errors during AKS cluster provisioning.

azure/services/managedclusters/managedclusters.go
alexeldeib: I said I would reply later so here I am 🙂 I think I said pretty much everything that mattered in slack, however. A lot of properties in the AKS API are, as you noticed, immutable. So it definitely makes sense to remove them from the diff and block them earlier, e.g. in webhooks. Up to you whether you want to tackle that in this PR :)

nader-ziada:
I think this is good to go now
lgtm

@CecileRobertMichon @alexeldeib for any last comments

nader-ziada:
@nprokopic I believe you need to rebase on #1260 to get the windows e2e working

once that passes, we can merge it in

azure/services/managedclusters/managedclusters.go
nprokopic: > Up to you whether you want to tackle validation in this PR, or stick to fixing the errors

Finally got some time to finish this. I have added update validations to the AzureManagedControlPlane webhook that are returning an error when an immutable field is changed.

nprokopic:
> New changes are detected. LGTM label has been removed.

After chatting with @alexeldeib about some fields being immutable in AzureManagedControlPlane because Azure does not allow them to be changed, I have added update validations to the AzureManagedControlPlane webhook that are returning an error when an immutable field is changed.

nprokopic:
> @nprokopic I believe you need to rebase on #1260 to get the windows e2e working
> 
> once that passes, we can merge it in

Rebased, but pull-cluster-api-provider-azure-e2e-windows failed again.

nader-ziada:
@alexeldeib would you mind taking a quick look at the new validations added

alexeldeib:
/lgtm

thanks a lot for your work here, this is great!

devigned (approved):
Thank you for the PR. Awesome work!
/lgtm

PR #1207, opened 2021-03-04
Allow API server LB frontend port to be configured

tuommaki:
/kind feature

**Description:**
This change allows API server LB frontend port to be configured separately from the API server listen port. The code uses `Cluster` CR's `Spec.ControlPlaneEndpoint.Port` as LB frontend port, given it is defined. Otherwise the API server listen port is used as the implementation was before.

This change implements following proposal: https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1240

**Common use case for this would be**:
API LB is listening in default HTTPS port (443) while workload cluster control plane node uses default (6443) for API server listen port.

Depending on situation, this might require user action on existing clusters if control plane endpoint port is configured but earlier LB defaulting has been used.

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Use `Cluster` CR's `Spec.ControlPlaneEndpoint.Port` as LB frontend port if it's defined. User action is needed if the said port is defined but differs from `Spec.ClusterNetwork.APIServerPort`.
```

### DIFF
--- a/azure/scope/cluster.go
+++ b/azure/scope/cluster.go
@@ -144,6 +144,7 @@ func (s *ClusterScope) LBSpecs() []azure.LBSpec {
 			Name:              s.APIServerLB().Name,
 			SubnetName:        s.ControlPlaneSubnet().Name,
 			FrontendIPConfigs: s.APIServerLB().FrontendIPs,
+			FrontendPort:      s.APIServerLBFrontendPort(),
 			APIServerPort:     s.APIServerPort(),
 			Type:              s.APIServerLB().Type,
 			SKU:               infrav1.SKUStandard,
@@ -455,6 +456,15 @@ func (s *ClusterScope) AdditionalTags() infrav1.Tags {
 		tags = s.AzureCluster.Spec.AdditionalTags.DeepCopy()
 	}
 	return tags
+}
+
+// APIServerLBFrontendPort returns the frontend listen port to use when creating the load balancer.
+func (s *ClusterScope) APIServerLBFrontendPort() int32 {
+	if s.Cluster.Spec.ControlPlaneEndpoint.Port > 0 {
+		return s.Cluster.Spec.ControlPlaneEndpoint.Port
+	}
+
+	return s.APIServerPort()
 }
 
 // APIServerPort returns the APIServerPort to use when creating the load balancer.
--- a/azure/services/loadbalancers/loadbalancers.go
+++ b/azure/services/loadbalancers/loadbalancers.go
@@ -259,7 +259,7 @@ func (s *Service) getLoadBalancingRules(lbSpec azure.LBSpec, frontendIDs []netwo
 				LoadBalancingRulePropertiesFormat: &network.LoadBalancingRulePropertiesFormat{
 					DisableOutboundSnat:     to.BoolPtr(true),
 					Protocol:                network.TransportProtocolTCP,
-					FrontendPort:            to.Int32Ptr(lbSpec.APIServerPort),
+					FrontendPort:            to.Int32Ptr(lbSpec.FrontendPort),
 					BackendPort:             to.Int32Ptr(lbSpec.APIServerPort),
 					IdleTimeoutInMinutes:    to.Int32Ptr(4),
 					EnableFloatingIP:        to.BoolPtr(false),
--- a/azure/services/loadbalancers/loadbalancers_test.go
+++ b/azure/services/loadbalancers/loadbalancers_test.go
@@ -77,6 +77,7 @@ func TestReconcileLoadBalancer(t *testing.T) {
 								},
 							},
 						},
+						FrontendPort:  443,
 						APIServerPort: 6443,
 					},
 				})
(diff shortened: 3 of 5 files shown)


CecileRobertMichon:
Thanks for the PR @tuommaki! Could you please document how to set both ports in https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/docs/book/src/topics/api-server-endpoint.md ? 

Also, does this change impact the guidance on allowing inbound ports in vnet security groups https://github.com/kubernetes-sigs/image-builder/pull/422 ? And does it impact the default ingress rules we generate https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/azure/scope/cluster.go#L507

tuommaki:
Hi Cecile and thanks for the initial feedback!

> Could you please document how to set both ports

Sure, I will.

> does this change impact the guidance on allowing inbound ports in vnet security groups

Based on my understanding no, because this change won't affect the CP node listen port.

> does it impact the default ingress rules we generate

This one I'm not sure and I'll perform further tests to find out. :+1:

tuommaki:
> does it impact the default ingress rules we generate

Ok. I did additional testing to verify this and I cannot observe any impacts. The default ingress rule linked earlier matches for apiserver listen port, but this PR doesn't change any behavior related to that.

This only affects the LB public facing configuration which affects LB rules but those work both with defaults and with configuration e.g. from 443 -> 6443.

I also created the issue and linked it in proposal.

tuommaki:
I'll write the missing documentation next.

azure/scope/cluster.go
```diff
@@ -457,6 +458,15 @@ func (s *ClusterScope) AdditionalTags() infrav1.Tags {
 	return tags
 }
 
+// APIServerLBFrontendPort returns the frontend listen port to use when creating the load balancer.
+func (s *ClusterScope) APIServerLBFrontendPort() int32 {
+	if s.Cluster.Spec.ControlPlaneEndpoint.Port > 0 {
```
CecileRobertMichon, line 463: What is still not 100% clear to me is the relationship between Cluster.Spec.ControlPlaneEndpoint and AzureCluster.Spec.ControlPlaneEndpoint

We set AzureCluster.Spec.ControlPlaneEndpoint in the AzureCluster controller here (to follow the CAPI provider contract): https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/controllers/azurecluster_controller.go#L256-L260

I had opened a discussion in CAPI a while back about moving it to Status since it's basically being used as status right now (and any user value will be overridden) https://github.com/kubernetes-sigs/cluster-api/issues/3715

CecileRobertMichon: in the CAPI cluster_controller, `Cluster.Spec.ControlPlaneEndpoint` is set to the value of `AzureCluster.Spec.ControlPlaneEndpoint` https://github.com/kubernetes-sigs/cluster-api/blob/master/controllers/cluster_controller_phases.go#L184

tuommaki: Hmm... I didn't realize to think from this perspective.

My original use case for this change was to provide migration compatibility when converting existing Kubernetes cluster (Giant Swarm flavor in this case) into one managed by CAPZ. The cluster is already reconciled by v1alpha3 CAPI + CAPZ CRs and there's DNS zone managed for the workload cluster.

I was also viewing the `Cluster` CR as main spec for provider independent workload cluster configuration since it also has several important network configuration fields (e.g. Service and POD CIDRs as well as API server listen port).

Now I see that my understanding / interpretation on that is slightly off from the present implementation and interoperability between different controllers.

Do you think this change would still make sense in CAPZ?

As far as I was able to reason & test, this change should be fully compatible with present state, but I do understand that there's the corner case that one might not be able to configure _just_ port because the host must be set as well and something needs to manage that. I'm not sure if that can be fixed in any way right now.

CecileRobertMichon: > I was also viewing the Cluster CR as main spec for provider independent workload cluster configuration since it also has several important network configuration fields (e.g. Service and POD CIDRs as well as API server listen port).

Right, that's why it's confusing. I think it's a bug that providers don't honor that field, that's why I opened https://github.com/kubernetes-sigs/cluster-api/issues/3715. 

I think as long as it has no effect on current behaviors for those who don't set the port & for existing clusters that have the default port, it should be ok to make this change to unblock your use case. However in that case we should remove the user facing documentation since it's not really a configuration option, but more of a workaround. Also let's make sure there are solid comments in the code so we know exactly why this was done and what the expected bahavior is when we look at this code again 6 months from now.

tuommaki: Sorry for long delays, we have been recently working on different areas so this was not high priority. Because of this controversial nature of this feature and the fact that we don't urgently need this right now, I'll take more time to consider this. I'll get back to this later. Thanks for your patience :slightly_smiling_face:

CecileRobertMichon:
> Sorry for long delays, we have been recently working on different areas so this was not high priority. Because of this controversial nature of this feature and the fact that we don't urgently need this right now, I'll take more time to consider this. I'll get back to this later. Thanks for your patience 🙂

Going to close the PR for now, feel free to reopen when it's ready for review again

/close

PR #1210, opened 2021-03-04
⚠️ Use custom service accounts instead of default for controllers

### LINKED ISSUE #1206: Use custom service accounts instead of default for controllers
As an operator, I would like to use a dedicated service account for each CAPZ controller (instead of default)

/kind feature

/priority important-soon
/milestone v0.4.0
/help

see https://github.com/kubernetes-sigs/cluster-api/pull/4245

furkatgofurov7:
**What type of PR is this?**
/kind feature

Uses a dedicated service account for each CAPZ controller (instead of default)

Fixes #1206 

_Please have a look thoroughly, in case I missed some important points to update in the PR. Thanks!._

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
⚠️ Use custom service accounts instead of default for controllers
```

### DIFF
--- a/config/default/aad-pod-identity-deployment.yaml
+++ b/config/default/aad-pod-identity-deployment.yaml
@@ -63,8 +63,8 @@ metadata:
     k8s-app: capz-aad-pod-id-nmi-binding
 subjects:
 - kind: ServiceAccount
-  name: default
-  namespace: capz-system
+  name: manager
+  namespace: system
 roleRef:
   kind: ClusterRole
   name: aad-pod-id-nmi-role
@@ -92,7 +92,7 @@ spec:
         component: nmi
         tier: node
     spec:
-      serviceAccountName: default
+      serviceAccountName: manager
       hostNetwork: true
       dnsPolicy: ClusterFirstWithHostNet
       volumes:
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -58,3 +58,4 @@ spec:
               fieldRef:
                 fieldPath: metadata.namespace      
       terminationGracePeriodSeconds: 10
+      serviceAccountName: manager
--- a/config/rbac/auth_proxy_role_binding.yaml
+++ b/config/rbac/auth_proxy_role_binding.yaml
@@ -8,5 +8,5 @@ roleRef:
   name: proxy-role
 subjects:
 - kind: ServiceAccount
-  name: default
+  name: manager
   namespace: system
--- a/config/rbac/kustomization.yaml
+++ b/config/rbac/kustomization.yaml
@@ -3,6 +3,7 @@ kind: Kustomization
 resources:
   - role.yaml
   - role_binding.yaml
+  - service_account.yaml
   - leader_election_role.yaml
   - leader_election_role_binding.yaml
   - auth_proxy_client_clusterrole.yaml
--- a/config/rbac/leader_election_role_binding.yaml
+++ b/config/rbac/leader_election_role_binding.yaml
@@ -8,5 +8,5 @@ roleRef:
   name: leader-election-role
 subjects:
 - kind: ServiceAccount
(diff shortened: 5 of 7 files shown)


nprokopic:
Should the service account for aad-pod-identity be updated as well? https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/config/default/aad-pod-identity-deployment.yaml#L65-L67

furkatgofurov7:
> Should the service account for aad-pod-identity be updated as well? https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/config/default/aad-pod-identity-deployment.yaml#L65-L67

@nprokopic hi! Thanks for the suggestion. I was mainly following [kubernetes-sigs/cluster-api/4245](https://github.com/kubernetes-sigs/cluster-api/pull/4245) and it changes only the SAs' in the `system` namespace, so I haven't touched it. But please do let me know if I am misunderstanding it.

nprokopic:
> I was mainly following kubernetes-sigs/cluster-api/4245 and it changes only the SAs' in the system namespace

I believe that the changes are for the `default` service account in multiple namespaces with names ending with `-system` (there is some `kustomize` templating involved).

Here are the default service accounts from `*-system` namespaces before the above linked CAPI change:

```
$ kubectl get serviceaccount -A | grep default | grep system
capi-kubeadm-bootstrap-system       default                              1         28d
capi-kubeadm-control-plane-system   default                              1         28d
capi-system                         default                              1         28d
capi-webhook-system                 default                              1         28d
capz-system                         default                              1         28d
kube-system                         default                              1         28d
```

So I believe that in CAPZ we should also update the aad-pod-identity deployment which is currently using default service account for `capz-system` namespace, that is of course unless aad-pod-identity doesn't have some specific requirements and it needs the default service account.

nader-ziada:
> So I believe that in CAPZ we should also update the aad-pod-identity deployment which is currently using default service account for `capz-system` namespace, that is of course unless aad-pod-identity doesn't have some specific requirements and it needs the default service account.

agree that aad-pod-identity service account should be updated as well

furkatgofurov7:
@nprokopic thanks for spotting this up, fixed it.

furkatgofurov7:
Seems a CI flakiness, re-triggering.

/retest

config/default/aad-pod-identity-deployment.yaml
```diff
@@ -63,7 +63,7 @@ metadata:
     k8s-app: capz-aad-pod-id-nmi-binding
 subjects:
 - kind: ServiceAccount
-  name: default
+  name: manager
```
nprokopic, line 66: The account that is set here I believe it should also be set in aad-pod-identity NMI DaemonSet `.Spec.Template.Spec.ServiceAccountName` field (in this same file), which is currently set to `default`.

furkatgofurov7: Set it to `manager`. PTAL.

config/default/aad-pod-identity-deployment.yaml
```diff
@@ -63,7 +63,7 @@ metadata:
...
   namespace: capz-system
```
nader-ziada, line 67:
```suggestion
  namespace: system
```

the `manager` service account is created in `system` namespace

furkatgofurov7: Fixed, thanks!

nader-ziada:
/lgtm

@CecileRobertMichon PTAL

CecileRobertMichon:
should this be marked as a breaking change? I see the corresponding CAPI commit is breaking

furkatgofurov7:
@CecileRobertMichon done!

CecileRobertMichon (approved):
/approve

PR #1212, opened 2021-03-05
Capz should not add tags to managed resource group

LochanRn:
**What type of PR is this?**
/kind bug

When delete capz cluster is triggered, capz try to delete the resource group, even through the RG is not created by capz. 
Capz should check if err and only continue to createOrUpdate RG when err is resource-not-found. otherwise when azure API have issue thus return some other error, the RG will be updated with owner-tags, then capz will think the RG is managed. 

Fixes None

**TODOs**:
- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```release-note
None
```


CecileRobertMichon (commented):
/ok-to-test

CecileRobertMichon:
Thanks for the PR @LochanRn. Can you please fix the merge conflict and add a unit test case that verifies the scenario you are fixing?

cloud/services/groups/groups.go
```diff
@@ -55,10 +55,16 @@ func (s *Service) Reconcile(ctx context.Context) error {
 	ctx, span := tele.Tracer().Start(ctx, "groups.Service.Reconcile")
 	defer span.End()
 
-	if _, err := s.client.Get(ctx, s.Scope.ResourceGroup()); err == nil {
+	_, err := s.client.Get(ctx, s.Scope.ResourceGroup())
+	if err == nil {
```
CecileRobertMichon, line 59: isn't this equivalent to what we were doing before with `if _, err := s.client.Get(ctx, s.Scope.ResourceGroup()); err == nil {` ?

LochanRn: Yes, but err will be local to `if _, err := s.client.Get(ctx, s.Scope.ResourceGroup()); err == nil {` block.
We will need the err in `if !azure.ResourceNotFound(err) {
 		return err
 	}
`
I will modify the PR with an else if block, to minimise the number of lines changed.

LochanRn:
closing this ticket as I raised a fresh PR #1218 without merge conflict and added the required changes suggested based on the reviews

Issue #603, opened 2020-05-08
Bootstrap failure detection


ncdc:
/kind feature

**Describe the solution you'd like**
If bootstrapping doesn't succeed, set AzureMachine.Status.FailureReason and FailureMessage to indicate there was an error. 

**Anything else you would like to add:**
More information in https://github.com/kubernetes-sigs/cluster-api/issues/2554.

We may need to amend the bootstrap provider contract to require them to write a sentinel file indicating success to a specific location because not all bootstrap providers will necessarily use cloud-init and have that as a consistent means for checking success/failure. I will probably write up a separate proposal for that.

ncdc:
We will need a way for the CAPZ controller to be able to determine the status for a given AzureMachine. Does Azure have any services where the bootstrap logic in the VM could send an update to an Azure service (such as a message queue) to indicate bootstrap success? Or maybe some way to label or tag the Azure VM? This would need to be properly secured, presumably.

alexeldeib:
I **think** the credentials in azure.json (or MSI) should do the trick for worker authentication, assuming we setup RBAC correctly.

There's an existing boot diagnostics option on VMs that streams cloud init + boot time logs into a plain text file in a storage account. Uploading a sentinel file that way or tagging a VM would probably be doable.

If you're thinking more like a message queue (perhaps to build on later?), Service Bus is a pretty traditional pub/sub message broker and Event Grid is a more loosely coupled event source/sink model with handlers.

nader-ziada:
azure also supports the option to [add a tag](https://docs.microsoft.com/en-us/rest/api/resources/tags/createorupdateatscope) to the VM

ncdc:
Whatever we do, we need to make sure that the information is available for inspection any time, and that CAPZ doesn't have to be running to receive a notification. I want to avoid a situation where bootstrapping fails, but CAPZ crashes/goes offline for whatever reason, and because it's down, we miss the notification that bootstrapping failed. It sounds like the options you suggested are all feasible, as long as messages are persisted until consumption.

CecileRobertMichon:
@nader-ziada the tags are only visible to the Azure user, that might be one downside of using tags (ie. if we wanted to ever gather data on percentage of bootstrap failures/success).

CecileRobertMichon:
/assign @jackfrancis
/assign

[...]

PR #1217, opened 2021-03-05
🐛  Ensure VM and VMSS extensions are applied once

CecileRobertMichon:
**What type of PR is this?**

**What this PR does / why we need it**: This is a temporary fix while #603 work is in progress to make sure that we're not making unnecessary PUT calls to VM and VMSS extensions that already exist. It also adds the extension profile directly as part of VMSS create so that the extensions are applied to the VMSS VMs as they are created instead of being applied to the model after they exist.

These changes are already included in #1105 for the most part, breaking them out in a separate PR to speed up the fix and facilitate backporting.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Ensure VM and VMSS extensions are applied once
```

### DIFF
--- a/azure/services/vmextensions/vmextensions.go
+++ b/azure/services/vmextensions/vmextensions.go
@@ -55,6 +55,11 @@ func (s *Service) Reconcile(ctx context.Context) error {
 	defer span.End()
 
 	for _, extensionSpec := range s.Scope.VMExtensionSpecs() {
+		if _, err := s.client.Get(ctx, s.Scope.ResourceGroup(), extensionSpec.VMName, extensionSpec.Name); err == nil {
+			// check for the extension and don't update if already exists
+			// TODO: set conditions based on extension status
+			continue
+		}
 		s.Scope.V(2).Info("creating VM extension", "vm extension", extensionSpec.Name)
 		err := s.client.CreateOrUpdate(
 			ctx,
@@ -81,6 +86,6 @@ func (s *Service) Reconcile(ctx context.Context) error {
 }
 
 // Delete is a no-op. Extensions will be deleted as part of VM deletion.
-func (s *Service) Delete(ctx context.Context) error {
+func (s *Service) Delete(_ context.Context) error {
 	return nil
 }
--- a/azure/services/scalesets/mock_scalesets/scalesets_mock.go
+++ b/azure/services/scalesets/mock_scalesets/scalesets_mock.go
@@ -445,6 +445,20 @@ func (mr *MockScaleSetScopeMockRecorder) V(level interface{}) *gomock.Call {
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "V", reflect.TypeOf((*MockScaleSetScope)(nil).V), level)
 }
 
+// VMSSExtensionSpecs mocks base method.
+func (m *MockScaleSetScope) VMSSExtensionSpecs() []azure.VMSSExtensionSpec {
+	m.ctrl.T.Helper()
+	ret := m.ctrl.Call(m, "VMSSExtensionSpecs")
+	ret0, _ := ret[0].([]azure.VMSSExtensionSpec)
+	return ret0
+}
+
+// VMSSExtensionSpecs indicates an expected call of VMSSExtensionSpecs.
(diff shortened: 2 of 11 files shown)


devigned (approved):
/lgtm

Thank you so much for opening this PR. I ran out of time on Friday.

azure/services/vmextensions/vmextensions.go
```diff
@@ -55,6 +55,11 @@ func (s *Service) Reconcile(ctx context.Context) error {
 	defer span.End()
 
 	for _, extensionSpec := range s.Scope.VMExtensionSpecs() {
+		if _, err := s.client.Get(ctx, s.Scope.ResourceGroup(), extensionSpec.VMName, extensionSpec.Name); err == nil {
```
alexeldeib, line 58: This is an interesting choice. The PR makes sense, I wonder why we wouldn't want to check/apply the extension with idempotency inside the VM reconciler? Maybe I missed some discussion on that, or wdyt?

CecileRobertMichon: it came out of [this conversation](https://kubernetes.slack.com/archives/CEX9HENG7/p1614873280120900) on slack. The extension was getting applied on every reconcile loop which, for VMs isn't a big deal, but for VMSS was causing the instances to always be in latest model "false" status since the extension was applied on the scale set after the VMSS VMs were created.

This also gets us closer to the behavior that will be implemented soon for #603 (PR incoming) since we'll have to do a GET to fetch the extension status and set the corresponding condition. Separated this fix into this PR so we can backport it to alpha3.

alexeldeib: but why reconcile the VMSS extension separately from the VMSS at all? In this PR I see you moved extension generation into the scaleset service and no-op'd the check here. Can we simply remove this code? https://github.com/CecileRobertMichon/cluster-api-provider-azure/blob/65b62e65c90a590fb362919d24317838bb05603e/exp/controllers/azuremachinepool_reconciler.go#L72-L74

And remove the vmss extensions service altogether? I think that's pretty much what AKS does, fwiw, and I don't think we hit the same issue with the extension being applied repeatedly. To get the extension status I _think_ you'll need the per instance view anyway.
 
relatively minor, not really worth blocking on, just thought I'd mention it.

alexeldeib: hmm, I see the extensions client actually does return the status. I was assuming we'd pluck it from the VM/VMSS instance view, which appears to have the same data:

https://pkg.go.dev/github.com/Azure/azure-sdk-for-go@v52.3.1+incompatible/services/compute/mgmt/2020-12-01/compute#VirtualMachineInstanceView

CecileRobertMichon: > And remove the vmss extensions service altogether? I think that's pretty much what AKS does, fwiw, and I don't think we hit the same issue with the extension being applied repeatedly. To get the extension status I think you'll need the per instance view anyway. I think you can do the same for individual VM too.

I see what you're saying now. I think you're right. I want to keep this PR with as little LOC changes as possible to facilitate backport but I'll look into this as part of #603.

nader-ziada:
lgtm

CecileRobertMichon:
```
sigs.k8s.io/cluster-api-provider-azure/azure/services/scalesets
  Incompatible changes:
  - ScaleSetScope.VMSSExtensionSpecs: added
  ```

**Comment by CecileRobertMichon** [comment]
```
/home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/test/e2e/azure_test.go:143
Unexpected error:
    <*errors.StatusError | 0xc0002c7040>: {
        ErrStatus: {
            TypeMeta: {Kind: "", APIVersion: ""},
            ListMeta: {
                SelfLink: "",
                ResourceVersion: "",
                Continue: "",
                RemainingItemCount: nil,
            },
            Status: "Failure",
            Message: "etcdserver: request timed out",
            Reason: "",
            Details: nil,
            Code: 500,
        },
    }
    etcdserver: request timed out
occurred
/home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/test/e2e/azure_net_pol.go:159
```

That's a new one... :sigh:

/retest

PR #1229, opened 2021-03-11
Make node outbound LB configurable

### LINKED ISSUE #541: Ensure that Standard LBs have configurable number of frontend IPs
/kind feature

Because Kubernetes on Azure implementations share a common Standard LoadBalancer resource to route outbound traffic across all node pools, it's important to be able to scale up the number of IP addresses shared by the backend pool members for outbound access, because outbound port allocation is capped per-IP address.

This is a feature request to make sure we don't harden the Standard LoadBalancer implementation so that it is statically set to 1 public IP address for outbound SNAT. If we do so, then clusters at scale will break (port exhaustion).

This PR will do this for aks-engine, which can be used as a reference:

https://github.com/Azure/aks-engine/pull/3085

**Environment:**

### LINKED ISSUE #1159: Additional Configurability with Azure Private clusters
/kind feature

CAPZ feels to currently be slightly opinionated with the set of resources it creates and does not offer configurability to disable the creation of some resources.  There could be instances where the creation of a resource is undesired by the end user, or even not permitted per an organization security policy. For instance:

1. A user may not want to use a private Azure DNS zone b/c their environment has others means of providing DNS (perhaps through deploying their own Activity Directory service). 
2. A user may not want to have public loadbalancers provisioned b/c the cluster will not have any externally visible services so an ingress path is not required (or ingress is provided through alternative means like Big-IP F5 devices w/ F5 ingress controller).
3. A user may not want to have public loadbalancers provisioned b/c outbound connectivity for worker and controlplane nodes is not permitted because the environment is air-gapped (container images, binaries, linux packages could be served through internal artifact repos like Artifactory or Nexus).  Some organizations also require that egress traffic travel through a proxy device to reach the internet so that additional network edge policies can be enforced. 
4. A user may not want to create public ip addresses b/c the organization enforces a policy in the Azure subscription that prevents the creation of public ips. 

**Environment:**

- cluster-api-provider-azure version: 0.4.10
- Kubernetes version: (use `kubectl version`): 1.18.12
- OS (e.g. from `/etc/os-release`): ubuntu 1804

shysank:
**What type of PR is this?**
/kind feature

~Adds a new field `LoadBalancerNodeOutboundIPs` in `AzureClusterSpec -> NetworkSpec` that specifies the number of frontend ips to be created for node outbound load balancer.~

Adds a separate section for node outbound LB configuration. This includes configuration to:
- Control the number of frontend ips created as part of this LB
- Disable node outbound LB for private clusters

Fixes #541
Also fixes #1159 partially.

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Support multiple frontend IPs for node outbound lb
```


api/v1alpha4/azurecluster_validation.go
```diff
@@ -38,6 +38,8 @@ const (
 	// described in https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules
 	subnetRegex       = `^[-\w\._]+$`
 	loadBalancerRegex = `^[-\w\._]+$`
+	// MaxLoadBalancerOutboundIPs is the maximum number of outbound IPs in a Standard LoadBalancer frontend configuration
+	MaxLoadBalancerOutboundIPs = 16
```
shysank, line 42: This number comes from [here](https://github.com/Azure/aks-engine/blob/f81f85fd9f8a3e83607c883204417685ddc2020e/pkg/api/common/const.go#L36)

azure/defaults.go
```diff
@@ -143,6 +143,11 @@ func GenerateAvailabilitySetName(clusterName, nodeGroup string) string {
 	return fmt.Sprintf("%s_%s-as", clusterName, nodeGroup)
 }
 
+// WithCount appends the count as suffix to a generated name
+func WithCount(name string, n int) string {
```
CecileRobertMichon, line 147:
```suggestion
func WithIndex(name string, n int) string {
```

is this more accurate?

api/v1alpha4/types.go
```diff
@@ -58,6 +58,10 @@ type NetworkSpec struct {
 	// APIServerLB is the configuration for the control-plane load balancer.
 	// +optional
 	APIServerLB LoadBalancerSpec `json:"apiServerLB,omitempty"`
+
+	// LoadBalancerNodeOutboundIPs is the configuration for multiple frontend ips for node outbound load balancer.
+	// +optional
+	LoadBalancerNodeOutboundIPs *int32 `json:"load_balancer_node_outbound_ips,omitempty"`
```
CecileRobertMichon, line 64: one of the other (sort of related) feature that has been requested that we might want to implement in the future is being able to disable the outbound LB all together https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1159

Not saying we should do it in this PR, but it would be good to think about how the UX might work for that and how this API can be modified later to include this, either by having a separate LB section for outbound, like `NodeOutboundLB.FrontendIPsCount`, or maybe reusing this field and allow setting it to 0 for no outbound IPs. WDYT?

azure/scope/cluster.go
```diff
@@ -394,6 +427,11 @@ func (s *ClusterScope) AvailabilitySetEnabled() bool {
 	return len(s.AzureCluster.Status.FailureDomains) == 0
 }
 
+// LoadBalancerNodeOutboundIPs returns the number node outbound ips that needs to be configured in the load balancer.
+func (s *ClusterScope) LoadBalancerNodeOutboundIPs() *int32 {
```
jackfrancis, line 431: Is it idiomatic to return the pointer type from this convenience method? Would returning an `int32` be easier for callers?

azure/scope/cluster.go
```diff
@@ -128,12 +131,24 @@ func (s *ClusterScope) PublicIPSpecs() []azure.PublicIPSpec {
...
+	// Node outbound public ip spec
+	// If no value is configured for LoadBalancerNodeOutboundIPs, or if it is set to 1, create one public ip
+	if s.LoadBalancerNodeOutboundIPs() == nil || *s.LoadBalancerNodeOutboundIPs() == 1 {
+		return append(publicIPSpecs, azure.PublicIPSpec{
 			Name: azure.GenerateNodeOutboundIPName(s.ClusterName()),
-		},
+		})
 	}
+
+	// For LoadBalancerNodeOutboundIPs > 1 create as many public ips as configured in LoadBalancerNodeOutboundIPs
```
jackfrancis, line 144: If we update `LoadBalancerNodeOutboundIPs()` to return `1` if the underlying data is `nil` (which I assume means the user did not provide a value, and we would provide a default value of `1`) then we could reduce some code surface area here and simply use the enumeration for all values of `LoadBalancerNodeOutboundIPs()`.

azure/defaults.go
shysank: Yeah, I think it makes more sense.

api/v1alpha4/types.go
shysank: I'm inclined to having a separate section for outbound LB because:
1. Setting `LoadBalancerNodeOutboundIPs` to 0 will mean that the LB will have 0 public IPs, but there'll still be an outbound LB. Instead we can do a simple nil check to disable outbound LB, avoids any ambiguity.
2. We can have more control over the LB config like we do for api server LB.

azure/scope/cluster.go
shysank: Yeah I think it makes sense to return 1 if nil.

azure/scope/cluster.go
shysank: The reason I've handled `LoadBalancerNodeOutboundIPs == 1` logic separately is I because I wanted the naming of azure resources to be the same as it was before (without the `-1` suffix). It's a bit of redundant code but I didn't want to behavior to change when there's no value set, which I believe will be, in most cases. Does that make sense?

api/v1alpha4/types.go
CecileRobertMichon: +1, a separate section makes sense to me as well

azure/scope/cluster.go
jackfrancis: Makes sense.

How about we capture the value as a variable so we only have to call `s.LoadBalancerNodeOutboundIPs()` only once.

And just a suggestion: IMO formulating these two append patterns (either == 1 or > 1) as an if/else will have the effect of reading the code more suggestive that these are mutually exclusive; as opposed to the early return statement in the `== 1` scenario. Keeping the common `return publicIPSpecs` at the end of the function as the only return invocation is a little easier to read/maintain.

shysank: Thanks for the feedback. I've made the changes, ptal.

api/v1alpha4/types.go
CecileRobertMichon: Does it make sense to create the new LB field now so we don't have to change the API again later?

shysank: yeah, make sense. I'll update the code.

CecileRobertMichon:
Can you please add docs that explain how and why a user should use this new field?

api/v1alpha4/azurecluster_default.go
```diff
@@ -148,6 +151,58 @@ func (c *AzureCluster) setAPIServerLBDefaults() {
...
+	lb := c.Spec.NetworkSpec.NodeOutboundLB
+	if lb.Type == "" {
+		lb.Type = Public
+	}
+	if lb.SKU == "" {
+		lb.SKU = SKUStandard
+	}
+
+	if lb.Name == "" {
+		lb.Name = c.ObjectMeta.Name
```
CecileRobertMichon, line 171: there's a catch here... Azure only allows you to attach one public (and one private) LB to each VM. Because the cloud provider already creates a public LB (and internal one) for worker nodes to expose services of type LoadBalancer, we have to use the same LB and can't create a separate one. Cloud provider makes the assumption that the LB name is <CLUSTER_NAME> https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/site/content/en/topics/assumptions.md so this isn't technically configurable and has to be `<CLUSTER_NAME>` (or whatever `--cluster-name` is set to in cloud-controller-manager https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/templates/flavors/base/cluster-template.yaml#L73).

CecileRobertMichon (commented):
@shysank adding `NodeOutboundLB *LoadBalancerSpec` and all its options seems to be adding a lot of complexity unrelated to what this PR is trying to enable (configurable frontendIP count). What do you think about changing it so that all the fields in `LoadBalancerSpec` are unsupported for now except `FrontendIPsCount` (other fields can be disabled via webhook validation). We can expand to supporting more user configurations in the future but for now, it might be easier to start with just `FrontendIPsCount`.

api/v1alpha4/types.go
```diff
@@ -160,11 +164,12 @@ type IngressRules []IngressRule
...
-	Name        string       `json:"name,omitempty"`
-	SKU         SKU          `json:"sku,omitempty"`
-	FrontendIPs []FrontendIP `json:"frontendIPs,omitempty"`
-	Type        LBType       `json:"type,omitempty"`
+	ID               string       `json:"id,omitempty"`
+	Name             string       `json:"name,omitempty"`
+	SKU              SKU          `json:"sku,omitempty"`
+	FrontendIPs      []FrontendIP `json:"frontendIPs,omitempty"`
+	Type             LBType       `json:"type,omitempty"`
+	FrontendIPsCount *int32       `json:"frontendIPsCount,omitempty"`
```
CecileRobertMichon, line 172: I could see this getting confusing for users, the list of frontendIPs above is redundant information with `FrontendIPsCount`... need to think more about this 🤔

api/v1alpha4/azurecluster_default.go
shysank: Thanks for the explanation, Cecile. I was wondering why we didn't have any suffix to node outbound lb name like we do for api-server. This makes sense. Will fix it. Probably will end up disabling mutation as suggested in your other comment.

api/v1alpha4/types.go
shysank: Yeah, I was earlier thinking something like `minFrontendIPsRequired`, and add some comments to make it clear that this field will not be used when frontendIps are configured. Also, instead of modifying `LoadBalancerSpec` to include FrontendIPsCount, maybe we could create a new type specific to NodeOutboundLb that composes LoadBalacerSpec with this new field. wdyt?

shysank: Another UX I don't like is the user having to specify an empty spec for node outbound lb for private cluster. Not sure how to avoid it though since we want to allow disabling node outbound LB.

CecileRobertMichon: > Another UX I don't like is the user having to specify an empty spec for node outbound lb for private cluster. Not sure how to avoid it though since we want to allow disabling node outbound LB.

I agree. We would need a specific "disable" field to make it explicit. I'm leaning towards separating the two types. @devigned @nader-ziada thoughts?

CecileRobertMichon: > add some comments to make it clear that this field will not be used when frontendIps are configured

this should also be enforced by validation. I think we can allow both as long as they are consistent with one another, or make them mutually exclusive completely.

shysank: > this should also be enforced by validation. I think we can allow both as long as they are consistent with one another, or make them mutually exclusive completely.

Since we're just going to enable `frontendIPsCount` to be mutated, this shouldn't be an issue at least for now.  I agree though that this should be made mutually exclusive once we open up other options.

docs/book/src/topics/node-outbound-lb.md
```diff
@@ -0,0 +1,64 @@
+# Node Outbound Load Balancer
+
+This document describes how to configure your clusters' node outbound load balancer.
+
+### Public Clusters
+
+For public clusters ie. clusters with api server load balancer type set to `Public`, CAPZ automatically configures a node outbound load balancer with the default settings.
+
+To provider custom settings for the node outbound load balacer, use the `nodeOutboundLB` section in cluster configuration.
```
CecileRobertMichon, line 9: To provide

docs/book/src/topics/node-outbound-lb.md
```diff
@@ -0,0 +1,64 @@
...
+      type: Internal
+    nodeOutboundLB:
+      frontendIPsCount: 1
+```
+
+<aside class="note warning">
+
+<h1> Warning </h1>
+
+Only `frontendIPsCount` is allowed to be configured for any node outbound load balancer. Trying to modify any other value will result in a validation error.
```
CecileRobertMichon, line 62: duplicate from above

docs/book/src/topics/node-outbound-lb.md
```diff
@@ -0,0 +1,64 @@
...
+
+</aside>
+
+### Private Clusters
+
+For private clusters ie. clusters with api server load balancer type set to `Internal`, CAPZ does not create a node outbound load balancer by default.
```
CecileRobertMichon, line 38: > For private clusters ie. clusters with api server load balancer type set to `Internal`, CAPZ does not create a node outbound load balancer by default

I don't think that's true. CAPZ always creates a node outbound LB to give outbound access to the worker nodes. For private clusters, it also creates a control plane outbound LB to give outbound access to the control plane nodes (in non-private clusters the API Server LB is public so it already gives outbound to the control planes).

docs/book/src/topics/node-outbound-lb.md
```diff
@@ -0,0 +1,64 @@
...
+
+This document describes how to configure your clusters' node outbound load balancer.
+
+### Public Clusters
+
+For public clusters ie. clusters with api server load balancer type set to `Public`, CAPZ automatically configures a node outbound load balancer with the default settings.
+
+To provider custom settings for the node outbound load balacer, use the `nodeOutboundLB` section in cluster configuration.
+
+Here is an example of a node outbound load balancer with `frontendIPsCount` set to 3. CAPZ will read this value and create 3 front end ips for this load balancer.
```
CecileRobertMichon, line 11: Let's include a node about why a user would want to increase the number of frontent IPs, something like:

> You may want more than one outbound IP address if you are running a large cluster that is processing lots of connections. See [here](https://docs.microsoft.com/en-us/azure/load-balancer/load-balancer-outbound-connections#multifesnat) for more documentation about how adding more outbound IP addresses can increase the number of SNAT ports available for use by the Standard Load Balancer in your cluster.

api/v1alpha4/azurecluster_default.go
```diff
@@ -148,6 +151,47 @@ func (c *AzureCluster) setAPIServerLBDefaults() {
 	}
 }
 
+func (c *AzureCluster) setNodeOutboundLBDefaults() {
+	if c.Spec.NetworkSpec.NodeOutboundLB == nil {
```
CecileRobertMichon, line 155: Not sure we want this (I believe it's a change from current behavior), see my comment below in docs

api/v1alpha4/types.go
nader-ziada: sorry for the late reply, missed this comment earlier, agreed its confusing, 

I agree that a new type for `NodeOutboundLbSpec` that would include the count (as mnetioned above) is much easier to understand, even if there duplication of fields with `LoadBalancerSpec`

and should be enforced by  validation

docs/book/src/topics/node-outbound-lb.md
shysank: IIUC, this pr changes the default behavior because we now want the ability to disable node outbound LB for private clusters by not including `nodeOutboundLB` section. So, unless the user explicitly sets nodeOutboundLB, capz will not create a node outbound lb. That's what I've tried to explain here. wdyt?

api/v1alpha4/azurecluster_default.go
shysank: Same comment as above about having the ability to disable node outbound lb for private clusters.

docs/book/src/topics/node-outbound-lb.md
CecileRobertMichon: Got it, I didn't realize you were including that change in this PR. In that case it also partly addresses #1159 right?

shysank: Yes, partially. I'll update the pr description to include this change.

CecileRobertMichon (commented):
/lgtm

great work @shysank, thanks for tackling this

nader-ziada:
experimented with this PR for a little bit and all looks good

/lgtm
/approve

PR #1233, opened 2021-03-12
Cleanup deprecated variables from v1alpha4

### LINKED ISSUE #1199: Audit "deprecated" variables/functions in v1alpha4
/kind feature

- Audit and cleanup any `deprecated` fields or functions as part of the main v1alpha4 branch

shysank:
**What type of PR is this?**
/kind cleanup

Cleanup deprecated variables from v1alpha4.

Fixes #1199 

The following deprecated variables were NOT removed since they were used in the code:
1. ~CidrBlock~
2. ~AvailabilityZone~

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Cleanup deprecated variables: AzureMachineSpec.Location, SubnetSpec.InternalLBIPAddress, VnetSpec.CidrBlock, AzureCluster.AvailabilityZone
```


CecileRobertMichon:
We should be able to also remove `CidrBlock` and `AvailabilityZone` as well since the only code that's using it should be for back compat reasons

CecileRobertMichon:
can you please make the release note more explicit and list the variables that are getting removed?

nader-ziada:
@shysank two quick questions 

 - I think we want to remove `CidrBlock` and `AvailabilityZone`, can we remove then from the code and use the new fields instead?
 
 - and if we remove these fields, we would also have to convert the removed types from v1alpha3 to the new types in v1alpha4 as part of the conversion webhook, right?

shysank:
> @shysank two quick questions
> 
> * I think we want to remove `CidrBlock` and `AvailabilityZone`, can we remove then from the code and use the new fields instead?

Yep, fixed it. As @CecileRobertMichon pointed out, those were mostly backward compatibility code. So, I've removed it. PTAL.

> * and if we remove these fields, we would also have to convert the removed types from v1alpha3 to the new types in v1alpha4 as part of the conversion webhook, right?

So, this is kind of an open question. Since we have removed deprecated variables from v1alpha4, `spoke-hub-spoke` tests would fail now because we don't preserve data when migrating from `v1alpha3` -> `v1alpha4`. So, when it is migrated back to `v1alpha3`, it wont have the deprecated fields. There are two solutions:
1. Preserve data like we do when we add new fields in newer spec.
2. Fix tests by overriding the fuzz funcs (This is what I've done). This is kind of hacky, but I thought doing (1) for deprecated variables seemed kind of an overkill. 

Please let me know you thoughts.

nader-ziada:
> 1. Preserve data like we do when we add new fields in newer spec.
> 2. Fix tests by overriding the fuzz funcs (This is what I've done). This is kind of hacky, but I thought doing (1) for deprecated variables seemed kind of an overkill.
> 
> Please let me know you thoughts.
> 
> On a related note, I remember disabling spoke-hub-spoke tests in capi, not sure how it's still there.

the tests enabled back by a recent PR, and for the conversion, since v1alpha3 is already out there and a lot of folks use, if they upgrade to v1alpha4, we should preserve the deprecated fields, so we can put them back. Is this too complicated? what do you think?

shysank:
> since v1alpha3 is already out there and a lot of folks use, if they upgrade to v1alpha4, we should preserve the deprecated fields, so we can put them back. Is this too complicated? what do you think?

It should not be hard to do now since there is no preservation happening. But as we update v1alpha4 spec, we'll run into problems because there'll be a conflict in handling the extra variables. Until now, we serialize and save the spec when downgrading, and do the other way around during upgrade. Now we'll have to do it in the other direction as well. Does that make sense?

nader-ziada:
> > since v1alpha3 is already out there and a lot of folks use, if they upgrade to v1alpha4, we should preserve the deprecated fields, so we can put them back. Is this too complicated? what do you think?
> 
> It should not be hard to do now since there is no preservation happening. But as we update v1alpha4 spec, we'll run into problems because there'll be a conflict in handling the extra variables. Until now, we serialize and save the spec when downgrading, and do the other way around during upgrade. Now we'll have to do it in the other direction as well. Does that make sense?

We have to do it when we introduce breaking changes, but I'm not sure what you mean by the other direction

shysank:
> We have to do it when we introduce breaking changes, but I'm not sure what you mean by the other direction

So  the conversion logic once we add breaking changes generally looks like this:

```
func (src *AzureCluster) ConvertTo(dstRaw conversion.Hub) error {
	dst := dstRaw.(*v1alpha4.AzureCluster)
       .....
        // Manually restore data.
	restored := &v1alpha4.AzureCluster{}
	if ok, err := utilconversion.UnmarshalData(src, restored); err != nil || !ok {
		return err
	}
        .....
	return nil
}

func (dst *AzureCluster) ConvertFrom(srcRaw conversion.Hub) error {
	src := srcRaw.(*v1alpha4.AzureCluster)
        ....
        // Preserve Hub data on down-conversion.
	if err := utilconversion.MarshalData(src, dst); err != nil {
		return err
	}
	return nil
}
```

If we are to preserve data during up-conversion (for preserving deprecated fields), we'll have to do `utilconversion.MarshalData` in `ConvertTo` method, and `utilconversion.UnmarshalData` in `ConvertFrom`, which is kind of conflicting.

Is my understanding correct?

nader-ziada:
> If we are to preserve data during up-conversion (for preserving deprecated fields), we'll have to do `utilconversion.MarshalData` in `ConvertTo` method, and `utilconversion.UnmarshalData` in `ConvertFrom`, which is kind of conflicting.
> 
> Is my understanding correct?

so here is how I understand this:
- if you are converting from v1alpha3 to v1alpha4 and there is a field that got deprecated, you set the new field. For example: `cidrBlock` is removed, so you set it on `cidrBlocks`
- if you are converting from v1alpha4 to v1alpha3, then you set it on v1alpha3 field if needed

nader-ziada:
After having a chat with @shysank, we both agree the deprecated fields should not be part of the conversion since they should not be used in `v1alpha3` anyways and was only there for backward compatibility with older versions

nader-ziada (commented):
looks like a valid failure for e2e
```
error: error validating "STDIN": error validating data: ValidationError(AzureMachineTemplate.spec.template.spec): unknown field "location" in io.x-k8s.cluster.infrastructure.v1alpha4.AzureMachineTemplate.spec.template.spec; if you choose to ignore these errors, turn validation off with --validate=false
```

you probably need to check the validation webhook

azure/scope/machine.go
```diff
@@ -269,9 +269,6 @@ func (m *MachineScope) AvailabilityZone() string {
 	if m.AzureMachine.Spec.FailureDomain != nil {
 		return *m.AzureMachine.Spec.FailureDomain
 	}
```
nader-ziada, line 271: small nit

remove the comment on this func about `AzureMachine.Spec.AvailabilityZone.ID (This is DEPRECATED)`

nader-ziada (commented):
can we also update the note in the doc here: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/docs/book/src/topics/failure-domains.md#explicit-placement 

+ the one small nit and I think it would be ready to go

nader-ziada:
I think you need to rebase on #1260 to get the windows e2e working

nader-ziada:
/lgtm

@CecileRobertMichon any comments?

CecileRobertMichon (commented):
/lgtm
/approve

thanks @shysank!

PR #1244, opened 2021-03-16
AzureEnvironment as part of the AzureClusterSpec instead of an environment variable 

### LINKED ISSUE #1227: Move AZURE_ENVIRONMENT to AzureCluster spec
/kind feature

Moving AZURE_ENVIRONMENT to AzureCluster spec will allow capz `infrastructure-components.yaml` to be deployed without requiring `envsubst` to set a default value

source: https://github.com/kubernetes-sigs/cluster-api/issues/3881#issuecomment-743370519 

**Environment:**

nader-ziada:
**What type of PR is this?**
/kind feature
/kind api-change

-  Moving AZURE_ENVIRONMENT to AzureCluster spec will allow capz infrastructure-components.yaml to be deployed without requiring envsubst to set a default value

Fixes #1227

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Set AzureEnvironment as part of the AzureClusterSpec instead of an environment variable. This is a BREAKING CHANGE to users not using AzurePublicCloud.
action required: Users will need to update the AzureCluster resource to include the appropriate value of AzureEnvironment 
```


devigned (changes requested):
Nice!

One thing that would be good to add here is a update hook to reject changes to the `AzureEnvironment`. To change this field would imply a delete of the cluster and rebuilding the cluster in another Azure cloud. I don't think we want to enable that.

nader-ziada:
> One thing that would be good to add here is a update hook to reject changes to the `AzureEnvironment`. To change this field would imply a delete of the cluster and rebuilding the cluster in another Azure cloud. I don't think we want to enable that.

good catch, didn't think about that, but makes perfect sense

devigned (approved):
/lgtm
/assign @CecileRobertMichon

nader-ziada:
oh wait, forgot to update relevant docs as well
/hold

devigned (approved):
/lgtm

config/default/manager_cloud_env_patch.yaml
```diff
@@ -1,16 +0,0 @@
...
-metadata:
-  name: controller-manager
-  namespace: system
-spec:
-  template:
-    spec:
-      containers:
-      - name: manager
-        env:
-          - name: AZURE_ENVIRONMENT
```
CecileRobertMichon, line 12: By removing this we're essentially making it not supported to set the AZURE_ENVIRONMENT value via en env variable right? Wasn't the idea to keep it as optional for back compat?

nader-ziada: If we keep it we would still need to use envsubst with the manifest, right?

CecileRobertMichon: Ok but this is a breaking change then for users not using AzurePublicCloud, let's make sure the release note calls that out

nader-ziada: should we leave this in and just not set a default for it using envsubst? would that be easier for not breaking users

CecileRobertMichon: > should we leave this in and just not set a default for it using envsubst? would that be easier for not breaking users

that would make it a required variable with clusterctl which might impact even more users (ie. everyone in AzurePublicCloud relying on the default, like quickstart users).

I think removing it and adding a clear release note for non-public cloud users to set the value is the best course of action.

PR #1261, opened 2021-03-23
ensure cluster is populated during a failed e2e test

devigned:
**What type of PR is this?**
/kind bug

During an e2e test where the cluster fails to provision the control plane, the cluster is nil since the ptr to the cluster will never be set. This PR fixes this by setting up the `clusterctl.ApplyClusterTemplateAndWaitResult` at a higher scope so that the `result.Cluster` field will be non-nil when trying to dump resource definitions at the end of the test.

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```


CecileRobertMichon:
/lgtm

thank you 🙏

nader-ziada:
you will have to wait for #1260 and rebase to get this in

Issue #664, opened 2020-06-01
Allow multiple subnets of role "node"


CecileRobertMichon:
/kind feature

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]
Related to https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/643 and https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/618.

Currently, CAPZ expects and only allows one subnet per "role", ie. one control plane subnet and one node subnet. In the future, there might be use cases where users need multiple subnets, one for each MachinePool for example. 

The challenges/ current blockers for supporting this on the top of my head are:

1. Subnets are defined at Cluster level, there isn't currently a good way to map subnets to machines besides their `Role`, which only allows one grouping for all Machines that are not control planes.
2. The `Subnets` type is `[]*SubnetSpec`, ie. a slice of pointers. This makes it difficult to handle operations on subnets where we need to "get" the subnet for a particular role. A slice of pointers is also not very go idiomatic and makes conversion difficult. See #618 for further discussion on how to change that type.
3. Currently, the AzureCluster reconciler only creates 2 subnets and will ignore any additional subnets. We need to change this to allow specifying more than 2 subnets without breaking the BYO vnet use case.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- cluster-api-provider-azure version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

alexeldeib:
cloud provider doesn't care about subnets except for ILB I think, based on my investigation in #773. Machines/MachinePools should probably be able to target a specific subnet (especially for the BYO scenario).

We would then only use the subnets in the cluster spec as an indication of what to provision. We could even select all machines owned by this cluster and ensure the corresponding subnets exist, but I don't think inverting that control is a good idea.

If we let machines target subnets directly, we should probably also have defaulting behavior so users don't **have** to fully qualify it every time.

rhockenbury:
Has there been any related updates on this issue?  My use case requires provisioning worker pools in different subnets.  For reference/context, the vsphere cluster api allows network settings to be defined in the machine template.

CecileRobertMichon:
@rhockenbury not yet, this is planned for v1alpha4

/milestone v0.5.x

rhockenbury:
/remove-lifecycle rotten

Still interested in this.

[...]

PR #1144, opened 2021-01-27
v1alpha4 types

nader-ziada:
**What type of PR is this?**

/kind feature
/kind api-change

**What this PR does / why we need it**:
- Generates types for v1alpha4
- Remove v1alpha2 types
- Updates imports to v1alpha4 packages

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #961 

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**TODOs**:

- [x] update min go version to 1.15.0
- [x] update controller-runtime to v0.8.1 when its merged in [capi](https://github.com/kubernetes-sigs/cluster-api/pull/4109) ?
- [x] update kind to v0.9.0
- [ ] what to do about v1alpha3 templates?

**Release note**:

```release-note
Add v1alpha4 types  
```

PR #1280, opened 2021-03-30
⚠️  Change Subnets type to map[string]SubnetSpec

CecileRobertMichon:
**What type of PR is this?**
/kind api-change

**What this PR does / why we need it**: Change `Subnets` type from `[]SubnetSpec` to `map[string]SubnetSpec`.
- Allows for easier retrieving of subnets, instead of having to loop through the map and comparing names
- Enforces the uniqueness of name requirement implicitly
- Makes #664 possible by not relying on "NodeSubnet()" and "ControlPlaneSubnet()" anymore in services

Fixes partially #618 

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Changed Subnets type to map[string]SubnetSpec
Action required: AzureCluster.Spec.NetworkSpec.Subnets should now be of type map[string]SubnetSpec. See https://capz.sigs.k8s.io/topics/custom-vnet.html for examples. Existing clusters will be automatically converted to use the new type.
```

### DIFF
--- a/azure/services/routetables/routetables.go
+++ b/azure/services/routetables/routetables.go
@@ -72,7 +72,7 @@ func (s *Service) Reconcile(ctx context.Context) error {
 			// currently don't support specifying your own routes via spec
 			routeTableSpec.Subnet.RouteTable.Name = to.String(existingRouteTable.Name)
 			routeTableSpec.Subnet.RouteTable.ID = to.String(existingRouteTable.ID)
-
+			s.Scope.SetSubnet(routeTableSpec.SubnetName, routeTableSpec.Subnet)
 			continue
 		}
 
--- a/azure/services/subnets/subnets.go
+++ b/azure/services/subnets/subnets.go
@@ -63,25 +63,16 @@ func (s *Service) Reconcile(ctx context.Context) error {
 			return errors.Wrapf(err, "failed to get subnet %s", subnetSpec.Name)
 		case err == nil:
 			// subnet already exists, update the spec and skip creation
-			var subnet infrav1.SubnetSpec
-			if subnetSpec.Role == infrav1.SubnetControlPlane {
-				subnet = s.Scope.ControlPlaneSubnet()
-			} else if subnetSpec.Role == infrav1.SubnetNode {
-				subnet = s.Scope.NodeSubnet()
-			} else {
-				continue
-			}
-
+			subnet := s.Scope.Subnet(subnetSpec.Name)
 			subnet.Role = subnetSpec.Role
-			subnet.Name = existingSubnet.Name
 			subnet.CIDRBlocks = existingSubnet.CIDRBlocks
 			subnet.ID = existingSubnet.ID
+			s.Scope.SetSubnet(subnetSpec.Name, subnet)
 
 		case !s.Scope.IsVnetManaged():
 			return fmt.Errorf("vnet was provided but subnet %s is missing", subnetSpec.Name)
 
 		default:
-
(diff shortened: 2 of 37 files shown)


azure/services/routetables/routetables.go
```diff
@@ -72,7 +72,7 @@ func (s *Service) Reconcile(ctx context.Context) error {
 			// currently don't support specifying your own routes via spec
 			routeTableSpec.Subnet.RouteTable.Name = to.String(existingRouteTable.Name)
 			routeTableSpec.Subnet.RouteTable.ID = to.String(existingRouteTable.ID)
-
+			s.Scope.SetSubnet(routeTableSpec.SubnetName, routeTableSpec.Subnet)
```
CecileRobertMichon, line 75: sneaked in a bug fix: subnet wasn't getting updated since SubnetSpec was changed to not be a pointer in #1144

azure/services/subnets/subnets.go
```diff
@@ -63,25 +63,16 @@ func (s *Service) Reconcile(ctx context.Context) error {
...
-			} else {
-				continue
-			}
-
+			subnet := s.Scope.Subnet(subnetSpec.Name)
 			subnet.Role = subnetSpec.Role
-			subnet.Name = existingSubnet.Name
 			subnet.CIDRBlocks = existingSubnet.CIDRBlocks
 			subnet.ID = existingSubnet.ID
+			s.Scope.SetSubnet(subnetSpec.Name, subnet)
```
CecileRobertMichon, line 70: same bug here: subnet wasn't getting updated since SubnetSpec was changed to not be a pointer in #1144

CecileRobertMichon:
@devigned pointed out that maps are not recommended in k8s APIs.  Instead, the convention is to use a list of subobjects containing a name field, which is what we are doing right now for subnets. IMO maps are a better fit for this particular use case for the reasons I mentioned above in the PR description. Particularly, we end up having to mimick a map when using a list and doing things like [putting all the names in a map](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1280/files#diff-a71ed33d224ff367e4e81a0e37aeb109f15b49cea3b6b535ed1deb33432da004L137) to validate the uniqueness of names. That being said, the API conventions are there for a reason and we should try to follow them as much as possible. Holding this for discussion since it's more controversial of a change than I initially thought.

> Using maps in Kubernetes API resources at all is not recommended (as per the sig-architecture api-conventions doc):
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#lists-of-named-subobjects-preferred-over-maps

https://github.com/kubernetes-sigs/controller-tools/pull/518#issuecomment-748376880 

/hold

shysank:
+1 for following conventions

> Allows for easier retrieving of subnets, instead of having to loop through the map and comparing names
> Enforces the uniqueness of name requirement implicitly
> Makes #664 possible by not relying on "NodeSubnet()" and "ControlPlaneSubnet()" anymore in services

For 1 and 2, I think we can write wrapper structs and write nice apis to make it more readable. For 3, I don't fully understand the use case yet, I'll read more about it. But in general, unless there is a performance bottleneck that we are facing when not using maps, I'm in favor of following conventions over maps.

CecileRobertMichon:
Thanks for the input @shysank, I agree with your analysis. Going to close this for now. Hopefully, we can find other ways to address 3) when it comes up.

/close

Issue #1286, opened 2021-03-30
Log dumping script for Windows clusters


chewong:
/kind feature

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]

Our log dumping script (https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/hack/log/log-dump.sh) should support Linux as well as Windows clusters 

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- cluster-api-provider-azure version: N/A
- Kubernetes version: (use `kubectl version`): N/A
- OS (e.g. from `/etc/os-release`): N/A

jsturtevant:
took a look and we are collecting some windows logs such as kubeproxy and flannel cni logs: https://gcsweb.k8s.io/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/1296/pull-cluster-api-provider-azure-e2e-windows/1387118255923007488/artifacts/clusters/capz-e2e-hy4mf8/kube-system/

there also seems to be some logs attempted to be collected but looks like assumptions were linux as most files are blank: https://gcsweb.k8s.io/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/1296/pull-cluster-api-provider-azure-e2e-windows/1387118255923007488/artifacts/clusters/capz-e2e-hy4mf8/machines/capz-e2e-hy4mf8-md-win-8678b7dd8-5btwg/ 

These files are collected by https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/test/e2e/azure_logcollector.go not the https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/hack/log/log-dump.sh which is only used in a few jobs where the e2e framework is not used (azure disk etc).  

Going to add Windows support to the `azure_logcollector.go`

PR #1296, opened 2021-04-08
Update Windows ssh configuration

jsturtevant:
**What type of PR is this?**
/kind bug

Image builder added support ssh configuration to the image which breaks the way we were configuring ssh.  This copies the ssh public key to the correct location and aligns the Windows ssh approach with other providers.

https://github.com/kubernetes-sigs/image-builder/blob/4b1de420ab2af76bac6580218b565e037904f5f9/images/capi/ansible/windows/roles/systemprep/tasks/main.yml#L94-L95

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Fix ssh for newer image builder images on windows
```

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -21,19 +21,20 @@
     - [Tilt for dev in both CAPZ and CAPI](#tilt-for-dev-in-both-capz-and-capi)
     - [Deploying a workload cluster](#deploying-a-workload-cluster)
     - [Viewing Telemetry](#viewing-telemetry)
-  - [Instrumenting Telemetry](#instrumenting-telemetry)
-    - [Distributed Tracing](#distributed-tracing)
-    - [Metrics](#metrics)
   - [Manual Testing](#manual-testing)
     - [Creating a dev cluster](#creating-a-dev-cluster)
       - [Building and pushing dev images](#building-and-pushing-dev-images)
       - [Customizing the cluster deployment](#customizing-the-cluster-deployment)
       - [Creating the cluster](#creating-the-cluster)
+  - [Instrumenting Telemetry](#instrumenting-telemetry)
+    - [Distributed Tracing](#distributed-tracing)
+    - [Metrics](#metrics)
   - [Submitting PRs and testing](#submitting-prs-and-testing)
     - [Executing unit tests](#executing-unit-tests)
   - [Automated Testing](#automated-testing)
     - [Mocks](#mocks)
     - [E2E Testing](#e2e-testing)
+    - [Running custom test suites on CAPZ clusters](#running-custom-test-suites-on-capz-clusters)
     - [Conformance Testing](#conformance-testing)
 
 <!-- /TOC -->
@@ -292,7 +293,10 @@ SSH_KEY_FILE=.sshkey
 rm -f "${SSH_KEY_FILE}" 2>/dev/null
 ssh-keygen -t rsa -b 2048 -f "${SSH_KEY_FILE}" -N '' 1>/dev/null
 echo "Machine SSH key generated in ${SSH_KEY_FILE}"
+# For Linux the ssh key needs to be b64 encoded because we use the azure api to set it
+# Windows doesn't support setting ssh keys so we use cloudbase-int to set which doesn't require base64 
 export AZURE_SSH_PUBLIC_KEY_B64=$(cat "${SSH_KEY_FILE}.pub" | base64 | tr -d '\r\n')
+export AZURE_SSH_PUBLIC_KEY=$(cat "${SSH_KEY_FILE}.pub" | tr -d '\r\n')
 ```
 
 ⚠️ Please note the generated templates include default values and therefore require the use of `clusterctl` to create the cluster
--- a/templates/cluster-template-machinepool-windows.yaml
+++ b/templates/cluster-template-machinepool-windows.yaml
@@ -252,15 +252,6 @@ spec:
       New-HnsNetwork -Type Overlay -AddressPrefix "192.168.255.0/30" -Gateway "192.168.255.1" -Name "External" -AdapterName "Ethernet 2" -SubnetPolicies @(@{Type = "VSID"; VSID = 9999; })
     path: C:/create-external-network.ps1
     permissions: "0744"
-  - content: |
-      [System.Text.Encoding]::UTF8.GetString([System.Convert]::FromBase64String("${AZURE_SSH_PUBLIC_KEY_B64:=''}")) | Add-Content C:/ProgramData/ssh/administrators_authorized_keys
-      icacls C:/ProgramData/ssh/administrators_authorized_keys /remove "NT AUTHORITY\Authenticated Users"
-      icacls C:/ProgramData/ssh/administrators_authorized_keys /inheritance:r
-      icacls C:/ProgramData/ssh/administrators_authorized_keys /grant SYSTEM:F
-      icacls C:/ProgramData/ssh/administrators_authorized_keys /grant BUILTIN\Administrators:F
-      restart-service sshd
-    path: C:/configure-ssh.ps1
-    permissions: "0744"
   joinConfiguration:
     nodeRegistration:
       kubeletExtraArgs:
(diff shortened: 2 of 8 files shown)


CecileRobertMichon:
lgtm but I don't have any windows expertise, @jsturtevant please assign someone who does for review

templates/cluster-template-windows.yaml
```diff
@@ -262,11 +262,8 @@ spec:
         path: C:/create-external-network.ps1
         permissions: "0744"
       - content: |
-          [System.Text.Encoding]::UTF8.GetString([System.Convert]::FromBase64String("${AZURE_SSH_PUBLIC_KEY_B64:=''}")) | Add-Content C:/ProgramData/ssh/administrators_authorized_keys
-          icacls C:/ProgramData/ssh/administrators_authorized_keys /remove "NT AUTHORITY\Authenticated Users"
```
marosset, line 266: Is there a reason to remove updated the perms on the authorized_keys file?

I think the goal was to ensure that a pod couldn't mount c:/ProgramData/ssh/ and steal the SSH keys used to connect to the node.
Can we double check that containers cannot mount c:/users/capi/.ssh/ and access the files?

perithompson:
Doesn't this work for capz? I need to check the capv provider but this is all I'm providing? 


users:
      - name: capi
        groups: Administrators
        sshAuthorizedKeys:
        - $SSH_AUTHORIZED_KEY
        sudo: ALL=(ALL) NOPASSWD:ALL

templates/cluster-template-windows.yaml
jsturtevant: these are only public keys so stealing them won't do much. I can verify that a pod can't drop a key there.

jsturtevant:
@perithompson where are you putting that data? There is no way for Azure to set Windows keys for VM's so it might be different?

jsturtevant:
Thanks! That plugin is enabled be default on image-builder so should be able to use that approach. https://github.com/kubernetes-sigs/image-builder/blob/2f4cfed5257aefca3048ba22d3f30546b4e4a9bf/images/capi/ansible/windows/roles/cloudbase-init/templates/cloudbase-init.conf#L34-L36

I think it is User section of the Userdata docs that does this: https://cloudbase-init.readthedocs.io/en/latest/userdata.html#cloud-config

marosset:
other than the above comments about making sure pods can't add public SSH keys to the node's authorirzed_hosts file these changes look good to me

jsturtevant:
As long as you are using `ContainerUser`, which is recommended for most scenarios, (or another user that is not `ContainerAdministrator` or part of the admin group in the container), you will not have access to these files.  This is the same on a Linux machine, if I am root and mount the host folder I can modify these files.

Using the following spec:

```
apiVersion: v1
kind: Pod
metadata:
  name: hostdir
spec:
  containers:
  - name: test
    image: e2eteam/busybox:1.29
    securityContext:
      windowsOptions:
        runAsUserName: ContainerUser
    command:
      - sleep
      - "3600"
    volumeMounts:
    - name: test-volume
      mountPath: /test-volume
  restartPolicy: Always
  volumes:
  - name: test-volume
    hostPath:
        path: "C:/"
        type: DirectoryOrCreate
  nodeSelector:
    "kubernetes.io/os": windows
```

accessing folders:
```
PS C:\test-volume\ProgramData\ssh> whoami
ContainerUser
PS C:\test-volume\ProgramData\ssh> echo test > sshd_config
out-file : Access to the path 'C:\test-volume\ProgramData\ssh\sshd_config' is denied.
At line:1 char:1
+ echo test > sshd_config
+ ~~~~~~~~~~~~~~~~~~~~~~~
+ CategoryInfo          : OpenError: (:) [Out-File], UnauthorizedAccessException
+ FullyQualifiedErrorId : FileOpenFailure,Microsoft.PowerShell.Commands.OutFileComman
```

```
PS C:\test-volume\Users\capi> ls
ls : Access to the path 'C:\test-volume\Users\capi' is denied.
At line:1 char:1
+ ls
+ ~~
+ CategoryInfo          : PermissionDenied: (C:\test-volume\Users\capi:String) [Get-ChildItem], UnauthorizedAccessException
+ FullyQualifiedErrorId : DirUnauthorizedAccessError,Microsoft.PowerShell.Commands.GetChildItemCommand
 
PS C:\test-volume\Users\capi> echo test > .ssh/file
out-file : Access to the path 'C:\test-volume\Users\capi\.ssh\file' is denied.
At line:1 char:1
+ echo test > .ssh/file
+ ~~~~~~~~~~~~~~~~~~~~~
+ CategoryInfo          : OpenError: (:) [Out-File], UnauthorizedAccessException
+ FullyQualifiedErrorId : FileOpenFailure,Microsoft.PowerShell.Commands.OutFileCommand
 
PS C:\test-volume\Users\capi> whoami
ContainerUser
```

In addition you can use a psp (though there is a replacement coming) and it will block either running as ContainerAdmin or mounting host dirs:

hostvolume psp:

```
error from server (Forbidden): error when creating "/home/jstur/projects/windows-k8s-playground/deployments/volumes/hostvolume.yaml": pods "hostdir" is forbidden: PodSecurityPolicy: unable to admit pod: [spec.volumes[0]: Invalid value: "hostPath": hostPath volumes are not allowed to be used]
```

`MustRunAsNonRoot` psp:

```
 Warning  Failed     62s (x8 over 2m31s)  kubelet            Error: container's runAsUser (ContainerAdministrator) which will be regarded as root identity and will break non-root policy (pod: "test_psp-example(a62cc596-7e57-4e83-8d68-f69f0d44f7d8)", container: test)
```

templates/cluster-template-windows.yaml
jsturtevant: see comment below https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1296#issuecomment-817014053

jsturtevant:
I think this is ready to go.  I've addressed the remaining comments on the access to the ssh keys.  Please let me know if there is anything else.

perithompson:
Good to be using the cloudbase functionality for ssh, thanks @jsturtevant!

/lgtm

docs/development.md
```diff
@@ -293,6 +293,7 @@ rm -f "${SSH_KEY_FILE}" 2>/dev/null
 ssh-keygen -t rsa -b 2048 -f "${SSH_KEY_FILE}" -N '' 1>/dev/null
 echo "Machine SSH key generated in ${SSH_KEY_FILE}"
 export AZURE_SSH_PUBLIC_KEY_B64=$(cat "${SSH_KEY_FILE}.pub" | base64 | tr -d '\r\n')
+export AZURE_SSH_PUBLIC_KEY=$(cat "${SSH_KEY_FILE}.pub" | tr -d '\r\n')
```
CecileRobertMichon, line 296: why do we need both B64 and not B64?

templates/cluster-template-machinepool-windows.yaml
```diff
@@ -270,4 +261,8 @@ spec:
       name: '{{ ds.meta_data["local_hostname"] }}'
   preKubeadmCommands:
   - powershell c:/create-external-network.ps1
-  - powershell c:/configure-ssh.ps1
+  users:
+  - groups: Administrators
+    name: capi
+    sshAuthorizedKeys:
+    - ${AZURE_SSH_PUBLIC_KEY:=""}
```
CecileRobertMichon, line 268: is this is expecting plain text (not base64 encoded)? 😕

jsturtevant: yup. I had the same expression when I figured that out.  We push the ssh key through the Azure api but we could use the cloud-init for linux in the same way.

CecileRobertMichon: @jsturtevant what happens when this is empty? For Linux there is a defaulting webhook that sets a random key but here that won't be the case... Do we need to update any e2e / conformance tests to have that var set?

docs/development.md
CecileRobertMichon: can you add a small comment here on why it needs both? I could see this being surprising to new devs in the future

templates/cluster-template-machinepool-windows.yaml
jsturtevant: the sshkey will be blank:
``` 
 users:
  - groups: Administrators
    name: capi
    sshAuthorizedKeys:
    - ""
```

> Do we need to update any e2e / conformance tests to have that var set?

We don't currently collect the windows logs in e2e: https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1286

docs/development.md
jsturtevant: done

PR #1299, opened 2021-04-09
Add support for custom egress rules

CecileRobertMichon:
**What type of PR is this?**

- Renames `ingressRule` to `securityRules` (note there was a missing s in the field name which represented a list of ingress rules) to be more consistent with Azure naming, and to include rules in both directions
- Adds a `direction` field to `SecurityRule` which can be either `Inbound` or `Outbound`. Defaults to `Inbound` to preserve existing behavior.

Feature request came out of [this slack thread](https://kubernetes.slack.com/archives/CEX9HENG7/p1617886293184900?thread_ts=1617708083.161200&cid=CEX9HENG7)

Fixes partly #618  

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Rename ingressRule to securityRules
Add optional direction field to securityRules, defaults to "Inbound"
```


cpanato (approved):
/lgtm

azure/services/securitygroups/securitygroups.go
```diff
@@ -77,10 +77,10 @@ func (s *Service) Reconcile(ctx context.Context) error {
 			// Check if the expected rules are present
 			update := false
 			securityRules = *existingNSG.SecurityRules
-			for _, rule := range nsgSpec.IngressRules {
-				if !ruleExists(securityRules, converters.IngresstoSecurityRule(rule)) {
+			for _, rule := range nsgSpec.SecurityRules {
+				if !ruleExists(securityRules, converters.SecurityRuleToSDK(rule)) {
```
whites11, line 81: Maybe worth storing the result of `converters.SecurityRuleToSDK(rule)` to a local variable to avoid calling it twice?

whites11 (approved):
For what it's worth LGTM

api/v1alpha3/azurecluster_conversion.go
```diff
@@ -223,14 +237,30 @@ func Convert_v1alpha3_SecurityGroup_To_v1alpha4_SecurityGroup(in *SecurityGroup,
...
+// Convert_v1alpha3_IngressRule_To_v1alpha4_SecurityRule
+func Convert_v1alpha3_IngressRule_To_v1alpha4_SecurityRule(in *IngressRule, out *infrav1alpha4.SecurityRule, _ apiconversion.Scope) error { //nolint
+	out.Name = in.Name
+	out.Description = in.Description
+	out.Protocol = v1alpha4.SecurityGroupProtocol(in.Protocol)
+	out.Priority = in.Priority
+	out.SourcePorts = in.SourcePorts
+	out.DestinationPorts = in.DestinationPorts
+	out.Source = in.Source
+	out.Destination = in.Destination
```
shysank, line 249: Just to confirm, `out.Direction = Inbound` is not needed here because it's set in defaults?

api/v1alpha3/azurecluster_conversion.go
```diff
@@ -195,10 +209,10 @@ func Convert_v1alpha4_SecurityGroup_To_v1alpha3_SecurityGroup(in *infrav1alpha4.
 	out.ID = in.ID
 	out.Name = in.Name
 
-	out.IngressRules = make(IngressRules, len(in.IngressRules))
-	for i := range in.IngressRules {
+	out.IngressRules = make(IngressRules, len(in.SecurityRules))
+	for i := range in.SecurityRules {
 		out.IngressRules[i] = IngressRule{}
```
shysank, line 214: Shouldn't we check for `securityRule.Direction == Inbound` here?

api/v1alpha4/azurecluster_validation.go
```diff
@@ -204,10 +204,10 @@ func validateInternalLBIPAddress(address string, cidrs []string, fldPath *field.
 		fmt.Sprintf("Internal LB IP address needs to be in control plane subnet range (%s)", cidrs))
 }
 
-// validateIngressRule validates an IngressRule
-func validateIngressRule(ingressRule IngressRule, fldPath *field.Path) *field.Error {
-	if ingressRule.Priority < 100 || ingressRule.Priority > 4096 {
-		return field.Invalid(fldPath, ingressRule.Priority, "ingress priorities should be between 100 and 4096")
+// validateSecurityRule validates a SecurityRule
+func validateSecurityRule(rule SecurityRule, fldPath *field.Path) *field.Error {
+	if rule.Priority < 100 || rule.Priority > 4096 {
```
shysank, line 209: nit: can we extract the priority vals to constants?

api/v1alpha3/azurecluster_conversion.go
```diff
@@ -58,6 +58,20 @@ func (src *AzureCluster) ConvertTo(dstRaw conversion.Hub) error { // nolint
 	dst.Spec.NetworkSpec.APIServerLB.FrontendIPsCount = restored.Spec.NetworkSpec.APIServerLB.FrontendIPsCount
 	dst.Spec.NetworkSpec.NodeOutboundLB = restored.Spec.NetworkSpec.NodeOutboundLB
 
+	for i, restoredSubnet := range restored.Spec.NetworkSpec.Subnets {
+		for j, dstSubnet := range dst.Spec.NetworkSpec.Subnets {
+			if dstSubnet.Name == restoredSubnet.Name {
+				for k, restoredSecurityRule := range restoredSubnet.SecurityGroup.SecurityRules {
+					for l, dstSecurityRule := range dstSubnet.SecurityGroup.SecurityRules {
+						if dstSecurityRule.Name == restoredSecurityRule.Name {
+							dst.Spec.NetworkSpec.Subnets[j].SecurityGroup.SecurityRules[l].Direction = restored.Spec.NetworkSpec.Subnets[i].SecurityGroup.SecurityRules[k].Direction
```
shysank, line 67: break here?

api/v1alpha3/azurecluster_conversion.go
```diff
@@ -58,6 +58,20 @@ func (src *AzureCluster) ConvertTo(dstRaw conversion.Hub) error { // nolint
...
+						}
+					}
+				}
```
shysank, line 70: ditto: break here?

shysank: Not sure if it'll make a huge difference. I'm fine with skipping it, if it's going to make the code more complex since we'll need to assign the loop to a variable to break.

docs/book/src/topics/custom-vnet.md
```diff
@@ -63,13 +63,13 @@ If no CIDR block is provided, `10.0.0.0/8` will be used by default, with default
 
 Whenever using custom vnet and subnet names and/or a different vnet resource group, please make sure to update the `azure.json` content part of both the nodes and control planes' `kubeadmConfigSpec` accordingly before creating the cluster.
 
-### Custom Ingress Rules
+### Custom Security Rules
```
shysank, line 66: Should we add a note about ingress rules being renamed to security rules somewhere?

api/v1alpha3/azurecluster_conversion.go
CecileRobertMichon: correct

api/v1alpha3/azurecluster_conversion.go
CecileRobertMichon: so this causes the fuzzy conversion test to fail, which I guess makes sense since we're dropping rules 🤔  the issue is that on round-trip conversion the entire rule won't be restored

api/v1alpha3/azurecluster_conversion.go
shysank: Looking at it again, I think it should be a straightforward break since we are just breaking the inner loop. Sorry about the confusion.

CecileRobertMichon: added

api/v1alpha3/azurecluster_conversion.go
CecileRobertMichon: ok I think I fixed it & and improved the conversion. @shysank I kept the last commit separate to make it easier to review, let me know what you think about the conversion before I squash.

api/v1alpha3/azurecluster_conversion.go
CecileRobertMichon: added it anyways, doesn't hurt to make it explicit

api/v1alpha3/azurecluster_conversion.go
```diff
@@ -58,6 +57,32 @@ func (src *AzureCluster) ConvertTo(dstRaw conversion.Hub) error { // nolint
...
+				var restoredOutboundRules []infrav1alpha4.SecurityRule
+				for k, restoredSecurityRule := range restoredSubnet.SecurityGroup.SecurityRules {
+					if restoredSecurityRule.Direction == infrav1alpha4.SecurityRuleDirectionInbound {
+						// For inbound rules, we add back the restored direction if the rule still exists.
+						for l, dstSecurityRule := range dstSubnet.SecurityGroup.SecurityRules {
+							if dstSecurityRule.Name == restoredSecurityRule.Name {
+								dst.Spec.NetworkSpec.Subnets[j].SecurityGroup.SecurityRules[l].Direction = restored.Spec.NetworkSpec.Subnets[i].SecurityGroup.SecurityRules[k].Direction
+								break
+							}
+						}
```
shysank, line 74: Isn't this logic redundant since inbound rules will get auto converted (now that you've set direction as well in line 263)? IIUC, I think we can simplify this to just handle outbound rules:
```
if restoredSecurityRule.Direction == infrav1alpha4.SecurityRuleDirectionOutbound {
  restoredOutboundRules = append(restoredOutboundRules, restoredSecurityRule)
}
```

wdyt?

api/v1alpha3/azurecluster_conversion.go
shysank: Thanks @CecileRobertMichon for fixing the conversion and adding comments! I've just added one small feedback to the conversion code, ptal.

api/v1alpha3/azurecluster_conversion.go
CecileRobertMichon: You are right! I had actually thought of that earlier and thought it would break the Fuzzy test because the directions set in the fuzzer are random strings. But, what I ended up doing is:

```
if restoredSecurityRule.Direction != infrav1alpha4.SecurityRuleDirectionInbound {
  restoredOutboundRules = append(restoredOutboundRules, restoredSecurityRule)
}
```

Which makes the Fuzzer happy, and, IMO, is more logically correct because the condition for "dropping" a rule in v1alpha4 --> v1alpha3 conversion is "if direction != Inbound" (we only include rules where direction == Ibound). So it makes sense to restore all rules that aren't inbound since they would all have been dropped theoretically speaking. And in the unlikely event that we need to add a new direction in v1alpha4 (for example, "both") we don't need to change the existing conversion.

shysank: +1 checking for `!= inbound` instead of `==outbound`, more future proof!

devigned (approved):
/lgtm
/appove

Looks great! Always love seeing docs updates with a PR :sunglasses:.

CecileRobertMichon:
/approve

:D

Issue #165, opened 2019-03-30
Implement Azure bastion host


justaugustus:
/kind feature

**Describe the solution you'd like**
Adding a bastion node will allow secure access to nodes, without having to rely on NAT rules on the public load balancer, laying the groundwork for non-public capz scenarios.

Use the capa bastion as a reference point: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/pkg/cloud/aws/services/ec2/bastion.go

https://docs.microsoft.com/en-us/azure/bastion/bastion-overview

Related: #104 

/priority important-soon
/milestone v1alpha

tahsinrahman:
I can work on this. How should it look like?

What i can think of now is,
- A bastion host with public ip
- A new subnet for bastion host, with security group rule that allows all traffice at port 22
- Remove ssh rule from controlplane sg
- Remove ssh nat rule from load balancer

justaugustus:
@tahsinrahman -- Appreciate the help! What you've described looks like a great first pass. Additionally, let's make sure that the bastion uses the same OS as the Cluster API machines (Ubuntu 18.04).

Tag me on the PR when you're ready for someone to review. :)

CecileRobertMichon:
Pasting @voor 's  feature request from #528 here:

```
[Azure Bastion](https://aka.ms/AboutBastion) is a neat feature that allows secure connections to the individual VMs without exposing their SSH ports to the internet.  The default configuration should add in the necessary address space, `AzureBastionSubnet`, and vnet-bastion setup to allow for this to work without any additional configuration.  Documentation should be updated to reflect this is a more secure and preferable way to connect to hosts instead of exposing them to the internet.
```

voor:
Only thing I want to add is to make sure we're not confusing bastion-the-concept with Azure Bastion-the-capability, since Azure Bastion is explicitly allowed in certain environments, whereas punching a hole for port 22 through a VM is not, the original issue was directed towards the hole-punching approach.

CecileRobertMichon:
@voor I think focusing on implementing Azure Bastion makes sense. I don't think it was available when this issue was originally opened. 

> since Azure Bastion is explicitly allowed in certain environments

You mean https://docs.microsoft.com/en-us/azure/bastion/bastion-overview#regions right?

whites11:
Are we still interested in this feature? I could work on it.

devigned:
[...]

PR #1300, opened 2021-04-12
Added support for Azure Bastion.

whites11:
/kind feature

Towards: https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/165

This PR adds support for the first of 3 bastion host implementations for CAPZ: the one using [Azure Bastion](https://azure.microsoft.com/en-us/services/azure-bastion/).

The PR adds a new field to the `AzureCluster` CR that looks like this:

```
apiVersion: infrastructure.cluster.x-k8s.io/v1alpha4
kind: AzureCluster
metadata:
  name: d3m3v
  namespace: default
spec:
  bastionSpec:
    azureBastion:
      name: "azurebastion1"
      subnet: *SubnetSpec
      publicIP:
        name: "azurebastion1-publicIP"
```

by default the `bastionSpec` field is empty and the `azureBastion` field is therefore null (that means the feature is disabled by default).
User can simply set a value of the `azureBastion` field to enable the feature (other fields are optional and defaulted).

Once the feature is enabled, removing it from a cluster is unsupported (see [slack thread](https://kubernetes.slack.com/archives/CEX9HENG7/p1617883958183000)).

It is possible to enable the azureBastion feature on an existing v1alpha4 cluster though.

Not sure where to put documentation about the feature, guidance would be appreciated.

No image version changes.

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Added support for using [Azure Bastion](https://azure.microsoft.com/en-us/services/azure-bastion/) to get console access to virtual machines in the cluster through the Azure Portal.
```


shysank:
@whites11 can you add the copyright info to any new files you created. I think the `pull-cluster-api-provider-azure-verify` test  is failing because of that.

whites11:
> @whites11 can you add the copyright info to any new files you created? I think `pull-cluster-api-provider-azure-verify` is failing because of that.

Fixed, thx

whites11:
> @whites11 can you add the copyright info to any new files you created? I think `pull-cluster-api-provider-azure-verify` is failing because of that.

Not sure how to fix the apidiff job. Any hints?

shysank:
> Not sure how to fix the apidiff job. Any hints?

It doesn't have to be fixed. They are just warnings when there are changes in api.

api/v1alpha3/azurecluster_conversion.go
```diff
@@ -58,6 +60,15 @@ func (src *AzureCluster) ConvertTo(dstRaw conversion.Hub) error { // nolint
 	dst.Spec.NetworkSpec.APIServerLB.FrontendIPsCount = restored.Spec.NetworkSpec.APIServerLB.FrontendIPsCount
 	dst.Spec.NetworkSpec.NodeOutboundLB = restored.Spec.NetworkSpec.NodeOutboundLB
 
+	if bastionSpecJson, ok := src.Annotations[bastionEnvironmentAnnotation]; ok {
```
devigned, line 63: Could this not be pull from `restored`?

api/v1alpha3/azurecluster_conversion.go
```diff
@@ -85,6 +96,19 @@ func (dst *AzureCluster) ConvertFrom(srcRaw conversion.Hub) error { // nolint
 		return err
```
devigned, line 96: Looks like there are 2 calls to utilconversion.MashalData. There should only be one.

api/v1alpha3/azurecluster_conversion.go
```diff
@@ -85,6 +96,19 @@ func (dst *AzureCluster) ConvertFrom(srcRaw conversion.Hub) error { // nolint
...
 	}
 
+	if src.Spec.BastionSpec.AzureBastion != nil {
```
devigned, line 99: I believe this should be marshaled as part of the `utilconversion.MarshalData` can and be available for restoration via `restored` above. I don't thing this block is needed.

devigned (changes requested):
I think the conversion code can be slimmed up a bit. I don't think the extra bastion annotation is needed as the entire converted structure should be available via the `utilconversion` annotation.

api/v1alpha3/azurecluster_conversion.go
whites11: Oh, thanks for opening my eyes. I struggled to make the conversion test to pass and it was so easy after all.
Should hopefully be ok now.

api/v1alpha3/azurecluster_conversion.go
```diff
@@ -58,6 +58,8 @@ func (src *AzureCluster) ConvertTo(dstRaw conversion.Hub) error { // nolint
 	dst.Spec.NetworkSpec.APIServerLB.FrontendIPsCount = restored.Spec.NetworkSpec.APIServerLB.FrontendIPsCount
 	dst.Spec.NetworkSpec.NodeOutboundLB = restored.Spec.NetworkSpec.NodeOutboundLB
```
devigned, line 60: nit: please remove the space between line 59 and 61.

devigned (changes requested):
One small nit, but overall the PR looks great!

I do have a concern about not having an e2e test to ensure the functionality doesn't regress. Would you consider adding a bastion to the existing e2e cluster deployment (perhaps, `./templates/test/ci/prow`) and a test verifying successful deployment of the bastion.

I wonder if the bastion shouldn't be part of the default template... @CecileRobertMichon wdyt?

api/v1alpha3/azurecluster_conversion.go
whites11: Blank line removed.
Regarding the e2e test, I will work on that.

CecileRobertMichon:
I promise to take a closer look tomorrow but in the meantime, would love to see some user-facing docs as part of this PR.

re: testing/flavors, I'm worried we're adding too many different flavors / test for each feature, we should start combining some. In what use case does it make the most sense for a user to turn this on? a private API server w/ no ssh maybe? or is it useful even when ssh access is possible? I wouldn't add it to the "default" flavor template if it's going to incur some extra cost for every user (especially quick start users), but it might make sense to add it to the default-prow template (or just the private cluster template if that makes more sense in terms of use case).

whites11:
During office hours earlier today we agreed enabling the azure bastion feature in the "internal networking" e2e test and adding additional checks into that test rather than having a dedicated e2e test just for this feature.

whites11:
Can anybody please help me a little bit with the e2e tests?
What do I need to test? I expected to test the fact that when the azure bastion feature is enabled in a cluster, the `Azure Bastion` cloud object gets created and becomes healthy,
Problem is that I don't see any example of this kind of tests in the code.
Can anybody please provide some guidance?

devigned:
@whites11 I don't think we are checking the Azure resources are created, but rather the functionality is working. For example, if we apply a machine deployment, we would then check if we can run workloads on those nodes. Perhaps, the same strategy could be used in the bastion. One example that comes to mind is: "Can you ssh into the bastion and then tunnel to one of the private nodes?". wdyt?

devigned:
Hmm... here's an example of setting up an Azure client and calling out to Azure Resource Manager (ARM) to get event logs. Perhaps, use a similar pattern where you create a client and request the bastion resource from ARM and verify the properties you expect are set. https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/e6edff7b16eab72bca70dc4d95b82feb407647ca/test/e2e/e2e_suite_test.go#L181-L232

I was thinking about a VM bastion scenario, not the Azure Bastion. My fault.

whites11:
I don't get why `pull-cluster-api-provider-azure-verify` fails: running `make generate` locally doesn't change files

devigned:
@whites11 perhaps, wipe out your `./hack/tools/bin` directory. Maybe it's using an older version of a bin than the CI server.

whites11:
> @whites11 perhaps, wipe out your `./hack/tools/bin` directory. Maybe it's using an older version of a bin than the CI server.

Didn't work unluckily.
Can it be related to this error I get in the output (both locally and in the CI)?

```
E0504 12:44:05.503496    1728 conversion.go:193] Rename function sigs.k8s.io/cluster-api-provider-azure/api/v1alpha3 Convert_v1alpha3_ManagedDisk_To_v1alpha4_ManagedDiskOptions -> Convert_v1alpha3_ManagedDisk_To_v1alpha4_ManagedDiskParameters to match expected conversion signature
E0504 12:44:05.503732    1728 conversion.go:193] Rename function sigs.k8s.io/cluster-api-provider-azure/api/v1alpha3 Convert_v1alpha4_ManagedDiskOptions_To_v1alpha3_ManagedDisk -> Convert_v1alpha4_ManagedDiskParameters_To_v1alpha3_ManagedDisk to match expected conversion signature
```

devigned:
> Can it be related to this error I get in the output (both locally and in the CI)?

I don't think so. It's only a warning. I'm seeing it locally too.

devigned:
@whites11 I'm seeing the issue in one of my PRs too. I'm looking into the resolution.

whites11:
> @whites11 I'm seeing the issue in one of my PRs too. I'm looking into the resolution.

OK, thanks for the heads up

CecileRobertMichon:
Did you rebase on the latest?

devigned:
Here's the diff I get in #1332 via [the verify build log](https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/1332/pull-cluster-api-provider-azure-verify/1389627379143413760/build-log.txt)

```
diff --git a/api/v1alpha3/zz_generated.conversion.go b/api/v1alpha3/zz_generated.conversion.go
index f124d9a7..ab9dfbd7 100644
--- a/api/v1alpha3/zz_generated.conversion.go
+++ b/api/v1alpha3/zz_generated.conversion.go
@@ -1299,7 +1299,7 @@ func autoConvert_v1alpha4_NetworkSpec_To_v1alpha3_NetworkSpec(in *v1alpha4.Netwo
 func autoConvert_v1alpha3_OSDisk_To_v1alpha4_OSDisk(in *OSDisk, out *v1alpha4.OSDisk, s conversion.Scope) error {
 	out.OSType = in.OSType
 	out.DiskSizeGB = in.DiskSizeGB
-	// WARNING: in.ManagedDisk requires manual conversion: inconvertible types (./api/v1alpha3.ManagedDisk vs *sigs.k8s.io/cluster-api-provider-azure/api/v1alpha4.ManagedDiskParameters)
+	// WARNING: in.ManagedDisk requires manual conversion: inconvertible types (sigs.k8s.io/cluster-api-provider-azure/api/v1alpha3.ManagedDisk vs *sigs.k8s.io/cluster-api-provider-azure/api/v1alpha4.ManagedDiskParameters)
 	out.DiffDiskSettings = (*v1alpha4.DiffDiskSettings)(unsafe.Pointer(in.DiffDiskSettings))
 	out.CachingType = in.CachingType
 	return nil
@@ -1308,7 +1308,7 @@ func autoConvert_v1alpha3_OSDisk_To_v1alpha4_OSDisk(in *OSDisk, out *v1alpha4.OS
 func autoConvert_v1alpha4_OSDisk_To_v1alpha3_OSDisk(in *v1alpha4.OSDisk, out *OSDisk, s conversion.Scope) error {
 	out.OSType = in.OSType
 	out.DiskSizeGB = in.DiskSizeGB
-	// WARNING: in.ManagedDisk requires manual conversion: inconvertible types (*sigs.k8s.io/cluster-api-provider-azure/api/v1alpha4.ManagedDiskParameters vs ./api/v1alpha3.ManagedDisk)
+	// WARNING: in.ManagedDisk requires manual conversion: inconvertible types (*sigs.k8s.io/cluster-api-provider-azure/api/v1alpha4.ManagedDiskParameters vs sigs.k8s.io/cluster-api-provider-azure/api/v1alpha3.ManagedDisk)
 	out.DiffDiskSettings = (*DiffDiskSettings)(unsafe.Pointer(in.DiffDiskSettings))
 	out.CachingType = in.CachingType
 	return nil
```

CecileRobertMichon:
Opened https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1363 to fix the conversion func name and get rid of the warning.

Are you rebased on latest @devigned ? I don't get any diff running make generate from master

devigned:
I am rebased on master. I also added the same clean up as #1363, but am still seeing the diff only in CI.

whites11:
> Did you rebase on the latest?

I did it a few hours ago, will do it again tomorrow

devigned:
I was still having an issue generating conversions when outside of the GOPATH. When inside of the GOPATH, everything works fine. 

@whites11 are you working from outside of your GOPATH? If so, can you repro the same thing from inside your GOPATH?

whites11:
> I was still having an issue generating conversions when outside of the GOPATH. When inside of the GOPATH, everything works fine.
> 
> @whites11 are you working from outside of your GOPATH? If so, can you repro the same thing from inside your GOPATH?

Witnessing the same. Running `make generate` from within GOPATH leads to some changes.
It also creates a weird folder tree within the repo:

```
$ git status
On branch azure-bastion
Your branch is up to date with 'origin/azure-bastion'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   templates/cluster-template-azure-bastion.yaml

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        github.com/

no changes added to commit (use "git add" and/or "git commit -a")
```

```
$ ls -R github.com/
github.com/:
kubernetes-sigs

github.com/kubernetes-sigs:
cluster-api-provider-azure

github.com/kubernetes-sigs/cluster-api-provider-azure:
api  exp

github.com/kubernetes-sigs/cluster-api-provider-azure/api:
v1alpha3

github.com/kubernetes-sigs/cluster-api-provider-azure/api/v1alpha3:
zz_generated.conversion.go

github.com/kubernetes-sigs/cluster-api-provider-azure/exp:
api

github.com/kubernetes-sigs/cluster-api-provider-azure/exp/api:
v1alpha3

github.com/kubernetes-sigs/cluster-api-provider-azure/exp/api/v1alpha3:
zz_generated.conversion.go
```

whites11:
Alright, e2e are finally green I guess that should be it? Thoughts anyone?

devigned (approved):
/lgtm

Congrats on adding the e2e test. Well done!

azure/services/bastionhosts/azurebastion.go
```diff
@@ -0,0 +1,225 @@
...
+	return nil
+}
+
+func (s *Service) ensureAzureBastionDeleted(ctx context.Context, azureBastionSpec azure.AzureBastionSpec) error {
+	s.Scope.V(2).Info("deleting bastion host", "bastion", azureBastionSpec.Name)
+
+	err := s.client.Delete(ctx, s.Scope.ResourceGroup(), azureBastionSpec.Name)
+	if err != nil && azure.ResourceNotFound(err) {
+		// already deleted
+		return nil
```
shysank, line 107: I don't think we can return here since we'll need to check for other resources

azure/scope/cluster.go
```diff
@@ -254,6 +254,20 @@ func (s *ClusterScope) SubnetSpecs() []azure.SubnetSpec {
 			Role:              s.NodeSubnet().Role,
 		},
 	}
+
+	if s.AzureCluster.Spec.BastionSpec.AzureBastion != nil {
+		azureBastionSubnet := s.AzureCluster.Spec.BastionSpec.AzureBastion.Subnet
```
shysank, line 259: IIUC this means azureBastionSubnet will be reconciled in `subnetsSvc` as well as in `bastionSvc`. Is this intentional?

shysank (commented):
A couple of nits and clarifications. I'm fine with opening up a followup pr for the nits as this has been around for a while, and don't want to get lgtm removed. Just want to make sure about the subnet and reconcileDelete clarifications before approving.

config/default/manager_image_patch.yaml
```diff
@@ -8,5 +8,5 @@ spec:
     spec:
       containers:
       # Change the value of image field below to your controller image URL
-      - image: gcr.io/k8s-staging-cluster-api-azure/cluster-api-azure-controller:latest
+      - image: gscapze2e.azurecr.io/cluster-api-azure-controller-amd64:20210504114244
```
CecileRobertMichon, line 11: revert this

config/default/manager_pull_policy.yaml
```diff
@@ -8,4 +8,4 @@ spec:
     spec:
       containers:
       - name: manager
-        imagePullPolicy: Always
+        imagePullPolicy: IfNotPresent
```
CecileRobertMichon, line 11: this too

CecileRobertMichon (commented):
/hold

api/v1alpha4/azurecluster_types.go
```diff
@@ -65,6 +65,10 @@ type AzureClusterSpec struct {
 	// - USGovernmentCloud: "AzureUSGovernmentCloud"
 	// +optional
 	AzureEnvironment string `json:"azureEnvironment,omitempty"`
+
+	// BastionSpec encapsulates all things related to the Bastions in the cluster.
+	// +optional
+	BastionSpec BastionSpec `json:"bastionSpec"`
```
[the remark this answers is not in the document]
whites11: yes makes sense

api/v1alpha4/azurecluster_default.go
```diff
@@ -204,6 +209,29 @@ func (c *AzureCluster) setNodeOutboundLBDefaults() {
 	}
 }
 
+func (c *AzureCluster) setBastionDefaults() {
+	if c.Spec.BastionSpec.AzureBastion != nil {
+		if c.Spec.BastionSpec.AzureBastion.Name == "" {
+			c.Spec.BastionSpec.AzureBastion.Name = generateAzureBastionName(c.ObjectMeta.Name)
+		}
+		// Ensure defaults for the Subnet settings.
+		{
```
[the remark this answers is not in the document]
whites11: default is no security group and no role.
The fields are there because they belong to the `SubnetSpec` type.

azure/services/bastionhosts/azurebastion.go
whites11: Good catch, thanks

config/default/manager_image_patch.yaml
whites11: done

config/default/manager_pull_policy.yaml
whites11: done

azure/scope/cluster.go
whites11: This is actually a very good point! Not sure how the PR ended up managing subnet and public IP that way, but I removed my handlers and let the other services reconcile the subnet and public IP.
It feels much better now, thanks @shysank

shysank: Can we leave bastion subnets and bastion public ip reconciliation to `subnteSvc` and `publicIpSvc` respectively and have `bastionSvc` only reconcile bastion host? Since we reconcile subnets and publicIps  are reconciled before bastion, you should not have to check for their existence, and it will make the code much simpler as well.  wdyt?

shysank: @whites11 Just checking to see if you need any help or is this something you feel is not needed? 
cc: @CecileRobertMichon for more 👀

nader-ziada: makes sense to reconcile bastion subnets and bastion public ip by the subnteSvc and publicIpSvc

nader-ziada:
tested the flow manually and confirmed the e2e tests works as expected, I don't see any outstanding comments 

/lgtm
/approve

nader-ziada:
@whites11 you are going to have to generate again to get the verify to pass after the update to capi PR

nader-ziada:
> @whites11 you are going to have to generate again to get the verify to pass after the update to capi PR

can you also add something to the docs about the bastion and how to use it?

whites11:
> > @whites11 you are going to have to generate again to get the verify to pass after the update to capi PR
> 
> can you also add something to the docs about the bastion and how to use it?

Sure, mind suggesting where I should put those docs?

nader-ziada:
> > > @whites11 you are going to have to generate again to get the verify to pass after the update to capi PR
> > 
> > 
> > can you also add something to the docs about the bastion and how to use it?
> 
> Sure, mind suggesting where I should put those docs?

you can either add a section in the private cluster section https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/docs/book/src/topics/api-server-endpoint.md , or add a new topic https://github.com/kubernetes-sigs/cluster-api-provider-azure/tree/master/docs/book/src/topics

whites11:
> > > > @whites11 you are going to have to generate again to get the verify to pass after the update to capi PR
> > > 
> > > 
> > > can you also add something to the docs about the bastion and how to use it?
> > 
> > 
> > Sure, mind suggesting where I should put those docs?
> 
> you can either add a section in the private cluster section https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/docs/book/src/topics/api-server-endpoint.md , or add a new topic https://github.com/kubernetes-sigs/cluster-api-provider-azure/tree/master/docs/book/src/topics

ok Added a new topic here https://github.com/whites11/cluster-api-provider-azure/blob/azure-bastion/docs/book/src/topics/ssh-access.md

nader-ziada:
thanks for adding the docs @whites11 

/lgtm
/approve

PR #1304, opened 2021-04-13
Add logs to e2e test

### LINKED ISSUE #1263: Add more log statements to E2E
#1260 Added a `Log()` and `Logf()` helper in the e2e package that prints timestamped logs. For now it's only used in the azure_lb test spec. We should add log lines to other parts of the E2E suite, like resource creation and other validations.

We should also consider adding durations in logs, for example: "service elb took 2.546s to become available".

mowangdk:
**What type of PR is this?**
/kind cleanup

Added a Log() and Logf() helper in the e2e package that prints timestamped logs

Fixes #1263 

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/test/e2e/azure_net_pol.go
+++ b/test/e2e/azure_net_pol.go
@@ -77,10 +77,12 @@ func AzureNetPolSpec(ctx context.Context, inputGetter func() AzureNetPolSpecInpu
 
 	nsDev, nsProd := "development", "production"
 	By("Creating development namespace")
+	Log("starting to create dev deployment namespace")
 	namespaceDev, err := e2e_namespace.CreateNamespaceDeleteIfExist(ctx, clientset, nsDev, map[string]string{"purpose": "development"})
 	Expect(err).NotTo(HaveOccurred())
 
 	By("Creating production namespace")
+	Log("starting to create prod deployment namespace")
 	namespaceProd, err := e2e_namespace.CreateNamespaceDeleteIfExist(ctx, clientset, nsProd, map[string]string{"purpose": "production"})
 	Expect(err).NotTo(HaveOccurred())
 
@@ -93,6 +95,7 @@ func AzureNetPolSpec(ctx context.Context, inputGetter func() AzureNetPolSpecInpu
 
 	// Front end Prod
 	frontendProdDeploymentName := fmt.Sprintf("frontend-prod-%v", randInt)
+	Log("starting to create frontend-prod deployments")
 	frontEndProd := deploymentBuilder.CreateDeployment("library/nginx:latest", frontendProdDeploymentName, namespaceProd.GetName())
 	frontEndProd.AddLabels(frontendLabels)
 	frontendProdDeployment, err := frontEndProd.Deploy(ctx, clientset)
@@ -100,6 +103,7 @@ func AzureNetPolSpec(ctx context.Context, inputGetter func() AzureNetPolSpecInpu
 
 	// Front end Dev
 	frontendDevDeploymentName := fmt.Sprintf("frontend-dev-%v", randInt+100000)
+	Log("starting to create frontend-dev deployments")
 	frontEndDev := deploymentBuilder.CreateDeployment("library/nginx:latest", frontendDevDeploymentName, namespaceDev.GetName())
 	frontEndDev.AddLabels(frontendLabels)
 	frontendDevDeployment, err := frontEndDev.Deploy(ctx, clientset)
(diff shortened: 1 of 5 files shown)


shysank (changes requested):
@mowangdk Thanks for the PR. The log statements that you've added will be really useful for debugging. What would also further help is, to know how much time creating/updating the resources took (an ilb example is cited in #1263 ). It'd be great if you can include that info also in the logs.

mowangdk:
@shysank  ptal, I'm not sure this is the best way to add execution time. thx~

shysank:
> @shysank ptal, I'm not sure this is the best way to add execution time. thx~

Can we not just calculate start time inside `WaitForServiceAvailable ` instead? I don't think we are going to lose any accuracy if that's what you were aiming for in this approach. Wondering if there is some kind of aop pattern where we can just annotate the function (go generate??) that will just calculate the execution time. @devigned any thoughts?

mowangdk:
> > @shysank ptal, I'm not sure this is the best way to add execution time. thx~
> 
> Can we not just calculate start time inside `WaitForServiceAvailable ` instead? I don't think we are going to lose any accuracy if that's what you were aiming for in this approach. Wondering if there is some kind of aop pattern where we can just annotate the function (go generate??) that will just calculate the execution time. @CecileRobertMichon @devigned any thoughts?

If there is no other updates, I will add the timing to  ```WaitForXXX``` method. @shysank

devigned (changes requested):
I agree with @shysank. I'd prefer to not overload `Log` functions with a return of time. I can see adding a timestamp by default to `Log*` funcs, but I don't like returning the time.

I'd prefer to see some `timed` func that might takes a func arg and have that do the timing. Perhaps, something like the following:
```go
func Timed(action func()) {
    //...
}
```

CecileRobertMichon:
@mowangdk can you please rebase when you get a chance?

mowangdk:
> @mowangdk can you please rebase when you get a chance?

ok~

test/e2e/azure_net_pol.go
```diff
@@ -177,27 +183,32 @@ func AzureNetPolSpec(ctx context.Context, inputGetter func() AzureNetPolSpecInpu
...
 	By("Applying a network policy to deny ingress access to app: webapp, role: backend pods in development namespace")
 	nwpolicyName, namespaceE2E, nwpolicyFileName := "backend-deny-ingress", nsDev, "backend-policy-deny-ingress.yaml"
+	Log("starting to applying a network policy to deny access to app: webapp, role: backend pods in development namespace")
 	e2e_networkpolicy.ApplyNetworkPolicy(ctx, clientset, nwpolicyName, namespaceE2E, nwpolicyFileName, PolicyDir)
 
 	By("Ensuring we no longer have ingress access from the network-policy pods to backend pods")
 	e2e_networkpolicy.EnsureConnectivityResultBetweenPods(clientset, config, nwpolicyPods, backendPods, false)
 
 	By("Cleaning up after ourselves")
+	Log("starting to cleaning up after ourselves")
```
shysank, line 193: can we be more specific about what we are cleaning up here?

shysank: same for other places as well.

test/e2e/azure_net_pol.go
```diff
@@ -177,27 +183,32 @@ func AzureNetPolSpec(ctx context.Context, inputGetter func() AzureNetPolSpecInpu
 
 	By("Applying a network policy to deny ingress access to app: webapp, role: backend pods in development namespace")
 	nwpolicyName, namespaceE2E, nwpolicyFileName := "backend-deny-ingress", nsDev, "backend-policy-deny-ingress.yaml"
+	Log("starting to applying a network policy to deny access to app: webapp, role: backend pods in development namespace")
```
shysank, line 186: I see that you're trying to use the test description for log statements. Would it be more useful if we can add details about the resource being applied. Eg:
`Logf("Applying network policy %s/%s", namespaceE2E, nwpolicyName)`

shysank: Same for other places as well.

shysank (changes requested):

shysank:
Thanks @mowangdk for being patient with the reviews!

mowangdk:
> Thanks @mowangdk for being patient with the reviews!
You're welcome~ I am willing to provide more help if there is a chance

PR #1309, opened 2021-04-14
ensure no latest Azure SDK API versions are allowed

devigned:
**What type of PR is this?**

/kind cleanup

It's very easy for a reviewer to miss Azure SDK "latest" import paths. This lint script will ensure none are committed.

Why do we not want a "latest" import path? The "latest" import paths are floating imports. This means the types in the latest import path are aliases which point to types in the newest stable Azure API version (latest --> 2020-12-01). By importing these paths, the version of the API used will change without making changes to the import path. This is advantageous for folks who would always like to stay on latest, and don't want to deal with updating import paths with each SDK Go module upgrade. However, in Cluster API Provider Azure it is beneficial to make a conscious choice which API versions we chose to depend upon to ensure compatibility across Azure clouds (public, sovereign, government, etc.).

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/lint-latest.sh
+++ b/hack/lint-latest.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+# Copyright 2021 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o nounset
+set -o pipefail
+
+root=$(dirname "${BASH_SOURCE[0]}")
+abs_root=$(cd "${root}/.." || exit 2; pwd)
+found=$(find "${abs_root}" -type f -name '*.go' -print0 | xargs -0 grep -Ei 'github.com/Azure/azure-sdk-for-go/.+/latest/.+')
+if [[ -n ${found} ]]; then
+  echo "Found usages of the 'latest' floating Azure API version. Only specific versions of the Azure APIs are allowed. Replace the following occurrences of latest with a specific date version described here: https://github.com/Azure/azure-sdk-for-go#versioning."
+  echo "${found}"
+  exit 1
+fi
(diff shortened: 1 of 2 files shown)


hack/lint-latest.sh
```diff
@@ -0,0 +1,26 @@
...
+# limitations under the License.
+
+set -o nounset
+set -o pipefail
+
+root=$(dirname "${BASH_SOURCE[0]}")
+abs_root=$(cd "${root}/.." || exit 2; pwd)
+found=$(find "${abs_root}" -type f -name '*.go' -print0 | xargs -0 grep -Ei 'github.com/Azure/azure-sdk-for-go/.+/latest/.+')
+if [[ -n ${found} ]]; then
+  echo "Found usages of the 'latest' floating Azure API version. Only specific versions of the Azure APIs are allowed. Replace the following occurrences of latest with a specific date, like 2020-12-01."
```
CecileRobertMichon, line 23: instead of stating a specific date maybe we should point to the Azure sdk repo and say to use a specific API version from the currently available APIs? 

Maybe I'm being too cautious but I could see a new contributor with no Azure SDK for Go knowledge trying to fix the error by literally replacing "latest" with "2020-12-01" and being confused when it doesn't work.

CecileRobertMichon:
Tried it locally on macOS, works great!

```
➜  cluster-api-provider-azure git:(pr/1309) ./hack/lint-latest.sh
➜  cluster-api-provider-azure git:(pr/1309) ✗ git status
On branch pr/1309
Your branch is up to date with 'devigned/latest-check'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   azure/services/virtualmachines/virtualmachines.go

no changes added to commit (use "git add" and/or "git commit -a")
➜  cluster-api-provider-azure git:(pr/1309) ✗ ./hack/lint-latest.sh
Found usages of the 'latest' floating Azure API version. Only specific versions of the Azure APIs are allowed. Replace the following occurrences of latest with a specific date, like 2020-12-01.
/Users/cerobert/go/src/sigs.k8s.io/cluster-api-provider-azure/azure/services/virtualmachines/virtualmachines.go:	"github.com/Azure/azure-sdk-for-go/profiles/latest/compute/mgmt/compute"
```

Besides my comment above on the error message, lgtm

hack/lint-latest.sh
devigned: Good idea! Changed it to point to the versioning documentation in the Azure SDK for Go. PTAL.

PR #1287, opened 2021-03-30
test: refactor ci-version templates

chewong:
Signed-off-by: Ernest Wong <chuwon@microsoft.com>

 

 
**What type of PR is this?**

**What this PR does / why we need it**:

/kind failing-test

- use `${CLUSTER_NAME}-md-0-azure-json` instead of `${CLUSTER_NAME}-control-plane-azure-json` for azure.json secret name
- centralize control-plane-related modifications into `templates/test/patches/control-plane-ci-version.yaml`
- remove MachinePool from prow-ci-version flavor. It should have MachineDeployment only
- remove `CNI_RESOURCES` environment variable from the following flavors: `prow`, `prow-ci-version`, `prow-machine-pool`, `prow-machine-pool-ci-version`, and use `configMapGenerator` instead

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
NONE
```

PR #1327, opened 2021-04-20
fix: remove calico installation step in Makefile

chewong:
Signed-off-by: Ernest Wong <chuwon@microsoft.com>

 

 
**What type of PR is this?**

Follow-up PR for https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1287. Looks like we are now installing calico twice on the workload cluster, causing the following flake in https://testgrid.k8s.io/provider-azure-master-signal#capz-azure-disk:

```log
Warning: resource serviceaccounts/calico-kube-controllers is missing the kubectl.kubernetes.io/last-applied-configuration annotation which is required by kubectl apply. kubectl apply should only be used on resources created declaratively by either kubectl create --save-config or kubectl apply. The missing annotation will be patched automatically.
serviceaccount/calico-kube-controllers configured
Error from server (AlreadyExists): error when creating "templates/addons/calico.yaml": deployments.apps "calico-kube-controllers" already exists
```

Fixes #

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -490,17 +490,6 @@ create-workload-cluster: $(ENVSUBST)
 	kubectl get secrets $(CLUSTER_NAME)-kubeconfig -o json | jq -r .data.value | base64 --decode > ./kubeconfig
 	timeout --foreground 600 bash -c "while ! kubectl --kubeconfig=./kubeconfig get nodes | grep master; do sleep 1; done"
 
-	@if [[ "${CLUSTER_TEMPLATE}" == *prow* ]]; then \
-		if [[ "${CLUSTER_TEMPLATE}" == *ipv6* ]]; then \
-			kubectl --kubeconfig=./kubeconfig apply -f templates/addons/calico-ipv6.yaml; \
-		elif [[ "${CLUSTER_TEMPLATE}" == *windows* ]]; then \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -496,8 +496,6 @@ create-workload-cluster: $(ENVSUBST)
 		elif [[ "${CLUSTER_TEMPLATE}" == *windows* ]]; then \
 			echo "windows being applied" \
 			kubectl --kubeconfig=./kubeconfig apply -f templates/addons/windows; \
-		else \
```
CecileRobertMichon, line 499: what about windows and ipv6 test templates? are those using cluster resource sets too?

chewong: Looks like they both are. Looks like I can completely remove the if statement?

CecileRobertMichon: If they're no longer needed let's remove the entire if statement.

I wonder if the conformance tests have the CRS as well though, running one below to confirm

CecileRobertMichon:
@jsturtevant this windows deployment timed out after 900.115s 😢 

/retest
/approve

PR #1360, opened 2021-05-04
Use AzureClusterIdentity when running ci e2e tests

nader-ziada:
**What type of PR is this?**
/kind feature

- use AzureClusterIdentity in ci to make sure it working in all cases

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
- Update aad-pod-identity to v1.8.0 which upgrades CRDs from apiextensions/v1beta1 to apiextensions/v1
- Update calico version used by templates and tests
```


nader-ziada:
> the `private cluster` test specifically is failing every time, looking into it

I believe the private cluster test is failing because of this issue here https://azure.github.io/aad-pod-identity/docs/troubleshooting/#token-requests-calls-fail-with-io-timeout

nader-ziada:
more info about issues with tests failing in this PR can be found in this slack thread https://kubernetes.slack.com/archives/CEX9HENG7/p1621362051002500

tldr; 
 - the nmi pod is having dns issues and can't authenticate when the management cluster is running on the Azure (not on a kind cluster) as in the case of the `private cluster` test. The dns issues happen when using `dnsPolicy=ClusterFirstWithHostNet` as recommended for any pod using `hostNetwork=true` 
 - The worker node has no issues connecting and the `nmi` pod works fine with `dnsPolicy=ClusterFirst`
 - but there seems to be an issue with either `how we configure the nmi pod`, or `the azure cluster vnet`

nader-ziada:
So the real issue here is kernel bug we get stuck on when using calico with vxlan

more details here: https://github.com/projectcalico/calico/issues/3145

nader-ziada:
I think we need to wait for the next release of calico that should include this fix here

https://github.com/projectcalico/felix/pull/2811

nader-ziada:
removed the identity ref from the private cluster template for now

nader-ziada:
@CecileRobertMichon ready for review

test/e2e/common.go
```diff
@@ -51,6 +51,8 @@ const (
 	AzureCPSubnetCidr        = "AZURE_CP_SUBNET_CIDR"
 	AzureNodeSubnetCidr      = "AZURE_NODE_SUBNET_CIDR"
 	MultiTenancyIdentityName = "MULTI_TENANCY_IDENTITY_NAME"
+	ClusterIdentityName      = "CLUSTER_IDENTITY_NAME"
+	VMSSHPort                = "VM_SSH_PORT"
```
CecileRobertMichon, line 55: are you using VMSSHPort anywhere? It was removed in another PR

nader-ziada: looks like the rebase messed up something, will clean up

nader-ziada: done

CecileRobertMichon (commented):
should we get rid of the `different SP identity` test since it's now equivalent to the others?

nader-ziada:
> should we get rid of the `different SP identity` test since it's now equivalent to the others?

yes, will remove that

fiunchinho:
I see that the release note for this PR is `None` but it's upgrading `aad-pod-identity` to `v1.8.0` which upgrades CRDs from apiextensions/v1beta1 to apiextensions/v1. Should this be included in the release note?

nader-ziada:
> I see that the release note for this PR is `None` but it's upgrading `aad-pod-identity` to `v1.8.0` which upgrades CRDs from apiextensions/v1beta1 to apiextensions/v1. Should this be included in the release note?

good point, it really should have been a different PR, but was done as part fixing this up here, will update release notes for now and if reviewers feel I should split it out in a different PR, I can do that

CecileRobertMichon (commented):
/lgtm
/assign @devigned 

@nader-ziada do you have any thoughts on how we can encourage users to use AzureClusterIdentites for their v1alpha4 clusters? Should the default template example have it? Related to #1445

nader-ziada:
> @nader-ziada do you have any thoughts on how we can encourage users to use AzureClusterIdentites for their v1alpha4 clusters? Should the default template example have it? Related to #1445

- document it in the quick start guide
- add it to the example templates?

CecileRobertMichon:
> 1) document it in the quick start guide

might not be as easy as setting env vars for quick start but if we do 2) the quickstart experience should stay the same

> 2) add it to the example templates?

Yeah I think we should consider doing this

> 3) add some deprecation warning (not sure how)

this we can do in the controller logs and/or as a k8s event whenever credentials are set through manager environment variables

nader-ziada:
I'll just try again before I debug too much 

nader-ziada:
the private cluster test is failing again because its getting stuck on creating the bastion host, 
```
I0622 21:57:02.456316       1 azurebastion.go:46] controllers/AzureCluster "msg"="creating bastion host" "azureCluster"="capz-e2e-jp3qx9" "cluster"="capz-e2e-jp3qx9" "namespace"="capz-e2e-tq69po" "bastion"="capz-e2e-jp3qx9-azure-bastion"
```
it eventually times out the private cluster is never created, 

still looking into it

CecileRobertMichon:
/lgtm

@nader-ziada can you follow up on 

> document it in the quick start guide
might not be as easy as setting env vars for quick start but if we do 2) the quickstart experience should stay the same
add it to the example templates?
Yeah I think we should consider doing this
add some deprecation warning (not sure how)
this we can do in the controller logs and/or as a k8s event whenever credentials are set through manager environment variables

nader-ziada:
> might not be as easy as setting env vars for quick start but if we do 2) the quickstart experience should stay the same



> document it in the quick start guide
> might not be as easy as setting env vars for quick start but if we do 2) the quickstart experience should stay the same
> add it to the example templates?
> Yeah I think we should consider doing this
> add some deprecation warning (not sure how)
> this we can do in the controller logs and/or as a k8s event whenever credentials are set through manager environment variables

@CecileRobertMichon it was thinking to wait until we also can do that for the private clusters so we don't have any exceptions in the quick start guide, wdyt?

CecileRobertMichon:
/approve
/hold 
feel free to unhold whenever we have confirmation that the conformance tests above are creating clusters without issue

nader-ziada:
/hold cancel

@CecileRobertMichon PTAL

all the tests are using AzureClusterIdentity except for `SelfHosted` and `PrivateCluster`

nader-ziada:
rebased on capi v0.4.0

azure/scope/identity.go
```diff
@@ -182,9 +181,9 @@ func (p *AzureCredentialsProvider) GetAuthorizer(ctx context.Context, resourceMa
 			Name:      fmt.Sprintf("%s-binding", copiedIdentity.Name),
 			Namespace: copiedIdentity.Namespace,
 			Labels: map[string]string{
-				clusterv1.ClusterLabelName:         clusterMeta.Name,
-				infrav1.ClusterLabelNamespace:      clusterMeta.Namespace,
-				clusterctl.ClusterctlMoveLabelName: "true",
+				clusterv1.ClusterLabelName:                   clusterMeta.Name,
+				infrav1.ClusterLabelNamespace:                clusterMeta.Namespace,
+				"clusterctl.cluster.x-k8s.io/move-hierarchy": "true",
```
CecileRobertMichon, line 186: why more away from using a constant here?

test/e2e/azure_test.go
```diff
@@ -145,6 +168,43 @@ var _ = Describe("Workload cluster creation", func() {
 				})
 			})
 		})
+
+		Context("Running the self-hosted spec", func() {
```
CecileRobertMichon, line 172: why not keep this in the capi_e2e suite like it was before? It's not really testing functionality that can be regressed by CAPZ PRs

nader-ziada: because for the identity secret to be moved when the cluster pivots to a management cluster running on Azure, it has to be in the same namespace as the cluster and running the test spec from capi doesn't allow me to know the namespace

azure/scope/identity.go
nader-ziada: I'll fix that

test/e2e/azure_test.go
CecileRobertMichon: right but could we define our own spec and run it as part of pull-cluster-api-provider-azure-capi-e2e ?

nader-ziada: oh okay, just move the calling it in capi_e2e?

CecileRobertMichon: yeah that's what I meant, sorry if it wasn't clear

nader-ziada: no worries, that makes better sense, will change it

CecileRobertMichon:
will this affect scripts/ci-entrypoint.sh? want to make sure we don't break https://testgrid.k8s.io/provider-azure-master-signal jobs that rely on it

nader-ziada:
> will this affect scripts/ci-entrypoint.sh? want to make sure we don't break https://testgrid.k8s.io/provider-azure-master-signal jobs that rely on it

you are right, `hack/create-dev-cluster.sh` doesn't create the secret or set the related `AzureClusterIdentity` fields, so would probably fail

nader-ziada:
@CecileRobertMichon I think the last two commits cover everything,   waiting for tests now🤞

hack/create-identity-secret.sh
```diff
@@ -0,0 +1,35 @@
...
+source "${REPO_ROOT}/hack/ensure-kubectl.sh"
+# shellcheck source=hack/parse-prow-creds.sh
+source "${REPO_ROOT}/hack/parse-prow-creds.sh"
+
+export CLUSTER_IDENTITY_SECRET_NAME="cluster-identity-secret"
+export CLUSTER_IDENTITY_NAME=${CLUSTER_IDENTITY_NAME:="cluster-identity"} 
+export AZURE_CLUSTER_IDENTITY_SECRET_NAMESPACE="default"
+export AZURE_CLUSTER_IDENTITY_CLIENT_ID="${AZURE_CLIENT_ID}"
+
+kubectl create secret generic "${CLUSTER_IDENTITY_SECRET_NAME}" --from-literal=clientSecret="${AZURE_CLIENT_SECRET}"
```
CecileRobertMichon, line 33: how do we make it obvious that users need to do this on their own? Should we document the steps in getting-started.md and the quickstart?

nader-ziada: I updated the docs in this PR #1477 which refers to this doc here: https://capz.sigs.k8s.io/topics/multitenancy.html

we would need to update the quick start but that's in capi, right?

CecileRobertMichon:
@nader-ziada did you test Tilt with these changes? I believe we also need to update the Tiltfile to create the secret on the management cluster

nader-ziada:
> @nader-ziada did you test Tilt with these changes? I believe we also need to update the Tiltfile to create the secret on the management cluster

I did not try tilt, will try it now

templates/test/ci/cluster-template-prow-multi-tenancy.yaml
```diff
@@ -418,6 +418,18 @@ data:
                             type: string
                         type: object
                       type: array
+                    serviceLoadBalancerIPs:
```
CecileRobertMichon, line 421: can we completely get rid of the multi-tenancy flavor now that all flavors are using multi-tenancy?

nader-ziada: I removed it from templates and removed the test, missed this one

templates/cluster-template.yaml
```diff
@@ -193,3 +197,17 @@ spec:
...
           name: '{{ ds.meta_data["local_hostname"] }}'
+---
+apiVersion: infrastructure.cluster.x-k8s.io/v1alpha4
+kind: AzureClusterIdentity
+metadata:
+  name: ${CLUSTER_IDENTITY_NAME}
+  namespace: default
+spec:
+  allowedNamespaces: {}
+  clientID: ${AZURE_CLUSTER_IDENTITY_CLIENT_ID}
```
CecileRobertMichon, line 208: would it make sense to keep this as `AZURE_CLIENT_ID` so existing setups (tilt, etc) can reuse the existing var?

nader-ziada: let me try

Tiltfile
```diff
@@ -169,6 +169,15 @@ def capz():
 
     k8s_yaml(blob(yaml))
 
+def create_identity_secret():
+    #create secret for identity password
+    os.putenv('CLUSTER_IDENTITY_SECRET_NAME', 'cluster-identity-secret')
```
CecileRobertMichon, line 174: will we need to add the namespace variable to tilt-settings (if so needs to be added to tilt docs), or could we just default it / set it here?

nader-ziada: its defaulting to `default` here currently, same as the script `hack/create-identity-secret.sh` 

for the e2e tests, its in the same namespace as the cluster

not sure what we need to change?

CecileRobertMichon: it looks like the templates expect `AZURE_CLUSTER_IDENTITY_SECRET_NAMESPACE` to be set, should that be set here? Same question for `CLUSTER_IDENTITY_NAME`.

Basically, I'm wondering if I pull the changes from this PR and run `make tilt-up` and then click on one of the flavors to deploy it, will everything work as it did before? or are there any additional vars that need to be defined in my tilt-settings.json file now? It would be nice if everything kept working as before.

nader-ziada: I think it should work the same, I added the setting of the namespace and identity name in the tiltfile similar to the script

CecileRobertMichon:
@nader-ziada tried out tilt locally, found two issues:
- if you run tilt up a second time it fails with `error: failed to create secret secrets "cluster-identity-secret" already exists`: we should delete the secret and recreate it every tilt up so it's idempotent (same thing we do with configmaps)
- the secret env var is getting printed in tilt logs, any way we can avoid that?

hack/create-identity-secret.sh
```diff
@@ -0,0 +1,34 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
```
CecileRobertMichon, line 19: `set +o xtrace`

we want to make sure the secret doesn't get printed

CecileRobertMichon:
found https://github.com/tilt-dev/tilt-extensions/tree/master/secret, not sure if it would be better to use that to create the secret (don't know if it logs the secret value the same way)

Tiltfile
nader-ziada: > @nader-ziada tried out tilt locally, found two issues:
> 
> * if you run tilt up a second time it fails with `error: failed to create secret secrets "cluster-identity-secret" already exists`: we should delete the secret and recreate it every tilt up so it's idempotent (same thing we do with configmaps)
> * the secret env var is getting printed in tilt logs, any way we can avoid that?

where does it delete configmaps? can't find that

CecileRobertMichon: here: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/Tiltfile#L174

nader-ziada:
> @nader-ziada tried out tilt locally, found two issues:
> 
> * if you run tilt up a second time it fails with `error: failed to create secret secrets "cluster-identity-secret" already exists`: we should delete the secret and recreate it every tilt up so it's idempotent (same thing we do with configmaps)
> * the secret env var is getting printed in tilt logs, any way we can avoid that?

- added the delete statement similar to the configmaps
- added the quiet=true option to prevent printing

Tiltfile
```diff
@@ -169,6 +169,20 @@ def capz():
...
+def create_identity_secret():
+    #create secret for identity password
+    local("kubectl delete secret cluster-identity-secret --ignore-not-found=true")
+
+    os.putenv('AZURE_CLUSTER_IDENTITY_SECRET_NAME', 'cluster-identity-secret')
+    os.putenv('AZURE_CLUSTER_IDENTITY_SECRET_NAMESPACE', 'default')
+    os.putenv('CLUSTER_IDENTITY_NAME', 'cluster-identity')
+
+    substitutions = settings.get("kustomize_substitutions", {})
+    os.environ.update(substitutions)
```
CecileRobertMichon, line 181: what is this update doing?

hack/create-identity-secret.sh
```diff
@@ -0,0 +1,34 @@
...
+source "${REPO_ROOT}/hack/ensure-kubectl.sh"
+# shellcheck source=hack/parse-prow-creds.sh
+source "${REPO_ROOT}/hack/parse-prow-creds.sh"
+
+export AZURE_CLUSTER_IDENTITY_SECRET_NAME="cluster-identity-secret"
+export CLUSTER_IDENTITY_NAME=${CLUSTER_IDENTITY_NAME:="cluster-identity"} 
+export AZURE_CLUSTER_IDENTITY_SECRET_NAMESPACE="default"
+
+kubectl create secret generic "${AZURE_CLUSTER_IDENTITY_SECRET_NAME}" --from-literal=clientSecret="${AZURE_CLIENT_SECRET}"
+kubectl label secret "${AZURE_CLUSTER_IDENTITY_SECRET_NAME}" "clusterctl.cluster.x-k8s.io/move-hierarchy=true" --overwrite=true
```
devigned, line 34: New line

hack/gen-flavors.sh
```diff
@@ -17,15 +17,15 @@
 set -o errexit
 set -o nounset
 set -o pipefail
-#set -o verbose
+# set -o verbose
```
devigned, line 20: delete?

templates/flavors/aks/kustomization.yaml
```diff
@@ -3,3 +3,6 @@ kind: Kustomization
 namespace: default
 resources:
 - cluster-template.yaml
+- ../../azure-cluster-identity
+patchesStrategicMerge:
+- ../../azure-cluster-identity/managedazurecluster-identity-ref.yaml
```
devigned, line 8: newline

templates/flavors/system-assigned-identity/kustomization.yaml
```diff
@@ -1,5 +1,6 @@
 namespace: default
 resources:
-  - ../default
+  - ../base
+  - ../default/machine-deployment.yaml
 patchesStrategicMerge:
   - patches/system-assigned-identity.yaml
```
devigned, line 6: newline

templates/flavors/user-assigned-identity/kustomization.yaml
```diff
@@ -1,5 +1,6 @@
 namespace: default
 resources:
-  - ../default
+  - ../base
+  - ../default/machine-deployment.yaml
 patchesStrategicMerge:
   - patches/user-assigned-identity.yaml
```
devigned, line 6: newline

templates/flavors/windows/kustomization.yaml
```diff
@@ -3,4 +3,8 @@ resources:
   - ../base-windows
   - machine-deployment.yaml
   - machine-deployment-windows.yaml
+  - ../../azure-cluster-identity
+
+patchesStrategicMerge:
+  - ../../azure-cluster-identity/azurecluster-identity-ref.yaml
```
devigned, line 10: weird newline

templates/test/ci/prow-ci-version/kustomization.yaml
```diff
@@ -7,3 +7,12 @@ patchesStrategicMerge:
...
   - patches/machine-deployment-ci-version.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: default
+resources:
+  - ../prow
+patchesStrategicMerge:
+  - ../patches/control-plane-ci-version.yaml
+  - ../patches/controller-manager.yaml
+  - patches/machine-deployment-ci-version.yaml
```
devigned, line 18: newline

templates/test/ci/prow-identity-from-env/mhc.yaml
```diff
@@ -0,0 +1,15 @@
...
+spec:
+  clusterName: "${CLUSTER_NAME}"
+  maxUnhealthy: 100%
+  selector:
+    matchLabels:
+      nodepool: "pool1"
+  unhealthyConditions:
+    - type: E2ENodeUnhealthy
+      status: "True"
+      timeout: 30s
```
devigned, line 15: newline

devigned (changes requested):
All nits.

CecileRobertMichon:
Do we want to keep all 7 commits in the final merge or are some of these squashable?

nader-ziada:
> Do we want to keep all 7 commits in the final merge or are some of these squashable?

I think they are for different parts, but I can squash if you prefer

CecileRobertMichon:
> I think they are for different parts, but I can squash if you prefer

nope, all good, just checking they are distinct

nader-ziada:
applied @devigned comments about newlines

devigned (approved):
/lgtm

nader-ziada:
thanks for all the help with debugging this PR @CecileRobertMichon @devigned @mboersma

devigned:
@nader-ziada great work on this one! This one was a tough task. Thank you for taking this on.

PR #1332, opened 2021-04-22
add Azure machine pool rolling upgrades with MaxSurge, MaxUnavailable and DeletePolicy

devigned:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
This PR introduces `AzureMachinePoolMachine` resource to replace `AzureMachinePool.Status.Instances` in accordance to [Azure Machine Pool Machines Proposal](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/adaf9c6d4b03bbe42174ad2dd1a7c0fba66a56dd/docs/proposals/20210222-azure-machinepool-machine.md).

This also introduces the machine pool rolling upgrade strategy which provides, MaxSurge, MaxUnavailable, and DeletePolicy modeled after machine deployments.

Functional changes:
- Ability to delete individual `AzureMachinePoolMachines`
- MaxSurge to surge machine pools upon upgrade of the machine pool template
- MaxUnavailable to ensure a minimum number of machines are available at any time
- DeletePolicy to enable preference on which nodes to delete, Oldest, Newest, Random
- Fully async reconciliation of `AzureMachinePool` resources

Non-goals of the PR (follow up PRs):
- implement cordon and drain per Azure machine pool machine
- implement instance update rather than delete / re-create

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
related #1105
fixes #1361

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
add Azure machine pool rolling upgrades with MaxSurge, MaxUnavailable and DeletePolicy. 
```

PR #1321, opened 2021-04-16
Make OSDisk Storage Type optional

CecileRobertMichon:
**What type of PR is this?**

**What this PR does / why we need it**: Making OS disk ManagedDisk.StorageAccountType optional as it is not actually a required field for Azure APIs.

VM sizes without "s" support both Standard HDD and Standard SSD, default is Standard HDD. VM sizes with an "s" support Premium SSD, Standard HDD and Standard SSD, defaults to Premium SSD. Some select VM sizes support Ultra SSD for data disks.

Also fixed the ephemeral OS disk flavor to use ReadOnly caching and added validation for caching types.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes partially #618 

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Make OSDisk Storage Type optional
```

PR #1368, opened 2021-05-05
 do not conversion-gen v1alpha3.OSDisk due to warning message non-determinism

devigned:
**What type of PR is this?**

/kind bug

When conversion-gen runs outside of the GOPATH, it will generate some resources with `./api/v1alpha3.TypeName` rather than the full import path of the resource `sigs.k8s.io/cluster-api-provider-azure/api/v1alpha3.TypeName`. This seems to work fine when running conversion-gen in the GOPATH.

#1300 and #1332 are running into this issue. It seems to have started with #1321, ~~but I don't see any evidence that the issue should be caused by anything in that change set.~~

~~I do not see a clear cause for why this is happening, and after a messing around with each of the flags for conversion-gen, I've thrown up my hands and just used sed to manipulate the conversion-gen output. I'm not super proud, but it will fix the problem until we find a better solution.~~

Due to OSDisk and DataDisk needing `*ManageDisk` and `ManagedDisk` conversions which would require the same conversion func name with two different signatures, conversion-gen warns that manual conversion is needed, which #1321 properly handled. This PR tells conversion-gen not to generate conversions for OSDisk and handles the conversion in the custom converter func for OSDisk, which eliminates the warning message with the relative path.

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
manually convert `OSDisk.ManagedDisk` to eliminate a non-deterministic warning from conversion-gen due to generating code outside of the GOPATH.
```

### DIFF
--- a/api/v1alpha3/types.go
+++ b/api/v1alpha3/types.go
@@ -358,6 +358,11 @@ const (
 )
 
 // OSDisk defines the operating system disk for a VM.
+//
+// WARNING: this requires any updates to ManagedDisk to be manually converted. This is due to the odd issue with
+// conversion-gen where the warning message generated uses a relative directory import rather than the fully
+// qualified import when generating outside of the GOPATH.
+// +k8s:conversion-gen=false
 type OSDisk struct {
 	OSType           string            `json:"osType"`
 	DiskSizeGB       int32             `json:"diskSizeGB"`
--- a/api/v1alpha3/azuremachine_conversion.go
+++ b/api/v1alpha3/azuremachine_conversion.go
@@ -18,7 +18,7 @@ package v1alpha3
 
 import (
 	apiconversion "k8s.io/apimachinery/pkg/conversion"
-	v1alpha4 "sigs.k8s.io/cluster-api-provider-azure/api/v1alpha4"
+	"sigs.k8s.io/cluster-api-provider-azure/api/v1alpha4"
 	utilconversion "sigs.k8s.io/cluster-api/util/conversion"
 	"sigs.k8s.io/controller-runtime/pkg/conversion"
 )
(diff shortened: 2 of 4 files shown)


CecileRobertMichon:
/hold

we are looking into fixing this another way

api/v1alpha3/types.go
```diff
@@ -358,6 +358,11 @@ const (
 )
 
 // OSDisk defines the operating system disk for a VM.
+//
+// WARNING: this requires any updates to ManagedDisk to be manually converted. This is due to the odd issue with
```
CecileRobertMichon, line 362: this warning should be in v1alpha4 types

devigned: I thought it should not be in the hub, but rather each spoke.

CecileRobertMichon: It's more likely that someone changes a v1alpha4 type which would require them to update conversion since that's the API version we're currently working on. Having the comment in v1alpha3 only means they risk not seeing it if they change api/v1alpha4/types.go

devigned: added

api/v1alpha3/types.go
```diff
@@ -358,6 +358,11 @@ const (
...
+// conversion-gen where the warning message generated uses a relative directory import rather than the fully
+// qualified import when generating outside of the GOPATH.
+// +k8s:conversion-gen=false
```
CecileRobertMichon, line 365: do we need `// +k8s:conversion-gen=false` so it doesn't generate `autoConvert_v1alpha3_OSDisk_To_v1alpha4_OSDisk` ?

devigned: Yes, that is precisely what `+k8s:conversion-gen=false` does.

PR #1369, opened 2021-05-06
POC: Parallel reconciliation of azure machine services

### LINKED ISSUE #1181: Parallel reconciliation of azure resources
/kind feature
/kind explore

We reconcile serially a bunch of resources in azurecluster and azuremachine reconcilers. Some of these resources have no dependency on each other, and can be reconciled simultaneously. Eg. After a vm is reconciled, vmextensions, tags and roleassignments can be reconciled simultaneously, and potentially save some time.

Reconcile mutually exclusive services in parallel using goroutines and channels.

This is more of an exploration than an actual requirement.

**Environment:**

shysank:
**What type of PR is this?**
/kind cleanup

This is a poc to reconcile azure machine service that are mutually exclusive to reconcile in parallel. If the concept is approved, we could extend this to other reconcilers. 

Fixes  #1181

This pr is not meant to be merged. If the concept is approved, There should be a follow up pr with proper implementation along with tests.

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note

```

### DIFF
--- a/controllers/azuremachine_reconciler.go
+++ b/controllers/azuremachine_reconciler.go
@@ -78,36 +78,26 @@ func (s *azureMachineService) Reconcile(ctx context.Context) error {
 	ctx, span := tele.Tracer().Start(ctx, "controllers.azureMachineService.Reconcile")
 	defer span.End()
 
-	if err := s.publicIPsSvc.Reconcile(ctx); err != nil {
-		return errors.Wrap(err, "failed to create public IP")
+	ar := newAsyncReconciler()
+
+	ar.submit(ctx, s.publicIPsSvc)
+	ar.submit(ctx, s.inboundNatRulesSvc)
+	ar.submit(ctx, s.networkInterfacesSvc)
+	ar.submit(ctx, s.availabilitySetsSvc)
+	if err := ar.wait(); err != nil {
+		return err
 	}
 
-	if err := s.inboundNatRulesSvc.Reconcile(ctx); err != nil {
-		return errors.Wrap(err, "failed to create inbound NAT rule")
+	ar.submit(ctx, s.virtualMachinesSvc)
+	if err := ar.wait(); err != nil {
+		return err
 	}
 
-	if err := s.networkInterfacesSvc.Reconcile(ctx); err != nil {
-		return errors.Wrap(err, "failed to create network interface")
-	}
-
-	if err := s.availabilitySetsSvc.Reconcile(ctx); err != nil {
-		return errors.Wrap(err, "failed to create availability set")
-	}
-
-	if err := s.virtualMachinesSvc.Reconcile(ctx); err != nil {
-		return errors.Wrap(err, "failed to create virtual machine")
-	}
-
-	if err := s.roleAssignmentsSvc.Reconcile(ctx); err != nil {
-		return errors.Wrap(err, "unable to create role assignment")
-	}
-
-	if err := s.vmExtensionsSvc.Reconcile(ctx); err != nil {
-		return errors.Wrap(err, "unable to create vm extension")
-	}
-
-	if err := s.tagsSvc.Reconcile(ctx); err != nil {
-		return errors.Wrap(err, "unable to update tags")
+	ar.submit(ctx, s.roleAssignmentsSvc)
+	ar.submit(ctx, s.vmExtensionsSvc)
+	ar.submit(ctx, s.tagsSvc)
+	if err := ar.wait(); err != nil {
+		return err
 	}
 
 	return nil
--- a/controllers/async_reconciler.go
+++ b/controllers/async_reconciler.go
@@ -1,31 +1,72 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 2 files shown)


controllers/azuremachine_reconciler.go
```diff
@@ -78,36 +78,26 @@ func (s *azureMachineService) Reconcile(ctx context.Context) error {
...
-		return errors.Wrap(err, "failed to create public IP")
+	ar := newAsyncReconciler()
+
+	ar.submit(ctx, s.publicIPsSvc)
+	ar.submit(ctx, s.inboundNatRulesSvc)
+	ar.submit(ctx, s.networkInterfacesSvc)
+	ar.submit(ctx, s.availabilitySetsSvc)
+	if err := ar.wait(); err != nil {
+		return err
 	}
```
nader-ziada, line 90: what about dependencies between these resources

shysank: The resources that are independent are submitted to the asyncReconciler which will be processed in parallel, and whenever there's a dependency, we wait for the tasks to finish before proceeding.

devigned (commented):
I see where this is going and it's not exactly what was meant by async reconciliation. Let me try to explain.

When the Azure SDK for Go does a PUT, PATCH and sometimes DELETE to Azure Resource Manager, a long running operation is started. The Azure SDK makes the PUT request, it then starts polling the resource until the Azure SDK determines the resource is in a "terminal" state (succeeded or failed) only then does the SDK return to the client. This is what we refer to when way say a synchronous request.

What you are doing this PR is a parallel set of synchronous requests.

I think what was intended by the async reconciliation of machine resources is more similar to what happens in the [AMPM pull request in ScaleSets](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/b30b86fd1113eada887eef8f5d73beb6adc62cc6/azure/services/scalesets/scalesets.go#L75-L144) where the Azure SDK is used to make the PUT, a future for the long running operation is captured, stored on the status of the reconciling resource, then used in subsequent reconciles to asynchronously reconcile the resource.

wdyt?

shysank:
I think we are trying optimize for two different reasons here: 1) time to do a full reconciliation, 2) don't block resources if reconciliation take long. 

#1181 was more aimed at solving (1), which might perhaps improve the overall machine creation time, and is more useful where the resource creation does not take much time. OTOH [#1067](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1067) solves (2), and is more suitable for resources that take a long time to create (vm, vmss etc.) 

Maybe we can use a combination of both? wdyt?

devigned:
> Maybe we can use a combination of both? wdyt?

Definitely could do both. The trick is tracking the state of multiple in-flight operations. 

Also, some reconciles are actually gating on a previous reconcile. Sending bad requests to Azure is a good way of inflating the number of API calls and possibly cause a busy signal from the Azure APIs.

devigned:
This might be a useful library to define the items that are serial and those that are parallel: https://github.com/johnsiilver/orchestration/tree/main/linear.

shysank:
> This might be a useful library to define the items that are serial and those that are parallel: https://github.com/johnsiilver/orchestration/tree/main/linear.

Thanks David! It looks very much similar to what we are trying to do here. I'll hack around and see how it goes.

nader-ziada:
@shysank @devigned are we going to close this PR? or use it?

devigned:
> @shysank @devigned are we going to close this PR? or use it?

Not in the current form. We should probably have a discussion during next office hours to get some points of view.

shysank:
@nader-ziada Agree with @devigned the pr is still very much incomplete. I was going to try out the library suggested but didn't have time this week. I'll try to do it sometime next week (still won't be mergeable, just to get feedback) and then we can close it.

alexeldeib:
@devigned fwiw I'm not totally sure about this: "parallelizing these activities isn't a huge gain". If you fire off one async request and have to requeue multiple times for the future to complete, it feels like a waste to _**not**_ start multiple async requests at once. I just skimmed the AMP PR and it looks like you'll end up looping multiple times before the future is actually done: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/b30b86fd1113eada887eef8f5d73beb6adc62cc6/azure/services/scalesets/scalesets.go#L141-L143

looking at the code, roleAssignments + vmExtensions seems like a natural one where the parallel initialization of the requests would help, and then checking the futures in e.g. a waitgroup or DAG or something.

devigned:
> fwiw I'm not totally sure about this: "parallelizing these activities isn't a huge gain". If you fire off one async request and have to requeue multiple times for the future to complete, it feels like a waste to not start multiple async requests at once. I just skimmed the AMP PR and it looks like you'll end up looping multiple times before the future is actually done. Looking at the code, roleAssignments + vmExtensions seems like a natural one where the parallel initialization of the requests would help, and then checking the futures in e.g. a waitgroup or DAG or something.

@alexeldeib I agree with your assessment. In the case where resources can be created concurrently, it would be an improvement. Perhaps, I was not as nuanced as I should have been. If you check out the office hours from yesterday, https://youtu.be/IDCDPuEMkR8?t=497, you'll hear a more nuanced discussion. 

The tl;dr; in machine provisioning, there are really 2, maybe 3, groups of resources for worker nodes. The nic, the VM, then the ext and role assignment which need each group to be provisioned serially, but can be concurrent inside of the group. The yield for this set of resources minimal since, as you pointed out, only the ext and the role assignment can be done concurrently. The biggest win for user experience would be async updates as the VM provisions. For the sake of time and complexity, we should proceed by first making the VM create async, then evaluate to see if we need further performance improvements.

nader-ziada:
@shysank does this need to be closed?

shysank:
yes, we can close this in favor of https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1610.

PR #1373, opened 2021-05-12
Expose cloud provider rate limit config

### LINKED ISSUE #815: expose cloud provider rate limiting configuration
/kind feature

The Azure cloud provider exposes a wide number of configuration fields: https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/docs/cloud-provider-config.md#cloud-provider-config

Most of these can be derived from infrastructure specs in CAPZ. The significant outliers all are around the rate limiting configuration for the cloud provider. We should surface these to users in a reasonable way without requiring them to set the entire file.

One approach would be to allow the subset of the CloudProviderConfig struct relevant to rate limiting to be defined in a config map. We would merge any user provided values there over our own defaults when we build the configuration.

Follow up to #802

shysank:
**What type of PR is this?**
/kind feature

This pr allows for customization of cloud provider rate limit config through `AzureCluster` spec. 

Fixes #815 

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Expose cloud provider rate limit config
```


api/v1alpha4/azurecluster_types.go
```diff
@@ -65,6 +65,13 @@ type AzureClusterSpec struct {
 	// - USGovernmentCloud: "AzureUSGovernmentCloud"
 	// +optional
 	AzureEnvironment string `json:"azureEnvironment,omitempty"`
+
+	// CloudProviderConfigOverrides is an optional set of configuration values that can be overridden in azure cloud provider config.
+	// This is only a subset of options that are available in azure cloud provider config.
+	// Some values for the cloud provider config are inferred from other parts of cluster api provider azure spec, and may not be available for overrides.
+	// See: https://kubernetes-sigs.github.io/cloud-provider-azure/install/configs
+	// +optional
+	CloudProviderConfigOverrides CloudProviderConfigOverrides `json:"cloudProviderConfigOverrides,omitempty"`
```
nader-ziada, line 74: should probably be a pointer since its optional

api/v1alpha4/types.go
```diff
@@ -507,3 +508,83 @@ type AddressRecord struct {
...
+	AvailabilitySetRateLimit = "availabilitySetRateLimit"
+)
+
+// allowedToOverrideRateLimits is a collection of rate limit spec names that are allowed to be overridden.
+var allowedToOverrideRateLimits = map[string]bool{
+	DefaultRateLimit: true, RouteRateLimit: true, SubnetsRateLimit: true, InterfaceRateLimit: true, RouteTableRateLimit: true,
+	LoadBalancerRateLimit: true, PublicIPAddressRateLimit: true, SecurityGroupRateLimit: true, VirtualMachineRateLimit: true,
+	StorageAccountRateLimit: true, DiskRateLimit: true, SnapshotRateLimit: true, VirtualMachineScaleSetRateLimit: true,
+	VirtualMachineSizesRateLimit: true, AvailabilitySetRateLimit: true,
+}
```
nader-ziada, line 576: would it be easier to use kubebuilder enum validation (similar to https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/api/v1alpha4/types.go#L324) 

makes it a littler easier to read

docs/book/src/topics/cloud-provider-config.md
```diff
@@ -6,6 +6,50 @@ CAPZ automatically generates this file based on user-provided values in AzureMac
 
 For AzureMachineTemplate and standalone AzureMachines, the generated secret will have the name "${RESOURCE}-azure-json", where "${RESOURCE}" is the name of either the AzureMachineTemplate or AzureMachine. The secret will have one data field, `azure.json`, with the raw content for that file. When the secret `${RESOURCE}-azure-json` already exists in the same namespace as an AzureCluster and does not have the label `"${CLUSTER_NAME}": "owned"`, CAPZ will not generate the default described above. Instead it will directly use whatever the user provides in that secret.
 
+### Overriding Cloud Provider Config
+
+While many of the cloud provider config values are inferred from the capz infrastructure spec, there are other configuration parameters that cannot be inferred, and hence default to the values set by the azure cloud provider. In order to provider custom values to such configuration options through capz, you must use the `spec.cloudProviderConfigOverrides` in `AzureCluster`. The following example overrides the load balancer rate limit configuration:
```
CecileRobertMichon, line 11: same comment here about clarifying order of precedence, if a custom azure-json secret is created these values will be ignored

docs/book/src/topics/cloud-provider-config.md
```diff
@@ -6,6 +6,50 @@ CAPZ automatically generates this file based on user-provided values in AzureMac
...
+          CloudProviderRateLimitBucketWrite: 2,
+          cloudProviderRateLimitQPS: 0,
+          CloudProviderRateLimitQPSWrite: 0
+```
+
+<aside class="note warning">
+
+<h1> Warning </h1>
+
+Presently, only rate limit configuration is supported for overrides, and this works only on clusters running kuberenetes versions above `v1.18.0`.
```
CecileRobertMichon, line 47: Kubernetes

api/v1alpha4/types.go
```diff
@@ -507,3 +508,63 @@ type AddressRecord struct {
 	Hostname string
 	IP       string
 }
+
+// CloudProviderConfigOverrides represents the fields that can be overridden in azure cloud provider config.
+type CloudProviderConfigOverrides struct {
+	RateLimits []RateLimitSpec `json:"rateLimits,omitempty"`
```
CecileRobertMichon, line 514: we might also want to add cloudProviderBackoff options here, specifically:


Name | Description | Remark
-- | -- | --
cloudProviderBackoff | Enable exponential backoff to manage resource request retries | Boolean value, default to false
cloudProviderBackoffRetries | Backoff retry limit | Integer value, valid if&nbsp;cloudProviderBackoff&nbsp;is true
cloudProviderBackoffExponent | Backoff exponent | Float value, valid if&nbsp;cloudProviderBackoff&nbsp;is true
cloudProviderBackoffDuration | Backoff duration | Integer value, valid if&nbsp;cloudProviderBackoff&nbsp;is true
cloudProviderBackoffJitter | Backoff jitter | Float value, valid if&nbsp;cloudProviderBackoff&nbsp;is true

CecileRobertMichon: this can also be done in a separate PR

shysank: Created [issue](https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1382) for tracking.

CecileRobertMichon (commented):
/cc @alexeldeib

CecileRobertMichon (commented):
Just one other question: what happens if you edit a value after the cluster is created? I assume the actual cloud provider config wouldn't get applied since it would require rolling the Machines to edit the file. Should we make these fields immutable and enforce it in webhooks until we can actually support updating an existing config?

shysank:
> Just one other question: what happens if you edit a value after the cluster is created? I assume the actual cloud provider config wouldn't get applied since it would require rolling the Machines to edit the file. Should we make these fields immutable and enforce it in webhooks until we can actually support updating an existing config?

I'm inclined to making it immutable. Just wondering if there will be a scenario where the user would want to change the limits without having to delete the cluster (roll out new machines and delete old ones) or perhaps as part of an upgrade?

CecileRobertMichon:
> Just wondering if there will be a scenario where the user would want to change the limits without having to delete the cluster (roll out new machines and delete old ones) or perhaps as part of an upgrade?

Ideally we do want it to be mutable, it's a valid use case to want to update those values after cluster is created (for example, if scaling up to many nodes). I'm just saying that in the current state of things it's not truly mutable since the update will just be ignored. So we should reflect reality in the webhook and disable updates until it's truly mutable. If you want to look into making it mutable as part of this PR that's fine, but I suspect it won't be a quick fix and updating the webhooks might be an easier first step.

PR #1383, opened 2021-05-19
Collect boot logs even when bootstrap fails

CecileRobertMichon:
**What type of PR is this?**

1. When the first control plane fails, the result cluster is nil which means we don't collect any logs. This adds an optional attempt to fetch the Cluster using the boostrapClusterProxy so we can proceed and grab logs on that control plane.
2. When a machine fails to bootstrap, and ssh fail, we're not collecting boot logs because the providerID in the Machine spec has not yet been set (eg https://gcsweb.k8s.io/gcs/kubernetes-jenkins/logs/periodic-cluster-api-provider-azure-e2e-full/1394441682971267072/artifacts/clusters/capz-e2e-4t733z/machines/capz-e2e-4t733z-control-plane-wnkbt/). Fix is to use the AzureMachine spec to get the provider ID of the VM. 

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/azure_logcollector.go
+++ b/test/e2e/azure_logcollector.go
@@ -45,11 +45,16 @@ type AzureLogCollector struct{}
 func (k AzureLogCollector) CollectMachineLog(ctx context.Context, managementClusterClient client.Client, m *clusterv1.Machine, outputPath string) error {
 	var errors []error
 
-	if err := collectLogsFromNode(ctx, managementClusterClient, m, outputPath); err != nil {
+	am, err := getAzureMachine(ctx, managementClusterClient, m)
+	if err != nil {
+		return err
+	}
+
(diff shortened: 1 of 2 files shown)


CecileRobertMichon:
This seems to be working, boot logs were collected for the first control plane machine with failed cloud-init --> https://gcsweb.k8s.io/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/1383/pull-cluster-api-provider-azure-e2e/1395136518770659328/artifacts/clusters/capz-e2e-gdyh7y/machines/capz-e2e-gdyh7y-control-plane-7kbrd/

Will revert the change to fail cloud init on purpose and remove WIP

PR #1390, opened 2021-05-25
move allowed namespaces check to fail early

### LINKED ISSUE #1381: Refactor IsClusterNamespaceAllowed check in multi tenancy to fail fast
/kind cleanup

For multi tenancy, we validate the identity's namespace with the cluster's namespace [here](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/430fb0fbeb1a2d3a9bb20e30d6edfb8996d30827/controllers/azurecluster_controller.go#L202). But, before this we already create a cluster scope which in turn creates a [credential provider](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/430fb0fbeb1a2d3a9bb20e30d6edfb8996d30827/azure/scope/cluster.go#L74) thereby creating an [AzureIdentityBinding](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/430fb0fbeb1a2d3a9bb20e30d6edfb8996d30827/azure/scope/identity.go#L82).

Ideally we don't want to create the binding if namespace is not allowed. So, fail fast before creating the cluster scope if the namespace validation fails.

**Environment:**

nader-ziada:
**What type of PR is this?**
/kind cleanup

- move the allowedNamespaces identity check to cluster scope to fail early if the namespace is not allowed 
- 
Fixes #1381 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/controllers/azurecluster_controller.go
+++ b/controllers/azurecluster_controller.go
@@ -151,6 +151,27 @@ func (r *AzureClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		return ctrl.Result{}, nil
 	}
 
+	if azureCluster.Spec.IdentityRef != nil {
+		identity, err := GetClusterIdentityFromRef(ctx, r.Client, azureCluster.Namespace, azureCluster.Spec.IdentityRef)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		if !scope.IsClusterNamespaceAllowed(ctx, r.Client, identity.Spec.AllowedNamespaces, azureCluster.Namespace) {
+			conditions.MarkFalse(azureCluster, infrav1.NetworkInfrastructureReadyCondition, infrav1.NamespaceNotAllowedByIdentity, clusterv1.ConditionSeverityError, "")
+			return reconcile.Result{}, errors.New("AzureClusterIdentity list of allowed namespaces doesn't include current cluster namespace")
+		}
+		if identity.Namespace == azureCluster.Namespace {
+			patchhelper, err := patch.NewHelper(identity, r.Client)
+			if err != nil {
+				return reconcile.Result{}, errors.Wrap(err, "failed to init patch helper")
+			}
+			identity.ObjectMeta.OwnerReferences = azureCluster.GetOwnerReferences()
+			if err := patchhelper.Patch(ctx, identity); err != nil {
+				return reconcile.Result{}, err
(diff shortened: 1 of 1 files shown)


azure/scope/identity.go
```diff
@@ -66,6 +67,11 @@ func NewAzureCredentialsProvider(ctx context.Context, kubeClient client.Client,
 		return nil, errors.New("AzureClusterIdentity is not of type Service Principal")
 	}
 
+	if !IsClusterNamespaceAllowed(ctx, kubeClient, identity.Spec.AllowedNamespaces, azureCluster.Namespace) {
+		conditions.MarkFalse(azureCluster, infrav1.NetworkInfrastructureReadyCondition, infrav1.NamespaceNotAllowedByIdentity, clusterv1.ConditionSeverityError, "")
```
shysank, line 71: I feel this will create an unexpected side effect that when instantiating a new credentials provider. I don't have any particularly good solution to it other than moving it to the controller. wdyt?

nader-ziada: what is the side effect? the condition

I was going to put in the controller first; but was trying to avoid doing an extra get call, 

I'll take another look at it

shysank: Yeah, the conditions. One thing I did for handling [vm deletion error](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/8ebf049dd162a74a579ac7d25db726f8238dea86/controllers/azuremachine_controller.go#L275) was to return a new error type and used that to set the appropriate condition.

nader-ziada: are you concerned that the `condition` is set in the scope package? or that we are setting a condition?

cause the I did it that way after seeing the machine scope set conditions as well

sorry for all the questions, trying to understand

shysank: Sorry about that bikeshedding. My thinking is that `NewAzureCredentialsProvider` should be just about creating a new CredentialProvider instance with the inputs passed. Since conditions kind of represent the state of an AzureCluster, I just thought it might not be the right place to set it. This is just an opinionated view. I'm fine with it remaining as is if you think otherwise.

nader-ziada: moved it to controller before creating scope, PTAL

PR #1401, opened 2021-05-31
Save subnet details back to scope

fiunchinho:
**What type of PR is this?**

/kind bug

When `SubnetSpec` was changed to not be a pointer in #1144 we were not saving the Subnet details in the scope.

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Release note**:

```release-note
Save subnet details back to scope when reconciling already existing subnets
```

### DIFF
--- a/azure/scope/managedcontrolplane.go
+++ b/azure/scope/managedcontrolplane.go
@@ -198,6 +198,12 @@ func (s *ManagedControlPlaneScope) NodeSubnet() infrav1.SubnetSpec {
 	}
 }
 
+// SetSubnet sets the passed subnet spec into the scope.
+// This is not used when using a managed control plane.
+func (s *ManagedControlPlaneScope) SetSubnet(subnetSpec infrav1.SubnetSpec) {
+	// no-op
+}
+
 // ControlPlaneSubnet returns the cluster control plane subnet.
 func (s *ManagedControlPlaneScope) ControlPlaneSubnet() infrav1.SubnetSpec {
 	return infrav1.SubnetSpec{}
--- a/azure/services/subnets/subnets.go
+++ b/azure/services/subnets/subnets.go
@@ -72,10 +72,10 @@ func (s *Service) Reconcile(ctx context.Context) error {
 				continue
 			}
 
-			subnet.Role = subnetSpec.Role
-			subnet.Name = existingSubnet.Name
-			subnet.CIDRBlocks = existingSubnet.CIDRBlocks
 			subnet.ID = existingSubnet.ID
+			subnet.CIDRBlocks = existingSubnet.CIDRBlocks
+
+			s.Scope.SetSubnet(subnet)
 
 		case !s.Scope.IsVnetManaged():
 			return fmt.Errorf("vnet was provided but subnet %s is missing", subnetSpec.Name)
--- a/azure/interfaces.go
+++ b/azure/interfaces.go
@@ -64,6 +64,7 @@ type NetworkDescriber interface {
 	IsVnetManaged() bool
 	NodeSubnet() infrav1.SubnetSpec
 	ControlPlaneSubnet() infrav1.SubnetSpec
+	SetSubnet(infrav1.SubnetSpec)
 	IsIPv6Enabled() bool
 	NodeRouteTable() infrav1.RouteTable
 	ControlPlaneRouteTable() infrav1.RouteTable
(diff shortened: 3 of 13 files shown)


nader-ziada:
the unit tests are failing on the lint because `routetables_test.go` is not `gofmt`-ed with `-s` (gofmt)

fiunchinho:
Fixed

azure/services/subnets/subnets.go
```diff
@@ -72,10 +72,10 @@ func (s *Service) Reconcile(ctx context.Context) error {
 				continue
 			}
 
-			subnet.Role = subnetSpec.Role
-			subnet.Name = existingSubnet.Name
-			subnet.CIDRBlocks = existingSubnet.CIDRBlocks
 			subnet.ID = existingSubnet.ID
+			subnet.CIDRBlocks = existingSubnet.CIDRBlocks
+
+			s.Scope.SetSubnet(subnet)
```
nader-ziada, line 78: This is being called from inside a loop going through the subnets, then it does a loop again, might be better to pass in the name or something like that and just updating that subnet, what do you think?

azure/scope/managedcontrolplane.go
```diff
@@ -198,6 +198,14 @@ func (s *ManagedControlPlaneScope) NodeSubnet() infrav1.SubnetSpec {
 	}
 }
 
+// SetSubnet sets the subnet spec for the subnet with the same role.
+func (s *ManagedControlPlaneScope) SetSubnet(subnetSpec infrav1.SubnetSpec) {
+	s.ControlPlane.Spec.VirtualNetwork.Subnet = infrav1exp.ManagedControlPlaneSubnet{
+		Name:      subnetSpec.Name,
+		CIDRBlock: subnetSpec.CIDRBlocks[0],
+	}
+}
```
nader-ziada, line 207: should this be a no op?

nader-ziada (commented):
thanks for the PR @fiunchinho

azure/scope/managedcontrolplane.go
```diff
@@ -198,6 +198,14 @@ func (s *ManagedControlPlaneScope) NodeSubnet() infrav1.SubnetSpec {
 	}
 }
 
+// SetSubnet sets the subnet spec for the subnet with the same role.
```
CecileRobertMichon, line 201: description isn't very accurate here

azure/services/subnets/subnets.go
CecileRobertMichon: the issue with this is that we don't have an efficient way to Get() a particular subnet in O(1)... that's what I attempted to fix in #1280 but it's not best practices for CRDs to use maps so we decided to not move forward with it. Even if we pass in a name we still have to loop through the subnets to find it. And if we were to collect a list of subnets to update and then make one call to SetSubnets() we'd still have a nested for loop to loop through the current subnets and the desired updates and compare names. We _could_ optionally store this "desired updates" list as a map and pass that into SetSubnets() but that's not going to bring that much optimization given that the number of subnets is a constant 2.

CecileRobertMichon:
Please add a release note, this is a user facing change

azure/services/subnets/subnets.go
fiunchinho: Taking into account that the number of subnets won't be a big number, I'd say it's fine. Going for a more complicated solution could harm the readability with not real gains in performance (again, due to the fact that usually we expect a low number of subnets).

azure/scope/managedcontrolplane.go
fiunchinho: Since the function is not used when using a managed control plane, I think you are right.

fiunchinho:
Done

azure/scope/managedcontrolplane.go
```diff
@@ -198,6 +198,12 @@ func (s *ManagedControlPlaneScope) NodeSubnet() infrav1.SubnetSpec {
 	}
 }
 
+// SetSubnet sets the passed subnet spec into the scope.
+// This is not used when using a managed control plane.
+func (s *ManagedControlPlaneScope) SetSubnet(subnetSpec infrav1.SubnetSpec) {
+
```
CecileRobertMichon, line 204:
```suggestion
    return // no-op
```

fiunchinho: If I add the `return` I get a linting error
```
S1023: redundant `return` statement (gosimple)
```

nader-ziada: yeah, I think you can't return because the func has no return type

nader-ziada: are you okay with keeping it empty @CecileRobertMichon ?

CecileRobertMichon: yeah that's fine

CecileRobertMichon:
@fiunchinho please squash

lgtm

fiunchinho:
Done

Issue #1113, opened 2021-01-11
Export traces in OpenTelemetry format and then to Azure Monitor


devigned:
/kind feature
/help

**Describe the solution you'd like**
As a user of Cluster API Provider Azure, I would like to be able to export trace spans in OpenTelemetry format so that they can then be exported to Azure Monitor via the [Azure Monitor Exporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/master/exporter/azuremonitorexporter).

Currently, when a developer runs `make tilt-up`, Jaeger is started on http://localhost:8080 and a Jaeger "all-in-one" sidecar is injected to into the controller pod. The CAPZ controller sends traces to the Jaeger sidecar for direct analysis.

The desired solution is to be able to run `make tilt-up` and have https://github.com/open-telemetry/opentelemetry-collector or https://github.com/open-telemetry/opentelemetry-operator setup with the CAPZ controller exporting traces to a collector configured to export to Azure Monitor using the Azure Monitor OpenTelemetry Exporter.

We should also document the components and configuration needed to create a CAPZ cluster with these components installed for non-development scenarios. (@mboersma, this might be an interesting idea for a CLI gesture)

**Anything else you would like to add:**
The first tracing PR #1019 added Jaeger support. It is a good starting point to learn from, but lacks the flexibility of design which comes with using the OpenTelemetry collector. Separating the collector from the exporter / analysis will enable users of CAPZ to have more control over their stack of tools.

PR #1412, opened 2021-06-04
Update opentelemetry library

mboersma:
**What type of PR is this?**:

/kind cleanup

Updates the [opentelemetry Go library](https://github.com/open-telemetry/opentelemetry-go) to current v0.20.0 ~~and updates the jaeger and prometheus componentry to current versions~~.

Opentelemetry is a moving target; this trains our sights on the most recently seen version.

Prerequisite to #1113

Since this only lights up in `make tilt-up` and there was a fair amount of refactoring required, it would be great to get others to try this out before we merge it.

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Update opentelemetry library
```


mboersma:
> I'm seeing a bunch of errors like this in the jaeger-operator logs

The newer jaeger-operator requires more permissions and more plumbing that I had missed. But after adding the cluster role and its binding, I was still seeing the same namespace-related error. So I removed the jaeger and prometheus component updates from this PR so all it does is update the opentelemetry-go library. Seems to work--PTAL when time permits.

mboersma:
/hold

I don't see Jaeger traces flowing yet, so let me do some more tinkering.

mboersma:
/hold cancel

I had to update two `otel` packages to a more recent commit to benefit from [a fix](https://github.com/open-telemetry/opentelemetry-go/pull/1853) for this jaeger UDP issue:

```
[manager] 2021/06/09 16:30:50 data does not fit within one UDP packet; size 172726, max 65000, spans 793
```

Things seem to be working now.

mboersma:
I had to rebase this because we added new references to the (old) opentelemetry API in code. I have a PR based on this that works with OpenTelemetry components to capture jaeger-compatible traces and send them to Azure Insights, so I'm confident this is working. PTAL if anyone has time.

main.go
```diff
@@ -257,16 +261,29 @@ func main() {
...
+
+		// Cleanly shutdown and flush telemetry when the application exits.
+		defer func(ctx context.Context) {
+			// Do not make the application hang when it is shutdown.
+			ctx, cancel = context.WithTimeout(ctx, time.Second*5)
+			defer cancel()
+			if err := tp.Shutdown(ctx); err != nil {
+				setupLog.Error(err, "failed to shut down Jaeger tracing")
+			}
+		}(ctx)
```
devigned, line 286: The use of `ctx` here is ineffectual. You might as well pass in `context.Background`. 

Please consider setting up something like [Cluster API's signal handler](https://github.com/kubernetes-sigs/cluster-api/blob/19b900defb84434e3a23d786c311c5daf9d94d6d/main.go#L200-L201) to gracefully stop and flush the telemetry when the process receives OS signals to terminate.

devigned (changes requested):
I really like these changes, but would like you to reconsider how the context is being used.

main.go
mboersma: Good call--I'll check out the signal handler.

main.go
```diff
@@ -321,12 +317,34 @@ func main() {
 	}
 
 	setupLog.Info("starting manager", "version", version.Get().String())
-	if err := mgr.Start(ctrl.SetupSignalHandler()); err != nil {
+	if err := mgr.Start(ctx); err != nil {
```
mboersma, line 320: We had just called `ctrl.SetupSignalHandler()` a few lines above, so I assume this code should have been using that `ctx` rather than calling the func again. That's how CAPI does it.

main.go
mboersma: I changed the naive code (copied from otel examples, where it's also ineffectual...) to instead use the context set up by the signal handler.

devigned (approved):
/lgtm
/approve

Looks great!

PR #1413, opened 2021-06-04
Make LB IdleTimeoutInMinutes configurable

### LINKED ISSUE #676: Look into making LB outbound rule IdleTimeoutInMinutes configurable and increase default
/kind feature

From https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/674#discussion_r434848567

> // DefaultOutboundRuleIdleTimeoutInMinutes determines the aks-engine provided default for IdleTimeoutInMinutes of the OutboundRule of the agent loadbalancer
// This value is set greater than the default Linux idle timeout (15.4 min): https://pracucci.com/linux-tcp-rto-min-max-and-tcp-retries2.html

We should:

1) Determine what is a sane default for this value. Should it be the same default for different LB roles? The Azure default is 4 minutes. AKS Engine sets it to 30 minutes, and points to that being greater than the default Linux timeout which is 15.4 minutes. It's possible that there is no action to be taken if 4 is already the right default.

2) Consider making this value user configurable. Before doing this, we need to answer the following questions: is there a user story for configuring this value? What is the best way to represent this in the AzureCluster Spec? Do we expect to expose other options for Load Balancers in the future? Currently, there is no Load Balancer configuration so this might involve creating a new struct to store Load Balancer Spec(s). Keep in mind that CAPZ currently manages the lifecycle of 3 different load balancers per cluster: 1 internal LB, 1 public apiserver LB (control planes in the backend pool), and 1 "node outbound" LB (nodes in the backend pool), which is also the same LB used by cloud provider to define LB rules for services.

**Environment:**

CecileRobertMichon:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**: Enables a user-override for the LoadBalancer idle timeout, which governs how long to keep idle connections open. Certain applications may be idle by design for longer than 4 minutes (the default), and we want to accommodate that. Conversely, certain clusters may not have any such applications but will expect a lot of aggregate outbound connections, in which case they will want to more aggressively close connections to better protect against SNAT port exhaustion.

Fixes #676 

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Make LB IdleTimeoutInMinutes configurable
```


azure/scope/cluster.go
```diff
@@ -208,10 +210,11 @@ func (s *ClusterScope) LBSpecs() []azure.LBSpec {
...
 		},
-		Type:            infrav1.Public,
-		SKU:             infrav1.SKUStandard,
-		BackendPoolName: s.OutboundPoolName(azure.GenerateControlPlaneOutboundLBName(s.ClusterName())),
-		Role:            infrav1.ControlPlaneOutboundRole,
+		Type:                 infrav1.Public,
+		SKU:                  infrav1.SKUStandard,
+		BackendPoolName:      s.OutboundPoolName(azure.GenerateControlPlaneOutboundLBName(s.ClusterName())),
+		Role:                 infrav1.ControlPlaneOutboundRole,
+		IdleTimeoutInMinutes: to.Int32Ptr(4),
```
nader-ziada, line 217: does the default value not apply here?

docs/book/src/topics/node-outbound-lb.md
```diff
@@ -6,7 +6,9 @@ This document describes how to configure your clusters' node outbound load balan
 
 For public clusters ie. clusters with api server load balancer type set to `Public`, CAPZ automatically configures a node outbound load balancer with the default settings.
 
-To provide custom settings for the node outbound load balacer, use the `nodeOutboundLB` section in cluster configuration.
+To provide custom settings for the node outbound load balancer, use the `nodeOutboundLB` section in cluster configuration.
+
+The `idleTimeoutInMinutes` specifies the number of minutes to keep a TCP connection open for the outbound rule. See [here](https://docs.microsoft.com/en-us/azure/load-balancer/load-balancer-tcp-reset#configurable-tcp-idle-timeout) for more details.
```
nader-ziada, line 11: should we mention the default value?

azure/scope/cluster.go
CecileRobertMichon: no, there is no spec for control plane outbound LB currently

PR #1415, opened 2021-06-07
Expose cloud provider backoff config 

### LINKED ISSUE #1382: Cloud provider config customization - cloudProviderBackoff
/kind feature

#1373 allowed rate limits to be customized in cloud provider config. We also might want to do this for other properties. This issue tracks cloud provider backoff config that we want to override. specifically the following ones:

Name | Description | Remark
-- | -- | --
cloudProviderBackoff | Enable exponential backoff to manage resource request retries | Boolean value, default to false
cloudProviderBackoffRetries | Backoff retry limit | Integer value, valid if&nbsp;cloudProviderBackoff&nbsp;is true
cloudProviderBackoffExponent | Backoff exponent | Float value, valid if&nbsp;cloudProviderBackoff&nbsp;is true
cloudProviderBackoffDuration | Backoff duration | Integer value, valid if&nbsp;cloudProviderBackoff&nbsp;is true
cloudProviderBackoffJitter | Backoff jitter | Float value, valid if&nbsp;cloudProviderBackoff&nbsp;is true

https://kubernetes-sigs.github.io/cloud-provider-azure/install/configs/

**Environment:**

Ankitasw:
**What type of PR is this?**
/kind feature

This pr allows for customization of cloud provider backoff config through AzureCluster spec.

Fixes #1382

**TODOs**:
- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:
```release-note
Expose cloud provider backoff config
```


api/v1alpha4/types.go
```diff
@@ -515,6 +515,26 @@ type AddressRecord struct {
 // CloudProviderConfigOverrides represents the fields that can be overridden in azure cloud provider config.
 type CloudProviderConfigOverrides struct {
 	RateLimits []RateLimitSpec `json:"rateLimits,omitempty"`
+	BackOffs   []BackOffSpec   `json:"backOffs,omitempty"`
```
shysank, line 518: Do we need `BackOffSpec` here? Unlike rate limits, there is not going to be different backoff configurations. Can we directly use `BackOffConfig` here?

controllers/helpers.go
```diff
@@ -301,6 +301,10 @@ func (cpc *CloudProviderConfig) overrideFromSpec(d azure.ClusterScoper) *CloudPr
 			cpc.AvailabilitySetRateLimit = toCloudProviderRateLimitConfig(rateLimit.Config)
 		}
 	}
+
+	for _, backOff := range d.CloudProviderConfigOverrides().BackOffs {
```
shysank, line 305: We won't need this loop once the above comment is addressed.

shysank (commented):
@Ankitasw Thanks for the PR! Just a couple of comments on the Backoffs data structure, otherwise looks good so far pending tests.

nader-ziada:
looks good

controllers/helpers.go
```diff
@@ -375,6 +378,31 @@ type RateLimitConfig struct {
 	CloudProviderRateLimitBucketWrite int     `json:"cloudProviderRateLimitBucketWrite,omitempty"`
 }
 
+// BackOffConfig indicates the back-off config options.
+// This is a copy of the struct used in cloud-provider-azure: https://github.com/kubernetes-sigs/cloud-provider-azure/blob/d585c2031925b39c925624302f22f8856e29e352/pkg/azureclients/azure_client_config.go#L48
+type BackOffConfig struct {
+	CloudProviderBackoff         bool    `json:"cloudProviderBackoff,omitempty"`
+	CloudProviderBackoffRetries  int     `json:"cloudProviderBackoffRetries,omitempty"`
+	CloudProviderBackoffExponent float32 `json:"cloudProviderBackoffExponent,omitempty"`
```
shysank, line 386: This should be `float64` to be in line with [cloud-provider-azure](https://github.com/kubernetes-sigs/cloud-provider-azure/blob/d585c2031925b39c925624302f22f8856e29e352/pkg/provider/azure.go#L186)

controllers/helpers.go
```diff
@@ -375,6 +378,31 @@ type RateLimitConfig struct {
...
+	CloudProviderBackoffDuration int     `json:"cloudProviderBackoffDuration,omitempty"`
+	CloudProviderBackoffJitter   float32 `json:"cloudProviderBackoffJitter,omitempty"`
```
shysank, line 388: Same as above => https://github.com/kubernetes-sigs/cloud-provider-azure/blob/d585c2031925b39c925624302f22f8856e29e352/pkg/provider/azure.go#L188

controllers/helpers.go
```diff
@@ -375,6 +378,31 @@ type RateLimitConfig struct {
...
+}
+
+// toCloudProviderBackOffConfig returns converts infrav1.BackOffConfig to BackOffConfig that is required with the cloud provider.
+func toCloudProviderBackOffConfig(source infrav1.BackOffConfig) *BackOffConfig {
```
devigned, line 392: Why return a ptr here if when it is used, it is dereferenced?
```go
cpc.BackOffConfig = *toCloudProviderBackOffConfig(d.CloudProviderConfigOverrides().BackOffs)
```

devigned (changes requested):
Consider returning a struct rather than a ptr from `toCloudProviderBackOffConfig`.

devigned (approved):
lgtm

Thank you!

PR #1417, opened 2021-06-07
Add support for using gen2 image when creating VM

### LINKED ISSUE #1003: Azure Generation 2 VM support
/kind feature

CAPZ should enable me to create clusters with [Azure Generation 2 VMs](https://docs.microsoft.com/en-us/azure/virtual-machines/generation-2).

I can't think of a use case where mixed gen 1 and gen 2 VMs are needed

nader-ziada:
**What type of PR is this?**
/kind feature

- add support to use a Gen2 image when creating a VM

Fixes #1003 

related to PR in image builder https://github.com/kubernetes-sigs/image-builder/pull/627

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Add support for using Gen2 VM images
```


CecileRobertMichon:
/assign @devigned 

to make sure this makes sense with the image defaulting stuff we were discussing for #1332

devigned:
I've update #1332 with the image defaulting behavior. See https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1332#issuecomment-856316521

azure/services/virtualmachines/virtualmachines.go
```diff
@@ -137,6 +137,17 @@ func (s *Service) Reconcile(ctx context.Context) error {
 			return errors.Wrap(err, "failed to generate OS Profile")
 		}
 
+		var generation compute.HyperVGenerationType
+		vmImage, err := s.Scope.GetVMImage()
+		if err != nil {
+			return errors.Wrap(err, "failed to get VM Image")
+		}
+		if vmImage.HyperVGeneration == "" {
```
devigned, line 145: What if someone sets `vmImage.HyperVGeneration == "V1"`? Should this be set in `s.Scope.GetVMImage()`? If it is, probably could get rid of the conditional and just have `compute.HyperVGenerationType(string(vmImage.HyperVGeneration))`.

devigned (changes requested):
Generation interpretation might lead to an invalid state for "V1". Outside of that the change looks solid.

azure/services/virtualmachines/virtualmachines.go
nader-ziada: if its set to `"v1"` the `if` condition would still be fine, basically, I'm setting it to v1 as a default since we don't publish "v2" images yet, eventually we should use v2.

As for moving this to the `scope`; usually translating the types to the `azure sdk` types that will be used in the call to create happens just before that call, so trying to be consistent.

devigned: > As for moving this to the scope; usually translating the types to the azure sdk types that will be used in the call to create happens just before that call, so trying to be consistent.

`s.generateOSProfile` calls `s.GetVMImage` to build the `*compute.StorageProfile`.

nader-ziada: the func `generateOSProfile` is part of virtualmachine.go. not scope.

I might be mis understanding, I though you want me to move the if condition to set the `generation` var to the scope.GetVMImage, so was just saying scope doesn't use anything from the `compute` types

devigned: By setting the `vmImage.HyperVGeneration = "V1" or "V2"` in `s.GetVMImage`, you can avoid compute types, encapsulate the behavior in `s.GetVMImage`, and have simpler code building the compute VM structure with a single line: 
```go
InstanceView: &compute.VirtualMachineInstanceView{
		HyperVGeneration: compute.HyperVGenerationType(string(vmImage.HyperVGeneration)),
},
```

devigned: My fault. I see what you were saying about `generateOSProfile`. Maybe `generateOSProfile` returns the `vmImage.HyperVGeneration` too.

nader-ziada: ok, I see what you are saying, will update the code

nader-ziada: thanks, your suggestions make the code better

azure/services/virtualmachines/virtualmachines.go
```diff
@@ -165,6 +170,9 @@ func (s *Service) Reconcile(ctx context.Context) error {
 						Enabled: to.BoolPtr(true),
 					},
 				},
+				InstanceView: &compute.VirtualMachineInstanceView{
```
CecileRobertMichon, line 173: do we want to support gen2 for vmss as well?

azure/services/virtualmachines/virtualmachines.go
nader-ziada: changed it, not sure if this is exactly what you had in mind, PTAL

azure/services/virtualmachines/virtualmachines.go
nader-ziada: yes we should

nader-ziada: is that even an option, can't find how to set it

CecileRobertMichon: not sure but I feel like it should be, @devigned do you know?

azure/services/virtualmachines/virtualmachines.go
devigned: wdyt about the structure below?

```go
// generateStorageProfile generates a pointer to a compute.StorageProfile which can utilized for VM creation.
func (s *Service) generateStorageProfile(ctx context.Context, vmSpec azure.VMSpec, sku resourceskus.SKU) (*compute.StorageProfile, HyperVGenerationType, error) {
        //....
	return storageProfile, vmImage.HyperVGeneration, nil
}
```

```go
osProfile, hyperVGeneration, err := s.generateOSProfile(ctx, vmSpec)
if err != nil {
     return errors.Wrap(err, "failed to generate OS Profile")
}

// ... omitted code //
      InstanceView: &amp;compute.VirtualMachineInstanceView{
            HyperVGeneration: compute.HyperVGenerationType(string(hyperVGeneration)),
      },
```

azure/services/virtualmachines/virtualmachines.go
devigned: I think it should be. One sec... :eyes:

devigned: I just created a VMSS with a the Ubuntu gen 2 image. There was nothing set in the VMSS PUT payload, so I'm thinking that VMSS gets that information from the image.

```
$ az vm image show -l westus2 -f UbuntuServer -p Canonical -s 19_10-daily-gen2 --version 19.10.202007100
{
  "automaticOsUpgradeProperties": {
    "automaticOsUpgradeSupported": false
  },
  "dataDiskImages": [],
  "disallowed": {
    "vmDiskType": "Unmanaged"
  },
  "hyperVGeneration": "V2",
  "id": "/Subscriptions/sub_id/Providers/Microsoft.Compute/Locations/westus2/Publishers/Canonical/ArtifactTypes/VMImage/Offers/UbuntuServer/Skus/19_10-daily-gen2/Versions/19.10.202007100",
  "location": "westus2",
  "name": "19.10.202007100",
  "osDiskImage": {
    "operatingSystem": "Linux",
    "sizeInBytes": 32213303808,
    "sizeInGb": 31
  },
  "plan": null,
  "tags": null
}
```

I'm questioning whether or not we need to update the `InstanceView`. I don't think that should be needed b/c the Compute resource provider should be responsible for determining the Hyper V generation based on the image used.

nader-ziada:
Update: after more testing, found out that Azure sdk/api will set the `VM generation` based on the image used, so looks like there is no need for us to set that field. 

Closing this PR for now and will investigate if there is any changes needed in capz, like maybe eventually changing the default image when we published gen2 images.

thanks for the review @devigned 

/close

PR #1425, opened 2021-06-08
Add script to run shellcheck on bash scripts in the repository

### LINKED ISSUE #1337: Add a shellcheck job as part of verify
A lot of CI jobs depend on our bash scripts, we should add shellcheck as part of presubmit to avoid small mistakes like this: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1336

/cc @chewong

fristonio:
* Add a bash script to run shellcheck on all the scripts in the repository. Derived from [Kubernetes Shellcheck](https://github.com/kubernetes/kubernetes/blob/master/hack/verify-shellcheck.sh).
* Add `verify-shellcheck` make target and include it as a part of `make verify`.
* Fix shellcheck warnings in the repository. 

/kind other

Fixes #1337

```release-note
NONE
```

### DIFF
--- a/hack/verify-shellcheck.sh
+++ b/hack/verify-shellcheck.sh
@@ -1,0 +1,130 @@
+#!/usr/bin/env bash
+
+# Copyright 2018 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This script lints each shell script by `shellcheck`.
+# Usage: `hack/verify-shellcheck.sh`.
+# Taken from - https://github.com/kubernetes/kubernetes/blob/master/hack/verify-shellcheck.sh
+# modified according to use.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+CAPZ_ROOT=$( cd "$(dirname "${BASH_SOURCE[0]}")/.." &> /dev/null && pwd )
+
+# allow overriding docker cli, which should work fine for this script
+DOCKER="${DOCKER:-docker}"
+SHELLCHECK_COLORIZED_OUTPUT="${SHELLCHECK_COLORIZED_OUTPUT:-auto}"
+
+# required version for this script, if not installed on the host we will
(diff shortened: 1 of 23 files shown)


mboersma:
This is great IMHO! I think `shellcheck` is a very useful tool.

It's failing because there's no `shellcheck` binary in the test environment, so it tries to run the Docker image but fails. I assume this is because we're already in a container that doesn't allow Docker-in-Docker.

I can think of three ways around that:
- change the `verify-shellcheck.sh` script to download the binary and put it in the $PATH if it's not found (maybe only when running in a  container?)
- prime the test container that runs the `pull-cluster-api-provider-azure-verify` job with the `shellcheck` binary
- enable Docker-in-Docker for the test container

I think #2 (make sure `shellcheck` is already there) seems simplest, but I'm not actually sure where to make that change. @fristonio what do you think?

devigned:
> I think option 2 (make sure shellcheck is already there) seems simplest, but I'm not actually sure where to make that change. 

I'd vote for adding a prow job with Docker in Docker support or adding it to the exiting job. Here's an example of a PR: https://github.com/kubernetes/test-infra/commit/4fcd52d5bbed6cc3bd7451894571e2d50c1b0aa6

CecileRobertMichon:
+1 on enabling dind on https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-sigs/cluster-api-provider-azure/cluster-api-provider-azure-presubmits.yaml#L167

(can be done by adding `labels:      preset-dind-enabled: "true"`) to the job definition in test-infra. @fristonio would you like to open that PR?

hack/verify-shellcheck.sh
```diff
@@ -0,0 +1,130 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This script lints each shell script by `shellcheck`.
+# Usage: `hack/verify-shellcheck.sh`.
+# Taken from - https://github.com/kubernetes/kubernetes/blob/master/hack/verify-shellcheck.sh
```
CecileRobertMichon, line 19: thanks for giving due credit here

CecileRobertMichon:
opened https://github.com/kubernetes/test-infra/pull/22559

mboersma:
Enabling Docker-in-Docker did the trick, thanks @CecileRobertMichon! Now it's just failing on actual `shellcheck` linting errors.

fristonio:
Thanks, @CecileRobertMichon for opening the PR to enable dind.
Will take a look at the remaining `shellcheck` failures.

mboersma (approved):
lgtm

devigned (changes requested):
lgtm, but please squash the commits.

CecileRobertMichon:
/lgtm
/approve
Thanks @fristonio !

PR #1035, opened 2020-11-13
✨ Add ability to provision Windows VMs

Danil-Grigorev:
Upstream changes from:
- https://github.com/openshift/cluster-api-provider-azure/pull/140
- https://github.com/openshift/cluster-api-provider-azure/pull/164

 

 
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

This PR provides alternative way to bring Windows VMs in CAPZ.

The functionality allows to bring Windows machines and configure SSH access via `customData` script.

This PR allows to pass `UnattendConfig` to Windows machines on Azure for autostarting custom bootstrap data 
Currently on Azure, the `CustomData` (also called bootstrap data) is written to `C:\AzureData\CustomData.bin`, but nothing will run it.

With this change, Windows machines on Azure will automatically execute the `CustomData` script if it exists.

This is the same behavior as Windows machines on AWS show.

This way of bootstrapping a machine with a provided script in the CustomData (e.g. to setup SSH access, and the kubelet) is leveraged by the [WMCO](https://github.com/openshift/windows-machine-config-operator) and does not require the kubeadm bootstrapping mechanism nor creating custom Windows OS images.

The PR is not requiring amendments to the [proposal](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1035#issuecomment-728258731) due to no plans in using kubeadm bootstrap provider.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/153 for WMCO.

**Special notes for your reviewer**:

cc @aravindhp @LorbusChris please review

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note

```

PR #1433, opened 2021-06-11
Add ability to autorun bootstrap/custom data script on Windows VMs

LorbusChris:
**What type of PR is this?**
/kind feature

This commit adds to the WindowsConfiguration an AdditionalUnattendContent
configuration for automatically logging in as admin user and running the
CustomData/BootstrapData script on first boot.

This aligns the behavior when provisioning machines in the case of a
Windows VM on Azure with the behavior on AWS, where the CustomData
script is also autostarted on first boot. The same is the case for
Linux Machines on both Azure and AWS as well.

With this change, a PowerShell script can be placed on the machine and
will be autorun on first boot which allows for a very flexible way of
e.g. configuring the machine, installing additional software on the
machine, etc., all without the need for a custom Windows boot image.

This PR is intended to supersede PR #1035

/cc @aravindhp
/cc @JoelSpeed
/cc @Danil-Grigorev 
/cc @jsturtevant 

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Add ability to autorun bootstrap/custom data script on Windows VMs
```

### DIFF
--- a/azure/services/virtualmachines/virtualmachines.go
+++ b/azure/services/virtualmachines/virtualmachines.go
@@ -39,6 +39,34 @@ import (
 	"sigs.k8s.io/cluster-api-provider-azure/util/tele"
 )
 
+const (
+	// winAutoLogonFormatString is the format string used to create the AutoLogon
+	// AdditionalUnattendContent configuration for Windows machines.
+	winAutoLogonFormatString = `<AutoLogon>
+			<Username>%s</Username>
+			<Password>
+				<Value>%s</Value>
+			</Password>
+			<Enabled>true</Enabled>
+			<LogonCount>1</LogonCount>
+		</AutoLogon>`
+
+	// winFirstLogonCommandsString is the string used to create the FirstLogonCommands
+	// AdditionalUnattendContent configuration for Windows machines.
+	winFirstLogonCommandsString = `<FirstLogonCommands>
+			<SynchronousCommand>
+				<Description>Copy user data secret contents to init script</Description>
+				<CommandLine>cmd /c "copy C:\AzureData\CustomData.bin C:\init.ps1"</CommandLine>
+				<Order>11</Order>
+			</SynchronousCommand>
+			<SynchronousCommand>
+				<Description>Launch init script</Description>
+				<CommandLine>powershell.exe -NonInteractive -ExecutionPolicy Bypass -File C:\init.ps1</CommandLine>
+				<Order>12</Order>
+			</SynchronousCommand>
+		</FirstLogonCommands>`
+)
+
 // VMScope defines the scope interface for a virtual machines service.
 type VMScope interface {
 	logr.Logger
@@ -468,6 +496,20 @@ func (s *Service) generateOSProfile(ctx context.Context, vmSpec azure.VMSpec) (*
 		osProfile.AdminPassword = to.StringPtr(generators.SudoRandomPassword(123))
 		osProfile.WindowsConfiguration = &compute.WindowsConfiguration{
 			EnableAutomaticUpdates: to.BoolPtr(false),
+			AdditionalUnattendContent: &[]compute.AdditionalUnattendContent{
+				{
+					PassName:      "OobeSystem",
+					ComponentName: "Microsoft-Windows-Shell-Setup",
+					SettingName:   "AutoLogon",
+					Content:       to.StringPtr(fmt.Sprintf(winAutoLogonFormatString, azure.DefaultUserName, *osProfile.AdminPassword)),
+				},
+				{
+					PassName:      "OobeSystem",
+					ComponentName: "Microsoft-Windows-Shell-Setup",
+					SettingName:   "FirstLogonCommands",
+					Content:       to.StringPtr(winFirstLogonCommandsString),
+				},
+			},
 		}
 	default:
 		osProfile.LinuxConfiguration = &compute.LinuxConfiguration{


LorbusChris:
looking at the e2e test logs, I'm not sure happened.

jsturtevant:
The end to end tests use [Kubeadm bootstrapper](https://cluster-api.sigs.k8s.io/developer/architecture/controllers/bootstrap.html) and cloudbase-init to launch based of what is in the CustomData field put there by CABPK.  Since the that data is [cloud config format ](https://cloudbase-init.readthedocs.io/en/latest/userdata.html#cloud-config) I don't think coping it directly over is going to work.  

Seems like we need a way to determine if it should be using Kubeadm bootstraper or something else because right now it just executes what ever is there which will interfere with cloudbase-init's bootstrapping mechanism.  We would also need to wire in a different bootstrap provider to be able to test this scenario.

LorbusChris:
The tests have all passed now...

We expressly do not wish to use Kubeadm bootstrapping here - this is intended to be orthogonal (or an alternative) to that.

We'd really like to align behavior of different CAPI providers when provisioning Windows VMs and we definitely require the ability to place a script on the machine and have it autostart on first boot, like it's done by e.g. the AWS provider.

I'm not very familiar with this code base, but it sounds like adding an option to disable the Kubeadm bootstraper would be desireble then.

We could also guard against trying to exec a non-PowerShell script by looking at the CustomData contents first, before actually running in.

jsturtevant:
> We expressly do not wish to use Kubeadm bootstrapping here - this is intended to be orthogonal (or an alternative) to that.

Yup, I understand that. The e2e tests currently use kubeadm bootstrapper and so I was surprised that they passed with this modification as at first glance looks like it would be conflicting.

It does seem in the tests, that the unattend file that is used by cloudbase-init is by-passing this change. If we look at the [output](https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/1433/pull-cluster-api-provider-azure-e2e-windows/1403378790633050112/artifacts/clusters/capz-e2e-ew10fv/machine-pools/capz-e2e-ew10fv-mp-win/win-p-win000000/cloudbase-init.log) of the tests runs we can see that the cloudbase-init and kubeadm scripts where used:

```
kubeadm join --config /run/kubeadm/kubeadm-join-config.yaml    && echo success  1>/run/cluster-api/bootstrap-success.complete \r\n[preflight] Running pre-flight checks\n[preflight] Reading configuration from the cluster...\n[preflight] FYI: You can look at this config file with \'kubectl -n kube-system get cm kubeadm-config -oyaml\'\n[kubelet-start] Writing kubelet configuration to file "\\\\var\\\\lib\\\\kubelet\\\\config.yaml"\n[kubelet-start] Writing kubelet environment file with flags to file "\\\\var\\\\lib\\\\kubelet\\\\kubeadm-flags.env"\n[kubelet-start] Starting the kubelet\n[kubelet-start] Waiting for the kubelet to perform the TLS Bootstrap...\n\nThis node has joined the cluster:\n* Certificate signing request was sent to apiserver and a response was received.\n* The Kubelet was informed of the new secure connection details.\n\nRun \'kubectl get nodes\' on the control-plane to see this node join the cluster.\n\n\r\nC:\\windows\\system32>nssm set kubelet start SERVICE_AUTO_START \r\nSet parameter "Start" for service "kubelet".\r\r\n' execute_user_data_script C:\Program Files\Cloudbase Solutions\Cloudbase-Init\Python\lib\site-packages\cloudbaseinit\plugins\common\userdatautils.py:93
```

I think we will want to ensure that this change is tested in someway and only turned on if using an alternative bootstrapper.  We would also want to update docs to show users how this works.

I am not sure what the best way to do that is, as we would need to have some type of alternative bootstrapper integrated into the code base here?

azure/services/virtualmachines/virtualmachines.go
```diff
@@ -468,6 +496,20 @@ func (s *Service) generateOSProfile(ctx context.Context, vmSpec azure.VMSpec) (*
 		osProfile.AdminPassword = to.StringPtr(generators.SudoRandomPassword(123))
 		osProfile.WindowsConfiguration = &compute.WindowsConfiguration{
 			EnableAutomaticUpdates: to.BoolPtr(false),
+			AdditionalUnattendContent: &[]compute.AdditionalUnattendContent{
```
devigned, line 499: Perhaps, this would be better exposed through an API, rather than statically defining.

devigned (changes requested):
I think this feature might be better served by adding an external API for providing `AdditionalUnattendContent` via `AzureMachine` and `AzureMachinePool`.

wdyt?

LorbusChris:
> I think this feature might be better served by adding an external API for providing `AdditionalUnattendContent` via `AzureMachine` and `AzureMachinePool`.
> 
> wdyt?

I like that idea!

Unfortunately this isn't very high up on my TODOs, so it might be some time till I can circle back to it, especially as the above will be a bit more involved than what's currently here.

Happy to help anyone who'd like to pick this up in the meantime :)

CecileRobertMichon:
> Unfortunately this isn't very high up on my TODOs, so it might be some time till I can circle back to it, especially as the above will be a bit more involved than what's currently here.

Please open an issue to detail what you'd want to see here. Whoever picks it up should open a new PR.

/close

aravindhp:
> > Unfortunately this isn't very high up on my TODOs, so it might be some time till I can circle back to it, especially as the above will be a bit more involved than what's currently here.
> 
> Please open an issue to detail what you'd want to see here. Whoever picks it up should open a new PR.
> 
> /close

@JoelSpeed @Danil-Grigorev can one of you pick this up?

PR #1435, opened 2021-06-11
cordon and drain azuremachinepoolmachines prior to delete

### LINKED ISSUE #1434: AzureMachinePoolMachine cordon and drain
/kind feature

An `AzureMachinePoolMachine` should cordon and drain nodes when a node is to be deleted with respect to `AzureMachinePool.Spec.NodeDrainTimeout`. If the node is not drained prior to issuing a DELETE to Azure, the node will be forcefully removed and could cause workloads to be forcefully terminated.

related #819

devigned:
**What type of PR is this?**

/kind feature

If an `AzureMachinePoolMachine` is deleted during scale down or upgrade, the workload needs to be migrated from the node running on that virtual machine to ensure the workload has time to gracefully terminate.

This PR introduces `AzureMachinePool.Spec.NodeDrainTimeout` which matches the same behavior as the CAPI machine drain behavior.

Thank you for the pairing, @mboersma.

Fixes #1434 
Related to #819 

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
cordon and drain AzureMachinePoolMachines prior to delete or upgrade operations
```


devigned:
I still need to add tests, but I just wanted to get this into a PR to show the direction and get early feedback.

devigned:
/retest

IPv6 failed (unrelated)

devigned:
/retest

prow flake fetching az cli

devigned:
/retest

azure cli download fail

azure/scope/machinepoolmachine.go
```diff
@@ -261,6 +274,152 @@ func (s *MachinePoolMachineScope) UpdateStatus(ctx context.Context) error {
...
+				"pod", fmt.Sprintf("%s/%s", pod.Name, pod.Namespace))
+		},
+		Out:    writer{klog.Info},
+		ErrOut: writer{klog.Error},
+		DryRun: false,
+	}
+
+	if noderefutil.IsNodeUnreachable(node) {
+		// When the node is unreachable and some pods are not evicted for as long as this timeout, we ignore them.
+		drainer.SkipWaitForDeleteTimeoutSeconds = 60 * 5 // 5 minutes
```
shysank, line 373: nit
```suggestion
		drainer.SkipWaitForDeleteTimeoutSeconds = int(5 * time.Minute)
```

azure/scope/machinepoolmachine.go
```diff
@@ -261,6 +274,152 @@ func (s *MachinePoolMachineScope) UpdateStatus(ctx context.Context) error {
...
+		return false
+	}
+
+	// if the draining succeeded condition does not exist
+	if conditions.Get(s.AzureMachinePoolMachine, clusterv1.DrainingSucceededCondition) == nil {
+		return false
+	}
+
+	now := time.Now()
+	firstTimeDrain := conditions.GetLastTransitionTime(s.AzureMachinePoolMachine, clusterv1.DrainingSucceededCondition)
```
shysank, line 418: Clarification:`DrainSucceededCondition` is marked false if there is an error as well [here](https://github.com/devigned/cluster-api-provider-azure/blob/6dae9549dd2949ace7ba1384eaf7e86798edc884/azure/scope/machinepoolmachine.go#L319) with a different severity and message. Will this update the transition time?

azure/scope/machinepoolmachine.go
```diff
@@ -261,6 +274,152 @@ func (s *MachinePoolMachineScope) UpdateStatus(ctx context.Context) error {
...
+	}
+
+	return nil
+}
+
+func (s *MachinePoolMachineScope) drainNode(ctx context.Context, node *corev1.Node) error {
+	ctx, span := tele.Tracer().Start(ctx, "scope.MachinePoolMachineScope.drainNode")
+	defer span.End()
+
+	restConfig, err := remote.RESTConfig(ctx, MachinePoolMachineScopeName, s.client, client.ObjectKey{
```
shysank, line 333: can we refactor to use workloadClusterProxy's client here?

shysank (commented):
Looks good overall with a couple of minor things.

azure/scope/machinepoolmachine.go
devigned: I don't think that will do what you think it does: https://play.golang.org/p/yk9BWxHPCk-

azure/scope/machinepoolmachine.go
devigned: Sure. I just lifted this code from the CAPI implementation. I didn't put much work into this section as I think https://github.com/kubernetes-sigs/cluster-api/issues/4063 will end up superseding this behavior.

devigned: `framework.ClusterProxy` is part of the test framework. I don't think we should have a dependency on the test frame work in the infra provider.

azure/scope/machinepoolmachine.go
devigned: This is a great catch. Based on the following, I think you are correct:
- Check in transition to see if conditions match
  - https://github.com/kubernetes-sigs/cluster-api/blob/59c620c36aead9e76157210a063c05d33c3bd276/util/conditions/setter.go#L53-L57
- Check sameness
  - https://github.com/kubernetes-sigs/cluster-api/blob/59c620c36aead9e76157210a063c05d33c3bd276/util/conditions/setter.go#L167-L175

devigned: Updated the `MarkFalse` error location with a hefty comment and a guard to ensure the condition is `nil` before setting.

azure/scope/machinepoolmachine.go
shysank: Ah, You're right. Thanks for the explanation!

azure/scope/machinepoolmachine.go
shysank: Maybe I misunderstood, I was referring to [workloadClusterProxy](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/128493dd756748e9d3efb0bd274064de48491a86/azure/scope/machinepoolmachine.go#L49) in MachinePoolMachineScope. I see that it already has a client for the workload cluster. So, I was wondering, maybe we can update the NodeGetter interface to get the client as well.  I don't think it's a big deal though, fine with creating a followup if it makes sense.

devigned: I think I misunderstood. I think I understand a little better now. I dried up the `getWorkloadClient` func to use `remote.NewClusterClient` which then depends on `remote.RESTConfig`.

I don't think we can use the remote client in the `workloadClusterProxy` because we need to construct `k8s.io/client-go/kubernetes.Interface` for: 
https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1435/files#diff-52e6076fd15ed8417a4824b81a78ef786b227161e9ad53fbe62268f21d7a3a3bR354

devigned:
/retest

azure cli fetch failure on e2e windows

nader-ziada:
nice comments around the drain conditions code! thanks :) 

/lgtm
/approve

nader-ziada:
the flaky timeouts hit again

PR #1460, opened 2021-06-18
Adding correlation IDs to reconcile loops

### LINKED ISSUE #1310: Azure API requests through the SDK should use a shared correlation ID for a given reconcile loop
/kind feature

arschles:
Signed-off-by: Aaron Schlesinger <aaron@ecomaz.net>

 

 
**What type of PR is this?**

/kind feature

This PR adds shared correlation IDs to `context.Context`s passed to reconcile loops, so that the API requests made to Azure get logged and tracked properly. That feature is especially useful for debugging.

Fixes #1310

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Add 'x-ms-correlation-id' headers to all Azure API calls via distributed traces.
```

### DIFF
--- a/controllers/azuremachine_controller.go
+++ b/controllers/azuremachine_controller.go
@@ -137,7 +137,8 @@ func (r *AzureMachineReconciler) Reconcile(ctx context.Context, req ctrl.Request
 			attribute.String("namespace", req.Namespace),
 			attribute.String("name", req.Name),
 			attribute.String("kind", "AzureMachine"),
-		))
+		),
+	)
 	defer span.End()
 
 	// Fetch the AzureMachine VM.
--- a/exp/controllers/azuremachinepool_controller.go
+++ b/exp/controllers/azuremachinepool_controller.go
@@ -168,7 +168,8 @@ func (ampr *AzureMachinePoolReconciler) Reconcile(ctx context.Context, req ctrl.
 			attribute.String("namespace", req.Namespace),
 			attribute.String("name", req.Name),
 			attribute.String("kind", "AzureMachinePool"),
-		))
+		),
+	)
 	defer span.End()
 
 	azMachinePool := &infrav1exp.AzureMachinePool{}
--- a/exp/controllers/azuremachinepoolmachine_controller.go
+++ b/exp/controllers/azuremachinepoolmachine_controller.go
@@ -136,7 +136,8 @@ func (ampmr *AzureMachinePoolMachineController) Reconcile(ctx context.Context, r
 			attribute.String("namespace", req.Namespace),
 			attribute.String("name", req.Name),
 			attribute.String("kind", "AzureMachinePoolMachine"),
-		))
+		),
+	)
 	defer span.End()
 
 	machine := &infrav1exp.AzureMachinePoolMachine{}
--- a/util/tele/corr_id.go
+++ b/util/tele/corr_id.go
@@ -1,0 +1,78 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package tele
+
+import (
+	"context"
+
+	"github.com/google/uuid"
+)
+
+type corrIDKey string
+
+// CorrID is a correlation ID that the cluster API provider
+// sends with all API requests to Azure. Do not create one
+// of these manually. Instead, use the CtxWithCorrelationID function
+// to create one of these within a context.Context.
+type CorrID string
+
+const corrIDKeyVal corrIDKey = "x-ms-correlation-id"
+
+// ctxWithCorrID creates a CorrID and creates a new context.Context
+// with the new CorrID in it. It returns the _new_ context and the
+// newly created CorrID. If there was a problem creating the correlation
+// ID, the new context will not have the correlation ID in it and the
+// returned CorrID will be the empty string.After you call this function, prefer to
+// use the newly created context over the old one. Common usage is
+// below:
+//
+// 	ctx := context.Background()
+//	ctx, newCorrID := CtxWithCorrID(ctx)
+//	fmt.Println("new corr ID: ", newCorrID)
+//	doSomething(ctx)
+func ctxWithCorrID(ctx context.Context) (context.Context, CorrID) {
+	currentCorrIDIface := ctx.Value(corrIDKeyVal)
+	if currentCorrIDIface != nil {
+		currentCorrID, ok := currentCorrIDIface.(CorrID)
+		if ok {
+			return ctx, currentCorrID
+		}
+	}
+	uid, err := uuid.NewRandom()
+	if err != nil {
+		return nil, CorrID("")
+	}
+	newCorrID := CorrID(uid.String())
+	ctx = context.WithValue(ctx, corrIDKeyVal, newCorrID)
+	return ctx, newCorrID
+}
+
+// CorrIDFromCtx attempts to fetch a correlation ID from the given
+// context.Context. If none exists, returns an empty CorrID and false.
+// Otherwise returns the CorrID value and true.
+func CorrIDFromCtx(ctx context.Context) (CorrID, bool) {
+	currentCorrIDIface := ctx.Value(corrIDKeyVal)
+	if currentCorrIDIface == nil {
+		return CorrID(""), false
+	}
+
+	if corrID, ok := currentCorrIDIface.(CorrID); ok {
+		return corrID, ok
+	}
+
+	return CorrID(""), false
+}
--- a/util/tele/tele.go
+++ b/util/tele/tele.go
@@ -17,11 +17,39 @@ limitations under the License.
 package tele
 
 import (
+	"context"
+
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/trace"
 )
 
-// Tracer returns the default opentelemetry tracer.
+type tracer struct {
+	trace.Tracer
+}
+
+// Start creates a new context with a new Azure correlation ID, then
+// creates a new trace.Span with that new context. This function then
+// returns the new Context and Span.
+func (t tracer) Start(
+	ctx context.Context,
+	op string,
+	opts ...trace.SpanOption,
+) (context.Context, trace.Span) {
+	ctx, _ = ctxWithCorrID(ctx)
+	return t.Tracer.Start(ctx, op, opts...)
+}
+
+// Tracer returns an OpenTelemetry Tracer implementation to be used
+// to create spans. Use this implementation instead of the "raw" one that
+// you could otherwise get from calling `otel.Tracer("whatever")`.
+//
+// Example usage:
+//
+//	ctx, span := tele.Tracer().Start(ctx, "myFunction")
+//	defer span.End()
+//	// use the span and context here
 func Tracer() trace.Tracer {
-	return otel.Tracer("capz")
+	return tracer{
+		Tracer: otel.Tracer("capz"),
+	}
 }


CecileRobertMichon:
/ok-to-test

Welcome @arschles!

devigned (commented):
WDYT about changing the way that `tele.Tracer.Start()` works such that anytime start is called and the context does not contain a correlationID, one is generated and added to the value bag. This could be done as a best effort kind of thing, so the error could be ignored, or a default value substituted if a new UUID can not be created?

Also, `pkg/trace/corr.go` requires a file header comment. You will see them littered through the other code files. If you run `make test`, it will generate files, lint, and run the local tests.

arschles:
@devigned I think that's a great idea! I'm working on both the `tele.Tracer.Start()` change and fixing build issues now.

pkg/trace/span.go
```diff
@@ -0,0 +1,93 @@
...
+// See below for an example:
+//
+//	ctx := context.Background()
+//	ctx, corrID, newSpan, err := StartSpan(ctx, tele.Tracer(), "myNewSpan")
+//	if err != nil {
+//		log.Fatal(err)
+//	}
+//	defer newSpan.End()
+//	doSomething(ctx)
+func StartSpan(
```
devigned, line 54: WDYT of something like this:

In `./util/tele.go`, we wrap the `otel.Tracer` with our own struct that implements `trace.Tracer`. We then write our own `Start(...)` func which would then delegate back to `otel.Tracer().Start(...)`. That way the interface can stay the same.
```go
package tele

type tracer struct {
    otel.Tracer
}

func (t *tracer) Start(ctx, opName string, opt ...otel.SpanOption) (context.Context, otel.Span) {
    // add correlationID to ctx and continue as normal
    // return the inner implementation of Start
    return t.Tracer.Start(ctx, opName, opt...)
}

```

Then either in the same package or elsewhere, add context helpers to set / fetch the correlationID.

```go
package corr  // or something else...

// FromCtx or some better name you think of
func FromCtx(ctx context.Context) (ctx, string) {
    // create new context.WithValue containing correlationID and return both if it does not exist, else return context and correlationID
}

```

arschles: Sure, happy to do that! A few questions before I get started, just to make sure I'm on the same page

- I'm assuming that you'd want to still call `tele.Tracer().Start(...)` in all the reconcilers, correct?
- Are you ok with creating our own tracer instance (the wrapper you mentioned) and then passing that -- instead of the default one -- to `otel.SetTracerProvider` in `main.go`?
- Do we need `FromCtx` to be exported at all? It seems like it would only be used inside the same package as the custom tracer, but not sure if you're thinking something more

devigned: > I'm assuming that you'd want to still call tele.Tracer().Start(...) in all the reconcilers, correct?

If possible, that would be great.

> Are you ok with creating our own tracer instance (the wrapper you mentioned) and then passing that -- instead of the default one -- to otel.SetTracerProvider in main.go?

:100: 

> Do we need FromCtx to be exported at all? It seems like it would only be used inside the same package as the custom tracer, but not sure if you're thinking something more

I was adding that because I saw you were returning the correlationID in your `StartSpan` func. This led me to think you wanted a way to access it from the context. I think one might need to access the correlationID from the context to be able to add that as a HTTP header for the Azure SDK for Go clients.

arschles: perfect - I'll make these changes.

regarding `FromCtx` - just in case someone wants to access the correlation ID, I'll make `FromCtx` exported like you said.

arschles: @devigned just an update, so you know I haven't forgotten about this! I'm looking through the implementation & docs trying to figure out a way to create a new [`TracerProvider`](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/trace#TracerProvider) that we can pass to the `SetTracerProvider` function in `main.go`. That would need to create a new `Tracer` which wraps the implementation you referenced above. I think that's going to require a lot of plumbing, which we could make unnecessary if we called a custom function to get either:

- get a new tracer that which wraps the underlying tracer implementation, like what you showed here
- create a new span, which is what's already in this PR

I'll keep digging, but let me know about going with either of the above options in case there isn't a good way to wrap everything we need

arschles: Another option would be to create a custom `TracerProvider` that wraps the jaeger one. Let me know your thoughts on that too?

devigned: Sorry for the delay, @arschles. I was thinking something more like this: https://github.com/kubernetes-sigs/cluster-api-provider-azure/compare/master...devigned:corr

The above diff should add the correlationID to the context without needing to manipulate the signature of the Tracer. wdyt?

arschles: Hi @devigned I see what you mean now! Thanks, I'll implement that.

arschles:
@devigned I made approximately the change you suggested. the only material difference is that I made the custom `tracer`'s `Start(...)` method create the new `context.Context` with the correlation ID in it, and also call the wrapped `trace.Tracer.Start()` function. let me know if this is more along the same lines as what you're thinking.

config/rbac/role.yaml
```diff
@@ -134,8 +134,6 @@ rules:
   - azuremachinepoolmachines/status
   verbs:
   - get
-  - patch
```
devigned, line 137: I don't think you meant to do this.

exp/controllers/azuremachinepool_controller.go
```diff
@@ -158,19 +157,16 @@ func (ampr *AzureMachinePoolReconciler) SetupWithManager(ctx context.Context, mg
 
 // Reconcile idempotently gets, creates, and updates a machine pool.
 func (ampr *AzureMachinePoolReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Result, reterr error) {
+	ctx, span := tele.Tracer().Start(
```
devigned, line 160: I'd hate to lose the trace attributes that were added to this span. Please consider reverting to the previous Start.

exp/controllers/azuremanagedcluster_controller.go
```diff
@@ -100,12 +98,7 @@ func (r *AzureManagedClusterReconciler) Reconcile(ctx context.Context, req ctrl.
 	defer cancel()
 	log := r.Log.WithValues("namespace", req.Namespace, "azureManagedCluster", req.Name)
 
-	ctx, span := tele.Tracer().Start(ctx, "controllers.AzureManagedClusterReconciler.Reconcile",
```
devigned, line 103: Again, the attributes are valuable. Please consider reverting the change to Start.

devigned (changes requested):
I like the work done in the `tele` package to create the tracer adapter and enhancing the `Start` func to do a bit more.

Please consider reverting the changes to files outside of `./util/tele`. Some of the changes to invocations of the `Start` func changed the attributes being supplied or were solely formatting changes.

config/rbac/role.yaml
arschles: wowza, the generators might have removed this. sorry about that. re-added in https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1460/commits/65ca5042956a85e50911c6a7f22d2975480921f6

exp/controllers/azuremanagedcluster_controller.go
arschles: sorry about that! re-added in https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1460/commits/f21e1c8a0c67c16b223d2d3e3ac57d00706a36a1

exp/controllers/azuremachinepool_controller.go
arschles: sorry about that! re-added in https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1460/commits/f21e1c8a0c67c16b223d2d3e3ac57d00706a36a1

arschles:
hey @devigned - would you mind taking another look when you have some time?

devigned:
> Please consider reverting the changes to files outside of `./util/tele`. Some of the changes to invocations of the `Start` func changed the attributes being supplied or were solely formatting changes.

I'm sorry for the delay. I was hoping you were going to revert the changes outside of `./util/tele` since the `Start` interface had stayed the same. If you feel strongly about those formatting changes, then I not opposed. I was just thinking it would make the change set tighter and leave less question about the intent of the changes.

This change set adds the correlationID, but it doesn't actually propagate it on the wire. That will need to be added in a future PR, or this one if you want to continue it.

Again, sorry for any confusion. I'm excited to get this pulled in.

CecileRobertMichon:
@arschles have you thought about how these correlation IDs would get propagated to the actual Azure API calls? I assume a we'd need to set the `x-ms-client-correlation-id` header on each request made to Azure?

arschles:
@devigned got it, I've removed most of the formatting changes made earlier - there are a few still here in which I moved the creation of cancellable contexts below the creation of the spans. that is so the new contexts can be based on the contexts that were created by the new span. I do think it's valuable - but not critical - to do that so the cancellable context has the new correlation IDs in it. let me know what you think of that.

also to address your and @CecileRobertMichon's point about percolating these correlation IDs up to the Azure API - I had originally planned on doing that in a follow-up PR when this PR was large, to draw a clear line between the change to add the IDs and the change to send them up to Azure. now that this is smaller, would you prefer I do that all in this PR?

I'm happy to do it either way, whatever makes it easier for you all

exp/controllers/azuremachinepoolmachine_controller.go
```diff
@@ -127,17 +127,17 @@ func (ampmr *AzureMachinePoolMachineController) SetupWithManager(ctx context.Con
 
 // Reconcile idempotently gets, creates, and updates a machine pool.
 func (ampmr *AzureMachinePoolMachineController) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Result, reterr error) {
-	ctx, cancel := context.WithTimeout(ctx, reconciler.DefaultedLoopTimeout(ampmr.ReconcileTimeout))
-	defer cancel()
-	logger := ampmr.Log.WithValues("namespace", req.Namespace, "azureMachinePoolMachine", req.Name)
-
-	ctx, span := tele.Tracer().Start(ctx, "controllers.AzureMachinePoolMachineController.Reconcile",
+	ctx, span := tele.Tracer().Start(ctx, "controllers.AzureMachinePoolController.Reconcile",
```
CecileRobertMichon, line 130:
```suggestion
	ctx, span := tele.Tracer().Start(ctx, "controllers. AzureMachinePoolMachineController.Reconcile",
```

CecileRobertMichon:
> I had originally planned on doing that in a follow-up PR when this PR was large, to draw a clear line between the change to add the IDs and the change to send them up to Azure. now that this is smaller, would you prefer I do that all in this PR?

separate PR is fine with me, up to you

exp/controllers/azuremachinepoolmachine_controller.go
arschles: Thanks @CecileRobertMichon !

devigned (approved):
/lgtm
/assign @CecileRobertMichon

arschles:
Thanks @CecileRobertMichon - I'll submit a follow-up PR with those changes

exp/controllers/azuremachinepool_controller.go
```diff
@@ -158,18 +158,18 @@ func (ampr *AzureMachinePoolReconciler) SetupWithManager(ctx context.Context, mg
...
 	ctx, span := tele.Tracer().Start(ctx, "controllers.AzureMachinePoolReconciler.Reconcile",
 		trace.WithAttributes(
 			attribute.String("namespace", req.Namespace),
 			attribute.String("name", req.Name),
 			attribute.String("kind", "AzureMachinePool"),
-		))
+		),
+	)
 	defer span.End()
+	ctx, cancel := context.WithTimeout(ctx, reconciler.DefaultedLoopTimeout(ampr.ReconcileTimeout))
```
CecileRobertMichon, line 169: why the change in order here but not in every controller? azuremachine and azurecluster still have the reverse order

arschles: I moved this context creation down to below the creation of the span so that the timeout context includes the correlation ID value. IIRC the others had this change but I reverted them by accident. Would it make more sense to save all these changes for a separate PR?

CecileRobertMichon: either way, as long as the changes are consistent (make the changes here to all the controllers or none of them)

arschles: ok, going to take this change out then because it's not strictly necessary

controllers/azuremachine_controller.go
```diff
@@ -137,7 +137,8 @@ func (r *AzureMachineReconciler) Reconcile(ctx context.Context, req ctrl.Request
 			attribute.String("namespace", req.Namespace),
 			attribute.String("name", req.Name),
 			attribute.String("kind", "AzureMachine"),
-		))
+		),
+	)
```
arschles, line 141: @CecileRobertMichon are you ok with this the the other similar format fixes?

Issue #1394, opened 2021-05-27
Support existing IP as node outbound IPs


yangl900:
/kind feature

**Describe the solution you'd like**

In many cases user would like to configure the cluster outbound IP to a set of pre-created static IPs or IP prefixes. External systems like SQL firewall will whitelist this IP in its ACL. 

We should allow customization of node outbound IPs to reuse existing IPs. This way new clusters, or re-created clusters can reuse the same set of IP, such that all remote ACLs won't need update.

Interface would be a new `outboundIPs` property in the `NodeOutboundLB` property from azurecluster

Summary
* Use outbound IP as ACL is common practice
* Modifying those ACLs are difficult, sometimes impossible
* Re-create cluster / migrating old clusters (from self-managed) to cluster API is common scenario
* To allow reusing same ACL, we should allow reuse existing IP as outbound

**Anything else you would like to add:**

**Environment:**

- cluster-api-provider-azure version: any
- Kubernetes version: (use `kubectl version`): any
- OS (e.g. from `/etc/os-release`): Ubuntu

CecileRobertMichon:
Related to #1159

shysank:
@Ankitasw are you still working on this one?

Ankitasw:
Hey @shysank no I have not got enough time to relook into this, if somebody is interested, they can take ahead the existing PR opened and work on it.

CecileRobertMichon:
Prior work: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1747

nawazkh:
Had an internal discussion with the team about this issue. 
We are trying to test another approach on exposing all the LB Specs, hosted by ASO, wrapped by CAPZ. 
If that succeeds, requests like these, where the user wants to pick different permutations of resources to acheive their green path (here that resource is LB and public IP) would get much easier to achieve. 
If that approach doesn't succeed, we will reopen this issue. 

For now, closing out the issue and the PR related to it. 
/close

PR #1466, opened 2021-06-23
Make control plane outbound lb configurable

shysank:
**What type of PR is this?**
/kind feature

- Adds a field `ControlPlaneOutboundLB` in `AzureCluster` to configure control plane outbound load balancer. This is useful for private clusters when we don't want a public control plane lb created by default
- Works very similar to `NodeOutboundLB`.

Fixes #
Addresses https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1159

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Make control plane outbound lb configurable
```


CecileRobertMichon:
@fiunchinho what's the impact of this PR on #894 if any?

fiunchinho:
> @fiunchinho what's the impact of this PR on #894 if any?

I don't think it has any impact, as we only reconcile nat gateways for worker nodes (we don't for the control plane).

CecileRobertMichon:
@shysank PR needs rebase

is this enough to close #1159 ?

shysank:
> is this enough to close #1159 ?

I think so, not sure if we should consider #1394 to be part of it.

CecileRobertMichon:
> I think so, not sure if we should consider #1394 to be part of it.

Let's keep #1394 separate since it has its own issue

api/v1alpha4/azurecluster_validation.go
```diff
@@ -420,6 +422,34 @@ func validateNodeOutboundLB(lb *LoadBalancerSpec, old *LoadBalancerSpec, apiserv
...
+func validateControlPlaneOutboundLB(lb *LoadBalancerSpec, apiserverLB LoadBalancerSpec, fldPath *field.Path) field.ErrorList {
+	var allErrs field.ErrorList
+
+	switch apiserverLB.Type {
+	case Public:
+		if lb != nil {
+			allErrs = append(allErrs, field.Forbidden(fldPath, "Control plane outbound load balancer cannot be set for public clusters."))
+		}
+	case Internal:
+		// Control plane outbound lb can be nil when it's disabled private clusters.
```
CecileRobertMichon, line 434:
```suggestion
		// Control plane outbound lb can be nil when it's disabled for private clusters.
```

CecileRobertMichon (commented):
/lgtm

PR #1397, opened 2021-05-28
 🐛 fix deletion, speed up creation for aks clusters

alexeldeib:
**What type of PR is this?**
/kind bug
/kind cleanup

**What this PR does / why we need it**:
- fixes [circular deletion dependencies](https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1395#issuecomment-849982900)
- fixes [CAPZ-managed vnet deletion for BYO resource groups](https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1395#issuecomment-850037786).
- Cluster infra ref needs to be ready before CAPI will reconcile the control plane ref. hardcode AzureManagedCluster.Status.Ready to true, since it's meaningless anyway. This speeds up provisioning, since CAPI controllers don't need to wait for AKS to be ready for the field to be true.
- AzureManagedCluster needs to be reconciled from AzureManagedControlPlane events so it can sync its status. We should implement https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1396 and avoid this dance entirely.
- AzureManagedMachinePool doesn't need to watch AzureManagedCluster, since cluster just copies from AzureManagedControlPlane, and it gets the owner cluster from the owning machine pool already.
- If the MachinePool owning the defaultPool of an AzureManagedControlPlane is nil when we try to reconcile, we need to re-reconcile. We can either return an error or watch machinePools and filter them down to the one owning the defaultPool (more accurate, but requires watching and filtering all machinePool events). I implemented the latter.
- Add a `ResourceGroupNotFound` error check during deletion. I accidentally put two CAPZ clusters in one RG. Deleting one deleted the RG, but then the other failed with "ResourceGroupNotFound" which unfortunately does not manifest as 404, because autorest wraps it as a StatusCode: 0.
- Don't fail tilt-up so long as the environment already has Azure credentials defined.
- Bump AKS version default to v1.20.5

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #1395

**Special notes for your reviewer**:

**TODOs**:
- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```release-note
Fix deletion and speed up provisioning for CAPI clusters using AzureManagedCluster + AzureManagedControlPlane (AKS).
```

PR #1475, opened 2021-06-25
added support for system node pools in managedClusters

### LINKED ISSUE #1376: [AKS] The default MachinePool should be of "system" type, not "user"
/kind bug

I create an AKS cluster with 2 machine pools.

The default machine pool should be a "system" node pool.

The node pool type should be a property of `AzureManagedMachinePool`, in order to allow its configuration.

**Environment:**
- cluster-api-provider-azure version: `v0.4.14`
- Kubernetes version: (use `kubectl version`): `v1.20.5`
- OS (e.g. from `/etc/os-release`): `Ubuntu 18.04.2 LTS`

LochanRn:
**What type of PR is this?**
/kind bug

system node pool will be of type system and not user.
added mode in the azuremannagedmachinepool api to set the mode of azure agent pools
removed defaultPoolRef from azuremanagedcontrolplane api. (provides more flexibility to manage system node pools.) 
validation for deletion of last system node pool is handled in the webhook.
bumped up the containerservice sdk from 2020-02-01 to 2021-03-01

Fixes #1376 #1416 

**TODOs**:

- [X] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Mode spec in the AzureManagedMachinePool is used to specify the mode of an agentPool i.e System or User.
Removed defaultPoolRef from AzureManagedControlPlane. 
```


config/crd/bases/infrastructure.cluster.x-k8s.io_azuremanagedmachinepools.yaml
```diff
@@ -42,6 +42,13 @@ spec:
             description: AzureManagedMachinePoolSpec defines the desired state of
               AzureManagedMachinePool.
             properties:
+              mode:
+                description: 'Mode - represents mode of an agent pool. Possible values
+                  include: ''System'', ''User'''
```
alexeldeib, line 47: nit: the quotes came out a bit weird here. maybe just drop the single quotes around the values themselves? or can we escape them better?

exp/api/v1alpha4/azuremanagedmachinepool_webhook.go
```diff
@@ -0,0 +1,179 @@
...
+		// Prevent OSDiskSizeGB modification if it was already set to some value
+		if r.Spec.OSDiskSizeGB == nil {
+			// unsetting the field is not allowed
+			allErrs = append(allErrs,
+				field.Invalid(
+					field.NewPath("Spec", "OSDiskSizeGB"),
+					r.Spec.OSDiskSizeGB,
+					"field is immutable, unsetting is not allowed"))
+		} else if *r.Spec.OSDiskSizeGB != *old.Spec.OSDiskSizeGB {
+			// changing the field is not allowed
```
alexeldeib, line 101: I think os disk size is mutable as long as you use managed disks, but I could be mistaken.

exp/api/v1alpha4/azuremanagedmachinepool_webhook.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+
+	if !ownerCluster.DeletionTimestamp.IsZero() {
+		return nil
+	}
+
+	opt1 := client.InNamespace(r.Namespace)
+	opt2 := client.MatchingLabels(map[string]string{
+		clusterv1.ClusterLabelName: clusterName,
+		LabelAgentPoolMode:         SystemNodePool,
```
alexeldeib, line 167: This is clever, but I'm curious how it works for a cluster in transition? Might need to add the label inside the reconciler for ammp created before this change went in to ensure they get it applied.

exp/controllers/helpers.go
```diff
@@ -408,11 +408,28 @@ func MachinePoolToAzureManagedControlPlaneMapFunc(ctx context.Context, c client.
...
+						Namespace: ref.Namespace,
+						Name:      ref.Name,
+					},
+				},
+			}
+		}
+
+		isSystemNodePool := ammp.Spec.Mode == infrav1exp.SystemNodePool
+
+		if groupMatches && kindMatches && isSystemNodePool {
```
alexeldeib, line 432: hmm...I'm not 100% sure I follow the thinking here. when this was the default pool, we reconciled it as part of the managedcluster itself (actually the managed service still tries to do this for all pools, maybe a TODO there). we'd actually double reconcile it I think, once as part of MC and once as part of AP.

if we modified the last system pool however, we don't want to reconcile the cluster, we want to reconcile just that pool. if we accidentally race and try to delete the last system pool before creating a new one, just let the retry happen on 400 (return err from reconcile)

we can probably also make the managedclusters client avoid patching agentpools after create, to cleanly separate that

exp/controllers/azuremangedcontrolplane_reconciler.go
```diff
@@ -209,27 +210,62 @@ func (r *azureManagedControlPlaneReconciler) reconcileManagedCluster(ctx context
...
+			}
+
+			ammps = append(ammps, ammp)
+		}
+		if len(ammps) == 0 {
+			return errors.New("owner ref for system machine pools not ready")
+		}
 		// Add to cluster spec
-		managedClusterSpec.AgentPools = []managedclusters.PoolSpec{defaultPoolSpec}
+		managedClusterSpec.AgentPools = ammps
```
alexeldeib, line 268: I know this is what we agreed to in #1376, WDYT about picking a random system pool instead and letting the AMMP reconciler parallelize creation of all the individual pools? I believe this would be faster since we need control plane + 1 pool, and then reconcile all user + system pools in parallel via AKS Agentpool API?

alexeldeib:
overall approach looks good, few small comments. I still need to try this out locally. looks like generated manifests are out of date and e2e need a few updates

config/crd/bases/infrastructure.cluster.x-k8s.io_azuremanagedmachinepools.yaml
LochanRn: ack

exp/api/v1alpha4/azuremanagedmachinepool_webhook.go
LochanRn: will check and get back on this.

exp/controllers/azuremangedcontrolplane_reconciler.go
LochanRn: correct that was the thought process.

exp/api/v1alpha4/azuremanagedmachinepool_webhook.go
LochanRn: I am not very clear what the doubt is here. 

The label is emeded into the ammp in the mutating webhook so i guess that should suffice and not require us to add labels in the reconciler.

alexeldeib: AMMP created before this commit and not updated after will not hit the webhook. They will be reconciled. For the webhook to work correctly, they all need to have the label. So it likely should be added inside reconciler.

alexeldeib: I guess there were no system pools before this commit, so that’s not really possible 😅 it’s fine as is, you are right

exp/controllers/helpers.go
LochanRn: MachinePoolToAzureManagedControlPlaneMapFunc was added by you in #1397 
This is the description, reasoning the PR has.
`If the MachinePool owning the defaultPool of an AzureManagedControlPlane is nil when we try to reconcile, we need to re-reconcile. We can either return an error or watch machinePools and filter them down to the one owning the defaultPool (more accurate, but requires watching and filtering all machinePool events). I implemented the latter.
`

So this is used in AMCP and not in MC.  

I found your reasoning fine hence i thought of working on the same lines for system node pool, i.e if the owner mp of system node pool is nil we try to reconcile as it is more accurate. 

But I guess I made a mistake in line 265 in azuremanagedcontrolplane_reconciler I should return a nil over there.

exp/controllers/azuremangedcontrolplane_reconciler.go
```diff
@@ -209,27 +210,62 @@ func (r *azureManagedControlPlaneReconciler) reconcileManagedCluster(ctx context
...
+			}
+
+			if ownerPool.Spec.Replicas != nil {
+				ammp.Replicas = *ownerPool.Spec.Replicas
+			}
+
+			ammps = append(ammps, ammp)
+		}
+		if len(ammps) == 0 {
+			return errors.New("owner ref for system machine pools not ready")
```
LochanRn, line 265: should return nil instead of error.

exp/api/v1alpha4/azuremanagedmachinepool_webhook.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+
+	ownerCluster := &clusterv1.Cluster{}
+	key := client.ObjectKey{
+		Namespace: r.Namespace,
+		Name:      clusterName,
+	}
+
+	if err := webhookClient.Get(ctx, key, ownerCluster); err != nil {
+		if !azure.ResourceNotFound(err) {
```
LochanRn, line 154: this should be azure.ResourceNotFound(err)

exp/controllers/helpers.go
alexeldeib: Yeah, I see. This makes sense, but then we should definitely make sure AMCP reconciler doesn't send all the AMMPs to AKS on updates, only update the control plane config. We don't want AMCP and AMMP reconcilers to compete to reconcile node pools *except* that we let AMCP create the very first one as part of the cluster.

LochanRn: yeah makes sense, i think we have a check for this in line 216 azuremanagedcontrolplane.go. In amcp we reconcile ammp only during creation.

CecileRobertMichon:
@LochanRn make sure you're rebasing and squashing commits instead of adding merge commits  to keep the main branch git history clean

CecileRobertMichon:
You can follow the instructions at https://www.kubernetes.dev/docs/guide/github-workflow/#squash-commits to rebase and squash

test/e2e/config/azure-dev.yaml
```diff
@@ -115,7 +115,7 @@ variables:
 intervals:
   default/wait-controllers: ["3m", "10s"]
   default/wait-cluster: ["20m", "10s"]
-  default/wait-control-plane: ["20m", "10s"]
+  default/wait-control-plane: ["35m", "10s"]
```
CecileRobertMichon, line 118: I don't think we want to change this value for every test, was it to address #1481 ? if so we should fix it separately and/or add a specific AKS timeout interval

LochanRn: Not sure if this is the only problem. But I have noticed several times for an AKS cluster to come up fully it takes a max of 30m I kept 5m buffer.

LochanRn:
> You can follow the instructions at https://www.kubernetes.dev/docs/guide/github-workflow/#squash-commits to rebase and squash

Thanks

test/e2e/config/azure-dev.yaml
alexeldeib: were there any other changes to make e2e pass? I noticed deletion and the full test sometimes takes more than 20min, but the cluster initialization should really take less time. it's fine if we want to debug it elsewhere, but let's be cognizant of that.

I'd probably prefer adding a separate interval for AKS if we need to increase it so high, like cecile mentioned.

azure/services/agentpools/agentpools.go
```diff
@@ -19,7 +19,7 @@ package agentpools
 import (
 	"context"
 
-	"github.com/Azure/azure-sdk-for-go/services/containerservice/mgmt/2020-02-01/containerservice"
+	"github.com/Azure/azure-sdk-for-go/services/containerservice/mgmt/2021-03-01/containerservice"
```
alexeldeib, line 22: any chance I can persuade you to upgrade to `2021-05-01` 😃 the diff should be basically nothing. if there's any non-trivial diff, you can leave it at `2021-03-01`.

docs/book/src/topics/managedcluster.md
```diff
@@ -12,9 +12,8 @@ custom resources:
 The combination of AzureManagedControlPlane/AzureManagedCluster
 corresponds to provisioning an AKS cluster. AzureManagedMachinePool
 corresponds one-to-one with AKS node pools. This also means that
-creating an AzureManagedControlPlane requires defining the default
-machine pool, since AKS requires at least one system pool at creation
-time.
+creating atleast one AzureManagedMachinePool with Spec.Mode System,
```
alexeldeib, line 15:
```suggestion
creating at least one AzureManagedMachinePool with Spec.Mode System,
```

exp/api/v1alpha3/azuremanagedmachinepool_types.go
```diff
@@ -23,6 +23,10 @@ import (
 
 // AzureManagedMachinePoolSpec defines the desired state of AzureManagedMachinePool.
 type AzureManagedMachinePoolSpec struct {
+	// Mode - represents mode of an agent pool. Possible values include: System, User.
+	// +kubebuilder:validation:Enum=System;User
+	Mode string `json:"mode"`
```
alexeldeib, line 28: should we default this? or require it? i'm fine as-is but thought I'd raise it

exp/api/v1alpha4/azuremanagedmachinepool_types.go
```diff
@@ -21,8 +21,21 @@ import (
 	capierrors "sigs.k8s.io/cluster-api/errors"
 )
 
+const (
+	// LabelAgentPoolMode represents mode of an agent pool. Possible values include: System, User.
+	LabelAgentPoolMode = "azurecluster.infrastructure.cluster.x-k8s.io/agentpoolmode"
```
alexeldeib, line 26: "azurecluster" ? maybe "azuremanagedmachinepool"? or at least azuremanagedcontrolplane?

exp/api/v1alpha4/azuremanagedmachinepool_types.go
```diff
@@ -21,8 +21,21 @@ import (
...
+
+	// SystemNodePool represents mode system for azuremachinepool.
+	SystemNodePool = "System"
```
alexeldeib, line 29: nit: you can probably just use the SDK values AgentPoolMode{System,User}, or else I'd prefer naming slightly closer to that (SystemNodePool is a bit too generic imo and could easily be a var name used elsewhere)

exp/controllers/azuremanagedcontrolplane_controller.go
```diff
@@ -225,13 +179,11 @@ func (r *AzureManagedControlPlaneReconciler) Reconcile(ctx context.Context, req
 
 	// Create the scope.
 	mcpScope, err := scope.NewManagedControlPlaneScope(ctx, scope.ManagedControlPlaneScopeParams{
-		Client:           r.Client,
-		Logger:           log,
-		Cluster:          cluster,
-		ControlPlane:     azureControlPlane,
-		MachinePool:      ownerPool,
-		InfraMachinePool: defaultPool,
```
alexeldeib, line 233: I think we can remove InfraMachinePool from ManagedControlPlaneScope now?

alexeldeib: you've done it here, but I think it can be removed from the struct definitions as well

exp/api/v1alpha4/azuremanagedmachinepool_types.go
alexeldeib: actually I think keeping the label definition here is better..

maybe Label{System,User}NodePool? not super picky on this one tbh

azure/services/agentpools/agentpools.go
LochanRn: ack 👍

docs/book/src/topics/managedcluster.md
LochanRn: ack 👍

exp/api/v1alpha4/azuremanagedmachinepool_types.go
LochanRn: Will change it to azuremanagedmachinepool.

exp/api/v1alpha3/azuremanagedmachinepool_types.go
LochanRn: I am okay with anything here I think. I guess the only difference would be that the user gets more knowledgeable about the mode of the machinepool if kept as required 😅.

exp/api/v1alpha4/azuremanagedmachinepool_types.go
LochanRn: ack 👍

exp/controllers/azuremanagedcontrolplane_controller.go
LochanRn: I did it remove it initially, but then had to add it back as the ammp controller also uses ManagedControlPlaneScope. The InfraMachinePool is required here.

exp/api/v1alpha3/azuremanagedmachinepool_types.go
LochanRn: I am okay with anything

LochanRn: Have we decided on this ? Should I add a default instead of required ?

alexeldeib: Changing from required => optional is non-breaking, let’s leave it required for now 👍

exp/api/v1alpha4/azuremanagedmachinepool_types.go
LochanRn: Will this be fine ? 

```
type NodePool string
const (
	// AgentPoolModeSystem represents mode system for azuremachinepool.
	AgentPoolModeSystem NodePool = "System"
)
```

test/e2e/config/azure-dev.yaml
alexeldeib: Nvm, answered my own question in the other issue/pr

alexeldeib: let’s revert the change here, I’ll fix the timeout

exp/api/v1alpha4/azuremanagedmachinepool_types.go
alexeldeib: I’d probably 

```go
type NodePoolMode string
const (
  NodePoolModeSystem …
  NodePoolModeUser…
)
```

or something to align the names a bit more closely

azure/services/agentpools/agentpools.go
LochanRn: The one minor difference while moving it from 2021-03-01 - 2021-05-01 is 

[ListClusterAdminCredentials](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/15da9c80de36e90f3f564e9c75e928c8003b7614/azure/services/managedclusters/client.go#L69) api has an extra param serverFqdn in 2021-05-01. I have kept it to empty string.

test/e2e/config/azure-dev.yaml
LochanRn: No, i did not make any other changes.  Yes it gets initialised quickly but for the first managedmachinepool or first system managed machine pool to come to a **ready** state it takes time. Like about 25-30 mins approx. I have not exactly timed the deletion but from the boot to until the first system pool comes up it would be approx 30 mins.

alexeldeib: I debugged this extensively in https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1488, there's a lag in Azure List VMSS which causes AMMP reconciliation to take way longer than expected. Both pools and the cluster should generally be done in ~10 minutes. I have a workaround and I'm also chasing it down internally with the owning team.

There's also a small issue with watches where if we reconcile AMMP with a stale AMCP in the cache, we won't requeue the 2nd machinepool for a long time (because our mapper only maps to system pools).

LochanRn:
unable to rebase correctly, closing this PR. Will open a fresh PR.

PR #1478, opened 2021-06-29
Add support for ultra disk for data disks

### LINKED ISSUE #1134: Add support for Ultra Disks for data disks
/kind feature

This depends on #964 

See also https://docs.microsoft.com/en-us/azure/virtual-machines/disks-types and https://github.com/Azure/aks-engine/pull/3872.

Azure ultra disks deliver high throughput, high IOPS, and consistent low latency disk storage for Azure IaaS VMs. Validation needs to take all limitations into account https://docs.microsoft.com/en-us/azure/virtual-machines/disks-types#ga-scope-and-limitations.

**Environment:**

Ankitasw:
**What type of PR is this?**
/kind feature

Add support for Ultra disk storage type for data disks.

Fixes #1134

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:
```release-note
Add option for ultra disks for data disks.
```

### DIFF
--- a/azure/services/resourceskus/sku.go
+++ b/azure/services/resourceskus/sku.go
@@ -66,6 +66,8 @@ const (
 	EncryptionAtHost = "EncryptionAtHostSupported"
 	// MaximumPlatformFaultDomainCount identifies the maximum fault domain count for an availability set in a region.
 	MaximumPlatformFaultDomainCount = "MaximumPlatformFaultDomainCount"
+	// UltraSSDAvailable identifies the capability for the support of UltraSSD data disks.
+	UltraSSDAvailable = "UltraSSDAvailable"
 )
 
 // HasCapability return true for a capability which can be either
@@ -127,4 +129,39 @@ func (s SKU) GetCapability(name string) (string, bool) {
 		}
 	}
 	return "", false
+}
+
+// HasLocationCapability returns true if the provided resource supports the location capability.
+func (s SKU) HasLocationCapability(capabilityName, location, zone string) bool {
+	if s.LocationInfo == nil {
+		return false
+	}
+
+	for _, info := range *s.LocationInfo {
+		if info.Location == nil || *info.Location != location || info.ZoneDetails == nil {
+			continue
+		}
+
+		for _, zoneDetail := range *info.ZoneDetails {
+			if zoneDetail.Capabilities == nil {
+				continue
+			}
+
+			for _, capability := range *zoneDetail.Capabilities {
+				if capability.Name != nil && *capability.Name == capabilityName {
+					if zoneDetail.Name == nil {
+						return false
+					}
+
+					for _, name := range *zoneDetail.Name {
+						if name == zone {
+							return true
+						}
+					}
+					return false
+				}
+			}
+		}
+	}
+	return false
 }
--- a/azure/services/scalesets/scalesets.go
+++ b/azure/services/scalesets/scalesets.go
@@ -37,6 +37,11 @@ import (
 	"sigs.k8s.io/cluster-api-provider-azure/util/tele"
 )
 
+const (
+	// UltraSSDStorageAccountType identifies the Ultra disk storage account type.
+	UltraSSDStorageAccountType = "UltraSSD_LRS"
+)
+
 type (
 	// ScaleSetScope defines the scope interface for a scale sets service.
 	ScaleSetScope interface {
@@ -314,6 +319,21 @@ func (s *Service) validateSpec(ctx context.Context) error {
 		return azure.WithTerminalError(errors.Errorf("encryption at host is not supported for VM type %s", spec.Size))
 	}
 
+	// check the support for ultra disks based on location and vm size
+	for _, disks := range spec.DataDisks {
+		location := s.Scope.Location()
+		zones, err := s.resourceSKUCache.GetZones(ctx, location)
+		if err != nil {
+			return azure.WithTerminalError(errors.Wrapf(err, "failed to get the zones for location %s", location))
+		}
+
+		for _, zone := range zones {
+			if disks.ManagedDisk != nil && disks.ManagedDisk.StorageAccountType == UltraSSDStorageAccountType && !sku.HasLocationCapability(resourceskus.UltraSSDAvailable, location, zone) {
+				return azure.WithTerminalError(fmt.Errorf("vm size %s does not support ultra disks in location %s. select a different vm size or disable ultra disks", spec.Size, location))
+			}
+		}
+	}
+
 	// Checking if selected availability zones are available selected VM type in location
 	azsInLocation, err := s.resourceSKUCache.GetZonesWithVMSize(ctx, spec.Size, s.Scope.Location())
 	if err != nil {
@@ -449,6 +469,14 @@ func (s *Service) buildVMSSFromSpec(ctx context.Context, vmssSpec azure.ScaleSet
 		vmss.Identity = &compute.VirtualMachineScaleSetIdentity{
 			Type:                   compute.ResourceIdentityTypeUserAssigned,
 			UserAssignedIdentities: userIdentitiesMap,
+		}
+	}
+
+	for _, dataDisk := range vmssSpec.DataDisks {
+		if dataDisk.ManagedDisk != nil && dataDisk.ManagedDisk.StorageAccountType == UltraSSDStorageAccountType {
+			vmss.VirtualMachineScaleSetProperties.AdditionalCapabilities = &compute.AdditionalCapabilities{
+				UltraSSDEnabled: to.BoolPtr(true),
+			}
 		}
 	}
 
--- a/azure/services/virtualmachines/virtualmachines.go
+++ b/azure/services/virtualmachines/virtualmachines.go
@@ -39,6 +39,11 @@ import (
 	"sigs.k8s.io/cluster-api-provider-azure/util/tele"
 )
 
+const (
+	// UltraSSDStorageAccountType identifies the Ultra disk storage account type.
+	UltraSSDStorageAccountType = "UltraSSD_LRS"
+)
+
 // VMScope defines the scope interface for a virtual machines service.
 type VMScope interface {
 	logr.Logger
@@ -190,6 +195,14 @@ func (s *Service) Reconcile(ctx context.Context) error {
 			virtualMachine.Identity = &compute.VirtualMachineIdentity{
 				Type:                   compute.ResourceIdentityTypeUserAssigned,
 				UserAssignedIdentities: userIdentitiesMap,
+			}
+		}
+
+		for _, dataDisk := range vmSpec.DataDisks {
+			if dataDisk.ManagedDisk != nil && dataDisk.ManagedDisk.StorageAccountType == UltraSSDStorageAccountType {
+				virtualMachine.VirtualMachineProperties.AdditionalCapabilities = &compute.AdditionalCapabilities{
+					UltraSSDEnabled: to.BoolPtr(true),
+				}
 			}
 		}
 
@@ -419,6 +432,12 @@ func (s *Service) generateStorageProfile(ctx context.Context, vmSpec azure.VMSpe
 
 			if disk.ManagedDisk.DiskEncryptionSet != nil {
 				dataDisks[i].ManagedDisk.DiskEncryptionSet = &compute.DiskEncryptionSetParameters{ID: to.StringPtr(disk.ManagedDisk.DiskEncryptionSet.ID)}
+			}
+
+			// check the support for ultra disks based on location and vm size
+			location := s.Scope.Location()
+			if disk.ManagedDisk.StorageAccountType == UltraSSDStorageAccountType && !sku.HasLocationCapability(resourceskus.UltraSSDAvailable, location, vmSpec.Zone) {
+				return nil, azure.WithTerminalError(fmt.Errorf("vm size %s does not support ultra disks in location %s. select a different vm size or disable ultra disks", vmSpec.Size, location))
 			}
 		}
 	}
--- a/azure/services/scalesets/scalesets_test.go
+++ b/azure/services/scalesets/scalesets_test.go
@@ -21,6 +21,8 @@ import (
 	"net/http"
 	"testing"
 
+	"k8s.io/utils/pointer"
+
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2020-06-30/compute"
 	"github.com/Azure/go-autorest/autorest"
 	"github.com/Azure/go-autorest/autorest/to"
@@ -248,12 +250,21 @@ func TestReconcileVMSS(t *testing.T) {
 			expectedError: "failed to get VMSS my-vmss after create or update: failed to get result from future: operation type PUT on Azure resource my-rg/my-vmss is not done",
 			expect: func(g *WithT, s *mock_scalesets.MockScaleSetScopeMockRecorder, m *mock_scalesets.MockClientMockRecorder) {
 				defaultSpec := newDefaultVMSSSpec()
+				defaultSpec.DataDisks = append(defaultSpec.DataDisks, infrav1.DataDisk{
+					NameSuffix: "my_disk_with_ultra_disks",
+					DiskSizeGB: 128,
+					Lun:        to.Int32Ptr(3),
+					ManagedDisk: &infrav1.ManagedDiskParameters{
+						StorageAccountType: "UltraSSD_LRS",
+					},
+				})
 				s.ScaleSetSpec().Return(defaultSpec).AnyTimes()
 				setupDefaultVMSSStartCreatingExpectations(s, m)
-				vmss := newDefaultVMSS()
+				vmss := newDefaultVMSS("VM_SIZE")
+				vmss.VirtualMachineScaleSetProperties.AdditionalCapabilities = &compute.AdditionalCapabilities{UltraSSDEnabled: pointer.Bool(true)}
 				m.CreateOrUpdateAsync(gomockinternal.AContext(), defaultResourceGroup, defaultVMSSName, gomockinternal.DiffEq(vmss)).
 					Return(putFuture, nil)
-				setupCreatingSucceededExpectations(s, m, newDefaultExistingVMSS(), putFuture)
+				setupCreatingSucceededExpectations(s, m, newDefaultExistingVMSS("VM_SIZE"), putFuture)
 			},
 		},
 		{
@@ -262,7 +273,7 @@ func TestReconcileVMSS(t *testing.T) {
 			expect: func(g *WithT, s *mock_scalesets.MockScaleSetScopeMockRecorder, m *mock_scalesets.MockClientMockRecorder) {
 				defaultSpec := newDefaultVMSSSpec()
 				s.ScaleSetSpec().Return(defaultSpec).AnyTimes()
-				createdVMSS := newDefaultVMSS()
+				createdVMSS := newDefaultVMSS("VM_SIZE")
 				instances := newDefaultInstances()
 				_ = setupDefaultVMSSInProgressOperationDoneExpectations(s, m, createdVMSS, instances)
 				s.SetLongRunningOperationState(nil)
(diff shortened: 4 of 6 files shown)


nader-ziada:
thanks for the PR @Ankitasw 

we might also want to add validation on the machine to make sure the user can select the ultra disk here:  https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/api/v1alpha4/types.go#L410

azure/services/virtualmachines/virtualmachines.go
```diff
@@ -393,6 +403,10 @@ func (s *Service) generateStorageProfile(ctx context.Context, vmSpec azure.VMSpe
 		}
 	}
 
+	if vmSpec.UltraSSDEnabled {
+		sku.HasLocationCapabilities("UltraSSDAvailable")
```
CecileRobertMichon, line 407: this is returning a bool, how is the output being used?

CecileRobertMichon (commented):
/ok-to-test

azure/services/virtualmachines/virtualmachines.go
Ankitasw: added the validation now

azure/services/virtualmachines/virtualmachines.go
```diff
@@ -421,6 +431,11 @@ func (s *Service) generateStorageProfile(ctx context.Context, vmSpec azure.VMSpe
 				dataDisks[i].ManagedDisk.DiskEncryptionSet = &compute.DiskEncryptionSetParameters{ID: to.StringPtr(disk.ManagedDisk.DiskEncryptionSet.ID)}
 			}
 		}
+		// check the support for ultra disks based on location and vm size
+		location := s.Scope.Location()
+		if disk.UltraSSDEnabled && !sku.HasLocationCapabilities("UltraSSDAvailable", vmSpec.Size, location) {
+			return nil, azure.WithTerminalError(fmt.Errorf("vm size %s does not support ultra disks. select a different vm size or disable ultra disks", vmSpec.Size))
```
CecileRobertMichon, line 437:
```suggestion
			return nil, azure.WithTerminalError(fmt.Errorf("vm size %s does not support ultra disks in %s. select a different vm size or disable ultra disks", vmSpec.Size, location))
```

azure/services/virtualmachines/virtualmachines.go
```diff
@@ -421,6 +431,11 @@ func (s *Service) generateStorageProfile(ctx context.Context, vmSpec azure.VMSpe
 				dataDisks[i].ManagedDisk.DiskEncryptionSet = &compute.DiskEncryptionSetParameters{ID: to.StringPtr(disk.ManagedDisk.DiskEncryptionSet.ID)}
 			}
 		}
+		// check the support for ultra disks based on location and vm size
+		location := s.Scope.Location()
+		if disk.UltraSSDEnabled && !sku.HasLocationCapabilities("UltraSSDAvailable", vmSpec.Size, location) {
```
nader-ziada, line 436: can make this a const

nader-ziada (commented):
you also need to do change in scaleset [create](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/azure/services/scalesets/scalesets.go#L380) similar to virtualmachine

azure/services/virtualmachines/virtualmachines.go
Ankitasw: you mean "UltraSSDAvailable"?

nader-ziada: yes

nader-ziada:
lgtm, but want one of @CecileRobertMichon or @devigned to take a look as well

azure/services/resourceskus/sku.go
```diff
@@ -128,3 +130,23 @@ func (s SKU) GetCapability(name string) (string, bool) {
 	}
 	return "", false
 }
+
+// HasLocationCapabilities returns true if the provided resource supports the location capability.
+func (s SKU) HasLocationCapabilities(name, size, location string) bool {
```
mboersma, line 135: Maybe this should be `HasLocationCapability()` for consistency with the related functions? It's also more descriptive since it only checks for one capability.

Ankitasw: sounds fair

mboersma (approved):
lgtm

CecileRobertMichon:
/approve

Thanks @Ankitasw 🎉

api/v1alpha3/azuremachine_conversion.go
```diff
@@ -45,6 +45,9 @@ func (src *AzureMachine) ConvertTo(dstRaw conversion.Hub) error { // nolint
 		}
 	}
 
+	if restored.Spec.DataDisks != nil {
+		dst.Spec.DataDisks = restored.Spec.DataDisks
```
CecileRobertMichon, line 49: sorry I just saw this, I think this would override any changes that were made to the Disk between the last conversion and now, we only want to restore UltraSSD disks . For example, let's say you create a data disk with another storage type, convert the Machine to v1alpha3, then delete the disk, then convert it back to v1alpha4. This line would re-add the deleted disk, which is incorrect. We only want to restore fields that were not supported in v1alpha3 when converting to v1alpha4.

CecileRobertMichon: here's an example you can use: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/api/v1alpha3/azurecluster_conversion.go#L66

CecileRobertMichon: this is the commit for reference https://github.com/kubernetes-sigs/cluster-api-provider-azure/commit/715d5fd7363c92713086a0304152a3c3612315d8#diff-513d36576b5dbcc8430d6b8a26b0f7cccda060c2d932cc1a0d38f7dc16a9e03e

api/v1alpha4/types.go
```diff
@@ -420,6 +420,8 @@ type DataDisk struct {
 	// NameSuffix is the suffix to be appended to the machine name to generate the disk name.
 	// Each disk name will be in format <machineName>_<nameSuffix>.
 	NameSuffix string `json:"nameSuffix"`
+	// UltraSSDEnabled is set to true if Ultra disk is enabled for data disks
+	UltraSSDEnabled bool `json:"ultraSSDEnabled,omitempty"`
```
CecileRobertMichon, line 424: now that I'm looking at this more closely, is there a reason we need this extra API field? why not use ManagedDisk. StorageAccountType? What's the expected value for a user to add a disk with ultra disk for StorageAccountType right now?

Ankitasw: sounds quiet right, didn't think of using StorageAccountType to confirm if ultra disk is enabled, we do not need this field.

CecileRobertMichon: >  is it possible that ManagedDiskParameters is left empty as it is optional? In that case, would we miss the info about ultra ssd?

if ManagedDiskParameters is empty we just use the Azure defaulted storage type as before. Ultra ssd would be disabled in that case since it's an opt in feature. Does that make sense?

Ankitasw: sounds fair, will remove the extra field `UltraSSDEnabled`  added and will make use of `StorageAccountType`

azure/services/resourceskus/sku.go
```diff
@@ -128,3 +130,23 @@ func (s SKU) GetCapability(name string) (string, bool) {
 	}
 	return "", false
 }
+
+// HasLocationCapability returns true if the provided resource supports the location capability.
+func (s SKU) HasLocationCapability(name, size, location string) bool {
+	if s.LocationInfo != nil && *s.Name == size {
```
alexeldeib, line 136: the check on sku.name shouldn't be necessary, since we fetch it from the cache by name already

azure/services/resourceskus/sku.go
```diff
@@ -128,3 +130,23 @@ func (s SKU) GetCapability(name string) (string, bool) {
...
+		for _, info := range *s.LocationInfo {
+			if info.ZoneDetails != nil && *info.Location == location {
+				for _, zone := range *info.ZoneDetails {
+					if zone.Capabilities != nil {
+						for _, capability := range *zone.Capabilities {
```
alexeldeib, line 141: this capability can be exposed in a subset of the zones in a region, should we check against the VM zone here? example:

```json
"locationInfo": [
      {
        "location": "southcentralus",
        "zoneDetails": [
          {
            "Name": [
              "3"
            ],
            "capabilities": [
              {
                "name": "UltraSSDAvailable",
                "value": "True"
              }
            ],
            "name": null
          }
        ],
        "zones": [
          "1",
          "2",
          "3"
        ]
      }
    ],
```

azure/services/resourceskus/sku.go
```diff
@@ -128,3 +130,23 @@ func (s SKU) GetCapability(name string) (string, bool) {
...
+							if *capability.Name == name {
```
alexeldeib, line 142: very technically (I've never seen it to be the case), `*capability.Value` can be false. see other funcs using it.

Ankitasw: sure we can add the check for Value as true

azure/services/resourceskus/sku.go
Ankitasw: i didn't quiet understand the check you are referring here, zoneDetails already contains the name of the zone in which the UltraSSD is available, so what we need to check ?

alexeldeib: where are we checking against the zone name? we are checking against location only as far as I see?

azure/services/resourceskus/sku.go
```diff
@@ -128,3 +130,23 @@ func (s SKU) GetCapability(name string) (string, bool) {
...
+								return true
+							}
+						}
+					}
+				}
+			}
+		}
+	}
+	return false
+}
```
alexeldeib, line 152:
```suggestion
func (s SKU) HasLocationCapability(name, size, location, zone string) bool {
	if s.LocationInfo != nil && *s.Name == size {
		for _, info := range *s.LocationInfo {
			if info.ZoneDetails != nil && *info.Location == location {
				for _, zone := range *info.ZoneDetails {
					if zone.Capabilities != nil {
						for _, capability := range *zone.Capabilities {
							if *capability.Name == name {
							        if zone.Names != nil {
							                for _, name := zone.Names {
							                        if name == zone {
							                                return true
							                        }
							                }
							        }
								return false
							}
						}
					}
				}
			}
		}
	}
	return false
}
```

alexeldeib: the diff is the func signature (added zone) and the inner loop on zone.Names, in case that code suggestion is hard to read

azure/services/resourceskus/sku.go
alexeldeib: we need to check zoneDetails.Names contains the zone we are deploying to

azure/services/resourceskus/sku.go
alexeldeib: modified a bit to reduce nesting, but same logic

Ankitasw: how can we accommodate zone in case of scaleset? i think there we have zone list, so for each zone in the list we have to check the capability?

azure/services/resourceskus/sku.go
Ankitasw: @alexeldeib please have look at the suggestion commit

azure/services/resourceskus/sku.go
alexeldeib: ah yeah, this is a bit tricky. what you said is the safest for sure, but does restrict some cases where the user would technically be able to deploy the VM. however, if they asked for ultrassd and didn't get it e.g. defaulted on, I think it's reasonable to fail the operation, since we can't fulfill their expectation that they will actually be able to get ultrassd.

@CecileRobertMichon or @jackfrancis wdyt?

jackfrancis: Makes sense to me

azure/services/scalesets/scalesets.go
```diff
@@ -452,6 +468,14 @@ func (s *Service) buildVMSSFromSpec(ctx context.Context, vmssSpec azure.ScaleSet
 		}
 	}
 
+	for _, dataDisk := range vmssSpec.DataDisks {
+		if dataDisk.ManagedDisk != nil && dataDisk.ManagedDisk.StorageAccountType == "UltraSSD_LRS" {
```
alexeldeib, line 472: nit: prefer a const

azure/services/virtualmachines/virtualmachines.go
```diff
@@ -193,6 +195,14 @@ func (s *Service) Reconcile(ctx context.Context) error {
 			}
 		}
 
+		for _, dataDisk := range vmSpec.DataDisks {
+			if dataDisk.ManagedDisk != nil && dataDisk.ManagedDisk.StorageAccountType == "UltraSSD_LRS" {
```
alexeldeib, line 199: nit: const

azure/services/virtualmachines/virtualmachines.go
```diff
@@ -420,6 +430,12 @@ func (s *Service) generateStorageProfile(ctx context.Context, vmSpec azure.VMSpe
 			if disk.ManagedDisk.DiskEncryptionSet != nil {
 				dataDisks[i].ManagedDisk.DiskEncryptionSet = &compute.DiskEncryptionSetParameters{ID: to.StringPtr(disk.ManagedDisk.DiskEncryptionSet.ID)}
 			}
+
+			// check the support for ultra disks based on location and vm size
+			location := s.Scope.Location()
+			if disk.ManagedDisk.StorageAccountType == "UltraSSD_LRS" && !sku.HasLocationCapability(resourceskus.UltraSSDAvailable, location, vmSpec.Zone) {
```
alexeldeib, line 436: same

alexeldeib:
https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1478#discussion_r666305387 is still relevant, even though the diff is marked as outdated BTW

api/v1alpha3/azuremachinetemplate_conversion.go
```diff
@@ -51,6 +51,10 @@ func (src *AzureMachineTemplate) ConvertTo(dstRaw conversion.Hub) error { // nol
 		dst.Spec.Template.Spec.Image.SharedGallery.SKU = restored.Spec.Template.Spec.Image.SharedGallery.SKU
 	}
 
+	if restored.Spec.Template.Spec.DataDisks != nil {
```
alexeldeib, line 54: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1478#discussion_r666305387

exp/api/v1alpha3/conversion.go
```diff
@@ -52,4 +52,4 @@ func Convert_v1alpha3_APIEndpoint_To_v1alpha4_APIEndpoint(in *clusterapiapiv1alp
 // Convert_v1alpha4_APIEndpoint_To_v1alpha3_APIEndpoint is an autogenerated conversion function.
 func Convert_v1alpha4_APIEndpoint_To_v1alpha3_APIEndpoint(in *clusterapiapiv1alpha4.APIEndpoint, out *clusterapiapiv1alpha3.APIEndpoint, s conversion.Scope) error {
 	return clusterapiapiv1alpha3.Convert_v1alpha4_APIEndpoint_To_v1alpha3_APIEndpoint(in, out, s)
-}
+}
```
devigned, line 55: nit; newline, please

azure/services/virtualmachines/virtualmachines.go
```diff
@@ -193,6 +200,14 @@ func (s *Service) Reconcile(ctx context.Context) error {
 			}
 		}
 
+		for _, dataDisk := range vmSpec.DataDisks {
+			if dataDisk.ManagedDisk != nil && dataDisk.ManagedDisk.StorageAccountType == UltraSSDStorageAccountType {
+				virtualMachine.VirtualMachineProperties.AdditionalCapabilities = &compute.AdditionalCapabilities{
+					UltraSSDEnabled: pointer.Bool(true),
```
devigned, line 206: The rest of the file uses `to.*`
```suggestion
					UltraSSDEnabled: to.BoolPtr(true),
```

azure/services/scalesets/scalesets.go
```diff
@@ -452,6 +473,14 @@ func (s *Service) buildVMSSFromSpec(ctx context.Context, vmssSpec azure.ScaleSet
 		}
 	}
 
+	for _, dataDisk := range vmssSpec.DataDisks {
+		if dataDisk.ManagedDisk != nil && dataDisk.ManagedDisk.StorageAccountType == UltraSSDStorageAccountType {
+			vmss.VirtualMachineScaleSetProperties.AdditionalCapabilities = &compute.AdditionalCapabilities{
+				UltraSSDEnabled: pointer.Bool(true),
```
devigned, line 479:
```suggestion
				UltraSSDEnabled: to.BoolPtr(true),
```

azure/services/resourceskus/sku.go
```diff
@@ -128,3 +130,34 @@ func (s SKU) GetCapability(name string) (string, bool) {
 	}
 	return "", false
 }
+
+// HasLocationCapability returns true if the provided resource supports the location capability.
+func (s SKU) HasLocationCapability(capabilityName, location, zone string) bool {
+	if s.LocationInfo == nil {
+		return false
+	}
+	for _, info := range *s.LocationInfo {
```
devigned, line 139: nit; newline between 138 and 139.

azure/services/resourceskus/sku.go
```diff
@@ -128,3 +130,34 @@ func (s SKU) GetCapability(name string) (string, bool) {
...
+		if info.Location == nil || *info.Location != location || info.ZoneDetails == nil {
+			continue
+		}
+		for _, zoneDetail := range *info.ZoneDetails {
```
devigned, line 143: nit; newline between 142 and 143

azure/services/resourceskus/sku.go
```diff
@@ -128,3 +130,34 @@ func (s SKU) GetCapability(name string) (string, bool) {
...
+			if zoneDetail.Capabilities == nil {
+				continue
+			}
+			for _, capability := range *zoneDetail.Capabilities {
```
devigned, line 147: nit; newline between 146 and 147

azure/services/resourceskus/sku.go
```diff
@@ -128,3 +130,34 @@ func (s SKU) GetCapability(name string) (string, bool) {
...
+				if capability.Name != nil && *capability.Name == capabilityName {
+					if zoneDetail.Name == nil {
+						return false
+					}
+					for _, name := range *zoneDetail.Name {
```
devigned, line 152: nit; newline between 151 and 152

azure/services/scalesets/scalesets.go
```diff
@@ -314,6 +321,20 @@ func (s *Service) validateSpec(ctx context.Context) error {
...
 	}
 
+	// check the support for ultra disks based on location and vm size
+	for _, disks := range spec.DataDisks {
+		location := s.Scope.Location()
+		zones, err := s.resourceSKUCache.GetZones(ctx, location)
+		if err != nil {
+			return azure.WithTerminalError(errors.Wrapf(err, "failed to get the zones for location %s", location))
+		}
+		for _, zone := range zones {
```
devigned, line 331: nit newline after conditional

devigned (changes requested):
Mostly nits. Please consider using vertical space between a conditional block and the following statement.

Ankitasw:
@alexeldeib @devigned @CecileRobertMichon please have a look at the latest changes.

Ankitasw:
> @Ankitasw thanks for updating the PR! This looks better. Could you please add a few lines about this feature in https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/docs/book/src/topics/data-disks.md ?

@CecileRobertMichon added the few lines, let me know if we want to specify some more info.

CecileRobertMichon (commented):
I tried it out and was able to build a cluster with an ultra disk data disk on the control plane VM. Thanks for all the hard work and being patient with our reviews!

/lgtm
/approve

PR #1492, opened 2021-07-06
SinglePlacementGroup=false for Azure VMSS (MachinePool)

### LINKED ISSUE #484: Support Multiple Placement Groups in VMSS
/kind feature

Many large customers, when they scale out their applications, run into the 100 VM limit in VMSS. There is a flag in VMSS to support scaling out across multiple Placement Groups to support up to 1000 VMs. I have run into this issue with customers on existing platforms and therefore believe it should go in before we hit it in the field.

[Azure Docs - Working with large virtual machine scale sets](https://docs.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-placement-groups)

ARM Template flag (`"singlePlacementGroup": false`):
```
{
  "type": "Microsoft.Compute/virtualMachineScaleSets",
  "location": "australiaeast",
  "name": "bigvmss",
  "sku": {
    "name": "Standard_DS1_v2",
    "tier": "Standard",
    "capacity": 1000
  },
  "properties": {
    "singlePlacementGroup": false,
    "upgradePolicy": {
      "mode": "Automatic"
    }
```

jackfrancis:
**What type of PR is this?**

/kind feature

This PR changes the AzureMachinePool implementation so that we do not create the underlying VMSS in a SinglePlacementGroup configuration.

A single placement group restricts the size of the VMSS (and thus the AzureMachinePool) to 100, whereas using multiple placement groups allows for a pool size of up to 1000.

Ref: https://docs.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-placement-groups#placement-groups

Fixes #484

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
SinglePlacementGroup=false for Azure VMSS (MachinePool)
```

### DIFF
--- a/azure/converters/vmss_test.go
+++ b/azure/converters/vmss_test.go
@@ -52,7 +52,8 @@ func Test_SDKToVMSS(t *testing.T) {
 						Location: to.StringPtr("westus2"),
 						Tags:     tags,
 						VirtualMachineScaleSetProperties: &compute.VirtualMachineScaleSetProperties{
-							ProvisioningState: to.StringPtr(string(compute.ProvisioningState1Succeeded)),
+							SinglePlacementGroup: to.BoolPtr(false),
+							ProvisioningState:    to.StringPtr(string(compute.ProvisioningState1Succeeded)),
 						},
 					},
 					[]compute.VirtualMachineScaleSetVM{
--- a/azure/services/scalesets/scalesets.go
+++ b/azure/services/scalesets/scalesets.go
@@ -387,6 +387,7 @@ func (s *Service) buildVMSSFromSpec(ctx context.Context, vmssSpec azure.ScaleSet
 		Zones: to.StringSlicePtr(vmssSpec.FailureDomains),
 		Plan:  s.generateImagePlan(),
 		VirtualMachineScaleSetProperties: &compute.VirtualMachineScaleSetProperties{
+			SinglePlacementGroup: to.BoolPtr(false),
 			UpgradePolicy: &compute.UpgradePolicy{
 				Mode: compute.UpgradeModeManual,
 			},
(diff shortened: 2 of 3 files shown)


jackfrancis:
cc @CecileRobertMichon @devigned I don't think we want to release v1alpha4 of capz without this change

CecileRobertMichon:
Fixes #484 ?

jackfrancis:
> Fixes #484 ?

I think so!

CecileRobertMichon:
> cc @CecileRobertMichon @devigned I don't think we want to release v1alpha4 of capz without this change

release is planned for tomorrow morning, will try to merge this before then but this is non-blocking, we can merge it and release it in a follow up patch (since it's not addressing a regression introduced in v1alpha4 and it's not a breaking change)

devigned:
/hold

With large clusters, the outbound ports will need to be expanded.

CecileRobertMichon:
> With large clusters, the outbound ports will need to be expanded.

would https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1229 suffice?

jackfrancis:
@CecileRobertMichon I think this is the change that @devigned is referring to:

https://github.com/Azure/aks-engine/pull/2526/files

Would that be appropriate in this PR? Or a distinct PR?

CecileRobertMichon:
> Would that be appropriate in this PR? Or a distinct PR?

let's do a separate PR since the two changes seem independent from each other and have different behavior impacts (even though they ultimately work towards the same goal)

CecileRobertMichon:
/lgtm

@devigned do we need to hold this for #1493 ? They look independent to me 

Also just curious, are you following https://docs.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-placement-groups#checklist-for-using-large-scale-sets for
> With large clusters, the outbound ports will need to be expanded.

devigned:
@CecileRobertMichon I don’t think we need to hold this pending the other PR. However, large clusters need the dynamic SNAT change too.

PR #1493, opened 2021-07-07
feat: automatic AllocatedOutboundPorts LB value

jackfrancis:
**What type of PR is this?**

/kind feature

This PR ensures that we use the Azure defaults for allocated outbound ports when creating a LoadBalancer, as different VMSS (AzureMachinePool) sizes are better suited to different values. Using "0" is the way to express "automatic".

To be clear, this configuration change will affect both AzureMachine (plain VMs) and AzureMachinePools (VMSS) resources. Azure's documentation guidance suggests VMSS, but our experience using this configuration in other Azure scenerios suggests it is appropriate for all VM types.

We will want to do some follow-up tests to validate how the "automatic" configuration adapts when the size of the AzureMachinePool changes. In other Azure + Kubernetes environments, using "0" when creating the cluster backed by VMSS works well in highly dynamic environments.

Ref:

https://docs.microsoft.com/en-us/azure/load-balancer/load-balancer-outbound-connections#outboundrules

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
LBs use automatic AllocatedOutboundPorts value
```

### DIFF
--- a/azure/services/loadbalancers/loadbalancers.go
+++ b/azure/services/loadbalancers/loadbalancers.go
@@ -240,6 +240,7 @@ func (s *Service) getOutboundRules(lbSpec azure.LBSpec, frontendIDs []network.Su
 				BackendAddressPool: &network.SubResource{
 					ID: to.StringPtr(azure.AddressPoolID(s.Scope.SubscriptionID(), s.Scope.ResourceGroup(), lbSpec.Name, lbSpec.BackendPoolName)),
 				},
+				AllocatedOutboundPorts: to.Int32Ptr(0),
 			},
 		},
 	}
--- a/azure/services/loadbalancers/loadbalancers_test.go
+++ b/azure/services/loadbalancers/loadbalancers_test.go
@@ -404,8 +404,9 @@ func newDefaultNodeOutboundLB() network.LoadBalancer {
 						BackendAddressPool: &network.SubResource{
 							ID: to.StringPtr("/subscriptions/123/resourceGroups/my-rg/providers/Microsoft.Network/loadBalancers/my-cluster/backendAddressPools/my-cluster-outboundBackendPool"),
 						},
-						Protocol:             network.LoadBalancerOutboundRuleProtocolAll,
-						IdleTimeoutInMinutes: to.Int32Ptr(30),
+						Protocol:               network.LoadBalancerOutboundRuleProtocolAll,
+						IdleTimeoutInMinutes:   to.Int32Ptr(30),
+						AllocatedOutboundPorts: to.Int32Ptr(0),
 					},
 				},
 			},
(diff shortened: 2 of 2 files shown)


devigned (changes requested):
Changes look reasonable to me. Please consider a slightly more descriptive release note and omit the `feat:` from the note.

CecileRobertMichon:
The PR description only mentions VMSS but the way it is implemented it will affect all CAPZ load balancers, including the ones attached to VMs, is that intentional? Is this change desirable for _all_ load balancers?

jackfrancis:
@CecileRobertMichon my point of reference is AKS Engine, which uses this configuration for all "Standard" Load Balancer scenarios, whether or not VMSS or standard VMs are behind that Load Balancer.

CecileRobertMichon:
@jackfrancis in that case let's update the PR description to make it clear that it's intentionally changing the behavior for every LB

jackfrancis:
@CecileRobertMichon PTAL at updated description

CecileRobertMichon:
@jackfrancis @devigned this is what I see in the portal with this change:

Screen Shot 2021-07-07 at 4 34 49 PM

Do you have a good understanding (I don't) of what that warning means and why this change is necessary besides "that's how AKS Engine does is" ? Especially since https://github.com/Azure/aks-engine/pull/2526 is from Jan 2020 and things may have change since then.

jackfrancis:
@CecileRobertMichon I think that is a warning to indicate that "Azure will choose an appropriate port number for you based on the number of VMSS instances (or, I think more accurately the number of backend pool members) at the time of LB creation. However, if you add more instances (and thus more backend pool members) the value Azure chose for you originally may no longer be valid."

I'm not sure what the solution is here. Can we update this LB configuration later (say, when the cluster scales out or in)? We could file that work as a follow-up item.

The reason I think this ("0") change is better than what we have currently is because in my tests, providing no configuration at all yielded the static value of "1024", which was actually preventing me from building maximally large AzureMachinePools (1000 replicas/instances).

tl;dr I think "let Azure choose the value based on the number of instances in the backend pool at the time of cluster creation" is better than "1024", which effectively prevents the cluster from ever going above 62 nodes (based on the hard limit of 64k SNAT ports across all members of the LB) assuming you only have 1 IP address.

CecileRobertMichon:
> I'm not sure what the solution is here. Can we update this LB configuration later (say, when the cluster scales out or in)? We could file that work as a follow-up item.

This would be complex to do since the LB is reconciled by the AzureCluster controller which has no knowledge (by design) of how many Machine/MachinePool replicas get created later. Also, how would we update it? Change from automatic to manual? In that case why not set to manual from the beginning and make it a dynamic value we control?

>which effectively prevents the cluster from ever going above 62 nodes (based on the hard limit of 64k SNAT ports across all members of the LB) assuming you only have 1 IP address.

What is the upside of doing this instead of having multiple node outbound IP addresses? That's already possible today https://capz.sigs.k8s.io/topics/node-outbound-lb.html#note

CecileRobertMichon:
BTW you may also want to look into how this large cluster scenario applies to NAT Gateways which is the Azure recommended way of doing outbound connectivity (https://docs.microsoft.com/en-us/azure/load-balancer/load-balancer-outbound-connections#scenarios) and was recently implemented in CAPZ (#898): https://capz.sigs.k8s.io/topics/node-outbound-lb.html#node-outbound-nat-gateway

CecileRobertMichon:
@jackfrancis are we good to close this given #1532 ?

PR #1496, opened 2021-07-07
CAEP for VMs as bastion hosts

whites11:
/kind design

Towards: https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/165#issuecomment-871256328

This is a CAEP to add support for virtual machines as bastion hosts in CAPZ.

Please be patient. This is my first CAEP and I am sure there is room for improvement.

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```


CecileRobertMichon (commented):
Thanks for opening the PR @whites11! I have not gotten to review it yet but I have not forgotten, it's on my list 

/cc @devigned

whites11:
> Thanks for opening the PR @whites11! I have not gotten to review it yet but I have not forgotten, it's on my list
> 
> /cc @devigned

No worries at all, thanks for the message.

docs/proposals/20210705-bastion-vms.md
```diff
@@ -0,0 +1,198 @@
...
+  azureBastion:
+    bastionHosts:
+    - bootstrapDataSecretName: "<secret name>"
+      azureMachineTemplate: {}
+      replicas: 2
+```
+
+With the above information, the `AzureCluster controller` can create an `AzureMachineTemplate` and a `MachineDeployment` and then the creation of VMs should happen automagically.
+
+Why a slice of `bastionHosts` rather than a single value? Because having multiple bastion hosts only makes sense if they are placed in different availability zones.
```
devigned, line 101: What do you think about making this a Virtual Machine Scale Set rather than individual machines. On a VMSS, you can specify the zones you would like and the platform will balance the machines. Also, if you update the cloud-init, you could have the VMSS automatically roll out the change.

docs/proposals/20210705-bastion-vms.md
```diff
@@ -0,0 +1,198 @@
...
+Frank is a Platform Engineer for a company offering Kubernetes as a Service.
+The company has a lot of kubernetes clusters and Frank, together with his colleagues, needs to access the machines
+forming the Kubernetes cluster via SSH in order to provide support to his company's customers.
+
+For security policy compliance reasons, the master node(s) of the clusters can't expose the SSH port to the public internet.
+
+Frank writes a `MutatingWebhook` that ensures every `AzureCluster` CR contains a `spec.bastionSpec.bastionHosts` to set up 
+one bastion host for every cluster.
+
+This way he ensures all Clusters have the same setup in terms of Bastions and he can create a tool to easily SSH into
```
devigned, line 119: Are all of the SSH keys in the cloud-init?

devigned (commented):
Can you go into a little bit detail on the networking and how it relates to the cluster. For example: 
- Will the bastion VMs be in the same VNet as the cluster or will the be in their own peered VNet? 
- What ports will be exposed to the public? Will they be configurable?
- Will there be network security rules to limit ingress to IP ranges?
- Will the VNet the bastions belong to have a VNet VPN, or is it open to all traffic?

Thank you for opening this proposal. I'm super excited about this functionality!

docs/proposals/20210705-bastion-vms.md
whites11: Potentially, but this is user choice.
For example, we (Giant Swarm) would not use ssh keys but use Hashicorp's Vault for SSH purposes.
The whole point of asking the client to give a full cloud init secret is to have max flexibility for virtually any use case

docs/proposals/20210705-bastion-vms.md
whites11: I am not sure I like it.
In general I'm all in for VMSSes in place of VMs, but the main point of having more than one bastion is to distribute them in different availability zones.
VMSS have multi AZ support, but customers don't have control in which AZs VMSS instances are run into.
Potentially all instances can end up in the same AZ and, in general, it is impossible to force AZ allocation.
This is why I went down the VM road.
WDYT?

whites11:
@devigned added my view about networking.

docs/proposals/20210705-bastion-vms.md
devigned: I believe we can achieve strict zone balancing and also enforce max spread of machines via https://docs.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-use-availability-zones#zone-balancing so that machines are required to be in difference AZs and different fault domains.

docs/proposals/20210705-bastion-vms.md
devigned: I see you point about having flexibility.

How do you envision the behavior when an operator will mutate or replace the cloud-init secret?

docs/proposals/20210705-bastion-vms.md
whites11: Didn't know about this. Then it totally makes sense to switch to VMSS to me.
Will work on that. Thank you David

docs/proposals/20210705-bastion-vms.md
whites11: The bastion hosts should be rolled in my opinion.
In general, I would stick to what CAPZ is doing with node pool nodes in this case.

docs/proposals/20210705-bastion-vms.md
whites11: I gave some more though time to this VMSS vs VM subject.
I started thinking that this whole CAEP might be useless.
Using `(Azure)MachinePools` to bootstrap bastion hosts requires the user to provide a whole `MachinePoolSpec` CR definition inside the `AzureCluster`. What is the point of adding it there? The user can simply create the `MachinePool` CR directly.
By the way I tried to do so, and it is currently impossible to create VMSSes as bastions using existing controllers, because:

- there is no support for adding public IPs to VMSS instances
- if you add support for public IPs bootstrap fails because of this https://github.com/Azure/aks-engine/issues/3521

I will try to work on his a little bit longer (trying to deploy VMs/VMSSes as bastions to get the feeling how it could work) but I feel like this CAEP is destined to be closed.

docs/proposals/20210705-bastion-vms.md
```diff
@@ -0,0 +1,209 @@
...
+Accessing virtual machines forming a Kubernetes cluster for administrative purposes is a potential security threat.
+A common pattern to provide SSH/RDP access to virtual machines without exposing them to the public internet is to use 
+a `Bastion host` or `Jump host`, a dedicated machine that is exposed to the internet acting as a bridge between the administrators'
+computers and the Kubernetes nodes (that can therefore stay behind a firewall).
+This CAEP aims at providing a way to bootstrap `Bastion Hosts` in a CAPZ cluster levereging existing Controllers as much as possible
+by reusing CRDs like `MachineDeployment` and `AzureMachineTemplate`.
+
+## Motivation
+
+By default CAPZ clusters expose one of the Master nodes to the internet making it possible to access all the other nodes in the cluster via SSH or RDP.
```
CecileRobertMichon, line 56:
```suggestion
By default CAPZ clusters expose one of the control plane nodes to the internet making it possible to access all the other nodes in the cluster via SSH or RDP.
```

docs/proposals/20210705-bastion-vms.md
```diff
@@ -0,0 +1,209 @@
...
+
+As an alternative, it is possible to use the [`AzureBastion`](https://azure.microsoft.com/en-us/services/azure-bastion/) feature to SSH/RDP to nodes using a web interface through 
+the `Azure Portal` (see [documentation](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/docs/book/src/topics/ssh-access.md)).
+
+Those solutions might not cover all use cases. As a matter of example:
+
+- Using one master node as a jump host might be not compliant to the security policies for some companies.
```
CecileRobertMichon, line 63:
```suggestion
- Using one control plane node as a jump host might be not compliant with the security policies for some companies.
```

docs/proposals/20210705-bastion-vms.md
```diff
@@ -0,0 +1,209 @@
...
+- Using `AzureBastion` might be inpractical or unusable in some use cases (machine to machine access would be nearly impossible).
```
CecileRobertMichon, line 64: > machine to machine access would be nearly impossible

can you clarify what this means?

docs/proposals/20210705-bastion-vms.md
```diff
@@ -0,0 +1,209 @@
...
+
+### Goals
+
+- Allow users to deploy virtual machines as bastion hosts in multiple availability zones using a custom cloud init file.
+
+### Non-Goals/Future Work
+
+- Make it possible to disable the default behaviour of exposing one master node to the internet.
```
CecileRobertMichon, line 72:
```suggestion
- Make it possible to disable the default behaviour of exposing one control plane node to the internet.
```

docs/proposals/20210705-bastion-vms.md
```diff
@@ -0,0 +1,209 @@
...
+```yaml
+apiVersion: infrastructure.cluster.x-k8s.io/v1alpha4
+kind: AzureCluster
+metadata:
+  name: cluster-name
+  namespace: default
+spec:
+  azureBastion:
+    bastionHosts:
+    - bootstrapDataSecretName: "<secret name>"
```
CecileRobertMichon, line 94: this could be confused with the field `DataSecretName` same name in CABPK

docs/proposals/20210705-bastion-vms.md
```diff
@@ -0,0 +1,209 @@
...
+the `Azure Portal` (see [documentation](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/docs/book/src/topics/ssh-access.md)).
+
+Those solutions might not cover all use cases. As a matter of example:
+
+- Using one master node as a jump host might be not compliant to the security policies for some companies.
+- Using `AzureBastion` might be inpractical or unusable in some use cases (machine to machine access would be nearly impossible).
+
+### Goals
+
+- Allow users to deploy virtual machines as bastion hosts in multiple availability zones using a custom cloud init file.
```
CecileRobertMichon, line 68: > custom cloud init file

does this imply the bastion VMs can be Linux only? should windows bastions / other OSes like Flatcar be added as a non-goal?

docs/proposals/20210705-bastion-vms.md
```diff
@@ -0,0 +1,209 @@
...
+
+### Non-Goals/Future Work
```
CecileRobertMichon, line 70: should we also add removing the existing AzureBastion service support as a non-goal?

docs/proposals/20210705-bastion-vms.md
CecileRobertMichon: > Using (Azure)MachinePools to bootstrap bastion hosts requires the user to provide a whole MachinePoolSpec CR definition inside the AzureCluster. What is the point of adding it there? The user can simply create the MachinePool CR directly.

The key difference is that the bastion VM/VMSS instance shouldn't itself be a k8s node. It shouldn't have azure.json on it, and it shouldn't run kubeadm.

> if you add support for public IPs bootstrap fails because of this Azure/aks-engine#3521

I don't think that applies to CAPZ, this is just a shortcoming of AKS Engine

devigned: Perhaps, adding a public IP for each VMSS instance is not needed. I think this could be accomplished by adding a NAT rule to the public LB to NAT traffic to individual instances.

docs/proposals/20210705-bastion-vms.md
whites11: No at all. Allowing users to specify the config secret directly grants maximum flexibility.
For example I've been able to test this using `flatcar linux` and `ignition` so I assume Windows will be perfectly supported.

docs/proposals/20210705-bastion-vms.md
whites11: That's exactly what it is. That field is meant to be a reference to a secret that is identical to the secret that CABPK would create for a "traditional" Machine.

docs/proposals/20210705-bastion-vms.md
whites11: I'd say no. In my vision `AzureBastion` has a different use case and it can be an alternative solution to whom that don't need/want a dedicated VM as bastion.

whites11:
> The key difference is that the bastion VM/VMSS instance shouldn't itself be a k8s node. It shouldn't have azure.json on it, and it shouldn't run kubeadm.

Exactly. That's why clients need to provide their pre-computed cloud-init (or equivalent) config. Kubeadm and CAPBK is not in play here and kubeadm is potentially not even installed in the VM image.

> I don't think that applies to CAPZ, this is just a shortcoming of AKS Engine

I faced this same error when:
- I edited CAPZ controller to enable public IPs for VMSSes.
- I tried to create one AzureMachinePool.

docs/proposals/20210705-bastion-vms.md
whites11: Added a hopefully clearer explanation

whites11:
@CecileRobertMichon @devigned let me try to clarify what is puzzling me and making me think to close this PR.

I can currently bootstrap a bastion host (with some limitations but let's ignore them for now) the way I am imagining it by using the following CRs:

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: bastions
  namespace: default
type: Opaque
stringData:
  value: |
    {
      "ignition": {
        "version": "2.2.0"
      },
      "storage": {
        "files": [
          {
            "path": "/run/cluster-api/bootstrap-success.complete",
            "filesystem": "root",
            "mode": "660",
            "user": {
              "id": 0
            },
            "group": {
              "id": 0
            }
          }
        ]
      },
      "passwd": {
        "users": [
          {
            "name": "username",
            "groups": [
              "sudo",
              "docker"
            ],
            "sshAuthorizedKeys": [
              "ssh-rsa AAAA..."
            ]
          }
        ]
      }
    }
---
apiVersion: cluster.x-k8s.io/v1alpha4
kind: MachinePool
metadata:
  name: ${CLUSTER_NAME}-bastions
  namespace: default
spec:
  clusterName: ${CLUSTER_NAME}
  replicas: 1
  template:
    spec:
      bootstrap:
        dataSecretName: "bastions"
      clusterName: ${CLUSTER_NAME}
      infrastructureRef:
        apiVersion: infrastructure.cluster.x-k8s.io/v1alpha4
        kind: AzureMachinePool
        name: ${CLUSTER_NAME}-bastions
      version: ${KUBERNETES_VERSION}
---
apiVersion: infrastructure.cluster.x-k8s.io/v1alpha4
kind: AzureMachinePool
metadata:
  name: ${CLUSTER_NAME}-bastions
  namespace: default
spec:
  location: ${AZURE_LOCATION}
  strategy:
    rollingUpdate:
      deletePolicy: Oldest
      maxSurge: 25%
      maxUnavailable: 1
    type: RollingUpdate
  template:
    osDisk:
      diskSizeGB: 30
      managedDisk:
        storageAccountType: Premium_LRS
      osType: Linux
    image:
      marketplace:
        offer: flatcar-container-linux-free
        publisher: kinvolk
        sku: stable
        version: 2765.2.6
        thirdPartyImage: true
    vmSize: ${AZURE_NODE_MACHINE_TYPE}
```

This CAEP is about embedding the above resource definitions in `AzureCluster/Spec/AzureBastion/BastionHosts` field, like this:

```yaml
apiVersion: infrastructure.cluster.x-k8s.io/v1alpha4
kind: AzureCluster
metadata:
  name: cluster-name
  namespace: default
spec:
  azureBastion:
    bastionHosts:
      machinePool:
        ... <machine pool contents here> ...
```

What is the advantage of embedding the `machinePool` definition inside `AzureCluster`? I can simply create the `MachinePool` CR in the management cluster to achieve the same goal. Hope I made my doubt clear, I look forward to your thoughts.

CecileRobertMichon:
Something that was brought up by @jsturtevant in this morning's office hours is that we should look into maybe leveraging Azure Service Operator for this use case. It could be deployed side by side with CAPZ controllers to deploy any additional Azure resources, including VMs to be used as jumphosts. This would also be extensible to future use cases folks might have to deploy all sorts of Azure resources that are not strictly part of the k8s cluster.

@devigned @mattchr @Porges @babbageclunk @theunrepentantgeek does that seem like a reasonable use case of ASO alongside CAPZ and a good fit to solve the bastion host problem?

whites11:
OK I think we have consensus here about the fact this CAEP should not move forwards because the approach is not "general" enough. 
I will go on and close this PR.
Thanks to all involved in the discussion.

matthchr:
@CecileRobertMichon - I see this is closed now but I think that's a reasonable use case for ASO. It doesn't currently support bastion but it obviously could. As users start wanting more "extra stuff" rather than getting that stuff modeled in CAPZ it may make sense to use ASO and then figure out some interoperation where it makes sense (setting up resource links, etc).

whites11:
> @CecileRobertMichon - I see this is closed now but I think that's a reasonable use case for ASO. It doesn't currently support bastion but it obviously could. As users start wanting more "extra stuff" rather than getting that stuff modeled in CAPZ it may make sense to use ASO and then figure out some interoperation where it makes sense (setting up resource links, etc).

@matthchr I closed this PR because I feel like if/when we'll integrate ASO into CAPZ it needs to be a bigger scope CAEP rather than this bastion-specific one. Feel free to reopen if you think otherwise

matthchr:
@whites11 no that seems reasonable to me. I don't really see anything that's going to be "Bastion specific" at least right now - I'm expecting the questions that need resolving are all pretty general and would apply to other resources that CAPZ is interested in interfacing with as well.

PR #1439, opened 2021-06-11
Export traces to App Insights via OpenTelemetry

mboersma:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

Removes jaeger from the `make tilt-up` workflow and replaces it with [opentelemetry-collector](https://github.com/open-telemetry/opentelemetry-collector) (still operating in jaeger-compatible collector mode). Adds the [azuremonitor exporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/azuremonitorexporter) component to send CAPZ controller-monitor tracing to App Insights for visualization in the "Transaction Search" function.

**Which issue(s) this PR fixes**:

Fixes #1113
Depends on #1412

**Special notes for your reviewer**:

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [ ] adds unit tests
- [x] Merge #1412 and rebase
- [x] Testing by others, since `make tilt-up` understandably isn't covered by any test harness
- [x] Documentation on how to install the Helm chart on an existing cluster
- [x] Documentation (with pictures!) on how to set up Azure Insights and retrieve the instrumentation key
- [x] Any other TODOs in the readme.md
- [x] Warn if `AZURE_INSTRUMENTATION_KEY` isn't set, but ensure the cluster continues to function (without tracing)
- [ ] Document a local trace destination -- maybe how to install the Jaeger all-in-one helm chart?
- [ ] Maybe switch to the otlp collector? Or make that a followup PR

**Release note**:

```release-note
Export traces to App Insights via OpenTelemetry
```

PR #1498, opened 2021-07-07
Export traces to Jaeger and update OpenTelemetry

### LINKED ISSUE #1113: Export traces in OpenTelemetry format and then to Azure Monitor

mboersma:
**What type of PR is this?**

/kind feature

See #1439 for part one of these changes.

Adds (back) a Jaeger all-in-one Pod for viewing CAPZ controller traces locally (in addition to App Insights). Updates all opentelemetry-go code to their latest split releases, v1.0.0-RC2 and v0.22.0. Adds "View traces" and "View metrics" links on the appropriate Tilt resources, as well as an "App Insights" link if `AZURE_INSTRUMENTATION_KEY` was set. Adds and updates lots of observability docs.

Fixes #1113
Depends on #1439 (or closes it, if we merge this directly)

#1439 is already `XXL` and so is this additional changeset. I think we should merge #1439 first and then I can rebase this for reviewing separately. Just wanted to get it out there for any feedback.

I thought it was important to have a local (Jaeger) solution for viewing traces, rather than require App Insights. Please let me know if the Tilt warning approach seems too heavy-handed.

Are there unit tests that could be written here (that would actually be worthwhile)?

**TODOs**:

- [ ] squashed commits
- [x] includes documentation
- [ ] adds unit tests
- [ ] add a readme.md for hack/observability/jaeger

**Release note**:

```release-note
Export traces to Jaeger and update OpenTelemetry
```


version/version.go
```diff
@@ -58,7 +58,7 @@ func Get() Info {
 		GoVersion:       runtime.Version(),
 		Compiler:        runtime.Compiler,
 		Platform:        fmt.Sprintf("%s/%s", runtime.GOOS, runtime.GOARCH),
-		AzureSdkVersion: fmt.Sprintf("v%s", azuresdkversion.Number),
```
mboersma, line 61: This was just a bug I noticed--`azuresdkversion.Number` is already "v55.2.0", so no there's no need to prepend a "v."

pkg/ot/traces.go
```diff
@@ -0,0 +1,91 @@
...
+	return nil
+}
+
+// otlpTracerProvider initializes an OTLP exporter and configures the corresponding tracer provider.
+func otlpTracerProvider(ctx context.Context, url string) (*sdktrace.TracerProvider, error) {
+	res, err := resource.New(ctx,
+		resource.WithAttributes(
+			semconv.ServiceNameKey.String("capz"),
+			attribute.String("exporter", "otlp"),
+			attribute.String("version", version.Get().String()),
```
mboersma, line 65: `version.Get().String()` returns "" for me in `make tilt-up`. You can see the same in the controller logs. Is this expected?

Also, I just added "version" and "azuresdk.version" here as an example of trace attributes, and to see how they show up in App Insights and Jaeger, but we may not want to add these.

mboersma:
I retested this PR and it's working well. I think three things remain:

- [ ] add a screenshot of Jaeger traces
- [ ] check that there are resource requests/limits on all observability components
- [ ] are there any opportunities here for unit test coverage?

mboersma:
I rebased this, updated to otel v1.0.0-RC2/v0.22.0, and retested locally.  It's working well, so it's ready for review IMHO.

mboersma:
/retest

A test cluster failed to provision, unrelated to this change AFAICT.

pkg/ot/traces.go
```diff
@@ -0,0 +1,91 @@
...
+		otlptracegrpc.WithInsecure(),
+		otlptracegrpc.WithEndpoint(url),
+	)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to create otlp trace exporter")
+	}
+
+	bsp := sdktrace.NewBatchSpanProcessor(traceExporter)
+	tracerProvider := sdktrace.NewTracerProvider(
+		sdktrace.WithSampler(sdktrace.AlwaysSample()),
```
devigned, line 83: We are going to want to think about this in the future. Perhaps, we should have some configuration around this to allow for a percentage based sampling. I don't think we should gate this PR on this, but just wanted to bring it up.

devigned (approved):
/lgtm

I did leave one piece of feedback about always sampling and that folks in production will likely want some control over the amount of sampling they are doing, rather than all or nothing as configuration allows today. This is something we should add in the future, but shouldn't gate this PR.

Awesome work on the readme(s)!

pkg/ot/traces.go
mboersma: 👍🏻 Totally agree, but since this initial implementation is just for devs, I think it's ok to hard-code some things for now.

mboersma:
I've had many e2e failures on this PR, but they all seem to be of the "timed out waiting for cluster to provision" variety which doesn't seem attributable to this PR. Are we just having a flaky time currently in CI or is there something I should dig into?

CecileRobertMichon (approved):
/lgtm
/approve

🙌  +1, great work on the readmes
Also, nice touch with the tilt links

PR #1501, opened 2021-07-08
add AzureClusterTemplate type

### LINKED ISSUE #1500: Add AzureClusterTemplate type
/kind feature

Add a new CRD named `AzureClusterTemplate`. This is needed for ClusterClass.
In the short term this type will not be used outside of ClusterClass.

ykakarap:
**What type of PR is this?**
/kind feature

Adds AzureClusterTemplate to support ClusterClass.

Fixes #1500

**Release note**:

```release-note
NONE
```


api/v1alpha4/azureclustertemplate_types.go
```diff
@@ -0,0 +1,94 @@
...
+// +kubebuilder:printcolumn:name="Location",type="string",priority=1,JSONPath=".spec.location"
+// +kubebuilder:resource:path=azureclustertemplates,scope=Namespaced,categories=cluster-api
+// +kubebuilder:storageversion
+
+// AzureClusterTemplate is the Schema for the azureclustertemplates API
+type AzureClusterTemplate struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec AzureClusterTemplateSpec `json:"spec,omitempty"`
```
shysank, line 80: The spec should refer to a `template` which in turn will refer to `AzureClusterSpec`.  Something like we do for [AzureMachineTemplate](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/d6a8184d4779bfddbc1efb87d0c910c55782e44a/api/v1alpha3/azuremachinetemplate_types.go#L32)

shysank (changes requested):

main.go
```diff
@@ -44,6 +44,7 @@ import (
 	cgrecord "k8s.io/client-go/tools/record"
 	"k8s.io/klog/v2"
 	"k8s.io/klog/v2/klogr"
+	infrastructurev1alpha4 "sigs.k8s.io/cluster-api-provider-azure/api/v1alpha4"
```
alexeldeib, line 47: dupe aliased import

main.go
```diff
@@ -85,6 +86,7 @@ func init() {
 	_ = infrav1alpha4exp.AddToScheme(scheme)
 	_ = clusterv1.AddToScheme(scheme)
 	_ = clusterv1exp.AddToScheme(scheme)
+	_ = infrastructurev1alpha4.AddToScheme(scheme)
```
alexeldeib, line 89: same as above, dupe aliased import

api/v1alpha4/azureclustertemplate_types.go
ykakarap: got it :) Will make the changes now.

fabriziopandini (changes requested):
@ykakarap thanks for this PR!

Could you kindly enable web hooks for the new type ensuring that template type applies the same defaulting & validation rules of the actual type, otherwise this could trigger continuous reconcile loop down the path when comparing the desired state (the AzureClusterTemplate object) with the actual state (the instance of AzureCluster generated from the template)?

ykakarap:
While working on @fabriziopandini comments I observed that `AzureMachine` and `AzureMachineTemplate` do not have the same validate function (ref [ValidateCreate](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/api/v1alpha4/azuremachine_webhook.go#L46-L78) of AzureMachine and [ValidateCreate](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/api/v1alpha4/azuremachinetemplate_webhook.go#L48-L50) of AzureMachineTemplate). To avoid a continuous reconcile loop the validations/defaulting logic should match across the template and the actual object, correct?

Thoughts?

cc @CecileRobertMichon

ykakarap:
> I think we should, to avoid reconcile failure loops in machine deployments/machine sets. I just saw capa code and noticed that it's not there either. Not sure if this was a conscious decision taken at that time.

cc @randomvariable

CecileRobertMichon:
> I just saw capa code and noticed that it's not there either. Not sure if this was a conscious decision taken at that time.

At the time they were added (v1alpha2), CAPZ code was a lot of copy/paste from CAPA code so that's probably why it was done that way. I'm personally not sure what the context was on not validating the templates on the CAPA side.

ykakarap:
From the discussion it looks like we do need the validation. 
I can add the validation logic to the machine templates :)

Should that be a separate PR or is it okay to be part of this PR?

CecileRobertMichon:
> Should that be a separate PR or is it okay to be part of this PR?

Separate PR please

ykakarap:
PR to have consistent defaulting/validation for machines and machinetemplates: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1516

ykakarap:
@CecileRobertMichon I addressed all the comments on the PR and all the pipeline are now green.
This PR is ready for review. 

cc @fabriziopandini

fabriziopandini (commented):
Changes lgmt to me, thanks for making default and validation rules consistent!
I leave final world to CAPZ maintainers

api/v1alpha4/azurecluster_default.go
```diff
@@ -41,91 +41,91 @@ const (
...
+	nodeSubnet, err := s.NetworkSpec.GetNodeSubnet()
 	if err != nil {
 		nodeSubnet = SubnetSpec{Role: SubnetNode}
-		c.Spec.NetworkSpec.Subnets = append(c.Spec.NetworkSpec.Subnets, nodeSubnet)
+		s.NetworkSpec.Subnets = append(s.NetworkSpec.Subnets, nodeSubnet)
 	}
 
 	if cpSubnet.Name == "" {
-		cpSubnet.Name = generateControlPlaneSubnetName(c.ObjectMeta.Name)
+		cpSubnet.Name = generateControlPlaneSubnetName(name)
```
ykakarap, line 97: I made code changes so that the template and the cluster types have the same defaulting and validation rules. However I am not sure if the template type should have all the defaulting rules (and validations) that the cluster has. 
In this case, when we use `setSubnetDefaults` on the template type we use the template name as the argument and therefore the subnetname is based on the clustertemplate name. When clusters are created using this template they will all share the same subnet name. Is that okay?

shysank: Good point. As long as they are in separate resource groups, they should be fine. But, the resource group also relies on the cluster template name. So, we might have to find a way around for that.

shysank: Should we even set defaults for the template? the generated name for azure cluster will be different from the template name.

ykakarap: > Should we even set name related defaults for the template?

I have similar doubts. As long as all the validations/defaulting we perform on the template are also performed on the cluster we wont end up in the continuous reconcile loop (i think). 
The validation/defaulting we have on template should be a subset of the cluster type. But the cluster might have more validation/defaulting on top of what the template already does. 
As long as the validation/defaulting do not conflict with the ones from the template we are probably good.

cc @fabriziopandini thoughts?

shysank: One way to work around this can be to add an annotation in azure cluster to signal that it was created from a template and handle defaults accordingly.

ykakarap: The ClusterClass proposal already suggests adding the following label
```yaml
cluster.x-k8s.io/topology: <cluster-name>
```
to the infrastructure cluster object. We can use this to detect if the cluster was created from a template.

Note that this only works for the current usecases (ClusterClass realm) we have and does not work for any future use cases outside of the clsuterclass workflow where infrastructure cluster objects would be created out of the template object.

Even if we do detect if the cluster is created using a template we might still need to have all the defaulting and validation that a normal cluster object would have. I believe it is the defaulting and validations of the template that might need to be adjusted.

fabriziopandini: I'm not sure I have all the context here, but I'm happy to talk about this; In the meantime let me refocus on the problem statement and some of the above comments.

ClusterClass requirements and expected behaviour:

1 For creating a CAPZ cluster using ClusterClass, it is necessary to introduce the AzureClusterTemplate.
2 The spec.template from AzureClusterTemplate will be used to create actual AzureCluster, and we should make sure this operation can succeed.
3 After creating the actual AzureCluster, ClusterClass will continuously reconcile its content to the original spec.template from AzureClusterTemplate, and we should make sure to avoid continuous reconciliation or other unexpected behaviour, like value override.
4 ClusterClass, who will manage 2 and 3, will not have any knowledge of the Azure specific types, and treat them as unstructured

> The validation/defaulting we have on template should be a subset of the cluster type. But the cluster might have more validation/defaulting on top of what the template already does.

I'm not sure about this. 
more validation rules in the cluster than in the template: what if someone sets a value in the template (no validation rule) that won't be accepted in the cluster (additional validation rule)? this will prevent 2 to successfully create a Cluster.

more defaulting rules in the cluster than in the template:  if you leave a field empty in the template (no defaulting rule), and the you default it in the actual cluster (additional validation rule), then 3 will continuously reconcile and try to set this value  back to empty.

> One way to work around this can be to add an annotation in azure cluster to signal that it was created from a template and handle defaults accordingly

This could make sense to me, but it really depends by the specific problem
There are also other options we can consider, like making 4 more smart, but this has to be discussed and the solution should be generic enough to work will all the providers

CecileRobertMichon: need to spend more time thinking about this but my initial thought on this is that we should _not_. share RG/vnet name/subnet names between AzureClusters created from the same template.

shysank: cc: @randomvariable @yastij as there could be similar concerns in other providers.

CecileRobertMichon:
@ykakarap @fabriziopandini do you need help moving this forward?

api/v1alpha4/azurecluster_default.go
ykakarap: This issue will be addressed according to the recommendations listed here: https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/proposals/202105256-cluster-class-and-managed-topologies.md#conventions-for-template-types-implementation

TLDR; `AzureClusterTemplate` will only contain a subset of fields of the `AzureCluster`. Other fields in `AzureCluster` that need to be set (that are cluster specific) will be set using defaulting. 
Note: This will not trigger a continuous reconciliation. Refer to:
>If a field does not have a value in the template, but the users/another controller sets this value in the object, the reconcile topology process will preserve the field value.


cc @shysank @CecileRobertMichon

ykakarap:
This PR is very old.
Moving to a new PR: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2072

shysank:
can we close this one to avoid confusion?

PR #1502, opened 2021-07-09
Don't delete AzureIdentities from other Clusters

CecileRobertMichon:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**: This is fixing a bug where the AzureIdentities/Bindings were getting deleted when there were multiple AzureClusters using AzureClusterIdentity in the same management cluster. Since we iterate through all the bindings to find the ones to delete, they won't all match that one AzureCluster's identity, which would cause all the other ones to get deleted. This caused identities and bindings to constantly get deleted and recreated.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Don't delete AzureIdentities from other Clusters
```

### DIFF
--- a/controllers/azureidentity_controller.go
+++ b/controllers/azureidentity_controller.go
@@ -33,7 +33,6 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/client-go/tools/record"
 	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1alpha4"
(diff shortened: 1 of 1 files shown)


nader-ziada:
nice catch, only delete the `AzureIdentity` if the `AzureCluster` or `ManagedCluster` is not found makes sense

/lgtm
/approve

PR #1504, opened 2021-07-09
Set identity related env vars in create dev cluster

nader-ziada:
**What type of PR is this?**
/kind bug

- set env variables needed for AzureClusterIdentity and secret in create-dev-cluster

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
none
```

### DIFF
--- a/hack/create-dev-cluster.sh
+++ b/hack/create-dev-cluster.sh
@@ -50,6 +50,11 @@ export WORKER_MACHINE_COUNT=${WORKER_MACHINE_COUNT:-2}
 export KUBERNETES_VERSION="${KUBERNETES_VERSION:-v1.19.7}"
 export CLUSTER_TEMPLATE="${CLUSTER_TEMPLATE:-cluster-template.yaml}"
 
+# identity secret settings.
+export AZURE_CLUSTER_IDENTITY_SECRET_NAME="cluster-identity-secret"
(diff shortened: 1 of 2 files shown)


hack/create-dev-cluster.sh
```diff
@@ -50,6 +50,11 @@ export WORKER_MACHINE_COUNT=${WORKER_MACHINE_COUNT:-2}
 export KUBERNETES_VERSION="${KUBERNETES_VERSION:-v1.19.7}"
 export CLUSTER_TEMPLATE="${CLUSTER_TEMPLATE:-cluster-template.yaml}"
 
+# identity secret settings.
+export AZURE_CLUSTER_IDENTITY_SECRET_NAME="cluster-identity-secret"
```
CecileRobertMichon, line 54: can we remove them from the other script? we shouldn't need to set them in both

nader-ziada: If we do that, running make create management cluster without the script would not work

CecileRobertMichon: that's true but currently running make create-cluster doesn't work without the variables set below either

nader-ziada: ok, moved them from the other script

CecileRobertMichon (commented):
/lgtm
/approve

PR #1506, opened 2021-07-09
Remove owner ref from AzureClusterIdentity

nader-ziada:
**What type of PR is this?**
/kind bug

- Remove AzureClusterIdentity owner reference to avoid issues with a reference to an owner (Cluster) in a different namespace
 
Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
AzureClusterIdentity should not have an owner reference of a Cluster
```

### DIFF
--- a/templates/cluster-template-aad.yaml
+++ b/templates/cluster-template-aad.yaml
@@ -210,6 +210,8 @@ spec:
 apiVersion: infrastructure.cluster.x-k8s.io/v1alpha4
 kind: AzureClusterIdentity
 metadata:
+  labels:
+    clusterctl.cluster.x-k8s.io/move-hierarchy: "true"
   name: ${CLUSTER_IDENTITY_NAME}
   namespace: default
 spec:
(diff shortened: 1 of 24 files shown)


devigned (approved):
lgtm

templates/cluster-template-aad.yaml
```diff
@@ -210,13 +210,15 @@ spec:
...
+    clusterctl.cluster.x-k8s.io/move-hierarchy: "true"
+  name: ${CLUSTER_IDENTITY_NAME:=cluster-identity}
   namespace: default
 spec:
   allowedNamespaces: {}
   clientID: ${AZURE_CLIENT_ID}
   clientSecret:
-    name: ${AZURE_CLUSTER_IDENTITY_SECRET_NAME}
-    namespace: ${AZURE_CLUSTER_IDENTITY_SECRET_NAMESPACE}
+    name: ${AZURE_CLUSTER_IDENTITY_SECRET_NAME:=cluster-identity-secret}
```
CecileRobertMichon, line 221: Why all the added defaults in this PR? I recall we decided to leave those without defaults so it would be a reminder to folks to create the secret when using the new templates

nader-ziada: oh ok, I'll remove them, thought it would be useful, but good point

CecileRobertMichon (commented):
/lgtm

PR #1488, opened 2021-07-02
[donotmerge] debug aks e2e

alexeldeib:
**What type of PR is this?**

**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note

```

PR #1509, opened 2021-07-12
add daemonset implementation of timesync validation

alexeldeib:
**What type of PR is this?**
/kind cleanup

Add an implementation of the timesync test which avoids using SSH. useful for private clusters and AKS clusters in which nodes do not have public IPs. 

As part of debugging #1488, I found VMSS List calls across regions have an unexpected lag in returning correct results right now. This also affects AKS e2e resource validation. I worked around the VMSS List issue in the AMMP reconciler. The test validation itself was List VMSS calls which hit the same issue. So instead of simply removing the existing e2e validation, I implemented the existing time sync validation using a daemonset + nsenter rather than SSH (which won't work when nodes all have private IPs). This should also work nicely for private clusters.

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/azure_test.go
+++ b/test/e2e/azure_test.go
@@ -460,9 +460,9 @@ var _ = Describe("Workload cluster creation", func() {
 				},
 			}, result)
 
-			Context("Validating AKS Resources", func() {
-				AKSResourcesValidationSpec(ctx, func() AKSResourcesValidationSpecInput {
-					return AKSResourcesValidationSpecInput{
+			Context("Validating AKS time synchronization", func() {
+				AzureDaemonsetTimeSyncSpec(ctx, func() AzureTimeSyncSpecInput {
+					return AzureTimeSyncSpecInput{
 						BootstrapClusterProxy: bootstrapClusterProxy,
 						Namespace:             namespace,
 						ClusterName:           clusterName,
--- a/test/e2e/azure_timesync.go
+++ b/test/e2e/azure_timesync.go
@@ -21,14 +21,21 @@ package e2e
 import (
 	"context"
 	"fmt"
+	"io/ioutil"
 	"strings"
 	"time"
 
 	. "github.com/onsi/gomega"
 	"github.com/pkg/errors"
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/types"
+	e2e_pod "sigs.k8s.io/cluster-api-provider-azure/test/e2e/kubernetes/pod"
 	"sigs.k8s.io/cluster-api/test/framework"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	kinderrors "sigs.k8s.io/kind/pkg/errors"
+	"sigs.k8s.io/yaml"
 )
 
 // AzureTimeSyncSpecInput is the input for AzureTimeSyncSpec.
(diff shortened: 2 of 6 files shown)


alexeldeib:
looks like it works:
```
Jul 12 08:17:42.860: INFO: want 2 instances, found 0 ready and 0 available. generation: 1, observedGeneration: 1
Jul 12 08:17:53.206: INFO: want 2 instances, found 2 ready and 2 available. generation: 1, observedGeneration: 1
Jul 12 08:17:53.325: INFO: mapping nsenter pods to hostnames for host-by-host execution
[1mSTEP[0m: checking that time synchronization is healthy on aks-agentpool0-38648403-vmss000000
[1mSTEP[0m: checking that time synchronization is healthy on aks-agentpool1-38648403-vmss000000
```

test/e2e/workloads/nsenter/daemonset.yaml
```diff
@@ -0,0 +1,26 @@
...
+      app: *name
+  template:
+    metadata:
+      labels:
+        app: *name
+    spec:
+      hostNetwork: true
+      hostPID: true
+      containers:
+      - image: alexeldeib/nsenter:latest
```
CecileRobertMichon, line 19: there's no more "official" image for this? Should we store one on mcr?

test/e2e/azure_timesync.go
```diff
@@ -94,3 +101,149 @@ func AzureTimeSyncSpec(ctx context.Context, inputGetter func() AzureTimeSyncSpec
...
 	}, thirty, thirty).Should(Succeed())
 }
+
+const (
+	nsenterWorkloadFile = "workloads/nsenter/daemonset.yaml"
+)
+
+// AzureDaemonsetTimeSyncSpec implements a test that verifies time synchronization is healthy for
+// the nodes in a cluster. It uses a privileged daemonset and nsenter instead of SSH.
+func AzureDaemonsetTimeSyncSpec(ctx context.Context, inputGetter func() AzureTimeSyncSpecInput) {
```
CecileRobertMichon, line 111: Should this just replace the SSH implementation? what's the upside of keeping the SSH validation around for other clusters?

test/e2e/workloads/nsenter/daemonset.yaml
alexeldeib: I don't know of a more official image, the existing nsenter ones are all random people. I built this over the weekend to be as minimal as possible, I'm happy to put it in MCR (especially since docker hub will get rate limited). https://github.com/alexeldeib/nsenter/blob/main/Dockerfile

It's a scratch dockerfile with static sleep + nsenter binaries. < 2MB for the whole image 🙂

test/e2e/azure_timesync.go
alexeldeib: I didn't want to be too presumptuous, but I'd say yes. I don't know what we'd do for Windows but I assume chrony doesn't apply there anyway.

I was hoping to follow up and make this an easy way to do other on-host validation, since you get access to basically anything you want.

alexeldeib: I also kind of wanted to get some validation from merging this first and making sure they were both reliable, but it seems to work well already to me. Too used to debugging flakes in the other e2e 😄

test/e2e/workloads/nsenter/daemonset.yaml
alexeldeib: most common images like ubuntu etc have nsenter and sleep installed, but are pretty beefy, and slimmer stuff like distroless usually doesn't have the binaries.

alexeldeib: better?

CecileRobertMichon (commented):
/cc @mboersma @shysank 
for review

alexeldeib:
meh, busybox didn't end up working because it doesn't nsenter all namespaces properly. If I can't find an existing image which works for sleep + nsenter, I'll push my minimal image to MCR.

alexeldeib:
yayyy it works again. now to squash and clean up

alexeldeib:
I think this is good from my perspective, just trying to get more test signal. There will still be flakes on the initial nodes coming up until I break up and merge all the changes I discovered in https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1488, but that can't be helped in this PR.

test/e2e/azure_test.go
```diff
@@ -460,9 +460,9 @@ var _ = Describe("Workload cluster creation", func() {
 				},
 			}, result)
 
-			Context("Validating AKS Resources", func() {
-				AKSResourcesValidationSpec(ctx, func() AKSResourcesValidationSpecInput {
-					return AKSResourcesValidationSpecInput{
+			Context("Validating AKS time synchronization", func() {
+				AzureDaemonsetTimeSyncSpec(ctx, func() AzureTimeSyncSpecInput {
```
CecileRobertMichon, line 464: So is this replacing the check we were doing before? If so, is that code still being used anywhere (looks like no) / should we remove it?

alexeldeib: ah good catch. Thought I got it in my patch when I split the diff from the other PR.

test/e2e/azure_timesync.go
```diff
@@ -94,3 +101,144 @@ func AzureTimeSyncSpec(ctx context.Context, inputGetter func() AzureTimeSyncSpec
...
+	Expect(err).NotTo(HaveOccurred(), "failed to unmarshal nsenter json bytes to unstructured")
+
+	nsenterDs.SetNamespace("default")
+
+	err = kubeclient.Create(ctx, &nsenterDs)
+	Expect(err).NotTo(HaveOccurred(), "failed to create daemonset for time sync check")
+
+	Eventually(func() error {
+		var nodes corev1.NodeList
+		if err := kubeclient.List(ctx, &nodes); err != nil {
```
shysank, line 145: do we need to check this every time? Can we move this outside of eventually?

shysank (commented):
lgtm overall with one minor refactoring. 
Looking at the test, it is still somewhat slow (around 19 minutes with ) although faster than what it was previously. Did we expect to improve this in this pr, or is that part of the other one related to watches?

alexeldeib:
> Did we expect to improve this in this pr, or is that part of the other one related to watches?

The watches are the much bigger impact. Each of those (and there were 2-3 I think) could cause 10+min gaps reconciling. 

The original reason for this workaround in the resource validation was the cross-region latency listing VMSS. I haven't checked recently, but the latency the week we had flakes was REALLY bad. So it's probably just coincidental that when i'm finally ready to merge this PR, the obvious impact is gone :D

take a look at the times in the other PR, the entire prowjob takes ~25min fairly reliably vs flaking/40+ minutes.

test/e2e/azure_timesync.go
alexeldeib: there are two separate eventually loops. this one just waits for the DS pods to come up on all nodes. the other one tries to exec them until it succeeds.

I wrote two loops so we can wait for all nodes/pods read, list out the full set of pods, and then match + retry the exec until success or failure, exactly *without* repeating this inside the full loop more than necessary.

shysank:
> take a look at the times in the other PR, the entire prowjob takes ~25min fairly reliably vs flaking/40+ minutes.

Will do, Thanks for the explanation!

test/e2e/azure_timesync.go
shysank: IIUC, the nodes should exist by the time it reaches here. So, I thought we could determine the `desired` count before the first eventually loop.

test/e2e/azure_timesync.go
CecileRobertMichon: > I was hoping to follow up and make this an easy way to do other on-host validation, since you get access to basically anything you want.

could you please file an issue so we can have context for having the two side by side in the meantime in case you don't get to it right away?

test/e2e/azure_timesync.go
alexeldeib: took another look in capi, you are right, I can just check the pod counts/readiness here I think 👍

alexeldeib:
probably won't get to finish this until monday :/

test/e2e/azure_timesync.go
alexeldeib: https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1543

test/e2e/azure_timesync.go
alexeldeib: think that's better? it was an easy fix actually, since we still need to loop on pod readiness, I think?

shysank: Yeah, this was my expectation as well. Thanks for fixing it!

CecileRobertMichon:
thanks for the thoughtful review @shysank 

/approve

PR #1514, opened 2021-07-12
Allow for using SP Identity directly embedded in a secret

### LINKED ISSUE #1414: Service Principal Identity without Pod Identity
/kind feature

The AzureClusterIdentity only works when Azure Pod Identity is running. In cases when AAD Pod Identity is not running, it would be nice to have the ability to simply use a Service Principal embedded in a secret. Looking at the code [here](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/bdd9f8c0c92505224215a2e5e234088ee8234d32/azure/scope/identity.go#L143-L150), I don't think this would be difficult.

Add another type of Identity.Spec.Type to represent a "manual service principal", and add something like below into identity.go.

```go
        if p.Identity.Spec.Type == infrav1.ServicePrincipal {
		spt, err = adal.NewServicePrincipalTokenFromMSIWithUserAssignedID(msiEndpoint, resourceManagerEndpoint, p.Identity.Spec.ClientID)
		if err != nil {
			return nil, errors.Errorf("failed to get token from service principal identity: %v", err)
		}
	}else if p.Identity.Spec.Type == infrav1.ManualServicePrincipal {
		spt, err = adal.NewServicePrincipalToken(oAuthConfig, p.Identity.Spec.ClientID, secret.ClientSecret, resourceManagerEndpoint)
		if err != nil {
			return nil, errors.Errorf("failed to get token from manual service principal identity: %v", err)
		}
	}else if p.Identity.Spec.Type == infrav1.UserAssignedMSI {
		return nil, errors.Errorf("UserAssignedMSI not supported: %v", err)
	}
```

**Environment:**

shysank:
**What type of PR is this?**
/kind feature

Fixes #1414 

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Allow for using SP Identity directly embedded in a secret without relying on aad pod identity
```

### DIFF
--- a/azure/scope/identity.go
+++ b/azure/scope/identity.go
@@ -44,7 +44,7 @@ const azureSecretKey = "clientSecret"
 
 // CredentialsProvider defines the behavior for azure identity based credential providers.
 type CredentialsProvider interface {
-	GetAuthorizer(ctx context.Context, resourceManagerEndpoint string) (autorest.Authorizer, error)
+	GetAuthorizer(ctx context.Context, resourceManagerEndpoint, activeDirectoryEndpoint string) (autorest.Authorizer, error)
 	GetClientID() string
 	GetClientSecret(ctx context.Context) (string, error)
 	GetTenantID() string
@@ -103,8 +103,8 @@ func NewAzureClusterCredentialsProvider(ctx context.Context, kubeClient client.C
 }
 
 // GetAuthorizer returns an Azure authorizer based on the provided azure identity. It delegates to AzureCredentialsProvider with AzureCluster metadata.
-func (p *AzureClusterCredentialsProvider) GetAuthorizer(ctx context.Context, resourceManagerEndpoint string) (autorest.Authorizer, error) {
-	return p.AzureCredentialsProvider.GetAuthorizer(ctx, resourceManagerEndpoint, p.AzureCluster.ObjectMeta)
+func (p *AzureClusterCredentialsProvider) GetAuthorizer(ctx context.Context, resourceManagerEndpoint, activeDirectoryEndpoint string) (autorest.Authorizer, error) {
+	return p.AzureCredentialsProvider.GetAuthorizer(ctx, resourceManagerEndpoint, activeDirectoryEndpoint, p.AzureCluster.ObjectMeta)
 }
 
 // NewManagedControlPlaneCredentialsProvider creates a new ManagedControlPlaneCredentialsProvider from the supplied inputs.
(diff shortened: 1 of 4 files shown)


azure/scope/identity.go
```diff
@@ -132,15 +140,62 @@ func NewManagedControlPlaneCredentialsProvider(ctx context.Context, kubeClient c
...
+		if err != nil {
+			return nil, err
+		}
+
+		var clientSecret corev1.Secret
+		if err := p.Client.Get(ctx, client.ObjectKey{Namespace: p.Identity.Spec.ClientSecret.Namespace, Name: p.Identity.Spec.ClientSecret.Name}, &clientSecret); err != nil {
+			return nil, err
+		}
+
+		clientSecretVal, ok := clientSecret.StringData["clientSecret"]
```
CecileRobertMichon, line 177: I see some reusable code with #1523 👀

shysank: Yes, many of the things from #1523 can be used here. I'll  refactor it once it's merged.

docs/book/src/topics/multitenancy.md
```diff
@@ -49,6 +49,27 @@ data:
...
+spec:
+  type: ManualServicePrincipal
+  tenantID: <azure-tenant-id>
+  clientID: <client-id-of-SP-identity>
+  clientSecret: {"name":"<secret-name-for-client-password>","namespace":"default"}
+  allowedNamespaces:
+    list:
+    - <cluster-namespace>
+```
+The rest of the configuration is the same as that of service principal identity. This useful in scenarios where you don't want to have a dependency on [aad-pod-identity](https://azure.github.io/aad-pod-identity).
```
CecileRobertMichon, line 71: Together with this change, we should also think about how to make installing aad pod identity optional for users who don't want it / already have it (#1545). I currently don't have any good ideas on how to do this since it's part of the release manifest...

shysank: Can we make the number of replicas for add pod identity deployment as an environment variable and use it as a proxy for disabling aad pod identity? It's still kind of a hack since the azure binding crds will be installed.

CecileRobertMichon (commented):
/cc @devigned

devigned (commented):
We should discuss this PR in the next CAPZ office hours. The implementation looks solid, but it seems like there are some wider questions around aad-pod-identity and use case.

CecileRobertMichon (commented):
/lgtm
/approve

shysank:
Created issue for optional nmi installation as discussed in last office hours meeting: https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1604

Issue #1521, opened 2021-07-13
Add support for cert/pwd auth for cloud provider azurejson


CecileRobertMichon:
/kind feature

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]

When the AzureClusterIdentity uses a Certificate & password, it should be able to pass those to the azurejson secret for cloud provider auth if no managed identity is specified.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- cluster-api-provider-azure version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

CecileRobertMichon:
The issue I'm running into here is that cloud provider only accepts 

```
	// The path of a client certificate for an AAD application with RBAC access to talk to Azure RM APIs
	AADClientCertPath string `json:"aadClientCertPath,omitempty" yaml:"aadClientCertPath,omitempty"`
```

which is the _filepath_ to a certificate file, which means that the file must be not only written to the VM but also mounted on the CCM container. I'm unable to do this from the CAPZ azurejson controller.

alexeldeib:
Volume mount the host path to ccm? Doesn’t it require a patch on ccm no matter what for the volume mount? I realize that means cluster template changes, but I think it’s reasonable to ask for that if you use this functionality

CecileRobertMichon:
> Doesn’t it require a patch on ccm no matter what for the volume mount? I realize that means cluster template changes, but I think it’s reasonable to ask for that if you use this functionality

It does, and yes that might be what have to end up doing but this is a step down from the original ask of getting the cert from the AzureClusterIdentity and putting it in the cloud provider config (like SP client secret). The other problem is that I have no good way of knowing which filepath it's at since the user can put it wherever they want so it's tricky to set the cloud provider config `AADClientCertPath` value for them.

CecileRobertMichon:
This is what I came up with: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1531/files#diff-2ecc665ce49ded769d9199f6979a596aaeb1554236e93cfbf7524596095ce359R126

PR #1523, opened 2021-07-13
Set Service Principal in azurejson secret when AzureCluster has IdentityRef but not using Managed Identity

### LINKED ISSUE #1517: azurejson controller doesn't have access to ClientID/Secret/TenantID when using AzureClusterIdentity
/kind bug

**What did you expect to happen:** The azure.json controller doesn’t have access to the Client secret/id/tenant ID when using AzureClusterIdentity and the manager creds aren’t set. This causes the azurejson secrets to be generated without aadClientID and aadClientSecret set, when not using managed identity, which causes cloud provider to fail (kube-controller-manager goes into crashloop).

Workaround is to keep setting manager credentials (base 64 encoded env vars), or to use managed identity.

Potential fixes to consider (just spitballing):

- Find a way to give access to the creds in the controller through aad-pod-identity
- create an identity for the machine (?)
- make managed system identity default and require env vars for SP cloud provider identity *but this requires the AzureClusterIdentity SP to have Owner access*
- ???

**Environment:**

CecileRobertMichon:
**What type of PR is this?**

**What this PR does / why we need it**: The azure.json controller didn't have access to the Client secret/id/tenant ID when using AzureClusterIdentity and the manager creds were not set. This caused the azurejson secrets to be generated without aadClientID and aadClientSecret set, when not using managed identity, which causes cloud provider to fail (kube-controller-manager goes into crashloop).

Also adds a warning when not using Managed Identity since it is less secure and requires writing creds to disk.

To ensure the integrity of these changes, the tilt, e2e and conformance scripts have been modified to not set base64 encoded variables so that the manager credentials are no longer set (this is what prevented us from catching this bug in the first place). Intentionally left ci-entrypoint untouched, will follow up on that one later.

Fixes #1517

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Get Cloud Provider config Service Principal credentials from AzureClusterIdentity when not using Managed Identity 
```

### DIFF
--- a/azure/scope/identity.go
+++ b/azure/scope/identity.go
@@ -21,6 +21,8 @@ import (
 	"fmt"
 	"reflect"
 
+	"k8s.io/apimachinery/pkg/types"
+
 	"github.com/Azure/go-autorest/autorest"
 	"github.com/Azure/go-autorest/autorest/adal"
 	"github.com/pkg/errors"
@@ -38,9 +40,14 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const azureSecretKey = "clientSecret"
+
 // CredentialsProvider defines the behavior for azure identity based credential providers.
 type CredentialsProvider interface {
 	GetAuthorizer(ctx context.Context, resourceManagerEndpoint string) (autorest.Authorizer, error)
+	GetClientID() string
+	GetClientSecret(ctx context.Context) (string, error)
+	GetTenantID() string
 }
 
 // AzureCredentialsProvider represents a credential provider with azure cluster identity.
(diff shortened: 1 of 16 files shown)


nader-ziada:
lint errors
```
azure/scope/identity.go:22: File is not `goimports`-ed (goimports)
	"k8s.io/apimachinery/pkg/types"
azure/scope/identity.go:250:1: comment on exported method `AzureCredentialsProvider.GetTenantID` should be of the form `GetTenantID ...` (golint)
// GetClientID returns the Tenant ID associated with the AzureCredentialsProvider's Identity.
^
```

CecileRobertMichon:
I expect the private cluster and self-hosted specs to fail because of known Calico issue identified in #1360 but running all the tests to make sure nothing else fails before I re-add the env vars for those two

CecileRobertMichon:
/assign @devigned @nader-ziada @shysank 

ready for review

nader-ziada:
great PR, will just wait and see what happens with the tests

controllers/azurejson_machinetemplate_controller.go
```diff
@@ -145,6 +146,14 @@ func (r *AzureJSONTemplateReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		userAssignedIdentityIfExists = azureMachineTemplate.Spec.Template.Spec.UserAssignedIdentities[0].ProviderID
 	}
 
+	if azureMachineTemplate.Spec.Template.Spec.Identity == infrav1.VMIdentityNone {
+		warningMessage := "You are using Service Principal authentication for Cloud Provider Azure which is less secure than Managed Identity."
```
devigned, line 150: Would you consider moving this into a const?

devigned (changes requested):
Just a consideration request for moving the warning message into a const. Outside of that, the pr lgtm.

nader-ziada:
The two failures here: `pull-cluster-api-provider-azure-e2e-full` don't seem to be related to this PR, the `GPU` and `AKS identity` have been know to have flakes

azure/scope/identity.go
```diff
@@ -219,6 +226,33 @@ func (p *AzureCredentialsProvider) GetAuthorizer(ctx context.Context, resourceMa
...
 }
 
+// GetClientID returns the Client ID associated with the AzureCredentialsProvider's Identity.
+func (p *AzureCredentialsProvider) GetClientID() string {
+	return p.Identity.Spec.ClientID
+}
+
+// GetClientSecret returns the Client Secret associated with the AzureCredentialsProvider's Identity.
+// NOTE: this only works if the Identity references a Service Principal Client Secret.
+// If using another type of credentials, such a Certificate, we return an empty string.
```
shysank, line 236: Clarification: How does cloud provider get the credentials when using other type of credentials (eg. certificate)? Is it expected to use a managed identity for such cases?

shysank (commented):
lgtm with one clarification

azure/scope/identity.go
nader-ziada: The answer to your question is coming here: https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1521

devigned (approved):
/lgtm

Issue #1481, opened 2021-06-30
AKS multi-tenancy e2e test flaky 


nader-ziada:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]

**What did you expect to happen:**
- the e2e test `Creating an AKS cluster using a different SP identity` has been failing a few times in the last few days when running e2e tests even if the change is unrelated to ManagedControlPlane or the exp/ folder
- Can't find anything obvious in the logs

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- cluster-api-provider-azure version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

alexeldeib:
/assign 

there are some minor fixes we can make in the reconciler to avoid extra requeues, but the most frequent failure I've found seems to be related to list VMSS returning stale data for a long time, and I'm investigating that.

alexeldeib:
ref: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1488

PR #1540, opened 2021-07-19
Make webhook registration accept v1 admission

### LINKED ISSUE #1528: Webhook registration should accept both v1 and v1beta1 admissions
Admission Review was previously set to only accept v1beta1 objects given
that Controller Runtime didn't support v1 (even though the two revisions
are effectively the same). We should add support for both within our
manifests before k8s 1.22.

We should implement the equivalent of https://github.com/kubernetes-sigs/cluster-api/pull/4942 for CAPZ webhhooks. 

To do so, we need to edit the kubebuilder annotations in each _webbook.go file and run `make generate` to update manifests.

/help
/good-first-issue
/priority-important-soon

Marking as good-first-issue because it's a small scope, well defined change but feel free to assign even if this isn't your first issue.

Evalle:
**What type of PR is this?**
/kind bug

Makes webhook registration accept v1 and v1beta1 admissions

Fixes #1528

**Release note**:

```release-note
Webhook registration accepts both v1 and v1beta1 admissions
```

### DIFF
--- a/api/v1alpha4/azurecluster_webhook.go
+++ b/api/v1alpha4/azurecluster_webhook.go
@@ -37,8 +37,8 @@ func (c *AzureCluster) SetupWebhookWithManager(mgr ctrl.Manager) error {
 		Complete()
 }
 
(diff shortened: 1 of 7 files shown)


CecileRobertMichon:
/ok-to-test

thanks for the PR @Evalle

CecileRobertMichon:
exp failure is a known flake (#1481)

CecileRobertMichon (commented):
/lgtm
/approve

PR #1541, opened 2021-07-20
Add proposal for Async Azure Resource Creation and Deletion

CecileRobertMichon:
**What type of PR is this?**

**What this PR does / why we need it**: This is a proposal to formalize some of the work that was already done in #1067, and to extend it to other Azure PUT/DELETE operations in the codebase. Working on a POC in parallel but opening this up to get some feedback asynchronously (😉 ).

 Credit goes to @devigned for kickstarting this effort with AzureMachinePools, and to @shysank for reflections in #1181.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```


docs/proposals/20210716-async-azure-resource-creation-deletion.md
```diff
@@ -0,0 +1,224 @@
...
+
+* Add an `AzureClientTimeout` which is the duration after which an Azure operation is considered a long running operation which should be handled asynchronously. Proposed starting value is 5 seconds.
+* For each Azure API call which returns a Future, wait for the operation to be completed for the above timeout duration. If the operation is not completed within the timeout duration, set Future of that resource in `LongRunningOperationStates` with the marshalled future data.
+* Set conditions at the end of each controller loop that describe the current state of the object and its associated Azure resources.
+
+For each Azure service, this is what the new asynchronous reconcile and delete flows will look like:
+
+##### Reconcile
+
+![Figure 1](./images/async-reconcile.png)
```
CecileRobertMichon, line 188: make sure that you open up the file to view the diagrams, a lot of the proposal subject matter is in those!

https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/f8276eec74b5bcea3756092f105131639c9a6874/docs/proposals/20210716-async-azure-resource-creation-deletion.md#reconcile

docs/Makefile
```diff
@@ -0,0 +1,36 @@
...
+.PHONY: diagrams
+diagrams: $(DIAGRAMS)
+
+%.png: %.plantuml
+	docker run \
+		--rm \
+		--volume ${ROOT_DIR}:/workdir$(DOCKER_VOL_OPTS) \
+		--user $(shell id -u):$(shell id -g) \
+		k8s.gcr.io/cluster-api/plantuml:1.2019.6 \
+		-v /workdir/$(shell echo '$^' | sed -e 's,.*docs/,,g' )
```
devigned, line 36: newline

docs/proposals/20210716-async-azure-resource-creation-deletion.md
```diff
@@ -0,0 +1,224 @@
...
+    // next reconciliation loop.
+    // +optional
+    LongRunningOperationStates Futures `json:"longRunningOperationState,omitempty"`
+}
+
+```
+
+#### <a name='ProposedControllerChanges'></a>Proposed Controller Changes
+
+* Add an `AzureClientTimeout` which is the duration after which an Azure operation is considered a long running operation which should be handled asynchronously. Proposed starting value is 5 seconds.
```
devigned, line 180: Would this timeout be for each call to Azure, the entire reconciliation loop, or a smaller scope, like all of the calls to Azure excluding other activities in the reconciliation loop? The reason I ask is that if the reconciliation loop contains multiple serial Azure requests, it's conceivable that the whole loop could take much longer than 5 seconds. Say there are 4 resources which take 5 seconds a piece, 5s + 5s + 5s + 5s = 20s, as a worse case it would not be very responsive. Also, a long running reconciliation loop including activities other than just Azure requests might not complete in 5s and end up in a state where it can no longer make progress. Thoughts?

docs/proposals/20210716-async-azure-resource-creation-deletion.md
devigned: Would be interesting to add a diagram for something like AzureCluster with multiple Azure resources.

devigned (commented):
Just have a couple comments and a small nit. I really like the tie between long running operation state and conditions, though it might be useful to provide a little more perscriptive guidance on whether or not a long running operation would tie to a given condition. For example, would a condition state be tied 1 to 1 with a long running operation and would that state change with the state of a long running operation.

Great work! This has the potential to make a significant improvement in the reactivity of the system and the experience of a user interacting with the controller.

docs/proposals/20210716-async-azure-resource-creation-deletion.md
CecileRobertMichon: that's a good question; my original thinking is that the 5s timeout, which is quite aggressive, would be per operation (each call). Then we'd have another overall reconcile loop timeout. Currently, that timeout is set at 90 minutes (which is also the ARM timeout) but I think with this proposal we could bring that down to something more responsive, like 30s.

The specific numbers are not set in stone, I need to do some performance testing with different values and calculate the P99 expected completion time of operations that are not long-running. 

The other question is whether we should have the same timeout value for all operations (the 5s) or curate per operation. For simplicity, I think we should start with a single value. Later on, we might want to optimize by calculating a dynamic timeout value for each operation based on heuristics. That would be better IMO than statically defining artificial timeout durations for each operation which might vary over time and might not be the same depending on region, subscription, etc.

CecileRobertMichon:
> though it might be useful to provide a little more perscriptive guidance on whether or not a long running operation would tie to a given condition. For example, would a condition state be tied 1 to 1 with a long running operation and would that state change with the state of a long running operation.

Will add. My current thinking is that conditions should _not_ be tied 1 to 1 with a long running operation, although we might have conditions that summarize the state of 1 or more resources created/deleted via long running operations. So long-running operations conditions might be a subset of all conditions.

I'll think about this further and come up with an initial list of added conditions.

docs/proposals/20210716-async-azure-resource-creation-deletion.md
```diff
@@ -0,0 +1,365 @@
...
+
+The existing conditions before this proposal can be seen [here](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/v0.5.0/api/v1alpha4/conditions_consts.go). Note that these existing conditions will be left unchanged, and are purposeffully left out below.
+
+Part of the proposed changes is to add new conditions for Azure CRDs. More granular conditions, paired with more responsive controllers, will allow for better visibility into the state of each resource. Initially, the following conditions will be added:
+
+```go
+// Azure Services Conditions and Reasons.
+const (
+    // ResourceGroupReadyCondition means the resource group exists and is ready to be used.
+    ResourceGroupReadyCondition clusterv1.ConditionType = "ResourceGroupReady"
```
CecileRobertMichon, line 206: I'm not sure if "Ready" is the right word. Other options I considered are "Created", "Succeeded", and "Running". I thought ready was a good choice because it also applies for BYO resource cases where the resource doesn't actually get created by the controller. "Running" also seems awkward since some of these resources don't actually "run". Thoughts?

docs/proposals/20210716-async-azure-resource-creation-deletion.md
```diff
@@ -0,0 +1,365 @@
...
+)
+```
+
+#### <a name='OpenQuestions'></a>Open Questions
+
+##### <a name='OpenQuestion1'></a>1. What would the scope of the timeout be? What should the timeout duration be?
+
+The 5s timeout, which is quite aggressive, would be per operation (each call). Then we'd have another overall reconcile loop timeout, which already exists. Currently, that timeout is set at 90 minutes (which is also the ARM deployment timeout). With this proposal, we could bring that overall timeout down to something more responsive, such as 30 seconds.
+
+The specific numbers are not set in stone, and should be revised after doing some performance testing with different values and calculating the P99 expected completion time of operations that are not long-running.
```
CecileRobertMichon, line 331: I'm expecting the numbers to become more aggressive with some testing, maybe closer to 2s per call / 15 seconds overall

docs/proposals/20210716-async-azure-resource-creation-deletion.md
CecileRobertMichon: Added some more details about this in the Context Timeouts section of proposed changes, and an open question

docs/proposals/20210716-async-azure-resource-creation-deletion.md
CecileRobertMichon: Added, PTAL

devigned: In the diagram, "UpdateStatus" state is the only one that updates the conditions. I think "StoreLongRunningState" should probably progress to "UpdateStatus", and "UpdateStatus" probably should not set long running operation state to nil. Perhaps consider adding a state between the CreateOrUpdate state, on the success side, the long running state is set to nil, but on the context expired side, keep it the same, but progress to the "UpdateStatus" state.

docs/proposals/20210716-async-azure-resource-creation-deletion.md
devigned: I like "Ready". It's convenient to remember as it's a common to many existing resources. It might not be 100% correct, but I think it conveys the right idea.

devigned: That's a lot of conditions. I wonder if there is some way we can remove toil in describing / setting these conditions. Do you think we could name them through convention rather than declaring all of them, or do you think that would take away from discoverability?

devigned (commented):
I like the updates. This is looking really solid. I'm added a couple comments and questions.

docs/proposals/20210716-async-azure-resource-creation-deletion.md
CecileRobertMichon: I agree it takes a lot of space in the proposal... but in the actual code it will be in its own constants file and I do think it's worth listing them explicitly. In the proposal I can simplify by just defining the pattern and providing an example or two, would that be better?

docs/proposals/20210716-async-azure-resource-creation-deletion.md
CecileRobertMichon: That makes sense to me

docs/proposals/20210716-async-azure-resource-creation-deletion.md
devigned: Nope. I just wanted us to discuss. I'm 100% good with listing them in the proposal and including the consts in the code.

docs/proposals/20210716-async-azure-resource-creation-deletion.md
```diff
@@ -0,0 +1,382 @@
...
+
+### <a name='UserStories'></a>User Stories
+
+#### <a name='Story1-UXofcreatinganAzureCluster'></a>Story 1 - UX of creating an AzureCluster
+
+Blake is a Program Manager trying out Cluster API for the first time. Blake is following the quickstart documentation in the Cluster API book and using Azure to create a cluster. Blake applies the cluster template on the management cluster and describes the resulting AzureCluster resource. The AzureCluster is in "Creating" state and the Conditions get updated as Azure resources are created to show the progress.
+
+#### <a name='Story2-ScalingupaMachineDeployment'></a>Story 2 - Scaling up a MachineDeployment
+
+Alex is an engineer in a large organization which has a MachineDeployment running. Alex needs to scale up the number of replicas of the MachineDeployment. Alex uses `kubectl` to scale the number of replicas in the MachineDeployment by two. Two new AzureMachines are created and their state quickly becomes "Creating". Alex checks the Conditions on the MachineDeployment and sees that the network interface was created successfully, and that the VM is being created.
```
CecileRobertMichon, line 90: note to self: add a story for scaling up many machines concurrently

docs/proposals/20210716-async-azure-resource-creation-deletion.md
```diff
@@ -0,0 +1,382 @@
...
+    // RoleAssignmentCreatingReason means the role assignment is being created.
+    RoleAssignmentCreatingReason clusterv1.ConditionType = "RoleAssignmentCreating"
+    // RoleAssignmentDeletingReason means the role assignment is being deleted.
+    RoleAssignmentDeletingReason clusterv1.ConditionType = "RoleAssignmentDeleting"
+    // RoleAssignmentFailedCondition means the role assignment failed to be created.
+    RoleAssignmentFailedCondition clusterv1.ConditionType = "RoleAssignmentFailed"
+)
+```
+
+### <a name='OpenQuestions'></a>Open Questions
```
CecileRobertMichon, line 344: Another question: how do we handle reconciler error so that transient "operation not done" errors, which are fully expected in this new model, do not pollute manager logs?

CecileRobertMichon: added

docs/proposals/20210716-async-azure-resource-creation-deletion.md
CecileRobertMichon: added

docs/proposals/20210716-async-azure-resource-creation-deletion.md
```diff
@@ -0,0 +1,399 @@
...
+- Make it easier for the user to understand the state of each resource by adding more granular Conditions
+- Apply the same asynchronous pattern to all resources
+
+### <a name='Non-GoalsFutureWork'></a>Non-Goals / Future Work
+
+- Increase or decrease overall duration of reconciliation
+- Increase the number of API calls to Azure
+- Start Azure operations in parallel
+- Predict how long each operation will take
+- Apply the same changes to the managed cluster (AKS) reconciler
```
CecileRobertMichon, line 78: have to do this for groups/vnets/subnets so might have to anyways

docs/proposals/20210716-async-azure-resource-creation-deletion.md
CecileRobertMichon: I actually ended up simplifying them by reusing the same Reasons across conditions.

docs/proposals/20210716-async-azure-resource-creation-deletion.md
```diff
@@ -0,0 +1,319 @@
...
+```go
+// Future contains the data needed for an Azure long-running operation to continue across reconcile loops.
+type Future struct {
+    // Type describes the type of future, update, create, delete, etc
+    Type string `json:"type"`
+    // ResourceGroup is the Azure resource group for the resource
+    // +optional
+    ResourceGroup string `json:"resourceGroup,omitempty"`
+    // Name is the name of the Azure resource
+    Name string `json:"name"`
```
shysank, line 141: Is this going to be used to identify the state of a long running service? As in, if there is a service with 3 resources, there will be 3 future objects with the name being the unique identifier. Is my understanding correct? 
If that's the case, should we also add a service name so that we can filter by service? It might be useful when we are updating conditions.

shysank (commented):
Thanks @CecileRobertMichon for formalizing the ideas. Looks good overall with a couple of nits/clarifications.

docs/proposals/20210716-async-azure-resource-creation-deletion.md
CecileRobertMichon: >  As in, if there is a service with 3 resources, there will be 3 future objects with the name being the unique identifier. Is my understanding correct?

Correct

> If that's the case, should we also add a service name so that we can filter by service? It might be useful when we are updating conditions.

that's a good idea, will add.

docs/proposals/20210716-async-azure-resource-creation-deletion.md
```diff
@@ -0,0 +1,399 @@
...
+    // FutureData is the base64 url encoded json Azure AutoRest Future
+    FutureData string `json:"futureData,omitempty"`
```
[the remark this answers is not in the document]
CecileRobertMichon: @devigned thoughts on this one?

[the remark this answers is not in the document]
CecileRobertMichon: nvm, just realized this is Future.FutureData, applied the change

docs/proposals/20210716-async-azure-resource-creation-deletion.md
CecileRobertMichon: removed

CecileRobertMichon:
There haven't been any comments on this for a while, I'd like to start a lazy consensus for end of next week if no other comments get added by then.

/hold for lazy consensus

/cc @jackfrancis @alexeldeib @mboersma in case you haven't had a chance to review this yet

devigned (approved):
lgtm. Great work!

nader-ziada:
a little bit late, but also lgtm 🎉

shysank:
lgtm from me as well, apologies for the delayed response.

docs/proposals/20210716-async-azure-resource-creation-deletion.md
```diff
@@ -0,0 +1,320 @@
...
+
+<!-- vscode-markdown-toc-config
+	numbering=false
+	autoSave=false
+	/vscode-markdown-toc-config -->
+<!-- /vscode-markdown-toc -->
+
+## <a name='Summary'></a>Summary
+
+CAPZ reconcilers currently call Azure and wait for each operation before proceeding. We should create/update/delete Azure resources asynchronously, especially for operations that take a long time to complete, such as Virtual Machine creation and deletion.
```
jackfrancis, line 58: Maybe this introductory statement ("CAPZ reconcilers currently call...") could be expanded to make it more clear exactly how bad things are. Maybe:

"Currently, when CAPZ reconciles any Azure resource it does so in a bulk, serialized operation, with the practical effect that reconciling many resources at once results takes a long time, as each operation blocks the next operation."

I can see that this isn't easy to formulate. :/

In any event, my 2¢ is that the current sentence understates things a bit.

docs/proposals/20210716-async-azure-resource-creation-deletion.md
```diff
@@ -0,0 +1,320 @@
...
+
+## <a name='Motivation'></a>Motivation
+
+Blocking on success is sometimes the right thing to do but most of the time, it's the equivalent of the UI freezing on an app because you have used the UI thread to fetch some data causing your user to wonder why and when the software will react. This proposal aims to make the reaction time of the CAPZ controller drastically faster and possibly, more resilient.
```
jackfrancis, line 62: I think we can remove the "possibly" part at the end of this paragraph. Doing these operations discretely and asynchronously should definitely add resiliency. In particular, operations that are TTL-bound will be less likely to encounter any upper bound timeout due to the fact that we are no longer doing a bunch of serialized operations inside a common context+timeout.

docs/proposals/20210716-async-azure-resource-creation-deletion.md
```diff
@@ -0,0 +1,320 @@
...
+
+### <a name='Goals'></a>Goals
+
+- Accelerate the feedback loop with the user so they can know that reconciliation is progressing without having to go check the Azure resources in the portal/CLI/etc.
```
jackfrancis, line 66: Do we not currently publish cluster-api-specific debug data about reconciliation progression? Is this because we don't have a good way to do that on a per-resource basis, and/or that we think it's clumsy to report progress for a sort of artificially serialized "set of recources" reconciliation operation?

Basically, is this feedback loop goal related to doing per-resource reconciliation? Or is it a discrete goal that shares a common surface area with the per-resource. stuff?

docs/proposals/20210716-async-azure-resource-creation-deletion.md
```diff
@@ -0,0 +1,320 @@
...
+- Make the time for the controller to react to a change much faster
+- Improve the resiliency of the controller by making it more fault tolerant
```
jackfrancis, line 68: Is the additional fault tolerance because we don't have to fail a bunch of resource reconciliations if just one resource reconciliation fails?

docs/proposals/20210716-async-azure-resource-creation-deletion.md
jackfrancis: Based on this Goal below:

"Apply the same asynchronous pattern to all resources"

I am inferring that the above "serial reconciliation" thing is not 100% consistent, and that some resources *do* reconcile discretely and asynchronously. If that's true, then we should probably also include that in the summary as a secondary foundational problem. Like, the fact that we don't implement reconciliation according to a common, consistent strategy/pattern is itself a problem (separate from the problem of picking the best strategy/pattern).

docs/proposals/20210716-async-azure-resource-creation-deletion.md
```diff
@@ -0,0 +1,320 @@
...
+- Make it easier for the user to understand the state of each resource by adding more granular Conditions
+- Apply the same asynchronous pattern to all resources
+
+### <a name='Non-GoalsFutureWork'></a>Non-Goals / Future Work
+
+- Increase or decrease overall duration of reconciliation
```
jackfrancis, line 74: I'm not sure that "decreasing the overall reconciliation time" should be a non-goal. If we make the above improvements and we don't observe quicker reconciliations, that would be very interesting and perplexing (it basically wouldn't make sense), and I think we would want to take a step back and pause if so. Basically, I think we should make "decreasing average reconciliation operations" a core goal here.

docs/proposals/20210716-async-azure-resource-creation-deletion.md
```diff
@@ -0,0 +1,320 @@
...
+- Increase the number of API calls to Azure
+- Start Azure operations in parallel
```
jackfrancis, line 76: Because this work will necessarily introduce more operational parallelism, perhaps we want to say "Optimize parallel Azure operations" here? I think what we mean to say is that we don't want to focus on making the parallel implementations "perfect" in this initial workstream. Is that right?

docs/proposals/20210716-async-azure-resource-creation-deletion.md
```diff
@@ -0,0 +1,320 @@
...
+- Accelerate the feedback loop with the user so they can know that reconciliation is progressing without having to go check the Azure resources in the portal/CLI/etc.
+- Make the time for the controller to react to a change much faster
+- Improve the resiliency of the controller by making it more fault tolerant
+- Make it easier for the user to understand the state of each resource by adding more granular Conditions
+- Apply the same asynchronous pattern to all resources
+
+### <a name='Non-GoalsFutureWork'></a>Non-Goals / Future Work
+
+- Increase or decrease overall duration of reconciliation
+- Increase the number of API calls to Azure
```
jackfrancis, line 75: Do we mean to say here "Optimize for the fewest required Azure API calls" or something like that?

docs/proposals/20210716-async-azure-resource-creation-deletion.md
```diff
@@ -0,0 +1,320 @@
...
+
+#### <a name='Story2-ScalingupaMachineDeployment'></a>Story 2 - Creating AzureMachines concurrently
+
+Alex is an engineer in a large organization which has a MachineDeployment running. Alex needs to scale up the number of replicas of the MachineDeployment by 200. Alex uses `kubectl` to scale the number of replicas in the MachineDeployment by 200. The AzureMachine controller in the management cluster is running with the default concurrency of 10. Ten new AzureMachines are created and their state quickly becomes "Creating". Shortly after, before the first ten machines are done creating, ten new ones start creating. The same thing happens until all 200 AzureMachines are in "Creating" state. Alex checks the Conditions on one of the creating AzureMachines and sees that the network interface was created successfully, and that the VM is being created. This allows Alex to quickly scale up the number of replicas as the new 200 VMs get created concurrently, without having to increase the controller concurrency.
```
jackfrancis, line 89: I don't understand this scenario. It seems that with the additional decomposition and asynchronicity of Azure resource reconciliation we are saying that the AzureMachine concurrency configuration is no longer operable?

docs/proposals/20210716-async-azure-resource-creation-deletion.md
```diff
@@ -0,0 +1,320 @@
...
+    // LongRunningOperationState saves the state for an Azure long-running operations so it can be continued on the
+    // next reconciliation loop.
+    // +optional
+    LongRunningOperationState *infrav1.Future `json:"longRunningOperationState,omitempty"`
+}
+```
+
+### <a name='ProposedAPIChanges'></a>Proposed API Changes
+
+The proposed changes below show the changes to AzureCluster, AzureMachine, AzureMachinePool, and AzureMachinePoolMachine. The existing `LongRunningOperationState` field in AzureMachinePoolStatus will be pluralized to `LongRunningOperationStates` to store a list of Futures, following a similar pattern than Conditions, and will be extended to other CAPZ CRDs. In addition, the `Name` field of the `Future` type will be made required, as it becomes the identifier for a Future.
```
jackfrancis, line 129: Can we add the proposed changes to `AzureMachinePoolStatus` (array of futures instead of single future) into the explicitly stated struct set below?

docs/proposals/20210716-async-azure-resource-creation-deletion.md
```diff
@@ -0,0 +1,320 @@
...
+    LongRunningOperationStates Futures `json:"longRunningOperationStates,omitempty"`
+}
+
+```
+
+### <a name='ProposedControllerChanges'></a>Proposed Controller Changes
+
+#### <a name='Contexttimeouts'></a>Context timeouts
+
+* Reduce the global controller reconcile loop context timeout to 15 seconds (currently 90 minutes).
```
jackfrancis, line 203: Can you explain this timeout reduction?

docs/proposals/20210716-async-azure-resource-creation-deletion.md
```diff
@@ -0,0 +1,320 @@
...
+)
+```
+
+### <a name='OpenQuestions'></a>Open Questions
+
+#### <a name='Whatshouldthetimeoutdurationsbe'></a>1. What should the timeout durations be?
+
+The specific numbers are not set in stone, and should be revised after doing some performance testing with different values and calculating the P99 expected completion time of operations that are not long-running.
+
+The other question is whether we should have the same timeout value for all operations (the 5s) or curate per operation. For simplicity, the proposal is to start with a single value. Later on, we might want to optimize by calculating a dynamic timeout value for each operation based on heuristics. That would be better than statically defining artificial timeout durations for each operation which might vary over time and might not be the same depending on region, subscription, etc.
```
jackfrancis, line 284: The reason to define a common timeout is that this timeout in practice is really just the timeout for getting a response from the Azure API. Is that correct?

If so, we want to make sure there are go Azure API GET calls that don't properly paginate when returning a set (or, probably more importantly, if the underlying resources themselves don't paginate their LIST responses). So long as Azure APIs return paginated sets generally, we should be able to make practical assumptions (e.g., 5 seconds) about how long before we want to timeout the API response. If certain APIs, however, don't paginate responses for LIST requests, then we could expect that the response time could scale with the number of items in the response.

docs/proposals/20210716-async-azure-resource-creation-deletion.md
```diff
@@ -0,0 +1,320 @@
...
+
+#### 2. How to handle transient errors in logs?
+
+The idea of short-circuiting the Reconcile loop when a long-running operation is in progress involves returning an error when an operation is not done. This means that the Reconcile loop will end in an error every time an operation is in progress. This is necessary because we need to requeue so that the reconcile loop can run again to check on the progress of the operation, but it also means that the user might will see the error message in the logs. How can we handle transient errors in logs without spamming the logs and therefore causing noise that reduces the user's ability to see actual errors in reconcile?
```
jackfrancis, line 288: Why are we returning an error? Is this because it's a well-established implementation pattern that we trust? It's not clear why we would *have* to represent a non-terminal response state as an error.

docs/proposals/20210716-async-azure-resource-creation-deletion.md
CecileRobertMichon: This proposal does _not_ aim to reconcile resources in a non-serial (parallel) way. See the Alternatives section where I went into why. Each operation would _still_ block the next operation. This proposal aims to change the fact that a single controller loop right can take up to 90 minutes.

CecileRobertMichon: > I am inferring that the above "serial reconciliation" thing is not 100% consistent, and that some resources do reconcile discretely and asynchronously.

That's not true. In AzureCluster/AzureMachine controllers, all resources are currently reconciling resources synchronously and serially. This goal highlights that we want to change all of them.

docs/proposals/20210716-async-azure-resource-creation-deletion.md
CecileRobertMichon: The keyword here is accelerate. We already have Status and Conditions and different ways for the user to get information about the progression but, by design, a controller only patches the object it's responsible for at the very end of the reconcile loop. That means if the controller is reconciling for 20 minutes, the user won't see any updates on the custom resource for 20 minutes.

docs/proposals/20210716-async-azure-resource-creation-deletion.md
CecileRobertMichon: > we don't have to fail a bunch of resource reconciliations if just one resource reconciliation fails?

If one resource fails, we short circuit the entire loop and try again later. However, if one resource is stuck creating in Azure or is taking longer than normal to create/update/delete, we won't waste the controller loop on that resource by polling continuously.

docs/proposals/20210716-async-azure-resource-creation-deletion.md
CecileRobertMichon: That is actually a non-goal which is why it's called out explicitly here. The main goal here is to make sure that a controller loop doesn't take more than a few minutes. The overall reconcilation time of 1 AzureCluster or 1 AzureMachine won't change, but the time it takes to, say, delete 100 AzureMachines will be overall shorter (without increasing the controller concurrency) because we'll be able to handle multiple ongoing operations at once.

docs/proposals/20210716-async-azure-resource-creation-deletion.md
CecileRobertMichon: it's a non-goal to increase the overall number of API calls to Azure, ie. the number of calls shouldn't be higher than what it was before.

docs/proposals/20210716-async-azure-resource-creation-deletion.md
CecileRobertMichon: Perhaps I should clarify Azure operations for a single AzureCluster or AzureMachine. For example, we _will not_ start the vnet and security group for a single cluster in parallel.

docs/proposals/20210716-async-azure-resource-creation-deletion.md
CecileRobertMichon: The concurrency limit will still be there and will still apply, but because a single loop will take 1 minute as opposed to 5-10 minutes (average), you can queue up all 200 Azure VM creates in Azure more quickly.

docs/proposals/20210716-async-azure-resource-creation-deletion.md
jackfrancis: Thanks for the clarification!

docs/proposals/20210716-async-azure-resource-creation-deletion.md
jackfrancis: Got it.

docs/proposals/20210716-async-azure-resource-creation-deletion.md
jackfrancis: Table this comment

docs/proposals/20210716-async-azure-resource-creation-deletion.md
jackfrancis: Understood

docs/proposals/20210716-async-azure-resource-creation-deletion.md
jackfrancis: Understood

docs/proposals/20210716-async-azure-resource-creation-deletion.md
jackfrancis: Right, if there are some operations that will be unchanged (not start in parallel) it could be worth calling those out, or calling out only the operations that will be changed to start in parallel

docs/proposals/20210716-async-azure-resource-creation-deletion.md
jackfrancis: Understood

docs/proposals/20210716-async-azure-resource-creation-deletion.md
jackfrancis: Table this comment

docs/proposals/20210716-async-azure-resource-creation-deletion.md
jackfrancis: Table this comment

docs/proposals/20210716-async-azure-resource-creation-deletion.md
CecileRobertMichon: It was already there but the name of the struct was wrong, was missing "Status", fixed

docs/proposals/20210716-async-azure-resource-creation-deletion.md
CecileRobertMichon: clarified

docs/proposals/20210716-async-azure-resource-creation-deletion.md
CecileRobertMichon: it's just the way we handle transient errors right now https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/controllers/azuremachine_controller.go#L299 it gets logged as an "error" which could be misleading in this case since a resource not being done provisioning is not an error. I think I have a good idea about how to avoid that though and make sure "still ongoing" requeues don't get logged as errors.

devigned (approved):
All feedback has been addressed and we seem to have a consensus of lgtm. Great work, @CecileRobertMichon!

/lgtm
/approve

PR #1544, opened 2021-07-20
🐛managedcluster/agentpool in non-terminal state should return error

alexeldeib:
**What type of PR is this?**
/kind bug

these are errors which should trigger a requeue, there is no point not requeuing since we will never get a notification from Azure the operation is complete without polling.

related to https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1488

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/azure/services/agentpools/agentpools.go
+++ b/azure/services/agentpools/agentpools.go
@@ -18,12 +18,14 @@ package agentpools
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/Azure/azure-sdk-for-go/services/containerservice/mgmt/2020-02-01/containerservice"
 	"github.com/google/go-cmp/cmp"
 	"github.com/pkg/errors"
 	"k8s.io/klog/v2"
 
+	infrav1alpha4 "sigs.k8s.io/cluster-api-provider-azure/api/v1alpha4"
 	"sigs.k8s.io/cluster-api-provider-azure/azure"
 	"sigs.k8s.io/cluster-api-provider-azure/util/tele"
 )
@@ -79,9 +81,10 @@ func (s *Service) Reconcile(ctx context.Context, spec interface{}) error {
 		}
 	} else {
 		ps := *existingPool.ManagedClusterAgentPoolProfileProperties.ProvisioningState
-		if ps != "Canceled" && ps != "Failed" && ps != "Succeeded" {
-			klog.V(2).Infof("Unable to update existing agent pool in non terminal state.  Agent pool must be in one of the following provisioning states: canceled, failed, or succeeded")
-			return nil
+		if ps != string(infrav1alpha4.Canceled) && ps != string(infrav1alpha4.Failed) && ps != string(infrav1alpha4.Succeeded) {
+			msg := fmt.Sprintf("Unable to update existing agent pool in non terminal state. Agent pool must be in one of the following provisioning states: canceled, failed, or succeeded. Actual state: %s", ps)
+			klog.V(2).Infof(msg)
+			return errors.New(msg)
 		}
 
 		// Normalize individual agent pools to diff in case we need to update
--- a/api/v1alpha4/types.go
+++ b/api/v1alpha4/types.go
@@ -245,6 +245,8 @@ const (
 	Succeeded ProvisioningState = "Succeeded"
 	// Updating ...
 	Updating ProvisioningState = "Updating"
+	// Canceled represents an action which was initiated but terminated by the user before completion.
+	Canceled ProvisioningState = "Canceled"
 	// Deleted represents a deleted VM
 	// NOTE: This state is specific to capz, and does not have corresponding mapping in Azure API (https://docs.microsoft.com/en-us/azure/virtual-machines/states-billing#provisioning-states)
 	Deleted ProvisioningState = "Deleted"
--- a/azure/services/agentpools/agentpools_test.go
+++ b/azure/services/agentpools/agentpools_test.go
@@ -92,7 +92,7 @@ func TestReconcile(t *testing.T) {
 				Name:          "my-agentpool",
 			},
 			provisioningStatesToTest: []string{"Deleting", "InProgress", "randomStringHere"},
-			expectedError:            "",
+			expectedError:            "Unable to update existing agent pool in non terminal state. Agent pool must be in one of the following provisioning states: canceled, failed, or succeeded. Actual state: randomStringHere",
 			expect: func(m *mock_agentpools.MockClientMockRecorder, provisioningstate string) {
 				m.Get(gomockinternal.AContext(), "my-rg", "my-cluster", "my-agentpool").Return(containerservice.AgentPool{ManagedClusterAgentPoolProfileProperties: &containerservice.ManagedClusterAgentPoolProfileProperties{
 					ProvisioningState: &provisioningstate,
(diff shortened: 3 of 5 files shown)


azure/services/agentpools/agentpools.go
```diff
@@ -80,8 +81,9 @@ func (s *Service) Reconcile(ctx context.Context, spec interface{}) error {
 	} else {
 		ps := *existingPool.ManagedClusterAgentPoolProfileProperties.ProvisioningState
 		if ps != "Canceled" && ps != "Failed" && ps != "Succeeded" {
-			klog.V(2).Infof("Unable to update existing agent pool in non terminal state.  Agent pool must be in one of the following provisioning states: canceled, failed, or succeeded")
-			return nil
+			msg := fmt.Sprintf("Unable to update existing agent pool in non terminal state. Agent pool must be in one of the following provisioning states: canceled, failed, or succeeded. Actual state: %s", ps)
+			klog.V(2).Infof(msg)
+			return errors.New(msg)
```
devigned, line 86: Perhaps, this error should return an transient error wrapper to inform the controller to requeue after a period of time. This would cause an immediate requeue, right?

devigned (commented):
Just concerned about immediate requeue based on returning an error.

azure/services/agentpools/agentpools.go
alexeldeib: immediate requeue isn't great, i'm kind of fine with the exponential backoff though tbh. I'm fine with doing an error wrapper here for the short term, if you'd like.

but...I really don't love generating errors in low level clients which higher level controller code needs to know about. I'd almost prefer to return reconcile.Results{} all the way up/down so the lower level clients can provide useful semantics to the controllers. It's annoying if we add new errors and have to check for them in the top level, we've already had bugs in because we didn't check for `ResourceGroupNotFound/ResourceNotFound` correctly at the top level after we added the change in the lower level (which you kindly fixed) https://github.com/kubernetes-sigs/cluster-api-provider-azure/commit/f1e81c1ac979ba57c24642ce9fd8430449817e58#diff-714daa93f2eede09b5665786a53bcddf09a13a8d0a19cd17fc61e3bc017f4b8aR198

alexeldeib: fwiw, I didn't notice much negative effect in #1488 -- the timing work out pretty closely to how long AKS take to create the cluster.

alexeldeib: well, I guess this is only hit on update/delete anyway, so maybe that's not meaningful.

azure/services/agentpools/agentpools.go
```diff
@@ -80,8 +81,9 @@ func (s *Service) Reconcile(ctx context.Context, spec interface{}) error {
 	} else {
 		ps := *existingPool.ManagedClusterAgentPoolProfileProperties.ProvisioningState
 		if ps != "Canceled" && ps != "Failed" && ps != "Succeeded" {
```
CecileRobertMichon, line 83: these should use the azure consts

azure/services/agentpools/agentpools.go
CecileRobertMichon: are we already handling terminal/transient errors in the managed cluster reconciler?

alexeldeib: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/e1faf08f76253e8b4200340a41b2e26dd86867f5/exp/controllers/azuremangedcontrolplane_reconciler.go#L207-L215

only to that extent in AMCP

https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/e1faf08f76253e8b4200340a41b2e26dd86867f5/exp/controllers/azuremanagedmachinepool_reconciler.go#L145

and 404s on VMSS. 

we aren't doing anything on provisioning state at the higher levels, which is why this is a problem(/bug), because there's no signal for a requeue when provisioning state has reached a terminal state.

CecileRobertMichon: this is what I meant (and I assume what David was also referring to) by transient/terminal error handling: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/controllers/azuremachine_controller.go#L284

I don't think these are being used anywhere outside of azuremachine and azuremachinepool controllers though (even azurecluster controller doesn't have logic to handle them) so I'm good with the bug fix as-is and we can revisit introducing terminal/transient error handling in the managedcluster controllers separately. I do think it would be beneficial as @devigned pointed out to be able to not requeue immediately on these types of errors where a bit of waiting is necessary.

CecileRobertMichon:
lgtm pending the comment about constants

azure/services/agentpools/agentpools.go
alexeldeib: fixed + added "Canceled" as a terminal state

alexeldeib:
I'm just re-running to see if it's a flake, I suspect this is the same List VMSS flake but for some reason the artifacts are missing :/

alexeldeib:
fwiw, @CecileRobertMichon I'd prefer to merge this one and continue debugging the failures in #1488 once we get a clean run.

CecileRobertMichon (commented):
/lgtm
/approve
/retest

CecileRobertMichon:
`Job default/curl-to-ilb-job8uag9 failed` :(

/retest

alexeldeib:
mmm flaky, what is this, a bakery? (i'll show myself out)

alexeldeib:
same failure as the job you mentioned before :/ one more go?

alexeldeib:
hrm...https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/1544/pull-cluster-api-provider-azure-e2e/1420241139092426752/build-log.txt
```


[91m[1m• Failure [2552.306 seconds][0m
Workload cluster creation
[90m/home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/test/e2e/azure_test.go:41[0m
  [91m[1mWith 3 control-plane nodes and 2 worker nodes [It][0m
  [90m/home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/test/e2e/azure_test.go:183[0m

  [91mExpected
      <bool>: false
  to equal
      <bool>: true[0m

  /home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/test/e2e/azure_accelnet.go:93

  [91mFull Stack Trace[0m
  sigs.k8s.io/cluster-api-provider-azure/test/e2e.AzureAcceleratedNetworkingSpec(0x24f7640, 0xc0000580d0, 0xc001086cd0)
  	/home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/test/e2e/azure_accelnet.go:93 +0xfee
  sigs.k8s.io/cluster-api-provider-azure/test/e2e.glob..func1.4.5()
  	/home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/test/e2e/azure_test.go:248 +0x5a
````

alexeldeib:
awkward, that failure is on checking the NIC having AN enabled: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/test/e2e/azure_accelnet.go#L93

alexeldeib:
same failure again :/ 

alexeldeib:
[accel net failed again ](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/1544/pull-cluster-api-provider-azure-e2e/1420298333582266368):/ i'm a little lazy to investigate it here, but probably will start a round of deflaking that soon in addition to aks if no one else does first :) 

PR #1563, opened 2021-07-27
Add template for testing out of tree cloud provider with Windows 

CecileRobertMichon:
**What type of PR is this?**

**What this PR does / why we need it**: Add a template for https://github.com/kubernetes-sigs/cloud-provider-azure/issues/705 

Fixes #

**Special notes for your reviewer**: This template has both a Linux pool and a Windows pool. I'm not sure how the CCM test suite handles testing LBs for Windows nodes or if it does at all. @nilo19 any recommentations? cc @jsturtevant 

/hold

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/scripts/ci-entrypoint.sh
+++ b/scripts/ci-entrypoint.sh
@@ -85,6 +85,9 @@ select_cluster_template() {
 
     if [[ -n "${TEST_CCM:-}" ]]; then
         export CLUSTER_TEMPLATE="test/ci/cluster-template-prow-external-cloud-provider.yaml"
+        if [[ -n "${WINDOWS:-}" ]]; then
+            export CLUSTER_TEMPLATE="test/ci/cluster-template-prow-external-cloud-provider-windows.yaml"
+        fi
         # shellcheck source=scripts/ci-build-azure-ccm.sh
         source "${REPO_ROOT}/scripts/ci-build-azure-ccm.sh"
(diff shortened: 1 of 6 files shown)


templates/flavors/external-cloud-provider-windows/cloud-controller-manager.yaml
```diff
@@ -0,0 +1,198 @@
+apiVersion: v1
```
jsturtevant, line 1: should these files come from the cloud manage repository so any changes made to them would be reflected in the tests?

CecileRobertMichon: we have some slight customizations in the cloud-controller-manager container, eg. image and clusterName. Do you know if kustomize let's you take a file url as base?

Side note, this should not have been duplicated in this PR though, I copied it into the new directory by accident, it should use the same file as we were using before.

jsturtevant: This is an example of using remote resources: https://github.com/kubernetes-sigs/kustomize/blob/master/examples/loadHttp.md. Admittedly haven't tried it my self

feiskyer: have someone confirmed whether the remote resources work or not?

CecileRobertMichon: @jsturtevant I'm using the files in a ConfigMapGenerator, not as a resource, so I'm not sure I can use the remove resource there

PR #1571, opened 2021-07-30
Remove accelerated networking test

CecileRobertMichon:
**What type of PR is this?**

**What this PR does / why we need it**: After investigation, it looks like the support for accelerated networking has changed recently and is no longer per VM size. Some VM sizes are supported in some regions but not others. The only way to get support for a specific SKU in a region is to use the SKU API, which is what we use in the AzureMachine reconciler to determine defaults for accelerated networking. The current test as it is, which checks that a statically defined list of VM SKUs has accelerated networking enabled as expected, is no longer relevant in its current form. After discussion (see comments below), we decided it made more sense to remove the test altogether and rely on our existing unit tests to validate this capability. 

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```


shysank:
@CecileRobertMichon I think the intermittent failure is because accelerated networking is enabled for `Standard_D2s_v3` in some locations while disabled in others.

```
○ → az vm list-skus -l westus | jq  -c ' .[] | select(.name | contains("Standard_D2s_v3")) | .capabilities[] | select(.name | contains("AcceleratedNetworkingEnabled")) '
{"name":"AcceleratedNetworkingEnabled","value":"True"}

 2021-07-29 19:18:57 ⌚  sankarans-a01 in ~
○ → az vm list-skus -l centralus | jq  -c ' .[] | select(.name | contains("Standard_D2s_v3")) | .capabilities[] | select(.name | contains("AcceleratedNetworkingEnabled")) '
{"name":"AcceleratedNetworkingEnabled","value":"False"}
```

CecileRobertMichon:
@shysank that's very surprising, and not what the documentation says here https://docs.microsoft.com/en-us/azure/virtual-network/create-vm-accelerated-networking-cli#limitations-and-constraints and here https://docs.microsoft.com/en-us/azure/virtual-machines/dv3-dsv3-series#dv3-series

CecileRobertMichon:
FWIW the Azure CLI also uses a static list of sizes for validation (not location dependent)... https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/vm/_validators.py#L723

alexeldeib:
what @shysank is saying aligns with recent issues we've seen in AKS (I thought I dropped a comment on the issue, sorry :/)

we've been using a static list of Accelerated Networking enabled sizes for many years, and recently AKS internal e2e started failing on the same issue (regional availability of AN differed for the same size). We switched to rely on the SKU API only (we allowed fallback on misses previously) as a fix. 

we didn't get an answer from CRP/NRP about why this changed, although historically I thought the rule was generally >= 4 cores for AN, that changed long ago IIRC, and I know we've used this size in many regions before.

alexeldeib:
However, in AKS we started to see errors like VMSSAcceleratedNetworkingNotSupported or something like that failing with 400 at frontend validation. We never got to the point of creating VMSS and listing them out. Is it possible we're hitting a similar failure here, and somehow masking it in the tests logic?

CecileRobertMichon:
Had a chat with @alexeldeib offline, it does look like there was a change on the Azure side at some point which caused some of these regions to not support accelerated networking on these SKUs that were previously supported everywhere. This makes our e2e less than ideal and not as relevant anymore.

We might want to disable the test for now and/or change the test to something like: listing skus -> pick enabled -> manually set it to enabled -> run workload requiring that functionality

shysank:
I'm +1 to disabling/removing the spec. The alternate of listing the skus and setting the capability seems to me like we'll be just repeating the code that is there in the actual implementation.

CecileRobertMichon:
```
/home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/test/e2e/azure_test.go:96
Timed out after 300.001s.
Expect the manually created resource group is empty after removing the manually created resources.
Error: Unexpected non-nil/non-zero extra argument at index 1:
	<autorest.DetailedError>: autorest.DetailedError{Original:(*azure.RequestError)(0xc0009acab0), PackageType:"resources.Client", Method:"GetByID", StatusCode:400, Message:"Failure responding to request", ServiceError:[]uint8(nil), Response:(*http.Response)(0xc0009ac990)}
/home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/test/e2e/azure_privatecluster.go:420
```

@devigned 

/retest

alexeldeib:
> The alternate of listing the skus and setting the capability seems to me like we'll be just repeating the code that is there in the actual implementation.

yeah, agreed. I think the key thing we want is not to test azure/our listing logic so much as asserting we actually get a VM with functional AN. That's why I suggested deploying a workload which requires it -- data plane vs control plane validation, basically.

shysank:
> yeah, agreed. I think the key thing we want is not to test azure/our listing logic so much as asserting we actually get a VM with functional AN. That's why I suggested deploying a workload which requires it -- data plane vs control plane validation, basically.

Would you be okay with getting this merged so that the ci pipelines are green, and then doing the workload test as a followup?

CecileRobertMichon:
/approve

let's unblock CI first and add any new test separately since it will test something distinct

PR #1574, opened 2021-07-30
[WIP] Percolate Azure correlation IDs to REST API calls

### LINKED ISSUE #1310: Azure API requests through the SDK should use a shared correlation ID for a given reconcile loop

arschles:
/kind feature

This is a follow-on to #1460. In that PR, we added functionality to set `x-ms-correlation-id` keys on all `context.Context`s returned by new traces returned by `tele.Tracer().Start()`. That change ensured that correlation IDs were created at the root of all reconciliation operations, but those correlation IDs did not escape the running process. This change ensures that, as the correlation IDs percolate to the REST API/`autorest` layer, they are picked up and sent over the wire to the Azure API.

Fixes #1310

This is a follow-on PR to #1460, according to [this comment](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1460#issuecomment-890097762). It should not be merged before that one. I am submitting this as a draft pull request until #1460 is merged.

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation (@devigned @CecileRobertMichon are there any places that the correlation ID needs to be documented?)
- [x] adds unit tests

**Release note**:

```release-note
Sending x-ms-correlation-id values to the Azure API
```


azure/defaults.go
```diff
@@ -322,10 +324,25 @@ func UserAgent() string {
 // SetAutoRestClientDefaults set authorizer and user agent for autorest client.
 func SetAutoRestClientDefaults(c *autorest.Client, auth autorest.Authorizer) {
 	c.Authorizer = auth
+	// Wrap the original Sender on the autorest.Client c.
+	// The wrapped Sender should set the x-ms-correlation-id on the given
+	// request, then pass the new request to the underlying Sender.
+	c.Sender = autorest.DecorateSender(c.Sender, msCorrelationIDSendDecorator)
```
arschles, line 330: @devigned related to our (offline) discussion today - you pointed out that autorest client creation code is centralized around this function. I this just wraps (wraps!) the raw client's sender in code that extracts the correlation ID out of the request context, puts it into the header, and then calls through to the underlying sender. I _think_ this does the trick. See below for a related comment on tests.

azure/defaults_test.go
```diff
@@ -235,3 +240,51 @@ func TestGetDefaultUbuntuImage(t *testing.T) {
 		})
 	}
 }
+
+func TestMSCorrelationIDSendDecorator(t *testing.T) {
```
arschles, line 244: @CecileRobertMichon @devigned would you prefer that I include a test here for `SetAutoRestClientDefaults` rather than just `msCorrelationIDSendDecorator`. I didn't want to blow up the scope too much, but I'm happy to go a bit bigger in tests here if you'd prefer.

CecileRobertMichon: this is fine with me, ideally we would add both but for this specific functionality a specific test is good enough

CecileRobertMichon (commented):
/assign @mboersma

CecileRobertMichon:
@arschles sorry for the delay getting back to you on this. The PR looks good to me, are you planning to make the required changes to all controllers in the PR? I've asked @mboersma for a review as well.

The tests are failing because of lint:
```
util/tele/corr_id.go:30:63: Comment should end in a period (godot)
// context.Contexts, HTTP headers, and other similar locations
```

arschles:
>@arschles sorry for the delay getting back to you on this. 

@CecileRobertMichon not a problem. 

>The PR looks good to me, are you planning to make the required changes to all controllers in the PR?

Great to hear. The changes in those controllers were unnecessary, so I've removed them in https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1574/commits/9a04ee46d34351b6253b13139a9ab68c6854387f (I'll squash later so you can see that in isolation for now). My apologies for forgetting to take those changes out.

>I've asked @mboersma for a review as well.

Great, and howdy @mboersma!

>The tests are failing because of lint:

Fixed in https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1574/commits/86ab11369bd8d06ed3c29ca50eb042d325a05c0f

devigned (approved):
The PR lgtm. @arschles, let's pair for a minute to verify the correlation ID is being propagated and collected in ARM.

mboersma (commented):
Looks good.

I'll try this in `tilt` locally to see if I can spot the Correlation ID being propagated.

CecileRobertMichon:
@mboersma @devigned is this on hold or ready to go?

mboersma:
The observability "stack" continues to work with this change, but I can't tell by looking at the traces (in Jaeger or in App Insights) if `x-ms-correlation-id` is being set for SDK requests, but we don't capture all the HTTP headers in the traces. There aren't ARM deployments to check AFAICT, so I'm not sure how to validate this.

Deferring to @devigned...

PR #1575, opened 2021-07-30
Adding correlation ID in reconcile loop loggers

### LINKED ISSUE #1310: Azure API requests through the SDK should use a shared correlation ID for a given reconcile loop

arschles:
/kind feature

This PR pulls the correlation ID out of trace contexts and includes it in reconcile loop loggers. This is necessary so that a user having an issue can convey valuable debugging information to someone just by submitting a log. A Microsoft employee can, using that log, look up the correlation ID on Azure's backend and gain a lot more visibility into what happened and what went wrong.

Fixes #1310 

This PR is part of a 3-part series including #1460 and #1574. All are part of the fix to #1310. This should be merged only after #1460. I am marking this as a draft PR until that time.

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation (@CecileRobertMichon @devigned can you point me to any documentation that should be updated?)
- [ ] adds unit tests

**Release note**:

```release-note
Including correlation ID in all log output
```

### DIFF
--- a/azure/scope/machinepool.go
+++ b/azure/scope/machinepool.go
@@ -212,8 +212,8 @@ func (m MachinePoolScope) MaxSurge() (int, error) {
 // updateReplicasAndProviderIDs ties the Azure VMSS instance data and the Node status data together to build and update
 // the AzureMachinePool replica count and providerIDList.
 func (m *MachinePoolScope) updateReplicasAndProviderIDs(ctx context.Context) error {
-	ctx, span := tele.Tracer().Start(ctx, "scope.MachinePoolScope.UpdateInstanceStatuses")
-	defer span.End()
+	ctx, _, done := tele.StartSpanWithLogger(ctx, "scope.MachinePoolScope.UpdateInstanceStatuses")
+	defer done()
 
 	machines, err := m.getMachinePoolMachines(ctx)
 	if err != nil {
@@ -235,8 +235,11 @@ func (m *MachinePoolScope) updateReplicasAndProviderIDs(ctx context.Context) err
 }
 
 func (m *MachinePoolScope) getMachinePoolMachines(ctx context.Context) ([]infrav1exp.AzureMachinePoolMachine, error) {
-	ctx, span := tele.Tracer().Start(ctx, "scope.MachinePoolScope.getMachinePoolMachines")
-	defer span.End()
+	ctx, _, done := tele.StartSpanWithLogger(
+		ctx,
+		"scope.MachinePoolScope.getMachinePoolMachines",
+	)
+	defer done()
 
 	labels := map[string]string{
 		clusterv1.ClusterLabelName:      m.ClusterName(),
@@ -251,8 +254,11 @@ func (m *MachinePoolScope) getMachinePoolMachines(ctx context.Context) ([]infrav
 }
 
 func (m *MachinePoolScope) applyAzureMachinePoolMachines(ctx context.Context) error {
-	ctx, span := tele.Tracer().Start(ctx, "scope.MachinePoolScope.applyAzureMachinePoolMachines")
-	defer span.End()
+	ctx, _, done := tele.StartSpanWithLogger(
+		ctx,
+		"scope.MachinePoolScope.applyAzureMachinePoolMachines",
+	)
+	defer done()
 
 	if m.vmssState == nil {
 		m.Info("vmssState is nil")
@@ -493,16 +499,19 @@ func (m *MachinePoolScope) SetAnnotation(key, value string) {
 
 // PatchObject persists the machine spec and status.
 func (m *MachinePoolScope) PatchObject(ctx context.Context) error {
-	ctx, span := tele.Tracer().Start(ctx, "scope.MachinePoolScope.PatchObject")
-	defer span.End()
+	ctx, _, done := tele.StartSpanWithLogger(
+		ctx,
+		"scope.MachinePoolScope.PatchObject",
+	)
+	defer done()
 
 	return m.patchHelper.Patch(ctx, m.AzureMachinePool)
 }
 
 // Close the MachineScope by updating the machine spec, machine status.
 func (m *MachinePoolScope) Close(ctx context.Context) error {
-	ctx, span := tele.Tracer().Start(ctx, "scope.MachinePoolScope.Close")
-	defer span.End()
+	ctx, _, done := tele.StartSpanWithLogger(ctx, "scope.MachinePoolScope.Close")
+	defer done()
 
 	if m.vmssState != nil {
 		if err := m.applyAzureMachinePoolMachines(ctx); err != nil {
@@ -521,8 +530,11 @@ func (m *MachinePoolScope) Close(ctx context.Context) error {
 
 // GetBootstrapData returns the bootstrap data from the secret in the Machine's bootstrap.dataSecretName.
 func (m *MachinePoolScope) GetBootstrapData(ctx context.Context) (string, error) {
-	ctx, span := tele.Tracer().Start(ctx, "scope.MachinePoolScope.GetBootstrapData")
-	defer span.End()
+	ctx, _, done := tele.StartSpanWithLogger(
+		ctx,
+		"scope.MachinePoolScope.GetBootstrapData",
+	)
+	defer done()
 
 	dataSecretName := m.MachinePool.Spec.Template.Spec.Bootstrap.DataSecretName
 	if dataSecretName == nil {
--- a/controllers/azurejson_machine_controller.go
+++ b/controllers/azurejson_machine_controller.go
@@ -23,8 +23,6 @@ import (
 
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
-	"go.opentelemetry.io/otel/attribute"
-	"go.opentelemetry.io/otel/trace"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -102,15 +100,17 @@ func (f filterUnclonedMachinesPredicate) Generic(e event.GenericEvent) bool {
 func (r *AzureJSONMachineReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Result, reterr error) {
 	ctx, cancel := context.WithTimeout(ctx, reconciler.DefaultedLoopTimeout(r.ReconcileTimeout))
 	defer cancel()
-	log := r.Log.WithValues("namespace", req.Namespace, "azureMachine", req.Name)
 
-	ctx, span := tele.Tracer().Start(ctx, "controllers.AzureJSONMachineReconciler.Reconcile",
-		trace.WithAttributes(
-			attribute.String("namespace", req.Namespace),
-			attribute.String("name", req.Name),
-			attribute.String("kind", "AzureMachine"),
-		))
-	defer span.End()
+	ctx, log, done := tele.StartSpanWithLogger(
+		ctx,
+		"controllers.AzureJSONMachineReconciler.Reconcile",
+		tele.KVP("namespace", req.Namespace),
+		tele.KVP("name", req.Name),
+		tele.KVP("kind", "AzureMachine"),
+	)
+	defer done()
+
+	log = log.WithValues("namespace", req.Namespace, "azureMachine", req.Name)
 
 	// Fetch the AzureMachine instance
 	azureMachine := &infrav1.AzureMachine{}
--- a/controllers/azuremachine_controller.go
+++ b/controllers/azuremachine_controller.go
@@ -22,8 +22,6 @@ import (
 
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
-	"go.opentelemetry.io/otel/attribute"
-	"go.opentelemetry.io/otel/trace"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/client-go/tools/record"
@@ -77,8 +75,8 @@ func NewAzureMachineReconciler(client client.Client, log logr.Logger, recorder r
 
 // SetupWithManager initializes this controller with a manager.
 func (amr *AzureMachineReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager, options Options) error {
-	ctx, span := tele.Tracer().Start(ctx, "controllers.AzureMachineReconciler.SetupWithManager")
-	defer span.End()
+	ctx, _, done := tele.StartSpanWithLogger(ctx, "controllers.AzureMachineReconciler.SetupWithManager")
+	defer done()
 
 	log := amr.Log.WithValues("controller", "AzureMachine")
 	var r reconcile.Reconciler = amr
@@ -139,16 +137,17 @@ func (amr *AzureMachineReconciler) SetupWithManager(ctx context.Context, mgr ctr
 func (amr *AzureMachineReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Result, reterr error) {
 	ctx, cancel := context.WithTimeout(ctx, reconciler.DefaultedLoopTimeout(amr.ReconcileTimeout))
 	defer cancel()
-	logger := amr.Log.WithValues("namespace", req.Namespace, "azureMachine", req.Name)
 
-	ctx, span := tele.Tracer().Start(ctx, "controllers.AzureMachineReconciler.Reconcile",
-		trace.WithAttributes(
-			attribute.String("namespace", req.Namespace),
-			attribute.String("name", req.Name),
-			attribute.String("kind", "AzureMachine"),
-		),
+	ctx, logger, done := tele.StartSpanWithLogger(
+		ctx,
+		"controllers.AzureMachineReconciler.Reconcile",
+		tele.KVP("namespace", req.Namespace),
+		tele.KVP("name", req.Name),
+		tele.KVP("kind", "AzureMachine"),
 	)
-	defer span.End()
+	defer done()
+
+	logger = logger.WithValues("namespace", req.Namespace, "azureMachine", req.Name)
 
 	// Fetch the AzureMachine VM.
 	azureMachine := &infrav1.AzureMachine{}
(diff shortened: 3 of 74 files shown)


controllers/azuremachine_controller.go
```diff
@@ -130,16 +130,21 @@ func (r *AzureMachineReconciler) SetupWithManager(ctx context.Context, mgr ctrl.
...
 		trace.WithAttributes(
 			attribute.String("namespace", req.Namespace),
 			attribute.String("name", req.Name),
 			attribute.String("kind", "AzureMachine"),
-		))
+		),
+	)
 	defer span.End()
 
+	logger := tele.CorrIDLogger(
```
arschles, line 143: @devigned @CecileRobertMichon I've purposely only made this change in only 1 reconcile loop to solicit feedback. Can you let me know what you think about this abstraction?

CecileRobertMichon:
@arschles the changes look good to me so far, could you please rebase the branch on the latest main branch so I can try it out in tilt? I'm having trouble bringing up tilt on the branch because it's outdated

controllers/azurejson_machine_controller.go
```diff
@@ -99,14 +99,18 @@ func (f filterUnclonedMachinesPredicate) Generic(e event.GenericEvent) bool {
 func (r *AzureJSONMachineReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Result, reterr error) {
 	ctx, cancel := context.WithTimeout(ctx, reconciler.DefaultedLoopTimeout(r.ReconcileTimeout))
 	defer cancel()
-	log := r.Log.WithValues("namespace", req.Namespace, "azureMachine", req.Name)
 
-	ctx, span := tele.Tracer().Start(ctx, "controllers.AzureJSONMachineReconciler.Reconcile",
+	ctx, span, log := tele.StartSpanWithLogger(
+		ctx,
+		"controllers.AzureJSONMachineReconciler.Reconcile",
+		tele.Tracer(),
```
devigned, line 106: I think I understand why you are adding `tele.Tracer()` and `r.Log` to the func signature. The signature you have exposed here makes it clear what the func depends upon. Transparency and ease of testing are always good.

The thing I struggle with is that if this function is used in every place with the same input, then perhaps, we need two functions. One that has all of these inputs, and another that defaults some of the inputs.

Further, the `r.Log` field should actually go away. We should not initialize a logger field on the reconciler, but rather use the context to carry the logger. I can see how this would be opaque to a caller, but it would be more consistent with [Cluster API](https://github.com/kubernetes-sigs/cluster-api/blob/074c2cb334ab54e4f8ef5ddcf40a15a3aa1044a3/controllers/cluster_controller.go#L97). Being part of the Cluster API ecosystem and staying as consistent as we can with CAPI would be a kindness to developers contributing across the projects.

Perhaps, this fits with having two functions, one that defaults in an opinionated way, and another that is referentially transparent. 

```go
      ctx, span, logger := tele.StartSpanWithLogger(
		ctx,
		"controllers.AzureMachineReconciler.Reconcile",
		trace.WithAttributes(
			attribute.String("namespace", req.Namespace),
			attribute.String("name", req.Name),
			attribute.String("kind", "AzureMachine"),
		),
	)
```

Perhaps, we could even go a step further. Since we only use one func on the `Span`, and that it would beneficial to get logging statements on the span (as you have done with the composite logger below :+1: ), consider returning a `SpanLogger`. Similar to the following.

```go
      ctx, spanLogger := tele.StartSpanWithLogger(
		ctx,
		"controllers.AzureMachineReconciler.Reconcile",
		trace.WithAttributes(
			attribute.String("namespace", req.Namespace),
			attribute.String("name", req.Name),
			attribute.String("kind", "AzureMachine"),
		),
	)
        defer spanLogger.End()
        spanLogger.Infof("something to log on both span and logger named %q", thing)
```

@CecileRobertMichon wdyt?

devigned (changes requested):
Added some thoughts on `tele.StartSpanWithLogger`. I'm curious what other maintainers / contributors think.

controllers/azurejson_machine_controller.go
shysank: +1 for having a default alternate. Not sure about the 2nd point though as we'll need the `compositeLogger` to embed `trace.Span`, which is not the intent here iiuc.

CecileRobertMichon: > The thing I struggle with is that if this function is used in every place with the same input, then perhaps, we need two functions. One that has all of these inputs, and another that defaults some of the inputs.

+1, seems like a good improvement

Not sure I follow the rest, @shysank can you help me understand what you mean by 
> we'll need the compositeLogger to embed trace.Span, which is not the intent here iiuc.

?

shysank: Maybe I misunderstood David's comment. My understanding was that the logger `spanLogger` returned from `StartSpanWithLogger` will still be a composite logger of some sort with `trace.Span` in it to call `span.end()`. So, either we modify `compositeLogger` to include `trace.Span`, or modify spanLogger to include composite logger functionality, both of which seem to be coupling two different things imo.

arschles: > One that has all of these inputs, and another that defaults some of the inputs.

We can avoid adding a second function to this API with functional options or similar. Would that be sufficient?

>Further, the r.Log field should actually go away. We should not initialize a logger field on the reconciler, but rather use the context to carry the logger

What do you mean by "initialize a logger"? `StartSpanWithLogger` creates two sub-loggers from the incoming `r.Log`.

The overall problem I have with this approach is the implicitness, as you probably could have guessed. We trade writing one extra parameter per call site -- more work to write and review initially, to be sure -- for a holistic increase in readability. Also, context values are intended for request scoped data, not for parameters to functions.

One other note - I don't see what is gained by using the embedded logger, from an API or community standpoint. This function wouldn't be any more compatible with the cluster API ecosystem if it simply extracted the logger from the context, would it?

>consider returning a `SpanLogger`

the `SpanLogger` as is currently written, or something different? as currently written, the `SpanLogger` won't write to standard out, which seems problematic.

>modify spanLogger to include composite logger functionality, both of which seem to be coupling two different things imo

agreed with you @shysank - my intended purpose for `spanLogger` is logging, not tracing. I think there would be a lot of long-term benefit to keeping those two functional areas orthogonal. The same applies to span logging and logs to standard out as well.

devigned: The composite logger already has the span embedded. Sure, the logger being returned fits the interface for a logger, but it has a dependency on both, std logger and span logger. They are already connected explicitly. Each log message that goes to standard out also goes to the span. The only thing missing is the `End()` func. 

As for fetching the logger from the context, it would be much better than the options we have now. At this point, we add the logger as field on structs, so we can avoid passing the logger into all of the methods all the way down the chain of calls. This is precisely where context works for us, because it is the one thing we have chosen to pass all the way down the call chain.

arschles: >The composite logger already has the span embedded

Yup, my intention with the composite logger was to provide it as a union of N loggers. The span and standard loggers are just `logr.Logger` implementations to the composite logger, and I think this setup probably shouldn't know anything about tracing.

>it would be much better than the options we have now

regarding this and the above point, we might be able to come to a good compromise by:

- Abstracting the magic that I don't like, while also...
- Providing an API that uses it under the hood and...
- Giving users the ability to easily `defer span.End()` while not tightly coupling logging and tracing concerns
Here's an idea:

```go
func StartSpan(
	ctx context.Context,
	spanName string,
	opts ...trace.SpanOption,
) (context.Context, logr.Logger, func()) {
	lggr := extractLoggerFromCtx(ctx) // this abstracts the "magic" of pulling a logger out of a context.Context
	c, span := tele.Tracer().Start(ctx, spanName, opts...)
	endFn := func() {
		span.End()
	}
	return c, &compositeLogger{
		loggers: []logr.Logger{
			corrIDLogger(ctx, lggr),
			&spanLogger{Span: span},
		},
	}, endFn
}
```

In reconcilers, this would be called as:

```go
ctx, log, done := StartSpan(ctx, "MyReconciler")
defer done()

log.Info("Starting!")
```

This follows a similar pattern as `context.WithCancel`, `context.WithTimeout`, etc... with its "done" return function, while hiding the existence of a Span completely. Users simply need to defer a call to the final return value and nothing else. It also abstracts and centralizes the functionality that pulls loggers out of the context.

WDYT?

devigned: Great suggestion! 👍

devigned: What do other think of https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1575#discussion_r704841410?

CecileRobertMichon: I like it, the done() func is really neat

shysank: +1 from me as well.

arschles: @devigned @CecileRobertMichon @shysank I'll build this out in a distinct commit within this PR, so that people can get a feel for the ergonomics

arschles: ok, see https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1575/commits/26c88a13d71d2c65264d6bbd44a3ae806c0f55b9

controllers/azurejson_machine_controller.go
```diff
@@ -99,15 +99,17 @@ func (f filterUnclonedMachinesPredicate) Generic(e event.GenericEvent) bool {
 func (r *AzureJSONMachineReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Result, reterr error) {
 	ctx, cancel := context.WithTimeout(ctx, reconciler.DefaultedLoopTimeout(r.ReconcileTimeout))
 	defer cancel()
-	log := r.Log.WithValues("namespace", req.Namespace, "azureMachine", req.Name)
 
-	ctx, span := tele.Tracer().Start(ctx, "controllers.AzureJSONMachineReconciler.Reconcile",
+	ctx, log, done := tele.StartSpanWithLogger(
+		ctx,
+		"controllers.AzureJSONMachineReconciler.Reconcile",
 		trace.WithAttributes(
```
devigned, line 106: What do folks think about leaking upward the `trace.WithAttributes` at this level? If we don't know about the span, we probably don't want to know about the OT trace attributes.

devigned (commented):
One more thought about the signature of `StartSpanWithLogger`.

controllers/azurejson_machine_controller.go
arschles: what if it was abstracted to API-native attributes and translated below. the API could be something like:

```go
tele.StartSpanWithLogger(ctx, "mything", tele.KVP("namespace", req.Namespace), tele.KVP(...))
```

arschles: @devigned see https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1575/commits/f3aeb5a50f9a8551aac7f13b906f9d13764f8ae3 for generic attributes and LMK what you think.

devigned (commented):
:+1: to the tele KVs. I support continuing with this approach.

arschles:
@devigned k, cool. I'll move forward a few reconcile loops at a time to test it out.

arschles:
Ok, after some incremental work, I didn't find anything in this repo that the new `tele.StartSpanWithLogger` won't work with. I did one big mechanical change, and committed it all at once. See https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1575/commits/db3372dd02b85f2045d6012672fc8b0c83149f38

CecileRobertMichon:
@arschles that looks great. Are there any remaining changes needed on this PR?

@devigned what do you say?

util/tele/span_logger.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+	ctx, span := Tracer().Start(
+		ctx,
+		spanName,
+		trace.WithAttributes(cfg.teleKeyValues()...),
+	)
+	endFn := func() {
+		span.End()
+	}
+	lggr := log.FromContext(ctx)
```
devigned, line 141: Please add the spanName to the logger via `WithName` or `WithVallues`.

devigned (changes requested):

util/tele/span_logger.go
arschles: @devigned done

devigned:
Let's get this passing the tests and get it pulled in!
```
util/tele/span_logger.go:31:18: Comment should end in a period (godot)
// data to a span
                 ^
util/tele/span_logger.go:102:6: exported type `Config` should have comment or be unexported (golint)
type Config struct {
     ^
util/tele/span_logger.go:116:6: exported type `Option` should have comment or be unexported (golint)
type Option func(*Config)
     ^
util/tele/span_logger.go:118:1: exported function `KVP` should have comment or be unexported (golint)
func KVP(key, value string) Option {
^
util/tele/span_logger.go:124:1: exported function `StartSpanWithLogger` should have comment or be unexported (golint)
func StartSpanWithLogger(
^
util/tele/tele.go:60:1: comment on exported function `Tracer` should be of the form `Tracer ...` (golint)
// Example usage:
^
azure/services/groups/client.go:93: unnecessary leading newline (whitespace)
func (ac *azureClient) DeleteAsync(ctx context.Context, spec azure.ResourceSpecGetter) (azureautorest.FutureAPI, error) {

level=info msg="File cache stats: 367 entries of total size 2.6MiB"
level=info msg="Memory: 546 samples, avg is 817.0MB, max is 2366.1MB"
level=info msg="Execution took 1m0.130999822s"
make: *** [Makefile:273: lint] Error 1
```

arschles:
@devigned cool. fixed the lint errors and squashed. I'll watch CI, but I think this is ready to go 🚀

devigned:
I'm going to pull down the PR and build some stuff to verify output / tracing info.

devigned:
Alright. I have good news and bad news. The good news is that I can see correlationIDs and was able to correlate requests and view them in both Application Insights as well as Jaeger. Unfortunately, log events are not exported to Application Insights, but they are in Jaeger. See the screen shots below.

## App Insights
image

## Jaeger
image

It's clear they are the same request, but one contains logs and the other does not. This is not a problem with this PR, but rather, something that needs to be added to https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/95f5018e0d57667fd04b79a30c45ab3d126cd17d/exporter/azuremonitorexporter/trace_to_envelope.go#L642.

I don't think that should hold up the PR. We can work to add events / logs to the App Insights exporter.

Well done!!

/lgtm
/assign @CecileRobertMichon @mboersma 

What do folks think?

devigned (approved):

azure/scope/machinepool.go
```diff
@@ -211,8 +211,11 @@ func (m MachinePoolScope) MaxSurge() (int, error) {
 // updateReplicasAndProviderIDs ties the Azure VMSS instance data and the Node status data together to build and update
 // the AzureMachinePool replica count and providerIDList.
 func (m *MachinePoolScope) updateReplicasAndProviderIDs(ctx context.Context) error {
-	ctx, span := tele.Tracer().Start(ctx, "scope.MachinePoolScope.UpdateInstanceStatuses")
-	defer span.End()
+	ctx, _, done := tele.StartSpanWithLogger(
+		ctx,
```
mboersma, line 215: Nitpick: I think this reads more easily as a one-liner, which it is in most of the changes here.

util/tele/span_logger.go
```diff
@@ -0,0 +1,168 @@
...
+	"fmt"
+	"time"
+
+	"github.com/go-logr/logr"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+// SpanLogger is a logr.Logger implementation that writes log
```
mboersma, line 30:
```suggestion
// spanLogger is a logr.Logger implementation that writes log
```

mboersma (approved):
/lgtm

This is great, thanks @arschles! I only had two trivial comments, so feel free to ignore and merge as-is.

azure/scope/machinepool.go
arschles: @mboersma done!

util/tele/span_logger.go
arschles: @mboersma done

devigned:
> @devigned mind if I open a new issue for implementing the change you suggested in [#1575 (comment)](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1575#issuecomment-924464966)? I can pick it up right after this PR

100% that would be awesome! We've been eying to contribute to OTEL for App Insights for a while, but just haven't gotten around to doing it. We've also been talking about adding metrics to the exporter too. Events would be a great contribution.

The thing that I wasn't clear on was how events would be correlated with spans. I'm guessing there's something in the API model for App Insights to do the correlation, but it wasn't clear to me from a quick glance.

CecileRobertMichon:
@arschles apologies, I merged #1691 without realizing it would cause conflicts with this one

azure/scope/machinepool.go
CecileRobertMichon: doesn't this apply to the other files as well?

mboersma:
lgtm after a squash and rebase. 👍🏻

azure/scope/machinepool.go
mboersma: It does--there are several more below that could be one-liners, but I don't want to block this PR on trivia.

devigned:
@arschles let me know if you want to pair to get through the rebase.

arschles:
@CecileRobertMichon no worries, I'll take care of the conflicts/squash. @devigned thanks for the offer. I'm gonna give it a shot today and if there's something thorny comes up I'll ping you. Thanks!

arschles:
@CecileRobertMichon @devigned @mboersma ok, all conflicts fixed and squashed. I found a few more places that needed to be converted from `tele.Tracer().Start(...)` calls to `tele.StartSpanWithLogger(...)` calls as well. There are also so places where I changed formatting more than strictly necessary; I'm happy to change to be closer to original formatting if you'd like. 

Let me know what you think.

devigned (approved):
/lgtm
/approve

arschles:
Thanks everyone!

devigned:
Great work, @arschles!! Thank you!

arschles:
my pleasure. I'll look next at some of the parallelism work and see where I can help there.

PR #1579, opened 2021-08-01
Tiltfile should tolerate creds set in base 64 encoding

### LINKED ISSUE #1573: Tiltfile should tolerate creds set in base 64 encoding
/kind feature

We should reorder the tiltfile so that auth validation (https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/Tiltfile#L97) happens after decoding B64 variables (https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/Tiltfile#L218-L221) so that B64 values are included when checking if the right creds are set.

Suggestion is to move

```
    for key in keys:
        if key[-4:] == "_B64":
            os.environ[key[:-4]] = base64_decode(os.environ[key])
```

to the beginning on `validate_auth`.           

**Environment:**

sayantani11:
**What type of PR is this?**

Fixes #1573 

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -96,7 +96,11 @@ def fixup_yaml_empty_arrays(yaml_str):
 
 def validate_auth():
     substitutions = settings.get("kustomize_substitutions", {})
-    missing = [k for k in keys if k not in substitutions and not os.environ.get(k)]
+    os.environ.update(substitutions)
+    for sub in substitutions:
+        if sub[-4:] == "_B64":
+            os.environ[sub[:-4]] = base64_decode(os.environ[sub])
+    missing = [k for k in keys if not os.environ.get(k)]
     if missing:
         fail("missing kustomize_substitutions keys {} in tilt-setting.json".format(missing))
 
@@ -130,8 +134,6 @@ def observability():
 # Build CAPZ and add feature gates
 def capz():
     # Apply the kustomized yaml for this provider
-    substitutions = settings.get("kustomize_substitutions", {})
-    os.environ.update(substitutions)
     yaml = str(kustomizesub("./hack/observability")) # build an observable kind deployment by default
 
     # add extra_args if they are defined
@@ -186,10 +188,8 @@ def create_identity_secret():
     os.putenv('AZURE_CLUSTER_IDENTITY_SECRET_NAMESPACE', 'default')
     os.putenv('CLUSTER_IDENTITY_NAME', 'cluster-identity')
 
-    substitutions = settings.get("kustomize_substitutions", {})
-    os.putenv('AZURE_CLIENT_SECRET_B64', base64_encode(substitutions.get("AZURE_CLIENT_SECRET")))
-
-    local("cat templates/azure-cluster-identity/secret.yaml | " + envsubst_cmd + " | kubectl apply -f -", quiet=True)
+    os.putenv('AZURE_CLIENT_SECRET_B64', base64_encode(os.environ.get("AZURE_CLIENT_SECRET")))
+    local("cat templates/azure-cluster-identity/secret.yaml | " + envsubst_cmd + " | kubectl apply -f -", quiet=True, echo_off=True)
     os.unsetenv('AZURE_CLIENT_SECRET_B64')
 
 def create_crs():
@@ -214,12 +214,8 @@ def flavors():
     config.define_string_list("worker-flavors")
     cfg = config.parse()
     worker_templates = cfg.get('templates-to-run', [])
-
+   
     substitutions = settings.get("kustomize_substitutions", {})
-    for key in keys:
-        if key[-4:] == "_B64":
-            os.environ[key[:-4]] = base64_decode(os.environ[key])
-
     ssh_pub_key_B64 = "AZURE_SSH_PUBLIC_KEY_B64"
     ssh_pub_key_path = "$HOME/.ssh/id_rsa.pub"
     if substitutions.get(ssh_pub_key_B64):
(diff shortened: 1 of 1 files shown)


Tiltfile
```diff
@@ -215,11 +218,6 @@ def flavors():
     cfg = config.parse()
     worker_templates = cfg.get('templates-to-run', [])
 
-    substitutions = settings.get("kustomize_substitutions", {})
```
CecileRobertMichon, line 218: we need to leave this line since substitutions is used later on line 223

`Tiltfile:223:8: undefined: substitutions`

Tiltfile
```diff
@@ -96,6 +96,9 @@ def fixup_yaml_empty_arrays(yaml_str):
 
 def validate_auth():
     substitutions = settings.get("kustomize_substitutions", {})
+    for key in keys:
```
CecileRobertMichon, line 99: I realize now this won't work because it's looking at the keys instead of the substitutions. Here's what this should be:

```
def validate_auth():
    os.environ.update(substitutions)
    for sub in substitutions:
        if sub[-4:] == "_B64":
            os.environ[sub[:-4]] = base64_decode(os.environ[sub])
    missing = [k for k in keys if not os.environ.get(k)]
    if missing:
        fail("missing kustomize_substitutions keys {} in tilt-setting.json".format(missing))
```

This will set all the variables in substitutions as environment variables and then use those to set the decoded equivalent variable if the substitution had a _B64 suffix.

Since we're doing `os.environ.update(substitutions)` early now, you can delete the line in `dev capz():` that does `os.environ.update(substitutions)` and the one before (lines 136-137).

CecileRobertMichon: we also need to change line 193 to

```
    os.putenv('AZURE_CLIENT_SECRET_B64', base64_encode(os.environ.get("AZURE_CLIENT_SECRET")))
```

and remove the `substitutions = settings.get("kustomize_substitutions", {})` line above (line 192)

So it gets the secret directly from the environment variables

CecileRobertMichon: Finally, let's add `echo_off=True` to all the local() functions that might print secrets so we don't output those in logs (I spotted 3 in the file, in `base64_encode`, `base64_decode` and `create_identity_secret`). For example:

```
    encode_blob = local("echo '{}' | tr -d '\n' | base64 - | tr -d '\n'".format(to_encode), quiet=True, echo_off=True)
```

CecileRobertMichon (changes requested):
@sayantani11 I pointed you in the wrong direction before, apologies for the confusion. I've added comments on what needs to be changed, let me know if you have questions.

Again, you can test this by running `make tilt-up` locally after following https://capz.sigs.k8s.io/developers/development.html#tilt-for-dev-in-capz to set up your tilt-settings.json file

sayantani11:
@CecileRobertMichon I was also thinking that, because although I was able to open it in local host, it showed tiltfile error.
But I thought, it was just my device :)

sayantani11:
``` <builtin>: in read_json
Error: error parsing JSON from tilt-settings.json: invalid character '}' looking for beginning of object key string ```
@CecileRobertMichon the local host still has this error.
Does tilt-setting.json need change? or the problem is from my side?

**Comment by CecileRobertMichon** [comment]
@sayantani11 that error looks to be a formatting error in your tilt-settings.json

Did you follow https://capz.sigs.k8s.io/developers/development.html#tilt-for-dev-in-capz to set it up? Do you have environment variables set for AZURE_SUBSCRIPTION_ID, etc. ? If not, you could set them to fake values for now just so you can test tilt up.

**Comment by sayantani11** [comment]
@CecileRobertMichon I did but let me look into it again.

**Comment by sayantani11** [comment]
@CecileRobertMichon Guess I am making some mistake in entering the environment variable.
So suppose,
``` "AZURE_SUBSCRIPTION_ID_B64": "$(echo "${AZURE_SUBSCRIPTION_ID}" | tr -d '\n' | base64 | tr -d '\n')" ```
I use the fake ID, how am I filling this up?

**Comment by CecileRobertMichon** [comment]
@sayantani11 try:

```
{
  "kustomize_substitutions": {
      "AZURE_SUBSCRIPTION_ID_B64": "ZmFrZXZhbHVlCg==",
      "AZURE_TENANT_ID_B64": "ZmFrZXZhbHVlCg==",
      "AZURE_CLIENT_SECRET_B64": "ZmFrZXZhbHVlCg==",
      "AZURE_CLIENT_ID_B64": "ZmFrZXZhbHVlCg==",
  }
}
```

`ZmFrZXZhbHVlCg==` is "fakevalue" in base 64.

sayantani11:
@CecileRobertMichon Yeah I actually did it but forgot to push. Doing it now.

sayantani11:
@CecileRobertMichon It came to notice that  https://capz.sigs.k8s.io/developers/development.html#tilt-for-dev-in-capz doesn't mention installation of Tilt as a Makefile dependency. So it would be better if it's added, as I was getting the error at first.

CecileRobertMichon:
>  It came to notice that https://capz.sigs.k8s.io/developers/development.html#tilt-for-dev-in-capz doesn't mention installation of Tilt as a Makefile dependency. So it would be better if it's added, as I was getting the error at first.

Let's do this as a follow-up. The rest of the changes look good, thanks again for your patience here!

/lgtm
/assign @nader-ziada

CecileRobertMichon:
@sayantani11 before this merges, can you please squash the commits?

/hold
/ok-to-test

Issue #1532, opened 2021-07-16
Use only nat gateway for outbound connections


shysank:
/kind feature

With https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1411, we are going to allow to configure multiple subnets for worker nodes. We also support nat gateway for outbound connections which means when some subnets have nat gateway configured, and some do not, we fallback to outbound load balancer. While this works, this adds some complexity to the code, and the network topology. Since we now have support for nat gateway, this is a good chance to get rid of outbound load balancer as they serve the same purpose. Also microsoft [recommends](https://docs.microsoft.com/en-us/azure/load-balancer/load-balancer-outbound-connections#scenarios) using nat gateway in favor of outbound loadbalacers.

**Describe the solution you'd like**
Remove all outbound lbs and only use nat gateway for outbound connectivity.

**Anything else you would like to add:**
Handling older clusters might be tricky.

- [ ] Replace Node outbound load balancer with nat gateway
- [ ] Replace Control Plane outbound load balancer with nat gateway (For private clusters)

**Environment:**

- cluster-api-provider-azure version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

CecileRobertMichon:
Two things we need to figure out before doing this:

1) How to handle back compat for existing clusters that have outbound LBs
2) are there any use cases in which you'd want an LB or _not_ want a NAT Gateway (but still want outbound)

@jackfrancis @Michael-Sinz tagging you in case you have any feedback or user data on this

jackfrancis:
I would love to get rid of unnecessary Load Balancers. LBs for the sole purpose of providing outbound egress + SNAT is overkill, IMO (I assume that's what we're talking about removing here).

That said, in Azure there are "tricks" that we rely upon the LB for in order to ensure that our clusters don't run out of outbound SNAT ports (in other words, when you have large clusters that maintain a bunch of connections that talk outside of the VNET, you accumulate SNAT ports that are counted against a hard limit, I think you have a certain amount of ports per "outbound IP" associated with the LB). Does the NAT Gateway solution present a comparable path forward for folks who need, let's call it, a whole boat ton of outbound SNAT ports?

CecileRobertMichon:
@jackfrancis one of the main benefits of NAT gateway is actually that it doesn't have the same SNAT port exhaustion limits

[...]

PR #1589, opened 2021-08-06
fix: don't create outbound LB if using NatGateway

### LINKED ISSUE #1587: Don't create outbound LB if nat-gateway is enabled
/kind bug

Building a cluster w/ only natgateway "node" subnets included an outbound LB w/ the nodes in the backend pool

The natgateway resource itself handles outbound connections from nodes, so the outbound LB is superfluous (and in fact constrains the node scale, as we run out of outbound SNAT ports quickly compared to natgateway).

**Environment:**

- cluster-api-provider-azure version: v1alpha4
- Kubernetes version: (use `kubectl version`): 1.21.3 (should repro for all versions) 

jackfrancis:
**What type of PR is this?**

/kind bug

This PR fixes the nat-gateway + subnet reconciliation logic so that we stop creating unnecessary outbound LB resources if we're exclusively using NatGateway across all of our node subnets.

Fixes #1587

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
don't create outbound LB if using NatGateway
```

### DIFF
--- a/api/v1alpha4/azurecluster_default.go
+++ b/api/v1alpha4/azurecluster_default.go
@@ -202,10 +202,10 @@ func (c *AzureCluster) setNodeOutboundLBDefaults() {
 			return
 		}
 
-		var oneSubnetWithoutNatGateway bool
+		var needsOutboundLB bool
 		for _, subnet := range c.Spec.NetworkSpec.Subnets {
 			if subnet.Role == SubnetNode && !subnet.IsNatGatewayEnabled() {
-				oneSubnetWithoutNatGateway = true
+				needsOutboundLB = true
 				break
 			}
 		}
@@ -213,7 +213,7 @@ func (c *AzureCluster) setNodeOutboundLBDefaults() {
 		// If we don't default the outbound LB when there are some subnets with nat gateway,
 		// and some without, those without wouldn't have outbound traffic. So taking the
 		// safer route, we configure the outbound LB in that scenario.
-		if len(c.Spec.NetworkSpec.Subnets) > 0 && !oneSubnetWithoutNatGateway {
+		if !needsOutboundLB {
 			return
 		}
 
--- a/api/v1alpha4/azurecluster_default_test.go
+++ b/api/v1alpha4/azurecluster_default_test.go
@@ -803,6 +803,20 @@ func TestNodeOutboundLBDefaults(t *testing.T) {
 				Spec: AzureClusterSpec{
 					NetworkSpec: NetworkSpec{
 						APIServerLB: LoadBalancerSpec{Type: Public},
+						Subnets: Subnets{
+							{
+								Role:          SubnetControlPlane,
+								Name:          "control-plane-subnet",
+								SecurityGroup: SecurityGroup{},
+								RouteTable:    RouteTable{},
+							},
+							{
+								Role:          SubnetNode,
+								Name:          "node-subnet",
+								SecurityGroup: SecurityGroup{},
+								RouteTable:    RouteTable{},
+							},
+						},
 					},
 				},
 			},
@@ -812,6 +826,20 @@ func TestNodeOutboundLBDefaults(t *testing.T) {
 				},
 				Spec: AzureClusterSpec{
 					NetworkSpec: NetworkSpec{
+						Subnets: Subnets{
+							{
+								Role:          SubnetControlPlane,
+								Name:          "control-plane-subnet",
+								SecurityGroup: SecurityGroup{},
+								RouteTable:    RouteTable{},
+							},
+							{
+								Role:          SubnetNode,
+								Name:          "node-subnet",
+								SecurityGroup: SecurityGroup{},
+								RouteTable:    RouteTable{},
+							},
+						},
 						APIServerLB: LoadBalancerSpec{
 							Type: Public,
 						},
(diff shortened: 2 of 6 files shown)


api/v1alpha4/azurecluster_default.go
```diff
@@ -202,18 +202,18 @@ func (c *AzureCluster) setNodeOutboundLBDefaults() {
 			return
 		}
 
-		var oneSubnetWithoutNatGateway bool
+		var needsOutboundLB bool
```
jackfrancis, line 205: this change aims to make the intent of this bool flag a little more semantic (though the current one is literally correct, it's not as obvious what we're aiming for)

CecileRobertMichon:
@jackfrancis looks like unit tests are failing, PTAL

api/v1alpha4/azurecluster_default.go
```diff
@@ -202,18 +202,18 @@ func (c *AzureCluster) setNodeOutboundLBDefaults() {
...
+				needsOutboundLB = true
 				break
 			}
 		}
 
 		// If we don't default the outbound LB when there are some subnets with nat gateway,
 		// and some without, those without wouldn't have outbound traffic. So taking the
 		// safer route, we configure the outbound LB in that scenario.
-		if len(c.Spec.NetworkSpec.Subnets) > 0 && !oneSubnetWithoutNatGateway {
+		if !needsOutboundLB {
```
jackfrancis, line 216: @CecileRobertMichon `TestNodeOutboundLBDefaults` is failing because it is implemented to expect a load balancer object even if there are no subnets in the input spec. (See the removal of the length requirement above.)

Is that a valid way to express a UT? Will a cluster ever not actually have any subnets by the time that `setNodeOutboundLBDefaults` is entered into?

We could either restore that length requirement (doesn't make sense to me or I could be missing something) or fix up the existing UT input/output for the default case to express what we expect given this change.

CecileRobertMichon: > Will a cluster ever not actually have any subnets by the time that setNodeOutboundLBDefaults is entered into

short answer is no, we set subnet defaults before setting LB defaults on line 57. The length requirement doesn't make sense to me either, if there are no subnets (that wouldn't happen but in theory) then I don't see why we need an outbound LB added. I vote we change the UT to make sure meaningful edge cases are covered instead: all subnets have a NAT GW, none have it, all but one, and all node subnets have it but not the control plane subnet.

jackfrancis: in terms of the "all subnets have a NAT GW", is that real scenario? will we ever have a cluster that doesn't have a control plane subnet, and/or do we support a control plane subnet that uses NAT GW?

CecileRobertMichon: you're right, my bad. NAT GW is only for node subnets. So all, some, none of the node subnets have NAT GW needs to be tested.

jackfrancis: O.K. cool, thanks!

jackfrancis: One more question on the "any one node subnet does not have NAT Gateway enabled" scenario. Currently, we seem to configure a single "default flavor" of outbound LB. Does that mean if any one node pool (assuming a 1:1 mapping of node pools to subnets) has NAT Gateway disabled, that we will create a single outbound LB that has *every single node in the cluster* in its backend pool? The logic as is doesn't seem smart enough to only use the outbound LB for node VMs that aren't configured for NAT Gateway.

Hope that makes sense. Basically, as I understand it, the current "some but not all nodes are using NAT Gateway" implementation seems sort of half-baked (if it even works at all).

CecileRobertMichon: See  https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1411#discussion_r670578995

CecileRobertMichon: cc @fiunchinho

jackfrancis: Thanks.

In the scenario where we have _n_ node pools with NAT Gateway and _n_ without, then essentially we create the outbound LB so that all nodes have outbound SNAT (not just those pools with NAT Gateway enabled). But in that case we don't actually use the NAT Gateway in practice, because the outbound LB has precedence over the NAT Gateway for all backend pool members that have a NAT Gateway (in our configuration the overlap is 100% — *all* of the NAT Gateway-enabled nodes will be members of this outbound LB).

If the above is correct, then I'm not sure why we support things in this way. It would make more sense to require the user "choose one or the other", because otherwise the user is effectively paying for a NAT Gateway that is not being used. (among other undesirable outcomes)

azure/services/natgateways/natgateways.go
```diff
@@ -118,12 +122,26 @@ func (s *Service) getExisting(ctx context.Context, spec azure.NatGatewaySpec) (*
 	if err != nil {
 		return nil, err
 	}
+	// We must have a non-nil PublicIPAddresses, and the underlying SubResource slice type must have a length of 1
+	// TODO do we want to eventually handle NatGateway resources w/ more than one public IP address?
```
CecileRobertMichon, line 126: eventually there might be use cases for having more than 1 IP (eg. if you need more than 64k connections) but currently, it's not supported. We should not aim to support it now but making it explicit is better than opening the door for future bugs IMO.

If there is more than one IP, is the right behavior to return an error and completely block reconciliation?

jackfrancis: Defer to project maintainers. We could leave this here until that support is actually added (as a very obvious reminder that there are implementation details to be sorted out) or it could passively accept a bunch of IP addresses and then always just use the first one (see L130 below).

api/v1alpha4/azurecluster_default.go
fiunchinho: If I understand correctly, that's the end goal with https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1532 . After that is fixed, we would only support Nat Gateways for outbound traffic from the node subnets.

jackfrancis: Does that include control plane nodes? So the idea is we don't re-use the control plane LB for outbound SNAT (assuming we're in a non-private cluster scenario where there _is_ a LB for control plane nodes) and instead use a dedicated NAT gateway for the control plane nodes?

CecileRobertMichon: For control plane nodes there are two cases:

- Public API Server (not a private cluster): the API Server LB also provides outbound, no need for another outbound mechanism
- Private cluster: The internal API Server LB does not provide outbound, we currently deploy a separate control plane outbound LB just for outbound. We could also switch this to instead creating a NAT gateway for the control plane subnet, but the SNAT port exhaustion issue isn't as much of a problem here since scaling control plane nodes to 60 VMs isn't a requirement.

api/v1alpha4/azurecluster_default_test.go
```diff
@@ -803,6 +803,253 @@ func TestNodeOutboundLBDefaults(t *testing.T) {
 				Spec: AzureClusterSpec{
 					NetworkSpec: NetworkSpec{
 						APIServerLB: LoadBalancerSpec{Type: Public},
+						Subnets: Subnets{
```
jackfrancis, line 806: @CecileRobertMichon can you sanity check these new scenarios, and that this is a complete representation of the cases we want to test for the outboundLB convenience func `setNodeOutboundLBDefaults`?

This is the new "default case", where we have a single control plane subnet and a single node pool without any NAT Gateway configuration. We should expect to get a default `NodeOutboundLB` resource injected into the cluster object.

api/v1alpha4/azurecluster_default_test.go
```diff
@@ -803,6 +803,253 @@ func TestNodeOutboundLBDefaults(t *testing.T) {
...
+							Type:                 Public,
+							FrontendIPsCount:     to.Int32Ptr(1),
+							IdleTimeoutInMinutes: to.Int32Ptr(DefaultOutboundRuleIdleTimeoutInMinutes),
+						},
+					},
+				},
+			},
+		},
+		{
+			name: "NAT Gateway enabled - no LB",
```
jackfrancis, line 864: This UT case includes a vanilla configuration (one control plane subnet and one node pool subnet) with NAT Gateway enabled on the node pool. We should expect to *not* get an outbound LB here.

api/v1alpha4/azurecluster_default_test.go
```diff
@@ -803,6 +803,253 @@ func TestNodeOutboundLBDefaults(t *testing.T) {
...
+						},
+						APIServerLB: LoadBalancerSpec{
+							Type: Public,
+						},
+					},
+				},
+			},
+		},
+		{
+			name: "NAT Gateway enabled on 1 of 2 node subnets",
```
jackfrancis, line 923: This scenario includes a single node pool without NAT Gateway enabled, so we should expect to get the single, default NodeOutboundLB injected.

api/v1alpha4/azurecluster_default_test.go
```diff
@@ -803,6 +803,253 @@ func TestNodeOutboundLBDefaults(t *testing.T) {
...
+							}},
+							Type:                 Public,
+							FrontendIPsCount:     to.Int32Ptr(1),
+							IdleTimeoutInMinutes: to.Int32Ptr(DefaultOutboundRuleIdleTimeoutInMinutes),
+						},
+					},
+				},
+			},
+		},
+		{
```
jackfrancis, line 1006: This UT expresses the expectation that even if we have a whole bunch of node pool subnets, we only get the one default NodeOutboundLB.

api/v1alpha4/azurecluster_default_test.go
```diff
@@ -858,7 +1310,7 @@ func TestNodeOutboundLBDefaults(t *testing.T) {
 			},
 		},
 		{
-			name: "frontendIPsCount > 1",
+			name: "NodeOutboundLB declared as input with non-default IdleTimeoutInMinutes and FrontendIPsCount values",
```
jackfrancis, line 1313: Changed the name of this case to express specifically what it's testing

api/v1alpha4/azurecluster_default_test.go
```diff
@@ -900,8 +1352,51 @@ func TestNodeOutboundLBDefaults(t *testing.T) {
...
-							IdleTimeoutInMinutes: to.Int32Ptr(15),
+							FrontendIPsCount:     to.Int32Ptr(2),  // we expect the original value to be respected here
+							IdleTimeoutInMinutes: to.Int32Ptr(15), // we expect the original value to be respected here
+						},
+					},
+				},
+			},
+		},
+		{
+			name: "NodeOutboundLB declared as input with non-respected values",
```
jackfrancis, line 1363: @CecileRobertMichon this is a weird case to test from a UX perspective but in fact it expresses literally what happens if you pass in a predefined LB object w/ either `Name`, `SKU`, or `Type` values. So if we _want_ to statically override those values we want this UT; or, if we don't want to do that now's the time to think about what we do want to do.

azure/services/subnets/subnets_test.go
```diff
@@ -190,6 +190,18 @@ func TestReconcileSubnets(t *testing.T) {
 			expectedError: "",
 			expect: func(s *mock_subnets.MockSubnetScopeMockRecorder, m *mock_subnets.MockClientMockRecorder) {
 				s.V(gomock.AssignableToTypeOf(2)).AnyTimes().Return(klogr.New())
+				s.Subnet("my-subnet").AnyTimes().Return(infrav1.SubnetSpec{
+					ID:         "subnet-id",
+					Name:       "my-subnet",
+					Role:       infrav1.SubnetNode,
+					CIDRBlocks: []string{"10.0.0.0/16"},
```
CecileRobertMichon, line 197: should we add a case to test that it doesn't overwrite existing NAT Gateways since that's the regression we're fixing. We could also test that if CIDRBlocks and ID are empty in the existing spec, it will set them with the Azure subnet values.

api/v1alpha4/azurecluster_default_test.go
CecileRobertMichon: The expected behavior, documented in https://capz.sigs.k8s.io/topics/node-outbound-lb.html#warning, is that these values are not configurable and modifying them should result in a validation error. It should _not_ silently override the values. If that's what is happening right now then it's a bug. I think we should leave these values as is if already set and let the validation webhook do it's job by failing if the values are not what we expect (validation should already be in place). This also applies to the control plane outbound LB.

This is a bit out of scope for this PR so feel free to open an issue if you don't want to address this as part of this PR.

api/v1alpha4/azurecluster_default_test.go
CecileRobertMichon: these scenarios look great, thanks for adding them

api/v1alpha4/azurecluster_default_test.go
CecileRobertMichon: ditto here on smaller number of subnets

azure/services/subnets/subnets_test.go
```diff
@@ -338,6 +362,122 @@ func TestReconcileSubnets(t *testing.T) {
 				}).Times(1)
 			},
 		},
+		{
+			name:          "doesn't overwrite existing NAT Gateway",
```
jackfrancis, line 366: @CecileRobertMichon is this a correct UT to express the existing NAT Gateway expectation?

azure/services/subnets/subnets_test.go
```diff
@@ -338,6 +362,121 @@ func TestReconcileSubnets(t *testing.T) {
...
+						Name: "existing-natgateway",
+						NatGatewayIP: infrav1.PublicIPSpec{
+							Name: "existing-natgateway-ip-name",
+						},
+					},
+				}).Times(1)
+			},
+		},
+		{
+			name:          "spec has empty CIDR and ID data",
```
jackfrancis, line 432: @CecileRobertMichon PTAL

CecileRobertMichon: not sure I understand what scenario this is testing

azure/services/subnets/subnets_test.go
CecileRobertMichon: looks good

azure/services/subnets/subnets_test.go
jackfrancis: > We could also test that if CIDRBlocks and ID are empty in the existing spec, it will set them with the Azure subnet values.

CecileRobertMichon: ah, sorry for the confusion

what I meant there was infrav1.SubnetSpec returned has no ID/CIDR when calling `s.Subnet` and the Azure GET returns the values, which are used in `s.SetSubnet` (which means we're updating the spec to have these values persisted in AzureCluster.Spec)

CecileRobertMichon (commented):
lgtm pending comment on UT

api/v1alpha4/azurecluster_default_test.go
```diff
@@ -832,6 +1117,173 @@ func TestNodeOutboundLBDefaults(t *testing.T) {
 				},
 			},
 		},
+		{
```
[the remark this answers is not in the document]
jackfrancis: You are correct, I'm just sort of "stress testing" this case (from a human perspective) that yeah, we're definitely introspecting and counting all of them. If you think it's in fact not human-friendly to express such a large quantity, I'll reduce to 3.

api/v1alpha4/azurecluster_default_test.go
jackfrancis: I think what we're observing is that this condition (user-configurable value overrides of properties that aren't user-overridable) is artificial; as you say, validation will prevent this condition from ever reaching this point in the code flow. I vote that we just not add this artificial test case (I'll delete it).

azure/services/natgateways/natgateways.go
jackfrancis: Since this error condition is new, I'll remove the novel "length must equal 1" condition and leave the comment that acknowledges that we're simply throwing away all IPs except the first one (index 0) if there are > 1.

api/v1alpha4/azurecluster_default_test.go
jackfrancis: reduced to 3

azure/services/natgateways/natgateways.go
```diff
@@ -118,12 +122,27 @@ func (s *Service) getExisting(ctx context.Context, spec azure.NatGatewaySpec) (*
 	if err != nil {
 		return nil, err
 	}
+	// We must have a non-nil, non-"empty" PublicIPAddresses
+	if !(existingNatGateway.PublicIPAddresses != nil && len(*existingNatGateway.PublicIPAddresses) > 0) {
```
jackfrancis, line 126: @CecileRobertMichon no longer requiring exactly 1 item in the `existingNatGateway.PublicIPAddresses` slice

CecileRobertMichon (commented):
/lgtm
/assign @shysank

shysank:
/lgtm
@jackfrancis I had some trouble figuring out the actual issue, It'd be great if you can add the steps to reproduce the bug. From slack conversation with @CecileRobertMichon 
> you can repro this by creating a cluster with nat gateway on the node subnet, the first reconcile the nat gateway gets created, no outbound LB is created. After a few minutes, the nat gateway disappears from the subnet spec in the AzureCluster and an outbound LB gets created.

jackfrancis:
I think the core issue is described in the issue: https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1587

In summary, a Load Balancer created solely for providing outbound SNAT is unnecessary in a cluster configuration where a NAT Gateway is declared for doing just that. Hopefully that's clear. Certainly happy to clarify if it's not already.

CecileRobertMichon:
@jackfrancis can you please add details about the root cause that we observed after investigating? It's not obvious from reading the issue that the LB doesn't get created immediately and is due to the nat gateway being erased from the spec. We're also fixing another bug in this PR (IP ID comparison with IP name) which is missing from the description.

/approve
/hold
until above is addressed

PR #1609, opened 2021-08-14
fix: cluster to ammp watch; infra -> ammp requeue

alexeldeib:
**What type of PR is this?**

/kind bug
/kind flake

This fixes a few subtle issues. 

In AMMP controller, we used predicates.ClusterUnpausedAndInfrastructureReady ([link](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/7e769db4c026002083c7633e174d518eb907931b/exp/controllers/azuremanagedmachinepool_controller.go#L110)). This only considers **transitions** between states, rather than a simpler `predicates.ResourceNotPaused`. Because AzureManagedCluster is a complete no op and is immediately ready, the infra provider transitions to ready before the control plane is ready. AMMP reconciler needs both to be ready. When the control plane is not ready but infrastructure is, AMMP will requeue without an error, expecting a new reconcile event. However when cluster does get re-reconciled from AMCP becoming ready, Cluster will not transition between Paused states or InfrastructureReady states (that's what we're re-queuing for!). This means AMMP will not get reconciled until a future loop, adding ~10m depending on the ordering of events.

Using predicates.ResourceNotPaused fixes the bug, but adds many unnecessary reconcile loops. Instead, I made AMC's status.ready depend on AMCP's control plane endpoint being populated, which itself implies AMCP (and thus cluster) have control plane in ready status. This means the next Cluster loop will transition infrastructure to Ready as expected by AMMP watch/predicates.

There was also a bug in AMMP controller watching Cluster and mapping reconcile requests to cluster Infra ref, not to the machine pools themselves. That + the above fix save ~15-20 minutes of reconcile time, from ~40 -> 25 min.

Right now we have mutual watches between the other Managed* objects which helps hide how bad this would be. In a follow up, I'll remove the other unnecessary watches.

Fixes #

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Fix watch behavior in AzureManagedMachinePool and AzureManagedCluster to provision AKS clusters more quickly.
```

### DIFF
--- a/exp/controllers/azuremanagedcluster_controller.go
+++ b/exp/controllers/azuremanagedcluster_controller.go
@@ -153,9 +153,13 @@ func (r *AzureManagedClusterReconciler) Reconcile(ctx context.Context, req ctrl.
 		return reconcile.Result{}, errors.Wrap(err, "failed to init patch helper")
 	}
 
-	// Infrastructure must be ready before control plane. We should also enqueue
-	// requests from control plane to infra cluster to keep control plane endpoint accurate.
-	aksCluster.Status.Ready = true
+	// we use control plane endpoint to drive status.ready to align with capi expectations around state transitions.
+	// we use this here to enforce cluster only picks up the endpoint when AMCP is already finished reconciling.
+	// this ensures AMMP will always have control plane ready when infra transitions to ready,
+	// enabling us to use predicates.ClusterUnpausedAndInfrastructureReady to filter many unnecessary reconcile loops.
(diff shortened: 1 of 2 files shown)


alexeldeib:
running a few times to check for any flakes

alexeldeib:
missing provider IDs again on all tests, looks like ARM/VMSS issue

/retest

alexeldeib:
still seems too long...maybe I missed porting something from my other PR 

/hold

CecileRobertMichon:
/close 

please reopen if this is still needed

Issue #1108, opened 2021-01-07
Add tests for scope structs


shysank:
/kind feature

The `scope` structs (machine, cluster etc.) look like they only have simple getters and setters, but some of them have some subtle logic which return different values based on the state of the cluster/machine. Eg. PublicIPSpecs. Since we mock this behavior in `services`, we don't actually test the behavior of scope methods.

**Describe the solution you'd like**
Unit tests for all the scope methods, at least for the ones that have conditional logic.

**Environment:**

- cluster-api-provider-azure version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

shysank:
This is slightly more complicated than a good first issue. But, it has low impact and would be a good way to get a feel of the codebase. I'd suggest opening a pr for one scope first, and then do it for the rest. Happy to pair or answer any questions on this.

sonasingh46:
@shysank -- Would like to work in this issue. Thanks.

jsturtevant:
this was fixed in #1649 
/close

PR #1613, opened 2021-08-16
return empty string for outbound lb name if there is no outbound lb

### LINKED ISSUE #1612: Private clusters fails with Resource Not found error when no outbound lb is configured
/kind bug

1. Create private cluster using --private flavor
2. Remove oubound lb config
3. Apply changes

cluster fails to reconcile with ResourceNotFound error for outbound lb 

Cluster should be created successfully

**Environment:**

shysank:
**What type of PR is this?**
/kind bug

Fixes #1612

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Return empty string for outbound LB name if there is no outbound LB configured
```

### DIFF
--- a/azure/scope/cluster.go
+++ b/azure/scope/cluster.go
@@ -455,16 +455,25 @@ func (s *ClusterScope) NodeOutboundLBName() string {
 // OutboundLBName returns the name of the outbound LB.
 func (s *ClusterScope) OutboundLBName(role string) string {
 	if role == infrav1.Node {
-		return s.ClusterName()
+		if s.NodeOutboundLB() == nil {
+			return ""
+		}
+		return s.NodeOutboundLB().Name
 	}
 	if s.IsAPIServerPrivate() {
-		return azure.GenerateControlPlaneOutboundLBName(s.ClusterName())
+		if s.ControlPlaneOutboundLB() == nil {
+			return ""
+		}
+		return s.ControlPlaneOutboundLB().Name
 	}
 	return s.APIServerLBName()
 }
 
 // OutboundPoolName returns the outbound LB backend pool name.
 func (s *ClusterScope) OutboundPoolName(loadBalancerName string) string {
+	if loadBalancerName == "" {
+		return ""
+	}
 	return azure.GenerateOutboundBackendAddressPoolName(loadBalancerName)
 }
 
--- a/azure/scope/cluster_test.go
+++ b/azure/scope/cluster_test.go
@@ -89,3 +89,128 @@ func TestGettingSecurityRules(t *testing.T) {
 	g.Expect(err).ToNot(HaveOccurred())
 	g.Expect(len(subnet.SecurityGroup.SecurityRules)).To(Equal(2))
 }
+
+func TestOutboundLBName(t *testing.T) {
+	tests := []struct {
+		clusterName            string
+		name                   string
+		role                   string
+		apiServerLB            *infrav1.LoadBalancerSpec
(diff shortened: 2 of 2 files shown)


mboersma:
Looks good. Could we add unit test coverage to this change in `cluster_test.go`? (Although I know it doesn't currently call this part of the API.)

shysank:
> Looks good. Could we add unit test coverage to this change in `cluster_test.go`? (Although I know it doesn't currently call this part of the API.)

done. We also have #1108 to address this on a broader level.

azure/scope/cluster.go
```diff
@@ -455,16 +455,25 @@ func (s *ClusterScope) NodeOutboundLBName() string {
 // OutboundLBName returns the name of the outbound LB.
 func (s *ClusterScope) OutboundLBName(role string) string {
 	if role == infrav1.Node {
+		if s.NodeOutboundLB() == nil {
```
CecileRobertMichon, line 458: re-reading this, shouldn't it return s.NodeOutboundLB().Name instead of ClusterName if s.NodeOutboundLB() is not nil? Aren't we defaulting s.NodeOutboundLB().Name to ClusterName already?

shysank: we default it for public clusters which will always have a node outbound lb, for private clusters we leave it if nil. https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/87d3a087dae4c42d915f9cdc0751f799af430a49/api/v1alpha4/azurecluster_default.go#L201

I can add a check for private cluster if it's more readable.

CecileRobertMichon: What I mean is shouldn't this func be changed to:

```
func (s *ClusterScope) OutboundLBName(role string) string {
	if role == infrav1.Node {
		if s.NodeOutboundLB() == nil {
		    return ""
		}
		return s.NodeOutboundLB().Name
}
```
?

Otherwise we return the wrong name when the user is using a custom LB name

shysank: oh I see what you mean. Yes, we should change that although it wouldn't make a huge difference since we ignore and default it to cluster name anyways 😄  https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/87d3a087dae4c42d915f9cdc0751f799af430a49/api/v1alpha4/azurecluster_default.go#L226. IIRC, there was some constraint in cloud provider to have this as cluster name. I'll create a followup issue to address this.

azure/scope/cluster.go
```diff
@@ -455,16 +455,25 @@ func (s *ClusterScope) NodeOutboundLBName() string {
...
+			return ""
+		}
+		return s.NodeOutboundLB().Name
 	}
 	if s.IsAPIServerPrivate() {
+		if s.ControlPlaneOutboundLB() == nil {
+			return ""
+		}
 		return azure.GenerateControlPlaneOutboundLBName(s.ClusterName())
```
CecileRobertMichon, line 467: same comment here, shouldn't this be `return ControlPlaneOutboundLB().Name` ?

CecileRobertMichon: especially because that one _is_ user configurable

CecileRobertMichon (commented):
/lgtm
/approve

CecileRobertMichon:
@shysank for some reason it didn't add the approve label and hasn't merged yet so you can still add a release note

shysank:
> @shysank for some reason it didn't add the approve label and hasn't merged yet so you can still add a release note

done

PR #1624, opened 2021-08-18
Automate release via Github Action

### LINKED ISSUE #1569: Automate release via GH Action
/kind feature

We should automate as much of the release process as possible to avoid human error. The current release steps are fairly automated but still require a maintainer to go through them one by one: https://capz.sigs.k8s.io/developers/releasing.html

Easy win: We should add a Github Action that does all the artifact generation and upload as well as release notes generation when a tag is created. 

Longer-term: how can we automate the process of opening PRs to promote images?

CecileRobertMichon:
**What type of PR is this?**

**What this PR does / why we need it**: Makes the release process more automated. Pushing a tag will automatically cause a draft release to be created with needed artifacts and auto-generated release notes.

Fixes #1569 

Tested on my fork: https://github.com/CecileRobertMichon/cluster-api-provider-azure/runs/3356321723?check_suite_focus=true
Resulting release - not sure if this is visible to all: https://github.com/CecileRobertMichon/cluster-api-provider-azure/releases/tag/untagged-aa4bb391782dcc5daa9e  (empty release notes because no changes)

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```


.github/workflows/release.yaml
```diff
@@ -0,0 +1,37 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set env
+        run:  echo "RELEASE_TAG=${GITHUB_REF:10}" >> $GITHUB_ENV
+      - name: checkout code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Install go
+        uses: actions/setup-go@v2
```
devigned, line 21: Interesting this is using a different go version than the cover.yaml

devigned (commented):
Looks good. Just one question about Go version.

.github/workflows/release.yaml
CecileRobertMichon: cover.yaml is using the latest go version from https://golang.org/VERSION?m=text which is 1.17 now. For this I'm using 1.16 because that's the min version for needed for the release-notes tool.

I asked @alexeldeib why cover.yaml isn't using the setup-go action to install go and it just wasn't around when he first set this up so I think we could switch cover to also using the action instead of having our own script.

CecileRobertMichon: bumped to go 1.17 and changed cover to also use the setup go action

CecileRobertMichon:
@devigned @alexeldeib any other comments on this? I'd like to get it merged so I can use it to release v0.5.2

PR #1635, opened 2021-08-20
update templates to use 8GB of etcd db data

jackfrancis:
**What type of PR is this?**

/kind other

^-- Test configuration change

This PR updates the base etcd config for all cluster templates to use an 8 GB data size configuration. This is based on two things:

1) the 2 GB data default can be problematic (i.e., too small) for large clusters
2) 8 GB is the recommended maximum to use (there are ways to use more data but for now using the recommended max seems good and 4x compared to the current 2 GB size is a non-trivial increase)

See:

https://etcd.io/docs/v3.4/dev-guide/limit/

I think the guiding motivation of this change is to help unlock large cluster scenarios by rigorously testing this configuration recommendation. Because we are already using a 256 GB data disk volume for etcd in these templates, the additional etcd data size does not require any changes there.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
cluster templates now use 8 GB of etcd db data (was 2 GB)
```

### DIFF
--- a/templates/cluster-template-aad.yaml
+++ b/templates/cluster-template-aad.yaml
@@ -73,6 +73,8 @@ spec:
       etcd:
         local:
           dataDir: /var/lib/etcddisk/etcd
+          extraArgs:
+            quota-backend-bytes: "8589934592"
     diskSetup:
       filesystems:
       - device: /dev/disk/azure/scsi1/lun0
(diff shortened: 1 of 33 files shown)


CecileRobertMichon:
> Test configuration change

This is more than a test configuration change, it's a change to all the published flavors so we should add a release note for this user-facing change

devigned (approved):
/lgtm
/approve

PR #1638, opened 2021-08-20
optimal defaults for UltraSSD data disk

jackfrancis:
**What type of PR is this?**

/kind feature

This PR updates the default data disk cachingType if using UltraSSD to the only allowable option of "None". This doesn't strictly fix a bug, as it's possible to use UltraSSD on an data disk (e.g., etcd) right now by explicitly including the `cachingType: None` configuration in your AzureMachineTemplate; but because "None" is the only valid value we should default to it.

Also we assign sensible IOPS and throughput configuration based on the size of the data disk. See:

- https://docs.microsoft.com/en-us/azure/virtual-machines/disks-types#disk-size

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
optimal defaults for UltraSSD data disk
```

### DIFF
--- a/azure/services/virtualmachines/virtualmachines.go
+++ b/azure/services/virtualmachines/virtualmachines.go
@@ -41,7 +41,11 @@ import (
 
 const (
 	// UltraSSDStorageAccountType identifies the Ultra disk storage account type.
-	UltraSSDStorageAccountType = "UltraSSD_LRS"
+	UltraSSDStorageAccountType           = "UltraSSD_LRS"
+	MaxUltraSSDIOPS                      = 160000
+	UltraSSDDiskSizeIOPSMultiplier       = 300
+	MaxUltraSSDThroughput                = 2000
+	UltraSSDDiskSizeThroughputMultiplier = 75
 )
 
 // VMScope defines the scope interface for a virtual machines service.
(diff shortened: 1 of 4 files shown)


jackfrancis:
If the updated code surface area is correct, then the next thing we want to do is add validation: we don't want to allow any value other than "None" to make it out to the Azure APIs.

azure/services/virtualmachines/virtualmachines.go
```diff
@@ -436,8 +440,24 @@ func (s *Service) generateStorageProfile(ctx context.Context, vmSpec azure.VMSpe
...
 			location := s.Scope.Location()
-			if disk.ManagedDisk.StorageAccountType == UltraSSDStorageAccountType && !sku.HasLocationCapability(resourceskus.UltraSSDAvailable, location, vmSpec.Zone) {
-				return nil, azure.WithTerminalError(fmt.Errorf("vm size %s does not support ultra disks in location %s. select a different vm size or disable ultra disks", vmSpec.Size, location))
+			if disk.ManagedDisk.StorageAccountType == UltraSSDStorageAccountType {
+				if !sku.HasLocationCapability(resourceskus.UltraSSDAvailable, location, vmSpec.Zone) {
+					return nil, azure.WithTerminalError(fmt.Errorf("vm size %s does not support ultra disks in location %s. select a different vm size or disable ultra disks", vmSpec.Size, location))
+				}
+				// Configure max IOPS and throughput as per this reference:
+				// https://docs.microsoft.com/en-us/azure/virtual-machines/disks-types#disk-size
+				iops := int(disk.DiskSizeGB) * UltraSSDDiskSizeIOPSMultiplier
```
jackfrancis, line 449: This configuration appears to be read-only in fact, if you read the SDK comments:

https://github.com/Azure/azure-sdk-for-go/blob/main/services/compute/mgmt/2020-06-30/compute/models.go#L1147

I did some manual cluster builds and observed the IOPS and throughput config for the etcd data disk was still the default.

I was able to manually change the value in the portal:

image

Perhaps there's an API to do that async after the VM and data disk attachment has been made?

Otherwise as far as I can tell we get a disk with UltraSSD capabilities, but tuned for very low throughput:

image

cc @devigned

devigned:
@jackfrancis I think this might require a newer version of the SDK to mutate. The link you posted above is for API version `2020-06-30`. If you take a look at API version `2021-04-01`, the fields have become read/write https://github.com/Azure/azure-sdk-for-go/blob/ea571ad17c6ba95853ecf8c27acc7830682c4f22/services/compute/mgmt/2021-04-01/compute/models.go#L17979-L17982. 

This is super cool stuff and could enable us to maybe mutate these on the fly if we were to notice some higher demand for IOPS on a given VM.

/cc @juan-lee

jackfrancis:
I've opened up a PR to rev the Azure SDK package version so we can use that newer compute API version:

https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1643

jackfrancis:
@devigned ha! it's still read-only for data disks:

https://github.com/Azure/azure-sdk-for-go/blob/ea571ad17c6ba95853ecf8c27acc7830682c4f22/services/compute/mgmt/2021-04-01/compute/models.go#L2676

devigned:
> @devigned ha! it's still read-only for data disks:
> 
> https://github.com/Azure/azure-sdk-for-go/blob/ea571ad17c6ba95853ecf8c27acc7830682c4f22/services/compute/mgmt/2021-04-01/compute/models.go#L2676

What about: https://github.com/Azure/azure-sdk-for-go/blob/ea571ad17c6ba95853ecf8c27acc7830682c4f22/services/compute/mgmt/2021-04-01/compute/models.go#L17963-L17983

devigned:
Oops... the above link was to a VMSS related model.

devigned:
Looks like the data disks with configurable IOPS is only available for VMSS. Interestingly enough, it looks like disks `Update` seems to allow for mutations of IOPS for ultra SSDs: https://github.com/Azure/azure-sdk-for-go/blob/ea571ad17c6ba95853ecf8c27acc7830682c4f22/services/compute/mgmt/2021-04-01/compute/disks.go#L693-L724

jackfrancis:
>  it looks like disks Update seems to allow for mutations of IOPS for ultra SSDs

That does make sense, as you *can* update those values in the portal after the data disk has been attached to the VM.

CecileRobertMichon:
@jackfrancis what's the status on this one?

jackfrancis:
This will require a different approach to set IOPS defaults (we'll have to do a follow-up PUT against the disk resource to increase the IOPS)

so tl;dr we should close this

PR #1640, opened 2021-08-21
Support APIServerAccessProfile in AzureManagedControlPlane

### LINKED ISSUE #1605: Allow user to specify APIServerAccessProfile
/kind feature

Allow user to specify [APIServerAccessProfile](https://github.com/Azure/azure-sdk-for-go/blob/v55.2.0/services/containerservice/mgmt/2021-05-01/containerservice/models.go#L2067) in [AzureManagedControlPlaneSpec](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/exp/api/v1alpha4/azuremanagedcontrolplane_types.go#L28)

richardchen331:
**What type of PR is this?**
/kind feature

Adds APIServerAccessProfile support in AzureManagedControlPlane. 

Fixes #1605

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Allow specifying APIServerAccessProfile in AzureManagedControlPlane.
```


exp/api/v1alpha3/azuremanagedcontrolplane_types.go
```diff
@@ -94,6 +98,23 @@ type AADProfile struct {
...
 }
 
+// APIServerAccessProfile - access profile for AKS API server.
+type APIServerAccessProfile struct {
+	// AuthorizedIPRanges - Authorized IP Ranges to kubernetes API server.
+	// +optional
+	AuthorizedIPRanges []string `json:"authorizedIPRanges,omitempty"`
+	// EnablePrivateCluster - Whether to create the cluster as a private cluster or not.
+	// +optional
+	EnablePrivateCluster bool `json:"enablePrivateCluster,omitempty"`
```
nprokopic, line 108: If this field is optional, I believe a bool pointer `*bool` would be better. Same for `EnablePrivateClusterPublicFQDN`. WDYT?

richardchen331: @nprokopic Thanks for your feedback! I'm pretty new to golang and cluster api, and my (limited) understanding of using a pointer is that we could differentiate between a zero value and a not specified value, but then we have to deal with `nil`. For `EnablePrivateCluster`, I think explicitly passing in a zero value (`false`) is equivalent to not passing it (both tells Azure to not enable private cluster). Therefore I'm not aware of the benefit of using a pointer.

Could you help explain a bit more why a bool pointer is better? I think I must have missed something and I'd love to learn :) Thank you!

CecileRobertMichon: It is part of the k8s API conventions:

> Therefore, we ask that pointers always be used with optional fields that do not have a built-in nil value.

richardchen331: Got it, thanks for the reference! I'll change the fields to use pointers.

nprokopic: > It is part of the k8s API conventions

I vaguely remembered something like that, and that we had a similar discussion for some other CAPZ type some time ago, just couldn't find it.

Thanks Cecile for referencing this.

exp/api/v1alpha4/azuremanagedcontrolplane_default.go
```diff
@@ -73,3 +73,10 @@ func (r *AzureManagedControlPlane) setDefaultSubnet() {
 		r.Spec.VirtualNetwork.Subnet.CIDRBlock = defaultAKSNodeSubnetCIDR
 	}
 }
+
+// setDefaultAPIServerAccessProfile sets the default APIServerAccessProfile for an AzureManagedControlPlane.
+func (r *AzureManagedControlPlane) setDefaultAPIServerAccessProfile() {
+	if r.Spec.APIServerAccessProfile == nil {
```
nprokopic, line 79: nit: Is this required? Wouldn't just leaving `nil` also work?

@CecileRobertMichon are there some other examples of setting default values like this?

exp/api/v1alpha4/azuremanagedcontrolplane_webhook.go
```diff
@@ -306,3 +312,71 @@ func (r *AzureManagedControlPlane) validateSSHKey() error {
...
+			agg := kerrors.NewAggregate(allErrs.ToAggregate().Errors())
+			azuremanagedcontrolplanelog.Info("Invalid apiServerAccessProfile: %s", agg.Error())
+			return agg
+		}
+	}
+	return nil
+}
+
+// validateAPIServerAccessProfileUpdate validates update to APIServerAccessProfile.
+func (r *AzureManagedControlPlane) validateAPIServerAccessProfileUpdate(old *AzureManagedControlPlane) field.ErrorList {
```
nprokopic, line 335: There are quite a few field combinations here, it would be good to have unit tests for this.

Also, maybe using `DeepEqual` would make this shorter, like here for example https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/v0.5.2/api/v1alpha4/azurecluster_webhook.go#L109. @alexeldeib @CecileRobertMichon is using `DeepEqual` for comparison recommended/acceptable, or is handpicked field comparison like here more preferable?

exp/api/v1alpha4/azuremanagedcontrolplane_default.go
richardchen331: You are right, I don't think this is required. (I was thinking setting it so there will be less working in validation webhook, but i don't think this is a good idea, since it's not good for unit test)

exp/api/v1alpha4/azuremanagedcontrolplane_webhook.go
richardchen331: Sure, I'll add unit tests for this.

Unfortunately I cannot use `DeepEqual` here since one of the field (`AuthorizedIPRanges`) is mutable. One option is to move `AuthorizedIPRanges` out of `APIServerAccessProfile` so we can use `DeepEqual` here. I don't have a strong preference here. Would love to know what do people think.

nprokopic: You could normalize values before comparison, i.e. set mutable fields to empty/`nil` values in both operands (you would have to copy them first and compare normalized copies to avoid operand mutation).

No strong opinion for using `DeepEqual`, just a suggestion.

richardchen331: Thanks for the suggestion! I like that, it makes the code much simpler.

The downside is that it will prevent certain updates that is validate (e.g. change `EnablePrivateCluster` from `nil` to a pointer to `false`).

I've updated the code to use `DeepEqual`, but I'm open to change it back if people have strong opinions.

exp/api/v1alpha4/azuremanagedcontrolplane_webhook_test.go
```diff
@@ -606,6 +620,44 @@ func TestAzureManagedControlPlane_ValidateUpdate(t *testing.T) {
...
+					DNSServiceIP: to.StringPtr("192.168.0.0"),
+					Version:      "v1.18.0",
+				},
+			},
+			amcp: &AzureManagedControlPlane{
+				Spec: AzureManagedControlPlaneSpec{
+					DNSServiceIP: to.StringPtr("192.168.0.0"),
+					Version:      "v1.18.0",
+					APIServerAccessProfile: &APIServerAccessProfile{
+						EnablePrivateCluster: &boolTrue,
```
LochanRn, line 636: use EnablePrivateCluster: to.BoolPtr(true),

LochanRn (changes requested):

LochanRn:
Please squash all your commits into a single commit once you are done with all your changes

richardchen331:
> Please squash all your commits into a single commit once you are done with all your changes

@LochanRn  Thanks for the feedback! I updated the code to use `to.BoolPtr(true)` and squashed all commits.

exp/api/v1alpha4/azuremanagedcontrolplane_webhook.go
LochanRn: Just a suggestion, one way to avoid DeepEqual while comparing is use cmp.Diff from github.com/google/go-cmp/cmp.

alexeldeib: cmp uses reflection quite a bit, it just wraps some of the tricky bits of comparisons nicely.

alexeldeib:
/approve

nice work!

alexeldeib:
will proactively lgtm since @LochanRn already did but got cleared 🙂 

/lgtm

exp/api/v1alpha4/azuremanagedcontrolplane_webhook.go
LochanRn: ohh okay

exp/api/v1alpha3/azuremanagedcontrolplane_types.go
```diff
@@ -94,6 +98,23 @@ type AADProfile struct {
 	AdminGroupObjectIDs []string `json:"adminGroupObjectIDs"`
 }
 
+// APIServerAccessProfile - access profile for AKS API server.
+type APIServerAccessProfile struct {
```
alexeldeib, line 102: similar comment to https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1679/files#r706664083

I think you should add this for v1a4 only, and implement the conversion webhook? which should basically be a noop except storing the data to roundtrip v1a4 -> v1a3 -> v1a4 conversion

richardchen331: Thanks @alexeldeib ! Reverted change in v1a3 and added one line in `azuremanagedcontrolplane_conversion.go` to restore the field during conversion.

LochanRn:
@richardchen331  can you please also update the managedclusters documentation about APIServerAccessProfile. 

@alexeldeib over all this PR looks fine to me now.  Over to you for a final review.

richardchen331:
> @richardchen331 can you please also update the managedclusters documentation about APIServerAccessProfile.
> 
> @alexeldeib over all this PR looks fine to me now. Over to you for a final review.

Sure. I'll rebase and update the documentation after the other open PR is merged since they modify similar set of files :)

richardchen331:
@alexeldeib I just rebased this PR and also added documentation. Could you help take a final look? Thanks!

azure/types.go
```diff
@@ -407,6 +410,18 @@ type LoadBalancerProfile struct {
...
+// APIServerAccessProfile is the access profile for AKS API server.
+type APIServerAccessProfile struct {
+	// AuthorizedIPRanges - Authorized IP Ranges to kubernetes API server.
+	AuthorizedIPRanges []string
+	// EnablePrivateCluster - Whether to create the cluster as a private cluster or not.
+	EnablePrivateCluster *bool
+	// PrivateDNSZone - Private dns zone mode for private cluster.
+	PrivateDNSZone *string
+	// EnablePrivateClusterPublicFQDN - Whether to create additional public FQDN for private cluster or not.
+	EnablePrivateClusterPublicFQDN *bool
```
alexeldeib, line 422: I believe this will default to true on AKS side if private cluster is enabled. should we let that happen, or default it inside CAPZ?

it would only mean that a user who explicitly sets it would trigger a no-op reconcile, and a user who wants to change it will be able to.

alexeldeib:

probably related to the failures we've been discussing. I chatted a bit with @LochanRn about watch issues unrelated to the AKS backend issue, we still have some debugging to do

alexeldeib:
/approve
/lgtm

(adding both since this was already approved)

one minor comment but not a blocker

azure/types.go
richardchen331: Thanks for the feedback! I would slightly prefer letting AKS to handle the default. My rationales are:
1. This is true for other fields as well (e.g. [managedOutboundIPs](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/exp/api/v1alpha4/azuremanagedcontrolplane_types.go#L127), [sku tier](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/exp/api/v1alpha4/azuremanagedcontrolplane_types.go#L115))
2. We need to add some logic when choosing the default values (e.g. set `EnablePrivateClusterPublicFQDN` to true when `EnablePrivateCluster` is true)
3. Given 1,2, this will add some extra complexity to the CAPZ code
The downside is that there will be some unnecessary no-op reconciles, but i feel that's probably fine?

I'm open to change it if you feel strongly about it :)

alexeldeib: Fine w me, I asked because I know this defaulting too true has surprised some folks in the past

alexeldeib: But that’s really an aks issue

PR #1641, opened 2021-08-23
Update MINIMUM_KIND_VERSION to v0.10.0

feiskyer:
**What type of PR is this?**

/kind failing-test

Update MINIMUM_KIND_VERSION to v0.10.0, So that the kind cluster would be created with version v1.20.2, which also supports IngressClass

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
fix: update MINIMUM_KIND_VERSION to v0.10.0
```

### DIFF
--- a/hack/ensure-kind.sh
+++ b/hack/ensure-kind.sh
@@ -19,7 +19,7 @@ set -o nounset
 set -o pipefail
 
 GOPATH_BIN="$(go env GOPATH)/bin/"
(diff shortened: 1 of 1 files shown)


CecileRobertMichon:
/lgtm
/approve

to unblock @feiskyer 

@devigned we should follow up on this and see if we should pin the version of ingress-nginx instead of applying the master version which might break us again like this in the future

Issue #1353, opened 2021-04-30
MTU for Linux Calico VXLAN deployments should be lowered to below 1400


jsturtevant:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]
As part of #1335 it was identified that Linux uses Calico's vxlan and the MTU is over 1400 (which is azure networks MTU).  This doesn't cause any direct issue but could cause throughput issues because the vxlan traffic will get fragmented.

**What did you expect to happen:**
Packets shouldn't get fragmented and throughput affected

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

the following should be lowered and validated that packets are not fragmented.

https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/2203c184b97eb8bd3e95bd9ac5ff099c6e235c04/templates/addons/calico.yaml#L13-L19

**Environment:**

- cluster-api-provider-azure version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

CecileRobertMichon:
should the title say below 1400?

CecileRobertMichon:
Do you have a recommendation for what the new value should be?

jsturtevant:
I believe this could be set to 1392 which would give room for the encapsulation but it would be good to validate it.

CecileRobertMichon:
/good-first-issue

the ask is to set the `veth_mtu` value to 1392. @jsturtevant do you have any guidance for validating that packets are not fragmented?

PR #1644, opened 2021-08-23
[WIP] test default calico 3.2.0 spec in capz

jackfrancis:
**What type of PR is this?**

/kind other

This PR replaces the vxlan-curated calico spec with the vanilla spec published by the Calico project here:

https://docs.projectcalico.org/v3.20/manifests/calico.yaml

Initially, this PR just overwrites the calico template with the above bits. If this test succeeds we can figure out a more elegant way of integrating it with capz CI.

Also, if we prove that this works and we want to make this change, we need to make sure to update the cluster-api book quick start guide for Azure.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note

```

### DIFF
--- a/templates/addons/calico.yaml
+++ b/templates/addons/calico.yaml
@@ -10,13 +10,13 @@ data:
   # Typha is disabled.
   typha_service_name: "none"
   # Configure the backend to use.
-  calico_backend: "vxlan"
-  # On Azure, the underlying network has an MTU of 1400, even though the network interface will have an MTU of 1500.
-  # We set this value to 1350 for “physical network MTU size minus 50” since we use VXLAN, which uses a 50-byte header.
-  # If enabling Wireguard, this value should be changed to 1340 (Wireguard uses a 60-byte header).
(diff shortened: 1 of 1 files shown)


templates/addons/calico.yaml
```diff
@@ -10,13 +10,13 @@ data:
...
-  # If enabling Wireguard, this value should be changed to 1340 (Wireguard uses a 60-byte header).
-  # https://docs.projectcalico.org/networking/mtu#determine-mtu-size
-  veth_mtu: "1350"
-  
+  calico_backend: "bird"
+
+  # Configure the MTU to use for workload interfaces and tunnels.
+  # By default, MTU is auto-detected, and explicitly setting this field should not be required.
+  # You can override auto-detection by providing a non-zero value.
+  veth_mtu: "0"
```
CecileRobertMichon, line 18: if we do go ahead with this change we still need the veth_mtu update (see #1353 and the comment above)

PR #1652, opened 2021-08-24
make NAT Gateway the default outbound SNAT template config

jackfrancis:
**What type of PR is this?**

/kind other

This PR updates the set of regularly tested cluster templates to use NAT Gateway for all cluster configurations that work with NAT Gateway (i.e., everything except ipv6 configs).

This PR doesn't change any underlying cluster configuration default enforcement, or add any additional validation. This is simply to increase the test coverage for NAT Gateway, which we plan to eventually make the *only* supported solution for node outbound SNAT.

Related to #1532

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
make NAT Gateway the default outbound SNAT template config
```


CecileRobertMichon:
Do you mind changing the release note to omit the word "default"? I would say something like "add NAT Gateway for outbound SNAT to all reference templates" to avoid confusion and make it clear that NAT GW is not actually the "default" yet.

templates/flavors/base/cluster-template.yaml
```diff
@@ -29,6 +29,19 @@ spec:
...
       name: "${AZURE_VNET_NAME:=${CLUSTER_NAME}-vnet}"
+      cidrBlocks:
+        - 10.0.0.0/8
+    subnets:
+    - cidrBlocks:
+      - 10.0.0.0/16
+      name: control-plane-subnet
+      role: control-plane
+    - cidrBlocks:
+      - 10.1.0.0/16
```
CecileRobertMichon, line 40: is the idea to make vnet / subnet cidr ranges more explicit in templates? These values aren't actually required to be set as far as I understand

templates/test/ci/cluster-template-prow-custom-vnet.yaml
```diff
@@ -47,6 +47,8 @@ spec:
       securityGroup:
         name: node-nsg
     vnet:
+      cidrBlocks:
+      - 10.0.0.0/8
```
CecileRobertMichon, line 51: for this one I would omit the cidrBlock, since the idea is that we are pre-creating the vnet out of band, we don't want the CIDR to be set (it will be ignored and overwritten with the actual value of the cidr range during reconcile)

CecileRobertMichon:
@shysank @nader-ziada thoughts on this? Does it make sense to you to opt-in the templates to NAT Gateway before actually making it the default so we can get some mileage on it before switching the default?

nader-ziada:
> @shysank @nader-ziada thoughts on this? Does it make sense to you to opt-in the templates to NAT Gateway before actually making it the default so we can get some mileage on it before switching the default?

yes, makes sense to use it for the tests which is the main purpose of the templates, but we just need to make it clear in docs somehow so users looking at the templates know (maybe a comment in the template)

shysank:
> @shysank @nader-ziada thoughts on this? Does it make sense to you to opt-in the templates to NAT Gateway before actually making it the default so we can get some mileage on it before switching the default?

+1, I think it's a good starting step to making nat gateway the default for outbound connections.

templates/flavors/base/cluster-template.yaml
jackfrancis: I just copied the relevant configuration stuffs from the existing NAT Gateway template spec.

Happy to make it more concise if that's better.

CecileRobertMichon: I don't think it's necessary to actually include it in here. It will get defaulted to these same values if not specified. I guess the upside on including it is that it makes the configuration more explicit and transparent, the downside is that it's less concise. WDYT?

CecileRobertMichon: @jackfrancis PTAL when you get a chance

jackfrancis: Is this a correct, more concise definition to put into the base and then apply everywhere?:

```
    vnet:
      name: "${AZURE_VNET_NAME:=${CLUSTER_NAME}-vnet}"
    subnets:
    - name: control-plane-subnet
      role: control-plane
    - name: node-subnet
      natGateway:
        name: node-natgateway
      role: node
```

CecileRobertMichon: yes that looks good

CecileRobertMichon:
/test pull-cluster-api-provider-azure-e2e-full
/test pull-cluster-api-provider-azure-capi-e2e
/test pull-cluster-api-provider-azure-conformance-v1alpha4
/test pull-cluster-api-provider-azure-upstream-v1alpha4-windows
/test pull-cluster-api-provider-azure-conformance-with-ci-artifacts
/test pull-cluster-api-provider-azure-windows-upstream-with-ci-artifacts

running all the tests since this changes templates

CecileRobertMichon:
/test pull-cluster-api-provider-azure-e2e-full

seems like the job was interrupted

PR #1658, opened 2021-08-26
Merge all Tilt settings into 'tilt-settings.json'

Jont828:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**: Removes unused settings and `tilt_config.json` and moves the settings into `tilt-settings.json` so there is only one config file. 

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*: N/A

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -220,34 +220,28 @@ def create_crs():
     local("kubectl apply -f templates/addons/calico-resource-set.yaml")
     local("kubectl apply -f templates/addons/flannel-resource-set.yaml")
 
-# run worker clusters specified from 'tilt up' or in 'tilt_config.json'
+# create flavor resources from cluster-template files in the templates directory 
 def flavors():
-    config.define_string_list("templates-to-run", args=True)
-    config.define_string_list("worker-flavors")
-    cfg = config.parse()
-    worker_templates = cfg.get('templates-to-run', [])
-   
     substitutions = settings.get("kustomize_substitutions", {})
     ssh_pub_key_B64 = "AZURE_SSH_PUBLIC_KEY_B64"
     ssh_pub_key_path = "$HOME/.ssh/id_rsa.pub"
     if substitutions.get(ssh_pub_key_B64):
         os.environ.update({ssh_pub_key_B64: substitutions.get(ssh_pub_key_B64)})
     else:
-        print("{} was not specified in tilt_config.json, attempting to load {}".format(ssh_pub_key_B64, ssh_pub_key_path))
+        print("{} was not specified in tilt-settings.json, attempting to load {}".format(ssh_pub_key_B64, ssh_pub_key_path))
         os.environ.update({ssh_pub_key_B64: base64_encode_file(ssh_pub_key_path)})
 
     ssh_pub_key = "AZURE_SSH_PUBLIC_KEY"
     if substitutions.get(ssh_pub_key):
         os.environ.update({ssh_pub_key: substitutions.get(ssh_pub_key)})
     else:
-        print("{} was not specified in tilt_config.json, attempting to load {}".format(ssh_pub_key_B64, ssh_pub_key_path))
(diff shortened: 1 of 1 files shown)


Tiltfile
```diff
@@ -220,34 +220,28 @@ def create_crs():
     local("kubectl apply -f templates/addons/calico-resource-set.yaml")
     local("kubectl apply -f templates/addons/flannel-resource-set.yaml")
 
-# run worker clusters specified from 'tilt up' or in 'tilt_config.json'
+# run worker clusters specified in 'tilt-settings.json'
```
CecileRobertMichon, line 223:
```suggestion
# create flavor resources from cluster-template files in the templates directory 
```

Tiltfile
```diff
@@ -220,34 +220,28 @@ def create_crs():
...
+        warn("{} was not specified in tilt-settings.json, attempting to load {}".format(ssh_pub_key_B64, ssh_pub_key_path))
         os.environ.update({ssh_pub_key: read_file_from_path(ssh_pub_key_path)})
 
-    templatelist = [ item for item in listdir("./templates") ]
-    for template in templatelist:
-        if template not in worker_templates and os.path.basename(template).startswith("cluster-template"):
-            worker_templates.append(template)
-    for template in worker_templates:
+    template_list = [ item for item in listdir("./templates") ]
+    template_list = [ template for template in template_list if os.path.basename(template).startswith("cluster-template") ]
```
CecileRobertMichon, line 242: this is unrelated to this change, but I'm wondering if we should change this to `endswith("yaml")` instead of `startswith("cluster-template")` so that it's less restrictive if you have something like "my-foo-template.yaml` 

thoughts for later

CecileRobertMichon: up to you, it's related enough to be part of the same PR as a separate commit so if you want to go ahead and make the change that's fine with me

Jont828: Added the change, does it look good to squash/merge?

CecileRobertMichon: had one more minor comment here https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1658#discussion_r700517620

CecileRobertMichon (commented):
/ok-to-test

Tiltfile
```diff
@@ -220,34 +220,28 @@ def create_crs():
...
-        print("{} was not specified in tilt_config.json, attempting to load {}".format(ssh_pub_key_B64, ssh_pub_key_path))
+        warn("{} was not specified in tilt-settings.json, attempting to load {}".format(ssh_pub_key_B64, ssh_pub_key_path))
         os.environ.update({ssh_pub_key_B64: base64_encode_file(ssh_pub_key_path)})
 
     ssh_pub_key = "AZURE_SSH_PUBLIC_KEY"
     if substitutions.get(ssh_pub_key):
         os.environ.update({ssh_pub_key: substitutions.get(ssh_pub_key)})
     else:
-        print("{} was not specified in tilt_config.json, attempting to load {}".format(ssh_pub_key_B64, ssh_pub_key_path))
+        warn("{} was not specified in tilt-settings.json, attempting to load {}".format(ssh_pub_key_B64, ssh_pub_key_path))
```
CecileRobertMichon, line 238: do we really want to warn here? I don't know if it's worth making the UI yellow when defaulting to the ssh key file, @mboersma what do you think?

Jont828: I decided to add it as a warning because if the user made a mistake in setting the Azure key or the Azure base 64 key, the code could fall through to the default key file without them noticing. For example, if they made a typo in "AZURE_SSH_PUBLIC_KEY," the variable would not be set.

mboersma: If we're just pursuing a fallback strategy, I don't think it's worth issuing a warning here. The yellow warnings in tilt feel serious to me.

CecileRobertMichon (commented):
/lgtm

PR #1610, opened 2021-08-16
Add long running operation types, conditions, and helpers

CecileRobertMichon:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**: This is the first PR to implement #1541. It implements LongRunningOperationStates, additional conditions, and async resource creator and deleter interfaces. ~It enables the async reconcile and delete for 3 services as a POC: resource groups, vnets, and security groups. Those 3 were chosen because they show how this would work across services with different levels of complexity (groups is very simple, vnets cares about managed vs. unmanaged, and NSG needs to merge with the existing state).~

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**TODOs**:

- [ ] unit tests
- [ ] update spec with Result()
- [ ] fix BYO vnet scenarios

**Release note**:

```release-note
Add long-running operation types, conditions, and helpers
```

PR #1667, opened 2021-09-01
Make group delete async

CecileRobertMichon:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**: Follow up on #1610, this PR enables async reconcile for groups. See #1541  for the proposal.

Fixes #

/hold

will rebase on top of #1610 once it merges.

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Make group delete async
```

### DIFF
--- a/api/v1alpha4/conditions_consts.go
+++ b/api/v1alpha4/conditions_consts.go
@@ -20,6 +20,8 @@ import clusterv1 "sigs.k8s.io/cluster-api/api/v1alpha4"
 
 // AzureCluster Conditions and Reasons.
 const (
+	// NetworkInfrastructureReadyCondition reports of current status of cluster infrastructure.
+	NetworkInfrastructureReadyCondition clusterv1.ConditionType = "NetworkInfrastructureReady"
 	// NamespaceNotAllowedByIdentity used to indicate cluster in a namespace not allowed by identity.
 	NamespaceNotAllowedByIdentity = "NamespaceNotAllowedByIdentity"
 )
--- a/azure/services/groups/client.go
+++ b/azure/services/groups/client.go
@@ -21,16 +21,20 @@ import (
 
 	"github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2019-05-01/resources"
 	"github.com/Azure/go-autorest/autorest"
+	azureautorest "github.com/Azure/go-autorest/autorest/azure"
+	"github.com/pkg/errors"
 
 	"sigs.k8s.io/cluster-api-provider-azure/azure"
+	"sigs.k8s.io/cluster-api-provider-azure/util/reconciler"
 	"sigs.k8s.io/cluster-api-provider-azure/util/tele"
 )
 
 // client wraps go-sdk.
 type client interface {
 	Get(context.Context, string) (resources.Group, error)
-	CreateOrUpdate(context.Context, string, resources.Group) (resources.Group, error)
-	Delete(context.Context, string) error
+	CreateOrUpdateAsync(context.Context, azure.ResourceSpecGetter) (azureautorest.FutureAPI, error)
+	DeleteAsync(context.Context, azure.ResourceSpecGetter) (azureautorest.FutureAPI, error)
+	IsDone(context.Context, azureautorest.FutureAPI) (bool, error)
 }
 
 // azureClient contains the Azure go-sdk Client.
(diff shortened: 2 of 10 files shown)


CecileRobertMichon:
https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1667/commits/de4e1992a80aa4a10f786b8db3d1b84ceeeea4da is the relevant commit for this PR (the other changes are already part of #1610)

api/v1alpha4/conditions_consts.go
```diff
@@ -20,6 +20,8 @@ import clusterv1 "sigs.k8s.io/cluster-api/api/v1alpha4"
 
 // AzureCluster Conditions and Reasons.
 const (
+	// NetworkInfrastructureReadyCondition reports of current status of cluster infrastructure.
+	NetworkInfrastructureReadyCondition clusterv1.ConditionType = "NetworkInfrastructureReady"
```
CecileRobertMichon, line 24: re-added so that the cluster doesn't show as Ready until all services are done reconciling (might remove once each service has its own condition in place to use as summary)

azure/services/groups/groups.go
```diff
@@ -55,65 +60,45 @@ func (s *Service) Reconcile(ctx context.Context) error {
...
+	return err
 }
 
-// Delete deletes the resource group with the provided name.
+// Delete deletes the resource group if it is managed by us.
 func (s *Service) Delete(ctx context.Context) error {
 	ctx, span := tele.Tracer().Start(ctx, "groups.Service.Delete")
 	defer span.End()
 
+	ctx, cancel := context.WithTimeout(ctx, reconciler.DefaultAzureServiceReconcileTimeout)
```
shysank, line 78: We should think about handling the cancellation signal here. Right now, we are just cleaning up the resources, but we should also think about what will happen if client.Get for eg. takes longer than "DefaultAzureServiceReconcileTimeout".  It's not going to break anything, but we are not returning even if it exceeds the timeout which was the original intention. wdyt?

controllers/azurecluster_controller.go
```diff
@@ -159,7 +159,7 @@ func (r *AzureClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request
 			return reconcile.Result{}, err
 		}
 		if !scope.IsClusterNamespaceAllowed(ctx, r.Client, identity.Spec.AllowedNamespaces, azureCluster.Namespace) {
-			conditions.MarkFalse(azureCluster, clusterv1.ReadyCondition, infrav1.NamespaceNotAllowedByIdentity, clusterv1.ConditionSeverityError, "")
+			conditions.MarkFalse(azureCluster, infrav1.NetworkInfrastructureReadyCondition, infrav1.NamespaceNotAllowedByIdentity, clusterv1.ConditionSeverityError, "")
```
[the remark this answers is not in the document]
CecileRobertMichon: the Ready condition gets set as the summary as part of PatchObject() which is called at the end of every loop https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1667/files#diff-b74c6acb4629abe58f8286da9939563cc4404e09fe9966eb61dbde9b87274189R553

this commit reverts part of #1610 which removed NetworkInfrastructureReadyCondition so that we have a generic condition for everything that isn't async yet so that the summary doesn't set Ready as "true" until all the resources are actually ready (Summary also takes NetworkInfrastructureReadyCondition as one of the conditions so Ready is true when both NetworkInfrastructureReadyCondition and all the individual resource conditions are True)

azure/services/groups/groups.go
CecileRobertMichon: > but we are not returning even if it exceeds the timeout which was the original intention

can you elaborate on this? 

maybe https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1684/files#diff-5bb14c98c52ecfa3dad6cf4745c9378b7624fdc60f0910343534ddac18408fb1R62 is a better illustration of how this works not just for creates... if the context expires in this function it will return an error which will cause that reconcile loop to fail. For long-running operations, like deletes and creates, we expect that to happen which is why we return a transient "operation not done" error which causes a requeue. If the overall loop times out, eg. a Get takes longer than it should, then maybe it's a good thing to error out and retry later? This is meant to be as an upper bound, ie. this service should never take more than 12 seconds to run the Delete/Reconcile function, unless something went wrong.

controllers/azurecluster_controller.go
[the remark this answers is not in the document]
shysank: Got it, Thanks for the explanation.

azure/services/groups/groups.go
shysank: > if the context expires in this function it will return an error which will cause that reconcile loop to fail

Not sure if that'll happen, though. We'll need to watch for `ctx.Done()` to get the cancellation signal, and return error accordingly. Something similar to what `future.WaitForCompletionRef` does [here](https://github.com/Azure/go-autorest/blob/492f94bc3a562197e229f044dbfabcc0ecb2489a/autorest/azure/async.go#L252). Does that make sense?

CecileRobertMichon: I see what you're saying now. I guess this isn't really a long-running function where we want to continuously do something and timeout though. The reason for doing a cxt timeout here is that the collection of all the shorter timeouts should never exceed this bigger timeout. When go sets a new timeout deadline, it will select the soonest between the parent context's deadline and the new deadline, so this is meant more as an intermediate timeout (in this case it doesn't really apply though because the group delete only performs one delete operation so it's a 1:1 mapping), same way that the reconcile loop timeout is a global timeout of all the long-running operations we have. Maybe we can discuss this at office hours tomorrow it will be easier to talk it through.

shysank: Ah I see, so this timeout is just to make sure that the `time(all async operations) < DefaultReconcileTimeout` , and not necessarily `time(service.Reoncile/Delete) < DefaultReconcileTimeout`. I think it makes sense, there shouldn't be any network calls outside of the async interface. ResourceGroups is a special case I guess to check for `IsManaged`, and for other services this will be picked up from a cache.

CecileRobertMichon:
would be great to get a second lgtm on this if anyone has time to put some 👀  on it 

@nader-ziada @devigned @whites11

CecileRobertMichon:
I changed the service name from "groups" to "group" so it reads "group deleting" instead of "groups deleting" in the logs (there is always only one group per cluster)

azure/services/groups/client.go
```diff
@@ -63,27 +67,94 @@ func (ac *azureClient) Get(ctx context.Context, name string) (resources.Group, e
...
+	if err != nil {
+		return false, errors.Wrap(err, "failed checking if the operation was complete")
+	}
+
+	return done, nil
+}
+
+// resourceGroupParams returns the desired resource group parameters from the given spec.
+func (ac *azureClient) resourceGroupParams(ctx context.Context, spec azure.ResourceSpecGetter) (*resources.Group, error) {
+	var params interface{}
```
devigned, line 131:
```suggestion
        ctx, span := tele.Tracer().Start(ctx, "groups.azureClient.resourceGroupParams")
	defer span.End()
	
	var params interface{}
```

azure/services/groups/spec.go
```diff
@@ -0,0 +1,66 @@
...
+// OwnerResourceName is a no-op for groups.
+func (s *GroupSpec) OwnerResourceName() string {
+	return "" // not applicable
+}
+
+// Parameters returns the parameters for the group.
+func (s *GroupSpec) Parameters(existing interface{}) (interface{}, error) {
+	if existing != nil {
+		// rg already exists, nothing to update.
+		return nil, nil
```
devigned, line 54: What if additional tags were added?

devigned (changes requested):
Just a couple suggestions. Overall looks great!

azure/services/groups/spec.go
nader-ziada: I don't think it would update currently

nader-ziada:
lgtm, pending David's comments

azure/services/groups/spec.go
CecileRobertMichon: indeed, but we weren't doing that before either: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1667/files#diff-036de5e90528343ff92467735b06ba4c22abbb0616cc3911820360bc37aea755L59 

I tried to avoid making any sort of bug fixes / behavioral changes (besides async stuff) as part of these PRs because they are already so big. How about I file a bug issue to fix it separately?

azure/services/groups/client.go
CecileRobertMichon: do we want to be starting spans for functions that don't make any outside calls ? this function is just a getter so execution should be quick.

azure/services/groups/spec.go
devigned: sgtm

azure/services/groups/client.go
devigned: I'd agree, but just below this is `existingRG, err := ac.Get(ctx, spec.ResourceName())`, which does make a call. If you were looking at the spans, you would see the parent span come from the grandparent func. The spans would not tell the code story as well if you don't fill in the span names all the way down. WDYT?

CecileRobertMichon: oh my bad, I thought this comment was on the paramaters func in spec.go

you're totally right, this deserves a span, adding it

CecileRobertMichon: added

azure/services/groups/spec.go
CecileRobertMichon: https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1696

devigned (approved):
/lgtm
/approve

CecileRobertMichon:
/retest

windows vmss flake

PR #1649, opened 2021-08-24
add unit tests for machine scope methods

sonasingh46:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:
This PR adds unit test for machine scope methods as described in this issue: 
https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1108

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes # https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1108

**Special notes for your reviewer**:

Have added UT for the following methods: 
```
PublicIPSpecs()
InboundNatSpecs()
RoleAssignmentSpecs()
VMExtensionSpecs()
Subnet()
AvailabilityZone()
AvailabilitySet()
Role()
IsControlPlane()
Namespace()
VMState()
GetVMImage()
```
Will be adding couple of thus removing WIP 
_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
None
```

PR #1676, opened 2021-09-08
Avoid duplicate NIC for AzureMachine if AllocatePublicIP is set to true.

whites11:
**What type of PR is this?**

/kind cleanup

Currently, when `AzureMachine.Spec.AllocatePublicIP` is set to true, it causes the Azure virtual machine to have 2 NIC cards attached.
Both NICs are attached to the same subnet, one uses the Public IP address and the other is connected to the Load Balancer for egress purposes.
The reason it is like this is that it is not supported to have both a public IP and a load balancer on the same NIC (error `NicWithPublicIpCannotReferencePoolWithOutboundRule`).

This PR suggests a modification to the dual NIC approach that works like this:

- If the `AllocatePublicIP` flag is false and the NAT gateway feature is disabled, then a load balancer is set up for egress purposes.
- If the `AllocatePublicIP` flag is true, the load balancer is NOT created and egress happens through the public IP.
- If the `AllocatePublicIP` flag is true and the client wants to have a uniform IP address for egress purposes, he/she can enable the NAT gateway feature on the subnet.

Having multiple NICs (especially if they are connected to the same subnet) is an unnecessary complication. It requires a special network configuration inside the virtual machines (route tables for example) and in general, might lead to unwanted/weird behavior. I doubt the generic CAPZ user would benefit from having multiple NICs on a VM.

Let me know what you think about this!

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Disable egress load balancer for AzureMachines when `AllocatePublicIP` is set to true.

CLEANUP ACTION REQUIRED: After you delete an AzureMachine created with a previous version, you might want to manually clean up the Network Interface named `<azuremachine name>-public-nic`.
```

### DIFF
--- a/azure/scope/machine.go
+++ b/azure/scope/machine.go
@@ -169,26 +169,17 @@ func (m *MachineScope) NICSpecs() []azure.NICSpec {
 		}
 	}
 
-	// If Nat Gateway is not enabled, then the NIC needs to reference the LB to get outbound traffic.
-	if m.Role() == infrav1.Node && !m.Subnet().IsNatGatewayEnabled() {
+	// If Nat Gateway is not enabled and node has no public IP, then the NIC needs to reference the LB to get outbound traffic.
+	if m.Role() == infrav1.Node && !m.Subnet().IsNatGatewayEnabled() && !m.AzureMachine.Spec.AllocatePublicIP {
 		spec.PublicLBName = m.OutboundLBName(m.Role())
 		spec.PublicLBAddressPoolName = m.OutboundPoolName(m.OutboundLBName(m.Role()))
 	}
-	specs := []azure.NICSpec{spec}
-	if m.AzureMachine.Spec.AllocatePublicIP {
-		specs = append(specs, azure.NICSpec{
-			Name:                  azure.GeneratePublicNICName(m.Name()),
-			MachineName:           m.Name(),
-			VNetName:              m.Vnet().Name,
-			VNetResourceGroup:     m.Vnet().ResourceGroup,
-			SubnetName:            m.AzureMachine.Spec.SubnetName,
-			PublicIPName:          azure.GenerateNodePublicIPName(m.Name()),
-			VMSize:                m.AzureMachine.Spec.VMSize,
-			AcceleratedNetworking: m.AzureMachine.Spec.AcceleratedNetworking,
-		})
+
+	if m.Role() == infrav1.Node && m.AzureMachine.Spec.AllocatePublicIP {
+		spec.PublicIPName = azure.GenerateNodePublicIPName(m.Name())
 	}
 
-	return specs
+	return []azure.NICSpec{spec}
 }
 
 // NICNames returns the NIC names.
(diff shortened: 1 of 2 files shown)


CecileRobertMichon:
@whites11 I agree this is a wanted simplification. To confirm, there is no additional required setup to get egress working when the node has a public IP, correct?

One thing I see with this change is that it could be "breaking" to some users who were previously using the allocatePublicIP functionality and were expecting the second NIC. The other issue is that scaling down or deleting machines could leak existing NICs since those won't get deleted anymore with this code. This should only impact a small number of users (if any) but we should document it clearly in the release notes section with an "action required".

whites11:
> @whites11 I agree this is a wanted simplification. To confirm, there is no additional required setup to get egress working when the node has a public IP, correct?

Yes, tested both with and without NAT gateway and it behaves as expected.

> One thing I see with this change is that it could be "breaking" to some users who were previously using the allocatePublicIP functionality and were expecting the second NIC. The other issue is that scaling down or deleting machines could leak existing NICs since those won't get deleted anymore with this code. This should only impact a small number of users (if any) but we should document it clearly in the release notes section with an "action required".

Makes complete sense

CecileRobertMichon:
@rsmitty, not sure if you all are still using the public IPs on nodes functionaility at talos but this affect you. Can you please loop in anyone else who needs to have eyes on this for awareness?

@whites11 I'll let you amend the release note in the PR description to include what we discussed above.

azure/scope/machine.go
```diff
@@ -169,26 +169,17 @@ func (m *MachineScope) NICSpecs() []azure.NICSpec {
 		}
 	}
 
-	// If Nat Gateway is not enabled, then the NIC needs to reference the LB to get outbound traffic.
-	if m.Role() == infrav1.Node && !m.Subnet().IsNatGatewayEnabled() {
+	// If Nat Gateway is not enabled and node has no public IP, then the NIC needs to reference the LB to get outbound traffic.
+	if m.Role() == infrav1.Node && !m.Subnet().IsNatGatewayEnabled() && !m.AzureMachine.Spec.AllocatePublicIP {
```
CecileRobertMichon, line 173: It would be great to add a small unit test for this since https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1649 merged

whites11: added tests

whites11:
> @whites11 I'll let you amend the release note in the PR description to include what we discussed above.

I added a sentence, just not sure if it fulfills what we normally do in these scenarios. Feedback appreciated, thx

CecileRobertMichon (commented):
/lgtm

whites11:
Dang, I forgot to ask about this PR during office hours. Can we maybe move it on?

CecileRobertMichon:
@whites11 looking for a second reviewer

/cc @shysank @mboersma

mboersma (approved):
/lgtm

PR #1677, opened 2021-09-09
Wire `TerminateNotificationTimeout` to enable termination events on VMSSes

whites11:
**What type of PR is this?**
/kind bug

I tried enabling termination event on a MachinePool by setting the `AzureMachinePool/spec/template/terminateNotificationTimeout` to 5 but the resulting Azure scale set came up with termination events disabled:

disabled

I then realised the mentioned field was unused in CAPZ.
This PR provides the needed wiring in order to allow enabling of termination events notification.

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Fixed wiring for AzureMachinePool's terminateNotificationTimeout field.
```


devigned (commented):
Hmm... looks like the missing functionality was an oversight. However, I'm curious how this will act without a component running on the VMSS VM that will [respond to the termination notification via IMDS](https://docs.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-terminate-notification). I think it should block for the entire time, rather than doing some work, then notifying IMDS the VM can be deprovisioned when the work has been done.

There probably needs to be a component baked into the nodes that listens to these notifications, cordons and drains the node, then responds to the notification telling Azure it can take the VM.

whites11:
> Hmm... looks like the missing functionality was an oversight. However, I'm curious how this will act without a component running on the VMSS VM that will [respond to the termination notification via IMDS](https://docs.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-terminate-notification). I think it should block for the entire time, rather than doing some work, then notifying IMDS the VM can be deprovisioned when the work has been done.
> 
> There probably needs to be a component baked into the nodes that listens to these notifications, cordons and drains the node, then responds to the notification telling Azure it can take the VM.

Yes absolutely. For example in giant swarm we developed an app for that (not linking to avoid advertising).
What happens if the app is not present and the termination event is never ACKed is that Azure waits until the timeout expires before it deletes the instance.
Considering this is an opt-in feature disabled by default, I assume whoever enables it is aware of how termination events work and is able to install an app to deal with events. WDYT?

whites11:
> > I'd love to see termination notification ACK functionality land in image-builder for CAPZ images. Seems like a clear benefit to all users of CAPZ. Definitely would make spot instances much better.
> 
> Uhm, I'm not sure I agree. This feels too opinionated to me.
> What about discussing this during next office hours to gather more opinions?

I'll try to summarize what we discussed during office hours (see [recording](https://www.youtube.com/watch?v=GjRCsdvvr0E)):

- we aim at providing a community-managed reference application users can install in their clusters to react to azure VM events
- we don't want to bundle such application in the VM reference images, but rather follow the same distribution strategy used for tools such as `cluster-autoscaler`.
- we don't want to block this PR because it fixes a bug and the presence of the aforementioned reference application shouldn't be a blocker for merging this PR.

devigned (approved):
/lgtm

/assign @CecileRobertMichon

whites11:
> > > I'd love to see termination notification ACK functionality land in image-builder for CAPZ images. Seems like a clear benefit to all users of CAPZ. Definitely would make spot instances much better.
> > 
> > 
> > Uhm, I'm not sure I agree. This feels too opinionated to me.
> > What about discussing this during next office hours to gather more opinions?
> 
> I'll try to summarize what we discussed during office hours (see [recording](https://www.youtube.com/watch?v=GjRCsdvvr0E)):
> 
> * we aim at providing a community-managed reference application users can install in their clusters to react to azure VM events
> * we don't want to bundle such application in the VM reference images, but rather follow the same distribution strategy used for tools such as `cluster-autoscaler`.
> * we don't want to block this PR because it fixes a bug and the presence of the aforementioned reference application shouldn't be a blocker for merging this PR.

I am curious how the community works in this scenario.how is the new event management app meant to be created? Should it be a new GitHub project? Should it aim to be provider independent like cluster-auroscaler?

CecileRobertMichon (commented):
/lgtm
/approve

PR #1678, opened 2021-09-09
Explicitly set EnableRBAC to true in AzureManagedControlPlaneSpec

### LINKED ISSUE #1663: Explicitly set EnableRBAC to true in AzureManagedControlPlaneSpec
/kind feature

It's unclear from the [Azure API doc ](https://docs.microsoft.com/en-us/rest/api/aks/managed-clusters/create-or-update)what default value for `EnableRBAC` is. Explicitly set [EnableRBAC](https://github.com/Azure/azure-sdk-for-go/blob/v55.2.0/services/containerservice/mgmt/2021-05-01/containerservice/models.go#L2055) to true in [AzureManagedControlPlaneSpec](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/exp/api/v1alpha4/azuremanagedcontrolplane_types.go#L28)

richardchen331:
**What type of PR is this?**
/kind feature

Explicitly set `EnableRBAC` to true in `AzureManagedControlPlaneSpec`.

Fixes #1663

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
None
```

### DIFF
--- a/azure/services/managedclusters/managedclusters.go
+++ b/azure/services/managedclusters/managedclusters.go
@@ -104,6 +104,7 @@ func (s *Service) Reconcile(ctx context.Context) error {
 		Tags:     *to.StringMapPtr(managedClusterSpec.Tags),
 		ManagedClusterProperties: &containerservice.ManagedClusterProperties{
 			NodeResourceGroup: &managedClusterSpec.NodeResourceGroupName,
+			EnableRBAC:        to.BoolPtr(true),
 			DNSPrefix:         &managedClusterSpec.Name,
 			KubernetesVersion: &managedClusterSpec.Version,
 			LinuxProfile: &containerservice.LinuxProfile{


azure/services/managedclusters/managedclusters.go
```diff
@@ -104,6 +104,7 @@ func (s *Service) Reconcile(ctx context.Context) error {
 		Tags:     *to.StringMapPtr(managedClusterSpec.Tags),
 		ManagedClusterProperties: &containerservice.ManagedClusterProperties{
 			NodeResourceGroup: &managedClusterSpec.NodeResourceGroupName,
+			EnableRBAC:        to.BoolPtr(true),
```
CecileRobertMichon, line 107: have you verified that the default is true?

richardchen331: Yes, I created an AKS cluster without `EnableRBAC` set, and verified that the created cluster has kubernetes RBAC enabled.

However we cannot guarantee that the default behavior won't change, but I think it probably makes sense to always enable it explicitly? If you think we should allow user to choose and default to enable, I can make the change :)

CecileRobertMichon: I'm good with enabling explicitly, and I don't foresee the default changing anytime soon. If a use case comes up to disable it, we can make it user-configurable then. For now, let's keep the change as-is.

PR #1679, opened 2021-09-09
Allow user to specify SKU for AKS

### LINKED ISSUE #1664: Allow user to specify SKU for AKS
/kind feature

Allow user to specify [SKU](https://github.com/Azure/azure-sdk-for-go/blob/v55.2.0/services/containerservice/mgmt/2021-05-01/containerservice/models.go#L997) in [AzureManagedControlPlaneSpec](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/exp/api/v1alpha4/azuremanagedcontrolplane_types.go#L28)

richardchen331:
**What type of PR is this?**
/kind feature

Allow user to specify [SKU](https://github.com/Azure/azure-sdk-for-go/blob/v55.2.0/services/containerservice/mgmt/2021-05-01/containerservice/models.go#L997) in [AzureManagedControlPlaneSpec](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/exp/api/v1alpha4/azuremanagedcontrolplane_types.go#L28)

Fixes #1664

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Allow specifying SKU in AzureManagedControlPlane.
```


azure/services/managedclusters/managedclusters.go
```diff
@@ -215,7 +228,21 @@ func (s *Service) Reconcile(ctx context.Context) error {
 			}
 		}
 
-		diff := cmp.Diff(propertiesNormalized, existingMCPropertiesNormalized)
+		clusterNormalized := &containerservice.ManagedCluster{
```
alexeldeib, line 231: do we actually need this normalization dance? we're comparing one user-provided field and another field which always has the same value (or we can enforce that)

exp/api/v1alpha3/azuremanagedcontrolplane_types.go
```diff
@@ -81,6 +81,10 @@ type AzureManagedControlPlaneSpec struct {
 	// AadProfile is Azure Active Directory configuration to integrate with AKS for aad authentication.
 	// +optional
 	AADProfile *AADProfile `json:"aadProfile,omitempty"`
+
+	// Sku is the SKU of the AKS to be provisioned.
+	// +optional
+	Sku *SKU `json:"sku,omitempty"`
```
alexeldeib, line 87: minor nit: capitalization inconsistency

azure/services/managedclusters/managedclusters.go
```diff
@@ -173,6 +173,19 @@ func (s *Service) Reconcile(ctx context.Context) error {
...
+		var tier containerservice.ManagedClusterSKUTier
+		if managedClusterSpec.Sku.Tier == "Paid" {
+			tier = containerservice.ManagedClusterSKUTierPaid
+		} else {
+			tier = containerservice.ManagedClusterSKUTierFree
+		}
+		managedCluster.Sku = &containerservice.ManagedClusterSKU{
+			Name: containerservice.ManagedClusterSKUNameBasic,
+			Tier: tier,
+		}
```
alexeldeib, line 186:
```suggestion
                skuName := containerservice.ManagedClusterSKUName(managedClusterSpec.Sku.Tier)
		managedCluster.Sku = &containerservice.ManagedClusterSKU{
			Name: skuName,
			Tier: tier,
		}
```

exp/api/v1alpha3/azuremanagedcontrolplane_types.go
alexeldeib: should we be adding this to v1a3 at this point? or just doing webhook conversion?

alexeldeib: cc @CecileRobertMichon 

i'd assume the latter

alexeldeib:
minor comments, nice work so far!

exp/api/v1alpha3/azuremanagedcontrolplane_types.go
richardchen331: @alexeldeib thanks for reviewing my PR!

capitalization inconsistency: i'll address it :)

v1a3 vs. v1a4: I just tried reverting the change to v1a3, but the hub-spoke-hub test of the v1a3 [conversion test](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/exp/api/v1alpha3/conversion_test.go) would fail since it's lossy (we don't store sku in v1a3). So my understanding is that we still need to add it to v1a3. Is that correct?

azure/services/managedclusters/managedclusters.go
richardchen331: @alexeldeib I believe that the `Name` field in `ManagedClusterSKU` is always [Basic](https://github.com/Azure/azure-sdk-for-go/blob/v55.8.0/services/containerservice/mgmt/2021-05-01/containerservice/enums.go#L210-L216), and we should set `Tier` according to the tier in our Spec. That's why I added the code here. Please let me know if that makes sense :)

alexeldeib: This is not correct, roundtrip conversion should not be lossy. See how it’s handled for the identity ref which is not present in v1a3. 

Stored the higher version and unmarshal it inside the conversion webhook, then rehydrate the field from the unmarshaled struct.

alexeldeib: Whoops, replied to the wrong comment.

alexeldeib: You’re correct about the code, I’m just simplifying the existing logic a bit. I think it’s the same?

exp/api/v1alpha3/azuremanagedcontrolplane_types.go
alexeldeib: This is not correct, roundtrip conversion should not be lossy. See how it’s handled for the identity ref which is not present in v1a3.

Stored the higher version and unmarshal it inside the conversion webhook, then rehydrate the field from the unmarshaled struct.

richardchen331: ah interesting, I reverted change in v1a3 and added one line in `azuremanagedcontrolplane_conversion.go` to restore the field, now it works. Thanks @alexeldeib !

azure/services/managedclusters/managedclusters.go
richardchen331: sorry but I'm not sure if I understand your suggested change, it looks like now the `Name` field is `skuName`, which is `containerservice.ManagedClusterSKUName(managedClusterSpec.Sku.Tier)`, but that isn't `containerservice.ManagedClusterSKUNameBasic`. Did I miss something here?

alexeldeib: my mistake -- I flipped tier and name in the suggestion. 

Basically, you are comparing tierName against a non-const string, but we already enforce the values coming in are enums exactly matching one of the supplied values. so we can just type coerce whatever is in the spec, rather than do a conditional check.

alexeldeib: updated the code snippet

richardchen331: Got it, this make the code much simpler, thanks for the suggestion!

alexeldeib:
/approve

pending e2e

alexeldeib:
/assign @shysank @LochanRn 

for a lgtm to merge :)

azure/types.go
```diff
@@ -358,6 +358,9 @@ type ManagedClusterSpec struct {
 
 	// AADProfile is Azure Active Directory configuration to integrate with AKS, for aad authentication.
 	AADProfile *AADProfile
+
+	// SKU is the SKU of the AKS to be provisioned.
+	SKU *SKU
```
LochanRn, line 363:
```suggestion
// Tier - Tier of a managed cluster SKU.
Tier string
```

azure/types.go
```diff
@@ -372,6 +375,12 @@ type AADProfile struct {
 	AdminGroupObjectIDs []string
 }
 
+// SKU - AKS SKU.
+type SKU struct {
+	// Tier - Tier of a managed cluster SKU.
+	Tier string
+}
+
```
LochanRn, line 383:
```suggestion

LochanRn: because we need only Tier information from the SKU struct, can we just add it in the ManagedClusterSpec and avoid using a separate struct for this ?

alexeldeib: I’m pretty neutral on this. The current approach aligns to the AKS api, which was intended to be extensible, but we also don’t have any plans to extend it currently.

LochanRn: any possibility that it can be extended in the future ??

alexeldeib: It can change from AKS side yes. If AKS changes it in the future, we will have to make a breaking change to align it to the impl already in this PR

LochanRn: yeah so I think it is better to not change then 👍

LochanRn: @alexeldeib apart from the above changes this PR lgtm :) Great work @richardchen331

PR #1681, opened 2021-09-09
fix RANDOM var being omitted on Linux

devigned:
**What type of PR is this?**
/kind bug

On Linux when running Tilt and deploying a flavor, the CLUSTER_NAME variable will not be populated with $RANDOM. This change shells out to bash to populate RANDOM env var in sh to be used to generate the CLUSTER_NAME.

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```


CecileRobertMichon:
confirmed this works on macOS

/lgtm
/approve

PR #1683, opened 2021-09-13
Align kubectl install and CI scripts to use local bin folder

### LINKED ISSUE #1192: Align install and CI scripts to use local bin folder for all tools in the repository
/kind bug

There are two different ways the repository installs tools that was found in #1097.  After discussion in office hours would like to align to install all tools via `makefile` into a `hack/tools/bin` and have all scripts use those tools.  

`make` and CI scripts all use the same version of tooling from local path.

see https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1097 for discussion

**Environment:**

sayantani11:
**What type of PR is this?**
This PR makes the CI scripts use local bin folder for kubectl

Fixes #1192

WIP

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```


sayantani11:
@CecileRobertMichon Something like this?

CecileRobertMichon:
@sayantani11 I was thinking something like this: https://github.com/CecileRobertMichon/cluster-api-provider-azure/commit/10be812e0f77ff6b9a615eefcf62abfbb23ba237

And this would apply to every other place/script in the code using "kubectl"

sayantani11:
> @sayantani11 I was thinking something like this: [CecileRobertMichon@10be812](https://github.com/CecileRobertMichon/cluster-api-provider-azure/commit/10be812e0f77ff6b9a615eefcf62abfbb23ba237)
> 
> And this would apply to every other place/script in the code using "kubectl"

Now I get it completely, I understood when you explained yesterday, but was having problems on how to apply the changes.

sayantani11:
@CecileRobertMichon I guess I made the required changes. Please verify and let me know if changes are required.

CecileRobertMichon (commented):
/ok-to-test

Makefile
```diff
@@ -264,6 +264,9 @@ $(KUBECTL): ## Build kubectl
 	ln -sf "$(KUBECTL)" "$(TOOLS_BIN_DIR)/$(KUBECTL_BIN)"
 	chmod +x "$(TOOLS_BIN_DIR)/$(KUBECTL_BIN)" "$(KUBECTL)"
 
+.PHONY: $(KUBECTL_BIN)
+$(KUBECTL_BIN): $(KUBECTL) ## Building kubectl from the tools folder
```
CecileRobertMichon, line 268: for this to work as you're using it below, we need to change KUBECTL_BIN on line 97 to `$(TOOLS_BIN_DIR)/kubectl` and then every place in the makefile that uses `$(TOOLS_BIN_DIR)/$(KUBECTL_BIN)` to just `$(KUBECTL_BIN)`

so that $(KUBECTL_BIN) becomes `/hack/tools/bin/kubectl`

sayantani11:
@CecileRobertMichon Something else needs to be fixed.

sayantani11:
@CecileRobertMichon Can you please inform what else is failing? Or am I supposed to run `make generate` and then push the changes?

devigned:
@sayantani11 if you haven't, consider digging into the failed test jobs. The failure results will quickly provide guidance about what failed and usually why it failed.

For example, the [verify job failed](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/1683/pull-cluster-api-provider-azure-verify/1437894540970168320) with the following.
```
In ./scripts/kind-with-registry.sh line 21:
REPO_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
^-------^ SC2034: REPO_ROOT appears unused. Verify use (or export if used externally).


In ./scripts/ci-entrypoint.sh line 33:
# shellcheck kubectl
^-- SC1073: Couldn't parse this shellcheck directive. Fix to allow more checks.
                    ^-- SC1072: Expected '=' after directive key. Fix any mentioned problems and try again.

For more information:
  https://www.shellcheck.net/wiki/SC2034 -- REPO_ROOT appears unused. Verify ...
  https://www.shellcheck.net/wiki/SC1072 -- Expected '=' after directive key....
  https://www.shellcheck.net/wiki/SC1073 -- Couldn't parse this shellcheck di...

Please review the above warnings. You can test via "./hack/verify-shellcheck.sh"
If the above warnings do not make sense, you can exempt this warning with a comment
 (if your reviewer is okay with it).
In general please prefer to fix the error, we have already disabled specific lints
 that the project chooses to ignore.
See: https://github.com/koalaman/shellcheck/wiki/Ignore#ignoring-one-specific-instance-in-a-file

make: *** [Makefile:620: verify-shellcheck] Error 1
```

Try running `make test` to verify your changes.

sayantani11:
@devigned I actually do go through the errors but mostly find it difficult to understand 😅

hack/log/log-dump.sh
```diff
@@ -23,8 +23,10 @@ cd "${REPO_ROOT}" || exit 1
 
 # shellcheck source=hack/ensure-kind.sh
 source "${REPO_ROOT}/hack/ensure-kind.sh"
-# shellcheck source=hack/ensure-kubectl.sh
-source "${REPO_ROOT}/hack/ensure-kubectl.sh"
+
+# installation of kubectl
+KUBECTL="hack/tools/bin/kubectl"
```
CecileRobertMichon, line 28: @sayantani11 it's not finding `make ${KUBECTL}` below because the make target is the full absolute path (TOOLS_BIN uses abspath in the makefile). You can fix it by changing the line above to

```suggestion
KUBECTL=$(realpath hack/tools/bin/kubectl)
```

devigned:
> @devigned I actually do go through the errors but mostly find it difficult to understand sweat_smile

@sayantani11, thank you for digging into the log output. Is there something we can do make the output more actionable / more approachable?

sayantani11:
> > @devigned I actually do go through the errors but mostly find it difficult to understand sweat_smile
> 
> @sayantani11, thank you for digging into the log output. Is there something we can do make the output more actionable / more approachable?

Yeah, when I go through them, if there are are multiple errors in files, they appear in the same font style, as of which I find it difficult to understand. If the lines pointing the errors have a different color, guess that helps.

hack/log/log-dump.sh
sayantani11: Is it applying to every script using kubectl?

sayantani11: And will it be something like this `$(${REPO_ROOT}/hack/tools/bin/kubectl)` or we use real path?

sayantani11:
@CecileRobertMichon I went through the errors, it is not able to find the kubectl folder. And the verify failure, I guess is a consequence of that. I mean the directory path, how can I write that to avoid the error.

hack/log/log-dump.sh
CecileRobertMichon: @sayantani11 you just need to add $(realpath ...)` where `...` is the path you had before. So if the path was `../hack/tools/bin/kubectl`,  it would become `$(realpath ../hack/tools/bin/kubectl)`.

Same thing in every script.

CecileRobertMichon:
the verify failure is because you need to add double quotes when using bash variables, replace every occurence of `${KUBECTL}` with `"${KUBECTL}"`

This is what is suggested in the output of the verify test job:

```
In ./scripts/kind-with-registry.sh line 25:
make ${KUBECTL} &>/dev/null
     ^--------^ SC2086: Double quote to prevent globbing and word splitting.
Did you mean: 
make "${KUBECTL}" &>/dev/null
```

sayantani11:
```realpath: ../hack/tools/bin/kubectl: No such file or directory
+ EXIT_VALUE=1```
I guess this isn't working...
should I try ```$(../../hack/tools/bin/kubectl)```?

File: hack/create-identity-secret.sh [hunk 3]
```diff
@@ -21,10 +21,12 @@ set +o xtrace
 REPO_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
 cd "${REPO_ROOT}" || exit 1
 
-# shellcheck source=hack/ensure-kubectl.sh
-source "${REPO_ROOT}/hack/ensure-kubectl.sh"
+# Installation of kubectl
+KUBECTL=$(realpath ../hack/tools/bin/kubectl)
```
CecileRobertMichon, line 25:
```suggestion
KUBECTL=$(realpath hack/tools/bin/kubectl)
```

hack/install-cert-manager.sh
```diff
@@ -46,16 +46,20 @@ spec:
 END
 )
 
+# Installation of kubectl
+KUBECTL=$(realpath ../hack/tools/bin/kubectl)
+make "${KUBECTL}" &>/dev/nul
```
CecileRobertMichon, line 51:
```suggestion
make "${KUBECTL}" &>/dev/nul
```
```suggestion
make "${KUBECTL}" &>/dev/null
```

hack/install-cert-manager.sh
```diff
@@ -46,16 +46,20 @@ spec:
 END
 )
 
+# Installation of kubectl
+KUBECTL=$(realpath ../hack/tools/bin/kubectl)
```
CecileRobertMichon, line 50:
```suggestion
KUBECTL=$(realpath hack/tools/bin/kubectl)
```

hack/log/log-dump.sh
```diff
@@ -23,8 +23,10 @@ cd "${REPO_ROOT}" || exit 1
 
 # shellcheck source=hack/ensure-kind.sh
 source "${REPO_ROOT}/hack/ensure-kind.sh"
-# shellcheck source=hack/ensure-kubectl.sh
-source "${REPO_ROOT}/hack/ensure-kubectl.sh"
+
+# installation of kubectl
+KUBECTL=$(realpath ../hack/tools/bin/kubectl)
```
CecileRobertMichon, line 28:
```suggestion
KUBECTL=$(realpath hack/tools/bin/kubectl)
```

scripts/ci-conformance.sh
```diff
@@ -30,8 +30,11 @@ cd "${REPO_ROOT}" || exit 1
 source "${REPO_ROOT}/hack/ensure-go.sh"
 # shellcheck source=hack/ensure-kind.sh
 source "${REPO_ROOT}/hack/ensure-kind.sh"
-# shellcheck source=hack/ensure-kubectl.sh
-source "${REPO_ROOT}/hack/ensure-kubectl.sh"
+
+# check installation of kubectl
+KUBECTL=$(realpath ../hack/tools/bin/kubectl)
```
CecileRobertMichon, line 35:
```suggestion
KUBECTL=$(realpath hack/tools/bin/kubectl)
```

scripts/ci-e2e.sh
```diff
@@ -30,8 +30,11 @@ cd "${REPO_ROOT}" || exit 1
 source "${REPO_ROOT}/hack/ensure-go.sh"
 # shellcheck source=hack/ensure-kind.sh
 source "${REPO_ROOT}/hack/ensure-kind.sh"
-# shellcheck source=hack/ensure-kubectl.sh
-source "${REPO_ROOT}/hack/ensure-kubectl.sh"
+
+# check installation of kubectl
+KUBECTL=$(realpath ../hack/tools/bin/kubectl)
```
CecileRobertMichon, line 35:
```suggestion
KUBECTL=$(realpath hack/tools/bin/kubectl)
```

scripts/ci-entrypoint.sh
```diff
@@ -29,8 +29,11 @@ cd "${REPO_ROOT}" || exit 1
 source "${REPO_ROOT}/hack/ensure-go.sh"
 # shellcheck source=hack/ensure-kind.sh
 source "${REPO_ROOT}/hack/ensure-kind.sh"
-# shellcheck source=hack/ensure-kubectl.sh
-source "${REPO_ROOT}/hack/ensure-kubectl.sh"
+
+# building kubectl from tools folder
+KUBECTL=$(realpath ../hack/tools/bin/kubectl)
```
CecileRobertMichon, line 34:
```suggestion
KUBECTL=$(realpath hack/tools/bin/kubectl)
```

scripts/kind-with-registry.sh
```diff
@@ -17,9 +17,12 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-
+# shellcheck disable=SC2034  # Unused variables left for readability
```
CecileRobertMichon, line 20: let's remove this and the REPO_ROOT line below

scripts/kind-with-registry.sh
```diff
@@ -17,9 +17,12 @@ set -o errexit
...
 REPO_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
-KUBECTL=$REPO_ROOT/hack/tools/bin/kubectl
+
+# building kubectl from tools folder
+KUBECTL=$(realpath ../hack/tools/bin/kubectl)
```
CecileRobertMichon, line 24:
```suggestion
KUBECTL=$(realpath hack/tools/bin/kubectl)
```

sayantani11:
``` realpath: hack/tools/bin/kubectl: No such file or directory``` This error is still there.

**Comment by CecileRobertMichon** [comment]
@sayantani11 the issue is that hack/tools/bin doesn't exist yet in the CI env so it's failing to resolve the path.

To fix it, you need to add `mkdir -p "${REPO_ROOT}/hack/tools/bin"` before the line that calls realpath so that it creates the hack/tools/bin directory in case it doesn't already exist. Not that this requires REPO_ROOT to be defined so you'll need to add it in scripts that don't already have it.

**Comment by sayantani11** [comment]
@CecileRobertMichon Most of the errors are probably done, but I guess it's failing a few tests, went through the logs but was a bit confused.

**Comment by devigned** [comment]
/retest

The failure looks like it might be a flake.

**Comment by devigned** [comment]
/retest

One more try before digging deeper into the error cause.

**Comment by sayantani11** [comment]
> /retest
> 
> One more try before digging deeper into the error cause.

Guess it worked. Finally!

File: hack/create-identity-secret.sh [hunk 12]
```diff
@@ -21,10 +21,13 @@ set +o xtrace
 REPO_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
 cd "${REPO_ROOT}" || exit 1
 
-# shellcheck source=hack/ensure-kubectl.sh
-source "${REPO_ROOT}/hack/ensure-kubectl.sh"
+# Installation of kubectl
+mkdir -p "${REPO_ROOT}/hack/tools/bin/kubectl"
+KUBECTL=$(realpath hack/tools/bin/kubectl)
+make "${KUBECTL}" &>/dev/null
```
jsturtevant, line 27: I don't think this is doing what is intended?  If I run these commands locally then try to run kubectl via the env variable I get:

```
$KUBECTL 
bash: /workspaces/cluster-api-provider-azure/hack/tools/bin/kubectl: Is a directory
```
It looks like this is only during the make create-management-cluster command. Have you tried running that?

https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/e5327290e4fd5df3fae5779296b8e1d2c73c508e/Makefile#L483-L491

scripts/ci-entrypoint.sh
```diff
@@ -108,17 +112,17 @@ wait_for_nodes() {
 
     # Ensure that all nodes are registered with the API server before checking for readiness
     local total_nodes="$((CONTROL_PLANE_MACHINE_COUNT + WORKER_MACHINE_COUNT))"
-    while [[ $(kubectl get nodes -ojson | jq '.items | length') -ne "${total_nodes}" ]]; do
+    while [[ $(${KUBECTL} get nodes -ojson | jq '.items | length') -ne "${total_nodes}" ]]; do
```
jsturtevant, line 115: Have you tested with with `ci-entrypoint.sh`? I am not sure this is going to work becuase `${KUBECTL}` points to `hack/tools/bin/kubectl` which is a folder.

sayantani11: I ran `make verify`, it runs all the scripts as well right? or am I missing something😶?

jsturtevant: `make verify` runs some checks to ensure files are upto date (has correct headers, go mod is upto date, go generate was run, linting of shell scripts) but doesn't run e2e scripts or other processes:

https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/4fe1b243d2c694a70b8a154ca9ae13f7d68a5149/Makefile#L603-L624

hack/create-identity-secret.sh
CecileRobertMichon: Line 25 should be `mkdir -p "${REPO_ROOT}/hack/tools/bin"` instead of `mkdir -p "${REPO_ROOT}/hack/tools/bin/kubectl"`

same everywhere else

sayantani11: @CecileRobertMichon removing kubectl from the path gives `realpath: hack/tools/bin/kubectl: No such file or directory`

scripts/ci-entrypoint.sh
sayantani11: I will test `ci-entrypoint.sh` once I understand is it supposed to be `mkdir -p hack/tools/bin/kubectl` or `mkdir -p hack/tools/bin`. The first one gives `folder already exists` and the latter `realpath:......: No such file or directory`.

CecileRobertMichon:
/lgtm
/approve

thanks for all the hard work @sayantani11

PR #1684, opened 2021-09-14
Make VNet and NSGs reconcile/delete async

CecileRobertMichon:
**What type of PR is this?**
/kind cleanup
/kind feature

**What this PR does / why we need it**: Follow up on #1610 and implementation of #1541 for virtual networks and security groups (bundled together because they were both part of the original #1610 PR and share the managed vnet logic).

Fixes # 

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Make VNet and NSGs reconcile/delete async
```


azure/services/securitygroups/spec.go
```diff
@@ -0,0 +1,117 @@
...
+	return network.SecurityGroup{
+		Location: to.StringPtr(s.Location),
+		SecurityGroupPropertiesFormat: &network.SecurityGroupPropertiesFormat{
+			SecurityRules: &securityRules,
+		},
+		Etag: etag,
+	}, nil
+}
+
+// TODO: review this logic and make sure it is what we want. It seems incorrect to skip rules that don't have a certain protocol, etc.
```
CecileRobertMichon, line 95: this logic smells fishy to me, not sure why we're doing a continue when the rule has certain properties. That being said I didn't want to introduce bug fixes/behavior changes as part of this already large refactor so I kept the logic as is from https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1684/files#diff-5bb14c98c52ecfa3dad6cf4745c9378b7624fdc60f0910343534ddac18408fb1L115

azure/services/securitygroups/securitygroups_test.go
```diff
@@ -22,213 +22,131 @@ import (
...
+		},
+		{
+			name:          "second security groups create not done",
+			expectedError: "failed to create resource test-group/test-nsg (service: securitygroups): this is an error",
+			expect: func(s *mock_securitygroups.MockNSGScopeMockRecorder, m *mock_securitygroups.MockclientMockRecorder) {
+				s.V(gomock.AssignableToTypeOf(2)).AnyTimes().Return(klogr.New())
+				s.IsVnetManaged().Return(true, nil)
+				s.NSGSpecs().Return(fakeNSGSpecs)
+				s.GetLongRunningOperationState("test-nsg", serviceName).Return(nil)
+				m.CreateOrUpdateAsync(gomockinternal.AContext(), &fakeNSG).Return(nil, errFake)
```
karuppiah7890, line 129: @CecileRobertMichon Looking at the name of the test `second security groups create fails`, I'm wondering if it matches the implementation. The implementation has an error for the creation of both the security groups and the result status is from the first failure. From the test name I was thinking the implementation might look like

first security group `CreateOrUpdateAsync` - passes, no errors, no futures as it's immediately created. But currently it's different

second security `CreateOrUpdateAsync` - has errors but it has future as it's create is not yet done. This is how it's currently done, except that `fakeFuture` has reference to `DELETE` HTTP Method - https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1684/files#diff-5bfc905db237a71559b05fbd32c6ec3f216c5b5060b8e1824abc27b97fddaab2R76

Also, generic question, looks like `CreateOrUpdateAsync` is [expected to return a non-nil error even when there's a future](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/services/async/async.go#L77-L79). Why is that the case? I see that `WaitForCompletionRef` used in the async implementations can only fail due to context cancellation (in our case due to timeout) or other timing reasons (exceeds polling duration etc). If that's the case, why not return nil error along with future? I'm asking as it's weird to see an error being returned in the test called `errFoo` and the only reason is because `async.CreateResource` expects `CreateOrUpdateAsync` to return error along with future which is what we do here https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1684/files#diff-53d4a612701b97cfb63f5e584b0e1392ee162a7bee333063721cccb1604b4c8dR109 . The first thought the came to mind was - why not return future with nil error when it's sure that there's no error other than "operation not done" or "operation pending" error

azure/services/securitygroups/securitygroups.go
```diff
@@ -52,111 +54,72 @@ func New(scope NSGScope) *Service {
...
-				s.Scope.V(2).Info("security group exists and no default rules are missing, skipping update", "security group", nsgSpec.Name)
-				continue
-			}
-		default:
-			s.Scope.V(2).Info("creating security group", "security group", nsgSpec.Name)
-			for _, rule := range nsgSpec.SecurityRules {
-				securityRules = append(securityRules, converters.SecurityRuleToSDK(rule))
+		if err := async.CreateResource(ctx, s.Scope, s.client, nsgSpec, serviceName); err != nil {
+			if !azure.IsOperationNotDoneError(err) || result == nil {
+				result = err
```
karuppiah7890, line 83: Maybe we could add some tests to capture the multiple errors scenario? As it's a bit of a tricky logic in here. Multiple errors can be a mix of operation-not-done errors and non operation-not-done errors and in any order. Depending on the order of errors and the kind of error, the result gets overwritten multiple times or it doesn't get overwritten

karuppiah7890:
@CecileRobertMichon I was looking at the implementation of async reconcile of NSGs in this PR to replicate it for public ips to fix #1716 . I have a question regarding the status updates of NSGs. I see that we are storing only [one error as a result](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1684/files#diff-5bb14c98c52ecfa3dad6cf4745c9378b7624fdc60f0910343534ddac18408fb1R78) for the [`UpdatePutStatus` call](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1684/files#diff-5bb14c98c52ecfa3dad6cf4745c9378b7624fdc60f0910343534ddac18408fb1R88). I was wondering about the case where we have multiple errors - errors which are non operation-not-done errors . I see that we would just store the last such non operation-not-done errors and then never overwrite it with any possible operation-not-done-errors. Why not capture all the errors which are non operation-not-done errors? So that we can show them all here - https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/4fe1b243d2c694a70b8a154ca9ae13f7d68a5149/azure/scope/cluster.go#L733

azure/services/securitygroups/securitygroups.go
```diff
@@ -52,111 +54,72 @@ func New(scope NSGScope) *Service {
...
+		return errors.Wrap(err, "failed to determine if network security groups are managed")
+	} else if !managed {
+		s.Scope.V(4).Info("Skipping network security groups reconcile in custom VNet mode")
 		return nil
 	}
 
+	// We go through the list of NSGSpecs to reconcile each one, independently of the result of the previous one.
+	// If multiple erros occur, we return the most pressing one
+	// order of precedence is: error creating -> creating in progress -> created (no error)
+	var result error
```
karuppiah7890, line 78: I just noticed this comment actually 🙈 , with respect to https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1684#issuecomment-931091181 . I guess the question still remains as to - When we have multiple errors in the category of `error creating`, should we think about showing them all? I understand it can be too much errors, but I don't think we can find the most pressing error. In the current case we just return the last error that occurs, which may or not be the most pressing errors. And operation-not-done is an error in code, but in terms of the whole operation, it's only a stage of creation, and same for created (no error). So I think there's just mainly one category of errors (excluding operation-not-done errors) which is error in creating the resource, in this case the NSG

azure/services/securitygroups/securitygroups_test.go
CecileRobertMichon: updated the test names to make what they're testing more clear. Also added a fake PUT future to use for reconcile tests. 

> why not return future with nil error when it's sure that there's no error other than "operation not done" or "operation pending" error
addressed in #1837

CecileRobertMichon:
Rebased/tests updated/fixed issues, this should be ready to go!

CecileRobertMichon:
Can't repro the private cluster deletion issue locally. Trying one more time...

/retest

azure/scope/cluster.go
```diff
@@ -392,8 +406,15 @@ func (s *ClusterScope) Vnet() *infrav1.VnetSpec {
 }
 
 // IsVnetManaged returns true if the vnet is managed.
-func (s *ClusterScope) IsVnetManaged() bool {
-	return s.Vnet().ID == "" || s.Vnet().Tags.HasOwned(s.ClusterName())
+func (s *ClusterScope) IsVnetManaged(ctx context.Context) (bool, error) {
+	var err error
+	if s.cache.IsVnetManaged == nil {
+		vnetSvc := virtualnetworks.New(s)
```
CecileRobertMichon, line 412: @shysank PTAL. I had to go back on our conversation in https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1610#discussion_r698808313 because it doesn't work for Delete() as we call services in the opposite order as Reconcile() (eg. the subnets need to be deleted before the vnet is deleted) so this means the vnet service isn't the first one to be reconciled which means we have to do this before we get to the vnet service. I'm not sure how to work around that constraint without calling the virtual network from either the scope or other services (eg. from the subnet service). Thoughts?

CecileRobertMichon: cc @devigned as well

shysank: can we think of `IsManaged` as something that identifies whether that resource is managed by capz instead of checking `IsVnetManaged`? Every Service can implement `IsManaged` behaviour, and internally a service can depend on another service or another service's client to determine whether it is managed or not. The cache will be shared among all services. In this way, we can make clear abstractions. Concretely: `SubnetSvc.IsManaged()` will check if vnet is managed either through `virtualNetworksClient` (we can write a helper if it is used in more than one place) or by exposing `virtualNetworksSvc.IsManaged`, and updates the cache through scope which can be used by subsequent services. wdyt?

CecileRobertMichon: let me try some things once #1874 is merged and I can rebase on top of it

CecileRobertMichon:
/close

replaced by https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1921 and #1918 

will tackle refactoring the IsManaged logic separately once all services are async.

PR #1686, opened 2021-09-14
Make route tables reconcile/delete async

CecileRobertMichon:
**What type of PR is this?**
/kind cleanup
/kind feature

**What this PR does / why we need it**: Implements async reconciliation and delete for the route tables service. Follow up on #1610 and #1541.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Make route tables reconcile/delete async
```


CecileRobertMichon:
/hold 

to rebase on top of #1684

azure/services/routetables/client.go
```diff
@@ -61,36 +65,104 @@ func (ac *azureClient) Get(ctx context.Context, resourceGroupName, rtName string
...
+	// if the operation completed, return a nil future
+	return nil, err
 }
 
-// Delete deletes the specified route table.
-func (ac *azureClient) Delete(ctx context.Context, resourceGroupName, rtName string) error {
+// DeleteAsync deletes a route table asynchronously. DeleteAsync sends a DELETE
+// request to Azure and if accepted without error, the func will return a Future which can be used to track the ongoing
+// progress of the operation.
+func (ac *azureClient) DeleteAsync(ctx context.Context, spec azure.ResourceSpecGetter) (azureautorest.FutureAPI, error) {
```
CecileRobertMichon, line 105: TODO (cecile): add timeout

azure/services/routetables/client.go
```diff
@@ -61,36 +65,104 @@ func (ac *azureClient) Get(ctx context.Context, resourceGroupName, rtName string
...
 }
 
-// CreateOrUpdate create or updates a route table in a specified resource group.
-func (ac *azureClient) CreateOrUpdate(ctx context.Context, resourceGroupName string, rtName string, rt network.RouteTable) error {
-	ctx, span := tele.Tracer().Start(ctx, "routetables.AzureClient.CreateOrUpdate")
+// CreateOrUpdateAsync creates or updates a route table asynchronously.
+// It sends a PUT request to Azure and if accepted without error, the func will return a Future which can be used to track the ongoing
+// progress of the operation.
+func (ac *azureClient) CreateOrUpdateAsync(ctx context.Context, spec azure.ResourceSpecGetter) (azureautorest.FutureAPI, error) {
+	ctx, span := tele.Tracer().Start(ctx, "routetables.AzureClient.CreateOrUpdateAsync")
```
CecileRobertMichon, line 72: TODO (cecile): add timeout

azure/services/routetables/routetables.go
```diff
@@ -44,78 +47,75 @@ type Service struct {
...
-		if err != nil {
-			return errors.Wrapf(err, "failed to create route table %s in resource group %s", routeTableSpec.Name, s.Scope.ResourceGroup())
+	for _, rtSpec := range s.Scope.RouteTableSpecs() {
+		if err := async.CreateResource(ctx, s.Scope, s.client, rtSpec, serviceName); err != nil {
+			if !azure.IsOperationNotDoneError(err) || result == nil {
+				result = err
+			}
 		}
-		s.Scope.V(2).Info("successfully created route table", "route table", routeTableSpec.Name)
 	}
```
karuppiah7890, line 84: I have the same thought as https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1684#issuecomment-931091181 , https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1684#discussion_r719205970 about multiple errors case especially multiple non operation-not-done errors. And also about tests for the current implementation

azure/services/routetables/client.go
```diff
@@ -54,43 +50,112 @@ func newRouteTablesClient(subscriptionID string, baseURI string, authorizer auto
...
+
+	if futureData == nil {
+		return nil, errors.Errorf("cannot get result from nil future")
+	}
+	var result func(client network.RouteTablesClient) (RT network.RouteTable, err error)
+
+	switch futureType {
+	case infrav1.PutFuture:
+		var future *network.RouteTablesCreateOrUpdateFuture
+		jsonData, err := futureData.MarshalJSON()
```
shysank, line 143: Is there a reason for not directly casting `futureData` to `RouteTablesCreateOrUpdateFuture`? I presume `RouteTablesCreateOrUpdateFuture` implements `FutureApi`.

azure/services/routetables/routetables.go
```diff
@@ -19,101 +19,93 @@ package routetables
...
+	if !s.Scope.IsVnetManaged() {
+		log.V(4).Info("Skipping route tables reconcile in custom vnet mode")
+	} else {
+		// We go through the list of route tables to reconcile each one, independently of the result of the previous one.
+		// If multiple erros occur, we return the most pressing one
+		// order of precedence is: error creating -> creating in progress -> created (no error)
+		for _, rtSpec := range s.Scope.RouteTableSpecs() {
+			if _, err := s.CreateResource(ctx, rtSpec, serviceName); err != nil {
+				if !azure.IsOperationNotDoneError(err) || resErr == nil {
+					resErr = err
```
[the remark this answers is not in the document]
CecileRobertMichon: yes, previous discussion on this in https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1900 and https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1745#discussion_r737561421

for now, this follows the pattern already established in other services

azure/services/routetables/client.go
CecileRobertMichon: Unfortunately this doesn't work right now because the future type is actually azureautorest.Future, not `RouteTablesCreateOrUpdateFuture`, because it was converted back to an Azure Future from the CAPZ infrav1.Future type stored in Status: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/converters/futures.go#L49

Now, we could change the above conversion by either a) adding a separate conversion function for each service, or b) have a switch statement that maps different Azure SDK future type based on the infrav1.Future Type and Name, and this would allow us to do the casting directly. However, that's arguably worse than the few extra lines of code we have to marshal and unmarshal the future above, since the matrix of Future types is so big (one for each type of operation PUT, PATCH, DELETE for each service), we'd have about ~ 25 services * 3 = 75 future types to handle (probably a bit less since not every service implements each future type), and it would be very error-prone.

It's unfortunate that Result() isn't part of the `FutureApi` interface, that would solve the problem, but it's not possible because the Result return is type-specific: https://github.com/Azure/azure-sdk-for-go/issues/14433#issuecomment-802539626

CecileRobertMichon: this is one of the alternatives: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/services/scalesets/client.go#L209

Instead of passing in a FutureAPI interface, we directly pass in the infrav1.Future. IMO the above implementation is a bit cleaner (and that's the approach I've adopted in refactors of other services so far) because it doesn't require as tight of a coupling between the client and the API types. Also, we had to jump through some hoops at the time and define our own adapter interfaces to be able to get the Result func from the future: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/47ead2a504fc0165035e9e8c020b54d7422d9b87/azure/services/scalesets/client.go#L59-L71

CecileRobertMichon: cc @devigned

azure/services/routetables/client.go
```diff
@@ -54,43 +50,112 @@ func newRouteTablesClient(subscriptionID string, baseURI string, authorizer auto
...
+	return ac.routetables.Get(ctx, spec.ResourceGroupName(), spec.ResourceName(), "")
 }
 
-// CreateOrUpdate create or updates a route table in a specified resource group.
-func (ac *azureClient) CreateOrUpdate(ctx context.Context, resourceGroupName string, rtName string, rt network.RouteTable) error {
-	ctx, _, done := tele.StartSpanWithLogger(ctx, "routetables.AzureClient.CreateOrUpdate")
+// CreateOrUpdateAsync creates or updates a route table asynchronously.
+// It sends a PUT request to Azure and if accepted without error, the func will return a Future which can be used to track the ongoing
+// progress of the operation.
+func (ac *azureClient) CreateOrUpdateAsync(ctx context.Context, spec azure.ResourceSpecGetter, parameters interface{}) (interface{}, azureautorest.FutureAPI, error) {
```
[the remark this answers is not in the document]
CecileRobertMichon: Great suggestion. I just pushed an update for this & applied the change to other services that are already refactored as well. I kept the commit separate since it's a bit of a scope creep and required renaming some variables here and there.

azure/services/routetables/routetables.go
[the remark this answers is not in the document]
shysank: Ah, ok thanks for the context. I'm just a bit confused by the comment above ` If multiple errors occur, we return the most pressing one. order of precedence is: error creating -> creating in progress -> created (no error)`. For eg. if it fails in `spec.Parameters` for the first spec and the second spec completes with no error, the comment suggests it should return no error, but we'd be returning the error for `spec.Parameters`. Is my understanding correct here?

azure/services/routetables/client.go
shysank: Thanks for the explanation, would it make sense to add a comment that explains this or atleast a link to your comment above?

CecileRobertMichon: Added a comment there and in the other places that do this, let me know what you think

shysank: Looks good, Thanks! Maybe in future, we can try to move these pieces of code to some util file, so that they are all in one place.

azure/services/routetables/routetables.go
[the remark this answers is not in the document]
CecileRobertMichon: >  if it fails in spec.Parameters for the first spec and the second spec completes with no error

in this case we should be returning the error from the first spec, which is what the code is doing

That's what I was trying to express in the comments but maybe the arrows are making it unclear. Any error always takes precedence over an operation not done, which itself takes precedence over nil error. 

How could I rephrase the comment to be more clear about what's actually going on?

azure/services/routetables/client.go
CecileRobertMichon: that's what I was thinking too but unfortunately we can't share code because the `RouteTablesCreateOrUpdateFuture` type is unique to the service and to the operation so we'd need a different helper func for each one... At least not until generics come along :)

CecileRobertMichon:
> failed to solve with frontend dockerfile.v0: failed to solve with frontend gateway.v0: frontend grpc server closed unexpectedly

that's a new one...

/retest

azure/services/routetables/routetables.go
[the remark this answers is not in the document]
shysank: Maybe just mention it's in the descending order of precedence (something like highest -> lowest priority). And also add the errors that are not of type `IsOperationNotDoneError `. So, it'll look something like 
`If multiple errors occur, we return the most pressing one. order of precedence (highest -> lowest) is: error not related to operation status -> error creating -> creating in progress -> created (no error)`. wdyt?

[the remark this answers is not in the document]
CecileRobertMichon: > Maybe just mention it's in the descending order of precedence (something like highest -> lowest priority)

👍 

what is the different between "error not related to operation status" and "error creating" for you? 

For me creating in progress <==> operationNotDoneError, while error creating <==> any other error.

How about

```
If multiple errors occur, we return the most pressing one. order of precedence (highest -> lowest) is: error that is not an operationNotDoneError (ie. error creating) -> operationNotDoneError (ie. creating in progress) -> no error (ie. created)
```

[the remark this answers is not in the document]
shysank: looks good 👍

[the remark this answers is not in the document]
CecileRobertMichon: updated here and other locations

azure/services/availabilitysets/client.go
```diff
@@ -30,7 +30,7 @@ import (
 
 // Client wraps go-sdk.
 type Client interface {
-	Get(ctx context.Context, spec azure.ResourceSpecGetter) (interface{}, error)
+	Get(ctx context.Context, spec azure.ResourceSpecGetter) (result interface{}, err error)
 	CreateOrUpdateAsync(context.Context, azure.ResourceSpecGetter, interface{}) (interface{}, azureautorest.FutureAPI, error)
```
Jont828, line 34: @CecileRobertMichon Should the named return be added in the declarations for `CreateOrUpdateAsync`, `DeleteAsync`, and `IsDone` as well?

CecileRobertMichon: They should, good catch thanks!

azure/services/vnetpeerings/vnetpeerings.go
```diff
@@ -59,8 +59,8 @@ func (s *Service) Reconcile(ctx context.Context) error {
 	defer cancel()
 
 	// We go through the list of VnetPeeringSpecs to reconcile each one, independently of the result of the previous one.
-	// If multiple errors occur, we return the most pressing one
-	// order of precedence is: error creating -> creating in progress -> created (no error)
+	// If multiple errors occur, we return the most pressing one.
+	//  Order of precedence (highest -> lowest) is: error that is not an operationNotDoneError (ie. error creating) -> operationNotDoneError (ie. creating in progress) -> no error (ie. created)
```
Jont828, line 63:
```suggestion
	//  Order of precedence (highest -> lowest) is: error that is not an operationNotDoneError (ie. error creating) -> operationNotDoneError (ie. creating in progress) -> no error (i.e. created)
```
(add dot)

azure/services/vnetpeerings/vnetpeerings.go
```diff
@@ -85,8 +85,8 @@ func (s *Service) Delete(ctx context.Context) error {
 	var result error
 
 	// We go through the list of VnetPeeringSpecs to delete each one, independently of the result of the previous one.
-	// If multiple errors occur, we return the most pressing one
-	// order of precedence is: error deleting -> deleting in progress -> deleted (no error)
+	// If multiple errors occur, we return the most pressing one.
+	//  Order of precedence (highest -> lowest) is: error that is not an operationNotDoneError (ie. error creating) -> operationNotDoneError (ie. creating in progress) -> no error (ie. created)
```
Jont828, line 89:
```suggestion
	//  Order of precedence (highest -> lowest) is: error that is not an operationNotDoneError (i.e. error deleting) -> operationNotDoneError (i.e. deleting in progress) -> no error (i.e. deleted)
```

devigned (approved):
Beyond the comments from @mboersma, I have nothing more to add. lgtm!

shysank:
> > failed to solve with frontend dockerfile.v0: failed to solve with frontend gateway.v0: frontend grpc server closed unexpectedly
> 
> that's a new one...
> 
just curious, was this a flake?

CecileRobertMichon:
yes it seemed like a prow infra flake, the test didn't even start running

CecileRobertMichon:
@shysank @devigned could one of you give you an approve as well please?

PR #1691, opened 2021-09-16
Enable coalescing reconciler for more controllers

### LINKED ISSUE #1688: Enable coalescing reconciler for all controllers
/kind feature

https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1332/commits/b6b38b0e6f81b57e98f9eb933e3e6a9cbf2f5936 added a coalescing reconciler to enable debouncing of reconciles of AzureMachinePools and AzureMachinePoolMachines. We should enable this for all controllers to make sure that we aren't reconciling too often.

**Environment:**

CecileRobertMichon:
Enable coalescing reconciler for AzureCluster, AzureMachine, AzureManagedControlPlane, AzureManagedCluster, and AzureManagedMachinePool

 

 
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**: #1332 (https://github.com/devigned/cluster-api-provider-azure/commit/b6b38b0e6f81b57e98f9eb933e3e6a9cbf2f5936) added a coalescing reconciler to debounce reconciles (in other words, make sure we don't run too many successful reconcile loops in short amounts of time). At the time, it was only enabled for AzureMachinePool and AzureMachinePoolMachine controllers. This PR enables it for more controllers, specifically all the ones that reconcile Azure resources (AzureCluster, AzureMachine, AzureManagedControlPlane, AzureManagedCluster, and AzureManagedMachinePool), in preparation for #1541.

Also fixes some duplicate code in main.go.

Fixes #1688 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Enable coalescing reconciler for more controllers
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -296,18 +296,10 @@ func main() {
 	}
 
 	registerControllers(ctx, mgr)
+
+	registerWebhooks(ctx, mgr)
+
 	// +kubebuilder:scaffold:builder
-
-	if err := mgr.AddReadyzCheck("webhook", mgr.GetWebhookServer().StartedChecker()); err != nil {
-		setupLog.Error(err, "unable to create ready check")
-		os.Exit(1)
-	}
-
-	if err := mgr.AddHealthzCheck("webhook", mgr.GetWebhookServer().StartedChecker()); err != nil {
-		setupLog.Error(err, "unable to create health check")
-		os.Exit(1)
-	}
-
 	setupLog.Info("starting manager", "version", version.Get().String())
 	if err := mgr.Start(ctx); err != nil {
 		setupLog.Error(err, "problem running manager")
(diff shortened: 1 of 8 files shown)


main.go
```diff
@@ -296,18 +296,10 @@ func main() {
 	}
 
 	registerControllers(ctx, mgr)
-	// +kubebuilder:scaffold:builder
 
-	if err := mgr.AddReadyzCheck("webhook", mgr.GetWebhookServer().StartedChecker()); err != nil {
```
CecileRobertMichon, line 301: @reviewers, please pay close attention here. I believe we were duplicating some code so I removed it, but let me know if there is a good reason for this being called twice (here and line 520)

same thing with mgr.Start (line 304 and line 514)

main.go
```diff
@@ -316,22 +308,30 @@ func main() {
...
 		mgr.GetEventRecorderFor("azuremachine-reconciler"),
 		reconcileTimeout,
 		watchFilterValue,
-	).SetupWithManager(ctx, mgr, controller.Options{MaxConcurrentReconciles: azureMachineConcurrency}); err != nil {
+	).SetupWithManager(ctx, mgr, controllers.Options{Options: controller.Options{MaxConcurrentReconciles: azureMachineConcurrency}, Cache: machineCache}); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", "AzureMachine")
 		os.Exit(1)
 	}
 
+	clusterCache, err := coalescing.NewRequestCache(5 * time.Second)
```
CecileRobertMichon, line 324: went with 5 seconds since I don't want the reconciles to be too slow but let me know if you think that's too aggressive

main.go
```diff
@@ -316,22 +308,30 @@ func main() {
 }
 
 func registerControllers(ctx context.Context, mgr manager.Manager) {
+	machineCache, err := coalescing.NewRequestCache(5 * time.Second)
```
devigned, line 311: Did you consider making this a configurable as a cmdline arg?

devigned (commented):
Looks good. Just a couple comments.

main.go
CecileRobertMichon: I didn't, that would be a good one to be able to make configurable. Do you have any thoughts on whether it should be configurable per controller or just a single value?

devigned: I had made it configurable for both AMP and AMPM. If it's not overkill for cmdline args, I'd vote for per controller.

CecileRobertMichon: AMP and AMPM are both hardcoded right now, I don't think they're configurable.

Per controller makes sense, but it might be overwhelming to the user to be able to configure all of them (or to have to configure each one separately to change all the values). What do you think about one common flag for now, and potentially make it more granular later if the use case arises?

devigned: You are right. I lied. I think I was thinking about doing that, but must have forgot or thought that I already had.

CecileRobertMichon: wishful thinking :) 

okay let me know what you think of this:

1) using a common var across all the controllers: this is slightly less granular but I think it's a better, easier to understand configuration from a user's perspective (since they're not supposed to know how the code of each controller works), 10 seconds being a middle ground default value. 
2) the flag name "debouncing-timer" and description, tried to make that as human developer-friendly as possible and tried to describe how it's actually useful (ie. what it does) and not _how_ it does it (ie. "cache").

CecileRobertMichon:
private cluster test had remaining resources after delete

/retest

devigned (approved):
/lgtm

devigned:
> How will this work when there are more than one controller instances running? AFAICT, the cache appears to be local to an instance. Does controller manager guarantee that an object's reconciliation request is always sent to the same instance?

Are you implying a scenario where more than one controller instance is watching and reconciling the same resources?

devigned:
> Yeah, as in, just scale my capz controller deployment to 2 (or more).

The controller should have only 1 leader elected based on our [manager configuration](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/de7cb8b67bee4219751baf7926dd836eee833325/config/manager/manager.yaml#L24). I don't know that we should support 2 controllers reconciling the same resources. If 2 controllers are run side by side, I would imagine that each would be responsible for reconciling their own exclusive set of resources.

^ is that assumption incorrect?

devigned:
Architecture docs to the rescue.

https://cluster-api.sigs.k8s.io/developer/architecture/controllers/support-multiple-instances.html#contract

shysank:
@devigned I think it's a fair assumption. Thanks for the explanation! The only edge case I can think of is if a new leader was elected, but that's going be rare, and even if it happens, the worst thing that could happen is that the cache will get invalidated, which is fine.
/lgtm

CecileRobertMichon:
Thanks for bringing this up @shysank, definitely a good scenario to think about. 

/approve

Issue #1648, opened 2021-08-23
GPU test is flaky


CecileRobertMichon:
/kind bug

[Before submitting an issue, have you checked the [Troubleshooting Guide](https://capz.sigs.k8s.io/topics/troubleshooting.html)?]

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]

https://testgrid.k8s.io/sig-cluster-lifecycle-cluster-api-provider-azure#capz-postsubmit-capi-e2e-full-main is failing at around 50% rate. The first step in investigating this would be to look at failures and add more details to this issue about what error(s) is(are) happening.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- cluster-api-provider-azure version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

mboersma:
/assign

I looked at the ten most recent GPU cluster failures.

<details>
  <summary>7 were Deployment failures:</summary>

```go
Workload cluster creation
/home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/test/e2e/azure_test.go:41
  Creating a GPU-enabled cluster
  /home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/test/e2e/azure_test.go:357
    with a single control plane node and 1 node [It]
    /home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/test/e2e/azure_test.go:358
    Timed out after 900.004s.
      
    Expected
        <int>: 0
    to equal
        <int>: 1
    /home/prow/go/pkg/mod/sigs.k8s.io/cluster-api/test@v0.4.1/framework/machinedeployment_helpers.go:121
    Full Stack Trace
    sigs.k8s.io/cluster-api/test/framework.WaitForMachineDeploymentNodesToExist(0x251d3a0, 0xc0000640d0, 0x7f3ca0342cc0, 0xc0003cebd0, 0xc0001ec000, 0xc000b58b00, 0x0, 0x0, 0x0)
    	/home/prow/go/pkg/mod/sigs.k8s.io/cluster-api/test@v0.4.1/framework/machinedeployment_helpers.go:121 +0x41d
[...]

PR #1693, opened 2021-09-16
Test if GPU resources are available before scheduling e2e job

mboersma:
**What type of PR is this?**

/kind failing-test

Fixes (one of) the problems causing the GPU-enabled cluster tests to fail in [testgrid e2e](https://testgrid.k8s.io/sig-cluster-lifecycle-cluster-api-provider-azure#capz-postsubmit-capi-e2e-full-main) by waiting for the GPU resources to be exposed before running the cuda-vector Job.

Refs #1648

The behavior I've observed with the nvidia-operator machinery is that after the node is available, it still takes several minutes for the "nvidia.com/gpu" resource to be available. This has been enough to make the GPU-requiring Job fail. This essentially is just extending the timeout, but will give more information if the failure relates to GPU resources.

I assume this is considered acceptable--I know the previous non-operator implementation also took a little bit to finish `PostKubeAdmCommands`. But if we would rather address it by trying to ensure the GPU resource is available earlier during provisioning, I could look at that instead.

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/azure_gpu.go
+++ b/test/e2e/azure_gpu.go
@@ -61,6 +61,20 @@ func AzureGPUSpec(ctx context.Context, inputGetter func() AzureGPUSpecInput) {
 	clientset := clusterProxy.GetClientSet()
 	Expect(clientset).NotTo(BeNil())
 
+	By("Waiting for a node to have an \"nvidia.com/gpu\" allocatable resource")
+	Eventually(func() bool {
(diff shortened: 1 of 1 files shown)


mboersma:
/test pull-cluster-api-provider-azure-e2e-full

Previous run failed on the AKS test. :-|

PR #1721, opened 2021-09-22
Fix resource group not getting updated if tags are added

### LINKED ISSUE #1696: Resource group should get updated if tags are added
/kind bug

Currently, we only create the resource group if it doesn't already exist: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/47caf96183f7d3bd1e4dda5944a06dae10318bc7/azure/services/groups/spec.go#L53. However, we should consider adding any additional tags that were mutated and updating it, **if and only if the group is managed by capz.**

- if the group already exists and is BYO, we should NOT add the owner tag to it (or any other tags)
- if the group already exists and is managed, we should update tags if and only if some are missing
- if the group already exists and no tags are missing (desired tags exist) we should do nothing
- if the group doesn't exist, we should create it

Someone working on this issue might want to take a look at how network security groups look at existing rules and only update them if some are missing, and use an etag to make sure there is no race with others also updating the NSG from other controllers: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/7a76e16a66d2953bc9467a19e952c5b4b27ee1d9/azure/services/securitygroups/spec.go#L63

**Environment:**

karuppiah7890:
**What type of PR is this?**

/kind bug

Fixes #1696 

Let me know if the release note makes sense. Specifically the mention of `tags` and not `additional tags` (the field in `Azure Cluster`). Also, do check the `TODO`s in the code which contain questions and some questions are also in the issue - #1696 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Fix resource group not getting updated if tags are added
```

### DIFF
--- a/api/v1alpha4/tags.go
+++ b/api/v1alpha4/tags.go
@@ -136,6 +136,12 @@ const (
 	// See https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/
 	// for annotation formatting rules.
 	VMTagsLastAppliedAnnotation = "sigs.k8s.io/cluster-api-provider-azure-last-applied-tags-vm"
+
+	// RGTagsLastAppliedAnnotation is the key for the Azure Cluster object annotation
+	// which tracks the AdditionalTags for Resource Group which is part in the Azure Cluster.
+	// See https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/
+	// for annotation formatting rules.
+	RGTagsLastAppliedAnnotation = "sigs.k8s.io/cluster-api-provider-azure-last-applied-tags-rg"
 )
 
 // SpecVersionHashTagKey is the key for the spec version hash used to enable quick spec difference comparison.
--- a/azure/services/tags/tags.go
+++ b/azure/services/tags/tags.go
@@ -31,7 +31,7 @@ import (
 // TagScope defines the scope interface for a tags service.
 type TagScope interface {
 	logr.Logger
-	azure.ClusterDescriber
+	azure.Authorizer
 	TagsSpecs() []azure.TagsSpec
 	AnnotationJSON(string) (map[string]interface{}, error)
 	UpdateAnnotationJSON(string, map[string]interface{}) error
@@ -94,7 +94,7 @@ func (s *Service) Reconcile(ctx context.Context) error {
 	return nil
 }
 
-// Delete is a no-op as the tags get deleted as part of VM deletion.
+// Delete is a no-op as the tags get deleted as part of resource deletion.
 func (s *Service) Delete(ctx context.Context) error {
 	_, span := tele.Tracer().Start(ctx, "tags.Service.Delete")
 	defer span.End()
--- a/controllers/azurecluster_reconciler.go
+++ b/controllers/azurecluster_reconciler.go
@@ -19,10 +19,12 @@ package controllers
 import (
 	"context"
 
+	"github.com/Azure/azure-sdk-for-go/profiles/latest/resources/mgmt/resources"
 	"github.com/pkg/errors"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1alpha4"
 
 	"sigs.k8s.io/cluster-api-provider-azure/azure"
+	"sigs.k8s.io/cluster-api-provider-azure/azure/converters"
 	"sigs.k8s.io/cluster-api-provider-azure/azure/scope"
 	"sigs.k8s.io/cluster-api-provider-azure/azure/services/bastionhosts"
 	"sigs.k8s.io/cluster-api-provider-azure/azure/services/groups"
@@ -34,6 +36,7 @@ import (
 	"sigs.k8s.io/cluster-api-provider-azure/azure/services/routetables"
 	"sigs.k8s.io/cluster-api-provider-azure/azure/services/securitygroups"
 	"sigs.k8s.io/cluster-api-provider-azure/azure/services/subnets"
+	"sigs.k8s.io/cluster-api-provider-azure/azure/services/tags"
 	"sigs.k8s.io/cluster-api-provider-azure/azure/services/virtualnetworks"
 	"sigs.k8s.io/cluster-api-provider-azure/util/tele"
 )
@@ -52,6 +55,7 @@ type azureClusterService struct {
 	bastionSvc       azure.Reconciler
 	skuCache         *resourceskus.Cache
 	natGatewaySvc    azure.Reconciler
+	tagsSvc          azure.Reconciler
 }
 
 // newAzureClusterService populates all the services based on input scope.
@@ -74,6 +78,7 @@ func newAzureClusterService(scope *scope.ClusterScope) (*azureClusterService, er
 		privateDNSSvc:    privatedns.New(scope),
 		bastionSvc:       bastionhosts.New(scope),
 		skuCache:         skuCache,
+		tagsSvc:          tags.New(scope),
 	}, nil
 }
 
@@ -131,7 +136,49 @@ func (s *azureClusterService) Reconcile(ctx context.Context) error {
 		return errors.Wrap(err, "failed to reconcile bastion")
 	}
 
+	if err := s.reconcileTags(ctx); err != nil {
+		return errors.Wrap(err, "failed to reconcile resource tags")
+	}
+
 	return nil
+}
+
+func (s *azureClusterService) reconcileTags(ctx context.Context) error {
+	// check that the resource group is not BYO.
+	managed, err := s.isResourceGroupManaged(ctx)
+	if err != nil {
+		return errors.Wrap(err, "could not get resource group management state")
+	}
+	if !managed {
+		s.scope.V(2).Info("Should not update resource group tags in unmanaged mode")
+		return azure.ErrNotOwned
+	}
+
+	if err := s.tagsSvc.Reconcile(ctx); err != nil {
+		return errors.Wrap(err, "unable to update tags")
+	}
+
+	return nil
+}
+
+func (s *azureClusterService) isResourceGroupManaged(ctx context.Context) (bool, error) {
+	// group service IsGroupManaged method is not available as we use group service as a azure.Reconciler
+	// which restricts us to only two methods
+
+	// Get resource group client - currently it's a non exported feature in groups package.
+	// Below is duplicated code without telemetry of azure API calls
+	groupsClient := resources.NewGroupsClientWithBaseURI(s.scope.BaseURI(), s.scope.SubscriptionID())
+	azure.SetAutoRestClientDefaults(&groupsClient.Client, s.scope.Authorizer())
+
+	// Get resource group
+	group, err := groupsClient.Get(ctx, s.scope.ResourceGroup())
+	if err != nil {
+		return false, err
+	}
+
+	// check if resource group is managed or not
+	tags := converters.MapToTags(group.Tags)
+	return tags.HasOwned(s.scope.ClusterName()), nil
 }
 
 // Delete reconciles all the services in a predetermined order.
--- a/azure/defaults.go
+++ b/azure/defaults.go
@@ -172,6 +172,11 @@ func WithIndex(name string, n int) string {
 	return fmt.Sprintf("%s-%d", name, n)
 }
 
+// ResourceGroupID returns the azure resource ID for a given resource group.
+func ResourceGroupID(subscriptionID, resourceGroup string) string {
+	return fmt.Sprintf("/subscriptions/%s/resourceGroups/%s", subscriptionID, resourceGroup)
+}
+
 // VMID returns the azure resource ID for a given VM.
 func VMID(subscriptionID, resourceGroup, vmName string) string {
 	return fmt.Sprintf("/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Compute/virtualMachines/%s", subscriptionID, resourceGroup, vmName)
--- a/azure/scope/cluster.go
+++ b/azure/scope/cluster.go
@@ -18,6 +18,7 @@ package scope
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"hash/fnv"
 	"strconv"
@@ -745,5 +746,56 @@ func (s *ClusterScope) UpdatePatchStatus(condition clusterv1.ConditionType, serv
 		conditions.MarkFalse(s.AzureCluster, condition, infrav1.UpdatingReason, clusterv1.ConditionSeverityInfo, "%s updating", service)
 	default:
 		conditions.MarkFalse(s.AzureCluster, condition, infrav1.FailedReason, clusterv1.ConditionSeverityError, "%s failed to update. err: %s", service, err.Error())
+	}
+}
+
+// AnnotationJSON returns a map[string]interface from a JSON annotation.
+func (s *ClusterScope) AnnotationJSON(annotation string) (map[string]interface{}, error) {
+	out := map[string]interface{}{}
+	jsonAnnotation := s.AzureCluster.GetAnnotations()[annotation]
+	if len(jsonAnnotation) == 0 {
+		return out, nil
+	}
+	err := json.Unmarshal([]byte(jsonAnnotation), &out)
+	if err != nil {
+		return out, err
+	}
+	return out, nil
+}
+
+// UpdateAnnotationJSON updates the `annotation` with
+// `content`. `content` in this case should be a `map[string]interface{}`
+// suitable for turning into JSON. This `content` map will be marshalled into a
+// JSON string before being set as the given `annotation`.
+func (s *ClusterScope) UpdateAnnotationJSON(annotation string, content map[string]interface{}) error {
+	b, err := json.Marshal(content)
+	if err != nil {
+		return err
+	}
+	s.SetAnnotation(annotation, string(b))
+	return nil
+}
+
+// SetAnnotation sets a key value annotation on the AzureCluster.
+func (s *ClusterScope) SetAnnotation(key, value string) {
+	if s.AzureCluster.Annotations == nil {
+		s.AzureCluster.Annotations = map[string]string{}
+	}
+	s.AzureCluster.Annotations[key] = value
+}
+
+func (s *ClusterScope) TagsSpecs() []azure.TagsSpec {
+	return []azure.TagsSpec{
+		{
+			Scope: azure.ResourceGroupID(s.SubscriptionID(), s.ResourceGroup()),
+			Tags: infrav1.Build(infrav1.BuildParams{
+				ClusterName: s.ClusterName(),
+				Lifecycle:   infrav1.ResourceLifecycleOwned,
+				Name:        to.StringPtr(s.ResourceGroup()),
+				Role:        to.StringPtr(infrav1.CommonRole),
+				Additional:  s.AdditionalTags(),
+			}),
+			Annotation: infrav1.RGTagsLastAppliedAnnotation,
+		},
 	}
 }
--- a/azure/scope/managedcontrolplane.go
+++ b/azure/scope/managedcontrolplane.go
@@ -19,6 +19,7 @@ package scope
 import (
 	"context"
 	"encoding/base64"
+	"encoding/json"
 	"fmt"
 	"net"
 	"strings"
@@ -602,4 +603,39 @@ func (s *ManagedControlPlaneScope) UpdatePutStatus(condition clusterv1.Condition
 // UpdatePatchStatus updates a condition on the AzureManagedControlPlane status after a PATCH operation.
 func (s *ManagedControlPlaneScope) UpdatePatchStatus(condition clusterv1.ConditionType, service string, err error) {
 	// TODO: add condition to AzureManagedControlPlane status
+}
+
+// AnnotationJSON returns a map[string]interface from a JSON annotation.
+func (s *ManagedControlPlaneScope) AnnotationJSON(annotation string) (map[string]interface{}, error) {
+	out := map[string]interface{}{}
+	jsonAnnotation := s.ControlPlane.GetAnnotations()[annotation]
+	if len(jsonAnnotation) == 0 {
+		return out, nil
+	}
+	err := json.Unmarshal([]byte(jsonAnnotation), &out)
+	if err != nil {
+		return out, err
+	}
+	return out, nil
+}
+
+// UpdateAnnotationJSON updates the `annotation` with
+// `content`. `content` in this case should be a `map[string]interface{}`
+// suitable for turning into JSON. This `content` map will be marshalled into a
+// JSON string before being set as the given `annotation`.
+func (s *ManagedControlPlaneScope) UpdateAnnotationJSON(annotation string, content map[string]interface{}) error {
+	b, err := json.Marshal(content)
+	if err != nil {
+		return err
+	}
+	s.SetAnnotation(annotation, string(b))
+	return nil
+}
+
+// SetAnnotation sets a key value annotation on the ControlPlane.
+func (s *ManagedControlPlaneScope) SetAnnotation(key, value string) {
+	if s.ControlPlane.Annotations == nil {
+		s.ControlPlane.Annotations = map[string]string{}
+	}
+	s.ControlPlane.Annotations[key] = value
 }
--- a/azure/services/groups/client.go
+++ b/azure/services/groups/client.go
@@ -44,7 +44,7 @@ type azureClient struct {
 
 var _ client = (*azureClient)(nil)
 
-// newClient creates a new VM client from subscription ID.
+// newClient creates a new Resource Group client from subscription ID.
 func newClient(auth azure.Authorizer) *azureClient {
 	c := newGroupsClient(auth.SubscriptionID(), auth.BaseURI(), auth.Authorizer())
 	return &azureClient{
--- a/azure/services/groups/spec.go
+++ b/azure/services/groups/spec.go
@@ -51,6 +51,7 @@ func (s *GroupSpec) OwnerResourceName() string {
 func (s *GroupSpec) Parameters(existing interface{}) (interface{}, error) {
 	if existing != nil {
 		// rg already exists, nothing to update.
+		// Note that rg tags are updated separately using tags service
 		return nil, nil
 	}
 	return resources.Group{


karuppiah7890:
I'll fix the CI issues and rerun the tests 👍

karuppiah7890:
TODO:
- Fix failing test(s) in `groups_test.go`
- Add new tests for this bug fix

azure/services/groups/groups.go
```diff
@@ -67,9 +97,34 @@ func (s *Service) Reconcile(ctx context.Context) error {
 
 	err := async.CreateResource(ctx, s.Scope, s.client, groupSpec, serviceName)
 	s.Scope.UpdatePutStatus(infrav1.ResourceGroupReadyCondition, serviceName, err)
+
+	if err != nil {
+		return err
+	}
+
+	err = s.ReconcileTags(ctx)
```
CecileRobertMichon, line 105: Instead of calling ReconcileTags here directly in the group service, what do you think about adding a step in the azurecluster reconciler to call tagsSvc.Reconcile() just like we do already in azuremachine_reconciler for VMs? That way, if we later want to expand the tag updates to other resource types (eg. VNet) in AzureCluster we can just append it to TagSpecs.

karuppiah7890: Makes sense! I'll do that 👍

azure/services/tags/tags.go
```diff
@@ -31,7 +31,7 @@ import (
 // TagScope defines the scope interface for a tags service.
 type TagScope interface {
 	logr.Logger
-	azure.ClusterDescriber
```
karuppiah7890, line 34: @CecileRobertMichon 
Notice - I also made a change in the `tags.go` where I changed `azure.ClusterDescriber` to `azure.Authorizer` as we weren't using any `azure.ClusterDescriber` interface methods in `tags.go`

azure/services/groups/groups.go
karuppiah7890: What about Azure Managed Control Plane reconciler? Should I add to that too? I have already started to handle that too in this PR. If we are looking for a separate issue and PR on that then I can remove those changes from this PR

karuppiah7890: Also, if we move tag reconciliation of resources to the Azure Cluster Reconciler, how do we handle the part about checking if the resource is managed or not? The tag service reconciliation simply loops over all the tag specs and does reconciliation and works only with Kubernetes level data, so if some resources are not managed at some point (owned value tag removed) then those have to be skipped in the tag reconciliation 🤔 while others have to reconcile

karuppiah7890: At Azure Cluster Reconciler level, we don't have full access to the services and azure API clients of the resources, so finding if a resource is managed or not from that level is going to be a bit tricky. I was thinking of two places to put the is-managed check code - either in the reconcileTags or in tag specs in each tag spec, but again, both are tricky and can bloat up pretty quickly as we want managed tags for more resources

controllers/azurecluster_reconciler.go
```diff
@@ -131,9 +136,51 @@ func (s *azureClusterService) Reconcile(ctx context.Context) error {
...
+	}
+
+	if err := s.tagsSvc.Reconcile(ctx); err != nil {
+		return errors.Wrap(err, "unable to update tags")
+	}
+
 	return nil
 }
 
+func (s *azureClusterService) isResourceGroupManaged(ctx context.Context) (bool, error) {
```
karuppiah7890, line 164: @CecileRobertMichon this is one of the ideas that I thought about, but yeah, does not look very great, especially if we start doing this check for each resource we add to tag spec it will get bloated up, also we still have to work out how to skip some reconciliations when we have more than just resource group and when they are not managed and when we don't want to update resource's tags if they are not managed

azure/services/groups/groups.go
karuppiah7890: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1721/files#r715302634

CecileRobertMichon: @karuppiah7890 here are two ways we could workaround this:

1) Do something like https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1684/files#diff-b74c6acb4629abe58f8286da9939563cc4404e09fe9966eb61dbde9b87274189R113 for the RG (store management state at the beginning of the reconcile and use it later to build the tag spec)
2) since we're already getting the existing tags in the tag service, and this is what we use to determine if a resource is managed, we could add a condition for only reconciling tags when the set of existing tags contains the cluster name owned tag. This would also apply to other resources using the tag service (just VMs right now) so we would need to either a) add owned tags to VMs, which we're not doing currently or b) add some bool field to the tagSpec to specify if we should check resource management state and set it to true for groups and false for VMs.

What do you think?

azure/services/tags/tags.go
CecileRobertMichon: good catch, thanks for changing it

azure/services/groups/groups.go
karuppiah7890: I was also thinking about doing the check in the tag service the only odd one out was VMs case, and yeah we could also use the boolean I thought about that too. But if VM is not an exception, that would make things better, just that it would mean touching other parts of the code not related to RG, which could probably done in a separate PR

karuppiah7890: About storing state at the start - I have to check it out, didn't think about that

karuppiah7890: Great ideas, I'll check them out and get back with any questions or thoughts, maybe along with some code

CecileRobertMichon: > add owned tags to VMs, which we're not doing currently

I take this back, looks like we are already adding the same tags to VMs so it shouldn't be an exception: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/services/virtualmachines/virtualmachines.go#L145

karuppiah7890: > I was also thinking about doing the check in the tag service

I was thinking about pulling tags by pulling the whole resource 🙈 I just realized what you meant about tags service pulling tag using `get-at-scope` Tag API which I didn't think about using

By the way, I'm planning to use the update-at-scope Tag API to avoid race conditions and use get-at-scope Tag API to just check if the resource group is managed or not. The only advantage of doing the API call and not using cache would be - updated data - as the it's possible that resource group can get created with owned tag but then the resource group can lose the owned tag value when reconciling the tags, say some external system deleted it. But yeah, it's not exactly a usual thing so the cache will still be helpful in most cases as long as no one is meddling with CAPZ's owned tag value and corresponding tag name

karuppiah7890:
I'm wondering if https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1721#issuecomment-929026734 is related to this PR. Seems like an unrelated failure, not sure though, I'll dig it up sometime

karuppiah7890:
I have updated the PR with latest code. The commit messages have enough information but let me know if it's not clear as to what has been done in the PR or why it has been done

Based on the implementation, I have some questions and thoughts -

Questions -

- Do we want to handle both Azure Cluster and Azure Managed Control Plane Resource Groups in this PR? I made the changes for both, but if we want to separate it out, I could do that too
- Should we write some tests for the Azure Cluster and Azure Managed Control Plane reconciler? Currently I noticed there are no tests. There's not much processing being done as part of this PR with regards to the reconciler and the Cluster scope and Managed Control Plane Scope, it's more of passing around data. Only in `AnnotationJSON` and `UpdateAnnotationJSON` - there's some JSON serialization and JSON parsing code. I haven't added any tests at these levels, as it was mostly just data passing and less processing but let me know what you folks think
- Should we name `created` in tags service to something like `createdOrUpdated`? Currently it's clear only in one place, where it says

```go
// Tracking for created/updated
created := map[string]string{}
```

everywhere else we use `created` though it includes `updated` tags too, where tag value is updated. Or we could just rename it as `updated` which can encompass a newly created tag too
- Currently as part of this PR's implementation, we do GET resource tags request to check if the resources are managed based on the owned tag value. In the case of managed resource, here we assume that the owned tag still remains while we do diff processing and then call APIs for updates based on diffs. I think this is a somewhat fair assumption. But yeah it's totally possible for an external entity to remove the owned tag value just after our GET request and just before our update API call. What do you folks think about this scenario?

Thoughts -

- Currently we do a diff using tag spec tags (desired tags) vs last applied tags which is present in the annotation. And then we do appropriate patch API calls (merge / delete) for the diff. There's one downside to this approach, it's possible that an external entity changed the CAPZ managed tags of the resource, and we would miss out to take care of that while reconciling. For example, with the current implementation in this PR -

Initial Azure Resource Tags: `{ c: c1, d: d1 }`
Tag Spec: `{ a: a1, b: b1 }`
Last Applied Tags from K8s Annotation: `{ a: a1, b: b1 }`
No diff, no calls to Azure API to make changes
Final Azure Resource Tags: `{ c: c1, d: d1 }`

In the above case, ideally, Azure Resource Tags should have been patched (merged) with `{ a: a1, b: b1 }` to finally get `{ a: a1, b: b1, c: c1, d: d1 }`. Why? As it looks like someone meddled with the tags that are managed by CAPZ and CAPZ has to reconcile it. Another example is below

Initial Azure Resource Tags: `{ c: c1, d: d1 }`
Tag Spec: `{ a: a1, b: b1 }`
Last Applied Tags from K8s Annotation: `{ a: a1 }`
There is diff, one patch (merge) call with `{ b: b1 }` to Azure API to make changes
Final Azure Resource Tags: `{ b: b1, c: c1, d: d1 }`

In the above case, ideally, Azure Resource Tags should have been patched (merged) with `{ a: a1, b: b1 }` to finally get `{ a: a1, b: b1, c: c1, d: d1 }`.

I didn't write test cases for such scenarios though, though I could have. I wanted to discuss it first without code / test code. Let me know if you want to look at test cases for the above examples in case the examples are not very clear or just to help with discussions, I could write the test cases

I think the diff should be done something like this -

For finding deletions, we can look at last applied tags from annotation and then compare it with the desired tag spec

For finding updated / created, we can look at current Azure resource tags from Azure API and then compare it with the desired tag spec

How does that sound?

- `tagsChanged` function returns `map[string]string` for created-or-update and deleted, which we later have to convert to `map[string]*string` for doing the update-at-scope API call for both created-or-updated and deleted. I was wondering if that's okay or if we should we just return `map[string]*string` from `tagsChanged`. Though it's sole responsibility is to help with finding with the tags changed, it could also help more by providing data of appropriate data type for it's only caller which is the tags service


Let me know what you folks think about the code, tests and the above questions and thoughts! Thanks!

karuppiah7890:
cc @CecileRobertMichon ^

CecileRobertMichon:
@karuppiah7890 first of all, thank you for all the detailed explanations and walking us through your thought process - makes my job as a reviewer a lot easier 🙂  I really like what you came up with, I think the code is more clean and reusable now that we're checking for the owned tags directly in the tag service reconcile. To answer your questions above:

> Do we want to handle both Azure Cluster and Azure Managed Control Plane Resource Groups in this PR? I made the changes for both, but if we want to separate it out, I could do that too

yes, let's do both in this PR

> Should we write some tests for the Azure Cluster and Azure Managed Control Plane reconciler?

more tests are always welcome but as you say not much is happening in the reconcilers. Perhaps let's add the tag svc to https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/341acbedc9074e179802bcf30c083768cba839e6/controllers/azurecluster_reconciler_test.go. And I see you already added test cases to tags_test.go for unmanaged/managed resources so 👍 

 
> Should we name created in tags service to something like createdOrUpdated? Currently it's clear only in one place, where it says

sure, that makes sense


In terms of the risk of someone modifying tags out of band, I think our assumption should be that if we're managing the resource and we're also providing a way for the user to add custom tags, we shouldn't allow for external actors to also manage them. Otherwise, we introduce an ambiguous source of truth and it becomes difficult to reconcile changes when they happen in both as you pointed out. If the user wants to manage their own tags out of band, they can remove the managed tag and take it from there. 

That being said, GETs are cheap and I think it's totally fair to do a GET on each loop to determine if anything has changed. If we're already doing the GET for create/update, we can use the same data for deleted tags too and the annotation becomes sort of useless. This PR is already an incremental improvement though so if you want to open an issue for someone else to work on this later that'd be totally fine.

For the issue of the race condition where someone updates the tag after we do the GET and before the PUT, does the tag CreateOrUpdateAtScope API have etags supported? if so, we can use those to make changes only if the etag hasn't changed. If not, I think it's not a huge problem to assume the owned tag doesn't get deleted in between.

karuppiah7890:
Thanks to you too @CecileRobertMichon for helping review this PR promptly with lot of help through comments here and in the issue

> Perhaps let's add the tag svc to https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/341acbedc9074e179802bcf30c083768cba839e6/controllers/azurecluster_reconciler_test.go.

I was going to do that. The Azure cluster reconciler tests passed regardless (with no nil pointer exceptions) and when I checked it out, I noticed we don't have tests for the `Reconcile` method and we only have tests for Azure cluster `Delete` method

> we shouldn't allow for external actors to also manage them

Ah okay. Based on the discussion in the issue #1696 , I was under the impression that CAPZ has it's own set of tags that it can manage and shouldn't be interfering (by overwriting) with the tags that are added and managed by users / other systems for other purposes, for example to help with billing? This is based on the comments from @devigned - https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1696#issuecomment-925060775 , https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1696#issuecomment-925078929 , https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1696#issuecomment-925130687 .

> Otherwise, we introduce an ambiguous source of truth and it becomes difficult to reconcile changes when they happen in both as you pointed out.

True, the thing is I have noticed this kind of feature/behavior with Kubernetes too where changes can happen outside the managed fields. I remember reading `kubectl apply` helps manage resources using YAML but also retains some external modifications by users which are not part of the YAML. I think [this doc](https://kubernetes.io/docs/concepts/cluster-administration/manage-deployment/#kubectl-apply) talks a bit about it, I think the term popularly used is the "three way diff" or "three way merge". So if we think about a similar narrative here - it's more like there are CAPZ managed tags, and then there could be tags managed and owned by other systems or users which they can modify. We track what we manage using the last-applied-tags annotation in the case of tags service, similar to Kubernetes last-applied annotation

> That being said, GETs are cheap and I think it's totally fair to do a GET on each loop to determine if anything has changed. If we're already doing the GET for create/update, we can use the same data for deleted tags too and the annotation becomes sort of useless.

Makes sense. If we are not looking to let other external systems / users manage their own separate set of tags which are different from CAPZ managed tags, then yeah, it's not needed to use the annotation and we can just use GET for current resource tags and compare it with the desired resource tags in the tag spec. Currently the implementation in `main` branch the annotation is already being used, and I believe it is exactly to support other external entities to manage their own separate tags. It does GET to get all resource tags - CAPZ managed and probably other tags managed by other external entities, and then does a diff between last-applied annotation tags and desired resource tags in the tag spec to get the diff only for tags managed by CAPZ. The only tricky thing is, with the implementation in `main` we might overwrite the data of other systems as mentioned in [this comment](https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1696#issuecomment-925060775) and it also has the downsides of the implementation in this PR, which is, it's possible that the annotation and desired don't have diff but the desired tags are not in current resource, maybe because some external entity removed it

> For the issue of the race condition where someone updates the tag after we do the GET and before the PUT, does the tag CreateOrUpdateAtScope API have etags supported? if so, we can use those to make changes only if the etag hasn't changed. If not, I think it's not a huge problem to assume the owned tag doesn't get deleted in between.

What do you mean by `someone`? You mentioned we expect only CAPZ to manage the tags and no external entities, so is it another instance of CAPZ? and `CreateOrUpdateAtScope` API does not support etags from what I see in the API docs [here](https://docs.microsoft.com/en-us/rest/api/resources/tags/create-or-update-at-scope)

Also, in this PR, I have replaced the `CreateOrUpdateAtScope` with two [`UpdateAtScope`](https://docs.microsoft.com/en-us/rest/api/resources/tags/update-at-scope) API calls - one for create or update using `Merge` [patch operation](https://docs.microsoft.com/en-us/rest/api/resources/tags/update-at-scope#tagspatchoperation) and another for delete using `Delete` [patch operation](https://docs.microsoft.com/en-us/rest/api/resources/tags/update-at-scope#tagspatchoperation). This is based on the [comment here](https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1696#issuecomment-925013265) from @devigned . This will help prevent us from race conditions [mentoned here](https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1696#issuecomment-925078929) I think - where we overwrite data of external systems by mistake. But I'm not sure if you were talking about external entities

So the big question is - when we say a resource is managed by CAPZ (using the `owned` tag value), should **all** the tags on the resource be managed by CAPZ? Or is it okay for some tags to be added and managed and owned by external entities? Assuming these tags by external entities are totally different (different tag keys) from what CAPZ manages, because if there's conflict / clash there, then CAPZ and external entities are going to have a hard time updating the tags unnecessarily. cc @CecileRobertMichon @devigned

karuppiah7890:
The API diff in https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1721#issuecomment-930058147 is expected based on the PR's changes. But yeah, it has breaking changes / backwards incompatible changes for some exported entities `TagScope`. I could revert it too, the change is because I noticed we don't need all the methods present in `TagsScope`. But if external users importing the library use `TagScope` and are using the methods that are being removed, then yes, this will be a breaking change and a bad experience. Let me know what you folks think about it. Commit - https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1721/commits/f4433aab408a02c83708c8e37e8f682494ebe57f

karuppiah7890:
I'll check the other failures 👍

karuppiah7890:
> If that's the case, I'm +1 on allowing external entities to also add and manage tags.

Cool!

> Let's give CAPZ priority for setting tags, if the key is the same, we should set the value to the one in the AzureCluster spec.

Makes sense. From CAPZ point of view, we will always ensure that the tags we manage are present in the CAPZ managed resources

> About race condition:
> 
> By someone I meant external entities updating the same Azure resource. Example, a user makes an Azure CLI call on the resource group. Will the patch operation prevent a race condition in the following scenario?
> 
>     * User creates CAPZ cluster with tags `foo: bar`
> 
>     * User adds tag `color: purple` in AzureCluster spec
> 
>     * Reconciliation starts, CAPZ GETs existing tags from API  `foo: bar`
> 
>     * User adds tag `billing: y` with Azure CLI
> 
>     * CAPZ tries to patch tags with `foo: bar` and `color: purple` (doesn't know about `billing: y`)
> 
> 
> would the operation fail? would `billing: y` get overwritten or would it be untouched?

Yeah, to solve this problem of overwriting we will be using the [Tags Update At Scope API](https://docs.microsoft.com/en-us/rest/api/resources/tags/update-at-scope) which has the concept of [tags patch operation](https://docs.microsoft.com/en-us/rest/api/resources/tags/update-at-scope#tagspatchoperation) where `Merge` is one option and it will only patch/merge the mentioned tags and not touch the other tags. We will be using the `Delete` option to delete mentioned tags without touching the other tags

karuppiah7890:
I was thinking about the updation of last-applied-tags annotation and when and how we do it. Currently we update the annotation whenever there's a change in tags - tag deletion or tag updation

https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/7f78c7bc945b8cabc35c476c09013fb1a9dab320/azure/services/tags/tags.go#L64-L65

https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/7f78c7bc945b8cabc35c476c09013fb1a9dab320/azure/services/tags/tags.go#L87-L90

And both tag deletion and tag updation are found out using the last-applied-tags annotation and the desired tags spec. So if the annotation is initially empty, there will be a change and the last-applied-tags annotation will get updated

But if we change the logic now as part of this PR to -

tag deletion - find by comparing last-applied-tags annotation and the desired tags spec
tag updation / creation - find by comparing the desired tags spec and the Azure resource tags from Azure API

and if we still update last-applied-tags annotation only when there's a change in tags, then the flow of operations for resource group tags will look like this -

* User creates CAPZ cluster with tags `{ a: a1, b: b1 }`
* AzureCluster Reconciliation starts
  * Resource Group Service Reconcile - Resource group gets created with tags `{ a: a1, b: b1 }`
  * Tags Service Reconcile
    * no last-applied-tags annotation, so comparing with desired tags spec `{ a: a1, b: b1 }` , there's no tag deletion that's needed, as according to the annotation, we didn't manage any tags previously
    * Since the resource group just got created with the tags `{ a: a1, b: b1 }` , comparing desired tags spec with Azure resource group tags `{ a: a1, b: b1 }` from Azure API, there's no tag updation that's needed
    * No changes, so Tags service does no Azure API calls
    * Since there's no changes / applies from Tags service, it also makes sense to NOT update last-applied-tags annotation as nothing was applied in this reconcile
* User removes `a: a1` tag in AzureCluster spec, making the tags as `{ b: b1 }`
  * Resource Group Service Reconcile - Resource group already exists and no update actions are taken
  * Tags Service Reconcile
    * no last-applied-tags annotation, so comparing with desired tags spec `{ b: b1 }`, there's no tag deletion that's needed, as according to the annotation, we didn't manage any tags previously. _Note that this should **not** happen ideally, as we need to delete `a: a1` tag in this case_
    * Comparing desired tags spec `{ b: b1 }` with Azure resource group tags `{ a: a1, b: b1 }` from Azure API, there's no tag creation or updation that's needed
    * No changes, so Tags service does no Azure API calls. _Note that this should **not** happen ideally, as we need to delete `a: a1` tag in this case_
    * Since there's no changes / applies from Tags service, it also makes sense to NOT update last-applied-tags annotation as nothing was applied in this reconcile. _Note that this should **not** happen ideally, as we need to delete `a: a1` tag in this case and update the last-applied-tags annotation with `{ a: a1, b: b1 }`_

To avoid this problem of wrong value of last-applied-tags annotation causing deletion of tags to not work, we need to think about how and when we update it.

For example, we could do one of the following
- When Resource groups are created along with tags - we can also update the last-applied-tags annotation as we are applying the tags to the resource group using the Azure API and that can be as a very valid last-applied-tags annotation value. This will help Tags service to work as is and update last-applied-tags annotation when any other creation/updation/deletion to tags happen. This would also mean that in the future every resource whose tags are going to be maintained and updated has to update the last-applied-tags annotation as part of it's create code - only create code. Update will be taken care by tags service - as Tags service can do patch updates without overwriting data of other entities
- Another ideas is - we can avoid creating tags as part of resource creation, and instead handle creation of tags in the tags service. This way - everything (create, update, delete) will be maintained in the tags service. And when the tags are created by tags service, there will be a change and it will update the last-applied-tags annotation value. This way every resource in the future can leave the tag creation, updation and deletion to the tags service and not have any code related to tags in the resource service code. But yeah, this would also mean that, when the resource is created, it will get it's tags as part of the resource creation. Also, we will have to think through the managed and unmanaged resources part. Managed resources - we find it out using tags, and if we don't create it during resource creation, that's very tricky. Or we can atleast create the `owned` tag value tag during resource creation and handle everything else separately in tags service. Not to mention, tags service also works only with managed resources, so if we miss to put `owned` value as part of resource creation, we will never know who created the resource including in the tag service which modifies and maintains tags only for managed resources
- Another idea is - Let resource group service create resource group along with tags. When tags service does reconcile, handle the special case of first time reconcile where last-applied-tags annotation is empty and update it. We have to think this through though. And ensure there are no problems by handling such a special case here. Not to mention, when handling this special case, it would mean that we would update the last-applied-tags annotation in first reconcile alone in tags service even when tags service does no Azure API calls (no applies) as resource group was already created with tags in resource group service

karuppiah7890:
@CecileRobertMichon Let me know if you have other ideas ^

CecileRobertMichon:
@karuppiah7890 how about creating the owned tags (and all other default CAPZ tags) as part of resource creation and handling all additionalTags create/update/delete in the tag service?

karuppiah7890:
Makes sense 👍

azure/services/tags/tags.go
```diff
@@ -57,31 +59,47 @@ func (s *Service) Reconcile(ctx context.Context) error {
...
-			if err != nil {
-				return errors.Wrap(err, "failed to get existing tags")
-			}
-			tags := make(map[string]*string)
-			if result.Properties != nil && result.Properties.Tags != nil {
-				tags = result.Properties.Tags
-			}
-			for k, v := range created {
-				tags[k] = to.StringPtr(v)
+			if len(createdOrUpdated) > 0 {
```
shysank, line 83: nit: not going to cause any huge performance difference, but I feel (update + delete) may be costlier than createAndUpdate except maybe in cases when there are large number of user created tags, then we'll be sending all those values across the network.

api/v1alpha4/tags.go
```diff
@@ -136,6 +136,12 @@ const (
 	// See https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/
 	// for annotation formatting rules.
 	VMTagsLastAppliedAnnotation = "sigs.k8s.io/cluster-api-provider-azure-last-applied-tags-vm"
+
+	// RGTagsLastAppliedAnnotation is the key for the Azure Cluster object annotation
+	// which tracks the AdditionalTags for Resource Group which is part in the Azure Cluster.
```
shysank, line 141: nit
```suggestion
	// which tracks the capz specific tags for Resource Group which includes the AdditionalTags 
	// set in AzureCluster, and AzureManagedControlPlane.
```

shysank (commented):
lgtm with a couple of nits, and pending squash.

api/v1alpha4/tags.go
karuppiah7890: I have to think about this change. This might be true with the current implementation, which isn't reflecting some of the changes that was discussed in this PR as I'm yet to work on updating the PR. Recently we had a conversation where we were talking about how the tags will be managed, specifically talking about how we will track tag deletion using the last-applied-tags annotation and what it means. I'll think on the implementation next week and get back. You can follow the conversation here - https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1721#issuecomment-934598442 and here - https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1721#issuecomment-934301110 . We settled on creating CAPZ default tags as part of resource creation in the appropriate resource service and then manage tag create, update and delete in tags service, and at first thought I think tags service will find created, update, deleted and finally update the last-applied-tags annotation with full desired spec - which currently seems to be default + additional tags, but I might change the implementation where the tag spec will just contain additional tags, then annotation would just track additional tags, then the above change won't be needed. But yeah, I'll first implement everything and then get back to this anyway, when the code is more finalized

azure/services/tags/tags.go
shysank: Yes, 2 calls vs 1 call was what I was meaning to say.

karuppiah7890:
Question: I'm not very familiar with the `api/` package and the different API versions in it. I'll give it a read sometime and check docs too and try to understand it. In the meanwhile, I wanted to ask which package the `RGTagsLastAppliedAnnotation` constant should be defined in. Previously it was in `sigs.k8s.io/cluster-api-provider-azure/api/v1alpha4`. Now I'm moving it to `sigs.k8s.io/cluster-api-provider-azure/api/v1beta1`. Should it be present in both? I'm yet to figure it out. If I do, I'll keep you folks posted, till then I'll keep the question here

PR #1742, opened 2021-09-30
Optionally specify agent pool name in AMMP

### LINKED ISSUE #1689: CAPZ AKS - cannot create node pools with the same name under multiple AKS clusters defined in the same namespace
/kind bug

I cannot create node pools with the same name under multiple AKS clusters defined in the same namespace.
In CAPZ to provision AKS cluster, the machinepool/Azuremanagedmachinepool object uses the k8s object name from metadata to create the AKS node pools. Therefore, it's not possible to create node pools with the same name in the same namespace even though these node pools belong to different AKS clusters.

I was trying to create multiple AKS clusters using CAPZ and put the CRs under same namespace (so that controller can only watch this namespace), but I cannot put same name for node pools belong to different AKS clusters.

We should be able to have node pools with same name under different AKS clusters, which is supported by AKS.

**Environment:**
Capz

- cluster-api-provider-azure version:  0.5.2
- Kubernetes version: (use `kubectl version`): 1.21.1 
- OS (e.g. from `/etc/os-release`): linux

richardchen331:
**What type of PR is this?**
/kind bug

Allow user to specify agent pool name in AMMP spec (instead of relying on the CR name in metadata)

Fixes #1689

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Allow user to specify agent pool name in AMMP spec (instead of relying on the CR name in metadata)
```


azure/scope/managedcontrolplane.go
```diff
@@ -526,8 +533,15 @@ func (s *ManagedControlPlaneScope) AgentPoolSpec() azure.AgentPoolSpec {
 		replicas = *s.MachinePool.Spec.Replicas
 	}
 
+	var name string
+	if s.InfraMachinePool.Spec.Name != nil {
```
shysank, line 537: can we instead do this in webhook to check for `spec.name`, and default to `metadata.name`? wdyt?

richardchen331: make sense to me! updated the webhook w/ defaulting logic

exp/api/v1alpha4/azuremanagedmachinepool_webhook.go
```diff
@@ -44,6 +44,10 @@ func (r *AzureManagedMachinePool) Default(client client.Client) {
 		r.Labels = make(map[string]string)
 	}
 	r.Labels[LabelAgentPoolMode] = r.Spec.Mode
+
+	if r.Spec.Name == nil {
```
shysank, line 48: just to be safe
```suggestion
	if r.Spec.Name == nil  || r.Spec.Name == ""{
```
Not sure

shysank: Can we also add/update tests for this?

richardchen331: Sure. Added test for this.

I left the defaulting logic as is, but added validation logic to check if `Spec.Name == ""`

shysank: I see what you're doing in `validate` but AFAIK the general way of handling empty strings in defaults has been to just assign a default like [here](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/7f78c7bc945b8cabc35c476c09013fb1a9dab320/api/v1alpha4/azurecluster_default.go#L77) for example. I'll leave to other reviewers on this.

shysank (commented):
Overall lgtm pending clarification on handling empty string. @richardchen331 can you please squash the commits when you get a chance.

exp/api/v1alpha4/azuremanagedmachinepool_webhook.go
richardchen331: oh i see, makes sense. Update the code according to your suggestion. Squashed the commits as well. Thanks for reviewing the PR @shysank !

exp/api/v1alpha4/azuremanagedmachinepool_webhook.go
```diff
@@ -114,6 +118,25 @@ func (r *AzureManagedMachinePool) ValidateDelete(client client.Client) error {
 	return errors.Wrapf(r.validateLastSystemNodePool(client), "if the delete is triggered via owner MachinePool please refer to trouble shooting section in https://capz.sigs.k8s.io/topics/managedcluster.html")
 }
 
+// Validate the AzureManagedMachinePool and return an aggregate error.
+func (r *AzureManagedMachinePool) Validate(client client.Client) error {
```
shysank, line 122: Is this needed now? Can we instead add a test for default webhook?

richardchen331: Thanks for catching that. Removed the validation logic. Added tests for default webhook.

PR #1744, opened 2021-09-30
Assign all azs to public ip for backwards compatibility

### LINKED ISSUE #1739: cluster upgrade fails when provider is updated from 0.4.14 to 0.5.x
/kind bug

1. Create a MC using CAPZ 1.4.14.
2. Create WC with k8s version 1.20
3. Upgrade MC to CAPZ 1.5.2
4. Upgrade WC to k8s version 1.21. 
5. fails in pip reconciliation with the following error 
`
Error: Code=\"ResourceAvailabilityZonesCannotBeModified\" Message=\"Resource /subscriptions/bla/resourceGroups/bla/providers/Microsoft.Network/publicIPAddresses/bla-apiserver has an existing availability zone constraint 1, 2, 3 and the request has availability zone constraint NoZone, which do not match. Zones cannot be added/updated/removed once the resource is created. The resource cannot be updated from regional to zonal or vice-versa.\
`
The cluster should upgrade 

**Environment:**

shysank:
**What type of PR is this?**
/kind bug

This pr assigns all available azs in the cluster's region to public ip. This is required to mimic the old sdk behaviour so that clusters created in version prior to v0.4.15 don't break.

Fixes #1739 

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Assign all azs to public ip for backwards compatibility
```

### DIFF
--- a/azure/scope/cluster.go
+++ b/azure/scope/cluster.go
@@ -605,6 +605,17 @@ func (s *ClusterScope) SetFailureDomain(id string, spec clusterv1.FailureDomainS
 	s.AzureCluster.Status.FailureDomains[id] = spec
 }
 
+// FailureDomains returns the failure domains for the cluster.
+func (s *ClusterScope) FailureDomains() []string {
+	fds := make([]string, len(s.AzureCluster.Status.FailureDomains))
+	i := 0
+	for id := range s.AzureCluster.Status.FailureDomains {
+		fds[i] = id
+		i++
+	}
+	return fds
+}
+
 // SetControlPlaneSecurityRules sets the default security rules of the control plane subnet.
 // Note that this is not done in a webhook as it requires a valid Cluster object to exist to get the API Server port.
 func (s *ClusterScope) SetControlPlaneSecurityRules() {
--- a/azure/interfaces.go
+++ b/azure/interfaces.go
@@ -79,6 +79,7 @@ type ClusterDescriber interface {
 	AdditionalTags() infrav1.Tags
 	AvailabilitySetEnabled() bool
 	CloudProviderConfigOverrides() *infrav1.CloudProviderConfigOverrides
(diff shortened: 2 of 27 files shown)


shysank:
I tested the following scenario locally using tilt:
1. Deploy capz v0.4.14, and capi 0.3.23
2. Create a v1alpha3 cluster
3. Update capi to 0.4.x and capz to the current version with the changes
4. Observe public ips reconciliation during cluster reconciliation. There were no errors.

azure/scope/cluster.go
```diff
@@ -605,6 +605,17 @@ func (s *ClusterScope) SetFailureDomain(id string, spec clusterv1.FailureDomainS
 	s.AzureCluster.Status.FailureDomains[id] = spec
 }
 
+// FailureDomains returns the failure domains for the cluster.
+func (s *ClusterScope) FailureDomains() []string {
+	fds := make([]string, len(s.AzureCluster.Status.FailureDomains))
+	i := 0
```
devigned, line 611: Total nit, but perhaps use
```go
	for i :=0; i<len(s.AzureCluster.Status.FailureDomains); i++ {
		fds[i] = s.AzureCluster.Stutus.FailureDomains[i]
	}
```

devigned (changes requested):
Looks great. Just a little nit.

azure/scope/cluster.go
shysank: I don't think we can do this since `s.AzureCluster.Status.FailureDomains` is a `map[string]FailureDomainSpec`.

devigned: You are correct. I was thinking it was a slice. Thank you for the correction.

devigned (approved):
/lgtm

PR #1740, opened 2021-09-28
[WIP] update api types to v1beta1

shysank:
**What type of PR is this?**
/kind api-change

**What this PR does / why we need it**:
- Updates capi to latest commit as of yesterday (09/27) 
- Adds new api type v1beta1
- Adds v1alpha3 -> v1beta1 conversion
- Adds v1alpha4 -> v1beta1 conversoin

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #1724 #1725 #1723 

**Special notes for your reviewer**:

I have made a separate commit for each logical block. Please let me know if I can make it easier to review.

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
update api types to v1beta1
```

PR #1747, opened 2021-10-01
Add NodeOutboundIPs customization in Azure cluster

### LINKED ISSUE #1394: Support existing IP as node outbound IPs

Ankitasw:
**What type of PR is this?**
/kind feature

This PR allows to add `frontendIPs` to `NodeOutboundLB` to allow users to use existing IPs as nodeoutbound IPs

Fixes #1394 

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Release note**:
```release-note
Add support for using existing IPs as nodeoutbound IPs in Azure cluster
```


CecileRobertMichon:
/hold cancel

#1740 has merged, please rebase and move api changes to v1beta1

CecileRobertMichon:
> Adding outboundIP property to FrontendIP.PublicIP seems fair to me. @CecileRobertMichon do you have any thoughts?

Sorry I missed this earlier. +1 to Shyam's suggestion.

api/v1beta1/azurecluster_validation.go
```diff
@@ -498,3 +508,12 @@ func validateCloudProviderConfigOverrides(old, new *CloudProviderConfigOverrides
 	}
 	return allErrs
 }
+
+func validateOutboundIP(frontendIP FrontendIP) error {
+	outboundIP := net.ParseIP(frontendIP.PublicIP.OutboundIP)
+	_, _, err := net.ParseCIDR(frontendIP.PublicIP.OutboundIP)
+	if outboundIP == nil && err != nil {
+		return err
```
shysank, line 516: can we wrap this into a error that says something about `outboundIP==nil` as well?

api/v1beta1/azurecluster_validation_test.go
```diff
@@ -1024,9 +1024,11 @@ func TestValidateNodeOutboundLB(t *testing.T) {
 			name: "FrontendIps can update when frontendIpsCount changes",
 			lb: &LoadBalancerSpec{
 				FrontendIPs: []FrontendIP{{
-					Name: "some-frontend-ip-1",
+					Name:             "some-frontend-ip-1",
+					PrivateIPAddress: "1.2.3.4",
```
shysank, line 1028: is this test related to the changes?

api/v1beta1/azurecluster_validation_test.go
```diff
@@ -1024,9 +1024,11 @@ func TestValidateNodeOutboundLB(t *testing.T) {
...
 				}, {
-					Name: "some-frontend-ip-2",
+					Name:             "some-frontend-ip-2",
+					PrivateIPAddress: "5.6.7.8",
```
shysank, line 1031: same comment as above

api/v1beta1/azurecluster_validation.go
```diff
@@ -498,3 +508,12 @@ func validateCloudProviderConfigOverrides(old, new *CloudProviderConfigOverrides
...
+	}
+	return nil
```
shysank, line 518: wdyt about this?
```suggestion
	outboundIP := net.ParseIP(frontendIP.PublicIP.OutboundIP)
	if outboundIP != nil {
	   return nil
	}
	
	_, _, err := net.ParseCIDR(frontendIP.PublicIP.OutboundIP)
	if err == nil {
	   return nil
	}
	
	return errors.Wrap("could not parse into an ip or prefix", err)
```

shysank: or atleast wrap the error with information about `outboundIP == nil` condition

azure/services/publicips/publicips.go
```diff
@@ -149,3 +149,29 @@ func (s *Service) isIPManaged(ctx context.Context, ipName string) (bool, error)
 	tags := converters.MapToTags(ip.Tags)
 	return tags.HasOwned(s.Scope.ClusterName()), nil
 }
+
+func getPublicIPAddressPropertiesFormat(addressVersion network.IPVersion, dnsSettings *network.PublicIPAddressDNSSettings, outboundIP, outboundIPPrefix net.IP) *network.PublicIPAddressPropertiesFormat {
+	if outboundIP != nil {
```
shysank, line 154: can we set the common fields outside of the conditions and append outbound ip or prefix based on the inputs?

shysank: I'd also suggest moving the parsing logic inside the helper and parse lazily only if `ip.OutboundIP != nil`

shysank: something like 
```
pipAddressProperties := &network.PublicIPAddressPropertiesFormat{
			PublicIPAddressVersion:   addressVersion,
			PublicIPAllocationMethod: network.IPAllocationMethodStatic,
			DNSSettings:              dnsSettings,
		}

if ip.outboundip == nil {
return pipAddressProperties
}

// append outbound ip/prefix here

```

api/v1alpha4/azurecluster_conversion.go
```diff
@@ -40,6 +40,38 @@ func (src *AzureCluster) ConvertTo(dstRaw conversion.Hub) error { // nolint
 	// Restore list of virtual network peerings
 	dst.Spec.NetworkSpec.Vnet.Peerings = restored.Spec.NetworkSpec.Vnet.Peerings
 
+	if restored.Spec.NetworkSpec.NodeOutboundLB != nil {
+		for i, _ := range restored.Spec.NetworkSpec.NodeOutboundLB.FrontendIPs {
+			dst.Spec.NetworkSpec.NodeOutboundLB.FrontendIPs[i].PublicIP = restored.Spec.NetworkSpec.NodeOutboundLB.FrontendIPs[i].PublicIP
```
shysank, line 45: shouldn't we only set publicIP.outboundIP?

api/v1beta1/azurecluster_validation.go
```diff
@@ -437,6 +437,16 @@ func validateNodeOutboundLB(lb *LoadBalancerSpec, old *LoadBalancerSpec, apiserv
 			fmt.Sprintf("Node outbound idle timeout should be between %d and %d minutes", MinLBIdleTimeoutInMinutes, MaxLoadBalancerOutboundIPs)))
 	}
 
+	for index, frontendIP := range lb.FrontendIPs {
```
shysank, line 440: Although we intend to allow outboundIP only for nodeoutboundLB, there is nothing stopping the user from setting it in other places where publicIP is used (eg. AzureBastion). Should we add validations there to prevent the user from setting it?

azure/services/publicips/publicips.go
```diff
@@ -75,6 +76,9 @@ func (s *Service) Reconcile(ctx context.Context) error {
 			}
 		}
 
+		outboundIP := net.ParseIP(ip.OutboundIP)
```
shysank, line 79: We should add mapping from api to scope for the new field `outboundIP` => https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/45519b18be2cb7a26034419df99d46c87b410285/azure/scope/cluster.go#L125

api/v1beta1/azurecluster_validation_test.go
Ankitasw: no will remove it, was added in previous changes

api/v1beta1/azurecluster_validation_test.go
Ankitasw: same as above

api/v1beta1/azurecluster_validation.go
Ankitasw: sounds fair to me

api/v1alpha4/azurecluster_conversion.go
Ankitasw: missed it, will add only outboundIP here

api/v1beta1/azurecluster_validation.go
Ankitasw: I notices PublicIP is used only in AzureBastion other than loadbalancer spec, do you know any other places where it is used or I have missed?

azure/services/publicips/publicips.go
Ankitasw: something like [this](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/57ca45a908419f8ae0ed2a78e8bacc53db25a41f/azure/scope/cluster.go#L734) ?

azure/services/publicips/publicips.go
Ankitasw: fixed now

azure/services/publicips/publicips.go
shysank: yeah something like that. To verify the expected behavior, I'd suggest creating a cluster using tilt and see if the appropriate resources are created.

Ankitasw: ok but how to use existing IP, I mean which IPs can I use for testing?

Issue #1692, opened 2021-09-16
PublicIP resources should not PUT on each AzureCluster reconcile


devigned:
/kind bug

[Before submitting an issue, have you checked the [Troubleshooting Guide](https://capz.sigs.k8s.io/topics/troubleshooting.html)?]

**What steps did you take and what happened:**
When reconciling an `AzureCluster` each reconcile loop will cause a PUT for each PublicIP resource.

**What did you expect to happen:**
PublicIP resources should only be mutated with a PUT if the desired state has changed.

/assign @CecileRobertMichon

karuppiah7890:
I'll see if this can be handled as part of the sync to async migration in #1716 or else migrate as is. I'll check the code and see the difference in effort with and without the bug fix

karuppiah7890:
cc @CecileRobertMichon

karuppiah7890:
@devigned I was just trying out the PUT API multiple times with same request body to see the behavior. I see that the public IP (I tried IPv6) is not affected. What's the aim of this issue? To reduce the API calls to Azure when there's no change? and to also avoid PUT request which is usually meant to do updates (generally by overwrites)? Or is there any other aim / extra aim for this issue?

If it's only about avoiding unnecessary PUT API calls, how do we plan to do it? Currently I see there's no GET API call in `publicips` service [in `Reconcile`](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/services/publicips/publicips.go#L55-L56). So, do we want to do a GET API call and compare the current spec from K8s with the current public IP resource from Azure and do PUT API call only if there's a diff?

karuppiah7890:
Maybe we should also talk about what kind of updates are allowed for Public IPs. I mean, looking at some of the validations being done [here](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/e5327290e4fd5df3fae5779296b8e1d2c73c508e/api/v1alpha4/azurecluster_validation.go) by using the [webhook](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/e5327290e4fd5df3fae5779296b8e1d2c73c508e/api/v1alpha4/azurecluster_webhook.go), looks like there's not much we can change for anything related to Public IPs

[...]

PR #1791, opened 2021-10-25
chore(private_dns): tag dns zone and delete only if owned

### LINKED ISSUE #1749: BYO DNS Zone
/kind feature

1. Created a private cluster with custom vnet
2. Create a dns zone through azure cli/ui and the appropriate links and records 
3. Delete the cluster
4. capz tries to delete the dns zone created in step 2, but fails because there are some active links.

CAPZ should support byo dns zone

**Environment:**

sonasingh46:
**What type of PR is this?**

/kind feature

This PR tags private DNS zone created by CAPZ and delete
only if it has been created by CAPZ.

Fixes #1749 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
None
```

### DIFF
--- a/azure/services/privatedns/client.go
+++ b/azure/services/privatedns/client.go
@@ -28,8 +28,10 @@ import (
 
 // Client wraps go-sdk.
 type client interface {
+	GetZone(context.Context, string, string) (privatedns.PrivateZone, error)
 	CreateOrUpdateZone(context.Context, string, string, privatedns.PrivateZone) error
 	DeleteZone(context.Context, string, string) error
+	GetLink(context.Context, string, string, string) (privatedns.VirtualNetworkLink, error)
 	CreateOrUpdateLink(context.Context, string, string, string, privatedns.VirtualNetworkLink) error
 	DeleteLink(context.Context, string, string, string) error
 	CreateOrUpdateRecordSet(context.Context, string, string, privatedns.RecordType, string, privatedns.RecordSet) error
@@ -74,11 +76,21 @@ func newRecordSetsClient(subscriptionID string, baseURI string, authorizer autor
 	return recordsClient
 }
 
+// GetZone returns a private zone.
+func (ac *azureClient) GetZone(ctx context.Context, resourceGroupName, zoneName string) (privatedns.PrivateZone, error) {
+	ctx, _, done := tele.StartSpanWithLogger(ctx, "privatedns.AzureClient.GetZone")
+	defer done()
+	zone, err := ac.privatezones.Get(ctx, resourceGroupName, zoneName)
+	if err != nil {
+		return privatedns.PrivateZone{}, err
+	}
+	return zone, nil
+}
+
 // CreateOrUpdateZone creates or updates a private zone.
 func (ac *azureClient) CreateOrUpdateZone(ctx context.Context, resourceGroupName string, zoneName string, zone privatedns.PrivateZone) error {
 	ctx, _, done := tele.StartSpanWithLogger(ctx, "privatedns.AzureClient.CreateOrUpdateZone")
 	defer done()
-
 	future, err := ac.privatezones.CreateOrUpdate(ctx, resourceGroupName, zoneName, zone, "", "")
 	if err != nil {
 		return err
@@ -106,6 +118,17 @@ func (ac *azureClient) DeleteZone(ctx context.Context, resourceGroupName, name s
 	}
 	_, err = future.Result(ac.privatezones)
 	return err
+}
+
+// GetLink returns a vnet link.
+func (ac *azureClient) GetLink(ctx context.Context, resourceGroupName, zoneName, vnetLinkName string) (privatedns.VirtualNetworkLink, error) {
+	ctx, _, done := tele.StartSpanWithLogger(ctx, "privatedns.AzureClient.GetLink")
+	defer done()
+	vnetLink, err := ac.vnetlinks.Get(ctx, resourceGroupName, zoneName, vnetLinkName)
+	if err != nil {
+		return privatedns.VirtualNetworkLink{}, err
+	}
+	return vnetLink, nil
 }
 
 // CreateOrUpdateLink creates or updates a virtual network link to the specified Private DNS zone.
--- a/azure/services/privatedns/privatedns.go
+++ b/azure/services/privatedns/privatedns.go
@@ -19,6 +19,8 @@ package privatedns
 import (
 	"context"
 
+	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
+
 	"github.com/Azure/azure-sdk-for-go/services/privatedns/mgmt/2018-09-01/privatedns"
 	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/pkg/errors"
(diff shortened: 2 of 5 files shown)


CecileRobertMichon (commented):
/ok-to-test

azure/services/privatedns/privatedns.go
```diff
@@ -59,7 +61,15 @@ func (s *Service) Reconcile(ctx context.Context) error {
...
-		err := s.client.CreateOrUpdateZone(ctx, s.Scope.ResourceGroup(), zoneSpec.ZoneName, privatedns.PrivateZone{Location: to.StringPtr(azure.Global)})
+		pDNS := privatedns.PrivateZone{
+			Location: to.StringPtr(azure.Global),
+			Tags: converters.TagsToMap(infrav1.Build(infrav1.BuildParams{
+				ClusterName: s.Scope.ClusterName(),
+				Lifecycle:   infrav1.ResourceLifecycleOwned,
+				Additional:  s.Scope.AdditionalTags(),
+			})),
+		}
+		err := s.client.CreateOrUpdateZone(ctx, s.Scope.ResourceGroup(), zoneSpec.ZoneName, pDNS)
```
CecileRobertMichon, line 72: Wouldn't this cause us to add the tags to any pre-existing DNS zone, including a user's BYO zone?

azure/services/privatedns/privatedns.go
```diff
@@ -59,7 +61,15 @@ func (s *Service) Reconcile(ctx context.Context) error {
...
 		if err != nil {
 			return errors.Wrapf(err, "failed to create private DNS zone %s", zoneSpec.ZoneName)
 		}
```
CecileRobertMichon, line 75: we should consider also not creating/deleting vnet links if both the DNS zone and the vnet are not managed by capz

azure/services/privatedns/privatedns.go
sonasingh46: This is good point. I was thinking on the similar lines. But when I looked at piublicip reconcile there was no such handling in the Reconcile for managed vs unmanaged. ( Not sure if it applies to that or not ) 
Ref: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/services/publicips/publicips.go#L83
But I will make the changes here so that BYO DNS does not get tagged ( and hence reconciled) by CAPZ

azure/services/privatedns/privatedns.go
sonasingh46: I was thinking, simply we should not be reconciling/deleting vnet links if the vnet is not managed. Does the management of vnet depends on wether DNS is managed or not?

sonasingh46: By the way, I have raised the commit according to above comment of mine. Let me know if this is incorrect or needs changes.

azure/services/privatedns/privatedns.go
```diff
@@ -142,3 +204,25 @@ func (s *Service) Delete(ctx context.Context) error {
 	}
 	return nil
 }
+
+// isPrivateDNSManaged returns true if the private DNS has an owned tag with the cluster name as value,
+// meaning that the DNS lifecycle is managed.
+func (s *Service) isPrivateDNSManaged(ctx context.Context, resourceGroup, zoneName string) (bool, error) {
+	gotZone, err := s.client.GetZone(ctx, resourceGroup, zoneName)
```
shysank, line 211: nit
```suggestion
	zone, err := s.client.GetZone(ctx, resourceGroup, zoneName)
```

azure/services/privatedns/privatedns.go
```diff
@@ -142,3 +204,25 @@ func (s *Service) Delete(ctx context.Context) error {
...
+		return false, err
+	}
+	tags := converters.MapToTags(gotZone.Tags)
+	return tags.HasOwned(s.Scope.ClusterName()), nil
+}
+
+// isVnetLinkManaged returns true if the vnet link has an owned tag with the cluster name as value,
+// meaning that the vnet link lifecycle is managed.
+func (s *Service) isVnetLinkManaged(ctx context.Context, resourceGroupName, zoneName, vnetLinkName string) (bool, error) {
+	gotZone, err := s.client.GetLink(ctx, resourceGroupName, zoneName, vnetLinkName)
```
shysank, line 222:
```suggestion
	vnetLink, err := s.client.GetLink(ctx, resourceGroupName, zoneName, vnetLinkName)
```

azure/services/privatedns/client.go
```diff
@@ -108,6 +120,17 @@ func (ac *azureClient) DeleteZone(ctx context.Context, resourceGroupName, name s
 	return err
 }
 
+// GetLink returns a vnet link.
+func (ac *azureClient) GetLink(ctx context.Context, resourceGroupName, zoneName, vnetLinkName string) (privatedns.VirtualNetworkLink, error) {
+	ctx, _, done := tele.StartSpanWithLogger(ctx, "privatedns.AzureClient.GetZone")
```
shysank, line 125:
```suggestion
	ctx, _, done := tele.StartSpanWithLogger(ctx, "privatedns.AzureClient.GetLink")
```

azure/services/privatedns/privatedns.go
```diff
@@ -57,15 +59,51 @@ func (s *Service) Reconcile(ctx context.Context) error {
...
+			isManaged = true
+		}
+		if !isManaged {
+			s.Scope.V(2).Info("Skipping reconciliation of unmanaged private DNS zone", "private DNS", zoneSpec.ZoneName)
+			return nil
+		}
 		// Create the private DNS zone.
 		s.Scope.V(2).Info("creating private DNS zone", "private dns zone", zoneSpec.ZoneName)
-		err := s.client.CreateOrUpdateZone(ctx, s.Scope.ResourceGroup(), zoneSpec.ZoneName, privatedns.PrivateZone{Location: to.StringPtr(azure.Global)})
+		pDNS := privatedns.PrivateZone{
```
shysank, line 78: Maybe not part of this pr, but should we call `CreateOrUpdateZone` if there is already a zone existing since we don't allow changes to any of its properties?

azure/services/privatedns/privatedns.go
CecileRobertMichon: Interesting, we should verify but I think that might be a bug. Also coincidentally would be fixed by https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1745/files#diff-101a2cb7fa509c5208b45ff64bc92af67c470f1a3de730a5a05f60b95b7b1681R58.

CecileRobertMichon: BTW the only issue with not tagging existing DNS zones is that we won't tag the ones created by earlier versions of CAPZ and thus it would cause the old DNS zones to be treated as BYO and not be deleted with this new code.

CecileRobertMichon: added a comment in https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1692

azure/services/privatedns/privatedns.go
sonasingh46: Opened up an issue to keep track of this one. 
https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1872

azure/services/privatedns/privatedns.go
sonasingh46: >BTW the only issue with not tagging existing DNS zones is that we won't tag the ones created by earlier versions of CAPZ and thus it would cause the old DNS zones to be treated as BYO and not be deleted with this new code.

Shall we cover this as part of upgrade? One of the things that I can think top off my head is, before upgrading CAPZ users can annotate DNS to adopt it for reconcile. May be we can automate some of this process but there has to be an user intervention unless the DNS that is NOT a BYO has some footprint that differentiates to that of DNS that is a BYO.

sonasingh46: Infact, IMO we can make a generic annotation key to solve similar issue for other resources or in general if we at some point of time want to adopt resources that are not reconciled or BYO 

What do you think?

shysank: >BTW the only issue with not tagging existing DNS zones is that we won't tag the ones created by earlier versions of CAPZ and thus it would cause the old DNS zones to be treated as BYO and not be deleted with this new code.

This would only affect a subset of clusters where resource group is also not managed. Perhaps, we can add some logging for those scenarios to nudge the user in the right direction?

sonasingh46: Sure. I am already logging for not managed DNS.

CecileRobertMichon: let's change that log to v(1) so it's printed even if the user doesn't change verbosity and add a little blurb about adding the tag to nudge the user to add the tag if they didn't create the zone themselves? Maybe we can even point them to a link in docs if we document this feature

CecileRobertMichon:
https://capz.sigs.k8s.io/topics/custom-dns.html would be a good place to document the feature

CecileRobertMichon:
@sonasingh46 please squash commits

CecileRobertMichon:
@sonasingh46 can you please rebase when you get a chance?

sonasingh46:
@CecileRobertMichon -- Done rebase.

CecileRobertMichon (commented):
/lgtm
/approve

PR #1800, opened 2021-10-27
fix managed cluster diff so aks added fields are omitted

devigned:
**What type of PR is this?**
/kind bug

When an Azure Managed Cluster reconciles, will get into a never ending update loop trying to update the managed cluster due to the spec not matching the AKS computed fields for SKU and LoadBalancer related fields.

This PR does the following:
- defaults the SKU if not specified in the spec
- Sets the existing LB properties for the managed cluster to nil if the spec generated properties are nil
- Sets the EffectiveOutboundIPs field to nil for the managed cluster if spec generated LB properties are not nil
- Adds support for transient error handling in CP and MP controllers to backoff for known transient failure states.
- AMMPs reconcile only after the CP is initialized since CP creation includes the initial specs for the AMMPs, this does not delay creation of those resources, but rather, waits until they first exist before checking state.

I don't particularly like this solution as it is brittle in the face of new computed defaults from the AKS side. I would much rather see a more robust solution where both the spec generated managed cluster and the fetched managed cluster are normalized, then diffed to ensure we are only comparing the properties that should trigger an update.

/cc @richardchen331 @alexeldeib 

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
fix managed cluster endless reconcile loop due to a differences in AKS computed fields and spec generated fields
```


CecileRobertMichon:
@devigned it's tough to merge this with confidence with a failing AKS test. Do you have a good understanding of what other stuff is remaining in order to restore the test signal?

devigned:
> @devigned it's tough to merge this with confidence with a failing AKS test. Do you have a good understanding of what other stuff is remaining in order to restore the test signal?

I need to explore a bit more. The change in this PR is definitely needed, but is not the whole story. The main problem is that the AzureManagedMachinePool (agent pools) are unable to find the VMSS associated, so the providerIDs for the VM instances are never populated. @alexeldeib brought up that he thought it may be Azure Resource Manager related, but I can't say for sure. 

There is no rush here. Let me take a bit more time to get to the root of the VMSS problem, and get a solid green signal.

devigned:
Ok! I think I have one of the issues:
```
The VM size of AgentPoolProfile:agentpool1,AgentPoolProfile:agentpool0 is not allowed in your subscription in location 'northcentralus'. For more details, please visit https://aka.ms/cpu-quota
```

When running AKS test in `northcentralus`.

devigned:
I think we have two issues going on. AKS K8s versions being deprecated as well as VM SKU capacity.

/cc @alexeldeib

devigned:
pull-cluster-api-provider-azure-e2e-exp passed after changing the second agent pool to be a system pool. Retrying to see if it was just luck.

devigned:
pull-cluster-api-provider-azure-e2e-exp passed again after changing the second agent pool to be a system pool. Retrying one more time.

devigned:

Failed, but was after provisioning. This is the third provision success in a row of pull-cluster-api-provider-azure-e2e-exp.

devigned:
### API Diff
sigs.k8s.io/cluster-api-provider-azure/exp/api/v1beta1
  Incompatible changes:
  - SKU.Tier: changed from string to AzureManagedControlPlaneSkuTier
  Compatible changes:
  - AzureManagedControlPlaneSkuTier: added
  - FreeManagedControlPlaneTier: added
  - PaidManagedControlPlaneTier: added

sigs.k8s.io/cluster-api-provider-azure/exp/controllers
  Incompatible changes:
  - IsAgentPoolVMSSNotFoundError: removed

devigned:
Update: After debugging with @CecileRobertMichon and @jackfrancis we have determined the VMSSs created by the AKS cluster exist in the nodeResourceGroup, but the API is responding with an empty list. We are determining the root cause of this unexpected API response.

LochanRn:
> Update: After debugging with @CecileRobertMichon and @jackfrancis we have determined the VMSSs created by the AKS cluster exist in the nodeResourceGroup, but the API is responding with an empty list. We are determining the root cause of this unexpected API response.
> 
> /hold

Is this only with all the type of nodepools ? or just user nodepools ?

devigned:
> Is this only with all the type of nodepools ? or just user nodepools ?

I don't think it has anything to do with the type of node pools. We have reached out to msft teams responsible for this API, so  I should have a definitive answer soon.

jackfrancis:
@devigned given that we have identified the source of test failures are we able to temporarily reduce regional coverage to those that are more likely to overcome these errors, and then move the changes in this PR forward?

In other words, these changes are not exclusively intended to "fix" test failures, right? I think we should remove the various debug changes that were added to help debug, restore this PR to its state of intended change, and retest the "exp" job until we get a passing test. (Like, let's not block these changes on Azure platform errors that are largely out of our control.)

jackfrancis (approved):
lgtm

CecileRobertMichon:
@devigned is this ready or did you want to run a couple more retests?

CecileRobertMichon:
actually

/hold

I'll let @devigned remove the hold when we're good from a testing standpoint

devigned:

give it one more just for good measure.

PR #1812, opened 2021-10-29
make all logging consistent across the controllers

### LINKED ISSUE #1806: Consistent Logging
/kind feature

I've found there are 3 ways we are currently logging.
1) directly via klog
2) via the logger on scopes
3) via the span logger

We should have a single consistent pattern for logging to improve the consistency of the code, but more so, to improve the logging experience for the operator of CAPZ.

I'd like to see that we consistently use the span logger, so that spans will be sent to both STDOUT, but also as events on our distributed traces spans. The span logger is directly tied to the context. In funcs where there is no context, we should provide the logger or the context as a argument.

devigned:
**What type of PR is this?**
/kind cleanup

This PR cleans up all of the logging inconsistencies and logs everything to the span / composite logger. This PR removes all of the Logger fields and Logger embedded interfaces in scopes, reducing the scope interface footprint.

All log events are added to the appropriate spans in the distributed traces.

**Which issue(s) this PR fixes**: Fixes #1806 

This is a terribly large PR touching almost all of the code base. I'm opening this to illustrate the scope of the changes and would be happy to break it up in whatever way makes the most sense. Please suggest.

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
make all logging consistent across the controller
```


nader-ziada:
nice cleanup 👍  

the tests are failing cause of this error:
```
controllers/azurecluster_controller.go:75:2: ineffectual assignment to ctx (ineffassign)
	ctx, log, done := tele.StartSpanWithLogger(ctx,
	^
controllers/azureidentity_controller.go:56:2: ineffectual assignment to ctx (ineffassign)
	ctx, log, done := tele.StartSpanWithLogger(ctx,
```

azure/services/scalesets/mock_scalesets/scalesets_mock.go
```diff
@@ -268,18 +236,18 @@ func (mr *MockScaleSetScopeMockRecorder) GetLongRunningOperationState(arg0, arg1
 }
 
 // GetVMImage mocks base method.
-func (m *MockScaleSetScope) GetVMImage() (*v1beta1.Image, error) {
+func (m *MockScaleSetScope) GetVMImage(arg0 context.Context) (*v1beta1.Image, error) {
```
mboersma, line 239: Why not `ctx`?

controllers/azureidentity_controller.go
```diff
@@ -99,12 +102,11 @@ func (r *AzureIdentityReconciler) SetupWithManager(ctx context.Context, mgr ctrl
...
 	ctx, cancel := context.WithTimeout(ctx, reconciler.DefaultedLoopTimeout(r.ReconcileTimeout))
 	defer cancel()
-	log := r.Log.WithValues("namespace", req.Namespace, "identityOwner", req.Name)
 
-	ctx, _, done := tele.StartSpanWithLogger(ctx, "controllers.AzureIdentityReconciler.Reconcile",
+	ctx, log, done := tele.StartSpanWithLogger(ctx, "controllers.AzureIdentityReconciler.Reconcile",
 		tele.KVP("namespace", req.Namespace),
 		tele.KVP("name", req.Name),
-		tele.KVP("kind", "AzureCluster"),
+		tele.KVP("kind", "AzureIdentity"),
```
mboersma, line 109: Fixing some telemetry metadata?

exp/controllers/azuremachinepoolmachine_controller_test.go
```diff
@@ -81,6 +80,7 @@ func TestAzureMachinePoolMachineReconciler_Reconcile(t *testing.T) {
 	os.Setenv(auth.TenantID, "fooTenant")
 
 	for _, c := range cases {
+		c := c
```
mboersma, line 83: 👍🏻  Nice catch there.

exp/controllers/common_controller_test.go
```diff
@@ -1,109 +0,0 @@
-/*
```
mboersma, line 1: Are we ceding any important coverage by deleting this (and other) logging-related unit tests?

util/tele/composite_logger.go
```diff
@@ -35,7 +35,11 @@ func (c *compositeLogger) Enabled() bool {
 
 func (c *compositeLogger) iter(fn func(l logr.Logger)) {
 	for _, l := range c.loggers {
-		fn(l)
+		if callDepthLogger, ok := l.(logr.CallDepthLogger); ok {
```
mboersma, line 38: You literally have to read to the end of this giant PR to find the one line of code that makes you go, "Hey, what's that do?"
You explained it to me, but maybe we should make a code comment here explaining why we try to default to the call-depth logger, 3 deep.

shysank: related to https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1827

azure/scope/cluster.go
```diff
@@ -214,7 +210,7 @@ func (s *ClusterScope) LBSpecs() []azure.LBSpec {
 
 // RouteTableSpecs returns the node route table.
 func (s *ClusterScope) RouteTableSpecs() []azure.RouteTableSpec {
-	routetables := []azure.RouteTableSpec{}
+	var routetables []azure.RouteTableSpec
```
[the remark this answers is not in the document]
devigned: They are subtly different: https://play.golang.org/p/Yd6jei06m9y. One is an empty slice, the other is nil.

azure/scope/machinepool.go
```diff
@@ -510,12 +507,12 @@ func (m *MachinePoolScope) PatchObject(ctx context.Context) error {
 
 // Close the MachineScope by updating the machine spec, machine status.
 func (m *MachinePoolScope) Close(ctx context.Context) error {
-	ctx, _, done := tele.StartSpanWithLogger(ctx, "scope.MachinePoolScope.Close")
+	ctx, log, done := tele.StartSpanWithLogger(ctx, "scope.MachinePoolScope.Close")
```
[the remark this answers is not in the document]
devigned: Yes'ish, because the `ctx` that is returned is a new `ctx` struct which may now contain a correlationID, span, and other values in the bag. We could change the name of the new context, but this ensure that the original is not used below. Only the new `ctx` would be used. Use of the original is in nearly all cases an error.

azure/services/scalesets/mock_scalesets/scalesets_mock.go
devigned: This is generated code. We don't control the argument names.

controllers/azureidentity_controller.go
devigned: Hmm... this is actually an AzureCluster. I'll change it back.

exp/controllers/common_controller_test.go
devigned: I don't think so. The tests were not super effective, and as the original author, I'm no longer sold on the fact they verify anything of value.

util/tele/composite_logger.go
devigned: I've added a comment to explain why we are using the `CallDepthLogger` interface, and what it does.

devigned:
I believe I've address all the feedback.

devigned:
I rebased these changes this morning and looks like all the tests are green. Let me know if you folks see anything that needs to change.

main.go
```diff
@@ -263,7 +262,11 @@ func main() {
 		}()
 	}
 
-	ctrl.SetLogger(klogr.New())
+	lggr := klogr.New()
+	if callDepthLogger, ok := lggr.(logr.CallDepthLogger); ok {
+		lggr = callDepthLogger.WithCallDepth(2)
```
CecileRobertMichon, line 267: why are we doing this here and why is it 2 and not 3 like in composite_logger.go?

controllers/azurecluster_controller.go
```diff
@@ -76,20 +73,23 @@ func NewAzureClusterReconciler(client client.Client, log logr.Logger, recorder r
...
+	)
 	defer done()
 
-	log := acr.Log.WithValues("controller", "AzureCluster")
 	var r reconcile.Reconciler = acr
 	if options.Cache != nil {
 		r = coalescing.NewReconciler(acr, options.Cache, log)
 	}
 
+	v6log := log.V(6)
```
CecileRobertMichon, line 87: why are we adding a v(6) here? the predicate functions are already setting different verbosity levels for logging (looks like mostly v(4) and v(6)) 

https://github.com/kubernetes-sigs/cluster-api/blob/3b839fd1c57b9155ef1ecac0380440d73d9ebd7b/util/predicates/generic_predicates.go#L189

controllers/azureidentity_controller.go
```diff
@@ -43,24 +35,36 @@ import (
...
 // SetupWithManager initializes this controller with a manager.
 func (r *AzureIdentityReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager, options controller.Options) error {
-	log := r.Log.WithValues("controller", "AzureIdentity")
+	_, log, done := tele.StartSpanWithLogger(ctx,
+		"controllers.AzureIdentityReconciler.SetupWithManager",
+		tele.KVP("controller", "AzureIdentity"),
+	)
+	defer done()
+
+	v6log := log.V(6)
```
CecileRobertMichon, line 63: same question here

controllers/azurecluster_controller.go
CecileRobertMichon: I think this might be causing the logs to be spammed with all the predicate logs https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/1812/pull-cluster-api-provider-azure-e2e/1462803034441519104/artifacts/clusters/bootstrap/controllers/capz-controller-manager/capz-controller-manager-6dd76454d5-tz2cv/manager.log

controllers/azureidentity_controller.go
```diff
@@ -43,24 +35,36 @@ import (
...
 	ReconcileTimeout time.Duration
 	WatchFilterValue string
 }
 
 // SetupWithManager initializes this controller with a manager.
 func (r *AzureIdentityReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager, options controller.Options) error {
-	log := r.Log.WithValues("controller", "AzureIdentity")
+	_, log, done := tele.StartSpanWithLogger(ctx,
+		"controllers.AzureIdentityReconciler.SetupWithManager",
+		tele.KVP("controller", "AzureIdentity"),
```
shysank, line 59: same comment as above for passing key value pairs, and for other occurrences below.

azure/services/scalesetvms/scalesetvms.go
```diff
@@ -92,7 +90,7 @@ func (s *Service) Delete(ctx context.Context) error {
 		instanceID    = s.Scope.InstanceID()
 	)
 
-	log := s.Scope.WithValues("resourceGroup", resourceGroup, "scaleset", vmssName, "instanceID", instanceID)
+	log = log.WithValues("resourceGroup", resourceGroup, "scaleset", vmssName, "instanceID", instanceID)
```
shysank, line 93: can we not plug this in `StartSpanWithLogger `? or is this intentionally done to not include the kvs in spans?

controllers/azuremachine_controller.go
```diff
@@ -76,25 +72,28 @@ func NewAzureMachineReconciler(client client.Client, log logr.Logger, recorder r
 
 // SetupWithManager initializes this controller with a manager.
 func (amr *AzureMachineReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager, options Options) error {
-	ctx, _, done := tele.StartSpanWithLogger(ctx, "controllers.AzureMachineReconciler.SetupWithManager")
+	ctx, log, done := tele.StartSpanWithLogger(ctx,
+		"controllers.AzureMachineReconciler.SetupWithManager",
```
shysank, line 76: clarification:
This logger previously had the name  created as`ctrl.Log.WithName("controllers").WithName("AzureMachine")`. Is this okay to change now? Same for other controllers as well.

azure/scope/machine.go
```diff
@@ -437,18 +429,21 @@ func (m *MachineScope) SetFailureReason(v capierrors.MachineStatusError) {
 }
 
 // SetBootstrapConditions sets the AzureMachine BootstrapSucceeded condition based on the extension provisioning states.
-func (m *MachineScope) SetBootstrapConditions(provisioningState string, extensionName string) error {
+func (m *MachineScope) SetBootstrapConditions(ctx context.Context, provisioningState string, extensionName string) error {
+	_, log, done := tele.StartSpanWithLogger(ctx, "scope.MachineScope.SetBootstrapConditions")
```
shysank, line 433: General question on the approach: All logs are now written through the span logger which also creates spans for the functions they are called in. This means we have already added spans to functions which previously didn't have one, and in future, we'll be adding more if we want to add logs. Will this create a flurry of spans and make it harder to visualize the flow?

main.go
devigned: This was not effective and has been removed.

azure/scope/machinepool.go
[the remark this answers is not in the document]
mboersma: > Use of the original is in nearly all cases an error.

Got it. Thanks for the clarification.

controllers/azurecluster_controller.go
devigned: I think you are right. I also think there was an issue with the composite logger not dealing with verbosity level appropriately. I've updated the composite logger to delegate verbosity down to sub loggers.

controllers/azureidentity_controller.go
devigned: Resolved by removing these V(6) calls and handling verbosity appropriately in the composite logger.

controllers/azuremachine_controller.go
devigned: I would assume so, but it sounds like you are concerned about something specific. What concerns you about changing these?

azure/scope/machine.go
devigned: Creating the spans and adding them to the traces in my opinion actually makes it easier to understand what is happening. If someone would like to filter spans, they are able to in the visualization or in the span data. Without producing the spans, the data would then be lost. For example, in Jaeger, I can just not expand items in the tree below a given level if I want. However, if I need to dig deeper I can. If we omit deeper span information, there is no option to dig deeper.

Additionally, exporters would likely be configured to do stochastic tracing only selecting a percentage of traces or a subset based on some criteria. 

Overall, I'd bias toward over instrumentation and giving folks the ability to reduce based on export criteria.

wdyt?

api/v1beta1/azurecluster_webhook.go
```diff
@@ -45,21 +41,16 @@ var _ webhook.Defaulter = &AzureCluster{}
...
 // Default implements webhook.Defaulter so a webhook will be registered for the type.
 func (c *AzureCluster) Default() {
-	clusterlog.Info("default", "name", c.Name)
-
 	c.setDefaults()
 }
 
 // ValidateCreate implements webhook.Validator so a webhook will be registered for the type.
 func (c *AzureCluster) ValidateCreate() error {
-	clusterlog.Info("validate create", "name", c.Name)
```
CecileRobertMichon, line 55: are we loosing any valuable information by removing these logs? would it make sense to downgrade them to v() > 1 instead?

controllers/azuremachine_controller.go
CecileRobertMichon: are we loosing info on the namespace and the name of the object? I see `logger.WithValues("namespace", req.Namespace, "azureMachine", req.Name)` was removed

api/v1beta1/azurecluster_webhook.go
devigned: I didn't see this kind of logging in CAPI, so I ripped them out. I'm happy to add them back if they add value.

controllers/azuremachine_controller.go
devigned: > are we loosing info on the namespace and the name of the object? I see logger.WithValues("namespace", req.Namespace, "azureMachine", req.Name) was removed

Indeed, that was removed, but it was added back in the tele.KVP attributes that are passed into `tele.StartSpanWithLogger`, which adds them as values to the composite logger.

CecileRobertMichon:
Given the size of the PR and the number of lines that it touches, I'm going to let it merge since this week is pretty slow and it should allow us to optimize for minimum collisions with other in-flight PRs. Most comments have been addressed and since this is not adding any features or APIs we can adjust and correct if needed. @shysank feel free to complete your review when you get back so any outstanding comments/questions you have can be addressed. 

/lgtm
/approve

devigned:

`Process did not finish before 2h0m0s timeout`

PR #1822, opened 2021-11-02
clean up kind with local registry and remove kind ingress

devigned:
**What type of PR is this?**
/kind cleanup

I recently picked up a new OSX machine and was setting up CAPZ development. I was surprised to see that when I ran `make tilt-up` there was already a service running on localhost 5000, AirTunes. Since others might run into this kind of situation, I wanted to introduce a env var to help out with configuring the port the kind local registry runs on.

Since I was already in the kind setup script, I also took a little time to reduce the things going on in the script to the bare minimum. I removed the ingress setup that was used for metrics and tracing. The Tilt link for metrics and tracing works without ingress, so it was no longer needed.

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
None
```

### DIFF
--- a/scripts/kind-with-registry.sh
+++ b/scripts/kind-with-registry.sh
@@ -27,32 +27,20 @@ make "${KUBECTL}" &>/dev/null
 
 # desired cluster name; default is "kind"
 KIND_CLUSTER_NAME="${KIND_CLUSTER_NAME:-capz}"
-export KIND_EXPERIMENTAL_DOCKER_NETWORK="bridge"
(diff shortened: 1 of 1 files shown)


mboersma (approved):
/lgtm

Much simpler. I tested this locally and didn't see any problems using the registry it creates with `make tilt-up`.

CecileRobertMichon (commented):
/approve

PR #1831, opened 2021-11-04
fix(windows-calico-containerd): make envsubst working

lippertmarkus:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**: Fix envsubst
envsubst was not working before as it can't replace `${KUBERNETES_VERSION/+/_}` and also accidently did replace the `$env:CONTAINER_SANDBOX_MOUNT_POINT` variables in the container arguments. Example:
```bash
$ export KUBERNETES_VERSION=v1.22.3

# old behaviour
$ echo "aaaaa \${KUBERNETES_VERSION/+/_} aaaa \$env:CONTAINER_SANDBOX_MOUNT_POINT aaa" | envsubst
aaaaa ${KUBERNETES_VERSION/+/_} aaaa :CONTAINER_SANDBOX_MOUNT_POINT aaa

# new behaviour with fix
$ echo "aaaaa \${KUBERNETES_VERSION} aaaa \$env:CONTAINER_SANDBOX_MOUNT_POINT aaa" | envsubst \$KUBERNETES_VERSION
aaaaa v1.22.3 aaaa $env:CONTAINER_SANDBOX_MOUNT_POINT aaa
```

Fixes #

-

**TODOs**:

- [x] squashed commits
- [x] includes documentation - not needed
- [x] adds unit tests - not needed

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -220,8 +220,8 @@ def create_crs():
 
     # need to set version for kube-proxy on windows.
     os.putenv("KUBERNETES_VERSION", settings.get("kubernetes_version", {}))
-    local("kubectl create configmap flannel-windows-addon --from-file=templates/addons/windows/flannel/ --dry-run=client -o yaml | " + envsubst_cmd + " | kubectl apply -f -")
-    local("kubectl create configmap calico-windows-addon --from-file=templates/addons/windows/calico/ --dry-run=client -o yaml | " + envsubst_cmd + " | kubectl apply -f -")
+    local("kubectl create configmap flannel-windows-addon --from-file=templates/addons/windows/flannel/ --dry-run=client -o yaml | " + envsubst_cmd + " \$KUBERNETES_VERSION | kubectl apply -f -")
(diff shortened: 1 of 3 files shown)


Tiltfile
```diff
@@ -220,8 +220,8 @@ def create_crs():
 
     # need to set version for kube-proxy on windows.
     os.putenv("KUBERNETES_VERSION", settings.get("kubernetes_version", {}))
```
lippertmarkus, line 222: @jsturtevant maybe we can replace `+` with `_` here instead?

jsturtevant:
@lippertmarkus I think this is working as expected.  cluster-api uses https://github.com/drone/envsubst in its tooling which seems to have a slightly different behavoir.  If I do the following (which is what the tiltfile and makefiles use) it works

```
echo "aaaaa \${KUBERNETES_VERSION/+/_} aaaa \$env:CONTAINER_SANDBOX_MOUNT_POINT aaa" | hack/tools/bin/envsubst
aaaaa v1.22.3 aaaa $env:CONTAINER_SANDBOX_MOUNT_POINT aaa
```

lippertmarkus:
got it, thought it was using the standard envsubst. Thanks for clarifying.

PR #1837, opened 2021-11-05
Check future regardless of error in async helpers and return VM result for existing VM

CecileRobertMichon:
**What type of PR is this?**

**What this PR does / why we need it**: Small followups on #1697 and comments in https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1684

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```


azure/services/async/async.go
```diff
@@ -85,16 +85,14 @@ func CreateResource(ctx context.Context, scope FutureScope, client Creator, spec
...
 	result, sdkFuture, err := client.CreateOrUpdateAsync(ctx, spec)
-	if err != nil {
-		if sdkFuture != nil {
-			future, err := converters.SDKToFuture(sdkFuture, infrav1.PutFuture, serviceName, resourceName, rgName)
-			if err != nil {
-				return nil, errors.Wrapf(err, "failed to create resource %s/%s (service: %s)", rgName, resourceName, serviceName)
-			}
-			scope.SetLongRunningOperationState(future)
-			return nil, azure.WithTransientError(azure.NewOperationNotDoneError(future), retryAfter(sdkFuture))
+	if sdkFuture != nil {
```
CecileRobertMichon, line 88: @karuppiah7890 please take a look. I believe this addresses your comment in https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1684#discussion_r719192451. I think it makes sense to not require an error (that way this will work whether an error is returned or not). However, I still think we shouldn't silence the error returned by `WaitForCompletionRef` in the clients since eventually it might return a different error besides context deadline exceeded.

azure/services/virtualmachines/client.go
```diff
@@ -78,12 +78,25 @@ func (ac *AzureClient) CreateOrUpdateAsync(ctx context.Context, spec azure.Resou
...
-	} else if vm == nil {
-		// nothing to do here.
-		return nil, nil, nil
+	}
+
+	vm, ok := params.(compute.VirtualMachine)
+	if !ok {
+		if params == nil {
+			// nothing to do here.
+			return existingVM, nil, nil
```
CecileRobertMichon, line 97: this addresses https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1697#discussion_r742238447

CecileRobertMichon:
/hold

need to investigate why e2e tests are failing

azure/services/virtualmachines/client.go
```diff
@@ -78,15 +78,29 @@ func (ac *AzureClient) CreateOrUpdateAsync(ctx context.Context, spec azure.Resou
 	ctx, _, done := tele.StartSpanWithLogger(ctx, "virtualmachines.AzureClient.CreateOrUpdate")
 	defer done()
 
-	vm, err := ac.vmParams(ctx, spec)
+	var existingVM interface{}
+
+	if existing, err := ac.Get(ctx, spec.ResourceGroupName(), spec.ResourceName()); err != nil && !azure.ResourceNotFound(err) {
+		return nil, nil, errors.Wrapf(err, "failed to get virtual machine %s in %s", spec.ResourceName(), spec.ResourceGroupName())
```
shysank, line 84: shouldn't we handle errors that are not `ResourceNotFound`?

azure/services/virtualmachines/client.go
shysank: can we move this to line 86 and invoke spec.Parameters() can be only for new vms? wdyt?

azure/services/virtualmachines/client.go
CecileRobertMichon: if it's a `ResourceNotFound` error, that means the VM doesn't exist, so we don't set existingVM. It's not an error we want to return, it's an expected error whenever we create the resource for the first time.

azure/services/virtualmachines/client.go
CecileRobertMichon: in this specific VM service we technically could because we never want to update existing VMs but this pattern works best for all services in general: we pass in the existing resource to `spec.Parameters(existing)` and that function determines if the resource needs to be created or updated based on the existing resource. For example, for NSGs we update the group the security rules are missing. IMO it's best to use the same pattern throughout all services so it's easier to refactor them later to have even more code in common.

azure/services/virtualmachines/client.go
shysank: oh yeah, this used to be in a different form in `virtualmachines.go`. Perhaps I should be more in sync with the new code, but alas the code itself is not in sync anymore.

devigned (approved):
/lgtm
/approve

PR #1842, opened 2021-11-08
Make CAPZ tooling work on arm64

### LINKED ISSUE #1836: Support for darwin/arm64 development
/kind bug

When trying to get started on a new arm64 OSX dev machine, I ran into some issues fetching envtest binaries and some other little oddities.

Most of the amd64 stuff works fine, but there are some edge cases where we can do a bit better.

See: https://github.com/kubernetes-sigs/cluster-api/pull/5547

darwin/arm64 has a native dev experience where as many of the dev tools are compiled targeting arm64

mboersma:
**What type of PR is this?**

/kind bug

Removes some hard-coding of `amd64` to allow dev tooling to work on all platforms that Go supports.

See kubernetes-sigs/cluster-api#5547

Fixes #1836

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -107,7 +107,7 @@ PROD_REGISTRY := us.gcr.io/k8s-artifacts-prod/cluster-api-azure
 IMAGE_NAME ?= cluster-api-azure-controller
 CONTROLLER_IMG ?= $(REGISTRY)/$(IMAGE_NAME)
 TAG ?= dev
-ARCH ?= amd64
+ARCH ?= $(GOARCH)
 ALL_ARCH = amd64 arm arm64 ppc64le s390x
 
 # Allow overriding manifest generation destination directory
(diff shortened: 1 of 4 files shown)


CecileRobertMichon:
Does this fix #1836?

/assign @devigned to make sure this works on arm64

mboersma:
/retest

AKS cluster failed to provision.

devigned (approved):
I think many of these ensure files can be replaced with `go install`. Also, they are adding the bin to the Go bin path, rather than `./hack/tools/bin`, which is something we have tried to get away from.

I don't think the above commentary should block this PR. This is still better than the current situation.

/lgtm
/approve

PR #1013, opened 2020-10-28
Cleanup disks service

shysank:
**What type of PR is this?**
/kind cleanup

**What this PR does / why we need it**:
Cleanup disks service
- Make `Client` interface private ( the methods are still exported to generate mocks)
- Move `DiskScope` and `Service` struct to `disks.go`
- `NewService()` -> `New()` 
- Remove redundant initialization code in `disks_test.go`

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #1001 (Partially)

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Cleanup disks service
```

PR #1844, opened 2021-11-09
Make disks delete async

### LINKED ISSUE #1711: async disks

### LINKED ISSUE #1829: async vnet peerings

Jont828:
**What type of PR is this?**

/kind feature
**What this PR does / why we need it**: Implementation of an async service for disks as a follow up for #1610 and #1541.

Fixes #1711 

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Make disks delete async
```

### DIFF
--- a/api/v1beta1/conditions_consts.go
+++ b/api/v1beta1/conditions_consts.go
@@ -106,6 +106,8 @@ const (
 	AvailabilitySetReadyCondition clusterv1.ConditionType = "AvailabilitySetReady"
 	// RoleAssignmentReadyCondition means the role assignment exists and is ready to be used.
 	RoleAssignmentReadyCondition clusterv1.ConditionType = "RoleAssignmentReady"
+	// DisksReadyCondition means the disks exist and are ready to be used.
+	DisksReadyCondition clusterv1.ConditionType = "DisksReady"
 
 	// CreatingReason means the resource is being created.
 	CreatingReason = "Creating"
--- a/azure/scope/cluster.go
+++ b/azure/scope/cluster.go
@@ -600,6 +600,7 @@ func (s *ClusterScope) PatchObject(ctx context.Context) error {
 			infrav1.ResourceGroupReadyCondition,
 			infrav1.NetworkInfrastructureReadyCondition,
 			infrav1.VnetPeeringReadyCondition,
+			infrav1.DisksReadyCondition,
 		),
 	)
 
@@ -611,6 +612,7 @@ func (s *ClusterScope) PatchObject(ctx context.Context) error {
 			infrav1.ResourceGroupReadyCondition,
 			infrav1.NetworkInfrastructureReadyCondition,
 			infrav1.VnetPeeringReadyCondition,
+			infrav1.DisksReadyCondition,
 		}})
 }
 
--- a/azure/services/disks/client.go
+++ b/azure/services/disks/client.go
@@ -21,48 +21,81 @@ import (
 
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2021-04-01/compute"
 	"github.com/Azure/go-autorest/autorest"
+	azureautorest "github.com/Azure/go-autorest/autorest/azure"
+	"github.com/pkg/errors"
+
 	"sigs.k8s.io/cluster-api-provider-azure/azure"
+	"sigs.k8s.io/cluster-api-provider-azure/util/reconciler"
 	"sigs.k8s.io/cluster-api-provider-azure/util/tele"
 )
 
-// Client wraps go-sdk.
+// client wraps go-sdk.
 type client interface {
-	Delete(context.Context, string, string) error
+	DeleteAsync(context.Context, azure.ResourceSpecGetter) (azureautorest.FutureAPI, error)
+	Result(context.Context, azureautorest.FutureAPI, string) (interface{}, error)
+	IsDone(context.Context, azureautorest.FutureAPI) (bool, error)
 }
 
-// AzureClient contains the Azure go-sdk Client.
+// azureClient contains the Azure go-sdk Client.
 type azureClient struct {
 	disks compute.DisksClient
 }
 
 var _ client = (*azureClient)(nil)
 
-// newClient creates a new VM client from subscription ID.
+// newClient creates a new VM Client from subscription ID.
 func newClient(auth azure.Authorizer) *azureClient {
-	c := newDisksClient(auth.SubscriptionID(), auth.BaseURI(), auth.Authorizer())
+	c := NewDisksClient(auth.SubscriptionID(), auth.BaseURI(), auth.Authorizer())
 	return &azureClient{c}
 }
 
-// newDisksClient creates a new disks client from subscription ID.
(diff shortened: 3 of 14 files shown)


api/v1beta1/conditions_consts.go
```diff
@@ -106,6 +106,8 @@ const (
 	AvailabilitySetReadyCondition clusterv1.ConditionType = "AvailabilitySetReady"
 	// RoleAssignmentReadyCondition means the role assignment exists and is ready to be used.
 	RoleAssignmentReadyCondition clusterv1.ConditionType = "RoleAssignmentReady"
+	// DisksReadyCondition means the disks exist and are ready to be used.
+	DisksReadyCondition clusterv1.ConditionType = "DisksReady"
```
CecileRobertMichon, line 110: This is an interesting one: would it ever become true if we're never actually creating the disks?

azure/services/disks/spec.go
```diff
@@ -0,0 +1,40 @@
+/*
```
CecileRobertMichon, line 1: verify test is failing because this header doesn't match the expected header

> Boilerplate header is wrong for: /home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/azure/services/disks/spec.go

azure/scope/cluster.go
```diff
@@ -600,6 +600,7 @@ func (s *ClusterScope) PatchObject(ctx context.Context) error {
 			infrav1.ResourceGroupReadyCondition,
 			infrav1.NetworkInfrastructureReadyCondition,
 			infrav1.VnetPeeringReadyCondition,
+			// TODO: add infrav1.DisksReadyCondition,
```
CecileRobertMichon, line 603: #1868 might help with this

azure/services/disks/spec.go
```diff
@@ -0,0 +1,43 @@
...
+func (s *DiskSpec) ResourceName() string {
+	return s.Name
+}
+
+// ResourceGroupName returns the name of the resource group.
+func (s *DiskSpec) ResourceGroupName() string {
+	return s.ResourceGroup
+}
+
+// OwnerResourceName is a no-op for virtual network peerings.
```
Jont828, line 35: TODO: fix the spec names

Jont828:
@CecileRobertMichon Do you think it would make sense to add `UpdatePutStatus(infrav1.DisksReadyCondition, serviceName, err)` to the reconciler for async VMs?

CecileRobertMichon:
@Jont828 please squash commits

azure/services/disks/disks.go
```diff
@@ -52,26 +57,31 @@ func (s *Service) Reconcile(ctx context.Context) error {
 	_, _, done := tele.StartSpanWithLogger(ctx, "disks.Service.Reconcile")
 	defer done()
 
+	// TODO: what happens if we don't update the put status for the ready condition?
```
CecileRobertMichon, line 60: remove this?

azure/services/disks/disks.go
```diff
@@ -52,26 +57,31 @@ func (s *Service) Reconcile(ctx context.Context) error {
...
+
 	return nil
 }
 
 // Delete deletes the disk associated with a VM.
 func (s *Service) Delete(ctx context.Context) error {
-	ctx, _, done := tele.StartSpanWithLogger(ctx, "disks.Service.Delete")
+	_, _, done := tele.StartSpanWithLogger(ctx, "disks.Service.Delete")
```
CecileRobertMichon, line 67: why not set ctx here? see #1806 for how we should be handling ctx and logging

azure/services/disks/spec.go
```diff
@@ -0,0 +1,43 @@
...
+func (s *DiskSpec) ResourceGroupName() string {
+	return s.ResourceGroup
+}
+
+// OwnerResourceName is a no-op for disks.
+func (s *DiskSpec) OwnerResourceName() string {
+	return ""
+}
+
+// Parameters returns the parameters for the disk.
```
CecileRobertMichon, line 40: consider adding a comment here to highlight that it's a no-op for disks

azure/services/virtualmachines/virtualmachines.go
```diff
@@ -83,6 +83,7 @@ func (s *Service) Reconcile(ctx context.Context) error {
 
 	result, err := async.CreateResource(ctx, s.Scope, s.Client, vmSpec, serviceName)
 	s.Scope.UpdatePutStatus(infrav1.VMRunningCondition, serviceName, err)
+	s.Scope.UpdatePutStatus(infrav1.DisksReadyCondition, serviceName, err)
```
CecileRobertMichon, line 86: I think we should consider only setting the Disk condition once the VM create is successful. The reason I say that is that if the overall operation fails, it's hard to know whether it's caused by the VM or the disk, and if we set the disk ready condition to `false` it could potentially shadow the VM ready condition in the overall Machine status.
Eg scenario: User creates a Machine. Azure VM provisioning fails. The AzureMachine shows `Ready` as "False" and reason as "disk failed to create". This is confusing because the disk reconcile is a no-op if they look at the code. It would be better to always show "vm failed to create" in those cases.

CecileRobertMichon: let's also add a comment to explain we set the disk condition in the VM service

CecileRobertMichon (commented):
lgtm besides two minor comments

azure/services/disks/spec.go
```diff
@@ -0,0 +1,43 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
CecileRobertMichon, line 2:
```suggestion
Copyright 2021 The Kubernetes Authors.
```

azure/services/disks/client.go
```diff
@@ -21,48 +21,81 @@ import (
...
-	Delete(context.Context, string, string) error
+type Client interface {
+	DeleteAsync(context.Context, azure.ResourceSpecGetter) (azureautorest.FutureAPI, error)
+	Result(context.Context, azureautorest.FutureAPI, string) (interface{}, error)
+	IsDone(context.Context, azureautorest.FutureAPI) (bool, error)
 }
 
 // AzureClient contains the Azure go-sdk Client.
-type azureClient struct {
+type AzureClient struct {
```
CecileRobertMichon, line 40: why change azureClient to public? only the clients that are being used in other packages should be public and it doesn't look like we are changing that requirement in this PR

ref https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1013

CecileRobertMichon (commented):
/lgtm
/approve
/retest

CecileRobertMichon:
possibly a flake?

/retest

CecileRobertMichon:

the test is going to fail, restarting to gain time. Going to look into the different failures.

CecileRobertMichon:
/retest

quota issues on the sub should be mitigated

CecileRobertMichon:
the latest one wasn't a quota issue, going to investigate separately (seems unrelated to this PR)

/retest

CecileRobertMichon:
this PR does not want to merge...

Issues we've run into so far:
- #1901 
- quota issues (resolved)
- control planes not creating/joining (looking)
- service curl flakes

/retest

PR #1845, opened 2021-11-09
Bump golangci-lint to 1.43.0

stmcginnis:
**What type of PR is this?**

/kind cleanup

This version of golangci-lint pulls in releases of its linters that
address issues when running under go 1.17. With these newer linters,
there are also better linting checks.

This bumps the golangci-lint version and addresses the one additional
issue this newer version raised.

Can be validated by running `make lint-full`

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```


controllers/helpers_test.go
```diff
@@ -14,6 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+//nolint:gosec
```
mboersma, line 17: Could you instead put this skip statement at line 405? Then the real code in the top half of this file will still be checked by `gosec`.

mboersma (changes requested):

controllers/helpers_test.go
stmcginnis: Sure thing - I was thinking this was all test code based on the name, so skipping it all. But in general it is probably better to be more specific anyway. Thanks!

mboersma (approved):
/lgtm

CecileRobertMichon (commented):
/approve

PR #1850, opened 2021-11-10
Use region specific API calls with VMSS

### LINKED ISSUE #1720: AKS test is flaky
/kind bug

AKS tests have been failing very frequently recently with error "System machine pools not ready". Example:https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/periodic-cluster-api-provider-azure-e2e-full/1440295663572094976

dashboard: https://testgrid.k8s.io/sig-cluster-lifecycle-cluster-api-provider-azure#capz-periodic-capi-e2e-full-main 

This is slowing down PRs to merge as they are unable to pass the full test suite.

**Environment:**

devigned:
**What type of PR is this?**
/kind bug

When listing VMSSes in a resource group, Azure responds with values it has in the regional cache. Unfortunately, there are sometimes spikes in regional replication for these caches. To avoid cross region replication issues, this PR introduces a region specific client for AzureManageMachinePools by transforming the client baseURI from something like `https://management.azure.com` to `https://{region}.management.azure.com` to ensure the request is directed to the region where the VMSSes should be.

fixes #1720

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
use a region specific Azure client for listing VMSS in AzureManagedClusters
```


devigned:
success 2, 22mins

devigned:
success 3, 23mins

exp/controllers/azuremanagedmachinepool_reconciler.go
```diff
@@ -146,7 +153,11 @@ func (s *azureManagedMachinePoolService) Delete(ctx context.Context) error {
...
 
-// IsAgentPoolVMSSNotFoundError returns true if the error is an AgentPoolVMSSNotFoundError.
-func IsAgentPoolVMSSNotFoundError(err error) bool {
-	return errors.Is(err, notFoundErr)
+func getRegionalBaseURI(region, baseURI string) (string, error) {
+	parsedURI, err := url.Parse(baseURI)
+	if err != nil {
+		return "", errors.Wrap(err, "failed to parse the base URI of client")
+	}
+
```
alexeldeib, line 161: Nice

devigned:
success 4, 23mins

alexeldeib:
23 min is right around what I’d expect for the total duration of this test. Looking good!

alexeldeib:
Actually, with the latest bits to reconcile pools all together on cluster create, I wonder if we can’t squeeze that lower. Delete is kind of slow though, so we’ll see what makes sense.

devigned:
5 passes, 5th 23mins

devigned:
@CecileRobertMichon the PR history is now solid green for the exp job. I think we are solid. So solid that we may want to consider adding the AKS related tests to the regular pr e2e test.

devigned:
> Actually, with the latest bits to reconcile pools all together on cluster create, I wonder if we can’t squeeze that lower. Delete is kind of slow though, so we’ll see what makes sense (unrelated to this PR)

I think with the transient error / requeue stuff that I added to the AzureManagedMachinePool reconciler, I think we are close to as good as we can get. The initial reconcile of AMMP only starts when control plane transitions to ready. Once ready, the reconciler will try to reach goal state persistently with requeues if it reaches what is deemed a transient error, a 404 for the VMSS other similar errors that will be resolved as infra comes online. I don't think this is the most elegant solution, but it seems to be resilient.

devigned:
success 7?, 22mins.

azure/regional_baseuri.go
```diff
@@ -0,0 +1,60 @@
...
+	}, nil
+}
+
+// BaseURI return a regional base URI, like `https://{region}.management.azure.com`.
+func (a *baseURIAdapter) BaseURI() string {
+	if a == nil || a.parsedURL == nil || a.Region == "" {
+		return a.aliasAuth.BaseURI()
+	}
+
+	sansScheme := path.Join(fmt.Sprintf("%s.%s", a.Region, a.parsedURL.Host), a.parsedURL.Path)
```
alexeldeib, line 58: Do we actually want path.join() (os filepath separator) here, versus a hardcoded slash? Wouldn’t this now be platform dependent?

alexeldeib: Nvm, I might be thinking filepath.join()

alexeldeib:
Before the PR which reconciled all AMMP with cluster creation, we would create the cluster, wait for it to reconcile fully, then trigger creation of nodepools which takes ~5 min. With that approach I would expect to see around 25min runs based on my previous testing.

since we don’t need that extra 5min anymore (all pools should be created), I wonder if we have room to squeeze the times lower somewhere, or if that’s just as good as it gets.

alexeldeib:
Nice work with the URL scheme/host/path parsing

/lgtm

azure/regional_baseuri.go
devigned: I think we are good: https://cs.opensource.google/go/go/+/refs/tags/go1.17.3:src/path/path.go;l=174

devigned:
> Before the PR which reconciled all AMMP with cluster creation, we would create the cluster, wait for it to reconcile fully, then trigger creation of nodepools which takes ~5 min. With that approach I would expect to see around 25min runs based on my previous testing.
> 
> since we don’t need that extra 5min anymore (all pools should be created), I wonder if we have room to squeeze the times lower somewhere, or if that’s just as good as it gets.

I like where this is going! I think we need to do some tinkering now that tests are more stable.

exp/controllers/azuremanagedmachinepool_reconciler.go
```diff
@@ -74,12 +74,21 @@ func (a *AgentPoolVMSSNotFoundError) Is(target error) bool {
 }
 
 // newAzureManagedMachinePoolService populates all the services based on input scope.
-func newAzureManagedMachinePoolService(scope *scope.ManagedControlPlaneScope) *azureManagedMachinePoolService {
+func newAzureManagedMachinePoolService(scope *scope.ManagedControlPlaneScope) (*azureManagedMachinePoolService, error) {
+	var authorizer azure.Authorizer = scope
+	if scope.ControlPlane != nil {
+		regionalAuthorizer, err := azure.WithRegionalBaseURI(scope, scope.ControlPlane.Spec.Location)
```
CecileRobertMichon, line 80: should we use `scope.Location()` instead?

exp/controllers/azuremanagedmachinepool_reconciler.go
```diff
@@ -74,12 +74,21 @@ func (a *AgentPoolVMSSNotFoundError) Is(target error) bool {
...
+		}
+		authorizer = regionalAuthorizer
+	}
+
 	return &azureManagedMachinePoolService{
 		scope:         scope,
 		agentPoolsSvc: agentpools.New(scope),
-		scaleSetsSvc:  scalesets.NewClient(scope),
-	}
+		scaleSetsSvc:  scalesets.NewClient(authorizer),
```
CecileRobertMichon, line 90: nice

CecileRobertMichon (commented):
looks great, just one comment about using the Location() func

CecileRobertMichon (commented):
/lgtm
/approve

CecileRobertMichon:
Ok that's fair @ cherry-pick bot :) 

@devigned do you think we should cherry-pick this to release-1.0?

devigned:
> @devigned do you think we should cherry-pick this to release-1.0?

/shrug

With the next release coming very soon and managed clusters being experimental, I lean toward dealing with the cherry pick merge conflict.

Issue #1828, opened 2021-11-03
Cache refresh may lead to terminal error


CecileRobertMichon:
clarification: is there a chance for the following happening: cache has not been refreshed -> errors when doing client.List? Asking because, in this case, the error would not be a terminal error.

_Originally posted by @shysank in https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1697#discussion_r742162339_

CecileRobertMichon:
I have not had a chance to work on this yet, marking it as help wanted in case someone else is able to pick it up.

The ask is to return only return a terminal error when skuCache.Get fails because the SKU doesn't exist. If the skuCache.Get fails because of a cache refresh issue, we should return a regular error, ie. ` return errors.Wrapf(err, "failed to get SKU %s in compute api", m.AzureMachine.Spec.VMSize)`.

https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/scope/machine.go#L132

/kind bug
/help

PR #1861, opened 2021-11-12
Make availability set reconcile/delete async

### LINKED ISSUE #1709: async availabilitysets

Jont828:
**What type of PR is this?**

/kind feature
**What this PR does / why we need it**: Implementation of an async service for virtual network peerings as a follow up for #1610 and #1541.

Fixes #1709 

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Make availability set reconcile/delete async
```

### DIFF
--- a/azure/scope/machine.go
+++ b/azure/scope/machine.go
@@ -23,6 +23,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2021-04-01/compute"
 	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
@@ -37,6 +38,7 @@ import (
 
 	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
 	"sigs.k8s.io/cluster-api-provider-azure/azure"
+	"sigs.k8s.io/cluster-api-provider-azure/azure/services/availabilitysets"
 	"sigs.k8s.io/cluster-api-provider-azure/azure/services/disks"
 	"sigs.k8s.io/cluster-api-provider-azure/azure/services/resourceskus"
 	"sigs.k8s.io/cluster-api-provider-azure/azure/services/virtualmachines"
@@ -94,9 +96,10 @@ type MachineScope struct {
 
 // MachineCache stores common machine information so we don't have to hit the API multiple times within the same reconcile loop.
 type MachineCache struct {
-	BootstrapData string
-	VMImage       *infrav1.Image
-	VMSKU         resourceskus.SKU
+	BootstrapData      string
+	VMImage            *infrav1.Image
+	VMSKU              resourceskus.SKU
+	availabilitySetSKU resourceskus.SKU
 }
 
 // InitMachineCache sets cached information about the machine to be used in the scope.
@@ -122,9 +125,16 @@ func (m *MachineScope) InitMachineCache(ctx context.Context) error {
 		if err != nil {
 			return err
 		}
+
 		m.cache.VMSKU, err = skuCache.Get(ctx, m.AzureMachine.Spec.VMSize, resourceskus.VirtualMachines)
 		if err != nil {
-			return azure.WithTerminalError(errors.Wrapf(err, "failed to get SKU %s in compute api", m.AzureMachine.Spec.VMSize))
+			return azure.WithTerminalError(errors.Wrapf(err, "failed to get VM SKU %s in compute api", m.AzureMachine.Spec.VMSize))
+		}
+
+		m.cache.availabilitySetSKU, err = skuCache.Get(ctx, string(compute.AvailabilitySetSkuTypesAligned), resourceskus.AvailabilitySets)
+		if err != nil {
+			// TODO: verify error message
+			return azure.WithTerminalError(errors.Wrapf(err, "failed to get availability set SKU %s in compute api", string(compute.AvailabilitySetSkuTypesAligned)))
 		}
 	}
 
@@ -361,6 +371,29 @@ func (m *MachineScope) ProviderID() string {
 		return ""
 	}
 	return parsed.String()
+}
+
+// AvailabilitySet returns the availability set for this machine if available.
+func (m *MachineScope) AvailabilitySetSpec() azure.ResourceSpecGetter {
+	availabilitySetName, ok := m.AvailabilitySet()
+	if !ok {
+		return nil
+	}
+
+	spec := &availabilitysets.AvailabilitySetSpec{
+		Name:           availabilitySetName,
+		ResourceGroup:  m.ResourceGroup(),
+		ClusterName:    m.ClusterName(),
+		Location:       m.Location(),
+		SKU:            nil,
+		AdditionalTags: m.AdditionalTags(),
+	}
+
+	if m.cache != nil {
+		spec.SKU = &m.cache.availabilitySetSKU
+	}
+
+	return spec
 }
 
 // AvailabilitySet returns the availability set for this machine if available.
--- a/azure/services/availabilitysets/availabilitysets.go
+++ b/azure/services/availabilitysets/availabilitysets.go
@@ -18,95 +18,63 @@ package availabilitysets
 
 import (
 	"context"
-	"strconv"
 
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2021-04-01/compute"
-	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/pkg/errors"
 	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
 	"sigs.k8s.io/cluster-api-provider-azure/azure"
-	"sigs.k8s.io/cluster-api-provider-azure/azure/converters"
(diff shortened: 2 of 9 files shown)


azure/scope/machine.go
```diff
@@ -129,9 +133,16 @@ func (m *MachineScope) InitMachineCache(ctx context.Context) error {
...
+
 		m.cache.VMSKU, err = skuCache.Get(ctx, m.AzureMachine.Spec.VMSize, resourceskus.VirtualMachines)
 		if err != nil {
-			return azure.WithTerminalError(errors.Wrapf(err, "failed to get SKU %s in compute api", m.AzureMachine.Spec.VMSize))
+			return azure.WithTerminalError(errors.Wrapf(err, "failed to get VM SKU %s in compute api", m.AzureMachine.Spec.VMSize))
+		}
+
+		m.cache.availabilitySetSKU, err = skuCache.Get(ctx, string(compute.AvailabilitySetSkuTypesAligned), resourceskus.AvailabilitySets)
+		if err != nil {
+			// TODO: verify error message
```
CecileRobertMichon, line 144: Related to #1828

azure/scope/machine.go
```diff
@@ -371,6 +382,51 @@ func (m *MachineScope) ProviderID() string {
 	return parsed.String()
 }
 
+// AvailabilitySet returns the availability set for this machine if available.
+func (m *MachineScope) AvailabilitySetSpec() (azure.ResourceSpecGetter, error) {
```
Jont828, line 386: @CecileRobertMichon What do you think about this function I added to get the spec? Unlike most of the spec getters, this could also return an error.

azure/scope/machine.go
```diff
@@ -371,6 +382,51 @@ func (m *MachineScope) ProviderID() string {
...
+	if !m.AvailabilitySetEnabled() {
+		// Availability set not enabled, not an error
+		return nil, nil
```
Jont828, line 389: The original function `AvailabilitySet() (string, bool)` returns the name and true/false whether the availabilty set is available for the machine. Since it's not an error, this return `nil` to keep the same behavior pattern. Is there a better way to handle this?

azure/scope/machine.go
CecileRobertMichon: I had a similar question for virtual machines spec. At the time, I ended up deciding not to return an error https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1861/files#diff-b6931ac61f37a13f66fc7836e039bf714a0553d5db4f313cc8ef31ca5f5647b9L156 because otherwise it might fail for Delete() as well, and I didn't want to block VM deletion on getting the VM sku from the cache, since delete should not be subject to validations. The other way to avoid this is to ignore the error for Delete but that's not ideal since ignoring an error could lead to bugs down the line. Instead, I think it's better to skip setting those values in the spec if the cache is nil, and in the Reconcile() / spec Parameters() function, return an error if required values are missing from the spec. Thoughts?

azure/scope/machine.go
CecileRobertMichon: why duplicate the code from the AvailabilitySet() function below and not just use it directly? ie

```
availabilitySetName, ok := s.Scope.AvailabilitySet()
if !ok {
	return nil
}

spec := &availabilitysets.AvailabilitySetSpec{
		Name: availabilitySetName,
		ResourceGroup:  m.ResourceGroup(),
		ClusterName:    m.ClusterName(),
		Location:       m.Location(),
		AdditionalTags: m.AdditionalTags(),
	}

...
```

azure/scope/machine.go
Jont828: That makes sense, I just pushed some changes. How does that look?

azure/services/availabilitysets/spec.go
```diff
@@ -0,0 +1,85 @@
...
+		// availability set already exists
+		return nil, nil
+	}
+
+	if s.FaultDomainCount == nil {
+		return nil, errors.New("unable to find required FaultDomainCount from machine cache")
+	}
+
+	asParams := compute.AvailabilitySet{
+		// TODO: should availability sets be given a name?
```
CecileRobertMichon, line 67: the name is part of the CreateOrUpdate function parameters (same with the resource group)

azure/scope/machine.go
```diff
@@ -363,6 +374,36 @@ func (m *MachineScope) ProviderID() string {
...
+		FaultDomainCount: nil,
+		AdditionalTags:   m.AdditionalTags(),
+	}
+
+	if m.cache != nil {
+		asSKU := m.cache.availabilitySetSKU
+		faultDomainCountStr, ok := asSKU.GetCapability(resourceskus.MaximumPlatformFaultDomainCount)
+		if ok {
+			faultDomainCount, err := strconv.ParseUint(faultDomainCountStr, 10, 32)
+			if err == nil {
```
CecileRobertMichon, line 398: how about just putting `SKU` in `AvailabilitySetSpec` and just doing 
```
if m.cache != nil {
    spec.SKU = m.cache.availabilitySetSKU
}
```

in here, that way we're not silencing the actual error from getting the fault domain count if there is one? 

Then in the spec.go file you can use the SKU to get the fault domain count and return an error if appropriate.

azure/services/availabilitysets/spec.go
```diff
@@ -0,0 +1,97 @@
...
+		return nil, nil
+	}
+
+	if s.SKU == nil {
+		return nil, errors.New("unable to find required availability set SKU from machine cache")
+	}
+
+	var faultDomainCount *int32
+	faultDomainCountStr, ok := s.SKU.GetCapability(resourceskus.MaximumPlatformFaultDomainCount)
+	if ok {
```
CecileRobertMichon, line 71: what if !ok ?

azure/services/availabilitysets/availabilitysets.go
```diff
@@ -18,134 +18,108 @@ package availabilitysets
...
 
-	if err != nil {
-		return errors.Wrapf(err, "failed to delete availability set %s in resource group %s", availabilitySetName, s.Scope.ResourceGroup())
-	}
-
-	log.V(2).Info("successfully delete availability set", "availability set", availabilitySetName)
-
-	return nil
+	err = s.DeleteResource(ctx, setSpec, serviceName)
+	s.Scope.UpdateDeleteStatus(infrav1.AvailabilitySetReadyCondition, serviceName, err)
```
CecileRobertMichon, line 123: it seems like there are 4 different places where we are calling `s.Scope.UpdateDeleteStatus`, can you find a way to reduce it to 1 or 2 times by calling it once when we return at the end, similar to what you did for Reconcile()?

Jont828: I just pushed some changes to only update the status at the end. Before, I structured it like that so it would be easy to read, just a linear series of `if` checks before deleting, and if any of them failed then we'd break out of it. I switched it to nested `if/else` blocks, do you think it's still intuitive to read?

azure/services/availabilitysets/availabilitysets.go
```diff
@@ -18,134 +18,98 @@ package availabilitysets
...
 type Service struct {
 	Scope AvailabilitySetScope
 	Client
+	async.Reconciler
 	resourceSKUCache *resourceskus.Cache
 }
 
 // New creates a new availability sets service.
 func New(scope AvailabilitySetScope, skuCache *resourceskus.Cache) *Service {
+	Client := NewClient(scope)
```
CecileRobertMichon, line 51:
```suggestion
	client := NewClient(scope)
```

this var doesn't need to be exported

azure/services/availabilitysets/availabilitysets.go
```diff
@@ -18,134 +18,98 @@ package availabilitysets
...
+	} else {
+		existingSet, err := s.Client.Get(ctx, setSpec)
+		if err != nil {
+			if !azure.ResourceNotFound(err) {
+				resultingErr = errors.Wrapf(err, "failed to get availability set %s in resource group %s", setSpec.ResourceName(), setSpec.ResourceGroupName())
+			}
+		} else {
+			availabilitySet, ok := existingSet.(compute.AvailabilitySet)
+			if !ok {
+				resultingErr = errors.Errorf("%T is not a compute.AvailabilitySet", existingSet)
```
CecileRobertMichon, line 101:
```suggestion
				return errors.Errorf("%T is not a compute.AvailabilitySet", existingSet)
```

azure/services/availabilitysets/spec_test.go
```diff
@@ -0,0 +1,92 @@
...
+		Name:           "test-as",
+		ResourceGroup:  "test-rg",
+		ClusterName:    "test-cluster",
+		Location:       "test-location",
+		SKU:            &resourceskus.SKU{},
+		AdditionalTags: map[string]string{},
+	}
+)
+
+func TestParameters(t *testing.T) {
```
CecileRobertMichon, line 39: thanks for adding this test 👍

azure/services/availabilitysets/availabilitysets.go
CecileRobertMichon: oh nevermind, I see you did this to call UpdateDeleteStatus even when there's an error 🤔

CecileRobertMichon (commented):
@Jont828 can you please squash the commits?

lgtm

Issue #1685, opened 2021-09-14
Mark all ID fields as READ-ONLY and standardize how we set resource IDs across services


CecileRobertMichon:
/kind feature

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]

Currently, some of the services (subnets, route tables for example) set the Azure resource ID of existing resources in the spec as part of Reconcile(). There are a couple of problems with this:
1) The ID field is part of the spec, not status, but is not user configurable. If a user were to set an ID, it would be overwritten by the controller. This field should not be in the spec or be clearly identified as read only.
2) The implementation is inconsistent across the codebase, some resources have this while others don't. Furthermore, even when the ID is set, we don't use it most of the time, instead reconstructing the ID from scratch when needed (eg. route table ID in the subnet service).
3) Modifying the spec mid reconcile can cause bugs if some of the previous fields are not respected, as seen in #1589 

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- cluster-api-provider-azure version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

CecileRobertMichon:
Interestingly it seems like the RG GET is actually timing out after 12 seconds. I wouldn't expect an Azure GET API to take 12 seconds.... cc @devigned @Jont828 

![Screen Shot 2021-12-01 at 12 12 06 PM](https://user-images.githubusercontent.com/8650424/144306833-a3a2f18f-baaa-43ac-8af7-d3f6beb83c2f.png)

Jont828:
Noting that there's a similar issue in #1865. I kept the behavior the same when refactoring but I'll come back to it when this issue gets fixed.

CecileRobertMichon:
I have not been able to work on this one. Will unassign for now in case you or someone else wants to pick it up.

/unassign
/help

aniruddha2000:
I can take this, but I need some help like, what resources I should look at or what could be my approach and some example code where there will be the ID change 😺

CecileRobertMichon:
We might want want to hold on this one until after the switch to ASO (https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/3402), @nojnhuh wdyt?

nojnhuh:
[...]

PR #1865, opened 2021-11-15
Make NAT Gateway reconcile/delete async

### LINKED ISSUE #1714: async natgateways

Jont828:
**What type of PR is this?**

/kind feature
**What this PR does / why we need it**: Implementation of an async service for NAT gateways as a follow up for #1610 and #1541.

Fixes #1714

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Make NAT gateway reconcile/delete async
```


Jont828:
/assign mboersma 
/assign shysank 

I'd appreciate feedback on the reconcile/delete logic. I'm curious about if we need to set the subnets at the end of a Reconcile() loop since I don't believe removing it would change any behavior. And btw I'm holding off on refactoring the tests until the reconcile/delete functions are at a good spot.

azure/services/natgateways/natgateways.go
```diff
@@ -55,119 +57,61 @@ func New(scope NatGatewayScope) *Service {
...
-	}, nil
+	s.Scope.UpdatePutStatus(infrav1.NATGatewaysReadyCondition, serviceName, result)
+	return result
 }
 
 // Delete deletes the nat gateway with the provided name.
 func (s *Service) Delete(ctx context.Context) error {
-	ctx, _, done := tele.StartSpanWithLogger(ctx, "natgateways.Service.Delete")
-	defer done()
+	ctx, span := tele.Tracer().Start(ctx, "vnetpeerings.Service.Delete")
```
mboersma, line 97:
```suggestion
	ctx, span := tele.Tracer().Start(ctx, "natgateways.Service.Delete")
```

azure/services/natgateways/client.go
```diff
@@ -18,19 +18,26 @@ package natgateways
 
 import (
 	"context"
+	"encoding/json"
 
-	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2019-06-01/network"
+	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2021-02-01/network"
```
mboersma, line 23: Looks like we were using `2021-02-01` everywhere else already. 👍🏻

azure/services/natgateways/client.go
```diff
@@ -55,42 +62,127 @@ func netNatGatewaysClient(subscriptionID string, baseURI string, authorizer auto
...
+	done, err := future.DoneWithContext(ctx, ac.natgateways)
+	if err != nil {
+		return false, errors.Wrap(err, "failed checking if the operation was complete")
+	}
+
+	return done, nil
+}
+
+// Result fetches the result of a long-running operation future.
+func (ac *azureClient) Result(ctx context.Context, futureData azureautorest.FutureAPI, futureType string) (interface{}, error) {
```
mboersma, line 161: Add `tele.Tracer().Start()` here?

Jont828: @CecileRobertMichon Should we be doing this for Result()? I don't think we've been adding it on the other async services.

azure/services/natgateways/natgateways.go
```diff
@@ -55,119 +57,61 @@ func New(scope NatGatewayScope) *Service {
...
+			// TODO: consider making OwnerResourceName() a no-op
+			subnetSpec := s.Scope.Subnet(natGatewaySpec.OwnerResourceName())
+			networkNatGateway, ok := natGateway.(network.NatGateway)
+			if !ok {
+				// TODO: maybe this error shouldn't take priority
+				result = errors.Errorf("created resource %T is not a network.NatGateway", natGateway)
+			}
+			// TODO: is it necessary to set the spec since it doesn't appear to change any behavior
+			subnetSpec.NatGateway.ID = *networkNatGateway.ID
+			s.Scope.SetSubnet(subnetSpec)
```
Jont828, line 87: @CecileRobertMichon @mboersma Since the NAT gateway should already exist in the subnet spec, do we need to update the ID and set the subnet spec? I don't think that'll change the behavior if we remove it, and that way it'll be easy to make OwnerResourceName() a no-op instead of giving the subnet name.

azure/services/natgateways/client.go
CecileRobertMichon: It should, it has a context. We should also add it to the others, can be a separate commit here or later in a separate PR.

azure/services/natgateways/natgateways.go
```diff
@@ -55,119 +57,61 @@ func New(scope NatGatewayScope) *Service {
...
-	}, nil
+	s.Scope.UpdatePutStatus(infrav1.NATGatewaysReadyCondition, serviceName, result)
+	return result
 }
 
 // Delete deletes the nat gateway with the provided name.
 func (s *Service) Delete(ctx context.Context) error {
-	ctx, _, done := tele.StartSpanWithLogger(ctx, "natgateways.Service.Delete")
-	defer done()
+	ctx, span := tele.Tracer().Start(ctx, "natgateways.Service.Delete")
```
CecileRobertMichon, line 97: Is there any reason why you reverted `tele.StartSpanWithLogger`?

azure/services/natgateways/client.go
CecileRobertMichon: actually it should be tele.StartSpanWithLogger

azure/services/natgateways/client.go
```diff
@@ -55,42 +62,127 @@ func netNatGatewaysClient(subscriptionID string, baseURI string, authorizer auto
...
 
-// Delete deletes the specified nat gateway.
-func (ac *azureClient) Delete(ctx context.Context, resourceGroupName, natGatewayName string) error {
-	ctx, _, done := tele.StartSpanWithLogger(ctx, "natgateways.AzureClient.Delete")
-	defer done()
+// DeleteAsync deletes a Nat Gateway asynchronously. DeleteAsync sends a DELETE
+// request to Azure and if accepted without error, the func will return a Future which can be used to track the ongoing
+// progress of the operation.
+func (ac *azureClient) DeleteAsync(ctx context.Context, spec azure.ResourceSpecGetter) (azureautorest.FutureAPI, error) {
+	ctx, span := tele.Tracer().Start(ctx, "natgateways.azureClient.Delete")
```
CecileRobertMichon, line 125: should be 
```suggestion
	ctx, _, done := tele.StartSpanWithLogger(ctx, "natgateways.AzureClient.Delete")
```

azure/services/natgateways/client.go
```diff
@@ -55,42 +62,127 @@ func netNatGatewaysClient(subscriptionID string, baseURI string, authorizer auto
...
+	defer span.End()
```
CecileRobertMichon, line 126:
```suggestion
	defer done()
```

azure/services/natgateways/client.go
CecileRobertMichon: same thing for all the others

azure/services/natgateways/spec.go
```diff
@@ -0,0 +1,105 @@
...
+					ID: to.StringPtr(azure.PublicIPID(s.SubscriptionID, s.ResourceGroupName(), s.NatGatewayIP.Name)),
+				},
+			},
+		},
+	}
+
+	return natGatewayToCreate, nil
+}
+
+func GetPublicIP(natGateway network.NatGateway) (string, error) {
```
CecileRobertMichon, line 88: an idea: we could change this to `hasPublicIP` (don't think it needs to be public, it's just used here) and pass in the publicIP name, and just return true if the nat gateway has that public IP in its `PublicIPAddresses`

azure/services/natgateways/spec.go
```diff
@@ -0,0 +1,105 @@
...
+func (s *NatGatewaySpec) ResourceGroupName() string {
+	return s.ResourceGroup
+}
+
+// OwnerResourceName returns the name of the associated subnet.
+func (s *NatGatewaySpec) OwnerResourceName() string {
+	return s.SubnetName
+}
+
+// Parameters returns the parameters for the Nat Gateway.
```
CecileRobertMichon, line 55: let's standardize the casing to `NAT gateway` (source: https://docs.microsoft.com/en-us/azure/virtual-network/nat-gateway/nat-gateway-resource) throughout code comments

azure/services/natgateways/natgateways.go
```diff
@@ -55,119 +57,61 @@ func New(scope NatGatewayScope) *Service {
...
-			NatGatewayPropertiesFormat: &network.NatGatewayPropertiesFormat{
-				PublicIPAddresses: &[]network.SubResource{
-					{
-						ID: to.StringPtr(azure.PublicIPID(s.Scope.SubscriptionID(), s.Scope.ResourceGroup(), natGatewaySpec.NatGatewayIP.Name)),
-					},
-				},
-			},
-		}
-		err = s.client.CreateOrUpdate(ctx, s.Scope.ResourceGroup(), natGatewaySpec.Name, natGatewayToCreate)
+		natGateway, err := async.CreateResource(ctx, s.Scope, s.client, natGatewaySpec, serviceName)
```
CecileRobertMichon, line 71:
```suggestion
		result, err := async.CreateResource(ctx, s.Scope, s.client, natGatewaySpec, serviceName)
```

consider renaming the error so it doesn't collide. I've used `resErr` in some other services.

azure/services/natgateways/natgateways.go
CecileRobertMichon: I agree with you it's not needed but in the spirit of not changing behavior as part of a refactor, let's keep setting the ID in case anyone was using it before / relying on it in some way.

You should still make `OwnerResourceName()` a no-op since the subnet isn't the owner resource of the NAT GW in this case.

What we could do here is add a func to the scope `SetNatGatewayID(name string,  ID string)` which ranges through the subnets and sets the ID for all the NAT gateways with a matching name to the spec. Note that one NAT GW may be used by more than one subnet so it's important to range through all of them, not just the first one we find (since we're not doing a 1:1 mapping anymore by passing in the subnet who asked for this NAT GW to be created).

azure/services/natgateways/spec.go
```diff
@@ -0,0 +1,105 @@
...
+
+	return natGatewayToCreate, nil
+}
+
+func GetPublicIP(natGateway network.NatGateway) (string, error) {
+	// We must have a non-nil, non-"empty" PublicIPAddresses
+	if !(natGateway.PublicIPAddresses != nil && len(*natGateway.PublicIPAddresses) > 0) {
+		return "", errors.New("failed to parse PublicIPAddresses")
+	}
+	// TODO: do we need to handle NatGateway resources w/ more than one public IP address?
```
CecileRobertMichon, line 93: Could you please open an issue for this? We're treating NAT gateways as if they only had 1 IP in the API but really they can have a list of associated public IPs.

azure/scope/cluster.go
```diff
@@ -230,18 +231,22 @@ func (s *ClusterScope) RouteTableSpecs() []azure.RouteTableSpec {
...
+			natGateways = append(natGateways, &natgateways.NatGatewaySpec{
+				Name:           subnet.NatGateway.Name,
+				ResourceGroup:  s.ResourceGroup(),
+				SubscriptionID: s.SubscriptionID(),
+				Location:       s.Location(),
 				NatGatewayIP: infrav1.PublicIPSpec{
 					Name: subnet.NatGateway.NatGatewayIP.Name,
 				},
-				Subnet: subnet,
+				SubnetName: subnet.Name,
```
CecileRobertMichon, line 249: if you're not using SubnetName anywhere after removing it from OwnerResourceName, this can probably be removed

azure/scope/cluster.go
```diff
@@ -230,18 +231,22 @@ func (s *ClusterScope) RouteTableSpecs() []azure.RouteTableSpec {
...
 
 // NatGatewaySpecs returns the node nat gateway.
-func (s *ClusterScope) NatGatewaySpecs() []azure.NatGatewaySpec {
-	natGateways := []azure.NatGatewaySpec{}
+func (s *ClusterScope) NatGatewaySpecs() []azure.ResourceSpecGetter {
+	natGateways := []azure.ResourceSpecGetter{}
 
 	// We ignore the control plane nat gateway, as we will always use a LB to enable egress on the control plane.
+	// TODO: do we need to use setSubnet() in natgateways.go, since it doesn't add/remove any NodeSubnets() or change IsNatGatewayEnabled()
 	for _, subnet := range s.NodeSubnets() {
```
CecileRobertMichon, line 239: if we switch to updating the ID for every subnet that has this NAT GW in its spec, we might also be able to de-dupicate the NAT gateways, and only add it to spec if it wasn't already added.

azure/services/natgateways/natgateways.go
```diff
@@ -55,119 +57,61 @@ func New(scope NatGatewayScope) *Service {
...
-			NatGatewayIP: infrav1.PublicIPSpec{
-				Name: natGatewaySpec.NatGatewayIP.Name,
-			},
+		if err == nil && result != nil {
+			// TODO: consider making OwnerResourceName() a no-op
+			subnetSpec := s.Scope.Subnet(natGatewaySpec.OwnerResourceName())
+			networkNatGateway, ok := natGateway.(network.NatGateway)
+			if !ok {
+				// TODO: maybe this error shouldn't take priority
+				result = errors.Errorf("created resource %T is not a network.NatGateway", natGateway)
```
CecileRobertMichon, line 83: In theory, this error should never happen because it should always return a NAT gateway type. That being said, if it did happen, we would want it to interrupt the loop I think because it's a fatal error and if it happened for one of the nat gateways, chances are it will happen for all of them. I would just return the error directly here.

azure/services/natgateways/spec.go
CecileRobertMichon: cc @jackfrancis

azure/services/natgateways/natgateways.go
Jont828: Do you mean colliding with the return value of `async.CreateResource`? I thought `natGateway` would be more specific than `result`.

azure/services/natgateways/natgateways.go
Jont828: I think that was just a mistake. Just to clarify, do we still want to use `tele.StartSpanWithLogger()` everwhere except for   `IsDone`, which is using `tele.Tracer().Start()` on all the async services so far.

CecileRobertMichon: I think `IsDone` should be using `tele.StartSpanWithLogger()` too, not sure why it's not using it on the other services.

Jont828:
Holding to rebase off of #1874 once it merges

Jont828:
#1874 has merged, don't need to hold this anymore

azure/scope/cluster.go
```diff
@@ -220,19 +221,22 @@ func (s *ClusterScope) RouteTableSpecs() []azure.RouteTableSpec {
 	return routetables
 }
 
-// NatGatewaySpecs returns the node nat gateway.
-func (s *ClusterScope) NatGatewaySpecs() []azure.NatGatewaySpec {
-	var natGateways []azure.NatGatewaySpec
+// NatGatewaySpecs returns the node NAT gateway.
+func (s *ClusterScope) NatGatewaySpecs() []azure.ResourceSpecGetter {
+	natGateways := []azure.ResourceSpecGetter{}
```
CecileRobertMichon, line 226:
```suggestion
	var natGateways []azure.ResourceSpecGetter
```

https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1812#discussion_r740309773

azure/services/natgateways/spec.go
Jont828: I pushed the changes to implement `hasPublicIP` instead, which I think gets us around this issue since we search all the public IPs in a NAT gateway.

azure/services/natgateways/client.go
```diff
@@ -54,43 +50,113 @@ func netNatGatewaysClient(subscriptionID string, baseURI string, authorizer auto
...
 }
 
-// Delete deletes the specified nat gateway.
-func (ac *azureClient) Delete(ctx context.Context, resourceGroupName, natGatewayName string) error {
-	ctx, _, done := tele.StartSpanWithLogger(ctx, "natgateways.AzureClient.Delete")
+// DeleteAsync deletes a Nat Gateway asynchronously. DeleteAsync sends a DELETE
+// request to Azure and if accepted without error, the func will return a Future which can be used to track the ongoing
+// progress of the operation.
+func (ac *azureClient) DeleteAsync(ctx context.Context, spec azure.ResourceSpecGetter) (azureautorest.FutureAPI, error) {
+	ctx, _, done := tele.StartSpanWithLogger(ctx, "natgateways.Service.DeleteAsync")
```
CecileRobertMichon, line 96:
```suggestion
	ctx, _, done := tele.StartSpanWithLogger(ctx, "natgateways. AzureClient.DeleteAsync")
```

azure/services/natgateways/client.go
```diff
@@ -54,43 +50,113 @@ func netNatGatewaysClient(subscriptionID string, baseURI string, authorizer auto
...
 }
 
-// CreateOrUpdate create or updates a nat gateway in a specified resource group.
-func (ac *azureClient) CreateOrUpdate(ctx context.Context, resourceGroupName string, natGatewayName string, natGateway network.NatGateway) error {
-	ctx, _, done := tele.StartSpanWithLogger(ctx, "natgateways.AzureClient.CreateOrUpdate")
+// CreateOrUpdateAsync creates or updates a Nat Gateway asynchronously.
+// It sends a PUT request to Azure and if accepted without error, the func will return a Future which can be used to track the ongoing
+// progress of the operation.
+func (ac *azureClient) CreateOrUpdateAsync(ctx context.Context, spec azure.ResourceSpecGetter, parameters interface{}) (interface{}, azureautorest.FutureAPI, error) {
+	ctx, _, done := tele.StartSpanWithLogger(ctx, "natgateways.Service.CreateOrUpdateAsync")
```
CecileRobertMichon, line 64:
```suggestion
	ctx, _, done := tele.StartSpanWithLogger(ctx, "natgateways. AzureClient.CreateOrUpdateAsync")
```

azure/services/natgateways/client.go
```diff
@@ -54,43 +50,113 @@ func netNatGatewaysClient(subscriptionID string, baseURI string, authorizer auto
...
+	if err != nil {
+		return false, errors.Wrap(err, "failed checking if the operation was complete")
+	}
+
+	return isDone, nil
+}
+
+// Result fetches the result of a long-running operation future.
+func (ac *azureClient) Result(ctx context.Context, futureData azureautorest.FutureAPI, futureType string) (interface{}, error) {
+	_, _, done := tele.StartSpanWithLogger(ctx, "natgateways.Service.Result")
```
CecileRobertMichon, line 133:
```suggestion
	_, _, done := tele.StartSpanWithLogger(ctx, "natgateways. AzureClient.Result")
```

Jont828:
@CecileRobertMichon Pushing some changes. Should it be `natgateways.azureClient` or `natgateways.AzureClient` since the azure client is called `azureClient` here and  not exported?

CecileRobertMichon:
> Should it be natgateways.azureClient or natgateways.AzureClient since the azure client is called azureClient here and not exported?

That's a good question, @mboersma do you have thoughts on that one? not sure if there's a convention for naming spans when the receiver is private. Right now it seems like we're mostly doing  `AzureClient` in other places even when `azureClient` is unexported, except for scalesetvms.

azure/scope/cluster.go
```diff
@@ -220,24 +221,33 @@ func (s *ClusterScope) RouteTableSpecs() []azure.RouteTableSpec {
...
+					NatGatewayIP: infrav1.PublicIPSpec{
+						Name: subnet.NatGateway.NatGatewayIP.Name,
+					},
+				}
+			}
 		}
 	}
 
-	return natGateways
+	result := make([]azure.ResourceSpecGetter, 0, len(natGateways))
```
CecileRobertMichon, line 245: suggestion for a small optimization in time and space: instead of having to loop through twice and storing the list of azure.ResourceSpecGetter twice (once in `natGateways` and once in `result`), you could instead define a `natGatewaysSet := map[string]struct{}` (cheaper since you're not storing the objects in there) and directly store the NatGatewaySpecs in the return slice. Each time you add a NAT GW to the specs you also add it to your set. And you use the set to determine if you need to add a nat GW to the specs for each subnet. That way you only loop once through the subnets.

```
for _, subnet := range s.NodeSubnets() {
	if subnet.IsNatGatewayEnabled() {
		if _, ok := natGatewaysSet[subnet.NatGateway.Name]; !ok {
			specs = append(specs, &natgateways.NatGatewaySpec{...})
			natGatewaysSet[subnet.NatGateway.Name] = struct{}{}
                }
        }
```

azure/services/natgateways/natgateways.go
```diff
@@ -18,156 +18,112 @@ package natgateways
...
 	ctx, log, done := tele.StartSpanWithLogger(ctx, "natgateways.Service.Reconcile")
 	defer done()
 
+	ctx, cancel := context.WithTimeout(ctx, reconciler.DefaultAzureServiceReconcileTimeout)
+	defer cancel()
+
 	if !s.Scope.Vnet().IsManaged(s.Scope.ClusterName()) {
 		log.V(4).Info("Skipping nat gateways reconcile in custom vnet mode")
+
+		s.Scope.UpdatePutStatus(infrav1.NATGatewaysReadyCondition, serviceName, nil)
```
CecileRobertMichon, line 68: we might want to get rid of this when #1868 is fixed

azure/services/natgateways/natgateways.go
```diff
@@ -18,156 +18,112 @@ package natgateways
...
 	ctx, log, done := tele.StartSpanWithLogger(ctx, "natgateways.Service.Delete")
 	defer done()
 
+	ctx, cancel := context.WithTimeout(ctx, reconciler.DefaultAzureServiceReconcileTimeout)
+	defer cancel()
+
 	if !s.Scope.Vnet().IsManaged(s.Scope.ClusterName()) {
 		log.V(4).Info("Skipping nat gateway deletion in custom vnet mode")
+
+		s.Scope.UpdateDeleteStatus(infrav1.NATGatewaysReadyCondition, serviceName, nil)
```
CecileRobertMichon, line 111: not sure we want to do this: if the NAT gateways aren't managed and we don't delete them, we shouldn't mark them as deleted in conditions

azure/services/natgateways/natgateways.go
```diff
@@ -18,156 +18,112 @@ package natgateways
...
-		}
-		log.V(2).Info("successfully created nat gateway", "nat gateway", natGatewaySpec.Name)
-		natGateway := infrav1.NatGateway{
-			ID:   azure.NatGatewayID(s.Scope.SubscriptionID(), s.Scope.ResourceGroup(), natGatewaySpec.Name),
-			Name: natGatewaySpec.Name,
-			NatGatewayIP: infrav1.PublicIPSpec{
-				Name: natGatewaySpec.NatGatewayIP.Name,
-			},
+			// TODO: ideally we wouldn't need to set the subnet spec based on the result of the create operation
+			s.Scope.SetNatGatewayIDInSubnets(natGatewaySpec.ResourceName(), *natGateway.ID)
```
CecileRobertMichon, line 92: can you please add a comment about this one in #1685 so we remember it when we fix the issue

azure/services/natgateways/natgateways.go
```diff
@@ -18,156 +18,112 @@ package natgateways
...
 			}
-		default:
-			// nat gateway doesn't exist but its name was specified in the subnet, let's create it
-			log.V(2).Info("nat gateway doesn't exist yet, creating it", "nat gateway", natGatewaySpec.Name)
 		}
+		if err == nil {
+			natGateway, ok := result.(network.NatGateway)
+			if !ok {
+				// Return out of loop since this would be an unexpcted fatal error
+				resultingErr = errors.Errorf("created resource %T is not a network.NatGateway", result)
```
CecileRobertMichon, line 87:
```suggestion
				return errors.Errorf("created resource %T is not a network.NatGateway", result)
```

azure/services/natgateways/spec.go
```diff
@@ -0,0 +1,98 @@
...
+			},
+		},
+	}
+
+	return natGatewayToCreate, nil
+}
+
+func hasPublicIP(natGateway network.NatGateway, publicIPName string) bool {
+	// We must have a non-nil, non-"empty" PublicIPAddresses
+	if !(natGateway.PublicIPAddresses != nil && len(*natGateway.PublicIPAddresses) > 0) {
```
CecileRobertMichon, line 84: since you're looping through the IPs below it should be okay if there is more than 1 public IP.  

should this be
```suggestion
	if !(natGateway.PublicIPAddresses != nil && len(*natGateway.PublicIPAddresses) == 0) {
```
? that's what I read from "non-"empty""

api/v1alpha4/types.go
```diff
@@ -133,7 +133,7 @@ type RouteTable struct {
 	Name string `json:"name,omitempty"`
 }
 
-// NatGateway defines an Azure Nat Gateway.
+// NatGateway defines an Azure NAT gateway.
```
CecileRobertMichon, line 136: let's not change v1alpha4 types, even if it's just code comments (those are actually used as CRD descriptions)

azure/services/natgateways/natgateways.go
Jont828: Should we still set the put status in that case? And if users BYO VNet, should we care about updating the delete status to reflect if they delete the NAT gateway themselves?

azure/services/natgateways/natgateways.go
Jont828: I used the `break` statement to avoid calling `UpdatePutStatus` again similar to the changes in #1861, do you think it's worth changing back?

azure/scope/cluster.go
```diff
@@ -220,20 +221,26 @@ func (s *ClusterScope) RouteTableSpecs() []azure.RouteTableSpec {
 	return routetables
 }
 
-// NatGatewaySpecs returns the node nat gateway.
-func (s *ClusterScope) NatGatewaySpecs() []azure.NatGatewaySpec {
-	var natGateways []azure.NatGatewaySpec
+// NatGatewaySpecs returns the node NAT gateway.
+func (s *ClusterScope) NatGatewaySpecs() []azure.ResourceSpecGetter {
+	natGatewaySet := make(map[string]struct{})
+	natGateways := make([]azure.ResourceSpecGetter, 0)
```
CecileRobertMichon, line 227:
```suggestion
	var natGateways []azure.ResourceSpecGetter
```

azure/services/natgateways/natgateways.go
```diff
@@ -18,156 +18,112 @@ package natgateways
...
+				resultingErr = err
 			}
-		default:
-			// nat gateway doesn't exist but its name was specified in the subnet, let's create it
-			log.V(2).Info("nat gateway doesn't exist yet, creating it", "nat gateway", natGatewaySpec.Name)
 		}
+		if err == nil {
+			natGateway, ok := result.(network.NatGateway)
+			if !ok {
+				// Return out of loop since this would be an unexpcted fatal error
```
CecileRobertMichon, line 86:
```suggestion
				// Return out of loop since this would be an unexpected fatal error
```

azure/services/natgateways/natgateways.go
CecileRobertMichon: got it

azure/services/natgateways/natgateways.go
CecileRobertMichon: I'm leaning towards not showing any conditions at all for resources we don't manage and only focus the Status on resources we create/update/delete, but I think we need #1868 in order to do that cleanly

CecileRobertMichon:
Please squash commits

CecileRobertMichon (commented):
/lgtm
/approve

api/v1alpha4/types.go
```diff
@@ -133,7 +133,7 @@ type RouteTable struct {
 	Name string `json:"name,omitempty"`
 }
 
-// NatGateway defines an Azure Nat Gateway.
+// NatGateway defines an Azure Nat gateway.
```
CecileRobertMichon, line 136: we shouldn't change v1alpha4, did the revert get lost in the rebase?

azure/services/disks/client.go
```diff
@@ -80,7 +80,7 @@ func (ac *azureClient) Result(ctx context.Context, futureData azureautorest.Futu
 
 // IsDone returns true if the long-running operation has completed.
 func (ac *azureClient) IsDone(ctx context.Context, future azureautorest.FutureAPI) (bool, error) {
-	ctx, _, done := tele.StartSpanWithLogger(ctx, "natgateways.Service.IsDone")
+	ctx, _, done := tele.StartSpanWithLogger(ctx, "disks.AzureClient.IsDone")
```
CecileRobertMichon, line 83:
```suggestion
	ctx, _, done := tele.StartSpanWithLogger(ctx, "disks.azureClient.IsDone")
```

api/v1alpha4/types.go
```diff
@@ -134,7 +134,7 @@ type RouteTable struct {
 }
 
 // NatGateway defines an Azure Nat Gateway.
-// NAT gateway resources are part of Vnet NAT and provide outbound Internet connectivity for subnets of a virtual network.
+// Nat Gateway resources are part of Vnet NAT and provide outbound Internet connectivity for subnets of a virtual network.
```
CecileRobertMichon, line 137: one more here

PR #1870, opened 2021-11-16
Make inbound NAT rules reconcile/delete async

### LINKED ISSUE #1712: async inboundnatrules

Jont828:
**What type of PR is this?**

/kind feature
**What this PR does / why we need it**: Implementation of an async service for inbound NAT rules as a follow up for #1610 and #1541.

Fixes #1712

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Make inbound NAT rules reconcile/delete async
```

### DIFF
--- a/azure/scope/machine.go
+++ b/azure/scope/machine.go
@@ -40,6 +40,7 @@ import (
 	"sigs.k8s.io/cluster-api-provider-azure/azure"
 	"sigs.k8s.io/cluster-api-provider-azure/azure/services/availabilitysets"
 	"sigs.k8s.io/cluster-api-provider-azure/azure/services/disks"
+	"sigs.k8s.io/cluster-api-provider-azure/azure/services/inboundnatrules"
 	"sigs.k8s.io/cluster-api-provider-azure/azure/services/resourceskus"
 	"sigs.k8s.io/cluster-api-provider-azure/azure/services/virtualmachines"
 	"sigs.k8s.io/cluster-api-provider-azure/util/futures"
@@ -193,16 +194,25 @@ func (m *MachineScope) PublicIPSpecs() []azure.PublicIPSpec {
 }
 
 // InboundNatSpecs returns the inbound NAT specs.
-func (m *MachineScope) InboundNatSpecs() []azure.InboundNatSpec {
+func (m *MachineScope) InboundNatSpecs(portsInUse map[int32]struct{}) []azure.ResourceSpecGetter {
+	// The existing inbound NAT rules are needed in order to find an available SSH port for each new inbound NAT rule.
 	if m.Role() == infrav1.ControlPlane {
-		return []azure.InboundNatSpec{
-			{
-				Name:             m.Name(),
-				LoadBalancerName: m.APIServerLBName(),
-			},
+		spec := &inboundnatrules.InboundNatSpec{
+			Name:                      m.Name(),
+			ResourceGroup:             m.ResourceGroup(),
+			LoadBalancerName:          m.APIServerLBName(),
+			FrontendIPConfigurationID: nil,
+			PortsInUse:                portsInUse,
 		}
+		if frontEndIPs := m.APIServerLB().FrontendIPs; len(frontEndIPs) > 0 {
+			ipConfig := frontEndIPs[0].Name
+			id := azure.FrontendIPConfigID(m.SubscriptionID(), m.ResourceGroup(), m.APIServerLBName(), ipConfig)
+			spec.FrontendIPConfigurationID = to.StringPtr(id)
+		}
+
+		return []azure.ResourceSpecGetter{spec}
 	}
-	return []azure.InboundNatSpec{}
+	return []azure.ResourceSpecGetter{}
 }
 
 // NICSpecs returns the network interface specs.
--- a/azure/scope/machine_test.go
+++ b/azure/scope/machine_test.go
@@ -18,7 +18,9 @@ package scope
 
 import (
 	"context"
+	"fmt"
 	"reflect"
+	"strings"
 	"testing"
 
 	autorestazure "github.com/Azure/go-autorest/autorest/azure"
@@ -32,7 +34,19 @@ import (
 	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
 	"sigs.k8s.io/cluster-api-provider-azure/azure"
 	"sigs.k8s.io/cluster-api-provider-azure/azure/services/disks"
+	"sigs.k8s.io/cluster-api-provider-azure/azure/services/inboundnatrules"
 )
+
+func specArrayToString(specs []azure.ResourceSpecGetter) string {
+	var sb strings.Builder
+	sb.WriteString("[ ")
+	for _, spec := range specs {
+		sb.WriteString(fmt.Sprintf("%+v ", spec))
+	}
+	sb.WriteString("]")
+
+	return sb.String()
+}
 
 func TestMachineScope_Name(t *testing.T) {
 	tests := []struct {
@@ -283,7 +297,7 @@ func TestMachineScope_InboundNatSpecs(t *testing.T) {
 	tests := []struct {
 		name         string
 		machineScope MachineScope
-		want         []azure.InboundNatSpec
+		want         []azure.ResourceSpecGetter
 	}{
 		{
 			name: "returns empty when infra is not control plane",
@@ -295,7 +309,7 @@ func TestMachineScope_InboundNatSpecs(t *testing.T) {
 					},
 				},
 			},
-			want: []azure.InboundNatSpec{},
+			want: []azure.ResourceSpecGetter{},
 		},
 		{
 			name: "returns InboundNatSpec when infra is control plane",
@@ -313,29 +327,48 @@ func TestMachineScope_InboundNatSpecs(t *testing.T) {
 					},
 				},
 				ClusterScoper: &ClusterScope{
+					AzureClients: AzureClients{
+						EnvironmentSettings: auth.EnvironmentSettings{
+							Values: map[string]string{
+								auth.SubscriptionID: "123",
+							},
+						},
+					},
 					AzureCluster: &infrav1.AzureCluster{
 						Spec: infrav1.AzureClusterSpec{
+							ResourceGroup:  "my-rg",
+							SubscriptionID: "123",
 							NetworkSpec: infrav1.NetworkSpec{
 								APIServerLB: infrav1.LoadBalancerSpec{
 									Name: "foo-loadbalancer",
+									FrontendIPs: []infrav1.FrontendIP{
+										{
+											Name: "foo-frontend-ip",
+										},
+									},
 								},
 							},
 						},
 					},
 				},
 			},
-			want: []azure.InboundNatSpec{
-				{
-					Name:             "machine-name",
-					LoadBalancerName: "foo-loadbalancer",
+			want: []azure.ResourceSpecGetter{
+				&inboundnatrules.InboundNatSpec{
+					Name:                      "machine-name",
+					LoadBalancerName:          "foo-loadbalancer",
+					ResourceGroup:             "my-rg",
+					FrontendIPConfigurationID: to.StringPtr(azure.FrontendIPConfigID("123", "my-rg", "foo-loadbalancer", "foo-frontend-ip")),
+					PortsInUse:                make(map[int32]struct{}),
 				},
 			},
 		},
 	}
 	for _, tt := range tests {
+		tt := tt
 		t.Run(tt.name, func(t *testing.T) {
-			if got := tt.machineScope.InboundNatSpecs(); !reflect.DeepEqual(got, tt.want) {
-				t.Errorf("InboundNatSpecs() = %v, want %v", got, tt.want)
+			t.Parallel()
+			if got := tt.machineScope.InboundNatSpecs(make(map[int32]struct{})); !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("InboundNatSpecs() = %s, want %s", specArrayToString(got), specArrayToString(tt.want))
 			}
 		})
 	}
--- a/azure/services/inboundnatrules/client.go
+++ b/azure/services/inboundnatrules/client.go
@@ -18,19 +18,28 @@ package inboundnatrules
 
 import (
 	"context"
+	"encoding/json"
+	"fmt"
 
 	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2021-02-01/network"
 	"github.com/Azure/go-autorest/autorest"
+	azureautorest "github.com/Azure/go-autorest/autorest/azure"
+	"github.com/pkg/errors"
 
+	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
 	"sigs.k8s.io/cluster-api-provider-azure/azure"
+	"sigs.k8s.io/cluster-api-provider-azure/util/reconciler"
 	"sigs.k8s.io/cluster-api-provider-azure/util/tele"
 )
 
 // client wraps go-sdk.
 type client interface {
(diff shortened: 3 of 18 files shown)


azure/services/inboundnatrules/client.go
```diff
@@ -55,42 +66,147 @@ func newInboundNatRulesClient(subscriptionID string, baseURI string, authorizer
...
 
 	return ac.inboundnatrules.Get(ctx, resourceGroupName, lbName, inboundNatRuleName, "")
 }
 
-// CreateOrUpdate creates or updates a inbound NAT rules.
-func (ac *azureClient) CreateOrUpdate(ctx context.Context, resourceGroupName string, lbName string, inboundNatRuleName string, inboundNatRuleParameters network.InboundNatRule) error {
-	ctx, _, done := tele.StartSpanWithLogger(ctx, "inboundnatrules.AzureClient.CreateOrUpdate")
+// List returns all inbound NAT rules on a load balancer
+func (ac *azureClient) List(ctx context.Context, resourceGroupName, lbName string) ([]network.InboundNatRule, error) {
+	ctx, _, done := tele.StartSpanWithLogger(ctx, "inboundnatrules.azureClient.Get")
```
CecileRobertMichon, line 77:
```suggestion
	ctx, _, done := tele.StartSpanWithLogger(ctx, "inboundnatrules.azureClient.List")
```

azure/services/inboundnatrules/inboundnatrules.go
```diff
@@ -19,133 +19,98 @@ package inboundnatrules
...
+	azure.AsyncStatusUpdater
+	APIServerLBName() string
+	InboundNatSpecs([]network.InboundNatRule) []azure.ResourceSpecGetter
 }
 
 // Service provides operations on Azure resources.
 type Service struct {
 	Scope InboundNatScope
 	client
-	loadBalancersClient loadbalancers.Client
```
CecileRobertMichon, line 43: 🥳

azure/services/inboundnatrules/inboundnatrules.go
```diff
@@ -19,133 +19,98 @@ package inboundnatrules
...
-	defer done()
-
-	for _, inboundNatSpec := range s.Scope.InboundNatSpecs() {
-		s.Scope.V(2).Info("creating inbound NAT rule", "NAT rule", inboundNatSpec.Name)
-
-		lb, err := s.loadBalancersClient.Get(ctx, s.Scope.ResourceGroup(), inboundNatSpec.LoadBalancerName)
-		if err != nil {
-			return errors.Wrapf(err, "failed to get Load Balancer %s", inboundNatSpec.LoadBalancerName)
-		}
+	ctx, span := tele.Tracer().Start(ctx, "inboundnatrules.Service.Reconcile")
```
CecileRobertMichon, line 61: let's keep using tele.StartSpanWithLogger everywhere

it allows us to get a custom logger with the correlation ID (see #1806)

azure/scope/machine.go
```diff
@@ -191,16 +194,21 @@ func (m *MachineScope) PublicIPSpecs() []azure.PublicIPSpec {
...
 
 // InboundNatSpecs returns the inbound NAT specs.
-func (m *MachineScope) InboundNatSpecs() []azure.InboundNatSpec {
+func (m *MachineScope) InboundNatSpecs(existingRules []network.InboundNatRule) []azure.ResourceSpecGetter {
 	if m.Role() == infrav1.ControlPlane {
-		return []azure.InboundNatSpec{
-			{
-				Name:             m.Name(),
-				LoadBalancerName: m.APIServerLBName(),
+		ipConfig := m.APIServerLB().FrontendIPs[0].Name
```
CecileRobertMichon, line 199: // TODO: need to figure out how this should be handles when the LB has more than one frontend IP

azure/services/inboundnatrules/spec.go
```diff
@@ -0,0 +1,117 @@
...
+func (s *InboundNatSpec) ResourceName() string {
+	return s.Name
+}
+
+// ResourceGroupName returns the name of the resource group.
+func (s *InboundNatSpec) ResourceGroupName() string {
+	return s.ResourceGroup
+}
+
+// OwnerResourceName is a no-op for virtual network peerings.
```
CecileRobertMichon, line 45: comment needs to be adjusted

azure/services/inboundnatrules/spec.go
```diff
@@ -0,0 +1,117 @@
...
+type InboundNatSpec struct {
+	Name                      string
+	LoadBalancerName          string
+	ResourceGroup             string
+	FrontendIPConfigurationID string
+	FrontendPort              string
+	ExistingRules             []network.InboundNatRule
+}
+
+// ResourceName returns the name of the virtual network peering.
```
CecileRobertMichon, line 35: comment needs to be adjusted

azure/services/inboundnatrules/spec.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+// ResourceGroupName returns the name of the resource group.
+func (s *InboundNatSpec) ResourceGroupName() string {
+	return s.ResourceGroup
+}
+
+// OwnerResourceName is a no-op for virtual network peerings.
+func (s *InboundNatSpec) OwnerResourceName() string {
+	return s.LoadBalancerName
```
CecileRobertMichon, line 47: 👍

azure/scope/machine.go
```diff
@@ -191,16 +194,21 @@ func (m *MachineScope) PublicIPSpecs() []azure.PublicIPSpec {
 }
 
 // InboundNatSpecs returns the inbound NAT specs.
-func (m *MachineScope) InboundNatSpecs() []azure.InboundNatSpec {
+func (m *MachineScope) InboundNatSpecs(existingRules []network.InboundNatRule) []azure.ResourceSpecGetter {
```
CecileRobertMichon, line 197: let's add a comment to explain what existingRules means in this context and why it's here

azure/scope/machine.go
```diff
@@ -101,9 +103,10 @@ type MachineScope struct {
 
 // MachineCache stores common machine information so we don't have to hit the API multiple times within the same reconcile loop.
 type MachineCache struct {
-	BootstrapData string
-	VMImage       *infrav1.Image
-	VMSKU         resourceskus.SKU
+	BootstrapData    string
+	VMImage          *infrav1.Image
+	VMSKU            resourceskus.SKU
+	ExistingNatRules []network.InboundNatRule
```
CecileRobertMichon, line 109: you probably don't need this anymore

azure/scope/machine.go
CecileRobertMichon: also wondering if we could avoid adding a dependency on the SDK in here by maybe passing in our own type that just has the rule name and frontend port (we would convert the []network. InboundNatRule in the service using a new func in the converters package). wdyt?

azure/converters/inboundnatrules.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2021-02-01/network"
+)
+
+type ExistingInboundNatSpec struct {
+	Name         string
+	FrontendPort int32
+}
+
+// SKUtoSDK converts infrav1.SKU into a network.LoadBalancerSkuName.
```
CecileRobertMichon, line 28: comment seems wrong here

azure/scope/machine.go
Jont828: It looks like we have a check in `azurecluster_validation.go` to ensure there's exactly one frontend IP, so it looks like we'd only need to handle that if our spec invariant changes.

```
// There should only be one IP config.
if len(lb.FrontendIPs) != 1 || pointer.Int32Deref(lb.FrontendIPsCount, 1) != 1 {
	allErrs = append(allErrs, field.Invalid(fldPath.Child("frontendIPConfigs"), lb.FrontendIPs,
		"API Server Load balancer should have 1 Frontend IP"))
} 
```

Jont828:
@CecileRobertMichon @mboersma Squashed the commits, would appreciate a review when you get the chance

CecileRobertMichon (commented):
will do another round of review on Monday but for now just dropping this comment from @shysank about using named returns from another PR that is relevant here:

https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1686#discussion_r777675199

Here is the commit that does this in the route tables PR: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1686/commits/b2302ea1e03fbbaf428f4a2987c2c5f850dc40ae

azure/scope/machine_test.go
```diff
@@ -313,30 +316,51 @@ func TestMachineScope_InboundNatSpecs(t *testing.T) {
...
+		tt := tt
 		t.Run(tt.name, func(t *testing.T) {
-			if got := tt.machineScope.InboundNatSpecs(); !reflect.DeepEqual(got, tt.want) {
-				t.Errorf("InboundNatSpecs() = %v, want %v", got, tt.want)
-			}
+			g := NewWithT(t)
+
+			t.Parallel()
+			result := tt.machineScope.InboundNatSpecs([]converters.ExistingInboundNatSpec{})
+			fmt.Printf("InboundNatSpecs() = %+v, want %+v\n", result, tt.want)
```
mboersma, line 362: Do we want to print the expectations always? Maybe this `fmt.Printf` should be a second argument to the `BeEquivalentTo()` on the next line, so it only prints when a test case fails. WDYT?

azure/services/inboundnatrules/client.go
```diff
@@ -42,55 +51,148 @@ var _ client = (*azureClient)(nil)
...
+		}
+	}
+
+	return natRules, nil
 }
 
-// CreateOrUpdate creates or updates a inbound NAT rules.
-func (ac *azureClient) CreateOrUpdate(ctx context.Context, resourceGroupName string, lbName string, inboundNatRuleName string, inboundNatRuleParameters network.InboundNatRule) error {
-	ctx, _, done := tele.StartSpanWithLogger(ctx, "inboundnatrules.AzureClient.CreateOrUpdate")
+// CreateOrUpdateAsync creates or updates a inbound NAT rule asynchronously.
```
mboersma, line 96:
```suggestion
// CreateOrUpdateAsync creates or updates an inbound NAT rule asynchronously.
```

azure/services/inboundnatrules/spec.go
```diff
@@ -0,0 +1,104 @@
...
+		},
+	}
+
+	return rule, nil
+}
+
+func getAvailablePort(portsInUse map[int32]struct{}) (int32, error) {
+	var i int32 = 22
+	if _, ok := portsInUse[22]; ok {
+		for i = 2201; i < 2220; i++ {
```
mboersma, line 94: I confess I don't know why we're constrained to this narrow range of ports. If it's possible to add an explanation in a code comment, that would be illuminating.

Jont828: I'm not sure either, it's just the existing code that I moved over into the spec.

azure/scope/machine_test.go
Jont828: Oh that was a debugging statement, I'll go ahead and just remove that.

Jont828: Actually I'm not sure how helpful that line is. Since it's returning a ResourceSpecGetter, the output just looks something like this. Do you have any suggestions on what to do here?
``` 
InboundNatSpecs([]converters.ExistingInboundNatSpec{}) = [0xc000112050], want [0xc0004812c0]
```

azure/converters/inboundnatrules.go
```diff
@@ -0,0 +1,34 @@
...
+limitations under the License.
+*/
+
+package converters
+
+import (
+	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2021-02-01/network"
+)
+
+type ExistingInboundNatSpec struct {
```
CecileRobertMichon, line 23: Looks like the only place we are using this is converting it to a map of ports in use https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1870/files#diff-64818b5b4ad20088a23cd553a259558754c7f3a3b73813d166b307c0f44a694cR64-R67

Have you considered getting rid of this intermediate struct and directly storing usedPorts or portsInUsed as a `map[int32]struct{}` ?

azure/services/inboundnatrules/inboundnatrules.go
```diff
@@ -19,136 +19,99 @@ package inboundnatrules
...
 
-		err = s.client.CreateOrUpdate(ctx, s.Scope.ResourceGroup(), to.String(lb.Name), inboundNatSpec.Name, rule)
-		if err != nil {
-			return errors.Wrapf(err, "failed to create inbound NAT rule %s", inboundNatSpec.Name)
+	// We go through the list of InboundNatSpecs to reconcile each one, independently of the result of the previous one.
+	// If multiple errors occur, we return the most pressing one.
+	//  Order of precedence (highest -> lowest) is: error that is not an operationNotDoneError (i.e. error creating) -> operationNotDoneError (i.e. creating in progress) -> no error (i.e. created)
+	var result error
+	for _, natRule := range s.Scope.InboundNatSpecs(existingRuleSpecs) {
+		if _, err := s.CreateResource(ctx, natRule, serviceName); err != nil {
```
CecileRobertMichon, line 84: thoughts for later: in theory if this was used to create multiple rules we'd have to update existingRules to include the previous ones that were just created since they are now taking ports that used to be available. This doesn't matter much in practice currently since we are only creating one rule per machine but could lead to errors down the road. Consider adding a TODO or a comment to that effect, or modifying the logic to include previously created rules in existing rules.

azure/services/inboundnatrules/spec.go
CecileRobertMichon: NAT rules need to use a unique port. Since we need one NAT rule per control plane and we expect to have 1,3,5, maybe 9 control planes, there should never be more than 9 ports in use. This is an artificial limit of 20 ports that we can pick from, which should be plenty enough (in reality we should never reach that limit). These NAT rules are used for SSH purposes which is why we start at 22 and then use 2201, 2202, etc.

Please update the TODO with this explanation. 

Perharps a good improvement in the future would be to pass in a port range in the InboundNatSpec so it can vary based on the use of the rules being created. Right now we assume SSH rules for control planes but it might not always be the case.

azure/scope/machine_test.go
```diff
@@ -313,29 +315,58 @@ func TestMachineScope_InboundNatSpecs(t *testing.T) {
...
 				},
 			},
 		},
 	}
 	for _, tt := range tests {
+		tt := tt
 		t.Run(tt.name, func(t *testing.T) {
-			if got := tt.machineScope.InboundNatSpecs(); !reflect.DeepEqual(got, tt.want) {
-				t.Errorf("InboundNatSpecs() = %v, want %v", got, tt.want)
+			t.Parallel()
```
CecileRobertMichon, line 357: +1 on making this test Parallel

azure/services/inboundnatrules/spec.go
```diff
@@ -0,0 +1,99 @@
...
+			Protocol:     network.TransportProtocolTCP,
+			FrontendPort: &sshFrontendPort,
+		},
+	}
+
+	return rule, nil
+}
+
+func getAvailablePort(portsInUse map[int32]struct{}) (int32, error) {
+	// TODO: should we use a different range of ports?
```
CecileRobertMichon, line 86: please update TODO based on discussion below

CecileRobertMichon (commented):
lgtm

/assign @mboersma

devigned (commented):
Code / tests look solid. Nice work!

lgtm

devigned (approved):

CecileRobertMichon:
/lgtm
/approve

🎉

CecileRobertMichon (approved):
/lgtm

Jont828:
Let's go!!!

PR #1873, opened 2021-11-17
chore(role_assignment): make roleassignment reconcile async

### LINKED ISSUE #1718: async roleassignments

sonasingh46:

 

 
**What type of PR is this?**

Implementation of async service for role assignment.

Fixes #1718 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Make roleassignment reconcile async
```


azure/scope/machine.go
```diff
@@ -23,6 +23,8 @@ import (
 	"strings"
 	"time"
 
+	"sigs.k8s.io/cluster-api-provider-azure/azure/services/roleassignments"
```
sonasingh46, line 26: This import here cause an import cycle.

azure/services/roleassignments/roleassignments.go
```diff
@@ -32,12 +31,14 @@ import (
 )
```
sonasingh46, line 31: Roleassginments imports scalesets. Scalesets in turn imports scope

sonasingh46: Ref: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/services/scalesets/scalesets.go#L36

azure/services/scalesets/scalesets_test.go
```diff
@@ -53,46 +47,47 @@ func init() {
...
-		Client:           client,
-		MachinePool:      new(clusterv1exp.MachinePool),
-		AzureMachinePool: new(infrav1exp.AzureMachinePool),
-		ClusterScope:     s,
-	})
-	g.Expect(err).NotTo(HaveOccurred())
-	actual := NewService(mps, resourceskus.NewStaticCache(nil, ""))
-	g.Expect(actual).ToNot(BeNil())
-}
+// ToDo: Fix this test case
```
sonasingh46, line 50: @CecileRobertMichon @shysank -- Any suggestions on fixing this test case. This test case requires to import scope package and hence causes import loop.

CecileRobertMichon: looking, I don't think we should be creating real scopes to test `NewService()`. `NewService` takes a ScaleSetScope, which we have a mock for, so we should be able to do:
```
mockCtrl := gomock.NewController(t)
defer mockCtrl.Finish()
 scopeMock := NewMockScaleSetScope(mockCtrl)
actual := NewService(scopeMock, resourceskus.NewStaticCache(nil, ""))
g.Expect(actual).ToNot(BeNil())
```

If that doesn't work, that test isn't testing much since the NewService func just sets the parameters into a struct and returns it, so worst case we can get rid of the test altogether.

Jont828:
Hey @sonasingh46, how's it going with the role assignments PR? I worked on a lot of the async refactor PRs with @CecileRobertMichon, and I'd be happy to answer any questions or to do a code review.

sonasingh46:
Hey @Jont828 -- Thanks for the help. 
I am trying to test the roleassignment part now. Meanwhile, you can take a look at the code or we can pair up on this one.

sonasingh46:
To test this change I am creating `system-assigned-identity` flavor from the tilt and it is failing with below error in tilt: 
```
Error from server (InternalError): error when creating "STDIN": Internal error occurred: failed calling webhook "default.kubeadmcontrolplane.controlplane.cluster.x-k8s.io": Post "https://capi-kubeadm-control-plane-webhook-service.capi-kubeadm-control-plane-system.svc:443/mutate-controlplane-cluster-x-k8s-io-v1beta1-kubeadmcontrolplane?timeout=10s": dial tcp 10.96.229.108:443: connect: connection refused
Error from server (InternalError): error when creating "STDIN": Internal error occurred: failed calling webhook "default.machinedeployment.cluster.x-k8s.io": Post "https://capi-webhook-service.capi-system.svc:443/mutate-cluster-x-k8s-io-v1beta1-machinedeployment?timeout=10s": dial tcp 10.96.52.90:443: connect: connection refused
```

api/v1beta1/zz_generated.deepcopy.go
```diff
@@ -22,7 +22,7 @@ limitations under the License.
 package v1beta1
 
 import (
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
Jont828, line 25: Not sure why this is changing since it'd still be named `v1`. Maybe go autosave changed some imports?

azure/scope/machine.go
```diff
@@ -289,17 +291,17 @@ func (m *MachineScope) DiskSpecs() []azure.ResourceSpecGetter {
...
+func (m *MachineScope) RoleAssignmentSpecs() []azure.ResourceSpecGetter {
+	roles := make([]azure.ResourceSpecGetter, 1)
 	if m.AzureMachine.Spec.Identity == infrav1.VMIdentitySystemAssigned {
-		return []azure.RoleAssignmentSpec{
-			{
-				MachineName:  m.Name(),
-				Name:         m.AzureMachine.Spec.RoleAssignmentName,
-				ResourceType: azure.VirtualMachine,
-			},
+		roles[0] = &roleassignments.RoleAssignmentSpec{
```
Jont828, line 297: This would be a good spot to assign these values so we don't need to in `Reconcile()`. Also that would mean you don't need to mock the `s.Scope.SubscriptionID()` return values in the test cases.

```
scope := fmt.Sprintf("/subscriptions/%s/", s.Scope.SubscriptionID())
rs.Scope = scope
contributorRoleDefinitionID := fmt.Sprintf("/subscriptions/%s/providers/Microsoft.Authorization/roleDefinitions/%s", s.Scope.SubscriptionID(), azureBuiltInContributorID)
rs.RoleDefinitionID = contributorRoleDefinitionID
```

azure/services/roleassignments/roleassignments.go
```diff
@@ -32,54 +34,65 @@ import (
...
+	defer cancel()
 
 	for _, roleSpec := range s.Scope.RoleAssignmentSpecs() {
-		switch roleSpec.ResourceType {
+		rs := roleSpec.(*RoleAssignmentSpec)
+		scope := fmt.Sprintf("/subscriptions/%s/", s.Scope.SubscriptionID())
+		rs.Scope = scope
+		contributorRoleDefinitionID := fmt.Sprintf("/subscriptions/%s/providers/Microsoft.Authorization/roleDefinitions/%s", s.Scope.SubscriptionID(), azureBuiltInContributorID)
+		rs.RoleDefinitionID = contributorRoleDefinitionID
+		switch rs.ResourceType {
```
Jont828, line 80: We could create a getter function such as `RoleAssignmentResourceType(name)` and add it to the role assignment scope to determine if you need to reconcile a VM or a VMSS.

azure/services/roleassignments/client.go
```diff
@@ -28,7 +32,12 @@ import (
 
 // client wraps go-sdk.
 type client interface {
-	Create(context.Context, string, string, authorization.RoleAssignmentCreateParameters) (authorization.RoleAssignment, error)
+	//Create(context.Context, string, string, authorization.RoleAssignmentCreateParameters) (authorization.RoleAssignment, error)
```
Jont828, line 35: Do we need this client interface? In the other async PRs, we've been able to remove them along with the client mocks if they're not directly called by a service.

azure/services/roleassignments/client.go
```diff
@@ -51,18 +60,54 @@ func newRoleAssignmentClient(subscriptionID string, baseURI string, authorizer a
...
+// Creating a roleassignment is not a long running operation, so we don't ever return a future.
+func (ac *azureClient) CreateOrUpdateAsync(ctx context.Context, spec azure.ResourceSpecGetter, parameters interface{}) (interface{}, azureautorest.FutureAPI, error) {
+	ctx, _, done := tele.StartSpanWithLogger(ctx, "groups.AzureClient.CreateOrUpdate")
 	defer done()
+	assignment, ok := parameters.(authorization.RoleAssignmentPropertiesWithScope)
+	if !ok {
+		return nil, nil, errors.Errorf("%T is not a authorization.RoleAssignment", parameters)
+	}
+	//scope := fmt.Sprintf("/subscriptions/%s/", *assignment.Scope)
+	roleAssignmentCreatePrams := authorization.RoleAssignmentCreateParameters{
```
Jont828, line 82: It's a bit awkward to need to recreate the parameters here like you said, but I think this is probably the best way to do it.

azure/services/roleassignments/roleassignments.go
```diff
@@ -123,27 +137,16 @@ func (s *Service) reconcileVMSS(ctx context.Context, roleSpec azure.RoleAssignme
 	return nil
 }
 
-func (s *Service) assignRole(ctx context.Context, roleAssignmentName string, principalID *string) error {
+func (s *Service) assignRole(ctx context.Context, roleSpec *RoleAssignmentSpec) error {
```
Jont828, line 140: We could probably remove this function since it just wraps `s.CreateResource()`

azure/services/roleassignments/roleassignments.go
```diff
@@ -32,54 +34,65 @@ import (
...
 func (s *Service) Reconcile(ctx context.Context) error {
 	ctx, _, done := tele.StartSpanWithLogger(ctx, "roleassignments.Service.Reconcile")
 	defer done()
+	ctx, cancel := context.WithTimeout(ctx, reconciler.DefaultAzureServiceReconcileTimeout)
+	defer cancel()
 
 	for _, roleSpec := range s.Scope.RoleAssignmentSpecs() {
-		switch roleSpec.ResourceType {
+		rs := roleSpec.(*RoleAssignmentSpec)
+		scope := fmt.Sprintf("/subscriptions/%s/", s.Scope.SubscriptionID())
```
Jont828, line 76: There are a ton of functions for generating IDs like this in [defaults.go](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/defaults.go) using `fmt.Sprintf("/subscriptions"...)`. You could probably add a function for getting the scope there.

azure/services/roleassignments/roleassignments.go
```diff
@@ -32,54 +34,65 @@ import (
...
+		rs.Scope = scope
+		contributorRoleDefinitionID := fmt.Sprintf("/subscriptions/%s/providers/Microsoft.Authorization/roleDefinitions/%s", s.Scope.SubscriptionID(), azureBuiltInContributorID)
```
Jont828, line 78: Same goes for here, you could add a `GenerateContributorRoleDefinitionID()` function too.

azure/services/roleassignments/roleassignments.go
```diff
@@ -94,7 +107,8 @@ func (s *Service) reconcileVM(ctx context.Context, roleSpec azure.RoleAssignment
 		return errors.Errorf("%T is not a compute.VirtualMachine", resultVMIface)
 	}
 
-	err = s.assignRole(ctx, roleSpec.Name, resultVM.Identity.PrincipalID)
+	roleSpec.PrincipalID = resultVM.Identity.PrincipalID
```
Jont828, line 110: It looks like this is the only spot where we can't get a spec value from `machine.go`. If we can pass this value in as an argument for `RoleAssignmentSpecs()` then that we could avoid the need to cast the spec type directly. Let me see if I can figure something out.

azure/services/roleassignments/roleassignments.go
```diff
@@ -32,54 +34,65 @@ import (
...
 }
 
-func (s *Service) reconcileVM(ctx context.Context, roleSpec azure.RoleAssignmentSpec) error {
+func (s *Service) reconcileVM(ctx context.Context, roleSpec *RoleAssignmentSpec) error {
 	ctx, log, done := tele.StartSpanWithLogger(ctx, "roleassignments.Service.reconcileVM")
 	defer done()
-
 	spec := &virtualmachines.VMSpec{
 		Name:          roleSpec.MachineName,
 		ResourceGroup: s.Scope.ResourceGroup(),
```
Jont828, line 98: I think we can get this value through `roleSpec.ResourceGroupName()` to avoid a scope call, which will simplify the test cases.

test/e2e/capi_test.go
```diff
@@ -1,3 +1,4 @@
+//go:build e2e
```
Jont828, line 1: We can remove this line, it's a weird issue where Go 1.17 will add it when generating files.

test/e2e/common.go
```diff
@@ -1,3 +1,4 @@
+//go:build e2e
```
Jont828, line 1: Same for this line.

test/e2e/data/infrastructure-azure/v1alpha4/cluster-template-kcp-adoption.yaml
```diff
@@ -0,0 +1,347 @@
+apiVersion: v1
```
Jont828, line 1: Where are these files coming from btw? Are they related to this PR? If not it might be something going on with `make generate`.

azure/services/roleassignments/spec.go
```diff
@@ -0,0 +1,72 @@
...
+func (s *RoleAssignmentSpec) ResourceName() string {
+	return s.RoleAssignmentName
+}
+
+// ResourceGroupName returns the name of the resource group.
+func (s *RoleAssignmentSpec) ResourceGroupName() string {
+	return s.ResourceGroup
+}
+
+// OwnerResourceName is a no-op for role assignment.
```
Jont828, line 50:
```suggestion
// OwnerResourceName is a no-op for role assignments.
```

azure/services/roleassignments/spec.go
```diff
@@ -0,0 +1,72 @@
...
+
+import (
+	"fmt"
+
+	"github.com/Azure/azure-sdk-for-go/profiles/2019-03-01/authorization/mgmt/authorization"
+	"github.com/Azure/go-autorest/autorest/to"
+	"github.com/pkg/errors"
+)
+
+// RoleAssignmentSpec defines the specification for a Role Assignment.
```
Jont828, line 27:
```suggestion
// RoleAssignmentSpec defines the specification for a role assignment.
```

azure/services/async/async.go
```diff
@@ -102,6 +101,7 @@ func (s *Service) CreateResource(ctx context.Context, spec azure.ResourceSpecGet
 
 	// Get the resource if it already exists, and use it to construct the desired resource parameters.
 	var existingResource interface{}
+
```
Jont828, line 104: Are these line changes needed?

api/v1beta1/zz_generated.deepcopy.go
sonasingh46: fixed it... I guess my editor was doing something.

azure/services/roleassignments/client.go
```diff
@@ -51,18 +60,52 @@ func newRoleAssignmentClient(subscriptionID string, baseURI string, authorizer a
...
-// for a resource.
-// roleAssignmentName - the name of the role assignment to create. It can be any valid GUID.
-// parameters - parameters for the role assignment.
-func (ac *azureClient) Create(ctx context.Context, scope string, roleAssignmentName string, parameters authorization.RoleAssignmentCreateParameters) (authorization.RoleAssignment, error) {
-	ctx, _, done := tele.StartSpanWithLogger(ctx, "roleassignments.AzureClient.Create")
+// Get gets the specified role assignment by the role assignment name.
+func (ac *azureClient) Get(ctx context.Context, spec azure.ResourceSpecGetter) (interface{}, error) {
+	ctx, span := tele.Tracer().Start(ctx, "roleassignments.AzureClient.Get")
+	defer span.End()
+	roleAssignmentSpec := spec.(*RoleAssignmentSpec)
```
sonasingh46, line 67: @Jont828 -- In this `Get` `scope` is required. I had to cast here to get the scope. 

On a different note -- this get was previously not implemented and as of now we can keep it as a no-op by returning `nil, nil` until we do some more modification to async framework to fetch arbitrary fields. 

Or we can keep as it is now -- and change once the async framework have such provision of fetching arbitrary fields.

azure/services/roleassignments/client.go
sonasingh46: Will check on this part.

azure/scope/machine_test.go
```diff
@@ -407,19 +409,29 @@ func TestMachineScope_RoleAssignmentSpecs(t *testing.T) {
...
+				&roleassignments.RoleAssignmentSpec{
+					ResourceType: azure.VirtualMachine,
 					MachineName:  "machine-name",
 					Name:         "azure-role-assignment-name",
-					ResourceType: azure.VirtualMachine,
+					//ToDo(@sonasingh46): Use mock at this place
+					ResourceGroup:    "",
+					Scope:            azure.GenerateScope(""),
+					RoleDefinitionID: azure.GenerateContributorRoleDefinitionID(""),
+					PrincipalID:      to.StringPtr("fakePrincipalID"),
```
sonasingh46, line 427: As we now assign `ResourceGroup `, `Scope `, `RoleDefinitionID ` and `PrincipalID` in the machine scope itself. (azure/scope/machine.go:func (m *MachineScope) RoleAssignmentSpecs(principalID *string) []azure.ResourceSpecGetter ) -- we may need to mock these here but to being with I am passing an empty string.

azure/scope/machine.go
sonasingh46: Added the values here. Only downside is in machine scope unit testing. 
See this https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1873/files#r794443824

azure/services/roleassignments/roleassignments.go
sonasingh46: I had to make little change in the code of `reconcileVM` and `reconcileVMSS` -- > I renamed them as `getVMPrincipalID` and `getVMSSPrincipalID` as there were simply fetching principal id from the api. 

I am doing the following: 
1. Defined a `RoleAssignmentResourceType` method in the RoleAssignment scope which should return the resoource type ( vm or vmss ) 

2. If the resource type is `VM` -- call `getVMPrincipalID` that fetched the principal ID for VM case

3. If the resource type is `VMSS` -- call `getVMSSPrincipalID` that fetched the principal ID for VMSS case

Now, I am passing this `principalID` to the `RoleAssignmentSpecs` as an argument so that the value gets set at the  RoleAssignmentSpec in machine scope. 
Due to this -- I do not have access to get the ResourceGroupName .

Let me know what you think.

azure/services/roleassignments/client.go
Jont828: Yeah I don't think we can implement Get() currently without casting to get the scope. I think we can leave it as is now since it's used in the async code to see if an existing resource is there. Otherwise if it's a no-op, Parameters() would always get `existing == nil` and we'd be making a CreateOrUpdateAsync() call everytime. I think it would make sense to address this in a PR to add to the ResourceSpecGetter interface. Any thoughts @CecileRobertMichon?

azure/scope/machine_test.go
Jont828: I think we're going to want to set those values in the machine scope so we can make sure the fields are getting populated correctly. Check out [this file](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1939/files#diff-2a71aa3df4645e00fede1673e7b5294c50dd82516ae3f79bbb20ed601e6d9839R1338) for an example on how to do that. Also you're going to have to pass the subscription ID through `ClusterScope.AzureClients.EnvironmentSettings.Values` instead of as a normal machine scope field.

azure/services/roleassignments/client.go
Jont828: I think you should be able to since it's not imported anywhere and it's not needed in your refactor of Reconcile() and Delete(). You could try removing it to verify that it doesn't break anything ,and make sure to remove client_mock.go and the references to it in doc.go.

azure/defaults.go
```diff
@@ -97,6 +98,16 @@ func GenerateBackendAddressPoolName(lbName string) string {
 	return fmt.Sprintf("%s-%s", lbName, "backendPool")
 }
 
+// GenerateScope generates the scope name.
```
CecileRobertMichon, line 101:
```suggestion
// GenerateSubscriptionScope generates a role assignment scope that applies to all resources in the subscription.
```
to make it more descriptive on what kind of scope it's generating (especially since the word "scope" is overused in this codebase)

azure/scope/machine.go
```diff
@@ -289,17 +292,35 @@ func (m *MachineScope) DiskSpecs() []azure.ResourceSpecGetter {
...
+			Name:             m.AzureMachine.Spec.RoleAssignmentName,
+			MachineName:      m.Name(),
+			ResourceType:     azure.VirtualMachine,
+			ResourceGroup:    m.ResourceGroup(),
+			Scope:            azure.GenerateScope(m.SubscriptionID()),
+			RoleDefinitionID: azure.GenerateContributorRoleDefinitionID(m.SubscriptionID()),
+			PrincipalID:      principalID,
 		}
+		fmt.Println(roles[0])
```
CecileRobertMichon, line 307: remember to remove this once the PR is ready

azure/scope/machine.go
```diff
@@ -289,17 +292,35 @@ func (m *MachineScope) DiskSpecs() []azure.ResourceSpecGetter {
...
+// RoleAssignmentResourceType returns the role assignment resource type.
+func (m *MachineScope) RoleAssignmentResourceType() string {
+	if m.AzureMachine.Spec.Identity == infrav1.VMIdentitySystemAssigned {
+		return azure.VirtualMachine
+	}
+	return ""
+}
+
+// MachineName returns the machine name.
+func (m *MachineScope) MachineName() string {
```
CecileRobertMichon, line 322: why not just use `m.Name()`?

sonasingh46: This is actually an implementation of `RoleAssignmentScope` interface method. 

See this https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1873/files#diff-5480fcd7b8b0c3331bc1e5049598731653f0133de8a39be5553d1b6e49a0026dR43

I had to do this because -- the `Service` has the following: 
```
type Service struct {
	Scope RoleAssignmentScope
	client
	async.Reconciler
	virtualMachinesClient        virtualmachines.Client
	virtualMachineScaleSetClient scalesets.Client
}
```
Now this `Scope` in the `Service` struct is an interface of type `RoleAssignmentScope`. So to get access of machine name, I had to implement the method. Other way would have been a cast of the value but that seemed not appropriate. Open to suggestions.

sonasingh46: And machine name is required to fetch the VM for principal ID in the reconcile block.

azure/services/roleassignments/roleassignments.go
```diff
@@ -31,119 +32,111 @@ import (
...
-	for _, roleSpec := range s.Scope.RoleAssignmentSpecs() {
-		switch roleSpec.ResourceType {
-		case azure.VirtualMachine:
-			return s.reconcileVM(ctx, roleSpec)
-		case azure.VirtualMachineScaleSet:
-			return s.reconcileVMSS(ctx, roleSpec)
-		default:
-			return errors.Errorf("unexpected resource type %q. Expected one of [%s, %s]", roleSpec.ResourceType,
-				azure.VirtualMachine, azure.VirtualMachineScaleSet)
+	for _, roleAssignmentSpec := range s.Scope.RoleAssignmentSpecs(principalID) {
```
Jont828, line 95: This is looking really good! I like that we don't need to explicitly cast the spec here anymore.

test/e2e/azure_test.go
```diff
@@ -141,7 +142,7 @@ var _ = Describe("Workload cluster creation", func() {
 
 	if os.Getenv("LOCAL_ONLY") != "true" {
 		Context("Creating a private cluster", func() {
-			It("Creates a public management cluster in the same vnet", func() {
+			It("Creates a public management cluster in a custom vnet", func() {
```
CecileRobertMichon, line 145: this should already be part of https://github.com/kubernetes-sigs/cluster-api-provider-azure/commit/c4064ec43e554d0979eee47bb256fac469257895#diff-93ce65d81f5c1510135f6035d43f678bf9b5962b8e4ee45105e3345a1852efc6R145, rebase error?

azure/services/roleassignments/client.go
CecileRobertMichon: +1 to implementing this GET so we don't do a PUT for every Reconcile(). 

Make sure you handle the error when casting and return an error if `spec` is not a `RoleAssignmentSpec`.

azure/services/scalesets/scalesets_test.go
sonasingh46: I tried something like below: 
```
scopeMock := mock_scalesets.NewMockScaleSetScope(mockCtrl)
	rec:=scopeMock.EXPECT()
	rec.SubscriptionID().Return("123")
	rec.BaseURI().Return("/dummy")
	rec.Authorizer().Return(scopeMock)
	actual := NewService(scopeMock, resourceskus.NewStaticCache(nil, ""))
	g.Expect(actual).ToNot(BeNil())
```

I added those mock `Returns` as they were expected but at the end I got error something like this:

```	
wrong type of argument 0 to Return for *mock_scalesets.MockScaleSetScope.Authorizer: *mock_scalesets.MockScaleSetScope is not assignable to autorest.Authorizer [/Users/kuashutosh/go/src/sigs.k8s.io/cluster-api-provider-azure/azure/services/scalesets/scalesets_test.go:66]
```

azure/defaults.go
```diff
@@ -81,7 +81,8 @@ const (
 const (
 	// ProviderIDPrefix will be appended to the beginning of Azure resource IDs to form the Kubernetes Provider ID.
 	// NOTE: this format matches the 2 slashes format used in cloud-provider and cluster-autoscaler.
-	ProviderIDPrefix = "azure://"
+	ProviderIDPrefix          = "azure://"
+	azureBuiltInContributorID = "b24988ac-6180-42a0-ab88-20f7382dd24c"
```
CecileRobertMichon, line 85: could you please add a quick comment about what this is / where it's from? I remember you asking about it earlier and it might not be obvious to everyone

azure/services/roleassignments/client.go
```diff
@@ -50,18 +45,57 @@ func newRoleAssignmentClient(subscriptionID string, baseURI string, authorizer a
...
+}
+
+// Result fetches the result of a long-running operation future.
+func (ac *azureClient) Result(ctx context.Context, futureData azureautorest.FutureAPI, futureType string) (interface{}, error) {
+	// Result is a no-op for role assignment as only Delete operations return a future.
+	return nil, nil
+}
+
+func (ac *azureClient) DeleteAsync(ctx context.Context, spec azure.ResourceSpecGetter) (azureautorest.FutureAPI, error) {
+	// ToDo: Complete this function
```
CecileRobertMichon, line 99: is this still a todo?

azure/services/roleassignments/client.go
CecileRobertMichon: not sure I follow, why are we not using the passed in parameters?

azure/services/roleassignments/spec.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+// ResourceGroupName returns the name of the resource group.
+func (s *RoleAssignmentSpec) ResourceGroupName() string {
+	return s.ResourceGroup
+}
+
+// OwnerResourceName is a no-op for role assignments.
+func (s *RoleAssignmentSpec) OwnerResourceName() string {
+	return ""
```
CecileRobertMichon, line 50: is role assignment not a sub resource of the VM/VMSS ? what if this returned the scope?

azure/services/scalesets/scalesets.go
```diff
@@ -59,6 +58,9 @@ type (
 	}
 )
 
+// ScalesetsServiceName is the name of the scalesets service.
+const ScalesetsServiceName = "scalesets"
```
CecileRobertMichon, line 62: is this used outside the scaleset package? If not no need to make it public. Also since scalesets is the name of the package already should be avoid the stutter

```suggestion
const servicename = "scalesets"
```

azure/services/roleassignments/spec.go
sonasingh46: >is role assignment not a sub resource of the VM/VMSS ? what if this returned the scope?
If it returns the scope -- it basically returns -- `fmt.Sprintf("/subscriptions/%s/", subscriptionID)`

CecileRobertMichon:
/retest

looks like all tests are failing? retrying in case results are outdated

azure/services/roleassignments/roleassignments.go
```diff
@@ -18,132 +18,121 @@ package roleassignments
...
+	"sigs.k8s.io/cluster-api-provider-azure/util/reconciler"
 	"sigs.k8s.io/cluster-api-provider-azure/util/tele"
 )
 
-const azureBuiltInContributorID = "b24988ac-6180-42a0-ab88-20f7382dd24c"
+const serviceName = "roleassignments"
 
 // RoleAssignmentScope defines the scope interface for a role assignment service.
 type RoleAssignmentScope interface {
 	azure.ClusterDescriber
```
CecileRobertMichon, line 36: is azure.ClusterDescriber still used anywhere? If not, let's remove it. If it is and only for one or two methods, can we list the specific methods needed in the interface instead of including `azure.ClusterDescriber` which is pretty broad?

azure/services/roleassignments/spec.go
CecileRobertMichon: I would vote for returning the scope since it would really simplify the code and avoid having to cast parameters in the client, especially since we're not using `OwnerResourceName` for anything. Technically the scope isn't an "owning resource" but it does define what parent resource the role assignment applies to, and it's part of the role assignment resource ID, just like a parent resource: `PUT https://management.azure.com/scope/providers/Microsoft.Authorization/roleAssignments/roleAssignmentName?api-version=2015-07-01`

https://docs.microsoft.com/en-us/rest/api/authorization/role-assignments/create#create-role-assignment


We could even consider renaming `OwnerResourceName` to something that includes both "scope" and "owner resource name" if that improves code readability. At the very least we would add a code comment on this func to explain that it returns the scope for role assignments.

CecileRobertMichon: found some more official docs on this: https://docs.microsoft.com/en-us/azure/role-based-access-control/scope-overview#scope-and-arm-templates:

> A role assignment is a special type in Azure Resource Manager called an extension resource. An extension resource is a resource that adds to another resource's capabilities. They always exist as an extension (like a child) of another resource. For example, a role assignment at subscription scope is an extension resource of the subscription. The name of a role assignment is always the name of the resource you are extending plus /Microsoft.Authorization/roleAssignments/{roleAssignmentId}

sonasingh46: Sure. Returning scope from `OwnerResourceName` makes sense to me. I will make the changes accordingly.

azure/services/roleassignments/roleassignments.go
sonasingh46: `ResourceGroup()` is what is used right now. 
I could have got rid of the `ClusterDescriber` in the `RoleAssignmentScope` interface but had to use due to the following reason:

- Reconcile RoleAssignment needs principal ID. To get the principal ID, an api call needs to be done which needs resource group. Now we can use `RoleAssignmentSpecs` to fetch the resource group but `RoleAssignmentSpecs` needs `PrincipalID`

CecileRobertMichon: Can we just replace `azure.ClusterDescriber` by `ResourceGroup()` then? If we just need that one func might as well be explicit instead of including the entire `ClusterDescriber` interface which is pretty broad

sonasingh46: Done

sonasingh46:
@CecileRobertMichon @Jont828 -- I have incorporated the comments and fixed the nil pointer issue. Please take a look.

azure/services/roleassignments/roleassignments.go
```diff
@@ -67,92 +68,85 @@ func (s *Service) Name() string {
...
 	defer done()
+	ctx, cancel := context.WithTimeout(ctx, reconciler.DefaultAzureServiceReconcileTimeout)
+	defer cancel()
+	log.V(2).Info("reconciling role assignment")
+
+	// Return early if there is no roleassignment spec to reconcile.
+	if len(s.Scope.RoleAssignmentSpecs(to.StringPtr(""))) < 1 {
+		log.V(2).Info("no role assignment spec to reconcile")
+		return nil
+	}
```
sonasingh46, line 81: This is a redundant call to role assignments spec that I had to make so that when we fetch principalID -- nil point exception does not occur. 

When there is not role assignment spec, while fetching the vm details, some of fields are nil and looks more simpler to return early in case there is no role assignment spec.

So this first call passes an empty principal ID and validates if role assignment spec is present and then below finds the principal ID and calls this spec function once again to finally get the correct role assignment specs. 
cc @CecileRobertMichon @Jont828

shysank: I'd suggest adding another method to scope that just returns true/false if the machine/machinepool has system assigned identity for better readability.

CecileRobertMichon: > I'd suggest adding another method to scope that just returns true/false if the machine/machinepool has system assigned identity for better readability.

+1

sonasingh46: Done

azure/services/roleassignments/spec.go
```diff
@@ -0,0 +1,67 @@
...
+// Parameters returns the parameters for the RoleAssignmentSpec.
+func (s *RoleAssignmentSpec) Parameters(existing interface{}) (interface{}, error) {
+	if existing != nil {
+		if _, ok := existing.(authorization.RoleAssignment); !ok {
+			return nil, errors.Errorf("%T is not a authorization.RoleAssignment", existing)
+		}
+		// RoleAssignmentSpec already exists
+		return nil, nil
+	}
+	return authorization.RoleAssignmentCreateParameters{
```
CecileRobertMichon, line 61: We shouldn't need `RoleAssignmentCreateParameters` anymore now that we don't need to include the scope, you should be able to just return `RoleAssignmentProperties` and get rid of extra step to cast it in client.go

CecileRobertMichon: my bad I misread this and thought it was still returning the custom type we had before...

CecileRobertMichon:
@sonasingh46 can you please squash?

shysank:
@sonasingh46 were you able to find the cause of failure for the vmss e2e test?

sonasingh46:
@shysank -- Not sure why it failed but when I rebased and pushed again the test passed.

PR #1874, opened 2021-11-18
Refactor async service to use an async reconciler interface and handle getting existing resources

### LINKED ISSUE #1871: Refactor Get existing logic into CreateResource() async helper
Currently, we're implementing the same logic to get the existing resource for each service in the client CreateOrUpdateAsync function that has been refactored by #1702. We should try to extract that logic to CreateResource() so it can be shared by all the services, have better unit testing. That will also ensure the clients have minimal business logic and stay as close to wrappers of the SDK as possible.

Jont828:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**: This PR refactors the async service into an implementation of an asynchronous reconciler interface. It also refactors the logic for fetching the existing resource into `CreateResource`, simplifying `CreateOrUpdateAsync`, which is meant to wrap the Azure client calls and not do a lot of heavy computation. The unit tests for the existing async services were also refactored and simplified using a mock of the async reconciler interface.

Fixes #1871 

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Add async reconciler interface and refactor async service to get existing resources
```


Jont828:
@CecileRobertMichon Here's an idea I was experimenting with for refactoring the logic for getting an existing resource into a helper. Do you think this is in the right direction?

azure/services/async/async.go
```diff
@@ -68,6 +68,36 @@ func processOngoingOperation(ctx context.Context, scope FutureScope, client Futu
...
 
+// GetExistingResource implements the logic for creating a resource.
+func GetExistingResource(ctx context.Context, scope FutureScope, client Creator, spec azure.ResourceSpecGetter, serviceName string) (interface{}, error) {
+	ctx, _, done := tele.StartSpanWithLogger(ctx, "async.Service.GetExistingResource")
+	defer done()
+
+	resourceName := spec.ResourceName()
+	rgName := spec.ResourceGroupName()
+
+	// TODO: should Get operations be blocked by a long running operation?
```
CecileRobertMichon, line 79: no need to check for long running operations here, if we reach this phase it means there is already no running operation

azure/services/async/async.go
```diff
@@ -68,6 +68,36 @@ func processOngoingOperation(ctx context.Context, scope FutureScope, client Futu
 	return result, err
 }
 
+// GetExistingResource implements the logic for creating a resource.
+func GetExistingResource(ctx context.Context, scope FutureScope, client Creator, spec azure.ResourceSpecGetter, serviceName string) (interface{}, error) {
```
CecileRobertMichon, line 72: does this need to be exported? If it's just a helper for CreateResource, it should start with a lower case

azure/services/async/async.go
```diff
@@ -68,6 +68,36 @@ func processOngoingOperation(ctx context.Context, scope FutureScope, client Futu
 	return result, err
 }
 
+// GetExistingResource implements the logic for creating a resource.
```
CecileRobertMichon, line 71: comment copy paste

azure/services/async/async.go
```diff
@@ -68,6 +68,36 @@ func processOngoingOperation(ctx context.Context, scope FutureScope, client Futu
...
+
+	var existingResource interface{}
+
+	if existing, err := client.Get(ctx, spec); err != nil && !azure.ResourceNotFound(err) {
+		return nil, errors.Wrapf(err, "failed to get existing resource %s/%s (service: %s)", rgName, resourceName, serviceName)
+	} else if err == nil {
+		existingResource = existing
+	}
+
+	scope.V(2).Info("successfully got existing resource", "service", serviceName, "resource", resourceName, "resourceGroup", rgName)
```
CecileRobertMichon, line 97: let's only print the success log if the error was nil

azure/services/vnetpeerings/client.go
```diff
@@ -61,29 +61,21 @@ func newPeeringsClient(subscriptionID string, baseURI string, authorizer autores
...
-	var existingPeering interface{}
-
-	if existing, err := ac.Get(ctx, spec.ResourceGroupName(), spec.OwnerResourceName(), spec.ResourceName()); err != nil && !azure.ResourceNotFound(err) {
-		return nil, nil, errors.Wrapf(err, "failed to get virtual network peering %s for %s in %s", spec.ResourceName(), spec.OwnerResourceName(), spec.ResourceGroupName())
-	} else if err == nil {
-		existingPeering = existing
-	}
-
-	params, err := spec.Parameters(existingPeering)
+	params, err := spec.Parameters(existingResource)
```
CecileRobertMichon, line 78: I wonder if we could go even further and move this logic up in async.go too, then `CreateOrUpdateAsync` can become ` CreateOrUpdateAsync(ctx context.Context, spec azure.ResourceSpecGetter, parameters interface{})`

azure/services/vnetpeerings/client.go
```diff
@@ -61,29 +61,21 @@ func newPeeringsClient(subscriptionID string, baseURI string, authorizer autores
 }
 
 // Get gets the specified virtual network peering by the peering name, virtual network, and resource group.
-func (ac *AzureClient) Get(ctx context.Context, resourceGroupName, vnetName, peeringName string) (network.VirtualNetworkPeering, error) {
+func (ac *AzureClient) Get(ctx context.Context, spec azure.ResourceSpecGetter) (interface{}, error) {
 	ctx, span := tele.Tracer().Start(ctx, "vnetpeerings.AzureClient.Get")
 	defer span.End()
 
-	return ac.peerings.Get(ctx, resourceGroupName, vnetName, peeringName)
+	return ac.peerings.Get(ctx, spec.ResourceGroupName(), spec.OwnerResourceName(), spec.ResourceGroupName())
```
CecileRobertMichon, line 68: 👌

azure/services/vnetpeerings/client.go
Jont828: Would we be able to do the type check for params, i.e. `natRule, ok := params.(network.InboundNatRule)` if we move this part into the interface?

CecileRobertMichon: no that part would stay in here but you could do the nil check in there, ie.:

async.go:

```
if existing, err := client.Get(ctx, spec); err != nil && !azure.ResourceNotFound(err) {
		return nil, errors.Wrapf(err, "failed to get existing resource %s/%s (service: %s)", rgName, resourceName, serviceName)
	} else if err == nil {
		existingResource = existing
	}

params, err := spec.Parameters(existingResource)
if err != nil {
		return nil, nil, errors.Wrapf(err, "failed to get desired parameters for resource %s/%s (service: %s)", rgName, resourceName, serviceName)
} else if params == nil {
   // nothing to do, return existing, don't call CreateOrUpdateAsync
}
```

and in here:

```
func (ac *AzureClient) CreateOrUpdateAsync(ctx context.Context, spec azure.ResourceSpecGetter, parameters interface{}) (interface{}, azureautorest.FutureAPI, error) {
...
peering, ok := parameters.(network.VirtualNetworkPeering)
if !ok {
    return nil, nil, errors.Errorf("%T is not a network.VirtualNetworkPeering", params)
}

future, err := ac.peerings.CreateOrUpdate(ctx, spec.ResourceGroupName(), spec.OwnerResourceName(), spec.ResourceName(), peering, network.SyncRemoteAddressSpaceTrue)

...
```

Jont828:
@CecileRobertMichon Just pushed some more changes. If that looks good, I can refactor all the current async services on main to use this logic and update the tests for them as well.

azure/services/vnetpeerings/client.go
```diff
@@ -61,40 +61,23 @@ func newPeeringsClient(subscriptionID string, baseURI string, authorizer autores
...
 
 // CreateOrUpdateAsync creates or updates a virtual network peering asynchronously.
 // It sends a PUT request to Azure and if accepted without error, the func will return a Future which can be used to track the ongoing
 // progress of the operation.
-func (ac *AzureClient) CreateOrUpdateAsync(ctx context.Context, spec azure.ResourceSpecGetter) (interface{}, azureautorest.FutureAPI, error) {
+func (ac *AzureClient) CreateOrUpdateAsync(ctx context.Context, spec azure.ResourceSpecGetter, parameters interface{}) (interface{}, azureautorest.FutureAPI, error) {
 	ctx, span := tele.Tracer().Start(ctx, "vnetpeerings.AzureClient.CreateOrUpdateAsync")
 	defer span.End()
 
-	var existingPeering interface{}
```
CecileRobertMichon, line 78: beautiful 👌

azure/services/async/async.go
```diff
@@ -82,9 +110,21 @@ func CreateResource(ctx context.Context, scope FutureScope, client Creator, spec
 		return processOngoingOperation(ctx, scope, client, resourceName, serviceName)
 	}
 
-	// No long running operation is active, so create the resource.
+	// No long running operation is active, so get the existing resource.
```
CecileRobertMichon, line 113: IMO the previous comment still makes sense, we're still creating the resource. Getting the existing resource to check if anything needs to be updated before doing a create is an implementation detail

maybe something like
```suggestion
	// No long-running operation is active, so get or create the resource.
```
makes sense?

azure/services/async/async.go
```diff
@@ -68,6 +68,34 @@ func processOngoingOperation(ctx context.Context, scope FutureScope, client Futu
 	return result, err
 }
 
+// getExistingResourceWithParameters is a helper that implements the logic for getting an a resource and constructing the parameters.
+func getExistingResourceWithParameters(ctx context.Context, scope FutureScope, client Creator, spec azure.ResourceSpecGetter, serviceName string) (interface{}, interface{}, error) {
```
CecileRobertMichon, line 72: this is one of those situations where it might not make sense to have a helper function. This function is very specific to the CreateResource function above, takes all the same function parameters as input, and returns everything as output, which makes the code a bit awkward. It might make more sense if we were only returning the parameters as output, but here we also need the existing resource to use as result if parameters is nil. Consider moving the logic directly into `CreateResource`, or finding another way to breakout the code into smaller funcs if that makes the function too big / hard to read.

azure/services/async/async.go
```diff
@@ -82,9 +110,21 @@ func CreateResource(ctx context.Context, scope FutureScope, client Creator, spec
...
+	// No long running operation is active, so get the existing resource.
+	existingResource, parameters, err := getExistingResourceWithParameters(ctx, scope, client, spec, serviceName)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to create resource %s/%s (service: %s)", rgName, resourceName, serviceName)
+	} else if parameters == nil && existingResource != nil {
+		// Nothing to do, return existing and don't call CreateOrUpdateAsync
+		scope.V(2).Info("resource up to date", "service", serviceName, "resource", resourceName, "resourceGroup", rgName)
+		return existingResource, nil
+	} else {
+		errors.Errorf("failed to create resource %s/%s (service: %s): no parameters and no existing resource found", rgName, resourceName, serviceName)
```
CecileRobertMichon, line 122: this isn't actually an error?

Jont828: I don't think we'd ever reach that block if everything is working correctly. If there are no parameters, then we'd have an existing resource that's up to date, and if we have parameters, then we'd be able to create or update using that. If both `parameters` and `existingResource` are `nil`, then it seems like we'd have violated the specs of the function in our code. Does that sound right to you, or is there a better way to handle that?

CecileRobertMichon: Right, there are really just 3 cases:

if err != nil {
     // something went wrong, return an error
} else if parameters == nil {
    // nothing to do, return existing resource 
}

// otherwise, keep going and create the resource

CecileRobertMichon:
@Jont828 once the PR is ready for review and tests are passing, please squash commits to reflect only distinct logical changes and update the PR title/description to remove WIP and reflect the current state of the PR (ie. the other changes that were made besides get existing logic)

azure/services/async/async.go
```diff
@@ -69,61 +85,83 @@ func processOngoingOperation(ctx context.Context, scope FutureScope, client Futu
...
 	if future != nil {
-		return processOngoingOperation(ctx, scope, client, resourceName, serviceName)
+		return processOngoingOperation(ctx, s.Scope, s.Creator, resourceName, serviceName)
 	}
 
 	// No long running operation is active, so create the resource.
-	scope.V(2).Info("creating resource", "service", serviceName, "resource", resourceName, "resourceGroup", rgName)
-	result, sdkFuture, err := client.CreateOrUpdateAsync(ctx, spec)
+
+	// Check if the resource already exists. If it does and matches the spec, we can return early.
```
CecileRobertMichon, line 103: > If it does and matches the spec, we can return early.

this is not always true, it is up to the `Parameters()` func to decide.

```suggestion
	// Get the resource if it already exists, and use it to construct the desired resource parameters.
```

azure/services/async/async.go
```diff
@@ -69,61 +85,83 @@ func processOngoingOperation(ctx context.Context, scope FutureScope, client Futu
...
+	parameters, err := spec.Parameters(existingResource)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to get desired parameters for resource %s/%s (service: %s)", rgName, resourceName, serviceName)
+	} else if parameters == nil {
+		// Nothing to do, return existing and don't call CreateOrUpdateAsync
+		s.Scope.V(2).Info("resource up to date", "service", serviceName, "resource", resourceName, "resourceGroup", rgName)
+		return existingResource, nil
+	}
+
+	// Now we can create the resource using the spec and the existing resource.
```
CecileRobertMichon, line 123:
```suggestion
	// Create or Update the resource with the desired parameters.
```

azure/services/async/async.go
```diff
@@ -69,61 +85,83 @@ func processOngoingOperation(ctx context.Context, scope FutureScope, client Futu
...
+		existingResource = existing
+		s.Scope.V(2).Info("successfully got existing resource", "service", serviceName, "resource", resourceName, "resourceGroup", rgName)
+	}
+
+	// Construct parameters using the resource spec and information from the existing resource, if there is one.
+	parameters, err := spec.Parameters(existingResource)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to get desired parameters for resource %s/%s (service: %s)", rgName, resourceName, serviceName)
+	} else if parameters == nil {
+		// Nothing to do, return existing and don't call CreateOrUpdateAsync
```
CecileRobertMichon, line 118:
```suggestion
		// Nothing to do, don't create or update the resource and return the existing resource.
```

CecileRobertMichon (approved):
/lgtm
/assign @shysank @mboersma @devigned 

This will make it a lot easier to switch further services to async + unit test them, and ensure we always try to get the existing resource before creating it.

azure/services/groups/groups_test.go
```diff
@@ -44,16 +43,8 @@ var (
 		ClusterName:    "test-cluster",
 		AdditionalTags: map[string]string{"foo": "bar"},
 	}
-	internalError  = autorest.NewErrorWithResponse("", "", &http.Response{StatusCode: 500}, "Internal Server Error")
-	notFoundError  = autorest.NewErrorWithResponse("", "", &http.Response{StatusCode: 404}, "Not Found")
-	errCtxExceeded = errors.New("ctx exceeded")
```
mboersma, line 49: I assume `errCtxExceeded` and `fakeFuture` are unused?

azure/services/async/interfaces.go
```diff
@@ -41,11 +41,18 @@ type FutureHandler interface {
 // Creator is a client that can create or update a resource asynchronously.
 type Creator interface {
 	FutureHandler
-	CreateOrUpdateAsync(ctx context.Context, spec azure.ResourceSpecGetter) (interface{}, azureautorest.FutureAPI, error)
+	CreateOrUpdateAsync(ctx context.Context, spec azure.ResourceSpecGetter, existingResource interface{}) (interface{}, azureautorest.FutureAPI, error)
+	Get(ctx context.Context, spec azure.ResourceSpecGetter) (interface{}, error)
```
mboersma, line 45: It's arguably odd to have `Get` in a `Creator` interface, but I'm ok with it here.

mboersma (approved):
/lgtm

This seems like a consistent refactoring. Nice work!

Jont828:
Sounds good, thanks for the reviews! I'm going to merge the async disks PR and rebase this before merging.

CecileRobertMichon:
/hold

for rebase after #1711 merges

azure/services/async/async.go
```diff
@@ -29,6 +29,22 @@ import (
 	"sigs.k8s.io/cluster-api-provider-azure/util/tele"
 )
 
+// Service is an implementation of the AsyncReconciler interface. It handles asynchronous creation and deletion of resources.
```
CecileRobertMichon, line 32:
```suggestion
// Service is an implementation of the Reconciler interface. It handles asynchronous creation and deletion of resources.
```

since it was renamed

azure/services/disks/client.go
```diff
@@ -30,19 +30,12 @@ import (
 )
 
 // client wraps go-sdk.
```
CecileRobertMichon, line 32: remove the comment?

azure/services/disks/client.go
```diff
@@ -30,19 +30,12 @@ import (
...
-}
 
 // azureClient contains the Azure go-sdk Client.
 type azureClient struct {
 	disks compute.DisksClient
 }
 
-var _ client = (*azureClient)(nil)
-
 // newClient creates a new VM Client from subscription ID.
```
CecileRobertMichon, line 39:
```suggestion
// newClient creates a new disk Client from subscription ID.
 ```

not your PR but while you're here...

File: azure/services/disks/mock_disks/client_mock.go [hunk 19]
```diff
@@ -14,6 +14,54 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+/*
```
CecileRobertMichon, line 17: what's going on here?

azure/services/disks/mock_disks/doc.go
```diff
@@ -15,7 +15,6 @@ limitations under the License.
 */
 
 // Run go generate to regenerate this mock.
-//go:generate ../../../../hack/tools/bin/mockgen -destination client_mock.go -package mock_disks -source ../client.go client
 //go:generate ../../../../hack/tools/bin/mockgen -destination disks_mock.go -package mock_disks -source ../disks.go DiskScope
 //go:generate /usr/bin/env bash -c "cat ../../../../hack/boilerplate/boilerplate.generatego.txt client_mock.go > _client_mock.go && mv _client_mock.go client_mock.go"
```
CecileRobertMichon, line 19: probably need to remove this line too

azure/services/disks/mock_disks/client_mock.go
CecileRobertMichon: this file should be deleted

azure/services/groups/groups.go
```diff
@@ -108,10 +112,15 @@ func (s *Service) IsGroupManaged(ctx context.Context) (bool, error) {
 	defer done()
 
 	groupSpec := s.Scope.GroupSpec()
-	group, err := s.client.Get(ctx, groupSpec.ResourceName())
+	groupIface, err := s.client.Get(ctx, groupSpec)
```
CecileRobertMichon, line 115: you might be able to replace this with `s.Reconciler.Get` or `s.Reconciler.Creator.Get` (I think first one should work) and get rid of the client too

azure/services/disks/client.go
CecileRobertMichon: this is still unresolved

CecileRobertMichon:
@devigned @mboersma PTAL when you get a chance. Hoping to merge this one soon so we can rebase and unblock a bunch of the open async services PRs

azure/services/virtualmachines/client.go
```diff
@@ -35,8 +35,8 @@ import (
 // Client wraps go-sdk.
 type (
 	Client interface {
```
CecileRobertMichon, line 37: do we still need the Client interface here?

CecileRobertMichon:
@Jont828 #1812 merged so you need to rebase this one, the conflicts should be too complex, just make sure you use the prefered logging mechanism from #1812 everywhere

test/e2e/data/infrastructure-azure/v1alpha4/cluster-template-kcp-adoption.yaml
```diff
@@ -0,0 +1,347 @@
+apiVersion: v1
```
CecileRobertMichon, line 1: these files shouldn't be part of the PR, were they removed from .gitignore?

azure/services/async/async.go
CecileRobertMichon: looks like the rename was reverted

CecileRobertMichon (commented):
/lgtm
/assign @devigned

devigned (approved):
The refactor is pretty clean. However, I'm really looking forward to Go 1.18 and generics to clean up the `interface{}` usage.

Solid work, @Jont828.

PR #1767, opened 2021-10-07
Add workload cluster upgrades spec

CecileRobertMichon:
**What type of PR is this?**

**What this PR does / why we need it**: Adds upgrade spec based on CAPI cluster upgrade test framework.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #1735 

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

PR #1879, opened 2021-11-19
e2e: Only check for valid image if version is a semver version

CecileRobertMichon:
**What type of PR is this?**

**What this PR does / why we need it**: Fixes a regression from #1767 that broke tests using `latest-x.x` as k8s version https://testgrid.k8s.io/provider-azure-master-signal#capz-conformance

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/e2e_suite_test.go
+++ b/test/e2e/e2e_suite_test.go
@@ -482,6 +482,7 @@ func parseImageSkuNames(skus []string) semver.Versions {
 func getLatestSkuForMinor(version string, skus semver.Versions) string {
 	isStable, match := validateStableReleaseString(version)
 	if isStable {
+		// if the version is in the format "stable-1.21", we find the latest 1.21.x version.
 		major, err := strconv.ParseUint(match[1], 10, 64)
 		Expect(err).NotTo(HaveOccurred())
 		minor, err := strconv.ParseUint(match[2], 10, 64)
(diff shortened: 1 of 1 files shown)


test/e2e/e2e_suite_test.go
```diff
@@ -493,7 +493,7 @@ func getLatestSkuForMinor(version string, skus semver.Versions) string {
 				break
 			}
 		}
-	} else {
+	} else if strings.HasPrefix(version, "v") {
```
jsturtevant, line 496: What are the possible values for version here?  

v1.21.0
latest-1.21

CecileRobertMichon: it could be either `vx.y.z` (eg. v1.21.0), `latest` or `latest-x.y` (eg. latest-1.21)

https://capz.sigs.k8s.io/developers/development.html#conformance-testing

If it's a latest- version (or anything else that could be added in the future), we don't check for a valid semver and an existing VM image and instead keep the previous behavior.

CecileRobertMichon: actually, wait

should I just change it back to what I had here originally: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1767#discussion_r749648252

/hold

CecileRobertMichon: /hold cancel

@jsturtevant check it out, I think this is better

devigned (approved):
/lgtm
/approve

test/e2e/e2e_suite_test.go
jsturtevant: the comments help alot thanks

test/e2e/e2e_suite_test.go
```diff
@@ -493,11 +494,11 @@ func getLatestSkuForMinor(version string, skus semver.Versions) string {
 				break
 			}
 		}
-	} else {
-		v, err := semver.ParseTolerant(version)
-		Expect(err).NotTo(HaveOccurred())
+	} else if v, err := semver.ParseTolerant(version); err == nil {
```
jsturtevant, line 497: parse tolerant allows for versions with or without a v in it.  I think this is ok here but just a note

https://pkg.go.dev/github.com/blang/semver#ParseTolerant

CecileRobertMichon: Yes, this is the only way to parse a version with a "v" prefix (which we use in k8s versions). `Parse` won't allow for a "v" prefix.

PR #1886, opened 2021-11-23
Make load balancer reconcile/delete async

### LINKED ISSUE #1713: async loadbalancers

Jont828:
**What type of PR is this?**

/kind feature
**What this PR does / why we need it**: Implementation of an async service for load balancers as a follow up for #1610 and #1541.

Fixes #1713

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Make load balancer reconcile/delete async
```


Jont828:
/hold
Waiting to rebase off of #1870 since that makes the inbound NAT rule service independent of the load balancer client.

Jont828:
Hold until #1870 is done to rebase off of that branch

Jont828:
@CecileRobertMichon @mboersma I rebased this off the inbound NAT PR and added tests, would appreciate some feedback when you have some time!

CecileRobertMichon:
/ok-to-test

sorry for the delay, this is on my list to review next week!

Also pasting this here:

> comment from @shysank about using named returns from another PR that is relevant here:
> #1686 (comment)
> Here is the commit that does this in the route tables PR: b2302ea

devigned (commented):
Code / approach look good. I left one comment about an outstanding todo. 

I would like to review this again after this is rebased on the nat rules PR.

azure/scope/cluster.go
```diff
@@ -164,9 +165,10 @@ func (s *ClusterScope) PublicIPSpecs() []azure.PublicIPSpec {
 }
 
 // LBSpecs returns the load balancer specs.
-func (s *ClusterScope) LBSpecs() []azure.LBSpec {
-	specs := []azure.LBSpec{
-		{
+func (s *ClusterScope) LBSpecs() []azure.ResourceSpecGetter {
+	// TODO: update the getter spec for the LBs
```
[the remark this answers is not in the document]
Jont828: Ah thanks, I forgot about that.

CecileRobertMichon:
PR has been rebased 

/hold cancel

azure/services/loadbalancers/spec.go
```diff
@@ -0,0 +1,313 @@
...
+			}
+		}
+
+		if !update {
+			// load balancer already exists with all required defaults
+			return nil, nil
+		}
+	}
+
+	frontendIPConfigs, frontendIDs = getFrontendIPConfigs(*s)
```
CecileRobertMichon, line 133: wouldn't this override the existing values if the LB exists and we just want to update it? 

Before the refactor there was a `default` switch statement around this which meant we'd only set these values if existing was nil and it hadn't gone into the above `if existing != nil` logic https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1886/files#diff-85bedcd738722859960dbecef775a3b135f6a27726c7c13005c29b3f07473628L135

azure/services/loadbalancers/spec_test.go
```diff
@@ -0,0 +1,335 @@
...
+			name:     "node outbound load balancer exists with all expected values",
+			spec:     &fakeNodeOutboundLBSpec,
+			existing: newDefaultNodeOutboundLB(),
+			expect: func(g *WithT, result interface{}) {
+				g.Expect(result).To(BeNil())
+			},
+			expectedError: "",
+		},
+		{
+			name:     "load balancer exists with missing load balancing rules",
```
CecileRobertMichon, line 96: If my comment above makes sense, let's check why this test case didn't catch the values being overwritten and fix it so it would have (and make sure it then passes after fixing the issue)

Jont828: I think your comment about it being overwritten makes sense. I believe the reason it wasn't failing was because all the fields are replaced w/ the same values since the spec is still the same.

azure/services/loadbalancers/client.go
```diff
@@ -54,61 +50,129 @@ func newLoadBalancersClient(subscriptionID string, baseURI string, authorizer au
...
 	if err != nil {
-		return err
+		return false, errors.Wrap(err, "failed checking if the operation was complete")
+	}
+
+	return isDone, nil
+}
+
+// Result fetches the result of a long-running operation future.
+func (ac *azureClient) Result(ctx context.Context, future azureautorest.FutureAPI, futureType string) (interface{}, error) {
```
CecileRobertMichon, line 148:
```suggestion
func (ac *azureClient) Result(ctx context.Context, future azureautorest.FutureAPI, futureType string) (result interface{}, err error) {
```

azure/services/loadbalancers/spec.go
```diff
@@ -0,0 +1,313 @@
...
+func getFrontendIPConfigs(lbSpec LBSpec) ([]network.FrontendIPConfiguration, []network.SubResource) {
+	frontendIPConfigurations := make([]network.FrontendIPConfiguration, 0)
+	frontendIDs := make([]network.SubResource, 0)
+	for _, ipConfig := range lbSpec.FrontendIPConfigs {
+		var properties network.FrontendIPConfigurationPropertiesFormat
+		if lbSpec.Type == infrav1.Internal {
+			properties = network.FrontendIPConfigurationPropertiesFormat{
+				PrivateIPAllocationMethod: network.IPAllocationMethodStatic,
+				Subnet: &network.Subnet{
+					// TODO: is the VNet resource group the same as the cluster resource group?
```
CecileRobertMichon, line 170: vnet RG is not always the same as the cluster RG since you can BYO pre-existing vnet. Can this TODO be removed?

CecileRobertMichon (commented):
lgtm besides two small nits

azure/scope/cluster.go
```diff
@@ -165,11 +166,17 @@ func (s *ClusterScope) PublicIPSpecs() []azure.PublicIPSpec {
 }
 
 // LBSpecs returns the load balancer specs.
-func (s *ClusterScope) LBSpecs() []azure.LBSpec {
-	specs := []azure.LBSpec{
-		{
+func (s *ClusterScope) LBSpecs() []azure.ResourceSpecGetter {
```
CecileRobertMichon, line 169: /lgtm cancel

whoops I just realized right after approving, we also need to add LoadBalancersReadyCondition to the list of owned conditions

azure/scope/machine.go
```diff
@@ -544,6 +544,7 @@ func (m *MachineScope) PatchObject(ctx context.Context) error {
 		conditions.WithConditions(
 			infrav1.VMRunningCondition,
 			infrav1.AvailabilitySetReadyCondition,
+			infrav1.LoadBalancersReadyCondition,
```
CecileRobertMichon, line 547: should be in the cluster.go patch object, not machine

CecileRobertMichon (commented):
/lgtm

PR #1890, opened 2021-11-24
Derive API server LB DNS name from user-defined private DNS zone name

### LINKED ISSUE #1885: If the cluster is configured with a private DNS zone name, this name is not used to derive the API server host
Support for a private DNS zone name was added in https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1289.

Although the private DNS zone name _is_ used to [create the private DNS zone](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/0579d7139f550b432f4912d60f91daaa0f4490c0/azure/scope/cluster.go#L360), the name is _not_ used to [derive the API server host]( 
https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/0579d7139f550b432f4912d60f91daaa0f4490c0/azure/scope/cluster.go#L639-L645). Instead, the API server host is [derived from the default private zone DNS name `.capz.io`](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/8508da989832c571b86d2a2fbe7a6e1ae59df327/azure/defaults.go#L137-L140). 

/kind bug

dlipovetsky:
**What type of PR is this?**
/kind bug

If the user defines a private DNS zone name, they expect the API server LB DNS name to use that zone. Prior to this change, the API server LB DNS name always used the default private DNS zone (`capz.io`).

Fixes #1885 

I'm not yet familiar with the code base, so the fix and/or tests might be off-track. Constructive criticism welcome :smile: 

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
The user-defined private DNS zone name is now used to derive the cluster's API server load balancer DNS name. 
```

### DIFF
--- a/azure/defaults.go
+++ b/azure/defaults.go
@@ -139,9 +139,9 @@ func GeneratePrivateDNSZoneName(clusterName string) string {
 	return fmt.Sprintf("%s.capz.io", clusterName)
 }
 
-// GeneratePrivateFQDN generates FQDN for a private API Server.
-func GeneratePrivateFQDN(clusterName string) string {
-	return fmt.Sprintf("%s.%s", PrivateAPIServerHostname, GeneratePrivateDNSZoneName(clusterName))
+// GeneratePrivateFQDN generates the FQDN for a private API Server based on the private DNS zone name.
+func GeneratePrivateFQDN(zoneName string) string {
+	return fmt.Sprintf("%s.%s", PrivateAPIServerHostname, zoneName)
 }
 
 // GenerateVNetLinkName generates the name of a virtual network link name based on the vnet name.
--- a/azure/scope/cluster.go
+++ b/azure/scope/cluster.go
@@ -639,7 +639,7 @@ func (s *ClusterScope) APIServerPort() int32 {
 // APIServerHost returns the hostname used to reach the API server.
 func (s *ClusterScope) APIServerHost() string {
 	if s.IsAPIServerPrivate() {
-		return azure.GeneratePrivateFQDN(s.ClusterName())
+		return azure.GeneratePrivateFQDN(s.GetPrivateDNSZoneName())
 	}
 	return s.APIServerPublicIP().DNSName
 }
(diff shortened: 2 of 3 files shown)


dlipovetsky:
```
*** Running go-apidiff ***

sigs.k8s.io/cluster-api-provider-azure/azure
  Incompatible changes:
  - GeneratePrivateFQDN: removed
+ EXIT_VALUE=1
+ set +o xtrace
```

I removed that function because its only argument (`clusterName`) is not enough to derive the FQDN, and the function is not a `ClusterScope` receiver, it needs to accept the private zone DNS name as an argument.

I could leave the function, but it would become (a) dead code, and (b) an incorrect implementation.

Open to suggestions!

azure/defaults.go
```diff
@@ -139,11 +139,6 @@ func GeneratePrivateDNSZoneName(clusterName string) string {
 	return fmt.Sprintf("%s.capz.io", clusterName)
 }
 
-// GeneratePrivateFQDN generates FQDN for a private API Server.
-func GeneratePrivateFQDN(clusterName string) string {
```
CecileRobertMichon, line 143: instead of removing the func, could we change the parameters to take the dns zone name instead of the cluster name? Then it would be used in cluster.go like this:

```
azure.GeneratePrivateFQDN(s.GetPrivateDNSZoneName())
```

IMO it's slightly better than directly constructing the name with `fmt.Sprintf` in cluster.go

dlipovetsky: Good point. Fixed, thanks!

CecileRobertMichon:
lgtm

@dlipovetsky please squash commits!

dlipovetsky:
I don't think the e2e failure is related to the change.

On a separate, and not really important note, the apidiff test _passed_. I guess that's because  `GeneratePrivateFQDN` changed its argument name, but not its type. That's a false negative, but I guess the only way to make the check accurate here is to use a specific type, not a generic `string`, for the argument.)

PR #1896, opened 2021-11-29
separate AKS control-plane and agentpool versions

### LINKED ISSUE #1883: Support control plane only upgrades for AKS
/kind feature

When upgrading the `AzureManagedControlPlane` it upgrades the Agentpools also present in AKS cluster. 
If a user has specified the version in MachinePool template spec, the `AzureManagedMachinPool`controller tries to downgrade the agentpool. 

We should use the default `AzureManagedControlPlane` upgrade as `control plane only` And let the machine pool be upgraded separately. 
Azure cli exposes an option through the [flag `--control-plane-only` to upgrade ](https://docs.microsoft.com/en-us/azure/architecture/operator-guides/aks/aks-upgrade-practices)the control plane only. 

error seen after upgrading the cluster
```
1 controller.go:317] controller/azuremanagedmachinepool "msg"="Reconciler error" "error"="error creating AzureManagedMachinePool cluster-registry/agentpool0: failed to reconcile machine pool agentpool0: failed to create or update agent pool: failed to begin operation: containerservice.AgentPoolsClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code=\"AgentPoolUpgradeVersionNotAllowed\" Message=\"Upgrade agent pool version from 1.22.2 to 1.21.2 is not allowed, you can only upgrade one minor version at a time. Please use [az aks get-upgrades] command to get the supported upgrade paths. For more information, please check https://aka.ms/version-skew-policy\"" "name"="agentpool0" "namespace"="cluster-registry" "reconciler group"="infrastructure.cluster.x-k8s.io" "reconciler kind"="AzureManagedMachinePool"
```
**Environment:**

- cluster-api-provider-azure version: 1.0.0

zmalik:
**What type of PR is this?**
/kind feature

If `MachinePool` `.spec.template.spec.version` is specified, this PR honours the version for nodepools and also separates the control plane upgrades from the agentpool upgrades. 

Fixes #1883

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```release-note
control-plane-only upgrades for AKS
```

### DIFF
--- a/azure/scope/managedcontrolplane.go
+++ b/azure/scope/managedcontrolplane.go
@@ -27,6 +27,7 @@ import (
 	"github.com/Azure/go-autorest/autorest"
 	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/pkg/errors"
+	"golang.org/x/mod/semver"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
@@ -509,6 +510,14 @@ func (s *ManagedControlPlaneScope) GetAgentPoolSpecs(ctx context.Context) ([]azu
 
 		if ownerPool.Spec.Replicas != nil {
 			ammp.Replicas = *ownerPool.Spec.Replicas
+		}
+
+		if ownerPool.Spec.Template.Spec.Version != nil {
+			version := *ownerPool.Spec.Template.Spec.Version
+			if semver.Compare(version, s.ControlPlane.Spec.Version) > 0 {
+				return nil, errors.New("MachinePool version cannot be greater than the AzureManagedControlPlane version")
+			}
+			ammp.Version = to.StringPtr(strings.TrimPrefix(version, "v"))
 		}
 
 		ammps = append(ammps, ammp)
--- a/azure/services/managedclusters/managedclusters.go
+++ b/azure/services/managedclusters/managedclusters.go
@@ -233,15 +233,16 @@ func (s *Service) Reconcile(ctx context.Context) error {
 	for i := range managedClusterSpec.AgentPools {
 		pool := managedClusterSpec.AgentPools[i]
 		profile := containerservice.ManagedClusterAgentPoolProfile{
-			Name:              &pool.Name,
-			VMSize:            &pool.SKU,
-			OsDiskSizeGB:      &pool.OSDiskSizeGB,
-			Count:             &pool.Replicas,
-			Type:              containerservice.AgentPoolTypeVirtualMachineScaleSets,
-			VnetSubnetID:      &managedClusterSpec.VnetSubnetID,
-			Mode:              containerservice.AgentPoolMode(pool.Mode),
-			AvailabilityZones: &pool.AvailabilityZones,
-			MaxPods:           pool.MaxPods,
+			Name:                &pool.Name,
+			VMSize:              &pool.SKU,
+			OsDiskSizeGB:        &pool.OSDiskSizeGB,
+			Count:               &pool.Replicas,
+			Type:                containerservice.AgentPoolTypeVirtualMachineScaleSets,
+			VnetSubnetID:        &managedClusterSpec.VnetSubnetID,
+			Mode:                containerservice.AgentPoolMode(pool.Mode),
+			AvailabilityZones:   &pool.AvailabilityZones,
+			MaxPods:             pool.MaxPods,
+			OrchestratorVersion: pool.Version,
 		}
 		*managedCluster.AgentPoolProfiles = append(*managedCluster.AgentPoolProfiles, profile)
 	}
@@ -313,6 +314,10 @@ func (s *Service) Reconcile(ctx context.Context) error {
 			// AKS.
 			existingMC.NetworkProfile.LoadBalancerProfile.EffectiveOutboundIPs = nil
 		}
+
+		// Avoid changing agent pool profiles through AMCP and just use the existing agent pool profiles
+		// AgentPool changes are managed through AMMP
+		managedCluster.AgentPoolProfiles = existingMC.AgentPoolProfiles
 
 		diff := computeDiffOfNormalizedClusters(managedCluster, existingMC)
 		if diff != "" {
--- a/azure/scope/managedcontrolplane_test.go
+++ b/azure/scope/managedcontrolplane_test.go
@@ -130,6 +130,135 @@ func TestManagedControlPlaneScope_Autoscaling(t *testing.T) {
 	}
 }
 
+func TestManagedControlPlaneScope_PoolVersion(t *testing.T) {
+	scheme := runtime.NewScheme()
+	_ = capiv1exp.AddToScheme(scheme)
+	_ = infrav1.AddToScheme(scheme)
+
+	cases := []struct {
+		Name     string
+		Input    ManagedControlPlaneScopeParams
+		Expected []azure.AgentPoolSpec
+		Err      string
+	}{
+		{
+			Name: "Without Version",
+			Input: ManagedControlPlaneScopeParams{
+				AzureClients: AzureClients{
+					Authorizer: autorest.NullAuthorizer{},
+				},
+				Cluster: &clusterv1.Cluster{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "cluster1",
+						Namespace: "default",
+					},
+				},
+				ControlPlane: &infrav1.AzureManagedControlPlane{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "cluster1",
+						Namespace: "default",
+					},
+					Spec: infrav1.AzureManagedControlPlaneSpec{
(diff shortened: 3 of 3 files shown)


azure/services/managedclusters/managedclusters.go
```diff
@@ -169,11 +169,11 @@ func (s *Service) Reconcile(ctx context.Context) error {
 	// clusters API at create time, not update.
 	if azure.ResourceNotFound(err) {
 		isCreate = true
-		// Add system agent pool to cluster spec that will be submitted to the API
-		managedClusterSpec.AgentPools, err = s.Scope.GetAgentPoolSpecs(ctx)
-		if err != nil {
-			return errors.Wrapf(err, "failed to get system agent pool specs for managed cluster %s", s.Scope.ClusterName())
-		}
+	}
```
alexeldeib, line 172: can you clarify the thinking here? shouldn't this still happen only on create?

if not, a code comment with the reasoning would be nice.

alexeldeib:
based on your [comment ](https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1883#issuecomment-981827129), I think we may want additional webhook validation in AMMP or reconciler to enforce AMMP version <= AMCP version? what do you think?

azure/services/managedclusters/managedclusters.go
zmalik: I was seeing empty agentpoolprofiles on updates. I will add a comment. Without this the ManagedCluster was setting orchestrator version empty for the agentpoolprofiles and upgrading them with the managed controlplane version

I will add a comment.

zmalik:
> I think we may want additional webhook validation in AMMP or reconciler to enforce AMMP version <= AMCP version? what do you think?

yes, we should def do it. Let me take a stab on that

azure/services/managedclusters/managedclusters.go
alexeldeib: got it, thanks. 

I think we might need to be careful here -- it's possible for a user to add AMMP, and for that to cause a race between AMCP and AMMP reconcile. if AMCP wins, we'll end up adding a new agentpool via AKS ManagedClusters API which will return a 400. so you may want to do a GET and make sure you only PUT existing agentpools on update.

I guess that might cause a race in the other direction, where we do 1) GET on AKS 2) AMMP creates new pool 3) AMCP reconciles old pools and we deleted one that AMMP just created so we get a 400...

I think either way it's possible to get a 400 and kind of hard to avoid due to the upgrade issue you mentioned...I'm fine with either solution for now (likely as-is, given it's less code and no additional changes). I might have to check/ask around on the AKS side if there's a better way to handle this.

alexeldeib: cc @devigned in case you have some brainy ideas on avoiding this

zmalik: thanks for the review!
> make sure you only PUT existing agentpools on update

good idea. What about if there is a mismatch between GET on AKS and AMMP count/name etc, throw an error and wait till it is same? This way we let AMMP manage the creation or deletion

azure/scope/managedcontrolplane.go
```diff
@@ -507,7 +508,11 @@ func (s *ManagedControlPlaneScope) GetAgentPoolSpecs(ctx context.Context) ([]azu
 		}
 
 		if ownerPool.Spec.Template.Spec.Version != nil {
-			ammp.Version = to.StringPtr(strings.TrimPrefix(*ownerPool.Spec.Template.Spec.Version, "v"))
+			version := *ownerPool.Spec.Template.Spec.Version
+			if semver.Compare(version, s.ControlPlane.Spec.Version) > 0 {
+				return nil, errors.New("MachinePool version cannot be greater than the AzureManagedMachinePool version")
```
zmalik, line 513: will fix the typo in next set of changes

azure/services/managedclusters/managedclusters.go
alexeldeib: Ideally if the cluster is capi managed the only mismatch should be in number of pools. I guess the pool config could change but again, we wouldn't want to modify that on cluster PUT. 

so we might need to treat 400s here as a retryable error and let re-reconcile fix it, which isn't ideal, but should work.

devigned: I would prefer if we were able to differentiate between a 400 that is caused by an agent pool add/delete vs a different 400. 

400 is usually a client error, which requires the client to make a different request. That would mean it would generally be a terminal error, rather than transient.

zmalik: I reverted the above change. And just use later `existingMC.AgentPoolProfiles` for the PUT which is coming from `GET on AKS`

alexeldeib: > I would prefer if we were able to differentiate between a 400 that is caused by an agent pool add/delete vs a different 400.

I believe AKS does categorize the error in some useful way, I can check the message and get back to you.

> 400 is usually a client error, which requires the client to make a different request. That would mean it would generally be a terminal error, rather than transient.

true, but on re-reconcile we would in fact make a different request due to the race I mentioned, which would be valid. 

>I reverted the above change. And just use later existingMC.AgentPoolProfiles for the PUT which is coming from GET on AKS

unfortunately this can still hit the race in the other direction I mentioned before. I'm fine with the behavior for this PR as long as we don't consider a 400 in MC reconciler as terminal. It'd be nice to open an issue to track that separately if we have this PR go in with the logic as it is.

zmalik: > I believe AKS does categorize the error in some useful way, I can check the message and get back to you.

I will wait for this and not consider this particular error as non-terminal.
to consider it non-terminal it would be wrapping it with `azure.WithTransientError` explicitly right?

jackfrancis (approved):
lgtm

CecileRobertMichon:
@zmalik looks like #1910 merged just before this one got in, can you please rebase?

zmalik:
rebased and squashed all in one commit also

azure/scope/managedcontrolplane.go
```diff
@@ -511,6 +512,14 @@ func (s *ManagedControlPlaneScope) GetAgentPoolSpecs(ctx context.Context) ([]azu
 			ammp.Replicas = *ownerPool.Spec.Replicas
 		}
 
+		if ownerPool.Spec.Template.Spec.Version != nil {
```
jackfrancis, line 515: We should really have a UT for this (useful!) validation failure

CecileRobertMichon: agreed, @zmalik would you be able to add a short UT for this? Sorry for not suggesting that earlier!

zmalik: sounds good! added the test now

CecileRobertMichon:
linter is failing with

```
azure/scope/managedcontrolplane_test.go:257: unnecessary trailing newline (whitespace)
		})
azure/scope/managedcontrolplane_test.go:250:7: ineffectual assignment to err (ineffassign)
			s, err := NewManagedControlPlaneScope(context.TODO(), c.Input)
```

zmalik:
my bad, fixed the linter issues

azure/services/managedclusters/managedclusters.go
```diff
@@ -314,6 +315,10 @@ func (s *Service) Reconcile(ctx context.Context) error {
 			existingMC.NetworkProfile.LoadBalancerProfile.EffectiveOutboundIPs = nil
 		}
 
+		// Avoid changing agent pool profiles through AMCP and just use the existing agent pool profiles
+		// AgentPool changes are managed through AMMP
+		managedCluster.AgentPoolProfiles = existingMC.AgentPoolProfiles
```
jackfrancis, line 320: Does this thread summarize the discussion around the above addition?

https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1896#pullrequestreview-818380859

@alexeldeib are we good with this as-is?

(Everything else in the PR lgtm)

CecileRobertMichon: @zmalik are you able to answer Jack's question above?

zmalik: yes, we agreed to send the AgentPoolProfiles as they come from Azure API and not use AMCP reconcile to update agentpoolprofiles but only use AMMP reconcile to update the agentpools. 

In the current solution, this can give us a corner race condition where while AMMP is adding an agentPool and AMCP doesn't get it from the Azure API and tries to delete it. This returns an error as per @alexeldeib but we need AKS team to tell us which error it would be. And return a transient error in that case. 
Right now it's returning a normal error and would reconcile again where Azure API will return the newly added agentpool.

I will open a separate PR to handle this corner case as soon as we know what error is returned.

CecileRobertMichon: Thanks! Can you please open an issue and assign you & Ace for the corner case so we remember to circle back after the holiday?

zmalik: yes! created https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1937

Issue #1840, opened 2021-11-08
Panic in VMSS cluster e2e spec


mboersma:
/kind bug

**What steps did you take and what happened:**

A memory access error sometimes causes a panic when e2e tests are running. [Here](https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/periodic-cluster-api-provider-azure-e2e-full-main/1455783680084545536) is one example.

<details>
  <summary>Stack Trace:</summary>
<pre>
•! Panic [1461.804 seconds]
Workload cluster creation
/home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/test/e2e/azure_test.go:43
  Creating a VMSS cluster
  /home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/test/e2e/azure_test.go:333
    with a single control plane node and an AzureMachinePool with 2 Linux and 2 Windows worker nodes [It]
    /home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/test/e2e/azure_test.go:334
    Test Panicked
    runtime error: invalid memory address or nil pointer dereference
    /usr/local/go/src/runtime/panic.go:212
    Full Stack Trace
    sigs.k8s.io/cluster-api-provider-azure/test/e2e.labelNodesWithMachinePoolName(0x2583200, 0xc000126018, 0x259dc40, 0xc0007dc000, 0xc000f9a720, 0x19, 0xc000f1e000, 0x3, 0x4)
    	/home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/test/e2e/azure_machinepool_drain.go:202 +0xc3
    sigs.k8s.io/cluster-api-provider-azure/test/e2e.testMachinePoolCordonAndDrain(0x2583200, 0xc000126018, 0x25a1810, 0xc0005cdb10, 0x25a1880, 0xc000a9b580, 0xc00121d420, 0x10, 0xc000cb9f80, 0x27, ...)
    	/home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/test/e2e/azure_machinepool_drain.go:147 +0x432
    sigs.k8s.io/cluster-api-provider-azure/test/e2e.AzureMachinePoolDrainSpec(0x2583200, 0xc000126018, 0xc0000e4cc8)
    	/home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/test/e2e/azure_machinepool_drain.go:89 +0x798
[...]

PR #1898, opened 2021-11-30
remove usage of realpath in scripts

CecileRobertMichon:
**What type of PR is this?**

**What this PR does / why we need it**: related to #1192 

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,8 +69,9 @@ CONVERSION_GEN_VER := v0.22.2
 CONVERSION_GEN_BIN := conversion-gen
 CONVERSION_GEN := $(TOOLS_BIN_DIR)/$(CONVERSION_GEN_BIN)-$(CONVERSION_GEN_VER)
 
+ENVSUBST_VER := v2.0.0-20210730161058-179042472c46
 ENVSUBST_BIN := envsubst
-ENVSUBST := $(TOOLS_BIN_DIR)/$(ENVSUBST_BIN)-drone
+ENVSUBST := $(TOOLS_BIN_DIR)/$(ENVSUBST_BIN)-$(ENVSUBST_VER)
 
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -69,8 +69,9 @@ CONVERSION_GEN_VER := v0.22.2
 CONVERSION_GEN_BIN := conversion-gen
 CONVERSION_GEN := $(TOOLS_BIN_DIR)/$(CONVERSION_GEN_BIN)-$(CONVERSION_GEN_VER)
 
+ENVSUBST_VER := latest
```
jsturtevant, line 72: there isn't a release but we may want to pin to the same version that capi uses to avoid any differences that might occur between ci and tilt setups: https://github.com/kubernetes-sigs/cluster-api/blob/6eb2002ed15f73570ed14c83daceb55d18b2f939/hack/tools/go.mod#L7

scripts/ci-conformance.sh
```diff
@@ -23,19 +23,15 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
+# Install kubectl
 REPO_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
-cd "${REPO_ROOT}" || exit 1
+KUBECTL="hack/tools/bin/kubectl"
```
jsturtevant, line 28: I realize we are cd'ing to repo_root right after, but if a CD happens else where in the script (maybe added in the future) the path could be wrong.  If we prepend this with repo_root we could avoid that type of mistake in future.

CecileRobertMichon:
ran into #1840 

/retest

PR #1690, opened 2021-09-15
Remove test regions where GPU SKU isn't available

mboersma:
**What type of PR is this?**

/kind failing-test

**What this PR does / why we need it**:

Fixes (one of) the problems causing the GPU-enabled cluster tests to fail in [testgrid e2e](https://testgrid.k8s.io/sig-cluster-lifecycle-cluster-api-provider-azure#capz-postsubmit-capi-e2e-full-main) by removing three regions where capz controller manager logs show the preferred, less expensive SKU isn't available:

```
09/08 message: failed to reconcile AzureMachine: failed to create network interface: reconcile error that cannot be recovered occurred: failed to get SKU Standard_NV6 in compute api: resource sku with name 'Standard_NV6' and category 'virtualMachines' not found in location 'canadacentral'. Object will not be requeued
09/09 message: failed to reconcile AzureMachine: failed to create network interface: reconcile error that cannot be recovered occurred: failed to get SKU Standard_NV6 in compute api: resource sku with name 'Standard_NV6' and category 'virtualMachines' not found in location 'francecentral'. Object will not be requeued
09/13 message: 	message: failed to reconcile AzureMachine: failed to create network interface: reconcile error that cannot be recovered occurred: failed to get SKU Standard_NV6 in compute api: resource sku with name 'Standard_NV6' and category 'virtualMachines' not found in location 'westus'. Object will not be requeued
```

**Which issue(s) this PR fixes**:

Refs #1648

**Special notes for your reviewer**:

I wanted to move to the next-gen GPU SKU in this PR since `Standard_NV6` is being retired in a year, but getting quota will take some time and maneuvering, so this seems like the simplest change to help.

This doesn't cover it all, but should slow the bleeding.

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

PR #1905, opened 2021-12-06
Add more e2e regions to spread quota usage

CecileRobertMichon:
**What type of PR is this?**

**What this PR does / why we need it**: Add a few more regions to the list of test regions so we can spread the load a bit as we have more tests running now with different release branch tests + all periodic jobs. I verified that the test subscription has quota in the added regions.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/scripts/ci-conformance.sh
+++ b/scripts/ci-conformance.sh
@@ -48,7 +48,7 @@ source "${REPO_ROOT}/hack/util.sh"
 : "${AZURE_CLIENT_SECRET:?Environment variable empty or not defined.}"
 
 get_random_region() {
-    local REGIONS=("eastus" "eastus2" "northcentralus" "northeurope" "uksouth" "westeurope" "westus2")
+    local REGIONS=("canadacentral" "eastus" "eastus2" "northcentralus" "northeurope" "southeastasia" "uksouth" "westeurope" "westus" "westus2")
     echo "${REGIONS[${RANDOM} % ${#REGIONS[@]}]}"
 }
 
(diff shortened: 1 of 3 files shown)


mboersma:
In #1690 I removed `canadacentral` and `westus` because they didn't offer the `Standard_NV6` GPU SKU that e2e tests use by default. I think that's still the case, but things are complicated by the fact that this SKU is deprecated and doesn't show up in `az vm list-skus` even in regions where it is still supported currently.

We could consider having a separate random region function for just the GPU test, although that sounds clunky. We need to move to another, supported SKU before summer of 2022, but getting quota is difficult.

CecileRobertMichon:
/retest

no region related issue

CecileRobertMichon:
> We could consider having a separate random region function for just the GPU test, although that sounds clunky. We need to move to another, supported SKU before summer of 2022, but getting quota is difficult.

@mboersma the GPU test has actually been failing https://testgrid.k8s.io/sig-cluster-lifecycle-cluster-api-provider-azure#capz-periodic-e2e-full-main, not sure if it's a SKU issue though

mboersma:
> the GPU test has actually been failing

It doesn't appear to be a SKU issue in testgrid, because the GPU nodes have boot logs, etc. I don't see the problem yet, but I'll dig into it.

For this PR, I'm pretty sure adding those two regions guarantees failure for the GPU tests if they're selected. We could add separate logic for finding a random GPU-enabled region, but really we need to use a new GPU SKU and ensure quota for it. This will probably introduce a different set of compatible/incompatible regions, but IIRC the new recommended type is more widely available.

mboersma:
@CecileRobertMichon I'm still looking at the GPU failure and thinking about how to transition us to [`Standard_NV12s_v3`](https://docs.microsoft.com/en-us/azure/virtual-machines/nv-series-migration-guide#nvsv3-series). Hopefully we can arrive at a configuration that includes more regions in the end.

PR #1815, opened 2021-11-01
add support for autoscaling in azuremanagedmachinepools

zmalik:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:
Adds support for Autoscaling in the Managed clusters (AKS). 
**Which issue(s) this PR fixes**:
Fixes # https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1669

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
added support for cluster-autoscaler for AKS 
```

PR #1910, opened 2021-12-09
add MaxPods configuration for AKS node pools

jackfrancis:
**What type of PR is this?**

/kind feature

This PR enables MaxPods configuration for AKS cluster (AzureManagedCluster) resources. Depending on the size of the cluster (especially for large clusters), you may want to tune down your kubelet's --max-pods configuration across a node pool to account for the limitation of IP addresses in an Azure VNET (large AKS clusters currently require Azure CNI, which uses real Azure VNET IP addresses to assign to pods).

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
add MaxPods configuration for AKS node pools
```


exp/api/v1beta1/azuremanagedmachinepool_types.go
```diff
@@ -65,6 +65,10 @@ type AzureManagedMachinePoolSpec struct {
 	// Scaling specifies the autoscaling parameters for the node pool.
 	// +optional
 	Scaling *ManagedMachinePoolScaling `json:"scaling,omitempty"`
+
+	// MaxPods specifies the kubelet --max-pods configuration for the node pool.
+	// +optional
+	MaxPods *int32 `json:"maxPods,omitempty"`
```
CecileRobertMichon, line 71: you'll also need to add this new type to conversion from v1alpha3 and v1alpha4 to preserve the field during round trip conversions, see an example here: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1815/files#diff-45673dfe0beba334668369fcf88dcfb9a531245efe19cfb6d413e9bb6420e04dR41

templates/cluster-template-aks.yaml
```diff
@@ -65,6 +65,7 @@ metadata:
   name: agentpool0
   namespace: default
 spec:
+  maxPods: 24
```
CecileRobertMichon, line 68: is this value for testing? should we leave it empty if it's not required and let the default be whatever AKS defaults to?

exp/api/v1beta1/azuremanagedmachinepool_webhook.go
```diff
@@ -99,6 +99,10 @@ func (r *AzureManagedMachinePool) ValidateUpdate(oldRaw runtime.Object, client c
 		}
 	}
 
+	// TODO is there a need to do MaxPods validation here?
+	// It's not obvious from kubelet code that there is an enforced "valid range",
+	// But perhaps AKS enforces its own limits
```
CecileRobertMichon, line 104: not sure, @alexeldeib thoughts?

azure/scope/managedcontrolplane_test.go
```diff
@@ -130,6 +130,102 @@ func TestManagedControlPlaneScope_Autoscaling(t *testing.T) {
...
+					Spec: infrav1.AzureManagedControlPlaneSpec{
+						SubscriptionID: "00000000-0000-0000-0000-000000000000",
+					},
+				},
+				MachinePool:      getMachinePool("pool0"),
+				InfraMachinePool: getAzureMachinePool("pool0", infrav1.NodePoolModeSystem),
+				PatchTarget:      getAzureMachinePool("pool0", infrav1.NodePoolModeSystem),
+			},
+			Expected: azure.AgentPoolSpec{
+
```
devigned, line 169: please remove this blank line or maybe just gofmt

templates/cluster-template-aks.yaml
jackfrancis: It's for testing, yes

exp/api/v1beta1/azuremanagedmachinepool_webhook.go
LochanRn: The maxPods range actually varies based on the [networkPlugin](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/26b3a43dcfd794c60bbe3736cddfd6ff09be7d70/exp/api/v1beta1/azuremanagedcontrolplane_types.go#L73) configured by the user. 

https://docs.microsoft.com/fi-fi/azure/aks/configure-azure-cni#configure-maximum---new-clusters

So it's a bit tricky to do this. If we have to validate it, we first need to get the amcp object to check the networkPlugin configured.

jackfrancis: I wouldn't want to block progress on solving this (based on (1) the precedent of the existing rather absent input validations for this cluster type and (2) the experimental nature of the AzureManagedCluster surface area).

But should we keep the // TODO comment as a reminder to revisit it later when it's appropriate to reinforce this stuff?

azure/scope/managedcontrolplane_test.go
jackfrancis: copy pasta extra profit

exp/api/v1beta1/azuremanagedmachinepool_webhook.go
LochanRn: I think we should validate for a range of 10-250, the aks api validates for the same. 
I tried this for the both kubenet and azure and returns the same :) 
```
lochanrn@Lochans-MacBook-Pro ~ % az aks create -g rnl-rg -n aks-test --max-pods 300 --network-plugin kubenet
(MaxPodsLimitExceeded) The AgentPoolProfile 'nodepool1' has an invalid maxPods, the maxPods(300) should be smaller than or equal to 250. Please refer to aka.ms/aks-min-max-pod for more details.
Code: MaxPodsLimitExceeded
Message: The AgentPoolProfile 'nodepool1' has an invalid maxPods, the maxPods(300) should be smaller than or equal to 250. Please refer to aka.ms/aks-min-max-pod for more details.
Target: agentPoolProfile.kubernetesConfig.kubeletConfig.maxPods
```

During the update we need to ensure it is [immutable](https://docs.microsoft.com/fi-fi/azure/aks/configure-azure-cni#configure-maximum---existing-clusters)

alexeldeib: I basically agree with @LochanRn's assessment, but I agree it's not a blocker given the existing validation. I do think we should go for the cross object validation eventually, or else have a simple validation on < 250 (the higher of the two limits) for now

jackfrancis: Thanks for the feedback, I've added validations that match the current AKS requirements.

LochanRn:
The PR looks good to me :) 
Can you please add some docs as well regarding the maxPods option.

Thanks 😊

jackfrancis:
@LochanRn docs added, PTAL

jackfrancis:
@LochanRn docs added, PTAL

docs/book/src/topics/managedcluster.md
```diff
@@ -136,6 +136,7 @@ spec:
   mode: System
   osDiskSizeGB: 512
   sku: Standard_D2s_v3
+  maxPods: 24
```
LochanRn, line 139:
```suggestion
```
nit: This section does not have any optional field, would be great to follow the same here wdyt ??

docs/book/src/topics/managedcluster.md
```diff
@@ -162,6 +163,7 @@ spec:
   mode: User
   osDiskSizeGB: 1024
   sku: Standard_D2s_v4
+  maxPods: 24
```
LochanRn, line 166:
```suggestion
```

same as above

docs/book/src/topics/managedcluster.md
jackfrancis: done

docs/book/src/topics/managedcluster.md
jackfrancis: done

LochanRn:
/lgtm 

over to @CecileRobertMichon and @alexeldeib 


Nice Job 👍  
Thanks 😊

templates/cluster-template-aks.yaml
CecileRobertMichon, line 68: does this need to be re-generated?

CecileRobertMichon: can we remove it now that the tests have run? or do we want to add them to the test template specifically but not the reference template maybe?

jackfrancis: I don't think so (correct me if I'm wrong), these changes are the result of running `make generate-flavors` after updating `templates/flavors/aks/cluster-template.yaml`.

jackfrancis: I have a slight preference for keeping them in to get regular E2E validation on this configuration input. If it breaks existing E2E patterns, happy to remove.

CecileRobertMichon: if the goal is to represent a best practice value to users in our published release flavor, it should be in here. If the goal is just to test some arbitrary value regularly in e2e so the functionality doesn't regress, it should be a patch in https://github.com/kubernetes-sigs/cluster-api-provider-azure/tree/main/templates/test/ci/prow-aks-multi-tenancy so that it only gets added to https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/templates/test/ci/cluster-template-prow-aks-multi-tenancy.yaml

jackfrancis: Done in latest change.

templates/test/ci/patches/maxpods-aks.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
```
CecileRobertMichon, line 1: sorry if I'm being annoying but... can you please move this to a `patches` directory under https://github.com/kubernetes-sigs/cluster-api-provider-azure/tree/main/templates/test/ci/prow-aks-multi-tenancy? that's the current pattern for other flavors, eg. https://github.com/kubernetes-sigs/cluster-api-provider-azure/tree/main/templates/test/ci/prow-custom-vnet/patches, it's just tags that's an exception because it's shared between all the flavors

PR #1914, opened 2021-12-10
Make subnets reconcile/delete async

### LINKED ISSUE #1708: async subnets

Jont828:
**What type of PR is this?**

/kind feature
**What this PR does / why we need it**: Implementation of an async service for subnets as a follow up for #1610 and #1541.

Fixes #1708

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Make subnets reconcile/delete async
```


CecileRobertMichon:
comment from @shysank about using named returns from another PR that is relevant here:
#1686 (comment)

Here is the commit that does this in the route tables PR: b2302ea

azure/scope/cluster.go
```diff
@@ -462,6 +464,20 @@ func (s *ClusterScope) SetNatGatewayIDInSubnets(name string, id string) {
 	}
 }
 
+// UpdateSubnetCIDRs updates the subnet CIDRs for the subnet with the same name.
+func (s *ClusterScope) UpdateSubnetCIDRs(name string, cidrBlocks []string) {
```
mboersma, line 468: This isn't a complex function, but would it still be worth adding a unit test for it in `cluster_test.go`?

Same question for `UpdateSubnetID` below.

azure/scope/managedcontrolplane.go
```diff
@@ -265,6 +268,18 @@ func (s *ManagedControlPlaneScope) SetSubnet(_ infrav1.SubnetSpec) {
 	// no-op
 }
 
+// UpdateSubnetCIDRs updates the subnet CIDRs for the subnet with the same name.
+// This is not used when using a managed control plane.
+func (s *ManagedControlPlaneScope) UpdateSubnetCIDRs(_ string, _ []string) {
```
mboersma, line 273: You could write a unit test for this function that tested that it didn't mutate `s` (or parsed its AST to see that its body was actually a no-op). But that's probably not worth the effort. 😄

azure/scope/cluster.go
Jont828: It seems like we don't have unit tests for the other getter/setter functions like `Subnet()` or `SetSubnet()`. In that case, do you think we should add tests for all of them? Or just leave it be for now?

mboersma: I think we can just leave it be then. 👍🏻

azure/scope/managedcontrolplane.go
mboersma: But let's not. :-p

azure/scope/cluster.go
```diff
@@ -266,20 +267,18 @@ func (s *ClusterScope) NSGSpecs() []azure.NSGSpec {
...
 // SubnetSpecs returns the subnets specs.
-func (s *ClusterScope) SubnetSpecs() []azure.SubnetSpec {
-	numberOfSubnets := len(s.AzureCluster.Spec.NetworkSpec.Subnets)
-	if s.AzureCluster.Spec.BastionSpec.AzureBastion != nil {
-		numberOfSubnets++
-	}
-
-	subnetSpecs := make([]azure.SubnetSpec, 0, numberOfSubnets)
+func (s *ClusterScope) SubnetSpecs() []azure.ResourceSpecGetter {
+	subnetSpecs := []azure.ResourceSpecGetter{}
```
devigned, line 271: It looks like you know the number of specs you are going to create. Perhaps, preallocate the slice using `make([]azure.ResourceSpecGetter, len(s.AzureCluster.Spec.NetworkSpec.Subnets))` or `make([]azure.ResourceSpecGetter, 0, len(s.AzureCluster.Spec.NetworkSpec.Subnets) + 1)`.

azure/services/subnets/spec.go
```diff
@@ -0,0 +1,109 @@
...
+		if len(s.CIDRs) == 0 { // empty slice or nil?
+			// This is a bit of a workaround. By returning, we force a CreateOrUpdate call, which will trigger
+			// UpdateSubnetSpec() to be called. That will update the infrav1 spec, and those changes will be
+			// reflected in this spec when SubnetSpecs() is called.
+			return existingSubnet, nil
+		}
+
+		return nil, nil
+	} else if !s.IsVNetManaged {
+		// TODO: should this be a terminal error to break out of the reconcile loop?
```
devigned, line 74: I think it seems reasonable to return a terminal error here. WDYT, @CecileRobertMichon.

azure/services/subnets/spec.go
```diff
@@ -0,0 +1,109 @@
...
+		return nil, errors.Errorf("custom vnet was provided but subnet %s is missing", s.Name)
+	}
+	subnetProperties := network.SubnetPropertiesFormat{
+		AddressPrefixes: &s.CIDRs,
+	}
+
+	// workaround needed to avoid SubscriptionNotRegisteredForFeature for feature Microsoft.Network/AllowMultipleAddressPrefixesOnSubnet.
```
devigned, line 81: When reconciling we could check to see if the feature is registered. If it isn't, report the error to the user. However, I don't think we do this elsewhere, and the error from the HTTP request will likely have that content.

Maybe we need to have condition associated to lacking feature registration. WDYT, @CecileRobertMichon?

devigned (changes requested):
Just a couple comments and an allocation nit. Outside of that, lgtm.

azure/services/subnets/spec.go
Jont828: Sounds good, I opened #1973. I can open a separate PR to support terminal errors in the AzureCluster controller and add a terminal error here.

azure/services/subnets/spec.go
Jont828: Opened issue #1975 so we can come back to this.

azure/services/bastionhosts/azurebastion.go
```diff
@@ -69,7 +79,7 @@ func (s *Service) ensureAzureBastion(ctx context.Context, azureBastionSpec azure
 						Name: to.StringPtr(bastionHostIPConfigName),
 						BastionHostIPConfigurationPropertiesFormat: &network.BastionHostIPConfigurationPropertiesFormat{
 							Subnet: &network.SubResource{
-								ID: subnet.ID,
+								ID: existingSubnet.ID,
```
CecileRobertMichon, line 82: if just using the subnet to grab it's ID we can use the `SubnetID` helper function to construct the ID from the sub ID, rg, vnet name and subnet name and get rid of that extra GET call

azure/services/subnets/spec.go
```diff
@@ -0,0 +1,109 @@
...
+// Parameters returns the parameters for the subnet.
+func (s *SubnetSpec) Parameters(existing interface{}) (parameters interface{}, err error) {
+	if existing != nil {
+		existingSubnet, ok := existing.(network.Subnet)
+		if !ok {
+			return nil, errors.Errorf("%T is not a network.Subnet", existing)
+		}
+
+		// If the subnet infrav1 spec has no CIDRs but the existing subnet has CIDRs, update the spec
+		if len(s.CIDRs) == 0 { // empty slice or nil?
```
CecileRobertMichon, line 65: > // empty slice or nil?

They are equivalent:
https://play.golang.com/p/Kmk6-p2SAna

azure/services/subnets/spec.go
```diff
@@ -0,0 +1,109 @@
...
+			// This is a bit of a workaround. By returning, we force a CreateOrUpdate call, which will trigger
+			// UpdateSubnetSpec() to be called. That will update the infrav1 spec, and those changes will be
+			// reflected in this spec when SubnetSpecs() is called.
```
CecileRobertMichon, line 68: not sure I'm following what's going on here

when are we expecting this situation to arise?

Jont828: This is meant to replicate the behavior that was previously in the Reconcile loop. Before, if the subnet existed, we would get the resource and use it to update the ID and CIDRBlocks in the infrav1 subnet spec.

```
existingSubnet, err := s.getExisting(ctx, s.Scope.Vnet().ResourceGroup, subnetSpec)
switch {
case err != nil && !azure.ResourceNotFound(err):
    return errors.Wrapf(err, "failed to get subnet %s", subnetSpec.Name)
case err == nil:
    // subnet already exists, update the spec and skip creation
    s.Scope.SetSubnet(*existingSubnet)
    continue
```
```
// getExisting provides information about an existing subnet.
func (s *Service) getExisting(ctx context.Context, rgName string, spec azure.SubnetSpec) (*infrav1.SubnetSpec, error) {
	ctx, _, done := tele.StartSpanWithLogger(ctx, "subnets.Service.getExisting")
	defer done()

	subnet, err := s.Client.Get(ctx, rgName, spec.VNetName, spec.Name)
	if err != nil {
		return nil, errors.Wrapf(err, "failed to fetch subnet named %s in vnet %s", spec.VNetName, spec.Name)
	}

	var addresses []string
	if subnet.SubnetPropertiesFormat != nil && subnet.SubnetPropertiesFormat.AddressPrefix != nil {
		addresses = []string{to.String(subnet.SubnetPropertiesFormat.AddressPrefix)}
	} else if subnet.SubnetPropertiesFormat != nil && subnet.SubnetPropertiesFormat.AddressPrefixes != nil {
		addresses = to.StringSlice(subnet.SubnetPropertiesFormat.AddressPrefixes)
	}

	subnetSpec := s.Scope.Subnet(spec.Name)
	subnetSpec.ID = to.String(subnet.ID)
	subnetSpec.CIDRBlocks = addresses

	return &subnetSpec, nil
}
```


The old Reconcile logic doesn't explain why this is done, but I believe it's to auto fill the spec with the CIDR blocks for a custom vnet when the user only provides the subnet names, i.e. the example below. What do you think? 
```
networkSpec:
  subnets:
  - name: ${AZURE_CUSTOM_VNET_NAME}-controlplane-subnet
    role: control-plane
  - name: ${AZURE_CUSTOM_VNET_NAME}-node-subnet
    role: node
  vnet:
    name: ${AZURE_CUSTOM_VNET_NAME}
    resourceGroup: ${AZURE_RESOURCE_GROUP}
```

azure/services/subnets/spec.go
Jont828: Sounds good. Just to double check, would the CIDRs be defaulted to an empty slice when it's not provided?

CecileRobertMichon: it would be defaulted to a non-empty slice containing the default CIDR

azure/services/subnets/spec.go
CecileRobertMichon: > The old Reconcile logic doesn't explain why this is done, but I believe it's to auto fill the spec with the CIDR blocks for a custom vnet when the user only provides the subnet names, i.e. the example below. What do you think?
 
 that is correct
 
 I don't quite follow why we need to trick the reconciler and "force a CreateOrUpdate call" though. Wouldn't CreareResource return the existing resource if parameters returns nil already exactly for this use case?

Jont828: Oh that makes sense, in that case it looks like that's redundant.

azure/scope/cluster.go
```diff
@@ -266,20 +267,19 @@ func (s *ClusterScope) NSGSpecs() []azure.NSGSpec {
 }
 
 // SubnetSpecs returns the subnets specs.
-func (s *ClusterScope) SubnetSpecs() []azure.SubnetSpec {
-	numberOfSubnets := len(s.AzureCluster.Spec.NetworkSpec.Subnets)
-	if s.AzureCluster.Spec.BastionSpec.AzureBastion != nil {
-		numberOfSubnets++
-	}
+func (s *ClusterScope) SubnetSpecs() []azure.ResourceSpecGetter {
+	subnetSpecs := make([]azure.ResourceSpecGetter, 0, len(s.AzureCluster.Spec.NetworkSpec.Subnets)+1)
```
CecileRobertMichon, line 271: it should be `len(s.AzureCluster.Spec.NetworkSpec.Subnets)+1` only if `s.AzureCluster.Spec.BastionSpec.AzureBastion != nil`, otherwise the slice should have capacity `len(s.AzureCluster.Spec.NetworkSpec.Subnets)`

what was there before seemed correct, why change it?

azure/services/subnets/subnets.go
```diff
@@ -18,150 +18,111 @@ package subnets
...
 }
 
 // New creates a new service.
 func New(scope SubnetScope) *Service {
+	Client := NewClient(scope)
 	return &Service{
-		Scope:  scope,
-		Client: NewClient(scope),
+		Scope:      scope,
+		Client:     Client,
```
CecileRobertMichon, line 55: is Client still needed?

Jont828: Yes, it's needed for the bastion host service, but the bastion host PR will remove the client when it merges.

azure/services/subnets/subnets.go
```diff
@@ -18,150 +18,111 @@ package subnets
...
 			}
 
-			log.V(2).Info("successfully created subnet in vnet", "subnet", subnetSpec.Name, "vnet", subnetSpec.VNetName)
+			s.Scope.UpdateSubnetID(subnetSpec.ResourceName(), to.String(subnet.ID))
+			s.Scope.UpdateSubnetCIDRs(subnetSpec.ResourceName(), addresses)
 		}
 	}
-	return nil
+
+	s.Scope.UpdatePutStatus(infrav1.SubnetsReadyCondition, serviceName, resultErr)
```
CecileRobertMichon, line 95: SubnetsReadyCondition should be added to the AzureCluster owned conditions

azure/services/subnets/spec_test.go
```diff
@@ -0,0 +1,146 @@
...
+
+func TestParameters(t *testing.T) {
+	testcases := []struct {
+		name          string
+		spec          *SubnetSpec
+		existing      interface{}
+		expect        func(g *WithT, result interface{})
+		expectedError string
+	}{
+		// TODO: add remaining subnet_test.go tests here, i.e. vnet provided but subnet missing, etc.
```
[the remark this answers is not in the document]
Jont828: Nope, there were a couple tests that don't really apply anymore and I just wanted to verify.

azure/services/subnets/subnets.go
Jont828: We need it for now, but the bastion host PR should remove it since that change removes the call.

CecileRobertMichon: we need the subnet Client interface but do we need it in the subnet Service struct? I don't see it being used anywhere in subnets.go

devigned (approved):
lgtm! I did notice at least one outstanding piece of feedback from @CecileRobertMichon. Otherwise, great work!

devigned:
/retest

Looks like VMSS didn't come up for some reason. hmm...

azure/services/subnets/client.go
```diff
@@ -54,43 +61,113 @@ func newSubnetsClient(subscriptionID string, baseURI string, authorizer autorest
 }
 
 // Get gets the specified subnet by virtual network and resource group.
-func (ac *AzureClient) Get(ctx context.Context, resourceGroupName, vnetName, snName string) (network.Subnet, error) {
+func (ac *AzureClient) Get(ctx context.Context, spec azure.ResourceSpecGetter) (result interface{}, err error) {
 	ctx, _, done := tele.StartSpanWithLogger(ctx, "subnets.AzureClient.Get")
 	defer done()
 
-	return ac.subnets.Get(ctx, resourceGroupName, vnetName, snName, "")
+	return ac.subnets.Get(ctx, spec.ResourceGroupName(), spec.OwnerResourceName(), spec.ResourceName(), "")
```
shysank, line 68: see above for comment on vnet resource group.

azure/scope/cluster.go
```diff
@@ -289,20 +290,24 @@ func (s *ClusterScope) NSGSpecs() []azure.NSGSpec {
...
 	}
 
-	subnetSpecs := make([]azure.SubnetSpec, 0, numberOfSubnets)
+	subnetSpecs := make([]azure.ResourceSpecGetter, 0, numberOfSubnets)
+
 	for _, subnet := range s.AzureCluster.Spec.NetworkSpec.Subnets {
-		subnetSpec := azure.SubnetSpec{
+		subnetSpec := &subnets.SubnetSpec{
 			Name:              subnet.Name,
+			ResourceGroup:     s.ResourceGroup(),
```
shysank, line 304: I think we may also have to pass vnet's resource group because that is where subnet will be created : https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/a8555bbe56ce398efdd984906f059abef8e89d00/azure/services/subnets/subnets.go#L103

This is going to be tricky since we don't have a place in `azure.ResourceSpecGetter` to accomodate this. 
cc @CecileRobertMichon

azure/scope/cluster.go
```diff
@@ -311,10 +316,13 @@ func (s *ClusterScope) SubnetSpecs() []azure.SubnetSpec {
 
 	if s.IsAzureBastionEnabled() {
 		azureBastionSubnet := s.AzureCluster.Spec.BastionSpec.AzureBastion.Subnet
-		subnetSpecs = append(subnetSpecs, azure.SubnetSpec{
+		subnetSpecs = append(subnetSpecs, &subnets.SubnetSpec{
 			Name:              azureBastionSubnet.Name,
+			ResourceGroup:     s.ResourceGroup(),
```
shysank, line 321: same comment as above for vnet resource rg.

azure/services/subnets/client.go
```diff
@@ -54,43 +61,113 @@ func newSubnetsClient(subscriptionID string, baseURI string, authorizer autorest
...
+	ctx, _, done := tele.StartSpanWithLogger(ctx, "subnets.AzureClient.CreateOrUpdateAsync")
 	defer done()
 
-	future, err := ac.subnets.CreateOrUpdate(ctx, resourceGroupName, vnetName, snName, sn)
+	subnet, ok := parameters.(network.Subnet)
+	if !ok {
+		return nil, nil, errors.Errorf("%T is not a network.Subnet", parameters)
+	}
+
+	createFuture, err := ac.subnets.CreateOrUpdate(ctx, spec.ResourceGroupName(), spec.OwnerResourceName(), spec.ResourceName(), subnet)
```
shysank, line 83: same comment as above for vnet resource rg.

shysank (changes requested):
Correct me if I'm wrong but we should use vnet's resource group for creating subnets in the same rg. With the current `ResourceGetter` api, we don't have a way to do this other than dumping it in `spec.Parameters`. wdyt?

azure/scope/cluster.go
CecileRobertMichon: actually shouldn't `s.Scope.Vnet().ResourceGroup` be the only RG that gets passed in since that's the RG for the subnet resource? Why do we need ` s.ResourceGroup()` at all?

shysank: I think it's needed to refer sub resource id's => https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/a8555bbe56ce398efdd984906f059abef8e89d00/azure/services/subnets/subnets.go#L90

shysank: we could probably work around that by passing the id directly when creating the spec

CecileRobertMichon: in any case `ResourceGroupName` should return the vnet RG for a subnet spec

Jont828: @shysank @CecileRobertMichon If the user is bringing a custom VNet then wouldn't the NAT gateways, route tables, and security groups be on the VNet's resource group since we're not creating them in the reconciler?

CecileRobertMichon:
exp test is broken in the main branch, see https://kubernetes.slack.com/archives/CEX9HENG7/p1643648764688369

azure/services/subnets/client.go
CecileRobertMichon: this should be resolved now that spec.ResourceGroup is being set to the vnet RG?

azure/scope/managedcontrolplane.go
```diff
@@ -250,12 +251,15 @@ func (s *ManagedControlPlaneScope) NodeNatGateway() infrav1.NatGateway {
...
-	return []azure.SubnetSpec{
-		{
-			Name:     s.NodeSubnet().Name,
-			CIDRs:    s.NodeSubnet().CIDRBlocks,
-			VNetName: s.Vnet().Name,
+func (s *ManagedControlPlaneScope) SubnetSpecs() []azure.ResourceSpecGetter {
+	return []azure.ResourceSpecGetter{
+		&subnets.SubnetSpec{
+			Name:           s.NodeSubnet().Name,
+			ResourceGroup:  s.ResourceGroup(),
```
shysank, line 258:
```suggestion
			ResourceGroup:  s.Vnet().ResourceGroup(),
```

azure/services/subnets/spec.go
```diff
@@ -0,0 +1,103 @@
...
+	// workaround needed to avoid SubscriptionNotRegisteredForFeature for feature Microsoft.Network/AllowMultipleAddressPrefixesOnSubnet.
+	if len(s.CIDRs) == 1 {
+		subnetProperties = network.SubnetPropertiesFormat{
+			AddressPrefix: &s.CIDRs[0],
+		}
+	}
+
+	if s.RouteTableName != "" {
+		subnetProperties.RouteTable = &network.RouteTable{
+			ID: to.StringPtr(azure.RouteTableID(s.SubscriptionID, s.ResourceGroup, s.RouteTableName)),
```
shysank, line 84: This will point to vnet's resource group, but we need cluster's resource group here. We'll need to also set cluster's resource group in SubnetSpec and use it construct sub resource ids. Perhaps something like https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/dc0c3fad698c7c788ae5a14cfbb8c3c94180f97b/azure/services/loadbalancers/spec.go#L39

azure/services/subnets/spec.go
```diff
@@ -0,0 +1,103 @@
...
+		}
+	}
+
+	if s.NatGatewayName != "" {
+		subnetProperties.NatGateway = &network.SubResource{
+			ID: to.StringPtr(azure.NatGatewayID(s.SubscriptionID, s.ResourceGroup, s.NatGatewayName)),
```
shysank, line 90: Same comment as above

azure/services/subnets/spec.go
```diff
@@ -0,0 +1,103 @@
...
+		}
+	}
+
+	if s.SecurityGroupName != "" {
+		subnetProperties.NetworkSecurityGroup = &network.SecurityGroup{
+			ID: to.StringPtr(azure.SecurityGroupID(s.SubscriptionID, s.ResourceGroup, s.SecurityGroupName)),
```
shysank, line 96: Same comment as above

azure/services/subnets/spec.go
CecileRobertMichon: Do we really need the cluster RG here? There are two cases:
1) the vnet is created by CAPZ: then the vnet RG _is_ the cluster RG 
2) the vnet is custom: then we don't create the subnet, so this doesn't matter and the subnet should already have the right RT, NAT GW and NSG attached to it)

Jont828: Yeah that's what I was thinking. I'm not really sure why we'd expect the route tables, nat gateways, and security groups to be in a separate RG if the vnet is custom.

shysank: can we not customize vnet rg even when it is managed by capz? iiuc, it's only set to cluster rg if it's empty https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/5c96e824f739bc8627cab4017312e2054d329d63/api/v1beta1/azurecluster_default.go#L77

Jont828: I'm not sure what you mean, if we specify the VNet RG here then the value should be set in `s.Vnet().ResourceGroup` right?

shysank: Yes, but it wont be the same as cluster rg, right? The current implementation also suggest this: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/5c96e824f739bc8627cab4017312e2054d329d63/azure/services/subnets/subnets.go#L84

shysank: Perhaps it's a bug in the current implementation. But, right now, the way it works is:
1. Vnet and Subnets are created in the virtual network's resource group - `NetworkSpec.Vnet.ResourceGroup`
2. All other resources (route tables, nat gateways, security groups) are created in the cluster resource group - `AzureCluster.ResourceGroup`

CecileRobertMichon: @shysank is correct, I did not think of the case where the user doesn't create the vnet but sets the RG in the spec to be different than the cluster RG. In that case we need to use cluster RG for these since it's not always the same as vnet RG.

CecileRobertMichon: @shysank i believe the reasoning is that you can share a vnet between multiple clusters (and so it makes sense to put it in a common place like a networking RG). Subnets have to be in the same RG as the vnet since they are a nested resource. As for everything else, it’s a 1:1 relationship with the cluster hence putting it all in the cluster RG. If this is unexpected we could rework that UX. Let’s tackle that in a separate issue though and keep the behavior the same for this PR though.

azure/services/subnets/spec.go
```diff
@@ -0,0 +1,104 @@
...
+	Role              infrav1.SubnetRole
+	NatGatewayName    string
+}
+
+// ResourceName returns the name of the subnet.
+func (s *SubnetSpec) ResourceName() string {
+	return s.Name
+}
+
+// ResourceGroupName returns the name of the vnet's resource group.
```
CecileRobertMichon, line 47:
```suggestion
// ResourceGroupName returns the name of the vnet's resource group which the subnet is part of.
```

? 

this might not be the best way of saying it but essentially want to make it clear that we're intentionally using the vnet RG so it doesn't seem like a bug to someone reading this code for the first time

Jont828:
@CecileRobertMichon @shysank I believe the e2e tests are passing now, does this look good to you?

CecileRobertMichon (approved):
/lgtm
/assign @shysank

shysank:
/lgtm
/approve
Thanks for being patient with the reviews :)

PR #1915, opened 2021-12-10
fix: MaxPods is an optional AzureManagedMachinePool configuration

jackfrancis:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it**:

This PR https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1910 included a bug in the webhook validation for AzureManagedMachinePool create: the `MaxPods` configuration is in fact optional, and we should not throw a validation error if that configuration is not present.

Sorry about that!

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

Issue #1338, opened 2021-04-26
Cascading type validations in webhooks


CecileRobertMichon:
> With all of the individual validate${TYPE} funcs, it feels like we should have a validate method added to the types via the code generator and then just have it wired up in the webhooks. Then the webhooks can call an empty validate on the top level type and have it cascade to all of the field types.

Originally posted by @devigned in https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1321#pullrequestreview-644040499

PR #1916, opened 2021-12-10
add OsDiskType configuration for AKS node pools

jackfrancis:
**What type of PR is this?**

/kind feature

This PR enables OSDiskType configuration for AKS cluster node pool (`AzureManagedMachinePool`) resources.

This is an optional configuration for `AzureManagedMachinePool ` resources, and validation will be enforce to ensure that only valid values are allowed: "Managed" or "Ephemeral".

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
add OSDiskType configuration for AKS node pools
```


templates/flavors/aks/cluster-template.yaml
```diff
@@ -74,6 +74,7 @@ spec:
   mode: System
   osDiskSizeGB: 512
   sku: "${AZURE_NODE_MACHINE_TYPE}"
+  osDiskType: "Managed"
```
jackfrancis, line 77: These are temporary changes so that we can run tests before merging. Once we have passing signal I'll remove them.

exp/api/v1beta1/azuremanagedmachinepool_webhook.go
```diff
@@ -193,8 +215,20 @@ func (r *AzureManagedMachinePool) validateLastSystemNodePool(cli client.Client)
 }
 
 func (r *AzureManagedMachinePool) validateMaxPods() error {
-	if to.Int32(r.Spec.MaxPods) < 10 || to.Int32(r.Spec.MaxPods) > 250 {
-		return errors.New("MaxPods must be between 10 and 250")
+	if r.Spec.MaxPods != nil {
```
jackfrancis, line 218: This (and the updates to UT) will be cleaned up once https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1915 merges

CecileRobertMichon:
#1915 merged

exp/api/v1beta1/azuremanagedmachinepool_webhook.go
jackfrancis: Unnecessary foo no longer present

templates/test/ci/cluster-template-prow-aks-multi-tenancy.yaml
```diff
@@ -101,8 +102,38 @@ metadata:
   name: agentpool1
   namespace: default
 spec:
+  maxPods: 64
```
jackfrancis, line 105: Since we're adding a 3rd pool to test the "no configuration specified" pool config scenario, we may as well add a differnt `maxPods` config value to the 2nd one that now serves to validate the 2nd possible valid `OsDiskType` flavor.

exp/api/v1beta1/azuremanagedmachinepool_webhook.go
```diff
@@ -202,6 +224,16 @@ func (r *AzureManagedMachinePool) validateMaxPods() error {
 	return nil
 }
 
+func (r *AzureManagedMachinePool) validateOSDiskType() error {
+	if r.Spec.OsDiskType != "" {
+		if !(r.Spec.OsDiskType == string(containerservice.OSDiskTypeEphemeral) || r.Spec.OsDiskType == string(containerservice.OSDiskTypeManaged)) {
+			return fmt.Errorf("Allowed values of OsDiskType are %s or %s", string(containerservice.OSDiskTypeEphemeral), string(containerservice.OSDiskTypeManaged))
```
CecileRobertMichon, line 230: Forgot to mention this before for maxPods validation but we should be using api-machinery field validation errors for these (similar to what we do above with field.Invalid). Error types are found here: https://github.com/kubernetes/apimachinery/blob/master/pkg/util/validation/field/errors.go#L93

`field.NotSupported()` seems like a good fit for this one

jackfrancis: Done.

Addressed the MaxPods implementation with this PR:

https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1920

azure/scope/managedcontrolplane_test.go
```diff
@@ -225,6 +226,101 @@ func TestManagedControlPlaneScope_MaxPods(t *testing.T) {
...
+						Name:      "cluster1",
+						Namespace: "default",
+					},
+					Spec: infrav1.AzureManagedControlPlaneSpec{
+						SubscriptionID: "00000000-0000-0000-0000-000000000000",
+					},
+				},
+				MachinePool:      getMachinePool("pool1"),
+				InfraMachinePool: getAzureMachinePoolWithOsDiskType("pool1", string(containerservice.OSDiskTypeEphemeral)),
+				PatchTarget:      getAzureMachinePoolWithScaling("pool1", 2, 10),
```
zmalik, line 296:
```suggestion
				PatchTarget:      getAzureMachinePoolWithOsDiskType("pool1", string(containerservice.OSDiskTypeEphemeral)),,
```

exp/api/v1beta1/azuremanagedmachinepool_types.go
```diff
@@ -69,6 +69,10 @@ type AzureManagedMachinePoolSpec struct {
 	// MaxPods specifies the kubelet --max-pods configuration for the node pool.
 	// +optional
 	MaxPods *int32 `json:"maxPods,omitempty"`
+
+	// OsDiskType specifies the OS disk type for each node in the pool. Allowed values are 'Ephemeral' and 'Managed'.
```
zmalik, line 73: lets add the openAPI schema validation for this field
// +kubebuilder:validation:Enum= Ephemeral; Managed

exp/api/v1beta1/azuremanagedmachinepool_webhook.go
```diff
@@ -202,6 +223,19 @@ func (r *AzureManagedMachinePool) validateMaxPods() error {
 	return nil
 }
 
+func (r *AzureManagedMachinePool) validateOSDiskType() error {
+	if r.Spec.OsDiskType != "" {
+		if !(r.Spec.OsDiskType == string(containerservice.OSDiskTypeEphemeral) || r.Spec.OsDiskType == string(containerservice.OSDiskTypeManaged)) {
```
zmalik, line 228: that can be delegated to openAPI schema validation

azure/scope/managedcontrolplane_test.go
jackfrancis: yikes, good catch, I wish the UT result would not have permitted that...

exp/api/v1beta1/azuremanagedmachinepool_webhook.go
jackfrancis: @CecileRobertMichon do you agree this webhook validation is superfluous if we add the enum allowed values to the api property spec?

exp/api/v1beta1/azuremanagedmachinepool_types.go
```diff
@@ -69,6 +69,11 @@ type AzureManagedMachinePoolSpec struct {
...
+	// +kubebuilder:validation:Enum=Ephemeral; Managed
+	// +optional
+	OsDiskType string `json:"osDiskType,omitempty"`
```
zmalik, line 76: should we use a pointer here to make it truly optional? 
and if it's not optional in sense, there should be always some default value can we add this to mutation webhook?

exp/api/v1beta1/azuremanagedmachinepool_webhook.go
CecileRobertMichon: +1, for simple enumeration validation, it's better to use openAPI schema validation and only use webhook validation for more complex validation.

exp/api/v1beta1/azuremanagedmachinepool_types.go
CecileRobertMichon: +1 

exp/api/v1beta1/azuremanagedmachinepool_webhook.go
```diff
@@ -131,6 +133,25 @@ func (r *AzureManagedMachinePool) ValidateUpdate(oldRaw runtime.Object, client c
...
+		} else if r.Spec.OsDiskType != old.Spec.OsDiskType {
+			// changing the field is not allowed
+			allErrs = append(allErrs,
+				field.Invalid(
+					field.NewPath("Spec", "OsDiskType"),
+					r.Spec.OsDiskType,
+					"field is immutable"))
+		}
+	}
+
```
AverageMarcus, line 154: Do we need to handle both situations individually? Can they not be combined into a single "immutable" error?

```suggestion
	if old.Spec.OsDiskType != "" && r.Spec.OsDiskType != old.Spec.OsDiskType  {
		// Prevent OSDiskType modification if it was already set to some value
		allErrs = append(allErrs,
			field.Invalid(
				field.NewPath("Spec", "OsDiskType"),
				r.Spec.OsDiskType,
				"field is immutable"))
	}

```

jackfrancis: I think your observation here is sound. If `A` is a non-empty string, then we can simply test `B` for being different from `A`. `B` being an empty string will always be different from `A` when `A` is a non-empty string, so we can exclusively use:

- `if r.Spec.OsDiskType != old.Spec.OsDiskType`

exp/api/v1beta1/azuremanagedmachinepool_types.go
jackfrancis: done

exp/api/v1beta1/azuremanagedmachinepool_webhook.go
jackfrancis: done

exp/api/v1beta1/azuremanagedmachinepool_webhook.go
jackfrancis: Because other validations in this surface area are doing this "test for empty string" validation, let's continue this pattern for now. If we want to change this approach let's get consensus and then change it everywhere so that we're consistent.

Thoughts?

jackfrancis:
@CecileRobertMichon @zmalik I think this is ready for another review

azure/scope/managedcontrolplane.go
```diff
@@ -501,6 +501,7 @@ func (s *ManagedControlPlaneScope) GetAgentPoolSpecs(ctx context.Context) ([]azu
 			OSDiskSizeGB:      0,
 			Mode:              pool.Spec.Mode,
 			AvailabilityZones: pool.Spec.AvailabilityZones,
+			OsDiskType:        pool.Spec.OsDiskType,
```
jackfrancis, line 504: @zmalik @CecileRobertMichon moved these assignments here (instead of lower in a "if not empty" condition) in response to this PR:

https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1945

exp/api/v1beta1/azuremanagedmachinepool_types.go
```diff
@@ -69,6 +69,11 @@ type AzureManagedMachinePoolSpec struct {
 	// MaxPods specifies the kubelet --max-pods configuration for the node pool.
 	// +optional
 	MaxPods *int32 `json:"maxPods,omitempty"`
+
+	// OsDiskType specifies the OS disk type for each node in the pool. Allowed values are 'Ephemeral' and 'Managed'.
+	// +kubebuilder:validation:Enum=Ephemeral; Managed
```
AverageMarcus, line 74:
```suggestion
	// +kubebuilder:validation:Enum=Ephemeral;Managed
```

Nitpicking as I don't _think_ it actually matters but for consistency sake :)

exp/api/v1beta1/azuremanagedmachinepool_webhook.go
```diff
@@ -42,6 +43,10 @@ func (r *AzureManagedMachinePool) Default(client client.Client) {
 	if r.Spec.Name == nil || *r.Spec.Name == "" {
 		r.Spec.Name = &r.Name
 	}
+
+	if r.Spec.OsDiskType == nil || to.String(r.Spec.OsDiskType) == "" {
```
AverageMarcus, line 47: Could this be handled using a default comment on the type for OpenAPI to handle for us? (I'm not actually sure if that would handle both nil and empty string)

exp/api/v1beta1/azuremanagedmachinepool_webhook.go
AverageMarcus: +1 for consistency. I hadn't noticed the similar pattern above.

exp/api/v1beta1/azuremanagedmachinepool_webhook.go
```diff
@@ -205,6 +230,10 @@ func (r *AzureManagedMachinePool) validateMaxPods() error {
 	return nil
 }
 
+func (r *AzureManagedMachinePool) validateOSDiskType() error {
```
AverageMarcus, line 233: Do we still need this function?

jackfrancis: I have a slight preference towards shipping a new feature w/ a "prototype" validation function to standardize how we maintain these validations (so that if we add validations over time they are structured/expressed in a consistent way across the codebase).

@CecileRobertMichon thoughts?

CecileRobertMichon: I'm leaning towards not keeping a function that's a no-op / not needed. If we come across the need for more validation that can't be done in OpenAPI later we can easily add this back (there are plenty of other examples of similar validation functions in the codebase)

See also #1338

exp/api/v1beta1/azuremanagedmachinepool_webhook.go
jackfrancis: I think we could add this:

`// +kubebuilder:default=Managed`

@CecileRobertMichon @zmalik is that preferable to enforcing in the `Default()` method of the webhook?

CecileRobertMichon: Both are valid ways to do defaulting. I don't know if one is preferable over the other. OpenAPI defaulting is slightly cleaner since it doesn't require code in the webhook, but I don't know that it's a huge gain. It might be good to check if there are any API conventions out there that recommend one over the other (there isn't anything about this in https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md).

jackfrancis: I would prefer using the openapi solution, I'll test drive that out.

@zmalik do you have any thoughts?

exp/api/v1beta1/azuremanagedmachinepool_webhook.go
jackfrancis: removed

exp/api/v1beta1/azuremanagedmachinepool_types.go
jackfrancis: fixed, thanks!

jackfrancis:
@AverageMarcus @zmalik @CecileRobertMichon this is ready for another (possibly final) round of reviews, thank you!

AverageMarcus (approved):
LGTM 👍

exp/api/v1beta1/azuremanagedmachinepool_webhook.go
```diff
@@ -42,6 +43,10 @@ func (r *AzureManagedMachinePool) Default(client client.Client) {
 	if r.Spec.Name == nil || *r.Spec.Name == "" {
 		r.Spec.Name = &r.Name
 	}
+
+	if r.Spec.OsDiskType == nil || to.String(r.Spec.OsDiskType) == "" {
+		r.Spec.OsDiskType = to.StringPtr(string(containerservice.OSDiskTypeManaged))
```
CecileRobertMichon, line 48: It looks like you added a default in the OpenAPI schema, can we remove this?

jackfrancis: Yes, thanks!

Because the schema is now injecting the default value, are we able to use validation to fail if nil or empty string?

templates/test/ci/prow-aks-multi-tenancy/patches/aks-agentpool0.yaml
```diff
@@ -3,4 +3,5 @@ kind: AzureManagedMachinePool
 metadata:
   name: "agentpool0"
 spec:
-  maxPods: 24
+  maxPods: 32
```
CecileRobertMichon, line 6: let's remove this and wait for #1971 to merge

exp/api/v1beta1/azuremanagedmachinepool_webhook.go
zmalik: I also think that both ways are valid. And not sure about the standard preferred way. If it can be done through OpenAPI I would also prefer that way. 

> I'm not actually sure if that would handle both nil and empty string

from 1.20 we have this: 
```
When nullable is true, null values will be conserved and won't be defaulted
```

jackfrancis: Based on prior discussions about preferring an opinionated approach vs a passive one (let AKS choose the default for us), I think we should keep the current definition and not set `OsDiskType` to nullable.

templates/test/ci/prow-aks-multi-tenancy/patches/aks-agentpool0.yaml
jackfrancis: rebased

CecileRobertMichon (commented):
/approve

PR #1918, opened 2021-12-13
Make security groups reconcile/delete async

CecileRobertMichon:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**: Broke out NSGs from #1684. This follows the pattern set in #1541 and applies it to security groups.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Make security groups reconcile/delete async
```


azure/services/securitygroups/spec.go
```diff
@@ -0,0 +1,117 @@
...
+		Location: to.StringPtr(s.Location),
+		SecurityGroupPropertiesFormat: &network.SecurityGroupPropertiesFormat{
+			SecurityRules: &securityRules,
+		},
+		Etag: etag,
+	}, nil
+}
+
+// TODO: review this logic and make sure it is what we want. It seems incorrect to skip rules that don't have a certain protocol, etc.
+func ruleExists(rules []network.SecurityRule, rule network.SecurityRule) bool {
```
CecileRobertMichon, line 96: this logic smells fishy to me, not sure why we're doing a continue when the rule has certain properties. That being said I didn't want to introduce bug fixes/behavior changes as part of this already large refactor so I kept the logic as is from https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1684/files#diff-5bb14c98c52ecfa3dad6cf4745c9378b7624fdc60f0910343534ddac18408fb1L115

azure/services/securitygroups/client.go
```diff
@@ -82,30 +85,91 @@ func (ac *azureClient) CreateOrUpdate(ctx context.Context, resourceGroupName str
...
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to marshal future")
+		}
+		if err := json.Unmarshal(jsonData, &future); err != nil {
+			return nil, errors.Wrap(err, "failed to unmarshal future data")
+		}
+		result = (*future).Result
+
+	case infrav1.DeleteFuture:
+		// Delete does not return a result vnet.
```
Jont828, line 167:
```suggestion
		// Delete does not return a result security group
```

azure/services/securitygroups/securitygroups.go
```diff
@@ -18,142 +18,98 @@ package securitygroups
...
+	// Only create the NSGs if their lifecycle is managed by this controller.
 	if !s.Scope.IsVnetManaged() {
-		log.V(4).Info("Skipping network security group reconcile in custom VNet mode")
+		s.Scope.UpdatePutStatus(infrav1.SecurityGroupsReadyCondition, serviceName, nil)
+		log.Info("Skipping network security groups reconcile in custom VNet mode")
 		return nil
 	}
 
+	// We go through the list of NSGSpecs to reconcile each one, independently of the result of the previous one.
+	// If multiple erros occur, we return the most pressing one
```
Jont828, line 70:
```suggestion
	// If multiple errors occur, we return the most pressing one
```

azure/services/securitygroups/securitygroups.go
```diff
@@ -18,142 +18,98 @@ package securitygroups
...
 	if !s.Scope.IsVnetManaged() {
-		log.V(4).Info("Skipping network security group delete in custom VNet mode")
+		log.V(4).Info("Skipping network security groups delete in custom VNet mode")
 		return nil
 	}
 
+	var result error
+
+	// We go through the list of NSGSpecs to delete each one, independently of the result of the previous one.
+	// If multiple erros occur, we return the most pressing one
```
Jont828, line 103:
```suggestion
	// If multiple errors occur, we return the most pressing one
```

azure/services/securitygroups/securitygroups_test.go
```diff
@@ -18,212 +18,117 @@ package securitygroups
...
+				r.CreateResource(gomockinternal.AContext(), &fakeNSG, serviceName).Return(nil, errFake)
+				r.CreateResource(gomockinternal.AContext(), &fakeNSG2, serviceName).Return(nil, nil)
+				s.UpdatePutStatus(infrav1.SecurityGroupsReadyCondition, serviceName, errFake)
 			},
-		}, {
-			name: "skipping network security group reconcile in custom VNet mode",
-			expect: func(s *mock_securitygroups.MockNSGScopeMockRecorder, m *mock_securitygroups.MockclientMockRecorder) {
+		},
+		{
+			name:          "first sg create fails, second sg create not done, should return create error",
```
Jont828, line 106: Just a thought: we have this block in several services where we have a loop and need to decide if we should overwrite the error message.
```
if !azure.IsOperationNotDoneError(err) || result == nil {
	result = err
}
```
We end up having to test the logic for handling `OperationNotDoneError`s and when `result == nil` for every service. Maybe we can try to refactor the logic to a helper that we can thoroughly test, similar to how we test all the errors related to `Get`, `CreateOrUpdateAsync`, and `Parameters` in `async_test.go`. For example, we might append all the errors to a list and create a function that picks the most important error out of it.

azure/services/securitygroups/spec.go
```diff
@@ -0,0 +1,117 @@
...
+	"strings"
+
+	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2021-02-01/network"
+	"github.com/Azure/go-autorest/autorest/to"
+	"github.com/pkg/errors"
+	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
+	"sigs.k8s.io/cluster-api-provider-azure/azure/converters"
+)
+
+// NSGSpec defines the specification for a Security Group.
```
Jont828, line 29: Is there a naming convention for the casing, i.e. Security group/security group/Security Group?

azure/services/securitygroups/spec.go
Jont828: Yeah the logic is a bit unintuitive. Do you think you could streamline the boolean logic without changing the behavior, i.e. if the names and all the other fields match, return true instead of continuing otherwise.

azure/services/securitygroups/client.go
```diff
@@ -54,26 +50,33 @@ func newSecurityGroupsClient(subscriptionID string, baseURI string, authorizer a
 }
 
 // Get gets the specified network security group.
-func (ac *azureClient) Get(ctx context.Context, resourceGroupName, sgName string) (network.SecurityGroup, error) {
+func (ac *azureClient) Get(ctx context.Context, spec azure.ResourceSpecGetter) (interface{}, error) {
 	ctx, _, done := tele.StartSpanWithLogger(ctx, "securitygroups.AzureClient.Get")
```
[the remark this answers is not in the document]
mboersma: I was reviewing this and just thought of that above... I saw that we've been capitalizing the second part of the logger label in other contexts too, but I think lowercase `azureClient` matching the struct name makes more sense.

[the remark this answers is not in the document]
CecileRobertMichon: Changed to lower case `azureClient`, let's try to stick to that in other places as well.

azure/services/securitygroups/spec.go
CecileRobertMichon: depends on the context but here it should be lower case. Changed.

azure/services/securitygroups/spec.go
CecileRobertMichon: I'd rather not mess with it and just rewrite that function all together in a separate PR

CecileRobertMichon:
rebased, this is ready for review again

/assign @mboersma @Jont828 @shysank

azure/services/securitygroups/spec.go
```diff
@@ -0,0 +1,117 @@
...
+	return s.ResourceGroup
+}
+
+// OwnerResourceName is a no-op for security groups.
+func (s *NSGSpec) OwnerResourceName() string {
+	return ""
+}
+
+// Parameters returns the parameters for the security group.
+func (s *NSGSpec) Parameters(existing interface{}) (interface{}, error) {
```
shysank, line 53: can we have some unit tests for this?

shysank:
Apart from one nit, and spec tests not being included, pr looks good.

azure/services/securitygroups/spec.go
CecileRobertMichon: added

CecileRobertMichon:
@shysank @Jont828 addressed all comments and rebased, PTAL

azure/services/securitygroups/spec_test.go
```diff
@@ -0,0 +1,177 @@
+/*
+Copyright 2021 The Kubernetes Authors.
```
shysank, line 2:
```suggestion
Copyright 2022 The Kubernetes Authors.
```

shysank (commented):
/lgtm
There are a couple of 2021 -> 2022 things, but since the pr was opened in dec 2021, I'm fine merging it.

CecileRobertMichon:
updated the copyrights :)

azure/services/securitygroups/securitygroups.go
```diff
@@ -18,142 +18,96 @@ package securitygroups
...
-					securityRules = append(securityRules, sdkRule)
+	var resErr error
+
+	// Only create the NSGs if their lifecycle is managed by this controller.
+	if !s.Scope.IsVnetManaged() {
+		log.Info("Skipping network security groups reconcile in custom VNet mode")
+	} else {
+		// We go through the list of security groups to reconcile each one, independently of the result of the previous one.
+		// If multiple errors occur, we return the most pressing one.
+		//  Order of precedence (highest -> lowest) is: error that is not an operationNotDoneError (ie. error creating) -> operationNotDoneError (ie. creating in progress) -> no error (ie. created)
```
Jont828, line 70:
```suggestion
		//  Order of precedence (highest -> lowest) is: error that is not an operationNotDoneError (i.e. error creating) -> operationNotDoneError (i.e. creating in progress) -> no error (i.e. created)
```

azure/services/securitygroups/securitygroups.go
```diff
@@ -18,142 +18,96 @@ package securitygroups
...
 	ctx, log, done := tele.StartSpanWithLogger(ctx, "securitygroups.Service.Delete")
 	defer done()
 
+	ctx, cancel := context.WithTimeout(ctx, reconciler.DefaultAzureServiceReconcileTimeout)
+	defer cancel()
+
+	// Only delete the NSG if its lifecycle is managed by this controller.
 	if !s.Scope.IsVnetManaged() {
-		log.V(4).Info("Skipping network security group delete in custom VNet mode")
+		log.V(4).Info("Skipping network security groups delete in custom VNet mode")
```
Jont828, line 94: It looks like the log level for skipping NSG reconcile is a `log.Info()` while this is a `log.V(4).Info()`, should they be the same log level?

azure/services/securitygroups/securitygroups.go
Jont828: It looks like this typo is still here

azure/services/securitygroups/securitygroups_test.go
```diff
@@ -322,16 +225,101 @@ func TestDeleteSecurityGroups(t *testing.T) {
...
+			DestinationAddressPrefix: to.StringPtr("*"),
+			SourceAddressPrefix:      to.StringPtr("*"),
+			Priority:                 to.Int32Ptr(100),
+			Direction:                network.SecurityRuleDirectionOutbound,
+		},
+	}
+)
 
-			g.Expect(s.Delete(context.TODO())).To(Succeed())
+func TestRuleExists(t *testing.T) {
```
Jont828, line 290: What do you think about moving the tests for this helper function to `spec_test.go` since it's defined in `spec.go`?

azure/services/securitygroups/securitygroups.go
```diff
@@ -18,142 +18,96 @@ package securitygroups
...
 		return nil
 	}
 
+	var result error
+
+	// We go through the list of security groups to delete each one, independently of the result of the previous one.
+	// If multiple erros occur, we return the most pressing one
```
Jont828, line 101: It looks like this typo is still here

azure/services/securitygroups/securitygroups.go
CecileRobertMichon: changed both to V(4) since route tables are also using V(4)

CecileRobertMichon:
rebased

/assign @Jont828 @shysank

CecileRobertMichon:
rebased on top of #2093

PR #1921, opened 2021-12-14
Make vnets reconcile/delete async

CecileRobertMichon:
**What type of PR is this?**
/kind feature
/kind cleanup

**What this PR does / why we need it**: Follow up on #1610 and implementation of #1541 for virtual networks.

Fixes #

**Special notes for your reviewer**: This PR replaces https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1684

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Make vnets reconcile/delete async
```


CecileRobertMichon:
rebased, this is ready for review again

/assign @shysank @mboersma

azure/services/virtualnetworks/virtualnetworks.go
```diff
@@ -18,151 +18,130 @@ package virtualnetworks
...
+	}
+
+	vnetIface, err := s.Get(ctx, spec)
 	if err != nil {
 		if azure.ResourceNotFound(err) {
-			log.V(2).Info(fmt.Sprintf("Resource not found for VNet %q from resource group %q", spec.Name, spec.ResourceGroup))
-			return nil, err
+			// if the vnet was already deleted, attempt to get previous management state from the spec.
+			if s.Scope.Vnet().ID != "" {
+				return s.Scope.Vnet().Tags.HasOwned(s.Scope.ClusterName()), nil
```
shysank, line 136: why should we do this? can we not just return the error since `azure.ResourceNotFound` is handled in the calling function?

shysank (commented):
looks good overall with one minor comment. Will give another review after rebase.

azure/services/virtualnetworks/virtualnetworks.go
CecileRobertMichon: I think I did this at the time I was sharing this function with other services trying to centralize "is vnet managed" because there was an edge case where if the vnet had already been deleted we wouldn't know whether we should delete the nat gateway. I don't think we need this for now though since I removed the caching part of this PR to simplify it. Will remove and update comment.

CecileRobertMichon: done

CecileRobertMichon:
@shysank @Jont828 rebased and addressed comments, PTAL

CecileRobertMichon:
/retest

aks flake

shysank (commented):
looks good overall with a couple of nits and a clarification on vnet peerings

azure/scope/cluster.go
```diff
@@ -362,11 +363,14 @@ func (s *ClusterScope) VnetPeeringSpecs() []azure.ResourceSpecGetter {
...
 
 // VNetSpec returns the virtual network spec.
-func (s *ClusterScope) VNetSpec() azure.VNetSpec {
-	return azure.VNetSpec{
-		ResourceGroup: s.Vnet().ResourceGroup,
-		Name:          s.Vnet().Name,
-		CIDRs:         s.Vnet().CIDRBlocks,
+func (s *ClusterScope) VNetSpec() azure.ResourceSpecGetter {
+	return &virtualnetworks.VNetSpec{
+		ResourceGroup:  s.Vnet().ResourceGroup,
```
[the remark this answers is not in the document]
CecileRobertMichon: yes, what was happening before was that [we were doing a full deep copy into the vnet](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/services/virtualnetworks/virtualnetworks.go#L78) stored in AzureCluster.Spec which meant we had to include every single field that had been set before otherwise we would risk overwriting it. This is not good because if a new feature gets added like vnet peerings there could be a bug where the vnet write erases the user configuration (and in fact we did run into that issue during vnet peering implementation). This is hopefully simplifying the code while keeping it functionally equivalent. We [only set the vnet, ID, tags and cidrs](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1921/files#diff-f45b26f09f47e6be08607d767760708b985d428a1dc4b9a81f0640ee3e7ee656R78-R85) from the existing vnet while keeping the rest of the VnetSpec the same.

devigned (commented):
lgtm outside of the failure in e2e when ensuring the cluster is deleted.

CecileRobertMichon:
> Operation could not be completed as it results in exceeding approved standardDSv3Family Cores quota

/retest

azure/scope/cluster.go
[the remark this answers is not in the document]
shysank: Oh okay, makes sense. Thanks for the explanation!

CecileRobertMichon:
AKS test is broken on main (see https://kubernetes.slack.com/archives/CEX9HENG7/p1643648764688369)

all other tests passed

/cc @devigned

CecileRobertMichon:
/retest

exp test fixed in #2025

CecileRobertMichon:
/retest

AKS timesync flake

Issue #1917, opened 2021-12-10
Use kustomize remote URL's to manage addons


jsturtevant:
/kind feature

**Describe the solution you'd like**

Currently we copy the addon's into our code base, modify and maintain them.  We can use kustomizes remote url and a make command to automate this as done in https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1903

example from pr:
```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
namespace: kube-system
resources:
  - https://github.com/kubernetes-sigs/metrics-server/releases/download/v0.5.2/components.yaml
```

```
.PHONY: generate-addons
generate-addons:
	$(KUSTOMIZE) build $(ADDONS_DIR)/metrics-server > $(ADDONS_DIR)/metrics-server/metrics-server.yaml
```

Possible files we could switch to this method:

- [ ] cloudprovider
- [x] [calico](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/templates/addons/calico.yaml)
- [x] [calico ipv6](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/templates/addons/calico-ipv6.yaml)
- [ ] [storage classes for azure disk/file](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/templates/addons/storageclass-azure-file.yaml)
- [ ] [windows files](https://github.com/kubernetes-sigs/cluster-api-provider-azure/tree/main/templates/addons/windows)

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- cluster-api-provider-azure version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

jsturtevant:
we've made a progress on this. I've checked off a few of the checkboxes

CecileRobertMichon:
For cloud-provider we might not be able to do this as there is no manifest artifact published as part of [releases](https://github.com/kubernetes-sigs/cloud-provider-azure/releases). There are [sample manifest in the repo ](https://github.com/kubernetes-sigs/cloud-provider-azure/tree/master/examples/out-of-tree) but @feiskyer pointed out those shouldn't be used directly as addon manifests since they use a `Pod` for CCM which should only be applied as a static Pod (instead of a `Deployment`).

feiskyer:
[...]

PR #1922, opened 2021-12-14
Update NVIDIA GPU operator componentry

### LINKED ISSUE #1907: GPU tests failing consistently in testgrid
/kind bug

The GPU-enabled cluster tests [have been failing](https://testgrid.k8s.io/sig-cluster-lifecycle-cluster-api-provider-azure#capz-periodic-e2e-full-main) in k8s testgrid. Need to find the root cause.

mboersma:
**What type of PR is this?**

/kind bug

The GPU-enabled test spec began failing a couple weeks ago. This updates the GPU operator components to recently released versions, which seems to fix what was essentially a runtime C code compilation error inside the nvidia-driver-daemonset pod.

Fixes #1907

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```


mboersma:
I made this a work-in-progress because I'm adding some test logging that would have clarified the error.

templates/flavors/nvidia-gpu/gpu-operator-components.yaml
```diff
@@ -344,7 +344,7 @@ spec:
         - name: node-feature-discovery-master
           securityContext:
             {}
-          image: "quay.io/kubernetes_incubator/node-feature-discovery:v0.6.0"
+          image: "k8s.gcr.io/nfd/node-feature-discovery:v0.9.0"
```
CecileRobertMichon, line 347: this is a good example of one we could switch to using the remote spec directly in the future #1917

CecileRobertMichon:
@mboersma planning on cutting v1.1 today, should we try to include this in there?

mboersma:
> should we try to include this...?

That would be great. I'm validating some changes locally (now that we have power again) and then I'll push them out to the PR.

CecileRobertMichon:
```
capz-e2e: Workload cluster creation Creating a GPU-enabled cluster with a single control plane node and 1 node | 21m43s
-- | --
```

The full test suite failed with an ipv6 and an AKS failure but the GPU test passed, which is what we're looking for here. Going to look into the two failures but in the interest of time and being able to include this in the release:

/lgtm
/approve

Issue #1701, opened 2021-09-20
AKS clusters should allow users to specify more optional configuration for node pools


meixingdb:
/kind feature

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]

Add optional configurations for users to specify specs for AKS node pools. The list of options include:
// MaxCount - Maximum number of nodes for auto-scaling
// +optional
MaxCount 
// MinCount - Minimum number of nodes for auto-scaling
// +optional
MinCount 
// EnableAutoScaling - Whether to enable auto-scaler
// +optional
EnableAutoScaling 
// EnableNodePublicIP - Enable public IP for nodes
// +optional
EnableNodePublicIP 
// EnableFIPS - Whether to use FIPS enabled OS
// +optional
EnableFIPS
// OsDiskType - OS disk type to be used for machines in a given agent pool. Allowed values are 'Ephemeral' and 'Managed'. If unspecified, defaults to 'Ephemeral' when the VM supports ephemeral OS and has a cache disk larger than the requested OSDiskSizeGB. Otherwise, defaults to 'Managed'. May not be changed after creation. Possible values include: 'Managed', 'Ephemeral'
// +optional
OsDiskType 
// NodeLabels - Agent pool node labels to be persisted across all nodes in agent pool.
// +optional
NodeLabels 
// NodeTaints - Taints added to new nodes during node pool create and scale. For example, key=value:NoSchedule.
// +optional
NodeTaints 
// VnetSubnetID - VNet SubnetID specifies the VNet's subnet identifier for nodes and maybe pods
// +optional
VnetSubnetID 
// AvailabilityZones - Availability zones for nodes. Must use VirtualMachineScaleSets AgentPoolType.
// +optional
AvailabilityZones 
// ScaleSetPriority - ScaleSetPriority to be used to specify virtual machine scale set priority. Default to regular. Possible values include: 'Spot', 'Regular'
// +optional
ScaleSetPriority 
// MaxPods - Maximum number of pods that can run on a node.
// +optional
MaxPods 
// KubeletConfig - KubeletConfig specifies the configuration of kubelet on agent nodes.
// +optional
KubeletConfig 

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- cluster-api-provider-azure version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

sadysnaat:
[...]

PR #1925, opened 2021-12-15
add support for spot instances in AKS

zmalik:
**What type of PR is this?**

/kind feature

- Adds support for spot instances node pool for AKS.

Fixes  part of #1701

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```release-note
add support for spot node pool in AKS
```


exp/api/v1beta1/azuremanagedmachinepool_webhook.go
```diff
@@ -42,6 +42,10 @@ func (r *AzureManagedMachinePool) Default(client client.Client) {
 	if r.Spec.Name == nil || *r.Spec.Name == "" {
 		r.Spec.Name = &r.Name
 	}
+
+	if r.Spec.ScaleSetPriority == nil {
```
jackfrancis, line 46: @CecileRobertMichon do we want to have capz to have a default opinion of this? The alternative would be to continue not including this configuration and defer to AKS for a default.

exp/api/v1beta1/azuremanagedmachinepool_webhook.go
```diff
@@ -205,6 +210,18 @@ func (r *AzureManagedMachinePool) validateMaxPods() error {
...
 
+func (r *AzureManagedMachinePool) ValidateSpotNodePool() error {
+	if r.Spec.ScaleSetPriority != nil && *r.Spec.ScaleSetPriority == "Spot" {
+		if r.Spec.Mode != string(NodePoolModeUser) {
+			return field.Invalid(
+				field.NewPath("Spec", "ScaleSetPriority"),
+				r.Spec.ScaleSetPriority,
+				"Spot ScaleSetPriority requires AzureManagedMachinePool mode User")
+		}
+	}
```
jackfrancis, line 221: Let's add further validation to ensure that only the literal strings "Spot" and "Regular" are allowed. If I use a value of "Special" (or some other random string) the validation should fail.

exp/api/v1beta1/azuremanagedmachinepool_webhook.go
CecileRobertMichon: +1 to defaulting because otherwise if AKS ever changed the default it wouldn't be backwards compatible if the field is immutable (I doubt they would do that for this particular setting but you never know)

exp/api/v1beta1/azuremanagedmachinepool_webhook.go
```diff
@@ -205,6 +210,18 @@ func (r *AzureManagedMachinePool) validateMaxPods() error {
 	return nil
 }
 
+func (r *AzureManagedMachinePool) ValidateSpotNodePool() error {
+	if r.Spec.ScaleSetPriority != nil && *r.Spec.ScaleSetPriority == "Spot" {
+		if r.Spec.Mode != string(NodePoolModeUser) {
+			return field.Invalid(
```
CecileRobertMichon, line 216: field.Forbidden might be a better fit for this error

exp/api/v1beta1/azuremanagedmachinepool_webhook.go
zmalik: we have open API schema validation for that

CecileRobertMichon (commented):
/lgtm
/cc @meixingdb

zmalik:
rebased and squashed

CecileRobertMichon:
@zmalik can you please check why test is failing on this one when you get a chance?

zmalik:
rebased to bring all fixes from `main`

zmalik:
will rebase after https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1927 lands as it will require another rebase after that

zmalik:
will pick this PR now, that taints and label PRs have been merged

jackfrancis:
@zmalik is on 🔥

zmalik:
failed with `No AKS versions found for v1.19` 
will try to bump in this PR just to test the fix

zmalik:
that fixed it

jackfrancis (approved):
/lgtm

zmalik:
removed the AKS version bump as https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2108 merged

zmalik:
added to e2e tests a spot agentpool

zmalik:
after doing more experiments looks like AKS returns errors when setting from empty scalesetpriority to `Regular`
```
Changing property 'properties.ScaleSetPriority' is not allowed
```
This is an issue for all existing nodepools that were created through old versions of CAPZ. which means we will need to make it by default empty also.

jackfrancis:
@zmalik can we use `*string` to our advantage anywhere to be able to differentiate between empty string and actually non-existent?

zmalik:
> @zmalik can we use *string to our advantage anywhere to be able to differentiate between empty string and actually non-existent?

yes, we can do that. But can that be confusing for users?

if it's nil --> we don't set any default value? (as that would be case for users upgrading from old version of CAPZ)
if it's not nil but empty --> we set a default value of `Regular` as by design in CAPZ we don't allow empty string for scalesetpriority?

was that the proposal or reverse?

jackfrancis:
@zmalik Yeah that's my thinking, though I'm just driving by and haven't looked at all the codepaths yet!

The reason to think hard about figuring out some solution of setting a capz default value is to avoid unexpected changes in the future when AKS decides to change its opinion of what should be default.

zmalik:
yeah, I agree. 

Mostly I'm thinking in case we introduce some default and user has to set empty string to trigger the default it might not be intuitive. As with `nil` value the old and new will still bypass any defaulting. 

Can we ask AKS team if changing from `""` to `Regular` in Azure API should be allowed as effectively `""` means `Regular` and the error message `Changing property 'properties.ScaleSetPriority' is not allowed` is purely an API error as user is not changing the instance type but setting the `ScaleSetPriority` which it already has.

jackfrancis:
@palma21 it seems like the following is happing in AKS:

1) user does not specify a `ScaleSetPriority` when creating a cluster/node pool, which then sets the cluster/pool data model representation to an "empty" value for that configuration property
2) AKS effectively sets the VMSS priority to be `"Regular"`, but does not update the customer cluster/pool data model `ScaleSetPriority ` property to reflect this configuration
3) if a user submits a PUT operation on his or her existing cluster, including the effective value of `ScaleSetPriority` (i.e., "Regular") results in a `Changing property 'properties.ScaleSetPriority' is not allowed` error, even though in fact that PUT data representation accurately reflects the fact that VMSS "Regular" priority is actually being used in practice

Is this a bug, or is this the behavior of AKS as designed?

Thanks!

CecileRobertMichon:
@alexeldeib any ideas on this one?

jackfrancis:
@zmalik reviving this after a while

1. do you want to pick this up again? (I'm happy to do so myself if you don't have cycles)
2. I would vote for introducing a change to the current behavior and explicitly choosing a default; I would also vote that we audit existing capz AzureManagedCluster features and do the same for all (i.e., not rely upon the AKS API to "default" us to one thing or the other, because that is subject to change in the future, with unpredictable side-effects)

jackfrancis:
I have another solution in place that addresses node pool labeling side-effects with spot-enabled pools.

PR #1936, opened 2021-12-17
Use hack/tools kubectl in Tiltfile

CecileRobertMichon:
**What type of PR is this?**

**What this PR does / why we need it**: Use the hack/tools kubectl binary in the Tiltfile instead of whatever version of kubectl is installed on the dev machine. 

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -1,6 +1,7 @@
 # -*- mode: Python -*-
 
 envsubst_cmd = "./hack/tools/bin/envsubst"
+kubectl_cmd = "./hack/tools/bin/kubectl"
 tools_bin = "./hack/tools/bin"
 
 #Add tools to path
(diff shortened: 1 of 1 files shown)


mboersma (approved):
/lgtm

I tried out `make tilt-up` locally and it seems fine.

jsturtevant:
Should we update the tilt modules as well? https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/tilt_modules/cert_manager/Tiltfile

CecileRobertMichon:
> Should we update the tilt modules as well? https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/tilt_modules/cert_manager/Tiltfile

My understanding is that the cert_manager module is copied directly from https://github.com/tilt-dev/tilt-extensions/blob/master/cert_manager/Tiltfile so not sure if we should avoid diverging from it, @mboersma thoughts?

mboersma:
What I've seen is that removing the `tilt_modules/` directory (or a subdirectory) will force `tilt` to fetch these tilt_extensions from the source that @CecileRobertMichon pointed out. 

AFAICT we are up-to-date with the tilt_extensions version anyway, so I don't think there's anything to do here.

jsturtevant:
sounds like it's not necessary.

/lgtm

PR #1945, opened 2022-01-03
Set max pods for the first time creation of ManagedControlplane (AKS) cluster

zmalik:
**What type of PR is this?**

/kind bug

When creating AKS clusters for first time we create the system agentpool through ManagedClusters and not through `AMMPController`. If we have a system agentpool created by user with `maxPods` value configured. It was being ignored. And later the `AMMPController` is trying to set the maxPods value but as it is immutable it stays in error. 

This PR sets the maxPods for the first time creation also. 

We will have to consider the above path when adding other immutable fields for agentpools also. 

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
set maxPods for first time creation of clusters
```

### DIFF
--- a/azure/scope/managedcontrolplane.go
+++ b/azure/scope/managedcontrolplane.go
@@ -500,6 +500,7 @@ func (s *ManagedControlPlaneScope) GetAgentPoolSpecs(ctx context.Context) ([]azu
 			Replicas:          1,
 			OSDiskSizeGB:      0,
 			Mode:              pool.Spec.Mode,
+			MaxPods:           pool.Spec.MaxPods,
 			AvailabilityZones: pool.Spec.AvailabilityZones,
 		}
 
@@ -557,6 +558,7 @@ func (s *ManagedControlPlaneScope) AgentPoolSpec() azure.AgentPoolSpec {
 			s.ControlPlane.Spec.VirtualNetwork.Subnet.Name,
 		),
 		Mode:              s.InfraMachinePool.Spec.Mode,
+		MaxPods:           s.InfraMachinePool.Spec.MaxPods,
 		AvailabilityZones: s.InfraMachinePool.Spec.AvailabilityZones,
 	}
 
@@ -568,10 +570,6 @@ func (s *ManagedControlPlaneScope) AgentPoolSpec() azure.AgentPoolSpec {
 		agentPoolSpec.EnableAutoScaling = to.BoolPtr(true)
 		agentPoolSpec.MaxCount = s.InfraMachinePool.Spec.Scaling.MaxSize
 		agentPoolSpec.MinCount = s.InfraMachinePool.Spec.Scaling.MinSize
-	}
-
-	if s.InfraMachinePool.Spec.MaxPods != nil {
-		agentPoolSpec.MaxPods = s.InfraMachinePool.Spec.MaxPods
 	}
 
 	return agentPoolSpec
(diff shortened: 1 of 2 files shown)


azure/scope/managedcontrolplane_test.go
```diff
@@ -388,7 +391,7 @@ func getAzureMachinePoolWithScaling(name string, min, max int32) *infrav1.AzureM
 }
 
 func getAzureMachinePoolWithMaxPods(name string, maxPods int32) *infrav1.AzureManagedMachinePool {
-	managedPool := getAzureMachinePool(name, infrav1.NodePoolModeUser)
+	managedPool := getAzureMachinePool(name, infrav1.NodePoolModeSystem)
```
jsturtevant, line 394: Why is this being updated?

jsturtevant:
/retitle Set max pods for the first time creation of ManagedControlplane (AKS) cluster

azure/scope/managedcontrolplane_test.go
zmalik: to make it a valid request. We cannot create a cluster with all machine pools with mode `User`

CecileRobertMichon: isn't the exact same thing being done above for `getAzureMachinePoolWithScaling`? Should we be validating both System and User pools or are the functionally equivalent for this purpose?

azure/scope/managedcontrolplane.go
```diff
@@ -570,10 +572,6 @@ func (s *ManagedControlPlaneScope) AgentPoolSpec() azure.AgentPoolSpec {
 		agentPoolSpec.MinCount = s.InfraMachinePool.Spec.Scaling.MinSize
 	}
 
-	if s.InfraMachinePool.Spec.MaxPods != nil {
-		agentPoolSpec.MaxPods = s.InfraMachinePool.Spec.MaxPods
```
CecileRobertMichon, line 574: is the above equivalent because we expect MaxPods to never be nil because of defaulting (just double checking my understanding)?

azure/scope/managedcontrolplane.go
```diff
@@ -500,6 +500,7 @@ func (s *ManagedControlPlaneScope) GetAgentPoolSpecs(ctx context.Context) ([]azu
 			Replicas:          1,
 			OSDiskSizeGB:      0,
 			Mode:              pool.Spec.Mode,
+			MaxPods:           pool.Spec.MaxPods,
```
CecileRobertMichon, line 503: this is honestly easy to miss and it's not very surprising that a bug like this was introduced. I can easily see more similar bugs happening in the future unless reviewers are extra careful. Doesn't have to be in this PR, but it would be great if we can find a way to make this code a bit less error-prone, maybe by extracting the common logic from these two functions, thoughts?

azure/scope/managedcontrolplane.go
zmalik: its actually just to simplify the code as if it's nil we are ok to pass nil. As both are `*int32`
e.g in case of other fields like `OSDiskSizeGB` the nil check makes sense as we are taking a `*int32` and passing it as `int32`

azure/scope/managedcontrolplane_test.go
zmalik: yes, we should be testing both actually. But in case of scaling its not an immutable field so we just need to make sure it works for `AgentPoolSpec()` and in this case we have to make sure it works also for `GetAgentPoolSpecs()` which is the one complaining about no system nodepool present.

azure/scope/managedcontrolplane.go
zmalik: yes, I can take a stab on separate PR as we are getting different specs of agentpool from two different methods. Ideally, we should get only one same spec. 

As there are other fields that needs to be tested carefully like `VnetSubnetID`, `ResourceGroup` or clustername etc

CecileRobertMichon (commented):
/lgtm

jackfrancis (approved):
lgtm

PR #1946, opened 2022-01-04
Support multiple CIDR blocks and multiple subnets in AzureManagedControlPlane

### LINKED ISSUE #1687: Support multiple CIDR blocks and multiple subnets in AzureManagedControlPlane
/kind feature

Today `AzureManagedControlPlane` allows user to specify [one CIDR block and one subnet](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/exp/api/v1alpha4/azuremanagedcontrolplane_types.go#L117-L118) for its networking. We'd like it to support multiple CIDR blocks and multiple subnets (by changing the current spec to use a slice of `CIDRBlock`s and `Subnet`s.

This is a breaking change.

richardchen331:
**What type of PR is this?**
/kind feature

Today AzureManagedControlPlane allows user to specify one CIDR block and one subnet for its networking. This PR allows user to specify multiple CIDR blocks and multiple subnets (by changing the current spec to use a slice of CIDRBlocks and Subnets.

Fixes #1687

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Support multiple CIDR blocks and multiple subnets in AzureManagedControlPlane
```


exp/api/v1beta1/azuremanagedcontrolplane_types.go
```diff
@@ -186,16 +186,16 @@ type APIServerAccessProfile struct {
...
-	Subnet ManagedControlPlaneSubnet `json:"subnet,omitempty"`
+	Subnets []ManagedControlPlaneSubnet `json:"subnets,omitempty"`
 }
 
 // ManagedControlPlaneSubnet describes a subnet for an AKS cluster.
 type ManagedControlPlaneSubnet struct {
-	Name      string `json:"name"`
-	CIDRBlock string `json:"cidrBlock"`
+	Name       string   `json:"name"`
+	CIDRBlocks []string `json:"cidrBlocks"`
```
zmalik, line 198: how we maintain retro-compatibility with this change? if we remove `cidrBlock` ? 

we should provide a way to migrate to this new way of expressing cidrBlocks

richardchen331: @zmalik The old apis (e.g. v1alpha3, v1alpha4) are not changed, and I added conversion between the old api `cidrBlock` and the new api `cidrBlocks` (see for example `exp/api/v1alpha4/azuremanagedcontrolplane_conversion.go`)

zmalik: but what about v1beta1 API? if someone is using those?

richardchen331: @zmalik Thanks for pointing this out. @alexeldeib Do you think if we can (and how we can) introduce breaking changes like this?

zmalik: That's where API versions could help us.
I think we should strictly follow [kubernetes deprecation policy](https://kubernetes.io/docs/reference/using-api/deprecation-policy/) 

Which has the first rule that `API elements may only be removed by incrementing the version of the API group.`

CecileRobertMichon: @zmalik is correct that breaking changes should (normally) not be made outside API version increments, especially now that CAPZ is in beta / v1. Only new API fields should be added. Removing or changing the type of an existing field is considered a breaking change.

That being said, the AzureManagedControlPlane API is (still) in the experimental folder. I see 3 different approaches we can take:
1) wait for v1beta2 (not ideal, it might be a while before we're ready to release a new API version)
2) add a new `CIDRBlocks` field and deprecate the existing field `CIDRBlock`. In the code, always use `CIDRBlocks` if it is specified and if not fallback to `CIDRBlock`. Add logic to convert `CIDRBlock` to `CIDRBlocks` (array of len 1) when it is specified. 
3) make this as a breaking change with a clear release note since the AzureManagedControlPlane API is clearly documented as experimental with no backward compatibility guarantees. 

I personally would prefer 2 or 3.

richardchen331: Thanks for your input @zmalik @CecileRobertMichon ! I could update the PR to go with either option 2 or 3. Please let me know what you think and I can go ahead and make the change.

NovemberZulu: After reading through  Kubernetes deprecation policy, and API [API versioning](https://kubernetes.io/docs/reference/using-api/#api-versioning), I believe option 2 is the better one.
>  The schema and/or semantics of objects may change in incompatible ways in a **subsequent beta or stable release** (emphasis mine -- NZ).

zmalik: I also believe option 2 is the better one. As we set a good precedent for stable API without blocking any new features till a new API version is available.

CecileRobertMichon:
@richardchen331 are you still working on this one?

Issue #1965, opened 2022-01-12
capz AKS cluster test is failing


jsturtevant:
/kind failing-test

[Before submitting an issue, have you checked the [Troubleshooting Guide](https://capz.sigs.k8s.io/topics/troubleshooting.html)?]

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]
The test `capz-e2e.Workload cluster creation Creating an AKS cluster with a single control plane node and 1 node` is failing consistently: https://testgrid.k8s.io/sig-cluster-lifecycle-cluster-api-provider-azure#capz-periodic-e2e-full-main

**What did you expect to happen:**

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]
Looks like it might be https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1945

https://github.com/kubernetes-sigs/cluster-api-provider-azure/compare/fdae34a56...c507e7d47

**Environment:**

- cluster-api-provider-azure version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

CecileRobertMichon:
/close 

fixed by #1971

PR #1971, opened 2022-01-12
fix managed clusters e2e tests

zmalik:
**What type of PR is this?**

/kind cleanup

we are seeing failures in managed clusters e2e tests. Looks like we break the tests in `main` branch. Im trying to understand how we missed this as all PRs that are merged are 🟢 
the test is:
```
Workload cluster creation Creating an AKS cluster with a single control plane node and 1 node
```
this means a `machinePool` has only `1` replicas. Which means the minimum `maxPods` for that nodepool is `30`. The usual minimum allowed is `10` pods, only if the total in nodepool is `30` 

```
The minimum value that can be set for maximum pods per node is 10 if and only if the configuration of each node pool has space for a minimum of 30 pods. For example, setting the maximum pods per node to the minimum of 10 requires each individual node pool to have a minimum of 3 nodes.
```
https://docs.microsoft.com/en-us/azure/aks/configure-azure-cni#configure-maximum---new-clusters

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
"NONE"
```


jsturtevant:
fixes https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1965

jackfrancis:
@zmalik these regexex are wrong :(

https://github.com/kubernetes/test-infra/pull/24621/files

To explain why we didn't catch it (we didn't run the "exp" test on your PR!)

I'm in the process of fixing it.

CecileRobertMichon:

until we fix the regex

templates/test/ci/cluster-template-prow-aks-multi-tenancy.yaml
```diff
@@ -70,7 +70,7 @@ metadata:
   name: agentpool0
   namespace: default
 spec:
-  maxPods: 24
+  maxPods: 30
```
CecileRobertMichon, line 73: verify test failed because you need to modify the kustomize patch and then run make generate instead of changing the generated template directly (or you can cheat and manually change both as long as the generated output produces no diff)

zmalik: 🤦 will fix it now

jackfrancis (approved):
lgtm

CecileRobertMichon:
/lgtm
/approve

exp test passed

PR #1976, opened 2022-01-13
build aks agentpool spec through one method

zmalik:
**What type of PR is this?**
/kind cleanup

we are building `azure.AgentPoolSpec` through two different ways. 
This PR tries to unify those two ways so we don't end up with different logics for first time creation of agentpool and later additions to the machinepool or updates. 

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/azure/scope/managedcontrolplane.go
+++ b/azure/scope/managedcontrolplane.go
@@ -471,9 +471,9 @@ func (s *ManagedControlPlaneScope) ManagedClusterSpec() (azure.ManagedClusterSpe
 	return managedClusterSpec, nil
 }
 
-// GetAgentPoolSpecs gets a slice of azure.AgentPoolSpec for the list of agent pools.
-func (s *ManagedControlPlaneScope) GetAgentPoolSpecs(ctx context.Context) ([]azure.AgentPoolSpec, error) {
-	ctx, log, done := tele.StartSpanWithLogger(ctx, "scope.ManagedControlPlaneScope.GetAgentPoolSpecs")
+// GetAllAgentPoolSpecs gets a slice of azure.AgentPoolSpec for the list of agent pools.
+func (s *ManagedControlPlaneScope) GetAllAgentPoolSpecs(ctx context.Context) ([]azure.AgentPoolSpec, error) {
+	ctx, log, done := tele.StartSpanWithLogger(ctx, "scope.ManagedControlPlaneScope.GetAllAgentPoolSpecs")
 	defer done()
 
 	if len(s.AllNodePools) == 0 {
@@ -495,7 +495,7 @@ func (s *ManagedControlPlaneScope) GetAgentPoolSpecs(ctx context.Context) ([]azu
 		ammps           = make([]azure.AgentPoolSpec, 0, len(s.AllNodePools))
 		foundSystemPool = false
 	)
-	for _, pool := range s.AllNodePools {
+	for i, pool := range s.AllNodePools {
 		// Fetch the owning MachinePool.
 
 		ownerPool, err := capiexputil.GetOwnerMachinePool(ctx, s.Client, pool.ObjectMeta)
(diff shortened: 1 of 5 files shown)


CecileRobertMichon:
/ok-to-test

I'm happy to sponsor your membership in kubernetes-sig if you can find another sponsor

zmalik:
> I'm happy to sponsor your membership in kubernetes-sig if you can find another sponsor

🙇 thank you! I will ask in slack

jackfrancis (approved):
lgtm

azure/scope/managedcontrolplane_test.go
```diff
@@ -169,10 +169,12 @@ func TestManagedControlPlaneScope_PoolVersion(t *testing.T) {
...
 				{
-					Name:     "pool0",
-					SKU:      "Standard_D2s_v3",
-					Replicas: 1,
-					Mode:     "System",
+					Name:         "pool0",
+					SKU:          "Standard_D2s_v3",
+					Replicas:     1,
+					Mode:         "System",
+					Cluster:      "cluster1",
```
CecileRobertMichon, line 176: the refactor is adding Cluster and VnetSubnetID to the spec, is that significant?

jackfrancis: So currently we are constructing the pools differently, this PR is standardizing to a single constrution:

```
	agentPoolSpec := azure.AgentPoolSpec{
		Name:          to.String(managedMachinePool.Spec.Name),
		ResourceGroup: managedControlPlane.Spec.ResourceGroupName,
		Cluster:       managedControlPlane.Name,
		SKU:           managedMachinePool.Spec.SKU,
		Replicas:      replicas,
		Version:       normalizedVersion,
		VnetSubnetID: azure.SubnetID(
			managedControlPlane.Spec.SubscriptionID,
			managedControlPlane.Spec.ResourceGroupName,
			managedControlPlane.Spec.VirtualNetwork.Name,
			managedControlPlane.Spec.VirtualNetwork.Subnet.Name,
		),
		Mode:              managedMachinePool.Spec.Mode,
		MaxPods:           managedMachinePool.Spec.MaxPods,
		AvailabilityZones: managedMachinePool.Spec.AvailabilityZones,
		OsDiskType:        managedMachinePool.Spec.OsDiskType,
	}
```

The above, proposed standardized way of constructing a pool introduces the following additive, explicit property assignments to the pool construction in the `(s *ManagedControlPlaneScope) GetAgentPoolSpecs` method:

- `ResourceGroup`
- `Cluster`
- `VnetSubnetID`

(It's not clear why the addition of `ResourceGroup` didn't require a UT update like the other two.)

If those additions are controversial, I could understand better how to grok the impact here by understanding the difference between `func (s *ManagedControlPlaneScope) AgentPoolSpec()` and `func (s *ManagedControlPlaneScope) GetAgentPoolSpecs(ctx context.Context)`.

CecileRobertMichon: > If those additions are controversial, I could understand better how to grok the impact here by understanding the difference between func (s *ManagedControlPlaneScope) AgentPoolSpec() and func (s *ManagedControlPlaneScope) GetAgentPoolSpecs(ctx context.Context).

Trying to understand the impact myself (if any). It's possible that there is no functional impact and that we're just making something explicit that was previously set on Update but not on Create (for eg.). That would be my guess but double-checking to be sure.

Side note, the two function names are very _very_ similar and there is no way to know what the difference is between the two without reading the code comments and even then it's not super clear. Feel free to tell me it's a scope creep but it wouldn't be a bad idea to rename one/both of them to make the difference more obvious. Maybe something like `GetAllAgentPoolSpecs`?

zmalik: > Side note, the two function names are very very similar and there is no way to know what the difference is between the two without reading the code comments and even then it's not super clear

➕ it is confusing. I will do the name change in this PR. The purpose of the PR was to improve the developer's experience when working with AKS `AgentPoolSpec`. So I guess this falls into this also.

jackfrancis (approved):
lgtm

azure/scope/managedcontrolplane_test.go
jackfrancis: Well on 2nd thought looking at the failed apidiff test job it seems like we have to update an interface spec in order to change the method name. IMO that's not a great thing to do given that the interface type spec is exported, but perhaps this is a change we can make and introduce in the next version release?

In any event I didn't fully review the renames, so I'll make some comments there. Thanks for keeping with this @zmalik! :)

azure/scope/managedcontrolplane.go
```diff
@@ -472,7 +472,7 @@ func (s *ManagedControlPlaneScope) ManagedClusterSpec() (azure.ManagedClusterSpe
 }
 
 // GetAgentPoolSpecs gets a slice of azure.AgentPoolSpec for the list of agent pools.
```
jackfrancis, line 474: the above comment needs to change to reflect the new method name if we're going to keep that, and ditto the `"scope.ManagedControlPlaneScope.GetAgentPoolSpecs"` string reference in the logger needs to be changed to `"scope.ManagedControlPlaneScope.GetAllAgentPoolSpecs"`

zmalik: good catch! thanks

azure/scope/managedcontrolplane_test.go
zmalik: > IMO that's not a great thing to do given that the interface type spec is exported, but perhaps this is a change we can make and introduce in the next version release?

that's a good point. honestly didn't think about it. I can revert the method name change as unifying the `AgentPoolSpec` definition feels more helpful now that we are introducing features to AKS. And revisit the method names later.

jackfrancis: @CecileRobertMichon wdyt?

In any event, even if there is an established pattern for updating an exported interface, I think that change should be done discretely (not as a side-effect in a larger PR).

CecileRobertMichon:
/lgtm
/approve

changes to the exported functions are acceptable in minor releases. the apidiff checker is mostly there to catch breaking changes in the API

azure/scope/managedcontrolplane.go
CecileRobertMichon: removed the lgtm to allow you to update this but otherwise lgtm

zmalik: updated! also rebased and squashed the commits

CecileRobertMichon (commented):
/lgtm

PR #1978, opened 2022-01-13
Set Control Plane Endpoint Optionally

### LINKED ISSUE #1977: Allow user to set Spec.ControlPlaneEndpoint
/kind feature

Users should be able to set the AzureCluster Spec.ControlPlaneEndpoint as part of their defined cluster. The provider should not write over this if it is set by the user. This allows users to bring their own infrastructure (WAF, Custom Load Balancer) between the control plane endpoint and the deployed workload cluster.

This issue is in response to discussion about the effects of setting this to early in the bring up process. We should test this to verify CAPI doesn't do that, if it does we should open tickets with them appropriately.

Comment from original post [here](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1959#discussion_r783553057)

**Environment:**

- cluster-api-provider-azure version: v1.1.0
- Kubernetes version: (use `kubectl version`): v1.22 (ish)
- OS (e.g. from `/etc/os-release`): Ubuntu, OSX, Windows, Redhat, etc...

dmlb2000:
**What type of PR is this?**

/kind feature

This sets the Control Plane Endpoint optionally if it's not already
set and makes sure the Host and Port can be set independently.

Fixes #1977

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Allow Control Plane Endpoint to be set by user.
```

### DIFF
--- a/api/v1beta1/azurecluster_types.go
+++ b/api/v1beta1/azurecluster_types.go
@@ -45,7 +45,8 @@ type AzureClusterSpec struct {
 
 	Location string `json:"location"`
 
-	// ControlPlaneEndpoint represents the endpoint used to communicate with the control plane.
+	// ControlPlaneEndpoint represents the endpoint used to communicate with the control plane. It is not recommended to set
+	// this when creating an AzureCluster as CAPZ will set this for you. However, if it is set, CAPZ will not change it.
 	// +optional
 	ControlPlaneEndpoint clusterv1.APIEndpoint `json:"controlPlaneEndpoint,omitempty"`
 
--- a/api/v1beta1/azurecluster_webhook.go
+++ b/api/v1beta1/azurecluster_webhook.go
@@ -75,6 +75,20 @@ func (c *AzureCluster) ValidateUpdate(oldRaw runtime.Object) error {
 		)
 	}
 
+	if old.Spec.ControlPlaneEndpoint.Host != "" && c.Spec.ControlPlaneEndpoint.Host != old.Spec.ControlPlaneEndpoint.Host {
+		allErrs = append(allErrs,
+			field.Invalid(field.NewPath("spec", "ControlPlaneEndpoint", "Host"),
+				c.Spec.ControlPlaneEndpoint.Host, "field is immutable"),
+		)
+	}
+
+	if old.Spec.ControlPlaneEndpoint.Port != 0 && c.Spec.ControlPlaneEndpoint.Port != old.Spec.ControlPlaneEndpoint.Port {
+		allErrs = append(allErrs,
+			field.Invalid(field.NewPath("spec", "ControlPlaneEndpoint", "Port"),
+				c.Spec.ControlPlaneEndpoint.Port, "field is immutable"),
+		)
+	}
+
 	if !reflect.DeepEqual(c.Spec.AzureEnvironment, old.Spec.AzureEnvironment) {
 		// The equality failure could be because of default mismatch between v1alpha3 and v1beta1. This happens because
 		// the new object `r` will have run through the default webhooks but the old object `old` would not have so.
--- a/api/v1beta1/azurecluster_webhook_test.go
+++ b/api/v1beta1/azurecluster_webhook_test.go
@@ -20,6 +20,7 @@ import (
 	"testing"
 
 	. "github.com/onsi/gomega"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 )
 
 func TestAzureCluster_ValidateCreate(t *testing.T) {
@@ -34,6 +35,18 @@ func TestAzureCluster_ValidateCreate(t *testing.T) {
 			name: "azurecluster with pre-existing vnet - valid spec",
 			cluster: func() *AzureCluster {
 				return createValidCluster()
+			}(),
+			wantErr: false,
+		},
+		{
+			name: "azurecluster with pre-existing control plane endpoint - valid spec",
+			cluster: func() *AzureCluster {
+				cluster := createValidCluster()
+				cluster.Spec.ControlPlaneEndpoint = clusterv1.APIEndpoint{
+					Host: "apiserver.example.com",
+					Port: 8443,
+				}
+				return cluster
 			}(),
 			wantErr: false,
 		},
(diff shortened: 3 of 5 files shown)


api/v1beta1/azurecluster_webhook.go
```diff
@@ -75,6 +75,13 @@ func (c *AzureCluster) ValidateUpdate(oldRaw runtime.Object) error {
 		)
 	}
 
+	if !reflect.DeepEqual(c.Spec.ControlPlaneEndpoint, old.Spec.ControlPlaneEndpoint) {
```
CecileRobertMichon, line 78: we should allow setting each value if it is empty in this case to preserve existing behavior since the controller might patch it after the cluster is created

dmlb2000: I thought I did that, but it missed somehow.

dmlb2000:
Just trying...

/ok-to-test

api/v1beta1/azurecluster_webhook.go
```diff
@@ -75,6 +75,20 @@ func (c *AzureCluster) ValidateUpdate(oldRaw runtime.Object) error {
 		)
 	}
 
+	if !reflect.DeepEqual(c.Spec.ControlPlaneEndpoint.Host, old.Spec.ControlPlaneEndpoint.Host) {
```
CecileRobertMichon, line 78: this wouldn't allow me to go from `""` to `"hostname"` because "" != hostname. This should be 

```suggestion
	if !old.Spec.ControlPlaneEndpoint.Host != "" && !reflect.DeepEqual(c.Spec.ControlPlaneEndpoint.Host, old.Spec.ControlPlaneEndpoint.Host) {
```

api/v1beta1/azurecluster_webhook.go
```diff
@@ -75,6 +75,20 @@ func (c *AzureCluster) ValidateUpdate(oldRaw runtime.Object) error {
...
+		allErrs = append(allErrs,
+			field.Invalid(field.NewPath("spec", "ControlPlaneEndpoint", "Host"),
+				c.Spec.ControlPlaneEndpoint.Host, "field is immutable"),
+		)
+	}
+
+	if !reflect.DeepEqual(c.Spec.ControlPlaneEndpoint.Port, old.Spec.ControlPlaneEndpoint.Port) {
```
CecileRobertMichon, line 85: ditto

CecileRobertMichon (commented):
/ok-to-test

would be great to add a small unit test to validate this

api/v1beta1/azurecluster_webhook.go
dmlb2000: So, is there something that `reflect.DeepEqual()` gives us when we are just comparing two strings? Host isn't that complex is it?

dmlb2000: And this is more of a style question...

Which do you prefer?

```
if a != 'foo' && a != b { ... }
```
or
```
if !(a == 'foo'  || a == b) { ... }
```

I personally like the latter because I my brain turns the 'if !' into 'unless'. Then I read the conditionals as 'if a equals foo or a equals b skip the block'. However, I'm mostly seeing the former in the code, is that a go'ism?

CecileRobertMichon: > So, is there something that reflect.DeepEqual() gives us when we are just comparing two strings?

not AFAIK. You can use `==`, should be the same.

> However, I'm mostly seeing the former in the code, is that a go'ism?

I don't think one is preferred over the other, it's mostly style / personal preference. Personally, I read the first as "if old is not empty and old is not equal to new, then return an error" which matches the intent of the validation here but logically they are exactly the same.

api/v1beta1/azurecluster_default_test.go
```diff
@@ -24,8 +24,76 @@ import (
...
+				},
+			},
+		},
+	}
+
+	for name := range cases {
+		c := cases[name]
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+			c.cluster.setResourceGroupDefault()
```
CecileRobertMichon, line 87: why call setResourceGroupDefault default here? You should be able to add the test case to the existing `TestAzureCluster_ValidateUpdate` function in azurecluster_webhook_test.go

dmlb2000: Yeah, I'm still working through the testing code... I may need to add a webhook test as well I'm guessing?

dmlb2000:
I think I'm good with where this pull request is at this point.

I started looking through the end-to-end testing to try and test the scenario, setting the control plane endpoint when creating the Azure Cluster document. But I got lost pretty quick, it's a very different system. Feels like standard Github Actions with standard K8S testing on top of CAPI testing on top of Azure testing... It'd be nice to have a document on where to start with that...

api/v1beta1/azurecluster_default_test.go
```diff
@@ -24,8 +24,73 @@ import (
 	"github.com/Azure/go-autorest/autorest/to"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 )
 
+func TestControlPlaneEndpointDefault(t *testing.T) {
```
CecileRobertMichon, line 30: what is this test testing? We're not setting any defaults for the control plane endpoint in the webhook AFAIK

dmlb2000: This was just a test to verify that the Host and Port could be set independently, if it doesn't make sense to keep that around I can delete that test.

CecileRobertMichon: let's remove it

api/v1beta1/azurecluster_webhook_test.go
```diff
@@ -105,6 +118,26 @@ func TestAzureCluster_ValidateUpdate(t *testing.T) {
 		cluster    *AzureCluster
 		wantErr    bool
 	}{
+		{
+			name: "azurecluster with pre-existing control plane endpoint - valid spec",
```
CecileRobertMichon, line 122: can we also add a test case that checks it fails and returns an error when host / port is modified? and also that it doesn't return an error if host/port is set from empty

CecileRobertMichon:
> t'd be nice to have a document on where to start with that...

We have an in-progress doc about which jobs we run for e2e test https://capz.sigs.k8s.io/developers/jobs.html

Our CI system uses the standard k8s Prow tooling which you can find docs for [here](https://github.com/kubernetes/test-infra/blob/master/prow/README.md)

I wouldn't recommend adding an e2 test for this one yet since we still need to figure out if anything needs to change on the CAPI side to support this and it's also an edge case, I'd expect most users to stay away from setting their own ControlPlaneEndpoint. 

Speaking of, it would be good to add a warning to the [description](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/api/v1beta1/azurecluster_types.go#L48) of the field about only setting this if you know what you're doing and telling users that CAPZ will set the host and port for them if it's left empty.

api/v1beta1/azurecluster_webhook_test.go
dmlb2000: So, I thought I did the former with this test, line 139 wants the error and oldCluster and cluster variables have different host/port for the control plane endpoints.

I will add the second test.

dmlb2000: I think I got the following tests in at this point, unless I'm not interpreting what this function is doing (which is entirely possible).

 * Check for error when changing a cluster with an existing endpoint
 * Check setting an empty endpoint works without error

dmlb2000:
@CecileRobertMichon did I address all your concerns?

I did finally get things up and going in our environment. I was hoping to use our on-site k8s to bootstrap cluster-api and build k8s in Azure from there. However, that didn't pan out... So, new architecture!

I settled on a small admin VM that has two jobs:

- Proxy and API Endpoint (using this patch)
- Installing Cluster API to build out the cluster.

I built out the admin VM using k0s and it seemed work out just fine.

If we merge this I think we can close #1959 as the rest of cluster API did wait nicely for the API Endpoint to come up.

CecileRobertMichon:
> I did finally get things up and going in our environment.
[...]
If we merge this I think we can close #1959 as the rest of cluster API did wait nicely for the API Endpoint to come up.

That's great to hear 🎉 

Thank you so much for working with me on figuring out a better solution, I'm happy with what we ended up with. PR lgtm, hoping to get another pair of 👀  on it though.

/lgtm
/assign @shysank

shysank (commented):
lgtm, with a small nit.

api/v1beta1/azurecluster_types.go
```diff
@@ -45,7 +45,8 @@ type AzureClusterSpec struct {
 
 	Location string `json:"location"`
 
-	// ControlPlaneEndpoint represents the endpoint used to communicate with the control plane.
+	// ControlPlaneEndpoint represents the endpoint used to communicate with the control plane. It is not recommended to set
+	// this when creating an AzureCluster as the CAPZ will set this for you. However, if it is set CAPZ will not change it.
```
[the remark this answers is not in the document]
dmlb2000: Should be fixed.

Issue #1982, opened 2022-01-18
golang-ci lint should be in parity with capi 


sonasingh46:
/kind bug

**What did you expect to happen:**
Did a comparison of golang-ci linters used in core capi vs capz and found out that couple of linters are missing in capz. If it makes sense, we could add those missing linters in capz too. The following section has the comparison matrix. 

TL;DR

Missing linters in CAPZ but present in CAPI:  
- gci
- gocritic
- ifshort
- predeclared
- stylecheck
- thelper
- unparam

We can discuss about these and if it makes sense then we can add these linters too. 

Ref link for what these linters do : 
https://golangci-lint.run/usage/linters/

**Anything else you would like to add:**
golang-ci lint comparison matrix: 

<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-24fe2445-7fff-3148-7a4d-deba751d0d33"><div dir="ltr" style="margin-left:0pt;" align="left">

S.no | Linters[golang-ci] | CAPI | CAPZ
-- | -- | -- | --
1 | asciicheck | ✔ | ✔
2 | bodyclose | ✔ | ✔
3 | deadcode | ✔ | ✔
4 | depguard | ✔ | ✔
5 | dogsled | ✔ | ✔
6 | errcheck | ✔ | ✔
7 | errorlint | ✖ | ✔
8 | exportloopref | ✔ | ✔
9 | gci | ✔ | ✖
10 | goconst | ✔ | ✔
11 | gocritic | ✔ | ✖
12 | gocyclo | ✔ | ✔
13 | godot | ✔ | ✔
14 | gofmt | ✔ | ✔
15 | goimports | ✔ | ✔
16 | goprintffuncname | ✔ | ✔
17 | gosec | ✔ | ✔
18 | gosimple | ✔ | ✔
19 | govet | ✔ | ✔
20 | ifshort | ✔ | ✖
21 | importas | ✔ | ✔
22 | ineffassign | ✔ | ✔
23 | misspell | ✔ | ✔
24 | nakedret | ✔ | ✔
25 | nilerr | ✔ | ✔
26 | nolintlint | ✔ | ✔
27 | prealloc | ✔ | ✔
28 | predeclared | ✔ | ✖
29 | revive | ✔ | ✔
30 | rowserrcheck | ✔ | ✔
31 | staticcheck | ✔ | ✔
32 | structcheck | ✔ | ✔
[...]

PR #1985, opened 2022-01-19
chore(golangci-lint): add unparam linter

sonasingh46:

/kind cleanup

This PR adds `unparam` linter 

Reference umbrella issue: https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1982

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
None
```

### DIFF
--- a/azure/services/scalesets/scalesets_test.go
+++ b/azure/services/scalesets/scalesets_test.go
@@ -265,7 +265,7 @@ func TestReconcileVMSS(t *testing.T) {
 				s.ScaleSetSpec().Return(defaultSpec).AnyTimes()
 				createdVMSS := newDefaultVMSS("VM_SIZE")
 				instances := newDefaultInstances()
-				_ = setupDefaultVMSSInProgressOperationDoneExpectations(s, m, createdVMSS, instances)
+				setupDefaultVMSSInProgressOperationDoneExpectations(s, m, createdVMSS, instances)
 				s.DeleteLongRunningOperationState(defaultSpec.Name, scope.ScalesetsServiceName)
 			},
 		},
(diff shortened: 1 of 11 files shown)


azure/services/scalesets/scalesets_test.go
```diff
@@ -265,7 +265,7 @@ func TestReconcileVMSS(t *testing.T) {
 				s.ScaleSetSpec().Return(defaultSpec).AnyTimes()
 				createdVMSS := newDefaultVMSS("VM_SIZE")
 				instances := newDefaultInstances()
-				_ = setupDefaultVMSSInProgressOperationDoneExpectations(s, m, createdVMSS, instances)
+				setupDefaultVMSSInProgressOperationDoneExpectations(s, m, createdVMSS, instances)
```
mboersma, line 268: Is this something that `unparam` enforces? I prefer assignment to `_` in general, because it documents that you meant to ignore the error (or other) return value. WDYT?

sonasingh46: The return value from that function is nowhere used that is why it is enforced. If we want to use the return value somewhere in the future we can modify accordingly. And wherever we do not want we can use the `_` assignment.

sonasingh46: Copying @sbueringer here in case he has some thoughts.

sbueringer: Just my two cents. In general I would also prefer to explicitly assign to _ when a return value is ignored (especially an error).

*But* if I see correctly this PR also removes the return parameter entirely because it's not used anywhere.

I think we have the following options:
* (as is) remove the return parameter because it's unused
* keep the return parameter because we assume it's useful in the future, then we can add ~ `//nolint:unparam // We want to keep the return parameter even though it is currently unused`

CecileRobertMichon: +1 to removing the `_` assignment since that function is no longer returning a value, we're not actually ignoring a return value

CecileRobertMichon (commented):
/lgtm

mboersma (approved):
/lgtm

devigned (approved):
/lgtm
/approve

PR #1991, opened 2022-01-20
skip updating the AKS agentpoolprofile count if autoscaling is enabled (managed clusters)

zmalik:
**What type of PR is this?**

/kind cleanup

when autoscaling is enabled for AKS we are still trying to update the agentpoolprofile with initial `replicas`.
autoscaling can continuously change the count of agentpool, so we should skip updating the count if autoscaling is enabled. 
Otherwise, we are updating the agentpool on each reconcile. 

Also doing a cleanup to use `klog` to print diff for any changes in agentpool

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/azure/services/agentpools/agentpools.go
+++ b/azure/services/agentpools/agentpools.go
@@ -25,6 +25,7 @@ import (
 	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/google/go-cmp/cmp"
 	"github.com/pkg/errors"
+	"k8s.io/klog/v2"
 
 	infrav1alpha4 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
 	"sigs.k8s.io/cluster-api-provider-azure/azure"
@@ -133,10 +134,16 @@ func (s *Service) Reconcile(ctx context.Context) error {
 			},
 		}
 
+		if profile.EnableAutoScaling != nil && *profile.EnableAutoScaling {
+			// unset the nodepool count if autoscaling is enabled
+			normalizedProfile.Count = nil
+			existingProfile.Count = nil
+		}
+
 		// Diff and check if we require an update
 		diff := cmp.Diff(normalizedProfile, existingProfile)
 		if diff != "" {
-			log.V(2).Info(fmt.Sprintf("Update required (+new -old):\n%s", diff))
+			klog.V(2).Info(fmt.Sprintf("Update required (+new -old):\n%s", diff))
 			err = s.Client.CreateOrUpdate(ctx, agentPoolSpec.ResourceGroup, agentPoolSpec.Cluster, agentPoolSpec.Name, profile)
 			if err != nil {
 				return errors.Wrap(err, "failed to create or update agent pool")
--- a/azure/services/agentpools/agentpools_test.go
+++ b/azure/services/agentpools/agentpools_test.go
@@ -261,6 +261,42 @@ func TestReconcile(t *testing.T) {
 				}, nil)
 			},
 		},
+		{
+			name: "skip PoolProfile count update if autoscaling is enabled",
+			agentPoolsSpec: azure.AgentPoolSpec{
+				Name:              "my-agent-pool-autoscaling",
+				ResourceGroup:     "my-rg",
+				Cluster:           "my-cluster",
+				SKU:               "Standard_D2s_v3",
+				Version:           to.StringPtr("9.99.9999"),
+				Replicas:          1,
+				EnableAutoScaling: to.BoolPtr(true),
+				MaxCount:          to.Int32Ptr(2),
+				MinCount:          to.Int32Ptr(1),
+				OSDiskSizeGB:      100,
+				MaxPods:           to.Int32Ptr(12),
+				OsDiskType:        to.StringPtr(string(containerservice.OSDiskTypeEphemeral)),
+			},
+			expectedError: "",
+			expect: func(m *mock_agentpools.MockClientMockRecorder) {
+				m.Get(gomockinternal.AContext(), "my-rg", "my-cluster", "my-agent-pool-autoscaling").Return(containerservice.AgentPool{
+					ManagedClusterAgentPoolProfileProperties: &containerservice.ManagedClusterAgentPoolProfileProperties{
+						Count:               to.Int32Ptr(2), // get the scaled up count
+						OsDiskSizeGB:        to.Int32Ptr(100),
+						VMSize:              to.StringPtr(string(containerservice.VMSizeTypesStandardD2sV3)),
+						OsType:              containerservice.OSTypeLinux,
+						OrchestratorVersion: to.StringPtr("9.99.9999"),
+						ProvisioningState:   to.StringPtr("Succeeded"),
+						VnetSubnetID:        to.StringPtr(""),
+						EnableAutoScaling:   to.BoolPtr(true),
+						MaxPods:             to.Int32Ptr(12),
+						MaxCount:            to.Int32Ptr(2),
+						MinCount:            to.Int32Ptr(1),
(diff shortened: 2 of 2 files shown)


azure/services/agentpools/agentpools.go
```diff
@@ -133,10 +134,16 @@ func (s *Service) Reconcile(ctx context.Context) error {
...
+			// unset the nodepool count if autoscaling is enabled
+			normalizedProfile.Count = nil
+			existingProfile.Count = nil
+		}
+
 		// Diff and check if we require an update
 		diff := cmp.Diff(normalizedProfile, existingProfile)
 		if diff != "" {
-			log.V(2).Info(fmt.Sprintf("Update required (+new -old):\n%s", diff))
+			klog.V(2).Info(fmt.Sprintf("Update required (+new -old):\n%s", diff))
```
CecileRobertMichon, line 146: why switch to klog here?

zmalik: will update in the description. Following the [managedclusters]( https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/services/managedclusters/managedclusters.go#L325)

`klogs` does a pretty print format 
```
I0120 21:01:07.657204       1 managedclusters.go:325] Update required (+new -old):
  &containerservice.ManagedCluster{
  	Response: {},
  		ProvisioningState: nil,
  		PowerState:        nil,
  		MaxAgentPools:     nil,
- 		KubernetesVersion: &"1.22.4",
+ 		KubernetesVersion: &"1.21.2",
  		DNSPrefix:         nil,
  		FqdnSubdomain:     nil,
...
```

while in agentpools we were getting in one line with escape chars

azure/services/agentpools/agentpools.go
```diff
@@ -133,10 +134,16 @@ func (s *Service) Reconcile(ctx context.Context) error {
 			},
 		}
 
+		if profile.EnableAutoScaling != nil && *profile.EnableAutoScaling {
+			// unset the nodepool count if autoscaling is enabled
```
jackfrancis, line 138: Is setting the count to nil idiomatic? It would seem more intuitive to simply inherit the current (at time of reconcile) count value from AKS itself. Not sure if there's a precedent for capz to passively consume (rather than enforce) state changes, but setting it to nil seems weird.

Lemme know if I'm missing something!

zmalik: the issue with setting any value is that we potentially end up in a race condition with cluster autoscaler. Where it scaled up or down a node and try to undo it with our updates later. 

I did a test with not setting the count and it didn't remove the count from Azure API, but it just didn't change it.  
```
        "count": 2,
         ...
        "maxPods": 60,
        "type": "VirtualMachineScaleSets",
        "maxCount": 999,
        "minCount": 0,
        "enableAutoScaling": true,
```

I think the intention is that behaviour, that we let CA do the job and don't interfere with the count. 
We do have precedent to passively consume in CAPZ but we introduce a possible race condition knowing there also. I feel here we can avoid it. But if we agree otherwise also sounds good.

azure/services/agentpools/agentpools.go
CecileRobertMichon: wonder if that's something we can fix in our logger instead... see #1806 

we've added traces to the logger and correlation IDs which are useful for debugging

CecileRobertMichon: cc @devigned

azure/services/agentpools/agentpools.go
jackfrancis: What if we did this instead earlier in the code flow, immediately after we have determined that this is not a "create" operation?:

```
		if to.Bool(profile.EnableAutoScaling) {
			if agentPoolSpec.Replicas != to.Int32(existingPool.Count) {
				s.scope.SetAgentPoolReplicas(to.Int32(existingPool.Count))
				log.V(2).Info("AKS AutoScaling enabled, updating Replicas to %d for Pool %s", agentPoolSpec.Replicas, agentPoolSpec.Name)
				log.V(2).Info("After ensuring that pool count is synchronized between capz and AKS, we will apply any other property updates in a future reconciliation")
				return nil
			}
		}
```

Does that break too many architectural boundaries?

cc @CecileRobertMichon @devigned

zmalik: even if `agentPoolSpec.Replicas == to.Int32(existingPool.Count)` doesn't guarantee that between the fetch of the existing pool and the update call to Azure API  cluster autoscaler might have changed the count. If we later set the count in the update request.

But I like `s.scope.SetAgentPoolReplicas(to.Int32(existingPool.Count))` which we should def do also 👍

azure/services/agentpools/agentpools.go
zmalik: got it! I will remove the change from this PR and later take a look at how we can fix that in our logger

azure/services/agentpools/agentpools.go
zmalik: stepping back, another approach that might be cleaner is not to allow setting replicas as it is optional, when autoscaling is enabled. 
we can add that validation for autoscaling in our webhooks.

zmalik: as updating replicas count to spec feels strange. We already have the actual count of nodes in status through cluster API which uses the clusterClient to fetch the actual nodes.

CecileRobertMichon: Looking at the CAPI + autoscaler integration, cluster-autoscaler takes control of spec.Replicas in that case. I think the issue here is that the AKS autoscaler is directly interacting with the Azure resources, not going through CAPZ CRDs to scale up/down.

I think this discussion is very similar to https://github.com/kubernetes-sigs/cluster-api/pull/5685 where there is a discussion ongoing about pro vs. against setting replicas to nil when autoscaling through the cloud provider is engaged. 

If we do decide to go the nil replicas route, I do agree it would be a better experience to not allow setting replicas at all, rather than overriding the replica count to nil and potentially silently ignoring manual replica count updates.

jackfrancis: We need to use the `Replicas` field during cluster or node pood creation even if autoscaling is enabled, right?

Can we simply allow `Replicas` to be a configurable field during create only, and then mark it as read-only thereafter? Whether or not to update it during reconciliations I suppose is up to whether or not we want to keep the field "current" with reality so that it can be used as a simple read-only node count property.

zmalik: > We need to use the Replicas field during cluster or node pood creation even if autoscaling is enabled, right?

we can skip it as it's optional, the default value on Azure API is 1. So we can delegat that to Azure API. Even if the nodepool starts from 0 nodes, with autoscaling enabled it should not be a problem. 

> Can we simply allow Replicas to be a configurable field during create only, and then mark it as read-only thereafter?

that is almost what this PR intention was. I didn't change the logic for `create`  but only when doing `update`, ignore(set to nil) the replicas count if autoscaling is enabled.

> If we do decide to go the nil replicas route, I do agree it would be a better experience to not allow setting replicas at all, rather than overriding the replica count to nil and potentially silently ignoring manual replica count updates.

This is a certain use-case when node autoscaling is enabled, so I feel going with this option would be the best. Given all other options.

zmalik: picking this up again.
not allowing setting replicas would be in cluster API level. But we could fail the cluster autoscaling feature validation if the machinepool object has replicas. 

the workflow of updating the replicas with the actual node count of VMSS can be done in separate PR? if we decide to go down that path or even use https://github.com/kubernetes-sigs/cluster-api/pull/5685

CecileRobertMichon (commented):
See #2444

Issue #1983, opened 2022-01-18
Support "managed" semantics for an existing private DNS zone


dlipovetsky:
/kind feature

**Describe the solution you'd like**
I have to use an existing private DNS zone. I am not authorized to delete the zone, but only to manage vnet links and records in the zone. I want CAPZ to manage the vnet link and records in this zone. When CAPZ deletes the cluster, I want it to only delete the records and vnet link it created, and to not delete the zone itself.

**Anything else you would like to add:**
CAPZ [sees a private DNS zone as either "managed" or "unmanaged."](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/d0605831a48a551f1292b21c1c6a3a1bdbec8fc5/azure/services/privatedns/privatedns.go#L53-L154) If the zone is "managed," then CAPZ
- manages a vnet link in the zone
- manages the records in the zone
- deletes the zone when it deletes the cluster

However, if I configure CAPZ to use an existing private DNS zone, then CAPZ will treat is as "unmanaged," by default. Afterward, I can add a tag to the private DNS zone, and CAPZ will treat it as "managed." However, CAPZ [deletes a "managed" zone](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/d0605831a48a551f1292b21c1c6a3a1bdbec8fc5/azure/services/privatedns/privatedns.go#L188-L194) when it deletes the cluster.

**Environment:**

- cluster-api-provider-azure version: v1.1.0
- Kubernetes version: (use `kubectl version`):
- OS (e.g. from `/etc/os-release`):

CecileRobertMichon:
@dlipovetsky this seems to be a bug. We shouldn't return [here](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/d0605831a48a551f1292b21c1c6a3a1bdbec8fc5/azure/services/privatedns/privatedns.go#L74) when the private zone is unmanaged, just skip creating/updating the zone, since the management of vnet links is separate.

Also, that log message about tagging the zone manually is misleading, it should only apply to DNS zones that were created with previous versions of CAPZ, before tags were being applied.

shysank:
/close
closing this as this addressed by https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2007

Issue #1872, opened 2021-11-17
CreateOrUpdate should not be  called for existing zone


sonasingh46:
`CreateOrUpdateZone` should not be called if there is already a zone existing since we don't allow changes to any of its properties

_Originally posted by @shysank in https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1791#discussion_r749571484_

PR #2007, opened 2022-01-26
Make private dns reconcile/delete async

### LINKED ISSUE #1715: async privatedns

shysank:
**What type of PR is this?**
/kind cleanup
/kind bug

This pr refactors privatedns reconciliation to use async pattern set in #1541. A few things to note:
- Private dns has 3 resources that needs to be reconciled: `Zone`, `VirtualNetworkLink`, and `Records`. This is done by treating each resource as a separate spec, and calling `createResource`/`deleteResource` independently for each of them.
- Although there are 3 resources, only one condition: `PrivateDnsReadyCondition` is updated to indicate status of reconciliation.
- `Delete` is **NOT** implemented for `Records` because it was not implemented in the current version. Implementing this is not trivial because deleting record requires a special parameter called `RecordType` apart from the usual suspects (resourceName, resourceGroupName, ownerResourceName). For create, I did a work around by reverse engineering the record type from `RecordSetProperties` set in parameters. 
- `createOrUpdateAsync` for `Records` is **Synchronous** operation. This is because azure api for records set does not have support for futures.
- I have split `client.go` into 3 files for each of the sub resources mentioned above as it became too large and difficult to navigate.
- This pr also fixes a [bug](https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1983) about managing zones and links separately.

Fixes #1715 #1983 #1872 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Make private dns reconcile/delete async
```


azure/services/privatedns/privatedns.go
```diff
@@ -19,205 +19,252 @@ package privatedns
...
+	PrivateDNSSpec() (zoneSpec azure.ResourceSpecGetter, linksSpec, recordsSpec []azure.ResourceSpecGetter)
 }
 
 // Service provides operations on Azure resources.
 type Service struct {
-	Scope Scope
-	client
+	Scope              Scope
+	zoneClient         async.Creator
+	vnetLinkClient     async.Creator
```
shysank, line 49: We need this because `deleteZone` and `deleteLink` requires to do a `GET` to check if the resource is managed or not. It seems a bit weird to use `Creator` to do a GET, though.

azure/services/privatedns/privatedns.go
```diff
@@ -19,205 +19,252 @@ package privatedns
...
-		// will allow the reconciliation to continue
-		if err != nil && azure.ResourceNotFound(err) {
-			isManaged = true
-		}
-		if !isManaged {
-			log.V(1).Info("Skipping reconciliation of unmanaged private DNS zone", "private DNS", zoneSpec.ZoneName)
-			log.V(1).Info("Tag the DNS manually from azure to manage it with capz."+
-				"Please see https://capz.sigs.k8s.io/topics/custom-dns.html#manage-dns-via-capz-tool", "private DNS", zoneSpec.ZoneName)
-			return nil
-		}
```
shysank, line 75: I have moved this logic to `spec.Parameters` since we already do a GET to fetch the existing resource. Same for vnet links.

azure/services/privatedns/spec.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+// ResourceGroupName returns the name of the resource group of the private dns zone.
+func (z ZoneSpec) ResourceGroupName() string {
+	return z.ResourceGroup
+}
+
+// Parameters returns the parameters for the private dns zone.
+func (z ZoneSpec) Parameters(existing interface{}) (params interface{}, err error) {
+	_, log, done := tele.StartSpanWithLogger(context.TODO(), "privatedns.ZoneSpec.Parameters")
```
shysank, line 74: Had to pass `context.TODO` to get a logger since we don't pass context here. Open to suggestions.

shysank:
cc @CecileRobertMichon @Jont828

azure/services/privatedns/privatedns.go
CecileRobertMichon: here's how I dealt with this for vnets: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1921/files#diff-f45b26f09f47e6be08607d767760708b985d428a1dc4b9a81f0640ee3e7ee656R56

https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1921/files#diff-f137c2f0a9f5eecc0fb2c4816d9ec073882fdc06681fe40855fea0dda7771324R35

shysank: did you consider refactoring `Creator` 
```
// Creator is a client that can create or update a resource asynchronously.
type Creator interface {
	FutureHandler
	CreateOrUpdateAsync(ctx context.Context, spec azure.ResourceSpecGetter, parameters interface{}) (result interface{}, future azureautorest.FutureAPI, err error)
        Getter
}

type Getter interface {
 Get(ctx context.Context, spec azure.ResourceSpecGetter) (result interface{}, err error)
}
```
so that we don't have to duplicate this?

shysank: Is it because, `Get` in this case is not part of async, that you created a new `Getter`?

CecileRobertMichon: > did you consider refactoring Creator

I hadn't, but that's a great idea

I was mostly trying to refactor the Client interface to make it more specific since we only need Get now and everything else is being handled by async.Reconciler

```
type Client interface {
	Get(context.Context, string, string) (network.VirtualNetwork, error)
	CreateOrUpdate(context.Context, string, string, network.VirtualNetwork) error
	Delete(context.Context, string, string) error
	CheckIPAddressAvailability(context.Context, string, string, string) (network.IPAddressAvailabilityResult, error)
}
```

shysank: Refactored to follow virtual networks pattern because it seems more appropriate.

CecileRobertMichon: implemented your suggestion in https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1921/commits/f29b0ff84cd7e2db739ab3a844f78976425f96c4

shysank: ah we swapped ideas 🙂  I'll rebase it once that pr is merged.

azure/services/privatedns/spec.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+
+	return set, nil
+}
+
+// isManaged returns true if tagsMap contains an owner tag for the cluster.
+func isManaged(tagsMap map[string]*string, clusterName string) bool {
+	tags := converters.MapToTags(tagsMap)
+	return tags.HasOwned(clusterName)
+}
```
sonasingh46, line 188: nit: 
This could move to a different package probably -- which can help reuse.

azure/services/privatedns/client.go
```diff
@@ -19,166 +19,13 @@ package privatedns
...
-// DeleteRecordSet deletes a record set within the specified Private DNS zone.
-func (ac *azureClient) DeleteRecordSet(ctx context.Context, resourceGroupName string, privateZoneName string, recordType privatedns.RecordType, name string) error {
-	ctx, _, done := tele.StartSpanWithLogger(ctx, "privatedns.AzureClient.DeleteRecordSet")
-	defer done()
-
-	_, err := ac.recordsets.Delete(ctx, resourceGroupName, privateZoneName, recordType, name, "")
-	return err
-}
+var _ Getter = (*azureZonesClient)(nil)
+var _ Getter = (*azureVirtualNetworkLinksClient)(nil)
```
sonasingh46, line 31: Don't we need to enforce type checks for `azureRecordsClient` also like we did for `azureZonesClient ` and `azureVirtualNetworkLinksClient ` ?

sonasingh46:
@shysank -- 
>This pr also fixes a bug about managing zones and links separately.

Is the link in `bug` in PR description correct? It opens up the same PR. Trying to see and learn about the bug that got fixed.

azure/services/privatedns/client_records.go
```diff
@@ -0,0 +1,94 @@
...
+func newRecordSetsClient(auth azure.Authorizer) *azureRecordsClient {
+	recordsClient := privatedns.NewRecordSetsClientWithBaseURI(auth.BaseURI(), auth.SubscriptionID())
+	azure.SetAutoRestClientDefaults(&recordsClient.Client, auth.Authorizer())
+	return &azureRecordsClient{
+		recordsets: recordsClient,
+	}
+}
+
+// CreateOrUpdateAsync creates or updates a record asynchronously.
+// Well, I lied, CreateOrUpdateAsync creates a record synchronously. This is because azure api for records does not futures yet.
```
Jont828, line 44:
```suggestion
// Well, I lied, CreateOrUpdateAsync creates a record synchronously. This is because azure api for records does not have futures yet.
```
Also side note I think we keep the same comment for `CreateOrUpdateAsync()` for all the services even if they're not actually returning a future.

azure/services/privatedns/client_records.go
```diff
@@ -0,0 +1,94 @@
...
+	return recordSet, nil, err
+}
+
+// Get gets the specified record set. Noop for records.
+func (arc *azureRecordsClient) Get(ctx context.Context, spec azure.ResourceSpecGetter) (result interface{}, err error) {
+	return nil, nil
+}
+
+// DeleteAsync deletes a record asynchronously. Noop for records.
+// TODO: implement Delete for records.
```
Jont828, line 81: Is this a no-op or does it still need to be implemented?

azure/services/privatedns/spec.go
Jont828: Any thoughts @CecileRobertMichon?

azure/services/privatedns/spec.go
```diff
@@ -0,0 +1,188 @@
...
+		zone, ok := existing.(privatedns.PrivateZone)
+		if !ok {
+			return nil, errors.Errorf("%T is not a privatedns.PrivateZone", existing)
+		}
+		if !isManaged(zone.Tags, z.ClusterName) {
+			log.V(1).Info("Skipping reconciliation of unmanaged private DNS zone", "private DNS", zone.Name)
+			log.V(1).Info("Tag the DNS manually from azure to manage it with capz."+
+				"Please see https://capz.sigs.k8s.io/topics/custom-dns.html#manage-dns-via-capz-tool", "private DNS", zone.Name)
+			return nil, nil
+		}
```
Jont828, line 87: Should we return `nil, nil` at the end of this block anyway if the resource exists? If that's the case, then the check for `isManaged` would be redundant.

azure/services/privatedns/spec.go
```diff
@@ -0,0 +1,188 @@
...
+	if existing != nil {
+		link, ok := existing.(privatedns.VirtualNetworkLink)
+		if !ok {
+			return nil, errors.Errorf("%T is not a privatedns.VirtualNetworkLink", existing)
+		}
+		if !isManaged(link.Tags, l.ClusterName) {
+			log.V(2).Info("Skipping vnet link reconciliation for unmanaged vnet link", "vnet link",
+				l.Name, "private dns zone", l.ZoneName)
+			return nil, nil
+		}
```
Jont828, line 129: Same goes for here, I'm not sure if we still need these checks if we return `nil, nil` to skip updating an existing spec anyway.

azure/services/privatedns/spec.go
```diff
@@ -0,0 +1,188 @@
...
+func (r RecordSpec) OwnerResourceName() string {
+	return r.ZoneName
+}
+
+// ResourceGroupName returns the name of the resource group of the virtual network link.
+func (r RecordSpec) ResourceGroupName() string {
+	return r.ResourceGroup
+}
+
+// Parameters returns the parameters for the virtual network link.
```
Jont828, line 163:
```suggestion
// Parameters returns the parameters for a record set.
```

azure/services/privatedns/spec.go
```diff
@@ -0,0 +1,188 @@
...
+func (r RecordSpec) ResourceName() string {
+	return r.Record.Hostname
+}
+
+// OwnerResourceName returns the zone name of the virtual network link.
+func (r RecordSpec) OwnerResourceName() string {
+	return r.ZoneName
+}
+
+// ResourceGroupName returns the name of the resource group of the virtual network link.
```
Jont828, line 158:
```suggestion
// ResourceGroupName returns the name of the resource group of a record set.
```

azure/services/privatedns/spec.go
```diff
@@ -0,0 +1,188 @@
...
+		Location: to.StringPtr(azure.Global),
+		Tags: converters.TagsToMap(infrav1.Build(infrav1.BuildParams{
+			ClusterName: l.ClusterName,
+			Lifecycle:   infrav1.ResourceLifecycleOwned,
+			Additional:  l.AdditionalTags,
+		})),
+	}, nil
+}
+
+// ResourceName returns the name of the virtual network link.
```
Jont828, line 148:
```suggestion
// ResourceName returns the name of a record set
```

azure/services/privatedns/spec.go
```diff
@@ -0,0 +1,188 @@
...
+func (r RecordSpec) ResourceName() string {
+	return r.Record.Hostname
+}
+
+// OwnerResourceName returns the zone name of the virtual network link.
```
Jont828, line 153:
```suggestion
// OwnerResourceName returns the zone name of a record set.
```

azure/services/privatedns/spec.go
```diff
@@ -0,0 +1,188 @@
...
+			log.V(2).Info("Skipping vnet link reconciliation for unmanaged vnet link", "vnet link",
+				l.Name, "private dns zone", l.ZoneName)
+			return nil, nil
+		}
+	}
+
+	return privatedns.VirtualNetworkLink{
+		VirtualNetworkLinkProperties: &privatedns.VirtualNetworkLinkProperties{
+			VirtualNetwork: &privatedns.SubResource{
+				ID: to.StringPtr(azure.VNetID(l.SubscriptionID, l.VNetResourceGroup, l.VNetName)),
```
Jont828, line 135: If you don't need the subscription ID and the VNet name/RG anywhere else, you could just compute the ID in the spec getter instead and remove those fields.

azure/services/privatedns/spec.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+// ResourceGroupName returns the name of the resource group of the virtual network link.
+func (r RecordSpec) ResourceGroupName() string {
+	return r.ResourceGroup
+}
+
+// Parameters returns the parameters for the virtual network link.
+func (r RecordSpec) Parameters(existing interface{}) (params interface{}, err error) {
+	set := privatedns.RecordSet{
```
Jont828, line 165: Do we want to check if `existing != nil` and make sure it's the right type, and return `nil, nil` to skip the create/update?

azure/services/privatedns/privatedns.go
```diff
@@ -19,205 +19,253 @@ package privatedns
...
+
+func (s *Service) reconcileLinks(ctx context.Context, links []azure.ResourceSpecGetter) error {
+	ctx, _, done := tele.StartSpanWithLogger(ctx, "privatedns.Service.reconcileLinks")
+	defer done()
+
+	var resErr error
+
+	// We go through the list of links to reconcile each one, independently of the result of the previous one.
+	// If multiple errors occur, we return the most pressing one.
+	// Order of precedence (highest -> lowest) is: error that is not an operationNotDoneError (ie. error creating) -> operationNotDoneError (ie. creating in progress) -> no error (ie. created)
```
Jont828, line 118:
```suggestion
	// Order of precedence (highest -> lowest) is: error that is not an operationNotDoneError (i.e. error creating) -> operationNotDoneError (i.e. creating in progress) -> no error (i.e. created)
```

azure/services/privatedns/privatedns.go
```diff
@@ -19,205 +19,253 @@ package privatedns
...
+
+func (s *Service) reconcileRecords(ctx context.Context, records []azure.ResourceSpecGetter) error {
+	ctx, _, done := tele.StartSpanWithLogger(ctx, "privatedns.Service.reconcileRecords")
+	defer done()
+
+	var resErr error
+
+	// We go through the list of links to reconcile each one, independently of the result of the previous one.
+	// If multiple errors occur, we return the most pressing one.
+	// Order of precedence (highest -> lowest) is: error that is not an operationNotDoneError (ie. error creating) -> operationNotDoneError (ie. creating in progress) -> no error (ie. created)
```
Jont828, line 138:
```suggestion
	// Order of precedence (highest -> lowest) is: error that is not an operationNotDoneError (i.e. error creating) -> operationNotDoneError (i.e. creating in progress) -> no error (i.e. created)
```

azure/services/privatedns/privatedns.go
```diff
@@ -19,205 +19,253 @@ package privatedns
...
+
+func (s *Service) deleteLinks(ctx context.Context, links []azure.ResourceSpecGetter) error {
+	ctx, log, done := tele.StartSpanWithLogger(ctx, "privatedns.Service.deleteLinks")
+	defer done()
+
+	var resErr error
+
+	// We go through the list of links to delete each one, independently of the result of the previous one.
+	// If multiple errors occur, we return the most pressing one.
+	// Order of precedence (highest -> lowest) is: error that is not an operationNotDoneError (ie. error creating) -> operationNotDoneError (ie. creating in progress) -> no error (ie. created)
```
Jont828, line 186:
```suggestion
	// Order of precedence (highest -> lowest) is: error that is not an operationNotDoneError (i.e. error creating) -> operationNotDoneError (i.e. creating in progress) -> no error (i.e. created)
```

azure/services/privatedns/privatedns.go
```diff
@@ -19,205 +19,253 @@ package privatedns
...
+		return errors.Wrapf(err, "could not get private DNS zone state of %s in resource group %s", zoneSpec.ResourceName(), zoneSpec.ResourceGroupName())
 	}
-	return nil
+
+	if !isManaged {
+		log.V(1).Info("Skipping deletion of unmanaged private DNS zone", "private DNS", zoneSpec.ResourceName())
+		return nil
+	}
+
+	err = s.zoneReconciler.DeleteResource(ctx, zoneSpec, serviceName)
```
Jont828, line 236: Maybe add this comment `// Delete the private DNS zone, which also deletes all records.` from the old reconciler here.

Jont828 (commented):
This is looking good! I like that the Reconciler is split into helper functions, and in general it looks way easier to read!

azure/services/privatedns/client_records.go
shysank: Actually I can do something like [this](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/a5505c9aec4bfa0bab54c8f078f5dbde7d404a1c/azure/services/securitygroups/client.go#L76) to make this async, let me try that.

azure/services/privatedns/client.go
shysank: Now that we are using `async.Getter`, we won't need this file anymore.

shysank:
> Is the link in `bug` in PR description correct? It opens up the same PR. Trying to see and learn about the bug that got fixed.

Fixed it, Thanks!

azure/services/privatedns/client_records.go
shysank: You're right, it's jut a noop, have updated the comment.

azure/services/privatedns/spec.go
shysank: I thought about this, but didn't want to change the current implementation. wdyt about doing it in a followup pr?

azure/services/privatedns/spec.go
shysank: same comment as above, didn't want to change current implementation. wdyt about doing it in a followup pr?

azure/services/privatedns/spec.go
shysank: This is a [pattern](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/7979a9705adabde2085801b04f87ad00e8888e1e/azure/services/loadbalancers/spec.go#L170) that's being used in other services, so didn't want to deviate from it.

azure/services/privatedns/spec.go
shysank: yeah, makes sense.

azure/services/privatedns/privatedns.go
shysank: Good catch, added it.

azure/services/privatedns/spec.go
sonasingh46: Sure. May be we can put a comment there by putting a TODO. This can help track in future if we want to reuse and refactor.

shysank:
@Jont828 @sonasingh46 I've addressed the review comments, ptal whenever you get a chance.

azure/services/privatedns/spec.go
Jont828: I think it should be fine in either case since if we don't update, we'd just be returning the existing resource anyway. In general, I've set up Parameters() to return nil, nil if we find an existing resource and also added any checks to the fields, i.e. for a subnet, if the VNet exists when it's not managed.

shysank: yeah, we [discussed](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1791#discussion_r749571484) about this earlier and created a followup for it. I'm fine with doing it here; I left it for now because we want to focus on other things. Happy to do it either way.

azure/services/privatedns/spec.go
```diff
@@ -0,0 +1,194 @@
...
+	_, log, done := tele.StartSpanWithLogger(context.TODO(), "privatedns.ZoneSpec.Parameters")
+	defer done()
+
+	if existing != nil {
+		zone, ok := existing.(privatedns.PrivateZone)
+		if !ok {
+			return nil, errors.Errorf("%T is not a privatedns.PrivateZone", existing)
+		}
+		if !isManaged(zone.Tags, z.ClusterName) {
+			log.V(1).Info("Skipping reconciliation of unmanaged private DNS zone", "private DNS", zone.Name)
```
CecileRobertMichon, line 83: We could add a TODO to remove this log in a future release of CAPZ. This is only a temporary warning due to #1791 adding tags previously created clusters not having the tags so this manual step is needed once for backwards compatibility.

azure/services/privatedns/spec.go
CecileRobertMichon: given my comment below about the logging being temporary, I'd vote to keep this as a context.TODO() for now and just remove the context when we remove the logging in an upcoming release.

azure/services/privatedns/spec.go
```diff
@@ -0,0 +1,194 @@
...
+}
+
+type RecordSpec struct {
+	Record        infrav1.AddressRecord
+	ZoneName      string
+	ResourceGroup string
+}
+
+// ResourceName returns the name of the private dns zone.
+func (z ZoneSpec) ResourceName() string {
```
CecileRobertMichon, line 58: nit: in other services we've used s as receiver name for "spec", if we care about consistency across services we should stick with s here and in the other private DNS specs as well.

azure/services/privatedns/client_records.go
```diff
@@ -0,0 +1,92 @@
...
+
+	// Determine record type.
+	var (
+		recordType privatedns.RecordType
+		aRecords   = set.RecordSetProperties.ARecords
+		aaaRecords = set.RecordSetProperties.AaaaRecords
+	)
+	if aRecords != nil && len(*aRecords) > 0 && (*aRecords)[0].Ipv4Address != nil {
+		recordType = privatedns.A
+	} else if aaaRecords != nil && len(*aaaRecords) > 0 && (*aaaRecords)[0].Ipv6Address != nil {
```
CecileRobertMichon, line 62: would it make sense to add an else condition that returns an error if we're unable to deduct the record type?

azure/services/privatedns/spec.go
shysank: Fixed

shysank:
@Jont828 @CecileRobertMichon I have addressed all the comments, ptal whenever you get a chance.

CecileRobertMichon:
@shysank could you please rebase on top of #2093

azure/services/privatedns/spec.go
Jont828: Okay sounds good.

azure/services/privatedns/spec.go
```diff
@@ -0,0 +1,197 @@
...
+	"github.com/Azure/go-autorest/autorest/to"
+	"github.com/pkg/errors"
+	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
+	"sigs.k8s.io/cluster-api-provider-azure/azure"
+	"sigs.k8s.io/cluster-api-provider-azure/azure/converters"
+	"sigs.k8s.io/cluster-api-provider-azure/util/tele"
+)
+
+// ZoneSpec defines the specification for private dns zone.
+type ZoneSpec struct {
```
Jont828, line 32: What do you think about breaking up the spec file into zone_spec.go, link_spec.go, and record_spec.go since the client files are split among the 3 types as well?

azure/services/privatedns/spec.go
Jont828: Sounds good!

azure/services/privatedns/privatedns.go
```diff
@@ -20,202 +20,250 @@ import (
...
+		s.Scope.UpdatePutStatus(infrav1.PrivateDNSReadyCondition, serviceName, err)
+		return err
+	}
+
+	if err := s.reconcileLinks(ctx, links); err != nil {
+		s.Scope.UpdatePutStatus(infrav1.PrivateDNSReadyCondition, serviceName, err)
+		return err
+	}
+
+	if err := s.reconcileRecords(ctx, records); err != nil {
```
Jont828, line 90: The same applies for here as well.

azure/services/privatedns/privatedns.go
```diff
@@ -20,202 +20,250 @@ import (
...
+	ctx, cancel := context.WithTimeout(ctx, reconciler.DefaultAzureServiceReconcileTimeout)
+	defer cancel()
+
+	zoneSpec, links, _ := s.Scope.PrivateDNSSpec()
+	if zoneSpec == nil {
+		return nil
+	}
+
+	if err := s.deleteLinks(ctx, links); err != nil {
+		s.Scope.UpdateDeleteStatus(infrav1.PrivateDNSReadyCondition, serviceName, err)
```
Jont828, line 161: I believe the same case applies to the deletion where we don't want to update the delete status if we don't have any managed link specs to delete.

azure/services/privatedns/privatedns.go
```diff
@@ -20,202 +20,250 @@ import (
...
-				return errors.Wrapf(err, "failed to create virtual network link %s", linkSpec.LinkName)
+	ctx, cancel := context.WithTimeout(ctx, reconciler.DefaultAzureServiceReconcileTimeout)
+	defer cancel()
+
+	zoneSpec, links, records := s.Scope.PrivateDNSSpec()
+	if zoneSpec == nil {
+		return nil
+	}
+
+	if err := s.reconcileZone(ctx, zoneSpec); err != nil {
```
Jont828, line 80: In [this PR](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2093) I discussed w/ @CecileRobertMichon that we don't want to update the put status for an unmanaged resource. In this case I think we want to add a check earlier to return or skip this block if the zone isn't managed.

azure/services/privatedns/privatedns.go
```diff
@@ -20,202 +20,250 @@ import (
...
+		return nil
+	}
+
+	if err := s.deleteLinks(ctx, links); err != nil {
+		s.Scope.UpdateDeleteStatus(infrav1.PrivateDNSReadyCondition, serviceName, err)
+		return err
+	}
+
+	if err := s.deleteZone(ctx, zoneSpec); err != nil {
+		s.Scope.UpdateDeleteStatus(infrav1.PrivateDNSReadyCondition, serviceName, err)
```
Jont828, line 166: Same for skipping delete on an unmanaged zone. So I guess tl;dr is that if we go through the reconciler or deletion and see that we don't have any managed specs at all, then we don't update the status. @CecileRobertMichon Does this approach make sense?

azure/services/privatedns/privatedns.go
```diff
@@ -20,202 +20,250 @@ import (
...
+	}
+
+	return resErr
+}
+
+func (s *Service) deleteZone(ctx context.Context, zoneSpec azure.ResourceSpecGetter) error {
+	ctx, log, done := tele.StartSpanWithLogger(ctx, "privatedns.Service.deleteZone")
+	defer done()
+
+	// Skip the reconciliation of private DNS zone which is not managed by capz.
```
Jont828, line 217:
```suggestion
	// Skip deleting the private DNS zone when it's not managed by capz.
```

azure/services/privatedns/privatedns.go
```diff
@@ -20,202 +20,250 @@ import (
...
+	result, err := s.zoneGetter.Get(ctx, spec)
 	if err != nil {
 		return false, err
 	}
+	zone, ok := result.(privatedns.PrivateZone)
+	if !ok {
+		return false, errors.Errorf("%T is not a privatedns.PrivateZone", zone)
+	}
+
 	tags := converters.MapToTags(zone.Tags)
```
Jont828, line 250: You could probably refactor these last two lines to use your `isManaged` helper function since it just checks a map for `tags.HasOwned(clusterName)`.

azure/scope/cluster.go
```diff
@@ -691,6 +710,7 @@ func (s *ClusterScope) PatchObject(ctx context.Context) error {
 			infrav1.BastionHostReadyCondition,
 			infrav1.VNetReadyCondition,
 			infrav1.SubnetsReadyCondition,
+			infrav1.PrivateDNSReadyCondition,
```
Jont828, line 713: In [this PR](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2093/files) we removed the SetSummary part. When you rebase, you can remove this line.

shysank:
To accomodate #2093, I've made the following changes:

1. Split PrivateDnsReadyCondition into `PrivateDnsZoneReadyCondition`, `PrivateDnsLinkReadyCondition`, `PrivateDnsRecordReadyCondition`. This is to make sure we only set the appropriate ready condition if the resource is managed by capz.
2. For resources with multiple items (eg. `vnetLinks`), I have considered it to be `Managed` even if one of them is managed. 
3. This means that we'll need to a [redundant get](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/d1d767654bd79dad873123dedbe3679dcd42dec9/azure/services/privatedns/zone_reconciler.go#L31) on the resource to check if it is managed or not. Since this is a common concern, we could perhaps move this into the async framework. Something like:`{Create,Delete}Resource` could return if the resource is managed or not? Or allow to register listeners during various stages of {Create,Delete}Resource?
4. Split spec,client, and reconciler logic for zone,link, and records into their own separate files, have created a separate commit for each one of them hoping it would make it easier to review. Also, would make it easier to refactor into separate services if needed.

 cc @CecileRobertMichon @Jont828

Jont828 (commented):
I normally wouldn't want to split up a ready condition into 3, but in this case I think it makes sense with how the code is split into 3 separate reconcilers, clients, and specs. Also in #2146 we changed the Reconcilers by adding a `Name()` and `Managed()` function. I think it makes sense to consider the resource managed if one of them is managed, which is what we're doing in 2146.

azure/services/privatedns/privatedns.go
```diff
@@ -20,202 +20,143 @@ import (
...
+	}
+
+	return err
 }
 
 // isPrivateDNSManaged returns true if the private DNS has an owned tag with the cluster name as value,
 // meaning that the DNS lifecycle is managed.
-func (s *Service) isPrivateDNSManaged(ctx context.Context, resourceGroup, zoneName string) (bool, error) {
-	zone, err := s.client.GetZone(ctx, resourceGroup, zoneName)
+func (s *Service) isPrivateDNSManaged(ctx context.Context, spec azure.ResourceSpecGetter) (bool, error) {
```
Jont828, line 133: I see what you're saying about the redundant Get() calls too, it looks like we have this pattern on several services like VNets, where we need to fetch the existing resource to see if it's managed or not. I'm open to amending the interface, did you have any ideas? Since the tags are all the same, I was thinking we could have some an interface function where it could cache the result for the next Get() call for Create(). Wdyt @shysank @CecileRobertMichon?

shysank: I'm inclined towards merging this one as is and updating the interface in a followup. Since it affects only private clusters, this shouldn't cause huge performance degradation imo.

CecileRobertMichon: yes +1 to caching as a separate issue, the scope is big enough here. #2146 (the second commit for IsManaged) should also help with that. I had tried to incorporate caching as part of https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1684/files#diff-b74c6acb4629abe58f8286da9939563cc4404e09fe9966eb61dbde9b87274189R110 but ended up reverting it to keep the changes small.

shysank:
@CecileRobertMichon @Jont828 I think I have addressed all review comments. PTAL, whenever you get a chance.

azure/services/privatedns/privatedns.go
```diff
@@ -57,173 +71,98 @@ func (s *Service) Name() string {
...
-	return nil
+
+	return err
 }
 
 // isPrivateDNSManaged returns true if the private DNS has an owned tag with the cluster name as value,
 // meaning that the DNS lifecycle is managed.
-func (s *Service) isPrivateDNSManaged(ctx context.Context, resourceGroup, zoneName string) (bool, error) {
-	zone, err := s.client.GetZone(ctx, resourceGroup, zoneName)
+func (s *Service) isPrivateDNSManaged(ctx context.Context, spec azure.ResourceSpecGetter) (bool, error) {
```
CecileRobertMichon, line 138: Can we reuse the isManaged func below instead for this one?

azure/services/privatedns/link_reconciler.go
```diff
@@ -0,0 +1,104 @@
...
+
+	// We go through the list of links to reconcile each one, independently of the result of the previous one.
+	// If multiple errors occur, we return the most pressing one.
+	// Order of precedence (highest -> lowest) is: error that is not an operationNotDoneError (i.e. error creating) -> operationNotDoneError (i.e. creating in progress) -> no error (i.e. created)
+	for _, linkSpec := range links {
+		isLinkManaged, err := s.isVnetLinkManaged(ctx, linkSpec)
+		if err != nil {
+			if azure.ResourceNotFound(err) {
+				isLinkManaged = true
+				// we consider VnetLinks as managed even if one of the links is managed.
```
CecileRobertMichon, line 41:
```suggestion
				// we consider VnetLinks as managed if at least one of the links is managed.
```

azure/services/privatedns/zone_reconciler.go
```diff
@@ -0,0 +1,78 @@
...
+	if err != nil {
+		if azure.ResourceNotFound(err) {
+			managed = true
+		} else {
+			return managed, err
+		}
+	}
+
+	if !managed {
+		log.V(1).Info("Skipping reconciliation of unmanaged private DNS zone", "private DNS", zoneSpec.ResourceName())
```
Jont828, line 41: Should this be `log.V(2)` instead?

azure/services/privatedns/privatedns.go
```diff
@@ -57,178 +71,103 @@ func (s *Service) Name() string {
...
+	if managed {
+		s.Scope.UpdatePutStatus(infrav1.PrivateDNSLinkReadyCondition, serviceName, err)
+	}
+	if err != nil {
+		return err
 	}
-	return nil
+
+	err = s.reconcileRecords(ctx, records)
+	s.Scope.UpdatePutStatus(infrav1.PrivateDNSRecordReadyCondition, serviceName, err)
```
Jont828, line 102: I think we only want to update the put status if we actually have specs. In this case, we could only call `UpdatePutStatus` if the number of specs > 0. The same goes for the delete status too.

Jont828: Okay sounds good. In that case I think it LGTM.

CecileRobertMichon (commented):
@shysank can you please squash?

azure/services/privatedns/privatedns.go
```diff
@@ -57,178 +71,103 @@ func (s *Service) Name() string {
...
+	}
+
+	tags := converters.MapToTags(link.Tags)
 	return tags.HasOwned(s.Scope.ClusterName()), nil
 }
 
-// IsManaged returns always returns true.
+// IsManaged returns true if the private DNS has an owned tag with the cluster name as value,
+// meaning that the DNS lifecycle is managed.
 // TODO: separate private DNS and VNet links so we can implement the IsManaged method for each.
```
CecileRobertMichon, line 155: can we remove the TODO?

shysank: done and sqashed

CecileRobertMichon (commented):
/lgtm
/approve

PR #2008, opened 2022-01-26
Replaced `go get` with `go install` to install binaries

### LINKED ISSUE #2004: Use go install instead of go get to install binaries
https://go.dev/doc/go-get-install-deprecation

Starting in Go 1.17, installing executables with go get is deprecated. go install may be used instead.

In Go 1.18, go get will no longer build packages; it will only be used to add, update, or remove dependencies in go.mod. 

we should stop using go get to install hack/tools and instead rely on go install here: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/scripts/go_install.sh#L54

/good-first-issue

jayesh-srivastava:
**What type of PR is this?**
/kind cleanup

The PR replaces `go get` with `go install` to install binaries.

Fixes #2004 

**Release note**:
```
```

### DIFF
--- a/scripts/go_install.sh
+++ b/scripts/go_install.sh
@@ -51,6 +51,6 @@ cd "${tmp_dir}"
 go mod init fake/mod
 
 # install the golang module specified as the first argument
-go get -tags tools "${1}@${3}"
+go install -tags tools "${1}@${3}"
 mv "${GOBIN}/${2}" "${GOBIN}/${2}-${3}"
 ln -sf "${GOBIN}/${2}-${3}" "${GOBIN}/${2}"


CecileRobertMichon:
It looks like there is a known issue with kustomize that prevents us from using go install with it: https://github.com/kubernetes-sigs/kustomize/issues/3618

we can either wait until it's fixed to merge this, or change the kustomize install to keep using go get for now.

/hold

sonasingh46:
@CecileRobertMichon -- I was thinking we can probably use `go install` to install binary and for `kustomize` we can still use  `go get`.

sonasingh46:
We can use hard coded `go get` here for kustomize 
https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/Makefile#L255

sonasingh46:
>we can either wait until it's fixed to merge this, or change the kustomize install to keep using go get for now.
I think you suggested the same.

CecileRobertMichon:
#2088 should unblock this

jayesh-srivastava:
@CecileRobertMichon I guess I can't remove the hold. Now that #2091 is merged, I think this should work now.

CecileRobertMichon:
/lgtm
/approve

Thanks @jayesh-srivastava!

PR #1995, opened 2022-01-21
e2e: Use the intersection of images available for windows and ubuntu to determine latest patch

CecileRobertMichon:
**What type of PR is this?**

**What this PR does / why we need it**: While new patch releases had been published in the capi offer for Ubuntu and the Windows offer publishing was still ongoing, some tests started failing with `The platform image 'cncf-upstream:capi-windows:k8s-1dot22dot6-windows-2019-containerd:latest is not available`. This is because the tests were selecting the latest available patch release for Ubuntu, but there wasn't an equivalent windows image yet.

This PR tweaks the logic in the e2e test helper to make sure we select the latest version available for _both_ windows and linux.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

PR #2011, opened 2022-01-26
Fix the custom vnet test template to provide the route table and nsg names

CecileRobertMichon:
**What type of PR is this?**
/kind failing-test

**What this PR does / why we need it**: This is fixing the flake observed in https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/1994/pull-cluster-api-provider-azure-e2e/1486169658842157056. We were creating a custom vnet with subnets and route table + nsg out of band, but not setting the route table and NSG name in the template. This caused CAPZ to skip reconciliation for route tables and NSGs, but then to pass default RT and NSG names to the cloud provider, and cloud provider would create the route table with default name to use it instead of the one we created. The flake was caused by the fact that the extra route table didn't get properly deleted which caused the cleanup step to fail.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/azure_test.go
+++ b/test/e2e/azure_test.go
@@ -1,3 +1,4 @@
+//go:build e2e
 // +build e2e
 
 /*
@@ -141,7 +142,7 @@ var _ = Describe("Workload cluster creation", func() {
 
 	if os.Getenv("LOCAL_ONLY") != "true" {
 		Context("Creating a private cluster", func() {
-			It("Creates a public management cluster in the same vnet", func() {
+			It("Creates a public management cluster a custom vnet", func() {
 				clusterName = getClusterName(clusterNamePrefix, "public-custom-vnet")
 				Context("Creating a custom virtual network", func() {
 					Expect(os.Setenv(AzureCustomVNetName, "custom-vnet")).NotTo(HaveOccurred())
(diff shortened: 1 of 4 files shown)


CecileRobertMichon:
/retest

same flake as #1995 , there might be something going on with AzureBastions

test/e2e/azure_test.go
```diff
@@ -141,7 +142,7 @@ var _ = Describe("Workload cluster creation", func() {
 
 	if os.Getenv("LOCAL_ONLY") != "true" {
 		Context("Creating a private cluster", func() {
-			It("Creates a public management cluster in the same vnet", func() {
+			It("Creates a public management cluster a custom vnet", func() {
```
jackfrancis, line 145: nit: "in a custom vnet"

If this test run passes lgtm and merge :)

jackfrancis (approved):
lgtm

test/e2e/azure_test.go
CecileRobertMichon: fixed!

jackfrancis (approved):
/lgtm

PR #2014, opened 2022-01-27
Update CAPI to v1.1.0

CecileRobertMichon:
**What type of PR is this?**

1) update CAPI to v1.1.0
2) fix KCP upgrade to use 0 worker nodes since https://github.com/kubernetes-sigs/cluster-api/commit/d3ac6d0e34673dc68ddb06e09bc89ad8d08bc068 is in the release
3) bump default upgrade versions from 1.22 to 1.23

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```


CecileRobertMichon:
/hold 

until #1995 merges to fix the stable- version behavior

test/e2e/capi_test.go
```diff
@@ -275,7 +276,7 @@ var _ = Describe("Running the Cluster API E2E tests", func() {
 				BootstrapClusterProxy:    bootstrapClusterProxy,
 				ArtifactFolder:           artifactFolder,
 				ControlPlaneMachineCount: pointer.Int64(3),
-				WorkerMachineCount:       pointer.Int64(1),
+				WorkerMachineCount:       pointer.Int64(0),
```
jsturtevant, line 279: I should probably remember this but why is this going to zero?

CecileRobertMichon: CAPI 1.1.0 deprecated the KCP only upgrade tests in favor of using the cluster upgrade tests. However when we bumped CAPI to 1.1.0-beta.1 there was a [bug](https://github.com/kubernetes-sigs/cluster-api/issues/4896#issuecomment-1010523122) which prevented us from using worker machine count 0 so I set to 1 temporarily. Now that the bug is [fixed](https://github.com/kubernetes-sigs/cluster-api/pull/5926), setting it back to 0 to optimize test duration without reducing coverage (we have a separate full cluster upgrade spec above that tests control plane + worker node upgrades, while these two specs are specific to control plane upgrade only with different rolling strategies).

CecileRobertMichon:
> ssh: rejected: connect failed (Temporary failure in name resolution),

/retest
/hold cancel

shysank:
@CecileRobertMichon should we hold for the new rc release with cert manager fix?

CecileRobertMichon:
@shysank there won't be another RC but I can bump this to v1.1.0 for when it gets released

the retest probably wouldn't work right now anyways so I won't be able to merge this as-is

CecileRobertMichon:
@shysank updated the PR, PTAL

we can trigger a retest once the release is actually out

go.mod
```diff
@@ -42,8 +42,8 @@ require (
 	k8s.io/klog/v2 v2.30.0
 	k8s.io/kubectl v0.23.0
 	k8s.io/utils v0.0.0-20210930125809-cb0fa318a74b
-	sigs.k8s.io/cluster-api v1.1.0-beta.1
-	sigs.k8s.io/cluster-api/test v1.1.0-beta.1
+	sigs.k8s.io/cluster-api v1.1.0-rc.0
```
jsturtevant, line 45: the rest of the references are `1.1.0` and this one is `1.1.0-rc.0`. I think these should be updated?

test/e2e/capi_test.go
jsturtevant: Since we are waiting for the release should we make comment in here that points to this explanation or gives one? I know I will forget this 6 months from now.

go.mod
CecileRobertMichon: yes they should. I need the actual release to be out so I can run go modules though

CecileRobertMichon: updated

test/e2e/capi_test.go
CecileRobertMichon: we're not waiting for the release anymore, this was a temporary workaround when we upgraded to beta.1, now that it's fixed we're setting the count to 0 as desired

CecileRobertMichon:
Release just dropped, this should be good to go pending tests

PR #2020, opened 2022-01-28
Support passing custom headers to AKS Managed Cluster and Node Pool create/update requests

### LINKED ISSUE #1979: Add support for GPU nodes in AKS (managed cluster)
/kind feature

Support for GPU nodepools in AKS
https://docs.microsoft.com/en-us/azure/aks/gpu-cluster 

one way to solve this is to allow custom headers for the requests to the agentpool creation. 

**Environment:**

### LINKED ISSUE #1981: add support for custom-headers for AKS
/kind feature

add support for custom-headers for AKS. 
Custom headers are used by Azure API to enable several features for AKS controlplane and AKS nodepools, few examples would be:

For control-plane:
```
- Enable Container Storage Interface (CSI) drivers for Azure disks and Azure Files 
```

For nodepools:
```
- Adding a node pool for GPU nodes
```

**Environment:**

michalno1:
**What type of PR is this?**
/kind feature

To use some AKS features one needs to pass custom headers when creating/updating a resource. One example is when you want to use a [GPU node pool](https://docs.microsoft.com/en-us/azure/aks/gpu-cluster), you need to register `GPUDedicatedVHDPreview` feature and pass a custom header to node pool creation request:
```
az aks nodepool add \
   ...
   --aks-custom-headers UseGPUDedicatedVHD=true \
   ...
```
This PR adds a mechanism where you can pass custom headers to cluster and node pool create/update requests by using annotations prefixed with `custom-headers.azure.dev/`. E.g. to create mentioned GPU node pool one would set `"custom-headers.azure.dev/UseGPUDedicatedVHD": true` annotation on `AzureManagedMachinePool`.

fixes #1981, fixes #1979

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Support passing custom headers to cluster and node pool create/update requests to enable additional features
```

### DIFF
--- a/azure/defaults.go
+++ b/azure/defaults.go
@@ -84,6 +84,14 @@ const (
 	ProviderIDPrefix = "azure://"
 )
 
+const (
+	// CustomHeaderPrefix is the prefix of annotations that enable additional cluster / node pool features.
+	// Whatever follows the prefix will be passed as a header to cluster/node pool creation/update requests.
+	// E.g. add `"infrastructure.cluster.x-k8s.io/custom-header-UseGPUDedicatedVHD": "true"` annotation to
+	// AzureManagedMachinePool CR to enable creating GPU nodes by the node pool.
+	CustomHeaderPrefix = "infrastructure.cluster.x-k8s.io/custom-header-"
+)
+
 var (
 	// LinuxBootstrapExtensionCommand is the command the VM bootstrap extension will execute to verify Linux nodes bootstrap completes successfully.
 	LinuxBootstrapExtensionCommand = fmt.Sprintf("for i in $(seq 1 %d); do test -f %s && break; if [ $i -eq %d ]; then return 1; else sleep %d; fi; done", bootstrapExtensionRetries, bootstrapSentinelFile, bootstrapExtensionRetries, bootstrapExtensionSleep)
--- a/azure/scope/managedcontrolplane.go
+++ b/azure/scope/managedcontrolplane.go
@@ -391,6 +391,10 @@ func (s *ManagedControlPlaneScope) FailureDomains() []string {
 	return []string{}
 }
 
+func (s *ManagedControlPlaneScope) ManagedClusterAnnotations() map[string]string {
+	return s.ControlPlane.Annotations
+}
+
 // ManagedClusterSpec returns the managed cluster spec.
 func (s *ManagedControlPlaneScope) ManagedClusterSpec() (azure.ManagedClusterSpec, error) {
 	decodedSSHPublicKey, err := base64.StdEncoding.DecodeString(s.ControlPlane.Spec.SSHPublicKey)
@@ -555,6 +559,10 @@ func (s *ManagedControlPlaneScope) GetAllAgentPoolSpecs(ctx context.Context) ([]
 	}
 
 	return ammps, nil
+}
+
+func (s *ManagedControlPlaneScope) AgentPoolAnnotations() map[string]string {
+	return s.InfraMachinePool.Annotations
 }
 
 // AgentPoolSpec returns an azure.AgentPoolSpec for currently reconciled AzureManagedMachinePool.
--- a/azure/services/agentpools/agentpools.go
+++ b/azure/services/agentpools/agentpools.go
@@ -27,6 +27,7 @@ import (
 	"github.com/pkg/errors"
 	infrav1alpha4 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
 	"sigs.k8s.io/cluster-api-provider-azure/azure"
+	"sigs.k8s.io/cluster-api-provider-azure/util/maps"
 	"sigs.k8s.io/cluster-api-provider-azure/util/tele"
 )
 
@@ -35,6 +36,7 @@ type ManagedMachinePoolScope interface {
 	azure.ClusterDescriber
 
 	NodeResourceGroup() string
+	AgentPoolAnnotations() map[string]string
 	AgentPoolSpec() azure.AgentPoolSpec
 	SetAgentPoolProviderIDList([]string)
 	SetAgentPoolReplicas(int32)
@@ -64,7 +66,6 @@ func (s *Service) Reconcile(ctx context.Context) error {
 	defer done()
 
 	agentPoolSpec := s.scope.AgentPoolSpec()
-
 	profile := containerservice.AgentPool{
 		ManagedClusterAgentPoolProfileProperties: &containerservice.ManagedClusterAgentPoolProfileProperties{
 			VMSize:              &agentPoolSpec.SKU,
@@ -96,8 +97,10 @@ func (s *Service) Reconcile(ctx context.Context) error {
 	// AKS will populate defaults and read-only values, which we want
 	// to strip/clean to match what we expect.
 
+	customHeaders := maps.FilterByKeyPrefix(s.scope.AgentPoolAnnotations(), azure.CustomHeaderPrefix)
 	if isCreate := azure.ResourceNotFound(err); isCreate {
-		err = s.Client.CreateOrUpdate(ctx, agentPoolSpec.ResourceGroup, agentPoolSpec.Cluster, agentPoolSpec.Name, profile)
+		err = s.Client.CreateOrUpdate(ctx, agentPoolSpec.ResourceGroup, agentPoolSpec.Cluster, agentPoolSpec.Name,
+			profile, customHeaders)
 		if err != nil && azure.ResourceNotFound(err) {
 			return azure.WithTransientError(errors.Wrap(err, "agent pool dependent resource does not exist yet"), 20*time.Second)
 		} else if err != nil {
@@ -138,7 +141,8 @@ func (s *Service) Reconcile(ctx context.Context) error {
 		diff := cmp.Diff(normalizedProfile, existingProfile)
 		if diff != "" {
 			log.V(2).Info(fmt.Sprintf("Update required (+new -old):\n%s", diff))
-			err = s.Client.CreateOrUpdate(ctx, agentPoolSpec.ResourceGroup, agentPoolSpec.Cluster, agentPoolSpec.Name, profile)
+			err = s.Client.CreateOrUpdate(ctx, agentPoolSpec.ResourceGroup, agentPoolSpec.Cluster, agentPoolSpec.Name,
+				profile, customHeaders)
 			if err != nil {
 				return errors.Wrap(err, "failed to create or update agent pool")
 			}
--- a/azure/services/managedclusters/client.go
+++ b/azure/services/managedclusters/client.go
@@ -30,7 +30,7 @@ import (
 type Client interface {
 	Get(context.Context, string, string) (containerservice.ManagedCluster, error)
 	GetCredentials(context.Context, string, string) ([]byte, error)
-	CreateOrUpdate(context.Context, string, string, containerservice.ManagedCluster) (containerservice.ManagedCluster, error)
+	CreateOrUpdate(context.Context, string, string, containerservice.ManagedCluster, map[string]string) (containerservice.ManagedCluster, error)
 	Delete(context.Context, string, string) error
 }
 
@@ -78,11 +78,19 @@ func (ac *AzureClient) GetCredentials(ctx context.Context, resourceGroupName, na
 }
 
 // CreateOrUpdate creates or updates a managed cluster.
-func (ac *AzureClient) CreateOrUpdate(ctx context.Context, resourceGroupName, name string, cluster containerservice.ManagedCluster) (containerservice.ManagedCluster, error) {
+func (ac *AzureClient) CreateOrUpdate(ctx context.Context, resourceGroupName, name string, cluster containerservice.ManagedCluster, headers map[string]string) (containerservice.ManagedCluster, error) {
 	ctx, _, done := tele.StartSpanWithLogger(ctx, "managedclusters.AzureClient.CreateOrUpdate")
 	defer done()
 
-	future, err := ac.managedclusters.CreateOrUpdate(ctx, resourceGroupName, name, cluster)
+	preparer, err := ac.managedclusters.CreateOrUpdatePreparer(ctx, resourceGroupName, name, cluster)
+	if err != nil {
+		return containerservice.ManagedCluster{}, errors.Wrap(err, "failed to prepare operation")
+	}
+	for key, value := range headers {
+		preparer.Header.Add(key, value)
+	}
+
+	future, err := ac.managedclusters.CreateOrUpdateSender(preparer)
 	if err != nil {
 		return containerservice.ManagedCluster{}, errors.Wrap(err, "failed to begin operation")
 	}
--- a/azure/services/managedclusters/managedclusters.go
+++ b/azure/services/managedclusters/managedclusters.go
@@ -30,6 +30,7 @@ import (
 	"k8s.io/klog/v2"
 	infrav1alpha4 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
 	"sigs.k8s.io/cluster-api-provider-azure/azure"
+	"sigs.k8s.io/cluster-api-provider-azure/util/maps"
 	"sigs.k8s.io/cluster-api-provider-azure/util/tele"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 )
@@ -42,6 +43,7 @@ var (
 // ManagedClusterScope defines the scope interface for a managed cluster.
 type ManagedClusterScope interface {
 	azure.ClusterDescriber
+	ManagedClusterAnnotations() map[string]string
 	ManagedClusterSpec() (azure.ManagedClusterSpec, error)
 	GetAllAgentPoolSpecs(ctx context.Context) ([]azure.AgentPoolSpec, error)
 	SetControlPlaneEndpoint(clusterv1.APIEndpoint)
@@ -292,8 +294,9 @@ func (s *Service) Reconcile(ctx context.Context) error {
 		}
 	}
 
+	customHeaders := maps.FilterByKeyPrefix(s.Scope.ManagedClusterAnnotations(), azure.CustomHeaderPrefix)
 	if isCreate {
-		managedCluster, err = s.Client.CreateOrUpdate(ctx, managedClusterSpec.ResourceGroupName, managedClusterSpec.Name, managedCluster)
+		managedCluster, err = s.Client.CreateOrUpdate(ctx, managedClusterSpec.ResourceGroupName, managedClusterSpec.Name, managedCluster, customHeaders)
 		if err != nil {
 			return fmt.Errorf("failed to create managed cluster, %w", err)
 		}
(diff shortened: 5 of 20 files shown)


azure/defaults.go
```diff
@@ -84,6 +84,14 @@ const (
 	ProviderIDPrefix = "azure://"
 )
 
+const (
+	// CustomHeaderPrefix is the prefix of annotations that enable additional cluster / node pool features.
+	// Whatever follows the prefix will be passed as a header to cluster/node pool creation/update requests.
+	// E.g. add `"custom-headers.azure.dev/UseGPUDedicatedVHD": "true"` annotation to AzureManagedMachinePool CR
+	// to enable creating GPU nodes by the node pool.
+	CustomHeaderPrefix = "custom-headers.azure.dev/"
```
sonasingh46, line 92: nit: 
Do annotation/labels follow a specific pattern?  In case of annotations I have seen `cluster.x-k8s.io` in CAPI. May be worth discussing as  his will help us being consistent while we may introduce more annotations in near future. WDYT?

azure/scope/managedcontrolplane.go
```diff
@@ -364,6 +364,10 @@ func (s *ManagedControlPlaneScope) FailureDomains() []string {
 	return []string{}
 }
 
+func (s *ManagedControlPlaneScope) ManagedClusterAnnotations() map[string]string {
+	return s.ControlPlane.Annotations
+}
+
```
sonasingh46, line 370: nit:
Do we need this function? Can't we simply use `s.ControlPlane.Annotations` at places wherever required?

azure/scope/managedcontrolplane.go
```diff
@@ -530,6 +534,10 @@ func (s *ManagedControlPlaneScope) GetAllAgentPoolSpecs(ctx context.Context) ([]
 	return ammps, nil
 }
 
+func (s *ManagedControlPlaneScope) AgentPoolAnnotations() map[string]string {
+	return s.InfraMachinePool.Annotations
+}
+
```
sonasingh46, line 540: Same as above

azure/scope/managedcontrolplane.go
sonasingh46: I see why it is so. You basically implemented it as part of `ManagedMachinePoolScope` interface.

azure/services/agentpools/agentpools.go
```diff
@@ -149,6 +153,17 @@ func (s *Service) Reconcile(ctx context.Context) error {
 	return nil
 }
 
+// get custom headers to pass to agent pool client from agent pool annotations.
+func (s *Service) getCustomHeaders() map[string]string {
+	var result = map[string]string{}
+	for key, value := range s.scope.AgentPoolAnnotations() {
+		if strings.HasPrefix(key, azure.CustomHeaderPrefix) {
+			result[strings.TrimPrefix(key, azure.CustomHeaderPrefix)] = value
```
sonasingh46, line 161: Just a thought: 
What if the value is empty? We can have probably move this function to some better place that takes input as `annotations` and `prefix` and returns the filtered values.  This could help reuse. 

And here we can probably just sanitise the values if required.

azure/services/managedclusters/managedclusters.go
```diff
@@ -368,3 +371,14 @@ func (s *Service) Delete(ctx context.Context) error {
...
+// get custom headers that will be passed to the client from control plane annotations.
+func (s *Service) getCustomHeaders() map[string]string {
+	var result = map[string]string{}
+	for key, value := range s.Scope.ManagedClusterAnnotations() {
+		if strings.HasPrefix(key, azure.CustomHeaderPrefix) {
+			result[strings.TrimPrefix(key, azure.CustomHeaderPrefix)] = value
+		}
+	}
+	return result
+}
```
sonasingh46, line 384: Refer: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2020/files#r795940312

CecileRobertMichon:
AKS test is broken, should be fixed as soon as #2025 merges

azure/defaults.go
michalno1: Definitely. Do you suggest opening a thread in `Disussions` or should we do it here? Do you have any suggestions? Maybe we can use `cluster-azure.x-k8s.io/` for all project annotations and `cluster-azure.x-k8s.io/custom-header-{HEADER_NAME}` specifically for this feature?

azure/scope/managedcontrolplane.go
michalno1: This particular method implements `ManagedClusterScope` rather than `ManagedMachinePoolScope`. This is done in the same spirit as `ManagedClusterSpec()`.

azure/services/agentpools/agentpools.go
michalno1: Refactored for reuse. What issue do you see with an empty value? We'll just pass a header with empty value if that is what someone intends.

azure/defaults.go
sonasingh46: `cluster-azure.x-k8s.io` -- This one sounds good to me. Copying @CecileRobertMichon @shysank to provide more input.

azure/services/agentpools/agentpools.go
sonasingh46: Sure. My only point was -- does the azure api tolerates the empty value. I do not know. Just thought to bring up, in case. If yes, I do not think we need to do anything here.

azure/services/managedclusters/client.go
```diff
@@ -78,11 +78,19 @@ func (ac *AzureClient) GetCredentials(ctx context.Context, resourceGroupName, na
 }
 
 // CreateOrUpdate creates or updates a managed cluster.
-func (ac *AzureClient) CreateOrUpdate(ctx context.Context, resourceGroupName, name string, cluster containerservice.ManagedCluster) (containerservice.ManagedCluster, error) {
+func (ac *AzureClient) CreateOrUpdate(ctx context.Context, resourceGroupName, name string, cluster containerservice.ManagedCluster, headers map[string]string) (containerservice.ManagedCluster, error) {
```
CecileRobertMichon, line 81: this is another case where it won't work super well with the [async interface](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/services/async/interfaces.go#L48) once we move this func to async... @shysank @Jont828 @sonasingh46 

For now I think it's fine to add it in here as-is but we'll need to have a way to expand ResourceSpecGetter when custom fields are needed like here

azure/defaults.go
CecileRobertMichon: how about `infrastructure.cluster.x-k8s.io`? `cluster-azure.x-k8s.io` is the API Group for Cluster API. CAPZ resources use `infrastructure.cluster.x-k8s.io` as its Group.


https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/2349883d1effe6344690459b40e5dc26138010df/api/v1beta1/groupversion_info.go#L29

azure/services/managedclusters/client.go
CecileRobertMichon: actually for this one particularly @jackfrancis pointed out it would be something useful to have for all clients, so we could just add it as a generic capability of ResourceSpecGetter (ie.` spec.CustomHeaders`). Not a comment for this PR, just thinking out loud.

azure/defaults.go
michalno1: @CecileRobertMichon makes sense, thanks. I've changed it.

azure/services/managedclusters/client.go
sonasingh46: >For now I think it's fine to add it in here as-is but we'll need to have a way to expand ResourceSpecGetter when custom fields are needed like here

Right. Also, I had some ideas on this one that I pasted on slack sometime back. I will open up an issue/discussion -- highlighting the challenges that we faced and we can actually brainstorm/groom to get to some solution.

azure/services/agentpools/agentpools.go
sonasingh46: Sure.

util/maps/maps.go
```diff
@@ -0,0 +1,35 @@
...
+	for key, value := range input {
+		if strings.HasPrefix(key, prefix) {
+			remainingKey := strings.TrimPrefix(key, prefix)
+			if len(remainingKey) > 0 {
+				result[remainingKey] = value
+			}
+		}
+	}
+	return result
+}
```
sonasingh46, line 35: Unit test for this will be great to have.

michalno1: Sure, done.

azure/defaults.go
```diff
@@ -84,6 +84,14 @@ const (
 	ProviderIDPrefix = "azure://"
 )
 
+const (
+	// CustomHeaderPrefix is the prefix of annotations that enable additional cluster / node pool features.
+	// Whatever follows the prefix will be passed as a header to cluster/node pool creation/update requests.
+	// E.g. add `"custom-headers.azure.dev/UseGPUDedicatedVHD": "true"` annotation to AzureManagedMachinePool CR
+	// to enable creating GPU nodes by the node pool.
+	CustomHeaderPrefix = "cluster-azure.x-k8s.io/custom-header-"
```
jackfrancis, line 92: I don't think we can include a trailing hyphen here, and in fact, I think we can't include a `/` in the prefix as well. According to this document, the `/` is meant to delimit the prefix from the annotation "key" (in a key/value context):

https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set

Why not use `azuremanagedcluster.cluster.x-k8s.io` as the prefix? This would yield something like this:

```
kind: AzureManagedMachinePool
metadata:
  annotations:
    azuremanagedcluster.cluster.x-k8s.io/UseGPUDedicatedVHD: true
```

michalno1: `CustomHeaderPrefix` is not meant to be an annotation prefix- it's almost a full annotation and only misses the name of the header (that is a simple string that doesn't contain `/` or `-`). I don't see a conflict between annotation syntax and this implementation. Also this implementation is future proof- it will be possible to enable existing and future preview features this way.

You suggest passing everything that is prefixed with `azuremanagedcluster.cluster.x-k8s.io/` as a API request header?

azure/defaults.go
```diff
@@ -84,6 +84,14 @@ const (
 	ProviderIDPrefix = "azure://"
 )
 
+const (
+	// CustomHeaderPrefix is the prefix of annotations that enable additional cluster / node pool features.
+	// Whatever follows the prefix will be passed as a header to cluster/node pool creation/update requests.
+	// E.g. add `"custom-headers.azure.dev/UseGPUDedicatedVHD": "true"` annotation to AzureManagedMachinePool CR
```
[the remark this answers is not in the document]
michalno1: Fixed

azure/defaults.go
jackfrancis: I understand now, thanks for the clarification!

jackfrancis (approved):
/lgtm

azure/services/agentpools/agentpools.go
```diff
@@ -137,7 +140,8 @@ func (s *Service) Reconcile(ctx context.Context) error {
 		diff := cmp.Diff(normalizedProfile, existingProfile)
 		if diff != "" {
 			log.V(2).Info(fmt.Sprintf("Update required (+new -old):\n%s", diff))
-			err = s.Client.CreateOrUpdate(ctx, agentPoolSpec.ResourceGroup, agentPoolSpec.Cluster, agentPoolSpec.Name, profile)
+			err = s.Client.CreateOrUpdate(ctx, agentPoolSpec.ResourceGroup, agentPoolSpec.Cluster, agentPoolSpec.Name,
```
shysank, line 143: can we enable/disable features after node pool creation?

azure/services/managedclusters/managedclusters.go
```diff
@@ -322,7 +325,7 @@ func (s *Service) Reconcile(ctx context.Context) error {
 		diff := computeDiffOfNormalizedClusters(managedCluster, existingMC)
 		if diff != "" {
 			klog.V(2).Infof("Update required (+new -old):\n%s", diff)
-			managedCluster, err = s.Client.CreateOrUpdate(ctx, managedClusterSpec.ResourceGroupName, managedClusterSpec.Name, managedCluster)
+			managedCluster, err = s.Client.CreateOrUpdate(ctx, managedClusterSpec.ResourceGroupName, managedClusterSpec.Name, managedCluster, customHeaders)
```
shysank, line 328: similar comment as above regarding enabling/disabling features after cluster creation?

shysank (commented):
Thanks for the pr @michalno1! The changes in general looks good. Just a couple of clarifications regarding cluster/nodepool update. wdyt about adding some [documentation](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/docs/book/src/topics/managedcluster.md) for this feature? I'm fine with doing it in a followup pr if it's easier.

azure/services/agentpools/agentpools.go
jackfrancis: I think the most sane way to move forward is to support "enable feature via custom header" on create only. There may be some aks nodepool update operations that are initiated via an "empty" PUT w/ HTTP header data, but that's kind of weird from a UX perspective and I would not expect to rely upon that.

As @zmalik suggests, we should consider feature configuration delivered via HTTP headers as an interface for AKS to rapidly deliver experimental features, and then eventually they may graduate to fully supported features (after which point capz would have to implement them in their final form).

Does that make sense? Is the PR as-is implemented to forbid updates of custom headers after cluster/nodepool creation?

shysank: The PR right now allows for updates, although not intentionally. If there is a non zero `diff` along with annotations changes, the new headers will be set. I'd rather we be more explicit by adding a validation webhook to make it immutable. wdyt?

jackfrancis: +1

michalno1:
> Thanks for the pr @michalno1! The changes in general looks good. Just a couple of clarifications regarding cluster/nodepool update. wdyt about adding some [documentation](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/docs/book/src/topics/managedcluster.md) for this feature? I'm fine with doing it in a followup pr if it's easier.

@shysank I've added some docs

azure/services/agentpools/agentpools.go
michalno1: @shysank added immutability check

azure/services/managedclusters/managedclusters.go
michalno1: @shysank added validation webhook with immutability check to AzureManagedCluster as well.

exp/api/v1beta1/azuremanagedcluster_webhook.go
```diff
@@ -0,0 +1,74 @@
...
+	old := oldRaw.(*AzureManagedCluster)
+	var allErrs field.ErrorList
+
+	// custom headers are immutable
+	oldCustomHeaders := maps.FilterByKeyPrefix(old.ObjectMeta.Annotations, azure.CustomHeaderPrefix)
+	newCustomHeaders := maps.FilterByKeyPrefix(r.ObjectMeta.Annotations, azure.CustomHeaderPrefix)
+	if !reflect.DeepEqual(oldCustomHeaders, newCustomHeaders) {
+		allErrs = append(allErrs,
+			field.Invalid(
+				field.NewPath("ObjectMetadata", "Annotations"),
```
shysank, line 59: We want the errors to suggest yaml path.
```suggestion
				field.NewPath("metadata", "annotations"),
```

exp/api/v1beta1/azuremanagedcluster_webhook_test.go
```diff
@@ -0,0 +1,127 @@
...
+					Annotations: map[string]string{
+						"infrastructure.cluster.x-k8s.io/custom-header-SomeFeature": "false",
+					},
+				},
+				Spec: AzureManagedClusterSpec{},
+			},
+			wantErr: true,
+		},
+		{
+			name: "AzureManagedCluster annotations passed as custom headers are immutable 2",
```
shysank, line 56: nit: can we be more specific with the naming?

exp/api/v1beta1/azuremanagedcluster_webhook_test.go
```diff
@@ -0,0 +1,127 @@
...
+			amc: &AzureManagedCluster{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: map[string]string{},
+				},
+				Spec: AzureManagedClusterSpec{},
+			},
+			wantErr: true,
+		},
+		{
+			name: "AzureManagedCluster annotations passed as custom headers are immutable 3",
```
shysank, line 74: nit: can we be more specific with the naming?

exp/api/v1beta1/azuremanagedmachinepool_webhook_test.go
```diff
@@ -254,6 +254,87 @@ func TestAzureManagedMachinePoolUpdatingWebhook(t *testing.T) {
...
+					Annotations: map[string]string{
+						"infrastructure.cluster.x-k8s.io/custom-header-SomeFeature": "false",
+					},
+				},
+				Spec: AzureManagedMachinePoolSpec{},
+			},
+			wantErr: true,
+		},
+		{
+			name: "custom headers are immutable 2",
```
shysank, line 278: nit: can we be more specific with the naming?

exp/api/v1beta1/azuremanagedmachinepool_webhook_test.go
```diff
@@ -254,6 +254,87 @@ func TestAzureManagedMachinePoolUpdatingWebhook(t *testing.T) {
...
+			new: &AzureManagedMachinePool{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: map[string]string{},
+				},
+				Spec: AzureManagedMachinePoolSpec{},
+			},
+			wantErr: true,
+		},
+		{
+			name: "custom headers are immutable 3",
```
shysank, line 296: nit: can we be more specific with the naming?

exp/api/v1beta1/azuremanagedmachinepool_webhook.go
```diff
@@ -103,6 +105,17 @@ func (r *AzureManagedMachinePool) ValidateUpdate(oldRaw runtime.Object, client c
 				"field is immutable"))
 	}
 
+	// custom headers are immutable
+	oldCustomHeaders := maps.FilterByKeyPrefix(old.ObjectMeta.Annotations, azure.CustomHeaderPrefix)
+	newCustomHeaders := maps.FilterByKeyPrefix(r.ObjectMeta.Annotations, azure.CustomHeaderPrefix)
+	if !reflect.DeepEqual(oldCustomHeaders, newCustomHeaders) {
+		allErrs = append(allErrs,
+			field.Invalid(
+				field.NewPath("ObjectMetadata", "Annotations"),
```
shysank, line 114:
```suggestion
				field.NewPath("metadata", "annotations"),
```

exp/api/v1beta1/azuremanagedcluster_webhook_test.go
michalno1: fixed

exp/api/v1beta1/azuremanagedmachinepool_webhook_test.go
michalno1: fixed

exp/api/v1beta1/azuremanagedmachinepool_webhook_test.go
michalno1: fixed

exp/api/v1beta1/azuremanagedcluster_webhook_test.go
michalno1: fixed

exp/api/v1beta1/azuremanagedcluster_webhook.go
michalno1: fixed

shysank:
/lgtm
/approve
Great job @michalno1! Thanks for being patient with the reviews.

PR #1807, opened 2021-10-29
Add externally managed infrastructure docs

enxebre:
**What type of PR is this?**

/kind documentation

**What this PR does / why we need it**:
Add docs for externally managed infrastructure.
Counter part for AWS https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/2895.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Include docs for externally managed infrastructure.
```

PR #2023, opened 2022-01-31
Properly respect externally managed annotation

alvaroaleman:
**What type of PR is this?**

/kind bug

We already have a predicate for events from AzureClusters that skip
their reconciliation if they are managed externally. We do however not
have such a check for events that come from a Cluster, resulting in
us doing a reconciliation for AzureClusters anyways.
    
The approach is copied from capa:
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/cf6fea721385d490b570b4cf119defea705b3320/controllers/awscluster_controller.go#L331-L377

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
The azurecluster reconciler now always respects the ExternallyManaged annotation
```


controllers/azurecluster_controller.go
```diff
@@ -97,7 +99,7 @@ func (acr *AzureClusterReconciler) SetupWithManager(ctx context.Context, mgr ctr
 	// Add a watch on clusterv1.Cluster object for unpause notifications.
 	if err = c.Watch(
 		&source.Kind{Type: &clusterv1.Cluster{}},
-		handler.EnqueueRequestsFromMapFunc(util.ClusterToInfrastructureMapFunc(infrav1.GroupVersion.WithKind("AzureCluster"))),
+		handler.EnqueueRequestsFromMapFunc(clusterToInfrastructureMapFunc(ctx, log, mgr.GetClient())),
```
sonasingh46, line 102: Any specific reason to pass `log` as an arg? I think you can get an instance of logger from ctx itself.
Ref: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/controllers/azurecluster_controller.go#L76

controllers/azurecluster_controller.go
```diff
@@ -296,3 +298,43 @@ func (acr *AzureClusterReconciler) reconcileDelete(ctx context.Context, clusterS
...
+		azureCluster := &infrav1.AzureCluster{}
+		key := types.NamespacedName{Namespace: cluster.Spec.InfrastructureRef.Namespace, Name: cluster.Spec.InfrastructureRef.Name}
+
+		if err := c.Get(ctx, key, azureCluster); err != nil {
+			log.V(4).Error(err, "Failed to get Azure cluster")
+			return nil
+		}
+
+		if annotations.IsExternallyManaged(azureCluster) {
+			log.V(4).Info("AWSCluster is externally managed, skipping mapping.")
```
sonasingh46, line 327: nit: AWSCluster --> AzureCluster

enxebre:
Can we please make sure the tag can't be removed via webhook? see https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/2383#diff-bede819718d1d34b23f65d8e8aad84aae8aa8311be130d14c8dc81d558ec44e1R108

controllers/azurecluster_controller.go
enxebre, line 102: wouldn't we need to just add an event filter as in https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2023/files#diff-c9f392795aea4d131d38cf4a128c8f6be728ca2ec8b8307cad6ecf1fb7ca5da1L91?

enxebre:
For ref this was already documented here https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1807

controllers/azurecluster_controller.go
alvaroaleman: No specific reason, updated, lmk what you think

controllers/azurecluster_controller.go
alvaroaleman: Fixed

controllers/azurecluster_controller.go
alvaroaleman: The reason a predicate can not be used is that the the annotation only exists on the azurecluster, but not on the capi cluster, so we have to fetch the corresponding azure cluster in order to find out if it has the annotation. That predicate only works for a watch on the azure cluster

alvaroaleman:
@sonasingh46 @enxebre addressed the feedback, please take another look

controllers/azurecluster_controller.go
```diff
@@ -296,3 +297,44 @@ func (acr *AzureClusterReconciler) reconcileDelete(ctx context.Context, clusterS
 
 	return reconcile.Result{}, nil
 }
+
+func clusterToInfrastructureMapFunc(ctx context.Context, c client.Client) handler.MapFunc {
+	log := ctrl.LoggerFrom(ctx)
```
CecileRobertMichon, line 302: in CAPZ we have a logger used for tracing, see #1806 

```suggestion
	ctx, log, done := tele.StartSpanWithLogger(ctx, "controllers.AzureClusterReconciler.clusterToInfrastructureMapFunc")
	defer done()
```

controllers/azurecluster_controller.go
```diff
@@ -296,3 +297,46 @@ func (acr *AzureClusterReconciler) reconcileDelete(ctx context.Context, clusterS
 
 	return reconcile.Result{}, nil
 }
+
+func clusterToInfrastructureMapFunc(ctx context.Context, c client.Client, controllerName string) handler.MapFunc {
+	return func(o client.Object) []ctrl.Request {
+		ctx, log, done := tele.StartSpanWithLogger(ctx, "controllers."+controllerName+".clusterToInfrastructureMapFunc")
```
CecileRobertMichon, line 303: this function is in the azurecluster_controller file, if it's meant to be used by multiple controllers should it be somewhere common instead? maybe helpers.go?

Also side question, if this needed for other providers too (sounds like it was added to CAPA), why not change the `ClusterToInfrastructureMapFunc` function in CAPI directly to skip externally managed clusters instead of overriding it with our own `clusterToInfrastructureMapFunc` func?

alvaroaleman: Good point, opened https://github.com/kubernetes-sigs/cluster-api/pull/6039 for that and will update this and capa once it merges

alvaroaleman: Updated to bump cluster-api and use the new ClusterToInfrastructureMapFuncWithExternallyManagedCheck, please have another look.

CecileRobertMichon (commented):
lgtm

should we wait for an actual release of CAPI to merge this?

alvaroaleman:
> should we wait for an actual release of CAPI to merge this?

Not sure, what is the common approach for a situation like this where we need a bugfix from an unreleased version?

sonasingh46 (approved):
lgtm

CecileRobertMichon:
/hold

let's wait for a CAPI release since we don't want to release CAPZ with an unreleased version of CAPI

go.mod
```diff
@@ -42,7 +42,7 @@ require (
 	k8s.io/klog/v2 v2.30.0
 	k8s.io/kubectl v0.23.0
 	k8s.io/utils v0.0.0-20210930125809-cb0fa318a74b
-	sigs.k8s.io/cluster-api v1.1.0
+	sigs.k8s.io/cluster-api v1.1.1-0.20220203212859-13fcc674c7ab
```
CecileRobertMichon, line 45: @alvaroaleman v1.1.1 is out with this change, we should be able to update and merge this now

thanks again

alvaroaleman: Thanks, updated

CecileRobertMichon (commented):
/lgtm
/approve
/hold cancel

Issue #2026, opened 2022-01-31
verify targets should be in parity with capi


sonasingh46:
/kind bug

**What did you expect to happen:**
Did a comparison of verify targets used in core capi vs capz and found out that couple of verify targets are missing in capz. If it makes sense, we could add those missing targets in capz too. This should help improve the developer experience in my opinion. The following section has the comparison matrix. 

TL;DR

Missing targets in CAPZ but present in CAPI:  
- lint-fix
- apidiff
- verify-conversions
- verify-tiltfile
- verify-book-links

We can discuss about these and if it makes sense then we can add these targets too. 

**Comparison Matrix:**
Validations (executed via make)
S.no | make target| CAPI | CAPZ| CAPZ
-- | -- | -- | --| --
1 | lint-fix | ✔ | ❌| runs golangci-lint with --fix 
2 | apidiff | ✔ | ❌| diffs API changes on PRs
3 | verify | ✔ | ✔| target which runs all the other verify targets
4 | verify-modules | ✔ | ✔| verifies go.mod’s are up-to-date
5 | verify-gen | ✔ | ✔| verifies all generated files are up-to-date
6 | verify-conversions | ✔ | ❌| verifies all API types implement conversions
7 | verify-boilerplate | ✔ | ✔| verifies file headers
8 | verify-shellcheck | ✔ | ✔| verifies shell scripts
9 | verify-tiltfile | ✔ | ❌| verifies tiltfile
10 | verify-book-links | ✔ | ❌| verifies links in books

**Note:** The `lint-fix` target will get fixed by PR [1987](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1987)

**Environment:**

- cluster-api-provider-azure version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

CecileRobertMichon:
apidiff is a script: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/scripts/ci-apidiff.sh

it's also a script in CAPI https://github.com/kubernetes-sigs/cluster-api/blob/main/scripts/ci-apidiff.sh, wonder if it's still being used

sonasingh46:
The script is used in PROW job. Thanks for pointing put the apidiff script in CAPZ. 
The corresponding target name in capz is `go-apidiff`. Wondering if it makes sense to have the same target name as that of CAPI to help improve contributor experience .

[...]

PR #2032, opened 2022-02-01
add apidiff and verify tilt make target

sonasingh46:
/kind cleanup

This PR adds the following make targets: 
1. `make apidiff` : This target in turn calls the existing `go-apidiff`
2. `make verify-tiltfile`: This verifies the tiltfile format.
3. `make format-tiltfile`: This formats the tiltfile. 

Reference issue #2026 

Open for suggestions if the `make apidiff` --> `make go-apidiff` makes sense. I feel it should help in the contributor experience. 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests
```release-note
None
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -664,8 +664,16 @@ clean-temporary: ## Remove all temporary files and folders
 clean-release: ## Remove the release folder
 	rm -rf $(RELEASE_DIR)
 
+.PHONY: apidiff
+apidiff: $(GO_APIDIFF) ## Check for API differences
+	$(GO_APIDIFF) $(shell git rev-parse origin/main) --print-compatible
+
+.PHONY: format-tiltfile
+format-tiltfile: ## Format the Tiltfile
+	./hack/verify-starlark.sh fix
+
 .PHONY: verify
-verify: verify-boilerplate verify-modules verify-gen verify-shellcheck verify-conversions
+verify: verify-boilerplate verify-modules verify-gen verify-shellcheck verify-conversions verify-tiltfile
 
 .PHONY: verify-boilerplate
 verify-boilerplate:
@@ -690,3 +698,7 @@ verify-shellcheck:
 .PHONY: verify-conversions
 verify-conversions: $(CONVERSION_VERIFIER)  ## Verifies expected API conversion are in place
 	$(CONVERSION_VERIFIER)
+
+.PHONY: verify-tiltfile
+verify-tiltfile: ## Verify Tiltfile format
+	./hack/verify-starlark.sh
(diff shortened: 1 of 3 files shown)


mboersma (approved):
/lgtm

I tried out these targets locally. I thought it odd that `verify-tiltfile` isn't a sub-target of `make verify`, but I see that's how CAPI has it as well, so :shrug:.

CecileRobertMichon:
>  I thought it odd that verify-tiltfile isn't a sub-target of make verify

+1 to adding it to make verify, if it's not run by make verify it won't run in CI which means we won't actually use it to gate PRs

sonasingh46:
+1 To add it to make verify. It was a miss from me. Also, capi does have the `verify-tiltfile` as a sub target of `make verify` -- just via little substitutions.

sonasingh46:
@CecileRobertMichon @mboersma -- Added it to `make verify` target.

jackfrancis:
@CecileRobertMichon should we update test-infra so that the `pull-cluster-api-provider-azure-apidiff` invokes `make apidiff` instead of calling the shell script explicitly? Ref:

- https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-sigs/cluster-api-provider-azure/cluster-api-provider-azure-presubmits-main.yaml#L422

(Ditto the three release presubmit specs that we currently maintain.)

CecileRobertMichon:
> should we update test-infra so that the pull-cluster-api-provider-azure-apidiff invokes make apidiff instead of calling the shell script explicitly?

Hmm good point. @sonasingh46 @sbueringer looks like CAPI also calls a shell script despite having a make target https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-sigs/cluster-api/cluster-api-presubmits-main.yaml#L35, thoughts?

sonasingh46:
@CecileRobertMichon @jackfrancis -- AFAIK, there is little difference in the `make <targets>` vs the `shell scripts`. There are some custom things that may need to be done. In this case, the script runs the `apidiff` on pull based sha. 
Ref: https://github.com/kubernetes-sigs/cluster-api/blob/main/scripts/ci-apidiff.sh#L28
This sha is exposed as an env var by the prow job.

sbueringer:
As Ashutosh said, there's a difference between the makefile target and the shell script.

The Makefile target is comparing the current state against origin/main, the shell script against `PULL_BASE_SHA`

Maybe we could improve this a bit and make the API_DIFF_BASE_SHA configurable in the makefile target (with a default of `shell git rev-parse origin/main`) and then call the makefile target from the shell script (with PULL_BASE_SHA)?

Related observation. In core CAPI afaik all Jobs are calling `scripts/ci-<>.sh` shell scripts and those are then running the makefile targets, in some cases they are also doing some additional stuff before. (although ideally they don't)
My impression is that this is done to decouple the Job configs from the Makefile targets.

Issue #1989, opened 2022-01-20
Conversion is not completely implemented for all API types


sbueringer:
/kind bug

Hey folks,
in core CAPI we have a tool called conversion-verifier which loops over all types and ensures the controller-runtime conversion interfaces are implemented correctly, i.e. it checks the types have the Hub / ConvertTo / ConvertFrom functions.

The results for CAPZ are:
* type "infrastructure.cluster.x-k8s.io/AzureClusterIdentityList" in package "sigs.k8s.io/cluster-api-provider-azure/api/v1alpha3" it's not convertible, missing ConvertFrom() and ConvertTo() methods
* type "infrastructure.cluster.x-k8s.io/AzureManagedClusterList" in package "sigs.k8s.io/cluster-api-provider-azure/exp/api/v1alpha3" it's not convertible, missing ConvertFrom() and ConvertTo() methods
* type "infrastructure.cluster.x-k8s.io/AzureManagedControlPlaneList" in package "sigs.k8s.io/cluster-api-provider-azure/exp/api/v1alpha3" it's not convertible, missing ConvertFrom() and ConvertTo() methods
* type "infrastructure.cluster.x-k8s.io/AzureMachinePoolList" in package "sigs.k8s.io/cluster-api-provider-azure/exp/api/v1alpha3" it's not convertible, missing ConvertFrom() and ConvertTo() methods
* type "infrastructure.cluster.x-k8s.io/AzureManagedMachinePoolList" in package "sigs.k8s.io/cluster-api-provider-azure/exp/api/v1alpha3" it's not convertible, missing ConvertFrom() and ConvertTo() methods

I think it would be great to:
* implement the missing methods
* integrate the conversion-verifier in CI

The conversion-verifier can be found here:
* Code: https://github.com/kubernetes-sigs/cluster-api/tree/main/hack/tools/conversion-verifier
* Usage: https://github.com/kubernetes-sigs/cluster-api/blob/main/Makefile#L339-L341 (core CAPI runs it as part of the verify job)

I can provide additional details if needed / if you want to implement this issue.

**Environment:**

- cluster-api-provider-azure version: latest
- Kubernetes version: (use `kubectl version`): -
- OS (e.g. from `/etc/os-release`): -

mboersma:
/close

Fixed by #2042

PR #2040, opened 2022-02-02
[WIP] Add missing conversion funcs

mboersma:
**What type of PR is this?**

/kind cleanup

**What this PR does / why we need it**:

Implements several missing conversion funcs suggested by the `conversion-verifier` tool, and does some related code cleanup.

**Which issue(s) this PR fixes**:

Fixes #1989

**Special notes for your reviewer**:

I am also trying to add CAPI's `conversion-verifier` tool to our `make verify` target, but am currently stumped because of a vendoring issue with `replace` statements in the hack/tools go.mod. We can consider this a WIP until I figure that out.

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

PR #2042, opened 2022-02-02
add verify-conversions target

sonasingh46:
/kind cleanup

This PR adds the following make targets: 
1. `make verify-conversions`: Verifies that the expected conversions are in place.

Reference issue #2026 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests
```release-note
None
```


CecileRobertMichon:
@sonasingh46 looks like this might overlap with #2040 (there was another issue opened a while ago about missing conversion func and I guess you and @mboersma both started working on it at the same time through different issues)

sonasingh46:
@CecileRobertMichon -- Sorry, I was not aware. This PR has some Makefile changes too. Either @mboersma can pick those from here in his same PR and then I will close this PR. Or once his PR gets merged, I can rebase and can get the Makefile changes in.

mboersma:
Ha! No, this is entirely similar to my PR, except that I added in a bunch of unrelated code cleanup. I'll close my PR and review this one instead, thanks @sonasingh46!

mboersma:
This PR fixes #1989.

mboersma (approved):
/lgtm

I tested it locally as well and everything looks good.

hack/tools/tools.go
```diff
@@ -0,0 +1,25 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// This package imports things required by build scripts, to force `go mod` to see them as dependencies
+package tools
+
+import (
+	_ "sigs.k8s.io/cluster-api/hack/tools"
```
CecileRobertMichon, line 24: do we need this? why not install it with go install how we install others like envsubst?

mboersma: The CAPI hack/tools package has a `replace` statement in it with a fake SHA that prevents `go install` (or `go get`) from working. Adding a local package in CAPZ and a local go.mod seems like the only solution, at least that's where I got to as well when looking at this.

PR #2046, opened 2022-02-03
Don't ignore kubeconfig directory in git

alexander-demichev:
**What type of PR is this?**
/kind bug

We are using `go mod vendor` for our internal builds and the current `.gitignore` excludes `sigs.k8s.io/cluster-api/util/kubeconfig` package from the vendor directory which breaks the build. I added `!kubeconfig/` to make it not ignore kubeconfig package and still ignore kubeconfig when it's a file.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Don't ignore kubeconfig directory in git
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,7 @@ envfile
 kind.kubeconfig
 minikube.kubeconfig
 kubeconfig
+!kubeconfig/
 
 # ssh keys
 .ssh*


mboersma (approved):
/lgtm

I agree the intent was to skip kubeconfig files, not a directory with that name. I guess this never caused problems because we don't do `go mod vendor` to actually pull code into vendor/.

PR #2050, opened 2022-02-04
Handle terminal errors in AzureCluster controller

### LINKED ISSUE #1973: Handle `TerminalErrors` in the AzureCluster controller
/kind feature

Currently, the AzureMachine controller handles both transient and terminal errors, but the AzureCluster controller only handles transient errors. By adding support for terminal errors, we can return terminal errors in Reconcile() for services like subnets.

**Environment:**

sayantani11:
**What type of PR is this?**

Fixes #1973 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Handle terminal errors in AzureCluster controller
```

### DIFF
--- a/controllers/azurecluster_controller.go
+++ b/controllers/azurecluster_controller.go
@@ -219,9 +219,15 @@ func (acr *AzureClusterReconciler) reconcileNormal(ctx context.Context, clusterS
 	}
 
 	if err := acs.Reconcile(ctx); err != nil {
-		// Handle transient errors
+		// Handle terminal & transient errors
 		var reconcileError azure.ReconcileError
 		if errors.As(err, &reconcileError) {
+			if reconcileError.IsTerminal() {
+				acr.Recorder.Eventf(clusterScope.AzureCluster, corev1.EventTypeWarning, "ReconcileErrror", errors.Wrapf(err, "failed to reconcile AzureCluster").Error())
+				log.Error(err, "failed to reconcile AzureCluster", "name", clusterScope.ClusterName())
+				conditions.MarkFalse(azureCluster, infrav1.NetworkInfrastructureReadyCondition, infrav1.FailedReason, clusterv1.ConditionSeverityError, "")
+				return reconcile.Result{}, nil
+			}
 			if reconcileError.IsTransient() {
 				if azure.IsOperationNotDoneError(reconcileError) {
 					log.V(2).Info(fmt.Sprintf("AzureCluster reconcile not done: %s", reconcileError.Error()))


controllers/azurecluster_controller.go
```diff
@@ -219,9 +219,14 @@ func (acr *AzureClusterReconciler) reconcileNormal(ctx context.Context, clusterS
 	}
 
 	if err := acs.Reconcile(ctx); err != nil {
-		// Handle transient errors
+		// Handle terminal & transient errors
 		var reconcileError azure.ReconcileError
 		if errors.As(err, &reconcileError) {
+			if reconcileError.IsTerminal() {
+				acr.Recorder.Eventf(clusterScope.AzureCluster, corev1.EventTypeWarning, "ReconcileErrror", errors.Wrapf(err, "failed to reconcile AzureCluster").Error())
+				log.Error(err, "failed to reconcile AzureCluster", "name", clusterScope.ClusterName())
```
CecileRobertMichon, line 227: let's also add a line to mark the `NetworkInfrastructureReadyCondition` condition as not ready so the status shows the cluster as not ready:

```
conditions.MarkFalse(azureCluster, infrav1.NetworkInfrastructureReadyCondition, infrav1.FailedReason, clusterv1.ConditionSeverityInfo, "")
```

sonasingh46:
Question: Are terminal errors applicable only for `AzureCluster` and `AzureMachine` controller? May be out of scope of this PR but if applicable we can consider adding them too as a followup. Will need to check on wether it is applicable or not. 
cc @Jont828 @CecileRobertMichon

controllers/azurecluster_controller.go
sayantani11: So its's getting added after `log.error`

jackfrancis: Curious why we use (re-use?) the `NetworkInfrastructureReadyCondition` const to indicate reconcile status for the high-level `AzureCluster` resource type. Are we missing a const like `ClusterReadyCondition`?

CecileRobertMichon: Great question. The `Ready` Condition is set at the end of each loop by the `SetSummary` call [here](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/scope/cluster.go#L658) (PatchObject is part of a defer in the AzureCluster Reconcile()). `SetSummary` sets the Ready condition based on all the other conditions we own (False if at least one is False, True if all are True, not set if no other Conditions are set). 

The `NetworkInfrastructureReadyCondition` was the only Condition defined on AzureCluster before we introduced more granular Conditions for each Service. It's used as a generic "Is the Azure Cluster infrastructure ready" condition in other places, eg. AzureClusterIdentity namespace list doesn't include the namespace this Cluster is in. We could directly set the `Ready` condition here instead of waiting for SetSummary() to take care of it as long as it doesn't get overwritten by SetSummary (I don't think it should, but would need to check).

Once #1541 is fully implemented and each service is setting its own conditions, we could potentially get rid of the generic `NetworkInfrastructureReadyCondition` since it would become redundant at that point.

CecileRobertMichon:
> Question: Are terminal errors applicable only for AzureCluster and AzureMachine controller. May be out of scope of this PR but if applicable we can consider adding them too as a followup. Will need to check on wether it is applicable or not.

Agreed, but since this is a good-first-issue PR let's keep the original scope of just AzureCluster controller.

controllers/azurecluster_controller.go
jackfrancis: Very useful backstory, thank you! I am going to file an "audit ready conditions" issue and link to this issue so we don't forget to do this, and remember what we were talking about.

jackfrancis: Another question: why are we using `clusterv1.ConditionSeverityInfo` for a failure condition instead of `clusterv1.ConditionSeverityError`?

CecileRobertMichon: that's a typo on my part, it should definitely be `clusterv1.ConditionSeverityError` 💯

controllers/azurecluster_controller.go
```diff
@@ -219,9 +219,15 @@ func (acr *AzureClusterReconciler) reconcileNormal(ctx context.Context, clusterS
...
+				conditions.MarkFalse(azureCluster, infrav1.NetworkInfrastructureReadyCondition, infrav1.FailedReason, clusterv1.clusterv1.ConditionSeverityError, "")
```
CecileRobertMichon, line 228:
```suggestion
				conditions.MarkFalse(azureCluster, infrav1.NetworkInfrastructureReadyCondition, infrav1.FailedReason, clusterv1.ConditionSeverityError, "")
```

jackfrancis (approved):
/lgtm

Issue #1801, opened 2021-10-28
Update CAPZ book to show debugging in Tilt


mboersma:
/kind feature

**Describe the solution you'd like**

Since kubernetes-sigs/cluster-api#5485 merged, it should be possible to debug the capz-controller-manager directly in the Tilt environment. We should make any changes necessary and document this thoroughly in the CAPZ book.

**Anything else you would like to add:**

**Environment:**

- cluster-api-provider-azure version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

PR #2053, opened 2022-02-07
docs: debugging information added

### LINKED ISSUE #588: Improve local debugging experience
/kind feature

It would be great if there was an easier way to debug locally, especially with the the webhooks still being used. 

There is a discussion on #572 around this with some ideas.

My current debugging workflow is:
1. make create-cluster
2.  Delete the deployment and webhook configuration:
```bash
kubectl delete deployment capz-controller-manager -n capz-system
kubectl delete ValidatingWebhookConfiguration capz-validating-webhook-configuration
```
3.  Set breakpoints in the capz code somewhere
4. Use a VSCode debug launch configuration to launch the manager
5. Apply cluster yaml

This approach is ok if you are ok ignoring the webhooks.

**Environment:**

/cc: @devigned

richardcase:
**What type of PR is this?**

/kind documentation

Documentation added on how to debug CAPZ using tilt.

Fixes #588 

**TODOs**:

- [ ] squashed commits
- [x] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Added documentation on debugging using Tilt.
```

### DIFF
--- a/docs/book/src/developers/development.md
+++ b/docs/book/src/developers/development.md
@@ -22,6 +22,7 @@
     - [Tilt for dev in both CAPZ and CAPI](#tilt-for-dev-in-both-capz-and-capi)
     - [Deploying a workload cluster](#deploying-a-workload-cluster)
     - [Viewing Telemetry](#viewing-telemetry)
+    - [Debugging](#debugging)
   - [Manual Testing](#manual-testing)
     - [Creating a dev cluster](#creating-a-dev-cluster)
       - [Building and pushing dev images](#building-and-pushing-dev-images)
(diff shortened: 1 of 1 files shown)


mboersma:
This also fixes #1801.

mboersma (approved):
/lgtm

mboersma:
This is great, thanks @richardcase!

By coincidence, I actually started making similar changes to enable debugging like this in the CAPZ-only setup. Would that also be useful, or should we just stick with CAPI's debugging configuration?

jackfrancis (approved):
/lgtm

richardcase:
> This is great, thanks @richardcase!
> 
> By coincidence, I actually started making similar changes to enable debugging like this in the CAPZ-only setup. Would that also be useful, or should we just stick with CAPI's debugging configuration?

@mboersma - i think we must have both seen the stale notification. It reminded me i hadn't done it. I like the consistency that the CAPI debug configuration gives as i can use that for CAPZ, CAPA, core etc. But guessing if there is benefit to something specific to CAPZ that could be be an addition?

mboersma:
Benefits to having a CAPZ-specific debugging configuration for Tilt would be convenience and keeping the project self-contained (although we document the CAPI+CAPZ debugging configuration already, obviously). But I'm not sure that outweighs the added complexity in CAPZ's Tiltfile, and it may be preferable to have one official way to debug CAPZ, which you have documented.

So I think I've talked myself out of it, but if you or anyone reading this sees more value in having a debugging configuration for the "Tilt with CAPZ" developer story, please say so. I'll keep the changes around.

Issue #2063, opened 2022-02-09
Update e2e tests to verify all system pods are ready before executing tests


jsturtevant:
/kind feature

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]
Our e2e suite should verify all pods in the kube-system namespace should be "ready" before running any tests.  This would have caught issues in the past due to conformance having this check. Calico was processing requests but not ready so our e2e tests passed where as conformance didn't.

See #2064 

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- cluster-api-provider-azure version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

jackfrancis:
@jsturtevant I'm not sure I understand the need for this, but I think I have an idea how we might do this. Do we still want to make this change to E2E tests that run in capz CI?

jsturtevant:
The reason for this is that we were able to create a cluster that past our CI but wasn't functional.  The Calico DS was deployed and was `Running`, but the pods were not `READY` (health check was failing).  This allowed our tests to pass which didn't exercise this network path.  We should make sure all pods in the system namespace (assumption is they are required for a fully functional cluster) are `READY` which would have caught this.

jackfrancis:
Got it.

I think for calico, once we implement a helm chart to install, we can add calico-specific validations into that flow (like we currently do for cloud-provider-azure to ensure cloud-controller-manager and cloud-node-manager pods).

If we want to do a global kube-system pods validation, we can create a standard set of `ControlPlaneWaiters` that all of our tests use so that we can wrap our validation requirements on top of the capi default waiters.

cc @CecileRobertMichon

jackfrancis:
@jsturtevant once we merge capi 1.2.0 support we can use this new interface to accomplish this:

https://github.com/kubernetes-sigs/cluster-api/pull/6855/files

cc @CecileRobertMichon

PR #2021, opened 2022-01-29
Generate calico manifests from source

CecileRobertMichon:
**What type of PR is this?**

**What this PR does / why we need it**: Instead of keeping a static Calico manifests in the repo, this PR uses the Kustomize feature to pull a base manifest from a url to get the vxlan manifest from https://docs.projectcalico.org/v3.20/manifests/calico-vxlan.yaml and makes a few patches necessary for Azure and IPv6. This will make sure we don't drift from the recommended Calico configuration and will make updating the Calico version a lot easier.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
partly addresses #1917 

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Generate calico manifests from source
```

PR #2064, opened 2022-02-09
Drop disable flag from calico spec so calico pod goes ready

jsturtevant:
**What type of PR is this?**
/kind failing-test

The PR https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2021 uses the calico manifests upstream. Those manifest have a bug in them that they reference AWS: https://github.com/projectcalico/calico/issues/5101

This wasn't caught in our e2e tests because we do not check if all the pods are ready (https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/2063) and even though it wasn't "ready" calico was functioning.  It was caught in the conformance tests because the test suite check all pods are ready.
 

Fixes #

this isn't really a nice solution but jsonpatch doesn't allow removing named fields from arrays.  The alternative is to copy all the variables and replace the env array completely.  As we plan to upgrade very quickly to the calico 3.22 release which doesn't have this issue and this is a quick fix to unblock conformance tests but happy to go other way.

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/templates/addons/calico/patches/remove-aws-reference.yaml
+++ b/templates/addons/calico/patches/remove-aws-reference.yaml
@@ -1,0 +1,5 @@
+# Drops the FELIX_AWSSRCDSTCHECK field
+# json patch doesn't allow for dropping fields in array by name
+# https://github.com/projectcalico/calico/issues/5101
+- op: remove
+  path: "/spec/template/spec/containers/0/env/12"
--- a/templates/addons/calico.yaml
+++ b/templates/addons/calico.yaml
@@ -3786,8 +3786,6 @@ spec:
             configMapKeyRef:
               key: veth_mtu
               name: calico-config
-        - name: FELIX_AWSSRCDSTCHECK
-          value: Disable
         - name: CALICO_DISABLE_FILE_LOGGING
           value: "true"
         - name: FELIX_DEFAULTENDPOINTTOHOSTACTION
--- a/templates/addons/calico/kustomization.yaml
+++ b/templates/addons/calico/kustomization.yaml
@@ -8,3 +8,9 @@ patches:
 - path: patches/calico-node.yaml
   target:
     kind: DaemonSet
+- target:
+    version: v1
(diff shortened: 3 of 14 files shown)


mboersma (approved):
/lgtm

Nice detective work @jsturtevant!

templates/addons/calico/patches/remove-aws-reference.yaml
```diff
@@ -0,0 +1,5 @@
+# Drops the FELIX_AWSSRCDSTCHECK field
+# json patch doesn't allow for dropping fields in array by name
+# https://github.com/projectcalico/calico/issues/5101
+- op: remove
+  path: "/spec/template/spec/containers/0/env/12"
```
jackfrancis, line 5: Is there any way to identify this by the key (the name) rather than the array index number? This is a pretty unreliable contract IMO.

jsturtevant: no [json patch](https://datatracker.ietf.org/doc/html/rfc6902#appendix-A.4) doesn't allow by name, only by index.  

> The alternative is to copy all the variables and replace the env array completely. As we plan to upgrade very quickly to the calico 3.22 release which doesn't have this issue and this is a quick fix to unblock conformance tests but happy to go other way.

It's not ideal but I also don't see this changing often.  I am happy to switch to coping all the envs to the patch and doing a full replace.

jackfrancis: Comments look good, and since this is pinned to a specific calico spec, this will work for the time being.

jackfrancis (approved):
/lgtm

devigned (approved):
/lgtm

jsturtevant:
This fixes the failing conformance tests at https://testgrid.k8s.io/provider-azure-master-signal#capz-conformance
/assign @shysank

PR #2065, opened 2022-02-09
:sparkling: Set InternalDNS address on machines

alvaroaleman:
Right now we never set the InternalDNS address for machines, which some
consumers expect to be there. This change fixes that.
The approach is copied from capa:
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/cf6fea721385d490b570b4cf119defea705b3320/controllers/awscluster_controller.go#L331-L377

 

 
**What type of PR is this?**

/kind feature

Fixes #

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Azuremachines .status.addresses now contains the InternalDNS address
```

### DIFF
--- a/azure/services/virtualmachines/virtualmachines.go
+++ b/azure/services/virtualmachines/virtualmachines.go
@@ -128,7 +128,12 @@ func (s *Service) getAddresses(ctx context.Context, vm compute.VirtualMachine, r
 	ctx, _, done := tele.StartSpanWithLogger(ctx, "virtualmachines.Service.getAddresses")
 	defer done()
 
-	var addresses []corev1.NodeAddress
+	addresses := []corev1.NodeAddress{
+		{
+			Type:    corev1.NodeInternalDNS,
+			Address: to.String(vm.Name),
+		},
+	}
 	if vm.NetworkProfile.NetworkInterfaces == nil {
 		return addresses, nil
 	}
--- a/azure/services/virtualmachines/virtualmachines_test.go
+++ b/azure/services/virtualmachines/virtualmachines_test.go
@@ -53,7 +53,8 @@ var (
 		BootstrapData:     "fake data",
 	}
 	fakeExistingVM = compute.VirtualMachine{
-		ID: to.StringPtr("test-vm-id"),
+		ID:   to.StringPtr("test-vm-id"),
+		Name: to.StringPtr("test-vm-name"),
 		VirtualMachineProperties: &compute.VirtualMachineProperties{
 			ProvisioningState: to.StringPtr("Succeeded"),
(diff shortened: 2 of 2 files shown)


alvaroaleman:
timeout waiting for controlplane nodes
/retest

azure/services/virtualmachines/virtualmachines.go
```diff
@@ -128,7 +128,10 @@ func (s *Service) getAddresses(ctx context.Context, vm compute.VirtualMachine, r
 	ctx, _, done := tele.StartSpanWithLogger(ctx, "virtualmachines.Service.getAddresses")
 	defer done()
 
-	var addresses []corev1.NodeAddress
+	addresses := []corev1.NodeAddress{{
+		Type:    corev1.NodeInternalDNS,
+		Address: *vm.Name,
```
CecileRobertMichon, line 133:
```suggestion
		Address: to.String(vm.Name),
```

azure/services/virtualmachines/virtualmachines.go
```diff
@@ -128,7 +128,10 @@ func (s *Service) getAddresses(ctx context.Context, vm compute.VirtualMachine, r
 	ctx, _, done := tele.StartSpanWithLogger(ctx, "virtualmachines.Service.getAddresses")
 	defer done()
 
-	var addresses []corev1.NodeAddress
+	addresses := []corev1.NodeAddress{{
```
CecileRobertMichon, line 131: nit: missing newline after the first {

azure/services/virtualmachines/virtualmachines.go
CecileRobertMichon, line 133: @jsturtevant is this also accurate for Windows VMs?

alvaroaleman: I would think that the DNS setup inside the vnet is the same regardless of the OS of the instance?

jsturtevant: the dns works the same but we do some funny things with the [windows name](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/b97fbf54f0b919e42f3c21eba7f0f12897623883/azure/scope/machine.go#L349-L352).  

This scope name is the name that gets assigned to the Azure VM: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/b97fbf54f0b919e42f3c21eba7f0f12897623883/azure/services/virtualmachines/virtualmachines.go#L78-L80  

so we are good.

CecileRobertMichon (commented):
/approve

Issue #1868, opened 2021-11-16
Only show relevant conditions for the AzureCluster / AzureMachine


CecileRobertMichon:
When building a cluster, the user should only see conditions that apply to that specific cluster in the Status and as part of summary. For example, if the cluster has no vnet peerings, it should not show a "vnet peerings ready" condition.

This is related to https://github.com/kubernetes-sigs/cluster-api/issues/5624.

Two ways we can go about achieving this:

1) Remove the WithConditions() option in SetSummary https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/scope/cluster.go#L599. This will ensure the controller only uses conditions that are set to set Ready status.
Pros: minimal code changes required, no changes required if we add or remove services later
Cons: external controllers setting conditions on the object may interfere, some glitches are possible where the status temporarily becomes Ready even though not all resources are ready if all the set conditions are true and some conditions haven't been set yet.

2) Add conditional logic to determine the set of Conditions needed based on the length of the different resource specs (ie. `if len(spec) != 0 { // append condition to list of conditions}`, sort of similar to what CAPA is doing here: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/461ef72c3b647e729298ec5507c3387dabf7b015/pkg/cloud/scope/cluster.go#L209-L233
Pros: well defined set of conditions, Ready status will always be correct
Cons: more logic required, logic will need to evolve whenever we add/remove services

In both cases, we will also need to change services to only set update the Status if the spec has > 0 resources https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/services/vnetpeerings/vnetpeerings.go#L75

Jont828:
Sure, I'd be happy to take a look!

CecileRobertMichon:
@Jont828 could we prioritize this one so we can merge it before releasing v1.2 which includes a lot of new async services? Let me know if you need any help.

PR #2066, opened 2022-02-09
:bug: Skip InboundNatRule reconciliation if no LB is configured

alvaroaleman:
Clusters might be externally managed in which case the apiserver
endpoint might be in a different Azure account or on a different
platform altogether. In this case, there are no inboundnatrules for the
LB to reconcile, so skip doing that if the LB name is empty.

 

 
**What type of PR is this?**

/kind bug

Fixes #

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Release note**:

```release-note
If a cluster has no APIServer LoadBalancer configured, the InboundNATRule reconciliaton for machines will be skippped
```

### DIFF
--- a/azure/services/inboundnatrules/inboundnatrules.go
+++ b/azure/services/inboundnatrules/inboundnatrules.go
@@ -56,8 +56,18 @@ func New(scope InboundNatScope) *Service {
 
 // Reconcile gets/creates/updates an inbound NAT rule.
 func (s *Service) Reconcile(ctx context.Context) error {
-	ctx, _, done := tele.StartSpanWithLogger(ctx, "inboundnatrules.Service.Reconcile")
+	ctx, log, done := tele.StartSpanWithLogger(ctx, "inboundnatrules.Service.Reconcile")
 	defer done()
+
+	// Externally managed clusters might not have an LB
+	if s.Scope.APIServerLBName() == "" {
+		log.V(4).Info("Skipping InboundNatRule reconciliation as the cluster has no LB configured")
+		// Until https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1868 is
+		// resolved, this needs to be set for the machine to be able to reach the ready condition:
+		// https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2066#discussion_r806150004
+		s.Scope.UpdatePutStatus(infrav1.InboundNATRulesReadyCondition, serviceName, nil)
+		return nil
+	}
 
 	ctx, cancel := context.WithTimeout(ctx, reconciler.DefaultAzureServiceReconcileTimeout)
 	defer cancel()
--- a/azure/services/inboundnatrules/inboundnatrules_test.go
+++ b/azure/services/inboundnatrules/inboundnatrules_test.go
@@ -125,6 +125,16 @@ func TestReconcileInboundNATRule(t *testing.T) {
 			},
 		},
 		{
+			name:          "No LB, Nat rule reconciliation is skipped",
+			expectedError: "",
+			expect: func(s *mock_inboundnatrules.MockInboundNatScopeMockRecorder,
+				m *mock_inboundnatrules.MockclientMockRecorder,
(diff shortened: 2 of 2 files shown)


azure/services/inboundnatrules/inboundnatrules.go
```diff
@@ -56,9 +56,14 @@ func New(scope InboundNatScope) *Service {
 
 // Reconcile gets/creates/updates an inbound NAT rule.
 func (s *Service) Reconcile(ctx context.Context) error {
-	ctx, _, done := tele.StartSpanWithLogger(ctx, "inboundnatrules.Service.Reconcile")
+	ctx, log, done := tele.StartSpanWithLogger(ctx, "inboundnatrules.Service.Reconcile")
 	defer done()
 
+	if s.Scope.APIServerLBName() == "" {
```
jackfrancis, line 62: How is it possible to create a cluster without a apiserver load balancer? In other words, how does the above condition (empty string value for `AzureCluster.Spec.NetworkSpec.APIServerLB.Name`) ever occur?

alvaroaleman: @jackfrancis as mentioned in the PR description, this can happen if the cluster is managed externally, see https://capz.sigs.k8s.io/topics/externally-managed-azure-infrastructure.html

jackfrancis: Thanks. The code lgtm. I am a bit confused about this line in the docs:

> If the AzureCluster resource includes a “cluster.x-k8s.io/managed-by” annotation then the [controller will skip any reconciliation](https://cluster-api.sigs.k8s.io/developer/providers/cluster-infrastructure.html#normal-resource).

If reconciliation is skipped, then why is this change needed?

alvaroaleman: The cluster controllers skips reconciling it, however the azuremachine-related controllers still do: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/a42074cdc1e82bb093ec2c79d99924bf5ef726dc/controllers/azuremachine_reconciler.go#L90-L92

We want to use capz to only manage worker machines, which is how we ran into this.

jackfrancis: cc @Jont828

jackfrancis: @Jont828 we were talking offline yesterday about your wanting this scenario to `s.Scope.UpdatePutStatus` even if it's a no-op (no LB present). Can you explain why?

CecileRobertMichon: Agree that setting the condition when there is no resource is confusing which is why we have #1868 open to address that. However in the meantime we need to set all conditions that are defined in https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/scope/machine.go#L550 to true otherwise the AzureMachine won't reach Ready condition as summary will include false conditions. However in this case it looks like `InboundNATRulesReadyCondition` is missing from the list of summary conditions (potentially a bug?).

alvaroaleman: I see, updated the change to include `UpdatePutStatus` even when there is no LB

azure/services/inboundnatrules/inboundnatrules.go
```diff
@@ -56,9 +56,18 @@ func New(scope InboundNatScope) *Service {
...
```
CecileRobertMichon, line 62: could you please also add a comment that explains this is to support the externally managed use case? I could see someone else reading the code later and having the exact same reaction as Jack did above

alvaroaleman: Added a comment, met me know what you think

alvaroaleman:
@CecileRobertMichon who should I ping for final approval?

alvaroaleman:
Thank you all for taking the time to review this, I appreciate it :)

Issue #2052, opened 2022-02-04
E2E k8s image resolution should skip checking the image for k8s prereleases


CecileRobertMichon:
/kind bug

[Before submitting an issue, have you checked the [Troubleshooting Guide](https://capz.sigs.k8s.io/topics/troubleshooting.html)?]

https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/test/e2e/e2e_suite_test.go#L511 checks that if the k8s version is a semver (eg. v1.21.1), we make sure we have a published OS image for it. However, this would fail for pre-releases of k8s which are used to test conformance in k8s PRs. We should not fail if the k8s version is a prerelease such as `v1.24.0-alpha.2.54+469c4c4a30ab57`.

In order to do this, we should add a condition to the line above to only check the image is available if the version is a semver AND not a prerelease tag.

CecileRobertMichon:
thanks for picking this up @sayantani11, let me know if you need help

/cc @chewong

CecileRobertMichon:
@sayantani11 this is now impacting upstream testing, if you haven't had a chance to work on it yet please let us know so we can pick it up.

/priority-critical-urgent
/cc @jackfrancis @Jont828

PR #2069, opened 2022-02-09
[release-1.1] e2e: Use the intersection of images available for windows and ubuntu to determine latest patch

jackfrancis:
This is a manual cherry-pick of https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1995

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/common.go
+++ b/test/e2e/common.go
@@ -67,6 +67,7 @@ const (
 	ManagedClustersResourceType    = "managedClusters"
 	capiImagePublisher             = "cncf-upstream"
 	capiOfferName                  = "capi"
+	capiWindowsOfferName           = "capi-windows"
 )
 
 func Byf(format string, a ...interface{}) {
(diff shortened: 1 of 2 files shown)


CecileRobertMichon:
@jackfrancis do we still want to hold this? Even if it isn't enough by itself and we also need #1879 to fully fix release-1.1 for presubmit jobs, we might still want to merge it to prevent running into the bug it fixes when image publishing is in progress.

jackfrancis:
> @jackfrancis do we still want to hold this? Even if it isn't enough by itself and we also need #1879 to fully fix release-1.1 for presubmit jobs, we might still want to merge it to prevent running into the bug it fixes when image publishing is in progress.

I held it because I didn't fully understand what the ramifications of merging it into `release-1.1` would be, after losing confidence that doing so would actually fix the blocked upstream k/k PRs (which we later determined will be fixed by #1879).

If we're confident this is a useful cherry-pick let's remove the hold!

CecileRobertMichon:
/hold cancel

This + https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/2052 should help fix upstream tests on release-1.1

PR #2085, opened 2022-02-15
AzureMachinePoolMachine: track concurrent deletes

jackfrancis:
**What type of PR is this?**

/kind bug

This PR improves the way that AzureMachinePoolMachines track deletes in the following way:

If capz observes that Azure is in an active state of deleting a VM in a VMSS, then we ensure that the corresponding AzureMachinePoolMachine is also deleted.
  - If we have *already* marked the AzureMachinePoolMachine for deletion (indicated by a non-nil `DeletionTimestamp` value in the `AzureMachinePoolMachine` data representation) then don't need to "re-delete" that AzureMachinePoolMachine

Furthermore, when taking an accounting of `Ready` machines that are candidates for deletion when the `AzureMachinePool` replica count is too high, we consider any machines that have been already marked for deletion as fulfilling the desired machine deletion goal.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
AzureMachinePoolMachine: track concurrent deletes
```

### DIFF
--- a/azure/scope/strategies/machinepool_deployments/machinepool_deployment_strategy.go
+++ b/azure/scope/strategies/machinepool_deployments/machinepool_deployment_strategy.go
@@ -149,6 +149,7 @@ func (rollingUpdateStrategy rollingUpdateStrategy) SelectMachinesToDelete(ctx co
 		"disruptionBudget", disruptionBudget,
 		"machinesWithoutTheLatestModel", len(machinesWithoutLatestModel),
 		"failedMachines", len(failedMachines),
+		"deletingMachines", len(deletingMachines),
 	)
 
 	// if we have failed or deleting machines, remove them
@@ -157,7 +158,7 @@ func (rollingUpdateStrategy rollingUpdateStrategy) SelectMachinesToDelete(ctx co
 		return append(failedMachines, deletingMachines...), nil
 	}
 
-	// if we have deleting machines, remove them
+	// if we have failed machines, remove them
 	if len(failedMachines) > 0 {
 		log.Info("failed machines", "desiredReplicaCount", desiredReplicaCount, "maxUnavailable", maxUnavailable, "failedMachines", getProviderIDs(failedMachines))
 		return failedMachines, nil
@@ -233,11 +234,16 @@ func getFailedMachines(machinesByProviderID map[string]infrav1exp.AzureMachinePo
 	return machines
 }
 
+// getDeletingMachines is responsible for identifying machines whose VMs are in an active state of deletion
+// but whose corresponding AzureMachinePoolMachine resource has not yet been marked for deletion.
 func getDeletingMachines(machinesByProviderID map[string]infrav1exp.AzureMachinePoolMachine) []infrav1exp.AzureMachinePoolMachine {
 	var machines []infrav1exp.AzureMachinePoolMachine
 	for _, v := range machinesByProviderID {
-		// ready status, with provisioning state Succeeded, and not marked for delete
-		if v.Status.ProvisioningState != nil && *v.Status.ProvisioningState == infrav1.Deleting {
+		if v.Status.ProvisioningState != nil &&
+			// provisioning state is Deleting
+			*v.Status.ProvisioningState == infrav1.Deleting &&
+			// Ensure that the machine has not already been marked for deletion
+			v.DeletionTimestamp.IsZero() {
 			machines = append(machines, v)
 		}
 	}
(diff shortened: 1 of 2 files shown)


jackfrancis:
cc @devigned

azure/scope/strategies/machinepool_deployments/machinepool_deployment_strategy.go
```diff
@@ -233,12 +234,17 @@ func getFailedMachines(machinesByProviderID map[string]infrav1exp.AzureMachinePo
...
+// but whose corresponding AzureMachinePoolMachine resource has not yet been marked for deletion.
 func getDeletingMachines(machinesByProviderID map[string]infrav1exp.AzureMachinePoolMachine) []infrav1exp.AzureMachinePoolMachine {
 	var machines []infrav1exp.AzureMachinePoolMachine
 	for _, v := range machinesByProviderID {
-		// ready status, with provisioning state Succeeded, and not marked for delete
+		// provisioning state is Deleting
 		if v.Status.ProvisioningState != nil && *v.Status.ProvisioningState == infrav1.Deleting {
-			machines = append(machines, v)
+			// If we've already marked the DeletionTimestamp then we have no work to do
+			if v.DeletionTimestamp.IsZero() {
```
devigned, line 245:
```suggestion
			if v.DeletionTimestamp.IsZero() || (v.Status.ProvisioningState != nil && *v.Status.ProvisioningState == infrav1.Deleting) {
```

I was thinking about this last night and in the case where the AMPM is not marked for deletion, but the AMPM last reconcile observed the VM is being deleted in Azure, this would ensure AMP reconcile loop doesn't delete an extra machine. 

This is an edge case optimization, but I think it might be worth the change.

devigned (changes requested):
Thank you for fixing this issue! I have just a bit of feedback about selecting a deleting machine.

azure/scope/strategies/machinepool_deployments/machinepool_deployment_strategy.go
jackfrancis: We shouldn't get to this point in the flow if we have any "Azure is deleting a VM but capz doesn't know about it yet", because we track those VMs when we call `getDeletingMachines()` earlier; if `getDeletingMachines()` returns a non-empty array we will short-circuit before we introspect the result from `readyMachines()`.

However...

The order of evaluation could change in the future and so to make this more reliable going forward we should probably take this into account when preparing our set of Ready machines. I'll make this change.

azure/scope/strategies/machinepool_deployments/machinepool_deployment_strategy.go
```diff
@@ -233,12 +234,17 @@ func getFailedMachines(machinesByProviderID map[string]infrav1exp.AzureMachinePo
...
 }
 
+// getDeletingMachines is responsible for identifying machines whose VMs are in an active state of deletion
+// but whose corresponding AzureMachinePoolMachine resource has not yet been marked for deletion.
 func getDeletingMachines(machinesByProviderID map[string]infrav1exp.AzureMachinePoolMachine) []infrav1exp.AzureMachinePoolMachine {
 	var machines []infrav1exp.AzureMachinePoolMachine
 	for _, v := range machinesByProviderID {
-		// ready status, with provisioning state Succeeded, and not marked for delete
+		// provisioning state is Deleting
 		if v.Status.ProvisioningState != nil && *v.Status.ProvisioningState == infrav1.Deleting {
```
devigned, line 243: I think this line should be deleted.

azure/scope/strategies/machinepool_deployments/machinepool_deployment_strategy.go
devigned: I guess what I was saying is that I'd prefer the if statement combined like what you did with ready machines.

azure/scope/strategies/machinepool_deployments/machinepool_deployment_strategy.go
jackfrancis: Which line?

devigned: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2085#discussion_r807316816

Consider combining the if statement.

azure/scope/strategies/machinepool_deployments/machinepool_deployment_strategy.go
jackfrancis: Sorry, I was totally confused here.

azure/scope/strategies/machinepool_deployments/machinepool_deployment_strategy.go
```diff
@@ -233,11 +234,16 @@ func getFailedMachines(machinesByProviderID map[string]infrav1exp.AzureMachinePo
...
 func getDeletingMachines(machinesByProviderID map[string]infrav1exp.AzureMachinePoolMachine) []infrav1exp.AzureMachinePoolMachine {
 	var machines []infrav1exp.AzureMachinePoolMachine
 	for _, v := range machinesByProviderID {
-		// ready status, with provisioning state Succeeded, and not marked for delete
-		if v.Status.ProvisioningState != nil && *v.Status.ProvisioningState == infrav1.Deleting {
+		if v.Status.ProvisioningState != nil &&
+			// provisioning state is Deleting
+			*v.Status.ProvisioningState == infrav1.Deleting &&
+			// If we've already marked the DeletionTimestamp then we have no work to do
+			v.DeletionTimestamp.IsZero() {
```
jackfrancis, line 246: @devigned so I've combined the if statements into one here, but you seemed to suggest that this would be improved if the "Azure VM is actively being deleted" + "capz has already marked this ampm for deletion" are OR'd together (instead of AND'ed). Can you clarify that subtlety, if you think we should make it OR?

devigned (approved):
/lgtm
/approve

PR #2093, opened 2022-02-15
Only show applicable conditions for AzureClusters and AzureMachines on async services

### LINKED ISSUE #1868: Only show relevant conditions for the AzureCluster / AzureMachine

Jont828:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**: Only show applicable conditions for AzureClusters and AzureMachines. It doesn't make sense to set a condition to ready if it doesn't apply, i.e. there are no specs.

Fixes #1868

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Only show applicable conditions for AzureClusters and AzureMachines
```

### DIFF
--- a/azure/scope/cluster.go
+++ b/azure/scope/cluster.go
@@ -679,20 +679,7 @@ func (s *ClusterScope) PatchObject(ctx context.Context) error {
 	ctx, _, done := tele.StartSpanWithLogger(ctx, "scope.ClusterScope.PatchObject")
 	defer done()
 
-	conditions.SetSummary(s.AzureCluster,
-		conditions.WithConditions(
-			infrav1.ResourceGroupReadyCondition,
-			infrav1.RouteTablesReadyCondition,
-			infrav1.NetworkInfrastructureReadyCondition,
-			infrav1.VnetPeeringReadyCondition,
-			infrav1.DisksReadyCondition,
-			infrav1.NATGatewaysReadyCondition,
-			infrav1.LoadBalancersReadyCondition,
-			infrav1.BastionHostReadyCondition,
-			infrav1.VNetReadyCondition,
-			infrav1.SubnetsReadyCondition,
-		),
-	)
+	conditions.SetSummary(s.AzureCluster)
 
 	return s.patchHelper.Patch(
 		ctx,
--- a/azure/services/availabilitysets/availabilitysets.go
+++ b/azure/services/availabilitysets/availabilitysets.go
@@ -70,6 +70,7 @@ func (s *Service) Reconcile(ctx context.Context) error {
 		_, err = s.CreateResource(ctx, setSpec, serviceName)
 	} else {
 		log.V(2).Info("skip creation when no availability set spec is found")
+		return nil
 	}
 
 	s.Scope.UpdatePutStatus(infrav1.AvailabilitySetReadyCondition, serviceName, err)
@@ -85,25 +86,27 @@ func (s *Service) Delete(ctx context.Context) error {
 	defer cancel()
 
 	var resultingErr error
-	if setSpec := s.Scope.AvailabilitySetSpec(); setSpec == nil {
+	setSpec := s.Scope.AvailabilitySetSpec()
+	if setSpec == nil {
 		log.V(2).Info("skip deletion when no availability set spec is found")
+		return nil
+	}
+
+	existingSet, err := s.Get(ctx, setSpec)
+	if err != nil {
+		if !azure.ResourceNotFound(err) {
+			resultingErr = errors.Wrapf(err, "failed to get availability set %s in resource group %s", setSpec.ResourceName(), setSpec.ResourceGroupName())
+		}
 	} else {
-		existingSet, err := s.Get(ctx, setSpec)
-		if err != nil {
-			if !azure.ResourceNotFound(err) {
-				resultingErr = errors.Wrapf(err, "failed to get availability set %s in resource group %s", setSpec.ResourceName(), setSpec.ResourceGroupName())
-			}
+		availabilitySet, ok := existingSet.(compute.AvailabilitySet)
+		if !ok {
+			resultingErr = errors.Errorf("%T is not a compute.AvailabilitySet", existingSet)
 		} else {
-			availabilitySet, ok := existingSet.(compute.AvailabilitySet)
-			if !ok {
-				resultingErr = errors.Errorf("%T is not a compute.AvailabilitySet", existingSet)
+			// only delete when the availability set does not have any vms
+			if availabilitySet.AvailabilitySetProperties != nil && availabilitySet.VirtualMachines != nil && len(*availabilitySet.VirtualMachines) > 0 {
+				log.V(2).Info("skip deleting availability set with VMs", "availability set", setSpec.ResourceName())
 			} else {
-				// only delete when the availability set does not have any vms
-				if availabilitySet.AvailabilitySetProperties != nil && availabilitySet.VirtualMachines != nil && len(*availabilitySet.VirtualMachines) > 0 {
-					log.V(2).Info("skip deleting availability set with VMs", "availability set", setSpec.ResourceName())
-				} else {
-					resultingErr = s.DeleteResource(ctx, setSpec, serviceName)
-				}
+				resultingErr = s.DeleteResource(ctx, setSpec, serviceName)
 			}
 		}
 	}
--- a/azure/services/inboundnatrules/inboundnatrules.go
+++ b/azure/services/inboundnatrules/inboundnatrules.go
@@ -62,10 +62,6 @@ func (s *Service) Reconcile(ctx context.Context) error {
 	// Externally managed clusters might not have an LB
 	if s.Scope.APIServerLBName() == "" {
 		log.V(4).Info("Skipping InboundNatRule reconciliation as the cluster has no LB configured")
-		// Until https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1868 is
-		// resolved, this needs to be set for the machine to be able to reach the ready condition:
-		// https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2066#discussion_r806150004
-		s.Scope.UpdatePutStatus(infrav1.InboundNATRulesReadyCondition, serviceName, nil)
 		return nil
 	}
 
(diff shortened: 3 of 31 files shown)


azure/services/inboundnatrules/inboundnatrules.go
```diff
@@ -65,7 +65,6 @@ func (s *Service) Reconcile(ctx context.Context) error {
 		// Until https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1868 is
 		// resolved, this needs to be set for the machine to be able to reach the ready condition:
 		// https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2066#discussion_r806150004
```
CecileRobertMichon, line 67: you can remove the comment here

azure/services/virtualmachines/virtualmachines.go
```diff
@@ -76,11 +77,16 @@ func (s *Service) Reconcile(ctx context.Context) error {
...
 	vmSpec := s.Scope.VMSpec()
+	if vmSpec == nil {
+		return nil
+	}
 
 	result, err := s.CreateResource(ctx, vmSpec, serviceName)
 	s.Scope.UpdatePutStatus(infrav1.VMRunningCondition, serviceName, err)
 	// Set the DiskReady condition here since the disk gets created with the VM.
-	s.Scope.UpdatePutStatus(infrav1.DisksReadyCondition, serviceName, err)
+	if len(s.Scope.DiskSpecs()) > 0 {
```
CecileRobertMichon, line 87: I don't think we even need to check this here. Since the disk is getting created as part of the VM, it's ready if the VM is ready.

azure/scope/cluster.go
```diff
@@ -679,20 +679,7 @@ func (s *ClusterScope) PatchObject(ctx context.Context) error {
...
-	conditions.SetSummary(s.AzureCluster,
-		conditions.WithConditions(
-			infrav1.ResourceGroupReadyCondition,
-			infrav1.RouteTablesReadyCondition,
-			infrav1.NetworkInfrastructureReadyCondition,
-			infrav1.VnetPeeringReadyCondition,
-			infrav1.DisksReadyCondition,
-			infrav1.NATGatewaysReadyCondition,
-			infrav1.LoadBalancersReadyCondition,
-			infrav1.BastionHostReadyCondition,
```
CecileRobertMichon, line 691: I don't see the bastion service being modified, did that one also need to be updated?

CecileRobertMichon: looks like there are 10+3 conditions being removed here and in machine.go but 10 services getting updated

Jont828: Oh yeah I think I missed those on accident, thanks for noticing!

CecileRobertMichon:
> Should we update the put status when we return early from a service in custom vnet mode? It looks like we currently do not update the delete status in VNets if it's not managed, so we could do the same with the other services. Wdyt?

I'm on the fence on this one. On one hand I could see us mark the VNet as "ready" if it's not managed once we check it exists. But at the same time if the CAPZ controller doesn't create/manage the VNet it probably shouldn't be showing its status either. Another way to look at this is to ask who is the user for these conditions and what are their use cases. Would it matter to them to see the VNet condition when using a pre-existing VNet? What would be less surprising for a user?

@sonasingh46 @shysank thoughts?

Jont828:
On the other hand it looks like we still reconcile the subnets whether or not the VNet is managed. It might look odd to only have a condition for subnets, but it's probably not a big deal.

shysank:
> Would it matter to them to see the VNet condition when using a pre-existing VNet? What would be less surprising for a user?

I think adding Ready condition for resources not managed by capz is kind of misleading, otoh there could be external (CI) systems relying on the ready condition as a trigger. If we were to take inspiration from how "externally managed cluster" is handled, the responsibility of updating "Ready" condition is given to the external system. This is not exactly similar to the situation here as we don't do any reconciliation for externally managed cluster, but the idea of not updating Ready condition for unmanaged resources can be replicated for consistency.

Jont828:
@CecileRobertMichon One thing that's different about the subnets is that even if they're not managed, we still want to Reconcile by updating the ID and CIDR blocks in the yaml spec. In that case, the subnet ready condition would signify that the spec has been updated. Do we want to keep the ready condition in this situation?

CecileRobertMichon:
> we still want to Reconcile by updating the ID and CIDR blocks in the yaml spec

We're only doing a GET API call in this case still, no PUT call. Updating the spec/status to reflect actual state isn't truly reconciling the resource in the sense that we're not making any changes to the Azure resource. I would vote to still opt out of setting a Ready condition in that case for now to simplify things.

Jont828:
@CecileRobertMichon I'm not sure if it's simpler to handle. In the event that a subnet isn't managed but there's an error in the Reconcile logic, do we still want to update the put status so the user can see the errors?

CecileRobertMichon:
>  In the event that a subnet isn't managed but there's an error in the Reconcile logic, do we still want to update the put status so the user can see the errors?

If the subnet isn't managed and there's an error, we should return the error but not update the put status for the subnet (because we are not truly doing a "PUT"), so it's not accurate to say that the subnet failed to "create or update" and it could mislead users. 

However, we need to make sure that the overall "Ready" condition is set to False if that's the case so that the Cluster doesn't show as "Ready". One thing we could do is set the generic `NetworkInfrastructureReadyCondition` to False in the azurecluster_controller whenever Reconcile returns an error (here: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/controllers/azurecluster_controller.go#L243) so that the summary Ready condition stays "False".

azure/services/subnets/subnets.go
```diff
@@ -104,23 +112,29 @@ func (s *Service) Delete(ctx context.Context) error {
...
+	for _, subnetSpec := range specs {
 		if err := s.DeleteResource(ctx, subnetSpec, serviceName); err != nil {
 			if !azure.IsOperationNotDoneError(err) || result == nil {
 				result = err
 			}
 		}
 	}
-	s.Scope.UpdateDeleteStatus(infrav1.SubnetsReadyCondition, serviceName, result)
+
+	if s.Scope.IsVnetManaged() {
```
CecileRobertMichon, line 135: we don't actually need this check since we're already returning early on line 115 when the vnet isn't managed

azure/services/virtualnetworks/virtualnetworks.go
```diff
@@ -66,6 +66,9 @@ func (s *Service) Reconcile(ctx context.Context) error {
 	defer cancel()
 
 	vnetSpec := s.Scope.VNetSpec()
+	if vnetSpec == nil {
+		return nil
+	}
 
 	result, err := s.CreateResource(ctx, vnetSpec, serviceName)
 	s.Scope.UpdatePutStatus(infrav1.VNetReadyCondition, serviceName, err)
```
CecileRobertMichon, line 74: looks like we'd still be updating the status here when the vnet is unmanaged

CecileRobertMichon: One suggestion: 

 to avoid the extra GET by calling IsManaged below, we should be able to use the information we have below in Tags to determine if the vnet is managed:

on line 83:
```
managed := tags.HasOwned(s.Scope.ClusterName()), nil
```

then below on line 90:

```
if managed {
    s.Scope.UpdatePutStatus(infrav1.VNetReadyCondition, serviceName, err)
}
return err
```

azure/services/virtualnetworks/virtualnetworks_test.go
```diff
@@ -67,6 +67,13 @@ func TestReconcileVnet(t *testing.T) {
 		expectedError string
 		expect        func(s *mock_virtualnetworks.MockVNetScopeMockRecorder, m *mock_async.MockGetterMockRecorder, r *mock_async.MockReconcilerMockRecorder)
 	}{
+		{
```
CecileRobertMichon, line 70: let's also add a test case for the scenario above (vnet is not managed)

azure/services/availabilitysets/availabilitysets.go
```diff
@@ -85,25 +86,27 @@ func (s *Service) Delete(ctx context.Context) error {
 	defer cancel()
 
 	var resultingErr error
-	if setSpec := s.Scope.AvailabilitySetSpec(); setSpec == nil {
+	setSpec := s.Scope.AvailabilitySetSpec()
+	if setSpec == nil {
```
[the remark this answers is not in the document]
Jont828: I believe the previous statement `if setSpec := s.Scope.AvailabilitySetSpec(); setSpec == nil {` would make `setSpec` a local variable to the if block, and I had to break up the if block b/c of a linting error.

azure/services/virtualnetworks/virtualnetworks.go
Jont828: Sure I'll remove that status update in that case. And wdym by an extra GET? IsVnetManaged() just does the same thing right? 
``` // IsVnetManaged returns true if the vnet is managed.
func (s *ClusterScope) IsVnetManaged() bool {
	return s.Vnet().ID == "" || s.Vnet().Tags.HasOwned(s.ClusterName())
}
```

CecileRobertMichon: `IsVnetManaged` would work too as long as it's run after line 82 so we've already set the tags

Jont828:
@CecileRobertMichon I've pushed some more changes, does this look ready to merge?

CecileRobertMichon (approved):
/lgtm
/assign @shysank

PR #2095, opened 2022-02-16
enable/disable AKS cluster add-ons

### LINKED ISSUE #1145: Add AKS addon support
/kind feature

Add support for installing/managing AKS addons as part of the control plane reconciliation.

michalno1:
**What type of PR is this?**
/kind feature

This PR allows enabling/disabling AKS cluster add-ons.

Fixes #1145 

**Release note**:

```release-note
Support enabling and disabling AKS cluster add-ons.
```


azure/scope/managedcontrolplane.go
```diff
@@ -470,6 +470,16 @@ func (s *ManagedControlPlaneScope) ManagedClusterSpec() (azure.ManagedClusterSpe
 		}
 	}
 
+	if s.ControlPlane.Spec.AddonProfiles != nil {
```
zmalik, line 473: add a unit test for this?

exp/api/v1beta1/azuremanagedcontrolplane_types.go
```diff
@@ -93,10 +93,14 @@ type AzureManagedControlPlaneSpec struct {
 	// +optional
 	IdentityRef *corev1.ObjectReference `json:"identityRef,omitempty"`
 
-	// AadProfile is Azure Active Directory configuration to integrate with AKS for aad authentication.
+	// AADProfile is Azure Active Directory configuration to integrate with AKS for aad authentication.
```
jackfrancis, line 96: I know this is nitpicky but let's not make his comment change here? (I assume your IDE did it for you :))

michalno1: Ok, removed.

azure/services/managedclusters/managedclusters.go
```diff
@@ -99,6 +99,26 @@ func computeDiffOfNormalizedClusters(managedCluster containerservice.ManagedClus
 		}
 	}
 
+	if managedCluster.AddonProfiles != nil {
+		propertiesNormalized.AddonProfiles = map[string]*containerservice.ManagedClusterAddonProfile{}
+		for k, v := range managedCluster.AddonProfiles {
+			propertiesNormalized.AddonProfiles[k] = &containerservice.ManagedClusterAddonProfile{
+				Enabled: v.Enabled,
```
jackfrancis, line 106: Where is the `Name` field?

azure/services/managedclusters/managedclusters.go
```diff
@@ -99,6 +99,26 @@ func computeDiffOfNormalizedClusters(managedCluster containerservice.ManagedClus
...
+				Config:  v.Config,
+			}
+		}
+	}
+
+	if existingMC.AddonProfiles != nil {
+		existingMCPropertiesNormalized.AddonProfiles = map[string]*containerservice.ManagedClusterAddonProfile{}
+		for k, v := range existingMC.AddonProfiles {
+			existingMCPropertiesNormalized.AddonProfiles[k] = &containerservice.ManagedClusterAddonProfile{
+				Enabled: v.Enabled,
```
jackfrancis, line 116: same question as above

azure/services/managedclusters/managedclusters.go
michalno1: The key in the map is the `Name`.

jackfrancis: yep, thanks :)

jackfrancis:
code lgtm

Can we add some test coverage to the AKS spec we run tests against at `templates/flavors/aks-multi-tenancy/cluster-template.yaml`?

Is there a "simple" addon we can enable in the `AzureManagedControlPlane` resource spec to test this new functionality that won't change any other functional aspects of the cluster?

After making that change run `make generate-flavors` to render the final template for prow CI to consume and be able to test.

Thanks!

CecileRobertMichon:
> Can we add some test coverage to the AKS spec we run tests against at templates/flavors/aks-multi-tenancy/cluster-template.yaml?

Heads up I'm removing the multi-tenancy flavor in https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2143 to replace it with the regular "aks" flavor since all templates have multi-tenancy now, it's redundant to have a separate "multi-tenancy" flavor. Going forward we'll be testing with https://github.com/kubernetes-sigs/cluster-api-provider-azure/tree/ab1aec6f47646a1f03839b4a4e1b3047cf3aea42/templates/test/ci/prow-aks.

michalno1:
@jackfrancis From the [addons that can be used](https://github.com/Azure/azure-cli/blob/78c6e2af33b0fa423830d806bc7a6574c2bbd895/src/azure-cli/azure/cli/command_modules/acs/_consts.py#L88) I think a good one for testing is `azurepolicy`- it shouldn't interfere with functional aspects of the cluster. Another good one would be `monitoring` but it does require creating a Log Analytics Workspace first. I'll use `azurepolicy` for now.

@jackfrancis @CecileRobertMichon It seems to me existing e2e tests just check if everything started successfully. Shouldn't they also somehow verify template specific settings are really in effect?

jackfrancis:
@michalno1 I have a slight preference for *not* testing functional outcomes and instead relying upon the AKS API to do that if it accepts our requests. The only reason is that it would add test maintenance for outcomes outside of our control (i.e., AKS operational behaviors).

But it's not a simple issue, as the project would benefit having more coverage signal to add customer confidence that "this particular way of using AKS will have the desired effect".

templates/cluster-template-aks-multi-tenancy.yaml
```diff
@@ -23,6 +23,9 @@ metadata:
   name: ${CLUSTER_NAME}
   namespace: default
 spec:
+  addonProfiles:
```
CecileRobertMichon, line 26: Ideally we shouldn't enable features for tests directly in the flavors as those are also exposed as examples for users, but we can add patches in the test/ directory for that sort of stuff

since this is likely to merge before #2143 and I'm removing this template anyways, let's keep this for now and I'll do some cleanup in my PR.

jackfrancis (approved):
/lgtm

PR #2098, opened 2022-02-16
Add control-plane role taints and tolerations

CecileRobertMichon:
**What type of PR is this?**
/kind failing-test

**What this PR does / why we need it**: 1.24 will soon remove the "master" role and our conformance job at k8s HEAD started failing after this set of changes: https://github.com/kubernetes/kubernetes/compare/a37b6fc8b488a0...400b1dea170d9d

Adding "control-plane" tolerations and taints in a couple of places where we were still using "master" and changing the makefile to look for the "control-plane" role in workload cluster creation.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Add control-plane role taints and tolerations
```

### DIFF
--- a/test/e2e/conformance_test.go
+++ b/test/e2e/conformance_test.go
@@ -1,3 +1,4 @@
+//go:build e2e
 // +build e2e
 
 /*
@@ -184,12 +185,22 @@ var _ = Describe("Conformance Tests", func() {
 			}
 
 			noScheduleTaint := &corev1.Taint{
-				Key:    "node-role.kubernetes.io/master",
+				Key:    "node-role.kubernetes.io/control-plane",
 				Value:  "",
 				Effect: "NoSchedule",
 			}
 
-			err := node.TaintNode(workloadProxy.GetClientSet(), options, noScheduleTaint)
+			if v, err := semver.ParseTolerant(kubernetesVersion); err == nil {
+				if v.LT(semver.MustParse("1.24.0-alpha.0.0")) {
+					noScheduleTaint = &corev1.Taint{
+						Key:    "node-role.kubernetes.io/master",
+						Value:  "",
+						Effect: "NoSchedule",
+					}
+				}
+			}
+
+			err = node.TaintNode(workloadProxy.GetClientSet(), options, noScheduleTaint)
 			Expect(err).NotTo(HaveOccurred())
 
 			// Windows requires a repo-list because some images are not in k8s gcr
--- a/Makefile
+++ b/Makefile
@@ -248,7 +248,7 @@ create-management-cluster: $(KUSTOMIZE) $(ENVSUBST) ## Create a management clust
 	./hack/create-identity-secret.sh
 
 	# Deploy CAPI
-	curl --retry $(CURL_RETRIES) -sSL https://github.com/kubernetes-sigs/cluster-api/releases/download/v1.1.0/cluster-api-components.yaml | $(ENVSUBST) | kubectl apply -f -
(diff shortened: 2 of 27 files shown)


test/e2e/conformance_test.go
```diff
@@ -183,13 +184,23 @@ var _ = Describe("Conformance Tests", func() {
...
-			noScheduleTaint := &corev1.Taint{
+			// This is kept to support versions of k8s < 1.20.
+			deprecatedNoScheduleTaint := &corev1.Taint{
 				Key:    "node-role.kubernetes.io/master",
 				Value:  "",
 				Effect: "NoSchedule",
 			}
 
-			err := node.TaintNode(workloadProxy.GetClientSet(), options, noScheduleTaint)
+			noScheduleTaint := &corev1.Taint{
```
jsturtevant, line 194: I think we either need to only apply this in 1.24+ or update e2e.test parameter `--NonblockingTaints` to include `node-role.kubernetes.io/control-plane` otherwise jobs < 1.24 will timeout because this new label will be applied.

https://github.com/kubernetes/kubernetes/compare/a37b6fc8b488a0...400b1dea170d9d#diff-c017e0c5e68c14cb966137bc87de3d3ab8475dc25f9b5fdb47a2b8174830054f

jsturtevant: since the other templates are now adding the flag, my guess is we need to add it to the test.e2e.

CecileRobertMichon: where do I add the `--NonblockingTaints` flag? 

I changed it to only apply the new taint for 1.24+, would that be enough?

jsturtevant: The viper config files pass the flags to the `e2e.test` binary: https://github.com/kubernetes-sigs/cluster-api-provider-azure/tree/main/test/e2e/data/kubetest 

see https://github.com/kubernetes-sigs/cluster-api/blob/dd3c2515c3edab63ab08b506c612932ecf3b08f8/test/framework/kubetest/run.go#L169



> I changed it to only apply the new taint for 1.24+, would that be enough?

I can't remember the behavior of the tolerations added to the addons.  If they schedule with those new tolerations added, it would be I believe.

CecileRobertMichon: looks like conformance is still failing, but I found out the calico deployment also has a `node-role.kubernetes.io/master` toleration so that might be the issue. Going to try adding a patch until we can update calico.

CecileRobertMichon:
/hold

testing CAPI nightly build, do not merge until v1.1.2 release is cut

CecileRobertMichon:
/hold cancel

updated to v1.1.2

jackfrancis:
@CecileRobertMichon I still see lots of `node-role.kubernetes.io/master` references in cluster templates. Are we not able to remove those?

CecileRobertMichon:
>  I still see lots of node-role.kubernetes.io/master references in cluster templates. Are we not able to remove those?

for now keeping both side by side to support k8s < 1.20 since the control-plane node role was introduced in 1.20. We're still testing upgrade from 1.19 [here](https://testgrid.k8s.io/sig-cluster-lifecycle-cluster-api-provider-azure#capi-periodic-upgrade-main-1-19-1-20) and since we are probably going to need to backport this change to fix upstream tests wanted to be cautious. Let's to follow up with a PR just for the main branch to remove 1.19 tests + all master refereces.

jackfrancis (approved):
/lgtm

PR #2100, opened 2022-02-16
test cluster flavors: required, optional, experimental

jackfrancis:
**What type of PR is this?**

/kind cleanup

This PR generalizes three classifications of cluster flavors for CI testing:

1. required
2. optional
3. experimental

And we've added corresponding ginkgo context strings surrounding the existing cluster tests to identify them. These new context string identifiers can be used by test-infra and other E2E tooling to more resiliently define "types" of test jobs. Rather than manually maintaing sets of "focus" or "skip" inputs as tests evolve, we will simply classify new cluster tests under one of the above types, and existing test jobs will keep working.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/azure_test.go
+++ b/test/e2e/azure_test.go
@@ -141,7 +141,7 @@ var _ = Describe("Workload cluster creation", func() {
 	})
 
 	if os.Getenv("LOCAL_ONLY") != "true" {
-		Context("Creating a private cluster", func() {
+		Context("Creating a private cluster [REQUIRED]", func() {
 			It("Creates a public management cluster in a custom vnet", func() {
 				clusterName = getClusterName(clusterNamePrefix, "public-custom-vnet")
 				Context("Creating a custom virtual network", func() {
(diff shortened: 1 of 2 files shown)


jackfrancis:
/hold

until https://github.com/kubernetes/test-infra/pull/25307 merges

test/e2e/azure_test.go
```diff
@@ -142,59 +142,61 @@ var _ = Describe("Workload cluster creation", func() {
...
-						ClusterName:              clusterName,
-						KubernetesVersion:        e2eConfig.GetVariable(capi_e2e.KubernetesVersion),
-						ControlPlaneMachineCount: pointer.Int64Ptr(1),
-						WorkerMachineCount:       pointer.Int64Ptr(1),
-					},
-					WaitForClusterIntervals:      e2eConfig.GetIntervals(specName, "wait-cluster"),
-					WaitForControlPlaneIntervals: e2eConfig.GetIntervals(specName, "wait-control-plane"),
-					WaitForMachineDeployments:    e2eConfig.GetIntervals(specName, "wait-worker-nodes"),
-				}, result)
+			Context("Required cluster flavor test", func() {
```
CecileRobertMichon, line 145: Just a thought: would it be better to add a [REQUIRED] prefix or suffix to the higher-level context instead of adding a separate Ginkgo context? Right now it reads like this in the e2e output:

`Workload cluster creation Creating a private cluster Required cluster flavor test Creates a public management cluster in a custom vnet`

which kind of stumbles

@mboersma thoughts?

jackfrancis: So basically re-organize the E2E tests so that all of the required tests are children of a single parent context? (and the same for optional and experimental)

CecileRobertMichon: Without reorganizing the test, we could add the [OPTIONAL] / [REQUIRED] prefix to the existing contexts. Or, having one parent context for each category of tests would also work. Essentially:

- Option 1: 

```
Workload cluster creation 
    Creating a private cluster [REQUIRED] 
        Creates a public management cluster in a custom vnet
```

- Option 2: 

```
Required workload cluster flavor test 
    Creating a private cluster 
        Creates a public management cluster in a custom vnet
```

jackfrancis: I slightly prefer option 2, but it would be tricker to make it backwards-compatible with the existing test-infra configs, and would be harder to review, so I'll opt for option 1 (which will be even easier to review compared to what's there right now :))

jackfrancis: Updated this PR, and the test-infra PR that will operate against these changes after this one merges:

https://github.com/kubernetes/test-infra/pull/25307

test/e2e/azure_test.go
```diff
@@ -193,7 +193,7 @@ var _ = Describe("Workload cluster creation", func() {
 		fmt.Fprintf(GinkgoWriter, "INFO: skipping test requires pushing container images to external repository")
 	}
 
-	It("With 3 control-plane nodes and 2 Linux and 2 Windows worker nodes", func() {
+	It("With 3 control-plane nodes and 2 Linux and 2 Windows worker nodes [REQUIRED]", func() {
```
jsturtevant, line 196: Why is this one at `It` level and others are `Context`?

jsturtevant: Ah looking further it looks like this isn't part of this PR but was created in a different way from the other tests.  minor but might be worth following the same pattern

jsturtevant (commented):
What is the distinction between these values (Required/optional/etc)?  Looks like the updates to test-infra don't make them optional in the sense that they only run if asked.  What does optional mean here then?

test/e2e/azure_test.go
jackfrancis: Right, I'm just injecting these additional substrings at the place where the current test strings are that we use for regex matching.

jackfrancis:
> What is the distinction between these values (Required/optional/etc)? Looks like the updates to test-infra don't make them optional in the sense that they only run if asked. What does optional mean here then?

- Required == cluster tests that run against any capz PR, and merging is only allowed on a passing signal
- Optional == cluster tests that are not required for a PR merge, but which test the remaining set of "supported" capz features
- Experimental == cluster tests that validate new features that aren't yet "supported" (current we only classify the AKS features this way)

Does that make sense? @CecileRobertMichon would you want to add or clarify anything above?

jsturtevant:
makes sense

> * Optional == cluster tests that are not required for a PR merge, but which test the remaining set of "supported" capz features

so we will update the prow jobs to make those not "required" but will still run?

jackfrancis:
@jsturtevant after we merged these changes we will review and merge https://github.com/kubernetes/test-infra/pull/25307

test/e2e/azure_test.go
CecileRobertMichon: Could we add a parent Context here and inject the suffix there instead?

```
Context("Creating a highly available cluster [REQUIRED]", func() {
```

jackfrancis: Yes, done

PR #2101, opened 2022-02-16
don’t check pre-release versions against known capz images

### LINKED ISSUE #2052: E2E k8s image resolution should skip checking the image for k8s prereleases

jackfrancis:
**What type of PR is this?**

/kind bug

This PR updates our logic which checks the Kubernetes version of a cluster E2E test spec against our known set of published test capz images: if the Kubernetes version has a pre-release on it, then we know we don't have an image for it and will rely upon downstream E2E foo to properly handle that as a "build of upstream k/k bits from an non-released commit".

Fixes #2052

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/e2e_suite_test.go
+++ b/test/e2e/e2e_suite_test.go
@@ -509,8 +509,10 @@ func getLatestSkuForMinor(version string, skus semver.Versions) string {
 			}
 		}
 	} else if v, err := semver.ParseTolerant(version); err == nil {
-		// if the version is in the format "v1.21.2", we make sure we have an existing image for it.
-		Expect(skus).To(ContainElement(v), fmt.Sprintf("Provided Kubernetes version %s does not have a corresponding VM image in the capi offer", version))
+		if len(v.Pre) == 0 {
+			// if the version is in the format "v1.21.2", we make sure we have an existing image for it.
+			Expect(skus).To(ContainElement(v), fmt.Sprintf("Provided Kubernetes version %s does not have a corresponding VM image in the capi offer", version))
+		}
 	}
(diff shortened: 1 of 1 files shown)


test/e2e/e2e_suite_test.go
```diff
@@ -509,8 +509,10 @@ func getLatestSkuForMinor(version string, skus semver.Versions) string {
 			}
 		}
 	} else if v, err := semver.ParseTolerant(version); err == nil {
-		// if the version is in the format "v1.21.2", we make sure we have an existing image for it.
-		Expect(skus).To(ContainElement(v), fmt.Sprintf("Provided Kubernetes version %s does not have a corresponding VM image in the capi offer", version))
+		if len(v.Pre) > 0 {
```
CecileRobertMichon, line 512: should be the opposite no? If v.Pre is empty, then check for a valid image

jackfrancis: Yep!

🙏

Fixed.

CecileRobertMichon (commented):
/lgtm
/approve
/cherry-pick release-1.1

PR #2102, opened 2022-02-17
Support specifying golang version for CI

lzhecheng:
Sometimes we need to specify golang version instead of
changing variants in test-infra repo.

**What type of PR is this?**

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Support specifying golang version for CI.
```

### DIFF
--- a/hack/ensure-go.sh
+++ b/hack/ensure-go.sh
@@ -42,7 +42,22 @@ EOF
   fi
 }
 
-verify_go_version
+install_go_with_version() {
+  echo "Install golang ${GO_VERSION}"
+  GO_TARBALL="go${GO_VERSION}.linux-amd64.tar.gz"
+  curl -LO "https://go.dev/dl/${GO_TARBALL}"
+  if [[ -d /usr/local/go ]]; then
+    mv /usr/local/go /usr/local/go.old
+  fi
+  tar xzf "${GO_TARBALL}" -C /usr/local
+  rm "${GO_TARBALL}"
+}
+
+if [[ -n "${GO_VERSION:-}" ]]; then
+  install_go_with_version
+else
+  verify_go_version
+fi
 
 # Explicitly opt into go modules, even though we're inside a GOPATH directory
 export GO111MODULE=on


lzhecheng:
I made this PR because when I migrated some cloud-provider-azure conformance tests to capz, I found that the golang version is too low sometimes. So I think it should help that golang version can be specified.
https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cloud-provider-azure/1152/pull-cloud-provider-azure-e2e-capz-1-22/1493849716142641152/build-log.txt

hack/ensure-go.sh
```diff
@@ -42,7 +42,22 @@ EOF
   fi
 }
 
-verify_go_version
+install_go_with_version() {
+  echo "Install golang ${GO_VERSION}"
+  GO_TARBALL="go${GO_VERSION}.linux-amd64.tar.gz"
```
shysank, line 47: If we are going to do this, we should also do it for other OSes as well. I'm a bit hesitant on this because it might become a maintenance burden. wdyt about [nudging](https://go.dev/doc/manage-install) the user to install the right version instead of installing it? cc @CecileRobertMichon

lzhecheng: I understand your concern. My intention is to update go version for tests defined in test-infra, as I mentioned above. So `ensure-go.sh` seems a proper place.

lzhecheng: Currently, capz release-1.0 requires go 1.16+, while kubekins for k8s 1.20 is still 1.15.
https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/release-1.0/hack/ensure-go.sh#L34
https://github.com/kubernetes/test-infra/blob/master/images/kubekins-e2e/variants.yaml#L49
https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cloud-provider-azure/1154/pull-cloud-provider-azure-e2e-capz-1-20/1494567585662898176/build-log.txt
I have no idea how to workaround this without code change in capz. Please help if there's a solution. Thank you!

CecileRobertMichon: > while kubekins for k8s 1.20

@lzhecheng what about using a newer version of kubekins? Nothing in CAPZ requires 1.20, we could switch those jobs to 1.22 or 1.23 and continually bump the version of the kubekins image as we update the CAPZ version/release on the jobs. CAPZ go version follows CAPI's, which itself follows k8s, so we'd never require a go version that isn't available in a kubekins image yet.

lzhecheng: Thank you. I am actually working on a cloud-provider-azure job which runs on k8s 1.20. According to k8s, v1.20 is built on golang v1.15. So I think the support for k8s 1.20 + golang 1.15 is needed?
Now in CAPZ `release-1.0` the min golang version is 1.16, can I say that it doesn't support k8s 1.20? Should I use an older CAPZ version like `release-0.4`?
I feel like `minimum_go_version` should be set according the k8s version?

CecileRobertMichon: CAPZ `release-1.0` supports k8s 1.20 (https://github.com/kubernetes-sigs/cluster-api-provider-azure#kubernetes-versions), but it does require go 1.16+. The kubekins image version doesn't need to match the k8s version of the CAPZ cluster that gets built, as long as it satisfies the go version requirement. So for a 1.20 cloud-provider-azure job, you should use the latest CAPZ release branch (right now that release-1.1) + whichever kubekins image has a compatible go version installed based on https://github.com/kubernetes/test-infra/blob/master/images/kubekins-e2e/variants.yaml#L31. 

Since we don't use the kubetest, it doesn't matter what the k8s_release in the kubekins image is.

bridgetkromhout: Given that k8s 1.20 is officially EOL as of this week, we probably shouldn't rely on it, but also probably don't need to focus on testing it? https://kubernetes.io/releases/#release-v1-20

lzhecheng: Discussed offline. Now the golang version issue has been fixed with kubekins 1.22. Thank you all! Closing this PR.

PR #2111, opened 2022-02-23
Don't default RoleAssignmentName on machine templates

fiunchinho:
/kind bug

While using `SystemAssigned` identity, I can't use more than 1 replica for my control plane. Or more than 1 replica on my `MachineDeployments`.

The problem is that `AzureMachineTemplate` has a `RoleAssignmentName` field for the name of the role assignment, which is then used when creating `AzureMachines`. But if more than one `AzureMachine` or VM is created out of this `AzureMachineTemplate` (i.e having more than 1 VM for the control plane, or even more than 1 replica in a `MachineDeployment`) the VM creation will fail with Azure API error `RoleAssignmentUpdateNotPermitted` saying `Tenant ID, application ID, principal ID, and scope are not allowed to be updated`.

I see that the templates used for e2e tests normally use only 1 replica for tests, that may explain why this hasn't been spotted so far.

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Fix 'SystemAssigned' identity by removing the defaulting of 'RoleAssignmentName' on 'AzureMachineTemplate' so that every 'AzureMachine'  defaults to a random 'RoleAssignmentName'.
```

### DIFF
--- a/api/v1beta1/azuremachinetemplate_webhook.go
+++ b/api/v1beta1/azuremachinetemplate_webhook.go
@@ -27,7 +27,10 @@ import (
 )
 
 // AzureMachineTemplateImmutableMsg ...
-const AzureMachineTemplateImmutableMsg = "AzureMachineTemplate spec.template.spec field is immutable. Please create new resource instead. ref doc: https://cluster-api.sigs.k8s.io/tasks/change-machine-template.html"
+const (
+	AzureMachineTemplateImmutableMsg          = "AzureMachineTemplate spec.template.spec field is immutable. Please create new resource instead. ref doc: https://cluster-api.sigs.k8s.io/tasks/change-machine-template.html"
+	AzureMachineTemplateRoleAssignmentNameMsg = "AzureMachineTemplate spec.template.spec.roleAssignmentName field can't be set"
+)
 
 // SetupWebhookWithManager sets up and registers the webhook with the manager.
 func (r *AzureMachineTemplate) SetupWebhookWithManager(mgr ctrl.Manager) error {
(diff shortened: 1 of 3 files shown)


api/v1beta1/azuremachinetemplate_webhook.go
```diff
@@ -92,5 +92,9 @@ func (r *AzureMachineTemplate) ValidateDelete() error {
 
 // Default implements webhookutil.defaulter so a webhook will be registered for the type.
 func (r *AzureMachineTemplate) Default() {
-	r.Spec.Template.Spec.SetDefaults()
+	if err := r.Spec.Template.Spec.SetDefaultSSHPublicKey(); err != nil {
+		ctrl.Log.WithName("SetDefault").Error(err, "SetDefaultSshPublicKey failed")
```
CecileRobertMichon, line 96:
```suggestion
		ctrl.Log.WithName("SetDefault").Error(err, "SetDefaultSSHPublicKey failed")
```

shysank:
Ideally, we want to remove `RoleAssignmentName` from `AzureMachineTemplate`, but can we atleast add a validation to make sure no value gets set?

shysank:
/lgtm
/assign @CecileRobertMichon 
can you squash commits please?

api/v1beta1/azuremachinetemplate_webhook.go
```diff
@@ -92,5 +104,9 @@ func (r *AzureMachineTemplate) ValidateDelete() error {
 
 // Default implements webhookutil.defaulter so a webhook will be registered for the type.
 func (r *AzureMachineTemplate) Default() {
-	r.Spec.Template.Spec.SetDefaults()
```
jackfrancis, line 95: Rather than changing this here, shouldn't we update the `func (s *AzureMachineSpec) SetDefaults()` method directly in `api/v1beta1/azuremachine_default.go`? That `SetDefaults()` method is also invoked in `api/v1beta1/azuremachine_webhook.go` (in the `func (m *AzureMachine) Default()` method)

jackfrancis:
So the core issue here seems to be that a static value in a `MachineTemplate` will render as-is on each `Machine` resource using it. Surely there are other such examples of machine-unique configuration, how is that handled?

There seems to be something fundamentally broken in the type spec here. What is the ideal way to define machine-specific configuration as distinct from `MachineTemplate` configuration? Should such properties even be present in the `MachineSpec` type definition at all?

cc @CecileRobertMichon @shysank

api/v1beta1/azuremachinetemplate_webhook.go
fiunchinho: I think it makes sense to default the `RoleAssignmentName` to a random UUID in the `AzureMachine` object, that's fine because all VMs would have a different one. It only doesn't make sense at the `AzureMachineTemplate` level.

fiunchinho:
> Can you squash commits please?
done

jackfrancis (approved):
/lgtm

api/v1beta1/azuremachinetemplate_webhook.go
jackfrancis: Thanks for the explanation, makes sense

PR #2123, opened 2022-02-25
add Eventually() to retryable k8s E2E operations

### LINKED ISSUE #2120: VMSS e2e test flake during drain
/kind bug

VMSS cluster test sometimes fails with:

```
1/24 Tests Failed. | expand_less
-- | --
capz-e2e: Workload cluster creation Creating a VMSS cluster [REQUIRED] with a single control plane node and an AzureMachinePool with 2 Linux and 2 Windows worker nodes expand_less23m13s/home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/test/e2e/azure_test.go:336 Unexpected error:     <*errors.StatusError \| 0xc00089f4a0>: {         ErrStatus: {             TypeMeta: {Kind: "", APIVersion: ""},             ListMeta: {                 SelfLink: "",                 ResourceVersion: "",                 Continue: "",                 RemainingItemCount: nil,             },             Status: "Failure",             Message: "nodes \"capz-e2e-e987kg-vmss-mp-0000000\" not found",             Reason: "NotFound",             Details: {                 Name: "capz-e2e-e987kg-vmss-mp-0000000",                 Group: "",                 Kind: "nodes",                 UID: "",                 Causes: nil,                 RetryAfterSeconds: 0,             },             Code: 404,         },     }     nodes "capz-e2e-e987kg-vmss-mp-0000000" not found occurred /home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/test/e2e/azure_machinepool_drain.go:204 | capz-e2e: Workload cluster creation Creating a VMSS cluster [REQUIRED] with a single control plane node and an AzureMachinePool with 2 Linux and 2 Windows worker nodes expand_less | 23m13s | /home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/test/e2e/azure_test.go:336 Unexpected error:     <*errors.StatusError \| 0xc00089f4a0>: {         ErrStatus: {             TypeMeta: {Kind: "", APIVersion: ""},             ListMeta: {                 SelfLink: "",                 ResourceVersion: "",                 Continue: "",                 RemainingItemCount: nil,             },             Status: "Failure",             Message: "nodes \"capz-e2e-e987kg-vmss-mp-0000000\" not found",             Reason: "NotFound",             Details: {                 Name: "capz-e2e-e987kg-vmss-mp-0000000",                 Group: "",                 Kind: "nodes",                 UID: "",                 Causes: nil,                 RetryAfterSeconds: 0,             },             Code: 404,         },     }     nodes "capz-e2e-e987kg-vmss-mp-0000000" not found occurred /home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/test/e2e/azure_machinepool_drain.go:204
capz-e2e: Workload cluster creation Creating a VMSS cluster [REQUIRED] with a single control plane node and an AzureMachinePool with 2 Linux and 2 Windows worker nodes expand_less | 23m13s
/home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/test/e2e/azure_test.go:336 Unexpected error:     <*errors.StatusError \| 0xc00089f4a0>: {         ErrStatus: {             TypeMeta: {Kind: "", APIVersion: ""},             ListMeta: {                 SelfLink: "",                 ResourceVersion: "",                 Continue: "",                 RemainingItemCount: nil,             },             Status: "Failure",             Message: "nodes \"capz-e2e-e987kg-vmss-mp-0000000\" not found",             Reason: "NotFound",             Details: {                 Name: "capz-e2e-e987kg-vmss-mp-0000000",                 Group: "",                 Kind: "nodes",                 UID: "",                 Causes: nil,                 RetryAfterSeconds: 0,             },             Code: 404,         },     }     nodes "capz-e2e-e987kg-vmss-mp-0000000" not found occurred /home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/test/e2e/azure_machinepool_drain.go:204

```

Example PR where this happened: #2119 

**Environment:**

jackfrancis:
**What type of PR is this?**

/kind failing-test

This PR introduces retries to E2E tests so that transient errors can be absorbed.

Fixes #2120

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/aks.go
+++ b/test/e2e/aks.go
@@ -188,7 +188,7 @@ func WaitForControlPlaneMachinesToExist(ctx context.Context, input WaitForContro
 		ammpList := &infrav1exp.AzureManagedMachinePoolList{}
 
 		if err := input.Lister.List(ctx, ammpList, opt1, opt2); err != nil {
-			Logf("Failed to get machinePool: %+v", err)
+			LogWarningf("Failed to get machinePool: %+v", err)
 			return false
 		}
 
@@ -202,7 +202,7 @@ func WaitForControlPlaneMachinesToExist(ctx context.Context, input WaitForContro
 				ownerMachinePool := &clusterv1exp.MachinePool{}
 				if err := input.Getter.Get(ctx, types.NamespacedName{Namespace: input.Namespace, Name: ref.Name},
 					ownerMachinePool); err != nil {
-					Logf("Failed to get machinePool: %+v", err)
+					LogWarningf("Failed to get machinePool: %+v", err)
 					return false
 				}
 				if len(ownerMachinePool.Status.NodeRefs) >= minReplicas.value(ownerMachinePool) {
--- a/test/e2e/azure_lb.go
+++ b/test/e2e/azure_lb.go
@@ -123,10 +123,16 @@ func AzureLBSpec(ctx context.Context, inputGetter func() AzureLBSpecInput) {
 	if !input.IPv6 {
 		By("creating an internal Load Balancer service")
 
-		ilbService := webDeployment.GetService(ports, deploymentBuilder.InternalLoadbalancer)
+		ilbService := webDeployment.CreateServiceResourceSpec(ports, deploymentBuilder.InternalLoadbalancer)
 		Log("starting to create an internal Load Balancer service")
-		_, err = servicesClient.Create(ctx, ilbService, metav1.CreateOptions{})
-		Expect(err).NotTo(HaveOccurred())
+		Eventually(func() error {
+			_, err := servicesClient.Create(ctx, ilbService, metav1.CreateOptions{})
+			if err != nil {
+				LogWarningf("failed creating service (%s):%s\n", ilbService.Name, err.Error())
+				return err
+			}
+			return nil
+		}, retryableOperationTimeout, retryableOperationSleepBetweenRetries).Should(Succeed())
 		ilbSvcInput := WaitForServiceAvailableInput{
 			Getter:    servicesClientAdapter{client: servicesClient},
 			Service:   ilbService,
@@ -136,14 +142,28 @@ func AzureLBSpec(ctx context.Context, inputGetter func() AzureLBSpecInput) {
 
 		By("connecting to the internal LB service from a curl pod")
 
-		svc, err := servicesClient.Get(ctx, ilbService.Name, metav1.GetOptions{})
-		Expect(err).NotTo(HaveOccurred())
+		var svc *corev1.Service
+		Eventually(func() error {
+			var err error
+			svc, err = servicesClient.Get(ctx, ilbService.Name, metav1.GetOptions{})
+			if err != nil {
+				LogWarningf("failed getting service (%s):%s\n", ilbService.Name, err.Error())
+				return err
+			}
+			return nil
+		}, retryableOperationTimeout, retryableOperationSleepBetweenRetries).Should(Succeed())
 		ilbIP := extractServiceIp(svc)
 
-		ilbJob := job.CreateCurlJob("curl-to-ilb-job", ilbIP)
+		ilbJob := job.CreateCurlJobResourceSpec("curl-to-ilb-job", ilbIP)
 		Log("starting to create a curl to ilb job")
-		_, err = jobsClient.Create(ctx, ilbJob, metav1.CreateOptions{})
-		Expect(err).NotTo(HaveOccurred())
+		Eventually(func() error {
+			_, err := jobsClient.Create(ctx, ilbJob, metav1.CreateOptions{})
+			if err != nil {
+				LogWarningf("failed creating job (%s):%s\n", ilbJob.Name, err.Error())
+				return err
+			}
+			return nil
+		}, retryableOperationTimeout, retryableOperationSleepBetweenRetries).Should(Succeed())
 		ilbJobInput := WaitForJobCompleteInput{
 			Getter:    jobsClientAdapter{client: jobsClient},
 			Job:       ilbJob,
@@ -154,19 +174,37 @@ func AzureLBSpec(ctx context.Context, inputGetter func() AzureLBSpecInput) {
 		if !input.SkipCleanup {
 			By("deleting the ilb test resources")
 			Logf("deleting the ilb service: %s", ilbService.Name)
-			err = servicesClient.Delete(ctx, ilbService.Name, metav1.DeleteOptions{})
-			Expect(err).NotTo(HaveOccurred())
+			Eventually(func() error {
+				err := servicesClient.Delete(ctx, ilbService.Name, metav1.DeleteOptions{})
+				if err != nil {
+					LogWarningf("failed deleting service (%s):%s\n", ilbService.Name, err.Error())
+					return err
+				}
+				return nil
+			}, deleteOperationTimeout, retryableOperationSleepBetweenRetries).Should(Succeed())
 			Logf("deleting the ilb job: %s", ilbJob.Name)
-			err = jobsClient.Delete(ctx, ilbJob.Name, metav1.DeleteOptions{})
-			Expect(err).NotTo(HaveOccurred())
+			Eventually(func() error {
+				err := jobsClient.Delete(ctx, ilbJob.Name, metav1.DeleteOptions{})
+				if err != nil {
+					LogWarningf("failed deleting job (%s):%s\n", ilbJob.Name, err.Error())
+					return err
+				}
+				return nil
+			}, deleteOperationTimeout, retryableOperationSleepBetweenRetries).Should(Succeed())
 		}
 	}
 
 	By("creating an external Load Balancer service")
-	elbService := webDeployment.GetService(ports, deploymentBuilder.ExternalLoadbalancer)
+	elbService := webDeployment.CreateServiceResourceSpec(ports, deploymentBuilder.ExternalLoadbalancer)
 	Log("starting to create an external Load Balancer service")
-	_, err = servicesClient.Create(ctx, elbService, metav1.CreateOptions{})
-	Expect(err).NotTo(HaveOccurred())
+	Eventually(func() error {
+		_, err := servicesClient.Create(ctx, elbService, metav1.CreateOptions{})
+		if err != nil {
+			LogWarningf("failed creating service (%s):%s\n", elbService.Name, err.Error())
+			return err
+		}
+		return nil
+	}, retryableOperationTimeout, retryableOperationSleepBetweenRetries).Should(Succeed())
 	elbSvcInput := WaitForServiceAvailableInput{
 		Getter:    servicesClientAdapter{client: servicesClient},
 		Service:   elbService,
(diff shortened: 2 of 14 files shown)


test/e2e/azure_machinepool_drain.go
```diff
@@ -198,13 +198,25 @@ func testMachinePoolCordonAndDrain(ctx context.Context, mgmtClusterProxy, worklo
...
+		Eventually(func() error {
+			err := workloadClient.Get(ctx, client.ObjectKey{
+				Name:      ampm.Status.NodeRef.Name,
+				Namespace: ampm.Status.NodeRef.Namespace,
+			}, n)
+			if err != nil {
+				return err
+			}
+			return nil
+		}, 30*time.Second, 3*time.Second).Should(Succeed())
```
CecileRobertMichon, line 210: would it be possible to make k8s operation retry timeout/interval constants?

jackfrancis: Definitely. Does this approach seem sensible? Retry between 5-10 times (depending on how long the failed op takes) and then timeout at 30 seconds?

CecileRobertMichon: I think it does, if we need more than 5-10 retries something bigger is wrong and we need to pay attention

jsturtevant: should we log any failures to see if there is a pattern?  Are we adding this due to the fact we are seeing failures on these calls?

jackfrancis: Strictly speaking we I wanna add retry logic to all i/o requests that are by definition retry-able. But yeah, we should log every error as a best practice to give folks an opportunity (however manual) to identify i/o flakes that were overcome w/ a retry.

test/e2e/azure_machinepool_drain.go
```diff
@@ -154,35 +156,21 @@ func testMachinePoolCordonAndDrain(ctx context.Context, mgmtClusterProxy, worklo
...
+			LogWarning(err.Error())
 			return err
 		}
 
 		decreasedReplicas := *owningMachinePool.Spec.Replicas - int32(1)
 		owningMachinePool.Spec.Replicas = &decreasedReplicas
 		return helper.Patch(ctx, owningMachinePool)
-	})
-
-	By(fmt.Sprintf("checking for a machine to start draining for machine pool: %s/%s", amp.Namespace, amp.Name))
```
jackfrancis, line 165: This test was never actually running, because the `Eventually()` was not actually running according to any success criteria. When I enabled the test using a retry + success criteria, it was failing consistently. After investigation, it seems pretty clear that evaluating for "started but hasn't yet finished draining" is a small enough window that it will regularly fail. Let's just get rid of this altogether and run the next test ("check for successful drain").

See https://kubernetes.slack.com/archives/CEX9HENG7/p1646686318723399

jackfrancis: Actually it's the "finished draining but hasn't yet deleted" window that is too small.

test/e2e/azure_machinepool_drain.go
```diff
@@ -176,35 +180,33 @@ func testMachinePoolCordonAndDrain(ctx context.Context, mgmtClusterProxy, worklo
...
 
-		for _, machine := range ampmls {
-			if conditions.Has(&machine, clusterv1.DrainingSucceededCondition) && conditions.IsTrue(&machine, clusterv1.DrainingSucceededCondition) {
-				return nil // started draining the node prior to delete
-			}
-		}
-
-		return errors.New("no machine has finished draining")
-	})
+	// TODO setup a watcher to detect the terminal drain success state
```
jackfrancis, line 185: Rather than checking for the expected state transitions serially, we can improve this in the future by setting a watch on the machine in the process of deleting. I'm making the judgment that doing that work is out of scope for this PR (this PR is about standardizing the usage of Eventually() for retryable operations).

Bottom line, this test coverage was never actually running due to the lack of `.Should(Succeed())`, so adding the "validate that drain begins after delete" is already net additive coverage.

test/e2e/aks.go
```diff
@@ -188,7 +188,7 @@ func WaitForControlPlaneMachinesToExist(ctx context.Context, input WaitForContro
 		ammpList := &infrav1exp.AzureManagedMachinePoolList{}
 
 		if err := input.Lister.List(ctx, ammpList, opt1, opt2); err != nil {
-			Logf("Failed to get machinePool: %+v", err)
+			LogWarningf("Failed to get machinePool: %+v", err)
```
CecileRobertMichon, line 191: side node: why is this func called `WaitForControlPlaneMachinesToExist` seems like it's not checking control planes (that wouldn't even make sense since it's for AKS)

test/e2e/aks.go
```diff
@@ -202,7 +202,7 @@ func WaitForControlPlaneMachinesToExist(ctx context.Context, input WaitForContro
 				ownerMachinePool := &clusterv1exp.MachinePool{}
 				if err := input.Getter.Get(ctx, types.NamespacedName{Namespace: input.Namespace, Name: ref.Name},
 					ownerMachinePool); err != nil {
-					Logf("Failed to get machinePool: %+v", err)
+					LogWarningf("Failed to get machinePool: %+v", err)
```
CecileRobertMichon, line 205: this seems like an actual failure and it's not in a retry block, why log it as a warning?

test/e2e/aks.go
jackfrancis: I know... (I mean, I don't actually know)

test/e2e/azure_machinepool_drain.go
```diff
@@ -176,35 +180,33 @@ func testMachinePoolCordonAndDrain(ctx context.Context, mgmtClusterProxy, worklo
...
 		n.Labels[clusterv1.OwnerKindAnnotation] = "MachinePool"
 		n.Labels[clusterv1.OwnerNameAnnotation] = mpName
-		Expect(workloadClient.Update(ctx, n)).ToNot(HaveOccurred())
+		Eventually(func() error {
+			err := workloadClient.Update(ctx, n)
+			if err != nil {
+				LogWarning(err.Error())
+			}
+			return err
+		}, 30*time.Second, 3*time.Second).Should(Succeed())
```
CecileRobertMichon, line 209: can this wait interval/retry be added as a const ? Seems like it's repeated in a few places

test/e2e/azure_lb.go
```diff
@@ -154,19 +174,37 @@ func AzureLBSpec(ctx context.Context, inputGetter func() AzureLBSpecInput) {
 		if !input.SkipCleanup {
 			By("deleting the ilb test resources")
 			Logf("deleting the ilb service: %s", ilbService.Name)
-			err = servicesClient.Delete(ctx, ilbService.Name, metav1.DeleteOptions{})
-			Expect(err).NotTo(HaveOccurred())
+			Eventually(func() error {
+				err := servicesClient.Delete(ctx, ilbService.Name, metav1.DeleteOptions{})
```
CecileRobertMichon, line 178: based on our previous conversation in slack, my understanding was that this Delete doesn't actually wait for the service to be deleted, it just waits for the client go Delete to return, which doesn't fully wait for the service to be gone. Is that correct?

If so, waiting 20 minutes doesn't seem appropriate. We should change it to actually wait for the service to be gone so that we avoid running into flakes later due to cloud provider being in the middle of reconciling the deleted service when we try to create a new service.

test/e2e/azure_machinepool_drain.go
```diff
@@ -275,14 +283,17 @@ func deployHttpService(ctx context.Context, clientset *kubernetes.Clientset, isW
 		Eventually(func() error {
 			version, err := node.GetWindowsVersion(ctx, clientset)
 			windowsVersion = version
+			if err != nil {
+				LogWarning(err.Error())
+			}
 			return err
 		}, 300*time.Second, 5*time.Second).Should(Succeed())
```
CecileRobertMichon, line 290: same here about using a const

test/e2e/azure_machinepool_drain.go
```diff
@@ -47,7 +47,9 @@ import (
 )
 
 const (
-	AzureMachinePoolDrainSpecName = "azure-mp-drain"
+	AzureMachinePoolDrainSpecName   = "azure-mp-drain"
+	waitForDrainOperationTimeout    = 20 * time.Minute
+	waitForDrainSleepBetweenRetries = 10 * time.Second
```
CecileRobertMichon, line 52: isn't 10 seconds pretty long in between retries if we're concerned about the node changing states too quickly?

test/e2e/azure_machinepool_drain.go
CecileRobertMichon: can we open an issue for this rather than leaving a TODO in the code? Also does this mean #2120 isn't fully fixed?

test/e2e/aks.go
jackfrancis: There is no existent error equivalent of `Logf` Do we want to create one?

test/e2e/azure_machinepool_drain.go
```diff
@@ -176,35 +180,33 @@ func testMachinePoolCordonAndDrain(ctx context.Context, mgmtClusterProxy, worklo
...
-	By(fmt.Sprintf("checking for a machine to successfully complete draining for machine pool: %s/%s", amp.Namespace, amp.Name))
-	Eventually(func() error {
-		ampmls, err := getAzureMachinePoolMachines(ctx, mgmtClusterProxy, workloadClusterProxy, amp)
-		if err != nil {
-			return errors.Wrap(err, "failed to list the azure machine pool machines")
-		}
+	}, waitForDrainOperationTimeout, waitForDrainSleepBetweenRetries).Should(Succeed())
 
-		for _, machine := range ampmls {
-			if conditions.Has(&machine, clusterv1.DrainingSucceededCondition) && conditions.IsTrue(&machine, clusterv1.DrainingSucceededCondition) {
```
CecileRobertMichon, line 189: if the issue is that the machine gets deleted too quickly, I wonder if we could add a finalizer to it before the test starts (in the setup above) so it doesn't get deleted until we've done all the checks and we're ready to let it delete

test/e2e/azure_lb.go
jackfrancis: It's actually not clear (to me) in the client-go code if the delete operation blocks on success here...

https://github.com/kubernetes/client-go/blob/v0.23.0/kubernetes/typed/core/v1/service.go#L160

I don't see a "wait" or "do not wait" option here

https://github.com/kubernetes/apimachinery/blob/v0.23.0/pkg/apis/meta/v1/types.go#L482

test/e2e/azure_machinepool_drain.go
jackfrancis: Fair, and since we're currently just testing for drain start signal, we can move the timeout down to something lower as well

test/e2e/azure_machinepool_drain.go
jackfrancis: What's the proper scope for this PR? Given that these existing drain tests don't actually work at all. Should we simply abandon any changes here and follow-up with targeted improvements to drain testing?

test/e2e/azure_machinepool_drain.go
jackfrancis: The referenced issue documents a flake that should be addressed w/ the new `Eventually()` block in L191 below

test/e2e/azure_machinepool_drain.go
jackfrancis: this is only used once (and this one is inherited, should we standardize it to the newly added "try every 3 seconds up to 30 seconds" config?)

jackfrancis: In fact the underlying node list call in `node.GetWindowsVersion` is now implementing retries. This is the right place for this, so I'll get rid of the `Eventually` wrapping here altogether.

test/e2e/azure_machinepool_drain.go
jackfrancis: done

test/e2e/aks.go
CecileRobertMichon: I see, Logf is actually Info which is even less acurrate. I thought we were decreasing severity.

test/e2e/azure_lb.go
CecileRobertMichon: Based on the timing in logs it seems very likely that the delete is not blocking:

```
Mar  9 20:30:50.686: INFO: job default/curl-to-ilb-joboupqs is complete, took 10.069781536s
[1mSTEP[0m: deleting the ilb test resources
Mar  9 20:30:50.686: INFO: deleting the ilb service: webg7snfx-ilb
Mar  9 20:30:50.741: INFO: deleting the ilb job: curl-to-ilb-joboupqs
[1mSTEP[0m: creating an external Load Balancer service
Mar  9 20:30:50.776: INFO: starting to create an external Load Balancer service
```

We should add a check to make sure the service is gone before we proceed. We don't have to fix it in this PR but let's at least reduce the timeout, 20 minutes is too long if it's just for the action of starting the delete (and not the delete actually completing).

test/e2e/aks.go
CecileRobertMichon: seems like either faulty copy/pasta or a "trick" to make it fit into the test framework but then it should really have better comments. My bet is on #1. 

anyways that's out of scope here.

CecileRobertMichon:
/lgtm
/approve

let's merge this and follow up with some targetted improvements for LB and VMSS drain tests

PR #2125, opened 2022-02-25
Allow for workflow of skipping conformance tests

jsturtevant:
**What type of PR is this?**
/kind feature

When deflaking upstream tests it is sometime necessary to create a cluster using the upstream kubernetes builds but the developer may not want to run tests as they will do this manually afterwards.  

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

/assign @marosset

### DIFF
--- a/docs/book/src/developers/development.md
+++ b/docs/book/src/developers/development.md
@@ -525,6 +525,7 @@ With the following environment variables defined, you can build a CAPZ cluster f
 | `WINDOWS_FLAVOR`     | Optional, can be `containerd` or `containerd-2022`, when not specified dockershim is used |
 | `KUBETEST_WINDOWS_CONFIG`     | Optional, can be `upstream-windows-serial-slow.yaml`, when not specified `upstream-windows.yaml` is used |
 | `WINDOWS_CONTAINERD_URL`     | Optional, can be any url to a `tar.gz` file containing binaries for containerd in the same format as upstream package |
+| `SKIP_E2E_TEST`           | Optional, if `true` then the cluster will be created but the conformance tests will not be run against the cluster.  |
 
 With the following environment variables defined, CAPZ runs `./scripts/ci-build-kubernetes.sh` as part of `./scripts/ci-conformance.sh`, which allows developers to build Kubernetes from source and run the Kubernetes Conformance test suite against a CAPZ cluster based on the custom build:
 
--- a/test/e2e/conformance_test.go
+++ b/test/e2e/conformance_test.go
@@ -212,6 +212,12 @@ var _ = Describe("Conformance Tests", func() {
 		ginkgoNodes, err := strconv.Atoi(e2eConfig.GetVariable("CONFORMANCE_NODES"))
 		Expect(err).NotTo(HaveOccurred())
 
+		skipTests := os.Getenv("SKIP_E2E_TEST")
+		if skipTests == "true" {
+			fmt.Fprintf(GinkgoWriter, "INFO: Skipping conformance tests")
+			return
+		}
+
 		runtime = b.Time("conformance suite", func() {
 			err := kubetest.Run(context.Background(),
 				kubetest.RunInput{


jackfrancis:
Is including the  `Conformance Tests` regex substring as the value to `GINKGO_SKIP` not sufficient?

marosset:
> Is including the `Conformance Tests` regex substring as the value to `GINKGO_SKIP` not sufficient?

That might be sufficient but is more of a burden for users IMO.

marosset:
> > Is including the `Conformance Tests` regex substring as the value to `GINKGO_SKIP` not sufficient?
> 
> That might be sufficient but is more of a burden for users IMO.

If we can just set `GINKGO_SKIP=*` that would be acceptable.

jsturtevant:
> Is including the `Conformance Tests` regex substring as the value to `GINKGO_SKIP` not sufficient?

This would skip the creation of the conformance cluster.  

Another option is to update docs that tell the user to modify the `ginkgo.focus` in the files: https://github.com/kubernetes-sigs/cluster-api-provider-azure/tree/main/test/e2e/data/kubetest but they would require doing this for every file and then have to avoid checking it in.

jackfrancis:
@jsturtevant thanks for the explanation, I see that this is just for the `ci-conformance.sh` script. I do think a more idiomatic way to do this, given that the script depends on ginkgo to do the actual work and testing, would be identify the "build the cluster" part and the "run the tests" part in GINKGO_SKIP/GINKGO_FOCUS-identifiable strings so that we could simply, easily re-use (and continue to build out) that existing facility.

Also, by doing it in the above way, we could prepare for a future use-case that is the inverse of what we're attempting to do here: run the tests but don't build a cluster (i.e., run on a pre-existing cluster).

docs/book/src/developers/development.md
```diff
@@ -525,6 +525,7 @@ With the following environment variables defined, you can build a CAPZ cluster f
 | `WINDOWS_FLAVOR`     | Optional, can be `containerd` or `containerd-2022`, when not specified dockershim is used |
 | `KUBETEST_WINDOWS_CONFIG`     | Optional, can be `upstream-windows-serial-slow.yaml`, when not specified `upstream-windows.yaml` is used |
 | `WINDOWS_CONTAINERD_URL`     | Optional, can be any url to a `tar.gz` file containing binaries for containerd in the same format as upstream package |
+| `SKIP_E2E_TEST`           | Optional, if `true` then the cluster will be created but the conformance tests will not be run against the cluster.  |
```
mboersma, line 528: Nitpick: I think `SKIP_E2E_TESTS` reads better, but this lgtm either way.

jsturtevant:
> I see that this is just for the `ci-conformance.sh` script. I do think a more idiomatic way to do this, given that the script depends on ginkgo to do the actual work and testing, would be identify the "build the cluster" part and the "run the tests" part in GINKGO_SKIP/GINKGO_FOCUS-identifiable strings so that we could simply, easily re-use (and continue to build out) that existing facility.

I don't think it will be useful to re-use the `GINKGO_*` capz specific variables here as they are two separate processes.  The conformance tests run in a docker container that is created and are passed the viper config files from https://github.com/kubernetes-sigs/cluster-api-provider-azure/tree/main/test/e2e/data/kubetest (which has the ginkgo skip/focus configuration)

I could see it making sense to allow for viper config `ginko.focus` overidable via an env variable something like `CONFORMANCE_GINKGO_SKIP` which substitutes those configuration files. This would also allow for PROW jobs to use this env variable as well which would be useful for Windows tests. wdyt?

jackfrancis:
@jsturtevant nah, I think your implementation makes sense, lgtm!

docs/book/src/developers/development.md
CecileRobertMichon: @jsturtevant are you planning on changing it or is this ready for merge?

jsturtevant: I will update this if we have agreement we should go with this over a way to provide conformance tests ginkgo parameters to e2e-test binary https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2125#issuecomment-1051262577

jsturtevant:
I've thought about this some more, and although it is a bit more work I think the ability to run individual conformance tests and skip them would be beneficial.  It would simplify the logic for multiple files and allow developers to target a subset of tests.  For instance, we have https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/test/e2e/data/kubetest/conformance-fast.yaml and https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/test/e2e/data/kubetest/conformance.yaml. 

We could have a single template and push that config up to the PROW jobs.  Same goes for https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/test/e2e/data/kubetest/upstream-windows-serial-slow.yaml and https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/test/e2e/data/kubetest/upstream-windows.yaml

jsturtevant:
For windows we are going a different route for running a majority of the tests.  This `ci-conformance.sh` will still be used to validate capz works against upstream but we will not override the tests here.  A developer can update the file if required for testing locally. Closing for now

/close

PR #2126, opened 2022-02-25
Add warning to docs about reference image availability

### LINKED ISSUE #2114: Create availability policy for reference images, so we can prune them
/kind feature

CAPZ documentation should make clear that the project provides reference images only for the last *X* patch releases of Kubernetes, for each of the last *Y* minor releases. Then maintainers should prune our stored images down to those covered by the availability policy.

**Background**

To facilitate tests, and to help new users kick the tires, the CAPZ project has published reference images for each version of Kubernetes released over the last couple years. But there are costs associated with keeping the entire set of reference images available.

The number of images involved has become quite large and poses some challenges for Azure's Marketplace. We run into issues frequently due to the size of the overall offer. We are also billed for the amount of storage used, which continues to grow.

Additionally, these images are for testing and reference. They aren't "supported" in the sense that Azure usually supports its services and binary artifacts. They are not updated with OS-level security patches, and many Kubernetes releases remain there after they have been effectively obsoleted by newer versions with CVE fixes.

The danger is that even with making the availability policy clear and taking a deliberate approach to removing old images (such as Kubernetes 1.16.10!), there may be users who depend on a particular image and are surprised. Since these images were never supported, and we have made clear that after a trial phase, users need to maintain their own images, I think we are covered. But I would love any other suggestions to help avoid this.

mboersma:
**What type of PR is this?**:

/kind documentation

Announces that Team CAPZ may begin removing reference images for old, unsupported releases of Kubernetes. Also clarifies some language about reference images and security, and uses HTML callout sections to be more consistent with other doc warnings.

Fixes #2114

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Add warning to docs about reference image availability
```

### DIFF
--- a/docs/book/src/topics/custom-images.md
+++ b/docs/book/src/topics/custom-images.md
@@ -6,13 +6,35 @@ This document will help you get a CAPZ Kubernetes cluster up and running with yo
 
 An *image* defines the operating system and Kubernetes components that will populate the disk of each node in your cluster.
 
-By default, images offered by "capi" in the [Azure Marketplace][azure-marketplace] are used. You can list these *reference images* with this command:
+By default, images offered by "capi" in the [Azure Marketplace][azure-marketplace] are used.
+
+You can list these *reference images* with this command:
 
 ```bash
 az vm image list --publisher cncf-upstream --offer capi --all -o table
 ```
 
-Note: These images are not updated for security fixes and it is recommended to always use the latest patch version for the Kubernetes version you wish to run. For production-like environments, and for more control over your nodes, it is highly recommended to build and use your own custom images.
+It is recommended to use the latest patch release of Kubernetes for a [supported minor release][supported-k8s].
+
+<aside class="note warning">
+
+<h1> Availability </h1>
+
+The Cluster API for Azure team publishes *reference images* for each Kubernetes release, for both Linux and Windows.
+
+Reference images for versions of Kubernetes which have known security issues or which are no longer [supported by Cluster API][supported-capi] will be removed from the Azure Marketplace.
+
+</aside>
+
+<aside class="note warning">
+
+<h1> Security </h1>
+
+The reference images are not updated with security fixes. They are intended only to facilitate testing and to help users try out Cluster API for Azure.
+
+The reference images should not be used in a production environment. It is highly recommended to [maintain your own custom image](#building-a-custom-image) instead.
+
+</aside>
 
 ## Building a custom image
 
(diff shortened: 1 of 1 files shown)


docs/book/src/topics/custom-images.md
```diff
@@ -6,13 +6,35 @@ This document will help you get a CAPZ Kubernetes cluster up and running with yo
...
-Note: These images are not updated for security fixes and it is recommended to always use the latest patch version for the Kubernetes version you wish to run. For production-like environments, and for more control over your nodes, it is highly recommended to build and use your own custom images.
+It is recommended to use the latest patch release of Kubernetes for a [supported minor release][supported-k8s].
+
+<aside class="note warning">
+
+<h1> Availability </h1>
+
+The Cluster API for Azure team publishes *reference images* for each Kubernetes release, for both Linux and Windows.
+
+Reference images for versions of Kubernetes that are no longer [supported by Cluster API][supported-capi] will be removed from the Azure Marketplace.
```
mboersma, line 25: I think in practice, we might want to keep more images around than this implies. @CecileRobertMichon also has a query that might be able to show us if a certain reference image has been used recently, as a further safety check before we actually delete them.

mboersma:
FYI @sonasingh46 since we were talking about creating the reference images.

docs/book/src/topics/custom-images.md
mboersma: Also I thought about justifying this statement with two reasons:

- Out-of-date versions of k8s often have unfixed CVEs or significant bugs, so making them always available is really a liability from our point of view
- It costs actual $$ to maintain this many images in storage

But I decided to omit the reasons, because I'm not sure it's necessary to justify it in the doc. But what do others think?

docs/book/src/topics/custom-images.md
```diff
@@ -6,13 +6,35 @@ This document will help you get a CAPZ Kubernetes cluster up and running with yo
...
+
+</aside>
+
+<aside class="note warning">
+
+<h1> Security </h1>
+
+The reference images are not updated with security fixes. They are intended only to facilitate testing and to help users try out Cluster API for Azure.
+
+The reference images should not be used in a production environment. It is highly recommended to maintain your own custom image instead.
```
CecileRobertMichon, line 35: link to image-builder book maybe?

mboersma: This was meant to lead directly into the next section "Building a custom image," so I linked it to that local anchor instead. We do link to the image-builder book there.

docs/book/src/topics/custom-images.md
bridgetkromhout: I think we should stop providing images that we know are unpatched CVEs waiting to happen. We can warn people before we remove them, but we should remove them and stop providing them going forward, because it's a liability we do not need to carry. Cost isn't the main reason; "we aren't going to provide things people should not use" is a completely defensible reason.

The published statement can just be "we provide images of what k8s upstream is patching; here's a link to their release schedule, full stop".

mboersma: What if I modify it to say this:

> Reference images for versions of Kubernetes which have known security issues or which  are no longer supported by Cluster API will be removed from the Azure Marketplace.

(We do link to the k8s supported releases doc just above.)

mboersma:
/retest

Pod timeout yadda yadda.

docs/book/src/topics/custom-images.md
mboersma: @bridgetkromhout I made the change above, let me know whatcha think.

CecileRobertMichon (commented):
/lgtm

jackfrancis (approved):
/lgtm

PR #2134, opened 2022-02-28
Populate ControlPlaneEndpoint when ManagedCluster update is not needed.

karthikbalasub:
**What type of PR is this?**
/kind feature

Fixes a bug in managed clusters service that results in control plane endpoint not getting updated correctly.
Before this fix, update flow will not read the FQDN from the the ManagedCluster fetched from cloud when no update was needed on the managed cluster.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Release note**:

```release-note
Fixed a bug in managed clusters service that results in control plane endpoint not getting updated correctly.
```


azure/services/managedclusters/managedclusters.go
```diff
@@ -321,6 +321,11 @@ func (s *Service) Reconcile(ctx context.Context) error {
 			if err != nil {
 				return fmt.Errorf("failed to update managed cluster, %w", err)
 			}
+		} else {
+			// No update required, but use the MC fetched from Azure for reading fields below.
+			// This is to ensure the read-only fields like Fqdn from the existing MC are used for updating the
+			// AzureManagedCluster.
+			managedCluster = existingMC
```
jackfrancis, line 328: Is this something that we only need to do some of the time in practice? If not, how did this updates ever work at all?

jackfrancis: Also, would it just be simpler to modify L335 below to use:

`Host: *existingMC.ManagedClusterProperties.Fqdn,`

CecileRobertMichon:
@karthikbalasub looks like there's a merge conflict, please rebase

azure/services/managedclusters/managedclusters.go
karthikbalasub: > Is this something that we only need to do some of the time in practice? If not, how did this updates ever work at all?

The issue happens when the create response doesn't include the FQDN (can happen due to timeout, for example). When that happens, subsequent update calls will fail to update the controlplane endpoint. 

> Also, would it just be simpler to modify L335 below to use:
> 
> `Host: *existingMC.ManagedClusterProperties.Fqdn,`
This would be incorrect when there was an update needed (`diff != ""`) as the response from the update call is assigned to `managedCluster` in this line above:
```
managedCluster, err = s.Client.CreateOrUpdate(ctx, managedClusterSpec.ResourceGroupName, managedClusterSpec.Name, managedCluster, customHeaders)
```

karthikbalasub:
> @karthikbalasub looks like there's a merge conflict, please rebase

Done

azure/services/managedclusters/managedclusters.go
jackfrancis: But the FQDN will never change, right? So even if there is a diff to process we can always conclude that the FQDN is not one of the updated properties.

karthikbalasub: `existingMC` would also be null for the create flow.

jackfrancis: This logic is in the `else` block of an `if isCreate` conditional block. So it should not affect any create flows.

karthikbalasub: If you feel that it is odd to do `managedCluster = existingMC` then I can refactor the logic to populate endpoint to a separate function and call it with either managedCluster or existingMC depending on whether the cluster was updated or not.

karthikbalasub: > This logic is in the `else` block of an `if isCreate` conditional block. So it should not affect any create flows.

But the lines 341-350 which updates the controlplane endpoint are outside the if-block (Common for create and update flows). So changing line 343 as you suggested would break the code?

jackfrancis: > But the lines 341-350 which updates the controlplane endpoint are outside the if-block (Common for create and update flows). So changing line 343 as you suggested would break the code?

Correct, you're right we can'd do that.

jackfrancis: So, to summarize, the issue is that this block does not always work as expected?

```
// Update control plane endpoint.
	if managedCluster.ManagedClusterProperties != nil && managedCluster.ManagedClusterProperties.Fqdn != nil {
		endpoint := clusterv1.APIEndpoint{
			Host: *managedCluster.ManagedClusterProperties.Fqdn,
			Port: 443,
		}
		s.Scope.SetControlPlaneEndpoint(endpoint)
	}
```

And the reason that is doesn't always work is because this call occasionally returns a nil `ManagedClusterProperties.Fqdn`:

```
managedCluster, err = s.Client.CreateOrUpdate(
```

So in summary, we are not doing this on 100% of cluster create operations:

`s.Scope.SetControlPlaneEndpoint(endpoint)`

?

Is that correct?

karthikbalasub: Yes, correct.

jackfrancis: Is there an error condition in this func that is hit in the edge case where we don't get the FQDN back from the create operation?

https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/v1.2.0/azure/services/managedclusters/client.go#L81

Basically, I'm thinking that we should fix this issue at the root, and ensure that we *always* get a valid FQDN. If we don't we need to retry, or do something else to guaranteed that the `containerservice.ManagedCluster` we return is sane.

karthikbalasub: Not getting FQDN back after create operation is one scenario for failure. The issue can also happen when create request fails but the cluster was created in the server (for example, the operation took long and the request timed out, or a network failure etc).
In such cases, the retry would go through the update flow and the FQDN will not be updated correctly without the fix in this PR.

Even if we fix this at the root, we still need this fix for the scenarios where the initial create request fails to return a response.

jackfrancis: @CecileRobertMichon @zmalik @michalno1 does anyone else have thoughts here? I think @karthikbalasub has done a great job of describing in detail what's going on.

CecileRobertMichon: so for other services we've refactored the reconcile logic to use a common CreateResource func which returns either 1) [the existing resource](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/services/async/async.go#L118) if no update was made or 2) [the resulting resource](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/services/async/async.go#L123) returned from the CreateOrUpdate code. Then each service can do what it needs with resulting resource (eg. update Status)

Looks like we're trying to achieve the same here and that's fine, but I think the code is slightly confusing just because it's not really clear what `managedCluster` is supposed to be since we use the same variable to send the PUT and to receive the return. What do you think of adding a new variable called `result` or maybe reusing the `existingMC` variable to assign the return of CreateOrUpdate functions (then we don't even need this else call)? 
This would achieve the same thing, just read a bit differently.

I'll look into refactoring managed clusters to async soon.

karthikbalasub: @CecileRobertMichon I refactored the code based on your suggestion. PTAL, thanks!

CecileRobertMichon: lgtm

@jackfrancis thoughts?

CecileRobertMichon:
@karthikbalasub could you please squash commits?

karthikbalasub:
> @karthikbalasub could you please squash commits?

Done

azure/services/managedclusters/managedclusters.go
jackfrancis: much cleaner, thanks @karthikbalasub !

jackfrancis (approved):
/lgtm

PR #2146, opened 2022-03-04
Use a list for Azure services in reconcilers

CecileRobertMichon:
**What type of PR is this?**
/kind cleanup

**What this PR does / why we need it**: Since we are reconciling (and deleting in reverse order) Azure services one by one in reconcilers, this is a small optimization to move the reconcile and delete to a list and range through the services instead of making a Reconcile/Delete call for each. This is possible now that we've refactored all services to implement `Azure.Reconciler`. Also adds unit tests.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Use a list for Azure services in AzureCluster and AzureMachine reconcilers
```

### DIFF
--- a/azure/interfaces.go
+++ b/azure/interfaces.go
@@ -24,18 +24,16 @@ import (
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 )
 
-// Reconciler is a generic interface used by components offering a type of service.
-// Example: virtualnetworks service would offer Reconcile/Delete methods.
+// Reconciler is a generic interface for a controller reconciler which has Reconcile and Delete methods.
 type Reconciler interface {
 	Reconcile(ctx context.Context) error
 	Delete(ctx context.Context) error
 }
 
-// CredentialGetter is a Service which knows how to retrieve credentials for an Azure
-// resource in a resource group.
-type CredentialGetter interface {
+// ServiceReconciler is an Azure service reconciler which can reconcile an Azure service.
+type ServiceReconciler interface {
+	Name() string
 	Reconciler
-	GetCredentials(ctx context.Context, group string, cluster string) ([]byte, error)
 }
 
 // Authorizer is an interface which can get the subscription ID, base URI, and authorizer for an Azure service.
--- a/azure/interfaces.go
+++ b/azure/interfaces.go
@@ -33,6 +33,7 @@ type Reconciler interface {
 // ServiceReconciler is an Azure service reconciler which can reconcile an Azure service.
 type ServiceReconciler interface {
 	Name() string
+	IsManaged(ctx context.Context) (bool, error)
 	Reconciler
 }
 
--- a/azure/services/privatedns/privatedns.go
+++ b/azure/services/privatedns/privatedns.go
@@ -28,6 +28,8 @@ import (
 	"sigs.k8s.io/cluster-api-provider-azure/util/tele"
 )
 
+const serviceName = "privatedns"
+
 // Scope defines the scope interface for a private dns service.
 type Scope interface {
 	azure.ClusterDescriber
@@ -46,6 +48,11 @@ func New(scope Scope) *Service {
 		Scope:  scope,
 		client: newClient(scope),
 	}
+}
+
+// Name returns the service name.
+func (s *Service) Name() string {
+	return serviceName
 }
 
 // Reconcile creates or updates the private zone, links it to the vnet, and creates DNS records.
--- a/azure/services/privatedns/privatedns.go
+++ b/azure/services/privatedns/privatedns.go
@@ -226,3 +226,9 @@ func (s *Service) isVnetLinkManaged(ctx context.Context, resourceGroupName, zone
 	tags := converters.MapToTags(zone.Tags)
 	return tags.HasOwned(s.Scope.ClusterName()), nil
 }
+
+// IsManaged returns always returns true.
+// TODO: separate private DNS and VNet links so we can implement the IsManaged method for each.
+func (s *Service) IsManaged(ctx context.Context) (bool, error) {
+	return true, nil
+}
--- a/azure/services/publicips/publicips.go
+++ b/azure/services/publicips/publicips.go
@@ -156,3 +156,8 @@ func (s *Service) isIPManaged(ctx context.Context, ipName string) (bool, error) 
 	tags := converters.MapToTags(ip.Tags)
 	return tags.HasOwned(s.Scope.ClusterName()), nil
 }
+
+// IsManaged returns always returns true as public IPs are managed on a one-by-one basis.
+func (s *Service) IsManaged(ctx context.Context) (bool, error) {
+	return true, nil
+}
--- a/azure/services/publicips/publicips.go
+++ b/azure/services/publicips/publicips.go
@@ -29,6 +29,8 @@ import (
 	"sigs.k8s.io/cluster-api-provider-azure/util/tele"
 )
 
+const serviceName = "publicips"
+
 // PublicIPScope defines the scope interface for a public IP service.
 type PublicIPScope interface {
 	azure.ClusterDescriber
@@ -47,6 +49,11 @@ func New(scope PublicIPScope) *Service {
 		Scope:  scope,
 		Client: NewClient(scope),
 	}
+}
+
+// Name returns the service name.
+func (s *Service) Name() string {
+	return serviceName
 }
 
 // Reconcile gets/creates/updates a public ip.
--- a/controllers/azurecluster_reconciler.go
+++ b/controllers/azurecluster_reconciler.go
@@ -102,26 +102,36 @@ func (s *azureClusterService) Delete(ctx context.Context) error {
 
 	groupSvc, err := s.getService(groups.ServiceName)
 	if err != nil {
-		return err
+		return errors.Wrap(err, "failed to get group service")
 	}
-	if err := groupSvc.Delete(ctx); err != nil {
-		if errors.Is(err, azure.ErrNotOwned) {
-			// If the resource group is not managed we need to delete resources inside the group one by one.
-			// services are deleted in reverse order from the order in which they are reconciled.
-			for i := len(s.services) - 1; i >= 1; i-- {
-				if err := s.services[i].Delete(ctx); err != nil {
-					return errors.Wrapf(err, "failed to delete AzureCluster service %s", s.services[i].Name())
-				}
-			}
-		} else {
+
+	managed, err := groupSvc.IsManaged(ctx)
+	if err != nil {
+		if azure.ResourceNotFound(err) {
+			// If the resource group is not found, there is nothing to delete, return early.
+			return nil
+		}
+		return errors.Wrap(err, "failed to determine if the AzureCluster resource group is managed")
+	}
+	if managed {
+		// if the resource group is managed, we delete the entire resource group directly.
+		if err := groupSvc.Delete(ctx); err != nil {
 			return errors.Wrap(err, "failed to delete resource group")
+		}
+	} else {
+		// If the resource group is not managed we need to delete resources inside the group one by one.
+		// services are deleted in reverse order from the order in which they are reconciled.
+		for i := len(s.services) - 1; i >= 0; i-- {
+			if err := s.services[i].Delete(ctx); err != nil {
+				return errors.Wrapf(err, "failed to delete AzureCluster service %s", s.services[i].Name())
+			}
 		}
 	}
 
 	return nil
 }
 
-func (s *azureClusterService) getService(name string) (azure.Reconciler, error) {
+func (s *azureClusterService) getService(name string) (azure.ServiceReconciler, error) {
 	for _, service := range s.services {
 		if service.Name() == name {
 			return service, nil
--- a/controllers/azurecluster_reconciler.go
+++ b/controllers/azurecluster_reconciler.go
@@ -41,20 +41,11 @@ import (
 
 // azureClusterService is the reconciler called by the AzureCluster controller.
 type azureClusterService struct {
-	scope            *scope.ClusterScope
-	groupsSvc        azure.Reconciler
-	vnetSvc          azure.Reconciler
-	securityGroupSvc azure.Reconciler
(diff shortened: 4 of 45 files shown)


controllers/azurecluster_reconciler.go
```diff
@@ -63,22 +54,23 @@ func newAzureClusterService(scope *scope.ClusterScope) (*azureClusterService, er
...
-		publicIPSvc:      publicips.New(scope),
-		loadBalancerSvc:  loadbalancers.New(scope),
-		privateDNSSvc:    privatedns.New(scope),
-		bastionSvc:       bastionhosts.New(scope),
-		skuCache:         skuCache,
-		peeringsSvc:      vnetpeerings.New(scope),
-		tagsSvc:          tags.New(scope),
+		scope: scope,
+		services: []azure.Reconciler{
+			groups.New(scope),
```
CecileRobertMichon, line 60: this is the same order as the Reconcile() calls below

controllers/azuremachine_reconciler.go
```diff
@@ -60,17 +54,19 @@ func newAzureMachineService(machineScope *scope.MachineScope) (*azureMachineServ
...
-		vmExtensionsSvc:      vmextensions.New(machineScope),
-		availabilitySetsSvc:  availabilitysets.New(machineScope, cache),
-		skuCache:             cache,
+		scope: machineScope,
+		services: []azure.Reconciler{
+			publicips.New(machineScope),
+			inboundnatrules.New(machineScope),
+			networkinterfaces.New(machineScope, cache),
+			availabilitysets.New(machineScope, cache),
+			disks.New(machineScope),
```
CecileRobertMichon, line 63: this is the same order as Reconcile() below except for disks that were not present before in Reconcile (it's a no-op) but needs to be deleted _after_ the VM (hence needs to be before VMs in the list)

devigned (approved):
I love this changeset! This improves the readability of the reconciler so much. 

/lgtm

mboersma (approved):
/lgtm

jackfrancis (approved):
/lgtm

/assign @shysank

controllers/azurecluster_reconciler.go
```diff
@@ -152,46 +102,14 @@ func (s *azureClusterService) Delete(ctx context.Context) error {
 	ctx, _, done := tele.StartSpanWithLogger(ctx, "controllers.azureClusterService.Delete")
 	defer done()
 
-	if err := s.groupsSvc.Delete(ctx); err != nil {
+	if err := s.services[0].Delete(ctx); err != nil {
```
Jont828, line 105: Could we set a variable like `groupSvc` before placing it in the list so we call that explicitly here?

controllers/azurecluster_reconciler.go
```diff
@@ -152,46 +102,14 @@ func (s *azureClusterService) Delete(ctx context.Context) error {
...
-				return errors.Wrap(err, "failed to delete network security group")
-			}
-
-			if err := s.vnetSvc.Delete(ctx); err != nil {
-				return errors.Wrap(err, "failed to delete virtual network")
+			// If the resource group is not managed we need to delete resources inside the group one by one.
+			// services are deleted in reverse order from the order in which they are reconciled.
+			for i := len(s.services) - 1; i >= 1; i-- {
+				if err := s.services[i].Delete(ctx); err != nil {
+					return errors.Wrap(err, "failed to delete AzureCluster service")
```
Jont828, line 111: Can we access the `serviceName` variable from `s.services[i]`? That way we could do something like `errors.Wrap(err, fmt.Sprintf("failed to delete AzureCluster service '%s'", serviceName)`.

controllers/azurecluster_reconciler_test.go
```diff
@@ -30,72 +30,107 @@ import (
 	"sigs.k8s.io/cluster-api-provider-azure/azure/scope"
 	"sigs.k8s.io/cluster-api-provider-azure/azure/services/resourceskus"
 	gomockinternal "sigs.k8s.io/cluster-api-provider-azure/internal/test/matchers/gomock"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 )
 
-type expect func(grp *mock_azure.MockReconcilerMockRecorder, vnet *mock_azure.MockReconcilerMockRecorder, sg *mock_azure.MockReconcilerMockRecorder, rt *mock_azure.MockReconcilerMockRecorder, sn *mock_azure.MockReconcilerMockRecorder, natg *mock_azure.MockReconcilerMockRecorder, pip *mock_azure.MockReconcilerMockRecorder, lb *mock_azure.MockReconcilerMockRecorder, dns *mock_azure.MockReconcilerMockRecorder, bastion *mock_azure.MockReconcilerMockRecorder, peer *mock_azure.MockReconcilerMockRecorder)
```
Jont828, line 35: I'm glad we don't need this giant block anymore!

controllers/azurecluster_reconciler_test.go
```diff
@@ -30,72 +30,107 @@ import (
...
+func TestAzureClusterServiceReconcile(t *testing.T) {
+	cases := map[string]struct {
+		expectedError string
+		expect        func(one *mock_azure.MockReconcilerMockRecorder, two *mock_azure.MockReconcilerMockRecorder, three *mock_azure.MockReconcilerMockRecorder)
```
Jont828, line 39: Is there any significance in having three services in the test cases? Or is it just for the sake of convenience so it'll cover most cases?

controllers/azuremachine_reconciler.go
Jont828: What was our behavior before this change? Would the disk get deleted anyway?

CecileRobertMichon: There was a disk delete call https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2146/files#diff-5a6d880b9a0be00329894fc2f52f15827dd6bdfe6b60aaec660728661023f41fL142 but no disk reconcile call. Disk reconcile is a no-op so adding it does not change behavior.

controllers/azurecluster_reconciler_test.go
CecileRobertMichon: no particular significance just chose a number big enough that it covers the basic edge cases (first, last, middle) but not too big that it makes the tests very long/repetitive and hard to read. 

Here since we're testing code with a for loop, the behavior should be the same whether there are 3, 4, 5, or 50 services. The goal is not to test the behavior of the for loop itself but more that for a given number of items our code is behaving as expected. I chose naming "one" "two" "three" since it makes the ordering aspect of the test very obvious and easy to follow (services are reconciled in order and deleted in reverse).

controllers/azurecluster_reconciler.go
shysank: Or can we make service registration a little bit more elaborate instead of just adding to an array. Something like
```
type RecocilerService struct {
 Name string,
 azure.Reconciler
}

func Register(serviceName, azure.Reconciler) {
}

```

This would help us easily add new apis to control reconciliation flow. For eg. for the case above, we can have a `GetService(name)` which iterates over the services list and returns the one with matching name. 
Or add a `Skip` option to the `ReconcilerService` if we want to skip reconciliation etc. wdyt?

CecileRobertMichon: even if we assign it to a variable it wouldn't be accessible from here unless we add it as a separate field on the `azureClusterService` struct since we only have access to the `s` receiver from this func. we could assign a new variable here to make it explicit that it's what we expect but the variable would only be used once and it would still make the same assumption on the order so not sure that's an improvement, like this:

```
groupSvc := s.services[0]
if err := s.groupSvc.Delete(ctx); err != nil {
```

controllers/azurecluster_reconciler.go
CecileRobertMichon: not currently. Each service in `s.services` is a `azure.Reconciler` which only implements `Reconcile()` and `Delete()`. I actually thought about this as well and considered adding `ServiceName()` to the interface so we could do this but eventually decided not to pursue that route since it requires modifying every single service and the error we get back already contains the service name/type.

controllers/azurecluster_reconciler.go
CecileRobertMichon: @shysank I like the idea of registering the services. How would you set order and iterate over all the services without a list?

Jont828: Could we make a pretty simple struct with an array and a map from a name to its element in the array? Something like
```
type ServiceWrapper struct {
    Services []azure.Reconciler 
    ServiceMap map[string]azure.Reconciler
}

func (s *ServiceWrapper) Add(name string, service azure.Reconciler) {
    s.Services = append(s.Services, service)
    s.ServiceMap[name] = service
}

func (s *ServiceWrapper) Get(name string) azure.Reconciler {
    return s.ServiceMap[name]
}
```

controllers/azurecluster_reconciler.go
Jont828: Where does the error add the service name? It looks like we add the `serviceName` in the errors coming out of `CreateResource()` and `DeleteResource()` in async.go, but if we just return an error like `errors.Errorf("%T is not a network.Subnet", result)` in the reconciler I'm not sure if that will include the service name.

controllers/azurecluster_reconciler.go
shysank: > How would you set order and iterate over all the services without a list?
There will still be some collection backing the services, just that it'd be something more than just []azure.Reonciler. Something like `[]ReconcilerService` (or servicesMap in @Jont828's example). If I want to add a new service, I'll just use the `Register` or `Add` api without having to care about the underlying collection. For ordering, we could start with the order in which the services were registered (add priorities later if needed). We can extend this further by tagging services from which we can extract groups. For example, there can be a group with services independent of each other, and services in that group can be reconciled concurrently while the group itself will be reconciled serially. (shameless plugin of a poc that I did earlier :) ).

Anyways, I'm fine with merging this PR with the current design since there has been so many reviews already, and we can discuss this in a separate issue.

CecileRobertMichon: Let me try something out. Concurrency is out of scope and we should discuss it separately but I think I can at least get a service name added so we can log the service name when there's an error to @Jont828's point below.

controllers/azurecluster_reconciler.go
CecileRobertMichon: this should be solved now, thanks for pointing that out

CecileRobertMichon:
Okay @Jont828 @shysank check out the latest commits. It is a significantly larger refactor but I think it addresses your concerns. I added a new interface called `ServiceReconciler` which is a `Reconciler` that also has `Name()` and `IsManaged()` . The `Name()` func allows us to retrieve the service name for better error logging and for identifying the groups service which avoid assuming ordering. After I did this I realized that we were still assuming ordering by skipping the group delete in the AzureCluster reconciler (loop was going from len-1 to 1 instead of 0). So I added another commit on top that adds `IsManaged` which allows us to check if the group is managed by CAPZ directly from the reconciler. Feel free to tell me this is a scope creep, I kept it unsquashed so we can extract the commit if needed. 

I also included AzureMachinePool and AzureManagedControlPlane reconcilers in the changes since they use the same pattern. 

I think this gets us closer to what we want and we can discuss adding "groups" of services (or priorities, or a dependency graph) if we want to move towards concurrency as a follow-up. The `IsManaged` interface is also something I had wanted to do previously in #1684 to allow us to do caching of the vnet tags and standardize the "is resource managed" story.

azure/interfaces.go
```diff
@@ -24,18 +24,17 @@ import (
...
-// Example: virtualnetworks service would offer Reconcile/Delete methods.
+// Reconciler is a generic interface for a controller reconciler which has Reconcile and Delete methods.
 type Reconciler interface {
 	Reconcile(ctx context.Context) error
 	Delete(ctx context.Context) error
 }
 
-// CredentialGetter is a Service which knows how to retrieve credentials for an Azure
-// resource in a resource group.
-type CredentialGetter interface {
```
CecileRobertMichon, line 36: this wasn't used anymore

azure/services/privatedns/privatedns.go
```diff
@@ -219,3 +226,9 @@ func (s *Service) isVnetLinkManaged(ctx context.Context, resourceGroupName, zone
 	tags := converters.MapToTags(zone.Tags)
 	return tags.HasOwned(s.Scope.ClusterName()), nil
 }
+
+// IsManaged returns always returns true.
+// TODO: separate private DNS and VNet links so we can implement the IsManaged method for each.
```
CecileRobertMichon, line 231: this one is a TODO for after #1715 is merged

azure/services/publicips/publicips.go
```diff
@@ -149,3 +156,8 @@ func (s *Service) isIPManaged(ctx context.Context, ipName string) (bool, error)
 	tags := converters.MapToTags(ip.Tags)
 	return tags.HasOwned(s.Scope.ClusterName()), nil
 }
+
+// IsManaged returns always returns true as public IPs are managed on a one-by-one basis.
```
CecileRobertMichon, line 160: this one is a little awkward. Each public IP can be either managed or unmanaged depending on its own tags and there is no overall "resource type is managed by CAPZ" or not, which makes it different from how the other services behave. Since we're not using this func anywhere (but it's there to satisfy the interface) I opted to always return true since that's the most cautious outcome: we assume public IPs are managed, but then we check on each one individually (https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2146/files#diff-fe943e74189aa531d871d7b774c371755043f5bd1b7e8ab14ad3d2b5fc023301R148). Open to suggestions if anyone has a better idea.

CecileRobertMichon:
looks like e2e tests are failing delete... looking

CecileRobertMichon:
should I rebase or keep the commits separate?

shysank:
> should I squash or keep the commits separate?

I'm fine with keeping the commits as is.

Jont828:
I like the interface changes!
/lgtm

PR #2147, opened 2022-03-04
Update Calico to v3.22.1

CecileRobertMichon:
**What type of PR is this?**
/kind cleanup

**What this PR does / why we need it**: This is a quick fix to unblock the verify test.

next step: find a way to not break the make verify test whenever a new version of calico is released. Because the source manifest used in our calico kustomization file as source manifest (https://docs.projectcalico.org/v3.22/manifests/calico-vxlan.yaml) is for a minor release of calico and not a specific release, it gets updated each time a patch is released and this cannot be used as a reliable source because it is mutable. We need to find a way to pin to a specific release while keeping the ability to update the manifest easily by just changing the version when needed.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Update Calico to v3.22.1
```

### DIFF
--- a/templates/addons/calico-ipv6.yaml
+++ b/templates/addons/calico-ipv6.yaml
@@ -4007,7 +4007,7 @@ spec:
           value: node
         - name: DATASTORE_TYPE
           value: kubernetes
-        image: docker.io/calico/kube-controllers:v3.22.0
+        image: docker.io/calico/kube-controllers:v3.22.1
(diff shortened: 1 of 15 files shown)


CecileRobertMichon:
updated the Windows images as well 

/assign @jsturtevant

PR #2149, opened 2022-03-04
Fetch calico manifests from release artifacts

CecileRobertMichon:
**What type of PR is this?**

**What this PR does / why we need it**: Currently, we fetch Calico manifest from https://docs.projectcalico.org/v3.22/manifests/ with kustomize and add patches to make it work for Azure (since #2021). The issue is that https://docs.projectcalico.org/v3.22/manifests/ is not pinned to a specific release, so the manifests are mutable. This means `make verify` will start failing whenever a new release of Calico causes the manifest to be updated and we'll be forced to update Calico to fix it. To be able to update the version of Calico on our own schedule, we need another level of indirection so we keep a copy of the source manifest in the repo and use it with kustomize to generate the Calico spec.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Fetch calico manifests from release artifacts
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -76,3 +76,6 @@ hack/boilerplate/*.pyc
 
 # release notes
 _releasenotes
+
+# calico manifests archive
+release-*/manifests/calico-*.yaml
--- a/Makefile
+++ b/Makefile
@@ -451,10 +451,20 @@ generate-e2e-templates: $(KUSTOMIZE) ## Generate Azure infrastructure templates 
 	$(KUSTOMIZE) build $(AZURE_TEMPLATES)/v1beta1/cluster-template-kcp-scale-in --load-restrictor LoadRestrictionsNone > $(AZURE_TEMPLATES)/v1beta1/cluster-template-kcp-scale-in.yaml
 
 .PHONY: generate-addons
-generate-addons: ## Generate metric-server, calico calico-ipv6 addons.
+generate-addons: fetch-calico-manifests ## Generate metric-server, calico calico-ipv6 addons.
 	$(KUSTOMIZE) build $(ADDONS_DIR)/metrics-server > $(ADDONS_DIR)/metrics-server/metrics-server.yaml
 	$(KUSTOMIZE) build $(ADDONS_DIR)/calico > $(ADDONS_DIR)/calico.yaml
 	$(KUSTOMIZE) build $(ADDONS_DIR)/calico-ipv6 > $(ADDONS_DIR)/calico-ipv6.yaml
+
+# When updating this, make sure to also update the Windows image version in templates/addons/windows/calico.
+CALICO_VERSION := v3.22.1
+
+.PHONY: fetch-calico-manifests
+fetch-calico-manifests: ## Get Calico release manifests and unzip them.
+	@echo "Fetching Calico release manifests from release artifacts, this might take a minute..."
+	wget -qO- https://github.com/projectcalico/calico/releases/download/$(CALICO_VERSION)/release-$(CALICO_VERSION).tgz | tar xz release-$(CALICO_VERSION)/manifests/calico-vxlan.yaml release-$(CALICO_VERSION)/manifests/calico-policy-only.yaml
+	mv release-$(CALICO_VERSION)/manifests/calico-vxlan.yaml $(ADDONS_DIR)/calico
+	mv release-$(CALICO_VERSION)/manifests/calico-policy-only.yaml $(ADDONS_DIR)/calico-ipv6
 
 .PHONY: modules
 modules: ## Runs go mod tidy to ensure proper vendoring.
--- a/templates/addons/calico-ipv6/calico-policy-only.yaml
+++ b/templates/addons/calico-ipv6/calico-policy-only.yaml
@@ -1,0 +1,4485 @@
+---
+# Source: calico/templates/calico-config.yaml
+# This ConfigMap is used to configure a self-hosted Calico installation.
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: calico-config
+  namespace: kube-system
+data:
+  # You must set a non-zero value for Typha replicas below.
+  typha_service_name: "calico-typha"
+
+  # Configure the MTU to use for workload interfaces and tunnels.
+  # By default, MTU is auto-detected, and explicitly setting this field should not be required.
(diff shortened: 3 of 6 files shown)


.gitignore
```diff
@@ -76,3 +76,6 @@ hack/boilerplate/*.pyc
 
 # release notes
 _releasenotes
+
+# calico manifests archive
+release-*/manifests/calico-*.yaml
```
jsturtevant, line 81: is this needed?

Makefile
```diff
@@ -451,11 +451,21 @@ generate-e2e-templates: $(KUSTOMIZE) ## Generate Azure infrastructure templates
...
 	$(KUSTOMIZE) build $(ADDONS_DIR)/metrics-server > $(ADDONS_DIR)/metrics-server/metrics-server.yaml
 	$(KUSTOMIZE) build $(ADDONS_DIR)/calico > $(ADDONS_DIR)/calico.yaml
 	$(KUSTOMIZE) build $(ADDONS_DIR)/calico-ipv6 > $(ADDONS_DIR)/calico-ipv6.yaml
 
+# When updating this, make sure to also update the Windows image version in templates/addons/windows/calico.
+CALICO_VERSION := v3.22.1
+
+.PHONY: fetch-calico-manifests
+fetch-calico-manifests: ## Get Calico release manifests and unzip them.
+	@echo "Fetching Calico release manifests from release artifacts, this might take a minute..."
```
jsturtevant, line 464: does this get fetched everytime? can us a file locally prevent that?

.gitignore
CecileRobertMichon: it's the temp location where the file is put when it's extracted, making sure it doesn't get checked in if the make target is interrupted before it gets moved for some reason

not 100% needed but more of a safety net, should I get rid of it?

Makefile
CecileRobertMichon: Everytime. The issue I'm trying to deal with is having to fetch it everytime vs. making sure generated files aren't outdated. If we don't fetch it how do we know the base manifest wasn't manually edited?

.gitignore
jsturtevant: thanks, makes sense to keep

Makefile
jsturtevant: can we drop a symlink to a versioned file name, to check if it is been downloaded and do a diff with previous version on the file to determine if it was changed?

CecileRobertMichon: > do a diff with previous version on the file 

Are you saying looking at a previous git version to compare the file? our CI just checks out the branch it's running on, we don't have any precedent with comparing things in previous versions... If I make it compare with "main" it wouldn't be accurate once this gets put into a release branch if we want to do a backport and the main file has changed.

CecileRobertMichon:
Gathering make verify duration data

/test pull-cluster-api-provider-azure-verify

Makefile
CecileRobertMichon: looking at the durations in job history for make verify (https://prow.k8s.io/job-history/gs/kubernetes-jenkins/pr-logs/directory/pull-cluster-api-provider-azure-verify) it doesn't look like this is adding that much extra time compared to previous runs, we're talking about ~ 1 minute difference so if we can't find a way around it I think we can take the hit

CecileRobertMichon:
Gathering make verify duration data

/test pull-cluster-api-provider-azure-verify

Makefile
jsturtevant: I was thinking from a developer experience perspective since generate flavors calls this: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/35782aad60f36cdac43260649692dd8ec3b29741/Makefile#L437

CecileRobertMichon: Here's the diff on my machine (it's pretty slow so might be faster for others):

```
➜  cluster-api-provider-azure git:(main) ✗ time make generate-flavors
/Users/cerobert/go/src/sigs.k8s.io/cluster-api-provider-azure/hack/tools/bin/kustomize-v4.5.2 build templates/addons/metrics-server > templates/addons/metrics-server/metrics-server.yaml
/Users/cerobert/go/src/sigs.k8s.io/cluster-api-provider-azure/hack/tools/bin/kustomize-v4.5.2 build templates/addons/calico > templates/addons/calico.yaml
/Users/cerobert/go/src/sigs.k8s.io/cluster-api-provider-azure/hack/tools/bin/kustomize-v4.5.2 build templates/addons/calico-ipv6 > templates/addons/calico-ipv6.yaml
./hack/gen-flavors.sh
make generate-flavors  37.28s user 2.41s system 98% cpu 40.261 total


➜  cluster-api-provider-azure git:(fix-calico-fetch) time make generate-flavors
Fetching Calico release manifests from release artifacts, this might take a minute...
wget -qO- https://github.com/projectcalico/calico/releases/download/v3.22.1/release-v3.22.1.tgz | tar xz release-v3.22.1/manifests/calico-vxlan.yaml release-v3.22.1/manifests/calico-policy-only.yaml
mv release-v3.22.1/manifests/calico-vxlan.yaml templates/addons/calico
mv release-v3.22.1/manifests/calico-policy-only.yaml templates/addons/calico-ipv6
/Users/cerobert/go/src/sigs.k8s.io/cluster-api-provider-azure/hack/tools/bin/kustomize-v4.5.2 build templates/addons/metrics-server > templates/addons/metrics-server/metrics-server.yaml
/Users/cerobert/go/src/sigs.k8s.io/cluster-api-provider-azure/hack/tools/bin/kustomize-v4.5.2 build templates/addons/calico > templates/addons/calico.yaml
/Users/cerobert/go/src/sigs.k8s.io/cluster-api-provider-azure/hack/tools/bin/kustomize-v4.5.2 build templates/addons/calico-ipv6 > templates/addons/calico-ipv6.yaml
./hack/gen-flavors.sh
make generate-flavors  43.69s user 4.54s system 49% cpu 1:36.70 total
```

I was using the `make verify` job for comparison because it runs a full `make generate` (which includes `make generate-flavors`)

CecileRobertMichon:
rebased on top of #2147

mboersma (approved):
/lgtm

CecileRobertMichon:
@jsturtevant thoughts?

PR #1994, opened 2022-01-21
Templates and E2E tests use external cloud-provider-azure by default

jackfrancis:
**What type of PR is this?**

**What this PR does / why we need it**:

Derived from https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1889

This PR re-orients test and reference templates so that external cloud-provider-azure is the "default" for all other templates, and that "in-tree" cloud-provider-azure configuration is only tested when that specific scenario is required.

The reason for this is twofold:

1) The Kubernetes community has agreed that all future cloud provider development (across all providers, not specifically Azure) will be maintained separate from the main upstream Kubernetes code. So this PR re-orients the capz tests and reference templates towards that community best-practice. More information here:

- https://kubernetes.io/blog/2019/04/17/the-future-of-cloud-providers-in-kubernetes/#in-tree-out-of-tree-providers

2) The development of new Azure cloud provider features and, more importantly, bug fixes, is prioritized for delivery in the external cloud-provider-azure project:

- https://github.com/kubernetes-sigs/cloud-provider-azure

As a result of the complexity of the years-long transition away from in-tree (i.e., cloud provider bits being tightly coupled to Kubernetes versions and delivered in common _with_ Kubernetes versions) cloud-provider-azure there are now material differences in the actual cloud-provider-azure runtime that is used depending on whether or not you are using in-tree or external cloud-provider-azure. The differences can be summarized most simply thusly: the external cloud-provider-azure is more stable (more frequently tested, more confirmed fixes actually included).

For the above reasons, we want to make it easier to rapidly test capz against external cloud-provider-azure, and to encourage capz users to create clusters running the external cloud-provider-azure components.

An important note for capz users: the capz reference templates use the experimental cluster-api `ClusterResourceSet` CRD. In order to use `ClusterResourceSet` resources in your capz templates, you must enable that feature flag in your cluster-api environment (management cluster).

Reference:

- https://cluster-api.sigs.k8s.io/tasks/experimental-features/cluster-resource-set.html

An example of how to enable `ClusterResourceSet` on your cluster-api management cluster is here:

- https://github.com/kubernetes-sigs/cluster-api/blob/v1.1.3/docs/book/src/tasks/experimental-features/experimental-features.md#enabling-experimental-features-on-existing-management-clusters

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

[...]

PR #2154, opened 2022-03-09
feat: add dual-stack support

### LINKED ISSUE #1266: Support ipv6 dual stack
/kind feature

#646 added ipv6 single stack support. 

This issue is to track adding IPv6 dual-stack support (https://kubernetes.io/docs/concepts/services-networking/dual-stack/) and enabling e2e testing for dual-stack clusters.

for prior Azure + dual stack art, refer to https://github.com/Azure/aks-engine/pulls?q=is%3Apr+dual+stack+is%3Aclosed

**Environment:**

aramase:
**What type of PR is this?**

/kind feature

- Adds dual-stack support
  - Added template and test CI flavors

```bash
➜ kubectl --kubeconfig=/tmp/capi-dual-stack.kubeconfig get nodes -owide
NAME                                  STATUS   ROLES                  AGE   VERSION   INTERNAL-IP   EXTERNAL-IP   OS-IMAGE             KERNEL-VERSION      CONTAINER-RUNTIME
capi-dual-stack-control-plane-g7xtb   Ready    control-plane,master   41m   v1.23.3   10.0.0.4      <none>        Ubuntu 20.04.3 LTS   5.11.0-1027-azure   containerd://1.5.7
capi-dual-stack-md-0-j96nr            Ready    <none>                 39m   v1.23.3   10.1.0.4      <none>        Ubuntu 20.04.3 LTS   5.11.0-1027-azure   containerd://1.5.7
capi-dual-stack-md-0-jjg68            Ready    <none>                 39m   v1.23.3   10.1.0.5      <none>        Ubuntu 20.04.3 LTS   5.11.0-1027-azure   containerd://1.5.7
```

Node has 2 internal IPs, one from each family
```bash
➜ kubectl --kubeconfig=/tmp/capi-dual-stack.kubeconfig get node capi-dual-stack-md-0-j96nr -o go-template --template='{{range .status.addresses}}{{printf "%s: %s \n" .type .address}}{{end}}'
Hostname: capi-dual-stack-md-0-j96nr 
InternalIP: 10.1.0.4 
InternalIP: 2001:1234:5678:9abd::4 
```

Node has 2 PodCIDRs, one from each family
```bash
➜ kubectl --kubeconfig=/tmp/capi-dual-stack.kubeconfig get node capi-dual-stack-md-0-j96nr -o go-template --template='{{range .spec.podCIDRs}}{{printf "%s\n" .}}{{end}}'
10.244.2.0/24
2001:1234:5678:9a42::/64
```

Pod has 2 PodIP, one from each family
```bash
➜ kubectl --kubeconfig=/tmp/capi-dual-stack.kubeconfig get pods nginx-7f76bfdf58-ftmb6 -o go-template --template='{{range .status.podIPs}}{{printf "%s \n" .ip}}{{end}}' 
10.244.2.37 
2001:1234:5678:9a42::25 
```

Fixes #1266 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Add dual-stack support
```

### DIFF
--- a/docs/book/src/topics/dual-stack.md
+++ b/docs/book/src/topics/dual-stack.md
@@ -1,0 +1,69 @@
+# Dual-stack clusters
+
+## Overview
+
+CAPZ enables you to create [dual-stack](https://kubernetes.io/docs/concepts/services-networking/dual-stack/) Kubernetes cluster on Microsoft Azure.
+
+- Dual-stack support is available for Kubernetes version 1.21.0 and later on Azure.
+
+To deploy a cluster using dual-stack, use the [dual-stack flavor template](../../../../templates/cluster-template-dual-stack.yaml).
+
+Things to try out after the cluster created:
+
+- Nodes have 2 internal IPs, one from each IP family.
+
+```bash
+kubectl get node <node name> -o go-template --template='{{range .status.addresses}}{{printf "%s: %s \n" .type .address}}{{end}}'
+Hostname: capi-dual-stack-md-0-j96nr 
+InternalIP: 10.1.0.4 
+InternalIP: 2001:1234:5678:9abd::4 
+```
+
+- Nodes have 2 `PodCIDRs`, one from each IP family.
+
+```bash
+kubectl get node <node name> -o go-template --template='{{range .spec.podCIDRs}}{{printf "%s\n" .}}{{end}}'
+10.244.2.0/24
+2001:1234:5678:9a42::/64
+```
+
+- Pods have 2 `PodIP`, one from each IP family.
+
+```bash
+kubectl get pods <pod name> -o go-template --template='{{range .status.podIPs}}{{printf "%s \n" .ip}}{{end}}' 
+10.244.2.37 
+2001:1234:5678:9a42::25 
+```
+
+- Able to reach other pods in cluster using IPv4 and IPv6.
+
+```bash
+# inside the nginx-pod
+/ # ifconfig eth0
+eth0      Link encap:Ethernet  HWaddr 8A:B2:32:92:4F:87
+          inet addr:10.244.2.2  Bcast:0.0.0.0  Mask:255.255.255.255
+          inet6 addr: 2001:1234:5678:9a42::2/128 Scope:Global
+          inet6 addr: fe80::88b2:32ff:fe92:4f87/64 Scope:Link
+          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
+          RX packets:9 errors:0 dropped:0 overruns:0 frame:0
+          TX packets:10 errors:0 dropped:1 overruns:0 carrier:0
+          collisions:0 txqueuelen:0
+          RX bytes:906 (906.0 B)  TX bytes:840 (840.0 B)
+
+/ # ping -c 2 10.244.1.2
+PING 10.244.1.2 (10.244.1.2): 56 data bytes
+64 bytes from 10.244.1.2: seq=0 ttl=62 time=1.366 ms
+64 bytes from 10.244.1.2: seq=1 ttl=62 time=1.396 ms
+
+--- 10.244.1.2 ping statistics ---
+2 packets transmitted, 2 packets received, 0% packet loss
+round-trip min/avg/max = 1.366/1.381/1.396 ms
+/ # ping -c 2 2001:1234:5678:9a41::2
+PING 2001:1234:5678:9a41::2 (2001:1234:5678:9a41::2): 56 data bytes
+64 bytes from 2001:1234:5678:9a41::2: seq=0 ttl=62 time=1.264 ms
+64 bytes from 2001:1234:5678:9a41::2: seq=1 ttl=62 time=1.233 ms
+
+--- 2001:1234:5678:9a41::2 ping statistics ---
+2 packets transmitted, 2 packets received, 0% packet loss
+round-trip min/avg/max = 1.233/1.248/1.264 ms
+```
--- a/templates/addons/calico-dual-stack/kustomization.yaml
+++ b/templates/addons/calico-dual-stack/kustomization.yaml
@@ -1,0 +1,6 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+    - ../calico-ipv6
+patchesStrategicMerge:
+    - patches/calico-config.yaml
--- a/templates/test/ci/cluster-template-prow-dual-stack.yaml
+++ b/templates/test/ci/cluster-template-prow-dual-stack.yaml
@@ -1,0 +1,4637 @@
+apiVersion: cluster.x-k8s.io/v1beta1
+kind: Cluster
+metadata:
+  labels:
+    cni: ${CLUSTER_NAME}-calico
+  name: ${CLUSTER_NAME}
+  namespace: default
+spec:
+  clusterNetwork:
+    pods:
+      cidrBlocks:
+      - 10.244.0.0/16
+      - 2001:1234:5678:9a40::/58
+    services:
+      cidrBlocks:
+      - 10.0.0.0/16
+      - fd00::/108
+  controlPlaneRef:
+    apiVersion: controlplane.cluster.x-k8s.io/v1beta1
+    kind: KubeadmControlPlane
+    name: ${CLUSTER_NAME}-control-plane
+  infrastructureRef:
+    apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
+    kind: AzureCluster
+    name: ${CLUSTER_NAME}
+---
+apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
+kind: AzureCluster
+metadata:
+  name: ${CLUSTER_NAME}
+  namespace: default
+spec:
+  additionalTags:
+    buildProvenance: ${BUILD_PROVENANCE}
+    creationTimestamp: ${TIMESTAMP}
+    jobName: ${JOB_NAME}
+  identityRef:
+    apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
+    kind: AzureClusterIdentity
+    name: ${CLUSTER_IDENTITY_NAME}
+  location: ${AZURE_LOCATION}
+  networkSpec:
+    subnets:
+    - cidrBlocks:
+      - 10.0.0.0/16
+      - 2001:1234:5678:9abc::/64
+      name: control-plane-subnet
+      role: control-plane
+    - cidrBlocks:
+      - 10.1.0.0/16
+      - 2001:1234:5678:9abd::/64
+      name: node-subnet
+      role: node
+    vnet:
+      cidrBlocks:
+      - 10.0.0.0/8
+      - 2001:1234:5678:9a00::/56
+      name: ${AZURE_VNET_NAME:=${CLUSTER_NAME}-vnet}
+  resourceGroup: ${AZURE_RESOURCE_GROUP:=${CLUSTER_NAME}}
+  subscriptionID: ${AZURE_SUBSCRIPTION_ID}
+---
+apiVersion: controlplane.cluster.x-k8s.io/v1beta1
+kind: KubeadmControlPlane
+metadata:
+  name: ${CLUSTER_NAME}-control-plane
+  namespace: default
+spec:
+  kubeadmConfigSpec:
+    clusterConfiguration:
+      apiServer:
+        extraArgs:
+          cloud-config: /etc/kubernetes/azure.json
+          cloud-provider: azure
+        extraVolumes:
+        - hostPath: /etc/kubernetes/azure.json
+          mountPath: /etc/kubernetes/azure.json
+          name: cloud-config
+          readOnly: true
+        timeoutForControlPlane: 20m
+      controllerManager:
+        extraArgs:
+          allocate-node-cidrs: "true"
+          cloud-config: /etc/kubernetes/azure.json
+          cloud-provider: azure
+          cluster-cidr: 10.244.0.0/16,2001:1234:5678:9a40::/58
+          cluster-name: ${CLUSTER_NAME}
+          configure-cloud-routes: "true"
+          v: "4"
+        extraVolumes:
+        - hostPath: /etc/kubernetes/azure.json
+          mountPath: /etc/kubernetes/azure.json
+          name: cloud-config
+          readOnly: true
+      etcd:
+        local:
+          dataDir: /var/lib/etcddisk/etcd
+          extraArgs:
+            quota-backend-bytes: "8589934592"
+    diskSetup:
+      filesystems:
+      - device: /dev/disk/azure/scsi1/lun0
+        extraOpts:
+        - -E
+        - lazy_itable_init=1,lazy_journal_init=1
+        filesystem: ext4
+        label: etcd_disk
+      - device: ephemeral0.1
+        filesystem: ext4
+        label: ephemeral0
+        replaceFS: ntfs
+      partitions:
+      - device: /dev/disk/azure/scsi1/lun0
+        layout: true
+        overwrite: false
+        tableType: gpt
+    files:
+    - contentFrom:
+        secret:
+          key: control-plane-azure.json
+          name: ${CLUSTER_NAME}-control-plane-azure-json
+      owner: root:root
+      path: /etc/kubernetes/azure.json
+      permissions: "0644"
+    initConfiguration:
+      localAPIEndpoint:
+        bindPort: 6443
+      nodeRegistration:
+        kubeletExtraArgs:
+          azure-container-registry-config: /etc/kubernetes/azure.json
+          cloud-config: /etc/kubernetes/azure.json
+          cloud-provider: azure
+        name: '{{ ds.meta_data["local_hostname"] }}'
+    joinConfiguration:
+      controlPlane:
+        localAPIEndpoint:
+          bindPort: 6443
+      nodeRegistration:
+        kubeletExtraArgs:
+          azure-container-registry-config: /etc/kubernetes/azure.json
+          cloud-config: /etc/kubernetes/azure.json
+          cloud-provider: azure
+        name: '{{ ds.meta_data["local_hostname"] }}'
+    mounts:
+    - - LABEL=etcd_disk
+      - /var/lib/etcddisk
+    postKubeadmCommands:
+    - sed -i '\#--listen-client-urls#s#$#,https://127.0.0.1:2379#' /etc/kubernetes/manifests/etcd.yaml
+    - echo "DNSStubListener=no" >> /etc/systemd/resolved.conf
+    - mv /etc/resolv.conf /etc/resolv.conf.OLD && ln -s /run/systemd/resolve/resolv.conf
+      /etc/resolv.conf
+    - systemctl restart systemd-resolved
+    preKubeadmCommands: []
+  machineTemplate:
+    infrastructureRef:
+      apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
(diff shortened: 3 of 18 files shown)


aramase:
Ran the dual-stack e2e tests from k/k to confirm cluster is functional. Test results are available [here](https://gist.github.com/aramase/420d90e35467af5d8b2c862e3af0596a)

```bash
➜ _output/local/bin/linux/amd64/e2e.test --ginkgo.focus="Feature:IPv6DualStack"
...
Ran 22 of 7119 Specs in 789.057 seconds
SUCCESS! -- 22 Passed | 0 Failed | 0 Pending | 7097 Skipped
PASS
```

templates/addons/calico-dual-stack/kustomization.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+    - https://docs.projectcalico.org/v3.22/manifests/calico-policy-only.yaml
```
CecileRobertMichon, line 4: let's change this to a static file since the manifest isn't immutable, see #2149 

we're already getting calico-policy-only.yaml for ipv6 so you can just reuse the same one

aramase: Do you recommend copying `calico-policy-only.yaml` to calico-dual-stack? I don't think we can reference it in kustomization.yaml from capi-ipv6 with a relative path from a different directory. AFAIK it only supports current or child directories.

CecileRobertMichon: you can either copy `calico-policy-only.yaml` to calico-dual-stack or use ../calico-ipv6 as the base for dual stack and then patch on top of it (then you can also get rid of the azure-mtu patch)

aramase:
> Where are we going to run periodic tests for this? Should we also run a CAPZ PR test (could be optional if we don't want to run it on every PR) to validate the dual-stack flavor?

Currently we're running periodic jobs for validating dual-stack in k/k using aks-engine. After we merge this PR and have a release, we'll migrate https://testgrid.k8s.io/sig-network-dualstack-azure-e2e#dualstack-azure-e2e-master-iptables and https://testgrid.k8s.io/sig-network-dualstack-azure-e2e#dualstack-azure-e2e-master-ipvs to use CAPZ. 

+1 for running an optional PR test for dual-stack flavor validation. Do we have any similar jobs today? Do we want to just validate if the flavor produces a functional cluster?

jsturtevant:
> Where are we going to run periodic tests for this?

the aks-engine test runs at https://testgrid.k8s.io/provider-azure-master-signal#dualstack-azure-e2e-master-iptables

@aramase is it possible to cover the IPVS case too in capz (not with this PR)? Ithink that is required to be off aks-enigne https://testgrid.k8s.io/provider-azure-master-signal#dualstack-azure-e2e-master-ipvs

aramase:
Opened https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2161 to fix the lint failure

CecileRobertMichon:
Here's the single stack equivalent test: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/test/e2e/azure_test.go#L270

it's marked as `[REQUIRED]` so it runs on every PR. Here's an example of an optional test: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/test/e2e/azure_test.go#L496

In order to use the flavor in the test framework you need to reference the template in azure-dev.yaml like this: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/test/e2e/config/azure-dev.yaml#L150

Ideally the test would 1) verify that the cluster build successfully and possibly 2) expose a dual stack LB service like we're doing for ipv6

test/e2e/azure_test.go
```diff
@@ -540,4 +540,56 @@ var _ = Describe("Workload cluster creation", func() {
 			})
 		})
 	})
+
+	Context("Creating a dual-stack cluster [REQUIRED]", func() {
```
aramase, line 544: Set it as `[REQUIRED]` because we're going to use CAPZ for k/k dual-stack tests and it'll be good to have it tested with every PR

aramase:
Thanks for the pointers @CecileRobertMichon! I've added the tests.

PR is ready for another round of review!

docs/book/src/topics/dual-stack.md
```diff
@@ -0,0 +1,67 @@
+# Dual-stack clusters
+
+## Overview
+
+CAPZ enables you to create [dual-stack](https://github.com/Azure/aks-engine/tree/master/examples/dualstack) Kubernetes cluster on Microsoft Azure.
```
CecileRobertMichon, line 5: why are we pointing to AKS Engine docs here? should we point to official k8s docs instead?

docs/book/src/topics/dual-stack.md
```diff
@@ -0,0 +1,67 @@
...
+
+To deploy a cluster using dual-stack, use the [dual-stack flavor template](https://raw.githubusercontent.com/kubernetes-sigs/cluster-api-provider-azure/main/templates/cluster-template-dual-stack.yaml).
```
CecileRobertMichon, line 7: this is going to fail the lint check because that file doesn't exist in main yet

templates/addons/calico-dual-stack/kustomization.yaml
```diff
@@ -0,0 +1,19 @@
...
+    - patches/calico-config.yaml
+patches:
+- target: 
+    group: apps
+    version: v1
+    kind: DaemonSet
+    name: calico-node
+    namespace: kube-system
+  path: patches/calico-node.yaml
+- path: patches/control-plane-tolerations.yaml
```
CecileRobertMichon, line 15: is this needed? I believe the ipv6 template already has the tolerations

test/e2e/azure_test.go
CecileRobertMichon: is it necessary to test both IPv6 and dual stack on every PR? Trying to keep the number of full cluster buildouts limited so we can keep the presubmit prow job duration as low as possible

aramase: If we only want to test one of it, I would prefer we test dual-stack. Dual-stack config will be more commonly used compared to single-stack IPv6.

Are optional tests run periodically?

templates/addons/calico-dual-stack/kustomization.yaml
aramase: Good point! Let me remove it. I think we can also get rid of `calico-node` because it's the same for ipv6

docs/book/src/topics/dual-stack.md
```diff
@@ -0,0 +1,67 @@
...
+
+Things to try out after the cluster created:
+
+- Nodes have 2 internal IPs, one from each IP family.
+
+```bash
+kubectl --kubeconfig=/tmp/capi-dual-stack.kubeconfig get node capi-dual-stack-md-0-j96nr -o go-template --template='{{range .status.addresses}}{{printf "%s: %s \n" .type .address}}{{end}}'
```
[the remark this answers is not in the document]
aramase: It was beta in `1.21` and GA in `1.23`. The feature flag is enabled by default in `1.21`. Added a note about it in the docs.

test/e2e/azure_test.go
CecileRobertMichon: yes the full suite of tests runs periodically https://testgrid.k8s.io/sig-cluster-lifecycle-cluster-api-provider-azure#capz-periodic-e2e-full-main

CecileRobertMichon:
@jackfrancis this PR is going to conflict with #1994 FYI

test/e2e/azure_test.go
aramase: perfect! i'll move dual-stack to `[OPTIONAL]` for now as it's a new test. We can decide at a later if we want to make this required and make ipv6 optional.

Based on the test run from last commit, we know the dual-stack ci flavor + tests work, so that's a good sign!

aramase:
lmk if you want me to squash the commits, or we can also do `/label tide/merge-method-squash`

CecileRobertMichon:
Great! You can use this command to trigger optional tests on PRs:

CecileRobertMichon:
yes please squash. For some reason the tide label doesn't play well with the k8s release notes automation.

/assign @Jont828 @jsturtevant

docs/book/src/topics/dual-stack.md
```diff
@@ -0,0 +1,69 @@
...
+```bash
+kubectl get node capi-dual-stack-md-0-j96nr -o go-template --template='{{range .spec.podCIDRs}}{{printf "%s\n" .}}{{end}}'
+10.244.2.0/24
+2001:1234:5678:9a42::/64
+```
+
+- Pods have 2 `PodIP`, one from each IP family.
+
+```bash
+kubectl get pods nginx-7f76bfdf58-ftmb6 -o go-template --template='{{range .status.podIPs}}{{printf "%s \n" .ip}}{{end}}'
```
Jont828, line 33: Small thing: since these commands can be copy/pasted for users to try, we can change the name of a pod to something like `<pod-name>` to fit w/ doc styling.

templates/test/ci/cluster-template-prow-dual-stack.yaml
```diff
@@ -0,0 +1,4637 @@
...
+metadata:
+  labels:
+    cni: ${CLUSTER_NAME}-calico
+  name: ${CLUSTER_NAME}
+  namespace: default
+spec:
+  clusterNetwork:
+    pods:
+      cidrBlocks:
+      - 10.244.0.0/16
```
Jont828, line 12: If we're using these for e2e tests, can we create env variables to substitute the CIDRs like in the custom-vnet and private cluster templates?

aramase: Do you have an example? I don't see environment variables for CIDS in the custom-vnet or private cluster prow templates.

aramase: Do we have a scenario where we would need to overwrite the CIDR block for the tests?

docs/book/src/topics/dual-stack.md
aramase: done!

test/e2e/azure_machinepool_drain.go
```diff
@@ -62,6 +63,7 @@ type (
 		ClusterName           string
 		SkipCleanup           bool
 		IPv6                  bool
+		IPFamilies            []corev1.IPFamily
```
CecileRobertMichon, line 66: do we need both IPv6 and IPFamilies here? Seems like they are both conveying the same info, could we get rid of the IPv6 bool and deduct it from the IPFamilies input?

docs/book/src/topics/dual-stack.md
```diff
@@ -0,0 +1,69 @@
+# Dual-stack clusters
```
CecileRobertMichon, line 1: Does it make sense to have a separate page for dual stack or should we put dual stack and ipv6 in the same place?

CecileRobertMichon (approved):
overall lgtm, added two minor comments

test/e2e/azure_machinepool_drain.go
aramase: If these tests are for 1.20+, I think we can rely on `IPFamilies[0]`. I've pushed an update to test it.

docs/book/src/topics/dual-stack.md
aramase: IMO a distinct page for dual-stack makes sense. Dual-stack is a blessed configuration in k/k which suggests dual ips for nodes, pods and services. Single-stack IPv6 would be all single stack, however the nodes having dual IPs (in case of Azure) is a cloud provider specific implementation detail. If we keep the docs distinct, it'll be easier to follow, validate and update in the future.

test/e2e/azure_lb.go
```diff
@@ -122,10 +123,10 @@ func AzureLBSpec(ctx context.Context, inputGetter func() AzureLBSpecInput) {
 
 	// TODO: fix and enable this. Internal LBs + IPv6 is currently in preview.
 	// https://docs.microsoft.com/en-us/azure/virtual-network/ipv6-dual-stack-standard-internal-load-balancer-powershell
-	if !input.IPv6 {
+	if !isIPv6 {
```
aramase, line 126: I think we should be able to remove this check as IPv6 ILB is supported: https://github.com/kubernetes-sigs/cloud-provider-azure/pull/703. I can remove this check in a follow-up.

CecileRobertMichon (commented):
/lgtm

test/e2e/kubernetes/deployment/deployment.go
```diff
@@ -1,3 +1,4 @@
+//go:build e2e
```
jackfrancis, line 1: I don't think we want your change (maybe your IDE is set for < golang 1.17?)

test/e2e/azure_machinepool_drain.go
```diff
@@ -1,3 +1,4 @@
+//go:build e2e
```
jackfrancis, line 1: ditto below comment on build tag change, I don't think we want this?

test/e2e/kubernetes/deployment/deployment.go
jackfrancis: Right, it's the opposite of what I said.

@CecileRobertMichon @mboersma do we care about these changes leaking in? Or would we rather do them all in one go in a discrete PR? (I assume we want to prefer go 1.17+ idioms going forward)

CecileRobertMichon: it's fine for them to leak in

https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2130 should catch the rest

jackfrancis (approved):
lgtm

pending success of `pull-cluster-api-provider-azure-e2e-optional` test run

lzhecheng:
@aramase is it possible to have a dual-stack OOT cloud-provider template?

CecileRobertMichon:
@lzhecheng we can't add example templates for every single combination of feature out there. Ideally you should maintain your own CI templates (I'm guessing this is for testing) so you have full control on the cluster configurations.

Side note, we need to migrate all example templates to use OOT by default.

CecileRobertMichon:
@lzhecheng there is nothing special about VMSS that should impact CNI/dual stack networking. How are you installing Calico? There is a special values file for Calico dual stack when using Helm (both our e2e test and ci-entrypoint.sh should be using the right one automatically). See https://capz.sigs.k8s.io/topics/addons.html#for-dual-stack-clusters.

lzhecheng:
> @lzhecheng there is nothing special about VMSS that should impact CNI/dual stack networking. How are you installing Calico? There is a special values file for Calico dual stack when using Helm (both our e2e test and ci-entrypoint.sh should be using the right one automatically). See https://capz.sigs.k8s.io/topics/addons.html#for-dual-stack-clusters.
> 
> Here are the logs from an e2e test using Calico with dual-stack: https://storage.googleapis.com/kubernetes-jenkins/logs/periodic-cluster-api-provider-azure-e2e-main/1622655939377106944/artifacts/clusters/capz-e2e-s8orha-dual-stack/calico-system/calico-node-2t6h7/calico-node.log

Thank you for the response. Yes, the calico installation is correct. It got only IPv4 while `CALICO_IPV6POOL_CIDR` is set
```
zhechengli@devbox:~$ kdpo -n calico-system      calico-node-gl4jp
Name:                 calico-node-gl4jp
Namespace:            calico-system
Priority:             2000001000
Priority Class Name:  system-node-critical
Node:                 zhecheng-206-1-mp-0000000/10.1.0.4
Start Time:           Mon, 06 Feb 2023 08:28:23 +0000
Labels:               app.kubernetes.io/name=calico-node
                      controller-revision-hash=85dc7f8d55
                      k8s-app=calico-node
                      pod-template-generation=1
Annotations:          hash.operator.tigera.io/cni-config: 2d19559df12928367220b4d9c76b3883ce3e1755
                      hash.operator.tigera.io/tigera-ca-private: cfb4753a251d65748d10675458f7b80a5b261f09
Status:               Running
IP:                   10.1.0.4
IPs:
  IP:           10.1.0.4
...
    Environment:
      DATASTORE_TYPE:                      kubernetes
      WAIT_FOR_DATASTORE:                  true
      CLUSTER_TYPE:                        k8s,operator
      CALICO_DISABLE_FILE_LOGGING:         false
      FELIX_DEFAULTENDPOINTTOHOSTACTION:   ACCEPT
      FELIX_HEALTHENABLED:                 true
      FELIX_HEALTHPORT:                    9099
      NODENAME:                             (v1:spec.nodeName)
      NAMESPACE:                           calico-system (v1:metadata.namespace)
      FELIX_TYPHAK8SNAMESPACE:             calico-system
      FELIX_TYPHAK8SSERVICENAME:           calico-typha
      FELIX_TYPHACAFILE:                   /etc/pki/tls/certs/tigera-ca-bundle.crt
      FELIX_TYPHACERTFILE:                 /node-certs/tls.crt
      FELIX_TYPHAKEYFILE:                  /node-certs/tls.key
      FIPS_MODE_ENABLED:                   false
      FELIX_TYPHACN:                       typha-server
      CALICO_MANAGE_CNI:                   true
      CALICO_IPV4POOL_CIDR:                10.244.0.0/16
      CALICO_IPV4POOL_IPIP:                Never
      CALICO_IPV4POOL_BLOCK_SIZE:          26
      CALICO_IPV4POOL_NODE_SELECTOR:       all()
      CALICO_IPV4POOL_DISABLE_BGP_EXPORT:  false
      CALICO_IPV6POOL_CIDR:                2001:1234:5678:9a40::/58
```
From portal, the VMSS doesn't have IPv6 address and I think that's the problem. Node doesn't have correct IPv6 address I think.
```
root@zhecheng-206-1-mp-0000000:/# ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host
       valid_lft forever preferred_lft forever
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc prio state UP group default qlen 1000
    link/ether 00:0d:3a:8a:40:7b brd ff:ff:ff:ff:ff:ff
    inet 10.1.0.4/16 metric 100 brd 10.1.255.255 scope global eth0
       valid_lft forever preferred_lft forever
    inet6 fe80::20d:3aff:fe8a:407b/64 scope link
       valid_lft forever preferred_lft forever
```
I checked your example e2e log and I think it is a dual-stack VMAS. https://gcsweb.k8s.io/gcs/kubernetes-jenkins/logs/periodic-cluster-api-provider-azure-e2e-main/1622655939377106944/artifacts/clusters/capz-e2e-s8orha-dual-stack/machines/capz-e2e-s8orha-dual-stack-md-0-5f7f76985d-6nmkw/
Is there any VMSS dual-stack e2e cluster? Thank you!

CecileRobertMichon:
> From portal, the VMSS doesn't have IPv6 address and I think that's the problem. Worker node doesn't have correct IPv6 address I think.

:o 

that could be a bug, can you please file an issue? this closed PR thread is not very visible

lzhecheng:
> > From portal, the VMSS doesn't have IPv6 address and I think that's the problem. Worker node doesn't have correct IPv6 address I think.
> 
> :o
> 
> that could be a bug, can you please file an issue? this closed PR thread is not very visible

Sure.

PR #2157, opened 2022-03-10
E2E: waiting for services to be deleted before proceeding

CecileRobertMichon:
**What type of PR is this?**
/kind failing-test

**What this PR does / why we need it**: following up on [this slack thread](https://kubernetes.slack.com/archives/CEX9HENG7/p1646154943464139), the Delete() client doesn't actually wait for a service to be deleted, it simply triggers the delete. cloud-provider-azure has a Finalizer on the Service resource as it needs to reconcile load balancers before the service can be deleted, which means it can take several minutes (sometimes even more) for the service to actually be gone. However, cloud-provider only reconciles one service at a time. This means before the ELB service starts reconciling, the ILB service first needs to be done deleting. By not waiting for the ILB to fully delete, we run into occasional flakes as the timeout for creating the ELB service also includes deleting the previous service.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/azure_lb.go
+++ b/test/e2e/azure_lb.go
@@ -1,3 +1,4 @@
+//go:build e2e
 // +build e2e
 
 /*
@@ -32,6 +33,7 @@ import (
 
 	"github.com/hashicorp/go-retryablehttp"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	"sigs.k8s.io/cluster-api/test/framework"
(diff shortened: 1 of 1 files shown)


test/e2e/azure_lb.go
```diff
@@ -267,7 +274,12 @@ func AzureLBSpec(ctx context.Context, inputGetter func() AzureLBSpecInput) {
 			return err
 		}
 		return nil
-	}, deleteOperationTimeout, retryableOperationSleepBetweenRetries).Should(Succeed())
+	}, retryableOperationTimeout, retryableOperationSleepBetweenRetries).Should(Succeed())
+	Logf("waiting for the external LB service to be deleted: %s", elbService.Name)
```
CecileRobertMichon, line 278: did this to be consistent but need to check if it adds a lot of time to e2e duration since we don't strictly need to wait on this one (it's the last service to be reconciled)

jackfrancis: Interesting, so we have actually not been validating ELB delete (ILBs and ELBs are ultimately the same Azure LoadBalancer type, so not super critical, but I agree it's best-practice to not make an exception here)

jackfrancis (approved):
/lgtm

jackfrancis:
@CecileRobertMichon see https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2160

CecileRobertMichon (commented):
/approve

PR #2163, opened 2022-03-10
enableUltraSSD for managed AKS clusters

TheNakedZealot:
**What type of PR is this?**

Feature

This PR is to give agentpools for Managed AKS Clusters Ultradisk support at creation. This is a feature that's available to non-managed clusters, and the support is important to those who use the managed solution as the Managed Premium SSDs used by default are much slower in disk latency dependent workloads. Also, the task didn't seem too much in effort!

Fixes #

This is submitted via request from the Slack channel #cluster-api-azure user Ace Eldeib. This particular PR is a resubmission of a previous PR that was having CLA complications with the OG author.

**TODOs**:

- [X] squashed commits
- [X] includes documentation
- [X] adds unit tests

**Release note**:

```release-note
NONE
```


jackfrancis:
@TheNakedZealot you'll have to generate the full api code surface area via `make generate-go` first of all...

alexeldeib:
few minor comments:
- like Jack said, you need to update the generated conversion code using `make generate-go`
- and then I think you'll want to update the conversion logic so the new types/old types are still roundtrippable: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/exp/api/v1alpha4/azuremanagedmachinepool_conversion.go
- I think this field is also immutable in AKS, would you mind checking that and if so add a webhook validation on changing the field value?

exp/api/v1beta1/azuremanagedmachinepool_webhook.go
```diff
@@ -172,6 +172,25 @@ func (m *AzureManagedMachinePool) ValidateUpdate(oldRaw runtime.Object, client c
 		}
 	}
 
+	if old.Spec.EnableUltraSSD != nil {
```
alexeldeib, line 175: I believe going from nil -> non-nil is also invalid?

alexeldeib: a unit test would be great for this too :)

exp/api/v1beta1/azuremanagedmachinepool_webhook_test.go
```diff
@@ -357,6 +357,90 @@ func TestAzureManagedMachinePoolUpdatingWebhook(t *testing.T) {
...
+			},
+			old: &AzureManagedMachinePool{
+				Spec: AzureManagedMachinePoolSpec{
+					EnableUltraSSD: to.BoolPtr(true),
+				},
+			},
+			wantErr: true,
+		},
+		{
+			name: "EnableUltraSSD feature is immutable",
```
alexeldeib, line 403: nit: it would be nice if these names were unique so it's easier to debug if they break in the future

exp/api/v1alpha3/azuremanagedmachinepool_conversion.go
```diff
@@ -43,6 +43,7 @@ func (src *AzureManagedMachinePool) ConvertTo(dstRaw conversion.Hub) error {
 	dst.Spec.MaxPods = restored.Spec.MaxPods
 	dst.Spec.OsDiskType = restored.Spec.OsDiskType
 	dst.Spec.NodeLabels = restored.Spec.NodeLabels
+	dst.Spec.EnableUltraSSD = restored.Spec.EnableUltraSSD
```
alexeldeib, line 46: I think you need this line for v1a4 as well, that will fix the failing test :)

alexeldeib:
Could you also squash commits after that the last edits?

exp/api/v1alpha3/azuremanagedmachinepool_conversion.go
TheNakedZealot: Thanks, added!

exp/api/v1beta1/azuremanagedmachinepool_webhook_test.go
TheNakedZealot: Made them more descriptive to the logic applied

jackfrancis:
@TheNakedZealot could you add this new ultrassd configuration to the "agentpool1" `AzureManagedMachinePool` spec that we test? That way we could get real-world test coverage of this change.

To do so, add the new configuration to "agentpool1" here:

- `templates/flavors/aks/cluster-template.yaml`

Then run `make generate-flavors` to render the AKS template with those changes.

Thanks!

CecileRobertMichon:
> To do so, add the new configuration to "agentpool1" here:
>
> templates/flavors/aks/cluster-template.yaml

@jackfrancis we should not add test configurations directly to the published flavor as those are meant to be examples. Instead, we can add specific patches for testing in https://github.com/kubernetes-sigs/cluster-api-provider-azure/tree/main/templates/test/ci/prow-aks/patches

jackfrancis:
@CecileRobertMichon thanks for the clarification.

@TheNakedZealot It looks like this is the `AzureManagedMachine` pool spec you want to add this new configuration to:

https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/templates/test/ci/prow-aks/patches/aks-agentpool0.yaml

After doing so, run `make generate-flavors`.

Thanks!

jackfrancis:
@TheNakedZealot `brew install wget` :)

(or equivalent)

TheNakedZealot:
> 

Wow, my failure on multiple levels. 😓

jackfrancis:
@TheNakedZealot Not at all, that same exact thing happened to me last week!

jackfrancis:
@TheNakedZealot test suggested that we have some work to do:

```
E0317 19:48:53.440285       1 controller.go:317] controller/azuremanagedcontrolplane "msg"="Reconciler error" "error"="error creating AzureManagedControlPlane capz-e2e-oxq9h5/capz-e2e-oxq9h5-aks: failed to reconcile AzureManagedControlPlane service managedclusters: failed to create managed cluster, failed to begin operation: Code=\"ErrorCodeEnableUltraSSD\" Message=\"Availability zone is required for UltraSSD setting.\"" "name"="capz-e2e-oxq9h5-aks" "namespace"="capz-e2e-oxq9h5" "reconciler group"="infrastructure.cluster.x-k8s.io" "reconciler kind"="AzureManagedControlPlane" 
```

Let's add the `availabilityZones: [1, 2]` configuration as well to the "aks-agentpool0.yaml` file (and re-run `make generate-flavors`) and see if that works.

jackfrancis:
```
Error from server: error when creating "STDIN": admission webhook "default.azuremanagedmachinepools.infrastructure.cluster.x-k8s.io" denied the request: json: cannot unmarshal number into Go struct field AzureManagedMachinePoolSpec.spec.availabilityZones of type string
```

Looks like I made a mistake, should be:

`availabilityZones: ["1", "2"]`

jackfrancis (approved):
/lgtm

templates/test/ci/prow-aks/patches/aks-agentpool0.yaml
```diff
@@ -5,3 +5,5 @@ metadata:
 spec:
   maxPods: 30
   osDiskType: "Managed"
+  enableUltraSSD: true
+  availabilityZones: ["1", "2"]
```
CecileRobertMichon, line 9: Do all the regions we test support availability zones @jackfrancis @mboersma ? Otherwise this might cause test failures when a region that doesn't support zones is selected

And was this added because zones are required for ultra ssd?

jackfrancis: Yes to the 2nd question.

@CecileRobertMichon is this where configured test regions lives?

- https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/07fbd76dbd7945edb2ef8af968d4693d6dff35f2/scripts/ci-entrypoint.sh#L46

If so, then northcentralus does not have availability zones (all other regions in the above list do). See:

- https://azure.microsoft.com/en-us/global-infrastructure/geographies/#geographies

CecileRobertMichon: ok in that case we can't merge this as-is, or tests will fail ~ 1 out of 7 times.

/hold

what can we do here? disable the test or remove the region I guess? Not super into the idea of removing a region since that will 1) increase load on quota from other regions and 2) we will loose coverage for "no AZ" regions (since that does have some different code paths and creates availability sets if zonal redundancy isn't available)

jackfrancis: IMO having coverage of AZ-dependent functionality is more important than proving that the set of our cluster configuration works on non-AZ-supported regions. The large majority of Azure regions support availability zones, and it is a fundamental building block of any mission critical operational implementation.

TheNakedZealot: Interestingly enough, I checked the region in question for support of Ultra Disk Storage, and this page claims North Central US is compatible. https://azure.microsoft.com/en-us/global-infrastructure/services/?regions=us-north-central%2cnon-regional&products=all

As is every location you all have listed in `REGIONS`; I checked them all for Ultra Disk Storage support, and the field is checked in each location's Products Page.

CecileRobertMichon: This doc https://docs.microsoft.com/en-us/azure/virtual-machines/disks-enable-ultra-ssd?tabs=azure-portal#ga-scope-and-limitations states:

> If a region in the following list lacks availability zones that support ultra disks, then a VM in that region must be deployed without infrastructure redundancy in order to attach an ultra disk.

So it's not that a region MUST support availability zones in order to use ultra disk storage, but it cannot use any other redundancy options (eg. availability sets). So in north central US, ultra SSD is available as long as VMs are single VMs. 

Could you please try to create an AKS cluster in north central US with ultra SSD enabled and see what happens? I wonder if we can get away with simply not listing zones explicitly (ie. does AKS do any sort of defaulting magic here that would place the VMs in zones whenever they are available?). Also wondering if VMSS counts as a "redundancy option", since I believe AKS agent pools are by default VMSS.

CecileRobertMichon: > does AKS do any sort of defaulting magic here that would place the VMs in zones whenever they are available

According to docs, that isn't the case: https://docs.microsoft.com/en-us/azure/aks/availability-zones#limitations-and-region-availability

It also confirms north central US is not one of the supported regions for availability zones in AKS.

jackfrancis: Before adding AZ config to the AKS PUT request, we got this error:

E0317 19:48:53.440285       1 controller.go:317] controller/azuremanagedcontrolplane "msg"="Reconciler error" "error"="error creating AzureManagedControlPlane capz-e2e-oxq9h5/capz-e2e-oxq9h5-aks: failed to reconcile AzureManagedControlPlane service managedclusters: failed to create managed cluster, failed to begin operation: Code=\"ErrorCodeEnableUltraSSD\" Message=\"Availability zone is required for UltraSSD setting.\"" "name"="capz-e2e-oxq9h5-aks" "namespace"="capz-e2e-oxq9h5" "reconciler group"="infrastructure.cluster.x-k8s.io" "reconciler kind"="AzureManagedControlPlane" 

So I concluded that we can't enable UltraSSD when we create an AKS cluster without declaring availability zones configuration. (And thus we can't test this new capz AKS functionality without AZ config, either)

CecileRobertMichon: @TheNakedZealot let's remove northcentral US from the list of test regions for now and add a comment above the list that all test regions must support AvailabilityZones

CecileRobertMichon:
@TheNakedZealot  please squash all commits when this is ready

scripts/ci-entrypoint.sh
```diff
@@ -42,8 +42,9 @@ source "${REPO_ROOT}/hack/parse-prow-creds.sh"
 # shellcheck source=hack/util.sh
 source "${REPO_ROOT}/hack/util.sh"
 
+# all test regions must support AvailabilityZones
 get_random_region() {
-    local REGIONS=("eastus" "eastus2" "northcentralus" "northeurope" "uksouth" "westeurope" "westus2")
```
CecileRobertMichon, line 46: the one we want to change is actually ci-e2e.sh (that's the script for running the CAPZ e2e suite) but for the sake of consistency, let's remove every occurrence (there should be 3 total, you can search for "get_random_region" in the codebase to find them).

TheNakedZealot: Should be looking good now - thanks!

CecileRobertMichon (approved):
/lgtm
/approve
/hold cancel

Thank you @TheNakedZealot

PR #2166, opened 2022-03-14
add unit tests for converters package

sonasingh46:

Adds unit tests for missing functions in converters package.
Increases the test coverage from 60% to 92.7%.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
None
```

### DIFF
--- a/azure/converters/image_test.go
+++ b/azure/converters/image_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package converters
 
 import (
+	"fmt"
 	"testing"
 
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2021-04-01/compute"
@@ -26,6 +27,97 @@ import (
 )
 
 func Test_ImageToPlan(t *testing.T) {
+	cases := []struct {
+		name      string
+		image     *infrav1.Image
+		expect    func(*GomegaWithT, *compute.ImageReference)
+		expectErr string
+	}{
+		{
+			name: "Should return a specific image reference if image id is specified",
+			image: &infrav1.Image{
+				ID: to.StringPtr("fake-id"),
+				SharedGallery: &infrav1.AzureSharedGalleryImage{
+					SubscriptionID: "fake-sub-id",
+					ResourceGroup:  "fake-rg",
+					Gallery:        "fake-gallery-name",
+					Name:           "fake-image-name",
+					Version:        "v1.0.0",
+				},
+			},
+			expect: func(g *GomegaWithT, result *compute.ImageReference) {
+				g.Expect(result).To(Equal(&compute.ImageReference{
+					ID: to.StringPtr("fake-id"),
+				}))
+			},
+			expectErr: "",
+		},
+		{
+			name: "Should return a market place image reference if Marketplace is specified",
+			image: &infrav1.Image{
+				Marketplace: &infrav1.AzureMarketplaceImage{
+					Publisher: "my-publisher",
+					Offer:     "my-offer",
+					SKU:       "my-sku",
+					Version:   "v1.0.0",
+				},
+			},
+			expect: func(g *GomegaWithT, result *compute.ImageReference) {
+				g.Expect(result).To(Equal(&compute.ImageReference{
+					Publisher: to.StringPtr("my-publisher"),
+					Offer:     to.StringPtr("my-offer"),
+					Sku:       to.StringPtr("my-sku"),
+					Version:   to.StringPtr("v1.0.0"),
+				}))
+			},
+			expectErr: "",
+		},
+		{
+			name: "Should return a SIG image reference if shared gallery is specified",
+			image: &infrav1.Image{
+				SharedGallery: &infrav1.AzureSharedGalleryImage{
+					SubscriptionID: "fake-sub-id",
+					ResourceGroup:  "fake-rg",
+					Gallery:        "fake-gallery-name",
+					Name:           "fake-image-name",
+					Version:        "v1.0.0",
+				},
+			},
+			expect: func(g *GomegaWithT, result *compute.ImageReference) {
+				ID := to.StringPtr(fmt.Sprintf("/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Compute/galleries/%s"+
+					"/images/%s/versions/%s", "fake-sub-id", "fake-rg", "fake-gallery-name", "fake-image-name", "v1.0.0"))
+				g.Expect(result.ID).To(Equal(ID))
+			},
+			expectErr: "",
+		},
+		{
+			name:      "Should return error if image is not specified",
+			expectErr: "unable to convert image as no options set",
+			image:     &infrav1.Image{},
+			expect: func(g *GomegaWithT, result *compute.ImageReference) {
+				// no-op
+			},
+		},
+	}
+
+	for _, c := range cases {
+		c := c
+		t.Run(c.name, func(t *testing.T) {
+			t.Parallel()
+			g := NewGomegaWithT(t)
+			result, err := ImageToSDK(c.image)
+			if c.expectErr != "" {
+				g.Expect(err).To(HaveOccurred())
+				g.Expect(err).To(MatchError(c.expectErr))
+			} else {
+				g.Expect(err).NotTo(HaveOccurred())
+			}
+			c.expect(g, result)
+		})
+	}
+}
+
+func Test_ImageToSDK(t *testing.T) {
 	cases := []struct {
 		name   string
 		image  *infrav1.Image
--- a/azure/converters/rules_test.go
+++ b/azure/converters/rules_test.go
@@ -1,0 +1,213 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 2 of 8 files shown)


azure/converters/vm_test.go
```diff
@@ -0,0 +1,141 @@
...
+	. "github.com/onsi/gomega"
+	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
+)
+
+func Test_SDKToVM(t *testing.T) {
+	cases := []struct {
+		name      string
+		sdkVM     compute.VirtualMachine
+		expect    func(*GomegaWithT, *VM)
+		expectErr bool
```
Jont828, line 33: Would this function ever return an error?

azure/converters/image_test.go
```diff
@@ -26,6 +27,96 @@ import (
...
+			expect: func(g *GomegaWithT, result *compute.ImageReference) {
+				ID := to.StringPtr(fmt.Sprintf("/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Compute/galleries/%s"+
+					"/images/%s/versions/%s", "fake-sub-id", "fake-rg", "fake-gallery-name", "fake-image-name", "v1.0.0"))
+				g.Expect(result.ID).To(Equal(ID))
+			},
+			expectErr: false,
+		},
+		{
+			name:      "Should return error if image is not specified",
+			expectErr: true,
```
Jont828, line 95: Wdyt about using a string for the expected error message instead of a bool? We do that [here](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/services/inboundnatrules/spec_test.go). The `expectedErr` string let's us check to verify the error message is correct, and we can set it to `""` to represent no error.

```
if tc.expectedError != "" {
	g.Expect(err).To(HaveOccurred())
	g.Expect(err).To(MatchError(tc.expectedError))
} else {
	g.Expect(err).NotTo(HaveOccurred())
	g.Expect(res).To(Equal(tc.expectedPortResult))
}
```

azure/converters/rules_test.go
```diff
@@ -0,0 +1,214 @@
...
+			securityRule: infrav1.SecurityRule{
+				Name:             "fake-rule",
+				Description:      "fake rule description",
+				Source:           to.StringPtr("fake-source"),
+				SourcePorts:      to.StringPtr("fake-source-ports"),
+				Destination:      to.StringPtr("fake-destination"),
+				DestinationPorts: to.StringPtr("fake-destination-port-ranges"),
+				Priority:         2,
+				Protocol:         infrav1.SecurityGroupProtocolAll,
+				//Direction: infrav1.SecurityRuleDirectionOutbound,
```
Jont828, line 45: Remove comment? I think it's fine to not test combinations of `rule.Protocol` and `rule.Direction` here since it'll have way too many combinations.

azure/converters/spotinstances.go
```diff
@@ -36,6 +37,7 @@ func GetSpotVMOptions(spotVMOptions *infrav1.SpotVMOptions) (compute.VirtualMach
 		if err != nil {
 			return "", "", nil, err
 		}
+		fmt.Println(err)
```
Jont828, line 40: Remove print statement?

azure/converters/spotinstances_test.go
```diff
@@ -0,0 +1,84 @@
...
+		billingProfile        *compute.BillingProfile
+	}
+	cases := []struct {
+		name          string
+		spotVMOptions *infrav1.SpotVMOptions
+		expectErr     bool
+		expect        func(*GomegaWithT, resultParams)
+	}{
+		{
+			name:          "should return vm priority, eviction policy and nil billing profile for spotVMOptions",
```
Jont828, line 42:
```suggestion
			name:          "should return vm priority, eviction policy and nil billing profile for spotVMOptions with nil MaxPrice",
```

azure/converters/spotinstances_test.go
```diff
@@ -0,0 +1,84 @@
...
+			spotVMOptions: &infrav1.SpotVMOptions{},
+			expectErr:     false,
+			expect: func(g *GomegaWithT, result resultParams) {
+				g.Expect(result.vmPriorityTypes).To(Equal(compute.VirtualMachinePriorityTypesSpot))
+				g.Expect(result.vmEvictionPolicyTypes).To(Equal(compute.VirtualMachineEvictionPolicyTypesDeallocate))
+				g.Expect(result.billingProfile).Should(BeNil())
+			},
+		},
+		{
+			name: "should return vm priority, eviction policy and billing profile for spotVMOptions",
```
Jont828, line 52:
```suggestion
			name: "should return vm priority, eviction policy and billing profile for spotVMOptions with set MaxPrice",
```

azure/converters/spotinstances_test.go
Jont828, line 42: Can we add a test for when `spotVMOptions == nil` for the base case?

azure/converters/image_test.go
```diff
@@ -26,6 +27,96 @@ import (
...
+				ID: to.StringPtr("fake-id"),
+				SharedGallery: &infrav1.AzureSharedGalleryImage{
+					SubscriptionID: "fake-sub-id",
+					ResourceGroup:  "fake-rg",
+					Gallery:        "fake-gallery-name",
+					Name:           "fake-image-name",
+					Version:        "v1.0.0",
+				},
+			},
+			expect: func(g *GomegaWithT, result *compute.ImageReference) {
```
Jont828, line 48: Wdyt about changing the `expect` part to `expectedResult`? That way, you can directly compare the expected conversion result in w/ something like this `g.Expect(res).To(Equal(tc.expectedResult))` in `t.Run()` and don't need to write it in each test case. Check out [this](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/services/inboundnatrules/spec_test.go) for an example.

azure/converters/tags_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+func TestMapToTags(t *testing.T) {
+	tests := []struct {
+		name   string
+		srcMap map[string]*string
+		want   v1beta1.Tags
+	}{
+		{
+			name: "convert map to tags",
+			srcMap: map[string]*string{
```
Jont828, line 35: Although this is really simple, can we add a test for (1) the map is empty, and (2) what if the map value is a nil string pointer?

Jont828 (commented):
I like that we're adding test cases here! Just a couple comments about how we could handle the results/errors.

azure/converters/vm_test.go
sonasingh46: This should not -- but that mean this functions needs refactoring.  Will do that.

azure/converters/image_test.go
sonasingh46: Either way should be fine. In all the test case -- we have to any way write the expected `infrav1.AzureSharedGalleryImage` struct. The only difference here is that it wrapped by a function. In fact, I thought about it while writing but it has been written the same in the existing code so did not break the pattern. 
Anyways, if you feel to go with the suggested approach -- I can make that change.

sonasingh46:
cc @CecileRobertMichon @jackfrancis @mboersma

azure/converters/image_test.go
Jont828: I got the `expectedResult` convention from the other tests like in `cluster_test.go`(https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/scope/cluster_test.go). IMO I prefer that method since it checks every field and is easier to understand at a glance. Wdyt @CecileRobertMichon

azure/converters/spotinstances_test.go
```diff
@@ -0,0 +1,95 @@
...
+	cases := []struct {
+		name          string
+		spotVMOptions *infrav1.SpotVMOptions
+		expectErr     bool
+		expect        func(*GomegaWithT, resultParams)
+	}{
+		{
+			name:          "should return empty vm priority, empty eviction policy and nil billing profile for nil spotVMOptions",
+			spotVMOptions: nil,
+			expectErr:     false,
```
Jont828, line 44: Can we use `expectErr` as a string like in image_test.go and add a test case where we are returning an error?

azure/converters/tags_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+func TestMapToTags(t *testing.T) {
+	tests := []struct {
+		name   string
+		srcMap map[string]*string
+		want   v1beta1.Tags
+	}{
+		{
+			name:   "convert nil map to tags",
+			srcMap: map[string]*string{},
```
Jont828, line 35: Is there a difference between `nil` map and empty map, i.e. set `srcMap` to `nil`? If not then we could merge this and the empty map test into one.

azure/converters/image_test.go
CecileRobertMichon: +1 to using `expectedResult` if we're trying to compare the entire struct (in this case looks like we are). In some other tests it's sometimes useful to have an `expect` func so we can customize the checks a bit more by test case and only check a specific desired outcome

CecileRobertMichon:
/close

in favor of #2564 etc.

/cc @willie-yao

Issue #1702, opened 2021-09-20
Make azure services async


shysank:
/kind feature

- #1541 introduced proposal to make azure service reconciliation asynchornous
- #1610 added a framework to achieve this by following a repeatable pattern

This issue aims to track individual services for the same:

- [ ] agentpools
- [ ] availabilitysets
- [ ] bastionhosts
- [ ] disks
- [x] groups - #1667 
- [ ] inboundnatrules
- [ ] loadbalancers
- [ ] managedclusters
- [ ] natgateways
- [ ] networkinterfaces
- [ ] privatedns
- [ ] publicips
- [ ] resourceskus
- [ ] roleassignments
- [ ] routetables - #1686 
- [ ] scalesets
- [ ] scalesetvms
- [ ] securitygroups - #1684 
- [ ] subnets
- [ ] tags
- [ ] virtualmachines - #1697 
- [ ] virtualnetworks - #1684 
- [ ] vmextensions

shysank:
cc @CecileRobertMichon

CecileRobertMichon:
thank you for opening this!

CecileRobertMichon:
Now that's I've opened a few PRs with some examples (and once the VM one merges, since that one has a few changes in the helpers), I would love some help to get this done. Since each service is self-contained, different people can open PRs in parallel, similar to what we did for #757. We just need to communicate on this issue before starting to work on a service so two people don't work on the same one.

/help

karuppiah7890:
How about we create separate GitHub issues for each service? More like sub tasks. GitHub supports that. Example - https://github.com/pingcap/tidb/issues/28010 (issue) , https://github.com/pingcap/tidb/issues/26747 (super issue for tracking), https://github.com/pingcap/tidb/issues/26022 (super super issue for tracking)

karuppiah7890:
Creating separate issues will avoid duplicate work or having to track assigned folks and the service they are assigned to from comments. We just have to ensure no one assigns themselves to this issue the super issue for tracking

shysank:
> How about we create separate GitHub issues for each service? More like sub tasks. GitHub supports that. Example - [pingcap/tidb#28010](https://github.com/pingcap/tidb/issues/28010) (issue) , [pingcap/tidb#26747](https://github.com/pingcap/tidb/issues/26747) (super issue for tracking), [pingcap/tidb#26022](https://github.com/pingcap/tidb/issues/26022) (super super issue for tracking)

[...]

PR #2168, opened 2022-03-14
[WIP] Make managedcluster reconcile and delete async

### LINKED ISSUE #2140: async managedclusters

CecileRobertMichon:
**What type of PR is this?**
/kind cleanup

**What this PR does / why we need it**: Implements #1702 for managedclusters.

Fixes #2140 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Make managedcluster reconcile and delete async
```


exp/api/v1beta1/azuremanagedcontrolplane_webhook.go
```diff
@@ -43,10 +48,9 @@ func (m *AzureManagedControlPlane) SetupWebhookWithManager(mgr ctrl.Manager) err
 
 // +kubebuilder:webhook:path=/mutate-infrastructure-cluster-x-k8s-io-v1beta1-azuremanagedcontrolplane,mutating=true,failurePolicy=fail,groups=infrastructure.cluster.x-k8s.io,resources=azuremanagedcontrolplanes,verbs=create;update,versions=v1beta1,name=default.azuremanagedcontrolplanes.infrastructure.cluster.x-k8s.io,sideEffects=None,admissionReviewVersions=v1;v1beta1
 
-var _ webhook.Defaulter = &AzureManagedControlPlane{}
-
 // Default implements webhook.Defaulter so a webhook will be registered for the type.
-func (m *AzureManagedControlPlane) Default() {
+func (m *AzureManagedControlPlane) Default(_ client.Client) {
```
alexeldeib, line 52: where is this client coming from? isn't this the defaulter without a client?

alexeldeib:
overall lgtm 🥳 huge PR, nice work!

I think splitting out the custom header interface change would be nice to reduce the size of the diff? Might be possible to break up some of the scope refactoring for machinepool too?

jackfrancis (changes requested):
is the `cluster-api-provider-azure` binary meant to be committed?

CecileRobertMichon:
> is the cluster-api-provider-azure binary meant to be committed?

oops that was not intentional, thanks for catching that. Removed.

jackfrancis:
> overall lgtm 🥳 huge PR, nice work!
> 
> I think splitting out the custom header interface change would be nice to reduce the size of the diff? Might be possible to break up some of the scope refactoring for machinepool too?

+1, an extraordinary amount of work

@zmalik @michalno1 @TheNakedZealot @thiagompc are we aware of any AKS scenarios that our E2E tests don't currently cover? For reference here is the current template that we use to test:

https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/a8a35fc9174816a320651e2e90e135ec42d6fc66/templates/test/ci/cluster-template-prow-aks.yaml

I'm thinking that if there are any particular use-cases we aren't testing (but that folks are reliably using) we will benefit by adding those scenarios given that the next release of capz will have the changes from this PR. These changes will optimize many reconciliations, but the scope is so large we'll be more likely to avoid regressions if we have more test coverage.

azure/services/managedclusters/spec.go
```diff
@@ -0,0 +1,423 @@
...
+
+// ManagedClusterSpec contains properties to create a managed cluster.
+type ManagedClusterSpec struct {
+	// Name is the name of this AKS Cluster.
+	Name string
+
+	// ResourceGroup is the name of the Azure resource group for this AKS Cluster.
+	ResourceGroup string
+
+	// NodeResourceGroupName is the name of the Azure resource group containing IaaS VMs.
```
jackfrancis, line 42: I see that you changed the name of `ResourceGroupName` to `ResourceGroup` while breaking out the existing `ManagedClusterSpec` into two types. Would this be a good time to consider changing the `NodeResourceGroupName` property name to something a little more accurately descriptive? For example, `InfraResourceGroupName` (or maybe better `InfraResourceGroup` to match the omission of the `Name` suffix above.

I'm wondering if having a more general name will make things less confusing in the future when folks are using this property to do things on non-node resources in the `MC_` resource group. Also (admittedly this is exceedingly speculative) if AKS were to ever implement a "control plane-only" feature (i.e., bring your own nodes) this name would then actually be misleading.

CecileRobertMichon: AKS API actually refers to it as "nodeResourceGroup" https://docs.microsoft.com/en-us/rest/api/aks/managed-clusters/get#managedcluster

properties.nodeResourceGroup | string | The name of the resource group containing agent pool nodes.
-- | -- | --

Will change it to `NodeResourceGroup` for consistency.

jackfrancis: adhering to AKS's API makes the most sense 👍

azure/services/managedclusters/spec_test.go
```diff
@@ -0,0 +1,208 @@
+/*
+Copyright 2021 The Kubernetes Authors.
```
sonasingh46, line 2: nit: 2022

azure/scope/managedcontrolplane.go
```diff
@@ -51,12 +48,15 @@ import (
...
+	Cluster             *clusterv1.Cluster
+	ControlPlane        *infrav1exp.AzureManagedControlPlane
+	ManagedMachinePools []ManagedMachinePool
+}
+
+type ManagedMachinePool struct {
 	InfraMachinePool *infrav1exp.AzureManagedMachinePool
-	MachinePool      *expv1.MachinePool
-	PatchTarget      conditions.Setter
+	OwnerMachinePool *clusterv1exp.MachinePool
```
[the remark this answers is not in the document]
CecileRobertMichon: No reason. Changed to `MachinePool`.

azure/services/managedclusters/spec.go
CecileRobertMichon: changed to `NodeResourceGroup`

azure/interfaces.go
```diff
@@ -106,6 +112,8 @@ type ResourceSpecGetter interface {
 	OwnerResourceName() string
 	// ResourceGroupName returns the name of the resource group the resource is in.
 	ResourceGroupName() string
+	// CustomHeaders returns the headers that should be added to Azure API calls.
+	CustomHeaders() map[string]string
```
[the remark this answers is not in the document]
CecileRobertMichon: @alexeldeib had also suggested splitting it out. I've added a commit to add a separate interface that has `CustomHeaders()`. The downside is that it required casting in client.go before using it but it's a little cleaner since it doesn't require touching any of the other services. Let me know what you think.

CecileRobertMichon:
looks like agent pools just took longer than expected 

/retest

azure/interfaces.go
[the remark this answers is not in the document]
sonasingh46: Yes. In separate interface -- it looks much cleaner. This should be fine to go with the PR.

CecileRobertMichon:
rebased + squashed 

@alexeldeib @jackfrancis @sonasingh46 @Jont828 PTAL

CecileRobertMichon:
rebased to fix merge conflict

azure/scope/managedcontrolplane.go
```diff
@@ -431,43 +422,19 @@ func (s *ManagedControlPlaneScope) ManagedClusterSpec() (azure.ManagedClusterSpe
 
 	if clusterNetwork := s.Cluster.Spec.ClusterNetwork; clusterNetwork != nil {
 		if clusterNetwork.Services != nil {
-			// A user may provide zero or one CIDR blocks. If they provide an empty array,
```
mboersma, line 434: Are we doing these "too many CIDRs" checks elsewhere now?

azure/services/managedclusters/spec.go
```diff
@@ -0,0 +1,430 @@
...
+	Managed bool
+
+	// EnableAzureRBAC - Whether to enable Azure RBAC for Kubernetes authorization.
+	EnableAzureRBAC bool
+
+	// AdminGroupObjectIDs - AAD group object IDs that will have admin role of the cluster.
+	AdminGroupObjectIDs []string
+}
+
+// AddonProfile - Profile of managed cluster add-on.
```
mboersma, line 112:
```suggestion
// AddonProfile is the profile of a managed cluster add-on.
```

azure/services/managedclusters/spec.go
```diff
@@ -0,0 +1,430 @@
...
+type AddonProfile struct {
+	Name    string
+	Config  map[string]string
+	Enabled bool
+}
+
+// SKU - AKS SKU.
```
mboersma, line 119:
```suggestion
// SKU is an AKS SKU.
```

azure/services/managedclusters/spec.go
```diff
@@ -0,0 +1,430 @@
...
+type SKU struct {
+	// Tier - Tier of a managed cluster SKU.
+	Tier string
+}
+
+// LoadBalancerProfile - Profile of the cluster load balancer.
```
mboersma, line 125:
```suggestion
// LoadBalancerProfile is the profile of the cluster load balancer.
```

azure/services/managedclusters/spec.go
```diff
@@ -0,0 +1,430 @@
...
+// AddonProfile - Profile of managed cluster add-on.
+type AddonProfile struct {
+	Name    string
+	Config  map[string]string
+	Enabled bool
+}
+
+// SKU - AKS SKU.
+type SKU struct {
+	// Tier - Tier of a managed cluster SKU.
```
mboersma, line 121: I'm pedantic enough to flag all these, but I prefer complete sentences for doc comments.

azure/services/managedclusters/spec.go
```diff
@@ -0,0 +1,430 @@
...
+			Type: containerservice.ResourceIdentityTypeSystemAssigned,
+		},
+		Location: &s.Location,
+		ManagedClusterProperties: &containerservice.ManagedClusterProperties{
+			NodeResourceGroup: &s.NodeResourceGroup,
+			EnableRBAC:        to.BoolPtr(true),
+			DNSPrefix:         &s.Name,
+			KubernetesVersion: &s.Version,
+			LinuxProfile: &containerservice.LinuxProfile{
+				AdminUsername: to.StringPtr("azureuser"),
```
mboersma, line 200: Should this be a string constant?

mboersma (approved):
This is a huge refactoring. I reviewed it all and I get the gist of the changes and didn't see anything obviously wrong.

shysank (commented):
looks good overall with one small clarification on AzureManagedControlPlane validation.

exp/api/v1beta1/azuremanagedcontrolplane_webhook.go
```diff
@@ -365,6 +368,74 @@ func (m *AzureManagedControlPlane) validateAPIServerAccessProfile() error {
...
+
+	ownerCluster := &clusterv1.Cluster{}
+	key := client.ObjectKey{
+		Namespace: m.Namespace,
+		Name:      clusterName,
+	}
+
+	if err := cli.Get(ctx, key, ownerCluster); err != nil {
+		if azure.ResourceNotFound(err) {
+			return nil
```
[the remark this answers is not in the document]
CecileRobertMichon: That's a good point. I had copied the pattern from https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/55a16d82fb5759e44ca1e1fa8601f778a5f23b46/exp/api/v1beta1/azuremanagedmachinepool_webhook.go#L235 so we should make both consistent if we do decide to change this. I wonder if we should retry a couple of times before failing in case the AzureManagedCluster and the Cluster are created simultaneously. Do you know if kubectl apply -f template.yaml is serial? Will it block creation if the AzureManagedControlPlane is applied before the Cluster or will it proceed to create the Cluster? 🤔

azure/services/managedclusters/spec.go
CecileRobertMichon: I agree - I have to admit I just lift and shifted those as-is from types.go. Will cleanup the comments.

azure/services/managedclusters/spec.go
CecileRobertMichon: yes it should be

exp/api/v1beta1/azuremanagedcontrolplane_webhook.go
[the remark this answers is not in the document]
shysank: > so we should make both consistent if we do decide to change this.
+1 to making it consistent. I'm okay with merging this one as is, and create a followup issue to address this concern.

> I wonder if we should retry a couple of times before failing
I'm a bit hesitant about this because we don't want to add too much complexity in validation webhook.

> Will it block creation if the AzureManagedControlPlane is applied before the Cluster or will it proceed to create the Cluster?
I think it's serial. However, it wouldn't block creation if AzureManagedControlPlane is defined before Cluster in the yaml.

azure/scope/managedmachinepool.go
```diff
@@ -0,0 +1,250 @@
...
+	machinePool *expv1.MachinePool,
+	managedMachinePool *infrav1exp.AzureManagedMachinePool) azure.AgentPoolSpec {
+	var normalizedVersion *string
+	if machinePool.Spec.Template.Spec.Version != nil {
+		v := strings.TrimPrefix(*machinePool.Spec.Template.Spec.Version, "v")
+		normalizedVersion = &v
+	}
+
+	replicas := int32(1)
+	if machinePool.Spec.Replicas != nil {
```
Jont828, line 133: What situation would we have `machinePool.Spec.Replicas` be nil? And if so, is there any particular reason why we want to set the default replicas to 1?

azure/scope/managedmachinepool.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+
+	agentPoolSpec := azure.AgentPoolSpec{
+		Name:          to.String(managedMachinePool.Spec.Name),
+		ResourceGroup: managedControlPlane.Spec.ResourceGroupName,
+		Cluster:       managedControlPlane.Name,
+		SKU:           managedMachinePool.Spec.SKU,
+		Replicas:      replicas,
+		Version:       normalizedVersion,
+		VnetSubnetID: azure.SubnetID(
```
Jont828, line 144: Nit: could we rename `VnetSubnetID` to `SubnetID`?

azure/scope/managedmachinepool_test.go
```diff
@@ -0,0 +1,561 @@
...
+				InfraMachinePool: getAzureMachinePool("pool0", infrav1.NodePoolModeSystem),
+			},
+			Expected: azure.AgentPoolSpec{
+
+				Name:         "pool0",
+				SKU:          "Standard_D2s_v3",
+				Replicas:     1,
+				Mode:         "System",
+				Cluster:      "cluster1",
+				VnetSubnetID: "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups//providers/Microsoft.Network/virtualNetworks//subnets/",
```
Jont828, line 73:
```suggestion
				VnetSubnetID: "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups//providers/Microsoft.Network/virtualNetworks/subnets/",
```

Remove double `//`

azure/services/managedclusters/managedclusters.go
```diff
@@ -18,37 +18,25 @@ package managedclusters
...
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 )
 
-const serviceName = "managedclusters"
-
-var (
-	defaultUser     = "azureuser"
-	managedIdentity = "msi"
-)
+const serviceName = "managedcluster"
```
Jont828, line 33: Is there a reason we're renaming this service? If so, do we want to rename the folder under `azure/services` for consistency?

azure/services/managedclusters/spec.go
```diff
@@ -0,0 +1,430 @@
...
+	EnablePrivateCluster *bool
+	// PrivateDNSZone is the private dns zone for private clusters.
+	PrivateDNSZone *string
+	// EnablePrivateClusterPublicFQDN defines whether to create additional public FQDN for private cluster or not.
+	EnablePrivateClusterPublicFQDN *bool
+}
+
+var _ azure.ResourceSpecGetterWithHeaders = (*ManagedClusterSpec)(nil)
+
+// ResourceName returns the name of the group.
```
Jont828, line 162:
```suggestion
// ResourceName returns the name of the managed cluster.
```
or
```suggestion
// ResourceName returns the name of the AKS cluster.
```

azure/services/managedclusters/spec.go
```diff
@@ -0,0 +1,430 @@
...
+func (s *ManagedClusterSpec) ResourceName() string {
+	return s.Name
+}
+
+// ResourceGroupName returns the name of the group.
```
Jont828, line 167:
```suggestion
// ResourceGroupName returns the name of the resource group.
```

azure/services/managedclusters/spec.go
```diff
@@ -0,0 +1,430 @@
...
+// Note that it is the same as the resource name in this case.
```
Jont828, line 168:
```suggestion

```

azure/services/managedclusters/spec.go
```diff
@@ -0,0 +1,430 @@
...
+func (s *ManagedClusterSpec) ResourceGroupName() string {
+	return s.ResourceGroup
+}
+
+// OwnerResourceName is a no-op for groups.
```
Jont828, line 173:
```suggestion
// OwnerResourceName is a no-op for managed clusters.
```

azure/services/managedclusters/spec.go
```diff
@@ -0,0 +1,430 @@
...
+func (s *ManagedClusterSpec) OwnerResourceName() string {
+	return "" // not applicable
+}
+
+// CustomHeaders returns custom headers to be added to the Azure API calls.
+func (s *ManagedClusterSpec) CustomHeaders() map[string]string {
+	return s.Headers
+}
+
+// Parameters returns the parameters for the group.
```
Jont828, line 183:
```suggestion
// Parameters returns the parameters for the managed cluster.
```

azure/services/managedclusters/spec_test.go
```diff
@@ -0,0 +1,208 @@
...
+				},
+				Version:         "v1.22.99",
+				LoadBalancerSKU: "Standard",
+			},
+			expect: func(g *WithT, result interface{}) {
+				g.Expect(result).To(BeAssignableToTypeOf(containerservice.ManagedCluster{}))
+				g.Expect(result.(containerservice.ManagedCluster).KubernetesVersion).To(Equal(to.StringPtr("v1.22.99")))
+			},
+		},
+		// TODO (cecile): add tests for specific parameters
```
Jont828, line 131: Is this TODO done?

exp/api/v1beta1/azuremanagedcontrolplane_webhook.go
Jont828, line 52: Why did we add an unused parameter here? Is it to fulfill an interface where it's needed somewhere else?

azure/scope/managedcontrolplane.go
```diff
@@ -51,12 +48,15 @@ import (
...
-	Client           client.Client
-	Cluster          *clusterv1.Cluster
-	ControlPlane     *infrav1exp.AzureManagedControlPlane
+	Client              client.Client
+	Cluster             *clusterv1.Cluster
+	ControlPlane        *infrav1exp.AzureManagedControlPlane
+	ManagedMachinePools []ManagedMachinePool
+}
+
+type ManagedMachinePool struct {
```
Jont828, line 57: Can we move this to `managedmachinepool.go` since we're trying to separate the scopes? Or would that cause an import cycle?

exp/api/v1beta1/azuremanagedcontrolplane_webhook_test.go
```diff
@@ -240,11 +243,11 @@ func TestValidatingWebhook(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			g := NewWithT(t)
 			t.Parallel()
-
+			var client client.Client
```
Jont828, line 246: Is this client supposed to be uninitialized?

azure/services/managedclusters/spec.go
```diff
@@ -0,0 +1,430 @@
...
+	AuthorizedIPRanges []string
+	// EnablePrivateCluster defines hether to create the cluster as a private cluster or not.
+	EnablePrivateCluster *bool
+	// PrivateDNSZone is the private dns zone for private clusters.
+	PrivateDNSZone *string
+	// EnablePrivateClusterPublicFQDN defines whether to create additional public FQDN for private cluster or not.
+	EnablePrivateClusterPublicFQDN *bool
+}
+
+var _ azure.ResourceSpecGetterWithHeaders = (*ManagedClusterSpec)(nil)
```
Jont828, line 160: What's this line for?

azure/services/managedclusters/managedclusters.go
```diff
@@ -184,240 +70,57 @@ func (s *Service) Reconcile(ctx context.Context) error {
...
-		if azure.ResourceNotFound(err) {
-			// already deleted
-			return nil
-		}
-		return errors.Wrapf(err, "failed to delete managed cluster %s in resource group %s", s.Scope.ClusterName(), s.Scope.ResourceGroup())
+	ctx, cancel := context.WithTimeout(ctx, reconciler.DefaultAzureServiceReconcileTimeout)
+	defer cancel()
+
+	managedClusterSpec := s.Scope.ManagedClusterSpec(ctx)
+	if managedClusterSpec == nil {
```
Jont828, line 117: Maybe add a log statement similar to the other async services?

azure/services/managedclusters/managedclusters.go
```diff
@@ -184,240 +70,57 @@ func (s *Service) Reconcile(ctx context.Context) error {
...
+	defer cancel()
 
-	isCreate := false
-	existingMC, err := s.Client.Get(ctx, managedClusterSpec.ResourceGroupName, managedClusterSpec.Name)
-	// Transient or other failure not due to 404
-	if err != nil && !azure.ResourceNotFound(err) {
-		return azure.WithTransientError(errors.Wrap(err, "failed to fetch existing managed cluster"), 20*time.Second)
+	managedClusterSpec := s.Scope.ManagedClusterSpec(ctx)
+	if managedClusterSpec == nil {
+		return nil
```
Jont828, line 78: Maybe add a log statement like in the other async services?

azure/services/managedclusters/managedclusters.go
```diff
@@ -184,240 +70,57 @@ func (s *Service) Reconcile(ctx context.Context) error {
...
-		if len(managedClusterSpec.LoadBalancerProfile.OutboundIPPrefixes) > 0 {
-			managedCluster.NetworkProfile.LoadBalancerProfile.OutboundIPPrefixes = &containerservice.ManagedClusterLoadBalancerProfileOutboundIPPrefixes{
-				PublicIPPrefixes: convertToResourceReferences(managedClusterSpec.LoadBalancerProfile.OutboundIPPrefixes),
-			}
-		}
-		if len(managedClusterSpec.LoadBalancerProfile.OutboundIPs) > 0 {
-			managedCluster.NetworkProfile.LoadBalancerProfile.OutboundIPs = &containerservice.ManagedClusterLoadBalancerProfileOutboundIPs{
-				PublicIPs: convertToResourceReferences(managedClusterSpec.LoadBalancerProfile.OutboundIPs),
+		// Update control plane endpoint.
+		if managedCluster.ManagedClusterProperties != nil && managedCluster.ManagedClusterProperties.Fqdn != nil {
```
Jont828, line 88: Would it be unexpected for the properties to be nil? If so maybe we could add a log.

azure/scope/managedmachinepool.go
CecileRobertMichon: this (and the entire file) was copied directly as-is from https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2168/files#diff-cfd94adbc2ff767742e457e64836e93e88c6e03c3a64c04619c54b7ed76f284eL574

exp/api/v1beta1/azuremanagedcontrolplane_webhook.go
CecileRobertMichon: correct https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/util/webhook/mutator.go#L32

we're just not using the client for default here, just for validate

azure/scope/managedmachinepool_test.go
CecileRobertMichon: I think it's empty because vnet name wasn't set, same with rg name. This was also a file split from https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2168/files/5dddbd410181418a2e1f95fd83a11b6dc894c6c7#diff-a851ceb007616d5d0254ac740f53cfb4dd97e16b30629e5d201ef40527d70c82L115

azure/services/managedclusters/managedclusters.go
CecileRobertMichon: If I remember correctly, this is just so logs are a bit more coherent since we're only creating one managed cluster per AzureManagedControlPlane, so instead of printing "managedclusters creating or updating" it will print "managedcluster creating or updating" in Conditions. Not a huge deal, I didn't want to rename the package since that's a bigger change but I can revert it if you think it's not worth it. I think groups are the same if that helps.

azure/services/managedclusters/spec.go
CecileRobertMichon: it's validating that the managed cluster spec implements the `ResourceSpecGetterWithHeaders` interface

it's a common go pattern to check that a type implements an interface https://golangbyexample.com/go-check-if-type-implements-interface/

azure/services/managedclusters/managedclusters.go
CecileRobertMichon: don't think we're logging anything for other services if there is no spec? It just means that service is not applicable for that object

azure/services/managedclusters/managedclusters.go
CecileRobertMichon: ditto

azure/services/managedclusters/managedclusters.go
CecileRobertMichon: not sure, I kept the same logic that was there before https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2168/files#diff-ea46b89333855bb06ff20351c648ad0a792c9fc36689b982e871b9f6fee2ef7aL366

CecileRobertMichon: removed this and the pointer dereference

CecileRobertMichon:
@Jont828 thanks for the review! I've addressed most of your comments, PTAL. I kept the new changes as a separate commit, let me know if this looks good and I can squash.

exp/api/v1beta1/azuremanagedcontrolplane_webhook.go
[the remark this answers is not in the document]
CecileRobertMichon: I ran some tests and it definitely works fine and doesn't block creation even if AzureManagedControlPlane is defined before Cluster

Changed both to return an error if Cluster is not found

CecileRobertMichon:
@Jont828 @shysank did you have any other comments? Let me know when I can squash

CecileRobertMichon:
@Jont828 @shysank did you have any other comments? Let me know when I can squash

exp/api/v1beta1/azuremanagedcontrolplane_webhook.go
```diff
@@ -270,7 +270,7 @@ func (m *AzureManagedControlPlane) Validate() error {
 }
 
 // validate DNSServiceIP.
```
Jont828, line 272: Nit: capitalize "validate" in comment to be consistent

Jont828 (commented):
Small thing but other than that I think we're good to go!

exp/api/v1beta1/azuremanagedcontrolplane_webhook.go
CecileRobertMichon: I fixed the comment to be consistent with the function name, and fixed the others that were using inconsistent capitalization

CecileRobertMichon:
Squashed

CecileRobertMichon:
@Jont828 @mboersma @shysank PTAL, addressed all comments and squashed

CecileRobertMichon:
@mboersma @shysank any objections on merging this one?

mboersma:
lgtm! Please merge, your patience is admirable.

PR #2177, opened 2022-03-17
Make VM extension reconcile async and move VMSS extension into scaleset service

### LINKED ISSUE #2141: async vmextensions

Jont828:
**What type of PR is this?**

/kind feature
**What this PR does / why we need it**: Implementation of an async service for VM extensions as a follow up for #1610 and #1541.

Fixes #2141 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Make VM extension reconcile async and move VMSS extension into scaleset service
```


Jont828:
@CecileRobertMichon I'm not sure if this is a scope creep but I think it would make sense to refactor VMSS extensions to be async as well. We have code in the scaleset service that wants to create VMSS extension parameters from the spec, but we haven't implemented a VMSSExtensionSpec to compute them.

```
for i, extensionSpec := range s.Scope.VMSSExtensionSpecs() {
    extensionSpec := extensionSpec
    extensions[i] = compute.VirtualMachineScaleSetExtension{
        Name: &extensionSpec.Name,
        VirtualMachineScaleSetExtensionProperties: &compute.VirtualMachineScaleSetExtensionProperties{
            Publisher:          to.StringPtr(extensionSpec.Publisher),
            Type:               to.StringPtr(extensionSpec.Name),
            TypeHandlerVersion: to.StringPtr(extensionSpec.Version),
            Settings:           nil,
            ProtectedSettings:  extensionSpec.ProtectedSettings,
        },
    }
}
````

File: azure/services/scalesets/scalesets.go [hunk 1]
```diff
@@ -544,22 +547,33 @@ func (s *Service) getVirtualMachineScaleSetIfDone(ctx context.Context, future *i
 	return converters.SDKToVMSS(vmss, vmssInstances), nil
 }
 
-func (s *Service) generateExtensions() []compute.VirtualMachineScaleSetExtension {
+func (s *Service) generateExtensions() ([]compute.VirtualMachineScaleSetExtension, error) {
 	extensions := make([]compute.VirtualMachineScaleSetExtension, len(s.Scope.VMSSExtensionSpecs()))
 	for i, extensionSpec := range s.Scope.VMSSExtensionSpecs() {
 		extensionSpec := extensionSpec
+		parameters, err := extensionSpec.Parameters(nil)
```
Jont828, line 554: @CecileRobertMichon This might be a scope creep but should we also refactor VMSS extensions to be async as well? `VMSSExtensionSpecs()` and `VMExtensionSpecs()` both used to return `ExtensionSpec` which is refactored into `vmextensions.ExtensionSpec`. But currently `vmextensions.ExtensionSpec.Parameters()` creates a `compute.VirtualMachineExtension` while here we want a `compute.VirtualMachineScaleSetExtension`. Another option would be to add a converter function that just copies the fields.

CecileRobertMichon: yes I think doing both VM and VMSS extensions in the same PR might actually make it easier since they share some code

Jont828: @CecileRobertMichon Cool I've made some changes to refactor VMSS. Could you take a look when you have some time?

azure/services/vmssextensions/spec.go
```diff
@@ -0,0 +1,73 @@
...
+	return compute.VirtualMachineScaleSetExtension{
+		Name: to.StringPtr(s.Name),
+		VirtualMachineScaleSetExtensionProperties: &compute.VirtualMachineScaleSetExtensionProperties{
+			Publisher:          to.StringPtr(s.Publisher),
+			Type:               to.StringPtr(s.Name),
+			TypeHandlerVersion: to.StringPtr(s.Version),
+			Settings:           nil,
+			ProtectedSettings:  s.ProtectedSettings,
+		},
+		// TODO: should we include location since it's used in VMExtensions too?
```
CecileRobertMichon, line 71: not sure why one has location and not the other, is location required for VM extensions?

azure/services/vmssextensions/spec.go
```diff
@@ -0,0 +1,73 @@
...
+
+import (
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2021-04-01/compute"
+	"github.com/Azure/go-autorest/autorest/to"
+	"github.com/pkg/errors"
+)
+
+// VMSSExtensionSpec defines the specification for a VM or VMScaleSet extension.
+type VMSSExtensionSpec struct {
+	Name              string
```
CecileRobertMichon, line 27: instead of duplicating fields here and in VMExtensionSpec (since they are the same) consider nesting the Azure extension spec in both

```
type VMSSExtensionSpec struct {
    azure.ExtensionSpec
}
```

azure/types.go
```diff
@@ -96,8 +96,8 @@ type PrivateDNSLinkSpec struct {
 	LinkName          string
 }
 
-// ExtensionSpec defines the specification for a VM or VMScaleSet extension.
-type ExtensionSpec struct {
+// BootstrapingExtensionSpec defines the specification for a CAPZ Bootstrapping VM or VMSS extension.
+type BootstrapingExtensionSpec struct {
```
CecileRobertMichon, line 100: Why the rename? the fields here could be reused later on for any other type of extension, nothing in here is bootstrapping specific AFAIK so I think the generic naming is appropriate

azure/services/vmssextensions/spec.go
Jont828: The code in VMSS created the parameters w/o a location like this while the VM extensions included it. I kept the original behavior since it's a refactor but I wasn't sure if it was a bug.
```
extensions[i] = compute.VirtualMachineScaleSetExtension{
	Name: &extensionSpec.Name,
	VirtualMachineScaleSetExtensionProperties: &compute.VirtualMachineScaleSetExtensionProperties{
		Publisher:          to.StringPtr(extensionSpec.Publisher),
		Type:               to.StringPtr(extensionSpec.Name),
		TypeHandlerVersion: to.StringPtr(extensionSpec.Version),
		Settings:           nil,
		ProtectedSettings:  extensionSpec.ProtectedSettings,
	},
}
```

azure/types.go
Jont828: I renamed it cause that struct is populated only by the `GetBootstrappingVMExtension()` function here and I thought it would be more clear where it's coming from. But if we want to use it more generically then we can change it back.
 https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/71a01a4acbbab2b37f8fc2010d3ceead5dadcfcc/azure/defaults.go#L361

azure/types.go
```diff
@@ -96,7 +96,7 @@ type PrivateDNSLinkSpec struct {
 	LinkName          string
 }
 
-// ExtensionSpec defines the specification for a VM or VMScaleSet extension.
+// ExtensionSpec defines the specification for a CAPZ Bootstrapping VM or VMSS extension.
```
CecileRobertMichon, line 99: nit: revert the comment about CAPZ Bootstrapping (VMSS change is good to keep)

azure/services/vmextensions/spec.go
```diff
@@ -0,0 +1,70 @@
...
+import (
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2021-04-01/compute"
+	"github.com/Azure/go-autorest/autorest/to"
+	"github.com/pkg/errors"
+	"sigs.k8s.io/cluster-api-provider-azure/azure"
+)
+
+// VMExtensionSpec defines the specification for a VM or VMScaleSet extension.
+type VMExtensionSpec struct {
+	Spec          azure.ExtensionSpec
```
CecileRobertMichon, line 28: This could just be:

```suggestion
	azure.ExtensionSpec
```

so you can refer to fields directly, for example `VMExtensionSpec.Publisher` instead of `VMExtensionSpec.Spec.Publisher`

azure/services/vmssextensions/spec.go
CecileRobertMichon: I'm not sure either. You could try 1) removing Location from VM extensions 2) adding Location to VMSS extensions, and see if either breaks?

I agree it doesn't seem right that they aren't consistent, but it's possible that this is due to the Azure API between VM and VMSS not being consistent (if it's not, let's make it consistent).

Jont828:
@CecileRobertMichon It looks like the reason we don't have a location in `compute.VirtualMachineScaleSetExtension` is because it simply doesn't have a location field while `compute.VirtualMachineExtension` does. So I guess it's something built in to the Azure Go API.

azure/services/vmextensions/client.go
```diff
@@ -52,36 +48,118 @@ func newVirtualMachineExtensionsClient(subscriptionID string, baseURI string, au
...
+	err = createFuture.WaitForCompletionRef(ctx, ac.vmextensions.Client)
+	if err != nil {
+		// if an error occurs, return the future.
+		// this means the long-running operation didn't finish in the specified timeout.
+		return nil, &createFuture, err
+	}
+	result, err = createFuture.Result(ac.vmextensions)
+	// if the operation completed, return a nil future.
+	return result, nil, err
+	// ctx, _, done := tele.StartSpanWithLogger(ctx, "vmextensions.AzureClient.CreateOrUpdate")
```
CecileRobertMichon, line 88: do you need to cleanup this commented out code?

azure/services/vmextensions/client.go
```diff
@@ -52,36 +48,118 @@ func newVirtualMachineExtensionsClient(subscriptionID string, baseURI string, au
...
+	return ac.vmextensions.Get(ctx, spec.ResourceGroupName(), spec.OwnerResourceName(), spec.ResourceName(), "")
 }
 
-// CreateOrUpdateAsync creates or updates the virtual machine extension.
-func (ac *azureClient) CreateOrUpdateAsync(ctx context.Context, resourceGroupName, vmName, name string, parameters compute.VirtualMachineExtension) error {
+// CreateOrUpdateAsync creates or updates a VM extension asynchronously.
+// It sends a PUT request to Azure and if accepted without error, the func will return a Future which can be used to track the ongoing
+// progress of the operation.
+func (ac *azureClient) CreateOrUpdateAsync(ctx context.Context, spec azure.ResourceSpecGetter, parameters interface{}) (result interface{}, future azureautorest.FutureAPI, err error) {
 	ctx, _, done := tele.StartSpanWithLogger(ctx, "vmextensions.AzureClient.CreateOrUpdate")
```
CecileRobertMichon, line 63:
```suggestion
	ctx, _, done := tele.StartSpanWithLogger(ctx, "vmextensions.AzureClient. CreateOrUpdateAsync")
```

azure/services/vmextensions/client.go
```diff
@@ -52,36 +48,118 @@ func newVirtualMachineExtensionsClient(subscriptionID string, baseURI string, au
...
+	// return err
 }
 
-// Delete removes the virtual machine extension.
-func (ac *azureClient) Delete(ctx context.Context, resourceGroupName, vmName, name string) error {
+// DeleteAsync deletes a VM extension asynchronously. DeleteAsync sends a DELETE
+// request to Azure and if accepted without error, the func will return a Future which can be used to track the ongoing
+// progress of the operation.
+func (ac *azureClient) DeleteAsync(ctx context.Context, spec azure.ResourceSpecGetter) (future azureautorest.FutureAPI, err error) {
 	ctx, _, done := tele.StartSpanWithLogger(ctx, "vmextensions.AzureClient.Delete")
```
CecileRobertMichon, line 99:
```suggestion
	ctx, _, done := tele.StartSpanWithLogger(ctx, "vmextensions.AzureClient.DeleteAsync")
```

azure/services/vmextensions/vmextensions.go
```diff
@@ -23,29 +23,34 @@ import (
...
 	"sigs.k8s.io/cluster-api-provider-azure/util/tele"
 )
 
 const serviceName = "vmextensions"
 
 // VMExtensionScope defines the scope interface for a vm extension service.
 type VMExtensionScope interface {
+	azure.Authorizer
+	azure.AsyncStatusUpdater
 	azure.ClusterDescriber
```
CecileRobertMichon, line 37: can we get rid of `ClusterDescriber` here?

azure/services/vmextensions/vmextensions_test.go
```diff
@@ -27,169 +27,140 @@ import (
...
-				}, nil)
-				s.SetBootstrapConditions(gomockinternal.AContext(), string(compute.ProvisioningStateFailed), "my-extension-1")
+			expect: func(s *mock_vmextensions.MockVMExtensionScopeMockRecorder, r *mock_async.MockReconcilerMockRecorder) {
+				s.VMExtensionSpecs().Return([]azure.ResourceSpecGetter{&extensionSpec1})
+				r.CreateResource(gomockinternal.AContext(), &extensionSpec1, serviceName).Return(extensionFailed1, nil)
+				s.SetBootstrapConditions(gomockinternal.AContext(), string(compute.ProvisioningStateFailed), extensionSpec1.ResourceName()).Return(nil)
 			},
 		},
 		{
+			// TODO: is this blocked by the operationNotDoneError?
```
CecileRobertMichon, line 126: is this still TODO?

azure/services/vmextensions/vmextensions.go
```diff
@@ -56,47 +61,45 @@ func (s *Service) Name() string {
...
+		} else {
+			vmextension, ok := result.(compute.VirtualMachineExtension)
+			if !ok {
+				return errors.Errorf("%T is not a compute.VirtualMachineExtension", result)
+			}
+
 			// check the extension status and set the associated conditions.
-			if retErr := s.Scope.SetBootstrapConditions(ctx, to.String(existing.ProvisioningState), extensionSpec.Name); retErr != nil {
+			if retErr := s.Scope.SetBootstrapConditions(ctx, to.String(vmextension.ProvisioningState), extensionSpec.ResourceName()); retErr != nil {
+				// TODO: what precedence should this error have?
```
CecileRobertMichon, line 93: `SetBootstrapConditions` might need a bit of refactoring. Instead of taking the error and using that to set the condition like we're doing for other resources, it's taking the provisioning state of the extension and using that to set the condition, then returning an error if the extension is done creating. Can we change it to always run (even if resultErr is not nil) and use the error to determine what the condition should be set to instead? Similar to what we're doing in `UpdatePutStatus`

azure/services/vmssextensions/spec.go
CecileRobertMichon: > Are VM/VMSS extensions covered in the e2e tests?

yes they are, since we create an extension for every single VM and every single VMSS to check kubeadm completion.

CecileRobertMichon: from your other comment sounds like there is a discrepancy in the API and we can just get rid of this comment?

CecileRobertMichon (commented):
Overall looking good, there are a few outstanding TODOs that need to be resolved before this can merge and looks like you need to rebase due to a conflict.

azure/services/scalesets/scalesets.go
```diff
@@ -144,8 +144,11 @@ func (s *Service) Reconcile(ctx context.Context) (retErr error) {
 		}
 	}
 
-	// if we get to here, we have completed any long running VMSS operations (creates / updates)
+	// If we get to here, we have completed any long running VMSS operations (creates / updates)
 	s.Scope.DeleteLongRunningOperationState(s.Scope.ScaleSetSpec().Name, serviceName)
+	// This also means that the VMSS extensions were successfully installed
+	s.Scope.UpdatePutStatus(infrav1.BootstrapSucceededCondition, serviceName, nil)
```
Jont828, line 150: @CecileRobertMichon I went ahead and added the UpdatePutStatus() call in the scaleset service. However, I'm not sure how we want to set the error message for VMSSExtensions. Do we want to just set it with whatever error comes from `s.createVMSS(ctx)`, `s.patchVMSSIfNeeded(ctx, fetchedVMSS)`, and `s.getVirtualMachineScaleSetIfDone(ctx, future)`? Or should we try to set specific errors caused by the VMSSExtension (and not necessarily the scaleset as a whole)?

api/v1beta1/conditions_consts.go
```diff
@@ -43,7 +43,7 @@ const (
 	// WaitingForBootstrapDataReason used when machine is waiting for bootstrap data to be ready before proceeding.
 	WaitingForBootstrapDataReason = "WaitingForBootstrapData"
 	// BootstrapSucceededCondition reports the result of the execution of the boostrap data on the machine.
-	BootstrapSucceededCondition = "BoostrapSucceeded"
+	BootstrapSucceededCondition clusterv1.ConditionType = "BootstrapSucceeded"
```
CecileRobertMichon, line 46: 👍

azure/services/scalesets/scalesets.go
CecileRobertMichon: That's a good question... looks like right now we're not setting conditions at all for the scale set itself because it hasn't yet been refactored (it was implemented as async before the async proposal landed so it's not fully following the same patterns). I would say let's leave this for now and add a note to handle both VMSS and VMSS extension conditions when an error occurs.

azure/services/scalesets/vmssextension_spec.go
```diff
@@ -0,0 +1,70 @@
...
+	return compute.VirtualMachineScaleSetExtension{
+		Name: to.StringPtr(s.Name),
+		VirtualMachineScaleSetExtensionProperties: &compute.VirtualMachineScaleSetExtensionProperties{
+			Publisher:          to.StringPtr(s.Publisher),
+			Type:               to.StringPtr(s.Name),
+			TypeHandlerVersion: to.StringPtr(s.Version),
+			Settings:           nil,
+			ProtectedSettings:  s.ProtectedSettings,
+		},
+		// TODO: should we include location since it's used in VMExtensions too?
```
CecileRobertMichon, line 68: remove the TODO since you've determined location is not included in the API?

CecileRobertMichon (approved):

CecileRobertMichon:
build is failing with 

```
exp/controllers/azuremachinepool_reconciler.go:28:2: no required module provides package sigs.k8s.io/cluster-api-provider-azure/azure/services/vmssextensions; to add it:
	go get sigs.k8s.io/cluster-api-provider-azure/azure/services/vmssextensions
```

CecileRobertMichon (commented):
/lgtm
/assign @mboersma

jackfrancis:
Thanks @Jont828!

Jont828:
@CecileRobertMichon Squashed and pushed, should be good to merge!

Issue #2179, opened 2022-03-18
Remove ClusterResourceSet dependencies from templates


jackfrancis:
/kind feature

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]

1. Decompose the `ClusterResourceSet` template fragments that currently exist inside test and reference templates (e.g., `templates/cluster-template-external-cloud-provider.yaml` and `templates/cluster-template-nvidia-gpu.yaml`) so that they exist on their own for more general re-use. We can create a new reference directory under `templates/` (maybe `templates/clusterresourcesets/` ?) to store these reference `ClusterResourceSet`s.
  - Make sure that the new directory we create for reference `ClusterResourceSet`s are added to the release templates.
2. Remove the existing in-line `ClusterResourceSet` resources in cluster templates under `templates/` and `templates/test/ci/` 
3. Update the E2E implementation so that the complete set of test template-dependent `ClusterResourceSet`s are available by the time our test templates are actually used in test jobs.
4. Update all documentation (including any developer docs about how to test) so that users always pre-install our reference `ClusterResourceSet`s prior to using any reference templates. This will actually encourage users to mix and match to create the cluster specs they need. These are the current reference `ClusterResourceSet`s
  - Include a note about upgrade limitations w/ `ClusterResourceSet`
5. After this merges, we'll want to ensure that we add lots of big headline matter in the release notes for the release that lands these changes.

Extra-credit: replace `ClusterResourceSet` resources with helm charts!

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- cluster-api-provider-azure version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

jackfrancis:
/remove-lifecycle stale

In progress: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2334

PR #2183, opened 2022-03-19
ci: stop referring to ClusterResourceSets as addons

jackfrancis:
**What type of PR is this?**

/kind cleanup

This PR replaces references to "addons" in the various CI-related places in the codebase that use the `ClusterResourceSet` CRD. These `ClusterResourceSets` are used by automation to make it easier to test capz integration stories for common Kubernetes components (e.g., GPU, calico CNI).

Instead of using the overloaded term "addons", let's just refer to clusterresourcesets explicitly. This will aid to effort to gradually decompose the various Kubernetes cluster-api integrations that we want to test from core cluster-api and capz reference templates.

Contributes to https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/2179

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```


CecileRobertMichon:
/hold

As discussed offline, we may not want to do this as it would break the reference in https://cluster-api.sigs.k8s.io/user/quick-start.html#deploy-a-cni-solution

CecileRobertMichon:
/close

As discussed we don't want to do this to avoid breaking users that rely on the current path and since ClusterResourceSets are a temporary solution it's not worth the effort.

PR #2188, opened 2022-03-23
AKS e2e: use latest stable k8s version

### LINKED ISSUE #2127: AKS E2E tests: use latest "stable" k8s version
As AKS is a service that deprecates k8s versions that are out of support policy, using a static (even latest patch) version in e2e test results in those tests eventually failing due to the version becoming unsupported as we saw in #2108.

We should allow specifying `latest` as the AKS k8s version and use the ListOrchestrators AKS API to fetch all supported versions and pick the latest one (excluding the ones marked as "preview").

/area managedcluster
/help

meghanajangi:
**What type of PR is this?**
/kind failing-test

Using a static version in e2e tests is resulting to failure when the version becomes unsupported. Changes in this PR helps fetch the latest stable k8s version for the test when the AKS_KUBERNETES_VERSION is set to latest.

Fixes #2127 

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Fixes the issue where using a static version in AKS e2e test results in failure when the version becomes unsupported.
```


test/e2e/aks.go
```diff
@@ -303,3 +310,48 @@ func GetWorkingAKSKubernetesVersion(ctx context.Context, subscriptionID, locatio
...
+	var maxVersion string
+
+	for _, o := range *result.Orchestrators {
+		orchestratorversions = append(orchestratorversions, *o.OrchestratorVersion)
+		isaksversioninpreview = append(isaksversioninpreview, o.IsPreview)
+	}
+
+	for i := range isaksversioninpreview {
+		if isaksversioninpreview[i] != nil {
+			orchVersion = orchestratorversions[i-1]
```
alexeldeib, line 344: why i-1 and not i here? shouldn't these match?

also this whole logic relies on the ordering returned by AKS, right? I would suggest actually parsing the resulting versions as semver, keep a pointer to the maximum, and do semver compare against each version, updating the max as you go.

alexeldeib:
Thanks for this PR! small comment on the approach, looking pretty good :)

test/e2e/aks.go
alexeldeib: ah, i-1 because the last versions would be preview in an ordered list. so we're finding the version preview version in the array and then taking one item back from the orchestrator version list, assuming that won't be a preview version. 

I would suggest the approach above, which will be much clearer to read in the future as well. I don't believe we guarantee the ordering you're relying on.

test/e2e/aks.go
```diff
@@ -303,3 +310,48 @@ func GetWorkingAKSKubernetesVersion(ctx context.Context, subscriptionID, locatio
 
 	return maxVersion, nil
 }
+
+// GetLatestStableAKSKubernetesVersion returns the latest stable available Kubernetes version of AKS.
+func GetLatestStableAKSKubernetesVersion(ctx context.Context, subscriptionID, location, version string) (string, error) {
```
CecileRobertMichon, line 315: why does this need "version" as input? should always be latest

test/e2e/aks.go
```diff
@@ -303,3 +310,48 @@ func GetWorkingAKSKubernetesVersion(ctx context.Context, subscriptionID, locatio
...
+	}
+
+	var orchestratorversions []string
+	var isaksversioninpreview []*bool
+	var foundWorkingVersion bool
+	var orchVersion string
+	var maxVersion string
+
+	for _, o := range *result.Orchestrators {
+		orchestratorversions = append(orchestratorversions, *o.OrchestratorVersion)
```
CecileRobertMichon, line 338: here could we just skip the ones where `o.IsPreview` is true and not even add them to `orchestratorversions`?

test/e2e/aks.go
```diff
@@ -303,3 +310,48 @@ func GetWorkingAKSKubernetesVersion(ctx context.Context, subscriptionID, locatio
...
+	}
+	authorizer, err := settings.GetAuthorizer()
+	if err != nil {
+		return "", err
+	}
+	containerServiceClient := containerservice.NewContainerServicesClient(subscriptionID)
+	containerServiceClient.Authorizer = authorizer
+	result, err := containerServiceClient.ListOrchestrators(ctx, location, ManagedClustersResourceType)
+	if err != nil {
+		return "", err
```
sonasingh46, line 328: nit: I prefer to wrap errors to have more context while debugging and viewing logs. 
Example: 
```
result, err := containerServiceClient.ListOrchestrators(ctx, location, ManagedClustersResourceType)
if err != nil {
		return "", errors.Wrap(err, "failed to list orchestrators")
	}
```

test/e2e/aks.go
meghanajangi: Sure, I have added this change. Thank you.

test/e2e/aks.go
meghanajangi: Thank you for your comments. 
I have made the change and used semver.Sort to find the maximum version.

test/e2e/aks.go
```diff
@@ -226,8 +226,14 @@ func GetAKSKubernetesVersion(ctx context.Context, e2eConfig *clusterctl.E2EConfi
 	settings, err := auth.GetSettingsFromEnvironment()
 	Expect(err).NotTo(HaveOccurred())
 	subscriptionID := settings.GetSubscriptionID()
-	maxVersion, err := GetWorkingAKSKubernetesVersion(ctx, subscriptionID, location, e2eAKSVersion)
-	Expect(err).NotTo(HaveOccurred())
+	var maxVersion string
+	if e2eAKSVersion == "latest" {
```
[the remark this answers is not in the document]
meghanajangi: I have updated the version to "latest" in azure-dev.yaml and validated that the test works as expected.

sonasingh46 (approved):
/lgtm

PR #2189, opened 2022-03-23
Secure bootstrapping for capz machines

### LINKED ISSUE #915: Secure sensitive bootstrap data
/kind feature

> CAPI generates sensitive cluster data (such as private keys) for the apiserver, etcd, etc. These are stored as secrets in Kubernetes. The kubeadm bootstrapper copies the contents of the secrets into bootstrap data in the KubeadmConfig resource, which is then copied into the Machine resource.
from https://github.com/kubernetes-sigs/cluster-api/issues/1739

CAPZ uses this bootstrap data as the user data for the VM/VMSS. If a user has read-only access to the VM, this could grant them access to the user data, and therefore access to the sensitive data.

Azure recommends _not_ placing any sensitive values in custom data https://docs.microsoft.com/en-us/azure/virtual-machines/custom-data#can-i-place-sensitive-values-in-custom-data.

We should secure the bootstrap data, for example by using Azure keyvault storage to store the data such that only the VM has access to that data, but not a user that has access to the VM.

/priority important-longterm
/milestone next 

**Environment:**

shysank:
**What type of PR is this?**
/kind feature

This PR provides an option to secure bootstrap data for capz machines. This is similar to the way capa works right now. On a high level, it works as follows: If `SecureBootstrapEnabled` is set to true,
1. Creates a azure key vault during cluster reconciliation
2. During machine reconciliation, saves the bootstrap data as secrets in the vault created in (1). The data is split into chunks to make sure the secret doesn't exceed azure keyvault limits.
3. Creates a cloudinit boothook script that: fetches the secrets created in (2) -> write them to a file -> restart cloudinit with the file created as source
4. As the vm is set to provision, set custom data to the script created in (3)

If `SecureBootstrapEnabled` is set to false (or not set), there is no change with the current behaviour.

Limitations:
1. This works only with UserAssigned identity. This is because the identity requires [key vault administrator](https://docs.microsoft.com/en-us/azure/role-based-access-control/built-in-roles#key-vault-administrator) roles for reading and deleting secrets. System assigned identity is not possible since we can only assign role after vm provisioning.
2. Cluster identity also requires [key vault administrator](https://docs.microsoft.com/en-us/azure/role-based-access-control/built-in-roles#key-vault-administrator) roles for creating vault and secrets.
3. `SecureBootstrapEnabled` works for the whole cluster, and not for individual machines (unlike capa). This is because we need to create a KeyVault for managing secrets, and doing it at the machine level will be too cumbersome. We could probably add an override at machine level to disable secure bootstrapping.

** Credits to @randomvariable and the CAPA team for the original implementation **

Fixes #915 (partially)

I have split the pr into logical commits for easier review. 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Secure bootstrapping for capz machines
```


shysank:
cc @CecileRobertMichon

CecileRobertMichon:
@shysank I haven't gone through the code yet but just from looking at the PR description, my initial questions are:

- Seems like this is making an assumption on cloud-init, what about other bootstrap mechanisms that don't leverage cloud-init, like cloudbase init for Windows nodes, Flatcar OS and ignition, or even other bootstrap providers outside of CABPK, like Talos? 
- How does this overlap with https://github.com/kubernetes-sigs/cluster-api/issues/5294 ? 

Perhaps it would make sense to start with a design proposal since this is such a big change/feature.

shysank:
> * Seems like this is making an assumption on cloud-init, what about other bootstrap mechanisms that don't leverage cloud-init, like cloudbase init for Windows nodes, Flatcar OS and ignition, or even other bootstrap providers outside of CABPK, like Talos?
> * How does this overlap with [Cleaner separation of kubeadm and machine bootstrapping cluster-api#5294](https://github.com/kubernetes-sigs/cluster-api/issues/5294) ?
>

The goal of this pr is to provide a way to improve the security of bootstrapping mechanism that leverage cloud init only; bring capz on par with capa in terms of bootstrap data security albeit with some limitations (as mentioned in the pr description).  This is not a generic solution to support other cloud initializers, perhaps we could implement `UserdDataResolver` for each cloud initializer, but we need more foundational changes to our design as mentioned in #5294 for a more robust solution . IMO, this is more of a stop gap solution until we are able separate kubeadm and machine bootstrapping. 


> Perhaps it would make sense to start with a design proposal since this is such a big change/feature.

I'll work on creating a design proposal.

jackfrancis:
@CecileRobertMichon are you able to assess if we should try to salvage this PR?

CecileRobertMichon:
I believe @sonasingh46 was looking into it https://kubernetes.slack.com/archives/CEX9HENG7/p1658756031736239

Regardless, we should start with a proposal as noted above and open a new PR once we're ready.

/close

sonasingh46:
I will work to raise a proposal PR for this.

PR #2194, opened 2022-03-24
ci: Ci entrypoint windows 2022 support

marosset:
**What type of PR is this?**

/kind feature

This PR allows for testing different versions of Windows server without the need to create multiple templates.

For `AzureMachineTemplate` objects with `osType: windows` if `osVersion` attribute will be used to pick which image SKU to use when creating a default image SKU.

Fixes # 2139

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
none
```

### DIFF
--- a/azure/defaults.go
+++ b/azure/defaults.go
@@ -53,6 +53,12 @@ const (
 )
 
 const (
+	// DefaultWindowsOsAndVersion is the default Windows Server version to use when
+	// genearating default images for Windows nodes.
+	DefaultWindowsOsAndVersion = "windows-2019"
+)
+
+const (
 	// Global is the Azure global location value.
 	Global = "global"
 )
@@ -277,12 +283,13 @@ func AvailabilitySetID(subscriptionID, resourceGroup, availabilitySetName string
 }
 
 // GetDefaultImageSKUID gets the SKU ID of the image to use for the provided version of Kubernetes.
-func getDefaultImageSKUID(k8sVersion, os, osVersion string) (string, error) {
+// note: osAndVersion is expected to be in the format of {os}-{version} (ex: unbuntu-2004 or windows-2022)
+func getDefaultImageSKUID(k8sVersion, osAndVersion string) (string, error) {
 	version, err := semver.ParseTolerant(k8sVersion)
 	if err != nil {
 		return "", errors.Wrapf(err, "unable to parse Kubernetes version \"%s\" in spec, expected valid SemVer string", k8sVersion)
 	}
-	return fmt.Sprintf("k8s-%ddot%ddot%d-%s-%s", version.Major, version.Minor, version.Patch, os, osVersion), nil
+	return fmt.Sprintf("k8s-%ddot%ddot%d-%s", version.Major, version.Minor, version.Patch, osAndVersion), nil
 }
 
 // GetDefaultUbuntuImage returns the default image spec for Ubuntu.
@@ -301,7 +308,7 @@ func GetDefaultUbuntuImage(k8sVersion string) (*infrav1.Image, error) {
 		osVersion = "1804"
 	}
 
-	skuID, err := getDefaultImageSKUID(k8sVersion, "ubuntu", osVersion)
+	skuID, err := getDefaultImageSKUID(k8sVersion, fmt.Sprintf("ubuntu-%s", osVersion))
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to get default image")
 	}
@@ -319,7 +326,7 @@ func GetDefaultUbuntuImage(k8sVersion string) (*infrav1.Image, error) {
 }
 
 // GetDefaultWindowsImage returns the default image spec for Windows.
-func GetDefaultWindowsImage(k8sVersion, runtime string) (*infrav1.Image, error) {
+func GetDefaultWindowsImage(k8sVersion, runtime, osAndVersion string) (*infrav1.Image, error) {
 	v122 := semver.MustParse("1.22.0")
 	v, err := semver.ParseTolerant(k8sVersion)
 	if err != nil {
@@ -331,7 +338,11 @@ func GetDefaultWindowsImage(k8sVersion, runtime string) (*infrav1.Image, error) 
 		return nil, errors.New("containerd image only supported in 1.22+")
 	}
 
-	skuID, err := getDefaultImageSKUID(k8sVersion, "windows", "2019")
+	if osAndVersion == "" {
+		osAndVersion = DefaultWindowsOsAndVersion
+	}
+
+	skuID, err := getDefaultImageSKUID(k8sVersion, osAndVersion)
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to get default image")
 	}
--- a/azure/defaults_test.go
+++ b/azure/defaults_test.go
@@ -34,8 +34,7 @@ func TestGetDefaultImageSKUID(t *testing.T) {
 
 	var tests = []struct {
 		k8sVersion     string
-		os             string
-		osVersion      string
+		osAndVersion   string
 		expectedResult string
 		expectedError  bool
 	}{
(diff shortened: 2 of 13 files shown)


marosset:
@CecileRobertMichon @mboersma @jsturtevant @devigned - Can you all take a look and weigh in on the general approach to this?
I split changes into two commits, one for the code changes and one to update the templates.
If people don't have concerns with this approach I can clean up/expand this PR.

My intentions were to figure out how to support different Windows Server OS versions in some of our templates without the need add more templates.

I am pretty sure that with only a little more work we can remove the hard-coded image versions from our CI templates (like [here](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/templates/test/ci/prow-ci-version-windows-containerd-2022/patches/windows-image-update.yaml)).

devigned (commented):
I don't love using annotations for stuff like this, however, I'm not sure where we would add this information to the spec. 

I do like how this makes selecting default Windows image easier.

Would it make sense to add this information to the `AzureMachineTemplate.Spec`?

marosset:
> I don't love using annotations for stuff like this, however, I'm not sure where we would add this information to the spec.
> 
> I do like how this makes selecting default Windows image easier.
> 
> Would it make sense to add this information to the `AzureMachineTemplate.Spec`?

Hmm, maybe we could add it as a new field on [AzureMachineTemplateSpec](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/5fd769c691697d610b29f3ce4cc941282a1444cf/api/v1beta1/azuremachinetemplate_types.go#L25-L27)

If we put this information into API objects we should do the same thing with the "runtime" annotation currently being used for selecting dochershim vs containerd (or wait until 1.23 is out of servicing)...

@devigned would you be opposed to adding the functionality in as an annotation, letting it bake for a while, then promoting the fields api fields?

marosset:
> > If we put this information into API objects we should do the same thing with the "runtime" annotation currently being used for selecting dochershim vs containerd (or wait until 1.23 is out of servicing)...
> 
> The runtime annotation was added as a temporary solution under the agreement that it would be removed once dockershim fell out of support. Is this Windows version annotation also intended to be short-lived?

I think the Windows version annotation is not going to be short-lived.

jsturtevant:
CAPI explicitly doesn't have information about runtime or OS versions in the API.  I don't believe we should introduce those types of things in our Azure API fields.  This would bring tighter coupling to the model and the underlying VMs.  One instance where this could cause trouble is where we encourage customers to build their own VM images. If we introduced a API we would get into situations where customer provided a different OS on the VM.  I would like to avoid this.  

These annotations are really hints out our defaulting logic for the selection of vm images.  We tell users to only use these images as a reference and so users can have a decent first impression without having to go through the process of building images just to try out CAPZ. In a true customer experience, they would be providing their own VMs with OS and not relying on our defaulting logic at all. 

In the Windows case, we (and partner teams) use capz heavily to test new OS features and so we need a way to provide some sort experience that enables these testing scenarios, while minimizing overhead. With Windows in particular the OS is important due to containers needing to match the underlying OS.

The runtime annotation will be short lived but I think the OS is going to be around for awhile (2019 still have several years of support and there will likely be a new WS version shortly after it expires). 

Given this I think annotations are a decent way to give these hints to our Image version defaulter but would be open to other ways.

marosset:
I updated the PR so that the annotation wouldn't be used in any of the 'flavors' templates (based on discussions in slack)

I set `WINDOWS_SERVER_VERSION` to `'"2022"'` and ran /scripts/ci-entrypoint.sh.
it used `cluster-api-provider-azure/templates/test/ci/cluster-template-prow.yaml` and I got a cluster with 2 Windows server 2022 nodes!

jsturtevant:
needing to use the string `'"2022"'` is awkward.  Mark and I looked at the way envsusbt and kustomize where interacting with the value but couldn't figure out. Anyone have thoughts?

jackfrancis:
@jsturtevant what's awkward about "2022" in particular?

jackfrancis:
> all of the quotes are needed so instead of `export WS="2022"` it is `export WS='"2022"'`

O.K. yeah that's horrible, thanks for calling that out! Is there any way we can inject any required double-quote chars later on in the data dispatch flow, so that users don't have to have that burden?

marosset:
> > all of the quotes are needed so instead of `export WS="2022"` it is `export WS='"2022"'`
> 
> O.K. yeah that's horrible, thanks for calling that out! Is there any way we can inject any required double-quote chars later on in the data dispatch flow, so that users don't have to have that burden?

I could not figure out how to get kustomize to create properly create valid deployment files with any other combination of quotes.
If I just set the value to `"2022"` then kustomize would leave not wrap the value in quotes and the deployment file would fail validation because annotations _need_ to be quoted.

Here is a related issue that was closed as won't fix.
https://github.com/kubernetes-sigs/kustomize/issues/3485

I don't have a lot of experience with kustomize so if anyone has any tips let me know!

judyliu-ms:
Hey @devigned @jackfrancis @mboersma @jsturtevant @CecileRobertMichon, is anyone able to help Mark out? We are hoping to merge this PR ASAP as it is blocking another team from making progress!

marosset:
I made an update to work around the quoting issues when trying to set WINDOWS_SERVER_VERSION to a number (ex: 2022).
Now the logic expects that WINDOWS_SERVER_VERSION is set to `windows-2019` or `windows-2022`.
This will keep the necessary quotes in the yaml so everything can deploy correctly.

devigned:
I don't like the annotations, but if they stay out of the flavors, I believe that should keep them relatively hidden. I'd like to see if we can find a better solution as we take on restructuring of marketplace image offers.

azure/defaults.go
```diff
@@ -331,7 +338,13 @@ func GetDefaultWindowsImage(k8sVersion, runtime string) (*infrav1.Image, error)
 		return nil, errors.New("containerd image only supported in 1.22+")
 	}
 
-	skuID, err := getDefaultImageSKUID(k8sVersion, "windows", "2019")
+	if osVersion == "" {
+		osVersion = DefaultWindowsServerVersion
+	} else {
+		osVersion = strings.ReplaceAll(strings.ToLower(osVersion), "windows-", "")
```
CecileRobertMichon, line 344: One thing I noticed: we're trimming the "windows" prefix here but then we end up using it as "os-version" anyways which gets converted back to `windows-osVersion` anyways: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2194/files#diff-b29d010635a6d1d4de222d1f49dfb5312909f7d391d5ada7ac131a84f6848b43L285

This would require just a bit of shuffling so feel free to not do this, but, if you wanted to make the code slightly cleaner we could just change `getDefaultImageSKUID(k8sVersion, os, osVersion string)` to ` getDefaultImageSKUID(k8sVersion, osVersion string)` and just pass in `osVersion` directly if it's set, else default `osVersion` to `"windows-" + DefaultWindowsServerVersion`. It would work for Ubuntu too, we just need to pass in `"ubuntu-" + osVersion` instead of two separate variables.

marosset: I can try doing these changes.

Out of curiosity, would similar functionality be useful for ubuntu too (possibility for testing 22.04 if/when it arrives)?

Right now it looks like the ubuntu version selected is only based on the K8s version (unless explicitly set in the marketplace image reference).

CecileRobertMichon: it could be, are you wondering if you should make the annotation generic? unless it's really easy to do both, I would keep this PR scoped to windows and we can add something similar for ubuntu later if the use case arises. As much as possible we don't want to encourage folks to use annotations as UX for specifying an OS version so let's keep it limited to where it's actually needed.

marosset: The latest commit incorporates these suggestions. PTAL

azure/defaults.go
```diff
@@ -331,7 +338,11 @@ func GetDefaultWindowsImage(k8sVersion, runtime string) (*infrav1.Image, error)
 		return nil, errors.New("containerd image only supported in 1.22+")
 	}
 
-	skuID, err := getDefaultImageSKUID(k8sVersion, "windows", "2019")
+	if osAndVersion == "" {
+		osAndVersion = fmt.Sprintf("windows-%s", DefaultWindowsServerVersion)
```
CecileRobertMichon, line 342: nit: can we just make `windows-` part of the constant?

CecileRobertMichon (commented):
lgtm

azure/defaults.go
marosset: I'll do the same for ubuntu

marosset: updated (and squashed commits)

marosset:
(rebase + push)

CecileRobertMichon (commented):
/approve
/assign @jsturtevant

andyzhangx:
thanks for the PR, so does following env work in test-infra now?
```
          env:
            - name: WINDOWS_FLAVOR
              value: "containerd-2022"
```

marosset:
> thanks for the PR, so does following env work in test-infra now?
> 
> ```
>           env:
>             - name: WINDOWS_FLAVOR
>               value: "containerd-2022"
> ```

For PROW jobs that use ci-entrypoint.sh this is needed to target Windows Server 2022 nodes

```
         env:
            - name: WINDOWS_SERVER_VERSION
              value: "windows-2022"
```

WINDOWS_FLAVOR is used for ci-conformance.sh.

I can make some PRs in test-infra to add a PR and some periodic jobs.

PR #2206, opened 2022-03-30
Apidiff test runs only if the changes are in api/ and exp/api/

### LINKED ISSUE #2051: Change apidiff test to only fail for actual breaking API changes
Right now the apidiff test fails as soon there is a breaking change in an exposed field or in an exposed function. While this is greatly useful for reviews, it results in a lot of false positives since we often have to change exported functions in the codebase and currently accept these changes.

We should consider modifying the job so it only looks at api/ changes since that's the area that is exposed to users where we want to avoid breaking changes. This might mean diverging from what CAPI is doing but it would be more appropriate for now until we either move types and functions we don't want exposed as a library to internal/ similarly to what CAPI did, and/or reach a stable codebase/library.

meghanajangi:
**What type of PR is this?**
/kind failing-test

This PR will limit the apidiff test to run only if there are changes made in the directory api/ and exp/api/

Fixes #2051 

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/scripts/ci-apidiff.sh
+++ b/scripts/ci-apidiff.sh
@@ -20,9 +20,12 @@ set -o pipefail
 
 REPO_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
 
-APIDIFF="${REPO_ROOT}/hack/tools/bin/go-apidiff"
-
-cd "${REPO_ROOT}" && make "${APIDIFF##*/}"
-echo "*** Running go-apidiff ***"
-
-${APIDIFF} "${PULL_BASE_SHA}" --print-compatible
+cd "${REPO_ROOT}"
+APICHANGE=$(git --no-pager diff --name-only FETCH_HEAD)
+if echo "${APICHANGE}" | grep "^api/\|/api/" 
+then
+    echo "*** Running go-apidiff ***"
+    APIDIFF_OLD_COMMIT="${PULL_BASE_SHA}" make apidiff
+else
+    echo "No files under api/ or exp/api/ changed."
+fi
--- a/Makefile
+++ b/Makefile
@@ -191,9 +191,11 @@ clean-temporary: ## Remove all temporary files and folders.
 clean-release: ## Remove the release folder.
 	rm -rf $(RELEASE_DIR)
 
+APIDIFF_OLD_COMMIT ?= $(shell git rev-parse origin/main)
+
 .PHONY: apidiff
 apidiff: $(GO_APIDIFF) ## Check for API differences.
-	$(GO_APIDIFF) $(shell git rev-parse origin/main) --print-compatible
+	$(GO_APIDIFF) $(APIDIFF_OLD_COMMIT) --print-compatible
 
 .PHONY: format-tiltfile
 format-tiltfile: ## Format the Tiltfile.


mboersma:
Thanks @meghanajangi! I've tested the essential command here `git --no-pager diff --name-only FETCH_HEAD | grep "api/"` locally and read up on `FETCH_HEAD`. This seems like it should do the right thing.

But looking more closely, I see that the `pull-cluster-api-provider-azure-apidiff` job still ran as usual. But since this PR itself didn't change anything under `api/` or `exp/api/`, it should have been a no-op.

The `Makefile` actually isn't used in this case. Instead, `./scripts/ci-apidiff.sh` is called directly by the job defined in [kubernetes/test-infra](https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-sigs/cluster-api-provider-azure/cluster-api-provider-azure-presubmits-main.yaml#L376).

If you can make a similar change to the `ci-apidiff.sh` script, it should work in all cases.

mboersma:
Another approach that might be preferable in this case is to [configure the presubmit prow job](https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#how-to-configure-new-jobs) with a `run_if_changed` regular expression. Then it won't even show up in the list of GitHub checks below unless it was needed.

Also see [testing prow jobs locally](https://github.com/kubernetes/test-infra/blob/master/config/jobs/README.md#testing-jobs-locally).

meghanajangi:
> Another approach that might be preferable in this case is to [configure the presubmit prow job](https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#how-to-configure-new-jobs) with a `run_if_changed` regular expression. Then it won't even show up in the list of GitHub checks below unless it was needed.
> 
> Also see [testing prow jobs locally](https://github.com/kubernetes/test-infra/blob/master/config/jobs/README.md#testing-jobs-locally).

Thank you for the comments @mboersma . I made the changes to ci-apidiff script as you suggested in the first comment. 
I also made a change in the script and Makefile similar to this [CAPI PR to deduplicate apidiff logic.](https://github.com/kubernetes-sigs/cluster-api/pull/6180)

mboersma:
/retest

Unrelated provisioning flake.

mboersma (approved):
/lgtm

CecileRobertMichon:
/approve

thank you!

scripts/ci-apidiff.sh
```diff
@@ -20,9 +20,12 @@ set -o pipefail
...
-APIDIFF="${REPO_ROOT}/hack/tools/bin/go-apidiff"
-
-cd "${REPO_ROOT}" && make "${APIDIFF##*/}"
-echo "*** Running go-apidiff ***"
-
-${APIDIFF} "${PULL_BASE_SHA}" --print-compatible
\ No newline at end of file
+cd "${REPO_ROOT}"
+APICHANGE=$(git --no-pager diff --name-only FETCH_HEAD)
+if echo "${APICHANGE}" | grep "api/"
```
jackfrancis, line 25: I think if we grep'd for `"/api/"` (both forward and trailing slashes) we'd be a bit more defensive (in a good way) going forward. For example, if someone created a subdirectory somewhere with a substring suffix of "capi" this would match, which we wouldn't want.

Thanks for doing this!

CecileRobertMichon:
/hold
/lgtm cancel

to address @jackfrancis' comment above

meghanajangi:
@jackfrancis @CecileRobertMichon I have made the requested change. PTAL :upside_down_face:

scripts/ci-apidiff.sh
mboersma: I don't think this change with the leading slash will work correctly, though:

```shell
% git --no-pager diff --name-only FETCH_HEAD | grep "/api/"
% echo dummy >> api/v1beta1/doc.go                         
% git --no-pager diff --name-only FETCH_HEAD | grep "/api/"
% git --no-pager diff --name-only FETCH_HEAD | grep "api/" 
api/v1beta1/doc.go

```

mboersma (approved):
/lgtm

This should work, as far as I can tell in local testing.

Thanks for your patience @meghanajangi!

jackfrancis (approved):
lgtm

jackfrancis:
@mboersma after this merges could you create a draft PR that validates this? (I'd do it myself, but I'm not familiar enough with how to trip up apidiff using present config)

mboersma:
> could you create a draft PR that validates this?

Good call–sure, I'll do that.

PR #1889, opened 2021-11-24
[WIP] Install external cloud-provider-azure in reference templates and tests

CecileRobertMichon:
**What type of PR is this?**

**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note

```

PR #2209, opened 2022-03-31
e2e: use helm to install out-of-tree cloud-provider-azure

jackfrancis:
**What type of PR is this?**

/kind feature

This PR updates the "external" (or out-of-tree) cloud-provider-azure templates and test implementation so that we use the official helm chart (see this PR: https://github.com/kubernetes-sigs/cloud-provider-azure/pull/1306).

This approach has the following advantages:

- rather than define static versions of cloud-provider-azure to install, we rely upon the helm chart to choose the right version for the version of Kubernetes running in the cluster
- no longer rely upon `ClusterResourceSet` CRD, which is alpha and may not be enabled on your cluster-api mgmt cluster

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
replace ClusterResourceSet with helm for external cloud-provider-azure templates
```


templates/cluster-template-external-cloud-provider.yaml
```diff
@@ -226,338 +226,3 @@ spec:
     namespace: ${AZURE_CLUSTER_IDENTITY_SECRET_NAMESPACE}
   tenantID: ${AZURE_TENANT_ID}
   type: ServicePrincipal
----
-apiVersion: addons.cluster.x-k8s.io/v1beta1
```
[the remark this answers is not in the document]
jackfrancis: Is that possible? Do we have a precedent for such a workflow after deploying the template using tilt?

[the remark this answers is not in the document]
jsturtevant: we do this for other CRS: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/c55fe5509ab16e191978b570084b1605e6a05217/Tiltfile#L230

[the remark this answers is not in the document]
CecileRobertMichon: Can you please update docs https://capz.sigs.k8s.io/topics/cloud-provider-config.html#external-cloud-provider too?

test/e2e/azure_test.go
```diff
@@ -450,6 +450,10 @@ var _ = Describe("Workload cluster creation", func() {
 				WaitForClusterIntervals:      e2eConfig.GetIntervals(specName, "wait-cluster"),
 				WaitForControlPlaneIntervals: e2eConfig.GetIntervals(specName, "wait-control-plane"),
 				WaitForMachinePools:          e2eConfig.GetIntervals(specName, "wait-machine-pool-nodes"),
+				ControlPlaneWaiters: clusterctl.ControlPlaneWaiters{
+					WaitForControlPlaneInitialized:   InstallCloudProviderAzureHelmChart,
```
CecileRobertMichon, line 454: Does this completely replace the helper to wait for certain conditions on the control planes or does is add the install cloud provider action _before_ waiting for the control plane to be initialized? ie. Are we still waiting for the Control Plane machines to be initialized + Ready in this test?

jackfrancis: I think what you may be implying is that we can use the default `WaitForControlPlaneMachinesReady` and simply "overwrite" the `WaitForControlPlaneMachinesReady` (which is the part where the missing providerID will short-circuit E2E progress). Is that right?

jackfrancis: Good find, will do

templates/cluster-template-external-cloud-provider.yaml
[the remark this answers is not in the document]
jsturtevant: does this mean the external cloud provider flavor in tilt will need a manual step to deploy the helm chart after the cluster is up?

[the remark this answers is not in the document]
jackfrancis: @jsturtevant yes

@CecileRobertMichon done

test/e2e/cloud-provider-azure.go
```diff
@@ -0,0 +1,116 @@
...
+	clusterProxy := input.ClusterProxy.GetWorkloadCluster(ctx, input.ConfigCluster.Namespace, input.ConfigCluster.ClusterName)
+	Eventually(func() error {
+		clientSet := clusterProxy.GetClientSet()
+		_, err := clientSet.CoreV1().Nodes().List(ctx, v1.ListOptions{})
+		return err
+	}, 20*time.Minute, 5*time.Second).Should(Succeed())
+	kubeConfigPath := clusterProxy.GetKubeconfigPath()
+	os.Setenv("KUBECONFIG", kubeConfigPath)
+	Eventually(func() error {
+		settings := cli.New()
```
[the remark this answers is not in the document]
jackfrancis: aliased all helm deps

test/e2e/cloud-provider-azure.go
```diff
@@ -0,0 +1,116 @@
...
+		actionConfig := new(action.Configuration)
+		if err := actionConfig.Init(settings.RESTClientGetter(), "default", os.Getenv("HELM_DRIVER"), Logf); err != nil {
+			return err
+		}
+		i := action.NewInstall(actionConfig)
+		i.RepoURL = "https://raw.githubusercontent.com/kubernetes-sigs/cloud-provider-azure/master/helm/repo"
```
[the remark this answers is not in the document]
jackfrancis: done

test/e2e/azure_test.go
jackfrancis: locally tested out this approach, thanks for the suggestion, a big improvement!

docs/book/src/topics/cloud-provider-config.md
```diff
@@ -81,6 +89,10 @@ kube-system   cloud-node-manager-mfsdg
 kube-system   cloud-node-manager-qrz74                                            1/1     Running   0          24s
 ```
 
+For more information on the `cloud-provider-azure` helm chart see the official documentation [here](https://github.com/kubernetes-sigs/cloud-provider-azure/tree/master/helm/cloud-provider-azure).
```
jsturtevant, line 92:
```suggestion
For more information on the `cloud-provider-azure` helm chart see the official [cloud provider documentation](https://github.com/kubernetes-sigs/cloud-provider-azure/tree/master/helm/cloud-provider-azure).
```

docs/book/src/topics/cloud-provider-config.md
```diff
@@ -81,6 +89,10 @@ kube-system   cloud-node-manager-mfsdg
...
+
+If you're not familiar with using helm to manage Kubernetes applications as packages, there's lots of good documentation [here](https://helm.sh/docs/).
```
jsturtevant, line 94:
```suggestion
If you're not familiar with using helm to manage Kubernetes applications as packages, there's lots of good [Helm documentation](https://helm.sh/docs/) website.
```

avoid using `here` for links to help users with screen readers

test/e2e/cloud-provider-azure.go
```diff
@@ -0,0 +1,143 @@
...
+	}, 20*time.Minute, 5*time.Second).Should(Succeed())
+	clusterProxy := input.ClusterProxy.GetWorkloadCluster(ctx, input.ConfigCluster.Namespace, input.ConfigCluster.ClusterName)
+	By("Waiting for nodes to come online")
+	Eventually(func() error {
+		clientSet := clusterProxy.GetClientSet()
+		n, err := clientSet.CoreV1().Nodes().List(ctx, v1.ListOptions{})
+		if err != nil {
+			return err
+		}
+		if len(n.Items) == (int(*input.ConfigCluster.WorkerMachineCount) + int(*input.ConfigCluster.ControlPlaneMachineCount)) {
```
jsturtevant, line 63: This test doesn't run for windows currently, but if it did we would need to include `WINDOWS_WORKER_MACHINE_COUNT` here.

templates/cluster-template-external-cloud-provider.yaml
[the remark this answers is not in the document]
jsturtevant: we might be able to extend the logic that is called on https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/c55fe5509ab16e191978b570084b1605e6a05217/Tiltfile#L336  to wait for cluster to finish provisioning then do some action (install helm).  But sounds like there are some other mechanism like the addon proposal in the works that would make this cleaner/easier.

test/e2e/cloud-provider-azure.go
jackfrancis: Actually this brings up a good question. On this test, for example, this `WorkerMachineCount` value seems to apply to each node pool (not total number of expected worker nodes):

https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/v1.2.1/test/e2e/azure_test.go#L216

And indeed, this value seems to be used to populate the named env var that kustomize uses to configure a single value which by convention is shared across all `MachinePool`s/`MachineDeployment`s:

https://github.com/kubernetes-sigs/cluster-api/blob/v1.1.3/cmd/clusterctl/client/config.go#L436

This check is really here to know that we're ready to reliably perform a `helm install` against the cluster. We can simply check for 1 Running node to do that.

This will work for a Windows-enabled cluster.

Do we want to just convert the reference "external-cloud-provider" template to have a Windows `MachineDeployment` as well? @CecileRobertMichon for thoughts as well

templates/cluster-template-external-cloud-provider.yaml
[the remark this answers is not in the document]
jackfrancis: Good idea! I've pushed up something that I tested out locally.

docs/book/src/topics/cloud-provider-config.md
jackfrancis: updated, thx

test/e2e/cloud-provider-azure.go
CecileRobertMichon: IMO we should not change the existing external cloud provider template in this PR but instead change _all_ the reference templates to use external cloud provider as we had started doing in #1889 in a follow PR. That will include various windows flavors.

jackfrancis: Makes sense

test/e2e/cloud-provider-azure.go
```diff
@@ -0,0 +1,171 @@
...
+		i.RepoURL = "https://raw.githubusercontent.com/kubernetes-sigs/cloud-provider-azure/master/helm/repo"
+		i.ReleaseName = "cloud-provider-azure-oot"
+		cp, err := i.ChartPathOptions.LocateChart("cloud-provider-azure", helmCli.New())
+		if err != nil {
+			return err
+		}
+		p := helmGetter.All(settings)
+		valueOpts := &helmVals.Options{}
+		valueOpts.Values = []string{fmt.Sprintf("infra.clusterName=%s", input.ClusterProxy.GetName())}
+		if imageRegistryFromEnv := os.Getenv("IMAGE_REGISTRY"); imageRegistryFromEnv != "" {
```
jackfrancis, line 93: @CecileRobertMichon @jsturtevant the below is a proposed solution to ensuring these changes are back-compat with upstream cloud-provider-azure test-infra jobs that rely upon capz to test out-of-tree. This is how the current tests currently work:

1. `TEST_CCM=true` is set when calling `scripts/ci-entrypoint.sh`:
  - https://github.com/kubernetes/test-infra/blob/f74bc266fbbff04f30769fbfcd736634c90d3b5f/config/jobs/kubernetes-sigs/cloud-provider-azure/cloud-provider-azure-config.yaml#L185
2. capz maintains a script to set the CI-built image data here:
  - https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/1352fa2be39e1431f6beb62ecaa014471435b5ca/scripts/ci-entrypoint.sh#L91
  - https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/1352fa2be39e1431f6beb62ecaa014471435b5ca/scripts/ci-build-azure-ccm.sh#L43
  - https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/1352fa2be39e1431f6beb62ecaa014471435b5ca/scripts/ci-build-azure-ccm.sh#L34-L36
  - https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/1352fa2be39e1431f6beb62ecaa014471435b5ca/scripts/ci-build-azure-ccm.sh#L44

These env vars aren't as specifically named as we'd ideally want, but re-using them as-is allows us to reduce the change surface area (don't have to update capz CI scripts and/or test-infra job defs).

templates/cluster-template-external-cloud-provider.yaml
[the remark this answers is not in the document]
jackfrancis: This is the tilt output for the `"external-cloud-provider"` flavor with the new changes:

```
wget -qO- https://github.com/projectcalico/calico/releases/download/v3.22.1/release-v3.22.1.tgz | tar xz release-v3.22.1/manifests/calico-vxlan.yaml release-v3.22.1/manifests/calico-policy-only.yaml
mv release-v3.22.1/manifests/calico-vxlan.yaml templates/addons/calico
mv release-v3.22.1/manifests/calico-policy-only.yaml templates/addons/calico-ipv6
/Users/jackfrancis/go/src/sigs.k8s.io/cluster-api-provider-azure/hack/tools/bin/kustomize-v4.5.2 build templates/addons/metrics-server > templates/addons/metrics-server/metrics-server.yaml
/Users/jackfrancis/go/src/sigs.k8s.io/cluster-api-provider-azure/hack/tools/bin/kustomize-v4.5.2 build templates/addons/calico > templates/addons/calico.yaml
/Users/jackfrancis/go/src/sigs.k8s.io/cluster-api-provider-azure/hack/tools/bin/kustomize-v4.5.2 build templates/addons/calico-ipv6 > templates/addons/calico-ipv6.yaml
/Users/jackfrancis/go/src/sigs.k8s.io/cluster-api-provider-azure/hack/tools/bin/kustomize-v4.5.2 build templates/addons/calico-dual-stack > templates/addons/calico-dual-stack.yaml
./hack/gen-flavors.sh
cluster.cluster.x-k8s.io/external-cloud-provider-29082 created
azurecluster.infrastructure.cluster.x-k8s.io/external-cloud-provider-29082 created
kubeadmcontrolplane.controlplane.cluster.x-k8s.io/external-cloud-provider-29082-control-plane created
azuremachinetemplate.infrastructure.cluster.x-k8s.io/external-cloud-provider-29082-control-plane created
machinedeployment.cluster.x-k8s.io/external-cloud-provider-29082-md-0 created
azuremachinetemplate.infrastructure.cluster.x-k8s.io/external-cloud-provider-29082-md-0 created
kubeadmconfigtemplate.bootstrap.cluster.x-k8s.io/external-cloud-provider-29082-md-0 created
azureclusteridentity.infrastructure.cluster.x-k8s.io/cluster-identity unchanged
Cluster 'external-cloud-provider-29082' created, don't forget to delete
NAME: cloud-provider-azure-1649106673
LAST DEPLOYED: Mon Apr  4 14:11:16 2022
NAMESPACE: default
STATUS: deployed
REVISION: 1
TEST SUITE: None
```

test/e2e/cloud-provider-azure.go
CecileRobertMichon: the out of tree cloud provider PR tests and conformance tests don't run CAPZ e2e so they won't call `InstallCloudProviderAzureHelmChart` as far as I know. We may need to 1) modify ci-entrypoint.sh directly to do the helm install for out of tree cases, and 2) change conformance_test.go to use this custom control plane waiter too when testing out of tree

CecileRobertMichon: /cc @lzhecheng

jackfrancis: I don't think we have to worry about the conformance stuff yet (in this PR), it seems the only template flavors that it currently uses are default and windows (not the external templates).

When we eventually get to the part of this effort where we're testing out-of-tree by default, then yeah, the conformance `ControlPlaneWaiters` will need to get updated as well.

scripts/ci-entrypoint.sh
```diff
@@ -152,6 +154,11 @@ create_cluster
 # export the target cluster KUBECONFIG if not already set
 export KUBECONFIG="${KUBECONFIG:-${PWD}/kubeconfig}"
 
+# install cloud-provider-azure components, if using out-of-tree
```
jackfrancis, line 157: @CecileRobertMichon @lzhecheng incorporated the helm stuff into the `ci-entrypoint.sh` script itself. The way this will work is basically the same way I implemented this in the `Tiltfile` (wait for kubeconfig secret, wait for successful get nodes). Reference:

https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/1352fa2be39e1431f6beb62ecaa014471435b5ca/Makefile#L283-L287

.gitignore
```diff
@@ -56,6 +56,7 @@ bazel-*
 /tilt.d
 tilt-settings.json
 tilt_config.json
+*.kubeconfig
```
CecileRobertMichon, line 59: can you please move this up to line 32 with the other kubeconfig files?

scripts/ci-entrypoint.sh
```diff
@@ -152,6 +154,11 @@ create_cluster
...
+if [[ -n "${TEST_CCM:-}" ]]; then
+  "${HELM}" install --repo https://raw.githubusercontent.com/kubernetes-sigs/cloud-provider-azure/master/helm/repo cloud-provider-azure --generate-name --set infra.clusterName="${CLUSTER_NAME}"
```
CecileRobertMichon, line 159: how does this work when using a custom image of CCM and CNM that we just built above?

jackfrancis: whoops

.gitignore
jackfrancis: done

scripts/ci-entrypoint.sh
jackfrancis: done, thanks for catching this!

CecileRobertMichon (commented):
/lgtm
/assign @jsturtevant @lzhecheng

scripts/ci-entrypoint.sh
```diff
@@ -152,6 +154,16 @@ create_cluster
 # export the target cluster KUBECONFIG if not already set
 export KUBECONFIG="${KUBECONFIG:-${PWD}/kubeconfig}"
 
+# install cloud-provider-azure components, if using out-of-tree
+if [[ -n "${TEST_CCM:-}" ]]; then
+  "${HELM}" install --repo https://raw.githubusercontent.com/kubernetes-sigs/cloud-provider-azure/master/helm/repo cloud-provider-azure --generate-name --set infra.clusterName="${CLUSTER_NAME}" --set cloudControllerManager.imageRepository="${IMAGE_REGISTRY}" \
```
CecileRobertMichon, line 159: @lzhecheng @jackfrancis is there still a plan to use gh pages to make an official helm repo for cloud-provider-azure?

jackfrancis: I'm not sure a plan was ever formulated. I'd be in favor of doing that, if the cloud-provider-azure maintainers agree.

Filed a proposal issue here:

https://github.com/kubernetes-sigs/cloud-provider-azure/issues/1418

docs/book/src/developers/kubernetes-developers.md
```diff
@@ -35,13 +35,56 @@ export CLUSTER_TEMPLATE="test/dev/cluster-template-custom-builds.yaml"
 
 ## Testing the out-of-tree cloud provider
 
-To test changes made to the [Azure cloud provider](https://github.com/kubernetes-sigs/cloud-provider-azure), first build and push images for cloud-controller-manager and/or cloud-node-manager from the root of the cloud-provider-azure repo.
+To test changes made to the [Azure cloud provider](https://github.com/kubernetes-sigs/cloud-provider-azure), first build and push images for cloud-controller-manager and/or cloud-node-manager from the branch of the cloud-provider-azure repo that the desired changes are in. Based on the repository, image name, and image tag you produce from your custom image build and push, set the appropriate environment variables below:
```
jackfrancis, line 38: @CecileRobertMichon FYI a new set of documentation changes since your most recent review

templates/flavors/external-cloud-provider/ccm-resource-set.yaml
```diff
@@ -1,27 +0,0 @@
-apiVersion: addons.cluster.x-k8s.io/v1beta1
```
jackfrancis, line 1: @CecileRobertMichon FYI these three yaml files are deleted since your last review

templates/flavors/external-cloud-provider/cloud-controller-manager.yaml
```diff
@@ -1,198 +0,0 @@
...
-  hostNetwork: true
-  nodeSelector:
-    node-role.kubernetes.io/master: ""
-  serviceAccountName: cloud-controller-manager
-  tolerations:
-    - key: node-role.kubernetes.io/master
-      effect: NoSchedule
-  containers:
-    - name: cloud-controller-manager
-      image: ${AZURE_CLOUD_CONTROLLER_MANAGER_IMG:=mcr.microsoft.com/oss/kubernetes/azure-cloud-controller-manager:v1.1.5}
```
jackfrancis, line 151: The `AZURE_CLOUD_CONTROLLER_MANAGER_IMG` env var is referenced in other parts of the codebase to fulfill this envsubst use case. Now that we're removing this template partial, we'll remove this env var from various foo as well. (Ditto `AZURE_CLOUD_NODE_MANAGER_IMG` in the `cloud-node-manager.yaml` partial.

scripts/ci-build-azure-ccm.sh
```diff
@@ -43,8 +43,6 @@ setup() {
     export IMAGE_REGISTRY=${REGISTRY}
     pushd "${AZURE_CLOUD_PROVIDER_ROOT}" && IMAGE_TAG=$(git rev-parse --short=7 HEAD) && export IMAGE_TAG && popd
     echo "Image Tag is ${IMAGE_TAG}"
-    export AZURE_CLOUD_CONTROLLER_MANAGER_IMG=${IMAGE_REGISTRY}/${CCM_IMAGE_NAME}:${IMAGE_TAG}
```
jackfrancis, line 46: We no longer need to compose the `AZURE_CLOUD_CONTROLLER_MANAGER_IMG` and `AZURE_CLOUD_NODE_MANAGER_IMG` env vars because the templates that were expecting those env vars no longer exist.

docs/book/src/developers/kubernetes-developers.md
CecileRobertMichon: @nilo19 @lzhecheng could you please review this part?

nilo19: lgtm

scripts/ci-entrypoint.sh
lzhecheng: Good idea!

docs/book/src/developers/kubernetes-developers.md
lzhecheng: lgtm. Thanks!

CecileRobertMichon (commented):
/lgtm
/assign @jsturtevant

Makefile
```diff
@@ -740,12 +743,23 @@ $(KUBECTL): ## Build kubectl from tools folder.
 	ln -sf $(KUBECTL) $(TOOLS_BIN_DIR)/$(KUBECTL_BIN)
 	chmod +x $(KUBECTL) $(TOOLS_BIN_DIR)/$(KUBECTL_BIN)
 
+$(HELM): ## Put helm into tools folder.
+	mkdir -p $(TOOLS_BIN_DIR)
+	rm -f "$(TOOLS_BIN_DIR)/$(HELM_BIN)*"
+	curl -fsSL -o $(TOOLS_BIN_DIR)/get_helm.sh https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3
```
jsturtevant, line 749: Do we want to pin a version to eliminate chance that something in helm breaks tests?

test/e2e/cloud-provider-azure.go
```diff
@@ -0,0 +1,149 @@
...
+	By("Installing the correct version of cloud-provider-azure components via helm")
+	Eventually(func() error {
+		settings := helmCli.New()
+		actionConfig := new(helmAction.Configuration)
+		if err := actionConfig.Init(settings.RESTClientGetter(), "default", os.Getenv("HELM_DRIVER"), Logf); err != nil {
+			return err
+		}
+		i := helmAction.NewInstall(actionConfig)
+		i.RepoURL = "https://raw.githubusercontent.com/kubernetes-sigs/cloud-provider-azure/master/helm/repo"
+		i.ReleaseName = "cloud-provider-azure-oot"
```
jsturtevant, line 85: The release here could be a different patch version each time? It may be hard to determine if a change in the cloud-provider broke or something in capz.  Not sure there is a good option here as we want to be able to have the helm chart pick the right version based on k8s verion just calling it out.

jsturtevant (commented):
/lgtm

Makefile
jackfrancis: The trade-off is:

1. Manually update helm versions regularly to ensure you have a patched version
2. Permissively get the latest

Based on the maturity of the helm project, I prefer option 2. Easy to do option 1 if that's what folks want.

test/e2e/cloud-provider-azure.go
jackfrancis: The name of the release is just a friendly, unique identifier for each "helm install" (and/or subsequent "helm upgrade"). Because the E2E flow is never helm install'ing more than once, I'm just choosing a static string (any string will do).

Makefile
CecileRobertMichon: This comment made me think of something I missed before: the way the Makefile works, because the binary is unversioned, it will install it once and then never install it again if the file "hack/tools/bin/helm" is already present. This means unless devs run `make clean` they won't get the latest version of Helm and their local hack/tools version will get stale. That's the reason why we use a versioned binary for the other tools, with a symlink to the simple binary name so it can be referenced in scripts without the version needing updating whenever we update the version.

For that reason, I'm voting we do 1) and pin to a version so that we ensure the local version folks use is always up to date with our CI version, and follow the same pattern as other hack/took binaries.

jackfrancis:
/hold

found a bug in ci-entrypoint.sh flow

Makefile
```diff
@@ -236,6 +241,9 @@ verify-tiltfile: ## Verify Tiltfile format.
 
 ##@ Development:
 
+.PHONY: install-tools # populate hack/tools/bin
+install-tools: $(ENVSUBST) $(KUSTOMIZE) $(KUBECTL) $(HELM) $(GINKGO)
```
jackfrancis, line 245: Unifying these installs together made it a lot easier to standardize dep installs across the various jobs and scripts.

Should I go ahead and add the remaining set of tools in hack/tools/bin to this list for best practice going forward (despite the fact that not every invocation of `make install-tools` will require *every single* binary)?

cc @CecileRobertMichon @jsturtevant

CecileRobertMichon: > despite the fact that not every invocation of make install-tools will require every single binary

I'm weak -1 on this, even if it improves the makefile a little, it would make running, say `make go-test`, a lot slower the first time and install a bunch of binaries that aren't needed, I slightly prefer the "just in time" install approach.

scripts/ci-entrypoint.sh
```diff
@@ -152,6 +155,17 @@ create_cluster
 # export the target cluster KUBECONFIG if not already set
 export KUBECONFIG="${KUBECONFIG:-${PWD}/kubeconfig}"
 
+# install cloud-provider-azure components, if using out-of-tree
+if [[ -n "${TEST_CCM:-}" ]]; then
```
jackfrancis, line 159: Validated this new ccm + helm scenario:

```sh
s$ TEST_CCM=true AZURE_LOADBALANCER_SKU=standard CONTROL_PLANE_MACHINE_COUNT=3 REGISTRY=docker.io/jackfrancis KUBERNETES_VERSION=1.23.5 ./scripts/ci-entrypoint.sh
...
# Get kubeconfig and store it locally.
kubectl get secrets capz-q4tba7-kubeconfig -o json | jq -r .data.value | base64 --decode > ./kubeconfig
timeout --foreground 600 bash -c "while ! kubectl --kubeconfig=./kubeconfig get nodes | grep control-plane; do sleep 1; done"
error: the server doesn't have a resource type "nodes"
capz-q4tba7-control-plane-s95kf   NotReady   control-plane,master   6s    v1.23.5
run "kubectl --kubeconfig=./kubeconfig ..." to work with the new target cluster
WARNING: Kubernetes configuration file is group-readable. This is insecure. Location: /Users/jackfrancis/go/src/sigs.k8s.io/cluster-api-provider-azure/kubeconfig
WARNING: Kubernetes configuration file is world-readable. This is insecure. Location: /Users/jackfrancis/go/src/sigs.k8s.io/cluster-api-provider-azure/kubeconfig
Installing cloud-provider-azure components via helm
NAME: cloud-provider-azure-1649354900
LAST DEPLOYED: Thu Apr  7 11:08:22 2022
NAMESPACE: default
STATUS: deployed
REVISION: 1
TEST SUITE: None
```

Manually validated that things looked as expected:

```sh
$ k get pods -A
NAMESPACE     NAME                                                      READY   STATUS    RESTARTS        AGE
kube-system   calico-kube-controllers-7c664f4648-sth6w                  1/1     Running   0               6m49s
kube-system   calico-node-79nft                                         1/1     Running   0               6m49s
kube-system   calico-node-mjvl4                                         1/1     Running   0               5m39s
kube-system   calico-node-sn5mg                                         1/1     Running   0               5m42s
kube-system   calico-node-svsnb                                         1/1     Running   0               3m9s
kube-system   calico-node-v5dhf                                         1/1     Running   0               5m9s
kube-system   cloud-controller-manager-69fcc57c84-b69pq                 1/1     Running   1 (4m57s ago)   6m43s
kube-system   cloud-node-manager-4n7cf                                  1/1     Running   0               6m43s
kube-system   cloud-node-manager-565r4                                  1/1     Running   0               5m42s
kube-system   cloud-node-manager-ck497                                  1/1     Running   0               3m9s
kube-system   cloud-node-manager-qr9zw                                  1/1     Running   0               5m8s
kube-system   cloud-node-manager-qvhlw                                  1/1     Running   0               5m39s
kube-system   coredns-64897985d-fhhqj                                   1/1     Running   0               6m50s
kube-system   coredns-64897985d-gtnjp                                   1/1     Running   0               6m50s
kube-system   etcd-capz-q4tba7-control-plane-7wss9                      1/1     Running   0               3m7s
kube-system   etcd-capz-q4tba7-control-plane-gpplv                      1/1     Running   0               5m7s
kube-system   etcd-capz-q4tba7-control-plane-s95kf                      1/1     Running   0               6m45s
kube-system   kube-apiserver-capz-q4tba7-control-plane-7wss9            1/1     Running   0               3m8s
kube-system   kube-apiserver-capz-q4tba7-control-plane-gpplv            1/1     Running   0               5m8s
kube-system   kube-apiserver-capz-q4tba7-control-plane-s95kf            1/1     Running   0               6m54s
kube-system   kube-controller-manager-capz-q4tba7-control-plane-7wss9   1/1     Running   0               3m8s
kube-system   kube-controller-manager-capz-q4tba7-control-plane-gpplv   1/1     Running   0               5m8s
kube-system   kube-controller-manager-capz-q4tba7-control-plane-s95kf   1/1     Running   1 (4m58s ago)   6m54s
kube-system   kube-proxy-74nrq                                          1/1     Running   0               5m9s
kube-system   kube-proxy-79bj8                                          1/1     Running   0               6m50s
kube-system   kube-proxy-88dl6                                          1/1     Running   0               3m9s
kube-system   kube-proxy-brhzh                                          1/1     Running   0               5m39s
kube-system   kube-proxy-x72bc                                          1/1     Running   0               5m42s
kube-system   kube-scheduler-capz-q4tba7-control-plane-7wss9            1/1     Running   0               3m8s
kube-system   kube-scheduler-capz-q4tba7-control-plane-gpplv            1/1     Running   0               5m7s
kube-system   kube-scheduler-capz-q4tba7-control-plane-s95kf            1/1     Running   1 (4m57s ago)   6m45s
$ k get pod cloud-controller-manager-69fcc57c84-b69pq -n kube-system -o yaml | grep 'image:'
    image: docker.io/jackfrancis/azure-cloud-controller-manager:4f3bab6
    image: docker.io/jackfrancis/azure-cloud-controller-manager:4f3bab6
k get pod cloud-node-manager-4n7cf -n kube-system -o yaml | grep 'image:'
    image: docker.io/jackfrancis/azure-cloud-node-manager:4f3bab6
    image: docker.io/jackfrancis/azure-cloud-node-manager:4f3bab6
```

Makefile
jackfrancis: Cool, yeah, that's the tradeoff.

Any concerns w/ this new `install-tools` target here? The practical problem is we have > 1 (actually like 3 or 4) places in > 1 file that need the above (or at least 4 of 5) installed, and during this effort I encountered how easy it is to miss a dep when adding a new one if the only option is to copy/paste those individual tool install commands around.

test/e2e/cloud-provider-azure.go
```diff
@@ -0,0 +1,149 @@
...
+	By("Waiting for workload cluster kubeconfig secret")
+	Eventually(func() error {
+		client := input.ClusterProxy.GetClient()
+		secret := &corev1.Secret{}
+		key := crclient.ObjectKey{
+			Name:      fmt.Sprintf("%s-kubeconfig", input.ConfigCluster.ClusterName),
+			Namespace: input.ConfigCluster.Namespace,
+		}
+		return client.Get(ctx, key, secret)
+	}, 20*time.Minute, 5*time.Second).Should(Succeed())
```
CecileRobertMichon, line 54: should this be using a config interval variable here instead of hardcoding to 20 minutes? same question for intervals below

test/e2e/cloud-provider-azure.go
```diff
@@ -0,0 +1,149 @@
+//go:build e2e
+// +build e2e
+
+/*
+Copyright 2020 The Kubernetes Authors.
```
CecileRobertMichon, line 5:
```suggestion
Copyright 2022 The Kubernetes Authors.
```

test/e2e/cloud-provider-azure.go
```diff
@@ -0,0 +1,149 @@
...
+				runningNodes++
+			}
+		}
+		if runningNodes > 0 {
+			return nil
+		}
+		return err
+	}, 20*time.Minute, 5*time.Second).Should(Succeed())
+	kubeConfigPath := clusterProxy.GetKubeconfigPath()
+	os.Setenv("KUBECONFIG", kubeConfigPath)
```
CecileRobertMichon, line 75: why do we need to export a var here? Any way we can pass in the kubeconfig to Helm another way so we don't mess with env vars? If not, we should add a defer statement to unset the var.

jackfrancis: There's gotta be a way to pass in the filepath somewhere in the flow, as the CLI (which uses these libs under the hood) supports that.

```sh
$ helm install --help | grep kubeconfig
WARNING: Kubernetes configuration file is group-readable. This is insecure. Location: /Users/jackfrancis/.kube/external-cloud-provider-24053.kubeconfig
WARNING: Kubernetes configuration file is world-readable. This is insecure. Location: /Users/jackfrancis/.kube/external-cloud-provider-24053.kubeconfig
      --kube-context string         name of the kubeconfig context to use
      --kubeconfig string           path to the kubeconfig file
```

I'll figure that out.

test/e2e/cloud-provider-azure.go
jackfrancis: thx!

test/e2e/cloud-provider-azure.go
jackfrancis: reusing the `WaitForControlPlaneIntervals` var everywhere here

arguably you can tune these down per retryable operations, but it's debatable whether or not it's worth the effort

Makefile
jackfrancis: Approach #1 implemented

test/e2e/cloud-provider-azure.go
jackfrancis: Replaced `KUBECONFIG` env var w/ object property assignment

CecileRobertMichon:
/assign @mboersma 

James is OOF this week

Tiltfile
```diff
@@ -331,9 +332,13 @@ def deploy_worker_templates(template, substitutions):
         yaml = yaml.replace("${" + substitution + "}", value)
 
     yaml = yaml.replace('"', '\\"')  # add escape character to double quotes in yaml
+    flavor_name = os.path.basename(flavor)
+    flavor_cmd = "RANDOM=$(bash -c 'echo $RANDOM'); CLUSTER_NAME=" + flavor.replace("windows", "win") + "-$RANDOM; make generate-flavors; echo \"" + yaml + "\" > ./.tiltbuild/" + flavor + "; cat ./.tiltbuild/" + flavor + " | " + envsubst_cmd + " | " + kubectl_cmd + " apply -f - && echo \"Cluster \'$CLUSTER_NAME\' created, don't forget to delete\""
```
mboersma, line 336: I wanted to suggest this would be better as an [f-string](https://docs.python.org/3/tutorial/inputoutput.html#formatted-string-literals), but then I remembered this is [Starlark](https://github.com/bazelbuild/starlark/blob/master/spec.md).

Makefile
mboersma: I'm also mildly opposed to this change if it adds any noticeable time to testing targets. Could we maybe have the new target install just the tools that are actually shared, or is this still a maintenance headache?

```Makefile
install-tools: $(ENVSUBST) $(KUSTOMIZE) $(KUBECTL)

test-e2e-run: generate-e2e-templates install-tools $(GINKGO)
```

docs/book/src/developers/kubernetes-developers.md
```diff
@@ -35,13 +35,56 @@ export CLUSTER_TEMPLATE="test/dev/cluster-template-custom-builds.yaml"
...
 ```bash
-AZURE_CLOUD_CONTROLLER_MANAGER_IMG=myrepo/my-ccm:v0.0.1 \
-AZURE_CLOUD_NODE_MANAGER_IMG=myrepo/my-cnm:v0.0.1 \
-CLUSTER_TEMPLATE=cluster-template-external-cloud-provider.yaml \
+$ export CLUSTER_NAME=my-cluster
+$ CLUSTER_TEMPLATE=cluster-template-external-cloud-provider.yaml \
 make create-workload-cluster
 ```
+
+Once your cluster deploys you should receive the kubeconfig to the workload cluster. Set you `KUBECONFIG` environment variable to point to the kubeconfig file, and then use the official cloud-provider-azure Helm chart to deploy the cloud-provider-azure components using your custom built images:
```
mboersma, line 55:
```suggestion
Once your cluster deploys, you should receive the kubeconfig to the workload cluster. Set your `KUBECONFIG` environment variable to point to the kubeconfig file, then use the official cloud-provider-azure Helm chart to deploy the cloud-provider-azure components using your custom built images:
```

mboersma (commented):
Just a few questions, but this looks good.

Makefile
jackfrancis: only `test-e2e-run` and `tilt-up` have been updated to include this new target, so I think that's a good trade-off (those are flows that are already sort of heavyweight and shouldn't slow down CI overall)

mboersma (approved):
/lgtm

PR #2213, opened 2022-04-04
Skip Existing Secret During Identity Creation

### LINKED ISSUE #2212: Manual Testing: Exit on existing identity
/kind bug

I've been working through the manual testing guide https://capz.sigs.k8s.io/developers/development.html#manual-testing.  While doing so, I was running `make create-cluster` multiple times because I was patching missing local environment vars, etc.  One scenario I hit was a dead end where `hack/create-identity-secret.sh` had already been ran so a secret existed, but because of this, the script would exit.
 
Screen Shot 2022-04-04 at 2 49 47 PM

I tend to expect that re-running should detect that a secret has already been created and move on (function as a no-op would).  I could `make clean kind-reset` and then `make create-cluster` each time but this saves some waiting.

If the expected behavior is the desirable outcome, this should be a simple fix.

evanfreed:
**What type of PR is this?**

/kind cleanup

Fixes #2212

Screen Shot 2022-04-04 at 2 54 33 PM

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

Signed-off-by: Evan Freed <evan.freed@protonmail.ch>

### DIFF
--- a/hack/create-identity-secret.sh
+++ b/hack/create-identity-secret.sh
@@ -26,5 +26,8 @@ cd "${REPO_ROOT}" && make "${KUBECTL##*/}"
 # shellcheck source=hack/parse-prow-creds.sh
 source "${REPO_ROOT}/hack/parse-prow-creds.sh"
 
-"${KUBECTL}" create secret generic "${AZURE_CLUSTER_IDENTITY_SECRET_NAME}" --from-literal=clientSecret="${AZURE_CLIENT_SECRET}"
+if ! "${KUBECTL}" get secret "${AZURE_CLUSTER_IDENTITY_SECRET_NAME}" > /dev/null; then
+  "${KUBECTL}" create secret generic "${AZURE_CLUSTER_IDENTITY_SECRET_NAME}" --from-literal=clientSecret="${AZURE_CLIENT_SECRET}"
+fi
+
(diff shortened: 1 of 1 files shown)


hack/create-identity-secret.sh
```diff
@@ -26,5 +26,10 @@ cd "${REPO_ROOT}" && make "${KUBECTL##*/}"
 # shellcheck source=hack/parse-prow-creds.sh
 source "${REPO_ROOT}/hack/parse-prow-creds.sh"
 
+if "${KUBECTL}" get secret "${AZURE_CLUSTER_IDENTITY_SECRET_NAME}" > /dev/null; then
```
jackfrancis, line 29: rather than exiting 0 early here (which might not be what we want if this script evolves in the future and more post-secret foo is added to the end of the flow) let's add a simply "if not exist" condition, like:

```sh
if ! "${KUBECTL}" get secret "${AZURE_CLUSTER_IDENTITY_SECRET_NAME}" > /dev/null; then
  "${KUBECTL}" create secret generic "${AZURE_CLUSTER_IDENTITY_SECRET_NAME}" --from-literal=clientSecret="${AZURE_CLIENT_SECRET}"
fi
```

And then we can preserve the existing `kubectl label secret` command at the end, which is idempotent as currently implemented (`--overwrite=true`).

Thanks for making this improvement!

evanfreed: Makes sense!

jackfrancis (approved):
/lgtm

CecileRobertMichon (commented):
/approve

PR #2214, opened 2022-04-04
Use MSI ClientID as userAssignedIdentityID in azure.json

### LINKED ISSUE #2164: `userAssignedIdentityID` in `azure.json` is wrong when using user assinged MSI
/kind bug

1. Create a control plane `AzureMachineTemplate` that specifies `userAssignedIdentities`. For example,

```yaml
kind: AzureMachineTemplate
metadata:
  name: foo-control-plane
  namespace: foo
spec:
  template:
    spec:
      dataDisks:
      - diskSizeGB: 256
        lun: 0
        nameSuffix: etcddisk
      identity: UserAssigned
      osDisk:
        diskSizeGB: 128
        osType: Linux
      sshPublicKey: 
      userAssignedIdentities:
      - providerID: /subscriptions/...
      - providerID: /subscriptions/...
      vmSize: Standard_F4s
```
2. Check the generated `azure.json` on control plane nodes. The `userAssignedIdentityID` looks like:

```json
{
    "userAssignedIdentityId": "/subscriptions/..."
}
```

There are two issues here:
* It should be `userAssignedIdentityID` rather than `userAssignedIdentityId`
* The value should be a **client id** (GUID) of the MSI. Not an **ARM resource ID** per cloud provider doc: https://kubernetes-sigs.github.io/cloud-provider-azure/install/configs/.

* Bug 1 location:
https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/c3d1ef191f8203adba1a58e920545f297f80037a/controllers/helpers.go#L373
* Bug 2 location:
https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/c3d1ef191f8203adba1a58e920545f297f80037a/controllers/azurejson_machine_controller.go#L184-L188

**Environment:**

- cluster-api-provider-azure version: v1.1.2
- Kubernetes version: (use `kubectl version`): 1.22
- OS (e.g. from `/etc/os-release`): ubuntu 20.04

mboersma:
**What type of PR is this?**

/kind bug

Uses a lookup function to translate providerID to ClientID when creating `azure.json` on nodes.

Fixes #2164

This is a hack and I'm open for suggestions as to how to make it conform better with our service and reconciler models. Also, it needs regression tests to ensure `azure.json` doesn't revert. (I've tested this interactively.)

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Use MSI ClientID as userAssignedIdentityID in azure.json
```

### DIFF
--- a/azure/services/identities/client.go
+++ b/azure/services/identities/client.go
@@ -1,0 +1,69 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package identities
+
+import (
+	"context"
+
+	"github.com/Azure/azure-sdk-for-go/services/msi/mgmt/2018-11-30/msi"
+	"github.com/Azure/go-autorest/autorest"
+	azuresdk "github.com/Azure/go-autorest/autorest/azure"
+	"sigs.k8s.io/cluster-api-provider-azure/azure"
+	"sigs.k8s.io/cluster-api-provider-azure/util/tele"
+)
+
+// AzureClient contains the Azure go-sdk Client.
+type AzureClient struct {
+	userAssignedIdentities msi.UserAssignedIdentitiesClient
+}
+
+// NewClient creates a new MSI client from auth info.
+func NewClient(auth azure.Authorizer) *AzureClient {
+	c := newUserAssignedIdentitiesClient(auth.SubscriptionID(), auth.BaseURI(), auth.Authorizer())
+	return &AzureClient{c}
+}
+
+// newUserAssignedIdentitiesClient creates a new MSI client from subscription ID, base URI, and authorizer.
+func newUserAssignedIdentitiesClient(subscriptionID string, baseURI string, authorizer autorest.Authorizer) msi.UserAssignedIdentitiesClient {
+	userAssignedIdentitiesClient := msi.NewUserAssignedIdentitiesClientWithBaseURI(baseURI, subscriptionID)
+	azure.SetAutoRestClientDefaults(&userAssignedIdentitiesClient.Client, authorizer)
+	return userAssignedIdentitiesClient
+}
+
+// Get returns a managed service identity.
+func (ac *AzureClient) Get(ctx context.Context, resourceGroupName, name string) (msi.Identity, error) {
+	ctx, _, done := tele.StartSpanWithLogger(ctx, "identities.AzureClient.Get")
+	defer done()
+
+	return ac.userAssignedIdentities.Get(ctx, resourceGroupName, name)
+}
+
+// GetClientID returns the client ID of a managed service identity, given its full URL identifier.
+func (ac *AzureClient) GetClientID(ctx context.Context, providerID string) (string, error) {
+	ctx, _, done := tele.StartSpanWithLogger(ctx, "identities.GetClientID")
+	defer done()
+
+	parsed, err := azuresdk.ParseResourceID(providerID)
+	if err != nil {
+		return "", err
+	}
+	ident, err := ac.Get(ctx, parsed.ResourceGroup, parsed.ResourceName)
+	if err != nil {
+		return "", err
+	}
+	return ident.ClientID.String(), nil
+}
--- a/controllers/azurejson_machinepool_controller.go
+++ b/controllers/azurejson_machinepool_controller.go
@@ -29,6 +29,7 @@ import (
 	"k8s.io/client-go/tools/record"
 	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
 	"sigs.k8s.io/cluster-api-provider-azure/azure/scope"
+	"sigs.k8s.io/cluster-api-provider-azure/azure/services/identities"
 	expv1 "sigs.k8s.io/cluster-api-provider-azure/exp/api/v1beta1"
 	"sigs.k8s.io/cluster-api-provider-azure/util/reconciler"
 	"sigs.k8s.io/cluster-api-provider-azure/util/tele"
@@ -131,12 +132,6 @@ func (r *AzureJSONMachinePoolReconciler) Reconcile(ctx context.Context, req ctrl
 		return reconcile.Result{}, err
 	}
 
-	// Construct secret for this machine
-	userAssignedIdentityIfExists := ""
-	if len(azureMachinePool.Spec.UserAssignedIdentities) > 0 {
-		userAssignedIdentityIfExists = azureMachinePool.Spec.UserAssignedIdentities[0].ProviderID
-	}
-
 	// Create the scope.
 	clusterScope, err := scope.NewClusterScope(ctx, scope.ClusterScopeParams{
 		Client:       r.Client,
@@ -145,6 +140,18 @@ func (r *AzureJSONMachinePoolReconciler) Reconcile(ctx context.Context, req ctrl
 	})
 	if err != nil {
 		return reconcile.Result{}, errors.Wrap(err, "failed to create scope")
+	}
+
+	// Construct secret for this machine
+	userAssignedIdentityIfExists := ""
+	if len(azureMachinePool.Spec.UserAssignedIdentities) > 0 {
+		// TODO: remove this ClientID lookup code when the fixed cloud-provider-azure is default
+		idsClient := identities.NewClient(clusterScope)
+		userAssignedIdentityIfExists, err = idsClient.GetClientID(
+			ctx, azureMachinePool.Spec.UserAssignedIdentities[0].ProviderID)
+		if err != nil {
+			return reconcile.Result{}, errors.Wrap(err, "failed to get user-assigned identity ClientID")
+		}
 	}
 
 	apiVersion, kind := infrav1.GroupVersion.WithKind("AzureMachinePool").ToAPIVersionAndKind()
--- a/go.mod
+++ b/go.mod
@@ -97,6 +97,7 @@ require (
 	github.com/go-openapi/swag v0.19.14 // indirect
 	github.com/gobuffalo/flect v0.2.4 // indirect
 	github.com/gobwas/glob v0.2.3 // indirect
+	github.com/gofrs/uuid v4.2.0+incompatible // indirect
 	github.com/gogo/protobuf v1.3.2 // indirect
 	github.com/golang-jwt/jwt/v4 v4.0.0 // indirect
 	github.com/golang/groupcache v0.0.0-20210331224755-41bb18bfe9da // indirect
--- a/test/e2e/azure_test.go
+++ b/test/e2e/azure_test.go
@@ -428,10 +428,12 @@ var _ = Describe("Workload cluster creation", func() {
 		})
 	})
 
-	// ci-e2e.sh and Prow CI skip this test by default.
-	// To include this test, set `GINKGO_SKIP=""`.
+	// ci-e2e.sh and Prow CI skip this test by default. To include this test, set `GINKGO_SKIP=""`.
+	// This spec expects a user-assigned identity named "cloud-provider-user-identity" in a "capz-ci"
+	// resource group. Override these defaults by setting the USER_IDENTITY and CI_RG environment variables.
 	Context("Creating a cluster that uses the external cloud provider [OPTIONAL]", func() {
 		It("with a 1 control plane nodes and 2 worker nodes", func() {
+			By("using user-assigned identity")
 			clusterName = getClusterName(clusterNamePrefix, "oot")
 			clusterctl.ApplyClusterTemplateAndWait(ctx, clusterctl.ApplyClusterTemplateAndWaitInput{
 				ClusterProxy: bootstrapClusterProxy,
@@ -499,6 +501,7 @@ var _ = Describe("Workload cluster creation", func() {
 		})
 	})
 
+	// ci-e2e.sh and Prow CI skip this test by default. To include this test, set `GINKGO_SKIP=""`.
 	Context("Creating a Windows Enabled cluster with dockershim [OPTIONAL]", func() {
 		// Requires 3 control planes due to https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/857
 		It("With 3 control-plane nodes and 1 Linux worker node and 1 Windows worker node", func() {
@@ -547,8 +550,12 @@ var _ = Describe("Workload cluster creation", func() {
 		})
 	})
 
+	// ci-e2e.sh and Prow CI skip this test by default. To include this test, set `GINKGO_SKIP=""`.
+	// This spec expects a user-assigned identity named "cloud-provider-user-identity" in a "capz-ci"
+	// resource group. Override these defaults by setting the USER_IDENTITY and CI_RG environment variables.
 	Context("Creating a dual-stack cluster [OPTIONAL]", func() {
 		It("With dual-stack worker node", func() {
+			By("using user-assigned identity")
 			clusterName = getClusterName(clusterNamePrefix, "dual-stack")
 			clusterctl.ApplyClusterTemplateAndWait(ctx, clusterctl.ApplyClusterTemplateAndWaitInput{
 				ClusterProxy: bootstrapClusterProxy,
--- a/controllers/azurejson_machine_controller.go
+++ b/controllers/azurejson_machine_controller.go
@@ -30,6 +30,7 @@ import (
 	"k8s.io/client-go/tools/record"
 	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
 	"sigs.k8s.io/cluster-api-provider-azure/azure/scope"
+	"sigs.k8s.io/cluster-api-provider-azure/azure/services/identities"
 	"sigs.k8s.io/cluster-api-provider-azure/util/reconciler"
 	"sigs.k8s.io/cluster-api-provider-azure/util/tele"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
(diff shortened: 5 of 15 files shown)


azure/services/identities/client.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+// GetClientID returns the client ID of a managed service identity, given its full URL identifier.
+func GetClientID(ctx context.Context, auth azure.Authorizer, providerID string) string {
+	ctx, _, done := tele.StartSpanWithLogger(ctx, "identities.GetClientID")
+	defer done()
+
+	// parse providerID into resource group and MSI name
+	rgAndName := regexp.MustCompile(`^/subscriptions/[\w-]+/resourceGroups/(?P<rg>.+)/providers/Microsoft.ManagedIdentity/userAssignedIdentities/(?P<name>.+)$`)
+	matches := rgAndName.FindStringSubmatch(providerID)
```
CecileRobertMichon, line 69: you can use `autorest.ParseResourceID` here instead of compiling your own regex (see natgateways spec.go for an example)

azure/services/identities/client.go
```diff
@@ -0,0 +1,82 @@
...
+// Get the managed service identity.
+func (ac *azureClient) Get(ctx context.Context, resourceGroupName, name string) (msi.Identity, error) {
+	ctx, _, done := tele.StartSpanWithLogger(ctx, "identities.azureClient.Get")
+	defer done()
+
+	return ac.userAssignedIdentities.Get(ctx, resourceGroupName, name)
+}
+
+// GetClientID returns the client ID of a managed service identity, given its full URL identifier.
+func GetClientID(ctx context.Context, auth azure.Authorizer, providerID string) string {
```
CecileRobertMichon, line 63: Consider making this func use the `ac *azureClient` receiver and instead of passing in the authorizer and constructing a new client in the func. We can then create a new client from the controller where we need it.

azure/services/identities/client.go
```diff
@@ -0,0 +1,82 @@
...
+	ctx, _, done := tele.StartSpanWithLogger(ctx, "identities.GetClientID")
+	defer done()
+
+	// parse providerID into resource group and MSI name
+	rgAndName := regexp.MustCompile(`^/subscriptions/[\w-]+/resourceGroups/(?P<rg>.+)/providers/Microsoft.ManagedIdentity/userAssignedIdentities/(?P<name>.+)$`)
+	matches := rgAndName.FindStringSubmatch(providerID)
+	if len(matches) < 3 {
+		// TODO: log error and return an empty string
+		return "failed to find resource group and name in provider ID URL"
```
CecileRobertMichon, line 72: This should return an error, we can handle the error from the controller

azure/services/identities/client.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+	rg, name := matches[1], matches[2]
+	// Look up the Client ID of the user-assigned MSI which is assigned to the underlying VMs.
+	ident, err := newClient(auth).Get(ctx, rg, name)
+	if err != nil {
+		// TODO: log error and return an empty string
```
CecileRobertMichon, line 78: This should return an error, we can handle the error from the controller

azure/services/identities/client.go
```diff
@@ -0,0 +1,82 @@
...
+	"regexp"
+
+	"github.com/Azure/azure-sdk-for-go/services/msi/mgmt/2018-11-30/msi"
+	"github.com/Azure/go-autorest/autorest"
+	"sigs.k8s.io/cluster-api-provider-azure/azure"
+	"sigs.k8s.io/cluster-api-provider-azure/util/tele"
+)
+
+// client wraps go-sdk.
+type client interface {
```
CecileRobertMichon, line 30: is this client interface used anywhere? If not, let's remove it

CecileRobertMichon (commented):
/assign @jackfrancis

mboersma:
I would really like to validate `azure.json` in an e2e test, but AFAICT we don't use the user-assigned id flavor in tests yet. I'll try writing a new spec for that. Suggestions welcome!

azure/services/identities/client.go
```diff
@@ -0,0 +1,69 @@
...
+func (ac *AzureClient) Get(ctx context.Context, resourceGroupName, name string) (msi.Identity, error) {
+	ctx, _, done := tele.StartSpanWithLogger(ctx, "identities.AzureClient.Get")
+	defer done()
+
+	return ac.userAssignedIdentities.Get(ctx, resourceGroupName, name)
+}
+
+// GetClientID returns the client ID of a managed service identity, given its full URL identifier.
+func (ac *AzureClient) GetClientID(ctx context.Context, providerID string) (string, error) {
+	ctx, _, done := tele.StartSpanWithLogger(ctx, "identities.GetClientID")
```
shysank, line 57:
```suggestion
	ctx, _, done := tele.StartSpanWithLogger(ctx, "identities.AzureClient.GetClientID")
```

go.mod
```diff
@@ -86,6 +86,7 @@ require (
 	github.com/go-openapi/jsonreference v0.19.5 // indirect
 	github.com/go-openapi/swag v0.19.14 // indirect
 	github.com/gobuffalo/flect v0.2.4 // indirect
+	github.com/gofrs/uuid v4.2.0+incompatible // indirect
```
shysank, line 89: Is this change intentional?

azure/services/identities/client.go
```diff
@@ -0,0 +1,69 @@
...
+
+// Get returns a managed service identity.
+func (ac *AzureClient) Get(ctx context.Context, resourceGroupName, name string) (msi.Identity, error) {
+	ctx, _, done := tele.StartSpanWithLogger(ctx, "identities.AzureClient.Get")
+	defer done()
+
+	return ac.userAssignedIdentities.Get(ctx, resourceGroupName, name)
+}
+
+// GetClientID returns the client ID of a managed service identity, given its full URL identifier.
```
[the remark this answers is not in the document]
shysank: how about adding an optional field for clientID in spec, and update it if it's not present? wdyt?

shysank (commented):
looks good overall with a couple of nits.

go.mod
mboersma: It was pulled in by `make modules` after I added

```go
import "github.com/Azure/azure-sdk-for-go/services/msi/mgmt/2018-11-30/msi"
```

azure/services/identities/client.go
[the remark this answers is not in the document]
mboersma: That's a clever solution. I was avoiding changing the spec because of the hassle. I don't think ClientID is considered more sensitive data than the full URL (ProviderID) that we already include, so it's probably fine to have it on the spec. I'll try that out.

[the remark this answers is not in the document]
mboersma: @shysank wouldn't we want to add an (optional) `ClientID` field to each `UserAssignedIdentity`? Although in practice the code expects there only to be one...

Also, that would prevent the case I described above from working, where an admin could delete the MSI and recreate it but expect the ClientID to be refreshed. That may be a contrived scenario or not something we should worry about supporting. The tradeoff is making more Azure API calls.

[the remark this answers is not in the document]
CecileRobertMichon: > where an admin could delete the MSI and recreate it but expect the ClientID to be refreshed

that definitely is a valid scenario

Another solution would be to move things around so that we get the Client ID Just-In-Time only when the cloud provider config secret doesn't exist and needs to be created. That way we don't do a GET every reconcile, only the first time when the secret is created or whenever it needs to be recreate if it was deleted (to refresh creds for example). It also optimizes things in that it doesn't require us to fetch the Client ID at all when the user is providing their own azure.json Secret and we don't need to create one.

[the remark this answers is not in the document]
CecileRobertMichon: @mboersma I see that you added a `ClientID` field in spec, did you look into the alternative of fetching the cliend ID only if the cloud provider secret doesn't exist? I'm concerned that having the ClientID in the spec has the potential to cause redundant/confusing UX and potentially inconsistency if the user sets both but they don't point to the same identity. It could also cause issues in the case you mentioned above where the ClientID needs to be refreshed.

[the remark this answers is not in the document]
mboersma: > did you look into the alternative of fetching the client ID only if the cloud provider secret doesn't exist?

I did. I'm not sure it's really feasible–it boils down to the deep-equals comparison on the secret data. I moved a ton of code around trying this, but didn't see a good way to get there.

It's true that this allows the user to specify the ClientID now (and perhaps to get it wrong). But to refresh the ClientID, you can just set it to empty and we will fetch it based on the ProviderID. These are both properties on a UserAssignedIdentity, so it seemed logical to add it to that struct (even though in practice, we only ever use the first ClientID).

mboersma:
See also kubernetes-sigs/cloud-provider-azure#1290.

mboersma:
This PR is working and has an e2e test that validates the userAssignedIdentityID value looks like a GUID.

While the proper fix for this issue is in cloud-provider-azure, it's better for this fix also to go in so CAPZ isn't dependent on their release schedule. After some amount of time when we're sure the fixed cloud-provider-azure is used everywhere, we should revert most of this change to simplify the code. We can leave the e2e test intact and just validate the field with `azuresdk.ParseResource` instead of a regex.

test/e2e/azure_test.go
```diff
@@ -468,6 +468,54 @@ var _ = Describe("Workload cluster creation", func() {
...
+				WaitForClusterIntervals:      e2eConfig.GetIntervals(specName, "wait-cluster"),
+				WaitForControlPlaneIntervals: e2eConfig.GetIntervals(specName, "wait-control-plane"),
+				WaitForMachinePools:          e2eConfig.GetIntervals(specName, "wait-machine-pool-nodes"),
+				ControlPlaneWaiters: clusterctl.ControlPlaneWaiters{
+					WaitForControlPlaneInitialized: InstallCloudProviderAzureHelmChart,
+				},
+			}, result)
+
+			Context("Validating the azure.json secret", func() {
+				AzureJSONSecretSpec(ctx, func() AzureJSONSecretSpecInput {
```
CecileRobertMichon, line 508: instead of validating the contents of the secret, have you considered reusing the LB test spec to expose a service with cloud provider and make sure it is able to use those credentials to create an Azure LB as expected?

controllers/azurejson_machinepool_controller.go
```diff
@@ -147,6 +142,17 @@ func (r *AzureJSONMachinePoolReconciler) Reconcile(ctx context.Context, req ctrl
 		return reconcile.Result{}, errors.Wrap(err, "failed to create scope")
 	}
 
+	// Construct secret for this machine
```
CecileRobertMichon, line 145: Can we add a TODO comment here about removing this once cloud-provider-azure supports user assigned IDs as resource IDs?

CecileRobertMichon: same thing in the other controllers

test/e2e/azure_test.go
mboersma: I had not, but that would be a more durable test that wouldn't have to be changed when we revert the rest. Great idea, I'll give it a try.

test/e2e/azure_json.go
```diff
@@ -0,0 +1,113 @@
...
+
+// AzureJSONSecretSpecInput is the input for AzureJSONSecretSpec.
+type AzureJSONSecretSpecInput struct {
+	BootstrapClusterProxy framework.ClusterProxy
+	Namespace             *corev1.Namespace
+	ClusterName           string
+}
+
+// AzureJSONSecretSpec implements a test that verifies the contents of the /etc/kubernetes/azure.json file.
+func AzureJSONSecretSpec(ctx context.Context, inputGetter func() AzureJSONSecretSpecInput) {
```
CecileRobertMichon, line 49: I was thinking we should get rid of this test completely in favor of the LB test. 

If you still want to do the extra validation step of checking the content of the AzureJson file, consider making this function more specific to what we are validating which is identity. With the current implementation, it assumes user-assigned identity even though the func name and comment make it seem like it's a generic test for azure.json content. It also checks for hardcoded values like `LoadBalancerSku` being "Regular" which might not apply to every cluster if we ever want to reuse this test spec. It would also be good to make sure that ClientID / ClientSecret are empty when using managed identity, and vice versa when using a service principal.

test/e2e/azure_json.go
```diff
@@ -0,0 +1,113 @@
...
+		Name:      clusterName,
+		Namespace: name,
+	})
+	Expect(azureCluster).NotTo(BeNil())
+
+	// Validate required fields.
+	Expect(config.Cloud).To(Equal("AzurePublicCloud"), "Invalid cloud name")
+	Expect(config.TenantID).To(MatchRegexp(guidRegex), "Invalid tenant ID")
+	Expect(config.SubscriptionID).To(MatchRegexp(guidRegex), "Invalid subscription ID")
+	Expect(config.ResourceGroup).To(Equal(azureCluster.Spec.ResourceGroup), "Invalid resource group")
```
CecileRobertMichon, line 103: you should already have access to the resource group name and location in the e2e config without having to get them from the AzureCluster object

test/e2e/azure_json.go
mboersma: > I was thinking we should get rid of this test completely in favor of the LB test.

That's fine with me. I just need to validate that the LB Spec + user-assigned ID test covers this and works.

I had meant this test to do a more general validation of `azure.json` based on the cluster configuration, but that's not specific to this case. Also I'm worried that involving SSH could introduce more flakiness into the tests overall, and we don't need that.

CecileRobertMichon: Looking through test again, I'm more convinced we should get rid of this test. If we want to test the content of the file, it should be covered via unit test (https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/controllers/helpers_test.go#L82). The goal of the e2e test is to make sure the cluster is functional. In this case a cluster using user-assigned identities should provide valid credentials for cloud provider to make Azure API calls. That is already validated by the LB spec.

test/e2e/azure_test.go
```diff
@@ -468,6 +468,65 @@ var _ = Describe("Workload cluster creation", func() {
 		})
 	})
 
+	// ci-e2e.sh and Prow CI skip this test by default.
+	// To include this test, set `GINKGO_SKIP=""`.
+	Context("Creating a cluster with user-assigned identity [OPTIONAL]", func() {
+		It("with a 1 control plane nodes and 0 worker nodes", func() {
```
CecileRobertMichon, line 474: @mboersma the LB spec won't work with 0 worker nodes, we need worker nodes to create a valid deployment with running pods

templates/test/ci/prow-user-assigned-identity/kustomization.yaml
```diff
@@ -0,0 +1,23 @@
...
+namespace: default
+resources:
+  - ../../../flavors/external-cloud-provider
+  - ../prow/cni-resource-set.yaml
+patchesStrategicMerge:
+  - ../patches/tags.yaml
+  - ../patches/controller-manager.yaml
+  - ../patches/cluster-cni.yaml
+  - ../patches/apiserver.yaml
+  - patches/azure-machine-template.yaml
```
CecileRobertMichon, line 12: we have a user-assigned identity flavor in the templates/ dir already which includes patches, wonder if we can reuse those since they seem to be doing the same thing?

templates/test/ci/prow-user-assigned-identity/kustomization.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: default
+resources:
+  - ../../../flavors/external-cloud-provider
```
CecileRobertMichon, line 5: you could base this on the prow external cloud provider flavor so you don't have to re-add all the patches below

templates/test/ci/prow-user-assigned-identity/kustomization.yaml
mboersma: It seemed to me that we were keeping them separate from the test/ci templates. The test framework expects a local clusterctl repository that is currently populated just from there.

templates/test/ci/prow-user-assigned-identity/kustomization.yaml
mboersma: Although that could confuse things when we make changes to the test/ci prow external cloud provider template. You'd have to run `make generate-flavors` twice to be sure.

mboersma:
I ran this locally and it succeeded:
```sh
GINKGO_SKIP="" GINKGO_FOCUS="Creating a cluster with user-assigned identity" SKIP_CLEANUP=true ./scripts/ci-e2e.sh
```
Then I reverted the clientID lookup changes and ran it again. It failed, waiting for the first internal LB to be available. Looking at the logs for cloud-controller-manager afterward, the error is `Status Code = '400'. Response body: {"error":"invalid_request","error_description":"Identity not found"}`, as expected.

I'm just reassuring myself that this fix allows user-assigned ID to work and that the LB test validates that.

templates/test/ci/prow-user-assigned-identity/kustomization.yaml
CecileRobertMichon: why would you have to run `make generate-flavors` twice? if you make changes to external-cloud-provider-prow base or patch yamls it would automatically get picked up when you run `make generate-flavor` by both templates

CecileRobertMichon: Another thing to consider: instead of adding a separate template for user-assigned identity, thoughts on making the existing external-cloud-provider prow template/test use user-assined id ?

mboersma: > why would you have to run `make generate-flavors` twice?

I guess I was imagining basing it off the generated external-cloud-provider template and assuming a one-pass generator, but I was probably wrong. I'll take another look; it would be preferable not to grow the list of templates and tests.

test/e2e/azure_test.go
```diff
@@ -499,9 +501,13 @@ var _ = Describe("Workload cluster creation", func() {
 		})
 	})
 
+	// ci-e2e.sh and Prow CI skip this test by default. To include this test, set `GINKGO_SKIP=""`.
+	// This spec expects a user-assigned identity named "cloud-provider-user-identity" in a "capz-ci"
+	// resource group. Override these defaults by setting the IDENTITY_NAME and CI_RG environment variables.
```
CecileRobertMichon, line 506: we're actually getting rid of this test in #2292, is there another in-tree optional test we can use to add user assigned id coverage?

mboersma: Dual-stack, or GPU. I vote for dual-stack.

mboersma:
Updated to validate user-assigned identity in the existing external-cloud-provider and dual-stack (internal cloud-provider) test specs.

This requires a user-assigned identity to exist before running tests, as described in the comments in azure_test.go. This is already true for the CI subscription.

CecileRobertMichon (commented):
/lgtm
/approve

CecileRobertMichon:
/cherry-pick release-1.3

@mboersma thoughts on cutting a release to include this one?

PR #2222, opened 2022-04-08
wait longer for private cluster to account for add’l bastion create time

jackfrancis:
**What type of PR is this?**

/kind failing-test

In observance of recent private cluster failures, let's increase the timeout that we enforce to wait for cluster provisioning. Using a common timeout value across private clusters and public clusters probably isn't the best idea, as the Azure bastion service can regularly take 5+ mins to come online (which is additive blocking time compared to public cluster scenario).

This PR increases the timeout to 40 mins to account for observed 20+ minute Azure bastion create times.

Examples of recent PR flakes:

- https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/2209/pull-cluster-api-provider-azure-e2e/1512495713999654912
- https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/2209/pull-cluster-api-provider-azure-e2e/1511121447714557952
- https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/2215/pull-cluster-api-provider-azure-e2e/1511128382417408000

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note

```

### DIFF
--- a/test/e2e/azure_privatecluster.go
+++ b/test/e2e/azure_privatecluster.go
@@ -149,7 +149,7 @@ func AzurePrivateClusterSpec(ctx context.Context, inputGetter func() AzurePrivat
 			ControlPlaneMachineCount: pointer.Int64Ptr(3),
 			WorkerMachineCount:       pointer.Int64Ptr(1),
 		},
-		WaitForClusterIntervals:      input.E2EConfig.GetIntervals(specName, "wait-cluster"),
+		WaitForClusterIntervals:      input.E2EConfig.GetIntervals(specName, "wait-private-cluster"),
 		WaitForControlPlaneIntervals: input.E2EConfig.GetIntervals(specName, "wait-control-plane"),
 		WaitForMachineDeployments:    input.E2EConfig.GetIntervals(specName, "wait-worker-nodes"),
 	}, result)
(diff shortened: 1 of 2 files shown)


CecileRobertMichon (commented):
would 30 minutes be enough to cover 99% of creates? 40 minutes to create a single cluster seems way too long.

jackfrancis:
> would 30 minutes be enough to cover 99% of creates? 40 minutes to create a single cluster seems way too long.

Easy enough to try! We can increase by 10 mins and see if the flake rate drops to non-annoying frequency.

jackfrancis:
@CecileRobertMichon timeout set to 30m

CecileRobertMichon:
/lgtm
/approve

@jackfrancis @shysank AzureBastion might be a good one to look into making parallel, or somehow non-blocking for creating AzureMachines since it's not strictly required for the Bastion to be created before creating nodes and other unrelated cluster infra.

PR #2223, opened 2022-04-09
fix for tags update and delete op

LochanRn:
**What type of PR is this?**
/kind bug

Allows aks tags to be updated and deleted 

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
allows aks tags to be updated and deleted.
```

### DIFF
--- a/azure/services/managedclusters/managedclusters.go
+++ b/azure/services/managedclusters/managedclusters.go
@@ -148,9 +148,11 @@ func computeDiffOfNormalizedClusters(managedCluster containerservice.ManagedClus
 
 	clusterNormalized := &containerservice.ManagedCluster{
 		ManagedClusterProperties: propertiesNormalized,
+		Tags:                     managedCluster.Tags,
 	}
 	existingMCClusterNormalized := &containerservice.ManagedCluster{
 		ManagedClusterProperties: existingMCPropertiesNormalized,
+		Tags:                     existingMC.Tags,
 	}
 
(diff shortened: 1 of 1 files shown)


jackfrancis:
@LochanRn thanks for looking into this! You'll have to update test cases here to deal w/ new `Tags` property: `azure/services/managedclusters/managedclusters_test.go`

LochanRn:
hey @jackfrancis thanks, will update the PR accordingly :)

LochanRn:
@jackfrancis can you please review this PR again, sorry for the delay.

azure/services/managedclusters/managedclusters.go
```diff
@@ -240,6 +241,10 @@ func (s *Service) Reconcile(ctx context.Context) error {
 		},
 	}
 
+	if tags := *to.StringMapPtr(managedClusterSpec.Tags); len(tags) != 0 {
```
jackfrancis, line 244: Can you explain why this change (only add tags property value if it is present in ManagedCluster spec) allows tags to be updated?

LochanRn: No it's because when tags are empty the existing object returned from the azure api has a nil value for tags but the current object has map[string]string with length 0, the diff function we use is returning true for this case, which triggers unnecessary updates.

jackfrancis (approved):
/lgtm

PR #2238, opened 2022-04-19
Delete long running operation state when resource is not found

CecileRobertMichon:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**: Fixing an edge case scenario where a resource is deleted after being created, before the CAPZ controller has a chance to fetch the result. In that case, `Result()` would return `404` and the controller would be stuck in a bad loop. If the resource was somehow deleted by an external process (eg. human or automated cleanup job) we should recreate it. Deleting the long-running state when the error is "not found" will make sure we don't get infinitely stuck trying to get a resource that doesn't exist and instead trying creating the resource again on the next loop.

Fixes #

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Delete long running operation state when resource is not found
```

### DIFF
--- a/azure/services/async/async.go
+++ b/azure/services/async/async.go
@@ -79,7 +79,10 @@ func processOngoingOperation(ctx context.Context, scope FutureScope, client Futu
 	// Resource has been created/deleted/updated.
 	log.V(2).Info("long running operation has completed", "service", serviceName, "resource", resourceName)
 	result, err = client.Result(ctx, sdkFuture, future.Type)
-	if err == nil {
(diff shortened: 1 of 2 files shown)


CecileRobertMichon (commented):
/assign @Jont828 @jackfrancis

CecileRobertMichon:
tests passed

jackfrancis (approved):
/lgtm

mboersma (approved):
/lgtm
/approve

PR #2240, opened 2022-04-19
Fix race condition when creating resource group with custom user tags

CecileRobertMichon:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**: Fixes a race condition with custom user-defined tags. Currently, if a user defines custom tags, the group is first created with the default set of tags and then the custom tags are added in the tags service. This can cause issues if the subscription has automation running that checks for certain tags to perform automated cleanup or apply policies. This makes sure the tags are added right away when the group is first created, then updated separately by the tags service as needed.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Ensure additional tags are added right away when the group is created
```

### DIFF
--- a/azure/services/groups/spec.go
+++ b/azure/services/groups/spec.go
@@ -56,13 +56,13 @@ func (s *GroupSpec) Parameters(existing interface{}) (params interface{}, err er
 	}
 	return resources.Group{
 		Location: to.StringPtr(s.Location),
-		// We create only CAPZ default tags. User defined additional tags
-		// are created and updated using tags service.
+		// User defined additional tags are created with the resource group and updated using tags service.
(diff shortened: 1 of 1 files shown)


jackfrancis (approved):
lgtm

jackfrancis:
looks like there are some spec_test.go UT in other services, we should probably spike on standardizing UT for all of these at some point

sonasingh46 (approved):

CecileRobertMichon:
@jackfrancis could I please get a /lgtm on this one?

jackfrancis (approved):
/lgtm

PR #2252, opened 2022-04-21
test: don’t panic during log collection

jackfrancis:
**What type of PR is this?**

/kind flake

This PR ensures that E2E log collection doesn't panic, as there are observations that the list of MachinePool `ProviderIDList` instances and its `NodeRefs` do not exactly correlate at the time that log collection occurs.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note

```


test/e2e/azure_logcollector.go
```diff
@@ -108,14 +109,18 @@ func (k AzureLogCollector) CollectMachinePoolLog(ctx context.Context, management
...
+				errors = append(errors, err)
+			}
 
-		if err := collectVMSSBootLog(ctx, instance, filepath.Join(outputPath, hostname)); err != nil {
-			errors = append(errors, err)
+			if err := collectVMSSBootLog(ctx, instance, filepath.Join(outputPath, hostname)); err != nil {
+				errors = append(errors, err)
+			}
+		} else {
+			Logf(fmt.Sprintf("INFO: MachinePool instance %s does not have a corresponding NodeRef", instance))
```
CecileRobertMichon, line 123: can we add something like "skipping node and boot log collection for MachinePool instance %: " to the log message so the impact is more obvious in the output?

jackfrancis: done

CecileRobertMichon: sorry if it wasn't clear, I meant in addition to the message you already had that way it's clear what we're doing and why

jackfrancis: all good, done

test/e2e/azure_logcollector.go
```diff
@@ -108,14 +109,19 @@ func (k AzureLogCollector) CollectMachinePoolLog(ctx context.Context, management
...
```
CecileRobertMichon, line 123: Logf is already printing "INFO:" so this would render as `INFO: INFO: `. Also it takes in format + args so you don't need to use `fmt.Sprintf`:

```suggestion
			Logf("MachinePool instance %s does not have a corresponding NodeRef", instance)
```

It seems that azure_logcollector.go is double printing INFO in a few places, eg. `Apr 20 23:09:07.082: INFO: INFO: Collecting boot logs for AzureMachine capz-e2e-002817-ipv6-md-0-vcpmh`, might be good to fix up those too while you're here.

https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/1703ef3434cdbe35a31062e1abb04e4291fdb9fc/test/e2e/log.go#L40

jackfrancis: good idea, done

CecileRobertMichon (commented):
/lgtm
/approve

Issue #1104, opened 2021-01-04
Multi-tenancy within one manager instance using Azure User Assigned Identities


nader-ziada:
/kind feature

**Describe the solution you'd like**

As described by the multi-tenancy design proposal https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/docs/proposals/20200720-single-controller-multitenancy.md (updated in #1089), 
the manager instance should also support `Azure User Assigned Identities` 

The current implementation (from pr #977) only added support for using Service Principal  using `aad-pod-identity`

nader-ziada:
@shysank there might be some changes need in the aad-pod-library before we can work on this, better to discuss with @devigned

devigned:
As @nader-ziada was mentioning and if I recall correctly, AAD Pod Identity is configured to work within a resource group when assigning managed identities and is configured through https://github.com/Azure/aad-pod-identity/blob/fb1bdba48ab5c1fa0be64382b6a74acdbcb41089/pkg/config/azureconfig.go#L10. After talking with @aramase, the reasoning behind this was mostly due to performance concerns of having to look up Azure resources outside of the resource group, which can be an expensive operation. I did not get the feeling scoping by resource group was functionally necessary if a resource lookup could be done in an efficient manner. 

Since CAPZ resources can be in any resource group a user desires and AAD Pod Identity is intended to be run as a singleton controller within a cluster, the constraint on resource group scoping of the Managed Identity Controller must be removed to unblock Managed Identities for CAPZ.

Happy to setup some time to discuss in adhoc meeting, in thread, or during our next office hours.

aramase:
The `resourcegroup` that's configured with MIC (MIC gets that value from `azure.json`) is for the compute resources only. MIC uses the resource group to list the VMSS instances and assign identity only to the VMSS in the cluster resource group. However the identities can be in any resource group.

@devigned The management cluster is the one that has the identities, right?

devigned:
> The management cluster is the one that has the identities, right?

Yes. 

So, are you saying that Azure compute resources hosting the management cluster would not span multiple resource groups, so having resourcegroup in the azure.json makes sense?

aramase:
> So, are you saying that Azure compute resources hosting the management cluster would not span multiple resource groups, so having resourcegroup in the azure.json makes sense?

[...]

PR #2258, opened 2022-04-22
Add support for Service Principal with Certificate auth using AAD pod identity

### LINKED ISSUE #2250: login trying too aggressively
/kind bug

Tried to create a new cluster.
```
kind: AzureClusterIdentity
metadata:
  labels:
    clusterctl.cluster.x-k8s.io/move-hierarchy: "true"
  name: xxx
  namespace: yyy
spec:
  allowedNamespaces:
    list:
    - zzz
  clientID: someid
  clientSecret:
    name: sp-cert
    namespace: yyy
  tenantID: someotherid
  type: ServicePrincipal

Then created the secret like so:
kubectl create secret generic -n yyy sp-cert --from-file=certificate=key.pem

If it were to fail to work, it should slow down

Instead capz-nmi is spewing constantly about:
E0420 20:34:00.878482       1 server.go:399] failed to get service principal token for pod: capz-system/capz-controller-manager-c75cd8b88-j7sww, error: adal: Refresh request failed. Status Code = '401'. Response body: {"error":"invalid_client","error_description":"AADSTS7000215: Invalid client secret provided. Ensure the secret being sent in the request is the client secret value, not the client secret ID, for a secret added to app 'someid'.\r\nTrace ID: someid\r\nCorrelation ID: 9c578966-9524-4a0b-9114-3fd1fa918609\r\nTimestamp: 2022-04-20 20:34:00Z","error_codes":[7000215],"timestamp":"2022-04-20 20:34:00Z","trace_id":"someid","correlation_id":"someid","error_uri":"https://login.microsoftonline.com/error?code=7000215"} Endpoint https://login.microsoftonline.com/someid/oauth2/token?api-version=1.0

It should probably back off at some point

**Environment:**

### PR DESCRIPTION
**Pull request by CecileRobertMichon**
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**: This fixes Certificate auth for AAD pod identity. It was already supported according to docs but never worked properly https://capz.sigs.k8s.io/topics/multitenancy.html#service-principal-identity.

Fixes #2250 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Add support for Service Principal with Certificate auth 
```


kfox1111:
lgtm

api/v1beta1/azureclusteridentity_types.go
```diff
@@ -43,7 +43,7 @@ type AllowedNamespaces struct {
 
 // AzureClusterIdentitySpec defines the parameters that are used to create an AzureIdentity.
 type AzureClusterIdentitySpec struct {
-	// UserAssignedMSI or Service Principal
```
jackfrancis, line 46: should we clean up all these comments so that the property name is the first string in the comment, e.g.
```
// Type is the identity type.
// Valid options are ServicePrincipal, ServicePrincipalCertificate, or ManualServicePrincipal.
Type IdentityType `json:"type"`
// ResourceID is the user-assigned MSI resource id.
// +optional
ResourceID string `json:"resourceID,omitempty"`
```

etc

api/v1beta1/types.go
```diff
@@ -422,18 +422,21 @@ const (
 )
 
 // IdentityType represents different types of identities.
-// +kubebuilder:validation:Enum=ServicePrincipal;ManualServicePrincipal;UserAssignedMSI
+// +kubebuilder:validation:Enum=ServicePrincipal;ManualServicePrincipal;ServicePrincipalCertificate
```
jackfrancis, line 425: this looks like a breaking change, is that intended/allowed?

CecileRobertMichon: It technically breaking but in practice, no one should be broken by the change. Right now if you set it to `UserAssignedMSI` it won't work and just throw a [validation error in the controller](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2258/files#diff-b6ab207506064afcbd1d777c1c7296e2953d2653a8843bfb39107af5cf5b0a87L90). It has never been supported and there is an [issue](https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1104) to add support for it (planning on working on that next). I could leave it as-is but IMO this is an improvement as it's better to fail the user right away if they try to use an unsupported value than to let them create the resource and later fail in the controller.

api/v1beta1/azureclusteridentity_types.go
CecileRobertMichon: done

api/v1beta1/types.go
jackfrancis: Makes sense!

jackfrancis (approved):
lgtm

api/v1beta1/azureclusteridentity_types.go
```diff
@@ -43,17 +43,20 @@ type AllowedNamespaces struct {
...
+	// Type is the type of Azure Identity used.
```
mboersma, line 46:
```suggestion
	// IdentityType is the type of Azure Identity used.
```

Although since this field is just `type` in its JSON representation, making this a proper Go comment here will create a naming discrepancy in the generated CRD description. I still think that would be more correct, but I am not invested in this change. 😄

mboersma (approved):
/lgtm

api/v1beta1/azureclusteridentity_types.go
CecileRobertMichon: `Type` is the name of the field too... `IdentityType` is the type of the field (this is confusing!) :D 

So I think `Type` is more correct

mboersma: You're right--so hard to read when things "stutter" like that.

PR #2261, opened 2022-04-22
test: fix wait.PollImmediate so it actually retries after errors

jackfrancis:
**What type of PR is this?**

/kind bug

This PR cleans up some of the mess I've left around by incompletely implementing `wait.PollImmediate`. Thanks everyone for your patience!

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/retry.go
+++ b/test/e2e/retry.go
@@ -33,17 +33,20 @@ const (
 	retryBackoffSteps           = 3
 )
 
-// retryWithTimeout retries the function until it returns true,
-// or a timeout is reached.
+// retryWithTimeout retries a function that returns an error until a timeout is reached
 func retryWithTimeout(interval, timeout time.Duration, fn func() error) error {
-	retryFn := func(fn func() error) func() (bool, error) {
-		return func() (bool, error) {
-			err := fn()
-			if err == nil {
-				return true, nil
-			}
-			return false, err
+	var pollError error
+	err := wait.PollImmediate(collectLogInterval, collectLogTimeout, func() (bool, error) {
+		pollError = nil
+		err := fn()
+		if err != nil {
+			pollError = err
+			return false, nil
 		}
+		return true, nil
+	})
+	if pollError != nil {
+		return pollError
 	}
-	return wait.PollImmediate(interval, timeout, retryFn(fn))
+	return err
 }


test/e2e/azure_logcollector.go
```diff
@@ -139,14 +141,21 @@ func collectLogsFromNode(ctx context.Context, managementClusterClient client.Cli
...
+			if err != nil {
+				return err
+			}
+			defer f.Close()
+			var retryError error
+			_ = wait.PollImmediate(collectLogInterval, collectLogTimeout, func() (bool, error) {
+				var execOnHostErr error
+				if execOnHostErr = execOnHost(controlPlaneEndpoint, hostname, sshPort, f, command, args...); err != nil {
+					retryError = execOnHostErr
+					return false, nil
```
CecileRobertMichon, line 154: is it possible to still reuse the helper function in retry.go but modify it so it does retry after errors? That way we can reuse it later if needed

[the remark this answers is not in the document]
jackfrancis: yeah, so I was stumped on this when I originally did this last Friday, after the weekend I think it's actually easy to do this without removing the general `retryWithTimeout` helper func

test/e2e/retry.go
```diff
@@ -33,17 +33,19 @@ const (
...
+	err := wait.PollImmediate(collectLogInterval, collectLogTimeout, func() (bool, error) {
+		err := fn()
+		if err != nil {
+			pollError = err
+			return false, nil
 		}
+		return true, nil
+	})
+	if pollError != nil {
+		return pollError
```
CecileRobertMichon, line 48: if one of the attempts failed wouldn't pollError be != nil? What sets it back to nil if a successive attempt works?

jackfrancis: yep, need to change this!

CecileRobertMichon (commented):
/assign @Jont828

test/e2e/retry.go
```diff
@@ -33,17 +33,20 @@ const (
...
-				return true, nil
-			}
-			return false, err
+	var pollError error
+	err := wait.PollImmediate(collectLogInterval, collectLogTimeout, func() (bool, error) {
+		pollError = nil
+		err := fn()
+		if err != nil {
+			pollError = err
+			return false, nil
```
Jont828, line 44: It looks like we're always returning `nil` for the error in this nested function. Would we ever want to return an error here?

jackfrancis: No, the purpose of this helper function is to absorb *every* error and only fail if the timeout has been reached.

Jont828:
I think this looks good!

/lgtm

test/e2e/retry.go
```diff
@@ -33,17 +33,20 @@ const (
...
 		}
+		return true, nil
+	})
+	if pollError != nil {
+		return pollError
```
CecileRobertMichon, line 49: I'm still not sure why we need to check the pollError outside the poll function.

I found this example which might be closer to what you're trying to do: https://github.com/kubernetes/kubernetes/blob/4ab90ccebb0e22396b5e10cfd58bb34641bc766f/cmd/kubeadm/app/phases/upgrade/health.go#L163

```
if err != nil {
    return errors.Wrapf(pollError, "some text about failing after retries")
}
```

jackfrancis: The idea here is to bubble up the last error received if there is one. So the order of precedence:

1) return the most recent specific error from the function invocation, if it exists
2) if no error was ever returned from any function invocation, then return the error from the retry func (the error will either be nil or "timed out")

Does that make sense? The thinking here is that the most likely scenario if an error from the function invocation exists (inside PollImmediate) is a recurring, common error, so it aids to debugging to return that specific error so we can triage.

CecileRobertMichon: I think I understand better now, thanks. Essentially, we don't want to return pollErr if `fn()` never returned and the retry just failed because of a timeout on the first attempt.

jackfrancis: Correct, in that case the "timed out" error *is* the fundamental error that we want to log/persist.

PR #2264, opened 2022-04-25
generalize helm install during E2E testing

jackfrancis:
**What type of PR is this?**

/kind cleanup

This PR follows up from https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2209

Here we generalize some of the "install helm chart" foo so that it can be composed in different ways to accommodate different test scenarios.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/cloud-provider-azure.go
+++ b/test/e2e/cloud-provider-azure.go
@@ -23,122 +23,47 @@ import (
 	"context"
 	"fmt"
 
-	"github.com/Azure/go-autorest/autorest/to"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-	helmAction "helm.sh/helm/v3/pkg/action"
-	helmLoader "helm.sh/helm/v3/pkg/chart/loader"
-	helmCli "helm.sh/helm/v3/pkg/cli"
-	helmVals "helm.sh/helm/v3/pkg/cli/values"
-	helmGetter "helm.sh/helm/v3/pkg/getter"
-	corev1 "k8s.io/api/core/v1"
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"sigs.k8s.io/cluster-api/test/framework"
 	"sigs.k8s.io/cluster-api/test/framework/clusterctl"
-	crclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	cloudProviderAzureHelmRepoURL    = "https://raw.githubusercontent.com/kubernetes-sigs/cloud-provider-azure/master/helm/repo"
+	cloudProviderAzureChartName      = "cloud-provider-azure"
+	cloudProvierAzureHelmReleaseName = "cloud-provider-azure-oot"
 )
 
 // InstallCloudProviderAzureHelmChart installs the official cloud-provider-azure helm chart
+// Fulfills the clusterctl.Waiter type so that it can be used as ApplyClusterTemplateAndWaitInput data
+// in the flow of a clusterctl.ApplyClusterTemplateAndWait E2E test scenario
 func InstallCloudProviderAzureHelmChart(ctx context.Context, input clusterctl.ApplyClusterTemplateAndWaitInput, result *clusterctl.ApplyClusterTemplateAndWaitResult) {
-	By("Waiting for workload cluster kubeconfig secret")
-	Eventually(func() error {
-		client := input.ClusterProxy.GetClient()
-		secret := &corev1.Secret{}
-		key := crclient.ObjectKey{
-			Name:      fmt.Sprintf("%s-kubeconfig", input.ConfigCluster.ClusterName),
-			Namespace: input.ConfigCluster.Namespace,
-		}
-		return client.Get(ctx, key, secret)
-	}, input.WaitForControlPlaneIntervals...).Should(Succeed())
-	clusterProxy := input.ClusterProxy.GetWorkloadCluster(ctx, input.ConfigCluster.Namespace, input.ConfigCluster.ClusterName)
-	By("Waiting for nodes to come online indicating that the cluster is ready to accept work")
-	Eventually(func() error {
-		clientSet := clusterProxy.GetClientSet()
-		var runningNodes int
-		list, err := clientSet.CoreV1().Nodes().List(ctx, v1.ListOptions{})
-		if err != nil {
-			return err
-		}
-		for _, n := range list.Items {
-			if n.Status.Phase == corev1.NodeRunning {
-				runningNodes++
-			}
-		}
-		if runningNodes > 0 {
-			return nil
-		}
-		return err
-	}, input.WaitForControlPlaneIntervals...).Should(Succeed())
+	By(fmt.Sprintf("Ensuring the kubeconfig secret for cluster %s/%s exists before installing cloud-provider-azure components", input.ConfigCluster.Namespace, input.ConfigCluster.ClusterName))
+	WaitForWorkloadClusterKubeconfigSecret(ctx, input)
 	By("Installing the correct version of cloud-provider-azure components via helm")
-	kubeConfigPath := clusterProxy.GetKubeconfigPath()
-	clusterName := input.ClusterProxy.GetName()
-	settings := helmCli.New()
-	settings.KubeConfig = kubeConfigPath
-	actionConfig := new(helmAction.Configuration)
-	err := actionConfig.Init(settings.RESTClientGetter(), "default", "secret", Logf)
-	Expect(err).To(BeNil())
-	i := helmAction.NewInstall(actionConfig)
-	i.RepoURL = "https://raw.githubusercontent.com/kubernetes-sigs/cloud-provider-azure/master/helm/repo"
-	i.ReleaseName = "cloud-provider-azure-oot"
-	Eventually(func() error {
-		cp, err := i.ChartPathOptions.LocateChart("cloud-provider-azure", helmCli.New())
-		if err != nil {
-			return err
-		}
-		p := helmGetter.All(settings)
-		valueOpts := &helmVals.Options{}
-		valueOpts.Values = []string{fmt.Sprintf("infra.clusterName=%s", clusterName)}
-		vals, err := valueOpts.MergeValues(p)
-		if err != nil {
-			return err
-		}
-		chartRequested, err := helmLoader.Load(cp)
-		if err != nil {
-			return err
-		}
-		release, err := i.RunWithContext(ctx, chartRequested, vals)
-		if err != nil {
-			return err
-		}
-		Logf(release.Info.Description)
-		return nil
-	}, input.WaitForControlPlaneIntervals...).Should(Succeed())
+	values := []string{fmt.Sprintf("infra.clusterName=%s", input.ConfigCluster.ClusterName)}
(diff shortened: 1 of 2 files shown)


jsturtevant:
Looks like the metric server has a chart.  Some of the window specific things like kube-proxy/calico don't have charts.

jackfrancis:
How much customization are we doing for metrics-server, kube-proxy, and calico? Depending upon the amount of "capz optimizations" that we're performing in these yaml specs it may be appropriate to host our own "capz-tested" charts for such components in this repo. And there's little work to maintain a canonical repo like we do here:

https://github.com/kubernetes-sigs/cloud-provider-azure/tree/master/helm

jsturtevant:
We can likely use the metric-server chart.  For Windows kube-proxy/calico the yaml is from https://github.com/kubernetes-sigs/sig-windows-tools/tree/master/hostprocess.  That would be the place where we could add those charts.  The long term goal is for sig-windows to not host the calico and kube-proxy images so I am hesitant to put much effort in to create charts.  If capz is moving away from CRS completely shorter term then I guess we might need too as it is still a ways out before we can get those images to proper owner repositories do to changes needed in the HostProcess containers implementation.

test/e2e/helpers.go
```diff
@@ -823,3 +829,94 @@ func getPodLogs(ctx context.Context, clientset *kubernetes.Clientset, pod corev1
 	}
 	return b.String()
 }
+
+func WaitForClusterKubeconfigSecret(ctx context.Context, input clusterctl.ApplyClusterTemplateAndWaitInput) {
```
CecileRobertMichon, line 833: this might be a helper in the Cluster API test framework already. If not, it should be. Nothing CAPZ specific here.

test/e2e/helpers.go
```diff
@@ -823,3 +829,94 @@ func getPodLogs(ctx context.Context, clientset *kubernetes.Clientset, pod corev1
...
+		secret := &corev1.Secret{}
+		key := client.ObjectKey{
+			Name:      fmt.Sprintf("%s-kubeconfig", input.ConfigCluster.ClusterName),
+			Namespace: input.ConfigCluster.Namespace,
+		}
+		return c.Get(ctx, key, secret)
+	}, input.WaitForControlPlaneIntervals...).Should(Succeed())
+}
+
+func WaitForRunningNodes(ctx context.Context, input clusterctl.ApplyClusterTemplateAndWaitInput) {
```
CecileRobertMichon, line 846: same here and for running pods below. As much as possible we should share e2e logic with the other providers in the CAPI framework instead of implementing our own logic.

test/e2e/helpers.go
jackfrancis: Good call-out. In fact we don't need to run these kubeconfig and running nodes checks, as the retryable "install helm chart" func will absorb the failures due to the workload cluster not being actually running yet.

test/e2e/helpers.go
jackfrancis: I was able to rely upon pre-existing cluster-api conveniences for a part of this, but there is the "ensure n pods are Running" that I had to implement here.

jackfrancis:
@CecileRobertMichon this should be ready for a review round

test/e2e/cloud-provider-azure.go
```diff
@@ -26,119 +26,49 @@ import (
...
-	}, input.WaitForControlPlaneIntervals...).Should(BeTrue())
+	cloudNodeManagerPodLabel, err := labels.Parse("k8s-app=cloud-node-manager")
+	Expect(err).ToNot(HaveOccurred())
+	framework.WaitForPodListCondition(ctx, framework.WaitForPodListConditionInput{
+		Lister: c,
+		ListOptions: &client.ListOptions{
+			LabelSelector: cloudNodeManagerPodLabel,
+			Namespace:     "kube-system",
+		},
+		Condition: podListHasNumPods(int(to.Int64(input.ConfigCluster.ControlPlaneMachineCount) + to.Int64(input.ConfigCluster.WorkerMachineCount))),
```
CecileRobertMichon, line 70: does this also count also need to add windows worker nodes?

test/e2e/helpers.go
```diff
@@ -823,3 +829,54 @@ func getPodLogs(ctx context.Context, clientset *kubernetes.Clientset, pod corev1
...
+		release, err := i.RunWithContext(ctx, chartRequested, vals)
+		if err != nil {
+			return err
+		}
+		Logf(release.Info.Description)
+		return nil
+	}, input.WaitForControlPlaneIntervals...).Should(Succeed())
+}
+
+func podListHasNumPods(numPods int) func(pl *corev1.PodList) error {
```
CecileRobertMichon, line 869: is the reason we need to do this because the helper func doesn't actually check the number of pods?

test/e2e/cloud-provider-azure.go
jackfrancis: I think what we're saying is that we need to multiply the `input.ConfigCluster.WorkerMachineCount` by the number of pools in the cluster.

The current OOT template only has a single node pool, which is why this is working.

I don't think we have a way to determine the number of pools from the input to this func, but because we enter into this after all the control plane nodes and machines are online with node references, we should be able to count the number of nodes in the cluster, right?

test/e2e/helpers.go
jackfrancis: This reformulated approach is based on trying to re-use existing capi E2E foo.

`framework.WaitForPodListCondition` is the most appropriate existing thing that I could find, and it takes a `framework.PodListCondition` to be able to enforce what we're looking for. In this particular test case validating that the resultant pod list has n number of running pods is what we want.

test/e2e/cloud-provider-azure.go
CecileRobertMichon: windows is its own variable so we should be able to at least add that to the count (even if it's zero for now, that will make it more future proof)

test/e2e/cloud-provider-azure.go
```diff
@@ -23,122 +23,54 @@ import (
...
+	By(fmt.Sprintf("Waiting for %d Running cloud-node-manager pods", len(machineList.Items)))
+	cloudNodeManagerPodLabel, err := labels.Parse("k8s-app=cloud-node-manager")
+	Expect(err).ToNot(HaveOccurred())
+	framework.WaitForPodListCondition(ctx, framework.WaitForPodListConditionInput{
+		Lister: workloadClusterClient,
+		ListOptions: &client.ListOptions{
+			LabelSelector: cloudNodeManagerPodLabel,
+			Namespace:     "kube-system",
+		},
+		Condition: podListHasNumPods(len(machineList.Items)),
```
jackfrancis, line 72: @CecileRobertMichon I think counting the machines in the cluster is probably the best approach for this

test/e2e/helpers.go
```diff
@@ -823,3 +829,68 @@ func getPodLogs(ctx context.Context, clientset *kubernetes.Clientset, pod corev1
...
+			Name:      fmt.Sprintf("%s-kubeconfig", input.ConfigCluster.ClusterName),
+			Namespace: input.ConfigCluster.Namespace,
+		}
+		err := cl.Get(ctx, key, secret)
+		if err != nil {
+			return err
+		}
+		return nil
+	}, input.WaitForControlPlaneIntervals...).Should(Succeed())
+	clusterProxy := input.ClusterProxy.GetWorkloadCluster(ctx, input.ConfigCluster.Namespace, input.ConfigCluster.ClusterName)
```
jackfrancis, line 848: Unfortunately in my testing I encountered an edge case where `input.ClusterProxy.GetWorkloadCluster` can produce a 404 grabbing the secret (and it's not retryable as there are embedded ginkgo expects in that func), so I put the above `Eventually` wrapper to ensure that we register a successful kubeconfig secret retrieval before proceeding.

test/e2e/cloud-provider-azure.go
CecileRobertMichon: what about machine pools? Should we count nodes instead?

jackfrancis: We should probably count both Machines and MachinePoolMachines, right? (a cluster can exist with both types, I assume)

CecileRobertMichon: IMO doing it by number of nodes is more future-proof. There's no guarantee CAPI won't add a new type of Machine in the future.

jackfrancis: The problem with using nodes is that they are not the authoritative unit in terms of calculating "expected number" outcomes. For example, a node may be offline, or not yet joined, or other states that are not as definitive in terms of gathering a cluster count as compared to number of machines (plus number of machinepoolmachines).

So, if we use nodes then we'd probably want to change the pod list criteria to >= to account for the fact that a new node may have joined after we calculated num nodes, but before we gathered a list of cloud-node-manager pods.

It would be nice to establish a resilient (hopefully simple to understand and maintain) pattern here, as I do think being able to both validate the helm install and validate expected outcomes of that helm install is valuable.

CecileRobertMichon: I see. In that case thoughts about adding helper functions to the CAPI test framework that return the number of expected control plane nodes and the number of expected worker nodes? Control planes can be the total number of control plane replicas of the control plane object (or 0 if the control plane does not implement replicas https://cluster-api.sigs.k8s.io/developer/architecture/controllers/control-plane.html?highlight=spec#required-spec-fields-for-implementations-using-replicas). Worker nodes can be the total of machines + machine pool replicas. 

It's a bit more resilient to abstract what a worker machine can be and have one central place to make that assumption than having to update every place in our CAPZ tests where we make that assumption later if a new type of Machine is added. 

Also worth checking if something like this doesn't already exist in CAPI, it's very possible it does.

jackfrancis: I checked and capi E2E doesn't have a convenience for this (there are things sort of similar, with lots of buried `Expect()` funcs that are in the process of being cleaned up, but not for this purpose explicitly). How about this as a way of getting this merged ASAP.

1) We verify the existence of the cloud-node-manager daemonset
2) We wait until all of its replicas are Ready

test/e2e/helpers.go
```diff
@@ -823,3 +829,90 @@ func getPodLogs(ctx context.Context, clientset *kubernetes.Clientset, pod corev1
...
+		err := cl.Get(ctx, key, secret)
+		if err != nil {
+			return err
+		}
+		return nil
+	}, input.WaitForControlPlaneIntervals...).Should(Succeed())
+}
+
+// WaitForDaemonset retries during E2E until a daemonset's pods are all Running
+func WaitForDaemonset(ctx context.Context, input clusterctl.ApplyClusterTemplateAndWaitInput, cl client.Client, name, namespace string) {
```
jackfrancis, line 889: @CecileRobertMichon this is the alternative solution to validate the `cloud-node-manager` pods. Rather than comparing against number of machines, let's just make sure the daemonset pods are all running. There are several reasons out of our control why the number of nodes running these pods may not be easily predictable. Rather than try to deal with all of those edge cases, we can simply rely upon the daemonset API to pick the appropriate schedulable nodes, and ensure that the pods reach a Running state.

CecileRobertMichon (commented):
Thanks @jackfrancis, looks great

/lgtm
/assign @Jont828

CecileRobertMichon:
@jackfrancis we should now be able to switch over Calico as well. I believe we just need a bit of kustomization in the values.yaml to make it work for Azure (mainly VXLAN encapsulation)

https://kubernetes.slack.com/archives/CEX9HENG7/p1652473459598119

Jont828:
This looks great! Was useful for getting us off the ground with HelmChartProxy as well!

/lgtm

PR #1729, opened 2021-09-24
Add Flatcar Container Linux template

jepio:
**What type of PR is this?**
/kind other

**What this PR does / why we need it**:

This PR adds a cluster template for Flatcar Container Linux based clusters.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Related to https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/629

**Special notes for your reviewer**:
The Flatcar cluster template works with images built using this image-builder PR https://github.com/kubernetes-sigs/image-builder/pull/701 and this cluster-api PR https://github.com/kubernetes-sigs/cluster-api/pull/4172.

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

No image versions are changed.

**TODOs**:

- [X] squashed commits <- *single commit*
- [ ] includes documentation 
- [ ] adds unit tests

**Release note**:

```release-note
Flatcar Container Linux is now supported as a cluster flavor. Try `clusterctl generate cluster my-cluster --flavor flatcar`.
```

PR #2277, opened 2022-05-02
Add ComputeGallery field and add community galleries support

### LINKED ISSUE #2267: Add support for consuming images from community galleries
/kind feature

Recently made available in [public preview community galleries](https://azure.microsoft.com/en-us/updates/communitygallerypreview/) feature for consuming the SIG images from SIGs in other tenants and subscriptions seem very useful for 3rd parties to publish their own images for public consumption, so there is no need to publish in the marketplace, which as far as I know is a complex process and also specifically to add the CI for Flatcar PR/templates #1729.

It seems this requires use of [dedicated API field in ImageReference](https://github.com/Azure/azure-sdk-for-go/blob/185e159d240e22132f83b1196aeebd5389993313/services/compute/mgmt/2021-12-01/compute/models.go#L9677) to consume those images. We may want to ensure we use new enough SDK version and expose those fields in CAPZ CRDs to make it available for consumption.

CC @t-lo @jepio

invidian:
**What type of PR is this?**

/kind feature

This PR adds support for creating machines using recently made available in [public preview community galleries](https://azure.microsoft.com/en-us/updates/communitygallerypreview/)  images, which can be shared between subscriptions and tenants, which makes it easier to share than publishing to marketplace.

Fixes #2267

Extracted from #1729, e2e tests are done there. Please let me know what kind of tests would you like to have added in this PR.

**TODOs**:

- [ ] squashed commits
- [x] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Added support for consuming community gallery images via new `ComputeGallery` field in `Image` reference field. It replaces `SharedGallery` field which is now deprecated.
```


azure/services/resourceskus/client.go
```diff
@@ -57,7 +57,7 @@ func (ac *AzureClient) List(ctx context.Context, filter string) ([]compute.Resou
 	ctx, _, done := tele.StartSpanWithLogger(ctx, "resourceskus.AzureClient.List")
 	defer done()
 
-	iter, err := ac.skus.ListComplete(ctx, filter)
+	iter, err := ac.skus.ListComplete(ctx, filter, "true")
```
invidian, line 60: I couldn't find documentation for this parameter and passing empty string was giving me errors, so I'm not sure what the correct value should be here.

alexeldeib: seems like it's supposed to be a string-bool as you eventually figured? 

the examples in azure-rest-api-specs seem to use "true" like you did here, did this not work?

invidian: `"true"` works indeed. but if it's string-bool, I don't know, perhaps it should be `"false"`?

invidian: So it seems this is about https://docs.microsoft.com/en-us/rest/api/extendedlocation/, so I guess it make sense to put `"true"` here to avoid potentially limiting users, though I don't know how well the resource SKUs work with extended locations.

api/v1beta1/types.go
```diff
@@ -324,6 +324,44 @@ type Image struct {
 	// Marketplace specifies an image to use from the Azure Marketplace
 	// +optional
 	Marketplace *AzureMarketplaceImage `json:"marketplace,omitempty"`
+
+	// CommunityGallery specifies an image to use from the Azure Community Gallery
+	// +optional
+	CommunityGallery *AzureCommunityGalleryImage `json:"communityGallery,omitempty"`
+}
+
+// AzureCommunityGalleryImage defines an image in the Azure Community Gallery ot use for VM creation.
```
jepio, line 333:
```suggestion
// AzureCommunityGalleryImage defines an image in the Azure Community Gallery to use for VM creation.
```

invidian: Good reminder to fix typos in entire repo! :smile: Thanks for catching this and for reviewing!

azure/services/resourceskus/client.go
invidian: On the other hand, it seems we are always listing by location name it seems, so hopefully this supports custom locations as well.

api/v1beta1/types.go
```diff
@@ -324,6 +324,44 @@ type Image struct {
...
 	// +optional
 	Marketplace *AzureMarketplaceImage `json:"marketplace,omitempty"`
+
+	// CommunityGallery specifies an image to use from the Azure Community Gallery
+	// +optional
+	CommunityGallery *AzureCommunityGalleryImage `json:"communityGallery,omitempty"`
+}
+
+// AzureCommunityGalleryImage defines an image in the Azure Community Gallery to use for VM creation.
+type AzureCommunityGalleryImage struct {
```
CecileRobertMichon, line 334: seems like there is a lot of overlap between community galleries and shared galleries (which have since been renamed to compute galleries), have you considered either reusing the existing type, creating a new one that applies to both and deprecating the old one, or extracting the common fields to a struct that can be included inline in both shared and community galleries that contains the gallery details?

invidian: I guess it could make sense, I'll have a look.

CecileRobertMichon: I'm personally not a fan of option 2 ("Deprecate both SharedGallery and Marketplace, move all fields into Image struct") because it puts a lot of burden on the user to figure out which fields are relevant for which image type. 

My suggestion was simply to avoid introducing a whole new type for Community galleries as they are logically the same as Azure Compute (private) galleries, previously known as Shared Image Galleries. In fact, according to docs, a Compute gallery can be mutated to become public after it exists (and vice versa): https://docs.microsoft.com/en-us/azure/virtual-machines/create-gallery?tabs=portal%2Ccli2#create-a-community-gallery-preview

+1 to option 1 as it is what best reflects the logical distinction of Azure Image types and allows us to better align with the official rename of Azure Compute Galleries. 

In order to share fields we could also do something like this: 

```
type Image struct {
        ID             *string                   `json:"id,omitempty"`
        SharedGallery  *AzureSharedGalleryImage  `json:"sharedGallery,omitempty"`
        Marketplace    *AzureMarketplaceImage    `json:"marketplace,omitempty"`
        ComputeGallery *AzureComputeGalleryImage `json:"computeGallery,omitempty"`
}

type AzureSharedGalleryImage struct {
        AzureComputeGalleryImage `json:",inline"`
        SubscriptionID *string `json:"subscriptionID"`
        ResourceGroup  *string `json:"resourceGroup"`
}

type AzureComputeGalleryImage struct {
        ImagePlan `json:",inline"`
        Gallery        string  `json:"gallery"`
        Name           string  `json:"name"`
        Version        string  `json:"version"`
}

type ImagePlan struct {
        Publisher      *string `json:"publisher,omitempty"`
        Offer          *string `json:"offer,omitempty"`
        SKU            *string `json:"sku,omitempty"`
}

type AzureMarketplaceImage struct {
        ImagePlan `json:",inline"`
        Version         *string `json:"version"`
        ThirdPartyImage bool    `json:"thirdPartyImage"`
}
```

invidian: Before I jump in into implementing all crazy conversion functions, what do you think about the types below?

So from what I saw, plan information must consist of Publisher, Offer and SKU, not having any of this fields populated results in an API error (correct me if there are cases when this is not true). Given that, let's make `ImagePlan` have only non-pointer required fields and control the requirements of those fields one level above.

So for `AzureMarketplaceImage`, `ImagePlan` is required, while for `AzureSharedGalleryImage` and `AzureComputeGalleryImage` fields are optional. I wonder if kubebuilder is smart enough about `+optional` on inlined fields.

`Version` field remains duplicated across all types for the time being.

And regardless of the change, for Go types this will be a breaking change. For YAML I think types remains backward compatible.

```go
// Image defines information about the image to use for VM creation.
// There are three ways to specify an image: by ID, Marketplace Image, ComputeGallery or SharedImageGallery
// One of ID, SharedImage, ComputeGallery or Marketplace should be set.
type Image struct {
        // ID specifies an image to use by ID
        // +optional
        ID *string `json:"id,omitempty"`

        // SharedGallery specifies an image to use from an Azure Shared Image Gallery
        // Deprecated: use computeGallery instead.
        // +optional
        SharedGallery *AzureSharedGalleryImage `json:"sharedGallery,omitempty"`

        // Marketplace specifies an image to use from the Azure Marketplace
        // +optional
        Marketplace *AzureMarketplaceImage `json:"marketplace,omitempty"`

        // ComputeGallery specifies an image to use from the Azure Compute Gallery
        // +optional
        ComputeGallery *AzureComputeGalleryImage `json:"computeGallery,omitempty"`
}

// AzureComputeGalleryImage defines an image in the Azure Compute Gallery to use for VM creation.
type AzureComputeGalleryImage struct {
        // ImagePlan contains `Plan` information for API request when creating the VM/VMSS resource.
        // +optional
        *ImagePlan `json:",inline"`

        // Gallery specifies the name of the compute gallery that contains the image. If specified without
        // subscriptionID and resourceGroup, image assumes reference to a community gallery.
        // +kubebuilder:validation:MinLength=1
        Gallery string `json:"gallery"`
        // Name is the name of the image
        // +kubebuilder:validation:MinLength=1
        Name string `json:"name"`
        // Version specifies the version of the compute gallery image. The allowed formats
        // are Major.Minor.Build or 'latest'. Major, Minor, and Build are decimal numbers.
        // Specify 'latest' to use the latest version of an image available at deploy time.
        // Even if you use 'latest', the VM image will not automatically update after deploy
        // time even if a new version becomes available.
        // +kubebuilder:validation:MinLength=1
        Version string `json:"version"`

        // SubscriptionID is the identifier of the subscription that contains the compute gallery.
        // Must be specified together with resourceGroup. if not specified, image assumes reference to
        // a community gallery.
        // +optional
        SubscriptionID *string `json:"subscriptionID"`
        // ResourceGroup specifies the resource group containing the compute gallery. Must be specified
        // together with subscriptionID. If not specified, image assumes reference to a community gallery.
        // +optional
        ResourceGroup *string `json:"resourceGroup"`
}

// ImagePlan contains `Plan` information for API request when creating the VM/VMSS resource.
type ImagePlan struct {
        // Publisher is the name of the organization that created the image.
        // This value will be used to add a `Plan` in the API request when creating the VM/VMSS resource.
        // This is needed when the source image from which this SIG image was built requires the `Plan` to be used.
        // +kubebuilder:validation:MinLength=1
        Publisher string `json:"publisher,omitempty"`
        // Offer specifies the name of a group of related images created by the publisher.
        // For example, UbuntuServer, WindowsServer
        // This value will be used to add a `Plan` in the API request when creating the VM/VMSS resource.
        // This is needed when the source image from which this SIG image was built requires the `Plan` to be used.
        // +kubebuilder:validation:MinLength=1
        Offer string `json:"offer,omitempty"`
        // SKU specifies an instance of an offer, such as a major release of a distribution.
        // For example, 18.04-LTS, 2019-Datacenter
        // This value will be used to add a `Plan` in the API request when creating the VM/VMSS resource.
        // This is needed when the source image from which this SIG image was built requires the `Plan` to be used.
        // +kubebuilder:validation:MinLength=1
        SKU string `json:"sku,omitempty"`
}

// AzureSharedGalleryImage defines an image in a Shared Image Gallery to use for VM creation.
type AzureSharedGalleryImage struct {
        // ImagePlan contains `Plan` information for API request when creating the VM/VMSS resource.
        // +optional
        *ImagePlan `json:",inline"`

        // SubscriptionID is the identifier of the subscription that contains the shared image gallery
        // +kubebuilder:validation:MinLength=1
        SubscriptionID string `json:"subscriptionID"`
        // ResourceGroup specifies the resource group containing the shared image gallery
        // +kubebuilder:validation:MinLength=1
        ResourceGroup string `json:"resourceGroup"`
        // Gallery specifies the name of the shared image gallery that contains the image
        // +kubebuilder:validation:MinLength=1
        Gallery string `json:"gallery"`
        // Name is the name of the image
        // +kubebuilder:validation:MinLength=1
        Name string `json:"name"`
        // Version specifies the version of the marketplace image. The allowed formats
        // are Major.Minor.Build or 'latest'. Major, Minor, and Build are decimal numbers.
        // Specify 'latest' to use the latest version of an image available at deploy time.
        // Even if you use 'latest', the VM image will not automatically update after deploy
        // time even if a new version becomes available.
        // +kubebuilder:validation:MinLength=1
        Version string `json:"version"`
}

// AzureMarketplaceImage defines an image in the Azure Marketplace to use for VM creation.
type AzureMarketplaceImage struct {
        // ImagePlan contains `Plan` information for API request when creating the VM/VMSS resource.
        ImagePlan `json:",inline"`

        // Version specifies the version of an image sku. The allowed formats
        // are Major.Minor.Build or 'latest'. Major, Minor, and Build are decimal numbers.
        // Specify 'latest' to use the latest version of an image available at deploy time.
        // Even if you use 'latest', the VM image will not automatically update after deploy
        // time even if a new version becomes available.
        // +kubebuilder:validation:MinLength=1
        Version string `json:"version"`
        // ThirdPartyImage indicates the image is published by a third party publisher and a Plan
        // will be generated for it.
        // +kubebuilder:default=false
        // +optional
        ThirdPartyImage bool `json:"thirdPartyImage"`
}
```

invidian: Honestly, I don't like existing plan fields definitions in `AzureSharedGalleryImage` either. They are currently defined as:
```go
	// +optional
	Publisher *string `json:"publisher,omitempty"`
	// +optional
	Offer *string `json:"offer,omitempty"`
	// +optional
	SKU *string `json:"sku,omitempty"`
```

Created #2286 around it.

I'd rather see those fields in a struct as they are currently defined in `AzureMarketplaceImage`:
```go
  // Publisher is the name of the organization that created the image
  // +kubebuilder:validation:MinLength=1
  Publisher string `json:"publisher"`
  // Offer specifies the name of a group of related images created by the publisher.
  // For example, UbuntuServer, WindowsServer
  // +kubebuilder:validation:MinLength=1
  Offer string `json:"offer"`
  // SKU specifies an instance of an offer, such as a major release of a distribution.
  // For example, 18.04-LTS, 2019-Datacenter
  // +kubebuilder:validation:MinLength=1
  SKU string `json:"sku"`
```

And in `AzureMarketplaceImage` they would be inlined, while in `AzureComputeGalleryImage`, they would be part of `+optional` `Plan` field. And we would deprecate `AzureSharedGalleryImage`. Thoughts?

invidian: I can update `AzureMarketplaceImage`, on the other hand it could be done completely independently after introducing the right type.

invidian: Done what proposed above. Also extracted compute update to #2288, so perhaps it can be merged sooner than this PR.

invidian:
/retest

`pull-cluster-api-provider-azure-e2e` and `pull-cluster-api-provider-azure-ci-entrypoint` seems like a flake, `pull-cluster-api-provider-azure-apidiff` is expected to be failing as I indeed made an incompatible change on Go level to the type.

invidian:
Something seems to be off with e2e tests, as they pass in #1729, which includes https://github.com/kubernetes-sigs/cluster-api-provider-azure/commit/abf8c8f21036d9a95190438a5056c9e3b7c71a2a.

CecileRobertMichon:
this is the error I'm seeing in logs

> I0510 14:54:52.210761       1 recorder.go:103] events "msg"="Warning"  "message"="failed to reconcile AzureMachine: failed to reconcile AzureMachine service virtualmachine: failed checking if the operation was complete: failed checking if the operation was complete: Code=\"ZonalAllocationFailed\" Message=\"Allocation failed. We do not have sufficient capacity for the requested VM size in this zone. Read more about improving likelihood of allocation success at http://aka.ms/allocation-guidance\"" "object"={"kind":"AzureMachine","namespace":"capz-e2e-ahqhol","name":"capz-e2e-ahqhol-ha-control-plane-s7fj6","uid":"e50859a4-e0f4-4fe7-912b-04af5efdad05","apiVersion":"infrastructure.cluster.x-k8s.io/v1beta1","resourceVersion":"2898"} "reason"="ReconcileError"

:(

api/v1beta1/types.go
CecileRobertMichon: Will do a full review soon, but this structure looks good to me 👍

api/v1alpha3/azuremachine_conversion.go
```diff
@@ -18,6 +18,7 @@ package v1alpha3
 
 import (
 	apiconversion "k8s.io/apimachinery/pkg/conversion"
+	apimachineryconversion "k8s.io/apimachinery/pkg/conversion"
```
CecileRobertMichon, line 21: this import is duplicated from above

api/v1alpha4/azuremachine_conversion.go
```diff
@@ -17,20 +17,41 @@ limitations under the License.
 package v1alpha4
 
 import (
+	apimachineryconversion "k8s.io/apimachinery/pkg/conversion"
```
CecileRobertMichon, line 20: let's keep the import naming consistent with other packages

```suggestion
	apiconversion "k8s.io/apimachinery/pkg/conversion"
```

api/v1beta1/azureimage_validation.go
```diff
@@ -62,14 +65,45 @@ func validateSingleDetailsOnly(image *Image, fldPath *field.Path) field.ErrorLis
 
 	if image.SharedGallery != nil {
 		if imageDetailsFound {
-			allErrs = append(allErrs, field.Forbidden(fldPath.Child("SharedGallery"), "SharedGallery cannot be used as an image ID or Marketplace images has been specified"))
+			allErrs = append(allErrs, field.Forbidden(fldPath.Child("SharedGallery"), "SharedGallery cannot be used as an image ID, Marketplace or ComputeGallery images has been specified"))
+		} else {
+			imageDetailsFound = true
```
CecileRobertMichon, line 70: Is it possible to add a [deprecation warning](https://kubernetes.io/blog/2020/09/03/warnings/#deprecation-warnings) when shared gallery is used?

api/v1beta1/azureimage_validation.go
```diff
@@ -62,14 +65,45 @@ func validateSingleDetailsOnly(image *Image, fldPath *field.Path) field.ErrorLis
...
+		if imageDetailsFound {
+			allErrs = append(allErrs, field.Forbidden(fldPath.Child("ComputeGallery"), "ComputeGallery cannot be used as an image ID, Marketplace or SharedGallery images has been specified"))
 		} else {
 			imageDetailsFound = true
 		}
 	}
 
 	if !imageDetailsFound {
-		allErrs = append(allErrs, field.Required(fldPath, "You must supply a ID, Marketplace or SharedGallery image details"))
+		allErrs = append(allErrs, field.Required(fldPath, "You must supply a ID, Marketplace, SharedGallery or ComputeGallery image details"))
```
CecileRobertMichon, line 83: let's put ComputeGallery first (only?) since we don't want to recommend SharedGallery going forward

api/v1beta1/azureimage_validation.go
```diff
@@ -62,14 +65,45 @@ func validateSingleDetailsOnly(image *Image, fldPath *field.Path) field.ErrorLis
...
+	}
+
+	return allErrs
+}
+
+func validateComputeGalleryImage(image *Image, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	if image.ComputeGallery.Gallery == "" {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("Gallery"), "", "Gallery cannot be empty when specifying an AzureComputeGalleryImage"))
```
CecileRobertMichon, line 93: I don't think we need this validation since it's already done from the OpenAPI validation?

api/v1beta1/types.go
```diff
@@ -318,16 +318,45 @@ type Image struct {
...
+	// +kubebuilder:validation:MinLength=1
+	Version string `json:"version"`
+	// SubscriptionID is the identifier of the subscription that contains the shared image gallery.
+	// +optional
+	SubscriptionID *string `json:"subscriptionID,omitempty"`
+	// ResourceGroup specifies the resource group containing the shared image gallery.
+	// +optional
+	ResourceGroup *string `json:"resourceGroup,omitempty"`
+	// Plan contains plan information.
+	Plan *ImagePlan `json:"plan,omitempty"`
```
CecileRobertMichon, line 356: should this be marked as optional?

azure/converters/image.go
```diff
@@ -68,6 +71,29 @@ func specificImageToSDK(image *infrav1.Image) (*compute.ImageReference, error) {
...
+	if image.ComputeGallery.ResourceGroup == nil {
+		return &compute.ImageReference{
+			CommunityGalleryImageID: to.StringPtr(fmt.Sprintf("/CommunityGalleries/%s/Images/%s/Versions/%s",
+				image.ComputeGallery.Gallery,
+				image.ComputeGallery.Name,
+				image.ComputeGallery.Version)),
+		}, nil
+	}
+
+	idTemplate := "/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Compute/galleries/%s/images/%s/versions/%s"
```
CecileRobertMichon, line 84: can we reuse `sigImageToSDK` for this?

exp/api/v1beta1/azuremachinepool_webhook_test.go
```diff
@@ -263,10 +263,12 @@ func TestAzureMachinePool_Default(t *testing.T) {
 func createMachinePoolWithtMarketPlaceImage(publisher, offer, sku, version string, terminateNotificationTimeout *int) *AzureMachinePool {
```
CecileRobertMichon, line 263: there is a typo in the function name if you want to fix it while you're here :)

```suggestion
func createMachinePoolWithMarketPlaceImage(publisher, offer, sku, version string, terminateNotificationTimeout *int) *AzureMachinePool {
```

api/v1beta1/azureimage_validation.go
invidian: Possibly, yeah. Though the OpenAPI validations are currently not tested (#2287). So I can either remove those validations here or add tests for them.

api/v1beta1/types.go
invidian: Hmm, it seems adding it has no effect on the manifests. But I'll add it for consistency.

azure/converters/image.go
invidian: We would have to map the parameters from `image.ComputeGallery` to `image.SharedGallery`, which I'd rather not do, I'd convert other way around. So I would either do:
```go
       if image.SharedGallery != nil {
               return computeImageToSDK(sigToCompute(image))
       if image.ComputeGallery != nil {
               return computeImageToSDK(image)
       }
```

or handle the conversion in `computeImageToSDK` and do:
```go
       if image.ComputeGallery != nil || image.SharedGallery != nil {
               return computeImageToSDK(image)
       }
...

func computeImageToSDK(image *infrav1.Image) (*compute.ImageReference, error) {
  idTemplate := "/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Compute/galleries/%s/images/%s/versions/%s"

  if image.SharedGallery != nil {
    return &compute.ImageReference{
      ID: to.StringPtr(fmt.Sprintf(idTemplate,
        image.SharedGallery.SubscriptionID,
        image.SharedGallery.ResourceGroup,
        image.SharedGallery.Gallery,
        image.SharedGallery.Name,
        image.SharedGallery.Version,
      )),
    }, nil
  }

  if image.ComputeGallery.ResourceGroup != nil {
    return &compute.ImageReference{
      ID: to.StringPtr(fmt.Sprintf(idTemplate,
        image.ComputeGallery.SubscriptionID,
        image.ComputeGallery.ResourceGroup,
        image.ComputeGallery.Gallery,
        image.ComputeGallery.Name,
        image.ComputeGallery.Version,
      )),
    }, nil
  }

  return &compute.ImageReference{
    CommunityGalleryImageID: to.StringPtr(fmt.Sprintf("/CommunityGalleries/%s/Images/%s/Versions/%s",
      image.ComputeGallery.Gallery,
      image.ComputeGallery.Name,
      image.ComputeGallery.Version)),
  }, nil
}
```

I'll push the 2nd approach to continue the discussion.

api/v1beta1/azureimage_validation.go
invidian: From what I see [kubebuilder only supports deprecating entire versions of objects](https://book.kubebuilder.io/reference/markers/crd.html) and generating custom warnings is currently not supported https://github.com/kubernetes-sigs/controller-runtime/issues/1896.

invidian:
Downloading helm fails on CI, though it works for me locally, let's try later...

api/v1beta1/types.go
```diff
@@ -318,16 +318,46 @@ type Image struct {
 	ID *string `json:"id,omitempty"`
 
 	// SharedGallery specifies an image to use from an Azure Shared Image Gallery
+	// Deprecated: use ComputeGallery instead.
```
jsturtevant, line 321: can we add something to the release note so this is called out?

jsturtevant: Thanks!

api/v1beta1/azureimage_validation.go
invidian: I assume :+1: means there is nothing left to do here, at least for now. Will mark as resolved.

api/v1beta1/azureimage_validation.go
CecileRobertMichon: My vote would be to remove them for simplicity since it's essentially unreachable code. I prefer to opt for OpenAPI validation whenever possible for straightforward validation and keep webhook validation for more involved validation logic.

Thanks for opening the issue to add tests. We have to think a bit more about it since it does add complexity to the tests and requires the extra binaries as you pointed out in the issue.

CecileRobertMichon:
cc @jsturtevant or @mboersma for a second pair of 👀 

lgtm pending openAPI comment

api/v1beta1/azureimage_validation.go
invidian: Sure, make sense. Removed then and marking it as resolved.

api/v1beta1/azureimage_validation.go
```diff
@@ -62,14 +65,35 @@ func validateSingleDetailsOnly(image *Image, fldPath *field.Path) field.ErrorLis
...
-			allErrs = append(allErrs, field.Forbidden(fldPath.Child("SharedGallery"), "SharedGallery cannot be used as an image ID or Marketplace images has been specified"))
+			allErrs = append(allErrs, field.Forbidden(fldPath.Child("SharedGallery"), "SharedGallery cannot be used as an image ID, Marketplace or ComputeGallery images has been specified"))
+		} else {
+			imageDetailsFound = true
+		}
+	}
+
+	if image.ComputeGallery != nil {
+		if imageDetailsFound {
+			allErrs = append(allErrs, field.Forbidden(fldPath.Child("ComputeGallery"), "ComputeGallery cannot be used as an image ID, Marketplace or SharedGallery images has been specified"))
```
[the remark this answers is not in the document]
jsturtevant: Found this error message hard to parse with the comma there

api/v1beta1/azureimage_validation.go
```diff
@@ -62,14 +65,35 @@ func validateSingleDetailsOnly(image *Image, fldPath *field.Path) field.ErrorLis
...
 		} else {
 			imageDetailsFound = true
 		}
 	}
 
 	if !imageDetailsFound {
-		allErrs = append(allErrs, field.Required(fldPath, "You must supply a ID, Marketplace or SharedGallery image details"))
+		allErrs = append(allErrs, field.Required(fldPath, "You must supply a ID, Marketplace or ComputeGallery image details"))
```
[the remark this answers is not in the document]
invidian: Out of scope for this PR, but fixed along the way :)

docs/book/src/topics/custom-images.md
```diff
@@ -76,7 +76,7 @@ spec:
   template:
     spec:
       image:
-        sharedGallery:
+        computeGallery:
```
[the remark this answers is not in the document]
invidian: >I think this should stay sharedGallery in the example.

This PR deprecates this field and `computeGallery` provides the same functionality if using SIG images without plan information. I think we shouldn't be promoting deprecated fields anymore.

At the same time, I don't think community galleries should be _recommended_ at the moment, as they are still in preview.

Could you also please elaborate what has #2293 to do with this PR? I can't see the connection?

docs/book/src/topics/custom-images.md
```diff
@@ -117,15 +117,16 @@ spec:
...
```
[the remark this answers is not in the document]
invidian: No, as far as I understand, Shared Image Galleries in Azure has been renamed to Azure Compute Gallery
, at least this is what for example [this](https://docs.microsoft.com/en-us/azure/virtual-machines/shared-image-galleries) URL suggests (see page title vs URL).

This PR reflects this by adding new field `computeGallery` to align the naming and also to use new convention of using separate time for Plan information, shared with Marketplace image to fight duplication of fields.

docs/book/src/topics/custom-images.md
```diff
@@ -171,6 +172,53 @@ spec:
           thirdPartyImage: true
 ```
 
+### Using Azure Community Gallery
```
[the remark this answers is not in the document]
invidian: Ditto, as community galleries are still in preview, I'd like to double check that it's actually something we want to recommend by default :)

Perhaps I should've split the commits different way in this PR, having one for new ComputeGallery type and another one for community galleries, perhaps that would make it more clear.

docs/book/src/topics/custom-images.md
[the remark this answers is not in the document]
jsturtevant: sorry the linked didn't copy paste properly, it should have been a link to a few lines above this section on line 66: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2277/files#diff-1bff25aa275cb44ca787e388c1774fcc0663c2be151dbbe9f3e44c6325a1fd30L66

> This PR deprecates this field and computeGallery provides the same functionality if using SIG images without plan information. I think we shouldn't be promoting deprecated fields anymore.

ok this makes sense. I was getting `Compute Image gallery` and` Community Galleries` conflated.  Maybe this section should be updated to `### Using Compute Image Gallery (Recommended)` instead of `### Using Shared Image Gallery (Recommended)` and add a note that it was renamed from SIG -> Compute Image Gallery

docs/book/src/topics/custom-images.md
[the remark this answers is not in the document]
jsturtevant: Yes I was getting the two conflated.  This makes sense to keep this change.

docs/book/src/topics/custom-images.md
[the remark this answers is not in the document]
invidian: >Maybe this section should be updated to ### Using Compute Image Gallery (Recommended) instead of ### Using Shared Image Gallery (Recommended) and add a note that it was renamed from SIG -> Compute Image Gallery

Yes, that definitely make sense. Perhaps something like:
```md
### Using Compute Gallery (Recommended)

To use an image from the [Compute Gallery][compute-gallery], previously known as Shared Image Gallery (SIG), fill in the `resourceGroup`, `name`, `subscriptionID`, `gallery`, and `version` fields:
```

I'm not sure if a warning or a dedicated note about the rename is needed here.

azure/converters/image.go
```diff
@@ -49,16 +49,38 @@ func mpImageToSDK(image *infrav1.Image) (*compute.ImageReference, error) {
...
+				image.SharedGallery.SubscriptionID,
+				image.SharedGallery.ResourceGroup,
+				image.SharedGallery.Gallery,
+				image.SharedGallery.Name,
+				image.SharedGallery.Version,
+			)),
+		}, nil
+	}
+
+	if image.ComputeGallery.ResourceGroup != nil {
```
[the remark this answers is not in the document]
jsturtevant: If a user is trying to specify a private image and they leave out the resource group, what happens? Does it fail to provision the VM because it can't find the image?

I am thinking a log here might be useful to someone debugging a scenario like that

docs/book/src/topics/custom-images.md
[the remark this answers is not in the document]
jsturtevant: lgtm, that way someone searching for the previous name will still find it

azure/converters/image.go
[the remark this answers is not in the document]
invidian: >If a user is trying to specify a private image and they leave out the resource group, what happens?

If subscription ID is specified or other way around (resource group given but no subscription ID), then validating webhook returns an error. If neither is specified, Gallery name is assumed to be the community gallery public name.

Do you think community gallery public name should have it's own field perhaps to make it more explicit? But then the only shared fields becomes plan and version, so we could as well introduce another type for the community galleries (which on it's own is not such a bad idea I guess and SIG -> Compute Gallery cleanup is still nice to have. 

@CecileRobertMichon since you suggested to extract common fields in this PR (as original version was adding only separate community gallery field), what do you think?

docs/book/src/topics/custom-images.md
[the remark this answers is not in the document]
invidian: >Perhaps I should've split the commits different way in this PR, having one for new ComputeGallery type and another one for community galleries, perhaps that would make it more clear.

I've committed it now separately, hope this helps. Marking as resolved.

docs/book/src/topics/custom-images.md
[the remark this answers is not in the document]
invidian: I'll mark it as resolved then.

docs/book/src/topics/custom-images.md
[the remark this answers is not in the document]
invidian: Fixed, so marking as resolved.

azure/converters/image.go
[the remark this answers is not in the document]
invidian: Hmm, I also had a look into https://github.com/hashicorp/packer-plugin-azure/pull/199, which is also about adding support for the directly shared galleries, which use yet another ID field. I think we should at least make the new type forward-compatible with those as well. As far as I understand, directly shared galleries also does not require subscription ID or resource group, which breaks my assumptions taken here.

Seems like directly shared galleries are referred using `SharedGalleryImageId` field in `ImageReference` field.

I guess we could introduce something like `Type` field let user specify how the image is shared or something.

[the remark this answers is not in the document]
invidian: I was thinking, if we could stat if a given Directly shared image exists or if a Community shared image is accessible, then we could save a hassle for the users to specify that and perhaps only require `sharingMethod` (new name which perhaps make most sense? I'm still trying to figure it out!) field when the information is ambiguous. 

Code like this seem to work just fine cross-tenants:
```go
  communityGalleryImage, err = client.Get(ctx, "westeurope", badPublicGalleryName, galleryImageName, imageVersion)
  if err != nil {
    derr := autorest.DetailedError{}
    if errors.As(err, &derr) && derr.StatusCode != 404 {
      panic(err)
    }

    fmt.Println("Image correctly not found")
  }
```

api/v1beta1/azureimage_validation.go
```diff
@@ -62,14 +65,35 @@ func validateSingleDetailsOnly(image *Image, fldPath *field.Path) field.ErrorLis
...
 
 	if !imageDetailsFound {
-		allErrs = append(allErrs, field.Required(fldPath, "You must supply a ID, Marketplace or SharedGallery image details"))
+		allErrs = append(allErrs, field.Required(fldPath, "You must supply an ID, Marketplace or ComputeGallery image details"))
+	}
+
+	return allErrs
+}
+
+func validateComputeGalleryImage(image *Image, fldPath *field.Path) field.ErrorList {
```
jsturtevant, line 89: did we add unit tests that validate this function?  I don't seem them but I might be missing something.

docs/book/src/topics/custom-images.md
```diff
@@ -171,6 +172,53 @@ spec:
...
+
+To use an image from [Azure Community Gallery][azure-community-gallery], set `name` field to gallery public name and don't set `subscriptionID` and `resourceGroup` fields:
```
jsturtevant, line 177:
```suggestion
To use an image from [Azure Community Gallery][azure-community-gallery], set `name` field to gallery's public name and don't set `subscriptionID` and `resourceGroup` fields:
```

jsturtevant (commented):
Thanks for the updates to the docs makes a lot more sense now.  lgtm minus a few small comments

api/v1beta1/azureimage_validation.go
invidian: Done.

agrawalkshitij (commented):
this sounds badly names, plan is different than pub/off/sku:
https://docs.microsoft.com/en-us/rest/api/compute/virtual-machines/create-or-update#plan

is this on prupose or outside of ur control to name this correctly?

docs/book/src/topics/custom-images.md
```diff
@@ -63,9 +63,9 @@ See [Upgrading workload clusters][upgrading-workload-clusters] for more details.
 
 To use a custom image, it needs to be referenced in an `image:` section of your `AzureMachineTemplate`. See below for more specific examples.
 
-### Using Shared Image Gallery (Recommended)
+### Using Compute Gallery (Recommended)
 
-To use an image from the [Shared Image Gallery][shared-image-gallery], fill in the `resourceGroup`, `name`, `subscriptionID`, `gallery`, and `version` fields:
+To use an image from the [Compute Gallery][compute-gallery], previously known as Shared Image Gallery (SIG), fill in the `resourceGroup`, `name`, `subscriptionID`, `gallery`, and `version` fields:
```
agrawalkshitij, line 68: the official name has "Azure" in it : Azure Compute Gallery

docs/book/src/topics/custom-images.md
```diff
@@ -171,6 +172,53 @@ spec:
...
+apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
+kind: AzureMachineTemplate
+metadata:
+  name: capz-community-gallery-example
+spec:
+  template:
+    spec:
+      image:
+        computeGallery:
+          gallery: testGallery-3282f15c-906a-4c4b-b206-eb3c51adb5be
```
agrawalkshitij, line 189: won't this be confusing for people? sometimes the gallery is uniquename, while other times it is not the unique name, but rather needs subid\rgname. Does it make sense to add another field like "gallery-unique-name" ?

azure/converters/image.go
```diff
@@ -49,16 +49,40 @@ func mpImageToSDK(image *infrav1.Image) (*compute.ImageReference, error) {
...
 
-func sigImageToSDK(image *infrav1.Image) (*compute.ImageReference, error) {
-	imageID := fmt.Sprintf("/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Compute/galleries/%s/images/%s/versions/%s",
-		image.SharedGallery.SubscriptionID,
-		image.SharedGallery.ResourceGroup,
-		image.SharedGallery.Gallery,
-		image.SharedGallery.Name,
-		image.SharedGallery.Version)
+func computeImageToSDK(image *infrav1.Image) (*compute.ImageReference, error) {
+	idTemplate := "/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Compute/galleries/%s/images/%s/versions/%s"
```
agrawalkshitij, line 53: the versions part is optional, even though it supports putting /versions/latest, the recommded way is to skip the versions part and just put the image-definition-id for using the "latest"

azure/converters/image_test.go
```diff
@@ -31,6 +31,25 @@ func Test_ImageToPlan(t *testing.T) {
...
+						Offer:     "my-offer",
+						SKU:       "my-sku",
+					},
+				},
+			},
+			expect: func(g *GomegaWithT, result *compute.Plan) {
+				g.Expect(result).To(Equal(&compute.Plan{
+					Name:      to.StringPtr("my-sku"),
+					Publisher: to.StringPtr("my-publisher"),
+					Product:   to.StringPtr("my-offer"),
```
agrawalkshitij, line 49: not sure if this is correct, productname and offer name are not the same.
https://docs.microsoft.com/en-us/rest/api/compute/virtual-machines/create-or-update#plan

azure/converters/image.go
invidian: `Version` is right now a _required_ parameter on CAPZ API, hence the formatting. I think in CAPZ case it is beneficial to always require an image version from the user instead of doing that implicitly.

But, given that we introduce a new block for compute galleries, we have a chance to reconsider it. I'll leave that up to maintainers.

azure/converters/image_test.go
invidian: This is just a copy-paste from other test, I think this is the local convention, even if it's not a 1:1 mapping with Azure API.

docs/book/src/topics/custom-images.md
invidian: I feel like `Azure` might be a redundant word here, given that it's the Azure provider in general.

docs/book/src/topics/custom-images.md
invidian: From the consumer perspective, it feels to me that it's _just a gallery name_. It should not matter whether it's a private name, resource name or unique name.

azure/converters/image_test.go
CecileRobertMichon: @agrawalkshitij https://docs.microsoft.com/en-us/rest/api/compute/virtual-machines/create-or-update#plan states explicitly states "This is the same value as Offer under the imageReference element", are we misunderstanding?

agrawalkshitij: Is Name also supposed to be Sku name? I don't think there is a 1:1 mapping like this from pub/off/sku to publisher/product/Name .. I can double check with my team.
I do see the doc mentions about product == offer, but not the name.

agrawalkshitij: since this test code, i guess it's fine.

azure/converters/image.go
agrawalkshitij: sounds good

invidian: I consider it resolved then.

docs/book/src/topics/custom-images.md
jsturtevant: I am not sure we have it clearly defined in how we write docs but sometime it's nice to use something like `<gallery-name>` , `<name>`, `<version>` so the user doesn't copy paste these values.

exp/api/v1beta1/azuremachinepool_webhook_test.go
```diff
@@ -260,13 +260,15 @@ func TestAzureMachinePool_Default(t *testing.T) {
...
 
-func createMachinePoolWithtMarketPlaceImage(publisher, offer, sku, version string, terminateNotificationTimeout *int) *AzureMachinePool {
+func createMachinePoolWithMarketPlaceImage(publisher, offer, sku, version string, terminateNotificationTimeout *int) *AzureMachinePool {
 	image := infrav1.Image{
 		Marketplace: &infrav1.AzureMarketplaceImage{
-			Publisher: publisher,
-			Offer:     offer,
-			SKU:       sku,
-			Version:   version,
+			ImagePlan: infrav1.ImagePlan{
```
jsturtevant, line 266: just to confirm,  is this a breaking change to the `AzureMarketPlaceImage` CRD?

docs/book/src/topics/custom-images.md
jsturtevant: either is probably fine here, `Azure Compute Gallery` is probably more technically correct

jsturtevant (commented):
this lgtm, pending one comment about the `AzureMarketPlaceImage` CRD.

exp/api/v1beta1/azuremachinepool_webhook_test.go
invidian: From programmatic perspective yes, from YAML it should remain the same as far as I'm aware, as the fields are inlined.

docs/book/src/topics/custom-images.md
CecileRobertMichon: +1 to using the official product name in docs, "Azure Compute Gallery" even if it's redundant

docs/book/src/topics/custom-images.md
CecileRobertMichon: +1 to using `<gallery-name>` etc to avoid copy paste confusion

CecileRobertMichon:
lgtm pending 2 minor docs comments

docs/book/src/topics/custom-images.md
CecileRobertMichon: SGTM

invidian: Actually, no other code example use placeholders, so I'll leave it as it is.

docs/book/src/topics/custom-images.md
invidian: Done. We now consistently use either `Azure Compute Gallery` or `private compute gallery`, which seem correct.

invidian:
All conversations are resolved. Thanks for reviewing it again :)

PR #2300, opened 2022-05-13
Support using a customized template outside CAPZ repo

### LINKED ISSUE #2299: Support customized templates outside CAPZ repo in ci-entrypoint.sh
/kind feature

`ci-entrypoint.sh` should support that a user can choose a customized template outside CAPZ repo by setting an environment variable to a URL. In this way, CAPZ can be better used to create different kinds of CI clusters.

Now we have `CLUSTER_TEMPLATE` but it seems to be limited to templates in this repo.

**Environment:**

lzhecheng:
A customized template can be one on the host or a link.

**What type of PR is this?**
/kind feature

Support using a customized template outside CAPZ repo
A customized template can be one on the host or a link.

Fixes #2299

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Support using a customized template outside CAPZ repo. A customized template can also be one on the host or a link.
```

### DIFF
--- a/docs/book/src/developers/development.md
+++ b/docs/book/src/developers/development.md
@@ -558,7 +558,7 @@ You can optionally set the following variables:
 | `EXP_MACHINE_POOL`          | Use [Machine Pool](../topics/machinepools.md) for worker machines.                                                                            |
 | `TEST_WINDOWS`                  | Build a cluster that has Windows worker nodes.          |
 | `REGISTRY`                  | Registry to push any custom k8s images or cloud provider images built.                                                                     |
-| `CLUSTER_TEMPLATE`          | Use a custom cluster template. By default, the script will choose the appropriate cluster template based on existing environment variables. |
+| `CLUSTER_TEMPLATE`          | Use a custom cluster template. It can be a path to a template under templates/, a path on the host or a link. If the value is not set, the script will choose the appropriate cluster template based on existing environment variables. |
 
 You can also customize the configuration of the CAPZ cluster (assuming that `SKIP_CREATE_WORKLOAD_CLUSTER` is not set). See [Customizing the cluster deployment](#customizing-the-cluster-deployment) for more details.
 
--- a/Makefile
+++ b/Makefile
@@ -287,7 +287,13 @@ create-management-cluster: $(KUSTOMIZE) $(ENVSUBST) ## Create a management clust
 .PHONY: create-workload-cluster
 create-workload-cluster: $(ENVSUBST) ## Create a workload cluster.
 	# Create workload Cluster.
-	$(ENVSUBST) < $(TEMPLATES_DIR)/$(CLUSTER_TEMPLATE) | kubectl apply -f -
+	@if [ -f "$(TEMPLATES_DIR)/$(CLUSTER_TEMPLATE)" ]; then \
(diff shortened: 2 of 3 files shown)


docs/book/src/developers/development.md
```diff
@@ -559,6 +559,7 @@ You can optionally set the following variables:
 | `TEST_WINDOWS`                  | Build a cluster that has Windows worker nodes.          |
 | `REGISTRY`                  | Registry to push any custom k8s images or cloud provider images built.                                                                     |
 | `CLUSTER_TEMPLATE`          | Use a custom cluster template. By default, the script will choose the appropriate cluster template based on existing environment variables. |
+| `CLUSTER_TEMPLATE_LINK` | Use a custom cluster template link outside of CAPZ repo. |
```
CecileRobertMichon, line 562: what do you think about reusing the `CLUSTER_TEMPLATE` variable instead of adding a new variable? 

That way we can just curl the template in the makefile and apply it directly if the template isn't a local file:

```
  if [ -f $(TEMPLATES_DIR)/$(CLUSTER_TEMPLATE) ]; then \
      $(ENVSUBST) < "$(TEMPLATES_DIR)/$(CLUSTER_TEMPLATE)" | kubectl apply -f -; \
else curl --retry $(CURL_RETRIES) $(CLUSTER_TEMPLATE) | $(ENVSUBST) | kubectl apply -f; - \
  fi
```

We could even add the option for a valid local file path that's not in the templates dir if we wanted.

CecileRobertMichon:
Thanks for the PR @lzhecheng

docs/book/src/developers/development.md
jsturtevant: > We could even add the option for a valid local file path that's not in the templates dir if we wanted.

I started to look at this for some e2e tests for sig-windows gmsa and this would helpful.  Right now we can't overload the `TEMPLATES_DIR=` https://github.dev/kubernetes-sigs/cluster-api-provider-azure/blob/c08e5e631bdb1831bcaf698525f3c423e48a706c/Makefile#L45-L46

so are limited to templates in the template folder here.

lzhecheng: Makes sense. I have updated the code.

lzhecheng:
@CecileRobertMichon @jsturtevant besides the change for more flexible `CLUSTER_TEMPLATE`, I also move the following logic out of `ci-entrypoint.sh:select_cluster_template()` 1) make ccm/cnm images; 2) export Windows variables. The reason is that if `CLUSTER_TEMPLATE`, `TEST_CCM` and `TEST_WINDOWS` are all set by users, images won't be built and win vars won't be exported with original code. I think it is a bug needing fix, right?

mboersma (approved):
/lgtm

I think your reasoning about moving the code makes sense, but I'll defer to other reviewers.

jsturtevant:
> 1) make ccm/cnm images; 2) export Windows variables. The reason is that if `CLUSTER_TEMPLATE`, `TEST_CCM` and `TEST_WINDOWS` are all set by users, images won't be built and win vars won't be exported with original code. I think it is a bug needing fix, right?

That looks like the correct fix.  

/lgtm

CecileRobertMichon (commented):
/lgtm
/approve

PR #2303, opened 2022-05-13
Add clusterv1.ReadyCondition to AzureMachinePool and AzureMachinePoolMachine

Jont828:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**: Add clusterv1.ReadyCondition to AzureMachinePool and AzureMachinePoolMachine so it can work properly in `clusterctl describe`. 

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Add clusterv1.ReadyCondition to AzureMachinePool and AzureMachinePoolMachine
```

### DIFF
--- a/azure/scope/machinepool.go
+++ b/azure/scope/machinepool.go
@@ -487,15 +487,25 @@ func (m *MachinePoolScope) SetAnnotation(key, value string) {
 	m.AzureMachinePool.Annotations[key] = value
 }
 
-// PatchObject persists the machine spec and status.
+// PatchObject persists the AzureMachinePool spec and status.
 func (m *MachinePoolScope) PatchObject(ctx context.Context) error {
 	ctx, _, done := tele.StartSpanWithLogger(ctx, "scope.MachinePoolScope.PatchObject")
 	defer done()
 
-	return m.patchHelper.Patch(ctx, m.AzureMachinePool)
+	conditions.SetSummary(m.AzureMachinePool)
+	return m.patchHelper.Patch(
+		ctx,
+		m.AzureMachinePool,
+		patch.WithOwnedConditions{Conditions: []clusterv1.ConditionType{
+			clusterv1.ReadyCondition,
+			infrav1.BootstrapSucceededCondition,
+			infrav1.ScaleSetDesiredReplicasCondition,
+			infrav1.ScaleSetModelUpdatedCondition,
+			infrav1.ScaleSetRunningCondition,
+		}})
 }
 
-// Close the MachineScope by updating the machine spec, machine status.
+// Close the MachinePoolScope by updating the AzureMachinePool spec and AzureMachinePool status.
 func (m *MachinePoolScope) Close(ctx context.Context) error {
 	ctx, log, done := tele.StartSpanWithLogger(ctx, "scope.MachinePoolScope.Close")
 	defer done()
@@ -512,17 +522,17 @@ func (m *MachinePoolScope) Close(ctx context.Context) error {
 		}
 	}
 
-	return m.patchHelper.Patch(ctx, m.AzureMachinePool)
+	return m.PatchObject(ctx)
 }
 
-// GetBootstrapData returns the bootstrap data from the secret in the Machine's bootstrap.dataSecretName.
+// GetBootstrapData returns the bootstrap data from the secret in the MachinePool's bootstrap.dataSecretName.
 func (m *MachinePoolScope) GetBootstrapData(ctx context.Context) (string, error) {
 	ctx, _, done := tele.StartSpanWithLogger(ctx, "scope.MachinePoolScope.GetBootstrapData")
 	defer done()
 
 	dataSecretName := m.MachinePool.Spec.Template.Spec.Bootstrap.DataSecretName
 	if dataSecretName == nil {
-		return "", errors.New("error retrieving bootstrap data: linked Machine Spec's bootstrap.dataSecretName is nil")
+		return "", errors.New("error retrieving bootstrap data: linked MachinePool Spec's bootstrap.dataSecretName is nil")
 	}
 	secret := &corev1.Secret{}
 	key := types.NamespacedName{Namespace: m.AzureMachinePool.Namespace, Name: *dataSecretName}
@@ -537,7 +547,7 @@ func (m *MachinePoolScope) GetBootstrapData(ctx context.Context) (string, error)
 	return base64.StdEncoding.EncodeToString(value), nil
 }
 
-// GetVMImage picks an image from the machine configuration, or uses a default one.
+// GetVMImage picks an image from the AzureMachinePool configuration, or uses a default one.
 func (m *MachinePoolScope) GetVMImage(ctx context.Context) (*infrav1.Image, error) {
 	_, log, done := tele.StartSpanWithLogger(ctx, "scope.MachinePoolScope.GetVMImage")
 	defer done()
@@ -601,7 +611,7 @@ func (m *MachinePoolScope) HasSystemAssignedIdentity() bool {
 	return m.AzureMachinePool.Spec.Identity == infrav1.VMIdentitySystemAssigned
 }
 
-// VMSSExtensionSpecs returns the vmss extension specs.
+// VMSSExtensionSpecs returns the VMSS extension specs.
 func (m *MachinePoolScope) VMSSExtensionSpecs() []azure.ResourceSpecGetter {
 	var extensionSpecs = []azure.ResourceSpecGetter{}
 	bootstrapExtensionSpec := azure.GetBootstrappingVMExtension(m.AzureMachinePool.Spec.Template.OSDisk.OSType, m.CloudEnvironment(), m.Name())
--- a/azure/scope/machinepoolmachine.go
+++ b/azure/scope/machinepoolmachine.go
@@ -236,17 +236,30 @@ func (s *MachinePoolMachineScope) IsReady() bool {
 
 // SetFailureMessage sets the AzureMachinePoolMachine status failure message.
 func (s *MachinePoolMachineScope) SetFailureMessage(v error) {
-	s.AzureMachinePool.Status.FailureMessage = pointer.StringPtr(v.Error())
+	s.AzureMachinePoolMachine.Status.FailureMessage = pointer.StringPtr(v.Error())
 }
 
 // SetFailureReason sets the AzureMachinePoolMachine status failure reason.
 func (s *MachinePoolMachineScope) SetFailureReason(v capierrors.MachineStatusError) {
-	s.AzureMachinePool.Status.FailureReason = &v
+	s.AzureMachinePoolMachine.Status.FailureReason = &v
 }
 
 // ProviderID returns the AzureMachinePool ID by parsing Spec.FakeProviderID.
 func (s *MachinePoolMachineScope) ProviderID() string {
 	return s.AzureMachinePoolMachine.Spec.ProviderID
+}
+
+// PatchObject persists the MachinePoolMachine spec and status.
+func (s *MachinePoolMachineScope) PatchObject(ctx context.Context) error {
+	conditions.SetSummary(s.AzureMachinePoolMachine)
+
+	return s.patchHelper.Patch(
+		ctx,
+		s.AzureMachinePoolMachine,
+		patch.WithOwnedConditions{Conditions: []clusterv1.ConditionType{
+			clusterv1.ReadyCondition,
+			clusterv1.MachineNodeHealthyCondition,
+		}})
 }
 
 // Close updates the state of MachinePoolMachine.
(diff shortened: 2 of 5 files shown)


Jont828:
/assign @CecileRobertMichon @mboersma 

Would appreciate it if you two could take a look!

azure/scope/machinepool.go
```diff
@@ -486,12 +486,19 @@ func (m *MachinePoolScope) SetAnnotation(key, value string) {
 	m.AzureMachinePool.Annotations[key] = value
 }
 
-// PatchObject persists the machine spec and status.
+// PatchObject persists the machine (pool?) spec and status.
```
devigned, line 489:
```suggestion
// PatchObject persists the Azure machine pool spec and status.
```

azure/scope/machinepoolmachine.go
```diff
@@ -249,6 +249,20 @@ func (s *MachinePoolMachineScope) ProviderID() string {
 	return s.AzureMachinePoolMachine.Spec.ProviderID
 }
 
+// TODO: does this need to be manually called in AMPM controller?
```
devigned, line 252: > does this need to be manually called in AMPM controller?

I'm not sure what "manually" implies, but yes, this func will need to be invoked somewhere. Perhaps, add a `defer scope.Close()` in the controller and add a call to `PatchObject` in the `Close` func.

devigned (changes requested):
+1 to better conditions / status reporting

azure/scope/machinepoolmachine.go
Jont828: Yep that makes sense. I just wanted to double check that we didn't need to call PatchObject() at the beginning.

azure/scope/machinepoolmachine.go
```diff
@@ -293,18 +307,27 @@ func (s *MachinePoolMachineScope) UpdateStatus(ctx context.Context) error {
 			APIVersion: node.APIVersion,
 		}
 
-		s.AzureMachinePoolMachine.Status.Ready = noderefutil.IsNodeReady(node)
+		nodeReady := noderefutil.IsNodeReady(node)
+		s.AzureMachinePoolMachine.Status.Ready = nodeReady
+		if nodeReady {
+			conditions.MarkTrue(s.AzureMachinePoolMachine, clusterv1.MachineNodeHealthyCondition)
+		} else {
+			conditions.MarkFalse(s.AzureMachinePoolMachine, clusterv1.MachineNodeHealthyCondition, clusterv1.NodeProvisioningReason, clusterv1.ConditionSeverityInfo, "")
```
Jont828, line 315: @CecileRobertMichon @devigned @mboersma Could use some feedback on which condition false reason we should be using here. Here are the reasons available:
```
// WaitingForNodeRefReason (Severity=Info) documents a machine.spec.providerId is not assigned yet.
WaitingForNodeRefReason = "WaitingForNodeRef"

// NodeProvisioningReason (Severity=Info) documents machine in the process of provisioning a node.
// NB. provisioning --> NodeRef == "".
NodeProvisioningReason = "NodeProvisioning"

// NodeNotFoundReason (Severity=Error) documents a machine's node has previously been observed but is now gone.
// NB. provisioned --> NodeRef != "".
NodeNotFoundReason = "NodeNotFound"

// NodeConditionsFailedReason (Severity=Warning) documents a node is not in a healthy state due to the failed state of at least 1 Kubelet condition.
NodeConditionsFailedReason = "NodeConditionsFailed"
```

azure/scope/machinepoolmachine.go
```diff
@@ -281,6 +294,7 @@ func (s *MachinePoolMachineScope) UpdateStatus(ctx context.Context) error {
 	}
 
 	if err != nil && !apierrors.IsNotFound(err) {
+		conditions.MarkFalse(s.AzureMachinePoolMachine, clusterv1.MachineNodeHealthyCondition, clusterv1.NodeNotFoundReason, clusterv1.ConditionSeverityError, "")
```
Jont828, line 297: @CecileRobertMichon @devigned @mboersma Same with here

azure/scope/machinepool.go
```diff
@@ -514,14 +525,14 @@ func (m *MachinePoolScope) Close(ctx context.Context) error {
 	return m.patchHelper.Patch(ctx, m.AzureMachinePool)
```
CecileRobertMichon, line 525: this should call `PatchObject` instead of Patching directly

azure/scope/machinepool.go
```diff
@@ -486,15 +486,26 @@ func (m *MachinePoolScope) SetAnnotation(key, value string) {
...
+	return m.patchHelper.Patch(
+		ctx,
+		m.AzureMachinePool,
+		patch.WithOwnedConditions{Conditions: []clusterv1.ConditionType{
+			clusterv1.ReadyCondition,
+			infrav1.BootstrapSucceededCondition,
+			infrav1.ScaleSetDesiredReplicasCondition,
+			infrav1.ScaleSetModelUpdatedCondition,
+			infrav1.ScaleSetRunningCondition,
+			// TODO: Add a role assignment condition in a later PR
```
CecileRobertMichon, line 504: Can you please file an issue for this instead of leaving the TODO in code? code TODOs tend to get forgotten...

azure/scope/machinepool.go
```diff
@@ -486,15 +486,26 @@ func (m *MachinePoolScope) SetAnnotation(key, value string) {
...
+		}})
+}
+
+// Close the MachinePoolScope by updating the AzureMachinePool spec, AzureMachinePool status.
```
CecileRobertMichon, line 508: we shouldn't be updating the spec in here, mainly the status?

azure/scope/machinepoolmachine.go
CecileRobertMichon: This is if `!apierrors.IsNotFound(err)`, so it's _not_ a NodeNotFound reason. It's that we failed to find the node because one of those func returned an error. Maybe we need to add an else statement if the node is actually not found, then we can mark the condition as NotFound. In this case, I think `WaitingForNodeRefReason` makes the most sense.

azure/scope/machinepoolmachine.go
CecileRobertMichon: I believe in this case it would be `NodeConditionsFailed`. `NodeProvisioningReason` would be above if the NodeRef is nil and we can't get the node yet.

azure/scope/machinepoolmachine.go
CecileRobertMichon: I think we need to split conditions for the two scenarios above:

`if nodeRef == nil || nodeRef.Name == "" `: Node ref doesn't exist yet, node could still be provisioning - `NodeProvisioningReason`
`else`:  node ref is set, but node is not found: `NodeNotFoundReason`

azure/scope/machinepool.go
Jont828: It looks like we are updating the spec by setting the provider IDs. There's a later line that calls `m.updateReplicasAndProviderIDs(ctx)`.

azure/scope/machinepoolmachine.go
```diff
@@ -274,14 +287,21 @@ func (s *MachinePoolMachineScope) UpdateStatus(ctx context.Context) error {
 		node    *corev1.Node
 		err     error
 	)
-	if nodeRef == nil || nodeRef.Name == "" {
+
+	if s.ProviderID() == "" {
```
Jont828, line 291: @CecileRobertMichon I made some changes to the if logic based on your comment. Is this what you were describing?

CecileRobertMichon: yep looks good

azure/scope/machinepoolmachine.go
```diff
@@ -274,14 +287,23 @@ func (s *MachinePoolMachineScope) UpdateStatus(ctx context.Context) error {
...
-	if nodeRef == nil || nodeRef.Name == "" {
+
+	switch {
+	case s.ProviderID() == "":
+		conditions.MarkFalse(s.AzureMachinePoolMachine, clusterv1.MachineNodeHealthyCondition, clusterv1.WaitingForNodeRefReason, clusterv1.ConditionSeverityInfo, "")
+	case nodeRef == nil || nodeRef.Name == "":
 		node, err = s.workloadNodeGetter.GetNodeByProviderID(ctx, s.ProviderID())
-	} else {
+		if err != nil && !apierrors.IsNotFound(err) {
+			// TODO: what should we set the condition to when node == nil and err == nil?
```
Jont828, line 297: @CecileRobertMichon @devigned Right now we call `conditions.MarkFalse(s.AzureMachinePoolMachine, clusterv1.MachineNodeHealthyCondition, clusterv1.NodeProvisioningReason, clusterv1.ConditionSeverityInfo, "")` when `err != nil && !apierrors.IsNotFound(err)`, but we don't set anything when `node == nil` and `err == nil`. Which condition should we mark in the latter case? I'm not sure what `node == nil` and `err == nil` is meant to represent.

devigned: If `node == nil`, then the node associated to the machine pool machine was not found, likely it has not joined the cluster yet. 

`err == nil` signifies there was no error when communicating with the API server.

The gist of what is going on here is that we are attempting to determine the state of the infrastructure and the health / presence of the corresponding node in the cluster.

azure/scope/machinepoolmachine.go
```diff
@@ -562,6 +591,7 @@ func getNodeByProviderID(ctx context.Context, workloadClient client.Client, prov
 		}
 	}
 
+	// TODO: should this be a not found err if it's not in the list?
```
devigned, line 594: In this case, I think `nil` response with `nil` error is effective in saying something is not in the list. An error might be confused as a 404 from API server.

azure/scope/machinepoolmachine.go
```diff
@@ -257,7 +270,7 @@ func (s *MachinePoolMachineScope) Close(ctx context.Context) error {
 	)
 	defer done()
 
-	return s.patchHelper.Patch(ctx, s.AzureMachinePoolMachine)
+	return s.PatchObject(ctx)
 }
 
 // UpdateStatus updates the node reference for the machine and other status fields. This func should be called at the
```
devigned, line 276:
```go
ctx, _, done := tele.StartSpanWithLogger(
		ctx,
		"scope.MachinePoolMachineScope.Get",
	)
```

Should change "scope.MachinePoolMachineScope.Get" to "scope.MachinePoolMachineScope.UpdateStatus"

azure/scope/machinepoolmachine.go
Jont828: Okay thanks for the info. So in this case I think if `node == nil` then we can set it to `clusterv1.NodeProvisioningReason`. In the latter case where `err != nil`, would we want to set a condition reason here? We don't want the condition to be ready, but I'm not sure what we can say about the node based on that info.

devigned: I don't know what reason to put there either. If we have a failure interacting with the API server, we should exit with error and require the reconcile (default behavior).

The reason could be unknown, perhaps.

devigned: As an aside, I never liked this func. There's too much going on, too many conditionals. I didn't put the time into refactoring. However, if you have it all queued up in your head, it might be a good opportunity to restructure a bit to make it easier on the next person.

Jont828: Yeah I can look into it and see if we can refactor a bit. In general, if we have a failure with the API server, would we want to mark the condition at all?

azure/scope/machinepoolmachine_test.go
```diff
@@ -232,41 +214,38 @@ func TestMachineScope_UpdateStatus(t *testing.T) {
...
-					LatestModelApplied: true,
+				g.Expect(scope.AzureMachinePoolMachine.Status.Ready).To(Equal(true))
+				g.Expect(scope.AzureMachinePoolMachine.Status.Version).To(Equal("1.2.3"))
+				g.Expect(scope.AzureMachinePoolMachine.Status.NodeRef).To(Equal(&corev1.ObjectReference{
+					Name: "node1",
 				}))
+				assertCondition(t, scope.AzureMachinePoolMachine, conditions.TrueCondition(clusterv1.MachineNodeHealthyCondition))
 			},
 		},
+		// {
```
devigned, line 225: Probably could delete this code.

devigned (changes requested):
lgtm other than the commented code block.

Jont828:
@devigned @mboersma Squashed changes and removed commented out code!

azure/scope/machinepoolmachine.go
```diff
@@ -52,6 +52,7 @@ const (
 type (
 	nodeGetter interface {
 		GetNodeByProviderID(ctx context.Context, providerID string) (*corev1.Node, error)
+		// GetNodeByProviderID(ctx context.Context, providerID string) (*corev1.Node, bool, error)
```
CecileRobertMichon, line 55: why is this commented out?

azure/scope/machinepoolmachine.go
```diff
@@ -235,11 +236,13 @@ func (s *MachinePoolMachineScope) IsReady() bool {
 }
 
 // SetFailureMessage sets the AzureMachinePoolMachine status failure message.
+// TODO: should this be AzureMachinePoolMachine or AzureMachinePool?
```
CecileRobertMichon, line 239: seems to me like it should be AzureMachinePoolMachine

azure/scope/machinepoolmachine.go
Jont828: Oops, forgot to remove that!

azure/scope/machinepoolmachine.go
CecileRobertMichon: Sorry should have been more clear, I was saying the opposite: it should be setting the status of the AzureMachinePoolMachine, not the AzureMachinePool. I agree with you that the comment doesn't match the function, I'm saying the comment is right but the code is wrong (because this is in the machinepoolmachine.go file). @devigned might be able to confirm.

devigned: I'm thinking this is probably copypasta that just happened to work since the scope contains a field that the MachinePoolScop contains.

mboersma (approved):
/lgtm
/approve

PR #2306, opened 2022-05-14
Simplify use of tar in scripts for compatibility

### LINKED ISSUE #2255: Fetch scripts for observability Helm charts don't work on Linux
/kind bug

**What steps did you take and what happened:** Running the scripts for fetching observability Helm charts results in a failure on Linux. For example, this script fails due to tar.

```
$ ./fetch-otel-resources.sh
tar: /^opentelemetry-collector/chart: Not found in archive
tar: -: Not found in archive
tar: Exiting with failure status due to previous errors
```

**Anything else you would like to add:** The issue is that the `tar` command in the scripts is the BSD version of tar while Linux uses GNU's version. If we want to keep the script the same, we could include a copy of BSD tar in `hack/tools/bin`.

**Environment:**

mboersma:
**What type of PR is this?**

/kind bug

In two resource update scripts, we use the `-s <pattern>` argument to `tar`. This is not supported on all versions of `tar`, but it can be replaced with a manual renaming of the target directory.

Fixes #2255

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/observability/jaeger/fetch-jaeger-resources.sh
+++ b/hack/observability/jaeger/fetch-jaeger-resources.sh
@@ -22,6 +22,8 @@ CHART_RELEASE=${CHART_RELEASE:-0.1.5}
 JAEGER_ROOT=$(dirname "${BASH_SOURCE[0]}")
 CHART_ROOT=$JAEGER_ROOT/chart
 
-rm -rf "${CHART_ROOT:?}/"*
+rm -rf "$CHART_ROOT"
+# "tar" has no POSIX standard, so use only basic options and test with both BSD and GNU.
 wget -qO- https://github.com/hansehe/jaeger-all-in-one/raw/master/helm/charts/jaeger-all-in-one-"$CHART_RELEASE".tgz \
-    | tar xvz -C "$JAEGER_ROOT" -s /^jaeger-all-in-one/chart/ -
+    | tar xvz -C "$JAEGER_ROOT"
+mv "$JAEGER_ROOT"/jaeger-all-in-one "$CHART_ROOT"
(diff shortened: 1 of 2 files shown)


hack/observability/jaeger/fetch-jaeger-resources.sh
```diff
@@ -22,6 +22,7 @@ CHART_RELEASE=${CHART_RELEASE:-0.1.5}
 JAEGER_ROOT=$(dirname "${BASH_SOURCE[0]}")
 CHART_ROOT=$JAEGER_ROOT/chart
 
-rm -rf "${CHART_ROOT:?}/"*
+rm -rf "$CHART_ROOT"
 wget -qO- https://github.com/hansehe/jaeger-all-in-one/raw/master/helm/charts/jaeger-all-in-one-"$CHART_RELEASE".tgz \
-    | tar xvz -C "$JAEGER_ROOT" -s /^jaeger-all-in-one/chart/ -
+    | tar xvz -C "$JAEGER_ROOT" -
```
Jont828, line 27: I ran this on Linux and I'm getting this error:
```
tar: -: Not found in archive
tar: Exiting with failure status due to previous errors
```

It looks like my tar command defaults to:
``` 
*This* tar defaults to:
--format=gnu -f- -b20 --quoting-style=escape --rmt-command=/usr/lib/tar/rmt
```
Do you have a Linux setup by any chance to verify this error?

mboersma: Turns out that using dash (`-`) to represent the stdin file handle isn't actually supported by GNU `tar`. But apparently it's not necessary either, so removing it seems to make this work on both versions of `tar` in my testing.

Thanks @Jont828, LMK if it works for you!

Jont828: Seems like it's working on my end!

Jont828:
Thanks for making this change! Could we add a comment about how `tar` works differently depending on if it's GNU or bsdtar for future reference?

Jont828:
Other than that LGTM!

/lgtm

PR #2316, opened 2022-05-19
ci: support external-cloud-provider plus latest k8s releases

jackfrancis:
**What type of PR is this?**

/kind failing-test

This PR adds a new `external-cloud-provider-ci-version` prow template to accommodate building out-of-tree cloud-provider-azure clusters running the latest release bits of Kubernetes.

Changes to `ci-entrypoint.sh` should be evaluated for back-compat with existing usage.

After this is merged, we want to cherry-pick into release-1.3 ASAP.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/scripts/ci-entrypoint.sh
+++ b/scripts/ci-entrypoint.sh
@@ -85,12 +85,16 @@ select_cluster_template() {
         # shellcheck source=scripts/ci-build-kubernetes.sh
         source "${REPO_ROOT}/scripts/ci-build-kubernetes.sh"
         export CLUSTER_TEMPLATE="test/dev/cluster-template-custom-builds.yaml"
-    elif [[ -n "${CI_VERSION:-}" ]] || [[ -n "${USE_CI_ARTIFACTS:-}" ]]; then
+    elif [[ -n "${CI_VERSION:-}" ]] || [[ -n "${USE_CI_ARTIFACTS:-}" ]] || [[ "${KUBERNETES_VERSION:-}" =~ "latest" ]]; then
         # export cluster template which contains the manifests needed for creating the Azure cluster to run the tests
         GOPATH="$(go env GOPATH)"
-        KUBERNETES_BRANCH="$(cd "${GOPATH}/src/k8s.io/kubernetes" && git rev-parse --abbrev-ref HEAD)"
+        if ls "${GOPATH}"/src/k8s.io/kubernetes; then
+            KUBERNETES_BRANCH="$(cd "${GOPATH}/src/k8s.io/kubernetes" && git rev-parse --abbrev-ref HEAD)"
+        fi
         if [[ "${KUBERNETES_BRANCH:-}" =~ "release-" ]]; then
             CI_VERSION_URL="https://dl.k8s.io/ci/latest-${KUBERNETES_BRANCH/release-}.txt"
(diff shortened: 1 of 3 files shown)


scripts/ci-entrypoint.sh
```diff
@@ -30,6 +30,7 @@ cd "${REPO_ROOT}" && make "${KUBECTL##*/}"; make "${HELM##*/}"
 # export the variables so they are available in bash -c wait_for_nodes below
 export KUBECTL
 export HELM
+export KUBERNETES_VERSION="${KUBERNETES_VERSION:-}"
```
CecileRobertMichon, line 33: do we need to do this or can we just replace `${KUBERNETES_VERSION}` with `${KUBERNETES_VERSION:-}` below?

jackfrancis: This is probably more resilient. I think it's correct that as long as we refer to `${KUBERNETES_VERSION:-}` the first time we use it we are O.K. (no unbound variable error), but the next maintainer of this file that introduces a novel `${KUBERNETES_VERSION}` may not follow that pattern and it may become the _new_ initial reference.

CecileRobertMichon: That's fair. The other variables are all defaulting for every use below but this works too. Let's maybe move it down to where the variable is used though, this section is specific to the binary tools and the comment on line 30 doesn't apply.

jackfrancis: I've gone ahead and does the inline default assignment thing as that's the existing pattern. I think we can trust our CI to complain if/when a future change breaks this.

scripts/ci-entrypoint.sh
```diff
@@ -85,12 +85,16 @@ select_cluster_template() {
...
+    elif [[ -n "${CI_VERSION:-}" ]] || [[ -n "${USE_CI_ARTIFACTS:-}" ]] || [[ "${KUBERNETES_VERSION:-}" =~ "latest" ]]; then
         # export cluster template which contains the manifests needed for creating the Azure cluster to run the tests
         GOPATH="$(go env GOPATH)"
-        KUBERNETES_BRANCH="$(cd "${GOPATH}/src/k8s.io/kubernetes" && git rev-parse --abbrev-ref HEAD)"
+        if ls "${GOPATH}"/src/k8s.io/kubernetes; then
+            KUBERNETES_BRANCH="$(cd "${GOPATH}/src/k8s.io/kubernetes" && git rev-parse --abbrev-ref HEAD)"
+        fi
         if [[ "${KUBERNETES_BRANCH:-}" =~ "release-" ]]; then
             CI_VERSION_URL="https://dl.k8s.io/ci/latest-${KUBERNETES_BRANCH/release-}.txt"
+        elif [[ "${KUBERNETES_VERSION}" =~ "latest" ]]; then
```
CecileRobertMichon, line 96:
```suggestion
        elif [[ "${KUBERNETES_VERSION:-}" =~ "latest" ]]; then
```

need to default here too

CecileRobertMichon:
> After this is merged, we want to cherry-pick into release-1.3 ASAP.

/cherry-pick release-1.3

CecileRobertMichon (commented):
/lgtm

CecileRobertMichon (approved):
/approve

PR #2325, opened 2022-05-23
simplify ci-entrypoint.sh

jackfrancis:
**What type of PR is this?**

/kind cleanup

This PR removes the `USE_CI_ARTIFACTS` option (by convention `USE_CI_ARTIFACTS=true` is supported; any non-zero value to `USE_CI_ARTIFACTS` effectively means boolean true) as the existing `KUBERNETES_VERSION=latest` interface provides the same functionality with more semantic expressiveness.

Additionally, we remove the inferential check to see if k/k is in the GOPATH, and to use the checked out sha as the `KUBERNETES_VERSION`. Going forward we will require usages of `ci-entrypoint.sh` to declare the desired version of Kubernetes to build via the `KUBERNETES_VERSION` environment variable (this is already the practice in the vast majority of test-infra jobs).

This test-infra PR needs to merge first so that we are sure we remove any automated usages of the deprecated config interface!:

https://github.com/kubernetes/test-infra/pull/26397

Finally, we add support for using the "stable" prefix to automatically gather the latest patch release. Using 1.23 as an example, "stable-1.23" will yield "v1.23.7" at the time of this writing. Simply using "stable" yields "v1.24.1" at the time of this writing.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```


docs/book/src/developers/development.md
```diff
@@ -552,8 +552,8 @@ You can optionally set the following variables:
 | `AZURE_SSH_PUBLIC_KEY_FILE` | Use your own SSH key.                                                                                                                      |
 | `SKIP_CLEANUP`              | Skip deleting the cluster after the tests finish running.                                                                                   |
 | `KUBECONFIG`                | Provide your existing cluster kubeconfig filepath. If no kubeconfig is provided, `./kubeconfig` will be used.                                  |
-| `USE_CI_ARTIFACTS`          | Use a CI version of Kubernetes, ie. not a released version (eg. `v1.19.0-alpha.1.426+0926c9c47677e9`)                                      |
-| `CI_VERSION`                | Provide a custom CI version of Kubernetes. By default, the latest master commit will be used.                                              |
+| `KUBERNETES_VERSION`        | Desired Kubernetes version to test. You can pass in a definitive released version, e.g., "v1.24.0"; or use the most recent built, releasable CI bits of a particular release by passing in e.g. "release-1.24"; or you may use the very latest built CI bits by passing in "latest".
```
CecileRobertMichon, line 555: does `release-1.24` actually work? FWIW in conformance test we use "stable-1.24" to indicate the latest released patch version of 1.24

CecileRobertMichon: https://dl.k8s.io/release/stable-1.24.txt does the job, but we'd have to add that into ci-entrypoint

marosset:
I think that specifying `KUBERNETES_VERSION` as anything with 'latest' will not work for Windows.

That would cause this envsubst to container a pre-release tag and therefore not match to any of the kubeproxy images for windows that we currently maintain.
https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/0298a7d2f4ed08b5a79063d471ce9a47e42ea7b3/templates/test/ci/cluster-template-prow-ci-version.yaml#L624

I'll try testing this out manually by checking out from this PR branch and trying to build a cluster.

marosset:
> I think that specifying `KUBERNETES_VERSION` as anything with 'latest' will not work for Windows.
> 
> That would cause this envsubst to container a pre-release tag and therefore not match to any of the kubeproxy images for windows that we currently maintain.
> 
> https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/0298a7d2f4ed08b5a79063d471ce9a47e42ea7b3/templates/test/ci/cluster-template-prow-ci-version.yaml#L624
> 
> I'll try testing this out manually by checking out from this PR branch and trying to build a cluster.

I was mis-informed and it looks like this DOES work for Windows.
It looks like we pull an arbitrary kube-proxy image and then tag it as `sigwindowstools/kube-proxy:${CI_VERSION/+/_}-calico-hostprocess
https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/ecb256236f2f0913e1a3a871eb24f3de3cc630a4/templates/test/ci/prow-ci-version/patches/kubeadm-bootstrap-windows.yaml#L33

docs/book/src/developers/development.md
jackfrancis:
```
$ curl -sSL https://dl.k8s.io/ci/latest-1.24.txt
v1.24.1-rc.0.29+1868a3a2eb9945
```

```
$ curl -sSL https://dl.k8s.io/release/stable-1.24.txt
v1.24.0
```

Do we want to add support to `ci-entrypoint.sh` for the 2nd pattern?

CecileRobertMichon: I think that'd be pretty useful, can be its own PR if you think that's a scope creep though

marosset: https://github.com/kubernetes/release/blob/9e1b9044533a09adeac1a6c4706b665db8cd5cd0/pkg/release/version_test.go#L141-L174

scripts/ci-entrypoint.sh
```diff
@@ -79,28 +79,22 @@ select_cluster_template() {
         # shellcheck source=scripts/ci-build-kubernetes.sh
         source "${REPO_ROOT}/scripts/ci-build-kubernetes.sh"
         export CLUSTER_TEMPLATE="test/dev/cluster-template-custom-builds.yaml"
-    elif [[ -n "${CI_VERSION:-}" ]] || [[ -n "${USE_CI_ARTIFACTS:-}" ]] || [[ "${KUBERNETES_VERSION:-}" =~ "latest" ]]; then
+    elif [[ "${KUBERNETES_VERSION:-}" =~ "latest" ]] || [[ -n "${CI_VERSION:-}" ]]; then
```
lzhecheng, line 82: Not introduced by this PR but could you please fix it? I think I found an issue here.

If `CLUSTER_TEMPLATE` is set, the ci-entrypoint.sh logic leads to `KUBERNETES_VERSION` error like "vlatest".
https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/cloud-provider-azure-master-vmss-capz/1529007312948170752

I think the below logic should be moved out of `select_cluster_template()`
https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/e108534f21de41501a2ec509e9c359ad63bbed7c/scripts/ci-entrypoint.sh#L85-L91
I think `select_cluster_template()` should only choose template with nothing else.

Please correct me if I'm wrong.

lzhecheng: Thank you Cecile and Jack. Now the request of cloud-provider-azure is to test with "latest-x" and I can update templates in its repo to follow the style in CAPZ.
But I think my suggested change is still needed to make it work? The change won't introduce problems for existing CAPZ templates, right? Meanwhile, it will make cloud-provider-azure's templates work.
Please correct me if I'm wrong.

CecileRobertMichon: +1 to move k8s version resolution into its own func, independent of select_cluster_template()

jackfrancis: The current implementation of `ci-entrypoint.sh` is sort of set up to do the following:

1) you may declare a version of kubernetes (including using a "latest" pattern) and we will choose the right template for you

Are we saying we want to introduce the following in addition:

2) you may declare a cluster template and we will choose the right kubernetes version for you

?

lzhecheng: I think it should support that a user specifies 1) a cluster template and 2) a k8s version (including "latest" pattern), then the script 1) use the template and 2) build k8s if version is of "latest" pattern.

jackfrancis: "latest" doesn't build kubernetes, it simply pulls the most recent CI bits built and hosted upstream.

If we want any cluster template to support any arbitrary build of Kubernetes then we would have to incorporate all of the foo in this template into every template:

https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/templates/test/ci/cluster-template-prow-ci-version.yaml

We don't have a good plan for how to do that at this time.

lzhecheng: I think most recent built is ok.

jackfrancis: Thanks for your feedback, I do think this is a problem we want to solve eventually.

docs/book/src/developers/development.md
CecileRobertMichon: in any case, let's remove `"release-1.24"` since it's not a thing and replace it with `"latest-1.24"`

scripts/ci-entrypoint.sh
```diff
@@ -79,28 +79,22 @@ select_cluster_template() {
...
         # export cluster template which contains the manifests needed for creating the Azure cluster to run the tests
-        GOPATH="$(go env GOPATH)"
-        if ls "${GOPATH}"/src/k8s.io/kubernetes; then
-            KUBERNETES_BRANCH="$(cd "${GOPATH}/src/k8s.io/kubernetes" && git rev-parse --abbrev-ref HEAD)"
-        fi
-        if [[ "${KUBERNETES_BRANCH:-}" =~ "release-" ]]; then
-            CI_VERSION_URL="https://dl.k8s.io/ci/latest-${KUBERNETES_BRANCH/release-}.txt"
-        elif [[ "${KUBERNETES_VERSION:-}" =~ "latest" ]]; then
+        export CLUSTER_TEMPLATE="test/ci/cluster-template-prow-ci-version.yaml"
+        if [[ "${KUBERNETES_VERSION:-}" =~ "latest" ]]; then
```
CecileRobertMichon, line 85: we can add another elif here if we want to support `stable` versions and resolve to `CI_VERSION_URL="https://dl.k8s.io/release/${KUBERNETES_VERSION}.txt`

scripts/ci-entrypoint.sh
```diff
@@ -79,28 +79,22 @@ select_cluster_template() {
...
             CI_VERSION_URL="https://dl.k8s.io/ci/${KUBERNETES_VERSION}.txt"
         else
             CI_VERSION_URL="https://dl.k8s.io/ci/latest.txt"
         fi
-        export CLUSTER_TEMPLATE="test/ci/cluster-template-prow-ci-version.yaml"
```
CecileRobertMichon, line 95: why remove this line?

CecileRobertMichon: nvm, it was moved up

jackfrancis:
I've audited tests since https://github.com/kubernetes/test-infra/pull/26397 merged and I don't observe any regressions.

This PR is ready for final review/merge.

/assign @CecileRobertMichon

scripts/ci-entrypoint.sh
CecileRobertMichon: actually for stable- versions we don't need the ci-version template since it will just resolve to the latest released patch version, example https://storage.googleapis.com/kubernetes-release/release/stable-1.23.txt is v1.23.7 right now, so it needs to be outside the if statement that sets the template to cluster-template-prow-ci-version.yaml.

We can do this in a separate PR too.

docs/book/src/developers/development.md
jackfrancis: Done

docs/book/src/developers/development.md
```diff
@@ -552,8 +552,8 @@ You can optionally set the following variables:
...
+| `CI_VERSION`                | Provide a custom CI version of Kubernetes (e.g., `v1.25.0-alpha.0.597+aa49dffc7f24dc`). By default, the latest master commit will be used.                                              |
```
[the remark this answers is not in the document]
jackfrancis: thx!

scripts/ci-entrypoint.sh
jackfrancis: yeah, you're right, see my latest change

I think this is a simple enough addition that is makes sense to incorporate it in here

scripts/ci-entrypoint.sh
```diff
@@ -79,28 +79,25 @@ select_cluster_template() {
...
         export CI_VERSION="${CI_VERSION:-$(curl -sSL ${CI_VERSION_URL})}"
         export KUBERNETES_VERSION="${CI_VERSION}"
     else
         export CLUSTER_TEMPLATE="test/ci/cluster-template-prow.yaml"
+        if [[ "${KUBERNETES_VERSION:-}" =~ "stable" ]]; then
+            CI_VERSION_URL="https://dl.k8s.io/release/${KUBERNETES_VERSION}.txt"
```
CecileRobertMichon, line 95: this needs to be before https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2325/files#diff-2009f7be5363812e22862771f557363795759e67b3ca2e32c50d9fd5f8c5669bR90 since that's where the URL is resolved to get the version... or maybe we just need to add:
```
export KUBERNETES_VERSION="$(curl -sSL ${CI_VERSION_URL})"
```

jackfrancis: O.K. this is a little more complicated than I anticipated. One observation is that we don't currently support both `KUBERNETES_VERSION` and `CI_VERSION` configuration at the same time. I've added some validation and updated docs, and put the overloaded `CI_VERSION` assignment below all this foo. Are we getting closer?

scripts/ci-entrypoint.sh
```diff
@@ -79,28 +83,31 @@ select_cluster_template() {
...
-            CI_VERSION_URL="https://dl.k8s.io/ci/latest.txt"
-        fi
         export CLUSTER_TEMPLATE="test/ci/cluster-template-prow-ci-version.yaml"
-        export CI_VERSION="${CI_VERSION:-$(curl -sSL ${CI_VERSION_URL})}"
-        export KUBERNETES_VERSION="${CI_VERSION}"
+        if [[ -z "${CI_VERSION:-}" ]]; then
+            if [[ "${KUBERNETES_VERSION:-}" =~ "latest" ]]; then
+                CI_VERSION_URL="https://dl.k8s.io/ci/${KUBERNETES_VERSION}.txt"
+            else
+                CI_VERSION_URL="https://dl.k8s.io/ci/latest.txt"
```
CecileRobertMichon, line 93: don't think we need this else here, `CI_VERSION_URL="https://dl.k8s.io/ci/${KUBERNETES_VERSION}.txt"` is equivalent to `CI_VERSION_URL="https://dl.k8s.io/ci/latest.txt"` when `KUBERNETES_VERSION is `latest` and there is no more defaulting to "latest" when `USE_CI_ARTIFACTS` is true.

This whole block (line 89-95) could just be 

```
if [[ "${KUBERNETES_VERSION:-}" =~ "latest" ]]; then
                CI_VERSION_URL="https://dl.k8s.io/ci/${KUBERNETES_VERSION}.txt"
fi
```

jackfrancis: good catch!

scripts/ci-entrypoint.sh
```diff
@@ -79,28 +83,29 @@ select_cluster_template() {
...
-        export CI_VERSION="${CI_VERSION:-$(curl -sSL ${CI_VERSION_URL})}"
-        export KUBERNETES_VERSION="${CI_VERSION}"
+        if [[ -z "${CI_VERSION:-}" ]]; then
+            if [[ "${KUBERNETES_VERSION:-}" =~ "latest" ]]; then
+                CI_VERSION_URL="https://dl.k8s.io/ci/${KUBERNETES_VERSION}.txt"
+            fi
+        fi
     else
         export CLUSTER_TEMPLATE="test/ci/cluster-template-prow.yaml"
+        if [[ "${KUBERNETES_VERSION:-}" =~ "stable" ]]; then
```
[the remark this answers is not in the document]
jackfrancis: Fair, it's not really usable/reliable until we can solve for more rapid reference image adoption

CecileRobertMichon (commented):
/lgtm

CecileRobertMichon:
/assign @mboersma 

@jackfrancis ready to remove the hold?

PR #2334, opened 2022-05-26
WIP ci: use helm to install calico CNI

### LINKED ISSUE #2063: Update e2e tests to verify all system pods are ready before executing tests

jackfrancis:
**What type of PR is this?**

/kind cleanup

This PR replaces the exising ClusterResourceSet-dependent solution for automating calico CNI delivery onto test clusters (and reference cluster templates) with a helm solution. For the private cluster scenario, we're keeping the CRS-dependent solution as a practical way to install the CNI from the management cluster (we have no direct access to the apiserver to `helm install` gestures and we don't yet have a E2E solution to indirectly access the apiserver in such scenarios).

Fixes #2063
Contributes to #2179

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
remove calico ClusterResourceSets from reference cluster templates
```


helm/calico-capz/templates/capz-calico.yaml
```diff
@@ -0,0 +1,1242 @@
...
+        k8s-app: kube-proxy-windows
+    spec:
+      serviceAccountName: kube-proxy
+      securityContext:
+        windowsOptions:
+          hostProcess: true
+          runAsUserName: "NT AUTHORITY\\system"
+      hostNetwork: true
+      containers:
+      - image: sigwindowstools/kube-proxy:{{ .Capabilities.KubeVersion }}-calico-hostprocess
```
jackfrancis, line 1206: @jsturtevant @marosset the existing implementation replaces `+` chars with `_` chars:

`${KUBERNETES_VERSION/+/_}`

Does that suggest that we need to do this string replacement foo in helm as well to ensure that pre-release CI versions work?

jsturtevant: Yes I believe so.

marosset: Yes, also need this updated https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/0187833c20b134404ec85b2280e7fd253effaba1/templates/test/ci/prow-ci-version/patches/kubeadm-bootstrap-windows.yaml#L34

lzhecheng:
Is this PR going to support installing calico on Windows cluster and fix the issue?
https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/2341

jackfrancis:
@lzhecheng Yes, definitely we will ensure that the installed calico helm chart has support for all Windows scenarios.

helm/calico-capz/templates/capz-calico.yaml
```diff
@@ -0,0 +1,1222 @@
+---
+# Source: calico/templates/calico-config.yaml
+# This ConfigMap is used to configure a self-hosted Calico installation.
+kind: ConfigMap
```
jsturtevant, line 4: Is the capz repo the right location for a calico helm chart? Is there an upstream chart we could use or contribute to?  

What is the plan to bring in upstream changes from calico?  This removes some of the automation, that helped keep the calico features upto date.

jackfrancis: My working plan is to start by proving out calico CNI + helm chart (in progress), after which I'll investigate if we can use the official upstream chart(s) to do this work.

test/e2e/common.go
```diff
@@ -212,3 +214,43 @@ func createRestConfig(ctx context.Context, tmpdir, namespace, clusterName string
...
+	})
+	kubeadmControlPlane := &kubeadmv1.KubeadmControlPlane{}
+	key := crclient.ObjectKey{
+		Namespace: cluster.Spec.ControlPlaneRef.Namespace,
+		Name:      cluster.Spec.ControlPlaneRef.Name,
+	}
+	Eventually(func() error {
+		return getter.Get(ctx, key, kubeadmControlPlane)
+	}, input.WaitForControlPlaneIntervals...).Should(Succeed(), "Failed to get KubeadmControlPlane object %s/%s", cluster.Spec.ControlPlaneRef.Namespace, cluster.Spec.ControlPlaneRef.Name)
+	if kubeadmControlPlane.Spec.KubeadmConfigSpec.ClusterConfiguration.ControllerManager.ExtraArgs["cluster-cidr"] == "10.244.0.0/16,2001:1234:5678:9a40::/58" {
```
jackfrancis, line 237: @aramase can you think of a general solution to introspect the `cluster-cidr` configuration to detect dual stack in any scenario? would it be to convert the string into a `,`-delimited array, and then to evaluate for len > 1 ?

test/e2e/common.go
```diff
@@ -212,3 +214,43 @@ func createRestConfig(ctx context.Context, tmpdir, namespace, clusterName string
...
+		InstallCalicoHelmChartForDualStack(ctx, input)
+	} else if kubeadmControlPlane.Spec.KubeadmConfigSpec.ClusterConfiguration.ControllerManager.ExtraArgs["cluster-cidr"] == "2001:1234:5678:9a40::/58" {
```
jackfrancis, line 239: Also @aramase is the equivalent check here to check for an array of len == 1, and to detect for a string that "is an ipv6 cidr"? (hopefully there's a standard library I can use for that, I'll look around)

test/e2e/common.go
jackfrancis: also lemme know if there's a better configuration vector to check than the value of `cluster-cider` for "is a dual stack cluster" or "is an ipv6 cluster"

thanks so much!

aramase: > @aramase can you think of a general solution to introspect the cluster-cidr configuration to detect dual stack in any scenario? would it be to convert the string into a ,-delimited array, and then to evaluate for len > 1 ?

That's right! We do the similar check in node ipam controller: https://github.com/kubernetes/kubernetes/blob/2a821d787b7f21fbb399b0a0bf6d3fa0c79f4293/cmd/cloud-controller-manager/nodeipamcontroller.go#L179.

aramase: Do you plan on making the check generic to see if there is an IPv4 and IPv6 CIDR instead of specific check with `"10.244.0.0/16,2001:1234:5678:9a40::/58"`?

test/e2e/common.go
aramase: > Also @aramase is the equivalent check here to check for an array of len == 1, and to detect for a string that "is an ipv6 cidr"? (hopefully there's a standard library I can use for that, I'll look around)

Yeah, checking the CIDR family is the right way. You can use the [net](https://pkg.go.dev/k8s.io/utils@v0.0.0-20220210201930-3a6ce19ff2f9/net) package in `k8s.io/utils` to check if it's [IPv6CIDRString](https://pkg.go.dev/k8s.io/utils@v0.0.0-20220210201930-3a6ce19ff2f9/net#IsIPv4CIDRString).

test/e2e/common.go
jackfrancis: yep

jackfrancis: done, thanks for the pro tipz! 🙏

templates/flavors/base/cluster-template.yaml
```diff
@@ -69,6 +67,7 @@ spec:
         extraArgs:
           cloud-provider: azure
           cloud-config: /etc/kubernetes/azure.json
+          feature-gates: "WindowsHostProcessContainers=true"
```
jackfrancis, line 70: This addition to the base `Cluster` flavor definition (and the resultant additive explosion to lots of cluster scenario templates) is to enable a single calico configuration that works whether or not you have Windows nodes in your cluster. For example:

```
Error: INSTALLATION FAILED: DaemonSet.apps "kube-proxy-windows" is invalid: spec.template.spec.securityContext.windowsOptions.hostProcess: Forbidden: not allowed when feature gate 'WindowsHostProcessContainers' is not enabled
helm.go:84: [debug] DaemonSet.apps "kube-proxy-windows" is invalid: spec.template.spec.securityContext.windowsOptions.hostProcess: Forbidden: not allowed when feature gate 'WindowsHostProcessContainers' is not enabled
INSTALLATION FAILED
```

More broadly, turning on this feature gate by default enables Windows nodes to be added to a cluster with little friction after the cluster has been created. We should be leaning into "Windows is a possible cluster configuration that we always enable for all clusters" rather than "Windows is an intrinsic feature that you either enable or disable at cluster creation time".

cc @jsturtevant @marosset @CecileRobertMichon

helm/calico-capz/README.md
```diff
@@ -0,0 +1,3 @@
+# calico-capz Helm Chart
+
+This Helm chart enables installation and maintenance of calico CNI components for capz (cluster-api-provider-azure) clusters.
```
lzhecheng, line 3: Could you add more detail in this doc, like how to install?

lzhecheng (commented):
Will the install command be added to scripts like `ci-entrypoint.sh`?

helm/calico-capz/README.md
jackfrancis: Yes, definitely, this is still WIP. We want to investigate using the official tigera calico chart(s) so we don't have to maintain our own.

To your above question: yes, I think we'll have to add this helm flow to the existing CI scripts so that a fully functional Kubernetes cluster is built.

PR #2339, opened 2022-05-27
Update subnet CIDRs when reconciling existing vnet

CecileRobertMichon:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**: This fixes an edge case scenario where the user specifies an existing BYO vnet along with subnets without specifying CIDR ranges. CAPZ will fill out the details in the controller when doing a GET of the existing vnet/subnet. However, in the case that the controller returns in between those reconciling the vnet and the subnets, which it often does as we reconcile resources async, it will attempt to patch the object with the updated vnet CIDR but the wrong defaulted subnet CIDRs, which results in a validation error similar to: `"validation.azurecluster.infrastructure.cluster.x-k8s.io" denied the request: AzureCluster.infrastructure.cluster.x-k8s.io "efreed-workload1" is invalid: [spec.networkSpec.subnets[0].cidrBlocks: Invalid value: "10.0.0.0/16": subnet CIDR not in vnet address space: [10.22.128.0/22], `.

This updates the reconciler code to update both vnet and subnet CIDR ranges together when an existing VNet with subnets matching the defined subnet names is found.

Thanks @evanfreed for reporting the issue.

slack thread: https://kubernetes.slack.com/archives/CEX9HENG7/p1653600096205839

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Update subnet CIDRs when reconciling existing vnet
```

### DIFF
--- a/azure/services/virtualnetworks/virtualnetworks.go
+++ b/azure/services/virtualnetworks/virtualnetworks.go
@@ -40,6 +40,7 @@ type VNetScope interface {
 	VNetSpec() azure.ResourceSpecGetter
 	ClusterName() string
 	IsVnetManaged() bool
+	UpdateSubnetCIDRs(string, []string)
 }
 
 // Service provides operations on Azure resources.
@@ -91,6 +92,15 @@ func (s *Service) Reconcile(ctx context.Context) error {
 			prefixes = to.StringSlice(existingVnet.VirtualNetworkPropertiesFormat.AddressSpace.AddressPrefixes)
 		}
 		vnet.CIDRBlocks = prefixes
+
+		// Update the subnet CIDRs if they already exist.
+		// This makes sure the subnet CIDRs are up to date and there are no validation errors when updating the VNet.
+		// Subnets that are not part of this cluster spec are silently ignored.
+		if existingVnet.Subnets != nil {
+			for _, subnet := range *existingVnet.Subnets {
+				s.Scope.UpdateSubnetCIDRs(to.String(subnet.Name), converters.GetSubnetAddresses(subnet))
+			}
+		}
 	}
 
 	if s.Scope.IsVnetManaged() {
--- a/azure/services/virtualnetworks/virtualnetworks_test.go
+++ b/azure/services/virtualnetworks/virtualnetworks_test.go
@@ -56,6 +56,28 @@ var (
 			"foo":       to.StringPtr("bar"),
 			"something": to.StringPtr("else"),
 		},
+		VirtualNetworkPropertiesFormat: &network.VirtualNetworkPropertiesFormat{
+			AddressSpace: &network.AddressSpace{
+				AddressPrefixes: &[]string{"fake-cidr"},
+			},
+			Subnets: &[]network.Subnet{
+				{
+					Name: to.StringPtr("test-subnet"),
+					SubnetPropertiesFormat: &network.SubnetPropertiesFormat{
+						AddressPrefix: to.StringPtr("subnet-cidr"),
+					},
+				},
+				{
+					Name: to.StringPtr("test-subnet-2"),
+					SubnetPropertiesFormat: &network.SubnetPropertiesFormat{
+						AddressPrefixes: &[]string{
+							"subnet-cidr-1",
+							"subnet-cidr-2",
+						},
+					},
+				},
+			},
+		},
 	}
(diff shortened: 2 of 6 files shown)


azure/services/virtualnetworks/virtualnetworks.go
```diff
@@ -40,6 +40,7 @@ type VNetScope interface {
 	VNetSpec() azure.ResourceSpecGetter
 	ClusterName() string
 	IsVnetManaged() bool
+	UpdateSubnetCIDRs(string, []string)
```
jackfrancis, line 43: Just a note, I read through the existing implementation of `UpdateSubnetCIDRs` and it will silently fail (ignore) when you pass in a name string that doesn't actually match an existing subnet. Not sure if we want to add some reinforcement around there.

CecileRobertMichon: yeah I think that's actually a good thing in this use case. Because the custom vnet might be shared for other purposes outside this cluster, it will likely have subnets we don't care about/don't recognize. We want to skip those and not fail when that happens.

evanfreed:
@CecileRobertMichon I can confirm your fix seems to work!

### Example spec used

```
kind: AzureCluster
metadata:
  name: ${CLUSTER_NAME}
  namespace: default
spec:
  location: ${AZURE_LOCATION}
  networkSpec:
    vnet:
      resourceGroup: ${AZURE_RESOURCE_GROUP}
      name: ${AZURE_VNET_NAME}
    subnets:
      - name: control_plane
        role: control-plane
        #cidrBlocks:
          #- 10.22.128.128/25
      - name: worker_nodes
        role: node
        #cidrBlocks:
          #- 10.22.130.0/23
  resourceGroup: ${AZURE_RESOURCE_GROUP}
```

The machine provisioned in the pre-created CIDR range that we self manage without needing to define it in the spec 🎉 thanks for the help!

/lgtm

azure/services/virtualnetworks/virtualnetworks.go
CecileRobertMichon: updated the comment to make it explicit that it's the desired behavior @jackfrancis

jackfrancis:
wow gfmt is picky

image

jackfrancis (approved):
/lgtm

jackfrancis:
/assign @mboersma

for final review and approval/merge

azure/services/virtualnetworks/virtualnetworks.go
```diff
@@ -91,6 +92,19 @@ func (s *Service) Reconcile(ctx context.Context) error {
 			prefixes = to.StringSlice(existingVnet.VirtualNetworkPropertiesFormat.AddressSpace.AddressPrefixes)
 		}
 		vnet.CIDRBlocks = prefixes
+
+		// Update the subnet CIDRs if they already exist.
+		// This makes sure the subnet CIDRs are up to date and there are no validation errors when updating the VNet.
+		// Subnets that are not part of this cluster spec are silently ignored.
+		if existingVnet.Subnets != nil {
+			for _, subnet := range *existingVnet.Subnets {
+				if subnet.SubnetPropertiesFormat != nil && subnet.SubnetPropertiesFormat.AddressPrefix != nil {
```
Jont828, line 101: We have a similar block in the subnet service where we have to handle the single and multiple CIDR cases. What do you think about adding a converter or other helper function that reads in the SubnetPropertiesFormat and returns a list of addresses?

azure/services/virtualnetworks/virtualnetworks_test.go
```diff
@@ -56,6 +56,25 @@ var (
...
+				{
+					Name: to.StringPtr("test-subnet"),
+					SubnetPropertiesFormat: &network.SubnetPropertiesFormat{
+						AddressPrefix: to.StringPtr("fake-subnet-cidr"),
+					},
+				},
+				{
+					Name: to.StringPtr("other-test-subnet"),
+					SubnetPropertiesFormat: &network.SubnetPropertiesFormat{
+						AddressPrefix: to.StringPtr("other-fake-subnet-cidr"),
```
Jont828, line 73: Could we add a subnet with multiple CIDRs to test the `AddressPrefixes` field as well?

CecileRobertMichon: added

PR #2373, opened 2022-06-09
use latest images for CI jobs

jsturtevant:
**What type of PR is this?**
/kind cleanup
/kind flake

These images have the [cloudbase-init fix](https://github.com/cloudbase/cloudbase-init/issues/94) that resolves windows nodes not coming online.  

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```
/sig windows


jackfrancis (approved):
Changes look sane. @mboersma and @marosset should probably sign off here

templates/test/ci/prow-ci-version-windows-containerd-2022/patches/windows-image-update.yaml
```diff
@@ -12,5 +12,5 @@ spec:
         marketplace:
           publisher: cncf-upstream
           offer: capi-windows
-          sku: k8s-1dot23dot3-windows-2022-containerd
-          version: "latest"
+          sku: windows-2022-containerd-gen1
+          version: "124.1.20220608"
```
CecileRobertMichon, line 16: what are your thoughts on just using version "latest" for these ci-version templates? That way we're always getting the latest image OS changes and since we don't actually use the k8s version on the image it doesn't matter which k8s version is used

CecileRobertMichon: also if do that it would make sense to change the comment above from `an older image` to `the latest published image`

CecileRobertMichon: we'd want to make this change for Linux as well if you want to change both in the same PR

jsturtevant: > what are your thoughts on just using version "latest" for these ci-version templates?

I did consider this and was hesitant, felt like it wasn't deterministic, would never know what image I got: could be 1.24 or 1.22 but I guess since we replace the Kubernetes bits in this case it doesn't matter to much. Prior to this I had a bit more control over the "latest" image since it would be the latest published for 1.23.3 image. Happy to try this route if you are and we can re-assess if it causes to many issues.

marosset:
/lgtm
/hold 
for tests

CecileRobertMichon (commented):
/lgtm

jsturtevant:
These tests failed due to the timeout on windows nodes.  This pr only fixes if for the latest patch releases for upstream CI kubernetes jobs.  Once we have confirmation that these cloudbase-init changes resolve the issue and we get a new release of CI we can rebuild the older versioned VHD's

PR #2224, opened 2022-04-11
Use latest stable

jsturtevant:
**What type of PR is this?**

/kind cleanup

**What this PR does / why we need it**:
The switched back to using the latest stable 1.22 version after https://github.com/kubernetes-sigs/image-builder/issues/853 was resolved.  Undoes: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2220

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

PR #2387, opened 2022-06-15
Change all CI templates to use latest reference images

CecileRobertMichon:
**What type of PR is this?**
/kind cleanup

**What this PR does / why we need it**: This is a follow up to #2373 to ensure that all CI jobs that use CI k8s versions are using latest images instead of pinning to an old reference image.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/templates/test/ci/cluster-template-prow-ci-version-windows-containerd-2022.yaml
+++ b/templates/test/ci/cluster-template-prow-ci-version-windows-containerd-2022.yaml
@@ -216,8 +216,8 @@ spec:
         marketplace:
           offer: capi
           publisher: cncf-upstream
-          sku: k8s-1dot18dot8-ubuntu-1804
-          version: 2020.08.17
+          sku: ubuntu-1804-gen1
+          version: latest
       osDisk:
         diskSizeGB: 128
         osType: Linux
--- a/templates/test/ci/cluster-template-prow-ci-version.yaml
+++ b/templates/test/ci/cluster-template-prow-ci-version.yaml
@@ -216,8 +216,8 @@ spec:
         marketplace:
           offer: capi
           publisher: cncf-upstream
-          sku: k8s-1dot18dot8-ubuntu-1804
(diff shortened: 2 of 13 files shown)


templates/test/ci/cluster-template-prow-ci-version-windows-containerd-2022.yaml
```diff
@@ -216,8 +216,8 @@ spec:
         marketplace:
           offer: capi
           publisher: cncf-upstream
-          sku: k8s-1dot18dot8-ubuntu-1804
-          version: 2020.08.17
+          sku: ubuntu-1804-gen1
+          version: latest
```
mboersma, line 220: `latest` here isn't going to be very predictable, unless I misunderstand. It will depend on the order of publishing, not on semantic version, so `latest` will give a recently published version but it may change from `v1.24.1` one day to `v1.22.10` the next.

CecileRobertMichon: Yes, that's a good point. I guess the point is that we don't care about k8s version on the image since we're replacing it with a CI version of k8s, we just need a fresh image (built with latest image-builder code) that isn't going to be removed and break tests in the future. We had a similar conversation with @jsturtevant in #2224. My main point is that the test is already unpredictable because from one run to the next the value of "latest" k8s ci version may have changed and it's better to test with latest image-builder code that an old image that was built months ago.

jsturtevant: the conversation was here: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2373#pullrequestreview-1003054672

CecileRobertMichon:
/retest

test failure tracked in #2391

PR #645, opened 2020-05-22
✨ support accelerated networking

mboersma:
**What this PR does / why we need it**:

Creates an `acceleratedNetworking` boolean in `AzureMachineTemplate` and in `AzureMachinePool` that enables [Azure accelerated networking](https://docs.microsoft.com/en-us/azure/virtual-network/create-vm-accelerated-networking-cli) on the machine's network interface or on the VMSS, respectively.
If `acceleratedNetworking` is omitted from the provisioning template, it is enabled or disabled after an API call is made to determine if the requested SKU supports it.

Fixes #307 

**Special notes for your reviewer**:

This could use more specific unit test coverage. I'm working on that, so maybe consider this a WIP.

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**Release note**:

```release-note
✨ support accelerated networking
```

Issue #467, opened 2020-03-20
Add support & documentation on using Azure CNI with capz


CecileRobertMichon:
/kind feature

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]
https://github.com/Azure/azure-container-networking/blob/master/docs/cni.md

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

- [ ] https://github.com/Azure/azure-container-networking/issues/1945
- [ ] https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/3343
- [x] https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/3342
- [x] https://github.com/Azure/azure-container-networking/issues/1897

A little lower priority issue, but must be solved to support multiple NICs on control plane.
- [ ] https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/3506

**Environment:**

- cluster-api-provider-azure version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

richardcase:
This would be good for installing the CNI: https://github.com/kubernetes-sigs/cluster-api/pull/3050

CecileRobertMichon:
@mboersma is this one you'd be interested in working on? It's a relatively hi-pri one because it's a blocker for IPv6 (cc @jsturtevant)

mboersma:
@CecileRobertMichon yes indeed, I can probably start on this tomorrow.

jackfrancis:
An implementation question: is the version of Azure CNI currently configurable?

cc @craiglpeters

CecileRobertMichon:
@jackfrancis not sure I understand the question? There is no Azure CNI implementation currently.

jackfrancis:
A reminder to use the "transparent" mode configuration when implementing a capz + Azure CNI scenario. This AKS Engine PR configures Azure CNI for "transparent" mode by default:

https://github.com/Azure/aks-engine/pull/3958

The vanilla installation method that AKS Engine follows (download the public release tarball and untar/gzip it) delivers a default configuration of "bridge" mode. AKS Engine modifies this to "transparent" by sed replacing the appropriate value:

https://github.com/Azure/aks-engine/blob/master/parts/k8s/cloud-init/artifacts/cse_config.sh#L288

Not pretty, but there you have it.

CecileRobertMichon:
[...]

PR #2411, opened 2022-06-22
Support for configurable Network Interfaces

### LINKED ISSUE #2327: Support configurable Network Interfaces
/kind feature

* Ability to configure the number of Network Interfaces to created and attached to a new VM upon AzureMachine creation / AzureMachinePool scale up
* Ability to configure multiple IPConfigurations on a Network Interface of an AzureMachine / AzureMachinePool
* Ability to attach a pre-existing Network Interface to a new AzureMachine

These features are useful for implementing more complex network topologies such as is needed to support use of the Azure CNI, separate pod subnets, and VM's with static IP's that must re-attach to a replacement VM.

Related to: #467 

**Environment:**

brianlieberman:
/kind feature

This PR adds the ability to configure one or more NetworkInterfaces for AzureMachines and AzureMachinePools, as well as pre-warming multiple IPConfigs on each interface.

Fixes #2327

**TODOs**:

- [ x] squashed commits
- [ x] includes documentation
- [ x] adds unit tests

**Release note**:

```release-note
Adds additional fields for AzureMachine, AzureMachineTemplate, and AzureMachinePool to configure multiple NetworkInterfaces. An example configuration:
networkInterfaces:
        - subnetName: control-plane-subnet
           acceleratedNetworking: false
         - subnetName: node-subnet
           acceleratedNetworking: true
           privateIPConfigs: 2
           publicIPConfigs: 1

will create two interfaces, with the node-subnet interface having two additional private IP addresses, and one additional public IP address
```


CecileRobertMichon:
Haven't looked at the code yet, but thanks for opening this!

One question about the UX:

```
 ipConfigs:
  - {}
  - {}
```

If a user needs to pre-allocate say 250 IPs to work with Azure CNI in a 250 max pods per node configuration, how would this scale? It might make sense to allow the user to specify an IP count instead?

api/v1beta1/azuremachine_webhook.go
```diff
@@ -136,6 +136,20 @@ func (m *AzureMachine) ValidateUpdate(oldRaw runtime.Object) error {
 		)
 	}
 
+	if !reflect.DeepEqual(m.Spec.SubnetName, old.Spec.SubnetName) {
```
jackfrancis, line 139: Is this new immutability requirement for `SubnetName` closely related to this change? Should it be a separate PR?

api/v1beta1/azuremachine_webhook_test.go
```diff
@@ -550,6 +621,17 @@ func TestAzureMachine_Default(t *testing.T) {
 	}
 }
 
+func createrMachineWithNetworkConfig(subnetName string, interfaces []AzureNetworkInterface) *AzureMachine {
```
jackfrancis, line 624: nit: s/createrMachineWithNetworkConfig/createMachineWithNetworkConfig

azure/scope/machine.go
```diff
@@ -220,47 +221,109 @@ func (m *MachineScope) InboundNatSpecs(portsInUse map[int32]struct{}) []azure.Re
...
-		VNetResourceGroup:     m.Vnet().ResourceGroup,
-		SubnetName:            m.AzureMachine.Spec.SubnetName,
-		AcceleratedNetworking: m.AzureMachine.Spec.AcceleratedNetworking,
-		IPv6Enabled:           m.IsIPv6Enabled(),
-		EnableIPForwarding:    m.AzureMachine.Spec.EnableIPForwarding,
-	}
+	nicSpecs := []azure.ResourceSpecGetter{}
+
+	if len(m.AzureMachine.Spec.NetworkInterfaces) < 1 {
+		spec := &networkinterfaces.NICSpec{
```
jackfrancis, line 227: Can we abstract L227-265 into a convenience fun (e.g., `getDefaultNetworkInterfaceSpec`) so that we can know more easily that if we enter into this flow we will return from `NICSpecs()` immediately after constructing the default NIC object?

```
if len(m.AzureMachine.Spec.NetworkInterfaces) < 1 {
    return append(nicSpecs, getDefaultNetworkInterfaceSpec(m))
}
...
```

azure/scope/machine.go
```diff
@@ -220,47 +221,109 @@ func (m *MachineScope) InboundNatSpecs(portsInUse map[int32]struct{}) []azure.Re
...
+			IPv6Enabled:           m.IsIPv6Enabled(),
+			EnableIPForwarding:    m.AzureMachine.Spec.EnableIPForwarding,
+			IPConfigs:             []networkinterfaces.IPConfig{},
+		}
 
-	return []azure.ResourceSpecGetter{spec}
+		spec.SubnetName = n.SubnetName
+
+		// Check for control plane interface setup on interface 0
+		if m.Role() == infrav1.ControlPlane && i == 0 {
```
jackfrancis, line 289: Do we have a guarantee that the first `NetworkInterface` in the `m.AzureMachine.Spec.NetworkInterfaces` slice will always be eth0?

brianlieberman:
Reworked the IPConfig functionality to use integer quantities and updated the example in the PR description

api/v1beta1/azuremachine_webhook.go
brianlieberman: The immutability isn't strictly changed with these changes, however it appeared to be an oversight I believe while looking at managing and validating the new interface level field?

api/v1beta1/azuremachine_types.go
```diff
@@ -114,6 +114,8 @@ type AzureMachineSpec struct {
 	// SubnetName selects the Subnet where the VM will be placed
 	// +optional
 	SubnetName string `json:"subnetName,omitempty"`
+
+	NetworkInterfaces []AzureNetworkInterface `json:"networkInterfaces,omitempty"`
```
Jont828, line 118: Should this be an optional field? Just to clarify, if apply an existing template w/o the `NetworkInterfaces` field do we want it to give an error or just create an empty slice?

api/v1beta1/types.go
```diff
@@ -579,6 +579,37 @@ type SubnetSpec struct {
 	SubnetClassSpec `json:",inline"`
 }
 
+// Network Interfaces to attach to each VM
+// +optional
+type AzureNetworkInterface struct {
+	// The subnet to place the interface in
```
Jont828, line 585:
```suggestion
	// The subnet to place the interface in.
```
Could we end the comments with periods? Not sure if the linters will complain but it's more consistent in any case.

api/v1beta1/types.go
```diff
@@ -579,6 +579,37 @@ type SubnetSpec struct {
...
+	AcceleratedNetworking *bool `json:"acceleratedNetworking,omitempty"`
+
+	// Attach an already provisioned interface by ID
+	// +optional
+	ID string `json:"id,omitempty"`
+}
+
+// IP Configuration defines options to confiure a network interface.
+type AzureIPConfig struct {
+	PrivateIP       string `json:"privateIP,omitempty"`
```
Jont828, line 608: Can we add comments for these fields as well?

api/v1beta1/types.go
```diff
@@ -579,6 +579,37 @@ type SubnetSpec struct {
...
+	// Enable acccelerated networking on the interface
+	// +optional
+	AcceleratedNetworking *bool `json:"acceleratedNetworking,omitempty"`
+
+	// Attach an already provisioned interface by ID
+	// +optional
+	ID string `json:"id,omitempty"`
+}
+
+// IP Configuration defines options to confiure a network interface.
```
Jont828, line 606:
```suggestion
// AzureIPConfig defines options to configure a network interface.
```

api/v1beta1/types.go
```diff
@@ -579,6 +579,37 @@ type SubnetSpec struct {
 	SubnetClassSpec `json:",inline"`
 }
 
+// Network Interfaces to attach to each VM
```
Jont828, line 582:
```suggestion
// AzureNetworkInterface defineds a network interface to attach to each VM.
```
Typically we want to start the documentation comment with the name of the type/function. I believe the linters will complain about it if it doesn't match.

azure/scope/machine.go
```diff
@@ -220,6 +221,91 @@ func (m *MachineScope) InboundNatSpecs(portsInUse map[int32]struct{}) []azure.Re
...
+			EnableIPForwarding:    m.AzureMachine.Spec.EnableIPForwarding,
+			IPConfigs:             []networkinterfaces.IPConfig{},
+		}
+
+		spec.SubnetName = n.SubnetName
+
+		// Check for control plane interface setup on interface 0
+		if m.Role() == infrav1.ControlPlane && i == 0 {
+			spec.PublicLBName = m.OutboundLBName(m.Role())
+			spec.PublicLBAddressPoolName = m.OutboundPoolName(m.OutboundLBName(m.Role()))
```
Jont828, line 253: Maybe move this line with `spec.PublicLBAddressPoolName` into the `if` block below since it would be overridden in the `else` block.

azure/scope/machine.go
```diff
@@ -220,6 +221,91 @@ func (m *MachineScope) InboundNatSpecs(portsInUse map[int32]struct{}) []azure.Re
...
+			if m.IsAPIServerPrivate() {
+				spec.InternalLBName = m.APIServerLBName()
+				spec.InternalLBAddressPoolName = m.APIServerLBPoolName(m.APIServerLBName())
+			} else {
+				spec.PublicLBNATRuleName = m.Name()
+				spec.PublicLBAddressPoolName = m.APIServerLBPoolName(m.APIServerLBName())
+			}
+		}
+
+		// If NAT gateway is not enabled and node has no public IP, then the NIC needs to reference the LB to get outbound traffic.
```
Jont828, line 263:
```suggestion
		// If the NAT gateway is not enabled and node has no public IP, then the NIC needs to reference the LB to get outbound traffic.
```
(nit)

azure/scope/machine.go
Jont828: Yeah I would definitely prefer a more reliable indication. Maybe we could make a helper function in the scope for getting eth0? That way it's less error prone and more readable.

azure/scope/machinepool_test.go
```diff
@@ -97,6 +97,48 @@ func TestMachinePoolScope_Name(t *testing.T) {
 		})
 	}
 }
+func TestMachinePoolScope_MultipleInterfaces(t *testing.T) {
```
Jont828, line 100: Could we add a couple different edge cases to this test, i.e. 0, 1, and >1?

azure/services/networkinterfaces/spec_test.go
```diff
@@ -311,12 +321,14 @@ func TestParameters(t *testing.T) {
...
 					Location: to.StringPtr("fake-location"),
 					InterfacePropertiesFormat: &network.InterfacePropertiesFormat{
+						Primary:                     nil,
 						EnableAcceleratedNetworking: to.BoolPtr(true),
 						EnableIPForwarding:          to.BoolPtr(true),
 						IPConfigurations: &[]network.InterfaceIPConfiguration{
 							{
 								Name: to.StringPtr("pipConfig"),
 								InterfaceIPConfigurationPropertiesFormat: &network.InterfaceIPConfigurationPropertiesFormat{
+									Primary:                         to.BoolPtr(true),
```
Jont828, line 331: Could we add some tests where we check the new ipConfigurations generated in `spec.go`?

azure/services/scalesets/scalesets.go
```diff
@@ -475,6 +452,123 @@ func (s *Service) buildVMSSFromSpec(ctx context.Context, vmssSpec azure.ScaleSet
...
+								PrivateIPAddressVersion: compute.IPVersionIPv4,
+								Subnet: &compute.APIEntityReference{
+									ID: to.StringPtr(azure.SubnetID(s.Scope.SubscriptionID(), vmssSpec.VNetResourceGroup, vmssSpec.VNetName, n.SubnetName)),
+								},
+								PublicIPAddressConfiguration: &compute.VirtualMachineScaleSetPublicIPAddressConfiguration{},
+							},
+						}
+						if j == 0 {
+							ipconfig.Primary = to.BoolPtr(true)
+							if i == 0 {
```
Jont828, line 499: Similar to what @jackfrancis said, is there any other identifier we can use besides the index to check for this case?

azure/services/scalesets/scalesets_test.go
```diff
@@ -255,6 +255,79 @@ func TestReconcileVMSS(t *testing.T) {
...
+				vmssIPConfigs := []compute.VirtualMachineScaleSetIPConfiguration{
+					{
+						Name: to.StringPtr("private-ipConfig0"),
+						VirtualMachineScaleSetIPConfigurationProperties: &compute.VirtualMachineScaleSetIPConfigurationProperties{
+							Primary:                 to.BoolPtr(true),
+							PrivateIPAddressVersion: compute.IPVersionIPv4,
+							Subnet: &compute.APIEntityReference{
+								ID: to.StringPtr("/subscriptions/123/resourceGroups/my-rg/providers/Microsoft.Network/virtualNetworks/my-vnet/subnets/subnet2"),
+							},
+							//LoadBalancerBackendAddressPools: &[]compute.SubResource{{ID: to.StringPtr("/subscriptions/123/resourceGroups/my-rg/providers/Microsoft.Network/loadBalancers/capz-lb/backendAddressPools/backendPool")}},
```
Jont828, line 303: Do we want to add this back or delete it?

azure/services/scalesets/scalesets_test.go
```diff
@@ -255,6 +255,79 @@ func TestReconcileVMSS(t *testing.T) {
...
+					},
+					{
+						Name: to.StringPtr("private-ipConfig1"),
+						VirtualMachineScaleSetIPConfigurationProperties: &compute.VirtualMachineScaleSetIPConfigurationProperties{
+							Primary:                 to.BoolPtr(false),
+							PrivateIPAddressVersion: compute.IPVersionIPv4,
+							Subnet: &compute.APIEntityReference{
+								ID: to.StringPtr("/subscriptions/123/resourceGroups/my-rg/providers/Microsoft.Network/virtualNetworks/my-vnet/subnets/subnet2"),
+							},
+							//LoadBalancerBackendAddressPools: &[]compute.SubResource{{ID: to.StringPtr("/subscriptions/123/resourceGroups/my-rg/providers/Microsoft.Network/loadBalancers/capz-lb/backendAddressPools/backendPool")}},
```
Jont828, line 314: Ditto

exp/api/v1alpha4/azuremachinepool_conversion.go
```diff
@@ -17,6 +17,7 @@ limitations under the License.
 package v1alpha4
 
 import (
+	machineryConversion "k8s.io/apimachinery/pkg/conversion"
```
Jont828, line 20:
```suggestion
	apiMachineryConversion "k8s.io/apimachinery/pkg/conversion"
```
Nit: for consistency with import name in other conversions.

exp/api/v1beta1/azuremachinepool_types.go
```diff
@@ -87,6 +87,10 @@ type (
 		// SubnetName selects the Subnet where the VMSS will be placed
 		// +optional
 		SubnetName string `json:"subnetName,omitempty"`
+
+		// Network Interfaces to attach to the to a virtual machine
```
Jont828, line 91:
```suggestion
		// NetworkInterfaces specifies the network interfaces to attach to the to a virtual machine.
```

exp/api/v1beta1/azuremachinepool_webhook.go
```diff
@@ -100,6 +101,13 @@ func (amp *AzureMachinePool) Validate(old runtime.Object) error {
 	return kerrors.NewAggregate(errs)
 }
 
+func (amp *AzureMachinePool) ValidateNetwork() error {
```
Jont828, line 104: Can we add a header comment for this function?

azure/services/networkinterfaces/spec.go
```diff
@@ -45,6 +47,14 @@ type NICSpec struct {
 	IPv6Enabled               bool
 	EnableIPForwarding        bool
 	SKU                       *resourceskus.SKU
+	IPConfigs                 []IPConfig
+	Primary                   *bool
+}
+
+type IPConfig struct {
```
Jont828, line 54: Can we add a header comment for this type?

api/v1beta1/azuremachine_types.go
Jont828: Also we're going to want to add a header comment for this as well.

Jont828 (changes requested):
Thanks for working on this PR, something that touches this much code definitely takes a while to put together. Here are some suggestions on some tweaks to make.

azure/scope/machine.go
dthorsen: Can you clarify what you are proposing? Utilizing the interface index appears to be the only current reliable way of ensuring your intended interface order in the OS.

dthorsen: Oh re-reading your comment I think I understand. The intent is to improve the code readability. Makes sense to me to have a helper function.

azure/scope/machine.go
```diff
@@ -221,6 +222,73 @@ func (m *MachineScope) InboundNatSpecs(portsInUse map[int32]struct{}) []azure.Re
...
+		return append(nicSpecs, defaultSpec)
+	}
+
+	for i, n := range m.AzureMachine.Spec.NetworkInterfaces {
+		var spec *networkinterfaces.NICSpec
+		if n.ID != "" {
+			continue
+		}
+		if i == 0 {
+			spec = m.DefaultNICSpec()
```
CecileRobertMichon, line 237: it's unclear to me why we're using the defaults for index 0 but not for the others. What's significant about the order the network interfaces are specified in? What's different between the default spec and the fields we're setting below in the `else` statement?

azure/scope/machine.go
```diff
@@ -221,6 +222,73 @@ func (m *MachineScope) InboundNatSpecs(portsInUse map[int32]struct{}) []azure.Re
...
+	nicSpecs := []azure.ResourceSpecGetter{}
+	defaultSpec := m.DefaultNICSpec()
+	if len(m.AzureMachine.Spec.NetworkInterfaces) < 1 {
+		return append(nicSpecs, defaultSpec)
+	}
+
+	for i, n := range m.AzureMachine.Spec.NetworkInterfaces {
+		var spec *networkinterfaces.NICSpec
+		if n.ID != "" {
+			continue
```
CecileRobertMichon, line 234: this deviates significantly from the pattern we've established for other BYO existing resource in CAPZ. The way it works for other resources is that the reconciler will add a tag on each resource it creates to allow it to identify the resources it created vs. the ones the user pre-created. See the virtual networks as an example. There are pros and cons for both ways of doing it, however because this is creating an inconsistent user experience I would recommend putting BYO network interface as out of scope for this PR and tackling it in its own issue/PR so we can iterate on the design.

api/v1beta1/azuremachine_types.go
```diff
@@ -118,6 +118,8 @@ type AzureMachineSpec struct {
...
```
Jont828, line 122: Can we add a doc comment for this field? And if it's optional we can also add `// +optional` as well for the kubebuilder generation.

api/v1beta1/types.go
```diff
@@ -579,6 +579,42 @@ type SubnetSpec struct {
 	SubnetClassSpec `json:",inline"`
 }
 
+// AzureNetworkInterface defineds a network interface.
+// +optional
```
Jont828, line 583: I don't think we need an `// +optional` comment tag for a struct, only fields within a struct

azure/services/networkinterfaces/spec_test.go
```diff
@@ -149,6 +149,54 @@ var (
...
+		MachineName:           "azure-test1",
+		SubnetName:            "my-subnet",
+		VNetName:              "my-vnet",
+		IPv6Enabled:           false,
+		VNetResourceGroup:     "my-rg",
+		PublicLBName:          "my-public-lb",
+		AcceleratedNetworking: nil,
+		SKU:                   &fakeSku,
+		EnableIPForwarding:    true,
+		IPConfigs:             []IPConfig{{}, {}},
```
Jont828, line 198: Can we try to add something in the `IPConfig` structs for these tests?

exp/api/v1beta1/azuremachinepool_types.go
```diff
@@ -87,6 +87,10 @@ type (
 		// SubnetName selects the Subnet where the VMSS will be placed
 		// +optional
 		SubnetName string `json:"subnetName,omitempty"`
+
+		// NetworkInterfaces to attach to the to a virtual machine.
```
Jont828, line 91:
```suggestion
		// NetworkInterfaces defines the network interfaces to attach to the to a virtual machine.
```

CecileRobertMichon:
@dthorsen looks like the PR still has a merge conflict and the CLA authorization is missing

api/v1beta1/types.go
```diff
@@ -579,6 +579,42 @@ type SubnetSpec struct {
...
+type AzureNetworkInterface struct {
+	// The subnet to place the interface in.
+	SubnetName string `json:"subnetName,omitempty"`
```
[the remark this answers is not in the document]
CecileRobertMichon: @dthorsen this comment still holds

nawazkh (commented):
@CecileRobertMichon and I were reviewing this PR and below are some of the thoughts on initial scanning.

api/v1beta1/azuremachine_types.go
```diff
@@ -118,6 +118,12 @@ type AzureMachineSpec struct {
 	// SubnetName selects the Subnet where the VM will be placed
 	// +optional
 	SubnetName string `json:"subnetName,omitempty"`
+
+	// NetworkInterfaces specifies a list of network interface configurations.
+	// If left unspecified, the VM will get a single network interface with a
+	// single IPConfig in the subnet specified in the spec.subnetName field.
+	// +optional
+	NetworkInterfaces []NetworkInterface `json:"networkInterfaces,omitempty"`
```
CecileRobertMichon, line 126: Thinking about this a bit more, what's potentially confusing here is that the naming and the comment of the field imply that this is defining _all_ network interfaces for that VM. That means as a user if I specify this field I would need to somehow specify the primary NIC and IPs as well. But looking at the code, it only is used for secondary NICs that get added on top of the default one (and that one is not configurable AFAICT). 

Would it make more sense to rename this to something like `SecondaryNetworkInterfaces` or `AdditionalNetworkInterfaces` and adjust the comment to make it match what the code it doing: adding these optional NICs to the VM if they are specified, in addition to the always present primary NIC.

azure/services/networkinterfaces/spec.go
```diff
@@ -132,6 +142,36 @@ func (s *NICSpec) Parameters(existing interface{}) (parameters interface{}, err
...
+			config.InterfaceIPConfigurationPropertiesFormat.PrivateIPAllocationMethod = network.IPAllocationMethodStatic
+			config.InterfaceIPConfigurationPropertiesFormat.PrivateIPAddress = &c.PrivateIP
+		}
+
+		if c.PublicIP && c.PublicIPAddress == "" {
+			config.InterfaceIPConfigurationPropertiesFormat.PublicIPAddress = &network.PublicIPAddress{
+				PublicIPAddressPropertiesFormat: &network.PublicIPAddressPropertiesFormat{
+					PublicIPAllocationMethod: network.IPAllocationMethodStatic,
+					IPAddress:                &c.PublicIPAddress,
+				}}
```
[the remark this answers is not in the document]
dthorsen: Good catch! This should have been `if c.PublicIP && c.PublicIPAddress != ""`. Fixing it.

azure/scope/machine.go
```diff
@@ -221,19 +221,17 @@ func (m *MachineScope) InboundNatSpecs(portsInUse map[int32]struct{}) []azure.Re
...
+	nicSpecs = append(nicSpecs, m.PrimaryNICSpec())
+	for _, spec := range m.SecondaryNICSpecs() {
+		nicSpecs = append(nicSpecs, spec)
 	}
+	return nicSpecs
+}
+
+// PrimaryNICSpec returns the NICSpec associated with the VM's primary network interface.
+func (m *MachineScope) PrimaryNICSpec() *networkinterfaces.NICSpec {
+	spec := m.DefaultNICSpec()
```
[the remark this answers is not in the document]
dthorsen: The name of that function is probably confusing. That spec is not user-facing, it is an internal structure and the `DefaultNICSpec()` function is just intended to initialize a new one with the default values commonly used across NICs. But many of the equivalent structures in this file are also named "Spec". `DiskSpecs`, `RoleAssignmentSpecs`, etc.

azure/scope/machinepool.go
```diff
@@ -128,9 +128,31 @@ func (m *MachinePoolScope) ScaleSetSpec() azure.ScaleSetSpec {
...
 	}
 }
 
+// PublicIPPrefixSpecs returns the specs for public IP prefixes to be associated with the VMSS.
+func (m *MachinePoolScope) PublicIPPrefixSpecs() []azure.PublicIPPrefixSpec {
+	if m.AzureMachinePool.Spec.Template.AllocatePublicIPPrefix {
+		return []azure.PublicIPPrefixSpec{
+			{
+				Name:         azure.GenerateMachinePoolPublicIPPrefixName(m.Name()),
+				PrefixLength: 28,
```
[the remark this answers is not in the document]
dthorsen: PrefixLength is setting the CIDR mask size of the PublicIPPrefix that Azure will allocate when creating the PublicIPPrefix resource. I can move it to a constant, though I don't think that will lend additional clarity. I chose 28 because it is the largest available, but actually now I am second-guessing this choice. I thought we needed this to get instance level, standard SKU public IPs but it looks like that is not the case per this doc. I am going to try to remove all the public ip prefix stuff to see if the instance level public IPs works without these static public ip prefixes.

exp/api/v1beta1/azuremachinepool_types.go
```diff
@@ -87,6 +87,14 @@ type (
...
+		// +optional
+		NetworkInterfaces []infrav1.NetworkInterface `json:"networkInterfaces,omitempty"`
+
+		// AllocatePublicIPPrefix enables the automatic creation and association of a public IP prefix for the nodes in the AzureMachinePool.
+		// +optional
+		AllocatePublicIPPrefix bool `json:"allocatePublicIPPrefix,omitempty"`
```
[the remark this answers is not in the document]
dthorsen: We needed instance level public IPs, and I was surprised to see that AzureMachines support them, but AzureMachinePools do not. I added it in this PR because my goal with this PR was to add the network configuration flexibility we needed for our use cases. We also have a need for bring your own NICs, but that will be in a follow up PR.

api/v1beta1/azuremachine_types.go
dthorsen: This field is actually defining all network interfaces for the VM. That is needed to specify the number of IPConfigs on the primary interface if you are using Azure CNI on the primary network interface.

azure/scope/machinepool.go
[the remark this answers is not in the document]
dthorsen: 😞 
I tried just using the method from the docs but it looks like I need to update the compute API version dependencies. I'll pull this logic out of this PR for now. Will need a new PR that bumps the compute API to the new version, which unfortunately has tons of breaking changes.

`Specified api-version 2021-11-01 does not meet the minimum required api-version 2022-03-01 to set this property skuOnPublicIPAddressConfiguration`

exp/api/v1beta1/azuremachinepool_types.go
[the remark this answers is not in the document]
CecileRobertMichon: @dthorsen can we please split that feature into a separate PR? This PR is already huge and hard to review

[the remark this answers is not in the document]
dthorsen: Agreed, doing that now

dthorsen:
I ran the e2e tests locally and they passed. Considering the timeouts I see in the prow job logs, I suspect something is flaky? Does anyone have any other insight? Should I just keep retrying that test?

api/v1beta1/azuremachine_types.go
CecileRobertMichon: In that case it makes https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2411#discussion_r914260114 and https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2411#discussion_r1012367038 even more relevant

azure/scope/machine.go
```diff
@@ -257,21 +253,62 @@ func (m *MachineScope) NICSpecs() []azure.ResourceSpecGetter {
...
+	if len(m.AzureMachine.Spec.NetworkInterfaces) < 1 {
+		// Retains for backwards compatibility with the old behavior, prior to the addition of configurable AzureNetworkInterfaces.
+		return spec
 	}
 
-	if m.cache != nil {
-		spec.SKU = &m.cache.VMSKU
+	// We always use the first network interface as the primary network interface.
+	primaryNICIndex := 0
+	nic := m.AzureMachine.Spec.NetworkInterfaces[primaryNICIndex]
```
CecileRobertMichon, line 272: we shouldn't rely on the position of the item in the list API field here, lists are not strictly ordered and there is nothing that guarantees the order won't change through conversions, users making changes, etc.

dthorsen: The `Primary: true` field in the ARM Compute API as I understand it does not enforce interface attachment order. Only the order of the config in the array does that. My understanding of the `Primary: true` field is that it indicates the interface that is used for traffic to the IMDS and maybe also `168.63.129.16`. I have observed that if you attempt to access the IMDS over an interface that is not designated as `Primary: true`, the IMDS is inaccessible.

Slices in Go and arrays in JSON are strictly ordered. If this were not the case this problem would already exist in many places. Ordinality in the list in the call to the ARM Compute API is also the only way to signal to the Azure platform which order the network interfaces should be attached to the virtual hardware so that the OS finds them in the expected order.

The same approach is taken in [terraform](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/linux_virtual_machine#network_interface_ids).

Similar is done [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/ca53d743a546c8849a1fde92cdfb39ce558aa35f/api/v1beta2/awsmachine_types.go#L134)

This makes for a very natural UX, because as the user configures their network interfaces, they naturally assume the first in the list is going to be the first in the Azure Compute API, and `eth0` in the OS.

On a slight tangent, it is theoretically possible that a CAPZ user may have a use case to set `primary: true` on an interface that is not the first in the list, but that seems very unlikely, and I believe it exposes an unnecessary foot-gun for the function of things like IMDS.

CecileRobertMichon: > The Primary: true field in the ARM Compute API as I understand it does not enforce interface attachment order. Only the order of the config in the array does that

this part is surprising to me, why would the API have a field `Primary` if that gets ignored? do you have a link to docs or code that states this? 

> Slices in Go and arrays in JSON are strictly ordered

this is true, but here I'm more concerned about the user experience of assuming order, and for round trip conversion. I thought I recalled something about k8s API conventions and not relying on order for fields of type array, but I may be misremembering. @sbueringer @vincepri do you know anything about this? Also are there any ClusterClass implications?

> Similar is done [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/ca53d743a546c8849a1fde92cdfb39ce558aa35f/api/v1beta2/awsmachine_types.go#L134)

I can't see anywhere in CAPA code where we are making any assumptions about the item at position 0 being different than the others, AFAICT it's just performing the same attach operation for every item in the list https://cs.github.com/kubernetes-sigs/cluster-api-provider-aws/blob/ca53d743a546c8849a1fde92cdfb39ce558aa35f/pkg/cloud/services/ec2/instances.go?q=.NetworkInterfaces#L246

Here it's different because we're not just passing through the list to Azure, we're configuring the NICs differently based on their position in the list.

> This makes for a very natural UX, because as the user configures their network interfaces, they naturally assume the first in the list is going to be the first in the Azure Compute API, and eth0 in the OS.

This is my main concern, we shouldn't make any implicit assumptions when it comes to UX, if we make an assumption about order being significant then it should be clearly documented in the docs and API field description. This is different from other existing fields in the API, such as [subnets](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/api/v1beta1/types.go#L140) and [security rules](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/api/v1beta1/types.go#L242), where order does not matter.

CecileRobertMichon:
> I ran the e2e tests locally and they passed. Considering the timeouts I see in the prow job logs, I suspect something is flaky? Does anyone have any other insight? Should I just keep retrying that test?

ipv6 cluster worker node kubeadm join failed with:

```
[2022-11-10 16:30:43] error execution phase preflight: unable to fetch the kubeadm-config ConfigMap: failed to get component configs: could not download the kubelet configuration from ConfigMap "kubelet-config" or "kubelet-config-1.23": Get "https://capz-e2e-r87hcg-ipv6-3560cb8d.westus2.cloudapp.azure.com:6443/api/v1/namespaces/kube-system/configmaps/kubelet-config-1.23?timeout=10s": net/http: request canceled (Client.Timeout exceeded while awaiting headers)
```

source: https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/2411/pull-cluster-api-provider-azure-e2e/1590739619538800640/artifacts/clusters/capz-e2e-r87hcg-ipv6/machines/capz-e2e-r87hcg-ipv6-md-0-856bdddf4c-zzk8n/cloud-init-output.log

That seems like an unrelated network flake to me but let's retry...

/retest

azure/scope/machine.go
dthorsen: The order should be stable. K8s core relies on stable order of lists too for things such as [executing initContainers in order](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/#detailed-behavior). Conversion logic looks correct to me to preserve order, relying on json marshal/unmarshal. 

Argo workflows CRD uses list order for ordered operations as well. 

The Primary field in the Compute Vm API isn't unused but it doesn't set interface index. I just looked it up and it is used to set influence the default route in the OS (I suspect by modifying the DHCP response for the VM). It is documented [here](https://learn.microsoft.com/en-us/azure/virtual-network/virtual-network-network-interface-vm#constraints).

sbueringer: Missed the mention somehow

> this is true, but here I'm more concerned about the user experience of assuming order, and for round trip conversion. I thought I recalled something about k8s API conventions and not relying on order for fields of type array, but I may be misremembering. @sbueringer @vincepri do you know anything about this? Also are there any ClusterClass implications?

I'm not aware of conventions around not relying on array ordering. Not sure if that answers the question but ClusterClass does preserve array ordering. So if you have a new array/slice field and that's passed through ClusterClass or set via a variable the array order will be preserved.

CecileRobertMichon:
@dthorsen I do see some errors in the latest e2e run controller [logs ](https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/2411/pull-cluster-api-provider-azure-e2e/1600167414060814336/artifacts/clusters/bootstrap/logs/capz-system/capz-controller-manager/capz-controller-manager-5b4c8867d5-v7n7t/manager.log)that seem related to this PR

```
E1206 17:04:54.815032       1 controller.go:326]  "msg"="Reconciler error" "error"="admission webhook \"validation.azuremachine.infrastructure.cluster.x-k8s.io\" denied the request: AzureMachine.infrastructure.cluster.x-k8s.io \"capz-e2e-dm3mea-ha-md-0-kqnqw\" is invalid: spec.networkInterfaces: Invalid value: \"null\": field is immutable" "AzureMachine"={"name":"capz-e2e-dm3mea-ha-md-0-kqnqw","namespace":"capz-e2e-dm3mea"} "controller"="azuremachine" "controllerGroup"="infrastructure.cluster.x-k8s.io" "controllerKind"="AzureMachine" "name"="capz-e2e-dm3mea-ha-md-0-kqnqw" "namespace"="capz-e2e-dm3mea" "reconcileID"="752409a6-3bb6-4712-979d-6307d9e24130"
```

api/v1beta1/azuremachine_types.go
```diff
@@ -115,7 +116,8 @@ type AzureMachineSpec struct {
 	// +optional
 	SecurityProfile *SecurityProfile `json:"securityProfile,omitempty"`
 
-	// SubnetName selects the Subnet where the VM will be placed
+	// SubnetName (deprecated) selects the Subnet where the VM will be placed.
+	// This field is deprecated. The subnetName should be set in the networkInterfaces field.
```
CecileRobertMichon, line 120:
```suggestion
	// // Deprecated: subnetName should be set in the networkInterfaces field.
```

to be consistent with other deprecated fields in the API

CecileRobertMichon: ditto for others

azure/scope/machine.go
```diff
@@ -228,22 +228,18 @@ func (m *MachineScope) InboundNatSpecs() []azure.ResourceSpecGetter {
...
-		VNetResourceGroup:     m.Vnet().ResourceGroup,
-		SubnetName:            m.AzureMachine.Spec.SubnetName,
-		AcceleratedNetworking: m.AzureMachine.Spec.AcceleratedNetworking,
-		DNSServers:            m.AzureMachine.Spec.DNSServers,
-		IPv6Enabled:           m.IsIPv6Enabled(),
-		EnableIPForwarding:    m.AzureMachine.Spec.EnableIPForwarding,
-		AdditionalTags:        m.AdditionalTags(),
-		ClusterName:           m.ClusterName(),
+	nicSpecs := []azure.ResourceSpecGetter{}
+	nicSpecs = append(nicSpecs, m.PrimaryNICSpec())
```
CecileRobertMichon, line 232: we shouldn't need two functions for primary vs secondary, This function should only range over `m.AzureMachine.Spec.NetworkInterfaces` and pass through the fields to convert them in a `[]*networkinterfaces.NICSpec`. There is still way too much business logic happening in here.

azure/scope/machine.go
```diff
@@ -680,7 +751,7 @@ func (m *MachineScope) GetVMImage(ctx context.Context) (*infrav1.Image, error) {
 // Note: this logic exists only for purposes of ensuring backwards compatibility for old clusters created without the `subnetName` field being
 // set, and should be removed in the future when this field is no longer optional.
 func (m *MachineScope) SetSubnetName() error {
-	if m.AzureMachine.Spec.SubnetName == "" {
+	if m.AzureMachine.Spec.SubnetName == "" && len(m.AzureMachine.Spec.NetworkInterfaces) == 0 {
```
CecileRobertMichon, line 754: does this need to be completely changed/removed now that we're no longer using the SubnetName field outside of NetworkInterfaces?

dthorsen: I believe this function is unfortunately still necessary, though I am going to make a small change to it to update it for the new field.

I don't think it is possible to implement this logic in the defaulting webhook. There is no guarantee that when the resources are applied to the management cluster that we can reliably find the AzureCluster resource to fetch the correct SubnetName. The order or resource apply is not enforced. This method will need to stay in the reconciler until the old fields are actually removed.

azure/scope/machine.go
dthorsen: I'm happy to combine them back into a single function with just one loop. Working on that now. 

Regarding the business logic, that is a more difficult question. The bulk of it predates this PR. That logic either needs to exist here, move somewhere else, or the features provided be removed or redesigned. Those problems are all unrelated to this PR though.

I am going to attempt to trim down the business logic I added now that we have the defaulting webhook logic in place.

dthorsen:
@CecileRobertMichon The tests are finally passing :). I did my best to incorporate your feedback. The business logic in the NICSpec builder is reduced.

azure/types.go
```diff
@@ -65,7 +65,7 @@ type ScaleSetSpec struct {
 	AdditionalCapabilities       *infrav1.AdditionalCapabilities
 	DiagnosticsProfile           *infrav1.Diagnostics
 	FailureDomains               []string
-	VMExtensions                 []infrav1.VMExtension
```
CecileRobertMichon, line 68: what happened here?

CecileRobertMichon (commented):
LGTM pending few minor comments

/assign @nawazkh for another round of review

Docs are missing but will work on docs + e2e tests with @nawazkh as part of #467 

@dthorsen thanks so much for doing the refactor and staying with us. Let's get this merged before January 10th so it can be included in v1.7.0. Could you please squash commits?

api/v1beta1/azuremachine_webhook.go
```diff
@@ -152,6 +154,24 @@ func (m *AzureMachine) ValidateUpdate(oldRaw runtime.Object) error {
 		}
 	}
 
+	if !reflect.DeepEqual(m.Spec.NetworkInterfaces, old.Spec.NetworkInterfaces) {
+		// The defaulting webhook may have migrated values from the old SubnetName field to the new NetworkInterfaces format.
+		old.Spec.SetNetworkInterfacesDefaults()
+
+		// The reconciler will populate the SubnetName on the first interface if the user left it blank.
```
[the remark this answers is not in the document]
dthorsen: This is running the Network Interfaces defaulter on the "old" resource. The defaulting webhook would have run against the "new" resource. By running the Network Interfaces defaulter against the "old" resource, we make the "old" and "new" identical in the case that the modification was due to the defaulter migrating values from the old SubnetName/AcceleratedNetworking fields to the new NetworkInterfaces fields. I used this as a way to bypass immutability for this very specific case since we need the validating webhook to permit this particular mutation.

azure/defaults.go
```diff
@@ -174,7 +174,10 @@ func GenerateVNetLinkName(vnetName string) string {
 }
 
 // GenerateNICName generates the name of a network interface based on the name of a VM.
-func GenerateNICName(machineName string) string {
+func GenerateNICName(machineName string, multiNIC bool, index int) string {
```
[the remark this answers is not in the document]
dthorsen: Do you mean the inverse? We assume multipleNIC if index is not nil? And assuming that, then the logic in NICSpecs does get a little more opaque, but still works. Do you prefer this method? or passing multiNIC to the GenerateNICName as another parameter? I feel like the extra parameter to GenerateNICName is a bit clearer personally i.e.:
```golang
// NICSpecs returns the network interface specs.
func (m *MachineScope) NICSpecs() []azure.ResourceSpecGetter {
	nicSpecs := []azure.ResourceSpecGetter{}

	// For backwards compatibility we need to ensure the NIC Name does not change on existing machines
	// created prior to multiple NIC support
	isMultiNIC := len(m.AzureMachine.Spec.NetworkInterfaces) > 1

	for i := 0; i < len(m.AzureMachine.Spec.NetworkInterfaces); i++ {
		isPrimary := i == 0
		var nicIndex *int = nil
		if isMultiNIC {
			nicIndex = &i
		}
		nicName := azure.GenerateNICName(m.Name(), nicIndex)
		nicSpecs = append(nicSpecs, m.BuildNICSpec(nicName, m.AzureMachine.Spec.NetworkInterfaces[i], isPrimary))
	}
	return nicSpecs
}
```

azure/scope/machine.go
```diff
@@ -681,7 +710,7 @@ func (m *MachineScope) GetVMImage(ctx context.Context) (*infrav1.Image, error) {
 // Note: this logic exists only for purposes of ensuring backwards compatibility for old clusters created without the `subnetName` field being
 // set, and should be removed in the future when this field is no longer optional.
 func (m *MachineScope) SetSubnetName() error {
-	if m.AzureMachine.Spec.SubnetName == "" {
+	if m.AzureMachine.Spec.NetworkInterfaces[0].SubnetName == "" {
```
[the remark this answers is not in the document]
dthorsen: It can certainly be removed once we can do a proper breaking API change. Maybe at v1beta2?

azure/types.go
dthorsen: 🤦‍♂️  Looks like I made an error during a rebase. My apologies, fixing.

azure/services/networkinterfaces/spec.go
```diff
@@ -52,6 +53,14 @@ type NICSpec struct {
...
 	AdditionalTags            infrav1.Tags
 	ClusterName               string
+	IPConfigs                 []IPConfig
+}
+
+// IPConfig defines the specification for an IP address configuration.
+type IPConfig struct {
+	PrivateIP       string
+	PublicIP        bool
+	PublicIPAddress string
```
nawazkh, line 63: Can we instead convert IPConfig struct as below? We will have one less field to manage and a nil pointer will intrinsicly imply weather it is assigned or not?
```suggestion
	PrivateIP       *string
	PublicIPAddress *string
```

controllers/azuremachine_reconciler.go
```diff
@@ -73,6 +73,9 @@ func (s *azureMachineService) Reconcile(ctx context.Context) error {
 	ctx, _, done := tele.StartSpanWithLogger(ctx, "controllers.azureMachineService.Reconcile")
 	defer done()
 
+	// Ensure that the deprecated networking field values have been migrated to the new NetworkInterfaces field.
+	s.scope.AzureMachine.Spec.SetNetworkInterfacesDefaults()
+
```
nawazkh, line 78: AzureMachine Defaulter should take care of this. I don't think we need this here.

exp/controllers/azuremachinepool_reconciler.go
```diff
@@ -57,6 +57,9 @@ func (s *azureMachinePoolService) Reconcile(ctx context.Context) error {
 	ctx, _, done := tele.StartSpanWithLogger(ctx, "controllers.azureMachinePoolService.Reconcile")
 	defer done()
 
+	// Ensure that the deprecated networking field values have been migrated to the new NetworkInterfaces field.
+	s.scope.AzureMachinePool.SetNetworkInterfacesDefaults()
+
```
nawazkh, line 62: Defaulter should take care of this, might not need it here.

nawazkh (commented):
Jotted down my thoughts on another pass. Please take a look!

api/v1beta1/azuremachine_types.go
```diff
@@ -131,6 +129,13 @@ type AzureMachineSpec struct {
 	// VMExtensions specifies a list of extensions to be added to the virtual machine.
 	// +optional
 	VMExtensions []VMExtension `json:"vmExtensions,omitempty"`
+
+	// NetworkInterfaces specifies a list of network interface configurations.
+	// If left unspecified, the VM will get a single network interface with a
+	// single IPConfig in the subnet specified in the cluster's node subnet field.
+	// The primary interface will be the first networkInterface specified (index 0) in the list.
+	// +optional
+	NetworkInterfaces []NetworkInterface `json:"networkInterfaces,omitempty"`
```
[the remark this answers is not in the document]
dthorsen: Slices are always pointers and always nilable. It is also notable that it is not possible to omit this field in the final persisted version of the object. The defaulting webhook will populate it if not set. In this case, omitempty really only applies to the client-side marshal/unmarshal for the purposes of maintaining backward compatibility. We do have tests for the defaulting behavior, so I think we should be covered here.

api/v1beta1/types.go
```diff
@@ -635,6 +635,24 @@ type ServiceEndpointSpec struct {
 	Locations []string `json:"locations"`
 }
 
+// NetworkInterface defines a network interface.
+type NetworkInterface struct {
+	// SubnetName specifies the subnet in which the new network interface will be placed.
+	SubnetName string `json:"subnetName,omitempty"`
+
+	// PrivateIPConfigs specifies the number of private IP addresses to attach to the interface.
+	// Defaults to 1 if not specified.
```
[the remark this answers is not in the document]
dthorsen: We could, though I hesitate to recommend that given the real limit is imposed by the cloud provider. We can certainly mirror the constraint of a maximum of 256 private IP Addresses from Azure, but then if it changes in Azure, it will need to be updated in CAPZ as well. Do we typically mirror Azure API limits in the CAPZ code? I had the impression we usually just pass it through to the Azure APIs and let the Azure APIs enforce the limits.

api/v1beta1/types.go
```diff
@@ -635,6 +635,24 @@ type ServiceEndpointSpec struct {
...
+	// +optional
+	PrivateIPConfigs int `json:"privateIPConfigs,omitempty"`
+
+	// AcceleratedNetworking enables or disables Azure accelerated networking. If omitted, it will be set based on
+	// whether the requested VMSize supports accelerated networking.
+	// If AcceleratedNetworking is set to true with a VMSize that does not support it, Azure will return an error.
```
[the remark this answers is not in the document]
dthorsen: I could be mistaken, but I don't think we do that in the controller. That is handled by default values in the Azure APIs themselves I believe. This comment was simply copied from the older acceleratedNetworking field.

api/v1beta1/azuremachine_default.go
```diff
@@ -115,6 +115,34 @@ func (s *AzureMachineSpec) SetDiagnosticsDefaults() {
...
+	if len(s.NetworkInterfaces) == 0 {
+		s.NetworkInterfaces = []NetworkInterface{
+			{
+				SubnetName:            s.SubnetName,
+				AcceleratedNetworking: s.AcceleratedNetworking,
+			},
+		}
+		s.SubnetName = ""
+		s.AcceleratedNetworking = nil
+	}
```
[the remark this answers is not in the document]
dthorsen: This is actually intentional. We are copying whatever was in the legacy fields to the new fields, even if those are an empty string and a nil pointer. Those are valid values. An empty SubnetName will cause the reconciler to patch the resource later with the node SubnetName from the Cluster scope. This is maintaining compatibility with existing behavior.

api/v1beta1/azuremachine_default.go
```diff
@@ -115,6 +115,34 @@ func (s *AzureMachineSpec) SetDiagnosticsDefaults() {
...
+
+	// Ensure that PrivateIPConfigs defaults to 1 if not specified.
+	for i := 0; i < len(s.NetworkInterfaces); i++ {
+		if s.NetworkInterfaces[i].PrivateIPConfigs == 0 {
+			s.NetworkInterfaces[i].PrivateIPConfigs = 1
+		}
+	}
```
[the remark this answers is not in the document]
dthorsen: That is correct. Every network interface must have at least 1 Primary private IP configuration. 

https://learn.microsoft.com/en-us/azure/virtual-network/ip-services/virtual-network-network-interface-addresses?tabs=nic-address-portal#ipv4

>Each network interface must have one [primary](https://learn.microsoft.com/en-us/azure/virtual-network/ip-services/virtual-network-network-interface-addresses?tabs=nic-address-portal#primary) IP configuration with an assigned [private](https://learn.microsoft.com/en-us/azure/virtual-network/ip-services/virtual-network-network-interface-addresses?tabs=nic-address-portal#private) [IPv4](https://learn.microsoft.com/en-us/azure/virtual-network/ip-services/virtual-network-network-interface-addresses?tabs=nic-address-portal#ipv4) address. You can add one or more [secondary](https://learn.microsoft.com/en-us/azure/virtual-network/ip-services/virtual-network-network-interface-addresses?tabs=nic-address-portal#secondary) IP configurations that each have an IPv4 private and (optionally) an IPv4 [public](https://learn.microsoft.com/en-us/azure/virtual-network/ip-services/virtual-network-network-interface-addresses?tabs=nic-address-portal#public) IP address.

controllers/azuremachine_reconciler.go
dthorsen: It is possible that the reconciler runs before a resource has been modified and therefore before the defaulter has run. This is a protection against that case. The code assumes those new fields are populated and we will cause the reconciler to panic with nil pointer references if a resource has not yet had these fields set.

exp/controllers/azuremachinepool_reconciler.go
dthorsen: Same as above. It is possible that the reconciler runs before a resource has been modified and therefore before the defaulter has run. This is a protection against that case. The code assumes those new fields are populated and we will cause the reconciler to panic with nil pointer references if a resource has not yet had these fields set.

api/v1beta1/azuremachine_webhook.go
```diff
@@ -152,6 +154,24 @@ func (m *AzureMachine) ValidateUpdate(oldRaw runtime.Object) error {
...
+		if old.Spec.NetworkInterfaces[0].SubnetName == "" && m.Spec.NetworkInterfaces[0].SubnetName != "" {
+			old.Spec.NetworkInterfaces[0].SubnetName = m.Spec.NetworkInterfaces[0].SubnetName
+		}
```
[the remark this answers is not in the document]
dthorsen: The controller itself patches the SubnetName field if it is empty, so we have to allow that specific mutation. This is for backwards compatibility. The defaulting webhook cannot set the SubnetName reliably, because we cannot guarantee that the AzureCluster has been persisted  when the defaulting webhook runs for the AzureMachine.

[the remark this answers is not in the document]
dthorsen: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2411/files#diff-b6931ac61f37a13f66fc7836e039bf714a0553d5db4f313cc8ef31ca5f5647b9R710-R713

controllers/azuremachine_reconciler.go
nawazkh: I don't think this code base follows "setting defaults in the reconciler" convention. @CecileRobertMichon What do you say?

dthorsen: @nawazkh the code already does this (predating this PR) here: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/controllers/azuremachine_reconciler.go#L76-L78

api/v1beta1/types.go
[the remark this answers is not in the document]
nawazkh: Makes sense to me. Would like to get @CecileRobertMichon thoughts too!

api/v1beta1/types.go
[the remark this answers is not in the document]
nawazkh: @CecileRobertMichon @jackfrancis What should be the behavior _"if AcceleratedNetworking is set to true with a VMSize that does not support it "_ ? 

Validating AcceleratedNetworking that via a webhook(if possible via Azure API) could be slow but can help user error out before persisting the object. What do you think?

api/v1beta1/azuremachine_default.go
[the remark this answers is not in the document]
nawazkh: Thanks for pinning this!

azure/defaults.go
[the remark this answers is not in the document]
CecileRobertMichon: fair point, let's leave it as-is

api/v1beta1/types.go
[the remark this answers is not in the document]
CecileRobertMichon: In some cases it makes sense to enforce Azure limits for things that are static and not susceptible to change (eg. naming) to accelerate the feedback loop but in this case it does seem likely that Azure will change/increase the max IP limit in the future so I'm +1 on deferring to Azure API for validation

api/v1beta1/types.go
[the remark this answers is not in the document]
CecileRobertMichon: here's the original PR that added the feature for context: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/645

I believe right now we just default to Azure and don't fail in the controller if a SKU that doesn't support AN has it explicitly enabled by the user. That could be a worthy improvement but out of scope for this PR IMO. Validating in the webhook would be more tricky as it requires an API call and Azure credentials. @nawazkh could you please file an issue so we can follow up?

controllers/azuremachine_reconciler.go
CecileRobertMichon: Agree, we have to do this to preserve backward compatibility

CecileRobertMichon:
@nawazkh thank you for the review 🙏 

@dthorsen happy new year! please let me know once all comments have been addressed and this is ready for a final pass

api/v1beta1/types.go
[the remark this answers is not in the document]
nawazkh: Created an issue to track ^ 
https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/2979

api/v1beta1/azuremachinetemplate_webhook.go
```diff
@@ -128,5 +132,6 @@ func (r *AzureMachineTemplate) Default(ctx context.Context, obj runtime.Object)
 	}
 	t.Spec.Template.Spec.SetDefaultCachingType()
 	t.Spec.Template.Spec.SetDataDisksDefaults()
+	t.Spec.Template.Spec.SetNetworkInterfacesDefaults()
```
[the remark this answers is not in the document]
dthorsen: azuremachinetemplate is immutable. And the new defaulting behavior should be permitted already through the immutability by the logic [here](https://github.com/newrelic-forks/cluster-api-provider-azure/blob/main/api/v1beta1/azuremachinetemplate_webhook.go#L102)

CecileRobertMichon (commented):
/lgtm

CecileRobertMichon:
@nawazkh are all your comments resolved?

If so @dthorsen can you please squash one more time?

dthorsen:
@nawazkh I believe I addressed all of your comments. Let me know if you have any other comments before I squash and push commits. Thanks!

api/v1beta1/azuremachinetemplate_webhook.go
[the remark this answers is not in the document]
nawazkh: Thanks for pinning that logic!

nawazkh (approved):
Mega job @dthorsen ! Thank you for patiently working on this and addressing all the nits 🙌🏼
/lgtm

nawazkh:
Persisting lgtm
/lgtm

CecileRobertMichon:
both of these test failures are known flakes getting worked on independently :( 

/retest
/approve

PR #2423, opened 2022-06-24
Bump to latest kubernetes patch version

jsturtevant:
**What type of PR is this?**
/kind cleanup
/kind flake

Seeing the ci-entrypoint.sh jobs fail on the PRs with windows timeouts.  The latest vm images for the patch releases of k8s has a fix in cloudbase-init that mitigates this (note it seem to improve but not full fix - follow https://github.com/cloudbase/cloudbase-init/issues/94 for latest)

Fixes #

This is not best fix I think.  We don't support release versions (only ci versions) in ci-entrypoint.sh but if we did then we could set this version in the test-infra job which would always pick the latest version: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/8040d124532067f2b933d18ec0b48a0b8fb5b263/scripts/ci-entrypoint.sh#L67-L75

Maybe the CI version is good enough since it doesn't change often but does open us up to breaking changes coming into releases.

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```


CecileRobertMichon:
> We don't support release versions (only ci versions) in ci-entrypoint.sh 

we talked about supporting "stable-1.22" type versions like we do in e2e, the only thing with that is that we'd have to either a) rely on the ci-version template to install the right released k8s version at runtime (possible but requires the extra kubeadm script foo) or b) have logic like we do in e2e that selects the latest patch we have an image for so we don't get broken the moment a new k8s version gets released that we haven't had time to publish an image for yet

For now this is good enough to mitigate

/lgtm
/approve

mboersma (approved):
/lgtm

jsturtevant:
job failed due to quota:

```
Code=\"OperationNotAllowed\" Message=\"Operation could not be completed as it results in exceeding approved standardDSv3Family Cores quota. Additional details - Deployment Model: Resource Manager, Location: centralus, Current Limit: 50, Current Usage: 50, Additional Required: 2, (Minimum) New Limit Required: 52. Submit a request for Quota increase at 
```
/retest

PR #2437, opened 2022-06-29
tests for conversion from previous API versions

### LINKED ISSUE #1365: Add tests for conversion from previous API versions
We should add similar tests for conversion webhooks than https://github.com/kubernetes-sigs/cluster-api/pull/4417 to make sure that v1alpha3 --> v1alpha4 conversion actually works for each CRD.

/help
/good-first-issue

sayantani11:
**What type of PR is this?**

Fixes #1365

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Test for conversion from API versions
```

### DIFF
--- a/api/v1alpha3/webhook_suite_test.go
+++ b/api/v1alpha3/webhook_suite_test.go
@@ -1,0 +1,60 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha3
+
+import (
+	"fmt"
+	"os"
+	"testing"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/cluster-api/test/helpers"
+	ctrl "sigs.k8s.io/controller-runtime"
+	// +kubebuilder:scaffold:imports
+)
+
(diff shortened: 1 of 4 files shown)


sayantani11:
> Thanks @sayantani11! Looks like you need to run `make modules` before `make verify` and `make test` could pass.
> 
> CAPZ stopped testing v1alpha3 in testgrid because it's no longer supported in general by Cluster API. What's the motivation in adding this test coverage?

It was just there in the backlog, so decided to just make the PR for it. I was thinking if we can have one for v1aplha4 -> v1beta1 but I guess CAPZ already has that if I am not wrong.

sayantani11:
It wasn't discouraging at all🙌🏻

CecileRobertMichon:
Even though we no longer support creating new v1alpha3 clusters, I still think it'd be valuable to have integration test coverage for v1alpha3 -> v1beta and v1alpha4 -> v1beta1 conversion since there could still be existing alpha clusters out there and we want to make sure we don't break conversion in the current releases. 

It would be valuable to also add the same test to v1alpha4, even though it looks like CAPI never added one for v1alpha4 https://github.com/prksu/cluster-api/tree/main/api/v1alpha4

sayantani11:
> Even though we no longer support creating new v1alpha3 clusters, I still think it'd be valuable to have integration test coverage for v1alpha3 -> v1beta and v1alpha4 -> v1beta1 conversion since there could still be existing alpha clusters out there and we want to make sure we don't break conversion in the current releases.
> 
> It would be valuable to also add the same test to v1alpha4, even though it looks like CAPI never added one for v1alpha4 https://github.com/prksu/cluster-api/tree/main/api/v1alpha4

Should I try adding the ones for v1Alpha3=>v1beta1 and v1Alpha4=>v1beta1?

mboersma:
> Should I try adding the ones for v1Alpha3=>v1beta1 and v1Alpha4=>v1beta1?

@sayantani11 if you're up for that, apparently it would be valuable. Could be a separate PR.

sayantani11:
> > Should I try adding the ones for v1Alpha3=>v1beta1 and v1Alpha4=>v1beta1?
> 
> @sayantani11 if you're up for that, apparently it would be valuable. Could be a separate PR.

Sure!!!

CecileRobertMichon:
/close

Doing some PR queue cleanup 🧹, please feel free to reopen when this is ready for review

PR #2439, opened 2022-06-29
Test: Removing healthcheck which interfers with cloudbase patch

jsturtevant:
**What type of PR is this?**
/kind cleanup

I've looked into a few recent flakes where the nodes don't come online for windows. I've found that the node was in the process of updating but then got hit by the health check.  This caused the nodes to get deleted and the tests to never fully provision.  

This occurred because the most recent patch we are using from cloudbase has a longer time out on the node provisioning which is interfering with the health check.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
none
```


jackfrancis (approved):
/lgtm
/approve

PR #2440, opened 2022-06-30
test: remotely retrieve VHD images once

jackfrancis:
**What type of PR is this?**

/kind cleanup

This PR ensures that we remotely pull for a list of regionally available images only once during an E2E test run. At present we pull both Linux and Windows images *four times* per run, once in the main `SynchronizedBeforeSuite` flow, and in addition once per node (ginkgo thread). In addition to causing unnecessary calls to the Azure APIs, we are slowing down our test runs by a long time (these pulls and data parsing take ~minutes!).

The E2E tests are currently implemented to build all clusters in a common region, so there is no reason to do "per-test-job image pulling".

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/e2e_suite_test.go
+++ b/test/e2e/e2e_suite_test.go
@@ -28,6 +28,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/blang/semver"
 	. "github.com/onsi/ginkgo"
 	"github.com/onsi/ginkgo/config"
 	"github.com/onsi/ginkgo/reporters"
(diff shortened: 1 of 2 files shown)


jackfrancis:
This is interesting, clearly the changeset is responsible for these test failures, but I can't reproduce them locally.

Does anything look suspicious to you here, @mboersma?

mboersma:
I'm not seeing the change in this PR that could cause those "invalid KubernetesVersion" errors, but obviously it broke something. I'll try locally.

jackfrancis:
@mboersma this must be what's going on... something about the fact that we're only pulling the versions once (instead of four) correlates a little too closely with the fact that we're only getting one successful cluster test (instead of four)

Issue #2443, opened 2022-07-01
When creating AKS clusters using autoscaler enabled, the update api is making too many requests


LochanRn:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]

Created an AKS cluster using autoscalar enabled. Initially the autoscalar worked fine, but after a while the capz controller started throwing the below error log.

`
E0701 18:40:03.669437       1 controller.go:304] controller/azuremanagedmachinepool "msg"="Reconciler error" "error"="error creating AzureManagedMachinePool cluster-62bd8e0fb040b7cb1d306dfa/workerpuwnk: failed to reconcile machine pool workerpuwnk: failed to create or update agent pool: failed to end operation: Code=\"ReconcileVMSSAgentPoolFailed\" Message=\"autorest/azure: Service returned an error. Status=429 Code=\\\"OperationNotAllowed\\\" Message=\\\"The server rejected the request because too many requests have been received for this subscription.\\\" Details=[{\\\"code\\\":\\\"TooManyRequests\\\",\\\"message\\\":\\\"{\\\\\\\"operationGroup\\\\\\\":\\\\\\\"HighCostGetVMScaleSet30Min\\\\\\\",\\\\\\\"startTime\\\\\\\":\\\\\\\"2022-07-01T18:39:49.9499033+00:00\\\\\\\",\\\\\\\"endTime\\\\\\\":\\\\\\\"2022-07-01T18:40:00+00:00\\\\\\\",\\\\\\\"allowedRequestCount\\\\\\\":971,\\\\\\\"measuredRequestCount\\\\\\\":1218}\\\",\\\"target\\\":\\\"HighCostGetVMScaleSet30Min\\\"}] InnerError={\\\"internalErrorCode\\\":\\\"TooManyRequestsReceived\\\"}\"" "name"="workerpuwnk" "namespace"="cluster-62bd8e0fb040b7cb1d306dfa" "reconciler group"="infrastructure.cluster.x-k8s.io" "reconciler kind"="AzureManagedMachinePool" 
`

Because of too many api calls, the cluster stops responding and becomes unstable.

On deeper investigation, the issue is because of the difference in the node count set. This diff is persistent and continuously triggers an update api call to the azure services.

```
Update required (+new -old):
   containerservice.AgentPool{
      Response: {},
      ManagedClusterAgentPoolProfileProperties: &containerservice.ManagedClusterAgentPoolProfileProperties{
[...]

PR #2444, opened 2022-07-01
When creating AKS clusters using autoscaler enabled, do not make an update api call to agentpool service based on difference in node count

LochanRn:
**What type of PR is this?**
/kind bug

Resolves https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/2443

Fixes https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/2443

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
None
```

### DIFF
--- a/azure/scope/managedmachinepool.go
+++ b/azure/scope/managedmachinepool.go
@@ -69,21 +69,28 @@ func NewManagedMachinePoolScope(ctx context.Context, params ManagedMachinePoolSc
 		return nil, errors.Wrap(err, "failed to init patch helper")
 	}
 
+	capiMachinePoolPatchHelper, err := patch.NewHelper(params.MachinePool, params.Client)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to init patch helper")
+	}
+
 	return &ManagedMachinePoolScope{
-		Client:               params.Client,
-		Cluster:              params.Cluster,
-		ControlPlane:         params.ControlPlane,
-		MachinePool:          params.MachinePool,
-		InfraMachinePool:     params.InfraMachinePool,
-		patchHelper:          helper,
-		ManagedClusterScoper: params.ManagedControlPlaneScope,
+		Client:                     params.Client,
+		Cluster:                    params.Cluster,
+		ControlPlane:               params.ControlPlane,
+		MachinePool:                params.MachinePool,
+		InfraMachinePool:           params.InfraMachinePool,
+		patchHelper:                helper,
+		capiMachinePoolPatchHelper: capiMachinePoolPatchHelper,
+		ManagedClusterScoper:       params.ManagedControlPlaneScope,
 	}, nil
 }
 
 // ManagedMachinePoolScope defines the basic context for an actuator to operate upon.
 type ManagedMachinePoolScope struct {
-	Client      client.Client
-	patchHelper *patch.Helper
+	Client                     client.Client
+	patchHelper                *patch.Helper
+	capiMachinePoolPatchHelper *patch.Helper
 
 	azure.ManagedClusterScoper
 	Cluster          *clusterv1.Cluster
@@ -231,7 +238,7 @@ func (s *ManagedMachinePoolScope) UpdateDeleteStatus(condition clusterv1.Conditi
 	}
 }
 
-// UpdatePutStatus updates a condition on the AzureManagedControlPlane status after a PUT operation.
+// UpdatePutStatus updates a condition on the AzureManagedMachinePool status after a PUT operation.
 func (s *ManagedMachinePoolScope) UpdatePutStatus(condition clusterv1.ConditionType, service string, err error) {
 	switch {
 	case err == nil:
@@ -243,7 +250,7 @@ func (s *ManagedMachinePoolScope) UpdatePutStatus(condition clusterv1.ConditionT
 	}
 }
 
-// UpdatePatchStatus updates a condition on the AzureManagedControlPlane status after a PATCH operation.
+// UpdatePatchStatus updates a condition on the AzureManagedMachinePool status after a PATCH operation.
 func (s *ManagedMachinePoolScope) UpdatePatchStatus(condition clusterv1.ConditionType, service string, err error) {
 	switch {
 	case err == nil:
@@ -253,4 +260,33 @@ func (s *ManagedMachinePoolScope) UpdatePatchStatus(condition clusterv1.Conditio
 	default:
 		conditions.MarkFalse(s.InfraMachinePool, condition, infrav1.FailedReason, clusterv1.ConditionSeverityError, "%s failed to update. err: %s", service, err.Error())
 	}
+}
+
+// PatchCAPIMachinePoolObject persists the capi machinepool configuration and status.
+func (s *ManagedMachinePoolScope) PatchCAPIMachinePoolObject(ctx context.Context) error {
+	return s.capiMachinePoolPatchHelper.Patch(
+		ctx,
+		s.MachinePool,
+	)
+}
+
+// UpdateCAPIMachinePoolReplicas updates the associated MachinePool replica count.
+func (s *ManagedMachinePoolScope) UpdateCAPIMachinePoolReplicas(replicas *int32) {
+	s.MachinePool.Spec.Replicas = replicas
+}
+
+// UpdateCAPIMachinePoolAnnotations updates the associated MachinePool annotation.
+func (s *ManagedMachinePoolScope) UpdateCAPIMachinePoolAnnotations(key, value string) {
+	s.MachinePool.Annotations[key] = value
+}
+
+// RemoveCAPIMachinePoolAnnotations removes the associated MachinePool annotation.
+func (s *ManagedMachinePoolScope) RemoveCAPIMachinePoolAnnotations(key string) {
+	delete(s.MachinePool.Annotations, key)
+}
+
+// GetCAPIMachinePoolAnnotations gets the associated MachinePool annotation.
+func (s *ManagedMachinePoolScope) GetCAPIMachinePoolAnnotation(key string) (bool, string) {
+	value, ok := s.MachinePool.Annotations[key]
+	return ok, value
 }
--- a/azure/services/agentpools/agentpools.go
+++ b/azure/services/agentpools/agentpools.go
@@ -22,6 +22,7 @@ import (
 	"time"
 
 	"github.com/Azure/azure-sdk-for-go/services/containerservice/mgmt/2021-05-01/containerservice"
+	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/google/go-cmp/cmp"
 	"github.com/pkg/errors"
 	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
@@ -43,6 +44,10 @@ type ManagedMachinePoolScope interface {
 	SetAgentPoolProviderIDList([]string)
 	SetAgentPoolReplicas(int32)
 	SetAgentPoolReady(bool)
+	UpdateCAPIMachinePoolReplicas(replicas *int32)
+	UpdateCAPIMachinePoolAnnotations(key, value string)
+	GetCAPIMachinePoolAnnotation(key string) (bool, string)
+	RemoveCAPIMachinePoolAnnotations(key string)
 }
 
 // Service provides operations on Azure resources.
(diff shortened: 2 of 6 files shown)


azure/services/agentpools/agentpools.go
```diff
@@ -125,6 +125,13 @@ func (s *Service) Reconcile(ctx context.Context) error {
 			},
 		}
 
+		// When autoscaling is set, the count of the nodes differ based on the autoscalar and should not depend on the
+		// count present in machinepool, azuremanagedmachinepool, hence we should not make an update api call based
+		// on difference in count.
+		if profile.EnableAutoScaling != nil && existingProfile.Count != nil {
```
devigned, line 131: This seems completely reasonable to me. 

However, I wonder if we could add a condition or something to let the user know what is going on. There are sure to be questions like, "Why is my cluster scaled up so high when I gave it the desired state of X?". I could imagine a user getting quite irritated by CAPZ ignoring their desired state.

devigned (commented):
Great catch and solution! Just a bit of feedback on how to surface the behavior to the operator.

azure/services/agentpools/agentpools.go
CecileRobertMichon: +1

I agree we should try to surface something observable to the user to let them know that the actual count is different as autoscaling is enabled. I think this is the same thing that https://github.com/kubernetes-sigs/cluster-api/pull/5685 is trying to solve.

CecileRobertMichon: also related to https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/2022

we should try to solve the issue consistently across providers as much as possible to ensure a consistent user experience across CAPI providers

LochanRn: After reading the above two threads, I feel along with the existing changes and the suggestions made above to surface something to the user to let know that the nodes have scaled up or scaled down because of the autoscaler we must also adjust the machinepool's desired replica count according to the node count present in cluster. 
In this way even the machinepools desired count and the phase spec will be accurate.

Jont828:
Hey thanks for making this change! I'm working on an agent pool service refactor in #2479. Are you planning on merging this PR soon or do you want to hold it and add additional changes to surface info to the user? If you plan on merging soon, I can hold my PR and rebase after it merges.

LochanRn:
> 

@Jont828 Sorry for the delay, will try to update the PR by tomorrow.

LochanRn:
@devigned @CecileRobertMichon can you please take a look at this PR again ?

azure/services/agentpools/agentpools.go
```diff
@@ -125,6 +128,19 @@ func (s *Service) Reconcile(ctx context.Context) error {
...
```
jackfrancis, line 134: I think this needs to be:

`if to.Bool(profile.EnableAutoScaling) && existingProfile.Count != nil {`

adding the `"github.com/Azure/go-autorest/autorest/to"` package to the existing import set in this file

Or this works without the imported library (but is more verbose):

`if profile.EnableAutoScaling != nil && if *profile.EnableAutoScaling && existingProfile.Count != nil {`

azure/services/agentpools/agentpools.go
```diff
@@ -125,6 +128,19 @@ func (s *Service) Reconcile(ctx context.Context) error {
 			},
 		}
 
+		// When autoscaling is set, the count of the nodes differ based on the autoscalar and should not depend on the
```
jackfrancis, line 131: nit: s/autoscalar/autoscaler

jackfrancis:
Overall lgtm, added some minor comments to address before final merging.

Thanks!

azure/services/agentpools/agentpools.go
LochanRn: ack

azure/scope/managedmachinepool.go
```diff
@@ -219,6 +219,16 @@ func (s *ManagedMachinePoolScope) DeleteLongRunningOperationState(name, service
 	futures.Delete(s.ControlPlane, name, service)
 }
 
+// UpdateMachinePoolReplicas updates capi machinepool replica count.
+func (s *ManagedMachinePoolScope) UpdateMachinePoolReplicas(replicas *int32) error {
+	patch := client.MergeFrom(s.MachinePool.DeepCopy())
```
jackfrancis, line 224: I'm unable to comment on whether or not this patch implementation is what we want, everything else lgtm in this PR.

Thanks again!

Maybe @CecileRobertMichon can give this a final thumbs up or offer feedback for improvement.

api/v1beta1/conditions_consts.go
```diff
@@ -82,6 +82,8 @@ const (
 	ManagedClusterRunningCondition clusterv1.ConditionType = "ManagedClusterRunning"
 	// AgentPoolsReadyCondition means the AKS agent pools exist and are ready to be used.
 	AgentPoolsReadyCondition clusterv1.ConditionType = "AgentPoolsReady"
+	// AutoScalerUpdatedMachinePoolsCondition means the Azure autoscaler scaledup or scaledDown the machinepools.
+	AutoScalerUpdatedMachinePoolReplicas clusterv1.ConditionType = "AutoScalerUpdatedMachinePoolReplicas"
```
CecileRobertMichon, line 86: I'm not sure if a Condition is the best way to signal this to the user. Cluster API condition conventions state:

> - Condition types MUST have a consistent polarity (i.e. "True = good");
>
> - Condition types SHOULD have one of the following suffix:
>    -  Ready, for resources which represent an ongoing status, like ControlplaneReady or MachineDeploymentsReady.
>    - Succeeded, for resources which run-to-completion, e.g. CreateVPCSucceeded
> 
> When the above suffix are not adequate for a specific condition type, other suffix with positive meaning COULD be used (e.g. Completed, Healthy); however, it is recommended to balance this flexibility with the objective to provide a consistent condition naming across all the Cluster API objects


In this case `AutoScalerUpdatedMachinePoolReplicas` isn't really a good/bad type of condition. In fact, we're never setting the condition to False.

How about adding an annotation to the MachinePool directly (as opposed to a condition on the AzureManagedMachinePool object) that indicates to the user that the MachinePool replica count is being managed by cluster-autoscaler (similar idea to https://github.com/kubernetes-sigs/cluster-api/pull/5685). That way it's directly on the object that the user might go modify to edit the replica count which is being ignored.

@Jont828 @jackfrancis thoughts?

azure/scope/managedmachinepool.go
```diff
@@ -219,6 +219,16 @@ func (s *ManagedMachinePoolScope) DeleteLongRunningOperationState(name, service
...
+	s.MachinePool.Spec.Replicas = replicas
+	if err := s.Client.Patch(context.Background(), s.MachinePool, patch); err != nil {
```
CecileRobertMichon, line 226: we shouldn't use `context.Background()` here, we have an actual context we can pass in from the parent func

CecileRobertMichon, line 226: We should use `patchHelper.Patch()` here

azure/scope/managedmachinepool.go
```diff
@@ -219,6 +219,16 @@ func (s *ManagedMachinePoolScope) DeleteLongRunningOperationState(name, service
 	futures.Delete(s.ControlPlane, name, service)
 }
 
+// UpdateMachinePoolReplicas updates capi machinepool replica count.
```
CecileRobertMichon, line 222:
```suggestion
// UpdateMachinePoolReplicas updates the associated MachinePool replica count.
```

azure/scope/managedmachinepool.go
jackfrancis: I assume you mean that the calling func (`func (s *Service) Reconcile`) has its own context object, and we should pass it along to `UpdateMachinePoolReplicas` here. I agree.

So we'd update the interface accordingly:

`UpdateMachinePoolReplicas(ctx context.Context, replicas *int32) error`

And then the actual method definition and its invocation in the flow of `Reconcile()` w/ the updated signature.

azure/scope/managedmachinepool.go
LochanRn: ack

azure/scope/managedmachinepool.go
LochanRn: ack

api/v1beta1/conditions_consts.go
jackfrancis: I like the idea of an annotation on the `MachinePool` (sort of mimics the comments masthead on generated code: // DO NOT MODIFY, WILL BE OVERWRITTEN!!!)

So the above would be something we'd continually reconcile so long as autoscaling was set to true on the underlying AKS node pool.

And then, additionally, I think a v2 info logging statement would be apropos for every time we inherit a different value from AKS and then update capi/capz accordingly. Is that log statement already happening somewhere in the flow as a result of the update, or do we need to add that additional verbosity?

LochanRn: Need some help in naming the annotation.
Is this ok for the same. ?
`sigs.k8s.io/autoscaler-is-managing-machinepool-replica: true`

jackfrancis: An annotation prefix that comes up a lot in the kubernetes/autoscaler project is `cluster-autoscaler.kubernetes.io/`

How about `cluster-api.cluster-autoscaler.kubernetes.io/replicas: true`?

CecileRobertMichon: my 2 cents: replicas is typically a number and it seems counter-intuitive to use it as a boolean. We should also not be using the autoscaler annotation prefix since this annotation is not actually set by the kubernetes/autoscaler controllers.

How about

`cluster.x-k8s.io/replicas-managed-by-autoscaler: true`

(`cluster.x-k8s.io` is the same prefix used by all CAPI annotations)

CecileRobertMichon: also we should probably align with whatever annotation CAPI decides to go with in https://github.com/kubernetes-sigs/cluster-api/pull/6991

One POC used `cluster.x-k8s.io/externally-managed-replicas`

LochanRn:
@CecileRobertMichon @jackfrancis updated the PR with new changes, could you please review it, I need to still test it though. 

@Jont828 not sure if this PR can be merged soon, if this is a blocker or keeping your PR waiting please go ahead and merge your PR I will rebase :)

CecileRobertMichon:
@LochanRn sorry for the delay on this, would you be able to rebase so we can merge it?

LochanRn:
@CecileRobertMichon @jackfrancis can you please make a final review on this :)

CecileRobertMichon (commented):
code lgtm, will let @jackfrancis confirm he was able to test this and it's ready to go

/lgtm

azure/services/agentpools/agentpools.go
```diff
@@ -127,6 +131,25 @@ func (s *Service) Reconcile(ctx context.Context) error {
 			},
 		}
 
+		// When autoscaling is set, the count of the nodes differ based on the autoscaler and should not depend on the
+		// count present in MachinePool or AzureManagedMachinePool, hence we should not make an update API call based
+		// on difference in count.
+		if to.Bool(profile.EnableAutoScaling) && existingProfile.Count != nil {
+			if to.Bool(profile.EnableAutoScaling) && s.scope.GetCAPIMachinePoolAnnotation(ctx, azure.ReplicasManagedByAutoscalerAnnotation) != "true" {
```
jackfrancis, line 138: We can remove this `if to.Bool(profile.EnableAutoScaling)` condition inside this block, as that condition is already declared in the parent block.

azure/services/agentpools/agentpools.go
```diff
@@ -127,6 +131,25 @@ func (s *Service) Reconcile(ctx context.Context) error {
...
+
+			if to.Int32(existingProfile.Count) != to.Int32(normalizedProfile.Count) {
+				s.scope.UpdateCAPIMachinePoolReplicas(ctx, existingProfile.Count)
+			}
+			normalizedProfile.Count = existingProfile.Count
+		}
+
+		// set ReplicasManagedByAutoscalerAnnotation to false as it is disabled by the user.
+		if !to.Bool(profile.EnableAutoScaling) && s.scope.GetCAPIMachinePoolAnnotation(ctx, azure.ReplicasManagedByAutoscalerAnnotation) == "true" {
+			s.scope.UpdateCAPIMachinePoolAnnotations(ctx, azure.ReplicasManagedByAutoscalerAnnotation, "false")
```
CecileRobertMichon, line 150: I realized while reviewing https://github.com/kubernetes-sigs/cluster-api/pull/7107/files#r952041386 we might not want to set a "false" value but instead remove the annotation altogether.

In any case this should be consistent with what https://github.com/kubernetes-sigs/cluster-api/pull/7107 ends up doing.

jackfrancis (approved):
/lgtm
/approve

thank you @LochanRn!

LochanRn:
> /lgtm /approve
> 
> thank you @LochanRn!

yay thank you as well :)

Issue #1237, opened 2021-03-15
NMI fails to run on hosts which have no /etc/default/kubelet


invidian:
/kind bug

**What steps did you take and what happened:**
1. Setup CAPZ with Tilt for development on cluster with no `/etc/default/kubelet` file.
2. Tilt never converges with NMI pod in crash loopback with following logs:
```console
$ klo capz-nmi-lc4g7
+ kubectl logs -f capz-nmi-lc4g7
F0312 16:12:33.667869       1 main.go:78] failed to check if CNI plugin is kubenet, error: read /etc/default/kubelet: is a directory
goroutine 1 [running]:
k8s.io/klog/v2.stacks(0xc0000a6001, 0xc0000ce0e0, 0x85, 0xda)
        /go/pkg/mod/k8s.io/klog/v2@v2.3.0/klog.go:996 +0xb9
k8s.io/klog/v2.(*loggingT).output(0x2188a40, 0xc000000003, 0x0, 0x0, 0xc00044b960, 0x20ebf7e, 0x7, 0x4e, 0x0)
        /go/pkg/mod/k8s.io/klog/v2@v2.3.0/klog.go:945 +0x191
k8s.io/klog/v2.(*loggingT).printf(0x2188a40, 0x3, 0x0, 0x0, 0x16cce4a, 0x34, 0xc0005bfea0, 0x1, 0x1)
        /go/pkg/mod/k8s.io/klog/v2@v2.3.0/klog.go:733 +0x17a
k8s.io/klog/v2.Fatalf(...)
        /go/pkg/mod/k8s.io/klog/v2@v2.3.0/klog.go:1463
main.main()
        /go/src/github.com/Azure/aad-pod-identity/cmd/nmi/main.go:78 +0x29b
goroutine 5 [chan receive]:
k8s.io/klog/v2.(*loggingT).flushDaemon(0x2188a40)
        /go/pkg/mod/k8s.io/klog/v2@v2.3.0/klog.go:1131 +0x8b
created by k8s.io/klog/v2.init.0
        /go/pkg/mod/k8s.io/klog/v2@v2.3.0/klog.go:416 +0xd8
goroutine 6 [select]:
go.opencensus.io/stats/view.(*worker).start(0xc00003a140)
        /go/pkg/mod/go.opencensus.io@v0.22.2/stats/view/worker.go:154 +0x105
created by go.opencensus.io/stats/view.init.0
[...]

PR #2446, opened 2022-07-01
Adding a DirectoryOrCreate to /etc/default in the AAD DaemonSet

luthermonson:
**What type of PR is this?**
/kind cleanup

The AAD daemonset has a hostpath mount for `/etc/default/kubelet` which is as far as I can tell is a file which is used in this function https://github.com/Azure/aad-pod-identity/blob/master/pkg/utils/utils.go#L30-L38 to determine if kubenet is being used.

The [suggested](https://kubernetes.io/docs/concepts/storage/volumes/#hostpath-fileorcreate-example) way to do FileOrCreate is to also put a DirectoryOrCreate on the paths leading up to the file so if the host doesn't have the `/etc/default` path it will create that too. Currently if you are using a kube cluster which doesn't generate this file on the host the AAD pod will fail to deploy with volume mount errors.

Fixes #

to test just build a cluster not built by kind and try and deploy capz into it, i was using k3d and this manifest changed fixed my issues.
 
**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Add DirectoryOrCreate to /etc/default for aad-pod-identity
```

### DIFF
--- a/config/default/aad-pod-identity-deployment.yaml
+++ b/config/default/aad-pod-identity-deployment.yaml
@@ -271,6 +271,10 @@ spec:
           path: /run/xtables.lock
           type: FileOrCreate
         name: iptableslock
+      - name: default-path
+        hostPath:
+          path: /etc/default
+          type: DirectoryOrCreate
       - name: kubelet-config
         hostPath:
           path: /etc/default/kubelet
@@ -316,6 +320,9 @@ spec:
         volumeMounts:
         - mountPath: /run/xtables.lock
           name: iptableslock
+        - name: default-path
+          mountPath: /etc/default
+          readOnly: true
         - name: kubelet-config
           mountPath: /etc/default/kubelet
           readOnly: true


CecileRobertMichon:
@aramase thoughts on this one? The example manifests for AAD pod ID have the same thing since we added it around the same time https://github.com/Azure/aad-pod-identity/issues/1002

In https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1237#issuecomment-800484033 you mentioned a `-allow-network-plugin-kubenet` flag would be added in the future, is it supported now and if yes would that be a more appropriate fix here?

aramase:
> In https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1237#issuecomment-800484033 you mentioned a -allow-network-plugin-kubenet flag would be added in the future, is it supported now and if yes would that be a more appropriate fix here?

Skipping check if `--allow-network-plugin-kubenet=true` was added in the follow-up [PR](https://github.com/Azure/aad-pod-identity/pull/999). If the flag is true, then NMI skips the volume check. If this is with kubenet, you can just enable the flag and don't require these volumes. 

ref doc: https://azure.github.io/aad-pod-identity/docs/configure/aad_pod_identity_on_kubenet/#how-to-run-aad-pod-identity-on-clusters-with-kubenet for the flag.

luthermonson:
@CecileRobertMichon @aramase https://azure.github.io/aad-pod-identity/docs/configure/aad_pod_identity_on_kubenet/#why-this-change the behavior is to not allow AAD to run in a kubenet enabled cluster. if someone runs capz in aks with kubenet and we default this to true won't we ignore the [check](https://github.com/Azure/aad-pod-identity/blob/master/cmd/nmi/main.go#L85-L87) and run AAD in a state susceptible to the ARP spoofing attack mentioned? i think im against defaulting this setting to true but if that's what you want i can update the PR

CecileRobertMichon:
@luthermonson I agree we shouldn't skip the check.

Trying to understand your use case better

> Currently if you are using a kube cluster which doesn't generate this file on the host the AAD pod will fail to deploy with volume mount errors.

in what situation are you running into this? Is it a kind cluster? is it an AKS kubenet cluster? something else?

luthermonson:
im testing it locally using https://github.com/k3d-io/k3d which boots k3s clusters in docker locally. in theory any distro that doesn't make /etc/default/kubelet on the host will fail so rke1 and rke2 probably would fail too just from my experience working on those distros

CecileRobertMichon:
Got it, thanks for clarifying @luthermonson 

/lgtm

can you please add a release note?

/assign @jackfrancis for approval

jackfrancis (approved):
/approve

jackfrancis:
/release-note-edit ```Add DirectoryOrCreate to /etc/default for aad-pod-identity```

PR #2450, opened 2022-07-05
Makefile: put Calico manifest into artifacts directory and cache them

### LINKED ISSUE #2449: Don't download Calico release manifest on every "make generate-flavors" execution
/kind feature

The manifest is 1GB in size and there is no point downloading it on every run. Likely required manifests should be saved to some hidden working directory (e.g. existing `_artifacts`?) and release archive should only be downloaded when expected files are missing from there.

Make should be able to "cache" the results and trigger re-download if Calico version changes.

This will also prevent polluting root project directory with ambiguous directories like `release-v3.23.0`.

invidian:
**What type of PR is this?**

/kind cleanup

Right now running targets like 'generate' or 'generate-flavors' which
are often used when adding new flavors, downloads Calico release
manifests each time and they are 1GB in size, which is slow and
wasteful.

To speed things up and save bandwidth, this commit puts unpacked Calico
manifests in _artifacts directory as a build cache and uses Make file
target feature to only re-download the manifests if the files are
missing, which should only occur on fresh repository clone or when
Calico version changes, which then justifies re-downloading the archive.

To avoid downloading Calico archive multiple times when make targets are
executed in parallel (-j), we unpack entire manifests directory instead
of just desired manifests, which also makes the process simpler overall.

The manifests directory is right now <5MB, which is a lot (it's mostly
whitespace!) but I think acceptable.

Fixes #2449

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -476,13 +476,22 @@ generate-addons: fetch-calico-manifests ## Generate metric-server, calico calico
 
 # When updating this, make sure to also update the Windows image version in templates/addons/windows/calico.
 CALICO_VERSION := v3.23.0
+# Where all downloaded Calico manifests are unpacked and stored.
+CALICO_RELEASES := $(ARTIFACTS)/calico
+# Path to manifests directory in a Calico release archive.
+CALICO_RELEASE_MANIFESTS_DIR := release-$(CALICO_VERSION)/manifests
+# Path where Calico manifests are stored which should be used for addons generation.
+CALICO_MANIFESTS_DIR := $(ARTIFACTS)/calico/$(CALICO_RELEASE_MANIFESTS_DIR)
 
 .PHONY: fetch-calico-manifests
-fetch-calico-manifests: ## Get Calico release manifests and unzip them.
+fetch-calico-manifests: $(CALICO_MANIFESTS_DIR) ## Get Calico release manifests and unzip them.
+	cp $(CALICO_MANIFESTS_DIR)/calico-vxlan.yaml $(ADDONS_DIR)/calico
+	cp $(CALICO_MANIFESTS_DIR)/calico-policy-only.yaml $(ADDONS_DIR)/calico-ipv6
+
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -476,13 +476,21 @@ generate-addons: fetch-calico-manifests ## Generate metric-server, calico calico
...
 CALICO_VERSION := v3.23.0
+CALICO_RELEASES := $(ARTIFACTS)/calico
+CALICO_MANIFEST_PATH := release-$(CALICO_VERSION)/manifests/calico-vxlan.yaml
+CALICO_MANIFEST_IPV6_PATH := release-$(CALICO_VERSION)/manifests/calico-policy-only.yaml
+CALICO_MANIFEST := $(CALICO_RELEASES)/$(CALICO_MANIFEST_PATH)
+CALICO_MANIFEST_IPV6 := $(CALICO_RELEASES)/$(CALICO_MANIFEST_IPV6_PATH)
 
 .PHONY: fetch-calico-manifests
-fetch-calico-manifests: ## Get Calico release manifests and unzip them.
+fetch-calico-manifests: $(CALICO_MANIFEST) $(CALICO_MANIFEST_IPV6) ## Get Calico release manifests and unzip them.
```
invidian, line 486: I'll actually tune it a bit so it plays well when running targets in parallel.

invidian: Done. As an alternative, I could make one target for regular manifests and then IPv6 target could depend on this one, however that seem a bit more hacky to me.

CecileRobertMichon:
Thanks @invidian! Downloading the calico release artifacts to get the manifests should hopefully be temporary (as soon as v.3.24.0 is out we can use the manifests URI directly for kustomize https://github.com/projectcalico/calico/issues/2587#issuecomment-1154498990, and longer-term we want to get rid of the manifests all together and use Helm charts instead #2334) but this is a better solution for now.

/lgtm
/approve

PR #2374, opened 2022-06-09
Use setup-envtest instead of fetch_ext_bins

evanfreed:
**What type of PR is this?**

**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #1990

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

/kind feature

Signed-off-by: Evan Freed <evan.freed@protonmail.ch>

PR #2452, opened 2022-07-05
Revert "Use setup-envtest instead of fetch_ext_bins"

CecileRobertMichon:
Reverts kubernetes-sigs/cluster-api-provider-azure#2374

> This will need to be reverted now. As I noted above, this doesn’t work as-is on Apple Silicon. I apologize—I should have put a hold on it but assumed comments were read. I don’t see that anyone has tested it successfully on an M1.

https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2374#issuecomment-1175595483

cc @evanfreed @mboersma

### DIFF
--- a/scripts/fetch_ext_bins.sh
+++ b/scripts/fetch_ext_bins.sh
@@ -1,0 +1,94 @@
+#!/usr/bin/env bash
+# Copyright 2018 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


scripts/fetch_ext_bins.sh
```diff
@@ -0,0 +1,94 @@
...
+# environment to any value:
+#
+# $ TRACE=1 test.sh
+TRACE=${TRACE:-""}
+if [[ -n "${TRACE}" ]]; then
+  set -x
+fi
+
+k8s_version=1.20.2
+goarch=amd64
```
CecileRobertMichon, line 30: @mboersma seems like goarch is hardcoded here too, curious how this worked for you before the PR landed

CecileRobertMichon:
/hold

I might have a better solution

CecileRobertMichon:
I think https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2453 should do the trick

jackfrancis:
#2453 merged

PR #2456, opened 2022-07-06
add e2e for csi migration upgrade scenarios

### LINKED ISSUE #2242: e2e test for k8s v1.22 --> v1.23 upgrade to use out of tree csi driver
/kind bug

From Kubernetes v1.23 AzureDiskCSI driver is enabled by default. 

An e2e test with following validations should be there: 

- Create a k8s cluster with v1.22.
- Create a deployment/sts that uses PVC created via in-tree AzureDiskCSI driver.
- Upgrade the k8s cluster to v1.23 and install out-of-tree AzureDiskCSI driver.
- Validate that the existing pod using PVC created via in-tree provider keeps functioning.
- Validate that a new pod using PVC created via the out-of-tree provider works. 

**Environment:**

sonasingh46:

 

 
**What type of PR is this?**
e2e tests

/kind feature

With Kubernetes 1.23, CSI migration for azure disk is enabled by default and is in beta in 1.23 ( GA in 1.24 ). 
This means that one can either opt to go with the default migration in 1.23 or not go with it by explicitly disabling it. Thus we can have following upgrade paths for 1.22 to 1.23. This PR validates the upgrade paths by adding e2e. 

So  an upgrade from, let us say from k8s v1.22 to v1.23 can mean any of the following combination: 

S.No | AzureDiskCSI | CloudProviderAzure | AzureDiskCSIMigration 
--- | --- | --- | ---
1 | In-tree | In-tree | Disabled 
2 | External | In-tree | Enabled 
3 | External  | External  | Enabled 
 
 For the last combination in the table mentioned ( i.e 3 ) `external-cloud-volume-plugin: azure` flag needs to be set on controller manager for old Kubelets to keep talking to in-tree CCM and upgrade fails without this is set.
 
Fixes #2242 

A next PR will be raised to document the flags to be used for different scenarios in case of upgrade from v1.22 to v.123
Adding reference of a similar PR in capa 
https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3267/files
**TODOs**:

- [x ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

```release-note
None
```

### DIFF
--- a/templates/test/ci/patches/azurediskcsi-migration-off.yaml
+++ b/templates/test/ci/patches/azurediskcsi-migration-off.yaml
@@ -1,0 +1,45 @@
+---
+apiVersion: cluster.x-k8s.io/v1beta1
+kind: Cluster
+metadata:
+  name: ${CLUSTER_NAME}
+  labels:
+    ccm: "intree"
+    csi: "intree"
+---
+kind: KubeadmControlPlane
+apiVersion: controlplane.cluster.x-k8s.io/v1beta1
+metadata:
+  name: "${CLUSTER_NAME}-control-plane"
+spec:
+  kubeadmConfigSpec:
+    initConfiguration:
+      nodeRegistration:
+        kubeletExtraArgs:
+          feature-gates: CSIMigrationAzureDisk=false
+          azure-container-registry-config: /etc/kubernetes/azure.json
+    joinConfiguration:
+      nodeRegistration:
+        kubeletExtraArgs:
+          feature-gates: CSIMigrationAzureDisk=false
+          azure-container-registry-config: /etc/kubernetes/azure.json
+    clusterConfiguration:
+      apiServer:
+        timeoutForControlPlane: 20m
+      controllerManager:
+        extraArgs:
+          feature-gates: CSIMigrationAzureDisk=false
+  version: "${KUBERNETES_VERSION}"
+---
+apiVersion: bootstrap.cluster.x-k8s.io/v1beta1
+kind: KubeadmConfigTemplate
+metadata:
+  name: "${CLUSTER_NAME}-md-0"
+spec:
+  template:
+    spec:
+      joinConfiguration:
+        nodeRegistration:
+          kubeletExtraArgs:
+            feature-gates: CSIMigrationAzureDisk=false
+            azure-container-registry-config: /etc/kubernetes/azure.json
--- a/templates/test/ci/prow-azurediskcsi-migration-off/kustomization.yaml
+++ b/templates/test/ci/prow-azurediskcsi-migration-off/kustomization.yaml
@@ -1,0 +1,8 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: default
+resources:
+  - ../prow-user-assigned-managed-identity
+patchesStrategicMerge:
+  - ../patches/azurediskcsi-migration-off.yaml
+
--- a/templates/test/ci/prow-external-azurediskcsi-driver/kustomization.yaml
+++ b/templates/test/ci/prow-external-azurediskcsi-driver/kustomization.yaml
@@ -1,0 +1,7 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: default
+resources:
+  - ../prow-user-assigned-managed-identity
+patchesStrategicMerge:
+  - ./patches/external-azurediskcsi-driver.yaml
--- a/templates/test/ci/prow-user-assigned-managed-identity/kustomization.yaml
+++ b/templates/test/ci/prow-user-assigned-managed-identity/kustomization.yaml
@@ -1,0 +1,23 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: default
+resources:
+  - ../../../flavors/default
+  - ../prow/cni-resource-set.yaml
+patchesStrategicMerge:
+  - ../patches/tags.yaml
+  - ../patches/controller-manager.yaml
+  - ../patches/cluster-cni.yaml
+  - ../patches/apiserver.yaml
+  - ../patches/uami-md-0.yaml
+  - ../patches/uami-control-plane.yaml
+configMapGenerator:
+  - name: cni-${CLUSTER_NAME}-calico
+    files:
+      - resources=../../../addons/calico.yaml
+generatorOptions:
+  disableNameSuffixHash: true
+  labels:
+    type: generated
+  annotations:
+    note: generated
--- a/test/e2e/azure_csidriver.go
+++ b/test/e2e/azure_csidriver.go
@@ -24,7 +24,9 @@ import (
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/kubernetes"
 	deploymentBuilder "sigs.k8s.io/cluster-api-provider-azure/test/e2e/kubernetes/deployment"
 	e2e_pvc "sigs.k8s.io/cluster-api-provider-azure/test/e2e/kubernetes/pvc"
 	e2e_sc "sigs.k8s.io/cluster-api-provider-azure/test/e2e/kubernetes/storageclass"
@@ -32,16 +34,21 @@ import (
 	"sigs.k8s.io/cluster-api/util"
 )
 
+const (
+	DriverTypeInternal = "Internal"
+)
+
 type AzureDiskCSISpecInput struct {
 	BootstrapClusterProxy framework.ClusterProxy
 	Namespace             *corev1.Namespace
 	ClusterName           string
 	SkipCleanup           bool
+	DriverType            string
 }
 
 // AzureDiskCSISpec implements a test that verifies out of tree azure disk csi driver
 // can be used to create a PVC that is usable by a pod.
-func AzureDiskCSISpec(ctx context.Context, inputGetter func() AzureDiskCSISpecInput) {
+func AzureDiskCSISpec(ctx context.Context, inputGetter func() AzureDiskCSISpecInput) *appsv1.Deployment {
 	specName := "azurediskcsi-driver"
 	input := inputGetter()
 	Expect(input.BootstrapClusterProxy).NotTo(BeNil(), "Invalid argument. input.BootstrapClusterProxy can't be nil when calling %s spec", specName)
@@ -52,21 +59,51 @@ func AzureDiskCSISpec(ctx context.Context, inputGetter func() AzureDiskCSISpecIn
 	clientset := clusterProxy.GetClientSet()
 	Expect(clientset).NotTo(BeNil())
 
-	By("Deploying managed disk storage class")
-	e2e_sc.Create("managedhdd").WithWaitForFirstConsumer().DeployStorageClass(clientset)
-
-	By("Deploying persistent volume claim")
-	b, err := e2e_pvc.Create("dd-managed-hdd-5g", "5Gi")
-	Expect(err).To(BeNil())
-	b.DeployPVC(clientset)
+	var pvcName string
+	if input.DriverType == DriverTypeInternal {
+		By("[In-tree]Deploying storage class and pvc")
+		By("Deploying managed disk storage class")
+		scName := "managedhdd" + util.RandomString(6)
+		e2e_sc.Create(scName).WithWaitForFirstConsumer().DeployStorageClass(clientset)
+		By("Deploying persistent volume claim")
+		pvcName = "dd-managed-hdd-5g" + util.RandomString(6)
+		pvcBuilder, err := e2e_pvc.Create(pvcName, "5Gi")
+		Expect(err).To(BeNil())
+		annotations := map[string]string{
+			"volume.beta.kubernetes.io/storage-class": scName,
+		}
+		pvcBuilder.WithAnnotations(annotations)
+		err = pvcBuilder.DeployPVC(clientset)
+		Expect(err).To(BeNil())
+	} else {
+		By("[External]Deploying storage class and pvc")
+		By("Deploying managed disk storage class")
+		scName := "oot-managedhdd" + util.RandomString(6)
+		e2e_sc.Create(scName).WithWaitForFirstConsumer().
+			WithOotProvisionerName().
+			WithOotParameters().
+			DeployStorageClass(clientset)
+		By("Deploying persistent volume claim")
+		pvcName = "oot-dd-managed-hdd-5g" + util.RandomString(6)
+		pvcBuilder, err := e2e_pvc.Create(pvcName, "5Gi")
+		Expect(err).To(BeNil())
+		pvcBuilder.WithStorageClass(scName)
+		err = pvcBuilder.DeployPVC(clientset)
+		Expect(err).To(BeNil())
+	}
 
 	By("creating a deployment that uses pvc")
 	deploymentName := "stateful" + util.RandomString(6)
-	statefulDeployment := deploymentBuilder.Create("nginx", deploymentName, corev1.NamespaceDefault).AddPVC("dd-managed-hdd-5g")
+	statefulDeployment := deploymentBuilder.Create("nginx", deploymentName, corev1.NamespaceDefault).AddPVC(pvcName)
 	deployment, err := statefulDeployment.Deploy(ctx, clientset)
 	Expect(err).NotTo(HaveOccurred())
+	waitForDeploymentAvailable(ctx, deployment, clientset, specName)
+	return deployment
+}
+
+func waitForDeploymentAvailable(ctx context.Context, deployment *appsv1.Deployment, clientset *kubernetes.Clientset, specName string) {
 	deployInput := WaitForDeploymentsAvailableInput{
-		Getter:     deploymentsClientAdapter{client: statefulDeployment.Client(clientset)},
+		Getter:     deploymentsClientAdapter{client: clientset.AppsV1().Deployments(deployment.Namespace)},
 		Deployment: deployment,
 		Clientset:  clientset,
 	}
--- a/test/e2e/config/azure-dev.yaml
+++ b/test/e2e/config/azure-dev.yaml
@@ -127,6 +127,12 @@ providers:
         targetName: "cluster-template-conformance-presubmit-artifacts-windows-containerd.yaml"
       - sourcePath: "${PWD}/templates/test/ci/cluster-template-prow-external-cloud-provider.yaml"
         targetName: "cluster-template-external-cloud-provider.yaml"
+      - sourcePath: "${PWD}/templates/test/ci/cluster-template-prow-azurediskcsi-migration-off.yaml"
+        targetName: "cluster-template-azurediskcsi-migration-off.yaml"
+      - sourcePath: "${PWD}/templates/test/ci/cluster-template-prow-external-azurediskcsi-driver.yaml"
+        targetName: "cluster-template-external-azurediskcsi-driver.yaml"
+      - sourcePath: "${PWD}/templates/test/ci/cluster-template-prow-user-assigned-managed-identity.yaml"
+        targetName: "cluster-template-user-assigned-managed-identity.yaml"
       - sourcePath: "${PWD}/templates/test/ci/cluster-template-prow-aks.yaml"
         targetName: "cluster-template-aks.yaml"
       - sourcePath: "${PWD}/templates/test/ci/cluster-template-prow-custom-vnet.yaml"
@@ -190,4 +196,8 @@ intervals:
   default/wait-job: ["5m", "10s"]
   default/wait-service: ["15m", "10s"]
   default/wait-machine-pool-nodes: ["30m", "10s"]
+  csi-migration/wait-controlplane-upgrade: ["60m", "10s"]
+  csi-migration/wait-worker-nodes: ["60m", "10s"]
+  csi-migration/wait-control-plane: ["60m", "10s"]
+  csi-migration/wait-cluster: ["60m", "10s"]
   node-drain/wait-machine-deleted: [ "10m", "10s" ]
--- a/test/e2e/csi_migration_test.go
+++ b/test/e2e/csi_migration_test.go
@@ -1,0 +1,344 @@
+//go:build e2e
+// +build e2e
+
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package e2e
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"strings"
+	"time"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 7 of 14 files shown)


test/e2e/azure_test.go
```diff
@@ -603,4 +604,192 @@ var _ = Describe("Workload cluster creation", func() {
...
 		})
 	})
+	Context("CSI=internal CCM=internal AzureDiskCSIMigration=false: upgrade to v1.23 [OPTIONAL]", func() {
+		It("should create volumes dynamically with intree cloud provider", func() {
+			By("Creating workload cluster v1.22 using user-assigned identity")
+			clusterName = getClusterName(clusterNamePrefix, "intree-providers")
+			configCluster := defaultConfigCluster(clusterName, namespace.Name)
+			configCluster.KubernetesVersion = e2eConfig.GetVariable(PreCSIKubernetesVer)
+			configCluster.WorkerMachineCount = pointer.Int64Ptr(1)
+			configCluster.Flavor = "user-assigned-managed-identity"
```
jackfrancis, line 614: Can we ensure that the cluster we upgrade has 3 control plane nodes so we have test coverage for the HA scenario?

sonasingh46: Sure. Will do. 
Out of curiosity -- What is the HA scenario here? Like these three tests don't have any.

jackfrancis: 3 control plane nodes

The most common way that we do that now is by setting this in the `ConfigCluster` of the test: `ControlPlaneMachineCount: pointer.Int64Ptr(3),`

sonasingh46: I see. I set the nodes to 1 intentionally as I do not know what is the benefit of having 3 control plane nodes. 
I am trying to ask myself : If I set the node count to 3, what is that extra thing( if HA, what is it) that gets validated by this upgrade e2e?

sonasingh46: Sure. Thanks. I will update the PR to have 3 CP nodes when we upgrade.

test/e2e/azure_test.go
```diff
@@ -621,4 +622,192 @@ var _ = Describe("Workload cluster creation", func() {
 
 		})
 	})
+	Context("CSI=internal CCM=internal AzureDiskCSIMigration=false: upgrade to v1.23 [OPTIONAL]", func() {
```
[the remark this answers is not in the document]
CecileRobertMichon: +1 to adding a separate `[CSI MIGRATION]` focus with its own job so it's easier to maintain, and for running it in its own job.

I also think we should move this out of the azure_test workload cluster creation describe spec and maybe put it in its own spec/file for CSI migration upgrade.

This especially makes sense given that we expect this test job to be somewhat short-lived (after a few k8s versions of CSI migration being GA we might want to remove these tests) and keeping it separate will make it easier to remove it.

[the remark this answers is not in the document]
sonasingh46: That make sense. Thank you. 
So : 
- Will add a "[CSI Migration]" string in the test description.
- Move the code in a different file. 
- To make it run in its own job -- do we need to raise a PR in test-infra?

sonasingh46:
The test infra PR: 
https://github.com/kubernetes/test-infra/pull/26993

templates/test/ci/prow-external-azurediskcsi-driver/kustomization.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: default
+resources:
+  - ../prow-user-assigned-managed-identity
+patchesStrategicMerge:
+  - ./patches/external-azurediskcsi-driver.yaml
```
Jont828, line 7: You might want to add a newline so the linter doesn't complain

templates/test/ci/patches/azurediskcsi-migration-off.yaml
```diff
@@ -0,0 +1,45 @@
...
+metadata:
+  name: "${CLUSTER_NAME}-md-0"
+spec:
+  template:
+    spec:
+      joinConfiguration:
+        nodeRegistration:
+          kubeletExtraArgs:
+            feature-gates: CSIMigrationAzureDisk=false
+            azure-container-registry-config: /etc/kubernetes/azure.json
```
Jont828, line 45: Same thing regarding newlines and in the other files as well/

test/e2e/azure_test.go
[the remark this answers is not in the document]
Jont828: Sure that makes sense to me. Maybe we can set some const values if we want to explicitly define the labels/categories too.

test/e2e/azure_csidriver.go
```diff
@@ -52,21 +59,50 @@ func AzureDiskCSISpec(ctx context.Context, inputGetter func() AzureDiskCSISpecIn
...
-	By("Deploying persistent volume claim")
-	b,err := e2e_pvc.Create("dd-managed-hdd-5g", "5Gi")
-	Expect(err).To(BeNil())
-	b.DeployPVC(clientset)
+		By("Deploying persistent volume claim")
+		pvcName = "dd-managed-hdd-5g" + util.RandomString(6)
+		b, err := e2e_pvc.Create(pvcName, scName, "5Gi")
+		Expect(err).To(BeNil())
+		annotations := map[string]string{
+			"volume.beta.kubernetes.io/storage-class":scName,
```
Jont828, line 74:
```suggestion
			"volume.beta.kubernetes.io/storage-class":scName,
```
```suggestion
			"volume.beta.kubernetes.io/storage-class": scName,
```

test/e2e/azure_csidriver.go
```diff
@@ -52,21 +59,50 @@ func AzureDiskCSISpec(ctx context.Context, inputGetter func() AzureDiskCSISpecIn
...
+		scName := "managedhdd" + util.RandomString(6)
+		e2e_sc.Create(scName).WithWaitForFirstConsumer().DeployStorageClass(clientset)
 
-	By("Deploying persistent volume claim")
-	b,err := e2e_pvc.Create("dd-managed-hdd-5g", "5Gi")
-	Expect(err).To(BeNil())
-	b.DeployPVC(clientset)
+		By("Deploying persistent volume claim")
+		pvcName = "dd-managed-hdd-5g" + util.RandomString(6)
+		b, err := e2e_pvc.Create(pvcName, scName, "5Gi")
```
Jont828, line 71: Just out of curiosity what does `b` stand for in this case?

test/e2e/azure_csidriver.go
```diff
@@ -52,21 +59,50 @@ func AzureDiskCSISpec(ctx context.Context, inputGetter func() AzureDiskCSISpecIn
...
+		Expect(err).To(BeNil())
+		annotations := map[string]string{
+			"volume.beta.kubernetes.io/storage-class":scName,
+		}
+		b.WithAnnotations(annotations)
+		b.DeployPVC(clientset)
+	}else{
```
Jont828, line 78:
```suggestion
	} else {
```

test/e2e/csi_migration_test.go
```diff
@@ -0,0 +1,333 @@
...
+		Expect(bootstrapClusterProxy).NotTo(BeNil(), "Invalid argument. bootstrapClusterProxy can't be nil when calling %s spec", specName)
+		Expect(os.MkdirAll(artifactFolder, 0755)).To(Succeed(), "Invalid argument. artifactFolder can't be created for %s spec", specName)
+		Expect(e2eConfig.Variables).To(HaveKey(capi_e2e.KubernetesVersion))
+
+		// CLUSTER_NAME and CLUSTER_NAMESPACE allows for testing existing clusters
+		// if CLUSTER_NAMESPACE is set don't generate a new prefix otherwise
+		// the correct namespace won't be found and a new cluster will be created
+		clusterNameSpace := os.Getenv("CLUSTER_NAMESPACE")
+		if clusterNameSpace == "" {
+			clusterNamePrefix = fmt.Sprintf("capz-e2e-%s", util.RandomString(6))
```
Jont828, line 70: Since this pattern of appending a random string comes up a lot, wdyt about defining a function like `generateName(base string, length int) string)` that just does `return base + "-" + util.RandomString(length)`?

test/e2e/helpers.go
```diff
@@ -911,3 +913,77 @@ func WaitForDaemonset(ctx context.Context, input clusterctl.ApplyClusterTemplate
 		return false
 	}, input.WaitForControlPlaneIntervals...).Should(Equal(true))
 }
+
+// podListHasNumPods fulfills the cluster-api PodListCondition type spec
```
Jont828, line 917:
```suggestion
// podListHasNumPods fulfills the Cluster API PodListCondition type spec.
```

test/e2e/helpers.go
```diff
@@ -911,3 +913,77 @@ func WaitForDaemonset(ctx context.Context, input clusterctl.ApplyClusterTemplate
...
+// given a list of pods, we validate for an exact number of those pods in a Running state
```
Jont828, line 918:
```suggestion
// Given a list of pods, we validate for an exact number of those pods in a Running state.
```
So linter doesn't complain, and same for the other functions here.

test/e2e/kubernetes/pvc/pvc.go
```diff
@@ -51,22 +52,34 @@ spec:
       storage: 5Gi
  */
 
+/*
+---
```
Jont828, line 56: Is this here as an example? If so maybe we can add a header comment describing this case and the case above ^.

test/e2e/kubernetes/storageclass/storageclass.go
```diff
@@ -77,6 +96,20 @@ func (d *Builder)WithWaitForFirstConsumer() *Builder  {
 	return d
 }
 
+// WithOotProvisionerName sets the correct external azure disk csi driver provisoner name
```
Jont828, line 99:
```suggestion
// WithOotProvisionerName sets the correct external azure disk csi driver provisioner name.
```

test/e2e/kubernetes/storageclass/storageclass.go
```diff
@@ -77,6 +96,20 @@ func (d *Builder)WithWaitForFirstConsumer() *Builder  {
...
+func (d *Builder)WithOotProvisionerName() *Builder  {
+	d.sc.Provisioner = OotAzureDiskProvisioner
+	return d
+}
+
+// WithOotParameters sets the sc parameters for out of tree csi drivers
```
Jont828, line 105:
```suggestion
// WithOotParameters sets the sc parameters for out of tree csi drivers.
```

test/e2e/kubernetes/storageclass/storageclass.go
```diff
@@ -77,6 +96,20 @@ func (d *Builder)WithWaitForFirstConsumer() *Builder  {
 	return d
 }
 
+// WithOotProvisionerName sets the correct external azure disk csi driver provisoner name
+func (d *Builder)WithOotProvisionerName() *Builder  {
```
Jont828, line 100:
```suggestion
func (d *Builder) WithOotProvisionerName() *Builder  {
```
Not sure why the existing functions didn't have a space but I think we can go back and add it.

Jont828 (commented):
Thanks for working on this! Just some minor nits/observations.

sonasingh46:
@CecileRobertMichon -- I have added [k8s-upgrade] label and skipping the test if k8s version upgrade is not from 1.22 --> 1.23. 

Question: Will this test even run in testgrid as I can see that upgrade from and upgrade to is pivoted to a particular stable release of k8s. 
https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/test/e2e/config/azure-dev.yaml#L150

CecileRobertMichon:
@sonasingh46 it will run in the periodic jobs (https://testgrid.k8s.io/sig-cluster-lifecycle-cluster-api-provider-azure#capi-periodic-upgrade-main-1-22-1-23)

test/e2e/config/azure-dev.yaml
```diff
@@ -145,6 +151,8 @@ variables:
   AKS_KUBERNETES_VERSION: "latest"
   KUBERNETES_VERSION: "${KUBERNETES_VERSION:-stable-1.23}"
   ETCD_VERSION_UPGRADE_TO: "3.5.3-0"
+  PRE_1_23_KUBERNETES_VERSION: "v1.22.4"
```
CecileRobertMichon, line 154: what is this / where is it used?

sonasingh46: It was used earlier. Let me remove that.

sonasingh46: Fixed it.

CecileRobertMichon (commented):
@andyzhangx could you please take a look?

test/e2e/helpers.go
```diff
@@ -911,3 +913,77 @@ func WaitForDaemonset(ctx context.Context, input clusterctl.ApplyClusterTemplate
...
+		Flavor:                   clusterctl.DefaultFlavor,
+		Namespace:                namespace,
+		ClusterName:              clusterName,
+		KubernetesVersion:        e2eConfig.GetVariable(capi_e2e.KubernetesVersion),
+		ControlPlaneMachineCount: pointer.Int64Ptr(3),
+		WorkerMachineCount:       pointer.Int64Ptr(0),
+	}
+}
+
+func createCluster(ctx context.Context, configCluster clusterctl.ConfigClusterInput, result *clusterctl.ApplyClusterTemplateAndWaitResult) (*clusterv1.Cluster, []*clusterv1.MachineDeployment, *controlplanev1.KubeadmControlPlane) {
```
CecileRobertMichon, line 966: Why not reuse the below function with a nil controlplane waiters parameter instead of creating a separate function?

templates/test/ci/prow-user-assigned-managed-identity/kustomization.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
CecileRobertMichon, line 1: why is this file being added? seems unrelated, maybe a rebase conflict?

templates/test/ci/prow-azurediskcsi-migration-off/kustomization.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: default
+resources:
+  - ../prow-user-assigned-managed-identity
```
CecileRobertMichon, line 5: why are these starting from a user-assigned template?

test/e2e/csi_migration_test.go
```diff
@@ -0,0 +1,347 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	clusterctlv1 "sigs.k8s.io/cluster-api/cmd/clusterctl/api/v1alpha3"
+	capi_e2e "sigs.k8s.io/cluster-api/test/e2e"
+	"sigs.k8s.io/cluster-api/test/framework/clusterctl"
+)
+
+var _ = Describe("[K8s-Upgrade] Running the workload cluster upgrade tests", func() {
+	var (
+		ctx                      = context.TODO()
+		specName                 = "create-workload-cluster"
```
CecileRobertMichon, line 47: spec name should be relevant to this test

test/e2e/csi_migration_test.go
```diff
@@ -0,0 +1,347 @@
...
+	)
+
+	BeforeEach(func() {
+		logCheckpoint(specTimes)
+		Expect(ctx).NotTo(BeNil(), "ctx is required for %s spec", specName)
+		Expect(e2eConfig).NotTo(BeNil(), "Invalid argument. e2eConfig can't be nil when calling %s spec", specName)
+		Expect(clusterctlConfigPath).To(BeAnExistingFile(), "Invalid argument. clusterctlConfigPath must be an existing file when calling %s spec", specName)
+		Expect(bootstrapClusterProxy).NotTo(BeNil(), "Invalid argument. bootstrapClusterProxy can't be nil when calling %s spec", specName)
+		Expect(os.MkdirAll(artifactFolder, 0755)).To(Succeed(), "Invalid argument. artifactFolder can't be created for %s spec", specName)
+		Expect(e2eConfig.Variables).To(HaveKey(capi_e2e.KubernetesVersion))
```
CecileRobertMichon, line 66: is this really used anywhere?

test/e2e/helpers.go
```diff
@@ -911,3 +913,77 @@ func WaitForDaemonset(ctx context.Context, input clusterctl.ApplyClusterTemplate
...
+			}
+		}
+		if runningPods >= numPods {
+			return errors.Errorf("expected %d Running pods, got %d", numPods, runningPods)
+		}
+		return nil
+	}
+}
+
+func defaultConfigCluster(clusterName, namespace string) clusterctl.ConfigClusterInput {
```
CecileRobertMichon, line 951: Thoughts about reusing this in more places? It's a good idea to refactor lines of code that come up often but if we're not using it everywhere it can lead to inconsistencies

test/e2e/kubernetes/pvc/pvc.go
```diff
@@ -83,6 +99,16 @@ func Create(pvcName string, storageRequest string) (*Builder,error) {
 	return pvcBuilder,nil
 }
 
+func (b *Builder)WithAnnotations(annotations map[string]string) *Builder{
```
Jont828, line 102:
```suggestion
func (b *Builder) WithAnnotations(annotations map[string]string) *Builder{
```
Nit

test/e2e/kubernetes/pvc/pvc.go
```diff
@@ -83,6 +99,16 @@ func Create(pvcName string, storageRequest string) (*Builder,error) {
...
+	b.pvc.Annotations = annotations
+	return b
+}
+
+func (b *Builder)WithStorageClass(scName string) *Builder{
```
Jont828, line 107:
```suggestion
func (b *Builder) WithStorageClass(scName string) *Builder{
```
Nit: Same for `DeployPVC` as well

Jont828 (commented):
A couple nits related to style but overall I don't see anything obviously wrong on my end.

andyzhangx:
AKS has enabled `AzureDiskCSIMigration` and installed Azure Disk CSI driver by default from 1.21, shall we follow same way on capz?

sonasingh46:
@andyzhangx -- Can you elaborate more? What do you mean by follow same way in capz? 
Are you saying that all the workload cluster creation e2e test should install external azure disk csi driver? 
Or Are you saying that we should test upgrade from 1.20 --> 1.21 also as azure disk csi driver is enabled by default in AKS

templates/test/ci/prow-azurediskcsi-migration-off/kustomization.yaml
sonasingh46: I figured out that we did not had a top level user assigned identity template. So added this to avoid repetition and use in other templates that needs a user assigned identity.  

I took inspiration from external cloud provider test that Jack wrote sometime back which is using user assigned identity. 

If we do not add this template, then in every template we will need to add following to the patch 
```
patchesStrategicMerge:
  - ../patches/uami-md-0.yaml
  - ../patches/uami-control-plane.yaml
 
 ```

File: templates/test/ci/prow-user-assigned-managed-identity/kustomization.yaml (hunk 17 above)
Comment (sonasingh46) [reply templates/test/ci/prow-user-assigned-managed-identity/kustomization.yaml:1]: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2456#discussion_r953604128

File: test/e2e/helpers.go (hunk 21 above)
Comment (sonasingh46) [reply test/e2e/helpers.go:951]: At the moment, it is not used elsewhere. But thought, would be a good start to start using this. I figured out that this make the e2e code more readable and make it really to understand what the test is doing on a high level. 
I can file up an issue so that we start using this defaultConfigCluster at other places in e2e and I can drive that to completion. 
My ideas is the following: 
- Get the defaultConfig using this function in our e2e test
- If we want to add/override -- do this e.g. `configCluster.WorkerMachineCount = pointer.Int64Ptr(3)`
wdyt?

File: test/e2e/kubernetes/pvc/pvc.go (hunk 23 above)
Comment (sonasingh46) [reply test/e2e/kubernetes/pvc/pvc.go:107]: This calls for adding linters for test files too. One by one -- in a next PR. 
But will fix the lint issue manually for this PR as of now. 
Thanks @Jont828

File: test/e2e/helpers.go (hunk 16 above)
Comment (sonasingh46) [reply test/e2e/helpers.go:966]: Makes sense.

**Comment by andyzhangx** [comment]
> AzureDiskCSIMigration

@sonasingh46 that depends on whether AzureDiskCSIMigration would be enabled on capz, if it's enabled on 1.23 (using default value), then Azure Disk CSI driver should be installed on capz 1.23 by default, otherwise there would be failure if user upgrade from 1.22 to 1.23 with in-tree disk PV.

**Comment by sonasingh46** [comment]
Right. As far as capz is concerned, the external drivers are not installed by default and no flags are set by default on controller manager that expects external csi driver for non managed clusters.

But with 1.23 the migration flag is enabled by default on core k8s.

So I don't think we need the test for version prior to 1.22 --> 1.23

If we need to do something similar for AKS as it might have implicitly enabled the migration flag -- sure we can have a test for managed cluster ( i.e AKS ). But I do not what is that inflection version to which we should test.
If I read it right -- you are saying for 1.21 k8s version in AKS -- the migration is made default-- so in this case 1.20 --> 1.21 upgrade for AKS make sense.

**Comment by andyzhangx** [comment]
@sonasingh46  for capz, only 1.22 --> 1.23 upgrade scenario is needed. And actually you could run the in-tree azure disk e2e test in every capz version, those in-tree azure disk e2e tests should pass in any version no matter it has enabled `AzureDiskCSIMigration` flag is enabled or not, no matter it's 1.21, 1.22 or 1.23.

**Comment by sonasingh46** [comment]
@andyzhangx -- Yes that makes sense. I can do that in a next PR as looks like this is bit out of scope of the current PR. wdyt?

**Comment by andyzhangx** [comment]
> @andyzhangx -- Yes that makes sense. I can do that in a next PR as looks like this is bit out of scope of the current PR. wdyt?

@sonasingh46 lgtm, thanks!

**Comment by sonasingh46** [comment]
https://github.com/kubernetes/test-infra/pull/27261

**Comment by CecileRobertMichon** [comment]
@sonasingh46 looks like the csi migration test is timing out, PTAL

**Comment by sonasingh46** [comment]
Checked the logs. Apparently the tests are passing but it is timing out due to something else e.g. collecting log. Will dig bit deeper to root cause it.

File: test/e2e/config/azure-dev.yaml [hunk 24]
```diff
@@ -196,4 +196,8 @@ intervals:
   default/wait-job: ["5m", "10s"]
   default/wait-service: ["15m", "10s"]
   default/wait-machine-pool-nodes: ["30m", "10s"]
+  csi-migration/wait-controlplane-upgrade: ["60m", "10s"]
+  csi-migration/wait-worker-nodes: ["60m", "10s"]
+  csi-migration/wait-control-plane: ["60m", "10s"]
```
CecileRobertMichon, line 201: do we really need 60 minutes to wait for the cluster, worker nodes, and control planes? 20 minutes should be way enough no?

CecileRobertMichon: also if we want to reuse the default/ times above we don't need to add a specific one for csi-migration

sonasingh46: Yes. Did not want to use the default ones. 

>do we really need 60 minutes to wait for the cluster, worker nodes, and control planes? 20 minutes should be way enough no?

I have seen that 30 mins was not enough for worker nodes to comes up especially during the upgrade due to rolling upgrade. 
I did couple of runs -- one time it took like 42 mins. So had set it to 60 mins just for csi-migration spec.

sonasingh46: All of the tests has 3 CP + 3 Worker node.

CecileRobertMichon: ok makes sense

CecileRobertMichon:
lgtm

please squash @sonasingh46

Issue #2422, opened 2022-06-24
Shorten AKS Windows pool names when AzureManagedMachinePool name is too long


jackfrancis:
/kind feature

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]

The AKS API enforces a max 6 character limit for node pool names. Because this is not necessarily an intuitive requirement, let's consider ways to do this automatically on behalf of users.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

A possible approach to follow is https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1408

**Environment:**

- cluster-api-provider-azure version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

CecileRobertMichon:
indeed!

/close

jackfrancis:
So actually the foo that @luthermonson links to is validation that the user-configured pool name is no longer than 6 characters. I originally opened up this issue to track consideration of doing this automatically (truncating).

If we want to consider doing that (user inputs "my-pool-name-win" and we truncate to "win-kX" or something) we can re-open this issue.

CecileRobertMichon:
Ah, understood. The title makes it sound like this is about validation I think.

/reopen
/retitle Shorten AKS Windows pool names when AzureManagedMachinePool name is too long

sayantani11:
@jackfrancis @CecileRobertMichon I would like to get some more insight on this issue if possible!

jackfrancis:
I'm not actually convinced we want to do this. @luthermonson do you have thoughts? Would you prefer a capz-enforced string concatenation and/or 6-char hash that replaces the inputted node pool name, or would you prefer capz to fail validation and force the user to purposefully choose a name that is 6 chars or fewer?

PR #2461, opened 2022-07-07
Adding Validators for ammp.Spec.OSType and amcp.Name

luthermonson:
**What type of PR is this?**
/kind cleanup

* Adds immutable checks for ammp.Spec.OSType, previously added OSType field was marked as immutable in docs etc. but the validator to confirm the update didn't let you change it was never added.
* The azuremanagedcontrolplane name is also immutable as AKS doesn't let you change a cluster name after it is created. Added a validator to bubble up errors if you try. In addition there is this nuance to Azure ` managed cluster name is invalid because 'MICROSOFT' and 'WINDOWS' can't be used as either a whole word or a substring in the name..` which I added a new validator for to help bubble up errors before they hit the Azure API.

Fixes #

try and make a cluster with `microsoft` or `windows` in the title, try and change the name of the cluster or the ammp ostype and you should see validation errors

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Add validators for ammp.Spec.OSType and amcp.Name
```

### DIFF
--- a/exp/api/v1beta1/azuremanagedmachinepool_webhook.go
+++ b/exp/api/v1beta1/azuremanagedmachinepool_webhook.go
@@ -76,6 +76,25 @@ func (m *AzureManagedMachinePool) ValidateUpdate(oldRaw runtime.Object, client c
 	old := oldRaw.(*AzureManagedMachinePool)
 	var allErrs field.ErrorList
 
+	if old.Spec.OSType != nil {
+		// Prevent OSType modification if it was already set to some value
+		if m.Spec.OSType == nil {
+			// unsetting the field is not allowed
+			allErrs = append(allErrs,
+				field.Invalid(
+					field.NewPath("Spec", "OSType"),
+					m.Spec.OSType,
+					"field is immutable, unsetting is not allowed"))
+		} else if *m.Spec.OSType != *old.Spec.OSType {
+			// changing the field is not allowed
+			allErrs = append(allErrs,
+				field.Invalid(
+					field.NewPath("Spec", "OSType"),
+					*m.Spec.OSType,
+					"field is immutable"))
+		}
+	}
+
 	if m.Spec.SKU != old.Spec.SKU {
 		allErrs = append(allErrs,
 			field.Invalid(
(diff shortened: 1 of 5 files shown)


CecileRobertMichon:
Thanks @luthermonson! Maybe out of scope for this PR but there is also #2422 if you want to add validation for that while you're at it

luthermonson:
@CecileRobertMichon can do... updating now

jsturtevant:
/lgtm
pending tests

exp/api/v1beta1/azuremanagedmachinepool_webhook.go
```diff
@@ -76,6 +76,14 @@ func (m *AzureManagedMachinePool) ValidateUpdate(oldRaw runtime.Object, client c
 	old := oldRaw.(*AzureManagedMachinePool)
 	var allErrs field.ErrorList
 
+	if m.Spec.OSType != old.Spec.OSType {
```
jackfrancis, line 79: `OSType` is a `*string`, so you'll have to follow the pattern from `OSDiskSizeGB` below, e.g.:

```
	if old.Spec.OSType != nil {
		// Prevent OSType removal if it was already set to some value
		if m.Spec.OSType == nil {
			// unsetting the field is not allowed
			allErrs = append(allErrs,
				field.Invalid(
					field.NewPath("Spec", "OSType"),
					m.Spec.OSType,
					"field is immutable, unsetting is not allowed"))
		} else if *m.Spec.OSType != *old.Spec.OSType {
			// changing the field is not allowed
			allErrs = append(allErrs,
				field.Invalid(
					field.NewPath("Spec", "OSType"),
					*m.Spec.OSType,
					"field is immutable"))
		}
	}
```

luthermonson: moof... my bad

luthermonson: fixed!

luthermonson:
@jackfrancis can this get an lgtm?

CecileRobertMichon:
@jackfrancis is on vacation but this lgtm

/lgtm
/approve

@luthermonson can you please add a release note in the PR description?

mboersma:
/lgtm

Thanks for your patience, this looks great!

PR #2471, opened 2022-07-12
user-assigned ID for all reference templates

### LINKED ISSUE #1534: Make user-assigned identity the default in flavors and e2e
/kind feature

Managed identity is more secure than writing SP creds on disk and is the recommended approach for VM identity. #1531 makes that more clear in docs. However, most templates/flavors and e2e templates are still using SP creds. We should switch them over to user-assigned so we can get better coverage on managed identity, improve our CI security, and eventually deprecate (maybe?) SP authentication for cloud-provider.

This will require a msft person to be involved on the e2e side as we'll need to provision a managed identity for Prow to be able to use with the test subscription.

**Environment:**

jackfrancis:
**What type of PR is this?**

/kind cleanup

This PR updates all reference and test templates so that a user-assigned identity configuration is used for all control plane machines and machines built from MachineDeployments and MachinePools. We already have this preferred configuration running in a variety of test templates, this PR promotes that to the default everywhere.

Fixes #1534

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```


templates/cluster-template.yaml
```diff
@@ -179,10 +182,13 @@ metadata:
 spec:
   template:
     spec:
+      identity: UserAssigned
       osDisk:
         diskSizeGB: 128
         osType: Linux
       sshPublicKey: ${AZURE_SSH_PUBLIC_KEY_B64:=""}
+      userAssignedIdentities:
+      - providerID: /subscriptions/${AZURE_SUBSCRIPTION_ID}/resourceGroups/${CI_RG:=capz-ci}/providers/Microsoft.ManagedIdentity/userAssignedIdentities/${USER_IDENTITY:=cloud-provider-user-identity}
```
CecileRobertMichon, line 191: this is modifying the base flavor templates which will require users to pre-create the user assigned ID in their subscription prior to using them (including in the quickstart.

It's also defaulting the rg to "capz-ci" or `CI_RG` which doesn't really apply in non-CI scenarios.

mboersma:
I think user-assigned ID is the way to go, but I'm concerned that it requires more up-front setup by users. 

[The documentation](https://capz.sigs.k8s.io/topics/vm-identity.html#user-assigned-managed-identity-recommended) CAPZ has for this is inadequate if this becomes the default path; it would need to be more prominent and show actual `az` commands rather than linking to another doc IMHO. 

And ideally someone deploying one of these templates without having set up user-assigned managed ID would get a perfectly explicit error about how to fix it (maybe even just-in-time?).

jackfrancis:
If we don't want to do this work we should probably say so in https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1534 and close that issue.

CecileRobertMichon:
+1, I think this is a worthwhile improvement but we need to do a bit of thinking around docs and quickstart so it's not too disruptive for new and existing users.

CecileRobertMichon:
/close

Let's circle back after #2205 is done, once all management clusters can use managed identities with identity federation support, that way it works out of the box for all user templates/examples (then we can use managed identities _everywhere_ and never require a service principal ever again)

jackfrancis:
sgtm

PR #2482, opened 2022-07-14
[WIP] Try setting e2e ginkgo parallel nodes to 24

CecileRobertMichon:
**What type of PR is this?**

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note

```


Makefile
```diff
@@ -649,7 +648,7 @@ test-cover: $(SETUP_ENVTEST) ## Run tests with code coverage and code generate r
 .PHONY: test-e2e-run
 test-e2e-run: generate-e2e-templates install-tools ## Run e2e tests.
 	$(ENVSUBST) < $(E2E_CONF_FILE) > $(E2E_CONF_FILE_ENVSUBST) && \
-    $(GINKGO) -v -trace -tags=e2e -focus="$(GINKGO_FOCUS)" -skip="$(GINKGO_SKIP)" -nodes=$(GINKGO_NODES) --noColor=$(GINKGO_NOCOLOR) $(GINKGO_ARGS) ./test/e2e -- \
+    $(GINKGO) -v -trace -tags=e2e -focus="$(GINKGO_FOCUS)" -skip="$(GINKGO_SKIP)" -p --noColor=$(GINKGO_NOCOLOR) $(GINKGO_ARGS) ./test/e2e -- \
```
CecileRobertMichon, line 651: this will automatically detect the optimal number of nodes to use

https://godocs.io/github.com/onsi/ginkgo/ginkgo

CecileRobertMichon:
Screen Shot 2022-07-18 at 11 46 41 AM

/assign @mboersma

Makefile
mboersma: Why are we hardcoding `GINKGO_NODES` to 10 instead of using `-p` now?

CecileRobertMichon: When using `-p` Ginkgo will automatically choose the "optimal" number of parallel nodes to use. However, I found that when using `-p` Ginkgo would run our test suites with 7 parallel nodes, including the ones that have more parallel tests which results in a longer overall time. When using `--nodes 10` we make sure that we're always running everything in parallel since at most our jobs are running 9 specs.

```
Will run 23 specs
Running in parallel across 7 nodes
```

https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/2482/pull-cluster-api-provider-azure-capi-e2e/1547996819181539328

Might be worth understanding why Ginkgo is choosing to use 7 parallel nodes. The "23" specs comes from the total number of specs in our suite (9 + 14 skipped) so it might not account for skipped test suites?

mboersma: > When using --nodes 10 we make sure that we're always running everything in parallel...

Makes sense to me.

As far as I can tell, Ginkgo just uses `runtime.NumCPU()` when given `-p`, but subtracts one if that number is 5 or more.

https://github.com/onsi/ginkgo/blob/master/types/config.go#L147-L160

Issue #1876, opened 2021-11-18
e2e tests logs are missing the entire controller manager logs 


jsturtevant:
/kind bug

[Before submitting an issue, have you checked the [Troubleshooting Guide](https://capz.sigs.k8s.io/topics/troubleshooting.html)?]

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]
The job https://testgrid.k8s.io/sig-cluster-lifecycle-cluster-api-provider-azure#capz-periodic-e2e-full-main is missing the full set of controller manager logs. If you look for vmss in 

- https://storage.googleapis.com/kubernetes-jenkins/logs/periodic-cluster-api-provider-azure-e2e-full-main/1461402439272894464/artifacts/clusters/bootstrap/controllers/capi-controller-manager/capi-controller-manager-6bdc78c4d4-nnxc7/manager.log 
- https://storage.googleapis.com/kubernetes-jenkins/logs/periodic-cluster-api-provider-azure-e2e-full-main/1461402439272894464/artifacts/clusters/bootstrap/controllers/capz-controller-manager/capz-controller-manager-5f7b59585c-m9lxd/manager.log 
 
you will not find relevant logs for that cluster even though the cluster was created: https://gcsweb.k8s.io/gcs/kubernetes-jenkins/logs/periodic-cluster-api-provider-azure-e2e-full-main/1461402439272894464/artifacts/clusters/capz-e2e-4e4w9c-win-vmss/machines/

**What did you expect to happen:**

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

My guess is the logs are rolling and we are missing the first part of the logs

**Environment:**

- cluster-api-provider-azure version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

jsturtevant:
I guess this currently is not in capz specifically. It looks like it is provided by the capi framework: https://github.com/kubernetes-sigs/cluster-api/blob/aafa086da595450ff1f75d4a29b213c873f15e4d/test/framework/clusterctl/clusterctl_helpers.go#L103-L109

jsturtevant:
So looking closer It seems like there maybe the stream is dropped at some point or restart of the agent logs and the stream doesn't re-establish itself.  Example:

[...]

PR #2483, opened 2022-07-14
Don't reload E2E config on each Ginkgo parallel node

CecileRobertMichon:
**What type of PR is this?**

**What this PR does / why we need it**: While trying to understand why we don't have full controller logs when running tests in parallel (#1876) I saw that we were reloading the e2e config on each parallel node instead of passing around the config. Because the gingko SynchronizedBeforeSuite allows us to pass []bytes from the first node to the others, we can encode the config and pass it around instead of reloading it each time. This happens to solve the issue #2440 was trying to solve.

Fixes #
Closes #2440 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/e2e_suite_test.go
+++ b/test/e2e/e2e_suite_test.go
@@ -20,7 +20,9 @@ limitations under the License.
 package e2e
 
 import (
+	"bytes"
 	"context"
+	"encoding/gob"
 	"flag"
 	"fmt"
 	"os"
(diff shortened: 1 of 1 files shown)


CecileRobertMichon:
quota issues

/retest

mboersma (approved):
/lgtm

PR #2466, opened 2022-07-08
Update to Ginkgo v2

mboersma:
**What type of PR is this?**:

/kind cleanup

**What this PR does / why we need it**:

Updates the Ginkgo test framework to v2 in CAPZ e2e tests.

This is mostly for discussion and learning. This change shouldn't merge until CAPI and other dependencies are also using Ginkgo v2, since the indirect import of v1 will conflict. I used an [import hack](https://github.com/onsi/ginkgo/issues/875#issuecomment-1036350659) to work around this for now.

CAPZ has been abusing nested [container nodes](https://onsi.github.io/ginkgo/#nodes-only-belong-in-container-nodes). This is now an error in Ginkgo v2, so we might consider getting out ahead of that by extracting those `Context` -> `By` changes into a separate PR.

This is a minimal refactoring, but there are more idiomatic changes such as [Spec Labels](https://onsi.github.io/ginkgo/#spec-labels) that we might also consider.

**Which issue(s) this PR fixes**:

Refs kubernetes-sigs/cluster-api#4897

**Special notes for your reviewer**:

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Update to Ginkgo v2
```

PR #2486, opened 2022-07-19
Update e2e tests for Ginkgo v2

mboersma:
**What type of PR is this?**

/kind cleanup

Fixes CAPZ's misuse of Ginkgo [container nodes](https://onsi.github.io/ginkgo/#nodes-only-belong-in-container-nodes) and updates a `ParallelNodes` reference such that both will work when we move to Ginkgo v2.

Nesting a `Context` inside an `It` container node is a warning in Ginkgo v1 but an error in v2. We can fix this now for correctness, to simplify the transition to v2, and to prevent more copy-and-pasting of this mistake.

Refs #2466

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/azure_test.go
+++ b/test/e2e/azure_test.go
@@ -149,7 +149,7 @@ var _ = Describe("Workload cluster creation", func() {
 		Context("Creating a private cluster [REQUIRED]", func() {
 			It("Creates a public management cluster in a custom vnet", func() {
 				clusterName = getClusterName(clusterNamePrefix, "public-custom-vnet")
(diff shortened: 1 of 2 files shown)


CecileRobertMichon:
Does this not need the CAPI version of these changes to merge first?

mboersma:
> Does this not need the CAPI version of these changes to merge first?

No, these are just future-proofing changes that work with either v1 or v2 of Ginkgo. Also we are using `Context` wrong and that should be corrected.

PR #2487, opened 2022-07-20
Upgrade golang to v1.18.4

Prajyot-Parab:
Signed-off-by: Prajyot-Parab <prajyot.parab2@ibm.com>

 

 
**What type of PR is this?**
/kind feature

- Upgrade golang to v1.18.4

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Upgrade golang to v1.18.4
```


Dockerfile
```diff
@@ -15,7 +15,7 @@
 # limitations under the License.
 
 # Build the manager binary
-FROM golang:1.17 as builder
+FROM golang:1.18.4 as builder
```
mboersma, line 18: Shouldn't we leave this at `golang:1.18` so it picks up the latest patch release of Go? Otherwise we might need to update all these references often.

Prajyot-Parab: Agreed, will fix it.

hack/ensure-go.sh
```diff
@@ -31,7 +31,7 @@ EOF
   local go_version
   IFS=" " read -ra go_version <<< "$(go version)"
   local minimum_go_version
-  minimum_go_version=go1.17.0
+  minimum_go_version=go1.18.4
```
mboersma, line 34: CAPI has this value as `go1.18.3`. While it's wise to be on the latest patch, I think we should match CAPI's minimum version so we don't impose additional developer requirements. If you're successfully writing code for CAPI with Go 1.18.3, that should also work for CAPZ in my opinion.

Prajyot-Parab: Fixed.

Makefile
```diff
@@ -354,7 +354,7 @@ delete-workload-cluster: $(KUBECTL) ## Deletes the example workload Kubernetes c
 .PHONY: docker-pull-prerequisites
 docker-pull-prerequisites: ## Pull prerequisites for building controller-manager.
 	docker pull docker/dockerfile:1.4
-	docker pull docker.io/library/golang:1.17
+	docker pull docker.io/library/golang:1.18.3
```
mboersma, line 357:
```suggestion
	docker pull docker.io/library/golang:1.18
```

Makefile
```diff
@@ -602,7 +602,7 @@ release-binary: $(RELEASE_DIR) ## Compile and build release binaries.
 		-e GOARCH=$(GOARCH) \
 		-v "$$(pwd):/workspace" \
 		-w /workspace \
-		golang:1.17 \
+		golang:1.18.3 \
```
mboersma, line 605:
```suggestion
		golang:1.18 \
```

Tiltfile
```diff
@@ -104,7 +104,7 @@ def validate_auth():
 
 tilt_helper_dockerfile_header = """
 # Tilt image
-FROM golang:1.17 as tilt-helper
+FROM golang:1.18.3 as tilt-helper
```
mboersma, line 107:
```suggestion
FROM golang:1.18 as tilt-helper
```

mboersma (changes requested):

mboersma:
@Prajyot-Parab this is failing on the `make verify-conversions` target. You can fix it by doing this:

```shell
pushd hack/tools
go get sigs.k8s.io/cluster-api/hack/tools/conversion-verifier@v0.0.0-20220202124422-fe7656c044e1
popd
```
and then committing the result.

Thanks for doing this work!

mboersma:
/hold cancel

kubernetes/test-infra#26864 is merged.

jackfrancis:
ClusterClass E2E test failure is known. See:

https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/2505

jackfrancis (approved):
/lgtm
/approve

PR #2495, opened 2022-07-22
Use Helm to install Calico CNI in e2e tests instead of ClusterResourceSets

CecileRobertMichon:
**What type of PR is this?**

/kind cleanup

This PR replaces the existing ClusterResourceSet-dependent solution for automating calico CNI delivery onto test clusters (and reference cluster templates) with a helm solution. For the private cluster scenario, we're keeping the CRS-dependent solution as a practical way to install the CNI from the management cluster (we have no direct access to the apiserver to helm install gestures and we don't yet have a E2E solution to indirectly access the apiserver in such scenarios).

This work is a continuation of @jackfrancis' work in #2334.

Fixes https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/2063
Contributes to https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/2179

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Use Helm to install Calico CNI
```


jackfrancis:
oooh 👀👀👀👀👀👀👀

test/e2e/helpers.go
```diff
@@ -879,6 +879,7 @@ func InstallHelmChart(ctx context.Context, input clusterctl.ApplyClusterTemplate
 		p := helmGetter.All(settings)
 		valueOpts := &helmVals.Options{}
 		valueOpts.Values = values
+		valueOpts.ValueFiles = valueFiles
```
jackfrancis, line 882: I assume the helm api handles an empty array here gracefully? And/or does the same if we include data in both `Values` and `ValueFiles` properties?

(Wondering if we need to do some input validation sanity-checking...)

CecileRobertMichon: Yes `MergeValues` will iterate over each `[]string` of values and merge them https://github.com/helm/helm/blob/main/pkg/cli/values/options.go#L41

test/e2e/helpers.go
```diff
@@ -858,29 +859,39 @@ func getPodLogs(ctx context.Context, clientset *kubernetes.Clientset, pod corev1
...
-	err := actionConfig.Init(settings.RESTClientGetter(), ns, "secret", Logf)
+	err := actionConfig.Init(settings.RESTClientGetter(), namespace, "secret", Logf)
 	Expect(err).To(BeNil())
 	i := helmAction.NewInstall(actionConfig)
 	if repoURL != "" {
 		i.RepoURL = repoURL
 	}
 	i.ReleaseName = releaseName
-	i.Namespace = ns
+	i.Namespace = namespace
```
Jont828, line 887: The Helm install client has a flag to create a namespace if it doesn't exist. You can add this below.
```
i.CreateNamespace = true
```

test/e2e/cni.go
```diff
@@ -0,0 +1,109 @@
...
+			ipv6 = cidr
+		} else {
+			ipv4 = cidr
+		}
+	}
+	addonsPath := e2eConfig.GetVariable(capi_e2e.CNIPath)
+	if ipv6 != "" && ipv4 != "" {
+		// dual-stack
+		values = &helmVals.Options{
+			StringValues: []string{fmt.Sprintf("installation.calicoNetwork.ipPools[0].cidr=%s", ipv4), fmt.Sprintf("installation.calicoNetwork.ipPools.ipPools[1].cidr=%s", ipv6)},
```
Jont828, line 92: Do we plan on using the ipPool CIDRs specified in the different values.yaml files? Or is it just there as a default?

test/e2e/cni.go
```diff
@@ -0,0 +1,109 @@
...
+		WaitForDeploymentsAvailable(ctx, waitInput, e2eConfig.GetIntervals(specName, "wait-deployment")...)
+	}
+	By("Waiting for Ready calico-node daemonset pods")
+	for _, ds := range []string{"calico-node"} {
+		WaitForDaemonset(ctx, input, workloadClusterClient, ds, CalicoSystemNamespace)
+	}
+}
+
+func getCalicoValues(cidrBlocks []string) *helmVals.Options {
+	var ipv6, ipv4 string
```
Jont828, line 79: It looks like this would overshadow the const `ipv4` and `ipv6` declared at the top of the file and are used to specify CIDR blocks.
```suggestion
	var ipv6CidrBlock, ipv4CidrBlock string
```

test/e2e/helpers.go
CecileRobertMichon: thank you, I didn't know, that's great!!

templates/addons/calico-dual-stack/values.yaml
```diff
@@ -0,0 +1,19 @@
...
+  cni:
+    type: Calico
+    ipam:
+      type: HostLocal
+  calicoNetwork:
+    bgp: Disabled
+    mtu: 1350
+    ipPools:
+    - blockSize: 26
+      cidr: 10.244.0.0/16
```
Jont828, line 11: Would there be a situation where a pod CIDR list has multiple IPv4 address blocks? If so, how would we set this field? Would it be the union of the two blocks?

test/e2e/cni.go
CecileRobertMichon: The ipPool CIDRs in the values.yaml are there as default (they match the default values in cluster-template.yaml) so the file can be used outside of e2e tests, for example in the quickstart instructions

templates/addons/calico-dual-stack/values.yaml
CecileRobertMichon: I don't think that's possible but let me do more research on that to verify

go.mod
```diff
@@ -44,7 +44,7 @@ require (
 	k8s.io/kubectl v0.24.0
 	k8s.io/utils v0.0.0-20220210201930-3a6ce19ff2f9
 	sigs.k8s.io/cluster-api v1.2.0
-	sigs.k8s.io/cluster-api/test v1.2.0
+	sigs.k8s.io/cluster-api/test v1.2.1-0.20220725220710-f3c6c6de4c40
```
CecileRobertMichon, line 47: TODO: update this to 1.2.1 once it is released

CecileRobertMichon, line 47: TODO: update this to 1.2.1 once it is released

docs/book/src/topics/addons.md
```diff
@@ -0,0 +1,158 @@
...
+| sed 's/namespace: kube-system/namespace: calico-system/' \
+| kubectl create -f -
+```
+
+</aside>
+
+
+For more information, see the [official Calico documentation](https://projectcalico.docs.tigera.io/getting-started/kubernetes/helm).
+
+## Flannel
```
CecileRobertMichon, line 54: note for reviewers: this section was copied as-is from flannel.md

docs/book/src/topics/addons.md
```diff
@@ -0,0 +1,158 @@
...
+    }
+```
+
+Apply kube-flannel.yml
+
+```bash
+kubectl apply -f kube-flannel.yml
+```
+
+# External Cloud Provider
```
CecileRobertMichon, line 122: note for reviewers: this section was copied as-is from cloud-provider-config.md

docs/book/src/topics/addons.md
```diff
@@ -0,0 +1,158 @@
...
+IPV6_CIDR_BLOCK=<cluster ipv6 pod cidr block> \
+helm repo add projectcalico https://projectcalico.docs.tigera.io/charts && \
+helm install calico projectcalico/tigera-operator -f https://raw.githubusercontent.com/kubernetes-sigs/cluster-api-provider-azure/main/templates/addons/calico-dual-stack/values.yaml --set-string installation.calicoNetwork.ipPools[0].cidr="$IPV4_CIDR_BLOCK",installation.calicoNetwork.ipPools[1].cidr="$IPV6_CIDR_BLOCK" --namespace tigera-operator --create-namespace
+```
+
+<aside class="note">
+
+<h1> Note </h1>
+
+For Windows nodes, you also need to copy the kubeadm-config configmap to the calico-system namespace so the calico-node-windows Daemonset can find it:
```
CecileRobertMichon, line 40: @marosset I'm concerned that this will break upstream tests out there that rely on the CAPZ windows calico resources, or that it will confuse users. It works as a workaround for our tests but it's not a great long-term solution. It would be much better if we can figure out a way to solve https://github.com/kubernetes-sigs/sig-windows-tools/blob/master/hostprocess/calico/install/calico-install.ps1#L13 and not rely on the kubeadm-config configmap.

CecileRobertMichon:
This is overall ready for review, added some docs and cleaned up commits.

The only thing blocking at this point is the self-hosted test which is failing, I still need to figure out why. In the meantime we can start reviewing and I'll keep this on hold so it doesn't merge.

/hold

/assign @jackfrancis @Jont828 @sonasingh46

CecileRobertMichon:
The self-hosted test is failing due to https://github.com/projectcalico/calico/issues/3145 which we worked around in the manifest by adding `featureDetectOverride: ChecksumOffloadBroken=true` to Felix configuration in manifest. It's not as easy to workaround with Helm chart as it doesn't allow direct configuration of the FelixConfiguration CRD. 

Trying to understand why the bug still happens in the first place since the root cause kernel bug has been fixed (https://github.com/torvalds/linux/commit/ea64d8d6c675c0bb712689b13810301de9d8f77a) since April 2020.

go.mod
CecileRobertMichon: done

CecileRobertMichon: done

CecileRobertMichon:
Current status of this:

1) needs rebase
2) needs reviews
3) two things making me hesitate to move this forward - requires workaround for Windows kube-proxy and workaround for FelixConfiguration because of the ChecksumOffloadBroken kernel bug. Not blockers per se since they both have workarounds but this is a regression in UX.

jackfrancis:
Are there any open calico issues tracking those workarounds?

Just a thought: now that we've established working proficiency with doing this via helm generally, would it make sense to timebox a cilium spike? Perhaps this has a chart that requires less hacking on our part?:

https://github.com/cilium/charts

AFAIK we don't have any particular reason (other than historical momentum) to prefer one standard IPAM CNI solution over another, my understanding is that cilium is functionally equivalent to calico in terms of IPAM.

CecileRobertMichon:
> Are there any open calico issues tracking those workarounds?

Yes for the calico FelixConfiguration: https://github.com/projectcalico/calico/issues/4727

For the Windows kube-proxy relying on the kubeadm configmap there is a [TODO](https://github.com/kubernetes-sigs/sig-windows-tools/blob/master/hostprocess/calico/install/calico-install.ps1#L13) but I'm not aware of an issue tracking it - @jsturtevant / @marosset do you know?

> Just a thought: now that we've established working proficiency with doing this via helm generally, would it make sense to timebox a cilium spike? Perhaps this has a chart that requires less hacking on our part?

Yes that's a great idea! I was actually going to investigate because of #2723 (which makes it seem like it doesn't work out of the box either). I think they aren't mutually exclusive though, I would advocate that we move this PR forward without removing the existing Calico install yaml used in the quickstart, but at least start by replacing ClusterResourceSets used in the e2e tests. And look into adding side by side e2e tests with Cilium in parallel. If we find that one is better than the other UX wise and they are both equivalent performance wise we can revisit what we use as "default" in docs and quickstart.

CecileRobertMichon (commented):
/assign @jackfrancis @marosset 

This is officially ready for review

/hold cancel

jsturtevant:
We don't have an issue (We probably should as this has come up recently now with HostProcess containers going to Stable in 1.26). There are details on the changes required in https://docs.google.com/document/d/1A6Gkyx5EvL86Z4L2P4sxnk9HQjdRgnLXvBg6tIMDB3s/edit# which was presented to sig-windows

docs/book/src/topics/addons.md
```diff
@@ -0,0 +1,162 @@
+# CNI
+
+By default, the CNI plugin is not installed for self-managed clusters, so you have to [install your own](https://cluster-api.sigs.k8s.io/user/quick-start.html#deploy-a-cni-solution).
+
+Some of the instructions below use [Helm](https://helm.sh) to install the addons. If you're not familiar with using Helm to manage Kubernetes applications as packages, there's lots of good [Helm documentation on the official website](https://helm.sh/docs/). You can install Helm by following the [official instructions](https://helm.sh/docs/intro/install/).
+
+## Calico
```
CecileRobertMichon, line 7: Going to work on adding a Cilium section here as a follow up, once I look into #2723

templates/addons/windows/calico/calico.yaml
```diff
@@ -11,7 +11,7 @@ kind: ConfigMap
 apiVersion: v1
 metadata:
   name: calico-static-rules
-  namespace: kube-system
```
CecileRobertMichon, line 14: @marosset this is important to note: because the Calico Helm chart puts everything in `calico-system` namespace, it requires the windows components to be in the same namespace which is why I changed all these manifests. We might want to check if these are being used anywhere outside CAPZ so this doesn't break users / tests when it merges...

templates/addons/calico-dual-stack/values.yaml
CecileRobertMichon: confirmed. CAPI is also adding validation to enforce this: https://github.com/kubernetes-sigs/cluster-api/pull/7420

CecileRobertMichon:
This is still ready for review

/lifecycle active

Preference for not merging today though, given that we're releasing tomorrow and this is a huge change.

Once Calico 1.25 is out we'll be able to revert the last commit for the FelixConfiguration workaround.

CecileRobertMichon:
Need to look into why tests are failing...

CecileRobertMichon:
@jackfrancis would you be able to give this one another review? Would be great to get it merge soon-ish so I can stop rebasing and we can have sufficient mileage before the v1.7 release 

(tests are failing due to ongoing docker [outage](https://status.docker.com/))

Tiltfile
```diff
@@ -350,8 +346,20 @@ def deploy_worker_templates(template, substitutions):
...
+    flavor_cmd += "; until " + kubectl_cmd + " get secret ${CLUSTER_NAME}-kubeconfig > /dev/null 2>&1; do sleep 5; done; " + kubectl_cmd + " get secret ${CLUSTER_NAME}-kubeconfig -o jsonpath={.data.value} | base64 --decode > ./${CLUSTER_NAME}.kubeconfig; chmod 600 ./${CLUSTER_NAME}.kubeconfig; until " + kubectl_cmd + " --kubeconfig=./${CLUSTER_NAME}.kubeconfig get nodes > /dev/null 2>&1; do sleep 5; done"
+
+    # install calico
+    if "ipv6" in flavor_name:
+        calico_values = "./templates/addons/calico-ipv6/values.yaml"
+    elif "dual-stack" in flavor_name:
+        calico_values = "./templates/addons/calico-dual-stack/values.yaml"
+    else:
+        calico_values = "./templates/addons/calico/values.yaml"
+    flavor_cmd += "; " + helm_cmd + " repo add projectcalico https://projectcalico.docs.tigera.io/charts; " + helm_cmd + " --kubeconfig ./${CLUSTER_NAME}.kubeconfig install calico projectcalico/tigera-operator -f " + calico_values + " --namespace tigera-operator --create-namespace; kubectl --kubeconfig ./${CLUSTER_NAME}.kubeconfig apply -f ./templates/addons/calico/felix-override.yaml"
```
jackfrancis, line 360: should we wait for some outcome from the helm install (like the existence of a CRD, for example) before applying the override? I assume there is some serialization required there?

docs/book/src/topics/addons.md
```diff
@@ -0,0 +1,162 @@
...
+### For IPv6 clusters
+
+```bash
+IPV6_CIDR_BLOCK=<cluster ipv6 pod cidr block> \
+helm repo add projectcalico https://projectcalico.docs.tigera.io/charts && \
+helm install calico projectcalico/tigera-operator -f https://raw.githubusercontent.com/kubernetes-sigs/cluster-api-provider-azure/main/templates/addons/calico-ipv6/values.yaml  --set-string installation.calicoNetwork.ipPools[0].cidr="$IPV6_CIDR_BLOCK" --namespace tigera-operator --create-namespace
+kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/cluster-api-provider-azure/main/templates/addons/calico/felix-override.yaml
+```
+
+### For IPv4/IPv6 dual-stack clusters
```
jackfrancis, line 30: I think I'd prefer referring to this as something like "dual-stack (IPv4 + IPv6)"

docs/book/src/topics/addons.md
```diff
@@ -0,0 +1,162 @@
...
+kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/cluster-api-provider-azure/main/templates/addons/calico/felix-override.yaml
+```
+
+<aside class="note">
+
+<h1> Note </h1>
+
+For Windows nodes, you also need to copy the kubeadm-config configmap to the calico-system namespace so the calico-node-windows Daemonset can find it:
+
+```bash
```
jackfrancis, line 46: yikes, is there no automated way to do this?

Tiltfile
CecileRobertMichon: that's a good point, from my observations as soon as the Helm chart is applied the CRD exists but let me make sure there isn't a possible race

docs/book/src/topics/addons.md
jackfrancis: The only thing I can think of is doing this automatically in the capz control loop. That's not better. :/

Tiltfile
jackfrancis: Thinking about this more... I think the helm CLI exits success when the apiserver has admitted the full set of resources. Because the CRDs we're talking about don't have any interesting state transitions, we should be able to reliably update `default` of type `FelixConfiguration` w/ `spec.featureDetectOverride: "ChecksumOffloadBroken=true"`.

But we should be aware that we don't fully understand all the side-effects that the set of k8s components that calico is installing via helm. It's _possible_ there's some conflicting update to that same CRD resource that happens during "calico bootstrap", so if we see weird results in our E2E tests we can check here.

CecileRobertMichon: Agreed. Also not that this workaround can be reverted as soon as Calico v1.25 comes out, the fix has already been merged, just waiting for the release

docs/book/src/topics/addons.md
CecileRobertMichon: @marosset @jsturtevant are there still plans to fix this for 1.26?

marosset: We need to wait until containerd v1.7 drops :(

marosset: Let's address this as a separate change

CecileRobertMichon:
Looks like some flakes on machine provisioning

/retest

docs/book/src/topics/addons.md
jackfrancis: ACK

docs/book/src/topics/addons.md
jackfrancis: sgtm

docs/book/src/topics/addons.md
jackfrancis: I assume we're good with this @marosset, correct?

scripts/ci-entrypoint.sh
```diff
@@ -131,6 +131,59 @@ create_cluster() {
...
+    echo "Installing Calico CNI via helm"
+    "${HELM}" repo add projectcalico https://projectcalico.docs.tigera.io/charts
+    "${HELM}" install calico projectcalico/tigera-operator -f "${CALICO_VALUES:-${REPO_ROOT}/templates/addons/calico/values.yaml}" --namespace tigera-operator --create-namespace
+
+    export -f wait_for_nodes
+    timeout --foreground 1800 bash -c wait_for_nodes
+
+    # Add FeatureOverride for ChecksumOffloadBroken in FelixConfiguration.
+    # This is the recommended workaround for https://github.com/projectcalico/calico/issues/3145.
+    "${KUBECTL}" apply -f "${REPO_ROOT}"/templates/addons/calico/felix-override.yaml
```
jackfrancis, line 151: Do we need to do the Windows foo here as well for Windows scenarios that use ci-entrypoint?

docs/book/src/topics/addons.md
jsturtevant: This is due to using the calico operator.  The images were build with the release manifests they publish on github.  We could probably clean the images to remove this but there are better ways to handle a lot of this once containerd 1.7 lands.

jackfrancis:
/lgtm
/approve

take some well deserved 🏖️ @CecileRobertMichon 😎

PR #2502, opened 2022-07-23
Fix Inbound NAT rule port allocation

### LINKED ISSUE #2492: Clusterctl thinking Machines are down when they aren't according to kubectl
/kind bug

Inconsistency between ./clusterctl describe cluster and kubectl --kubeconfig \kubeconfig get nodes. This may be an error with CAPI and not CAPZ, somewhat unsure.

Clusterctl should see that the machine deployments created exist but likely cannot due to a networking problem, but all of the machine deployments are still up according to kubectl and seeing as we've successfully deployed on this cluster.

Logs:
clusterctl describe cluster capi-quickstart:
NAME                                                                READY  SEVERITY  REASON  SINCE  MESSAGE 

  

Cluster/capi-quickstart                                             True                     5h32m 

  

├─ClusterInfrastructure - AzureCluster/capi-quickstart              True                     2d10h 

  

├─ControlPlane - KubeadmControlPlane/capi-quickstart-control-plane  True                     5h32m 

  

│ └─5 Machines...                                                   True                     3d8h   See capi-quickstart-control-plane-22rkx, capi-quickstart-control-plane-k447l, ... 

└─Workers 

  

  └─MachineDeployment/capi-quickstart-md-0                          True                     19d 

  

    └─10 Machines...                                                False  Error     Failed  9d     See capi-quickstart-md-0-98984f5f4-cptnd, capi-quickstart-md-0-98984f5f4-d5wbv, ... 

kubectl --kubeconfig \capi-diff.kubeconfig get nodes:

NAME                                  STATUS   ROLES           AGE   VERSION 

capi-quickstart-control-plane-ctdrf   Ready    control-plane   19d   v1.24.0 

capi-quickstart-control-plane-d9vvx   Ready    control-plane   19d   v1.24.0 

capi-quickstart-control-plane-dskvf   Ready    control-plane   19d   v1.24.0 

capi-quickstart-control-plane-gsgb6   Ready    control-plane   19d   v1.24.0 

capi-quickstart-control-plane-jngj9   Ready    control-plane   19d   v1.24.0 

capi-quickstart-md-0-4vdjr            Ready    <none>          19d   v1.24.0 

capi-quickstart-md-0-6czbd            Ready    <none>          19d   v1.24.0 

capi-quickstart-md-0-6nhmq            Ready    <none>          19d   v1.24.0 

capi-quickstart-md-0-bt8wl            Ready    <none>          19d   v1.24.0 

capi-quickstart-md-0-cdsjd            Ready    <none>          19d   v1.24.0 

capi-quickstart-md-0-fbcdt            Ready    <none>          19d   v1.24.0 

capi-quickstart-md-0-fw8v2            Ready    <none>          19d   v1.24.0 

capi-quickstart-md-0-jx8nl            Ready    <none>          19d   v1.24.0 

capi-quickstart-md-0-p7mcn            Ready    <none>          19d   v1.24.0 

capi-quickstart-md-0-v42g7            Ready    <none>          19d   v1.24.0 

**Environment:**

- cluster-api-provider-azure version: Deployed on v1.3.2, also deployed using clusterctl v1.1.4
- Kubernetes version: (use `kubectl version`):  1.24.2
- OS (e.g. from `/etc/os-release`): Azure Cloud Shell Bash

CecileRobertMichon:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**: This fixes a TODO for inbound NAT rule reconciliation where collisions might happen. Also stops doing unnecessary `List` calls for every AzureMachine, even if there are no inbound nat rules to reconcile (eg. worker machines).

Fixes #2492 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Fix Inbound NAT rule port allocation
```

### DIFF
--- a/azure/services/inboundnatrules/inboundnatrules.go
+++ b/azure/services/inboundnatrules/inboundnatrules.go
@@ -34,7 +34,7 @@ type InboundNatScope interface {
 	azure.ClusterDescriber
 	azure.AsyncStatusUpdater
 	APIServerLBName() string
-	InboundNatSpecs(map[int32]struct{}) []azure.ResourceSpecGetter
+	InboundNatSpecs() []azure.ResourceSpecGetter
 }
 
 // Service provides operations on Azure resources.
@@ -70,6 +70,11 @@ func (s *Service) Reconcile(ctx context.Context) error {
 		return nil
 	}
 
+	specs := s.Scope.InboundNatSpecs()
+	if len(specs) == 0 {
+		return nil
+	}
+
 	ctx, cancel := context.WithTimeout(ctx, reconciler.DefaultAzureServiceReconcileTimeout)
 	defer cancel()
 
@@ -85,19 +90,23 @@ func (s *Service) Reconcile(ctx context.Context) error {
 		portsInUse[*rule.InboundNatRulePropertiesFormat.FrontendPort] = struct{}{} // Mark frontend port as in use
 	}
 
-	specs := s.Scope.InboundNatSpecs(portsInUse)
-	if len(specs) == 0 {
-		return nil
-	}
-
 	// We go through the list of InboundNatSpecs to reconcile each one, independently of the result of the previous one.
 	// If multiple errors occur, we return the most pressing one.
 	//  Order of precedence (highest -> lowest) is: error that is not an operationNotDoneError (i.e. error creating) -> operationNotDoneError (i.e. creating in progress) -> no error (i.e. created)
 	var result error
-	for _, natRule := range specs {
-		// If we are creating multiple inbound NAT rules, we could have a collision in finding an available frontend port since the newly created rule takes an available port, and we do not update portsInUse in the specs.
-		// It doesn't matter in this case since we only create one rule per machine, but for multiple rules, we could end up restarting the Reconcile function each time to get the updated available ports.
-		// TODO: We can update the available ports and recompute the specs each time, or alternatively, we could deterministically calculate the ports we plan on using to avoid collisions, i.e. rule #1 uses the first available port, rule #2 uses the second available port, etc.
+	for _, spec := range specs {
+		// Find an available SSH port for the rule.
+		sshFrontendPort, err := getAvailableSSHFrontendPort(portsInUse)
+		if err != nil {
+			return errors.Wrapf(err, "failed to find available SSH Frontend port for NAT Rule %s in load balancer %s", spec.ResourceName(), spec.OwnerResourceName())
+		}
+		natRule, ok := spec.(*InboundNatSpec)
+		if !ok {
+			result = errors.Errorf("%T is not of type InboundNatSpec", spec)
+		}
+		natRule.SSHFrontendPort = &sshFrontendPort
+		// Add the SSH frontend port to the list of ports in use
+		portsInUse[sshFrontendPort] = struct{}{}
 		if _, err := s.CreateResource(ctx, natRule, serviceName); err != nil {
 			if !azure.IsOperationNotDoneError(err) || result == nil {
 				result = err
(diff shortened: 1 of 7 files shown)


CecileRobertMichon:
This is ready for review, just need to fix unit tests

azure/services/inboundnatrules/inboundnatrules.go
```diff
@@ -85,19 +90,23 @@ func (s *Service) Reconcile(ctx context.Context) error {
...
+		if err != nil {
+			return errors.Wrapf(err, "failed to find available SSH Frontend port for NAT Rule %s in load balancer %s", spec.ResourceName(), spec.OwnerResourceName())
+		}
+		natRule, ok := spec.(*InboundNatSpec)
+		if !ok {
+			result = errors.Errorf("%T is not of type InboundNatSpec", spec)
+		}
+		natRule.SSHFrontendPort = &sshFrontendPort
+		// Add the SSH frontend port to the list of ports in use
+		portsInUse[sshFrontendPort] = struct{}{}
```
jackfrancis, line 109: So we're now keeping an accounting of this during the runtime enumeration instead of in the spec. That makes sense as any individual NAT rule won't have authoritative knowledge of the ports in use at any given time T for TCP 2201-2220.

It does seem like there may be an edge case in our accounting, however. If we store `portsInUse` and re-use during each iteration in `specs`, then we may be unnecessarily storing a `sshFrontendPort`  if the below `s.CreateResource()` fails (which is what would be responsible for definitively "reserving" the `sshFrontendPort` on the NAT rule). Because we're not aborting out of the iteration in the failure case we would continue along and then `getAvailableSSHFrontendPort(portsInUse)` would grab the _next_ port in the allowable range, leaving "holes" in the `SSHFrontendPort` across our set of NAT rules.

Not sure if this is actually a problem, but tl;dr I wonder if we should put `portsInUse[sshFrontendPort] = struct{}{}` only after we verify that `err` from `s.CreateResource(ctx, natRule, serviceName)` is `nil`.

jackfrancis:
@CecileRobertMichon can you clarify where in this change we're removing the unnecessary `AzureMachine` `List()` calls?

CecileRobertMichon:
>  can you clarify where in this change we're removing the unnecessary AzureMachine List() calls?

Here: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2502/files#diff-e4f681f94e93f028e3e4db064105e746d5cdce9ac5dad43e902453ae219e025fR75

We now return early if the length of specs returned is 0, meaning we don't need a NAT rule for that VM (which is the case for all worker machines). Before, we would do a `List` API call right away, then get the spec and return early if it was empty. This is perhaps better illustrated by this change in UTs: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2502/files#diff-be961a0a68ef279a187dd2ceef838175ad5f94f1cd64cf626779cbc329b33b11L103

azure/services/inboundnatrules/inboundnatrules.go
CecileRobertMichon: Indeed this scenario _could_ happen. However, I don't think it's a problem for the following reasons:

- the find available port function is designed to be resilient to "holes" and finds the first available port in the range. This is because when machines are deleted, their NAT rule / port allocation is also deleted, which does create "holes". The next machine that gets created needs to be able to use those free ports. If we were to always increment we'd eventually run out of free ports. Each reconcile function does a List for that reason, so we can always get the latest accurate state of things.
- In practice, we're only creating one NAT rule per control plane VM. There are rarely more than 5 control plane VMs, we have lots of extra space in the range. This also means that we're not actually looping over a set of specs since the length of specs is always 0 or 1 (but still good to future proof).
- CreateResource() is asynchronous meaning we likely will not know if the rule was actually created successfully until a later reconcile loop. Given that, it is safer to assume that a port is taken as soon as we start an operation to create a rule with it.

jackfrancis: Thanks for the explanation, I'm convinced!

jackfrancis (approved):
/lgtm

Jont828 (approved):
Thanks for making this change, the spec function looks much better without an argument!

PR #2508, opened 2022-07-26
Bump golangci-lint to v1.47.2

Prajyot-Parab:
Signed-off-by: Prajyot-Parab <prajyot.parab2@ibm.com>

 

 
**What type of PR is this?**
/kind cleanup

- Bump golangci-lint to v1.47.2

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Bump golangci-lint to v1.47.2
```


api/v1beta1/types_class.go
```diff
@@ -128,7 +128,7 @@ func (sc *SubnetClassSpec) setDefaults(cidr string) {
 }
 
 // setDefaults sets default values for SecurityGroupClass.
-func (sgc *SecurityGroupClass) setDefaults(dir SecurityRuleDirection) {
+func (sgc *SecurityGroupClass) setDefaults(dir SecurityRuleDirection) { //nolint
```
mboersma, line 131:
```suggestion
func (sgc *SecurityGroupClass) setDefaults(dir SecurityRuleDirection) { //nolint: unparam
```

We don't need to ignore all linting, just this annoying message from `unparam` pointing out that this function is always called with `SecurityRuleDirectionInbound`.

Prajyot-Parab: Fixed.

jackfrancis: I actually think it's valid lint feedback. Thoughts on changing the func thusly:

``` 
// setDefaults sets default values for SecurityGroupClass.
func (sgc *SecurityGroupClass) setDefaults() {
	for i := range sgc.SecurityRules {
		if sgc.SecurityRules[i].Direction == "" {
			sgc.SecurityRules[i].Direction = SecurityRuleDirectionInbound
		}
	}
}
```

And then update all callers accordingly to invoke the func w/ no args `()`.

Thoughts @mboersma?

mboersma: I thought it was acceptable to silence the linter, since it's possible future code might call `setDefaults` with `SecurityRuleDirectionOutbound`.

But [KISS](https://en.wikipedia.org/wiki/KISS_principle) and [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) (and the linter, and you) suggest we should simplify the code, and I am persuaded by that argument. @Prajyot-Parab what do you think?

Prajyot-Parab: @mboersma IMO refactoring the code instead of silencing the linter seems like a wise choice here, will update the changes.

jackfrancis: ❤️ all around, and especially to the future maintainer who has to invoke `setDefaults` with a different context (e.g., `SecurityRuleDirectionOutbound`) and thinks "this should have been setup flexibly in the first place!"

Prajyot-Parab: Agreed. Fixed it.

mboersma (commented):
/lgtm

mboersma (approved):
/approve

Issue #2477, opened 2022-07-12
async agentpools


Jont828:
/kind feature

Make agent pool service async as a part of #1702.

PR #2521, opened 2022-07-29
refactor: standardize AKS node pool nomenclature

jackfrancis:
**What type of PR is this?**

/kind cleanup
/kind api-change

This PR standardizes the main development surface area so that we consistently refer to AKS node pools as "node pools" and not "agent pools".

Due to the fact that the AKS API interfaces *do* in fact refer to "AgentPools" we retain some of those references when thinly wrapping our code on top of those API types and function calls.

There are some breaking API changes here, which should be O.K. given that AzureManagedCluster is still declared as "experimental".

The primary motivating factor for making these changes now is to set up the development surface area so that we are able to more effectively maintain a sane Separation of Concerns from the capz and Azure (AKS) abstractions. Secondarily, we want to make it easier to maintain capz "self-managed" from capz "externally managed", as those two functional implementations are necessarily solving slightly different problems, or solving the same problem in slightly different ways. This secondary consideration informs this example change:

`azure/services/agentpools/agentpools.go → azure/services/managedmachinepools/managedmachinepools.go`

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
refactor: standardize AKS node pool nomenclature
```


jackfrancis:
cc @zmalik @LochanRn @austintackaberry

azure/services/managedmachinepools/managedmachinepools.go
```diff
@@ -31,18 +31,18 @@ import (
 	"sigs.k8s.io/cluster-api-provider-azure/util/tele"
 )
 
-const serviceName = "agentpools"
+const serviceName = "managedmachinepools"
```
CecileRobertMichon, line 34: the services under azure/services match the Azure resource type they are reconciling so here it would be more consistent to keep the agenpools naming since we're reconciling Azure agentpool resources with the Azure APIs.

CecileRobertMichon: To be clear by "here" I meant the entire file/package name. "ManagedMachinePools" is a purely CAPZ concept, it's the CRD that implements `MachinePool` for the Azure managed infra provider. 

Whatever is in azure/services should refer to Azure SDK terms. Whether that's agentpools or nodepools I don't know, but it seems that your intention here is to rename it to nodepools. 

Side note: it's confusing where the node pools term comes from since the APIs refer to "agent pools" but everything else from user facing docs to CLI refers to "node pools" https://docs.microsoft.com/en-us/azure/aks/use-multiple-node-pools... Is this an inconsistency in the AKS APIs?

CecileRobertMichon:
/cc @Jont828 

this conflicts with #2477 FYI

azure/services/managedmachinepools/managedmachinepools.go
jackfrancis: My intention with this particular bit of work is to make it easier to distinguish from "self-managed"-dependent services.

The AKS API is definitely called "agentpools", and that is inherited from acs-engine, which originally defined the semantic concepts of "MasterProfile" and "AgentPoolProfile". Because the acs-engine code evolved to become the authoritative type system for ACS, and then AKS, those concepts stuck around when the feature "node pool" (as it is referred to everywhere from a user perspective) was introduced.

As far as I can tell, all (well, maybe not _all_) user-facing documentation, including the Azure CLI, will refer to a "node pool" as the first class representation of what we're talking about. IMO the most meaningful places that we can update in our code to "node pool" are the places that capz users will be interacting with (CRD properties, log files, capz docs). When interesting things happen to folks' capz-managed AKS clusters, we want to give them an easy path to investigate in all the standard places, like the official AKS documentation. I think when we refer users to "agent pools" we will likely inhibit their ability to get more comprehensive info from the AKS documentation and user community, and/or redirect them to low-level API documentation instead, which we don't want.

We also want to make things consistent (at present "node pool" and "agent pool" are used interchangeably thoughout the AzureManagedCluster implementation, with no obvious rhyme or reaason).

As far as how to organize structure under `azure/services`, I would suggest we can improve things by adding the parent organizational namespace. E.g.:

```
- azure/services/agentpools --> azure/services/containerservice/agentpools
- azure/services/availabilitysets --> azure/services/compute/availabilitysets
- azure/services/disks --> azure/services/compute/disks
- azure/services/bastionhosts --> azure/services/network/bastionhosts
```

Etc.

Without that namespacing we are vulnerable to a child-level collision (these API libraries are clearly not exactly named to match to names of the service products, nor are they strictly cross-referenced across services to enforce uniqueness at the leaf node level), or the type of non-obvious organization of "agentpools" under `azure/services`, which doesn't indicate any AKS-specific service dependency until you are familiar with the particulars of the codebase.

If the above "node pool" vs "agent pool" makes sense let's definitely prioritize getting @Jont828's async refactor in soon and then I can do a more fine-grained audit of the user-facing surface area to ensure we're consistently referring to AKS "node pools" as "node pools".

Not sure if the proposal to move all the service code under `azure/services` underneath its parent constitutes a breaking change, I do think that would be a non-trivial improvement to better organize this stuff for the long term.

CecileRobertMichon:
@jackfrancis do we still want this change?

jackfrancis:
@CecileRobertMichon, no, I'll take up a subset of this scope in a future PR, pending completion of other higher priority items.

Issue #2484, opened 2022-07-15
AzureManagedControlPlane with existing vnet and existing subnet attempts to delete subnet when cluster is deleted


mtougeron:
/kind bug

[Before submitting an issue, have you checked the [Troubleshooting Guide](https://capz.sigs.k8s.io/topics/troubleshooting.html)?]

**What steps did you take and what happened:**

If you use a `AzureManagedControlPlane` with an existing `virtualNetwork.name` and an existing `virtualNetwork.subnet.name` it will properly use the existing resources. However, when deleting the cluster it will attempt to delete the subnet specified in `virtualNetwork.subnet.name`. You are then stuck and unable to finish the deletion of the cluster without also losing the existing subnet (if it already has resources in it).

**What did you expect to happen:**

While the documentation doesn't specifically say that you can use an existing subnet, I read it as implied because you can use an existing vnet. If that's not accurate, I can open a PR to update the documentation accordingly. However, it would be nice if you could use an existing subnet inside of an existing vnet when creating new AKS clusters.

In our environment, we have a single subnet in our vnets that we'd like to use for new AKS clusters being created with CAPZ as we do a migration so that we don't have to redo any peering connections.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

This seems similar to https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1590

**Environment:**

- cluster-api-provider-azure version: registry.k8s.io/cluster-api-azure/cluster-api-azure-controller:v1.4.0
- Kubernetes version: (use `kubectl version`): 1.22.11
- OS (e.g. from `/etc/os-release`): n/a

jackfrancis:
Hi @mtougeron, I think the first thing we want to do is validate that AKS itself supports the create flow you outlined (install into a custom (pre-existing) VNET + custom (pre-existing) subnet. Does the delete flow using vanilla AKS preserve both the previously existing subnet + VNET, only deleting the AKS-specific resources?

mtougeron:
> AKS itself supports the create flow you outlined (install into a custom (pre-existing) VNET + custom (pre-existing) subnet

It worked with CAPZ at least but I haven't tried with doing it via the CLI or Azure Portal. I'll try that.

> Does the delete flow using vanilla AKS preserve both the previously existing subnet + VNET, only deleting the AKS-specific resources

[...]

PR #2522, opened 2022-07-29
Check if the vnet is managed

mtougeron:
**What type of PR is this?**
/kind bug

Stops CAPZ from trying to delete subnets & vnets that are not managed by it when using an AzureManagedControlPlane

Fixes https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/2484

I'm not sure how to handle the error situation at line 334. Currently I have it fail back to the legacy behavior but that doesn't feel correct to me but I don't have a better answer. Any suggestions/guidance would be greatly appreciated.

Thank you to @CecileRobertMichon for identifying the root cause of https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/2484

**TODOs**:

- [x] squashed commits
- [n/a] includes documentation
- [n/a] adds unit tests

**Release note**:

```release-note
CAPZ will no longer try to delete subnets & vnets that are not managed by an AzureManagedControlPlane
```


azure/scope/managedcontrolplane.go
```diff
@@ -328,7 +328,12 @@ func (s *ManagedControlPlaneScope) IsIPv6Enabled() bool {
 
 // IsVnetManaged returns true if the vnet is managed.
 func (s *ManagedControlPlaneScope) IsVnetManaged() bool {
-	return true
+	isManaged, err := virtualnetworks.New(s).IsManaged(context.TODO())
```
CecileRobertMichon, line 331: instead of making a new service on the fly in here, have you considered putting this directly into the subnets.go `IsManaged` function? That would allow for proper handling of the error and you'd be able to pass in the context instead of using context.TODO.

The downside is that it would also affect self-managed clusters but the current implementation of `IsVnetManaged()` in `cluster.go` uses the value of ID and tags in the Vnet spec part of `AzureCluster.Spec` which  is not ideal since Spec could be modified by a user. It would be better to always check from the source of truth like we're doing here, i.e. the actual Azure vnet resource.

I would also be careful with where `IsVnetManaged` is being called since we are now making an Azure API call every time (more costly than a hardcoded return true :) ) and from glancing at the subnets code I can already see a few redundant uses that could be removed (for example we're embedding it in the SubnetSpec but then never using it [here](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/2a711856de20735337d59ee391202cb45d8f49e2/azure/scope/managedcontrolplane.go#L256)). Long term we might even want to cache the result so we don't have to make the same API call multiple time per reconcile loop (GETs are cheap but still...)

mtougeron: > have you considered putting this directly into the subnets.go IsManaged function

I looked at that and it's already calling the Scope's `IsVnetManaged` so I took that to mean that the `AzureManagedControlPlane` Scope is the one that needed to be updated since it's working for `AzureCluster`. I'm new enough to the code though that I might be misunderstanding the way the resource Scope(s) work.

> for example we're embedding it in the SubnetSpec but then never using it

I had been wondering about that myself. It led me through a few false starts when trying to fix this :)

> Long term we might even want to cache the result so we don't have to make the same API call multiple time per reconcile loop

Is there anything special to the way it does caching or is it just a new static field on the ManagedControlPlaneScope struct (like the `MachineCache` does)? I'm happy to add that now if you'd like.

CecileRobertMichon: > I looked at that and it's already calling the Scope's IsVnetManaged so I took that to mean that the AzureManagedControlPlane Scope is the one that needed to be updated since it's working for AzureCluster

totally reasonable, that's what my first instinct would have told me too. I just happen to have the context that `IsVnetManaged` is legacy code and that we've reworked the way to check if a resource is managed since.

> Is there anything special to the way it does caching or is it just a new static field on the ManagedControlPlaneScope struct (like the MachineCache does)? I'm happy to add that now if you'd like.

Nothing special, just a way to save the value on scope just like we do for VM SKUs in MachineCache. If you want to try and add it in this PR go for it!

mtougeron: I updated the `subnets.go`'s `IsManaged` however I'm running into an order of operations issue where it's calling a check to the subnet's IsManaged after the Vnet has already been deleted (when not using BYO Vnet). I added a fall back in that scenario to the old logic, which allows it to move on, but I'm not happy about it. I will debug a bit deeper on Monday to find a better way.

jackfrancis: I tried both approaches (update `IsVnetManaged()` in `azure/scope/managedcontrolplane.go` and update `IsManaged` in `azure/services/subnets/subnets.go`. They both work for this particular use case, but in the `azure/scope/managedcontrolplane.go` scenario we do end up making a bunch of regular GET calls which is not desireable. 

The reason for this seems obvious enough, as the subnets implementation only invokes `IsManaged()` in its `Delete` reconciler.

So functionally I think the current solution solves the bug. My current concern is the following scenario:

- AKS template with a reference to a pre-existing VNET
- no reference to a pre-existing subnet (so we expect a "default" subnet to be created in the pre-existing VNET)

Is that actually a known-working scenario? If so, it seems like relying on the VNET's "managed" status (looking for the owned tag) to stand in for the subnet's "managed" status wouldn't work. We'd actually *want* to clean up such a subnet is this scenario.

If I'm describing something that isn't possible (for example, if it's required that when you BYO VNET you also must BYO subnet), then this lgtm.

mtougeron: > If I'm describing something that isn't possible (for example, if it's required that when you BYO VNET you also must BYO subnet), then this lgtm.

Good point. Let me see if there's something I can do for the scenario when you don't BYO subnet. It's definitely harder since you can't tag/label subnets. :/

jackfrancis: I think this documentation accurately describes the "self-managed" requirements (not necessarily AKS) for vnets and subnets:

https://capz.sigs.k8s.io/topics/custom-vnet.html#pre-existing-vnet-and-subnets

Specifically:

> When providing a vnet, it is required to also provide the two subnets that should be used for control planes and nodes.

We can definitely add an AKS-contextualized version of that for the AKS scenario (for example, there's no "control plane nodes" in an AKS scenario) that matches how BYO subnet actually works with AKS, e.g. from the `az` CLI help text:

```
    --vnet-subnet-id                           : The ID of a subnet in an existing VNet into which
                                                 to deploy the cluster.
```

tl;dr my scenario described above where you bring your own VNET but rely upon capz to create the subnet is not actually a relevant AKS scenario

mtougeron: > tl;dr my scenario described above where you bring your own VNET but rely upon capz to create the subnet is not actually a relevant AKS scenario

Is that going to cause any backwards compatibility issues? If not, I won't worry about it and leave it as-is.

jackfrancis: @CecileRobertMichon and I were discussing this offline (we should probably try not to do that as much as possible!) and she had the idea to simply turn the subnet delete into a no-op as a potentially more elegant solution to this. I've scaffolded up that idea here:

- https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2525

What do we think about that?

mtougeron: I was wondering something similar when I noticed that the vnet deletion finished and deleted the subnet before the subnet deletion fully reconciled. The approach works for me but I defer to the folks with more experience in the code base. :)

jackfrancis: Yes, your observation/insight there was key. Not only does the VNET delete its subnets automatically, but in an environment where concurrent reconcilers are in-progress, it means that if we try do delete subnets and VNETs discretely we have to account for the actual Azure dependency graph relationship between the two and either (1) strictly serialize the order of deletions (we aren't and don't want to do this IMO) or (2) add sane error handling to account for the fact that the subnet may have already been deleted by the VNET and thus return successful from `Delete`.

So it is simpler to not delete the subnet discretely, it seems. However, given the architecture of the type system that Azure resources conform to (subnets are something we treat in capz like a first class service resource) it's a bit "weird" that we're no-op'ing during `Delete`; so I'll track that PR and see what folks think about that approach.

I didn't mean by submitting my own PR to prefer *my* code over *yours*, hopefully that didn't come across! Whichever solution lands (and we'll land it soon) it's certainly entirely due to your observation, investigation, and persistence that we can put this bug behind us. Thanks again for finding this gremlin!

mtougeron: > I didn't mean by submitting my own PR to prefer my code over yours, hopefully that didn't come across!

Not at all. Team effort! I was just waiting on closing this PR until I was sure the other one was getting merged.

> Thanks again for finding this gremlin!

Glad I could help. :)

jackfrancis (approved):
/lgtm
/approve

mtougeron:
> pull-cluster-api-provider-azure-apidiff failure

ah shoot. Forgot to regenerate stuff once I refactored from my first approach. :(

mtougeron:
Looks like I didn't test the non-managed cluster as thoroughly as I thought I did. I'll revisit that now that I can see the e2e failures.

mtougeron:
Ya! All the tests are passing. Let me know what you think of the refactored approach.

azure/services/virtualmachineimages/cache.go
```diff
@@ -119,7 +119,7 @@ func (c *Cache) Get(ctx context.Context, location, publisher, offer, sku string)
 			return compute.ListVirtualMachineImageResource{}, err
 		}
 	} else {
-		log.Info("VM images cache hit", "location", key.location, "publisher", key.publisher, "offer", key.offer, "sku", key.sku)
+		log.V(4).Info("VM images cache hit", "location", key.location, "publisher", key.publisher, "offer", key.offer, "sku", key.sku)
```
mtougeron, line 122: In case you're wondering, see https://kubernetes.slack.com/archives/CEX9HENG7/p1659637807401169?thread_ts=1659637507.302629&cid=CEX9HENG7 for why I changed this.

azure/scope/managedcontrolplane.go
```diff
@@ -254,6 +254,7 @@ func (s *ManagedControlPlaneScope) SubnetSpecs() []azure.ResourceSpecGetter {
 			VNetName:          s.Vnet().Name,
 			VNetResourceGroup: s.Vnet().ResourceGroup,
 			IsVNetManaged:     s.IsVnetManaged(),
+			IsSubnetManaged:   s.Vnet().Tags.HasOwned(s.ClusterName()),
```
CecileRobertMichon, line 257: unfortunately this won't work for managed clusters because `.Vnet()` [returns a new Vnet](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2522/files#diff-cfd94adbc2ff767742e457e64836e93e88c6e03c3a64c04619c54b7ed76f284eR199) with no tags set since the AzureManagedControlPlane spec doesn't actually have a Vnet spec that is persisted, so this will return `false` every time. The only way to is to check the actual tags on the vnet by calling the virtualnetworks service `IsManaged` function.

mtougeron: Okay, I'll go back to the previous approach and see if I can spend some more time trying to figure out how to properly mock and pass around the virtualnetworks service & scope while in the tests for the subnets. But TBH it kind of feels like that's crossing some boundary lines by doing that. I'll noodle on it and see what I come up with...

CecileRobertMichon: Yeah it's definitely not ideal 😕 Happy to brainstorm in a pairing session next week if that helps. 

In other news I don't know if you saw my [comment](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2525#issuecomment-1207082254 ) in the other PR but the alternative strategy of not deleting subnets at all might not actually work out because Azure isn't happy about route tables attached to a subnet being deleted before the subnet (and the route table has to be deleted before the vnet so can know if it's managed) so this complicates things quite a bit. This current PR might be our best way forward.

mtougeron: > Happy to brainstorm in a pairing session next week if that helps.

That might be very helpful. Thank you. I'll let you know where my brain ends up as it simmers over the weekend.

> the alternative strategy of not deleting subnets at all might not actually work out

yeah, I saw that. I was originally going to suggest that we go with that approach instead but then I saw the other comment and got all :( face.

This should be doable, it's more my limited experience with mocking things in Go than anything else.

mtougeron:
Closing in favor of @jackfrancis 's solution

PR #2525, opened 2022-08-01
don’t explicitly delete subnet, routetable, nsg

jackfrancis:
**What type of PR is this?**

/kind bug

This PR changes the subnet `Delete` implementation so that it is a no-op. There are a couple reasons why this is an attractive idea:

1. A subnet, properly speaking, are a dependent/child/sub- resource of a Virtual Network. Whenever we delete a Virtual Network, we will always delete any subnets it contains.
2. Subnets don't support resource tags, which means that we can't positively identify whether or not a subnet resource is "owned" by capz (and should thus be deleted by a capz controller when its relevant capz CRD resource is deleted) after it is created. Thus we aren't able to properly, conditionally delete a subnet _only if it is owned by capz_.

Unfortunately there are a couple undesirable side-effects of this change:

1. Making a resource `Delete` operation a static no-op in practice is hacky
2. Because subnet like all service resources fulfills the type `azure.ServiceReconciler`, we need to implement `IsManaged()`, which as stated above we don't actually know, and in fact with this change we actually aren't going to invoke. It seems like the least hacky way to do this is to always return `false`.

Fixes https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/2484

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/azure/services/routetables/routetables.go
+++ b/azure/services/routetables/routetables.go
@@ -96,39 +96,11 @@ func (s *Service) Reconcile(ctx context.Context) error {
 
 // Delete deletes route tables.
 func (s *Service) Delete(ctx context.Context) error {
-	ctx, log, done := tele.StartSpanWithLogger(ctx, "routetables.Service.Delete")
+	_, log, done := tele.StartSpanWithLogger(ctx, "routetables.Service.Delete")
 	defer done()
 
-	ctx, cancel := context.WithTimeout(ctx, reconciler.DefaultAzureServiceReconcileTimeout)
-	defer cancel()
-
-	// Only delete the route tables if their lifecycle is managed by this controller.
-	// route tables are managed if and only if the vnet is managed.
-	if managed, err := s.IsManaged(ctx); err == nil && !managed {
-		log.V(4).Info("Skipping route table deletion in custom vnet mode")
-		return nil
-	} else if err != nil {
-		return errors.Wrap(err, "failed to check if route tables are managed")
-	}
-
-	specs := s.Scope.RouteTableSpecs()
-	if len(specs) == 0 {
-		return nil
-	}
-
-	// We go through the list of RouteTableSpecs to delete each one, independently of the result of the previous one.
-	// If multiple errors occur, we return the most pressing one
-	// order of precedence is: error deleting -> deleting in progress -> deleted (no error)
-	var result error
-	for _, rtSpec := range specs {
-		if err := s.DeleteResource(ctx, rtSpec, serviceName); err != nil {
-			if !azure.IsOperationNotDoneError(err) || result == nil {
-				result = err
-			}
-		}
-	}
-	s.Scope.UpdateDeleteStatus(infrav1.RouteTablesReadyCondition, serviceName, result)
-	return result
+	log.V(4).Info("Route table will be deleted when its parent Virtual Network is deleted")
+	return nil
 }
 
 // IsManaged returns true if the route tables' lifecycles are managed.
--- a/azure/services/routetables/routetables_test.go
+++ b/azure/services/routetables/routetables_test.go
@@ -140,55 +140,10 @@ func TestDeleteRouteTable(t *testing.T) {
 		name          string
 		tags          infrav1.Tags
 		expectedError string
-		expect        func(s *mock_routetables.MockRouteTableScopeMockRecorder, r *mock_async.MockReconcilerMockRecorder)
 	}{
 		{
-			name:          "noop if no route table specs are found",
+			name:          "delete always returns nil error",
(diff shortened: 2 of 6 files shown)


sonasingh46:
+1 
I think if we are not able to determine if the subnet was created by capz or not and plus if deleting the vnet deletes the subnet - it looks good to me making the delete as a no-op

sonasingh46:
I was looking at the logs -- looks like that the private cluster that got created was not successfully deleted. 
As per the logs -- 
- The delete command worked fine as did not see any error log for that. 
See https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/test/e2e/azure_privatecluster.go#L164

- Looks like the cluster resource is still hanging out there  -- as per the logs it error out here
https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/test/e2e/azure_privatecluster.go#L179

I looked around the artifacts and logs but that did not help figure out why the cluster resource was not yet deleted. [May be we can also put the CR resource dumps]

Anyways, one way would be to test it locally and see live what it going around.

CecileRobertMichon:
> I was looking at the logs -- looks like that the private cluster that got created was not successfully deleted.

That's an important one to look into as it may be the only coverage we get for this change. All other clusters get deleted by deleting the entire resource group so we never actually go down the vnet/subnet delete code paths. However, the private cluster test creates a cluster in a pre-existing RG, which means it happens to test the route of deleting each resource individually.

jackfrancis:
> > I was looking at the logs -- looks like that the private cluster that got created was not successfully deleted.
> 
> That's an important one to look into as it may be the only coverage we get for this change. All other clusters get deleted by deleting the entire resource group so we never actually go down the vnet/subnet delete code paths. However, the private cluster test creates a cluster in a pre-existing RG, which means it happens to test the route of deleting each resource individually.

The actual test in the private cluster scenario executes against the Cluster resource, but I get what you're saying that this is the only scenario that would trigger a "Don't delete the Virtual Network" path that would functionally intersect with these changes.

I'll try to run these tests locally (my manual local BYO VNET scenario didn't repro anything like this).

If this continues to stump us we should probably merge #2522 first and then continue to work on this in the background.

CecileRobertMichon:
> I'll try to run these tests locally (my manual local BYO VNET scenario didn't repro anything like this).

To repro this you'll likely need to BYO RG but _not_ the VNet since we want VNet deletion to get triggered

jackfrancis:
I did a quick local build-out using tilt and couldn't repro.

1. Pre-created the RG
2. Configured `$AZURE_RESOURCE_GROUP` input in the "aks" flavor to use the above existing RG
3. Created a cluster using the "aks" flavor
4. Verified the cluster was up and running
5. Deleted the cluster

```sh
$ time k delete cluster aks-mine
cluster.cluster.x-k8s.io "aks-mine" deleted

real	6m9.683s
user	0m0.168s
sys	0m0.117s
```

Delete succeeded in a timely manner and the pre-existing RG is still around.

So there must be something else about the private cluster scenario that's happening. (It happened twice in succession in E2E, so I assume something interesting is going on.)

@CecileRobertMichon should we consider merging #2522 first while we figure out what's going on here so that we can unblock @mtougeron ?

CecileRobertMichon:
@jackfrancis did some digging and found this particular cluster failed to delete because of the following error:

```
error\":{\"code\":\"InUseRouteTableCannotBeDeleted\",\"message\":\"Route table capz-e2e-6x8c3t-private-node-routetable is in use and cannot be deleted.\"
```

I'm guessing this has to do with the order of operations. We delete subnet, route table, nsg and then finally vnet. Now that we are no longer deleting subnets separately, the route tables and nsgs get deleted before the subnet which fails because they are still in use by the subnet. 

Pretty sure this only affects self-managed clusters (we don't create route tables for AKS) which is why you couldn't repro with an AKS cluster.

jackfrancis:
/retitle don’t explicitly delete subnet, routetable, nsg

jackfrancis:
@CecileRobertMichon @mtougeron I've modified this PR to exclude route tables and NSGs as well as subnets from explicit deletion (they will all be deleted by VNET).

My initial (self-managed) tests suggest that this change is functionally equivalent (i.e., delete VNET _does_ delete the subnet and its associated route table(s) + NSG(s)). And in fact this will solve the equivalent problem we are trying to solve for subnets: when a pre-existing VNET is used in an AKS scenario, any pre-existing route tables and NSGs will be preserved.

CecileRobertMichon:
/close 

Closing in favor of #2543

PR #2529, opened 2022-08-02
Standardize more package import aliases

### LINKED ISSUE #2289: Enforce consistent import names across the module
/kind feature

Right now, packages are often imported using different name between the packages, which makes it more difficult to navigate and refactor the code base. Example:
```console
$ git grep sigs.k8s.io/cluster-api-provider-azure/api/v1beta1 | cut -d: -f2 | grep sigs | sort | uniq -c | sort -h
      1         apiv1beta1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
      2         clusterapiproviderazureapiv1beta1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
      2         infrav1alpha4 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
      8         infrav1beta1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
      8         "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
     28         v1beta1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
    142         infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
```

Package `sigs.k8s.io/cluster-api-provider-azure/api/v1beta1` is imported 7 different ways, including craziness like `infrav1alpha4 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"`

To promote better consistency and predictability, I suggest using `importas` linter https://github.com/golangci/golangci-lint/blob/7c1d8b75bd4aed28783a7b6a4dd852b7be9dc3d5/.golangci.example.yml#L903-L922, which allows enforcing style of imports.

**Environment:**

- cluster-api-provider-azure version: c5b5f632

mboersma:
**What type of PR is this?**

/kind cleanup

Standardizes several popular import aliases by adding them to the `importas` linter configuration. This is an area of code where creativity is ultimately unhelpful. The CAPI sections were copied from [CAPI's golangci config](https://github.com/kubernetes-sigs/cluster-api/blob/main/.golangci.yml#L59).

Refs #2289

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -72,6 +72,34 @@ linters-settings:
       # Controller Runtime
       - pkg: sigs.k8s.io/controller-runtime
         alias: ctrl
+      # CAPI
+      - pkg: sigs.k8s.io/cluster-api/api/v1alpha3
+        alias: clusterv1alpha3
+      - pkg: sigs.k8s.io/cluster-api/api/v1alpha4
+        alias: clusterv1alpha4
+      - pkg: sigs.k8s.io/cluster-api/api/v1beta1
+        alias: clusterv1
+      # CAPI exp
+      - pkg: sigs.k8s.io/cluster-api/exp/api/v1alpha3
+        alias: expv1alpha3
+      - pkg: sigs.k8s.io/cluster-api/exp/api/v1alpha4
+        alias: expv1alpha4
+      - pkg: sigs.k8s.io/cluster-api/exp/api/v1beta1
+        alias: expv1
+      # CAPZ
+      - pkg: sigs.k8s.io/cluster-api-provider-azure/api/v1alpha3
+        alias: infrav1alpha3
+      - pkg: sigs.k8s.io/cluster-api-provider-azure/api/v1alpha4
+        alias: infrav1alpha4
+      - pkg: sigs.k8s.io/cluster-api-provider-azure/api/v1beta1
+        alias: infrav1
+      # CAPZ exp
+      - pkg: sigs.k8s.io/cluster-api-provider-azure/exp/api/v1alpha3
+        alias: infrav1alpha3exp
+      - pkg: sigs.k8s.io/cluster-api-provider-azure/exp/api/v1alpha4
+        alias: infrav1alpha4exp
+      - pkg: sigs.k8s.io/cluster-api-provider-azure/exp/api/v1beta1
+        alias: infrav1exp
   godot:
     #   declarations - for top level declaration comments (default);
     #   toplevel     - for top level comments;
--- a/azure/services/agentpools/agentpools.go
+++ b/azure/services/agentpools/agentpools.go
@@ -24,7 +24,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/services/containerservice/mgmt/2021-05-01/containerservice"
 	"github.com/google/go-cmp/cmp"
 	"github.com/pkg/errors"
-	infrav1alpha4 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
+	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
 	"sigs.k8s.io/cluster-api-provider-azure/azure"
 	"sigs.k8s.io/cluster-api-provider-azure/azure/converters"
 	"sigs.k8s.io/cluster-api-provider-azure/util/maps"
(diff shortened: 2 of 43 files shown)


azure/services/agentpools/agentpools.go
```diff
@@ -24,7 +24,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/services/containerservice/mgmt/2021-05-01/containerservice"
 	"github.com/google/go-cmp/cmp"
 	"github.com/pkg/errors"
-	infrav1alpha4 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
```
mboersma, line 27: This was a misleading alias.

.golangci.yml
```diff
@@ -72,6 +72,34 @@ linters-settings:
...
+      # CAPI
+      - pkg: sigs.k8s.io/cluster-api/api/v1alpha3
+        alias: clusterv1alpha3
+      - pkg: sigs.k8s.io/cluster-api/api/v1alpha4
+        alias: clusterv1alpha4
+      - pkg: sigs.k8s.io/cluster-api/api/v1beta1
+        alias: clusterv1
+      # CAPI exp
+      - pkg: sigs.k8s.io/cluster-api/exp/api/v1alpha3
+        alias: expv1alpha3
```
invidian, line 84: Maybe it should be `expclusterv1alpha3` to be consistent with the above?

.golangci.yml
```diff
@@ -72,6 +72,34 @@ linters-settings:
...
+        alias: infrav1alpha3
+      - pkg: sigs.k8s.io/cluster-api-provider-azure/api/v1alpha4
+        alias: infrav1alpha4
+      - pkg: sigs.k8s.io/cluster-api-provider-azure/api/v1beta1
+        alias: infrav1
+      # CAPZ exp
+      - pkg: sigs.k8s.io/cluster-api-provider-azure/exp/api/v1alpha3
+        alias: infrav1alpha3exp
+      - pkg: sigs.k8s.io/cluster-api-provider-azure/exp/api/v1alpha4
+        alias: infrav1alpha4exp
```
invidian, line 100: It's `expv1alpha4` with `exp` _prefix_, but `infrav1alpha4exp` with `exp` _suffix_, maybe we can make it more consistent?

.golangci.yml
```diff
@@ -72,6 +72,34 @@ linters-settings:
       # Controller Runtime
       - pkg: sigs.k8s.io/controller-runtime
         alias: ctrl
+      # CAPI
+      - pkg: sigs.k8s.io/cluster-api/api/v1alpha3
```
invidian, line 76: I wonder, if source file imports only single package named `v1alpha3`, will the linter still require an alias?

invidian (commented):
Thanks for the PR @mboersma! I left some thoughts in the comments.

.golangci.yml
mboersma: It will, because we have [`no-unaliased: true`](https://golangci-lint.run/usage/linters/#importas) set in [.golangci.yml](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/.golangci.yml#L59).

.golangci.yml
mboersma: The CAPI rules for `importas` were copied from CAPI's golangci-lint configuration. I think we should just use them as-is to make it easier to move between codebases. I notice that's what CAPA did as well.

.golangci.yml
invidian: All good if that's deliberate. I think I like it this way, similar to `,` ending the last entry in the lists. It won't make you change unrelated stuff when you add new import. Feel free to mark as resolved.

.golangci.yml
mboersma: Yes, we could--I chose the `infrav1exp` pattern for CAPZ imports because it was already the most-used, so that minimized the churn in the codebase. But `infraexpv1` and `infraexpv1alpha4` are arguably more consistent with CAPI's imports.

.golangci.yml
invidian: Well, some say you should always question things and there is always room for the improvement... :smile: But perhaps less optimal consistency is better than better inconsistency, so I guess that's OK. Thanks for explaining! Feel free to mark as resolved.

mboersma:
@invidian I noticed that this only really fixes half of the problem, since there are still randomly-named imports of these packages in `conversion.go` and related files. We completely ignore those files in linting, but that's overkill as far as I can tell; I have a PR ready to follow this one that enables linting and standardizes those imports as well.

When we're all happy with the naming here, I'll update that branch and put a followup PR out.

jackfrancis (approved):
/lgtm
/approve

PR #2543, opened 2022-08-10
AKS: enable isVnetManaged, add caching

### LINKED ISSUE #2484: AzureManagedControlPlane with existing vnet and existing subnet attempts to delete subnet when cluster is deleted

jackfrancis:
**What type of PR is this?**

/kind bug
/kind feature

This PR enables the `IsVnetManaged()` method for AKS scenarios, which results in actual GET calls against the Azure API to determine in real-time whether the expected capz owner tag (tag is applied one time during cluster create) is present on the Virtual Network.

This will address the subnet delete issue during custom VNET scenarios.

Additionally, in order to account for additional calls against the Azure APIs we've implemented a local cache to ensure we don't call the Azure APIs more than once for VNET information per reconciliation.

Fixes #2484

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
AKS: enable isVnetManaged, add caching
```

### DIFF
--- a/azure/scope/cluster.go
+++ b/azure/scope/cluster.go
@@ -56,6 +56,7 @@ type ClusterScopeParams struct {
 	Client       client.Client
 	Cluster      *clusterv1.Cluster
 	AzureCluster *infrav1.AzureCluster
+	Cache        *ClusterCache
 }
 
 // NewClusterScope creates a new Scope from the supplied parameters.
@@ -87,6 +88,10 @@ func NewClusterScope(ctx context.Context, params ClusterScopeParams) (*ClusterSc
 		}
 	}
 
+	if params.Cache == nil {
+		params.Cache = &ClusterCache{}
+	}
+
 	helper, err := patch.NewHelper(params.AzureCluster, params.Client)
 	if err != nil {
 		return nil, errors.Errorf("failed to init patch helper: %v", err)
@@ -98,6 +103,7 @@ func NewClusterScope(ctx context.Context, params ClusterScopeParams) (*ClusterSc
 		Cluster:      params.Cluster,
 		AzureCluster: params.AzureCluster,
 		patchHelper:  helper,
+		cache:        params.Cache,
 	}, nil
 }
 
@@ -105,10 +111,16 @@ func NewClusterScope(ctx context.Context, params ClusterScopeParams) (*ClusterSc
 type ClusterScope struct {
 	Client      client.Client
 	patchHelper *patch.Helper
+	cache       *ClusterCache
 
 	AzureClients
 	Cluster      *clusterv1.Cluster
 	AzureCluster *infrav1.AzureCluster
+}
+
+// ClusterCache stores ClusterCache data locally so we don't have to hit the API multiple times within the same reconcile loop.
+type ClusterCache struct {
+	isVnetManaged *bool
 }
 
 // BaseURI returns the Azure ResourceManagerEndpoint.
@@ -524,7 +536,12 @@ func (s *ClusterScope) Vnet() *infrav1.VnetSpec {
 
 // IsVnetManaged returns true if the vnet is managed.
 func (s *ClusterScope) IsVnetManaged() bool {
-	return s.Vnet().ID == "" || s.Vnet().Tags.HasOwned(s.ClusterName())
+	if s.cache.isVnetManaged != nil {
+		return to.Bool(s.cache.isVnetManaged)
+	}
+	isVnetManaged := s.Vnet().ID == "" || s.Vnet().Tags.HasOwned(s.ClusterName())
+	s.cache.isVnetManaged = to.BoolPtr(isVnetManaged)
+	return isVnetManaged
 }
 
 // IsIPv6Enabled returns true if IPv6 is enabled.
--- a/azure/scope/managedcontrolplane.go
+++ b/azure/scope/managedcontrolplane.go
@@ -22,6 +22,7 @@ import (
 	"strings"
 
 	"github.com/Azure/go-autorest/autorest"
+	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/pkg/errors"
 	"golang.org/x/mod/semver"
 	corev1 "k8s.io/api/core/v1"
@@ -51,6 +52,7 @@ type ManagedControlPlaneScopeParams struct {
 	Cluster             *clusterv1.Cluster
 	ControlPlane        *infrav1exp.AzureManagedControlPlane
 	ManagedMachinePools []ManagedMachinePool
+	Cache               *ManagedControlPlaneCache
 }
 
 // NewManagedControlPlaneScope creates a new Scope from the supplied parameters.
@@ -82,6 +84,10 @@ func NewManagedControlPlaneScope(ctx context.Context, params ManagedControlPlane
 		}
 	}
 
+	if params.Cache == nil {
+		params.Cache = &ManagedControlPlaneCache{}
+	}
+
 	helper, err := patch.NewHelper(params.ControlPlane, params.Client)
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to init patch helper")
@@ -94,6 +100,7 @@ func NewManagedControlPlaneScope(ctx context.Context, params ManagedControlPlane
 		ControlPlane:        params.ControlPlane,
 		ManagedMachinePools: params.ManagedMachinePools,
 		patchHelper:         helper,
+		cache:               params.Cache,
 	}, nil
 }
 
@@ -102,11 +109,17 @@ type ManagedControlPlaneScope struct {
 	Client         client.Client
 	patchHelper    *patch.Helper
 	kubeConfigData []byte
+	cache          *ManagedControlPlaneCache
 
 	AzureClients
 	Cluster             *clusterv1.Cluster
 	ControlPlane        *infrav1exp.AzureManagedControlPlane
 	ManagedMachinePools []ManagedMachinePool
+}
+
+// ManagedControlPlaneCache stores ManagedControlPlane data locally so we don't have to hit the API multiple times within the same reconcile loop.
+type ManagedControlPlaneCache struct {
+	isVnetManaged *bool
 }
 
 // ResourceGroup returns the managed control plane's resource group.
@@ -328,7 +341,20 @@ func (s *ManagedControlPlaneScope) IsIPv6Enabled() bool {
 
 // IsVnetManaged returns true if the vnet is managed.
 func (s *ManagedControlPlaneScope) IsVnetManaged() bool {
-	return true
+	if s.cache.isVnetManaged != nil {
+		return to.Bool(s.cache.isVnetManaged)
+	}
+	// TODO refactor `IsVnetManaged` so that it is able to use an upstream context
+	// see https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/2581
+	ctx := context.Background()
+	ctx, log, done := tele.StartSpanWithLogger(ctx, "scope.ManagedControlPlaneScope.IsVnetManaged")
+	defer done()
+	isManaged, err := virtualnetworks.New(s).IsManaged(ctx)
+	if err != nil {
+		log.Error(err, "Unable to determine if ManagedControlPlaneScope VNET is managed by capz", "AzureManagedCluster", s.ClusterName())
+	}
+	s.cache.isVnetManaged = to.BoolPtr(isManaged)
+	return isManaged
 }
 
 // APIServerLBName returns the API Server LB spec.
--- a/azure/scope/cluster_test.go
+++ b/azure/scope/cluster_test.go
@@ -793,6 +793,7 @@ func TestRouteTableSpecs(t *testing.T) {
 						},
 					},
 				},
+				cache: &ClusterCache{},
 			},
 			want: nil,
 		},
@@ -828,6 +829,7 @@ func TestRouteTableSpecs(t *testing.T) {
 						},
 					},
 				},
+				cache: &ClusterCache{},
 			},
 			want: []azure.ResourceSpecGetter{
 				&routetables.RouteTableSpec{
@@ -875,6 +877,7 @@ func TestNatGatewaySpecs(t *testing.T) {
 						},
 					},
 				},
+				cache: &ClusterCache{},
 			},
 			want: nil,
 		},
@@ -922,6 +925,7 @@ func TestNatGatewaySpecs(t *testing.T) {
 						},
 					},
 				},
+				cache: &ClusterCache{},
 			},
 			want: []azure.ResourceSpecGetter{
 				&natgateways.NatGatewaySpec{
@@ -999,6 +1003,7 @@ func TestNatGatewaySpecs(t *testing.T) {
 						},
 					},
 				},
+				cache: &ClusterCache{},
 			},
 			want: []azure.ResourceSpecGetter{
 				&natgateways.NatGatewaySpec{
@@ -1075,6 +1080,7 @@ func TestNatGatewaySpecs(t *testing.T) {
 						},
 					},
 				},
+				cache: &ClusterCache{},
 			},
 			want: []azure.ResourceSpecGetter{
 				&natgateways.NatGatewaySpec{
@@ -1154,6 +1160,7 @@ func TestNSGSpecs(t *testing.T) {
 						},
 					},
 				},
+				cache: &ClusterCache{},
 			},
 			want: []azure.ResourceSpecGetter{
 				&securitygroups.NSGSpec{
@@ -1199,6 +1206,7 @@ func TestSubnetSpecs(t *testing.T) {
 						},
 					},
 				},
+				cache: &ClusterCache{},
 			},
 			want: []azure.ResourceSpecGetter{},
 		},
@@ -1260,6 +1268,7 @@ func TestSubnetSpecs(t *testing.T) {
 						},
 					},
 				},
+				cache: &ClusterCache{},
 			},
 			want: []azure.ResourceSpecGetter{
 				&subnets.SubnetSpec{
@@ -1362,6 +1371,7 @@ func TestSubnetSpecs(t *testing.T) {
 						},
 					},
 				},
+				cache: &ClusterCache{},
 			},
 			want: []azure.ResourceSpecGetter{
 				&subnets.SubnetSpec{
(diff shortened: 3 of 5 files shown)


Makefile
```diff
@@ -276,7 +276,7 @@ create-management-cluster: $(KUSTOMIZE) $(ENVSUBST) $(KUBECTL) $(KIND) ## Create
 	./hack/create-custom-cloud-provider-config.sh
 
 	# Deploy CAPI
-	curl --retry $(CURL_RETRIES) -sSL https://github.com/kubernetes-sigs/cluster-api/releases/download/v1.1.4/cluster-api-components.yaml | $(ENVSUBST) | $(KUBECTL) apply -f -
+	curl --retry $(CURL_RETRIES) -sSL https://github.com/kubernetes-sigs/cluster-api/releases/download/v1.2.0/cluster-api-components.yaml | $(ENVSUBST) | $(KUBECTL) apply -f -
```
CecileRobertMichon, line 279: this branch might need to be rebased, not sure why v1.2.0 changes are showing up here

azure/services/virtualnetworks/virtualnetworks.go
```diff
@@ -150,19 +167,34 @@ func (s *Service) IsManaged(ctx context.Context) (bool, error) {
 	ctx, _, done := tele.StartSpanWithLogger(ctx, "virtualnetworks.Service.IsManaged")
 	defer done()
 
+	// Report errors if the client is rate limited.
+	if !s.RateLimiter.Reader.TryAccept() {
```
CecileRobertMichon, line 171: This is good for mitigation when we do run into rate limiting. I agree we should try to do this generically across all services (we're already using the retry-after header to requeue create/delete async calls but we need to also use it across all GETs).

However, there is also an aspect of prevention to avoid running into rate limits as much as possible, by reducing the number of calls we make per reconcile (and reducing how often we reconciling which we already do by debouncing successful reconciles thanks to https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/pkg/coalescing/reconciler.go).

Low-hanging fruit for this PR would be to remove any extra calls to isVnetManaged since each one makes an API call. The ideal would be to cache the result of isVnetManaged in scope so we only make a single call per AzureCluster reconcile (ie. we don't need to check if it's managed once in the vnet service, AND for every subnet, we can get the result once and then reuse the value for all services, similar to how we cache VM skus in the machine scope.

Makefile
jackfrancis: yuck

azure/services/virtualnetworks/virtualnetworks.go
jackfrancis: is "cache" in scope as simple as assigning the API result to the scope object instance? (i.e., a new scope object is instantiated in every reconcile loop)

CecileRobertMichon: could be the API result or even just the boolean value of `isVnetManaged` since that's the only thing we need

CecileRobertMichon: see https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/scope/machine.go#L102 for an example

azure/scope/managedcontrolplane.go
```diff
@@ -328,7 +351,17 @@ func (s *ManagedControlPlaneScope) IsIPv6Enabled() bool {
...
+		return to.Bool(s.cache.isVnetManaged)
+	}
+	ctx := context.Background()
+	ctx, log, done := tele.StartSpanWithLogger(ctx, "scope.ManagedControlPlaneScope.IsVnetManaged")
+	defer done()
+	isManaged, err := virtualnetworks.New(s).IsManaged(ctx)
+	if err != nil {
+		log.Error(err, "Unable to determine if ManagedControlPlaneScope VNET is managed by capz", "AzureManagedCluster", s.ClusterName())
+	}
+	return isManaged
```
CecileRobertMichon, line 364: can we also set the cache here if it's unset?

azure/services/ratelimiter.go
```diff
@@ -0,0 +1,89 @@
...
+const (
+	AzureServiceRateLimit            bool    = true
+	AzureServiceRateLimitQPS         float32 = 1.0
+	AzureServiceRateLimitBucket      int     = 1
+	AzureServiceRateLimitQPSWrite    float32 = 1.0
+	AzureServiceRateLimitBucketWrite int     = 1
+)
+
+// NewRateLimiter creates new read and write flowcontrol.RateLimiter from RateLimitConfig.
+func NewRateLimiter(config *RateLimitConfig) (flowcontrol.RateLimiter, flowcontrol.RateLimiter) {
```
CecileRobertMichon, line 60: What do you think of moving this to a separate PR? it's a great improvement but I think for purposes of merging this quickly and getting it cherry-picked as a bug fix it would be beneficial to keep this PR as lean as possible. 

It would also be good to implement this across all services and doing a separate PR would also us to do that more easily.

azure/scope/managedcontrolplane.go
```diff
@@ -102,13 +105,33 @@ type ManagedControlPlaneScope struct {
...
 	ManagedMachinePools []ManagedMachinePool
 }
 
+// ManagedControlPlaneCache stores common ManagedControlPlane information so we don't have to hit the API multiple times within the same reconcile loop.
+type ManagedControlPlaneCache struct {
+	isVnetManaged *bool
+}
+
+// InitManagedControlPlaneCache sets cached information about the ManagedControlPlane to be used in the scope.
+func (s *ManagedControlPlaneScope) InitManagedControlPlaneCache(ctx context.Context) error {
```
CecileRobertMichon, line 122: I don't know if this init function is necessary given that we're fetching the info in IsVnetManaged() if cache is unset (seems redundant), we could probably get away with either:
a) having the init function set the cache at the beginning of the reconcile and then always return the cache value directly in IsVnetManaged() since cache would always be set, or
b) the first time IsVnetManaged() is called, it sets the value in the cache Just In Time and the rest of the time it gets the existing value from the cache

azure/scope/managedcontrolplane.go
CecileRobertMichon: (only if you go with approach b) from https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2543/files#r946138219)

azure/scope/managedcontrolplane.go
```diff
@@ -102,13 +105,33 @@ type ManagedControlPlaneScope struct {
...
+	cache          *ManagedControlPlaneCache
 
 	AzureClients
 	Cluster             *clusterv1.Cluster
 	ControlPlane        *infrav1exp.AzureManagedControlPlane
 	ManagedMachinePools []ManagedMachinePool
 }
 
+// ManagedControlPlaneCache stores common ManagedControlPlane information so we don't have to hit the API multiple times within the same reconcile loop.
+type ManagedControlPlaneCache struct {
```
CecileRobertMichon, line 117: can we do the same thing in cluster scope?

azure/services/ratelimiter.go
jackfrancis: +1

azure/scope/managedcontrolplane.go
jackfrancis: I don't have a strong opinion in the abstract. In my view I'd take the practical approach: are there a non-trivial number of reconciliation contexts that won't actually need these cached values? If so, then definitely we should not init first.

I do have an opinion that we should do caching consistently across service scopes. In which case maybe we punt this as well to a separate workstream (not just vnet)? In my manual tests the impact of these vnet GET calls does not materially decrement the allowed requests limits at a speed necessary to contribute to subscription throttling, so I think we can reasonably land this improvement without necessarily needing to couple the caching optimization.

CecileRobertMichon: > are there a non-trivial number of reconciliation contexts that won't actually need these cached values? If so, then definitely we should not init first.

the answer is no, every single reconcile loop will need at least call isVnetManaged at least once (both reconcile and delete) 

> I do have an opinion that we should do caching consistently across service scopes

agreed, do you have other examples of API GET calls that we need to do across multiple services that you're thinking of? We do have some redundant calls to isVnetManaged (especially one per subnet, which could grow as the number of subnets grows for large clusters with many node pools) so I do think caching is a worthy optimization this time around.

jackfrancis: I'm happy to scope the caching to this change (and this will be a good precedent going forward for net new changes that result in API calls).

The only reason I did the init is a basic copy/paste of the machine cache reference:

https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/scope/machine.go#L102

I think doing these on-demand is the more intuitive approach (simpler as well), so I'll just modify this to do that.

azure/services/ratelimiter.go
mtougeron: +1 to splitting the PR

azure/scope/managedcontrolplane.go
mtougeron: +1 to being cached instead of making frequent API calls during the reconcile loop.

azure/services/ratelimiter.go
jackfrancis: rate limiting will end up here: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2572

azure/scope/managedcontrolplane.go
jackfrancis: done

mtougeron:
Not sure if this is relevant, but should the commit message also be changed to be more relevant to what the PR is actually doing now? I'm not sure if that's being too picky in reviews...

jackfrancis:
> Not sure if this is relevant, but should the commit message also be changed to be more relevant to what the PR is actually doing now? I'm not sure if that's being too picky in reviews...

Yes, will definitely change the title!

CecileRobertMichon:
> Yes, will definitely change the title!

can you please also amend the commit? That's what will show up in commit history

Let's also add a release note for this change

azure/scope/managedcontrolplane.go
jackfrancis: done

azure/scope/managedcontrolplane.go
```diff
@@ -328,7 +341,18 @@ func (s *ManagedControlPlaneScope) IsIPv6Enabled() bool {
...
 // IsVnetManaged returns true if the vnet is managed.
 func (s *ManagedControlPlaneScope) IsVnetManaged() bool {
-	return true
+	if s.cache.isVnetManaged != nil {
+		return to.Bool(s.cache.isVnetManaged)
+	}
+	ctx := context.Background()
+	ctx, log, done := tele.StartSpanWithLogger(ctx, "scope.ManagedControlPlaneScope.IsVnetManaged")
+	defer done()
+	isManaged, err := virtualnetworks.New(s).IsManaged(ctx)
```
mtougeron, line 350: In `cluster.go` you called this variable `isVnetManaged`.

azure/scope/managedcontrolplane.go
```diff
@@ -82,6 +84,10 @@ func NewManagedControlPlaneScope(ctx context.Context, params ManagedControlPlane
 		}
 	}
 
+	if params.Cache == nil {
+		return nil, errors.New("failed to generate new scope from nil Cache")
```
mtougeron, line 88: nit: `failed to generate new scope due to nil Cache`

azure/scope/cluster.go
```diff
@@ -87,6 +88,10 @@ func NewClusterScope(ctx context.Context, params ClusterScopeParams) (*ClusterSc
 		}
 	}
 
+	if params.Cache == nil {
+		return nil, errors.New("failed to generate new scope from nil Cache")
```
mtougeron, line 92: nit: `failed to generate new scope due to nil Cache`

azure/services/backoff.go
```diff
@@ -0,0 +1,42 @@
+/*
```
mtougeron, line 1: I thought the retry handling was going to wait until a followup PR? Or is this just going to be the first place where implemented?

azure/services/virtualnetworks/virtualnetworks.go
```diff
@@ -150,19 +154,29 @@ func (s *Service) IsManaged(ctx context.Context) (bool, error) {
...
 
 	vnetIface, err := s.Get(ctx, spec)
-	if err != nil {
-		return false, err
-	}
 	vnet, ok := vnetIface.(network.VirtualNetwork)
 	if !ok {
 		return false, errors.Errorf("%T is not a network.VirtualNetwork", vnetIface)
 	}
+	if err != nil {
```
mtougeron, line 171: Doesn't this error check need to happen before it tries to cast the type into the vnet variable in case it didn't get a response due to rate limiting?

azure/scope/cluster.go
```diff
@@ -87,6 +88,10 @@ func NewClusterScope(ctx context.Context, params ClusterScopeParams) (*ClusterSc
 		}
 	}
 
+	if params.Cache == nil {
```
CecileRobertMichon, line 91: since Cache isn't strictly required, why fail if it's nil? we could simply initialize it to an empty cache if nil (that way no need to add the empty value everywhere in the controllers)

azure/services/backoff.go
CecileRobertMichon: +1 let's remove this file

azure/services/virtualnetworks/virtualnetworks.go
```diff
@@ -150,19 +154,29 @@ func (s *Service) IsManaged(ctx context.Context) (bool, error) {
 	ctx, _, done := tele.StartSpanWithLogger(ctx, "virtualnetworks.Service.IsManaged")
 	defer done()
 
+	// Report errors if the client is throttled.
+	if s.RetryAfter.Reader.After(time.Now()) {
```
CecileRobertMichon, line 158: should this also be removed?

CecileRobertMichon: Note: we have a better way to do this with controllers by scheduling a the controller to requeue on the retryAfter time, but let's do that in #2572 and keep this simple for now

azure/scope/cluster.go
jackfrancis: This will work, thanks!

azure/scope/managedcontrolplane.go
```diff
@@ -328,7 +341,18 @@ func (s *ManagedControlPlaneScope) IsIPv6Enabled() bool {
 
 // IsVnetManaged returns true if the vnet is managed.
 func (s *ManagedControlPlaneScope) IsVnetManaged() bool {
-	return true
+	if s.cache.isVnetManaged != nil {
+		return to.Bool(s.cache.isVnetManaged)
+	}
+	ctx := context.Background()
```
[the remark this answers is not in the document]
jackfrancis: It's really a question of trying to limit the refactor scope, but we can definitely include, I agree it would be better.

azure/scope/managedcontrolplane.go
jackfrancis: There's nothing mysterious here, I'm just preferring to use a lowercase local var to capture the response from the method name that we're calling.

azure/scope/managedcontrolplane.go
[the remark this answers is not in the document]
CecileRobertMichon: If it's going to be too big of a refactor, I'll settle with a TODO and a github issue to pass in a real context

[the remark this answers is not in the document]
jackfrancis: Yeah, it spans multiple interfaces and is going to be a significant amount of boilerplate change:

image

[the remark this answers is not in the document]
jackfrancis: https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/2581

azure/services/virtualnetworks/virtualnetworks.go
jackfrancis: no longer an open question as 429 foo is gone!

azure/services/backoff.go
jackfrancis: done

azure/scope/cluster.go
jackfrancis: no longer returning error for nil Cache

azure/scope/managedcontrolplane.go
jackfrancis: no longer returning error for nil Cache

azure/scope/cluster.go
```diff
@@ -524,7 +536,12 @@ func (s *ClusterScope) Vnet() *infrav1.VnetSpec {
 
 // IsVnetManaged returns true if the vnet is managed.
 func (s *ClusterScope) IsVnetManaged() bool {
-	return s.Vnet().ID == "" || s.Vnet().Tags.HasOwned(s.ClusterName())
+	if s.cache.isVnetManaged != nil {
+		return to.Bool(s.cache.isVnetManaged)
+	}
+	isVnetManaged := s.Vnet().ID == "" || s.Vnet().Tags.HasOwned(s.ClusterName())
```
CecileRobertMichon, line 542: I wonder if we should change this behavior to be the same as managed clusters eventually, although I agree this is the least dangerous change for now since it preserves existing behavior

jackfrancis: I was honestly a bit confused when implementing this for clusters, so glad to hear that suggestion.

I can throw up really quick PR after this one lands so that the scope of this PR leans a bit more towards managed clusters only.

CecileRobertMichon (commented):
/lgtm

CecileRobertMichon:
> Additionally, in order to account for additional calls against the Azure APIs I've wired in a rate limiting + HTTP 429 retry solution (inspired by cloud-provider-azure).

Remove this from PR description?

jackfrancis:
> > Additionally, in order to account for additional calls against the Azure APIs I've wired in a rate limiting + HTTP 429 retry solution (inspired by cloud-provider-azure).
> 
> Remove this from PR description?

Done

jackfrancis:
@mtougeron @CecileRobertMichon this is ready for another (hopefully final? 🤞) review round

CecileRobertMichon (commented):
/lgtm

mtougeron (approved):
looks good to me.

/approve

PR #2546, opened 2022-08-10
clean up hack/tools go.mod

jackfrancis:
**What type of PR is this?**

/kind cleanup

**What this PR does / why we need it**:

Unfortunately I approved https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2487 without noticing that the `go.mod` in `hack/tool` was not sanitized via `go mod tidy`. This PR addresses that.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

PR #2547, opened 2022-08-10
Update and verify go modules in hack/tools

mboersma:
**What type of PR is this?**

/kind failing-test

Adds the go.mod and go.sum files under hack/tools to the `make modules` and `make verify-modules` Makefile targets.

Refs #2546

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```


mboersma:
The above failure is correct and desirable, since it is now checking hack/tools/go.mod and #2)546 hasn't yet merged. (Perhaps this should be combined into a single PR with that one?)

PR #2560, opened 2022-08-12
Fix hack/tools imports so "make modules" works

mboersma:
**What type of PR is this?**

/kind bug

Fixes hack/tools imports so `make clean verify-modules verify-conversions` works.

Currently in CAPZ's `main` branch, doing `make modules` fails:

```shell
% make modules
go mod tidy
pushd hack/tools && go mod tidy; popd
~/projects/cluster-api-provider-azure/hack/tools ~/projects/cluster-api-provider-azure
go: downloading sigs.k8s.io/cluster-api v0.0.0-00010101000000-000000000000
sigs.k8s.io/cluster-api-provider-azure/hack/tools imports
	sigs.k8s.io/cluster-api/hack/tools: sigs.k8s.io/cluster-api@v0.0.0-00010101000000-000000000000: invalid version: unknown revision 000000000000
~/projects/cluster-api-provider-azure
```

I pinned an import for go-multierror and reverted the version of `conversion-verifier` to the last-known-working version. This seems to satisfy all the requirements, although we should investigate why we can't use CAPI's hack/tools code from `main` currently.

Fixes #

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```


hack/tools/go.mod
```diff
@@ -2,24 +2,25 @@ module sigs.k8s.io/cluster-api-provider-azure/hack/tools
 
 go 1.18
 
-require sigs.k8s.io/cluster-api/hack/tools v0.0.0-20220810165624-90923e703fac
+require (
+	github.com/hashicorp/go-multierror v1.1.1
```
CecileRobertMichon, line 6: Just curious, what's causing us to need to pin this one? Is it temporary and, if so, should we add a comment about conditions for reverting it?

mboersma: It's a package needed to build CAPI's `conversion-verifier`, but `go mod tidy` can't see that and removes it. I can't claim to understand the root importing problem 100%, but I know how it behaves.

We could perhaps revert it by importing a newer version of CAPI's hack/tools, but that appears to have its own set of problems. I did that in https://github.com/kubernetes-sigs/cluster-api-provider-azure/commit/d9863531d7bc9373cbdf5ddc879cf21435ecf6b2 but it just ended up creating a problem when checked in to `main`.

CecileRobertMichon (approved):
let's merge this to fix make modules and then figure out why the latest import is broken

@mboersma I wonder if this is related to the apidiff failure we're observing in release-1.3

/lgtm
/approve

PR #2564, opened 2022-08-15
Add unit tests for VM converter and remove unused err from SDKToVM

### LINKED ISSUE #2538: Add unit tests for VM converter
/kind feature

**Describe the solution you'd like:** Add unit tests for [vm.go](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/converters/vm.go) in the converters package similar to the other converters.

**Environment:**

willie-yao:
**What type of PR is this?**
/kind feature

This PR adds unit tests for the VM converter and removes unused `err` return from SDKToVM

Fixes #2538 

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/azure/converters/vm.go
+++ b/azure/converters/vm.go
@@ -44,7 +44,7 @@ type VM struct {
 }
 
 // SDKToVM converts an Azure SDK VirtualMachine to the CAPZ VM type.
-func SDKToVM(v compute.VirtualMachine) (*VM, error) {
+func SDKToVM(v compute.VirtualMachine) *VM {
 	vm := &VM{
 		ID:    to.String(v.ID),
 		Name:  to.String(v.Name),
@@ -63,5 +63,5 @@ func SDKToVM(v compute.VirtualMachine) (*VM, error) {
 		vm.Tags = MapToTags(v.Tags)
 	}
 
-	return vm, nil
+	return vm
 }
--- a/azure/converters/vm_test.go
+++ b/azure/converters/vm_test.go
@@ -1,0 +1,137 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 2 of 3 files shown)


Jont828:
By the way since the VM converter never actually returns an error I think could probably just remove it from the function spec. Wdyt about that?

willie-yao:
I agree with that. Should we do that in a separate PR?

mboersma:
> the VM converter never actually returns an error...just remove it from the spec
>> Should we do that in a separate PR?

My test for that is "is there a possibility we might need to revert these changes separately?" That illuminates whether the changes are logically distinct and ought to have separate commits and PRs.

In this case, the changes are distinct, but I think there's little chance we would need to revert them separately. So IMHO it's a judgment call. It's always defensible to make separate PRs, but I think you could make both changes in one PR in this case and reviewers wouldn't complain...at least I wouldn't. 😄 

This lgtm, btw.

willie-yao:
@mboersma @Jont828 Just pushed changes removing the unused `err` from `SDKToVM`. Let me know if this looks good, and I'll squash it.

mboersma:
Hmm...so changing the signature of `SDKToVM` is actually a public-facing API change, so apidiff is correct. I think it's entirely likely that no one is consuming that API beyond the single reference that you updated in this PR, but we can't be sure. 

Maybe I was wrong about combining the two commits? I'm curious what other reviewers think. ping @Jont828 @CecileRobertMichon @jackfrancis

willie-yao:
@mboersma That's correct. One interesting thing is that my PR to change the apidiff test would have made this test pass, since we are changing that test to only flag breaking changes within the api/ or exp/api/ directories. I'm assuming this means that this change would be okay to merge?

mboersma:
You are correct, I totally forgot that we really only care about those dirs for apidiff. (#2567 can't merge soon enough.)

mboersma:
@willie-yao if you can rebase this to fix the apidiff error, I think it's good to merge.

mboersma (commented):
/lgtm

CecileRobertMichon (commented):
/lgtm
/approve

great work 🎉

PR #2567, opened 2022-08-17
Apidiff test runs only if the changes are in api/ and exp/api/

### LINKED ISSUE #2051: Change apidiff test to only fail for actual breaking API changes

willie-yao:
**What type of PR is this?**
/kind failing-test

The apidiff test fails as soon as there is a breaking change in an exposed field or exposed function. This PR changes the test to only look at changes in the api/ and exp/api/ directories. Used #2206 as a reference.

Fixes #2051 

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -213,9 +213,11 @@ clean-temporary: ## Remove all temporary files and folders.
 clean-release: ## Remove the release folder.
 	rm -rf $(RELEASE_DIR)
 
+APIDIFF_OLD_COMMIT ?= $(shell git rev-parse origin/main)
+
 .PHONY: apidiff
 apidiff: $(GO_APIDIFF) ## Check for API differences.
-	$(GO_APIDIFF) $(shell git rev-parse origin/main) --print-compatible
+	$(GO_APIDIFF) $(APIDIFF_OLD_COMMIT)
 
 .PHONY: format-tiltfile
 format-tiltfile: ## Format the Tiltfile.
--- a/scripts/ci-apidiff.sh
+++ b/scripts/ci-apidiff.sh
@@ -14,15 +14,19 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-set -o errexit
 set -o nounset
 set -o pipefail
 
 REPO_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
 
-APIDIFF="${REPO_ROOT}/hack/tools/bin/go-apidiff"
+cd "${REPO_ROOT}" || exit
 
-make --directory="${REPO_ROOT}" "${APIDIFF##*/}"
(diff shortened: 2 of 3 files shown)


willie-yao:
/retitle Apidiff test runs only if the changes are in api/ and exp/api/

Jont828:
Nice work!

/lgtm

jackfrancis:
The shell script seems a bit magical to me, could you make a change under api/ and/or /exp/api so we can see the test fail as expected?

willie-yao:
> The shell script seems a bit magical to me, could you make a change under api/ and/or /exp/api so we can see the test fail as expected?

Yes, I have tested each scenario by making a breaking change inside those directories, as well as expecting a pass when making a change elsewhere. @jackfrancis 

You can see the expected behavior in the test history:

- [Expected failure due to change in api folder](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/2567/pull-cluster-api-provider-azure-apidiff/1560361165459034112)
- [Expected pass when changing something outside of the api or exp/api folder](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/2567/pull-cluster-api-provider-azure-apidiff/1560362571041607680)

mboersma:
/lgtm

I tested this locally and it seems to work as advertised.

Thanks @willie-yao!

willie-yao:
@mboersma Thank you! I'm having a bit of trouble trying to squash a merge commit, so I apologize for all the force pushes. Is it required to squash a merge commit?

mboersma:
>  Is it required to squash a merge commit?

Ideally, PRs here are rebased and contain a single commit. But we can also tell the prow bot to squash when it decides to merge if need be. So as long as there aren't extraneous commits in the PR, it's fine.

willie-yao:
Got it. I figured out a way to squash the merge commit and the PR will need a re-lgtm. The changes are the same @mboersma. Apologies for the inconvenience!

mboersma:
/lgtm

No worries about the `git` housekeeping, thanks again!

PR #2573, opened 2022-08-18
fix error getting activity logs in e2e for managed clusters

### LINKED ISSUE #2530: E2E test artifacts missing activity logs for AKS clusters
The e2e activity log collection assumes `AzureCluster` to find the resource group name: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/test/e2e/azure_clusterproxy.go#L220

This causes activity logs to fail to be collected when the cluster is a managed cluster (`AzureManagedCluster`).

Proposed fix: update the `collectActivityLogs` to fall back to finding AzureManagedClusters if no AzureCluster is found so we can get the resource group name for AKS clusters and collect logs.

nojnhuh:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**: Currently, when the e2e framework grabs the activity logs based on the resource group name, it derives that from an `AzureCluster` resource. For managed clusters, the resource group is instead defined on an `AzureManagedControlPlane` resource. These changes fallback to getting the resource group name from an `AzureManagedControlPlane` when the expected `AzureCluster` does not exist.

Fixes #2530

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/azure_logcollector.go
+++ b/test/e2e/azure_logcollector.go
@@ -183,8 +183,19 @@ func getAzureCluster(ctx context.Context, managementClusterClient client.Client,
 	}
 
 	azCluster := &v1beta1.AzureCluster{}
-	err := managementClusterClient.Get(context.TODO(), key, azCluster)
+	err := managementClusterClient.Get(ctx, key, azCluster)
 	return azCluster, err
+}
+
+func getAzureManagedControlPlane(ctx context.Context, managementClusterClient client.Client, namespace, name string) (*expv1alpha4.AzureManagedControlPlane, error) {
+	key := client.ObjectKey{
+		Namespace: namespace,
+		Name:      name,
+	}
+
+	azManagedControlPlane := &expv1alpha4.AzureManagedControlPlane{}
+	err := managementClusterClient.Get(ctx, key, azManagedControlPlane)
+	return azManagedControlPlane, err
 }
 
 func getAzureMachine(ctx context.Context, managementClusterClient client.Client, m *clusterv1.Machine) (*v1beta1.AzureMachine, error) {
--- a/test/e2e/azure_clusterproxy.go
+++ b/test/e2e/azure_clusterproxy.go
@@ -38,6 +38,7 @@ import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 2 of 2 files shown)


test/e2e/azure_logcollector.go
```diff
@@ -183,10 +183,21 @@ func getAzureCluster(ctx context.Context, managementClusterClient client.Client,
 	}
 
 	azCluster := &v1beta1.AzureCluster{}
-	err := managementClusterClient.Get(context.TODO(), key, azCluster)
+	err := managementClusterClient.Get(ctx, key, azCluster)
```
nojnhuh, line 186: This isn't strictly related to the rest of these changes, but seems like a mistake given the `ctx` parameter is otherwise not used and the other similar functions here follow this pattern.

nojnhuh:
I wasn't seeing anything in the activity logs when I was running the AKS test locally, but I'm seeing plenty of messages there in the CI run, so there must be some quirk with my sub or service principal that I can look into if that ever becomes a problem for me.

test/e2e/azure_logcollector.go
```diff
@@ -183,10 +183,21 @@ func getAzureCluster(ctx context.Context, managementClusterClient client.Client,
...
 	return azCluster, err
 }
 
+func getAzureManagedControlPlane(ctx context.Context, managementClusterClient client.Client, namespace, name string) (*expv1alpha4.AzureManagedControlPlane, error) {
+	key := client.ObjectKey{
+		Namespace: namespace,
+		Name:      name,
+	}
+
+	azManagedControlPlane := &expv1alpha4.AzureManagedControlPlane{}
```
CecileRobertMichon, line 196: There is an issue with the import name, this shouldn't be `expv1alpha4`, it's using the v1beta1 API version

nojnhuh: #2574 will address the import alias, but either this PR or that one will need an update here when the other is merged. Should I mark this PR as blocked on that one to make sure we don't forget?

mboersma: Either way is fine, I can easily rebase #2574 if this merges first.

CecileRobertMichon:
you'll need to rebase on top of #2575 to make codespell happy

CecileRobertMichon:
/lgtm

thanks @nojnhuh!

mboersma (approved):
/approve

PR #2575, opened 2022-08-18
Fix spelling errors pointed out by Codespell

mboersma:
**What type of PR is this?**:

/kind cleanup

Fixes spelling in two places to placate the updated Codespell GH action.

Although that job is not required, ignoring the red on every job isn't good either.

N/A

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/observability/prometheus/resources/bundle.yaml
+++ b/hack/observability/prometheus/resources/bundle.yaml
@@ -4486,7 +4486,7 @@ spec:
                               If underlying persistent volume is being resized then
                               the Condition will be set to 'ResizeStarted'.
                             items:
-                              description: PersistentVolumeClaimCondition contails
+                              description: PersistentVolumeClaimCondition contains
                                 details about state of pvc
                               properties:
                                 lastProbeTime:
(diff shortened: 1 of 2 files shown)


hack/observability/prometheus/resources/bundle.yaml
```diff
@@ -4486,7 +4486,7 @@ spec:
                               If underlying persistent volume is being resized then
                               the Condition will be set to 'ResizeStarted'.
                             items:
-                              description: PersistentVolumeClaimCondition contails
+                              description: PersistentVolumeClaimCondition contains
```
mboersma, line 4489: This change is to a YAML resource downloaded by a script, so it could be overwritten. But:
- the script hasn't been run in a long time; we're ok with prometheus staying right here
- if the script is run, the subsequent code change will presumably fail as a PR and not get merged (the current version of this prometheus bundle retains the same spelling error, I checked)

CecileRobertMichon (commented):
/lgtm
/approve
/retest

PR #2588, opened 2022-08-22
feat: respect externally managed annotation on unmanaged MachinePools

mweibel:
**What type of PR is this?**
/kind feature

This adds support for marking `MachinePool` resources as externally managed. Similar to #2444 but for unmanaged clusters. It does not rely on a property in `AzureMachinePool` since, unlike in AzureManagedMachinePool`, this property does not exist.

Please note that this is just a Draft PR in order to get some feedback on the approach. I reused code and methodology from #2444 in the hope that it matches the existing code style. 
I did not yet add tests since I just added the code and I'm trying it on a local cluster. 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
adds the ability to annotate a MachinePool with `cluster.x-k8s.io/replicas-managed-by-autoscaler` to synchronize VMSS capacity with MachinePool replicas automatically.
```


CecileRobertMichon:
https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2444 got merged so you should be able to reuse the code from it now if you rebase

See also https://github.com/kubernetes-sigs/cluster-api/pull/7107 which is relevant / in progress

mweibel:
> See also https://github.com/kubernetes-sigs/cluster-api/pull/7107 which is relevant / in progress

yes I saw that, thanks. Once that PR is merged we can adjust CAPZ code to use the const from CAPI I guess.

mweibel:
I tested using a separate empty commit and the tests ran through. I squashed them together and now there's again a conflict in the e2e test. 

What can I do about this?

CecileRobertMichon:
/retest

the failure above looks like a flake

azure/scope/machinepool.go
```diff
@@ -688,3 +701,39 @@ func (m *MachinePoolScope) UpdatePatchStatus(condition clusterv1.ConditionType,
 		conditions.MarkFalse(m.AzureMachinePool, condition, infrav1.FailedReason, clusterv1.ConditionSeverityError, "%s failed to update. err: %s", service, err.Error())
 	}
 }
+
+// PatchCAPIMachinePoolObject persists the capi machinepool configuration and status.
+func (m *MachinePoolScope) PatchCAPIMachinePoolObject(ctx context.Context) error {
```
CecileRobertMichon, line 706: just like in #2444, we need to make sure that the controller has appropriate rbac permissions to patch MachinePools

mweibel: is this something we need to do additionally? The changes to `azuremachinepool_controller.go` and `rbac.yaml`  in #2444 should be enough to work also in this case, if I'm not mistaken.

jackfrancis:
@mweibel this is an interesting use-case. So the idea is that you might want to use cluster-autoscaler on your cluster, but not leverage the Cluster API provider, and instead use the Azure provider (which updates replicas directly via the Azure VMSS API)?

Or are there other use cases you're thinking of?

I'm wondering if there's a better, general workflow to indicate this rather than requiring the user to manually add a well-known annotation to the corresponding capi MachinePool resource?

mweibel:
> @mweibel this is an interesting use-case. So the idea is that you might want to use cluster-autoscaler on your cluster, but not leverage the Cluster API provider, and instead use the Azure provider (which updates replicas directly via the Azure VMSS API)?

yes exactly. I contributed the same to CAPA already. The use case we have currently is that the cluster-api autoscaler does not yet have enough functionality to replace the native azure one. Crucial features like taints/labels and separate cool down times etc. are not available yet and will probably not for a little while (especially the taints since that's onlyin a proposal how to manage it in CAPI itself). 

> I'm wondering if there's a better, general workflow to indicate this rather than requiring the user to manually add a well-known annotation to the corresponding capi MachinePool resource?

I settled for an annotation since there was discussion on CAPA about this (annotations vs. a separate field) and they preferred an annotation. It also kind of fits together with the changes in kubernetes-sigs/cluster-api#7107.

jackfrancis:
@mweibel this lgtm, I think we need to rebase this PR on top of latest `main` to overcome the E2E flake

Thanks again for this, let's try to land this before mid-next week so it can be included in the 1.7. release!

azure/scope/machinepool.go
```diff
@@ -704,3 +717,39 @@ func (m *MachinePoolScope) UpdatePatchStatus(condition clusterv1.ConditionType,
...
+}
+
+// UpdateCAPIMachinePoolReplicas updates the associated MachinePool replica count.
+func (m *MachinePoolScope) UpdateCAPIMachinePoolReplicas(ctx context.Context, replicas *int32) {
+	m.MachinePool.Spec.Replicas = replicas
+}
+
+// HasReplicasExternallyManaged returns true if the externally managed annotation is set on the CAPI MachinePool resource.
+func (m *MachinePoolScope) HasReplicasExternallyManaged(ctx context.Context) bool {
+	return m.MachinePool.Annotations[azure.ReplicasManagedByAutoscalerAnnotation] == "true"
```
CecileRobertMichon, line 736: @jackfrancis do we need a PR to get rid of this CAPZ specific annotation in favor of the CAPI one that you added?

CecileRobertMichon: we should even be able to get rid of this function altogether and directly use the `ReplicasManagedByExternalAutoscaler()` helper from CAPI in `ReconcileReplicas` below

CecileRobertMichon:
> I think we need to rebase this PR on top of latest main to overcome the E2E flake

Prow does the rebase when running tests

/retest

CecileRobertMichon:
as discussed in https://kubernetes.slack.com/archives/CEX9HENG7/p1673034488923059, let's merge this and refactor to use the CAPI helper function in a separate PR

/lgtm
/approve

PR #2593, opened 2022-08-23
make AKS node taints mutable

zmalik:
**What type of PR is this?**
/kind feature

Allow AKS nodepool taints to be mutable. AKS allows the taints to be modified https://github.com/Azure/AKS/issues/1515 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
allow AKS nodepool taints to be updated.
```

### DIFF
--- a/azure/scope/managedmachinepool_test.go
+++ b/azure/scope/managedmachinepool_test.go
@@ -406,7 +406,7 @@ func TestManagedMachinePoolScope_Taints(t *testing.T) {
 				Mode:         "User",
 				Cluster:      "cluster1",
 				Replicas:     1,
-				NodeTaints:   []string{"key1=value1:NoSchedule"},
+				NodeTaints:   &[]string{"key1=value1:NoSchedule"},
 				VnetSubnetID: "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups//providers/Microsoft.Network/virtualNetworks//subnets/",
 				Headers:      map[string]string{},
 			},
--- a/azure/services/agentpools/agentpools_test.go
+++ b/azure/services/agentpools/agentpools_test.go
@@ -103,6 +103,43 @@ func TestReconcileAgentPools(t *testing.T) {
 				s.UpdatePutStatus(infrav1.AgentPoolsReadyCondition, serviceName, internalError)
 			},
 		},
+		{
+			name: "update needed on nodepool taints change",
+			agentPoolsSpec: azure.AgentPoolSpec{
+				Name:              "my-agent-pool",
+				ResourceGroup:     "my-rg",
+				Cluster:           "my-cluster",
+				SKU:               "Standard_D2s_v3",
+				Version:           to.StringPtr("9.99.9999"),
+				EnableAutoScaling: to.BoolPtr(true),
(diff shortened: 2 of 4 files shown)


jackfrancis:
I ran a manual test using this branch and I was able to update a taints definition on an AzureManagedMachinePool (added a 2nd taint). This is what happened:

1) the existing nodes were not changed
2) when I scaled out the corresponding MachinePool, the *new* nodes did reflect the taint changes (they had 2 taints)

I'm not sure this is what we want. We probably want a change to the taints on the AzureManagedMachinePool to trigger a rolling upgrade across all existing nodes to ensure that the configuration change is universally applied across its nodes.

@CecileRobertMichon thoughts?

zmalik:
good catch @jackfrancis !

yeah, that's the case I will fix it. Was too deep into testing the taints merge use-case https://github.com/zmalik/cluster-api-provider-azure/commit/e8e2352b5d06eab920c8865c614074adf36a2cd6 

I will update it now.

zmalik:
there are some slight changes required as `[]string` vs `*[]string` always ended up as a drift detection. We do not want to trigger a call to Azure API because of empty taints.

jackfrancis:
@zmalik let me know when this is ready for another round of smoke testing, happy to do so.

Thanks again!

zmalik:
@jackfrancis thank you! 
Can you validate that this version of the changes isn't enough? 

https://kubernetes.slack.com/archives/CEX9HENG7/p1661346766050419?thread_ts=1661262907.252129&cid=CEX9HENG7

This version tries to unset all of the AKS systems taints. Not sure if we can test it quickly as I used scalesetpriority settings which add an AKS system taints to the nodepools but that feature isn't present in main branch yet. To test that any AKS system taints would be enough.

jackfrancis:
@zmalik I tested add and delete, they worked great, AKS does the "rolling update" part of the equation by updating the nodes in-place and relying on this behavior:

https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/#taint-based-evictions

One thing I did observe is that we don't seem to have a way to "delete all taints" from a pool. If you remove the taints array entirely from the `AzureManagedMachinePool` resource, the `omitempty` JSON marshalling declaration means that AKS doesn't actually see any changes to the PUT body data and it doesn't update its taints configuration accordingly. It's not actually obvious to me _how_ one might entirely delete all taints from a nodepool given the AKS API, so I asked that question here:

https://github.com/Azure/AKS/issues/1515#issuecomment-1226203666

Not sure if we should block forward progress of this PR until we can answer the above question.

cc @CecileRobertMichon @nojnhuh

CecileRobertMichon:
@alexeldeib any ideas on the question above "It's not actually obvious to me how one might entirely delete all taints from a nodepool" ^?

jackfrancis:
@zmalik I would advocate that we move forward with this change. Are you able to rebase on top of main?

zmalik:
I will come back to this early PR next week. Sorry for the delay

zmalik:
this PR https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2682 will takeover this one
thanks @jackfrancis

PR #2599, opened 2022-08-24
Dump resources in private cluster test

### LINKED ISSUE #2584: Refactor dumpSpecResourcesAndCleanup for private cluster spec
/kind feature

Currently, the private cluster e2e spec has its own [cleanup function](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/test/e2e/azure_privatecluster.go#L158) instead of using [`dumpSpecResourcesAndCleanup`](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/test/e2e/common.go#L136), which all other e2e test specs use. This is because [`ExpectResourceGroupToBe404`](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/test/e2e/common.go#L191) expects all clusters to own the resource group, which is not the case for the private cluster. We can pass in an additional flag to the `cleanupInput` to skip `ExpectResourceGroupToBe404` for the private cluster spec.

This will also solve a problem where the private cluster spec resources were not being dumped. This will greatly improve our logging experience.

willie-yao:
**What type of PR is this?**
/kind bug

The private cluster spec doesn't dump resources before cleaning up as expected. We can solve this by having the test use the same cleanup function as the rest of the test specs.

Fixes #2584 

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/azure_privatecluster.go
+++ b/test/e2e/azure_privatecluster.go
@@ -38,7 +38,6 @@ import (
 	. "github.com/onsi/gomega"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/utils/pointer"
 	"sigs.k8s.io/cluster-api-provider-azure/azure"
@@ -59,6 +58,7 @@ type AzurePrivateClusterSpecInput struct {
 	E2EConfig             *clusterctl.E2EConfig
 	ArtifactFolder        string
 	SkipCleanup           bool
+	CancelWatches         context.CancelFunc
 }
 
 // AzurePrivateClusterSpec implements a test that creates a workload cluster with a private API endpoint.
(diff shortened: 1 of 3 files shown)


test/e2e/azure_privatecluster.go
```diff
@@ -160,25 +160,20 @@ func AzurePrivateClusterSpec(ctx context.Context, inputGetter func() AzurePrivat
 		// cluster is deleted at the end of the test. If we don't delete this cluster, the Azure resource delete
 		// verification will fail.
 		if !input.SkipCleanup {
```
CecileRobertMichon, line 162: since we're passing in the skip cleanup value to the cleanupInput now, we can remove this if statement so we always collect resources/logs regardless of skipCleanup value but only delete the cluster if skipCleanup is true

test/e2e/azure_privatecluster.go
```diff
@@ -160,25 +160,20 @@ func AzurePrivateClusterSpec(ctx context.Context, inputGetter func() AzurePrivat
...
+				ClusterProxy:           publicClusterProxy,
+				Namespace:              input.Namespace,
+				CancelWatches:          input.CancelWatches,
+				IntervalsGetter:        e2eConfig.GetIntervals,
+				SkipCleanup:            input.SkipCleanup,
+				ArtifactFolder:         input.ArtifactFolder,
+				SkipResourceGroupCheck: true,
+			}
+			dumpSpecResourcesAndCleanup(ctx, cleanInput)
 			Logf("deleting private cluster %q in namespace %q", cluster.Name, cluster.Namespace)
```
CecileRobertMichon, line 176: should this log line be removed? it's printed after the actual delete happened

test/e2e/azure_privatecluster.go
```diff
@@ -160,25 +160,20 @@ func AzurePrivateClusterSpec(ctx context.Context, inputGetter func() AzurePrivat
...
+			cleanInput := cleanupInput{
+				SpecName:               specName,
+				Cluster:                cluster,
+				ClusterProxy:           publicClusterProxy,
+				Namespace:              input.Namespace,
+				CancelWatches:          input.CancelWatches,
+				IntervalsGetter:        e2eConfig.GetIntervals,
+				SkipCleanup:            input.SkipCleanup,
+				ArtifactFolder:         input.ArtifactFolder,
+				SkipResourceGroupCheck: true,
```
CecileRobertMichon, line 173: let's add a comment here to explain why we're skipping the resource group check

mboersma (commented):
/lgtm

CecileRobertMichon (commented):
/lgtm
/approve

PR #2615, opened 2022-08-31
set AzureManagedMachinePool not ready for in-progress reconciliations

### LINKED ISSUE #2561: AzureManagedCluster: machinepools have negative replica count
/kind bug

A machinepool that was recently scaled out from 1 to 100:

```
  observedGeneration: 4
  phase: Running
  readyReplicas: 100
  replicas: 100
  unavailableReplicas: -99
```

I expect the `unavailableReplicas` count to converge downwards toward zero when scale out operations are in progress, and to always be equal to `0` when `phase: Running`. I don't expect it to ever be a negative integer.

**Environment:**

nojnhuh:
**What type of PR is this?**
/kind bug

The `status.ready` field for AzureManagedMachinePools is set to `true` in the first successful reconciliation:
https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/03c509372370de7e9fe810f7d12a0469f4a34798/exp/controllers/azuremanagedmachinepool_reconciler.go#L146
https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/03c509372370de7e9fe810f7d12a0469f4a34798/exp/controllers/azuremanagedmachinepool_controller.go#L282

After, it is never set to `false`. This causes an issue during scaling operations where Cluster API's MachinePool controller does not short-circuit [here](https://github.com/kubernetes-sigs/cluster-api/blob/8b4214d72762394144b83dd6d14986ff7e274095/exp/internal/controllers/machinepool_controller_phases.go#L270-L273) and proceeds to construct replica counts for the MachinePool's status [here](https://github.com/kubernetes-sigs/cluster-api/blob/8b4214d72762394144b83dd6d14986ff7e274095/exp/internal/controllers/machinepool_controller_noderef.go#L96-L99) which may not be correct while the infrastructure is still being reconciled. This manifests in #2561 where scaling an AzureManagedMachinePool up by some n number of replicas would sometimes result in the MachinePool's `status.unavailableReplicas` converging to -n and other times to 0 as expected.

For reference, the AzureMachinePool controller sets `status.ready` to false based on the status of the backing VMSS and desired/actual replicas in [this function](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/03c509372370de7e9fe810f7d12a0469f4a34798/azure/scope/machinepool.go#L390).

Fixes #2561

**Special notes for your reviewer**: I'm not exactly sure how best to express this fix, so feedback on whether there's a better way or place to do this is appreciated.

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/exp/controllers/azuremanagedmachinepool_controller.go
+++ b/exp/controllers/azuremanagedmachinepool_controller.go
@@ -258,6 +258,8 @@ func (ammpr *AzureManagedMachinePoolReconciler) reconcileNormal(ctx context.Cont
 	}
 
 	if err := svc.Reconcile(ctx); err != nil {
+		scope.SetAgentPoolReady(false)
+
 		// Handle transient and terminal errors
 		log := log.WithValues("name", scope.InfraMachinePool.Name, "namespace", scope.InfraMachinePool.Namespace)
 		var reconcileError azure.ReconcileError
@@ -279,7 +281,7 @@ func (ammpr *AzureManagedMachinePoolReconciler) reconcileNormal(ctx context.Cont
 	}
 
 	// No errors, so mark us ready so the Cluster API Cluster Controller can pull it
-	scope.InfraMachinePool.Status.Ready = true
+	scope.SetAgentPoolReady(true)
 	ammpr.Recorder.Eventf(scope.InfraMachinePool, corev1.EventTypeNormal, "AzureManagedMachinePool available", "agent pool successfully reconciled")
 	return reconcile.Result{}, nil
 }


azure/services/agentpools/agentpools.go
```diff
@@ -77,6 +77,7 @@ func (s *Service) Reconcile(ctx context.Context) error {
 		result, err := s.CreateResource(ctx, agentPoolSpec, serviceName)
 		if err != nil {
 			resultingErr = err
+			s.scope.SetAgentPoolReady(false)
```
CecileRobertMichon, line 80: Have you considered doing this at the end of the reconcile loop in the same place where we set this to true when there is no reconcile error? https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/exp/controllers/azuremanagedmachinepool_controller.go#L282 (perhaps that should also use the `s.scope.SetAgentPoolReady(true)` since that's equivalent)

The function name `SetAgentPoolReady` is maybe misleading, since it actually sets the Ready status field on the AzureManagedMachinePool CR, not the Azure agent pool resource.

nojnhuh: I thought about that spot, but since this particular scenario goes through the codepath ending at https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/24b93aacd3caf08d25d0bf1368e9eba4bb0e2d84/exp/controllers/azuremanagedmachinepool_controller.go#L272, it didn't seem like there was a clear way to catch where we should set ready to false.

And I'm still not sure in exactly which scenarios we'd want to set ready to false. Setting it based on whether or not this `err` is nil looks like it would include more scenarios where `status.ready` would be set to false than where I'm setting it now.
https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/24b93aacd3caf08d25d0bf1368e9eba4bb0e2d84/exp/controllers/azuremanagedmachinepool_controller.go#L260

Is that closer to what we'd want anyway? Or is there some additional introspection of that `err` we'd want to do to decide what the status should be?

CecileRobertMichon: > Is that closer to what we'd want anyway? Or is there some additional introspection of that err we'd want to do to decide what the status should be?

A question that might help answer this: are there any cases where you'd want the Status to still show `Ready` even though the Resource is not fully reconciled or there was an error returned from Reconcile?

nojnhuh: My gut reaction is to say "no," though I'm not familiar with the range of possible errors that might return. I noticed that for unmanaged AzureMachinePools, their `status.ready` is based solely on the provisioning state of the VMSS, desired replica count, and actual replica count. I can see where setting `status.ready` to false more aggressively may not necessarily be a bad thing though.

CecileRobertMichon: Here is how the [Cluster API contract](https://cluster-api.sigs.k8s.io/developer/providers/machine-infrastructure.html) defines "ready" (this definition is for Machines but the same should apply to MachinePools):

`ready (boolean): indicates the provider-specific infrastructure has been provisioned and is ready`

>  for unmanaged AzureMachinePools, their status.ready is based solely on the provisioning state of the VMSS, desired replica count, and actual replica count

There might be an opportunity to standardize Ready status for AzureMachinePools and AzureManagedMachinePools.

nojnhuh: Ok, I think that makes sense. I took a pass at moving where we set ready to false a bit lower to right after that `svc.Reconcile` call. Let me know if that seems better. The light manual testing I've done doesn't seem to show any difference in behavior.

CecileRobertMichon:
lgtm

/assign @jackfrancis 

@nojnhuh please squash commits before merge if @jackfrancis has no objections

jackfrancis:
this lgtm, will be great to get this included in 1.5

jackfrancis:
@nojnhuh to optimize re-test thrashing, squash commits sooner and we can get this in (pending test signal of course)

thank you!!!

jackfrancis (approved):
/lgtm
/approve

PR #2623, opened 2022-09-02
Restore "latest" for Ubuntu images in test templates

mboersma:
**What type of PR is this?**:

/kind cleanup

In #2618 we pinned some Ubuntu image references to the same Last Known Good version as Windows, but that version doesn't exist for the "capi" Linux offer. I think it should work to leave Ubuntu at "latest" since we have logic that negotiates agreement on a version between Windows and Linux. (Alternately, we can pin these to the correct Ubuntu version, published several days earlier.)

(Bug reported in chat by @lzhecheng.)

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```


jackfrancis (approved):
/lgtm
/approve

PR #2628, opened 2022-09-06
Document the CAPZ release cadence and support policy

### LINKED ISSUE #2295: Document the new release cadence and support policy
1. What release cadence should we adopt going forward? CAPI is [discussing formalizing a release cadence](https://github.com/kubernetes-sigs/cluster-api/issues/6442). Current proposal from discussion is to not follow CAPI release cadence but to define our own CAPZ release cadence for minor (feature) releases of either 1 minor release every month or every 2 months
2. How many minor release branches should we “support”  at any given time? For example, if we release features every 2 months, we could keep backporting bug fixes to the last 2 releases at any given time, meaning each minor release would be “in support” for ~4 months.

For release 1.3, we decided to try out using a 2 months release cadence, where the milestone due date is used to indicate the next release date.

We are also going to support (backport fixes to) the last two minor releases. For example, v1.2 and v1.3. As soon as v1.4 comes out, v1.2 becomes EOL and we maintain release branches for v1.3 and v1.4.

This issue tracks documenting that process after the initial trial and error period, if we decide to stick to it. Feedback welcome.

mboersma:
**What type of PR is this?**:

/kind documentation

Fixes #2295

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Document the CAPZ release cadence and support policy
```

### DIFF
--- a/docs/book/src/developers/releasing.md
+++ b/docs/book/src/developers/releasing.md
@@ -1,18 +1,50 @@
-# Release Process
+# CAPZ Releases
 
-## Update metadata.yaml (skip for patch releases)
+## Release Cadence
+
+CAPZ minor versions (that is, 1.**5**.0 versus 1.**4**.x) are released every two months.
+
+CAPZ patch versions (for example, 1.5.**2** versus 1.5.**1**) are released as often as weekly. Each week at the open office hours meeting, maintainers decide whether or not a patch release is called for based on community input. A patch release may bypass this cadence if circumstances warrant.
+
+## Release Support
+
+The two most recent minor releases of CAPZ will be supported with bug fixes. Since minor releases arrive every two months, each minor release receives fixes for four months.
+
+For example, let's assume CAPZ v1.4.2 is the current release, and v1.3.2 is the latest in the previous minor release line. When v1.5.0 is released, it becomes the current release. v1.4.2 becomes the previous release line and remains supported. And v1.3.2 reaches end-of-life and no longer receives support through bug fixes.
+
+Note that "support" in this context refers strictly to whether or not bug fixes are backported to a release line. Please see [the support documentation](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/SUPPORT.md) for more general information about how to get help with CAPZ.
+
+### Bug Fixes and Test Improvements
+
+Any significant user-facing bug fix that lands in the `main` branch should be backported to the current and previous release lines. Security-related fixes are automatically considered significant and user-facing.
+
+Improvements or significant changes to tests should be backported to the current release line. This is intended to minimize friction in the event of a critical test fix. Test improvements or changes may sometimes need to be backported to the previous release line in the event that tests break on all release branches.
+
+### Experimental API Changes
+
+Experimental Cluster API features (for example, `AzureManagedCluster`) may evolve more rapidly than graduated v1 features. CAPZ allows general changes, enhancements, or additions in this area to be cherry-picked into the current release branch for inclusion in patch releases. This will accelerate the effort to graduate experimental features to the stable API by allowing faster adoption and iteration.
+
+Breaking changes are also allowed in experimental APIs; those changes will not be included in a patch release, but will be introduced in a new minor release, with appropriate release notes.
+
+### Timing of Merges
+
+Sometimes pull requests touch a large number of files and are more likely to create challenges for the automated cherry-pick process. In such cases, maintainers may prefer to delay merging such changes until the end of a minor release cycle.
+
+## Release Process
+
+### Update metadata.yaml (skip for patch releases)
 
 - Make sure the [metadata.yaml](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/metadata.yaml) file is up to date and contains the new release with the correct cluster-api contract version.
   - If not, open a [PR](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1928) to add it.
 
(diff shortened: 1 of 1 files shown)


docs/book/src/developers/releasing.md
```diff
@@ -1,18 +1,40 @@
-# Release Process
+# CAPZ Releases
 
-## Update metadata.yaml (skip for patch releases)
+## Release Cadence
+
+CAPZ minor versions (that is, 1.**5**.0 versus 1.**4**.x) are released every two months.
+
+CAPZ patch versions (for example, 1.5.**2** versus 1.5.**1**) are released every two weeks. A patch release may bypass this cadence if circumstances warrant.
```
jackfrancis, line 7: We agreed during a recent office hours that we would aim for a weekly release "check in". In other words, every week we will poll the community for consensus around whether or not we should cut a patch given what's in the release branches. So if we have a regular routine of cherry-picking into the release branch w/ community support, we will in effect have a weekly release cadence. Or if the community prefers less frequency, it will be less frequent.

2022 August 25th office hours discussion notes:

https://docs.google.com/document/d/1P2FrRjuCZjGy0Yh72lwWCwmXekSEkqliUVTmJy_ETIk

docs/book/src/developers/releasing.md
```diff
@@ -1,18 +1,40 @@
...
+
+Note that "support" in this context refers strictly to whether or not bug fixes are backported to a release line. Please see [the support documentation](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/SUPPORT.md) for more general information about how to get help with CAPZ.
+
+### Bug Fixes and Test Improvements
+
+Any significant user-facing bug fix that lands in the `main` branch should be backported to the current and previous release lines. Security-related fixes are automatically considered significant and user-facing.
+
+Improvements or significant changes to tests should be backported to the current release line. This is intended to minimize friction in the event of a critical test fix. Test improvements or changes may sometimes need to be backported to the previous release line in the event that tests break on all release branches.
+
```
jackfrancis, line 22: Let's add a discrete section to document that for experimental APIs (e.g., `AzureManagedCluster`) we may include both _features or bug fixes_ in patch releases. This an exception to the normal semver definition, but it will accelerate our effort to graduate experimental features to v1 by allowing faster feature adoption and functional iteration.

cc @CecileRobertMichon @nojnhuh @devigned @zmalik @mtougeron @LochanRn @JoelSpeed for any add'l thoughts/dissent here

jackfrancis: And how about another 2-3 sentences explaining that due to our minor release support contract, we want to optimize the automated cherry-pick path from main to those two minor release branches. That may include delaying codebase-wide changes like housekeeping PRs (updating downstream APIs, linting, etc) until the end of a minor release cycle. If we integrate these housekeeping PRs in the middle of a release cycle, we can then be stuck with only manual cherry-picks for the rest of the release cycle after that.

mtougeron: So long as the new features are backwards compatible and don't significantly change structure or functionality I personally don't see a problem with it for experimental APIs. It really depends on what level of contract we want to provide to the users of the experimental features. I think we need to be more careful once a feature reaches beta but I definitely agree for alpha features.

nojnhuh: I agree with @mtougeron re: not breaking experimental APIs in a patch release. I can see where certain backwards-incompatible changes might be needed a bit sooner than the next minor release, though I think keeping experimental APIs backwards-compatible between patch releases is a good rule of thumb.

jackfrancis: I think this is a fair distinction:

1) additive features/functionality
2) changed/updated/removed features/functionality in response to the normal process of iterating in an experimental functional workstream

Part of our work in `AzureManagedCluster` will yield stuff that falls under #2 above. When we have a defensible reason for actually changing (i.e., breaking not just additive) an implementation classified as experimental, we should do that type of explicit breakage in a new minor release (with all-caps release notes for folks taking part in the experiment).

Does this agree with your points @nojnhuh @mtougeron?

mboersma:
/release-note Document the CAPZ release cadence and support policy

mboersma:
/release-note-edit ```release-note Document the CAPZ release cadence and support policy```

docs/book/src/developers/releasing.md
```diff
@@ -1,18 +1,48 @@
...
+### Experimental API Changes
+
+Experimental Cluster API features (for example, `AzureManagedCluster`) are not expected to be stable, so CAPZ allows general changes, enhancements, or additions in this area to be cherry-picked into the current release branch. This will accelerate the effort to graduate experimental features to the stable API by allowing faster adoption and iteration.
```
jackfrancis, line 25: Let's walk this back a little bit. How about:

> Experimental Cluster API features (for example, `AzureManagedCluster`) are expected to evolve more rapidly than graduated v1 features. CAPZ allows general changes, enhancements, or additions in this area to be cherry-picked into the current release branch for inclusion in patch releases. This will accelerate the effort to graduate experimental features to the stable API by allowing faster adoption and iteration. Breaking changes are also allowed in experimental APIs; those changes will not be included in a patch release, but will be introduced in a new minor release, with appropriate release notes.

mboersma: I like it–updated.

jackfrancis (approved):
/lgtm
/approve

PR #2631, opened 2022-09-06
Add support for custom vm extensions

### LINKED ISSUE #2221: Support Custom VM Extensions for AzureMachines
/kind feature

Support deployment of custom VM extensions from the Azure Extension gallery as part of a VM deployment. Currently if you want to add any extensions to a deployed VM you have to deploy afterwards via Azure Portal or az cli(https://docs.microsoft.com/en-us/azure/virtual-machines/extensions/overview). 

A custom extension is already deployed to facilitate capz bootstrapping however additional extensions are not supported. The use case I am targeting directly is the deployment of the "Microsoft.Azure.ActiveDirectory.AADSSHLoginForLinux" extensions to facilitate AAD SSH login to the virtual machines. 

**Environment:**

- cluster-api-provider-azure version: 1.2.0
- Kubernetes version: (use `kubectl version`): 1.23.5 
 ```
NAME="Ubuntu"
VERSION="20.04.4 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.4 LTS"
VERSION_ID="20.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal
```

### PR DESCRIPTION
**Pull request by willie-yao**
**What type of PR is this?**
/kind feature

This PR adds support for deployment of custom VM extensions from the Azure Extension gallery as part of a VM and VMSS deployment.

Fixes #2221 

In order to test the extension provisions successfully, I used Tilt to deploy the default flavor. I added the VMExtensions field to the default cluster template's AzureMachineTemplate spec like so:
```
vmExtensions:
      - name: CustomScript
        publisher: Microsoft.Azure.Extensions
        version: '2.1'
        settings:
          skipDos2Unix: false
        protectedSettings:
          commandToExecute: |
            #!/bin/sh
            echo "Updating packages ..."
            apt update
            apt upgrade -y
```

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Add support for custom vm extensions
```


azure/scope/machinepool.go
```diff
@@ -127,6 +127,7 @@ func (m *MachinePoolScope) ScaleSetSpec() azure.ScaleSetSpec {
 		SpotVMOptions:                m.AzureMachinePool.Spec.Template.SpotVMOptions,
 		FailureDomains:               m.MachinePool.Spec.FailureDomains,
 		TerminateNotificationTimeout: m.AzureMachinePool.Spec.Template.TerminateNotificationTimeout,
+		VMExtensions:                 m.AzureMachinePool.Spec.Template.VMExtensions,
```
CecileRobertMichon, line 130: why is this change needed?

azure/services/virtualmachines/spec.go
```diff
@@ -54,6 +54,7 @@ type VMSpec struct {
 	Image                  *infrav1.Image
 	BootstrapData          string
 	ProviderID             string
+	VMExtensions           []infrav1.VMExtension
```
CecileRobertMichon, line 57: where is this used ?

config/default/manager_pull_policy.yaml
```diff
@@ -8,4 +8,4 @@ spec:
     spec:
       containers:
       - name: manager
-        imagePullPolicy: Always
+        imagePullPolicy: IfNotPresent
```
CecileRobertMichon, line 11: you might have pushed this change by accident

(aside: we really need to find a a way to not modify this file when running e2e...)

exp/api/v1beta1/azuremachinepool_types.go
```diff
@@ -87,6 +87,10 @@ type (
 		// SubnetName selects the Subnet where the VMSS will be placed
 		// +optional
 		SubnetName string `json:"subnetName,omitempty"`
+
+		// VMExtensions specifies a list of extensions to be added to the virtual machine.
```
CecileRobertMichon, line 91:
```suggestion
		// VMExtensions specifies a list of extensions to be added to the scale set.
```

api/v1beta1/azuremachine_types.go
```diff
@@ -122,6 +122,10 @@ type AzureMachineSpec struct {
 	// DNSServers adds a list of DNS Server IP addresses to the VM NICs.
 	// +optional
 	DNSServers []string `json:"dnsServers,omitempty"`
+
+	// VMExtensions specifies a list of extensions to be added to the virtual machine.
+	// +optional
+	VMExtensions []VMExtension `json:"vmExtensions,omitempty"`
```
CecileRobertMichon, line 128: I wonder if it would make sense to call this just "Extensions" since everything in here is pretty much in the context of a VM... but on the other hand I could see just "extensions" being too vague as it's used in other contexts 🤔 

wonder if others have thoughts?

templates/test/ci/patches/azuremachinepool-vmextension.yaml
```diff
@@ -0,0 +1,17 @@
...
+spec:
+  template:
+    vmExtensions:
+      - name: CustomScript
+        publisher: Microsoft.Azure.Extensions
+        version: '2.1'
+        protectedSettings:
+          commandToExecute: |
+            #!/bin/sh
+            echo "Updating packages ..."
```
CecileRobertMichon, line 16: let's maybe make the echo statement less random and explain what this is for?

e.g. `echo "this script is a no-op used for extension testing purposes ..."`

templates/test/ci/patches/azuremachinetemplate-vmextension.yaml
```diff
@@ -0,0 +1,18 @@
...
+  template:
+    spec:
+      vmExtensions:
+      - name: CustomScript
+        publisher: Microsoft.Azure.Extensions
+        version: '2.1'
+        protectedSettings:
+          commandToExecute: |
+            #!/bin/sh
+            echo "Updating packages ..."
```
CecileRobertMichon, line 17: ditto

templates/test/ci/prow/kustomization.yaml
```diff
@@ -18,6 +18,7 @@ patchesStrategicMerge:
   - ../patches/machine-deployment-worker-counts.yaml
   - ../../../azure-cluster-identity/azurecluster-identity-ref.yaml
   - ../../../flavors/base-windows-containerd/kubeadm-control-plane.yaml
+  - ../patches/azuremachinetemplate-vmextension.yaml
```
CecileRobertMichon, line 21: so I'm slightly concerned about the number of templates this change ended up modifying, given that the "prow" test flavor is used as the basis for most other things... another argument for restructuring our test templates 

Not sure how we can avoid this without doing a major refactor of templates though, and there isn't a single template that makes more sense to use as _the_  test template for extensions.

test/e2e/azure_vmextensions.go
```diff
@@ -0,0 +1,120 @@
...
+	// list all of the requested objects within the cluster namespace with the cluster name label
+	Logf("Listing machines in namespace %s with label %s=%s", input.Namespace.Name, clusterv1.ClusterLabelName, workloadClusterProxy.GetName())
+	err := mgmtClient.List(ctx, machineList, client.InNamespace(input.Namespace.Name), client.MatchingLabels{clusterv1.ClusterLabelName: workloadClusterProxy.GetName()})
+	Expect(err).NotTo(HaveOccurred())
+
+	By("Ensuring the machine spec contains expected VM extensions")
+	for _, machine := range machineList.Items {
+		if strings.HasPrefix(machine.Name, input.VMName) {
+			Expect(len(machine.Spec.VMExtensions)).To(Equal(1))
+			Expect(machine.Spec.VMExtensions[0].Name).To(Equal("CustomScript"))
```
CecileRobertMichon, line 77: so here our test is essentially validating the yaml template and expecting a very precise extension to be present on a very specific machine at a hardcoded `0` position. This could cause issues in the future if we add extensions, rename the machine, etc.

Since the main goal of the e2e test is to test CAPZ code (in this case that adding a VM extension in the AzureMachine Spec results in the Azure VM that gets created to have that extension), what would you think of rewriting the test to be a bit more generic, ie. For all extensions defined in all the AzureMachines in the cluster, verify that the extension exists in the corresponding VM in Azure. In otherwords, we could iterate through the machines like you're doing here, and then check that the VM for each machine has whatever extensions defined in spec.

That way, this test could also be reused for multiple cluster templates that have different extensions and AzureMachine configurations (maybe even several in AzureMachines in one cluster) and wouldn't need to be enabled just for the specific cluster test / template combination that has the extension in the template.

api/v1beta1/azuremachine_types.go
```diff
@@ -122,6 +122,10 @@ type AzureMachineSpec struct {
 	// DNSServers adds a list of DNS Server IP addresses to the VM NICs.
 	// +optional
 	DNSServers []string `json:"dnsServers,omitempty"`
+
+	// VMExtensions specifies a list of extensions to be added to the virtual machine.
```
CecileRobertMichon, line 126: Would be good to add a short doc on this feature to https://github.com/kubernetes-sigs/cluster-api-provider-azure/tree/main/docs/book/src/topics so it can show up in the book

CecileRobertMichon (commented):
Great work @willie-yao, overall this looks good to me - just some minor comments + a suggestion to improve testing

azure/scope/machinepool.go
willie-yao: I thought that we would need to change the scale set spec since we added the VMExtensions field to the spec in case it is needed anywhere. It is currently not being used. Should I remove it if that's the case?

templates/test/ci/prow/kustomization.yaml
willie-yao: That's true. I considered creating a whole new template just for testing extensions, but that might be too much make the E2E test run slower. Adding the test to the default template made the most sense to me, but it may be better design-wise to incorporate it into a different test.

test/e2e/azure_vmextensions.go
willie-yao: This is a much better approach! I'm trying to implement this now and I'm having some trouble trying to get the machine spec, since the windows machine names on Azure are different than the machine names defined in the spec. Is there a way to get a Machine other than the name itself?

willie-yao: I think my idea right now is to create a mapping of ProviderIDs to the extension names from the spec, then iterate through the machines and verify the extensions using the ID as a key. I'm assuming that the spec's ProviderID is equal to the Virtual Machine's ID

CecileRobertMichon: > I'm assuming that the spec's ProviderID is equal to the Virtual Machine's ID

that's correct

api/v1beta1/azuremachine_types.go
willie-yao: Just added in my latest revision. Let me know what you think!

api/v1beta1/azuremachine_types.go
willie-yao: I saw that other pre-existing functions referring to VM extensions called it "VMExtensions" rather than "Extensions", such as the [VMExtensionSpecs() function](https://pkg.go.dev/sigs.k8s.io/cluster-api-provider-azure/azure/scope#MachineScope.VMExtensionSpecs) in `machine.go`. In order to keep things consistent, I called it `VMExtensions`. I'm not sure if other Azure resources have extensions, so maybe I'm being too specific here.

test/e2e/azure_vmextensions.go
willie-yao: I did this by creating a map of machineIDs to an array of expected VM extensions from the spec, and ensured that the VM extensions from Azure matched it. I'm not sure if this is the best way to go about it, but let me know what you think! @CecileRobertMichon 

[Code](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2631/files#diff-2bea7aae68fda3b77b9d0180bd026269d0804ded551d4d9d5897e528cf20da31R72-R78)

api/v1beta1/azuremachine_types.go
jackfrancis: I prefer having the "VM" prefix (disclaimer: I sort of really don't like the fact that we have to use such a generic word like "extension" to mean something very specific like "runs any arbitrary executable code on a VM at create time")

test/e2e/azure_vmextensions.go
```diff
@@ -0,0 +1,112 @@
...
+	for _, machine := range vmListResults.Values() {
+		vmExtensionListResult, err := extensionsClient.List(ctx, input.ClusterName, *machine.Name, "")
+		Expect(err).NotTo(HaveOccurred())
+		vmExtensionList := *vmExtensionListResult.Value
+		var vmExtensionNames []string
+		for _, vmExtension := range vmExtensionList {
+			vmExtensionNames = append(vmExtensionNames, *vmExtension.Name)
+		}
+		osName := string(machine.VirtualMachineProperties.StorageProfile.OsDisk.OsType)
+		Expect(vmExtensionNames).To(ContainElements("CAPZ." + osName + ".Bootstrapping"))
```
jackfrancis, line 109: So one thing that occurs to me: because the bootstrapping extension is present on (non-AKS) clusters we build, we could run this validation on all (non-AKS) cluster scenarios. It looks like you've already wired this up to detect the extensions defined on the capi resources, so we shouldn't have to curate which cluster scenarios we invoke `AzureVMExtensionsSpec` against.

api/v1beta1/azuremachine_types.go
CecileRobertMichon: > "runs any arbitrary executable code on a VM at create time"

that's only the case for the custom script extension (CSE), this would allow users to add any published Azure Marketplace extensions to the VMs, including custom script but also others that have very precise functionality. I also think the word "extension" is overloaded and vague but it's consistent with official Azure naming...

jackfrancis: How about `AzureExtension`?

jackfrancis: Is this doc only talking about CSE?

https://learn.microsoft.com/en-us/azure/virtual-machines/extensions/features-linux?tabs=azure-cli

(it regularly refers to "VM Extensions")

CecileRobertMichon: that one is referring to all VM extensions, not just custom script
> Each Azure VM extension has a specific use case. Examples include:
> ...
> In addition to process-specific extensions, a Custom Script extension is available for both Windows and Linux virtual machines. The [Custom Script extension for Linux](https://learn.microsoft.com/en-us/azure/virtual-machines/extensions/custom-script-linux) allows any Bash script to be run on a VM

https://learn.microsoft.com/en-us/azure/virtual-machines/extensions/features-linux?tabs=azure-cli#use-cases-and-samples

CecileRobertMichon: I think given above, `VMExtensions` makes the most sense

test/e2e/azure_vmextensions.go
willie-yao: That's true, should I added it to all other E2E test scenarios besides the AKS one? Theoretically this should work with other test scenarios as well, as it checks for extensions that are defined in the spec.

test/e2e/azure_test.go
```diff
@@ -234,6 +234,17 @@ var _ = Describe("Workload cluster creation", func() {
 				},
 			}, result)
 
+			By("Verifying a VM extension is present on the node", func() {
```
CecileRobertMichon, line 237:
```suggestion
			By("Verifying expected VM extensions are present on the node", func() {
```

same below

test/e2e/azure_vmssextensions.go
```diff
@@ -0,0 +1,111 @@
...
+// AzureVMSSExtensionsSpecInput is the input for AzureVMSSExtensionsSpec.
+type AzureVMSSExtensionsSpecInput struct {
+	BootstrapClusterProxy framework.ClusterProxy
+	Namespace             *corev1.Namespace
+	VMName                string
+	ClusterName           string
+}
+
+// AzureVMSSExtensionsSpec implements a test that verifies VM extensions are created and deleted.
+func AzureVMSSExtensionsSpec(ctx context.Context, inputGetter func() AzureVMSSExtensionsSpecInput) {
```
CecileRobertMichon, line 45: Since this has some overlapping code with the VM version of this, it might make sense to combine them into one spec. Since we're ranging over machines and machinepools it will simply skip those if the cluster has none.

willie-yao: Ah I didn't think about that! There might still be a lot of duplicated code since machines and machinepools use different types, but it does make sense to combine them into one spec.

docs/book/src/topics/custom-vm-extensions.md
```diff
@@ -0,0 +1,68 @@
+# Custom VM Extensions
+
+## Overview
```
CecileRobertMichon, line 3: Thanks for adding this doc, looks good 💯 

Can you please add it to https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/7d8dc8583b1c8db99a84768e5ed7f2f140983a05/docs/book/src/SUMMARY.md?plain=1 under Topics so it shows up in the book Table Of Content?

willie-yao: Got it! I'll add it under the other "custom" topics. I'm not sure if there is a specific ordering of topics, but I felt like that would be the best place to add it!

templates/test/ci/prow/kustomization.yaml
CecileRobertMichon: Let's not tackle this in this PR but we're overdue for a template cleanup/simplification. Some of the cluster test templates could probably be combined (anything that tests a machine / machinepool feature doesn't need its own cluster)

CecileRobertMichon (commented):
/lgtm

docs/book/src/topics/custom-vm-extensions.md
willie-yao: I thought so too, but it seems like a few are out of order. "Custom Images" is under "Custom Private DNS Zone name", and there are a few others like "SSH Access to nodes". Should we standardize how we order the topics?

CecileRobertMichon: I'm guessing "Custom Private DNS Zone" is ordered by file name "custom-dns" (d comes before i) and "SSH Access to nodes" was likely just added at the end by someone who didn't realize the list was in order. Feel free to reorganize/rename however you see fit, the ordered list is becoming less practical now that we have so many topics, maybe we need sub-sections now. This can be a separate PR so it doesn't block this one.

willie-yao:
@CecileRobertMichon I saw that there is a failed test with broken links. I'm assuming this is intended because the link for the new docs page hasn't been published yet. Is that correct?

jackfrancis:
> @CecileRobertMichon I saw that there is a failed test with broken links. I'm assuming this is intended because the link for the new docs page hasn't been published yet. Is that correct?

It's just bad luck, I got the same error on my PR, here's the fix:

https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2680/files#diff-4d554fade8fefc047dcf883418cc72311cb5b62f3e0d20f7cbbfb849faa7335aR525

tl;dr k8s maintains 4 versions of that link, and then garbage collects the oldest one when a new release is needed

Not sure if we wanna include it in this PR or rebase this on top of a PR that fixes it :(

willie-yao:
@jackfrancis Oh that makes sense! If this blocks merging, I'm not sure what's the better way to do it. I'm leaning towards including this in a separate PR since it's not related to custom vm extensions.

jackfrancis:
@willie-yao you can rebase on top of latest main as this PR delivered the needful: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2680

willie-yao:
@jackfrancis @CecileRobertMichon Did the rebase and the test passes now. Does everything look good for approval?

jackfrancis (approved):
/lgtm
/approve

PR #2636, opened 2022-09-07
apply SSA changes for capi 1.2

jackfrancis:
**What type of PR is this?**

/kind api-change

This PR updates api/v1beta1 slice types so that they are enabled for SSA, which ClusterClass requires as of capi 1.2. See:

- https://cluster-api.sigs.k8s.io/developer/providers/v1.1-to-v1.2.html#required-api-changes-for-providers

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
apply SSA changes for capi 1.2
```

### DIFF
--- a/api/v1beta1/types.go
+++ b/api/v1beta1/types.go
@@ -135,6 +135,8 @@ func (v *VnetSpec) IsManaged(clusterName string) bool {
 }
 
 // Subnets is a slice of Subnet.
+// +listType=map
+// +listMapKey=name
 type Subnets []SubnetSpec
 
 // SecurityGroup defines an Azure security group.
@@ -230,6 +232,8 @@ type SecurityRule struct {
 }
 
 // SecurityRules is a slice of Azure security rules for security groups.
+// +listType=map
+// +listMapKey=name
 type SecurityRules []SecurityRule
 
 // LoadBalancerSpec defines an Azure load balancer.
--- a/api/v1beta1/types_template.go
+++ b/api/v1beta1/types_template.go
@@ -86,10 +86,15 @@ type VnetTemplateSpec struct {
 }
 
 // VnetPeeringsTemplateSpec defines a list of peerings of the newly created virtual network with existing virtual networks.
+// +listType=map
+// +listMapKey=remoteVnetName
 type VnetPeeringsTemplateSpec []VnetPeeringClassSpec
 
 // SubnetTemplateSpec specifies a template for a subnet.
 type SubnetTemplateSpec struct {
+	// Name is the name for this subnet.
+	Name string `json:"name"`
+
 	SubnetClassSpec `json:",inline"`
 
 	// SecurityGroup defines the NSG (network security group) that should be attached to this subnet.
@@ -107,6 +112,8 @@ func (s SubnetTemplateSpec) IsNatGatewayEnabled() bool {
 }
 
 // SubnetTemplatesSpec specifies a list of subnet templates.
+// +listType=map
+// +listMapKey=name
 type SubnetTemplatesSpec []SubnetTemplateSpec
 
 // BastionTemplateSpec specifies a template for a bastion host.
--- a/Makefile
+++ b/Makefile
@@ -69,7 +69,7 @@ ifneq ($(abspath $(ROOT_DIR)),$(GOPATH)/src/sigs.k8s.io/cluster-api-provider-azu
 endif
 
 # Binaries.
-CONTROLLER_GEN_VER := v0.8.0
+CONTROLLER_GEN_VER := v0.9.2
 CONTROLLER_GEN_BIN := controller-gen
 CONTROLLER_GEN := $(TOOLS_BIN_DIR)/$(CONTROLLER_GEN_BIN)-$(CONTROLLER_GEN_VER)
 
--- a/config/crd/bases/infrastructure.cluster.x-k8s.io_azureclusteridentities.yaml
+++ b/config/crd/bases/infrastructure.cluster.x-k8s.io_azureclusteridentities.yaml
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.8.0
+    controller-gen.kubebuilder.io/version: v0.9.2
   creationTimestamp: null
   name: azureclusteridentities.infrastructure.cluster.x-k8s.io
 spec:
@@ -64,6 +64,7 @@ spec:
                       name must be unique.
                     type: string
                 type: object
+                x-kubernetes-map-type: atomic
               resourceID:
                 description: User assigned MSI resource id.
                 type: string
@@ -224,6 +225,7 @@ spec:
                           "value". The requirements are ANDed.
                         type: object
                     type: object
+                    x-kubernetes-map-type: atomic
                 type: object
               clientID:
                 description: Both User Assigned MSI and SP can use this field.
@@ -241,6 +243,7 @@ spec:
                       name must be unique.
                     type: string
                 type: object
+                x-kubernetes-map-type: atomic
               resourceID:
                 description: User assigned MSI resource id.
                 type: string
@@ -401,6 +404,7 @@ spec:
                           "value". The requirements are ANDed.
                         type: object
                     type: object
+                    x-kubernetes-map-type: atomic
                 type: object
               clientID:
                 description: ClientID is the service principal client ID. Both User
@@ -419,6 +423,7 @@ spec:
                       name must be unique.
                     type: string
                 type: object
+                x-kubernetes-map-type: atomic
               resourceID:
                 description: ResourceID is the Azure resource ID for the User Assigned
                   MSI resource. Only applicable when type is UserAssignedMSI.
@@ -495,9 +500,3 @@ spec:
     storage: true
     subresources:
       status: {}
-status:
-  acceptedNames:
-    kind: ""
-    plural: ""
-  conditions: []
-  storedVersions: []
--- a/config/crd/bases/infrastructure.cluster.x-k8s.io_azureclusters.yaml
+++ b/config/crd/bases/infrastructure.cluster.x-k8s.io_azureclusters.yaml
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.8.0
+    controller-gen.kubebuilder.io/version: v0.9.2
   creationTimestamp: null
   name: azureclusters.infrastructure.cluster.x-k8s.io
 spec:
@@ -121,6 +121,7 @@ spec:
                     description: 'UID of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#uids'
                     type: string
                 type: object
+                x-kubernetes-map-type: atomic
               location:
                 type: string
               networkSpec:
@@ -778,6 +779,7 @@ spec:
                     description: 'UID of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#uids'
                     type: string
                 type: object
+                x-kubernetes-map-type: atomic
               location:
                 type: string
               networkSpec:
@@ -1508,6 +1510,9 @@ spec:
                                   - protocol
                                   type: object
                                 type: array
+                                x-kubernetes-list-map-keys:
+                                - name
+                                x-kubernetes-list-type: map
                               tags:
                                 additionalProperties:
                                   type: string
@@ -1667,6 +1672,7 @@ spec:
                     description: 'UID of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#uids'
                     type: string
                 type: object
+                x-kubernetes-map-type: atomic
               location:
                 type: string
               networkSpec:
@@ -2057,6 +2063,9 @@ spec:
                                 - protocol
                                 type: object
                               type: array
+                              x-kubernetes-list-map-keys:
+                              - name
+                              x-kubernetes-list-type: map
                             tags:
                               additionalProperties:
                                 type: string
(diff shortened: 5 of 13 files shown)


sbueringer:
Changes look good to me so far, assuming those are the relevant slices.

The diff is expected as there were some changes in controller-gen:
* status is not rendered anymore
* more atomic markers are set (but atomic the default if not set, so no effectively that doesn't change anything)

Not sure if the 2nd point is also in scope of this PR, but otherwise lgtm

api/v1beta1/types.go
```diff
@@ -135,6 +135,8 @@ func (v *VnetSpec) IsManaged(clusterName string) bool {
 }
 
 // Subnets is a slice of Subnet.
+// +listType=map
+// +listMapKey=role
```
CecileRobertMichon, line 139: why are we using role as map key? role isn't unique but name is so name might be a more appropriate map key if map key needs to be unique

api/v1beta1/types_template.go
```diff
@@ -86,6 +86,8 @@ type VnetTemplateSpec struct {
 }
 
 // VnetPeeringsTemplateSpec defines a list of peerings of the newly created virtual network with existing virtual networks.
+// +listType=map
+// +listMapKey=remoteVnetName
```
CecileRobertMichon, line 90: same question here, I don't know if remoteVnetName is a unique identifier

jackfrancis: it's the only property in the underlying struct inside the slice, is it possible we'd be retaining more than one items in this slice that are identical?

api/v1beta1/types.go
jackfrancis: `name` seemed to work here (controller-gen didn't complain, at least), updated

api/v1beta1/types_template.go
```diff
@@ -107,6 +109,8 @@ func (s SubnetTemplateSpec) IsNatGatewayEnabled() bool {
 }
 
 // SubnetTemplatesSpec specifies a list of subnet templates.
+// +listType=map
+// +listMapKey=role
```
jackfrancis, line 113: @CecileRobertMichon do you have any thoughts on this one? If `role` isn't sufficiently unique we'll probably have to define a union of properties for unique identification

api/v1beta1/types.go
sbueringer: The problem with name is that it doesn't exist in the AzureClusterTemplate struct

api/v1beta1/types_template.go
sbueringer: Could it be that atomic merge behavior is good enough for this field, as there are no two clients trying to set different parts of the slice?

api/v1beta1/types_template.go
CecileRobertMichon: SubnetTemplate spec is only a subset of the SubnetSpec fields that are shared across clusters and was added for ClusterClass in order to be used in AzureClusterTemplate only. AFAIK there is no unique field combination in `SubnetTemplatesSpec` as-is... The uniqueness will only come from `name`, which will be generated to be a combination of cluster name + role + unique index.

TBH, I'm not too sure why `name` wasn't included in the template since you could have repeated subnet names that are the same across clusters, I think? Maybe @ykakarap remembers more than I do since he authored https://github.com/kubernetes-sigs/cluster-api-provider-azure/commit/99a589cef3e119ff7cbf890e9514da5309965dba

CecileRobertMichon: For reference: `name` uniqueness validation is here: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/b59c9198f45c3cb13a0b1209cb8167ebcc4156ac/api/v1beta1/azurecluster_validation.go#L180

api/v1beta1/types_template.go
CecileRobertMichon: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/b59c9198f45c3cb13a0b1209cb8167ebcc4156ac/api/v1beta1/azurecluster_validation.go#L266

uniqueness comes from resource group + remote vnet name for a vnet peering, this again the same problem as below where VnetPeeringsTemplateSpec is only a subset of VnetPeeringsSpec and is missing resource group, which might be an issue if you want to specify more than one vnet with the same name (which would be a valid thing to do if you have two vnets in two different RG with the same name).

api/v1beta1/types.go
jackfrancis: @sbueringer are you thinking of this type?

```
// SubnetTemplatesSpec specifies a list of subnet templates.
// +listType=map
// +listMapKey=name
type SubnetTemplatesSpec []SubnetTemplateSpec
```

Indeed the `SubnetTemplateSpec` type doesn't have a `Name` property, I wonder why code generation isn't complaining...

sbueringer: Probably the code generation isn't verifying that. But looks like the apiserver is:

> unable to install CRDs onto control plane: unable to create CRD instances: unable to create CRD "azureclustertemplates.infrastructure.cluster.x-k8s.io": CustomResourceDefinition.apiextensions.k8s.io "azureclustertemplates.infrastructure.cluster.x-k8s.io" is invalid: spec.validation.openAPIV3Schema.properties[spec].properties[template].properties[spec].properties[networkSpec].properties[subnets].x-kubernetes-list-map-keys: Invalid value: []string{"name"}: entries must all be names of item properties

api/v1beta1/types_template.go
sbueringer: > TBH, I'm not too sure why name wasn't included in the template since you could have repeated subnet names that are the same across clusters, I think? Maybe @ykakarap remembers more than I do since he authored https://github.com/kubernetes-sigs/cluster-api-provider-azure/commit/99a589cef3e119ff7cbf890e9514da5309965dba

Yup. If we could introduce name here as well and make it the map key, this would solve all our problems.

@ykakarap do you know if there was a strong reason not to add it here?

sbueringer: > The uniqueness will only come from name, which will be generated to be a combination of cluster name + role + unique index.

@CecileRobertMichon  That's how it's done in the AzureCluster webhook, right? Would it be a problem if the names are hard-coded to unique values in the AzureClusterTemplate and then used for multiple clusters? (aka does the name have to be unique across clusters?)

CecileRobertMichon: >  this would solve all our problems.

I like the sound of that

> Would it be a problem if the names are hard-coded to unique values in the AzureClusterTemplate and then used for multiple clusters? (aka does the name have to be unique across clusters?)

no it shouldn't be, subnet name just needs to be unique per cluster (ie. two subnets in the same cluster can't have the same name), maybe that's where the confusion came from

sbueringer: @CecileRobertMichon Found this comment: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1501/files#r673287133

Do you know why? :)

api/v1beta1/types_template.go
```diff
@@ -86,10 +86,15 @@ type VnetTemplateSpec struct {
...
 
 // VnetPeeringsTemplateSpec defines a list of peerings of the newly created virtual network with existing virtual networks.
+// +listType=map
+// +listMapKey=remoteVnetName
 type VnetPeeringsTemplateSpec []VnetPeeringClassSpec
 
 // SubnetTemplateSpec specifies a template for a subnet.
 type SubnetTemplateSpec struct {
+	// Name is the name for this subnet.
+	Name string `json:"name"`
```
jackfrancis, line 96: @CecileRobertMichon @sbueringer @ykakarap I assume this is the change you're talking about. Obviously we can't simply update this type spec without a lot more considerations, but wanted to make sure I understand that conversation

sbueringer: Yup exactly

api/v1beta1/types_template.go
CecileRobertMichon: So in Azure, there are naming scopes based on each resource type: https://learn.microsoft.com/en-us/azure/cloud-adoption-framework/ready/azure-best-practices/resource-naming#example-names-networking

For subnets, the scope is vnet. Meaning two subnets can share the same name as long as they are in different vnets.

For vnets, the scope is resource group. Meaning two vnets can share the same name as long as they are in different resource groups.

For resource groups, the scope is subscription. Meaning two resource groups can share the same name as long as they are in different subscriptions.

So in theory, a cluster class template could have hard-coded unique subnet names as long as a different vnet and/or resource group and/or subscription is used for each cluster. I think each cluster in a clusterclass template should definitely have its own resource group, because chances of using a different subscription for each cluster in practice are pretty low. And if that's the case (one resource group == one cluster), then we can safely reuse subnet names across clusters.

Does that make sense?

api/v1beta1/types_template.go
CecileRobertMichon: I think it should be added inside https://github.com/jackfrancis/cluster-api-provider-azure/blob/2ce64d384564971e96c83718cf9a9aee6ba1ce39/api/v1beta1/types_class.go#L74 no? it's nested inline so in practice same result but it's part of the subnet definition

jackfrancis: Happy to do either. Should we file an issue to enumerate how we can safely implement this breaking change?

CecileRobertMichon: FWIW ClusterClass is experimental and behind a feature flag and these types are not used anywhere outside ClusterClass so breaking changes are fair game, as documented in experimental policy

jackfrancis: Ooohhhhh 💪 

Great, so I'll move the `Name` into the `SubnetClassSpec` type itself, and then audit the code and reference templates and make all the changes.

We'll definitely want to call this out in the release.

api/v1beta1/types_template.go
sbueringer: Sounds good to me thx

ykakarap: >@ykakarap do you know if there was a strong reason not to add it here?

Did that because the last time we had a discussion about thing when introducing the AzureClusterTemplate it was told that the names had to be unique (thanks for the link to the old comment @sbueringer). That was the only reason. I do not recall any other reason for doing that. 

If the names can be duplicate among multiple clusters then I don't see a problem with making it part of `SubnetTemplatesSpec`.

**Note:** Once we move `Name` into `SubnetTemplatesSpec` we will need to introduce validation in `[validateSubnetTemplates](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/ecf9115bb4230527f8a93098b8590f5e5d15443e/api/v1beta1/azureclustertemplate_validation.go#L77-L78)` to make sure that there are no duplicate name within the same AzureClusterTemplate.

**Note:** We will also need to introduce defaulting for this field in AzureClusterTemplate. Otherwise it could lead to infinite reconcile (?).

sbueringer: @ykakarap Thx for the infos!

> Note: We will also need to introduce defaulting for this field in AzureClusterTemplate. Otherwise it could lead to infinite reconcile (?).

I'm not sure we have to introduce defaulting, but I'm definitely not against it. By introducing `name` as a mandatory field it always has to be set. So I'm not sure how this would lead to infinite reconciles. I would probably lean towards not having defaulting so that the subnet name really has to be defined in the AzureClusterTemplate which should make it easier to match the config in AzureClusterTemplate to the resulting AzureCluster (as the names would have to exist in the local AzureClusterTemplate YAML and in the resulting AzureClusters).

If I understand the current state in AzureCluster correctly the field is mandatory there but users don't have to set it on client-side as the webhook auto-generates the name if they don't exist

api/v1beta1/types.go
sbueringer: I think we can resolve this conversation in favor of the other one

api/v1beta1/types_template.go
sbueringer: Sounds like we really don't have a unique identifier in `VnetPeeringClassSpec` and in this case we shouldn't introduce resourceGroup to get to a map key consisting of resourceGroup and remoteVnetName. (as resourceGroup should be different per cluster)

Given:
* that per default we have atomic merge behavior and 
* we only need a map key to allow the `map` merging behavior ("At any point in time, a single manager owns the list. If set or map, different managers can manage entries separately.")

Do we have a use caseI were we want different managers (controllers) to be able to manage/own different entries of this list?

api/v1beta1/types_template.go
sbueringer: @CecileRobertMichon We currently have the following struct hierarchies;

AzureCluster > NetworkSpec > Subnets > SubnetSpec (contains Name) > SubnetClassSpec

AzureClusterTemplate > NetworkTemplateSpec > SubnetTemplatesSpec > SubnetTemplateSpec (missing Name) > SubnetClassSpec

I think there are two options:

1. Add the `Name` to `SubnetClassSpec`, drop it from `SubnetSpec`
2. Add the `Name` to `SubnetTemplateSpec`

Given that the idea of `SubnetClassSpec` is to contain the shared fields between AzureCluster and AzureClusterTemplate it would be the right place to keep this pattern, we just should't forget to drop `Name` from `SubnetSpec`)

api/v1beta1/types_template.go
ykakarap: Oh you are right! Since it is mandatory field we don't have do defaulting. Just validation should be enough.

sonasingh46:
Thanks for driving this one Jack. As discussed, I think just little more is required and I will add a new PR by taking the changes that you did here. Will add you as a co-author there. 
/assign sonasingh46

sbueringer:
One last finding.

Can you check the YAML files for e2e tests. All subnets now have to set the `name `field in all AzureClusterTemplates, e.g.:
* https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/15e070c6633ea0bcc7804806e90854e9551a02c0/templates/test/ci/cluster-template-prow-clusterclass-ci-default.yaml#L466-L470
* https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/15e070c6633ea0bcc7804806e90854e9551a02c0/templates/test/ci/prow-clusterclass-ci-default/base.yaml#L153-L157
(not sure if there are more)

I assume there is a corresponding e2e tests. Is it possible to run it on this PR?

jackfrancis:
#2708 is what we want!

PR #2641, opened 2022-09-08
AzureMachineTemplate webhooks dry-run

jackfrancis:
**What type of PR is this?**

/kind api-change

**What this PR does / why we need it**:

This PR wires up the `AzureMachineTemplate` to support server-side dry run reconciliation, as specified here:

- https://cluster-api.sigs.k8s.io/developer/providers/v1.1-to-v1.2.html#required-api-changes-for-providers

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
AzureMachineTemplate webhooks dry-run
```

PR #2646, opened 2022-09-12
Enable race detector for unit tests

mboersma:
**What type of PR is this?**

/kind failing-test

Enables the Go [race detector](https://go.dev/doc/articles/race_detector) in unit tests, because that seems like a best practice. However, it does more than double the time:

```shell
% git co ill-race-you && time TEST_ARGS="-race -count=1" make go-test
...
TEST_ARGS="-race -count=1" make go-test  518.22s user 31.92s system 447% cpu 2:02.92 total
% git co main && time TEST_ARGS="-count=1" make go-test
...
TEST_ARGS="-count=1" make go-test  144.72s user 20.40s system 343% cpu 48.067 total
```

Maybe we could leave `-race` enabled for CI but users can disable it locally via `TEST_ARGS`? I don't want to make tests take longer, but this flag already found two data races.

N/A

> If your tests have incomplete coverage, you may find more races by running a binary built with -race under a realistic workload.

This sentence in the race detector docs is very intriguing to me. I wonder if we could build a debug binary with `-race` that we used in CI; I have a strong intuition it would find *something*.

Or maybe add `RACE_DETECTOR` as a toggle for Tilt binaries? *Edit*: see #2727.

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -634,15 +634,19 @@ promote-images: $(KPROMO) ## Promote images.
 
 ##@ Testing:
 .PHONY: test
-test: generate lint go-test ## Run "generate", "lint" and "go-tests" rules.
+test: generate lint go-test-race ## Run "generate", "lint" and "go-test-race" rules.
+
+.PHONY: go-test-race
+go-test-race: TEST_ARGS+= -race
+go-test-race: go-test ## Run go tests with the race detector enabled.
 
 .PHONY: go-test
 go-test: $(SETUP_ENVTEST) ## Run go tests.
 	KUBEBUILDER_ASSETS="$(KUBEBUILDER_ASSETS)" go test ./... $(TEST_ARGS)
 
 .PHONY: test-cover
-test-cover: $(SETUP_ENVTEST) ## Run tests with code coverage and code generate reports.
-	$(MAKE) test TEST_ARGS="$(TEST_ARGS) -coverprofile=coverage.out"
+test-cover: TEST_ARGS+= -coverprofile coverage.out
+test-cover: go-test-race ## Run tests with code coverage and generate reports.
 	go tool cover -func=coverage.out -o coverage.txt
 	go tool cover -html=coverage.out -o coverage.html
 
@@ -667,7 +671,7 @@ test-e2e-local: ## Run "docker-build" rule then run e2e tests.
 	$(MAKE) docker-build \
 	test-e2e-run
 
-CONFORMANCE_FLAVOR ?= 
+CONFORMANCE_FLAVOR ?=
 CONFORMANCE_E2E_ARGS ?= -kubetest.config-file=$(KUBETEST_CONF_PATH)
 CONFORMANCE_E2E_ARGS += $(E2E_ARGS)
 .PHONY: test-conformance
--- a/api/v1beta1/azuremachinetemplate_webhook_test.go
+++ b/api/v1beta1/azuremachinetemplate_webhook_test.go
@@ -337,7 +337,6 @@ func TestAzureMachineTemplate_ValidateUpdate(t *testing.T) {
 	for _, amt := range tests {
 		amt := amt
 		t.Run(amt.name, func(t *testing.T) {
-			t.Parallel()
 			ctx := admission.NewContextWithRequest(context.Background(), admission.Request{AdmissionRequest: admissionv1.AdmissionRequest{DryRun: pointer.Bool(true)}})
 			err := amt.template.ValidateUpdate(ctx, amt.oldTemplate, amt.template)
 			if amt.wantErr {
--- a/azure/services/inboundnatrules/inboundnatrules_test.go
+++ b/azure/services/inboundnatrules/inboundnatrules_test.go
@@ -179,7 +179,7 @@ func TestReconcileInboundNATRule(t *testing.T) {
 		tc := tc
 		t.Run(tc.name, func(t *testing.T) {
 			g := NewWithT(t)
-			t.Parallel()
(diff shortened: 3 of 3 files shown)


mboersma:
/retitle [WIP] Enable race detector for unit tests

api/v1beta1/azuremachinetemplate_webhook_test.go
```diff
@@ -332,7 +332,6 @@ func TestAzureMachineTemplate_ValidateUpdate(t *testing.T) {
 	for _, amt := range tests {
 		amt := amt
 		t.Run(amt.name, func(t *testing.T) {
-			t.Parallel()
```
mboersma, line 335: This was my (bad) advice in #2641. There are two loops which reuse the same test data, so `t.Parallel()` in the first block meant that the `dry-run=false` cases could run simultaneously with the `dry-run=true` cases.

CecileRobertMichon:
This is cool! Which part is still [WIP]?

mboersma:
> Which part is still [WIP]?

Just that I'm wary of more than doubling the time for `make go-test`. Increasing that part of the developer loop or a CI job from 3 to 8 minutes (for example) is painful.

Users can always disable race checking with `TEST_ARGS=""`. If we're ok with that option and with the default for unit tests taking longer, I think we should merge this.

mboersma:
/retitle Enable race detector for unit tests

CecileRobertMichon:
> Just that I'm wary of more than doubling the time for make go-test. Increasing that part of the developer loop or a CI job from 3 to 8 minutes (for example) is painful.

That's not great. What if we add it to the `make test` target but not `make go-test`? That way devs can still use `make go-test` locally to run tests quickly but the PR job will run this with generate and lint?

jackfrancis:
My vote is a new periodic job

mboersma:
> What if we add it to the `make test` target but not `make go-test`? 

I like this idea: we should run it in CI but not make developers wait longer. I made those changes by adding a `go-test-race` target. @jackfrancis wdyt?

CecileRobertMichon (commented):
/lgtm
/lifecycle active

jackfrancis:
@mboersma are we sure this is working? the most test run for this PR was within the expected time duration range of previous runs (withough the `-race` option)

CecileRobertMichon:
indeed

> make test TEST_ARGS=" -coverprofile=coverage.out"

Looks like test-cover is setting its own args

/lgtm cancel

mboersma:
Oh, I see we actually use `make test-cover` in CI. We can set the args for that similarly.

CecileRobertMichon:
> IDK why the e2e-exp job ran

it runs whenever we change the Makefile and other common CI files

that last run was suspiciously fast and I don't see anything in the output of the test run that indicates it ran the race detector, how do we know it's working?

mboersma:
I added temp commits to log the `go test` command line and to reintroduce the race conditions. I see the `-race` flag being passed through and indeed the unit tests failed:

```
--- FAIL: TestReconcileInboundNATRule (0.00s)
    --- FAIL: TestReconcileInboundNATRule/fail_to_create_NAT_rule (0.00s)
        testing.go:1319: race detected during execution of test
    --- FAIL: TestReconcileInboundNATRule/NAT_rule_successfully_created_with_with_no_existing_rules (0.00s)
        testing.go:1319: race detected during execution of test
    --- FAIL: TestReconcileInboundNATRule/No_LB,_Nat_rule_reconciliation_is_skipped (0.00s)
        testing.go:1319: race detected during execution of test
```

It does run relatively quickly in CI: 6 or 7 minutes. But I'll take it since it's working as intended.

mboersma:
@CecileRobertMichon @jackfrancis this is ready for review.

Makefile
```diff
@@ -634,15 +634,19 @@ promote-images: $(KPROMO) ## Promote images.
 
 ##@ Testing:
 .PHONY: test
-test: generate lint go-test ## Run "generate", "lint" and "go-tests" rules.
+test: generate lint go-test-race ## Run "generate", "lint" and "go-test-race" rules.
+
+.PHONY: go-test-race
+go-test-race: TEST_ARGS+= -race
```
mboersma, line 640: This is a [target-specific variable value](https://www.gnu.org/software/make/manual/html_node/Target_002dspecific.html), which seems like the right way to reuse the `go-test` target while adding arguments to `$(TEST_ARGS}`.

azure/services/inboundnatrules/inboundnatrules_test.go
```diff
@@ -179,7 +179,7 @@ func TestReconcileInboundNATRule(t *testing.T) {
 		tc := tc
 		t.Run(tc.name, func(t *testing.T) {
 			g := NewWithT(t)
-			t.Parallel()
+			// TODO: investigate why t.Parallel() trips the race detector here.
```
CecileRobertMichon, line 182: could you please open an issue for this one?

CecileRobertMichon: could potentially be a good-first-issue too

CecileRobertMichon (commented):
/lgtm

azure/services/inboundnatrules/inboundnatrules_test.go
jackfrancis: https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/2919

jackfrancis (approved):
/lgtm
/approve

PR #2604, opened 2022-08-25
Implement ScaleSetPriority for AzureManagedMachinePool

jackfrancis:
**What type of PR is this?**

/kind feature
/kind bug

**What this PR does / why we need it**:

This PR implements the `ScaleSetPriority` feature for AzureManagedMachinePool. See the official AKS documentation for Spot node pools here:

https://docs.microsoft.com/en-us/azure/aks/spot-node-pool

This PR ensures that we don't override (e.g., accidentally delete) any AKS-enforced labels that may make it to the user-configurable part of the the node pool labels configuration surface area.

Until this AKS issue is addressed, capz must have this adaptive update behavior:

- https://github.com/Azure/AKS/issues/3152

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #2460

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Implement ScaleSetPriority for AzureManagedMachinePool
```

PR #2657, opened 2022-09-15
standardize scope local cache

jackfrancis:
**What type of PR is this?**

/kind feature
/kind cleanup

This PR picks up where this one left off:

https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2604

Here we standardize the way we maintain a local scope cache (in other words, we cache values that are retrieved externally so that we only have to fetch them once in any given scope lifecycle [e.g., one reconciliation loop]).

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
standardize scope local cache
```

### DIFF
--- a/azure/scope/machine.go
+++ b/azure/scope/machine.go
@@ -72,6 +72,9 @@ func NewMachineScope(params MachineScopeParams) (*MachineScope, error) {
 	if params.AzureMachine == nil {
 		return nil, errors.New("azure machine is required when creating a MachineScope")
 	}
+	if params.Cache == nil {
+		params.Cache = &MachineCache{}
+	}
 
 	helper, err := patch.NewHelper(params.AzureMachine, params.Client)
 	if err != nil {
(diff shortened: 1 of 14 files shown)


azure/scope/machine.go
```diff
@@ -101,59 +104,21 @@ type MachineScope struct {
...
 type MachineCache struct {
-	BootstrapData      string
+	BootstrapData      *string
 	VMImage            *infrav1.Image
-	VMSKU              resourceskus.SKU
-	availabilitySetSKU resourceskus.SKU
-}
-
-// InitMachineCache sets cached information about the machine to be used in the scope.
-func (m *MachineScope) InitMachineCache(ctx context.Context) error {
```
jackfrancis, line 111: @devigned do you have any concerns about replacing this "do a one-time initiatlization of all cached properties when we instantiate scope" vs implement the property value caching on demand?

My thinking (and this was @CecileRobertMichon 's thinking when we implemented caching in a prior PR) is that not all scope lifecycles are going to require access to all of these properties, and so in the act of caching them we are actually *de*optimizing things by fetching data unnecessarily.

azure/scope/machine.go
```diff
@@ -168,10 +133,14 @@ func (m *MachineScope) VMSpec() azure.ResourceSpecGetter {
 		AdditionalCapabilities: m.AzureMachine.Spec.AdditionalCapabilities,
 		ProviderID:             m.ProviderID(),
 	}
-	if m.cache != nil {
-		spec.SKU = m.cache.VMSKU
-		spec.Image = m.cache.VMImage
-		spec.BootstrapData = m.cache.BootstrapData
+	if bootstrapData, err := m.GetBootstrapData(ctx); err == nil {
```
CecileRobertMichon, line 136: We should be handling the errors here and below

if I remember correctly that was one of the main reasons we decided to go with a separate `InitMachineCache` func which would be able to handle errors and pass in context without requiring those on the Spec functions.

azure/scope/machinepool.go
```diff
@@ -68,13 +69,20 @@ type (
...
+		cache            *MachinePoolCache
 	}
 
 	// NodeStatus represents the status of a Kubernetes node.
 	NodeStatus struct {
 		Ready   bool
 		Version string
 	}
+
+	// MachinePoolCache stores common machine information so we don't have to hit the API multiple times within the same reconcile loop.
```
CecileRobertMichon, line 81:
```suggestion
	// MachinePoolCache stores common machine pool information so we don't have to hit the API multiple times within the same reconcile loop.
```

jackfrancis:
/hold

Not sure we actually want this change

CecileRobertMichon:
> Not sure we actually want this change

Going to close it to remove noise for reviewers in PR queue, please reopen if/when it's ready for review 

/close

PR #2662, opened 2022-09-22
Show granular conditions during AzureCluster delete

CecileRobertMichon:
**What type of PR is this?**
/kind cleanup

**What this PR does / why we need it**: Small cosmetic improvement so `AzureClusters` show "Deleting" instead of "Deleted" as status while deleting, as well as more granular condition message (which resource deletion is in progress).

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Show granular conditions during AzureCluster delete
```

### DIFF
--- a/controllers/azurecluster_controller.go
+++ b/controllers/azurecluster_controller.go
@@ -267,10 +267,6 @@ func (acr *AzureClusterReconciler) reconcileDelete(ctx context.Context, clusterS
 	log.Info("Reconciling AzureCluster delete")
 
 	azureCluster := clusterScope.AzureCluster
-	conditions.MarkFalse(azureCluster, infrav1.NetworkInfrastructureReadyCondition, clusterv1.DeletedReason, clusterv1.ConditionSeverityInfo, "")
-	if err := clusterScope.PatchObject(ctx); err != nil {
-		return reconcile.Result{}, err
-	}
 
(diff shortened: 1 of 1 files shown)


controllers/azurecluster_controller.go
```diff
@@ -267,10 +267,6 @@ func (acr *AzureClusterReconciler) reconcileDelete(ctx context.Context, clusterS
 	log.Info("Reconciling AzureCluster delete")
 
 	azureCluster := clusterScope.AzureCluster
-	conditions.MarkFalse(azureCluster, infrav1.NetworkInfrastructureReadyCondition, clusterv1.DeletedReason, clusterv1.ConditionSeverityInfo, "")
```
CecileRobertMichon, line 270: this condition was useful before we had conditions for each resource type but now it's just overshadowing other conditions that have the same severity in the "Ready" summary and it happened to be shown first because of lexicographical order

jackfrancis:
This lgtm. Seems we're not classifying this as a kind/bug so no cherry-pick, does that sound like the best approach?

CecileRobertMichon:
> Seems we're not classifying this as a kind/bug so no cherry-pick, does that sound like the best approach?

It's not a functional bug so I'm fine with letting this release as part of the normal release cycle, just a minor UX improvement

PR #2663, opened 2022-09-23
support ccm to read config from secret

monianshouhou:
**What type of PR is this?**
/kind cleanup

Cloud-provider-azure needs to add e2e tests of cloud-config, which requires ccm to read config from secret.
If LOAD_CLOUD_CONFIG_FROM_SECRET is set, ccm will read config from secret rather than from /etc/kubernetes/azure.json.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/scripts/ci-entrypoint.sh
+++ b/scripts/ci-entrypoint.sh
@@ -141,6 +141,18 @@ wait_for_nodes() {
     "${KUBECTL}" get nodes -owide
 }
 
+copy_secret() {
+    # point at the management cluster
+    unset KUBECONFIG
+    "${KUBECTL}" get secret "${CLUSTER_NAME}-control-plane-azure-json" -o jsonpath='{.data.control-plane-azure\.json}' | base64 --decode > azure_json
+    
+    # set KUBECONFIG back to the workload cluster
+    export KUBECONFIG="${KUBECONFIG:-${PWD}/kubeconfig}"
+    "${KUBECTL}" create secret generic "${CONFIG_SECRET_NAME}" -n kube-system \
+      --from-file=cloud-config=azure_json
+    rm azure_json
+}
+
 # cleanup all resources we use
 cleanup() {
     timeout 1800 "${KUBECTL}" delete cluster "${CLUSTER_NAME}" || true
@@ -185,6 +197,17 @@ if [[ -n "${TEST_CCM:-}" ]]; then
         # "app=calico" is the label only for calico-node-windows pods
         "${KUBECTL}" wait --for=condition=Ready pod -l app=calico -n kube-system --timeout=10m
     fi
+
+    CLOUD_CONFIG="/etc/kubernetes/azure.json"
+    CONFIG_SECRET_NAME=""
+    ENABLE_DYNAMIC_RELOADING=false
+    if [[ -n "${LOAD_CLOUD_CONFIG_FROM_SECRET:-}" ]]; then
+        CLOUD_CONFIG=""
+        CONFIG_SECRET_NAME="azure-cloud-provider"
+        ENABLE_DYNAMIC_RELOADING=true
+        copy_secret
+    fi
+
     echo "Installing cloud-provider-azure components via helm"
     "${HELM}" install --repo https://raw.githubusercontent.com/kubernetes-sigs/cloud-provider-azure/master/helm/repo cloud-provider-azure --generate-name \
 --set infra.clusterName="${CLUSTER_NAME}" \
@@ -194,7 +217,11 @@ if [[ -n "${TEST_CCM:-}" ]]; then
 --set cloudNodeManager.imageName="${CNM_IMAGE_NAME}" \
 --set-string cloudControllerManager.imageTag="${IMAGE_TAG}" \
 --set-string cloudNodeManager.imageTag="${IMAGE_TAG}" \
---set cloudControllerManager.replicas="${CCM_COUNT}"
+--set cloudControllerManager.replicas="${CCM_COUNT}" \
+--set cloudControllerManager.enableDynamicReloading="${ENABLE_DYNAMIC_RELOADING}"  \
+--set cloudControllerManager.cloudConfig="${CLOUD_CONFIG}" \
+--set cloudControllerManager.cloudConfigSecretName="${CONFIG_SECRET_NAME}"
+
     echo "Waiting for all kube-system pods to be ready"
     "${KUBECTL}" wait --for=condition=Ready pod -n kube-system --all --timeout=10m
 fi


monianshouhou:
@jackfrancis PTAL. Cloud-provider-azure needs these changes to enrich e2e tests. Thanks a lot.

scripts/ci-entrypoint.sh
```diff
@@ -141,6 +141,16 @@ wait_for_nodes() {
     "${KUBECTL}" get nodes -owide
 }
 
+copy_secret() {
+    unset KUBECONFIG
+    "${KUBECTL}" get secret "${CLUSTER_NAME}-control-plane-azure-json" -o jsonpath='{.data.control-plane-azure\.json}' | base64 --decode > azure_json
+    
+    export KUBECONFIG="${KUBECONFIG:-${PWD}/kubeconfig}"
+    kubectl create secret generic "${CONFIG_SECRET_NAME}" -n kube-system \
```
CecileRobertMichon, line 149: I'm not sure I understand why we need to copy the secret since there is already an existing secret with this information, is the issue that CCM expects the key to be "cloud-config"? Is that configurable?

scripts/ci-entrypoint.sh
```diff
@@ -194,7 +215,11 @@ if [[ -n "${TEST_CCM:-}" ]]; then
 --set cloudNodeManager.imageName="${CNM_IMAGE_NAME}" \
 --set-string cloudControllerManager.imageTag="${IMAGE_TAG}" \
 --set-string cloudNodeManager.imageTag="${IMAGE_TAG}" \
---set cloudControllerManager.replicas="${CCM_COUNT}"
+--set cloudControllerManager.replicas="${CCM_COUNT}" \
+--set cloudControllerManager.enableDynamicReloading="${ENABLE_DYNAMIC_RELOADING}"  \
+--set cloudControllerManager.cloudConfig="${CLOUD_CONFIG}" \
+--set cloudControllerManager.cloudConfigSecretName="${CONFIG_SECRET_NAME}"
```
CecileRobertMichon, line 221: is this a new feature in cloud-provider? do you recommend to enable it outside e2e tests too (for end users)?

scripts/ci-entrypoint.sh
monianshouhou: Thanks for the review. Because the existing secret is in the mgmt cluster, CCM expects the secret from the worker cluster. So we need to copy secret from mgmt cluster to worker cluster.

monianshouhou: > I'm not sure I understand why we need to copy the secret since there is already an existing secret with this information, is the issue that CCM expects the key to be "cloud-config"? Is that configurable?

Thanks for the review. Because the existing secret is in the mgmt cluster, CCM expects the secret from the worker cluster. So we need to copy secret from mgmt cluster to worker cluster.

scripts/ci-entrypoint.sh
monianshouhou: > is this a new feature in cloud-provider? do you recommend to enable it outside e2e tests too (for end users)?

Actually it is an old feature https://cloud-provider-azure.sigs.k8s.io/install/configs/#dynamically-reloading-cloud-controller-manager, which is already exposed to end users. 
Currently, we need test cloud-config in cloud-provider-azure. It is much easier to change cloud config in the test code if CCM reads the config from the secret.

scripts/ci-entrypoint.sh
CecileRobertMichon, line 149:
```suggestion
    "${KUBECTL}" create secret generic "${CONFIG_SECRET_NAME}" -n kube-system \
```
to make sure we're using the hack/tools/bin kubectl binary

scripts/ci-entrypoint.sh
CecileRobertMichon: Got it. I wonder if we should just do this always since it seems better than having the secret on the VM and also makes it easier to change it as you point out. @jackfrancis thoughts?

@monianshouhou do you know of any downsides of using a k8s secret for cloud config? Maybe ACR login won't work?

scripts/ci-entrypoint.sh
```diff
@@ -141,6 +141,16 @@ wait_for_nodes() {
     "${KUBECTL}" get nodes -owide
 }
 
+copy_secret() {
+    unset KUBECONFIG
```
CecileRobertMichon, line 145: could you please add comments to explain this is to point at the management cluster and that below we're setting the kubeconfig back to the workload cluster?

CecileRobertMichon (commented):
/ok-to-test

scripts/ci-entrypoint.sh
monianshouhou: > Got it. I wonder if we should just do this always since it seems better than having the secret on the VM and also makes it easier to change it as you point out. @jackfrancis thoughts?
> 
> @monianshouhou do you know of any downsides of using a k8s secret for cloud config? Maybe ACR login won't work?

Thanks for the reply! Sorry about that I don't know any downside of using the secret. However, reading from secret is a feature proposed 3 years ago and nowadays azure.json isn't deprecated. Maybe there exist some reasons. I suggest that we could give user freedom to choose load cloud-config either from azure.json or from secret, just like in cloud-provider-azure.
@lzhecheng What do you think?

scripts/ci-entrypoint.sh
monianshouhou: > to make sure we're using the hack/tools/bin kubectl binary

Thanks for correcting it! Updated.

scripts/ci-entrypoint.sh
monianshouhou: > could you please add comments to explain this is to point at the management cluster and that below we're setting the kubeconfig back to the workload cluster?

Of course. Updated

scripts/ci-entrypoint.sh
jackfrancis: I'm pretty sure that the helm template layer uses sprig for the `hasKey` dict func:

- https://github.com/Masterminds/sprig/blob/master/dict.go#L25

This matters because the current cloud-provider-azure helm chart uses that to determine whether or not to include the `--cloud-config-secret-name` runtime configuration:

- https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/helm/cloud-provider-azure/templates/cloud-provider-azure.yaml#L199-L201

So, this change to capz's `ci-entrypoint.sh` will mean that we'll be invoking the `cloud-controller-manager` runtime in the following way when not using the secret:

```
cloud-controller-manager --cloud-config="/etc/kubernetes/azure.json" --cloud-config-secret-name=""...
```

compared to the way we do it now:

```
cloud-controller-manager --cloud-config="/etc/kubernetes/azure.json"...
```

Is this new change a supported, equivalent configuration for `cloud-controller-manager`? In other words, is it equivalent to pass in `--cloud-config-secret-name=""` (empty string) vs not passing in `--cloud-config-secret-name` configuration at all (this latter pattern is what we do right now)?

monianshouhou: > I'm pretty sure that the helm template layer uses sprig for the `hasKey` dict func:
> 
> * https://github.com/Masterminds/sprig/blob/master/dict.go#L25
> 
> This matters because the current cloud-provider-azure helm chart uses that to determine whether or not to include the `--cloud-config-secret-name` runtime configuration:
> 
> * https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/helm/cloud-provider-azure/templates/cloud-provider-azure.yaml#L199-L201
> 
> So, this change to capz's `ci-entrypoint.sh` will mean that we'll be invoking the `cloud-controller-manager` runtime in the following way when not using the secret:
> 
> ```
> cloud-controller-manager --cloud-config="/etc/kubernetes/azure.json" --cloud-config-secret-name=""...
> ```
> 
> compared to the way we do it now:
> 
> ```
> cloud-controller-manager --cloud-config="/etc/kubernetes/azure.json"...
> ```
> 
> Is this new change a supported, equivalent configuration for `cloud-controller-manager`? In other words, is it equivalent to pass in `--cloud-config-secret-name=""` (empty string) vs not passing in `--cloud-config-secret-name` configuration at all (this latter pattern is what we do right now)?

Thanks for the review! Yes, because if `cloud-config` is set, `cloud-config-secret-name` will be ignored.

jackfrancis (approved):
/lgtm
/approve

PR #2664, opened 2022-09-23
Add nodePublicIPPrefixID to AzureManagedMachinePool

### LINKED ISSUE #2266: Add support for public ips features for nodepools in AKS
/kind feature

AKS allows creating nodepools, where each node has a public ExternalIP address. AKS also support using a fixed IP pool of public IPs for a nodepool. 

CAPZ doesn't support currently those features. We can add support in `AzureManagedMachinePool` API to enable both of those features.

**Environment:**

nojnhuh:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**: Adds the `nodePublicIPPrefixID` field from the AKS API to enable AKS node pools to use public IPs from a pre-reserved block.

Fixes #2266

I noticed some small issues while testing failure scenarios like creating machine pools with more nodes than available IP addresses. I have a draft of the fixes for those in #2665.

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
AKS: Add nodePublicIPPrefixID
```


exp/api/v1beta1/azuremanagedmachinepool_webhook.go
```diff
@@ -218,6 +211,12 @@ func (m *AzureManagedMachinePool) ValidateUpdate(oldRaw runtime.Object, client c
 		m.Spec.EnableNodePublicIP); err != nil {
 		allErrs = append(allErrs, err)
 	}
+	if err := validateStringPtrImmutable(
```
CecileRobertMichon, line 214: I like how you made future rebasing easier by preemptively adopting #2718 :)

test/e2e/aks.go
```diff
@@ -348,3 +354,119 @@ func GetLatestStableAKSKubernetesVersion(ctx context.Context, subscriptionID, lo
...
 	return maxVersion, nil
 }
+
+type AKSPublicIPPrefixSpecInput struct {
+	Cluster           *clusterv1.Cluster
+	KubernetesVersion string
+	WaitIntervals     []interface{}
+}
+
+func AKSPublicIPPrefixSpec(ctx context.Context, inputGetter func() AKSPublicIPPrefixSpecInput) {
```
CecileRobertMichon, line 364: Great work on adding e2e testing here 💯 

I love how you added a reusable spec that reuses the existing test AKS cluster to add a new MachinePool with the specific feature you want to test, instead of creating a whole new cluster. 

Would you be willing to show/demo this concept at CAPZ office hours? I think we should consider adopting this concept more widely in our e2e tests.

CecileRobertMichon (commented):
Great work here @nojnhuh 

lgtm

exp/api/v1beta1/azuremanagedmachinepool_webhook.go
```diff
@@ -363,3 +383,21 @@ func validateBoolPtrImmutable(path *field.Path, oldVal, newVal *bool) *field.Err
 
 	return nil
 }
+
+func validateStringPtrImmutable(path *field.Path, oldVal, newVal *string) *field.Error {
```
mtougeron, line 387: Does this perhaps belong as something like `util/strings/strings.go` util package like we do for maps & slices so that it can be reused by other validation webhooks?

mtougeron (commented):
Just a Q about how `validateStringPtrImmutable` can be used by other webhooks. I don't know enough about the e2e tests to comment on that part of the code but otherwise it seems good to me.

test/e2e/aks.go
nojnhuh: Thanks! I could definitely share this at office hours. @willie-yao has done a ton with this pattern to test VM extensions, so I could tag team with him if he wants.

exp/api/v1beta1/azuremanagedmachinepool_webhook.go
nojnhuh: That sounds like a good idea to me. @jackfrancis do you have any opinions on how that might relate to #2718?

jackfrancis: +1, my vote would be to fold the effort of generalizing this foo into a common library into #2718 so we can land this change.

Also we want to have func-specific UT at some point, I can take that on in #2718 as well.

jackfrancis (approved):
/lgtm
/approve

PR #2479, opened 2022-07-13
Make agent pools reconcile/delete async

Jont828:
**What type of PR is this?**

/kind feature
**What this PR does / why we need it**: Implementation of an async service for managed agent pools as part of an effort to make all services async. See #1610 and #1541.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #2477

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Make agent pools reconcile/delete async
```

PR #2665, opened 2022-09-23
fix irrecoverable errors in async operations

nojnhuh:
**What type of PR is this?**
/kind bug

Reconciliations involving asynchronous operations can fall into a loop
where an eventual "Failed" result can block future reconciliations from
making any further changes to that particular resource to fix the
problem.

This change sets `longRunningOperationStates` for agent pools on the
corresponding AzureManagedMachinePool instead of the
AzureManagedControlPlane, since changes to that resource were not being
persisted. It also only blocks starting new operations on the status of
existing operations of the same type. In-progress PUT operations will no
longer block new DELETEs and in-progress DELETEs will not block
in-progress PUTs.

In cases where polling a future from the Azure API would eventually
return both `isDone==true` and a non-nil error, a "failed checking if
the operation was complete" message would be logged and the error would
refer to the ultimate failure unrelated to polling the future itself.
This change treats all `isDone==true` polling checks as successful and
relies on the operation's error to be captured in the future's `Result`.
The future will always be deleted from the status when the operation is
done so it can be retried the next reconciliation if it failed.

Fixes #

I found these issues while working on #2664 and testing failure scenarios like creating an AKS cluster with more nodes than available public IP addresses. Incorporating that feature here will exercise these changes.

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/azure/scope/managedmachinepool.go
+++ b/azure/scope/managedmachinepool.go
@@ -221,20 +221,20 @@ func (s *ManagedMachinePoolScope) SetAgentPoolReady(ready bool) {
 	s.InfraMachinePool.Status.Ready = ready
 }
 
-// SetLongRunningOperationState will set the future on the AzureManagedControlPlane status to allow the resource to continue
+// SetLongRunningOperationState will set the future on the AzureManagedMachinePool status to allow the resource to continue
 // in the next reconciliation.
 func (s *ManagedMachinePoolScope) SetLongRunningOperationState(future *infrav1.Future) {
-	futures.Set(s.ControlPlane, future)
+	futures.Set(s.InfraMachinePool, future)
 }
 
-// GetLongRunningOperationState will get the future on the AzureManagedControlPlane status.
-func (s *ManagedMachinePoolScope) GetLongRunningOperationState(name, service string) *infrav1.Future {
-	return futures.Get(s.ControlPlane, name, service)
+// GetLongRunningOperationState will get the future on the AzureManagedMachinePool status.
+func (s *ManagedMachinePoolScope) GetLongRunningOperationState(name, service, futureType string) *infrav1.Future {
+	return futures.Get(s.InfraMachinePool, name, service, futureType)
 }
 
-// DeleteLongRunningOperationState will delete the future from the AzureManagedControlPlane status.
-func (s *ManagedMachinePoolScope) DeleteLongRunningOperationState(name, service string) {
-	futures.Delete(s.ControlPlane, name, service)
+// DeleteLongRunningOperationState will delete the future from the AzureManagedMachinePool status.
+func (s *ManagedMachinePoolScope) DeleteLongRunningOperationState(name, service, futureType string) {
+	futures.Delete(s.InfraMachinePool, name, service, futureType)
 }
 
 // UpdateDeleteStatus updates a condition on the AzureManagedControlPlane status after a DELETE operation.
--- a/azure/services/async/async.go
+++ b/azure/services/async/async.go
@@ -47,11 +47,11 @@ func New(scope FutureScope, createClient Creator, deleteClient Deleter) *Service
 
 // processOngoingOperation is a helper function that will process an ongoing operation to check if it is done.
 // If it is not done, it will return a transient error.
-func processOngoingOperation(ctx context.Context, scope FutureScope, client FutureHandler, resourceName string, serviceName string) (result interface{}, err error) {
+func processOngoingOperation(ctx context.Context, scope FutureScope, client FutureHandler, resourceName string, serviceName string, futureType string) (result interface{}, err error) {
 	ctx, log, done := tele.StartSpanWithLogger(ctx, "async.Service.processOngoingOperation")
 	defer done()
 
-	future := scope.GetLongRunningOperationState(resourceName, serviceName)
+	future := scope.GetLongRunningOperationState(resourceName, serviceName, futureType)
 	if future == nil {
 		log.V(2).Info("no long running operation found", "service", serviceName, "resource", resourceName)
 		return nil, nil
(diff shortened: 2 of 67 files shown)


azure/services/managedclusters/client.go
```diff
@@ -156,12 +156,7 @@ func (ac *azureClient) IsDone(ctx context.Context, future azureautorest.FutureAP
 	ctx, _, done := tele.StartSpanWithLogger(ctx, "managedclusters.azureClient.IsDone")
 	defer done()
 
-	isDone, err := future.DoneWithContext(ctx, ac.managedclusters)
-	if err != nil {
-		return false, errors.Wrap(err, "failed checking if the operation was complete")
-	}
-
-	return isDone, nil
+	return future.DoneWithContext(ctx, ac.managedclusters)
```
nojnhuh, line 159: I imagine this is a change we'd make to all the service clients at once. I'll only include the change to managedclusters for now and once we decide how these should look then I'll make any necessary changes to the others.

azure/scope/managedmachinepool.go
```diff
@@ -281,6 +288,14 @@ func (s *ManagedMachinePoolScope) PatchCAPIMachinePoolObject(ctx context.Context
 	)
 }
 
+// PatchControlPlaneObject persists the AzureManagedControlPlane configuration and status.
+func (s *ManagedMachinePoolScope) PatchControlPlaneObject(ctx context.Context) error {
```
CecileRobertMichon, line 292: I was wondering why you had to do this and then I saw https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2665/files#diff-5607bc7afbe8f18686e0e62796b8da6371e999006ddb598c5034dd7ef7040a47R234... I think the bug is actually that we're setting/getting the future on the wrong object above no? Probably a bad copy paste :(

azure/services/async/async.go
```diff
@@ -85,7 +94,7 @@ func processOngoingOperation(ctx context.Context, scope FutureScope, client Futu
 		// This can happen if the resource was deleted by another process before we could get the result.
 		scope.DeleteLongRunningOperationState(resourceName, serviceName)
 	}
-	return result, err
+	return result, resultErr{err}
```
CecileRobertMichon, line 97: instead of having to differentiate the error here, it might be easier to call `DeleteLongRunningOperationState` in all cases where done is true (regardless of the error return) that way we'll attempt a new operation on the next loop if the current one finished and won't keep the erroneous future in Status.

azure/services/async/async.go
```diff
@@ -149,9 +162,11 @@ func (s *Service) DeleteResource(ctx context.Context, spec azure.ResourceSpecGet
 
 	// Check if there is an ongoing long running operation.
 	future := s.Scope.GetLongRunningOperationState(resourceName, serviceName)
-	if future != nil {
-		_, err := processOngoingOperation(ctx, s.Scope, s.Deleter, resourceName, serviceName)
-		return err
+	// PUT operations still in progress should not be waited on if a delete was requested.
+	if future != nil && future.Type == infrav1.DeleteFuture {
```
CecileRobertMichon, line 166: we also want the reverse to be true (don't block a PUT on a delete operation in progress). What do you think of modifying `GetLongRunningOperationState` to only return futures for the right operation type so this applies to all operations?

nojnhuh: That makes sense, I'll add that.

I noticed there are a couple of places this is called outside of this file for scalesets:

Since it looks like this is looking for a future of any type, is it worth baking in some capability to search for multiple types? Or maybe just do that inline here?
https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/d28ff9a7d44b25fe10ca3e13a7e2725b50b35fad/azure/services/scalesets/scalesets.go#L94

The second one is in a `Delete` method, so I'm guessing we'd only want delete futures here too?
https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/d28ff9a7d44b25fe10ca3e13a7e2725b50b35fad/azure/services/scalesets/scalesets.go#L185

CecileRobertMichon: The first use (L94) is in Reconcile so that's either a PUT or PATCH future, and the second use (L185) should be just DELETE futures.

Scalesets still needs to be refactored as part of #1702, after that it will also be using async.go, so I think we can just do it inline for now and check both.

nojnhuh:
Testing this a bit more, I see some less-than-ideal behavior when I manage to create an agent pool in a "Failed" state. When the create operation finishes, the appropriate error is logged and set on the AzureManagedMachinePool status as expected. Then in the next reconciliation, the resource is found to exist with the same parameters which is assumed to be successful, so the error status gets cleared out and everything indicates success even though the agent pool is still in a failed state. The only sign something is wrong is a transient error log saying the underlying VMSS doesn't exist.

That seems like it might be a separate issue from what these changes attempt to address, so I'll leave that as a gap here for now and tackle that separately.

azure/services/managedclusters/client.go
nojnhuh: Updated the rest.

CecileRobertMichon:
> Testing this a bit more, I see some less-than-ideal behavior when I manage to create an agent pool in a "Failed" state

We might need to handle agent pools the same way we're handling VMs: in addition to the operation result which might be "failed" if it failed to create all together, in which case we do want to retry the operation the current way, there is also a case of the resource itself being in "failed" state, which requires us to inspect the provisioning state of the resource on every reconcile and mark the object as not ready if the resource is not healthy https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/services/virtualmachines/virtualmachines.go#L113

CecileRobertMichon:
@nojnhuh also tangent for managed cluster resource health: https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/2673

azure/services/async/async.go
```diff
@@ -66,26 +66,31 @@ func processOngoingOperation(ctx context.Context, scope FutureScope, client Futu
...
 	}
+	if err != nil {
+		log.V(2).Error(err, "error checking long running operation status after it finished")
+	}
+
+	// Once the operation is done, we can delete the long running operation state.
+	// If the operation failed, this will allow it to be retried during the next reconciliation.
+	// If the resource is not found, we also reset the long-running operation state so we can attempt to create it again.
+	// This can happen if the resource was deleted by another process before we could get the result.
+	scope.DeleteLongRunningOperationState(resourceName, serviceName)
```
CecileRobertMichon, line 89: Should we also add futureType to delete to make sure that we are differentiating between delete and put futures and not blindly deleting the first one we find?

CecileRobertMichon (commented):
lgtm pending comment on DeleteLongRunningOperationState

/cc @jackfrancis

jackfrancis:
cherry-pick release-1.4
cherry-pick release-1.5

azure/services/async/async.go
nojnhuh: Yeah that makes sense. I've made that change.

CecileRobertMichon (commented):
/lgtm
/assign @jackfrancis 

Great work @nojnhuh 🙌

jackfrancis (approved):
/approve

in addition to addressing pathologies, this is a nice collection of clearer, simplified flows

jackfrancis:
@CecileRobertMichon @nojnhuh We can't cherry-pick this into `release-1.4` because the async refactor of agentpools was not cherry-picked back. See:

https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2479

I think this makes sense, as the refactor work performed in #2479 isn't a candidate for a patch release cherry-pick.

We could cherry-pick the changes into release-1.4 only for those services that are using the `async` pattern, but that might be more frankenstein than we want for a patch release two versions ago. Wdyt?

PR #2688, opened 2022-09-30
detect Retry-After during async “does resource exist?” flow

jackfrancis:
**What type of PR is this?**

/kind feature

This PR adds a single additional vector of `Retry-After` HTTP response detection during async resource reconciliation. When we enter into the `CreateResource` async service method (which is in fact a "create or update resource" flow) then we initiate a GET against the Azure API to determine if the resource already exists. Because this GET may yield a HTTP 429 response (or any other response w/ `Retry-After` header data) we want to check that, to ensure that we inform the parent controller that we should requeue the request for later at a time after the specified `Retry-After` value from Azure.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
detect Retry-After during async “does resource exist?” flow
```

### DIFF
--- a/azure/services/async/async.go
+++ b/azure/services/async/async.go
@@ -18,8 +18,11 @@ package async
 
 import (
 	"context"
+	"net/http"
+	"strconv"
 	"time"
 
+	"github.com/Azure/go-autorest/autorest"
 	azureautorest "github.com/Azure/go-autorest/autorest/azure"
 	"github.com/pkg/errors"
 	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
@@ -76,7 +79,7 @@ func processOngoingOperation(ctx context.Context, scope FutureScope, client Futu
 
 		// Operation is still in progress, update conditions and requeue.
 		log.V(2).Info("long running operation is still ongoing", "service", serviceName, "resource", resourceName)
-		return nil, azure.WithTransientError(azure.NewOperationNotDoneError(future), retryAfter(sdkFuture))
+		return nil, azure.WithTransientError(azure.NewOperationNotDoneError(future), getRequeueAfterFromFuture(sdkFuture))
 	}
 	if err != nil {
 		log.V(2).Error(err, "error checking long running operation status after it finished")
@@ -111,7 +114,8 @@ func (s *Service) CreateResource(ctx context.Context, spec azure.ResourceSpecGet
 	// Get the resource if it already exists, and use it to construct the desired resource parameters.
 	var existingResource interface{}
 	if existing, err := s.Creator.Get(ctx, spec); err != nil && !azure.ResourceNotFound(err) {
-		return nil, errors.Wrapf(err, "failed to get existing resource %s/%s (service: %s)", rgName, resourceName, serviceName)
+		errWrapped := errors.Wrapf(err, "failed to get existing resource %s/%s (service: %s)", rgName, resourceName, serviceName)
+		return nil, azure.WithTransientError(errWrapped, getRetryAfterFromError(err))
 	} else if err == nil {
 		existingResource = existing
 		log.V(2).Info("successfully got existing resource", "service", serviceName, "resource", resourceName, "resourceGroup", rgName)
@@ -136,7 +140,7 @@ func (s *Service) CreateResource(ctx context.Context, spec azure.ResourceSpecGet
 			return nil, errors.Wrapf(err, "failed to create resource %s/%s (service: %s)", rgName, resourceName, serviceName)
 		}
 		s.Scope.SetLongRunningOperationState(future)
-		return nil, azure.WithTransientError(azure.NewOperationNotDoneError(future), retryAfter(sdkFuture))
+		return nil, azure.WithTransientError(azure.NewOperationNotDoneError(future), getRequeueAfterFromFuture(sdkFuture))
 	} else if err != nil {
 		return nil, errors.Wrapf(err, "failed to create resource %s/%s (service: %s)", rgName, resourceName, serviceName)
 	}
@@ -170,7 +174,7 @@ func (s *Service) DeleteResource(ctx context.Context, spec azure.ResourceSpecGet
 			return errors.Wrapf(err, "failed to delete resource %s/%s (service: %s)", rgName, resourceName, serviceName)
 		}
 		s.Scope.SetLongRunningOperationState(future)
-		return azure.WithTransientError(azure.NewOperationNotDoneError(future), retryAfter(sdkFuture))
+		return azure.WithTransientError(azure.NewOperationNotDoneError(future), getRequeueAfterFromFuture(sdkFuture))
 	} else if err != nil {
 		if azure.ResourceNotFound(err) {
 			// already deleted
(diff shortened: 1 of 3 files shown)


CecileRobertMichon:
> CreateResource async service method (which is in fact a "create or update resource" flow)

Feel free to rename it (here if you don't plan or backporting, or another PR if this one is meant to be backported), I honestly can't remember why it's called "CreateResource" (either `ReconcileResource` or `CreateOrUpdateResource` probably makes more sense)

azure/services/async/async.go
```diff
@@ -183,12 +194,27 @@ func (s *Service) DeleteResource(ctx context.Context, spec azure.ResourceSpecGet
...
+
+// getRetryAfterFromDetailedError returns the time.Duration from the http.Response in the autorest.DetailedError
+// If there is no Response object, or if there is no meaningful Retry-After header data, we return a zero value time.Duration
+func getRetryAfterFromDetailedError(detailedError autorest.DetailedError) (ret time.Duration) {
+	if detailedError.Response != nil {
+		if retryAfter := detailedError.Response.Header.Get("Retry-After"); retryAfter != "" {
+			if rai, _ := strconv.Atoi(retryAfter); rai > 0 {
+				ret = time.Duration(rai) * time.Second
+			}
+		} else if detailedError.Response.StatusCode == http.StatusTooManyRequests {
```
CecileRobertMichon, line 215: do we expect that `StatusTooManyRequests` error and `Retry-After` header being set are 2 mutually exclusive events?

jackfrancis:
> > CreateResource async service method (which is in fact a "create or update resource" flow)
> 
> Feel free to rename it (here if you don't plan or backporting, or another PR if this one is meant to be backported), I honestly can't remember why it's called "CreateResource" (either `ReconcileResource` or `CreateOrUpdateResource` probably makes more sense)

I tagged this as a feature, so I don't plan to backport it. I think I'd prefer to do a rename in a separate PR. That way we can spend some time w/ the community getting feedback on a better name as distinct from landing this change (which should be reviewed on its own merits).

azure/services/async/async.go
```diff
@@ -183,12 +194,27 @@ func (s *Service) DeleteResource(ctx context.Context, spec azure.ResourceSpecGet
...
+			ret = 1 * time.Minute
+		}
+	}
+	return
```
CecileRobertMichon, line 219: could we return the Default requeue time instead of 0 if no retry after is found? Like above https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2688/files#diff-0e657fbf13cf152e97cf8871a3baf550199d64f99f94316e7e1b9eeb5d6cc8e4R202

jackfrancis: So the idea here would be that if we receive a strongly typed `autorest.DetailedError` error, then we should always classify that as a transient error w/ default requeue?

azure/services/async/async.go
```diff
@@ -183,12 +197,27 @@ func (s *Service) DeleteResource(ctx context.Context, spec azure.ResourceSpecGet
...
 	return retryAfter
 }
+
+// getRetryAfterFromDetailedError returns the time.Duration from the http.Response in the autorest.DetailedError.
+// If there is no Response object, or if there is no meaningful Retry-After header data, we return a zero value time.Duration.
+func getRetryAfterFromDetailedError(detailedError autorest.DetailedError) (ret time.Duration) {
+	if detailedError.Response != nil {
+		if retryAfter := detailedError.Response.Header.Get("Retry-After"); retryAfter != "" {
+			if rai, _ := strconv.Atoi(retryAfter); rai > 0 {
+				ret = time.Duration(rai) * time.Second
```
CecileRobertMichon, line 216: why are we doing * time.Second here ?

azure/services/async/async.go
CecileRobertMichon: Makes sense. Could we add code comments? It wasn't completely clear to me without reading your explanation.

jackfrancis: Good point, done!

azure/services/async/async.go
jackfrancis: See my latest changes, which account for both flavors of `Retry-After` data. Ref:

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After#examples

azure/services/async/async.go
jackfrancis: I've moved all of this into the `getRetryAfterFromError` func, and now the GET in the flow of `CreateResource` *always* returns a transient `azure.ReconcileError`.

Is that correct? That's a change from the present, where non-404 errors are always returned as vanilla errors (`err != nil && !azure.ResourceNotFound(err)`. Is there a chance that changing this to transient `azure.ReconcileError` will result in a mis-classification of non-transient errors?

azure/services/async/async.go
```diff
@@ -29,6 +32,9 @@ import (
 	"sigs.k8s.io/cluster-api-provider-azure/util/tele"
 )
 
+// defaultHTTP429RetryAfter is a default backoff wait time when we get a HTTP 429 response with no Retry-After data.
+const defaultHTTP429RetryAfter = 1 * time.Minute
```
CecileRobertMichon, line 36: can we move this const to `reconciler/defaults.go` where the other requeue consts are defined?

azure/services/async/async.go
```diff
@@ -183,12 +190,40 @@ func (s *Service) DeleteResource(ctx context.Context, spec azure.ResourceSpecGet
...
 	retryAfter, _ := sdkFuture.GetPollingDelay()
 	if retryAfter < reconciler.DefaultReconcilerRequeue {
 		retryAfter = reconciler.DefaultReconcilerRequeue
 	}
 	return retryAfter
 }
+
+// getRetryAfterFromError returns the time.Duration from the http.Response in the autorest.DetailedError.
+// If there is no Response object, or if there is no meaningful Retry-After header data, we return a default.
+func getRetryAfterFromError(err error) (ret time.Duration) {
```
CecileRobertMichon, line 205: nit: can we be consistent with the sibling func above for named return parameters, either use named return above too or not use a named return here

azure/services/async/async.go
```diff
@@ -183,12 +190,40 @@ func (s *Service) DeleteResource(ctx context.Context, spec azure.ResourceSpecGet
...
+	// In case we aren't able to introspect Retry-After from the error type, we'll return this default
+	ret = reconciler.DefaultReconcilerRequeue
+	var detailedError autorest.DetailedError
+	// if we have a strongly typed autorest.DetailedError then we can introspect the HTTP response data
+	if errors.As(err, &detailedError) {
+		if detailedError.Response != nil {
+			// If we have Retry-After HTTP header data for any reason, prefer it
+			if retryAfter := detailedError.Response.Header.Get("Retry-After"); retryAfter != "" {
+				// This handles the case where Retry-After data is in the form of units of seconds
```
CecileRobertMichon, line 214: are we expecting both scenarios to be possible (units of seconds and absolute time)? Same question with `autorest.DetailedError` vs not ? Are there any SDK contracts for those error responses / headers that we can follow? Seems strange that we have to handle multiple possibilities, it's like we're guessing what the reponse might look like instead of expecting a specific format/unit/type.

azure/services/async/async.go
```diff
@@ -183,12 +190,40 @@ func (s *Service) DeleteResource(ctx context.Context, spec azure.ResourceSpecGet
...
+				if rai, err := strconv.Atoi(retryAfter); err == nil {
+					ret = time.Duration(rai) * time.Second
+					// This handles the case where Retry-After data is in the form of absolute time
+				} else if t, err := time.Parse(time.RFC1123, retryAfter); err == nil {
+					ret = time.Until(t)
+				}
+				// If we didn't find Retry-After HTTP header data but the response type is 429,
+				// we'll have to come up with our sane default.
+			} else if detailedError.Response.StatusCode == http.StatusTooManyRequests {
```
CecileRobertMichon, line 223: same here, is that a real scenario or are we doing that just in case?

azure/services/async/async.go
jackfrancis: done

azure/services/async/async.go
jackfrancis: done, dropped the named return var in the new func

azure/services/async/async.go
jackfrancis: > are we expecting both scenarios to be possible (units of seconds and absolute time)

As far as I can tell from research, the spec is overloaded to expect both value type flavors, so I think it's the best practice to deal with both types wherever we parse `Retry-After` HTTP header data.

> Same question with autorest.DetailedError vs not

The `autorest.DetailedError` error implementation is one that we definitely know about from the usage above in the async flow, based on the specific SDK API we're currently re-using. In that sense you could say that this helper function is sort of tightly coupled to the particular implementation of capz at this point in time (initially this foo was inline in the `CreateResource` func but I split it out for code maintenance reasons). So tl;dr we only have access to `Retry-After` data in this particular case via the `err` response from the underlying service `Get` implementations, and we only know how to get at it (as of right now) if the err is of "type" `autorest.DetailedError`. Hope that makes sense!

azure/services/async/async.go
jackfrancis: This is definitely a real scenario, as a `autorest.DetailedError` error has a strongly typed HTTP response object in it, which gives us access to the HTTP response status code.

The larger idea here is that not all API responses are well behaving, and if we get an HTTP 429 without `Retry-After` data we can at least trust that the HTTP 429 indicates we're sending too many requests, which is (IMO) sufficient justification for telling the controller to requeue the next request a little later in time than it would otherwise.

azure/defaults.go
```diff
@@ -100,6 +101,11 @@ const (
 	CustomHeaderPrefix = "infrastructure.cluster.x-k8s.io/custom-header-"
 )
 
+const (
+	// DefaultHTTP429RetryAfter is a default backoff wait time when we get a HTTP 429 response with no Retry-After data.
+	DefaultHTTP429RetryAfter = 1 * time.Minute
```
CecileRobertMichon, line 106: `util/reconcile/defaults.go` is where I was pointing to, sorry should have been explicit!

jackfrancis: done

jackfrancis:
Yay here's a PR where UT coverage data is gathered and displayed as part of the single "-test" job (no extra "-coverage" job).

cc @mboersma @CecileRobertMichon (thanks again @fabriziopandini)

PR #2689, opened 2022-09-30
Add evictionPolicy field for spot VMs

### LINKED ISSUE #2570: Allow configuration of Spot instance Eviction Policy
> ⚠️ Cluster API Azure maintainers can ask to turn an issue-proposal into a [CAEP](https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/proposals/YYYYMMDD-template.md) when necessary. This is to be expected for large changes that impact multiple components, breaking changes, or new large features.

***Goals***
1. Enable configuration of spot instance eviction policy 
2. Enable use of spot instances with ephemeral disk

***Non-Goals/Future Work***
1. Non-Goal 1
1. Non-Goal 2

**User Story**

As an operator I would like to be able to configure the eviction policy (Deallocate or Delete) of my spot instances for being able to use spot instances together with ephemeral disks which requires eviction policy to be Delete.

**Detailed Description**

Extend the `AzureMachineTemplate` resource with the property `spec.template.spec.spotVMOptions.evictionPolicy` that allows the values `Deallocate` and `Delete` and has a default value of `Deallocate`.

Example:

```yaml
kind: AzureMachineTemplate
metadata:
  name: spot
spec:
  template:
    spec:
      vmSize: Standard_DS3_v2
      acceleratedNetworking: true
      enableIPForwarding: true
      osDisk:
        osType: linux
        diskSizeGB: 128
        cachingType: ReadOnly
        diffDiskSettings:
          option: Local
      spotVMOptions:
        maxPrice: -1
        evictionPolicy: Delete
      sshPublicKey: <base64>
```

[Describe contract changes between Cluster API Azure Provider controllers, if applicable.]

Changes the model of [AzureMachine](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/api/v1beta1/azuremachine_types.go) with an additional property with a default value.

/kind proposal

willie-yao:
**What type of PR is this?**
/kind feature

This PR adds a field `evictionPolicy` to `spotVMOptions`. This can either be set to Delete or Deallocate.

Fixes #2570

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Add evictionPolicy field for spot VMs
```


api/v1beta1/azuremachine_types.go
```diff
@@ -133,6 +134,10 @@ type SpotVMOptions struct {
 	// MaxPrice defines the maximum price the user is willing to pay for Spot VM instances
 	// +optional
 	MaxPrice *resource.Quantity `json:"maxPrice,omitempty"`
+
+	// EvictionPolicy defines the behavior of the virtual machine when it is evicted. It can be either Delete or Deallocate.
+	// +optional
+	EvictionPolicy *compute.VirtualMachineEvictionPolicyTypes `json:"evictionPolicy,omitempty"`
```
CecileRobertMichon, line 140: we want to avoid adding a dependency on the Azure SDK in the api package. This is so that the SDK can we swapped out easily for another (for example this might be applicable to support other clouds besides Azure public cloud). Here instead of using `compute.VirtualMachineEvictionPolicyTypes` you can define a new type

CecileRobertMichon (commented):
/lgtm

jackfrancis:
FYI we actually have to update the service code flow so this new feature actually does something. Search for `SpotVMOptions` under `azure/scope/*` and `/azure/services/*` to see how this is done for the existing spot priority config for vm/vmss.

Also we want to update the docs in `docs/book/src/topics/spot-vms.md`.

willie-yao:
I tested this option through custom templates on Tilt and it was able to be configured. Does the feature not work for you?

docs/book/src/topics/spot-vms.md
```diff
@@ -48,6 +48,18 @@ spec:
       maxPrice: 0.04 # Price in USD per hour (up to 5 decimal places)
 ```
 
+In addition, you are able to explicitly set the eviction policy for the Spot VM.
+The default policy is `Delete` which will delete the VM when it is evicted. You
```
CecileRobertMichon, line 52: should that say "The default policy is `Deallocate`"? https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2689/files#diff-2128b31583d63daf5c4f86f34b7f90859ec634490812674658d179694a88bf2bR44

azure/converters/spotinstances.go
```diff
@@ -40,8 +40,12 @@ func GetSpotVMOptions(spotVMOptions *infrav1.SpotVMOptions, diffDiskSettings *in
 			MaxPrice: &maxPrice,
 		}
 	}
+
+	// Set the spot vm eviction policy if provided, otherwise default to deallocate for regular disks and delete for ephemeral disks
 	evictionPolicy := compute.VirtualMachineEvictionPolicyTypesDeallocate
```
CecileRobertMichon, line 45: rethinking this, it would be better to explicitly default the field in the defaulting webhook so that the value is persisted in the Spec, even if the user didn't explicitly set it. That way, if we decide to change the default later, we won't break existing users.

CecileRobertMichon: Webhook defaulting logic for AzureMachines: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/api/v1beta1/azuremachine_default.go
Webhook defaulting logic for AzureMachinePools: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/exp/api/v1beta1/azuremachinepool_default.go

More info on webhooks: https://cluster-api.sigs.k8s.io/developer/providers/webhooks.html

CecileRobertMichon:
> FYI we actually have to update the service code flow so this new feature actually does something

I had the same initial reaction when reviewing this, but in this case it's actually not necessary because the whole SpotVMOptions struct is being passed through to the service (whether we should be doing this is a separate discussion...) so it's already getting the new spot VM option without any additional changes needed to the spec: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/scope/machine.go#L165

azure/converters/spotinstances.go
willie-yao: If we define the default in the webhook, does that mean we won't have to do any checks inside the spotinstances converter, and just return the eviction policy from the spotVMOptions parameter directly? I'm assuming that by the time the converter is called when building the VM from spec, the webhook will set the default spotVMOptions if it isn't defined in the spec.

willie-yao: I'm also wondering if we should still default to delete if an ephemeral disk is being used. If the spec is already defined by the webhook by default, I'm not sure how to check if that's the case before setting it to delete for an ephemeral disk.

CecileRobertMichon: > If we define the default in the webhook, does that mean we won't have to do any checks inside the spotinstances converter, and just return the eviction policy from the spotVMOptions parameter directly

correct

> I'm also wondering if we should still default to delete if an ephemeral disk is being used

you can make the webhook default a "smart" default where it defaults to delete if ephemeral disk is being used

willie-yao: @CecileRobertMichon I was able to make changes to the webhooks to set the default spot eviction policy. However, many of the unit tests started failing because I don't think the defaulting webhook is run for them. Is there a way to run the webhook in a unit test environment?

azure/converters/spotinstances.go
```diff
@@ -40,9 +40,9 @@ func GetSpotVMOptions(spotVMOptions *infrav1.SpotVMOptions, diffDiskSettings *in
 			MaxPrice: &maxPrice,
 		}
 	}
-	evictionPolicy := compute.VirtualMachineEvictionPolicyTypesDeallocate
-	if diffDiskSettings != nil && diffDiskSettings.Option == "Local" {
-		evictionPolicy = compute.VirtualMachineEvictionPolicyTypesDelete
-	}
+
+	// Set the spot vm eviction policy if provided.
+	evictionPolicy := compute.VirtualMachineEvictionPolicyTypes(*spotVMOptions.EvictionPolicy)
```
CecileRobertMichon, line 45: we want to make sure we guarding against any potential nil pointer dereferencing here (even though we don't expect this to ever be nil in practice)

```
// Set the spot vm eviction policy if provided.
if spotVMOptions != nil {
    evictionPolicy := compute.VirtualMachineEvictionPolicyTypes(*spotVMOptions.EvictionPolicy)
}
```


This will also fix the unit test failures you're seeing along the lines of `panic: runtime error: invalid memory address or nil pointer dereference`

azure/converters/spotinstances.go
CecileRobertMichon: see my comment below https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2689/files#r990558486

azure/converters/spotinstances.go
willie-yao: This makes sense! I implemented this, but the unit tests that test if the spot vm converter returns a default eviction policy is now broken, since the defaults are being set in the webhook. Should I just remove the [test](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/2a0afb510a3554f4aacfaa411d9c66cd2245ecbe/azure/services/virtualmachines/spec_test.go#L273) entirely?

willie-yao:
Just rebased on main. Not sure why but Git is saying I changed [azure_vmextensions.go](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2689/files#diff-2bea7aae68fda3b77b9d0180bd026269d0804ded551d4d9d5897e528cf20da31) with the same changes from #2698 . It must be a bug since main has the changes shown on the diff for this PR.

CecileRobertMichon (commented):
/lgtm
/assign @jackfrancis

CecileRobertMichon:
> Not sure why but Git is saying I changed [azure_vmextensions.go](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2689/files#diff-2bea7aae68fda3b77b9d0180bd026269d0804ded551d4d9d5897e528cf20da31) with the same changes from https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2698 . It must be a bug since main has the changes shown on the diff for this PR.

hmm that's strange... could you please try git fetch and git rebase again?

jackfrancis (approved):
/approve

PR #2700, opened 2022-10-04
fix agent pool created in failed state becoming ready

nojnhuh:
**What type of PR is this?**
/kind bug

Agent pools for AKS clusters might be created but in a "Failed" provisioning state. In those cases, the error from the create operation is set on the AzureManagedMachinePool's status, logged, and the object is requeued. During the next reconciliation, the agent pool is found to exist and with the same parameters as when it was created which is assumed to indicate success. The "AgentPoolsReady" condition is then [set to true](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/7c3081b339073cb038faafb1a3d5047bbca6cb69/azure/services/agentpools/agentpools.go#L97), even though a later error indicates [the backing VMSS does not exist](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/7c3081b339073cb038faafb1a3d5047bbca6cb69/exp/controllers/azuremanagedmachinepool_reconciler.go#L126). This causes the MachinePool's "InfrastructureReady" condition to turn true even though no infra exists. This change will reset the condition to false if an error occurs after the create/update operation finishes.

Following up from https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2665#issuecomment-1259916539

Fixes #

**Special notes for your reviewer**: I came across this particular scenario while testing #2664 by creating a new node pool (in either a new or existing AKS cluster) with more replicas than available IP addresses in the public IP prefix resource.

This is related to #2615.

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/exp/controllers/azuremanagedmachinepool_controller.go
+++ b/exp/controllers/azuremanagedmachinepool_controller.go
@@ -35,6 +35,7 @@ import (
 	expv1 "sigs.k8s.io/cluster-api/exp/api/v1beta1"
 	"sigs.k8s.io/cluster-api/util"
 	"sigs.k8s.io/cluster-api/util/annotations"
+	"sigs.k8s.io/cluster-api/util/conditions"
 	"sigs.k8s.io/cluster-api/util/predicates"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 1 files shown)


jackfrancis (approved):
/lgtm
/approve

PR #2707, opened 2022-10-06
add “max version” config for getting “latest” AKS version

jackfrancis:
**What type of PR is this?**

/kind failing-test

This PR adds a "max version" to pair with our "get the latest available AKS version" foo, so that we can declare that beyond a certain version we don't want to passively pull in the latest version that AKS supports.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```


test/e2e/config/azure-dev.yaml
```diff
@@ -149,6 +149,7 @@ providers:
 
 variables:
   AKS_KUBERNETES_VERSION: "latest"
+  AKS_KUBERNETES_MAX_VERSION: "1.23"
```
CecileRobertMichon, line 152: I might be missing something but any particular reason why we're not just setting `AKS_KUBERNETES_VERSION` to `stable-1.23` to achieve the same result?

jackfrancis:
Not needed!

PR #2711, opened 2022-10-07
AzureManagedCluster spec.controlPlaneEndpoint is immutable

jackfrancis:
**What type of PR is this?**

/kind bug

While auditing AzureManagedCluster property CRUD operations, I observed that you're able to edit both the `azuremanagedcluster` resource and the `azuremanagedcontrolplane` resource and change the values of the `spec.controlPlaneEndpoint` configuration. The changes are accepted, and then it seems that the controllers are smart enough to revert the changes (or possibly they are calling the AKS API and resetting it to its authoritative value).

Rather than the above, we should fail updates on that configuration via webhook.

For cluster creates, we explicity fail (via webhook) if either of the `ControlPlaneEndpoint` property values are present, which expresses the actual API contract (these properties are designed to be user-facing).

For cluster updates, we explicitly fail all updates that change non-zero type values, which has the upside of preventing user modifications to this AKS-owned property surface area. Because this is an intrinsic, immutable configuration property from AKS's perspective, it is appropriate to enforce this immutability even if updates originat from AKS API (if CAPZ detected that the AKS API wanted to change these properties that would not an unexpected violation of the AKS feature contract).

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
AzureManagedCluster spec.controlPlaneEndpoint is immutable
```

### DIFF
--- a/exp/api/v1beta1/azuremanagedcluster_webhook.go
+++ b/exp/api/v1beta1/azuremanagedcluster_webhook.go
@@ -26,6 +26,7 @@ import (
 	"sigs.k8s.io/cluster-api-provider-azure/azure"
 	"sigs.k8s.io/cluster-api-provider-azure/feature"
 	"sigs.k8s.io/cluster-api-provider-azure/util/maps"
+	webhookutils "sigs.k8s.io/cluster-api-provider-azure/util/webhook"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 )
@@ -51,6 +52,18 @@ func (r *AzureManagedCluster) ValidateCreate() error {
 			"can be set only if the AKS feature flag is enabled",
 		)
 	}
+	if r.Spec.ControlPlaneEndpoint.Host != "" {
+		return field.Forbidden(
+			field.NewPath("Spec", "ControlPlaneEndpoint", "Host"),
+			controlPlaneEndpointErrorMessage,
+		)
+	}
+	if r.Spec.ControlPlaneEndpoint.Port != 0 {
+		return field.Forbidden(
+			field.NewPath("Spec", "ControlPlaneEndpoint", "Port"),
+			controlPlaneEndpointErrorMessage,
+		)
+	}
 	return nil
 }
 
@@ -68,6 +81,24 @@ func (r *AzureManagedCluster) ValidateUpdate(oldRaw runtime.Object) error {
 				field.NewPath("metadata", "annotations"),
 				r.ObjectMeta.Annotations,
 				fmt.Sprintf("annotations with '%s' prefix are immutable", azure.CustomHeaderPrefix)))
+	}
+
+	if old.Spec.ControlPlaneEndpoint.Host != "" {
+		if err := webhookutils.ValidateImmutable(
+			field.NewPath("Spec", "ControlPlaneEndpoint", "Host"),
+			old.Spec.ControlPlaneEndpoint.Host,
+			r.Spec.ControlPlaneEndpoint.Host); err != nil {
+			allErrs = append(allErrs, err)
+		}
+	}
+
+	if old.Spec.ControlPlaneEndpoint.Port != 0 {
+		if err := webhookutils.ValidateImmutable(
+			field.NewPath("Spec", "ControlPlaneEndpoint", "Port"),
+			old.Spec.ControlPlaneEndpoint.Port,
+			r.Spec.ControlPlaneEndpoint.Port); err != nil {
+			allErrs = append(allErrs, err)
+		}
 	}
 
 	if len(allErrs) != 0 {
--- a/exp/api/v1beta1/azuremanagedcontrolplane_webhook.go
+++ b/exp/api/v1beta1/azuremanagedcontrolplane_webhook.go
@@ -37,7 +37,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-var kubeSemver = regexp.MustCompile(`^v(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)([-0-9a-zA-Z_\.+]*)?$`)
+var (
+	kubeSemver                       = regexp.MustCompile(`^v(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)([-0-9a-zA-Z_\.+]*)?$`)
+	controlPlaneEndpointErrorMessage = "can not be set by the user, will be set automatically by AKS after the cluster is Ready"
+)
 
 // SetupWebhookWithManager sets up and registers the webhook with the manager.
 func (m *AzureManagedControlPlane) SetupWebhookWithManager(mgr ctrl.Manager) error {
@@ -90,6 +93,20 @@ func (m *AzureManagedControlPlane) ValidateCreate(client client.Client) error {
 			"can be set only if the AKS feature flag is enabled",
 		)
 	}
+
+	if m.Spec.ControlPlaneEndpoint.Host != "" {
+		return field.Forbidden(
+			field.NewPath("Spec", "ControlPlaneEndpoint", "Host"),
+			controlPlaneEndpointErrorMessage,
+		)
+	}
+	if m.Spec.ControlPlaneEndpoint.Port != 0 {
+		return field.Forbidden(
+			field.NewPath("Spec", "ControlPlaneEndpoint", "Port"),
+			controlPlaneEndpointErrorMessage,
+		)
+	}
+
 	return m.Validate(client)
 }
 
@@ -190,6 +207,24 @@ func (m *AzureManagedControlPlane) ValidateUpdate(oldRaw runtime.Object, client 
 						m.Spec.AADProfile.AdminGroupObjectIDs,
 						"length of AADProfile.AdminGroupObjectIDs cannot be zero"))
 			}
+		}
+	}
+
+	if old.Spec.ControlPlaneEndpoint.Host != "" {
+		if err := webhookutils.ValidateImmutable(
+			field.NewPath("Spec", "ControlPlaneEndpoint", "Host"),
+			old.Spec.ControlPlaneEndpoint.Host,
+			m.Spec.ControlPlaneEndpoint.Host); err != nil {
+			allErrs = append(allErrs, err)
+		}
+	}
+
+	if old.Spec.ControlPlaneEndpoint.Port != 0 {
+		if err := webhookutils.ValidateImmutable(
+			field.NewPath("Spec", "ControlPlaneEndpoint", "Port"),
+			old.Spec.ControlPlaneEndpoint.Port,
+			m.Spec.ControlPlaneEndpoint.Port); err != nil {
+			allErrs = append(allErrs, err)
 		}
 	}
 
--- a/azure/scope/managedcontrolplane.go
+++ b/azure/scope/managedcontrolplane.go
@@ -532,7 +532,12 @@ func (s *ManagedControlPlaneScope) GetAllAgentPoolSpecs() ([]azure.ResourceSpecG
 
 // SetControlPlaneEndpoint sets a control plane endpoint.
 func (s *ManagedControlPlaneScope) SetControlPlaneEndpoint(endpoint clusterv1.APIEndpoint) {
-	s.ControlPlane.Spec.ControlPlaneEndpoint = endpoint
+	if s.ControlPlane.Spec.ControlPlaneEndpoint.Host == "" {
+		s.ControlPlane.Spec.ControlPlaneEndpoint.Host = endpoint.Host
+	}
+	if s.ControlPlane.Spec.ControlPlaneEndpoint.Port == 0 {
+		s.ControlPlane.Spec.ControlPlaneEndpoint.Port = endpoint.Port
+	}
 }
 
 // MakeEmptyKubeConfigSecret creates an empty secret object that is used for storing kubeconfig secret data.
--- a/exp/api/v1beta1/azuremanagedcluster_webhook_test.go
+++ b/exp/api/v1beta1/azuremanagedcluster_webhook_test.go
@@ -23,6 +23,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilfeature "k8s.io/component-base/featuregate/testing"
 	"sigs.k8s.io/cluster-api-provider-azure/feature"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 )
 
 func TestAzureManagedCluster_ValidateUpdate(t *testing.T) {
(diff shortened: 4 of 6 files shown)


exp/api/v1beta1/azuremanagedcontrolplane_webhook.go
```diff
@@ -141,77 +141,41 @@ func (m *AzureManagedControlPlane) ValidateUpdate(oldRaw runtime.Object, client
 
 	if old.Spec.DNSServiceIP != nil {
 		// Prevent DNSServiceIP modification if it was already set to some value
-		if m.Spec.DNSServiceIP == nil {
-			// unsetting the field is not allowed
+		if !reflect.DeepEqual(m.Spec.DNSServiceIP, old.Spec.DNSServiceIP) {
```
jackfrancis, line 144: These `reflect.DeepEqual` changes are in response to linter complaining about this func's cyclomatic complexity with the addition of more validations. Next step might be to refactor this more, but this works for now.

exp/api/v1beta1/azuremanagedcluster_webhook.go
```diff
@@ -71,6 +71,26 @@ func (r *AzureManagedCluster) ValidateUpdate(oldRaw runtime.Object) error {
 				fmt.Sprintf("annotations with '%s' prefix are immutable", azure.CustomHeaderPrefix)))
 	}
 
+	if old.Spec.ControlPlaneEndpoint.Host != "" {
```
CecileRobertMichon, line 74: Are there any cases where the controller would need to change either of these? Because webhook validations also apply to patches made by controllers, this essentially prevents the CAPZ controllers from changing these values too.

jackfrancis: I did some investigation and it seems that AKS does not have any notion of ever being able to change the apiserver endpoint.

exp/api/v1beta1/azuremanagedcontrolplane_webhook.go
CecileRobertMichon: We should really refactor this to not do all the validation in one big function (maybe add a `ValidateImmutableFields`?), could you please open an issue so we don't forget? Could be a good first issue maybe

we're losing some error differentiation with this change so it's not fully equivalent, are we okay with that?

jackfrancis: Is something like this an improvement (imagine this applied across all webhook validations that check for immutability):

https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2718

jackfrancis: I'll finish #2718 to completion (audit all webhooks) after this PR lands (I think this PR is ready for final review).

And then after #2718 I'll file an issue that advocates this surface area is a potential good opportunity for introducing go generics into capz codebase.

CecileRobertMichon: > I'll finish https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2718 to completion (audit all webhooks) after this PR lands (I think this PR is ready for final review)

In that case I would advocate that we temporarily fix the go cyclomatic linter some other way (maybe extract some stuff to a function) so we don't temporarily regress the error messaging

jackfrancis: How about we just land #2718 first? (this PR is not highest priority)

CecileRobertMichon: that works too 👍

jackfrancis: Prereq work is done, this PR is ready for another review round

exp/api/v1beta1/azuremanagedcluster_webhook.go
nawazkh: How would this change affect a scenario where the user has put in wrong/invalid/ill-formed `ControlPlaneEndpoint` ? Do we validate initial `ControlPlaneEndpoint` ?

jackfrancis: Good question @nawazkh!

In fact this is one of those "weird" fields where we defer to the AKS service itself to populate the value (and thus we trust that the AKS service will always populate the right value).

@CecileRobertMichon @nojnhuh is there any additional foo we should put in place here? It does feel a bit weird not having any create webhook validations for such an important field, but on the other hand this is an exceptional case where we're not actually able to implement what we want:

1. Never allow the user to ever populate this field!
2. Always trust the AKS service and accept whatever it provides!

Is there a better way we could be doing this, or is "don't enforce CREATE webhooks" the most appropriate given the state of AKS + capz?

CecileRobertMichon: The real issue here is that controlPlaneEndpoint should be a Status field, not a Spec field, since it's meant for the CAPZ controller to communicate back info to CAPI, not to be user configurable.

I had a write-up on this in https://github.com/kubernetes-sigs/cluster-api/issues/3715 but unfortunately in never moved forward.

jackfrancis: That makes sense. My interpretation of that is that we're doing the best we can right now.

exp/api/v1beta1/azuremanagedcluster_webhook.go
```diff
@@ -70,6 +71,24 @@ func (r *AzureManagedCluster) ValidateUpdate(oldRaw runtime.Object) error {
...
```
willie-yao, line 74: Is there a case where `old.Spec.ControlPlaneEndpoint` is nil? I saw that the control plane endpoint is an optional field for the azure managed cluster spec

jackfrancis: Yes (I assume you mean `""` [empty string], not nil), during initial cluster creation (because this data is never populated by the end user). We can't enforce immutability for transitions like `""` --> `<a valid URL>` because that's the update vector we need to expose to allow the AKS API to provide us with data to do an automated update of this value after the cluster is created (the control plane URL is one piece of output data from the AKS API after a successful cluster create).

nojnhuh: `ControlPlaneEndpoint` is a plain struct value, not a pointer, so there isn't a risk of dereferencing a nil pointer in this case.

exp/api/v1beta1/azuremanagedcluster_webhook.go
nojnhuh: Is the behavior currently that if the `ControlPlaneEndpoint` is changed, that will trigger a reconciliation where the controller will immediately override the provided value with AKS's value? If so, then to @nawazkh's point, it seems like this removes the possibility for the controller to automatically reset the value in case it does happen to change. e.g. If a user mistakenly sets this before the controller does, it seems like the only way to resolve this is to delete and recreate the control plane. Overall I think this check is valuable, but worth being mindful of that change.

One other idea: Is there a way to make this trigger a warning instead of a hard error? Then the webhook could say something to the effect of, "You're trying to change this value, but ultimately we're going to ignore that and set it to what AKS tells us it should be."

jackfrancis: If the *user* is responsible for changing from `""` --> `<some value>`, then that `<some value>` will become the durable one, and the automated AKS API-informed update that happens during cluster create reconciliation will be webhook-rejected with this change.

I'm certainly open to any comments on how we can improve this! It seemed to my initially naive understanding that the current situation, with no enforcement upon this rather critical configuration vector, left room for improvement.

nojnhuh (approved):
I agree this isn't a perfect solution and also that "perfect" likely isn't attainable here, but this is definitely a positive step overall!
/lgtm

willie-yao (approved):
/lgtm

jackfrancis:
Would love to get one more sign-off from @CecileRobertMichon to make sure merging this is preferable to the current status quo. Thank you!

CecileRobertMichon:
Any chance we could discuss during office hours tomorrow? I'm not super clear on how this is an improvement. My current understanding:

1. Currently, the AzureManagedCluster controlPlaneEndpoint gets set by the controller and overrides whatever value is already there. This means if a user sets a value it gets ignored and overwritten, and there are no errors (but user gets ignored).
2. With this change, AzureManagedCluster controlPlaneEndpoint becomes immutable so updates get rejected by the webhook. If a user sets a value that is different from the AKS API Server, the controller would run into an error trying to update the field as the update would get rejected by the webhook (user is allowed to shoot themselves in the foot).
3. This is different from how the same validation is handled for AzureCluster. AzureCluster controlPlaneEndpoint is immutable in the webhook so the webhook rejects any updates, but, the controller doesn't try to set it unless the value is unset. This means the controller will never run into a webhook error trying to set an immutable field (but also wrong value could be set by the user and issues would happen).

I think my main questions are: similar to AzureCluster, should we change the managed control plane controller to only try to update the field if it's empty so it doesn't run into its own webhook validation error? If the intent is to never let the user set the value, is there a way to reject the user trying to set the field before the controller has a chance to? Maybe we can use Status and only allow setting the endpoint if the cluster is ready or provisioned? Is there ever a valid scenario for the user wanting to set the endpoint themselves before cluster exists?

jackfrancis:
@CecileRobertMichon let's discuss during office hours. In fact this PR was my attempt to do what you describe in point 3 above.

jackfrancis:
@CecileRobertMichon incorporated discussion points from yesterday's office hours, PTAL

exp/api/v1beta1/azuremanagedcluster_webhook.go
```diff
@@ -51,6 +52,21 @@ func (r *AzureManagedCluster) ValidateCreate() error {
 			"can be set only if the AKS feature flag is enabled",
 		)
 	}
+	if !r.Status.Ready {
```
nojnhuh, line 55: Is there a reason we're only checking the status on creates but not updates?

jackfrancis: Naively I was thinking to protect the "user accidentally includes a value in the `AzureManagedCluster` resource spec when creating the cluster, but not including this check in update leaves open the edge case of "user updates this property after cluster create but before the AKS API does *its* update", and so yeah, I'll add it there as well.

nawazkh: In addition to your thought process, would it be worth it to hide `AzureManagedCluster.Spec.ControlPlaneEndpoint` in the CRD from the user ? 
It would [technically be part of the spec](https://cluster-api.sigs.k8s.io/developer/providers/cluster-infrastructure.html#infracluster-resources), just not visible to the user. Could add a psychological barrier on updating a non-exposed field.

jackfrancis: That sounds like a breaking API change, no?

nawazkh: > hide AzureManagedCluster.Spec.ControlPlaneEndpoint in the CRD from the user

Hmm, yeah, that would be a breaking change. We can ignore my comment https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2711#discussion_r1046525392

jackfrancis:
@CecileRobertMichon fair point, I've applied the change there as well

nawazkh:
/lgtm
🚀

exp/api/v1beta1/azuremanagedcontrolplane_webhook.go
```diff
@@ -193,6 +193,24 @@ func (m *AzureManagedControlPlane) ValidateUpdate(oldRaw runtime.Object, client
 		}
 	}
 
+	if old.Spec.ControlPlaneEndpoint.Host != "" {
```
CecileRobertMichon, line 196: for AzureManagedControlPlane is there is no "ready" equivalent to make sure the user doesn't set it on the initial create? If so that's unfortunate, since that's where the user would be most likely to set it in the first place since that's where they set all other configurations

jackfrancis: There isn't actually. Well, the resource spec includes that field but we're not setting it.

jackfrancis: Would we want to address this gap in this PR? Something like this would do the trick

```
$ git diff
diff --git a/azure/scope/managedcontrolplane.go b/azure/scope/managedcontrolplane.go
index 251d77c59..4d1a01579 100644
--- a/azure/scope/managedcontrolplane.go
+++ b/azure/scope/managedcontrolplane.go
@@ -532,6 +532,7 @@ func (s *ManagedControlPlaneScope) GetAllAgentPoolSpecs() ([]azure.ResourceSpecG
 
 // SetControlPlaneEndpoint sets a control plane endpoint.
 func (s *ManagedControlPlaneScope) SetControlPlaneEndpoint(endpoint clusterv1.APIEndpoint) {
+       s.ControlPlane.Status.Ready = true
        if s.ControlPlane.Spec.ControlPlaneEndpoint.Host == "" {
                s.ControlPlane.Spec.ControlPlaneEndpoint.Host = endpoint.Host
        }
diff --git a/exp/api/v1beta1/azuremanagedcontrolplane_webhook.go b/exp/api/v1beta1/azuremanagedcontrolplane_webhook.go
index 451df1c78..a81a0185c 100644
--- a/exp/api/v1beta1/azuremanagedcontrolplane_webhook.go
+++ b/exp/api/v1beta1/azuremanagedcontrolplane_webhook.go
@@ -90,6 +90,21 @@ func (m *AzureManagedControlPlane) ValidateCreate(client client.Client) error {
                        "can be set only if the AKS feature flag is enabled",
                )
        }
+       if !m.Status.Ready {
+               errorMessage := "can not be set by the user, will be set automatically by AKS after the cluster is Ready"
+               if m.Spec.ControlPlaneEndpoint.Host != "" {
+                       return field.Forbidden(
+                               field.NewPath("Spec", "ControlPlaneEndpoint", "Host"),
+                               errorMessage,
+                       )
+               }
+               if m.Spec.ControlPlaneEndpoint.Port != 0 {
+                       return field.Forbidden(
+                               field.NewPath("Spec", "ControlPlaneEndpoint", "Port"),
+                               errorMessage,
+                       )
+               }
+       }
        return m.Validate(client)
 }
 
@@ -193,7 +208,7 @@ func (m *AzureManagedControlPlane) ValidateUpdate(oldRaw runtime.Object, client
                }
        }
 
-       if old.Spec.ControlPlaneEndpoint.Host != "" {
+       if old.Spec.ControlPlaneEndpoint.Host != "" || !m.Status.Ready {
                if err := webhookutils.ValidateImmutable(
                        field.NewPath("Spec", "ControlPlaneEndpoint", "Host"),
                        old.Spec.ControlPlaneEndpoint.Host,
@@ -202,7 +217,7 @@ func (m *AzureManagedControlPlane) ValidateUpdate(oldRaw runtime.Object, client
                }
        }
 
-       if old.Spec.ControlPlaneEndpoint.Port != 0 {
+       if old.Spec.ControlPlaneEndpoint.Port != 0 || !m.Status.Ready {
                if err := webhookutils.ValidateImmutable(
                        field.NewPath("Spec", "ControlPlaneEndpoint", "Port"),
                        old.Spec.ControlPlaneEndpoint.Port,
```

jackfrancis: cc @nojnhuh

CecileRobertMichon: For AzureCluster we do it right after setting the endpoint (before that data gets persisted and the webhook is called) https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/controllers/azurecluster_controller.go#L253, we could do something similar here.

> Would we want to address this gap in this PR?

A separate PR is probably better since this is a bit of a scope creep, we can hold this one until it gets merged.

jackfrancis: I was wrong about that!

Can address the webhook readiness condition in this PR.

jackfrancis: O.K., I did more investigation here and it doesn't appear we can (easily) enforce Status.Ready=true as a requirement. Status=True and ControlPlaneEndpoint data populations happen during the same reconciliation, and it appears in practice the webhooks are engaged with a `m *AzureManagedControlPlane` context (and the same for `r *AzureManagedCluster` in that webhook) that does not have the Status=True data. So if we add a requirement in the webhook that says "only allow a `"` --> `"<actual endpoint data>"` update if Status=True it will always fail, thus short-circuiting the reconciliation, thus blocking the entire cluster creation flow.

I'd be curious if I'm overlooking things.

jackfrancis:
@nojnhuh @CecileRobertMichon I think we should scope this PR down to the updated PR description. Integrating "cluster readiness" and/or "detecting whether an update originates from the AKS API vs a user" is proving to be difficult, and the real-world payoff for this additional webhook enforcement is (IMO) not worth the effort at this time.

Ready for final review.

nojnhuh (approved):
/lgtm

CecileRobertMichon (commented):
/lgtm
/approve

PR #2714, opened 2022-10-09
Allow to set custom backend pool names for load balancers

Fedosin:
**What type of PR is this?**
/kind feature

Currently, there is no way to specify the backend pool name. This PR makes LB's [backend pool](https://docs.microsoft.com/en-us/azure/load-balancer/backend-pool-management) name customizable. The changes introduced in this PR are backward compatible and should not break existing deployments.

Adding a struct `Backend` to API allows for future expansion of the configuration. If we later wanted to add some additional config, it would be preferable to have them all in a struct rather than BackendPoolName, BackendPoolPort, etc

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Release note**:

```release-note
Allow to set custom backend pool names for load balancers
```


azure/scope/cluster.go
```diff
@@ -228,6 +228,11 @@ func (s *ClusterScope) PublicIPSpecs() []azure.ResourceSpecGetter {
 
 // LBSpecs returns the load balancer specs.
 func (s *ClusterScope) LBSpecs() []azure.ResourceSpecGetter {
+	apiBackendPoolName := s.APIServerLBPoolName(s.APIServerLB().Name)
+	if s.APIServerLB().BackendPool.Name != nil {
```
CecileRobertMichon, line 232: Let's move this defaulting to a webhook, see https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2432#discussion_r915240838

CecileRobertMichon:
Thanks for picking this up @Fedosin! it looks like this is missing some of the changes that were made as part of the review that already happened on https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2432, could you please make sure you're reconciling those?

Fedosin:
@CecileRobertMichon Hi! I updated the PR as you asked. Could you take a look again?

api/v1beta1/azurecluster_default.go
```diff
@@ -193,6 +193,10 @@ func (c *AzureCluster) setAPIServerLBDefaults() {
 			}
 		}
 	}
+
+	if lb.BackendPool.Name == nil {
```
CecileRobertMichon, line 197: thanks for moving this defaulting to the webhook!

api/v1beta1/types.go
```diff
@@ -768,6 +771,14 @@ type AzureBastion struct {
 	PublicIP PublicIPSpec `json:"publicIP,omitempty"`
 }
 
+// BackendPool describes the backend pool of the load balancer.
+type BackendPool struct {
+	// Name specifies the name of backend pool for the load balancer. If not specified, the default name will
+	// be set, depending on the load balancer role.
+	// +optional
+	Name *string `json:"name,omitempty"`
```
CecileRobertMichon, line 779: is there value in making this a string pointer given that `""` will never be a valid name? If it was just a ` string`  it would avoid having to dereference it in a bunch of places (and it would be more consistent with other `Name` fields in the API)

azure/scope/cluster.go
```diff
@@ -244,7 +244,7 @@ func (s *ClusterScope) LBSpecs() []azure.ResourceSpecGetter {
 			Type:                 s.APIServerLB().Type,
 			SKU:                  infrav1.SKUStandard,
 			Role:                 infrav1.APIServerRole,
-			BackendPoolName:      s.APIServerLBPoolName(s.APIServerLB().Name),
+			BackendPoolName:      *s.APIServerLB().BackendPool.Name,
```
CecileRobertMichon, line 247: we shouldn't dereference a pointer directly to avoid panics, let's use `to.String()` instead, or get rid of the pointer altogether (see my comment above)

api/v1beta1/types.go
Fedosin: Yes, I agree. I copied it from the original [PR](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2432) where Name is a pointer. I updated it as you asked.

nawazkh (approved):
Thanks for putting this together!
/lgtm

CecileRobertMichon:
@Fedosin please squash

Fedosin:
@CecileRobertMichon now it's squashed

CecileRobertMichon (commented):
/lgtm
/approve

Thank you @nawazkh @willie-yao for reviewing 🙌

PR #1520, opened 2021-07-13
Added support for system node pools in managedClusters

LochanRn:
**What type of PR is this?**
/kind bug
/area managedclusters

**What this PR does / why we need it**:
system node pool will be of type system and not user.
added mode in the azuremannagedmachinepool api to set the mode of azure agent pools
removed defaultPoolRef from azuremanagedcontrolplane api. (provides more flexibility to manage system node pools.)
validation for deletion of last system node pool is handled in the webhook.
bumped up the containerservice sdk from 2020-02-01 to 2021-05-01
updated the aks docs

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes  #1376 #1418 #1536

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Mode spec in the AzureManagedMachinePool is used to specify the mode of an agentPool i.e System or User.
Removed defaultPoolRef from AzureManagedControlPlane. 
```

PR #2717, opened 2022-10-10
AKS: clarify that one system pool is required

jackfrancis:
**What type of PR is this?**

/kind cleanup

This is a suggested improvement to the language that we print out when trying to set all node pools to `User`. The simplest way to run into this error is to build a single node pool AKS cluster, and then change the `mode` configuration from `System` to `User`.

I was a little confused with the existing error message:

```
# Please edit the object below. Lines beginning with a '#' will be ignored,
# and an empty file will abort the edit. If an error occurs while saving this file will be
# reopened with the relevant failures.
#
# azuremanagedmachinepools.infrastructure.cluster.x-k8s.io "pool0" was not valid:
# * Spec.Mode: Invalid value: "User": Last system node pool cannot be mutated to user node pool
#
```

I think this one might be a bit clearer:

```
# Please edit the object below. Lines beginning with a '#' will be ignored,
# and an empty file will abort the edit. If an error occurs while saving this file will be
# reopened with the relevant failures.
#
# azuremanagedmachinepools.infrastructure.cluster.x-k8s.io "pool0" was not valid:
# * Spec.Mode: Invalid value: "User": Cannot change node pool mode to User, you must have at least one System node pool in your cluster
#
```

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/exp/api/v1beta1/azuremanagedmachinepool_webhook.go
+++ b/exp/api/v1beta1/azuremanagedmachinepool_webhook.go
@@ -154,10 +154,9 @@ func (m *AzureManagedMachinePool) ValidateUpdate(oldRaw runtime.Object, client c
 	if m.Spec.Mode != string(NodePoolModeSystem) && old.Spec.Mode == string(NodePoolModeSystem) {
 		// validate for last system node pool
 		if err := m.validateLastSystemNodePool(client); err != nil {
-			allErrs = append(allErrs, field.Invalid(
+			allErrs = append(allErrs, field.Forbidden(
 				field.NewPath("Spec", "Mode"),
-				m.Spec.Mode,
-				"Last system node pool cannot be mutated to user node pool"))
+				"Cannot change node pool mode to User, you must have at least one System node pool in your cluster"))
 		}
 	}
 


jackfrancis:
I'd love for @LochanRn to weigh in on whether or not he things this is an improvement, as he authored the original feature:

https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1520/files

(belated thank you for that, btw!)

exp/api/v1beta1/azuremanagedmachinepool_webhook.go
```diff
@@ -157,7 +157,7 @@ func (m *AzureManagedMachinePool) ValidateUpdate(oldRaw runtime.Object, client c
 			allErrs = append(allErrs, field.Invalid(
```
CecileRobertMichon, line 157: I wonder if `Forbidden` would be a more appropriate error type here

> // Forbidden returns a *Error indicating "forbidden".  This is used to
// report valid (as per formatting rules) values which would be accepted under
// some conditions, but which are not permitted by current conditions (e.g.
// security policy).

jackfrancis: I like it

CecileRobertMichon (commented):
/lgtm
/approve

PR #2718, opened 2022-10-10
standarize strict enforcement for immutable properties

jackfrancis:
**What type of PR is this?**

/kind cleanup

This PR standardizes the way we validate immutable fields via webhook, and refactors existing, one-off validation implementations to re-use common funcs.

Created a new file `util/webhook/validator.go` to host these exportable funcs so that we can easily re-use them across `exp/api/` and `api/`.

There are intentional, material changes in this PR to properly handle property immutability for AKS cluster properties that are immutable. The following `AzureManagedControlPlane` properties are now being checked for strict immutability:

- `Spec.SSHPublicKey`
- `Spec.DNSServiceIP`
- `Spec.NetworkPlugin`
- `Spec.NetworkPolicy`
- `Spec.LoadBalancerSKU`

And the following `AzureManagedMachinePool` properties:

- `Spec.OSDiskSizeGB`
- `Spec.MaxPods`
- `Spec.OsDiskType`

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
standarize strict enforcement for immutable properties
```

### DIFF
--- a/exp/api/v1beta1/azuremanagedmachinepool_webhook.go
+++ b/exp/api/v1beta1/azuremanagedmachinepool_webhook.go
@@ -31,6 +31,7 @@ import (
 	"sigs.k8s.io/cluster-api-provider-azure/azure"
 	azureutil "sigs.k8s.io/cluster-api-provider-azure/util/azure"
 	"sigs.k8s.io/cluster-api-provider-azure/util/maps"
+	webhookutils "sigs.k8s.io/cluster-api-provider-azure/util/webhook"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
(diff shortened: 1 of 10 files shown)


jackfrancis:
/hold

under active dev

exp/api/v1beta1/webhook.go
```diff
@@ -0,0 +1,117 @@
...
+	"k8s.io/apimachinery/pkg/util/validation/field"
+)
+
+const (
+	unsetMessage     = "field is immutable, unsetting is not allowed"
+	setMessage       = "field is immutable, setting is not allowed"
+	immutableMessage = "field is immutable"
+)
+
+func ensureStringSlicesAreEqual(a []string, b []string) bool {
```
nojnhuh, line 29: (acknowledging that this was just copy-pasted and not changed in this PR)

It looks like this will interpret something like `["a", "a", "b"]` and `["a", "b", "b"]` as equal, which may not be expected. If some other mechanism is preventing duplicate elements or otherwise handling this appropriately then I don't think this is an issue.

exp/api/v1beta1/webhook.go
```diff
@@ -0,0 +1,117 @@
+/*
+Copyright 2022 The Kubernetes Authors.
```
CecileRobertMichon, line 2: Would `webhook_helper.go` or `webhook_util.go` be a more accurate filename?

exp/api/v1beta1/azuremanagedmachinepool_webhook.go
```diff
@@ -271,7 +223,7 @@ func (m *AzureManagedMachinePool) validateLastSystemNodePool(cli client.Client)
 		return err
 	}
 
-	if len(ammpList.Items) <= 1 {
+	if len(ammpList.Items) < 1 {
```
CecileRobertMichon, line 226: seems like we're changing core validation logic, is this fixing a bug? If so, can you please split the commits?

exp/api/v1beta1/webhook.go
jackfrancis: yeah I don't know what's going on here, I'm just going to fix this

exp/api/v1beta1/azuremanagedmachinepool_webhook.go
jackfrancis: yeah, this was an oversight, thanks!

exp/api/v1beta1/webhook.go
jackfrancis: With the expanded scope, moved all of this into `util/webhook/validator.go` where we get util/helper naming vibes for free.

CecileRobertMichon:
> There are intentional, material changes in this PR to properly handle property immutability for AKS cluster properties that are immutable. The following AzureManagedControlPlane properties are now being checked for strict immutability

We should keep behavior changes separate from refactor changes, especially validation behavior which is considered a breaking change for the user API fields and should have its own release note. Please split those into a separate PR with an explicit release note about which fields are now immutable/required.

jackfrancis:
@CecileRobertMichon the part of the work that simply generalizes immutability checks into a common set of convenience funcs is here: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2741

jackfrancis:
Opened https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2795 instead

PR #2728, opened 2022-10-14
add shouldnot to codespellignore

jackfrancis:
**What type of PR is this?**

/kind failing-test

I have definitive evidence that I work for codespell and not the other way around.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/.codespellignore
+++ b/.codespellignore
@@ -4,3 +4,4 @@ geting
 ot
 intepreted
 updat
+shouldnot


CecileRobertMichon:
> Error: ./.codespellignore:7: ShouldNot ==> should not, shouldn't

hmmmm what?

.codespellignore
```diff
@@ -4,3 +4,4 @@ geting
 ot
 intepreted
 updat
+ShouldNot
```
CecileRobertMichon, line 7: could you try changing this to

```suggestion
shouldnot
```

Seems that there is a known bug where it only wants lowercase https://github.com/codespell-project/codespell/issues/2451#issuecomment-1269314113

jackfrancis: that works!

jackfrancis:
@CecileRobertMichon you have my permission to admin merge this if you're looking for more GitHub badge achievements

CecileRobertMichon:
/lgtm
/approve

This fixes it for now. Let's follow up on 
1) Why shouldnot is getting flagged, it's a common testing word and is a regression in the latest codespell release IMO
2) Whether we think there is value in keeping codespell enabled in general

PR #2612, opened 2022-08-30
Revert "Remove kcp adoption test"

jackfrancis:
Reverts kubernetes-sigs/cluster-api-provider-azure#2601

Issue #2596, opened 2022-08-23
Remove KCP adoption test


CecileRobertMichon:
The below test was added to ensure existing clusters with control plane machines before KCP (KubeadmControlPlane) was introduced could be adopted by a KCP object: 

https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/test/e2e/capi_test.go#L158

All new clusters should be using KCP for clusters built since ~ v1alpha4 so we should deprecate (remove) the test from the CAPI e2e test suite, and remove the associated e2e test templates (https://github.com/kubernetes-sigs/cluster-api-provider-azure/tree/main/test/e2e/data/infrastructure-azure/v1beta1/cluster-template-kcp-adoption) and all references to it.

willie-yao:
@CecileRobertMichon would we want to remove [cluster-template-kcp-adoption.yaml](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/test/e2e/data/infrastructure-azure/v1beta1/cluster-template-kcp-adoption.yaml) and [cluster-with-kcp.yaml](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/test/e2e/data/infrastructure-azure/v1beta1/bases/cluster-with-kcp.yaml) as well?

CecileRobertMichon:
@willie-yao yes

willie-yao:
Update: we do not want to remove `cluster-with-kcp.yaml` as it is the basis for several other templates

PR #2731, opened 2022-10-15
Remove kcp adoption test

CecileRobertMichon:
Signed-off-by: willie-yao <williamyao@microsoft.com>

 

 
**What type of PR is this?**

**What this PR does / why we need it**: Attempt to re-apply #2612 and understand why it broke upgrade tests.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```


CecileRobertMichon:
@willie-yao jfyi I tried opening a PR to re-apply your changes from #2596 but it makes the upgrade test fail still, I have no idea why 🤷‍♀️

CecileRobertMichon:
I may have found something...

willie-yao:
> @willie-yao jfyi I tried opening a PR to re-apply your changes from #2596 but it makes the upgrade test fail still, I have no idea why 🤷‍♀️

That's very interesting.. It doesn't seem like the two tests are related in any way but I'll take a closer look!

nojnhuh:
I think Willie and I figured out what's going on here. The KCP adoption spec from CAPI modifies some global state setting the default timeout for Gomega's `Eventually` outside of an `It` block which is evaluated when the test is defined, not in its regular execution flow:
https://github.com/kubernetes-sigs/cluster-api/blob/v1.2.4/test/e2e/kcp_adoption.go#L81-L82

We found out that the CAPI upgrade spec uses an interval definition (`wait-machine-pool-upgrade`) that doesn't exist in our e2e config, so Gomega's default 1s timeout/100ms interval take over:
https://github.com/kubernetes-sigs/cluster-api/blob/v1.2.4/test/e2e/cluster_upgrade.go#L208

So when the KCP adoption test is defined, the default timeout gets set before the upgrade test runs. When the KCP adoption test is removed, the default is never changed and the test fails because the timeout isn't long enough.

The simplest fix for now might be to add a `default/wait-machine-pool-upgrade` interval to the e2e config. Willie tested that and it seemed to work.

devigned:
> I think Willie and I figured out what's going on here. The KCP adoption spec from CAPI modifies some global state setting the default timeout for Gomega's `Eventually` outside of an `It` block which is evaluated when the test is defined, not in its regular execution flow: https://github.com/kubernetes-sigs/cluster-api/blob/v1.2.4/test/e2e/kcp_adoption.go#L81-L82
> 
> We found out that the CAPI upgrade spec uses an interval definition (`wait-machine-pool-upgrade`) that doesn't exist in our e2e config, so Gomega's default 1s timeout/100ms interval take over: https://github.com/kubernetes-sigs/cluster-api/blob/v1.2.4/test/e2e/cluster_upgrade.go#L208
> 
> So when the KCP adoption test is defined, the default timeout gets set before the upgrade test runs. When the KCP adoption test is removed, the default is never changed and the test fails because the timeout isn't long enough.
> 
> The simplest fix for now might be to add a `default/wait-machine-pool-upgrade` interval to the e2e config. Willie tested that and it seemed to work.

Nice find!

willie-yao:
@CecileRobertMichon The line to fix this is `default/wait-machine-pool-upgrade: ["60m", "10s"]`. The CAPI test reads the intervals from our `azure-dev.yaml` file. Would you be able to add this under the "intervals" section? I can test this on my end to make sure it still works with all the other files removed.

CecileRobertMichon:
@willie-yao added, PTAL

Issue #2673, opened 2022-09-27
Add resource health and service health in AzureManagedCluster status


CecileRobertMichon:
/kind feature

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]

AzureManagedCluster provides a way to create AKS clusters, but the health status of AzureManagedCluster CR can be improved by incorporating Resource Health Check & Service Health status that are already available for AKS clusters on Azure.

Resource Health documentation: https://docs.microsoft.com/en-us/azure/service-health/resource-health-overview
Service Health documentation: https://docs.microsoft.com/en-us/azure/service-health/service-health-overview
Service Health (Events) API: https://docs.microsoft.com/en-us/rest/api/resourcehealth/events/list-by-single-resource#events

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- cluster-api-provider-azure version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

nojnhuh:
A couple questions I have as I'm getting acquainted with these health services:
- At a high level, what information do we want to expose in CAPZ? It looks like the API exposes "availability statuses," "emerging issues," "events," "metadata," and "operations." I haven't taken a deep dive to figure out exactly what each of those means yet, but some guiding principles for what we want would be helpful to have in mind as I learn more.
- It looks like there's a way to wire up events to webhooks and [similar async mechanisms](https://learn.microsoft.com/en-us/azure/azure-monitor/alerts/alerts-common-schema-integrations). Is that something we could reasonably fit in CAPZ and would prefer over polling the Azure API for that data?

dtzar:
The resource health is completed by #2738, but the service health has not been started.  Can we please separate this work into two issues for tracking/prioritization purposes?

CecileRobertMichon:
opened https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/3001

/close

PR #2738, opened 2022-10-19
add AKS resource health to AzureManagedControlPlane

nojnhuh:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**: This PR adds Azure's [Resource Health](https://learn.microsoft.com/en-us/azure/service-health/resource-health-overview) service and wires it up to mirror the current [availability status](https://learn.microsoft.com/en-us/rest/api/resourcehealth/availability-statuses/get-by-resource?tabs=HTTP) for an AKS cluster in a condition of the `status` of its corresponding AzureManagedControlPlane.

Part of #2673

When testing this, I found that there's consistently an "Unknown" status immediately after the AKS cluster is fully provisioned citing an error connecting to the cluster's metrics server. This status goes to the expected "Available" after about 10 minutes.

The other way I found I could generate a non-Available status is to delete an underlying VMSS instance of a machine pool. After about 30 minutes, this status from the API starts being reflected:
```json
{
  "id": "/subscriptions/[redacted secret capz-manager-bootstrap-credentials:subscription-id]/resourcegroups/jon-12709/providers/microsoft.containerservice/managedclusters/jon-12709/providers/Microsoft.ResourceHealth/availabilityStatuses/current",
  "name": "current",
  "type": "Microsoft.ResourceHealth/AvailabilityStatuses",
  "location": "eastus",
  "properties": {
    "availabilityState": "Degraded",
    "title": "Customer Node Power Off",
    "summary": "A node could not be found in the running/powered on state.",
    "reasonType": "Customer Initiated",
    "rootCauseAttributionTime": "2022-10-19T18:48:21.0367057Z",
    "resolutionETA": "2022-10-19T19:08:21Z",
    "reasonChronicity": "Persistent",
    "reportedTime": "2022-10-19T19:52:31.9720417Z"
  }
}
``` 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
AKS resource health added to AzureManagedControlPlane status
```


nojnhuh:
It looks like the e2e test is timing out when it's checking for the right number of replicas of the MachinePool to exist because it takes a while after the AKS instance is created for the Resource Health availability status to turn green. Disregarding that that's probably a bug in Azure, is the test failing in the "correct" place in this scenario?

The node pools themselves in the Azure portal look fully provisioned and in a successful state, but the lingering "Unknown" availability status is blocking the AzureManagedMachinePool reconciler from kicking in and recording that. In this case, an error like "machine pool node doesn't exist" isn't accurate, though in the other scenario I outlined in the PR description where I shut down an underlying VMSS instance, that error seems more reasonable. Is it worth doing some deeper inspection of the actual status to determine how it should be handled? I have a feeling that would make things considerably more messy.

And in the short-term, how should we compensate for the initial "Unavailable" status in the test? Would it be better to increase the timeout, wait until this is fixed in Azure, or somehow work around this in CAPZ?

jackfrancis:
@nojnhuh we might need to update our custom `DiscoverAndWaitForAKSControlPlaneInitialized` func that we use to block progress of the E2E tests on "cluster is up and ready!" status

test/e2e/aks.go
```diff
@@ -87,6 +87,12 @@ func DiscoverAndWaitForAKSControlPlaneInitialized(ctx context.Context, input Dis
 	})
 	Expect(controlPlane).NotTo(BeNil())
 
+	Logf("Waiting for 'Available' status from Azure Resource Health for control plane %s/%s", controlPlane.Namespace, controlPlane.Name)
+	WaitForControlPlaneResourceHealthAvailable(ctx, WaitForControlPlaneResourceHealthAvailableInput{
```
CecileRobertMichon, line 91: Does this considerably slow down the test or is it manageable?

nojnhuh: It's considerable. About 15-20 minutes last I tried.

jackfrancis: I just repro'd this locally and it took 20 minutes.

The cluster in every other way seems fine. I'm unsure if this is something that will be user-impacting (obviously it's E2E-impacting and this additional timeout tolerance is appropriate).

What capi and/or capz CRDs are impacted by this change?

nojnhuh: Until the status turns green and the AzureManagedControlPlane's "Ready" condition turns true, the associated CAPI Cluster and MachinePools and CAPZ AzureManagedMachinePools will also not be marked ready. I'm not sure if any of that would prevents the cluster from actually being usable though.

jackfrancis: Delaying those transitions to Ready by 20 minutes almost certainly has some knock-on effects. How about we turn this into a `AzureManagedControlPlane` bool configuration property that defaults to false? That way we can expose it to folks who definitely want it, and understand the tradeoffs.

nojnhuh: I agree, that seems like a good compromise.

exp/api/v1beta1/azuremanagedcontrolplane_types.go
```diff
@@ -116,6 +116,10 @@ type AzureManagedControlPlaneSpec struct {
 	// APIServerAccessProfile is the access profile for AKS API server.
 	// +optional
 	APIServerAccessProfile *APIServerAccessProfile `json:"apiServerAccessProfile,omitempty"`
+
+	// AzureAvailabilityStatusEnabled determines whether or not the AzureResourceAvailable status condition will be reported.
+	// +optional
+	AzureAvailabilityStatusEnabled bool `json:"azureAvailabilityStatusEnabled"`
```
CecileRobertMichon, line 122: +1 to the suggestion to add a toggle so this feature can be enabled by users who understand it for now until latency is fixed

However, we need to be very careful with adding API fields to implement temporary toggles like this: once it's in the API spec, we can't remove it without making a breaking change which requires a new API version. How about instead using a feature flag at the controller level, similar to the AKS feature flag itself?

The only downside there is that you would enable it for all your clusters at once, not per cluster. I don't know if that's really a downside though, since I can't really see a use case for someone wanting to enable resource health for some clusters but not others.

jackfrancis: Let's bring this up in office hours this week and solicit feedback from the user community about how to implement this. I can definitely see a reason why a user would prefer their own "resource health" instrumentation (perhaps there are other vectors for latency given how much metrics tooling has the opportunity to touch every network packet). And so this would be a case where including it in the API does make sense (it would be a durable configuration interface).

But let's ask and see what users think.

cc @mtougeron @zmalik @sonasingh46 @luthermonson

CecileRobertMichon: > I can definitely see a reason why a user would prefer their own "resource health" instrumentation

I don't fully understand this comment, why would one prevent you from doing the other? The `Status` field is just extra information, once AKS fixes the initial 10 minute initial API response of `Unknown` this would be useful information for any user, regardless of network stack or metric tooling as far as I can tell.

IMO a feature flag would be a better fit, in this case, at least to start, and then we can decide to move it into the API later if we see a need. Doing the reverse would be much harder and breaking. If we do want to move it into a durable API user exposed Spec field, we also want to do this with a durable API design, potentially add a new struct so we don't later end up with a flat list of toggles in the CRD `Spec`.

nojnhuh: @mtougeron @zmalik @sonasingh46 @luthermonson Any thoughts on the above? The tl;dr so far is:
- Azure has a generic "Resource Health" service that AKS hooks into to report certain cluster issues via the Azure API.
- The goal of this PR is to report this same info in the `status` of the AzureManagedControlPlane.
- There's a quirk in Azure or AKS where a harmless (as far as we know) error gets reported for the first ~20 minutes after a cluster is fully provisioned, blocking the `Ready` condition on the AzureManagedControlPlane for that period of time. When we mentioned the delay issue to AKS, they thought 20 minutes is definitely worth looking into improving, but I got the sense that it would still likely end up being some single-digit number of minutes at best.

So the questions we're looking to answer are:
- Is this feature something we need to have a way to disable?
- If so, how should it be disabled? The two options presented so far are via the API in something like the AzureManagedControlPlane's `spec.azureAvailabilityStatusEnabled` or via a feature flag.

mtougeron: Sorry, I missed this the other week when I was first flagged...

I think I agree with @CecileRobertMichon where this should be a feature flag rather than an API field. I don't think it is something that should be set on the cluster level to override what Azure is telling CAPZ. I think it's more appropriate to tell CAPZ how to handle the response/status from Azure. IMO, it's more of a change of application behavior than a change of cluster definition.

azure/services/resourcehealth/resourcehealth.go
```diff
@@ -0,0 +1,139 @@
...
+
+	return nil
+}
+
+// IsManaged always returns true.
+func (s *Service) IsManaged(ctx context.Context) (bool, error) {
+	return true, nil
+}
+
+func azureAvailabilityStatusToCondition(avail resourcehealth.AvailabilityStatus) *clusterv1.Condition {
```
CecileRobertMichon, line 107: this helper could probably go in the converters package

nojnhuh: Done!

azure/scope/managedcontrolplane.go
```diff
@@ -655,3 +657,23 @@ func (s *ManagedControlPlaneScope) TagsSpecs() []azure.TagsSpec {
...
+// AvailabilityStatusResourceURI constructs the ID of the underlying AKS resource.
+func (s *ManagedControlPlaneScope) AvailabilityStatusResourceURI() string {
+	return azure.ManagedClusterID(s.SubscriptionID(), s.ResourceGroup(), s.ControlPlane.Name)
+}
+
+// AvailabilityStatusFilter ignores the health metrics connection error that
+// occurs on startup for every AKS cluster.
+func (s *ManagedControlPlaneScope) AvailabilityStatusFilter(cond *clusterv1.Condition) *clusterv1.Condition {
+	if time.Since(s.ControlPlane.CreationTimestamp.Time) < 1*time.Hour &&
+		strings.Contains(cond.Message, "We've temporarily lost connection to the health metrics of this AKS cluster.") {
```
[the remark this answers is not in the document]
nojnhuh: Yeah, that seems reasonable. Just made that change.

azure/converters/resourcehealth.go
```diff
@@ -0,0 +1,61 @@
...
+	"strings"
+
+	"github.com/Azure/azure-sdk-for-go/services/resourcehealth/mgmt/2020-05-01/resourcehealth"
+	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"sigs.k8s.io/cluster-api/util/conditions"
+)
+
+// SDKAvailabilityStatusToCondition converts an Azure Resource Health availability status to a status condition.
+func SDKAvailabilityStatusToCondition(avail resourcehealth.AvailabilityStatus) *clusterv1.Condition {
```
willie-yao, line 29: This is pretty subjective, but I think a name like `status` might be a bit more descriptive of the variable than `avail`.

nojnhuh:
@CecileRobertMichon Thanks for taking a look so far! I've been keeping the old commits around just in case we wanted to go back to adding an API field or feature flag to disable this behavior, but the current approach of ignoring that specific error seems to be working well. Should I go ahead and squash now before asking for LGTMs?

exp/api/v1beta1/azuremanagedcontrolplane_types.go
zmalik: sorry, I also missed this 
This is actually great as we would like to see it in the CR. And toggle feature is also a good way. The reasonings is that Azure Resource Health also counts in for the API requests which adds up to the total limit of requests. We are trying to reduce that as we end up hitting subscription limits in some cases. Having that toggle available would be helpful. 

As feature flag or inside spec, I don't have strong opinions. But would be really helpful as having a separate path such as Azure Metrics Exporter would allow us to keep this disabled. To not to do same query twice.

jackfrancis: +1 to the feature flag.

I suppose the right thing to do is to turn this off by default. @CecileRobertMichon do we have an existing pattern where we would introduce a feature flag default-enabled?

CecileRobertMichon: we don't have an existing one that on by default but the way you'd do that it to set the environment variable default value to true in manager.yaml https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/config/manager/manager.yaml#L26

nojnhuh: Sounds good, I'll plan to add the feature flag back. I think that would also give us a path toward moving the toggle into the API if we find that should be a more permanent option.

Is it worth still explicitly ignoring this health metrics error we see for every cluster created even when the feature is enabled?

azure/converters/resourcehealth.go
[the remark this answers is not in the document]
nojnhuh: I think I was trying to distinguish this with the `status` field on the resources, but something like `availStatus` might be a bit more clear while still keeping that distinction.

exp/api/v1beta1/azuremanagedcontrolplane_types.go
CecileRobertMichon: +1

azure/converters/resourcehealth.go
```diff
@@ -0,0 +1,61 @@
...
+
+	if state == resourcehealth.AvailabilityStateValuesAvailable {
+		return conditions.TrueCondition(infrav1.AzureResourceAvailableCondition)
+	}
+
+	var reason string
+	if availStatus.Properties.ReasonType != nil {
+		// CAPI specifies Reason should be CamelCase, though the Azure API
+		// response may include spaces (e.g. "Customer Initiated")
+		reason = strings.ReplaceAll(*availStatus.Properties.ReasonType, " ", "")
```
jackfrancis, line 44: Sounds like we trust that the Azure API will always capitalize the words in the `ReasonType` string, or should we pre-capitalize before we strip out the `" "`?

azure/scope/managedcontrolplane.go
```diff
@@ -655,3 +657,23 @@ func (s *ManagedControlPlaneScope) TagsSpecs() []azure.TagsSpec {
...
+
+// AvailabilityStatusResourceURI constructs the ID of the underlying AKS resource.
+func (s *ManagedControlPlaneScope) AvailabilityStatusResourceURI() string {
+	return azure.ManagedClusterID(s.SubscriptionID(), s.ResourceGroup(), s.ControlPlane.Name)
+}
+
+// AvailabilityStatusFilter ignores the health metrics connection error that
+// occurs on startup for every AKS cluster.
+func (s *ManagedControlPlaneScope) AvailabilityStatusFilter(cond *clusterv1.Condition) *clusterv1.Condition {
+	if time.Since(s.ControlPlane.CreationTimestamp.Time) < 1*time.Hour &&
```
jackfrancis, line 674: nit, let's change this `1*time.Hour` to a const like `resourceHealthEventualConsistencyTolerance` or something

jackfrancis:
overall lgtm, left a couple comments

jackfrancis:
Due to the added time that this feature requires (and our quickly growing list of E2E test scenarios), I opened this in test-infra to bump the test timeout to 4h (I think I'm bumping it? Maybe the default is higher, in any event for tests that we expect to run for a while I think it's more appropriate to declare an explicit, long timeout.)

https://github.com/kubernetes/test-infra/pull/28256

willie-yao (approved):
/lgtm

Thanks for all the great work on this feature!

CecileRobertMichon:
> Test started last Tuesday at 11:57 AM passed after 1h43m12s

> Due to the added time that this feature requires (and our quickly growing list of E2E test scenarios), I opened this in test-infra to bump the test timeout to 4h

That seems quite long for a required PR test. Right now it's optional and only runs for AKS exp changes but as we migrate the feature out of experimental and make this test required on every PR, we may want to think about how we can run a "lean" subset of the test in PR and keep the more lengthy (eg. upgrade) tests as periodic jobs + optional PR jobs that only trigger when the AKS codepaths get touched.

azure/converters/resourcehealth.go
nojnhuh: That's all I've seen, but I can't find any docs that describe what the possible values here are, so I think it's best to assume it's basically the wild west. I just added that.

jackfrancis (approved):
/lgtm
/approve

PR #2741, opened 2022-10-20
common funcs webhook immutable properties

jackfrancis:
**What type of PR is this?**

/kind cleanup

This PR introduces common property immutability convenience functions, and updates the scattershot implementations of immutability to use these common funcs.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/util/webhook/validator.go
+++ b/util/webhook/validator.go
@@ -20,13 +20,21 @@ import (
 	"context"
 	"errors"
 	"net/http"
+	"reflect"
 
 	admissionv1 "k8s.io/api/admission/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/validation/field"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+const (
+	unsetMessage     = "field is immutable, unable to set an empty value if it was already set"
+	setMessage       = "field is immutable, unable to assign a value if it was already empty"
+	immutableMessage = "field is immutable"
 )
 
 // Validator defines functions for validating an operation.
@@ -135,4 +143,27 @@ func validationResponseFromStatus(allowed bool, status metav1.Status) admission.
 			Result:  &status,
 		},
 	}
+}
+
+// ValidateImmutable validates equality across two values,
+// and returns a meaningful error to indicate a changed value, a newly set value, or a newly unset value.
+func ValidateImmutable(path *field.Path, oldVal, newVal any) *field.Error {
+	if reflect.TypeOf(oldVal) != reflect.TypeOf(newVal) {
+		return field.Invalid(path, newVal, "unexpected error")
+	}
+	if !reflect.ValueOf(oldVal).IsZero() {
+		// Prevent modification if it was already set to some value
+		if reflect.ValueOf(newVal).IsZero() {
+			// unsetting the field is not allowed
+			return field.Invalid(path, newVal, unsetMessage)
+		}
+		if !reflect.DeepEqual(oldVal, newVal) {
+			// changing the field is not allowed
+			return field.Invalid(path, newVal, immutableMessage)
+		}
+	} else if !reflect.ValueOf(newVal).IsZero() {
+		return field.Invalid(path, newVal, setMessage)
+	}
+
+	return nil
 }
--- a/util/webhook/validator_test.go
+++ b/util/webhook/validator_test.go
@@ -1,0 +1,241 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package webhook
+
+import (
+	"testing"
+
+	"github.com/Azure/go-autorest/autorest/to"
+	. "github.com/onsi/gomega"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+)
+
+func TestValidateImmutableBoolPtr(t *testing.T) {
+	testPath := field.NewPath("Spec", "Foo")
+
+	tests := []struct {
+		name           string
+		input1         *bool
+		input2         *bool
+		expectedOutput *field.Error
+	}{
+		{
+			name:   "nil",
+			input1: nil,
+			input2: nil,
+		},
+		{
+			name:   "no change",
+			input1: to.BoolPtr(true),
+			input2: to.BoolPtr(true),
+		},
+		{
+			name:           "can't unset",
+			input1:         to.BoolPtr(true),
+			input2:         nil,
+			expectedOutput: field.Invalid(testPath, nil, unsetMessage),
+		},
+		{
+			name:           "can't set from empty",
+			input1:         nil,
+			input2:         to.BoolPtr(true),
+			expectedOutput: field.Invalid(testPath, nil, setMessage),
+		},
+		{
+			name:           "can't change",
+			input1:         to.BoolPtr(true),
+			input2:         to.BoolPtr(false),
+			expectedOutput: field.Invalid(testPath, nil, immutableMessage),
+		},
+	}
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			g := NewWithT(t)
+			err := ValidateImmutable(testPath, tc.input1, tc.input2)
+			if tc.expectedOutput != nil {
+				g.Expect(err).To(HaveOccurred())
+				g.Expect(err.Detail).To(Equal(tc.expectedOutput.Detail))
+				g.Expect(err.Type).To(Equal(tc.expectedOutput.Type))
+				g.Expect(err.Field).To(Equal(tc.expectedOutput.Field))
+			} else {
+				g.Expect(err).NotTo(HaveOccurred())
+			}
+		})
+	}
+}
+
+func TestValidateImmutableString(t *testing.T) {
+	testPath := field.NewPath("Spec", "Foo")
+
+	tests := []struct {
+		name           string
+		input1         string
+		input2         string
+		expectedOutput *field.Error
+	}{
+		{
+			name:   "empty string",
+			input1: "",
+			input2: "",
+		},
+		{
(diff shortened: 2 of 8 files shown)


CecileRobertMichon (commented):
/lgtm
/assign @nojnhuh

util/webhook/validator_test.go
```diff
@@ -0,0 +1,265 @@
...
+import (
+	"testing"
+
+	"github.com/Azure/go-autorest/autorest/to"
+	. "github.com/onsi/gomega"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+)
+
+func TestValidateBoolPtrImmutable(t *testing.T) {
+	g := NewWithT(t)
```
nojnhuh, line 28: I tend to put this inside the subtest so the `t` it's passed is scoped to only that subtest. That way when the test fails, the error that shows up is a bit more clear since it includes the subtest's name. Gomega gives a message alluding to this and stops execution before getting to the rest of the tests.

e.g. If I make some change to force some tests to fail:
```
% go test ./util/webhook
--- FAIL: TestValidateBoolPtrImmutable (0.00s)
    validator_test.go:81:
        Unexpected error:
            ...
    --- FAIL: TestValidateBoolPtrImmutable/nil (0.00s)
        testing.go:1343: test executed panic(nil) or runtime.Goexit: subtest may have called FailNow on a parent test
FAIL
FAIL    sigs.k8s.io/cluster-api-provider-azure/util/webhook     0.025s
FAIL
```
And then after I move the `g := NewWithT(t)`:
```
% go test ./util/webhook
--- FAIL: TestValidateBoolPtrImmutable (0.00s)
    --- FAIL: TestValidateBoolPtrImmutable/nil (0.00s)
        validator_test.go:81:
            Unexpected error:
                <*field.Error | 0xc00044ff40>: {
                    Type: "FieldValueInvalid",
                    Field: "Spec.Foo",
                    BadValue: <*bool | 0x0>nil,
                    Detail: "field is immutable, unable to assign a value if it was already empty",
                }
                Spec.Foo: Invalid value: "null": field is immutable, unable to assign a value if it was already empty
            occurred
    --- FAIL: TestValidateBoolPtrImmutable/no_change (0.00s)
        validator_test.go:81:
            Unexpected error:
                <*field.Error | 0xc0002fa040>: {
                    Type: "FieldValueInvalid",
                    Field: "Spec.Foo",
                    BadValue: <*bool | 0x0>nil,
                    Detail: "field is immutable, unable to assign a value if it was already empty",
                }
                Spec.Foo: Invalid value: "null": field is immutable, unable to assign a value if it was already empty
            occurred
    --- FAIL: TestValidateBoolPtrImmutable/can't_set_from_empty (0.00s)
        validator_test.go:76:
            Expected an error to have occurred.  Got:
                <*field.Error | 0x0>: nil
FAIL
FAIL    sigs.k8s.io/cluster-api-provider-azure/util/webhook     0.023s
FAIL
```

util/webhook/validator_test.go
```diff
@@ -0,0 +1,265 @@
...
+			path:   testPath,
+			input1: nil,
+			input2: nil,
+		},
+		{
+			name:   "no change",
+			path:   testPath,
+			input1: nil,
+			input2: nil,
+		},
```
nojnhuh, line 49: Since this test case is the same as the one above it, I wonder if setting the inputs to the same non-nil value might provide a bit more value.

util/webhook/validator_test.go
```diff
@@ -0,0 +1,265 @@
...
+	"k8s.io/apimachinery/pkg/util/validation/field"
+)
+
+func TestValidateBoolPtrImmutable(t *testing.T) {
+	g := NewWithT(t)
+	testPath := field.NewPath("Spec", "Foo")
+
+	tests := []struct {
+		name           string
+		path           *field.Path
```
nojnhuh, line 33: If this is the same for all tests, I suppose it could be inlined if not having to add this to each test case would be worth it.

nojnhuh (commented):
Making use of these utils is definitely going to make the webhooks easier to work with, so thank you for doing this!

I found a few more places where it looks like we could apply these helpers. Would it make sense to update these too?
https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/d89ad7c6d35a44b3b2d583814a81a5f8576111ad/api/v1beta1/azurecluster_webhook.go#L102
https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/d89ad7c6d35a44b3b2d583814a81a5f8576111ad/api/v1beta1/azurecluster_webhook.go#L110
https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/d89ad7c6d35a44b3b2d583814a81a5f8576111ad/exp/api/v1beta1/azuremanagedmachinepool_webhook.go#L111

The checks in these functions:
https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/d89ad7c6d35a44b3b2d583814a81a5f8576111ad/exp/api/v1beta1/azuremanagedcontrolplane_webhook.go#L96
https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/d89ad7c6d35a44b3b2d583814a81a5f8576111ad/exp/api/v1beta1/azuremanagedcontrolplane_webhook.go#L488
https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/d89ad7c6d35a44b3b2d583814a81a5f8576111ad/api/v1beta1/azurecluster_webhook.go#L142

util/webhook/validator.go
```diff
@@ -136,3 +144,82 @@ func validationResponseFromStatus(allowed bool, status metav1.Status) admission.
 		},
 	}
 }
+
+// EnsureStringSlicesAreEquivalent returns if two string slices have equal lengths,
+// and that they have the exact same items; it does not enforce strict ordering of items.
+func EnsureStringSlicesAreEquivalent(a []string, b []string) bool {
```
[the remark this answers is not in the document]
jackfrancis: I think because we don't want to include any functional changes in this PR I'll do this in a follow-up PR (I'll remove the new, still-unused `EnsureStringSlicesAreEquivalent` func from this PR).

util/webhook/validator_test.go
jackfrancis: fair

util/webhook/validator_test.go
jackfrancis: 💯

util/webhook/validator_test.go
jackfrancis: Thanks! updated

util/webhook/validator.go
```diff
@@ -136,3 +144,82 @@ func validationResponseFromStatus(allowed bool, status metav1.Status) admission.
...
+	for i := range a {
+		if a[i] != b[i] {
+			return false
+		}
+	}
+
+	return true
+}
+
+// ValidateBoolPtrImmutable validates equality across two *bools,
```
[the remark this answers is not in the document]
jackfrancis: I did an audit and included all these validations, and fixed up a new, single `ValidateImmutable` func that is not type-specific (though it does have some limited type-awareness for the purposes of evaluating "unset value --> set value" or "set value --> unset value" specific errors.

util/webhook/validator.go
```diff
@@ -136,3 +144,43 @@ func validationResponseFromStatus(allowed bool, status metav1.Status) admission.
 		},
 	}
 }
+
+// ValidateImmutable validates equality across two values,
+// and returns a meaningful error to indicate a changed value, a newly set value, or a newly unset value.
+func ValidateImmutable(path *field.Path, oldVal, newVal any) *field.Error {
```
nojnhuh, line 150: This is definitely much less verbose and repetitive than having separate funcs for every type, though I can see where this introduces some foot-guns like using this for a not-explicitly-listed type and reasoning about how fields of something like `type VMIdentity string` behave.

Overall I could be convinced to start looking into generics for this function since I think that would address most of those concerns.

util/webhook/validator.go
```diff
@@ -136,3 +144,43 @@ func validationResponseFromStatus(allowed bool, status metav1.Status) admission.
...
+		var t int32
+		unsetVal = t
+	case string:
+		var t string
+		unsetVal = t
+	case *bool:
+		var t *bool
+		unsetVal = t
+	default:
+		unsetVal = nil
```
nojnhuh, line 169: For any type not explicitly listed here, I think every comparison of equality to `unsetVal` will be false because a nil slice (or any typed `nil`) is not equivalent to untyped `nil`. I think this example demonstrates this: https://go.dev/play/p/0UQbIuroQeh

Overall I don't think this is causing any major problems since the `reflect.DeepEqual` call is catching the type discrepancy, but it does hide the "can't set previously unset field"/"can't unset previously set field" context.

I tried modifying the AzureMachine webhook tests to print the exact error being returned and found that to be the case. Removing this definition of a string yields `field is immutable, unable to assign a value if it was already empty`:
https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/d77c067f5542624a0de5d0b8c0288d0f7fbc74fa/api/v1beta1/azuremachine_webhook_test.go#L232
And removing this definition of a struct pointer yields only `field is immutable`:
https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/d77c067f5542624a0de5d0b8c0288d0f7fbc74fa/api/v1beta1/azuremachine_webhook_test.go#L132-L134

jackfrancis: These observations match mine. For types that we don't explicitly have a case for, we're essentially wrapping `reflect.DeepEqual`. Strictly speaking this adds a bit of overhead when we change to this new func, with the benefit of using a common boilerplate pattern.

util/webhook/validator.go
jackfrancis: I'd love to see the go generic equivalent for this!

util/webhook/validator.go
nojnhuh: I think that's reasonable to essentially default to `reflect.DeepEqual`. If we can boil this down into a comment that makes sense I think that would be worth adding.

util/webhook/validator.go
nojnhuh: For reference, I took a pass at what that might look like and came up with this:
```go
func ValidateImmutable[T any](path *field.Path, oldVal, newVal T) *field.Error {
	var unsetVal T
	if !reflect.DeepEqual(oldVal, unsetVal) {
		// Prevent modification if it was already set to some value
		if reflect.DeepEqual(newVal, unsetVal) {
			// unsetting the field is not allowed
			return field.Invalid(path, newVal, unsetMessage)
		}
		if !reflect.DeepEqual(oldVal, newVal) {
			// changing the field is not allowed
			return field.Invalid(path, newVal, immutableMessage)
		}
	} else if !reflect.DeepEqual(newVal, unsetVal) {
		return field.Invalid(path, newVal, setMessage)
	}

	return nil
}
```

nojnhuh: Actually, [`reflect.IsZero()`](https://pkg.go.dev/reflect#Value.IsZero) might make this just as easy but without generics at all.

jackfrancis: IsZero ftw, thanks for the suggestion!

CecileRobertMichon (commented):
/approve

PR #2743, opened 2022-10-20
Add AzureMachine condition when userAssignedIdentity is missing from the VM

### LINKED ISSUE #2319: Add AzureMachine condition when userAssignedIdentity is missing from the VM
/kind bug

Deployed an `AzureMachine` via `AzureMachineTemplate` with:

```yaml
kind: AzureMachineTemplate
metadata:
  name: {name}
  namespace: {namespace}
spec:
  template:
    spec:
      vmSize: Standard_DS3_v2
      osDisk:
        diskSizeGB: 128
        osType: Linux
        cachingType: ReadOnly
      dataDisks:
      - diskSizeGB: 256
        lun: 0
        nameSuffix: etcddisk
        managedDisk:
          storageAccountType: StandardSSD_LRS
      sshPublicKey: {sshPublicKey}
      identity: UserAssigned
      userAssignedIdentities:
        - providerID: azure:///subscriptions/{subscription_id}/resourceGroups/{resource_group_name}/providers/Microsoft.ManagedIdentity/userAssignedIdentities/{managed_indentity_name}
```

On deployment the pre-created user-assigned identity gets attached to the VM's.
When removing the attachment the managed identity doesn't automatically get re-attached.

The `capz-controller` should reconcile the desired state by re-attaching the managed identity.

**Environment:**

- cluster-api-provider-azure version: v1.3.1
- Kubernetes version: (use `kubectl version`): 1.22.8
- OS (e.g. from `/etc/os-release`): Ubuntu 20.04.4 LTS (Focal Fossa)

willie-yao:
**What type of PR is this?**
/kind feature

When a `userAssignedIdentity` is removed, it doesn't automatically get re-attached to the VM. Since Machines are immutable in CAPI, this PR adds an `AzureMachine` unhealthy condition that triggers a `MachineHealthCheck` when the `userAssignedIdentity` does not match the spec.

Fixes #2319 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Add "VMIdentitiesReadyCondition" AzureMachine condition
```


CecileRobertMichon (commented):
/assign @nojnhuh

nojnhuh (commented):
Left a few questions and small comments inline, but nothing I would fight to block this on. Looks good!

azure/services/virtualmachines/virtualmachines.go
```diff
@@ -47,6 +49,8 @@ type VMScope interface {
 	SetProviderID(string)
 	SetAddresses([]corev1.NodeAddress)
 	SetVMState(infrav1.ProvisioningState)
+	SetConditionTrue(clusterv1.ConditionType)
```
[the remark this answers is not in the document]
willie-yao: That makes sense! It was leftover from when I tried to set a `VMUnhealthy` condition to True instead of setting `VMRunning` to false. I'll go ahead and remove it!

controllers/azuremachine_controller.go
```diff
@@ -284,6 +284,15 @@ func (amr *AzureMachineReconciler) reconcileNormal(ctx context.Context, machineS
 		return reconcile.Result{}, errors.Wrap(err, "failed to init machine scope cache")
 	}
 
+	// If a node is unhealthy, trigger a MachineHealthCheck to replace the node.
```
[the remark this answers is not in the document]
willie-yao: Setting up the MachineHealthCheck is an extra step that you have to add in the Spec. This [page](https://cluster-api.sigs.k8s.io/tasks/automated-machine-management/healthchecking.html?highlight=healthchecking) in the CAPI book gives a good overview on how it works.

controllers/azuremachine_controller.go
```diff
@@ -284,6 +284,15 @@ func (amr *AzureMachineReconciler) reconcileNormal(ctx context.Context, machineS
...
+	cond := conditions.Get(machineScope.AzureMachine, infrav1.VMRunningCondition)
+	if cond != nil && cond.Status == corev1.ConditionFalse && cond.Reason == infrav1.VMUnhealthyReason {
+		amr.Recorder.Eventf(machineScope.AzureMachine, corev1.EventTypeWarning, "VMUnhealthy", "VM is unhealthy")
```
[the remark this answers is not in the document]
willie-yao: Nice catch! That is much better than using the hard-coded string.

api/v1beta1/conditions_consts.go
```diff
@@ -38,6 +38,8 @@ const (
 	VMDeletingReason = "VMDeleting"
 	// VMProvisionFailedReason used for failures during vm provisioning.
 	VMProvisionFailedReason = "VMProvisionFailed"
+	// VMUnhealthyReason used for failures when the VM is in an unhealthy state.
+	VMUnhealthyReason = "VMUnhealthy"
```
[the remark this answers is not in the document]
willie-yao: Yes that's the reason! At first, I created a VMUnhealthy condition, but realized that it would be redundant with the VMRunning condition, since when the VM is unhealthy, we would want to replace it with another machine anyways. In order to do that we would need to set the VMRunning condition to false. I think the general rule is that if the reason doesn't fall under a pre-existing condition, we would want to create a new one. @CecileRobertMichon would you be able to confirm that my understanding is correct?

On a separate note, there is another rule with conditions saying that all of them must be positive, meaning VMUnhealthy would break that rule. If we did want to add another condition, we would name it VMHealthy instead.

azure/services/virtualmachines/virtualmachines.go
```diff
@@ -111,6 +115,34 @@ func (s *Service) Reconcile(ctx context.Context) error {
...
+
+		// Check if any userAssignedIdentities are missing from the VM and set a condition if so
+		if len(spec.UserAssignedIdentities) == len(infraVM.UserAssignedIdentities) {
+			for _, expectedIdentity := range spec.UserAssignedIdentities {
+				expectedClientID, err := idsClient.GetClientID(ctx, expectedIdentity.ProviderID)
+				if err != nil {
+					return errors.Wrap(err, "failed to get client ID")
+				}
+				found := false
+				for _, actualIdentity := range infraVM.UserAssignedIdentities {
```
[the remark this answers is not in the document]
willie-yao: I'm actually not sure about this one. I had to use a nested for loop since we can't assume the userAssignedIdentities would be in the same order between the vm and the machine. I think I may be able to optimize this further using a map. cc @CecileRobertMichon on whether this might be too inefficient.

azure/converters/vm.go
```diff
@@ -63,5 +65,13 @@ func SDKToVM(v compute.VirtualMachine) *VM {
 		vm.Tags = MapToTags(v.Tags)
 	}
 
+	if v.Identity != nil {
+		for _, identity := range v.Identity.UserAssignedIdentities {
+			vm.UserAssignedIdentities = append(vm.UserAssignedIdentities, infrav1.UserAssignedIdentity{
+				ProviderID: *identity.ClientID,
```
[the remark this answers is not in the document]
willie-yao: A basic Go question: would I also have to check if `v.Identity.UserAssignedIdentities` is nil before iterating over it?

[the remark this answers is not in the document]
nojnhuh: Nope, `range` over a nil slice won't panic: https://go.dev/play/p/2Tkns7WlT_9

api/v1beta1/conditions_consts.go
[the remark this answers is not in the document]
CecileRobertMichon: > I think the general rule is that if the reason doesn't fall under a pre-existing condition, we would want to create a new one.

That's correct.

In this case though, I'm leaning towards adding a new condition since the user-assigned identity doesn't really affect the VM running state (the user would still see the VM as `Running` in the Azure Portal). I also wouldn't call the VM "unhealthy", but maybe more something specific like `VMIdentitiesReady` with a reason `UserAssignedIdentityMissing` or something like that.

controllers/azuremachine_controller.go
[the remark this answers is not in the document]
CecileRobertMichon: This comment might be a bit confusing since the code itself isn't triggering a MachineHealthCheck and that needs to be set up separately as you point out. This also isn't checking for the node health, but the VM conditions. How about something like:

```suggestion
	// Mark the AzureMachine as failed if any conditions are false.
```

azure/services/virtualmachines/virtualmachines.go
```diff
@@ -111,6 +114,34 @@ func (s *Service) Reconcile(ctx context.Context) error {
...
 		s.Scope.SetAddresses(addresses)
 		s.Scope.SetVMState(infraVM.State)
+
+		spec, ok := vmSpec.(*VMSpec)
+		if !ok {
+			return errors.Errorf("%T is not a valid VM spec", vmSpec)
+		}
+		idsClient := identities.NewClient(s.Scope)
+
+		// Check if any userAssignedIdentities are missing from the VM and set a condition if so
```
CecileRobertMichon, line 124: could we extract the logic checking for missing identities into a helper function and write a focused unit test for it?

azure/services/virtualmachines/virtualmachines.go
```diff
@@ -111,6 +114,34 @@ func (s *Service) Reconcile(ctx context.Context) error {
...
+				}
+				found := false
+				for _, actualIdentity := range infraVM.UserAssignedIdentities {
+					if expectedClientID == actualIdentity.ProviderID {
+						found = true
+						break
+					}
+				}
+				if !found {
+					s.Scope.SetConditionFalse(infrav1.VMRunningCondition, infrav1.VMUnhealthyReason, clusterv1.ConditionSeverityWarning, "VM is missing expected user assigned identity")
```
CecileRobertMichon, line 139: What would you think about including the identity name or resource ID in the message to help the user know which one is missing?

azure/services/virtualmachines/virtualmachines.go
```diff
@@ -111,6 +114,34 @@ func (s *Service) Reconcile(ctx context.Context) error {
 		}
 		s.Scope.SetAddresses(addresses)
 		s.Scope.SetVMState(infraVM.State)
+
+		spec, ok := vmSpec.(*VMSpec)
+		if !ok {
+			return errors.Errorf("%T is not a valid VM spec", vmSpec)
+		}
+		idsClient := identities.NewClient(s.Scope)
```
CecileRobertMichon, line 122: here instead of creating a client on the fly, let's make it part of the Service, see how we do it above for `publicIPsGetter`

azure/services/virtualmachines/virtualmachines.go
willie-yao: I think that would be great! I do have a check for the length of the two lists, and if they aren't equal, I immediately mark the condition as fault. However, in this case, I don't know which identity is missing. Should I remove this check and favor knowing which identity is missing? Or should we prioritize the efficiency here.

willie-yao: Here's the [line](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/5c806269df45b9a25682784ae275fef46088874a/azure/services/virtualmachines/virtualmachines.go#L142) I'm talking about

azure/services/virtualmachines/virtualmachines.go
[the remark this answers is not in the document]
CecileRobertMichon: I wasn't able to find any public docs about how many identities you can have assigned to a single VM but I agree that we don't expect more than a few per VM

azure/services/virtualmachines/virtualmachines.go
CecileRobertMichon: Since we don't expect more than a few identities per VM I think I'm leaning towards optimizing for UX here and always printing the name of the missing one. Maybe this is also good motivation to optimize the for loops as you pointed out above https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2743#discussion_r1002207518

willie-yao:
@CecileRobertMichon I refactored the [`checkUserAssignedIdentities` method](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2743/commits/2a2872dca02b64495a529199fda3abeca75e18a5#diff-bd4c935d5db79644e733f4f321bda2011252db8da4cc4bdfef5fffe3e5cd0710R156) to be slightly more efficient. I also realized the previous function didn't account for the vm having more userAssignedIdentities than the spec, so now I'm doing a separate check for that. Is that the expected behavior, such as when a user adds a userAssignedIdentity from the portal manually?

CecileRobertMichon:
> I also realized the previous function didn't account for the vm having more userAssignedIdentities than the spec, so now I'm doing a separate check for that

@willie-yao great catch, we actually don't want to set the VM as failed if there are extras, just if one of the ones defined in CAPZ is missing.

azure/services/virtualmachines/virtualmachines.go
```diff
@@ -138,6 +153,35 @@ func (s *Service) Delete(ctx context.Context) error {
...
+	}
+
+	// Create a map of the actual identities from the vm.
+	for _, actualIdentity := range vmIdentities {
+		actualMap[actualIdentity.ProviderID]++
+	}
+
+	// Check if the expected identities are present in the vm.
+	for expectedKey, expectedVal := range expectedMap {
+		if actualMap[expectedKey] != expectedVal {
```
nojnhuh, line 176: I'm not familiar with the semantics around specifying the same identity more than once (or if that's even possible), so this is kind of a spitball, but would it be acceptable to have the same identity listed more times in the actual VM than in the CAPZ spec?
```suggestion
		if actualMap[expectedKey] < expectedVal {
```

azure/services/virtualmachines/virtualmachines_test.go
```diff
@@ -283,3 +291,78 @@ func TestDeleteVM(t *testing.T) {
...
+		{
+			name:             "mismatched user assigned identities by content",
+			specIdentities:   []infrav1.UserAssignedIdentity{fakeUserAssignedIdentity},
+			actualIdentities: []infrav1.UserAssignedIdentity{fakeUserAssignedIdentity2},
+			expect: func(s *mock_virtualmachines.MockVMScopeMockRecorder, i *mock_identities.MockClientMockRecorder) {
+				i.GetClientID(gomockinternal.AContext(), fakeUserAssignedIdentity.ProviderID).AnyTimes().Return(fakeUserAssignedIdentity.ProviderID, nil)
+				s.SetConditionFalse(infrav1.VMIdentitiesReadyCondition, infrav1.UserAssignedIdentityMissingReason, clusterv1.ConditionSeverityWarning, "VM is missing expected user assigned identity with client ID: "+fakeUserAssignedIdentity.ProviderID).Times(1)
+			},
+		},
+	}
```
nojnhuh, line 345: If the same identity could be listed more than once in either the VM or CAPZ spec, then adding some test cases like that would be worthwhile.

azure/services/virtualmachines/virtualmachines.go
willie-yao: I don't think it's allowed to have the same identity more than once, but I implemented this using maps for efficiency. However, it would make more sense to use `<` since we are always seeing if the identity is missing.

azure/services/virtualmachines/virtualmachines_test.go
willie-yao: I definitely can't add the same identity twice on the portal, but I'll try specifying the same one in the spec and see if this is something we need to test.

azure/services/virtualmachines/virtualmachines.go
nojnhuh: If that's the case, I think using `map[string]struct{}` might make the semantics a bit more clear.

willie-yao: That makes sense! What would I add to the struct to indicate that the identity exists? Would I just add the identity itself?

nojnhuh: `struct{}` by itself is a complete type where `struct{}{}` is the only possible value. Then the identity is used as the key and can be checked with something like:
```go
_, exists := actualMap[expectedKey]
```

azure/services/virtualmachines/virtualmachines.go
```diff
@@ -138,6 +153,36 @@ func (s *Service) Delete(ctx context.Context) error {
...
 
+func (s *Service) checkUserAssignedIdentities(ctx context.Context, specIdentities []infrav1.UserAssignedIdentity, vmIdentities []infrav1.UserAssignedIdentity) error {
+	expectedMap := make(map[string]struct{})
+	actualMap := make(map[string]struct{})
+
+	// Create a map of the expected identities. The ProviderID is converted to match the format of the VM identity.
+	for _, expectedIdentity := range specIdentities {
+		expectedClientID, err := s.identitiesGetter.GetClientID(ctx, expectedIdentity.ProviderID)
+		if err != nil {
+			return errors.Wrap(err, "failed to get client ID")
```
nojnhuh, line 164: It's probably worth adding a quick test case exercising this code path for that one extra line of sweet, sweet code coverage (and the actual benefits of verifying and documenting the behavior in this scenario).

azure/services/virtualmachines/virtualmachines_test.go
willie-yao: Looks like specifying duplicate identities in the spec just results in one identity being listed on the VM. This won't cause any errors with the current implementation, but I'll go ahead and add a test case for it since it's possible to list them on the spec.

jackfrancis:
@willie-yao could you kindly squash commits? thx!

jackfrancis:
It seems like this is a bug. Should we classify as such and cherry-pick? cc @CecileRobertMichon @mboersma

CecileRobertMichon:
> It seems like this is a bug

I'd say it's in between a bug and an improvement, @willie-yao what do you think?

willie-yao:
I'm not sure. I think it's more of an improvement because we're just adjusting the intended behavior of the vm. I think it'll be nice to have for previous versions, but maybe some users of previous versions would expect their VMs to not be deleted. Would that be an issue if we cherry pick this?

jackfrancis:
@willie-yao We definitely don't want to be changing behaviors into a previous release branch as a patch release if those behavior changes aren't 100% beneficial. If some users may be undesirably surprised with these changes then they should be included with the next minor release (v1.7.0).

PR #2744, opened 2022-10-21
Fix UTC meeting time in README

zioproto:
README: 09:00 (PT) is the same time as 16:00 (UTC)

**Release note**:

```release-note
NONE
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ The community holds office hours every week, with sessions open to all users and
 developers.
 
 Office hours are hosted on a zoom video chat every Thursday
-at 09:00 (PT) / 12:00 (ET) / 17:00 (UTC),
+at 09:00 (PT) / 12:00 (ET) / 16:00 (UTC),
 and are published on the [Kubernetes community meetings calendar][gcal]. Please add your questions or ideas to [the agenda][capz_agenda].
 
 ### Other ways to communicate with the contributors


CecileRobertMichon:
Will this still be accurate when the US enters [Daylight Savings time](https://www.timeanddate.com/time/change/usa?year=2022) in a few weeks? I wonder if that's what the current times are based on

zioproto:
> Will this still be accurate when the US enters [Daylight Savings time](https://www.timeanddate.com/time/change/usa?year=2022) in a few weeks? I wonder if that's what the current times are based on


"PT" refers not to a time, but to a time zone, and can therefore be used (implicitly and unambiguously) to refer to the current time in the Pacific time zone. If you specify a time (rather than a time zone), you should use the correct one (PDT or PST).
UTC is also a fixed reference that does not change during the year with Daylight or Standard time changes.

zioproto:
This is a one line change in the README, so I guess the test failing is the CI currently broken ?

README.md
```diff
@@ -93,7 +93,7 @@ The community holds office hours every week, with sessions open to all users and
 developers.
 
 Office hours are hosted on a zoom video chat every Thursday
-at 09:00 (PT) / 12:00 (ET) / 17:00 (UTC),
+at 09:00 (PT) / 12:00 (ET) / 16:00 (UTC),
```
mboersma, line 96: > If you specify a time (rather than a time zone), you should use the correct one (PDT or PST).

So to be accurate, should we merge this change, then revert it on November 6 when Daylight Savings ends?

mboersma (commented):
/approve

This fixes something that's just wrong currently, so I'm merging it. We should adjust the README.md again after Nov. 6.

Issue #2611, opened 2022-08-29
Show resource diffs for AKS clusters in the capz-controller-manager logs when making updates


AAkindele:
/kind bug

[Before submitting an issue, have you checked the [Troubleshooting Guide](https://capz.sigs.k8s.io/topics/troubleshooting.html)?]

**What steps did you take and what happened:**

1. Create an AKS cluster following the documentation <https://capz.sigs.k8s.io/topics/managedcluster.html>.
1. Increase the capz-controller-manager deployment log level from 0 to 5 by changing the `--v` flag.
1. After the cluster has been created, make a change to the cluster outside of Cluster API.
    - ex: Add tags, `environment=testing` to the AKS cluster from the Azure portal.
1. Make a change to the AzureManagedControlPlane configuration yaml so the changes are different from the ones made in the portal.
    - ex: Add tags. `environment=development`
1. Wait till Cluster API updates the cluster
1. Review the capz-controller-manager logs to see that there are no log entries showing the diff between tags.

**What did you expect to happen:**

When looking at capz-controller-manager logs, there should be log entries when diffs are identified between current AKS cluster state and target state.

**Anything else you would like to add:**

Example of expected log entries from a different Azure resource.
<https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/85c1951c4f1cfdec5fabb5b5a5b9a865082421f0/azure/services/agentpools/agentpools.go#L157>

**Environment:**

- cluster-api-provider-azure version: 1.4.1
- Kubernetes version: (use `kubectl version`): 1.22.3
- OS (e.g. from `/etc/os-release`): Ubuntu 20.04.3 LTS

nawazkh:
@AAkindele Can you talk more about point 4? 
> Make a change to the AzureManagedControlPlane configuration yaml so the changes are different from the ones made in the portal.
ex: Add tags. environment=development

What did you mean by _Add tags. environment=development_? Did you add it to the annotations or labels of  AzureManagedControlPlane manifest?

AAkindele:
@nawazkh Apologies for the lack of details. I added the tags to AzureManagedControlPlane using `spec.additionalTags`.

Example below with most of the fields removed for brevity.

```yaml

apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
kind: AzureManagedControlPlane
metadata:
  name: capi-quickstart
spec:
  additionalTags:
    environment: development

```

[...]

PR #2745, opened 2022-10-21
Add AdditionalTags property to CRD AzureManagedMachinePool

### LINKED ISSUE #1778: Add additionalTags property to CRD AzureManagedMachinePool
***Goals***
1. Tag all resources

**User Story**

As a user I would like to tag all AKS-related resources.

**Detailed Description**

I noticed the property `additionalTags` is defined in most CRDs but not in `AzureManagedMachinePool`. Why? Doesn’t it make sense in this context? 
```
$ grep -rnw config/crd/ -e additionalTags
config/crd//bases/infrastructure.cluster.x-k8s.io_azuremachinetemplates.yaml:57:                      additionalTags:
config/crd//bases/infrastructure.cluster.x-k8s.io_azuremachinetemplates.yaml:407:                      additionalTags:
config/crd//bases/infrastructure.cluster.x-k8s.io_azuremachinetemplates.yaml:780:                      additionalTags:
config/crd//bases/infrastructure.cluster.x-k8s.io_azureclusters.yaml:66:              additionalTags:
config/crd//bases/infrastructure.cluster.x-k8s.io_azureclusters.yaml:463:              additionalTags:
config/crd//bases/infrastructure.cluster.x-k8s.io_azureclusters.yaml:1294:              additionalTags:
config/crd//bases/infrastructure.cluster.x-k8s.io_azuremachinepools.yaml:76:              additionalTags:
config/crd//bases/infrastructure.cluster.x-k8s.io_azuremachinepools.yaml:595:              additionalTags:
config/crd//bases/infrastructure.cluster.x-k8s.io_azuremachinepools.yaml:1350:              additionalTags:
config/crd//bases/infrastructure.cluster.x-k8s.io_azuremanagedcontrolplanes.yaml:62:              additionalTags:
config/crd//bases/infrastructure.cluster.x-k8s.io_azuremanagedcontrolplanes.yaml:222:              additionalTags:
config/crd//bases/infrastructure.cluster.x-k8s.io_azuremanagedcontrolplanes.yaml:527:              additionalTags:
config/crd//bases/infrastructure.cluster.x-k8s.io_azuremachines.yaml:77:              additionalTags:
config/crd//bases/infrastructure.cluster.x-k8s.io_azuremachines.yaml:535:              additionalTags:
config/crd//bases/infrastructure.cluster.x-k8s.io_azuremachines.yaml:1048:              additionalTags:
```

It confuses me since there is such property in `AWSManagedMachinePool`.

/kind proposal

zioproto:
This PR adds the AdditionalTags to AzureManagedMachinePool for consistency with other CRDs

Fixes #1778 

/kind feature
/kind api-change

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

Release note:

```release-note
Add AdditionalTags property to AzureManagedMachinePool 
```


config/crd/bases/infrastructure.cluster.x-k8s.io_azuremanagedmachinepools.yaml
```diff
@@ -199,6 +199,13 @@ spec:
             description: AzureManagedMachinePoolSpec defines the desired state of
               AzureManagedMachinePool.
             properties:
+              additionalTags:
```
zioproto, line 202: Is this file autogenerated or it is correct to update this yaml file ?

exp/api/v1beta1/azuremanagedmachinepool_types.go
```diff
@@ -44,6 +44,11 @@ type NodePoolMode string
 // AzureManagedMachinePoolSpec defines the desired state of AzureManagedMachinePool.
 type AzureManagedMachinePoolSpec struct {
 
+	// AdditionalTags AdditionalTags is an optional set of tags to add to Azure resources managed by the
```
zioproto, line 47: Is it correct to update just the `v1beta1` file ?
was I supposed to backport this to the `v1alpha*` versions ?
or was I supposed to create a `v1beta2` version when introducing this change ?

config/crd/bases/infrastructure.cluster.x-k8s.io_azuremanagedmachinepools.yaml
sonasingh46: It is auto generated. If you update it causing the autogen to give the same output -- the ci won't complain though.

But as a best practice it should be generated.
See `make generate` command in makefile.

exp/api/v1beta1/azuremanagedmachinepool_types.go
sonasingh46: As the APIs are still experimental, I think it should be fine to add in v1beta1 alone.
cc @mtougeron @jackfrancis to comment more on this one.

CecileRobertMichon: You'll want to change `v1beta1` types + conversion to and from alpha API versions

As a general rule, we only update the latest API version types on the main branch. The alpha types are there for backwards compatibility purposes (conversion to and from the old types).

exp/api/v1beta1/azuremanagedmachinepool_types.go
```diff
@@ -44,6 +44,11 @@ type NodePoolMode string
...
+	// Azure provider, in addition to the ones added by default.
+	// +optional
+	AdditionalTags []string `json:"additionalTags,omitempty"`
```
CecileRobertMichon, line 50: @zioproto you'll also need to use this field in the controller to make sure it's applied on the Azure resource. You can look at AzureMachine additionalTags for an example.

exp/api/v1beta1/azuremanagedmachinepool_types.go
CecileRobertMichon, line 47:
```suggestion
	// AdditionalTags is an optional set of tags to add to Azure resources managed by the
```

azure/services/agentpools/spec.go
```diff
@@ -31,6 +31,9 @@ import (
 
 // AgentPoolSpec contains agent pool specification details.
 type AgentPoolSpec struct {
+	// AdditionalTags is an optional set of tags to add to Azure resources managed by the Azure provider, in addition to the ones added by default.
+	AdditionalTags infrav1.Tags
```
zioproto, line 35: I changed this from my previous version using the `type` defined in `infrav1.Tags`.
Is this the correct approach ?

zioproto:
I tested on Linux the unit tests with:
```
make generate
make test
```
I also created a new unit test.

Can anyone check the conversion part ? I am not sure it is complete.

Would be also possible to enable the e2e test in the CI ?

Thanks

zioproto:
I am testing it on a dev cluster, but I dont see the tags added to the nodepools.

I think I missed to do some changes here:
https://github.com/zioproto/cluster-api-provider-azure/blob/issue-1778/azure/services/agentpools/spec.go#L128

In this block:
https://github.com/zioproto/cluster-api-provider-azure/blob/issue-1778/azure/services/agentpools/spec.go#L144-L168

and probably also here:
https://github.com/zioproto/cluster-api-provider-azure/blob/issue-1778/azure/services/agentpools/spec.go#L212-L236

zioproto:
I have done some more changes and now I tested that tags are actually added to the nodepools.

I tested like this:

```
minikube start
# Base64 encode the variables
export AZURE_SUBSCRIPTION_ID_B64="$(echo -n "$AZURE_SUBSCRIPTION_ID" | base64 | tr -d '\n')"
export AZURE_TENANT_ID_B64="$(echo -n "$AZURE_TENANT_ID" | base64 | tr -d '\n')"
export AZURE_CLIENT_ID_B64="$(echo -n "$AZURE_CLIENT_ID" | base64 | tr -d '\n')"
export AZURE_CLIENT_SECRET_B64="$(echo -n "$AZURE_CLIENT_SECRET" | base64 | tr -d '\n')"

# Settings needed for AzureClusterIdentity used by the AzureCluster
export AZURE_CLUSTER_IDENTITY_SECRET_NAME="cluster-identity-secret"
export CLUSTER_IDENTITY_NAME="cluster-identity"
export AZURE_CLUSTER_IDENTITY_SECRET_NAMESPACE="default"

export AZURE_CONTROL_PLANE_MACHINE_TYPE="Standard_DC4s_v2"
export AZURE_LOCATION="westeurope"
export AZURE_NODE_MACHINE_TYPE="Standard_DC4s_v2"

# Create a secret to include the password of the Service Principal identity created in Azure
# This secret will be referenced by the AzureClusterIdentity used by the AzureCluster
kubectl create secret generic "${AZURE_CLUSTER_IDENTITY_SECRET_NAME}" --from-literal=clientSecret="${AZURE_CLIENT_SECRET}" --namespace "${AZURE_CLUSTER_IDENTITY_SECRET_NAMESPACE}"

# Management cluster
EXP_AKS="true" EXP_CLUSTER_RESOURCE_SET="true" EXP_MACHINE_POOL="true" clusterctl init --infrastructure azure

kubectl patch deployment -n capz-system capz-controller-manager --patch '{"spec":{"template":{"spec":{"containers":[{"name":"manager","image":"zioproto/cluster-api-azure-controller-amd64:dev"}]}}}}'

kubectl apply -f https://raw.githubusercontent.com/zioproto/cluster-api-provider-azure/issue-1778/config/crd/bases/infrastructure.cluster.x-k8s.io_azuremanagedmachinepools.yaml
```

Then I created an AKS cluster using this kind of `AzureManagedMachinePool`:

```
---
kind: AzureManagedMachinePool
metadata:
  name: pool0
  namespace: default
spec:
  enableNodePublicIP: false
  mode: System
  osDiskSizeGB: 30
  sku: Standard_DC4s_v2
  additionalTags:
    owner: "saverio"
```

I can read the tag back from the AKS API

```
% az aks nodepool list --cluster-name test -g test -o json | jq '.[0].tags'
{
  "owner": "saverio"
}
```

zioproto:
Could anyone please help me to understand why the `pull-cluster-api-provider-azure-e2e-exp` fails ? It seems to fail every run with a different Azure infrastructure problem not related to this change I am testing. In the failed runs there is always a timeout to some operation against the Azure API. Is it possible that is a problem of the CI and not a problem of the specific PR I am testing ? Thank you

mtougeron:
According to the [controller logs](https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/2745/pull-cluster-api-provider-azure-e2e-exp/1585869194644164608/artifacts/clusters/bootstrap/controllers/capz-controller-manager/capz-controller-manager-6d949555b8-nrzbx/manager.log) of the most recent run, I'm seeing some 429s

```
E1028 06:28:04.554047       1 async.go:86] async.Service.processOngoingOperation "msg"="error checking long running operation status after it finished" "error"="Code=\"ReconcileStandardLoadBalancerError\" Message=\"Reconcile standard load balancer failed. Details: outboundReconciler retry failed: Category: ClientError; Code: Unspecified; SubCode: OperationNotAllowed_TooManyRequests; Message: ; InnerMessage: ; Dependency: Microsoft.Compute/VirtualMachineScaleSet; OrginalError: autorest/azure: Service returned an error. Status=429 Code=\\\"OperationNotAllowed\\\" Message=\\\"The server rejected the request because too many requests have been received for this subscription.\\\" Details=[{\\\"code\\\":\\\"TooManyRequests\\\",\\\"message\\\":\\\"{\\\\\\\"operationGroup\\\\\\\":\\\\\\\"HighCostGetVMScaleSet30Min\\\\\\\",\\\\\\\"startTime\\\\\\\":\\\\\\\"2022-10-28T06:27:55.9583635+00:00\\\\\\\",\\\\\\\"endTime\\\\\\\":\\\\\\\"2022-10-28T06:35:00+00:00\\\\\\\",\\\\\\\"allowedRequestCount\\\\\\\":728,\\\\\\\"measuredRequestCount\\\\\\\":1232}\\\",\\\"target\\\":\\\"HighCostGetVMScaleSet30Min\\\"}] InnerError={\\\"internalErrorCode\\\":\\\"TooManyRequestsReceived\\\"}; AKSTeam: Networking, Retriable: false.\"" "azureManagedMachinePool"={"name":"pool3","namespace":"capz-e2e-mop8kn"} "controller"="azuremanagedmachinepool" "controllerGroup"="infrastructure.cluster.x-k8s.io" "controllerKind"="AzureManagedMachinePool" "name"="pool3" "namespace"="capz-e2e-mop8kn" "reconcileID"="0efed988-6092-41bd-99bf-9134a7b48cfa" "x-ms-correlation-request-id"="df167f9a-c404-493c-9f9a-61ac7a0b529d" 
```

and

```
E1028 06:34:01.905609       1 controller.go:326]  "msg"="Reconciler error" "error"="error creating AzureManagedMachinePool capz-e2e-mop8kn/pool2: failed to reconcile machine pool pool2: containerservice.AgentPoolsCreateOrUpdateFuture#Result: Polling failure: StatusCode=0 -- Original Error: Code=\"ReconcileStandardLoadBalancerError\" Message=\"Reconcile standard load balancer failed. Details: outboundReconciler retry failed: Category: ClientError; Code: Unspecified; SubCode: OperationNotAllowed_TooManyRequests; Message: ; InnerMessage: ; Dependency: Microsoft.Compute/VirtualMachineScaleSet; OrginalError: autorest/azure: Service returned an error. Status=429 Code=\\\"OperationNotAllowed\\\" Message=\\\"The server rejected the request because too many requests have been received for this subscription.\\\" Details=[{\\\"code\\\":\\\"TooManyRequests\\\",\\\"message\\\":\\\"{\\\\\\\"operationGroup\\\\\\\":\\\\\\\"HighCostGetVMScaleSet30Min\\\\\\\",\\\\\\\"startTime\\\\\\\":\\\\\\\"2022-10-28T06:33:59.2485233+00:00\\\\\\\",\\\\\\\"endTime\\\\\\\":\\\\\\\"2022-10-28T06:35:00+00:00\\\\\\\",\\\\\\\"allowedRequestCount\\\\\\\":728,\\\\\\\"measuredRequestCount\\\\\\\":1414}\\\",\\\"target\\\":\\\"HighCostGetVMScaleSet30Min\\\"}] InnerError={\\\"internalErrorCode\\\":\\\"TooManyRequestsReceived\\\"}; AKSTeam: Networking, Retriable: false.\"" "azureManagedMachinePool"={"name":"pool2","namespace":"capz-e2e-mop8kn"} "controller"="azuremanagedmachinepool" "controllerGroup"="infrastructure.cluster.x-k8s.io" "controllerKind"="AzureManagedMachinePool" "name"="pool2" "namespace"="capz-e2e-mop8kn" "reconcileID"="3c2398f8-d74d-4142-9a47-3865e40b0d3b" 
```

mtougeron:
@zioproto in case you don't know how to get to the controller logs, you get to Prow by click on the job `Details` link, then choosing `Artifacts` at the top, then navigate to `artifacts/clusters/bootstrap/controllers/capz-controller-manager/<controller pod name>/manager.log`

CecileRobertMichon:
@nojnhuh I do see this error in the e2e output:

> Oct 31 16:38:38.798: INFO: Got err network.PublicIPPrefixesCreateOrUpdateFuture: asynchronous operation has not completed
Oct 31 16:38:49.118: INFO: Got err <nil>

is that expected?

nojnhuh:
> @nojnhuh I do see this error in the e2e output:
> 
> > Oct 31 16:38:38.798: INFO: Got err network.PublicIPPrefixesCreateOrUpdateFuture: asynchronous operation has not completed
> > Oct 31 16:38:49.118: INFO: Got err
> 
> is that expected?

Yes, that tends to be how that looks normally. For reference, this is where that comes from: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/7eb7c4281575049e06c17a0effb95c774f234289/test/e2e/aks.go#L397-L401

Maybe it's worth tweaking that to only log if it's not a nil or "operation not complete" error.

CecileRobertMichon:
> Maybe it's worth tweaking that to only log if it's not a nil or "operation not complete" error.

+1, I would even say just log the last error if the eventually block times out

azure/services/agentpools/spec.go
```diff
@@ -148,6 +152,7 @@ func (s *AgentPoolSpec) Parameters(existing interface{}) (params interface{}, er
 				MaxCount:            existingPool.MaxCount,
 				NodeLabels:          existingPool.NodeLabels,
 				NodeTaints:          existingPool.NodeTaints,
+				Tags:                existingPool.Tags,
```
CecileRobertMichon, line 155: @nojnhuh @zioproto 

I think this is what may be causing the throttling issue

Looking at https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/2745/pull-cluster-api-provider-azure-e2e-exp/1587117062692540416/artifacts/clusters/capz-e2e-eju8ih-aks/azure-activity-logs/capz-e2e-eju8ih-aks.log, there hundreds (!) of agentpool requests, which is unexpected.

I started looking at what could be causing this and I believe it's because we're comparing all the existing tags on the agentpool with our spec tags, which is always going to cause a diff since AKS adds built in tags. Because of this, the agentpools are getting updated non-stop on every loop, and eventually causing throttling, which causes the e2e test to fail.

zioproto: I looked into this but I don't understand the "AKS adds built in tags".

I tried to create an AKS cluster with minimal options:

```
az aks create -g debug -n myAKSCluster
az aks list -o table
az aks nodepool list --cluster-name myAKSCluster -g debug -o json | jq '.[].tags'
```
and the NodePool does not have any default tag.

I also tried creating a cluster with Cluster API with the stable version and I dont see any tag associated to the Node Pools.

CecileRobertMichon: Could you add a log statement in here so we can see what diff is triggering the updates in the e2e test logs?

Note: #2611 is being worked on which will help with things like this in the future, but let's add a temporary log here so we can understand what's going on

zioproto: I added this ugly patch in my local env:
```
diff --git a/azure/services/agentpools/spec.go b/azure/services/agentpools/spec.go
index 3511a848..14d6e718 100644
--- a/azure/services/agentpools/spec.go
+++ b/azure/services/agentpools/spec.go
@@ -171,8 +171,10 @@ func (s *AgentPoolSpec) Parameters(existing interface{}) (params interface{}, er
                if diff == "" {
                        // agent pool is up to date, nothing to do
+                       fmt.Print("agent pool is up to date, nothing to do\n")
                        return nil, nil
                }
+               fmt.Printf("Should merge tags here diff: %s", diff)
        }
```
And you are right there is a diff:
```
capz-controller-manager-86fd5fcfc-kwb7k manager Should merge tags here diff:   containerservice.AgentPool{
capz-controller-manager-86fd5fcfc-kwb7k manager   	Response: {},
capz-controller-manager-86fd5fcfc-kwb7k manager   	ManagedClusterAgentPoolProfileProperties: &containerservice.ManagedClusterAgentPoolProfileProperties{
capz-controller-manager-86fd5fcfc-kwb7k manager   		... // 24 identical fields
capz-controller-manager-86fd5fcfc-kwb7k manager   		ScaleSetEvictionPolicy: "",
capz-controller-manager-86fd5fcfc-kwb7k manager   		SpotMaxPrice:           nil,
capz-controller-manager-86fd5fcfc-kwb7k manager - 		Tags:                   map[string]*string{},
capz-controller-manager-86fd5fcfc-kwb7k manager + 		Tags:                   nil,
capz-controller-manager-86fd5fcfc-kwb7k manager   		NodeLabels:             nil,
capz-controller-manager-86fd5fcfc-kwb7k manager   		NodeTaints:             nil,
capz-controller-manager-86fd5fcfc-kwb7k manager   		... // 7 identical fields
capz-controller-manager-86fd5fcfc-kwb7k manager   	},
capz-controller-manager-86fd5fcfc-kwb7k manager   	ID:   nil,
capz-controller-manager-86fd5fcfc-kwb7k manager   	Name: nil,
capz-controller-manager-86fd5fcfc-kwb7k manager   	Type: nil,
capz-controller-manager-86fd5fcfc-kwb7k manager   }
```

zioproto: I added some tags to understand better the output:
```
---
kind: AzureManagedMachinePool
metadata:
  name: pool1
  namespace: default
spec:
  enableNodePublicIP: false
  mode: User
  osDiskSizeGB: 40
  sku: Standard_DC4s_v2
  additionalTags:
    owner: "one"
 ```

```
capz-controller-manager-86fd5fcfc-kwb7k manager Should merge tags here diff:   containerservice.AgentPool{
capz-controller-manager-86fd5fcfc-kwb7k manager   	Response: {},
capz-controller-manager-86fd5fcfc-kwb7k manager   	ManagedClusterAgentPoolProfileProperties: &containerservice.ManagedClusterAgentPoolProfileProperties{
capz-controller-manager-86fd5fcfc-kwb7k manager   		... // 24 identical fields
capz-controller-manager-86fd5fcfc-kwb7k manager   		ScaleSetEvictionPolicy: "",
capz-controller-manager-86fd5fcfc-kwb7k manager   		SpotMaxPrice:           nil,
capz-controller-manager-86fd5fcfc-kwb7k manager - 		Tags:                   map[string]*string{"owner": &"one"},
capz-controller-manager-86fd5fcfc-kwb7k manager + 		Tags:                   nil,
capz-controller-manager-86fd5fcfc-kwb7k manager   		NodeLabels:             nil,
capz-controller-manager-86fd5fcfc-kwb7k manager   		NodeTaints:             nil,
capz-controller-manager-86fd5fcfc-kwb7k manager   		... // 7 identical fields
capz-controller-manager-86fd5fcfc-kwb7k manager   	},
capz-controller-manager-86fd5fcfc-kwb7k manager   	ID:   nil,
capz-controller-manager-86fd5fcfc-kwb7k manager   	Name: nil,
capz-controller-manager-86fd5fcfc-kwb7k manager   	Type: nil,
capz-controller-manager-86fd5fcfc-kwb7k manager   }
```

Comment (zioproto) [reply azure/services/agentpools/spec.go:155]: I figured out the state above is transient because I just had added the Tag.

The update forever happens only when there are no tags associated with the nodepool because there is a comparison between `nil` and `map[string]*string{}`.

```
capz-controller-manager-86fd5fcfc-kwb7k manager - 		Tags:                   map[string]*string{},
capz-controller-manager-86fd5fcfc-kwb7k manager + 		Tags:                   nil,
```

Comment (zioproto) [reply azure/services/agentpools/spec.go:155]: When all my nodepools have some tag I correctly see:

```
capz-controller-manager-86fd5fcfc-kwb7k manager agent pool is up to date, nothing to do
capz-controller-manager-86fd5fcfc-kwb7k manager agent pool is up to date, nothing to do
```

Comment (zioproto) [reply azure/services/agentpools/spec.go:155]: @CecileRobertMichon problem comes from here:
https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/2349883d1effe6344690459b40e5dc26138010df/azure/converters/tags.go#L35-L41

This will always return a `map[string]*string{}` right ?

Comment (zioproto) [reply azure/services/agentpools/spec.go:155]: I pushed this commit:
https://github.com/zioproto/cluster-api-provider-azure/commit/5a08d8c7aaee8025f79e45f5065a7e19e7f35142

Comment (CecileRobertMichon) [reply azure/services/agentpools/spec.go:155]: > This will always return a map[string]*string{} right

IMO this is a bug, it should return nil if the src is nil, could you please update that function and add a unit test to validate that behavior? That way you can drop the workaround

**Review by CecileRobertMichon — COMMENTED** [review]
LGTM aside from comments on nil map issue

File: azure/services/agentpools/spec.go (hunk 5 above)
Comment (zioproto) [reply azure/services/agentpools/spec.go:155]: For better readability and better future audit of the code changes, I think it is better to address the converters issue in a different PR. I understand if you want to hold this PR until we fix the converters. I suspect at least `TagsToMap` and `MapToTags` are broken in the same way.

Comment (zioproto) [reply azure/services/agentpools/spec.go:155]: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2802

**Review by nojnhuh — COMMENTED** [review]
Took this for a spin and it works great! Looks good to me once Cecile's remaining comments are addressed.

**Comment by zioproto** [comment]
I addressed Cecile's comments on the order no being alphabetical. Rebased on current `main` branch.

**Comment by dtzar** [comment]
@zioproto Looks like the only thing blocking is the one E2E test failing.  Do you know if this is related to your PR?

```
Nov 21 13:23:07.674: INFO: starting to cleaning up network policy development/backend-deny-egress after ourselves
STEP: Applying a network policy to allow egress access to app: webapp, role: frontend pods in any namespace from pods with app: webapp, role: backend labels in development namespace
Nov 21 13:23:07.857: INFO: starting to applying a network policy development/backend-allow-egress-pod-label to allow egress access to app: webapp, role: frontend pods in any namespace from pods with app: webapp, role: backend labels in development namespace
STEP: Ensuring we have egress access from pods with matching labels
STEP: Ensuring we don't have ingress access from pods without matching labels
curl: (7) Failed to connect to 192.168.167.7 port 80: Connection timed out
```

**Comment by zioproto** [comment]
@dtzar I don't think it is related.

Now that #2802 is merged, I was able to revert the commit with the workaround for the converters. I also rebased to the current main branch.

I added the release notes. If the tests pass this PR is now ready to merge. Thanks

**Comment by zioproto** [comment]
I figured out I had to revert some changes to `azure/services/agentpools/spec_test.go`:

What I reverted is here:
https://github.com/kubernetes-sigs/cluster-api-provider-azure/commit/cdda4d3fa6a3be0be69fa0783d325ceb623baa61#diff-224617dc1cf5a3e1abbc71a49b79c42441871bef5a14e810921526267e6bde82

All those `AdditionalTags:    map[string]string{},` were needed because before merging #2802 a `nil` list of tags was converted to `map[string]string{}` because of the buggy behavior, and without a clear understanding of what I was doing, I patched the unit test to accomodate the buggy behaviour.

Now that  #2802 is merged I reverted  `azure/services/agentpools/spec_test.go` to the original content.

@CecileRobertMichon please take a look again. Thanks

File: azure/services/agentpools/spec.go [hunk 6]
```diff
@@ -201,6 +207,10 @@ func (s *AgentPoolSpec) Parameters(existing interface{}) (params interface{}, er
 	if s.VnetSubnetID != "" {
 		vnetSubnetID = &s.VnetSubnetID
 	}
+	var additionalTags infrav1.Tags
```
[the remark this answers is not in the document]
nojnhuh: I think the pattern you're seeing here is only for the fields that are a pointer in the Azure API types but are plain values in the CAPZ `AgentPoolSpec`. The type of some fields like `s.MaxPods` matches the Azure API types and those values are passed through without a similar check here.

Since `converters.TagsToMap()` handles that type conversion between `map[string]string` and `map[string]*string` already and your other PR fixed the `nil` map case, I think this code here could be removed and the value could be passed through directly in the return like:
```go
			Tags:                 converters.TagsToMap(s.AdditionalTags),
```

zioproto:
@CecileRobertMichon @nojnhuh all comments addressed, all tests passing. Release notes added. Missing your approvals for merge. Thank you !

nojnhuh (approved):
/lgtm
Thanks @zioproto!

PR #2748, opened 2022-10-21
Use azidentity instead of ADAL

r4f4:
ADAL is being deprecated, so use azidentity with an adapter so the new authentication can work with v1 SDK clients. The only impact for users is to update the environment variables used for certificate authentication, from `AZURE_CERTIFICATE_` to `AZURE_CLIENT_CERTIFICATE_`.

Signed-off-by: Rafael Fonseca <r4f4rfs@gmail.com>

 

 
**What type of PR is this?**
/kind other

**What this PR does / why we need it**: ADAL is being deprecated, so use azidentity with an adapter so the new authentication can work with v1 SDK clients.

Fixes #

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Replace deprecated ADAL authentication with MSAL.
action required: for users relying on certificate authentication, the environment variables used by the libraries changed from AZURE_CERTIFICATE_ to AZURE_CLIENT_CERTIFICATE_
```


azure/scope/identity.go
```diff
@@ -23,8 +23,11 @@ import (
 
 	aadpodid "github.com/Azure/aad-pod-identity/pkg/apis/aadpodidentity"
 	aadpodv1 "github.com/Azure/aad-pod-identity/pkg/apis/aadpodidentity/v1"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud"
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
 	"github.com/Azure/go-autorest/autorest"
-	"github.com/Azure/go-autorest/autorest/adal"
+	"github.com/jongio/azidext/go/azidext"
```
CecileRobertMichon, line 30: what is this repo? it seems that it's in someone's personal GitHub org?

CecileRobertMichon: should we be following https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azidentity/MIGRATION.md instead?

r4f4: It's what I did. See the section https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azidentity/MIGRATION.md#use-azidentity-credentials-with-older-packages where they recommend jongio's repo for getting an adapter so the new authorizer can work with older SDK clients.

CecileRobertMichon: Got it, thanks for clarifying

sonasingh46:
This would be nice to get in. I have mentioned about this in workload identity proposal as a future work -- but prior work is fine.
https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2814

With azure workload identity, it seem we have to have use azidentity.

azure/scope/identity.go
```diff
@@ -136,44 +139,50 @@ func (p *ManagedControlPlaneCredentialsProvider) GetAuthorizer(ctx context.Conte
 
 // GetAuthorizer returns an Azure authorizer based on the provided azure identity and cluster metadata.
 func (p *AzureCredentialsProvider) GetAuthorizer(ctx context.Context, resourceManagerEndpoint, activeDirectoryEndpoint string, clusterMeta metav1.ObjectMeta) (autorest.Authorizer, error) {
-	var spt *adal.ServicePrincipalToken
+	var authErr error
+	var cred azcore.TokenCredential
```
sonasingh46, line 143: Any particular reason for declaring these variables explicitly vs using shorthand in the switch case?  

e.g

```
cred, err := azidentity.NewManagedIdentityCredential(&options)
```

azure/scope/identity.go
```diff
@@ -136,44 +139,50 @@ func (p *ManagedControlPlaneCredentialsProvider) GetAuthorizer(ctx context.Conte
...
+		return nil, errors.Errorf("failed to get token from service principal identity: %v", authErr)
+	}
+	authorizer := azidext.NewTokenCredentialAdapter(cred, []string{endpointToScope(resourceManagerEndpoint)})
+	return authorizer, nil
+}
+
+func endpointToScope(endpoint string) string {
+	if len(endpoint) > 0 && endpoint[len(endpoint)-1] != '/' {
+		endpoint += "/"
+	}
```
aramase, line 184: Why is this additional `/` added here? `resourceManagerEndpoint` in the `GetAuthorizer` call should already have the exact url that the token needs to be requested for? FYI, adding an extra slash or removing an existing one can render the token unusable because the resource identifier has to be an exact match.

xref: https://github.com/Azure/azure-workload-identity/blob/main/pkg/proxy/proxy.go#L194-L203

aramase: I was just looking at https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2814 and found a ref to this, so figured i'll review it. I didn't mean to lurk 😆

azure/scope/identity.go
r4f4: Yes. If I don't do this then I can't use them on lines https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2748/files#diff-b6ab207506064afcbd1d777c1c7296e2953d2653a8843bfb39107af5cf5b0a87R174-R177.

The alternative would be to duplicate those lines in each `case` branch, which I tried to avoid.

azure/scope/identity.go
r4f4: Thanks for the review. I've changed the code to be similar to the one you linked.

sonasingh46:
@r4f4 -- The PR is still titled `WIP`. Do you see any more work that needs to be done? Also, let me know if you need some help on getting this to a merge state. 
Thanks for doing this. It will be really helpful in getting the azwi integration.

azure/scope/identity.go
```diff
@@ -136,44 +140,47 @@ func (p *ManagedControlPlaneCredentialsProvider) GetAuthorizer(ctx context.Conte
...
 		return nil, errors.Errorf("identity type %s not supported", p.Identity.Spec.Type)
 	}
 
-	return autorest.NewBearerAuthorizer(spt), nil
+	if authErr != nil {
+		return nil, errors.Errorf("failed to get token from service principal identity: %v", authErr)
+	}
+	scope := resourceManagerEndpoint
+	if !strings.HasPrefix(scope, "/.default") {
+		scope = scope + "/.default"
```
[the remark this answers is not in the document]
r4f4: Right. I have this fixed locally, just waiting for more changes I need to commit before pushing them all.

sonasingh46:
[1] Re the env variables -- if MSAL uses a different env var name that will mean to communicate this to users to set up a env variable with the new name if a capz version containing this change is used or upgraded to.
I don't think this is something that can be or should be controlled from within capz.  

[2] I do not see ordering as a problem too. 

But it is worth to put these in release note and as well as in PR description. 

@aramase @CecileRobertMichon Any thoughts?

r4f4:
I'll wait for the test results before squashing commits and editing PR description with a release note.

r4f4:
I have also made some changes to the code to support StackCloud after my findings from testing https://github.com/openshift/installer/pull/6003 and added another commit to replace ADAL in the e2e tests which I have previously missed.

aramase:
> [1] Re the env variables -- if MSAL uses a different env var name that will mean to communicate this to users to set up a env variable with the new name if a capz version containing this change is used or upgraded to.
I don't think this is something that can be or should be controlled from within capz.

I haven't looked at the rest of the PR yet or the pkg docs, but this seems to be a breaking change. Can this be handled in CAPZ itself to set the env var `AZURE_CLIENT_CERTIFICATE_` when `AZURE_CERTIFICATE_` is set to ensure backward compatibility? (FYI, I haven't looked at the code yet to see who/where `AZURE_CERTIFICATE_` is being set/used today)

jackfrancis:
@r4f4 @sonasingh46 how far along is this? Should we add to v1.7 milestone (ETA: early Jan)?

Thanks!

r4f4:
> @r4f4 @sonasingh46 how far along is this? Should we add to v1.7 milestone (ETA: early Jan)?

From my side it's ready, unless it's decided I should implement what's suggest in https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2748#issuecomment-1329877714. I think early Jan milestone makes sense, since the EOL of ADAL is happening in Dec 31.

r4f4:
Fixed a typo `s/HasPrefix/HasSuffix`

sonasingh46:
@r4f4 - The PR looks good to me. Please rebase it. 

>I haven't looked at the rest of the PR yet or the pkg docs, but this seems to be a breaking change. Can this be handled in CAPZ itself to set the env var AZURE_CLIENT_CERTIFICATE_ when AZURE_CERTIFICATE_ is set to ensure backward compatibility? (FYI, I haven't looked at the code yet to see who/where AZURE_CERTIFICATE_ is being set/used today)


@aramase -- I don't see any such major issue in handling the env vars via capz, just that it looks bit odd and adding to maintenance overhead in capz ( with regard to how the MSAL library will move further ) 
But I also understand the downside, that if we choose to not do via capz, that introduces an extra step while upgrading.  

Also @CecileRobertMichon -- Do you have any thoughts re the env variable part.

r4f4:
Rebased and squashed fixup commits.

go.mod
```diff
@@ -5,8 +5,9 @@ go 1.19
 require (
 	github.com/Azure/aad-pod-identity v1.8.9
 	github.com/Azure/azure-sdk-for-go v67.1.0+incompatible
+	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.0.0
```
sonasingh46, line 8: Hey @r4f4 -- Can we please use v1.2.0 for azcore?

go.mod
```diff
@@ -5,8 +5,9 @@ go 1.19
...
+	github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.0.0
```
sonasingh46, line 9: For azidentity also, please use v1.2.0

sonasingh46:
@r4f4 -- Added one minor comment. Please take a look. Also, let us try to get this PR in next week by discussing about the env variable part.

go.mod
r4f4: Done

go.mod
r4f4: Done

r4f4:
> @r4f4 -- Added one minor comment. Please take a look. Also, let us try to get this PR in next week by discussing about the env variable part.

I've added a commit that sets the new vars in case the old ones are present. I'll be happy to drop it in case we decide that is not what we want to do.

azure/scope/clients.go
```diff
@@ -122,6 +122,16 @@ func (c *AzureClients) setCredentials(subscriptionID, environmentName string) er
...
+
+		// azidentity uses different envvars for certificate authentication
+		// Let's set it according to the envvars used by autorest, in case they are present
+		_, azidSet := os.LookupEnv("AZURE_CLIENT_CERTIFICATE_PATH")
+		path, autorestSet := os.LookupEnv("AZURE_CERTIFICATE_PATH")
+		if !azidSet && autorestSet {
+			os.Setenv("AZURE_CLIENT_CERTIFICATE_PATH", path)
+			os.Setenv("AZURE_CLIENT_CERTIFICATE_PASSWORD", os.Getenv("AZURE_CERTIFICATE_PASSWORD"))
+		}
+
```
sonasingh46, line 134: 1. If we are going to go with this approach, please include a comment on what env var is supported by MSAL vs autorest. 

@CecileRobertMichon -- Thoughts?

r4f4: Done

r4f4:
Rebased and added the comment as requested.

CecileRobertMichon:
Are we keeping the 3 distinct commits intentionally or would it make sense to squash those?

sonasingh46:
I think that should be squashed and @r4f4 kept it for easy reviews. @CecileRobertMichon 

@r4f4 -- Please squash the commits.

r4f4:
Squashed commits.

CecileRobertMichon (commented):
just a minor code cleanup comment from me, otherwise LGTM

test/e2e/common.go
```diff
@@ -192,13 +197,64 @@ func dumpSpecResourcesAndCleanup(ctx context.Context, input cleanupInput) {
 	}
 }
 
+// GetAuthorizer returns an autorest.Authorizer-compatible object from MSAL
+func GetAuthorizer(settings auth.EnvironmentSettings) (autorest.Authorizer, error) {
```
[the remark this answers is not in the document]
r4f4: Done

mboersma (approved):
/lgtm
/approve

CecileRobertMichon:
another flake?

/retest

PR #2751, opened 2022-10-26
fix scaling managed machine pool to zero

nojnhuh:
**What type of PR is this?**
/kind bug

Scaling a MachinePool backed by AzureManagedMachinePool to 0 replicas would never successfully apply those changes because the internal `AgentPoolSpec` type used to convert Cluster API types to the Azure resource cannot distinguish between a specified 0 and unspecified number of replicas, and assumes 0 means unspecified. This causes the resulting PUT operation during reconciliation to omit the replica count instead of setting it to 0.

For reference, this is where the `AgentPoolSpec` is constructed:
https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/4b5768a12ee9d14c174c8f208bddeaeac5d1acf3/azure/scope/managedmachinepool.go#L153-L156

Fixes #

I'm working on a [set of e2e tests](https://github.com/kubernetes-sigs/cluster-api-provider-azure/compare/main...nojnhuh:cluster-api-provider-azure:aks-topology-test) which cover this scenario among other related scaling tests. Please let me know if those tests should be included in this PR.

Since only User node pools can be scaled to 0, I verified that scaling a System node pool to 0 fails reasonably, with this condition getting set on the MachinePool:
```yaml
- lastTransitionTime: "2022-10-26T18:05:37Z"
  message: 'agentpools failed to create or update. err: failed to update resource
    aks-22378/pool0 (service: agentpools): failed to begin operation: Code="InvalidParameter"
    Message="The value of parameter agentPoolProfile.count is invalid. Please see
    https://aka.ms/aks-naming-rules for more details." Target="agentPoolProfile.count"'
  reason: Failed
  severity: Error
  status: "False"
  type: InfrastructureReady
```

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Fixed an issue preventing AKS "User" node pools from successfully scaling to 0 replicas
```

/area managedclusters

### DIFF
--- a/azure/services/agentpools/spec.go
+++ b/azure/services/agentpools/spec.go
@@ -189,10 +189,6 @@ func (s *AgentPoolSpec) Parameters(existing interface{}) (params interface{}, er
 	if len(s.AvailabilityZones) > 0 {
 		availabilityZones = &s.AvailabilityZones
 	}
-	var replicas *int32
-	if s.Replicas > 0 {
-		replicas = &s.Replicas
-	}
 	var nodeTaints *[]string
 	if len(s.NodeTaints) > 0 {
 		nodeTaints = &s.NodeTaints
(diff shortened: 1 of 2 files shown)


mboersma:
> I'm working on a [set of e2e tests](https://github.com/kubernetes-sigs/cluster-api-provider-azure/compare/main...nojnhuh:cluster-api-provider-azure:aks-topology-test) which cover this scenario

There is a `TestParameters` unit test that calls this func. Maybe it could be improved to cover this change?

nojnhuh:
> There is a `TestParameters` unit test that calls this func. Maybe it could be improved to cover this change?

Good call, I just added a test that passes with this change but failed before. I think we'll probably still want to do some e2e tests scaling machine pools up and down but a unit test definitely makes sense to include here.

mboersma (commented):
/lgtm

CecileRobertMichon (commented):
/lgtm
/approve

PR #2753, opened 2022-10-27
Stop using the beta.kubernetes.io/os

pacoxu:
/kind cleanup
See https://github.com/kubernetes/kubernetes/issues/89477, https://github.com/kubernetes/kubernetes/issues/89477#issuecomment-603911496.

```release-note
None
```


mboersma (commented):
/lgtm
/approve

I test this locally through `make tilt-up` and prometheus still works fine, as expected. Thanks @pacoxu!

PR #2755, opened 2022-10-27
Handle different base64 tools in Tiltfile

mboersma:
**What type of PR is this?**:

/kind bug

After upgrading to macOS 13 Ventura, I found `make tilt-up` failed due to a new `base64` binary showing up which doesn't understand `-` without `-i` preceding it:

```shell
% which base64
/usr/bin/base64
% base64 --help
Usage:	base64 [-hDd] [-b num] [-i in_file] [-o out_file]
  -h, --help     display this message
  -Dd, --decode   decodes input
  -b, --break    break encoded string into num character lines
  -i, --input    input file (default: "-" for stdin)
  -o, --output   output file (default: "-" for stdout)
```

For comparison, on Ubuntu Linux:

```
% base64 --help
Usage: base64 [OPTION]... [FILE]
Base64 encode or decode FILE, or standard input, to standard output.

With no FILE, or when FILE is -, read standard input.

Mandatory arguments to long options are mandatory for short options too.
  -d, --decode          decode data
  -i, --ignore-garbage  when decoding, ignore non-alphabet characters
  -w, --wrap=COLS       wrap encoded lines after COLS character (default 76).
                          Use 0 to disable line wrapping

      --help     display this help and exit
      --version  output version information and exit
```

N/A

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -361,11 +361,11 @@ def deploy_worker_templates(template, substitutions):
     )
 
 def base64_encode(to_encode):
-    encode_blob = local("echo '{}' | tr -d '\n' | base64 - | tr -d '\n'".format(to_encode), quiet = True, echo_off = True)
+    encode_blob = local("echo '{}' | tr -d '\n' | base64 | tr -d '\n'".format(to_encode), quiet = True, echo_off = True)
     return str(encode_blob)
 
 def base64_encode_file(path_to_encode):
-    encode_blob = local("cat {} | tr -d '\n' | base64 - | tr -d '\n'".format(path_to_encode), quiet = True)
+    encode_blob = local("cat {} | tr -d '\n' | base64 | tr -d '\n'".format(path_to_encode), quiet = True)
     return str(encode_blob)
 
 def read_file_from_path(path_to_read):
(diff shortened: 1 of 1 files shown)


Tiltfile
```diff
@@ -360,20 +360,26 @@ def deploy_worker_templates(template, substitutions):
...
 
+def base64_local_args():
+    # Some versions of base64 require the "-i/--input" flag to read from stdin.
+    if local("base64 --help | grep '\\-i '", quiet = True, echo_off = True):
+        return "-i "
+    return ""
+
 def base64_encode(to_encode):
-    encode_blob = local("echo '{}' | tr -d '\n' | base64 - | tr -d '\n'".format(to_encode), quiet = True, echo_off = True)
+    encode_blob = local("echo '{}' | tr -d '\n' | base64 {}- | tr -d '\n'".format(to_encode, base64_local_args()), quiet = True, echo_off = True)
```
nojnhuh, line 370: Would it be reasonable to omit this flag altogether and let the stdin default apply? I played around with that on Ubuntu and pre-Ventura macOS and it seems to work as expected.
```suggestion
    encode_blob = local("echo '{}' | tr -d '\n' | base64 | tr -d '\n'".format(to_encode), quiet = True, echo_off = True)
```

mboersma: Yes, simpler == better if that works. Let do some more local testing and I'll update this.

mboersma: Yep, much simpler and seems to work portably (between Ubuntu 22.04 and macOS 13 at least). 

Thanks @nojnhuh! Could you try this on macOS 12 or earlier?

nojnhuh: Working for me!

CecileRobertMichon:
> Rebased in order to get green e2e tests

next time you don't need to do this, prow will do a rebase before every test run so you should be able to just `/retest`

/lgtm

PR #2760, opened 2022-10-28
clusterctl: make explicit use of --worker-machine-count=${WORKER_MACH…

### LINKED ISSUE #2759: Default MachinePool spec.replicas=0 default value is invalid
/kind question

Generating the cluster resources with this command:
```
clusterctl generate cluster test --kubernetes-version 1.23.12 --flavor aks
```
By default the `MachinePool` has 0 replicas.

```
---
apiVersion: cluster.x-k8s.io/v1beta1
kind: MachinePool
metadata:
  name: pool0
  namespace: default
spec:
  clusterName: test
  replicas: 0
  template:
    metadata: {}
    spec:
      bootstrap:
        dataSecretName: ""
      clusterName: test
      infrastructureRef:
        kind: AzureManagedMachinePool
        name: pool0
      version: 1.23.12
```

The AKS cluster fails with `the value of parameter agentPoolProfile.count is invalid`

The number of replicas can be changed with the env var `WORKER_MACHINE_COUNT`. However the docs dont specify that this env variable is mandatory. https://capz.sigs.k8s.io/topics/managedcluster.html

The docs are a bit confusing because the guidance is:
```
# Kubernetes values
export CLUSTER_NAME="my-cluster"
export WORKER_MACHINE_COUNT=2
export KUBERNETES_VERSION="v1.19.6"
clusterctl init --infrastructure azure
clusterctl generate cluster ${CLUSTER_NAME} --kubernetes-version ${KUBERNETES_VERSION} --flavor aks > cluster.yaml
```
[It seems KUBERNETES_VERSION also dont need command line arguments to be evaluated](https://github.com/kubernetes-sigs/cluster-api/blob/main/cmd/clusterctl/cmd/util.go#L100). Because the other variables are passed as command line arguments it would be great to make also `WORKER_MACHINE_COUNT` explicit in the docs:

```
clusterctl generate cluster ${CLUSTER_NAME} --kubernetes-version ${KUBERNETES_VERSION} --worker-machine-count=${WORKER_MACHINE_COUNT} --flavor aks > cluster.yaml
```

Why the default template does not have `replicas: 1` ?
Is it the normal behaviour for other providers to deploy with 0 replicas ? Is this an AKS only limit ?

Why the naming `WORKER_MACHINE_COUNT` ?  Should `MACHINE_POOL_REPLICAS` be a more consistent name for this variable ?

**Environment:**

- cluster-api-provider-azure version: v1.5.3
- Kubernetes version: 1.23.12

This was discussed also here:
https://github.com/Azure/aks-engine/issues/2311

zioproto:
/kind documentation

Improves the documentation to use `clusterctl` to create an AKS cluster.

Fixes #2759

### DIFF
--- a/docs/book/src/topics/managedcluster.md
+++ b/docs/book/src/topics/managedcluster.md
@@ -74,7 +74,7 @@ Execute clusterctl to template the resources, then apply to your kind management
 
 ```bash
 clusterctl init --infrastructure azure
-clusterctl generate cluster ${CLUSTER_NAME} --kubernetes-version ${KUBERNETES_VERSION} --flavor aks > cluster.yaml
+clusterctl generate cluster ${CLUSTER_NAME} --kubernetes-version ${KUBERNETES_VERSION} --worker-machine-count=${WORKER_MACHINE_COUNT} --flavor aks > cluster.yaml
 
 # assumes an existing management cluster
 kubectl apply -f cluster.yaml


mboersma:
Adding this argument to the command doesn't change the actual behavior, right? The template should still respect the environment variable.

Is this just to make it more explicitly documented?

zioproto:
> Adding this argument to the command doesn't change the actual behavior, right? The template should still respect the environment variable.
> 
> Is this just to make it more explicitly documented?

Correct, this is just to make it more explicitly documented

docs/book/src/topics/managedcluster.md
```diff
@@ -58,7 +58,7 @@ Execute clusterctl to template the resources, then apply to a management cluster
 
 ```bash
 clusterctl init --infrastructure azure
-clusterctl generate cluster ${CLUSTER_NAME} --kubernetes-version ${KUBERNETES_VERSION} --flavor aks > cluster.yaml
+clusterctl generate cluster ${CLUSTER_NAME} --kubernetes-version ${KUBERNETES_VERSION} --worker-machine-count=${WORKER_MACHINE_COUNT} --flavor aks > cluster.yaml
```
CecileRobertMichon, line 61: while you're here, it looks like the `KUBERNETES_VERSION` in the example is outdated and no longer supported

jackfrancis: I found a whole bunch of errors with this example and fixed them here: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2776

FWIW I didn't observe that including `--worker-machine-count=${WORKER_MACHINE_COUNT}` was necessary, clusterctl picks up that env var automatically and does the right thing in the aks flavor template.

zioproto: I mean also `--kubernetes-version` is not necessary, because it is read from the env:

https://github.com/kubernetes-sigs/cluster-api/blob/bda7e5cca7b57767437f3627e8cf2d3d389e173f/cmd/clusterctl/cmd/util.go#L89-L107
I would say even we make both explicit or both implicit in the example

CecileRobertMichon:
/close in favor of https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2776

please reopen if you think this one is still relevant

PR #2764, opened 2022-10-31
add AKS nodepool autoscale test

nojnhuh:
**What type of PR is this?**
/kind cleanup

**What this PR does / why we need it**: This PR adds a simple flow to the AKS e2e test to enable and disable the cluster autoscaler for one of the cluster's machine pools.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/aks.go
+++ b/test/e2e/aks.go
@@ -355,6 +355,96 @@ func GetLatestStableAKSKubernetesVersion(ctx context.Context, subscriptionID, lo
 	return maxVersion, nil
 }
 
+type AKSAutoscaleSpecInput struct {
+	Cluster       *clusterv1.Cluster
+	MachinePool   *expv1.MachinePool
+	WaitIntervals []interface{}
+}
+
+func AKSAutoscaleSpec(ctx context.Context, inputGetter func() AKSAutoscaleSpecInput) {
+	input := inputGetter()
+
+	settings, err := auth.GetSettingsFromEnvironment()
+	Expect(err).NotTo(HaveOccurred())
+	subscriptionID := settings.GetSubscriptionID()
+	auth, err := settings.GetAuthorizer()
+	Expect(err).NotTo(HaveOccurred())
+	agentpoolClient := containerservice.NewAgentPoolsClient(subscriptionID)
+	agentpoolClient.Authorizer = auth
+	mgmtClient := bootstrapClusterProxy.GetClient()
+	Expect(mgmtClient).NotTo(BeNil())
+
+	amcp := &infrav1exp.AzureManagedControlPlane{}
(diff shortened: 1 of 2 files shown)


test/e2e/aks.go
```diff
@@ -355,6 +355,72 @@ func GetLatestStableAKSKubernetesVersion(ctx context.Context, subscriptionID, lo
...
+	amcp := &infrav1exp.AzureManagedControlPlane{}
+	err = mgmtClient.Get(ctx, types.NamespacedName{
+		Namespace: input.Cluster.Spec.ControlPlaneRef.Namespace,
+		Name:      input.Cluster.Spec.ControlPlaneRef.Name,
+	}, amcp)
+	Expect(err).NotTo(HaveOccurred())
+
+	ammp := &infrav1exp.AzureManagedMachinePool{}
+	err = mgmtClient.Get(ctx, client.ObjectKeyFromObject(input.MachinePool), ammp)
+	Expect(err).NotTo(HaveOccurred())
```
jackfrancis, line 386: How about instead of hard-coding an initial expectation that autoscaling is disabled we simply get the current configuration at the beginning of this flow, and then invert it?

jackfrancis: This way we can establish a flexible configuration A/B pattern that adapts to however the underlying cluster template evolves over time.

I can also imagine something like the following funcs that support this pattern:

`func ValidateAKSAutoscaleDisabled(ctx context.Context, inputGetter func() AKSAutoscaleSpecInput) {`

`func ValidateAKSAutoscaleEnabled(ctx context.Context, inputGetter func() AKSAutoscaleSpecInput) {`

If we detect in our initial `AKSAutoscaleSpec` flow that autoscaling is disabled for the pool in question, then we can invoke `ValidateAKSAutoscaleDisabled`, which would do something like add more pod replicas than can be scheduled on this pool, and validate that no scaling occurs. After that validation succeeds, we could then enabled the autoscaler (which is a part of our original test to confirm that mutable configurations can in fact be mutated), and then invoke `ValidateAKSAutoscaleEnabled`, which would do a similar pod replica scheduling thing, but this time validate a successful, automatic pool scale out.

And reverse the order of the above if the original cluster template config had the autoscaler enabled.

nojnhuh: That makes sense. I just pushed a change that I think is along these lines.

jackfrancis: This is nice! The next thing we want to do is to actually engage the autoscaler by scheduling pods. We can do that in a follow-up PR, let's land this ASAP so we have a nice pattern in the E2E suite that demonstrates this type of on/off/on/off feature testing.

PR #2765, opened 2022-11-01
add managed machine pool scale tests

nojnhuh:
**What type of PR is this?**
/kind cleanup

**What this PR does / why we need it**: This PR adds scaling of the MachinePools in an AzureManagedCluster up and down to replicate CAPI's MachinePool tests.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```


test/e2e/aks.go
```diff
@@ -355,6 +355,68 @@ func GetLatestStableAKSKubernetesVersion(ctx context.Context, subscriptionID, lo
...
+	Cluster         *clusterv1.Cluster
+	MachinePools    []*expv1.MachinePool
+	InitialReplicas int32
+	WaitIntervals   []interface{}
+}
+
+func AKSMachinePoolSpec(ctx context.Context, inputGetter func() AKSMachinePoolSpecInput) {
+	input := inputGetter()
+
+	By("Scaling the machine pools up")
```
jackfrancis, line 368: nit: s/up/out

test/e2e/aks.go
```diff
@@ -355,6 +355,68 @@ func GetLatestStableAKSKubernetesVersion(ctx context.Context, subscriptionID, lo
...
+	framework.ScaleMachinePoolAndWait(ctx, framework.ScaleMachinePoolAndWaitInput{
+		ClusterProxy:              bootstrapClusterProxy,
+		Cluster:                   input.Cluster,
+		Replicas:                  input.InitialReplicas + 1,
+		MachinePools:              input.MachinePools,
+		WaitForMachinePoolToScale: input.WaitIntervals,
+	})
+
+	By("Scaling the machine pools down")
```
jackfrancis, line 377: nit: s/down/in

test/e2e/aks.go
```diff
@@ -355,6 +355,68 @@ func GetLatestStableAKSKubernetesVersion(ctx context.Context, subscriptionID, lo
...
+}
+
+func AKSMachinePoolSpec(ctx context.Context, inputGetter func() AKSMachinePoolSpecInput) {
+	input := inputGetter()
+
+	By("Scaling the machine pools out")
+	framework.ScaleMachinePoolAndWait(ctx, framework.ScaleMachinePoolAndWaitInput{
+		ClusterProxy:              bootstrapClusterProxy,
+		Cluster:                   input.Cluster,
+		Replicas:                  input.InitialReplicas + 1,
```
jackfrancis, line 372: Seems like the scaling test failed which is interesting of itself, but I would echo my comments on the prior E2E PR that rather than take an `InitialReplicas` as input let's just get the input replicas count in real-time by getting the machinepool resource and then adding 1 to that for the scale out value.

nojnhuh: Yeah that makes sense. And since this test scales all of the machine pools, do you think it's also worth removing the assumption that all machine pools have the same number of replicas? Or should we only scale one machine pool at a time?

jackfrancis: I think AKS only allows one node pool scaling event at a time (that assumption should be validated), in which case we'll have to serialize.

Yeah, I think we just enumerate through all pools in the cluster, scale out by <whatever>, validate, then scale back in, validate. Ideally we can do those all concurrently in discrete goroutines to economize for E2E duration, but as I suggested above that might not be possible.

nojnhuh: When I was first testing this, scaling all the machine pools at once seemed to work. I can see where iterating over each one would make it easier to more granularly tweak replica counts, and that would probably be simple enough to wrap in a loop and do in parallel.

test/e2e/azure_test.go
```diff
@@ -571,7 +572,7 @@ var _ = Describe("Workload cluster creation", func() {
 					ClusterName:              clusterName,
 					KubernetesVersion:        kubernetesVersion,
 					ControlPlaneMachineCount: pointer.Int64Ptr(1),
-					WorkerMachineCount:       pointer.Int64Ptr(1),
+					WorkerMachineCount:       replicas,
```
willie-yao, line 575: nit: Is there a reason why `replicas` is stored as a variable here, but in other tests in this file, they are declared directly?

willie-yao (commented):
Everything looks good to me! I just had one small comment that I left inline.

test/e2e/azure_test.go
nojnhuh: I think this is left over from when I was using this value in another place too. I'll change this back.

jackfrancis (approved):
/lgtm
/approve

test/e2e/aks.go
nawazkh: Wow. What does `s/up/out` mean?

nojnhuh: "Replace 'up' with 'out'" a la `sed`:
```
% sed 's/up/out/' <<< "Scaling the machine pools up"
Scaling the machine pools out
```

nawazkh: Thanks!

PR #2767, opened 2022-11-01
Cleanup remote peerings when resource group is deleted

### LINKED ISSUE #2638: Remote Peerings are left in a disconnected state instead of being deleted
/kind bug

given the scenario:
  - Cluster is deployed with Capz
  - Resource group is owned by Capz
  - vnet peerings were specified to be created
  - the vnet peerings are to a vnet in a differnet resource group

when the cluster is deleted it will delete the resource group which deletes everything underneath it. however, if any of the vnets are peered with a vnet in a different resource group the other side of the peering will be left in a disconnected state instead of being deleted. At this point redeploying the cluster fails due to the left over peering being stuck in a disconnected state and must be manually cleaned up before being deleted

However, if the resource group is not owned by capz, the reconciler will run through each object inside of the resource group for deletion, in this scenario the remote peerings are successfully deleted

The logic for this may be found at https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/controllers/azurecluster_reconciler.go#L116

When a cluster is deleted, all peerings are cleaned up

**Environment:**

willie-yao:
**What type of PR is this?**
/kind bug

When a cluster owned by CAPZ is deleted, the resource group is deleted entirely, but the vnet remote peering is left in a disconnected state instead of deleted. This is because the remote peering is technically not part of the resource group itself. This PR adds an explicit delete for vnet peerings when the resource group is deleted.

Fixes #2638 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Cleanup remote peerings when resource group is deleted
```

### DIFF
--- a/controllers/azurecluster_reconciler.go
+++ b/controllers/azurecluster_reconciler.go
@@ -114,7 +114,16 @@ func (s *azureClusterService) Delete(ctx context.Context) error {
 		return errors.Wrap(err, "failed to determine if the AzureCluster resource group is managed")
 	}
 	if managed {
-		// if the resource group is managed, we delete the entire resource group directly.
+		// If the resource group is managed, delete it.
+		// We need to explicitly delete vnet peerings, as it is not part of the resource group.
+		vnetPeeringsSvc, err := s.getService(vnetpeerings.ServiceName)
+		if err != nil {
+			return errors.Wrap(err, "failed to get vnet peerings service")
+		}
+		if err := vnetPeeringsSvc.Delete(ctx); err != nil {
+			return errors.Wrap(err, "failed to delete peerings")
+		}
+		// Delete the entire resource group directly.
 		if err := groupSvc.Delete(ctx); err != nil {
 			return errors.Wrap(err, "failed to delete resource group")
 		}
--- a/azure/services/vnetpeerings/vnetpeerings.go
+++ b/azure/services/vnetpeerings/vnetpeerings.go
@@ -26,7 +26,8 @@ import (
 	"sigs.k8s.io/cluster-api-provider-azure/util/tele"
 )
 
-const serviceName = "vnetpeerings"
+// ServiceName is the name of this service.
+const ServiceName = "vnetpeerings"
 
 // VnetPeeringScope defines the scope interface for a subnet service.
 type VnetPeeringScope interface {
@@ -52,7 +53,7 @@ func New(scope VnetPeeringScope) *Service {
 
 // Name returns the service name.
 func (s *Service) Name() string {
-	return serviceName
+	return ServiceName
 }
 
 // Reconcile idempotently creates or updates a peering.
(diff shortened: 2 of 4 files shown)


controllers/azurecluster_reconciler.go
```diff
@@ -114,7 +114,16 @@ func (s *azureClusterService) Delete(ctx context.Context) error {
 		return errors.Wrap(err, "failed to determine if the AzureCluster resource group is managed")
 	}
 	if managed {
-		// if the resource group is managed, we delete the entire resource group directly.
+		// If the resource group is managed, delete it.
+		// We need to explicitly delete vnet peerings, as it is not part of the resource group.
+		vnetPeeringsSvc, err := s.getService(vnetpeerings.ServiceName)
```
jackfrancis, line 119: Does an AzureCluster resource _always_ have an associated VNET peering service?

willie-yao: It does not. I saw that the older version of this delete operation deleted resources like this [one by one](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2146/files#diff-5cf036f0f896f15c8f934e257474a75b1e3a4fd6f9e513a9fdfca962715564baL127), so I assumed that it would not return an error in the case of the resource not having a VNET peering service. I will test again to see if this is the intended behavior.

jackfrancis: If it is fine for a cluster not to have an associated VNET peering service, then I think we don't want to short-circuit the delete flow when we receive and error GET'ing a non-existent VNET peering service.

willie-yao: I just tested this while deleting a cluster without a peering, and the error isn't reached if there are no vnet peerings on the resource. It doesn't short circuit the delete flow, so I think the error is only reached when the deletion fails. Would this be okay or should I handle the deletion in a different way?

jackfrancis: Was your test of a "managed" type, without a peering? I'm just reading through the code and it seems like for every such type, during deletion, we get the associated VNET peering service and return early with an error if the GET fails.

willie-yao: Yes it was. I added debugging print statements within the code and the `getService` is not failing if there is no vnet peering. I could be understanding this incorrectly, but that's how it's working through my test.

It looks like the service list is being defined [here](https://github.com/willie-yao/cluster-api-provider-azure/blob/076783ee4a3010a414677efc88570e1201a87cf2/controllers/azurecluster_reconciler.go#L60) regardless of whether a particular service exists in the resource group or not, so `[getService](https://github.com/willie-yao/cluster-api-provider-azure/blob/076783ee4a3010a414677efc88570e1201a87cf2/controllers/azurecluster_reconciler.go#L143)` won't return an error.

CecileRobertMichon: Delete operation is a no-op if no vnet peerings exist and is short-circuited here: https://github.com/willie-yao/cluster-api-provider-azure/blob/main/azure/services/vnetpeerings/vnetpeerings.go#L97

There is always a service defined, for every resource type, but each service short circuits reconcile/delete whenever there are no resources of that type for that AzureCluster, so this should be fine 👍

nojnhuh (commented):
/lgtm

PR #2802, opened 2022-11-09
tags converters: return nil if src is nil

### LINKED ISSUE #2801: Tag converters return map[string]*string instead of nil when the source is nil
/kind bug

Tag converters should return `nil` when the source is `nil`. Instead the function returns an empty `map[string]*string` type. You see the following snippets `return tags` where `tags` cannot be `nil`.

https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/2349883d1effe6344690459b40e5dc26138010df/azure/converters/tags.go#L24-L33

https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/2349883d1effe6344690459b40e5dc26138010df/azure/converters/tags.go#L35-L44

Return `nil` when the source is `nil`.

zioproto:
Fixes #2801 
/kind bug

these converter functions had a bug where the conversion of a nil source was returning an empty `map[string]*string`
this behaviour triggered bugs when doing diff operations on converted objects

This is where this problem was encountered the first time:
https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2745/files#r1012288112

Release notes:
```release-note
Fixed tags converters functions MapToTags and TagsToMap to return nil on nil input. Added unit tests for tag converters.
```

### DIFF
--- a/azure/converters/tags.go
+++ b/azure/converters/tags.go
@@ -23,6 +23,10 @@ import (
 
 // MapToTags converts a map[string]*string into a infrav1.Tags.
 func MapToTags(src map[string]*string) infrav1.Tags {
+	if src == nil {
+		return nil
+	}
+
 	tags := make(infrav1.Tags, len(src))
 
 	for k, v := range src {
@@ -34,6 +38,10 @@ func MapToTags(src map[string]*string) infrav1.Tags {
 
 // TagsToMap converts infrav1.Tags into a map[string]*string.
 func TagsToMap(src infrav1.Tags) map[string]*string {
+	if src == nil {
+		return nil
+	}
+
 	tags := make(map[string]*string, len(src))
 
 	for k, v := range src {
--- a/azure/converters/tags_test.go
+++ b/azure/converters/tags_test.go
@@ -1,0 +1,151 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package converters
+
+import (
+	"testing"
+
+	"github.com/Azure/go-autorest/autorest/to"
+	"github.com/onsi/gomega"
+	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
+)
+
+func Test_TagsToMap(t *testing.T) {
+	cases := []struct {
+		tags   infrav1.Tags
+		expect map[string]*string
+	}{
+		{
+			tags:   nil,
+			expect: nil,
+		},
+		{
+			tags:   infrav1.Tags{},
+			expect: map[string]*string{},
+		},
+		{
+			tags:   infrav1.Tags{"env": "prod"},
+			expect: map[string]*string{"env": to.StringPtr("prod")},
+		},
+	}
+
+	for _, c := range cases {
+		c := c
+		t.Run("name", func(t *testing.T) {
+			t.Parallel()
+			g := gomega.NewGomegaWithT(t)
+			result := TagsToMap(c.tags)
+			g.Expect(c.expect).To(gomega.BeEquivalentTo(result))
+		})
+	}
+}
+
+func Test_MapToTags(t *testing.T) {
+	cases := []struct {
+		tags   map[string]*string
+		expect infrav1.Tags
+	}{
+		{
+			tags:   nil,
+			expect: nil,
+		},
+		{
+			tags:   map[string]*string{},
+			expect: infrav1.Tags{},
+		},
+		{
+			tags:   map[string]*string{"env": to.StringPtr("prod")},
+			expect: infrav1.Tags{"env": "prod"},
+		},
+	}
+
+	for _, c := range cases {
+		c := c
+		t.Run("name", func(t *testing.T) {
+			t.Parallel()
+			g := gomega.NewGomegaWithT(t)
+			result := MapToTags(c.tags)
+			g.Expect(c.expect).To(gomega.BeEquivalentTo(result))
+		})
+	}
+}
+
+// convert the value to map then back to tags
+// and make sure that the value we get is strictly equal
+// to the original value (ie. round trip conversion).
+func Test_TagsToMapRoundTrip(t *testing.T) {
+	cases := []struct {
+		tags   infrav1.Tags
+		expect infrav1.Tags
+	}{
+		{
+			tags:   nil,
+			expect: nil,
+		},
+		{
+			tags:   infrav1.Tags{},
+			expect: infrav1.Tags{},
+		},
+		{
+			tags:   infrav1.Tags{"env": "prod"},
+			expect: infrav1.Tags{"env": "prod"},
+		},
+	}
+
+	for _, c := range cases {
+		c := c
+		t.Run("name", func(t *testing.T) {
+			t.Parallel()
+			g := gomega.NewGomegaWithT(t)
+			result := MapToTags(TagsToMap(c.tags))
+			g.Expect(c.expect).To(gomega.BeEquivalentTo(result))
+		})
+	}
+}
+
+// convert the value to tags then back to map
+// and make sure that the value we get is strictly equal
+// to the original value (ie. round trip conversion).
+func Test_MapToTagsMapRoundTrip(t *testing.T) {
+	cases := []struct {
+		tags   map[string]*string
+		expect map[string]*string
+	}{
+		{
+			tags:   nil,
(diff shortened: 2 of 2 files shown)


azure/converters/tags.go
```diff
@@ -23,6 +23,10 @@ import (
 
 // MapToTags converts a map[string]*string into a infrav1.Tags.
 func MapToTags(src map[string]*string) infrav1.Tags {
+	if src == nil {
```
CecileRobertMichon, line 26: can you please add a unit test that covers this scenario?

zioproto: @CecileRobertMichon I now added unit tests for `MapToTags` and `TagsToMap`. Unit tests were missing completely for tag converters if I looked in the correct files, so I created a completely new file. Please review. thanks

mboersma:
When I grep the code for usages of `MapToTags`, I find several places where code seems to expect a non-`nil` return value:

```golang
func (s *Service) isResourceManaged(tags map[string]*string) bool {
	return converters.MapToTags(tags).HasOwned(s.Scope.ClusterName())
}
```

We should add `nil` checks or other appropriate refactoring to account for this change.

jackfrancis:
> When I grep the code for usages of `MapToTags`, I find several places where code seems to expect a non-`nil` return value:
> 
> ```go
> func (s *Service) isResourceManaged(tags map[string]*string) bool {
> 	return converters.MapToTags(tags).HasOwned(s.Scope.ClusterName())
> }
> ```
> 
> We should add `nil` checks or other appropriate refactoring to account for this change.

Would returning a zero value tags object (`infrav1.Tags{}`) be a better choice here instead of returning `nil`?

nawazkh:
> > When I grep the code for usages of `MapToTags`, I find several places where code seems to expect a non-`nil` return value:
> > ```go
> > func (s *Service) isResourceManaged(tags map[string]*string) bool {
> > 	return converters.MapToTags(tags).HasOwned(s.Scope.ClusterName())
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > We should add `nil` checks or other appropriate refactoring to account for this change.
> 
> Would returning a zero value tags object (`infrav1.Tags{}`) be a better choice here instead of returning `nil`?

Returning a zero value tags object (`infrav1.Tags{}`) might lead to throttling again as [being discussed here](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2745#discussion_r1012288112)?

zioproto:
I started this PR because the assumption was "it should return nil if the src is nil":
https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2745#discussion_r1017067725

Before doing additional changes I need to understand if this assumption is correct.

@CecileRobertMichon ☝️ 

Thanks :)

CecileRobertMichon:
> We should add nil checks or other appropriate refactoring to account for this change.

+1

Let's audit the code to make sure we are nil checking places where the func is used as needed. 

> Would returning a zero value tags object (infrav1.Tags{}) be a better choice here instead of returning nil?

This is what we're currently doing and that is leading to inconsistencies when the src is nil as discussed [here](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2745#discussion_r1017067725). Since the function's role is to convert a map variable that can have a nil value, IMO it's a bug to not return nil when the src variable is nil, and the functions in the codebase relying on that faulty behavior should be updated.

azure/converters/tags_test.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/onsi/gomega"
+	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
+)
+
+func Test_TagsToMap(t *testing.T) {
+	cases := []struct {
+		tags   infrav1.Tags
+		expect map[string]*string
+	}{
+		{
```
CecileRobertMichon, line 32: let's add a bit more unit test cases to cover other scenarios when the src isn't nil as well

CecileRobertMichon: maybe a good test would be to convert the value to map then back to tags and make sure that the value we get is strictly equal to the original value (ie. round trip conversion), since that's the bug we're fixing here.

sonasingh46:
If we are planning to return a nil for nil then we should also keep a track of this method here. 
https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/api/v1beta1/tags.go#L39

For example here, 
https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/services/privatedns/privatedns.go#L149
We are never passing a nil to the `MapsToTags` method so it is safe to call `HasOwned`

But here,
https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/services/groups/groups.go#L133
If a nil tag gets passed to `MapsToTags` then the call to `HasOwned` will panic if  `MapsToTags` returns a nil.

sonasingh46:
> Returning a zero value tags object (infrav1.Tags{}) might lead to throttling
Also @nawazkh can you please help me grok this.

nojnhuh:
> When I grep the code for usages of `MapToTags`, I find several places where code seems to expect a non-`nil` return value:
> 
> ```go
> func (s *Service) isResourceManaged(tags map[string]*string) bool {
> 	return converters.MapToTags(tags).HasOwned(s.Scope.ClusterName())
> }
> ```
> 
> We should add `nil` checks or other appropriate refactoring to account for this change.

FWIW, calling a method on or accessing a nil map won't necessarily panic. e.g. https://go.dev/play/p/-iu_DXwnIrE

So we may not need to change much (if anything). Unit test coverage to verify would be good though.

sonasingh46:
@nojnhuh -- I believe it will panic as the method is value receiver.

sonasingh46:
Ok.. My bad. I don't why I am under impression that when method is value receiver and called over a nil pointer, it panics. ( Could be wrong though )

zioproto:
I pushed the following additional changes:

* nil checking where tag converters are used
* Implemented round trip conversion unit test

I did not squash commits yet to make it easier to audit changes in this PR and eventually rollback individual commits.

Please review and give me feedback when I can squash commits.

Thank you !

azure/converters/tags_test.go
```diff
@@ -75,3 +75,61 @@ func Test_MapToTags(t *testing.T) {
...
+// and make sure that the value we get is strictly equal
+// to the original value (ie. round trip conversion)
+func Test_TagsToMapRoundTrip(t *testing.T) {
+	cases := []struct {
+		tags   infrav1.Tags
+		expect infrav1.Tags
+	}{
+		{
+			tags:   nil,
+			expect: nil,
```
sonasingh46, line 89: Also worth to add one more case that has empty tags. 
```go
{

tags:   map[string]*string{},
expect: map[string]*string{},

}

```

sonasingh46:
@zioproto -- You can squash the commits. Looks good to me.

zioproto:
@sonasingh46 commits squashed and rebased on current `origin/main`. Please apply the lgtm label again because the automation removed it with the squash

sonasingh46 (approved):

api/v1alpha3/types.go
```diff
@@ -87,6 +87,9 @@ type VnetSpec struct {
 
 // IsManaged returns true if the vnet is managed.
 func (v *VnetSpec) IsManaged(clusterName string) bool {
+	if v.Tags == nil {
```
CecileRobertMichon, line 90: I don't think this is needed as @nojnhuh pointed out

` Tags`  is a `map[string]string` , not a pointer to a map, so even if the map is nil it won't panic:

https://go.dev/play/p/-iu_DXwnIrE

azure/converters/tags_test.go
CecileRobertMichon: thank you for adding those!

azure/converters/vm.go
```diff
@@ -61,7 +61,7 @@ func SDKToVM(v compute.VirtualMachine) *VM {
 		vm.AvailabilityZone = to.StringSlice(v.Zones)[0]
 	}
 
-	if len(v.Tags) > 0 {
+	if v.Tags != nil && len(v.Tags) > 0 {
```
CecileRobertMichon, line 64: why do we need this extra check? a nil v.Tags will already not pass the length check https://go.dev/play/p/QoprOQrAlzV

CecileRobertMichon:
> FWIW, calling a method on or accessing a nil map won't necessarily panic. e.g. https://go.dev/play/p/-iu_DXwnIrE

@zioproto, @nojnhuh is correct that calling a method on a nil map doesn't panic so I don't think we need all the extra nil pointer checks. Thanks for adding all the unit tests.

zioproto:
@nojnhuh @CecileRobertMichon I pushed a new commit to address the latest review comments

zioproto:
@sonasingh46 @CecileRobertMichon please take a look again. It should be ready for merge now. Thanks

CecileRobertMichon:
/lgtm
/assign @sonasingh46 @nojnhuh 

@zioproto this is missing a release note

sonasingh46 (approved):
/lgtm

nawazkh:
Looks good. Thanks!

zioproto:
Release note added, thanks

CecileRobertMichon:
/retest

Error curling Helm 👀

PR #2803, opened 2022-11-09
refresh custom data automatically

### LINKED ISSUE #2683: MachinePool bootstrap token do not get refreshed automatically when VMSS is manually/externally scaled
/kind bug

A cluster with MachinePools and [externally managed autoscaler](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2588) is necessary for this bug to appear. 

The reconciliation loop for AzureMachinePool does not automatically refresh bootstrap tokens once they get rotated. Writing of new bootstrap tokens into custom data is done only when [there is a surge change or the VMSS model changes](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/services/scalesets/scalesets.go#L265-L283). 
When scaling a VMSS manually or externally via cluster-autoscaler set to provider azure, the token in the VMSS custom data may already be outdated and therefore the new node can not join the cluster. 

I believe there might be two separate issues:
1. [`patchVMSSIfNeeded`](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/071dd134f1c4b93a937fb76d94b7722cc7343f98/azure/services/scalesets/scalesets.go#L252) does not verify if custom data changed and therefore does not update it. To do this, we'd need to store e.g. a hash of custom data in the `AzureMachinePool.Status` and compare the hashes.
2. auto rotation of Kubeadm bootstrap tokens does not result in a reconciliation of AzureMachinePool. Only a change in MachinePool leads to reconciliation of AzureMachinePool and might result in updating the VMSS custom data. 

When Kubeadm bootstrap token refreshes, the VMSS custom data gets updated automatically and new nodes can join without issues.

I'm aware that the prerequisites (MachinePools and externally managed cluster autoscaler) are a special case. I'm not sure how many users want to have an externally managed autoscaler and I'm personally testing whether I can switch to the cluster-api cluster-autoscaler provider. 
However, I do think custom data changes should be considered in `patchVMSSIfNeeded`. 

To solve point two, I believe the AzureMachinePool controller would need to watch KubeadmConfig and then kick off a reconciliation. This might be even worth to consider in CAPI itself, since I expect this to be an issue for all CAP* providers.

mweibel:
/kind feature

Stores a hash of the current custom data contents into AzureMachinePool.Status.CustomDataHash. This allows to update the VMSS data if the custom data changes. 

Fixes #2683 (partially)

This is a partial fix for #2683. It only fixes it partially because there's a missing piece. 
Let me elaborate:

KubeadmconfigBootstrap Controller [rotates MachinePool tokens every TokenTTL](https://github.com/kubernetes-sigs/cluster-api/blob/main/bootstrap/kubeadm/internal/controllers/kubeadmconfig_controller.go#L267). 
In the usual case, this updated token gets saved to the VMSS custom data if [`MachinePool.spec.replicas` change](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/071dd134f1c4b93a937fb76d94b7722cc7343f98/azure/services/scalesets/scalesets.go#L252). 
When the number of instances get scaled up externally (e.g. external cluster autoscaler), the bootstrap token does not get saved automatically. This leads to the issue that after TokenTTL the token in the VMSS is expired and new nodes can't join the cluster. 

Idea of this PR is to save & compare the hash of the custom data (since the custom data can only be saved, not retrieved anymore) and update the VMSS custom data if it changed. 

However, when the bootstrap controller changes a token, it does not lead to a reconciliation of the MachinePool. Only the secret changes. 
I added [a change](https://github.com/helio/cluster-api-provider-azure/commit/c94aa50a80701b39fcfcd5b8aa039630ebdba62b) which watches the bootstrap secret and triggers a reconcile loop of AzureMachinePool in this case. This feels hacky though.

Question: Is there a better approach? Did I misunderstand something?

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
MachinePool: ensures updated custom data gets saved to VMSS on every reconciliation
```


jackfrancis:
@mweibel FYI, I've added this workstream to the v1.7.0 release milestone. Let us know if this is too aggressive, and we can push it out further.

mweibel:
@jackfrancis would be great to have this done in 1.7 but I'm not so sure about it since it doesn't work yet. I'll probably join CAPZ office hours one of these days to discuss it. I also wanted to verify how other infra providers do this to make sure I don't miss something obvious.

jackfrancis:
@mweibel is this PR actively nearing completion? If not, should we remove from the 1.7 milestone (ETA: early January)?

Thanks!

mweibel:
@jackfrancis Let's remove the milestone for now. I don't currently find the time to advance this further without guidance.

primeroz:
>  Let's remove the milestone for now. I don't currently find the time to advance this further without guidance.

I might pick this up since is a requirement for the other change i need to push from https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/2972

> but I'm not so sure about it since it doesn't work yet.

Is there anything specific not working yet or just have not completed the changes ?

mweibel:
> Is there anything specific not working yet or just have not completed the changes ?

At the moment it does only ensures (with the hash) that custom data changes actually get persisted in the VMSS. My initial thinking was that the bootstrap token changes reach the controller but the customData change is just not persisted. However, after actually trying it out, I saw that the token change do not reach the controller unless we add some kind of watcher for the secret. 
The watcher I implemented earlier but did not complete it either: https://github.com/helio/cluster-api-provider-azure/commit/236f2e59ec675e37d49b72b6644eeabe61fea4d2

So, still some things to investigate and figure out. Does this help at all?

primeroz:
> So, still some things to investigate and figure out. Does this help at all?

sure it does :+1: 

when i get onto this i might reach out for more info :)

primeroz:
For the moment internally we decided to move back to `MachineDeployments` to focus on other things so i won't do any work on fixing this issue for now. 

I am planning to get back to it as soon as it becomes relevant for us again

CecileRobertMichon:
/close

in favor of #3134

PR #2805, opened 2022-11-10
Fix meeting time in README

zioproto:
Removing UTC time because we would have to change it every 6 months at the transitions PDT to PST and PST to PDT


mboersma (commented):
/lgtm

PR #2807, opened 2022-11-10
Default Ubuntu to 22.04 for new k8s versions

mboersma:
**What type of PR is this?**

/kind feature

Updates the default Ubuntu version to 22.04 LTS for newer Kubernetes versions that have published reference images to Azure Marketplace for it.

N/A

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Default Ubuntu to 22.04 for new k8s versions
```

### DIFF
--- a/azure/services/virtualmachineimages/images.go
+++ b/azure/services/virtualmachineimages/images.go
@@ -180,14 +180,17 @@ func (s *Service) getSKUAndVersion(ctx context.Context, location, publisher, off
 
 // getUbuntuOSVersion returns the default Ubuntu OS version for the given Kubernetes version.
 func getUbuntuOSVersion(major, minor, patch uint64) string {
-	// Default to Ubuntu 20.04 LTS, except for k8s versions which have only 18.04 reference images.
-	osVersion := "2004"
-	if (major == 1 && minor == 21 && patch < 2) ||
-		(major == 1 && minor == 20 && patch < 8) ||
-		(major == 1 && minor == 19 && patch < 12) ||
-		(major == 1 && minor == 18 && patch < 20) ||
-		(major == 1 && minor < 18) {
+	// Default to Ubuntu 22.04 LTS for Kubernetes v1.25.3 and later.
+	osVersion := "2204"
+	if major == 1 && minor == 21 && patch < 2 ||
+		major == 1 && minor == 20 && patch < 8 ||
+		major == 1 && minor == 19 && patch < 12 ||
+		major == 1 && minor == 18 && patch < 20 ||
+		major == 1 && minor < 18 {
 		osVersion = "1804"
+	} else if major == 1 && minor == 25 && patch < 3 ||
+		major == 1 && minor < 25 {
+		osVersion = "2004"
 	}
 	return osVersion
 }
--- a/azure/services/virtualmachineimages/images_test.go
+++ b/azure/services/virtualmachineimages/images_test.go
@@ -116,6 +116,16 @@ func TestGetDefaultUbuntuImage(t *testing.T) {
 			},
 		},
 		{
+			k8sVersion:      "v1.22.16",
+			expectedSKU:     "ubuntu-2004-gen1",
+			expectedVersion: "122.16.20221117",
+			versions: compute.ListVirtualMachineImageResource{
+				Value: &[]compute.VirtualMachineImageResource{
+					{Name: to.StringPtr("122.16.20221117")},
+				},
+			},
+		},
+		{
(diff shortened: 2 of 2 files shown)


azure/services/virtualmachineimages/images.go
```diff
@@ -180,14 +180,19 @@ func (s *Service) getSKUAndVersion(ctx context.Context, location, publisher, off
...
 	if (major == 1 && minor == 21 && patch < 2) ||
 		(major == 1 && minor == 20 && patch < 8) ||
 		(major == 1 && minor == 19 && patch < 12) ||
 		(major == 1 && minor == 18 && patch < 20) ||
 		(major == 1 && minor < 18) {
 		osVersion = "1804"
+	} else if (major == 1 && minor == 25 && patch < 3) ||
+		(major == 1 && minor == 24 && patch < 7) ||
+		(major == 1 && minor == 23 && patch < 13) ||
+		(major == 1 && minor < 23) {
```
willie-yao, line 194: Does this if statement cover v1.22? I think it'll fall under line 194, but I'm not sure how explicit this needs to be.

nawazkh: `v1.22.x` definitely falls in the `else if` block. I tested it here https://go.dev/play/p/LrFDOouAVUC
I also noticed that `osVersion = "1804"` gets set only for SemVers `[v1.21.0 - v1.21.1]`, any SemVers belonging to `[v1.21.2 - 1.21.x]` would also fall in the `else if` block. I am not sure if we intend that?

mboersma: I think that's the desired behavior actually. 

This is funky code, because it's capturing something external: which versions are available in the real world as reference images from the `cncf-upstream` publisher.  For v1.21.x, we starting publishing Ubuntu 20.04 versions with v1.21.2. I added some unit tests to validate this, but it's possible I missed some case.

willie-yao: That makes sense. Would we also want to add a unit test case for `v1.22.x`?

mboersma:
Updated with a case for `<v1.22.16` since we'll actually publish it for Ubuntu 22.04, and with more unit test cases.

willie-yao (approved):
lgtm!

jackfrancis (approved):
/lgtm

jackfrancis:
@mboersma feel free to cancel the hold once the referenced 22.04 images are actually live and we're able to merge this

azure/services/virtualmachineimages/images.go
```diff
@@ -180,14 +180,20 @@ func (s *Service) getSKUAndVersion(ctx context.Context, location, publisher, off
...
+	osVersion := "2204"
 	if (major == 1 && minor == 21 && patch < 2) ||
 		(major == 1 && minor == 20 && patch < 8) ||
 		(major == 1 && minor == 19 && patch < 12) ||
 		(major == 1 && minor == 18 && patch < 20) ||
 		(major == 1 && minor < 18) {
 		osVersion = "1804"
+	} else if (major == 1 && minor == 25 && patch < 3) ||
+		(major == 1 && minor == 24 && patch < 7) ||
+		(major == 1 && minor == 23 && patch < 13) ||
```
mtougeron, line 193: My understanding is that cgroupv2 is enabled by default on ubuntu 22.04. If so, I don't think we should publish images for < k8s 1.24.

CecileRobertMichon:
/lgtm cancel

Per discussion in office hours this morning

azure/services/virtualmachineimages/images.go
mboersma: Right you are (according to my Googling), thanks for pointing this out. On the [Kubernetes cgroups doc](https://kubernetes.io/docs/concepts/architecture/cgroups/#cgroup-v2), it says:

> **FEATURE STATE**: Kubernetes v1.25 [stable]

Should we restrict Ubuntu 22.04 to >= v1.25 then?

mboersma:
After researching cgroups v2 support, I updated this PR so Ubuntu 22.04 is the default only for Kubernetes v1.25.3 and forward. (We didn't actually publish 22.04 images for v1.25.2 or earlier patches.) Let me know if that's playing it too safe.

mboersma:
/approve cancel

Need fresh reviews since the intent of this PR has changed.

mboersma:
/remove-approve

Why "approve" commands aren't parallel to "lgtm" commands, I don't know.

Issue #2720, opened 2022-10-11
async scalesetvms refactor


CecileRobertMichon:
We should refactor the scalesetvms service to reuse the async.go framework instead of using its own implementation.

This might need to wait until https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/999 is done first so it doesn't interfere.

Jont828:
@CecileRobertMichon @mboersma I'm thinking I can get started on this now that #3111 is done and ready for final review/merge.

PR #2813, opened 2022-11-15
VMSS Flex support for MachinePools

### LINKED ISSUE #999: Support VMSS Flex
/kind feature

https://docs.microsoft.com/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-orchestration-modes

Once this feature is generally available (timeline unknown), we should support VMs being added to VM Scale Sets, in addition to the current way of managing Scale Sets with instances defined as part of the scale set configuration. This work should start with a design proposal. There are currently some unknowns, for example: How will this be supported by Azure Cloud provider? Will the "VM" orchestration map to `Machines` and "VMScaleSet" to `MachinePools`, or should they both be a `MachinePool` story? How will the Scale Set resource be created for a MachineDeployment for the former?

**Environment:**

### LINKED ISSUE #2987: VMSS Flex mode should be validated by webhook
/kind bug
/priority important-soon

This is a reminder to follow up to #2813 with proper validation for `orchestrationMode: Flexible`. Specifically, Kubernetes >= v1.26.0 is required, and ideally `cloud-provider: external` should be checked too.

mboersma:
**What type of PR is this?**

/kind feature

Implements VMSS Flex mode support for MachinePools.

~~There is still work to do here, I'm just getting a WIP branch out to help with collaboration. For example:~~

- ~~Flex mode still on by default~~ Fixed
- ~~MP ReadyReplicas not always syncing with actual VM count~~ Fixed
- ~~faultDomainCount still hard-coded at 3 because cache code isn't working~~ Fixed

Fixes #999
Fixes #2987

This work was started and mostly completed by @devigned and @jackfrancis. Thanks team!

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
VMSS Flex support for MachinePools
```


Tiltfile
```diff
@@ -358,8 +358,10 @@ def deploy_worker_templates(template, substitutions):
...
     flavor_cmd += "; " + helm_cmd + " repo add projectcalico https://projectcalico.docs.tigera.io/charts; " + helm_cmd + " --kubeconfig ./${CLUSTER_NAME}.kubeconfig install calico projectcalico/tigera-operator -f " + calico_values + " --namespace tigera-operator --create-namespace; kubectl --kubeconfig ./${CLUSTER_NAME}.kubeconfig apply -f ./templates/addons/calico/felix-override.yaml"
-    if "external-cloud-provider" in flavor_name:
-        flavor_cmd += "; " + helm_cmd + " --kubeconfig ./${CLUSTER_NAME}.kubeconfig install --repo https://raw.githubusercontent.com/kubernetes-sigs/cloud-provider-azure/master/helm/repo cloud-provider-azure --generate-name --set infra.clusterName=${CLUSTER_NAME}"
+    if "external-cloud-provider" in flavor_name or "flex" in flavor_name:
+        flavor_cmd += "; " + helm_cmd + " --kubeconfig ./${CLUSTER_NAME}.kubeconfig install --repo https://raw.githubusercontent.com/kubernetes-sigs/cloud-provider-azure/master/helm/repo cloud-provider-azure --generate-name --set infra.clusterName=${CLUSTER_NAME}" + \
+                      " --set cloudControllerManager.imageRepository=selfhostedk8s.azurecr.io --set cloudControllerManager.imageTag=d8d65f1" + \
```
CecileRobertMichon, line 363: Can we add a TODO code comment to remove this once the release of cloud-provider we're waiting for gets released?

azure/converters/vmss.go
```diff
@@ -61,6 +61,38 @@ func SDKToVMSS(sdkvmss compute.VirtualMachineScaleSet, sdkinstances []compute.Vi
 	return vmss
 }
 
+// SDKVMToVMSSVM converts an Azure SDK VM to a VMSS VM.
+func SDKVMToVMSSVM(sdkInstance compute.VirtualMachine) *azure.VMSSVM {
```
CecileRobertMichon, line 65: missing unit test for this?

azure/scope/machinepool.go
```diff
@@ -19,8 +19,11 @@ package scope
 import (
 	"context"
 	"encoding/base64"
+	"fmt"
+	"strings"
 	"time"
 
+	azure2 "github.com/Azure/go-autorest/autorest/azure"
```
CecileRobertMichon, line 26: do we have a better convention for naming this anywhere else in the repo already? azure2 isn't very explicit. If so this would be a good one to add to https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/.golangci.yml#L59

azure/scope/machinepool.go
```diff
@@ -330,17 +334,14 @@ func (m *MachinePoolScope) applyAzureMachinePoolMachines(ctx context.Context) er
...
-		return errors.New("machine.Name must not be empty")
+	parsed, err := azure2.ParseResourceID(machine.ID)
+	if err != nil {
+		return errors.Wrap(err, fmt.Sprintf("failed to parse resource id %q", machine.ID))
 	}
 
 	ampm := infrav1exp.AzureMachinePoolMachine{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      m.AzureMachinePool.Name + "-" + machine.InstanceID,
+			Name:      strings.ReplaceAll(parsed.ResourceName, "_", "-"),
```
CecileRobertMichon, line 344: could this break existing machines? just wondering if we're introducing a new way of naming machines that would change exisitng machines

azure/services/scalesets/scalesets.go
```diff
@@ -271,10 +270,23 @@ func (s *Service) patchVMSSIfNeeded(ctx context.Context, infraVMSS *azure.VMSS)
...
 	hasModelChanges := hasModelModifyingDifferences(infraVMSS, vmss)
-	if maxSurge > 0 && (hasModelChanges || !infraVMSS.HasEnoughLatestModelOrNotMixedModel()) {
+	var isFlex bool
+	for _, instance := range infraVMSS.Instances {
+		if instance.InstanceID == "" {
+			isFlex = true
+			break
+		}
+	}
+	var hasEnoughLatestModelOrNotMixedModel bool
```
CecileRobertMichon, line 280: Not sure what `hasEnoughLatestModelOrNotMixedModel ` means at first read. Consider adding a code comment to explain?

azure/services/scalesets/scalesets.go
```diff
@@ -795,3 +810,10 @@ func getSecurityProfile(vmssSpec azure.ScaleSetSpec, sku resourceskus.SKU) (*com
 func (s *Service) IsManaged(ctx context.Context) (bool, error) {
 	return true, nil
 }
+
+func getOrchestrationMode(modeType infrav1.OrchestrationModeType) compute.OrchestrationMode {
```
CecileRobertMichon, line 814: this probably belongs better in the converters package

CecileRobertMichon: also let's add a simple unit test even if it's trivial

azure/services/scalesetvms/scalesetvms.go
```diff
@@ -122,24 +237,53 @@ func (s *Service) Delete(ctx context.Context) error {
...
+
+// ResourceGroupName returns the name of the resource group the resource is in.
+func (vm *VMSSFlexVMGetter) ResourceGroupName() string {
+	return vm.ResourceGroup
+}
+
+// Parameters takes the existing resource and returns the desired parameters of the resource.
+// If the resource does not exist, or we do not care about existing parameters to update the resource, existing should be nil.
+// If no update is needed on the resource, Parameters should return nil.
+func (vm *VMSSFlexVMGetter) Parameters(existing interface{}) (params interface{}, err error) {
```
CecileRobertMichon, line 287: what's going on with this no-op Parameters func here? Was this an attempt to implement the ResourceSpecGetter interface but without using it yet since we're not yet using the async package?

azure/services/virtualmachines/client.go
```diff
@@ -31,11 +34,35 @@ import (
 	"sigs.k8s.io/cluster-api-provider-azure/util/tele"
 )
 
-// AzureClient contains the Azure go-sdk Client.
-type AzureClient struct {
-	virtualmachines compute.VirtualMachinesClient
+type (
+	// Client wraps go-sdk.
+	Client interface {
```
CecileRobertMichon, line 39: Do we need this to interface to be exported? Is it used anywhere outside of this package?

azure/types.go
```diff
@@ -135,6 +137,13 @@ func (vmss VMSS) InstancesByProviderID() map[string]VMSSVM {
 
 // ProviderID returns the K8s provider ID for the VMSS instance.
 func (vm VMSSVM) ProviderID() string {
+	if vm.InstanceID == "" {
+		// build providerID a bit differently for VMSS Flex
```
CecileRobertMichon, line 141: What does "a bit differently" mean? this code is not very human friendly, a more explicit code comment with a concrete example could be helpful

azure/scope/machinepool.go
mboersma: See #2958

azure/services/virtualmachines/client.go
mboersma: Yes, it's used in [scalesetvms.go](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2813/files#diff-afa3aa43d3541f80b10af246e174440ced390fc557b6a907525183d4d9dd5106R51).

Tiltfile
```diff
@@ -359,7 +359,9 @@ def deploy_worker_templates(template, substitutions):
         calico_values = "./templates/addons/calico/values.yaml"
     flavor_cmd += "; " + helm_cmd + " repo add projectcalico https://projectcalico.docs.tigera.io/charts; " + helm_cmd + " --kubeconfig ./${CLUSTER_NAME}.kubeconfig install calico projectcalico/tigera-operator -f " + calico_values + " --namespace tigera-operator --create-namespace; kubectl --kubeconfig ./${CLUSTER_NAME}.kubeconfig apply -f ./templates/addons/calico/felix-override.yaml"
     if "external-cloud-provider" in flavor_name:
-        flavor_cmd += "; " + helm_cmd + " --kubeconfig ./${CLUSTER_NAME}.kubeconfig install --repo https://raw.githubusercontent.com/kubernetes-sigs/cloud-provider-azure/master/helm/repo cloud-provider-azure --generate-name --set infra.clusterName=${CLUSTER_NAME}"
+        # TODO: remove extra config once the azure-cloud-provider Helm chart release supports VMSS Flex.
```
CecileRobertMichon, line 362: Cloud-provider-azure released v1.26 last week so this should only be a matter of updating the Helm chart https://github.com/kubernetes-sigs/cloud-provider-azure/releases/tag/v1.26.0

cc @lzhecheng

Tiltfile
```diff
@@ -359,7 +359,9 @@ def deploy_worker_templates(template, substitutions):
...
+        flavor_cmd += "; " + helm_cmd + " --kubeconfig ./${CLUSTER_NAME}.kubeconfig install --repo https://raw.githubusercontent.com/kubernetes-sigs/cloud-provider-azure/master/helm/repo cloud-provider-azure --generate-name --set infra.clusterName=${CLUSTER_NAME}" + \
+                      " --set cloudControllerManager.imageTag=v1.26.0 --set cloudNodeManager.imageTag=v1.26.0"
```
CecileRobertMichon, line 364: this forces the version of k8s to 1.26 which might not be the case (in fact the tiltfile default for k8s version right now is 1.24 https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2813/files#diff-c2ee8653e1d6b85f0aadf87cd438a9250806c052877248442be4d434cbc52425R24). I think it might make sense to revert this change and then enforce "you must use k8s 1.26+ to enable vmss flex"

azure/services/scalesetvms/scalesetvms.go
```diff
@@ -143,3 +257,32 @@ func (s *Service) Delete(ctx context.Context) error {
...
+
+// ResourceGroupName returns the name of the resource group the resource is in.
+func (vm *VMSSFlexVMGetter) ResourceGroupName() string {
+	return vm.ResourceGroup
+}
+
+// Parameters takes the existing resource and returns the desired parameters of the resource.
+// If the resource does not exist, or we do not care about existing parameters to update the resource, existing should be nil.
+// If no update is needed on the resource, Parameters should return nil.
+func (vm *VMSSFlexVMGetter) Parameters(ctx context.Context, existing interface{}) (params interface{}, err error) {
```
CecileRobertMichon, line 286: can you add a comment here on why Parameters is a no-op? Assuming it's because #2720 hasn't been implemented yet

azure/services/virtualmachines/client.go
```diff
@@ -169,3 +211,52 @@ func (ac *AzureClient) Result(ctx context.Context, future azureautorest.FutureAP
 		return nil, errors.Errorf("unknown future type %q", futureType)
 	}
 }
+
+// GetResultIfDone fetches the result of a long-running operation future if it is done.
+func (ac *AzureClient) GetResultIfDone(ctx context.Context, future *infrav1.Future) (compute.VirtualMachine, error) {
```
CecileRobertMichon, line 216: we can probably clean this up a bit but let's leave that for #2720

templates/flavors/external-cloud-provider-machinepool/kustomization.yaml
```diff
@@ -0,0 +1,10 @@
+namespace: default
+resources:
+  - ../base
```
CecileRobertMichon, line 3: If you start this from https://github.com/kubernetes-sigs/cluster-api-provider-azure/tree/main/templates/flavors/machinepool it would already have the machinepool and azureclusteridentity (then we can just patch it for flex and OOT)

templates/test/ci/prow-external-cloud-provider/kustomization.yaml
```diff
@@ -2,10 +2,11 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 namespace: default
 resources:
-  - ../../../flavors/external-cloud-provider
+  - ../../../flavors/external-cloud-provider-machinepool
```
CecileRobertMichon, line 5: would it make sense to just have one template that has both md and mp to test OOT cloud provider? Seems like we're getting rid of md by doing this? The base prow template has both I believe

templates/test/ci/cluster-template-prow-external-cloud-provider.yaml
```diff
@@ -240,3 +243,17 @@ spec:
     namespace: ${AZURE_CLUSTER_IDENTITY_SECRET_NAMESPACE}
   tenantID: ${AZURE_TENANT_ID}
   type: ServicePrincipal
+---
+apiVersion: addons.cluster.x-k8s.io/v1beta1
+kind: ClusterResourceSet
+metadata:
+  name: ${CLUSTER_NAME}-calico-windows
```
CecileRobertMichon, line 250: are there windows nodes in this template?

templates/test/ci/prow-external-cloud-provider/kustomization.yaml
CecileRobertMichon: Also, are we testing both Flex and Uniform?

test/e2e/cloud-provider-azure.go
```diff
@@ -45,7 +45,13 @@ func InstallCalicoAndCloudProviderAzureHelmChart(ctx context.Context, input clus
 	specName := "cloud-provider-azure-install"
 	By("Installing the correct version of cloud-provider-azure components via helm")
 	options := &helmVals.Options{
-		Values: []string{fmt.Sprintf("infra.clusterName=%s", input.ConfigCluster.ClusterName), fmt.Sprintf("cloudControllerManager.clusterCIDR=%s", strings.Join(cidrBlocks, `,`))},
+		Values: []string{
+			fmt.Sprintf("infra.clusterName=%s", input.ConfigCluster.ClusterName),
+			fmt.Sprintf("cloudControllerManager.clusterCIDR=%s", strings.Join(cidrBlocks, `,`)),
+			// TODO: remove extra config once the azure-cloud-provider Helm chart release supports VMSS Flex.
+			"cloudControllerManager.imageTag=v1.26.0",
```
CecileRobertMichon, line 52: same comment as Tilt here, we should be able to fix this now that cloud provider 1.26 is out, but we shouldn't use v1.26 unless k8s version is also 1.26 as it may lead to incompatibility. This also means we can't test Flex unless we use k8s v1.26 in theory.

mboersma: In theory, although in practice it seems to work fine with k8s 1.24 and 1.25. Previously we had it hard-coded to a pre-release SHA.

But I agree, we should limit this to k8s >= 1.26. I'll make that change.

Tiltfile
mboersma: That release did not update the Helm chart, so provisioning with Kubernetes v1.26.0 gives an error about `imageTag required` without this workaround.

CecileRobertMichon: Right, I'm saying we should update the Helm chart (@lzhecheng @jackfrancis are usually the ones who maintain it but anyone can make a PR).

mboersma: See kubernetes-sigs/cloud-provider-azure#2994

test/e2e/cloud-provider-azure.go
CecileRobertMichon: Could we also document this limitation and the requirement for OOT cloud-provider?

azure/services/scalesets/scalesets.go
```diff
@@ -517,6 +526,19 @@ func (s *Service) buildVMSSFromSpec(ctx context.Context, vmssSpec azure.ScaleSet
 		},
 	}
 
+	if orchestrationMode == compute.OrchestrationModeUniform {
```
devigned, line 529: total nit, but in this case, I'd probably lean toward a switch statement.

config/crd/bases/infrastructure.cluster.x-k8s.io_azuremachinepools.yaml
```diff
@@ -1380,6 +1380,14 @@ spec:
                   meaning that the node can be drained without any time limitations.
                   NOTE: NodeDrainTimeout is different from `kubectl drain --timeout`'
                 type: string
+              orchestrationMode:
+                default: Uniform
```
devigned, line 1384: I thought we discussed that we'd default to Flexible going forward and convert previous versions setting Uniform on them. I believe this line will default to Uniform if not specified.

devigned (changes requested):
Little nit, and quick clarification on defaulting of orchestration mode.

config/crd/bases/infrastructure.cluster.x-k8s.io_azuremachinepools.yaml
mboersma: When I discussed with @CecileRobertMichon, we thought it was safer to leave the default as `Uniform` and require users to opt in. Especially since VMSS Flex support requires out-of-tree cloud provider and Kubernetes >= 1.26.

devigned: Makes sense. Thank you for the explanation.

devigned (approved):
/lgtm

Glad to see this functionality nearing merge.

templates/test/ci/prow-external-cloud-provider/kustomization.yaml
mboersma: I reverted this changed and instead created a new `prow-external-cloud-provider-machinepool` test template. It inherits from the `external-cloud-provider-machinepool` flavor, so they're both just testing VMSS Flex with Linux currently.

I also added a new e2e test spec for this. I would like to expand it to include both Flex and Uniform for Linux and Windows (4 node pools total), but it seems safer to start small.

templates/test/ci/cluster-template-prow-external-cloud-provider.yaml
mboersma: No, but this change should be reverted now.

Tiltfile
mboersma: I added this restriction in documentation, but adding it to validation seems tricky. The natural place to do this is when validating an `AzureMachinePool` and seeing `orchestrationMode: Flexible`. 

But the Kubernetes `version` is on the owner `MachinePool`. I can find it by iterating the owners, but I'm not sure how to fetch the actual resource through the OwnerRef or if that would be appropriate in the validation code.

It would also be ideal if this validation could check for external cloud provider, but that faces similar challenges.

zmyzheng (approved):

Tiltfile
CecileRobertMichon: Copy/Pasting since the answer is similar: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2965#discussion_r1061046246

Want to open a follow up issue to add validation via webhook?

mboersma: See #2987.

test/e2e/azure_test.go
```diff
@@ -556,6 +556,71 @@ var _ = Describe("Workload cluster creation", func() {
 		})
 	})
 
+	// ci-e2e.sh and Prow CI skip this test by default. To include this test, set `GINKGO_SKIP=""`.
+	// This spec expects a user-assigned identity named "cloud-provider-user-identity" in a "capz-ci"
+	// resource group. Override these defaults by setting the USER_IDENTITY and CI_RG environment variables.
+	Context("Creating a cluster that uses the external cloud provider and machinepools [OPTIONAL]", func() {
+		It("with 1 control plane node and 1 machinepool", func() {
+			By("using user-assigned identity")
+			clusterName = getClusterName(clusterNamePrefix, "oot")
```
CecileRobertMichon, line 565:
```suggestion
			clusterName = getClusterName(clusterNamePrefix, "oot-flex")
```

so we can differentiate with the other oot cluster in logs

CecileRobertMichon:
lgtm however the flex test failed with a failure to provision the LB in CCM, @mboersma @zmyzheng PTAL

https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/2813/pull-cluster-api-provider-azure-e2e-optional/1610780765266644992/artifacts/clusters/capz-e2e-aclv4h-oot/kube-system/cloud-controller-manager-6555775b6d-4hwk6/cloud-controller-manager.log

mboersma:
I switched to an...errr...`switch` statement as suggested and renamed the test spec to "flex". Hopefully the e2e-optional test failure was a flake–it's been passing for me locally.

zmyzheng:
> lgtm however the flex test failed with a failure to provision the LB in CCM, @mboersma @zmyzheng PTAL
> 
> https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/2813/pull-cluster-api-provider-azure-e2e-optional/1610780765266644992/artifacts/clusters/capz-e2e-aclv4h-oot/kube-system/cloud-controller-manager-6555775b6d-4hwk6/cloud-controller-manager.log

I took a look at the failed test and got a little confused: 
image

Is the test creating an external LB or Internal LB? The resource name is elb but the error shows it is create an internel LB somehow

mboersma:
The LB test failed again on the Flex machinepool spec, I'll dig in to it.

mboersma:
@zmyzheng good eye--I hadn't noticed that. Maybe we're mislabeling something in the test spec? I'll look at it.

This is the specific line where the test failed: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/test/e2e/azure_lb.go#L213

azure/services/scalesets/scalesets.go
```diff
@@ -517,6 +526,19 @@ func (s *Service) buildVMSSFromSpec(ctx context.Context, vmssSpec azure.ScaleSet
...
 	}
 
+	switch orchestrationMode {
+	case compute.OrchestrationModeUniform:
+		vmss.VirtualMachineScaleSetProperties.Overprovision = to.BoolPtr(false)
+		vmss.VirtualMachineScaleSetProperties.UpgradePolicy = &compute.UpgradePolicy{Mode: compute.UpgradeModeManual}
+	case compute.OrchestrationModeFlexible:
+		vmss.VirtualMachineScaleSetProperties.VirtualMachineProfile.NetworkProfile.NetworkAPIVersion =
+			compute.NetworkAPIVersionTwoZeroTwoZeroHyphenMinusOneOneHyphenMinusZeroOne
+		vmss.VirtualMachineScaleSetProperties.PlatformFaultDomainCount = to.Int32Ptr(1)
```
jackfrancis, line 536: So basically the VMSS Flex API exposes a new `PlatformFaultDomainCount` property that is a peer property to `Zones`, and we always want to ship a VMSS Flex config where those properties have the same value?

mboersma: Yes, that's my understanding: `PlatformFaultDomainCount` is required for `OrchestrationMode: Flexible` and shouldn't be greater than the number of zones.

azure/services/scalesetvms/scalesetvms.go
```diff
@@ -100,6 +127,89 @@ func (s *Service) Delete(ctx context.Context) error {
...
+	return instanceID == ""
+}
+
+func (s *Service) deleteVMSSFlexVM(ctx context.Context, resourceID string) error {
+	ctx, log, done := tele.StartSpanWithLogger(ctx, "scalesetvms.Service.deleteVMSSFlexVM")
+	defer done()
+
+	defer func() {
+		if vm, err := s.VMClient.GetByID(ctx, resourceID); err == nil && vm.VirtualMachineProperties != nil {
+			log.V(4).Info("updating vmss vm state", "state", vm.ProvisioningState)
```
jackfrancis, line 146: Do we want to be more explicit here and suggest "delete in progress" rather than just the generic "updating vmss vm state"?

CecileRobertMichon:
@mboersma what fixed the Flex LB test?

azure/types.go
```diff
@@ -136,9 +138,24 @@ func (vmss VMSS) InstancesByProviderID() map[string]VMSSVM {
 
 // ProviderID returns the K8s provider ID for the VMSS instance.
 func (vm VMSSVM) ProviderID() string {
+	if vm.IsFlex() {
```
jackfrancis, line 141: Can we get some UT around this fun?

test/e2e/azure_test.go
```diff
@@ -556,6 +556,71 @@ var _ = Describe("Workload cluster creation", func() {
...
+			By("using user-assigned identity")
+			clusterName = getClusterName(clusterNamePrefix, "flex")
+			clusterctl.ApplyClusterTemplateAndWait(ctx, clusterctl.ApplyClusterTemplateAndWaitInput{
+				ClusterProxy: bootstrapClusterProxy,
+				ConfigCluster: clusterctl.ConfigClusterInput{
+					LogFolder:                filepath.Join(artifactFolder, "clusters", bootstrapClusterProxy.GetName()),
+					ClusterctlConfigPath:     clusterctlConfigPath,
+					KubeconfigPath:           bootstrapClusterProxy.GetKubeconfigPath(),
+					InfrastructureProvider:   clusterctl.DefaultInfrastructureProvider,
+					Flavor:                   "external-cloud-provider-machinepool",
```
jackfrancis, line 573: Would it be better to name this new flavor "external-cloud-provider-vmss-flex" to be clear about the fact that we're testing an Azure scenario (flex) vs a CAPI scenario (machinepool)?

mboersma:
> @mboersma what fixed the Flex LB test?

@CecileRobertMichon nothing unfortunately. I commented it out and replaced it with a new test spec that validates providerIDs in an AzureMachinePool to prove that it's either in Flexible or Uniform mode. The LBSpec test was copy-paste from another spec, not a deliberate decision, but it may have uncovered something pointing to cloud-provider-azure. *But* I cannot reproduce it locally if I run just that spec–it succeeded all six times I ran it yesterday.

mboersma:
@CecileRobertMichon @jackfrancis I have a working webhook validator that rejects `Flexible` if the Kubernetes version is < v1.26.0. (Validating external-cloud-provider seems problematic, so I skipped that.) I can add that to this or keep it out as a separate PR for followup.

test/e2e/azure_test.go
mboersma: That would be more on target for this PR. As I commented elsewhere, I intended this template to test the matrix of Windows|Linux * Uniform|Flexible, but started small. I'll rename it and we can rename it again when its functionality changes.

jackfrancis:
@mboersma +1 to including it in this PR

test/e2e/azure_test.go
jackfrancis: Makes sense to me, thanks!

azure/services/scalesets/scalesets.go
jackfrancis: ACK

If we wanted to expose this to the user later will this current default enforcement be forward-compatible with such a change in the future? I think so.

CecileRobertMichon:
> nothing unfortunately. I commented it out and replaced it with a new test spec that validates providerIDs

that's unfortunate, that LB spec is essentially what validated cloud-provider-azure is working since it exposes a service via LoadBalancer and expects cloud-provider-azure to add the vmss instances to the LB. It's very curious that it passes locally and not on the PR 🤔

mboersma:
> It's very curious that it passes locally and not on the PR

It made me wonder if there is somehow test pollution going on, since the LB Spec is reused. But I don't see where that could happen. It's reused several times in the suite.

It failed 100% of the time in CI and succeeded 100% of the time locally, the main differences being I ran only that single test spec locally, and I run with `SKIP_CLEANUP=true`. Let me re-enable the test and try running the full suite locally without that variable.

zmyzheng:
> > It's very curious that it passes locally and not on the PR
> 
> It made me wonder if there is somehow test pollution going on, since the LB Spec is reused. But I don't see where that could happen. It's reused several times in the suite.
> 
> It failed 100% of the time in CI and succeeded 100% of the time locally, the main differences being I ran only that single test spec locally, and I run with `SKIP_CLEANUP=true`. Let me re-enable the test and try running the full suite locally without that variable.

Hi @mboersma, is the azure.json file generated by capz? if so, could you check what's the content of the azure.json file? One thing I can think of is you might need to add a new property to that field: "enableVmssFlexNodes": true

Reference: https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/tests/k8s-azure/manifest/cluster-api/cloud-config-vmss-default.json#L24

Could you try if this can fix the test failure?

CecileRobertMichon:
@zmyzheng why does cloud-provider config require a special property to use flex? Is it a feature flag that is going to be removed in the future?

zmyzheng:
> @zmyzheng why does cloud-provider config require a special property to use flex? Is it a feature flag that is going to be removed in the future?

Hi @CecileRobertMichon, this is to prevent querying VmssFlex in the resource group when only VMSS Uniform is used in the cluster to reduce the API calls.  @nilo19 suggests using pure VMSS Uniform in CCM should be the default behavior which aligns with what AKS does today. Let me share some contexts:
CCM needs to remove VMSS from backend pool of loadbalancers (LB) before the LB is deleted. Because of the API difference between Flex and Uniform, CCM batch queries VMSS Flex and VMSS Uniform with two seperate calls and handle the LB backend pool adding/deletion seperately. This means, if we know only VMSS Uniform is used, then there is no need to query the VMSS Flex in the resource group anymore so that some VMSS Flex releated API calls can be saved.

Since CCM is used by both AKS and self-hosted K8s cluster at the same time, @nilo19 suggests my code change in CCM should not affect the behavior or performance of the current AKS pure VMSS Uniform logic. So, he suggests I add a new property "enableVmssFlexNodes" and set it as false by default. In this way, we can assume VMSS Flex is disabled by default and when AKS starts to use the latest CCM, there is no need to handle VMSS Flex related quries. This will make sure the performance of AKS will not be affected when starting using the latest CCM.

This newly added property is documented in this page: https://cloud-provider-azure.sigs.k8s.io/install/configs/

CecileRobertMichon:
Seems like Uniform VMSS test is failing...

let's see if it was a flake or an issue with the new property

/retest

mboersma:
That uniform test isn't using OOT cloud-provider though, so wouldn't this new property be ignored by the old in-tree code?

zmyzheng:
> @zmyzheng I added the `enableVmssFlexNodes: true` to azure.json and the LB spec test finally passed. Thank you!
> 
> What should we do in the (unlikely) case of a cluster that has a Flex VMSS as well as a Uniform VMSS? CAPZ uses one control-plane `azure.json` config and one `azure.json` config for all nodes; we don't have a way currently to customize it for specific nodes.
> 
> That is, if `enableVmssFlexNodes` is enabled on a Uniform VMSS cluster, does it amount to just a de-optimization, or will it cause actual problems?

It will not cause problems. If enableVmssFlexNodes is set to true, the CCM will query the VMSS Flex in the resource group. If it discovers there is no VMSS Flex, then it will not do anything.

CecileRobertMichon:
> CAPZ uses one control-plane azure.json config and one azure.json config for all nodes; we don't have a way currently to customize it for specific nodes

The azure.json file is used by the CCM pod, which is shared by all nodes, there isn't a concept of "per node" config

mboersma:
/retest

```
Service kube-system/csi-azuredisk-node-win failed
```

I don't think this failure or the previous one were related to the Flex changes, but it is suspicious...

CecileRobertMichon:
the last failure is the same flake we just fixed in #2992 

@mboersma could you please squash commits for merge?

mboersma:
/test pull-cluster-api-provider-azure-test

Linter timeout, yes really.

mboersma:
/test pull-clsuter-api-provider-azure-e2e-optional

e2e failed to create an HA cluster and e2e-optional failed on the clusterclass spec. Neither appears to be related to these changes, but that's a worrisome number of flakes. ❄️

mboersma:
I updated the `azure.json`-related code not to add `enableVmssFlexNodes` if the `EXP_MACHINE_POOLS` feature flag is off (and tested this locally).

CecileRobertMichon (commented):
/assign @jackfrancis

mboersma:

Failed to provision HA cluster.

jackfrancis (approved):
/lgtm

CecileRobertMichon:
> Error from server (NotFound): configmaps "kubeadm-config" not found

seems like a race unrelated to this change, will investigate

/retest

mboersma:
/retest

I haven't seen this before, yay. I don't think this PR caused it, so retesting.

```go
INFO: "With ipv6 worker node" started at Tue, 10 Jan 2023 20:36:35 UTC on Ginkgo node 6 of 10 and junit test report to file /logs/artifacts/test_e2e_junit.e2e_suite.1.xml
  << Timeline
  [FAILED] Timed out after 1800.000s.
  Expected success, but got an error:
      <*errors.fundamental | 0xc003fa89c0>: {
          msg: "cannot re-use a name that is still in use",
          stack: [0x34dfe66, 0x34dda45, 0x352c418, 0x14ef085, 0x14ee57c, 0x190a67a, 0x190b582, 0x1908d2d, 0x352c09b, 0x351e558, 0x3522130, 0x2f9caf0, 0x35342c8, 0x18e639b, 0x18f9e98, 0x147c741],
      }
      cannot re-use a name that is still in use
  In [It] at: /home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/test/e2e/helpers.go:879 @ 01/10/23 20:30:17.977
  Full Stack Trace
    sigs.k8s.io/cluster-api-provider-azure/test/e2e.InstallHelmChart({_, _}, {{0x4223910, 0xc000428190}, {{0xc000d03e30, 0x22}, {0xc000531510, 0x31}, {0xc000533cf3, 0x17}, ...}, ...}, ...)
    	/home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/test/e2e/helpers.go:879 +0x5db
    sigs.k8s.io/cluster-api-provider-azure/test/e2e.InstallAzureDiskCSIDriverHelmChart({_, _}, {{0x4223910, 0xc000428190}, {{0xc000d03e30, 0x22}, {0xc000531510, 0x31}, {0xc000533cf3, 0x17}, ...}, ...}, ...)
    	/home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/test/e2e/cloud-provider-azure.go:79 +0x1f8
    sigs.k8s.io/cluster-api-provider-azure/test/e2e.EnsureControlPlaneInitialized({_, _}, {{0x4223910, 0xc000428190}, {{0xc000d03e30, 0x22}, {0xc000531510, 0x31}, {0xc000533cf3, 0x17}, ...}, ...}, ...)
    	/home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/test/e2e/common.go:268 +0xb90
    sigs.k8s.io/cluster-api/test/framework/clusterctl.ApplyClusterTemplateAndWait({_, _}, {{0x4223910, 0xc000428190}, {{0xc000d03e30, 0x22}, {0xc000531510, 0x31}, {0xc000533cf3, 0x17}, ...}, ...}, ...)
    	/home/prow/go/pkg/mod/sigs.k8s.io/cluster-api/test@v1.3.1/framework/clusterctl/clusterctl_helpers.go:334 +0xd30
    sigs.k8s.io/cluster-api-provider-azure/test/e2e.glob..func1.5.1()
    	/home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/test/e2e/azure_test.go:303 +0x488
------------------------------
[SynchronizedAfterSuite] PASSED [0.000 seconds]
[SynchronizedAfterSuite] 
/home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/test/e2e/e2e_suite_test.go:116
```

jackfrancis:
@mboersma fwiw I saw that last week (on another PR)

CecileRobertMichon:
that shouldn't be happening anymore since I made the helm install idempotent in #2915, I'll look into it

jackfrancis:
👁️ 🩸

PR #2814, opened 2022-11-15
proposal for workload identity integration

sonasingh46:

Proposal for Azure Workload Identity integration to Azure.

Design proposal for #2205 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
proposal for workload identity integration
```

### DIFF
--- a/docs/proposals/20221611-workload-identity-integration.md
+++ b/docs/proposals/20221611-workload-identity-integration.md
@@ -1,0 +1,481 @@
+```yaml
+title: Workload Identity Integration
+authors:
+    - @sonasingh46
+reviewers:
+    - @aramase
+    - @CecileRobertMichon
+    - @yastij
+    - @fabriziopandini 
+
+creation-date: 2022-11-16
+last-updated: 2023-05-23
+status: implementable
+see-also:
+    - https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/2205
+```
+
+# Workload Identity Integration
+
+## <a name='TableofContents'></a>Table of Contents
+
+<!-- vscode-markdown-toc -->
+* [Table of Contents](#TableofContents)
+* [Acronyms](#Acronyms)
+* [Summary](#Summary)
+* [Motivation](#Motivation)
+	* [Goals](#Goals)
+	* [Future Goals](#FutureGoals)
+* [Personas](#Personas)
+* [User Stories](#UserStories)
+* [Proposal](#Proposal)
+	* [Implementation Details/Notes/Constraints](#ImplementationDetailsNotesConstraints)
+		* [Key Generation](#KeyGeneration)
+		* [OIDC URL Setup](#OidcUrlSetup)
+		* [Set Service Account Signing Flags](#SetServiceAccountSigningFlags)
+		* [Federated Credential](#FederatedCredential)
+		* [Distribute Keys To Management Cluster](#DistributeKeys)
+	* [Cloud Provider Azure Integration](#CloudProviderAzureIntegration)
+	* [Proposed API Changes](#ProposedApiChanges)
+	* [Proposed Deployment Configuration Changes](#ProposedConfigurationChanges)
+	* [Proposed Controller Changes](#ProposedControllerChanges)
+		* [Identity](#Identity)
+	* [Open Questions](#OpenQuestions)
+		* [1. How to achieve multi-tenancy?](#Howtomultitenancy)
+		* [2. How to distribute key pair to management cluster?](#Howtodistributekeys)
+		* [3. User Experience](#UserExperience)
+	* [Migration Plan](#MigrationPlan)
+	* [Test Plan](#TestPlan)
+* [Implementation History](#ImplementationHistory)
+
+<!-- vscode-markdown-toc-config
+	numbering=false
+	autoSave=false
+	/vscode-markdown-toc-config -->
+<!-- /vscode-markdown-toc -->
+
+## <a name='Acronyms'></a>Acronyms
+| Acronym      | Full Form               |
+| ------------ | ------------------------|
+| AD           | Active Directory        |
+| AAD          | Azure Active Directory  |
+| AZWI         | Azure Workload Identity |
+| OIDC         | OpenID Connect          |
+| JWKS         | JSON Web Key Sets       |
+
+## <a name='Summary'></a>Summary
+
+Workloads deployed in Kubernetes cluster may require Azure Active Directory application credential or managed identities to access azure protected resource e.g. Azure Key Vault, Virtual Machines etc. AAD Pod Identity helps access azure resources without the need of a secret management via Azure Managed Identities. AAD Pod Identity is now deprecated and Azure AD Workload Identity is the next iteration of the former. This design proposal aims to define the way for AZWI integration into capz for self managed clusters with keeping in mind other factor e.g. User Experience, Multi Tenancy, Backwards Compatibility etc.
+
+For more details about AAD Pod Identity please visit this [link](https://github.com/Azure/aad-pod-identity)  
+
+## <a name='Motivation'></a>Motivation
+
+AZWI provides the capability to federate the identity with external identity providers in a Kubernetes native way. This approach overcomes several limitations of AAD Pod Identity as mentioned below.
+- Removes scale and performance issues that existed for identity assignment.
+- Supports K8s clusters hosted in any cloud or on premise.
+- Supports both Linux and Windows workloads and removes the need for CRD and pods that intercept IMDS traffic.
+
+From CAPZ perspective
+- The AAD pod identity pod has to be deployed on all the nodes where the capz pod can be potentially scheduled.
+- It has a dependency with the CNI and requires modifying iptables on the node.
+- AAD pod identity is now deprecated. 
+
+To learn more about AZWI please visit this link https://azure.github.io/azure-workload-identity/docs/introduction.html
+
+### <a name='Goals'></a>Goals
+
+- Enable CAPZ pod to use AZWI on the management cluster to authenticate to Azure to create/update/delete resources as part of workload cluster lifecycle management
+
+### <a name='NonGoals'></a>Non Goals
+
+- Use workload identity for cloud provider azure once supported.
+
+- Migrate to using workload identity in CI pipelines.
+
+- Automation for migration from AAD pod identity to workload identity.
+
+- Bootstrapping components (storage account, discovery document and JWKS) and enabling WI in the workload cluster.
+
+## <a name='Personas'></a>Personas
+
+The following personas are available when writing user stories. 
+
+- John - Cloud Admin
+	- Installs, configures and maintains, management and workload clusters on Azure using CAPZ. 
+
+## <a name='UserStories'></a>User Stories
+
+- [S1] As a cloud admin I want to use workload identity in the management cluster in order to enhance security by not using the static azure credentials. I prefer to use CAPI pivoting to create management cluster which means creating a management cluster on Kind and then create a workload cluster on Azure from that and then convert the workload cluster to a management cluster.
+
+- [S2] I am a cloud admin and I want to install CAP* on an existing Kubernetes cluster and make this a management cluster to create and manage workload clusters by using workload identity. For example, creating a management cluster on an already existing Kubernetes cluster on Azure.
+
+- [S3] As a cloud admin I want to migrate to using workload identity for my management cluster which is still using older AAD pod identity.   
+
+## <a name='Proposal'></a>Proposal
+
+In AZWI model, Kubernetes cluster itself becomes the token issuer issuing tokens to Kubernetes service accounts. These service accounts can be configured to be trusted on Azure AD application or user assigned managed identity. Workload pods can use this service account token which is projected to it's volume and can exchange the projected service account token for an Azure AD access token.
+
+The first step for creating a Kubernetes cluster using capz is creating a management cluster. On a high level the workflow looks the following to be able to use AZWI in the management cluster.
+
+**Management Cluster on Kind**
+
+Notes: 
+- Often, management cluster created on Kind is also termed as Bootstrap cluster.
+- Cloud Provider azure deployment is not required in this case.
+- This is also used in CAPI pivoting i.e creating a management cluster on Kind and then later creating a workload cluster from it and again this workload cluster is then converted to a management cluster and the Kind cluster is decommissioned. Refer to user story S1(#UserStories). 
+
+User Workflow:
+
+- The operator/admin generates signing key pair or BYO key pair. 
+
+- A Kind cluster is created with appropriate flags on kube-apiserver and kube-controller-manager and the key pairs are mounted on the container path for control plane node. See [this](#set-service-account-signing-flags) section for more details on this.
+	- kube-apiserver flags
+		- --service-account-issuer
+		- --service-account-signing-key-file
+		- --service-account-key-file
+	- kube-controller-manager
+		- --service-account-private-key-file
+
+- The operator/admin uploads the following two documents in a blob storage container. These documents are accessible publicly on a URL and this URL is commonly referred as issuer URL in this context. More about this [here](https://azure.github.io/azure-workload-identity/docs/installation/self-managed-clusters/oidc-issuer.html)  
+  - Generate and upload the Discovery Document.
+  - Generate and upload the JWKS Document.
+
+- A federated identity credential should be created between the identity and <Issuer URL, service account namespace, service account name>. More on this [here](https://azure.github.io/azure-workload-identity/docs/topics/federated-identity-credential.html)
+
+- CAPI and CAPZ are deployed on the Kind cluster that supports workload identity.
+
+- CAPZ pod uses the client ID and tenant ID of the Azure AD or user-assigned Identity by passing it in AzureClusterIdentity CR. The AzureClusterIdentity also has a capability to specify type of identity as `WorkloadIdentity` on the field `type`. 
+
+- The management cluster is now configured to use workload identity. A workload cluster can now be created by referencing the AzureClusterIdentity.
+
+
+**Management Cluster on Azure Via Pivoting**
+
+Notes:
+- Cloud provider azure will be required to run on Kubernetes cluster in this case as the management cluster is created on Azure cloud.
+
+User Workflow:
+
+- All the steps are followed as described above in `Management Cluster on Kind` with a exception that a secret is created with name `<cluster-name>-sa` encompassing the key pairs that is generated in the previous step. This is done so that the key pairs gets distributed on the control plane node. More details on it [here](https://cluster-api.sigs.k8s.io/tasks/certs/using-custom-certificates.html)
+
+- A workload cluster is created using Azure static credentials or aad pod identity. 
+
+- After the workload cluster is created, to convert it into a management cluster `clusterctl init` and `clusterctl move` commands are executed. More on this [here](https://cluster-api.sigs.k8s.io/clusterctl/commands/move.html?highlight=pivot#bootstrap--pivot)
+
+- The workload cluster is now converted into a management cluster but still using static credentials. A migration step can be followed to migrate to using workload identity.
+
+### <a name='ImplementationDetailsNotesConstraints'></a>Implementation Details/Notes/Constraints
+
+- AAD pod identity can co exist with AZWI
+- Migration plan to AZWI for existing cluster. Refer to the [MigrationPlan](#migration-plan) section at the bottom of the document.
+- For AZWI to work the following prerequisites must be met for self managed cluster. This is not required for managed cluster and follow this [link](https://azure.github.io/azure-workload-identity/docs/installation/managed-clusters.html) to know more about managed cluster setup.
+  - Key Generation
+  - OIDC URL Setup
+  - Set Service Account Signing Flags
+
+#### <a name='KeyGeneration'></a>Key Generation
+
+Admin should generate signing key pairs by using a tool such as openssl or bring their own public and private keys. 
+These keys will be mounted on a path on the containers running on the control plane node. These keys are required for signing the service account tokens that will be used by the capz pod. 
+
+#### <a name='OidcUrlSetup'></a>OIDC URL Setup
+
+Two documents i.e Discovery and JWKS json documents needs to be generated and published to a public URL. The OIDC discovery document contains the metadata of the issuer. The JSON Web Key Sets (JWKS) document contains the public signing key(s) that allows AAD to verify the authenticity of the service account token.
+
+Refer to [link](https://azure.github.io/azure-workload-identity/docs/installation/self-managed-clusters/oidc-issuer.html) for steps to setup and OIDC issuer URL.
+
+The steps on a high level to setup is the following
+- Create an azure blob storage account.
+- Create a storage container.
+- Generate the OIDC and JWKS document.
+- The document should be accessible on the public accessible URL which will be used later.
+
+#### <a name='SetServiceAccountSigningFlags'></a>Set Service Account Signing Flags
+
+Setup the flags on the kind cluster. An example is shown below
+
+```yaml
+kind: Cluster
+apiVersion: kind.x-k8s.io/v1alpha4
+nodes:
+- role: control-plane
+  extraMounts:
+      # path on node where the public key exists
+    - hostPath: ${SERVICE_ACCOUNT_KEY_FILE}
+      containerPath: /etc/kubernetes/pki/sa.pub
+      # path on node where the private key exists
+    - hostPath: ${SERVICE_ACCOUNT_SIGNING_KEY_FILE}
+      containerPath: /etc/kubernetes/pki/sa.key
+  kubeadmConfigPatches:
+  - |
+    kind: ClusterConfiguration
+    apiServer:
+      extraArgs:
+        # the oidc url after it has been set up
+        service-account-issuer: ${SERVICE_ACCOUNT_ISSUER}
+        service-account-key-file: /etc/kubernetes/pki/sa.pub
+        service-account-signing-key-file: /etc/kubernetes/pki/sa.key
+    controllerManager:
+      extraArgs:
+        service-account-private-key-file: /etc/kubernetes/pki/sa.key
+```
+
+#### <a name='FederatedCredential'></a>Federated Credential
+
+A federated identity should be created using the azure cli ( or via the azure portal).
+Please see [this](https://azure.github.io/azure-workload-identity/docs/topics/federated-identity-credential.html) for reference.
+
+```bash
+az identity federated-credential create \
+  --name "kubernetes-federated-credential" \
+  --identity-name "${USER_ASSIGNED_IDENTITY_NAME}" \
+  --resource-group "${RESOURCE_GROUP}" \
+  --issuer "${SERVICE_ACCOUNT_ISSUER}" \
+  --subject "system:serviceaccount:${SERVICE_ACCOUNT_NAMESPACE}:${SERVICE_ACCOUNT_NAME}"
+```
+
+#### <a name='DistributeKeys'></a>Distribute Keys
+
+Key pair can be distributed to a workload cluster by creating a secret with the name as `<cluster-name>-sa` encompassing the key pair. 
+Follow this [link](https://cluster-api.sigs.k8s.io/tasks/certs/using-custom-certificates.html) for more details.
+
+### <a name='CloudProviderAzureIntegration'></a>Cloud Provider Azure Integration
+
+- Cloud provider azure should be deployed with projected service account token config in the config YAML.
+
+- CAPZ should create cloud config by setting the following values if workload identity is used. 
+```go
+	AADFederatedTokenFile string `json:"aadFederatedTokenFile,omitempty" yaml:"aadFederatedTokenFile,omitempty"`
+	UseFederatedWorkloadIdentityExtension bool `json:"useFederatedWorkloadIdentityExtension,omitempty" yaml:"useFederatedWorkloadIdentityExtension,omitempty"`
+```
+
+
+### <a name='ProposedApiChanges'></a>Proposed API Changes
+
+The AzureClusterIdentity spec has a `Type` field that can be used to define what type of Azure identity should be used.
+
+```go
+// AzureClusterIdentitySpec defines the parameters that are used to create an AzureIdentity.
+type AzureClusterIdentitySpec struct {
+	// Type is the type of Azure Identity used.
+	// ServicePrincipal, ServicePrincipalCertificate, UserAssignedMSI or ManualServicePrincipal.
+	Type IdentityType `json:"type"`
+
+	// ...
+	// ...
+}
+```
+
+- Introducing one more acceptable value for `Type` in AzureClusterIdentity spec for workload identity is proposed.
+
+```go
+// IdentityType represents different types of identities.
+// +kubebuilder:validation:Enum=ServicePrincipal;UserAssignedMSI;ManualServicePrincipal;ServicePrincipalCertificate;WorkloadIdentity
+type IdentityType string
+
+const (
+	// UserAssignedMSI represents a user-assigned managed identity.
+	UserAssignedMSI IdentityType = "UserAssignedMSI"
+
+	// ServicePrincipal represents a service principal using a client password as secret.
+	ServicePrincipal IdentityType = "ServicePrincipal"
+
+	// ManualServicePrincipal represents a manual service principal.
+	ManualServicePrincipal IdentityType = "ManualServicePrincipal"
+
+	// ServicePrincipalCertificate represents a service principal using a certificate as secret.
+	ServicePrincipalCertificate IdentityType = "ServicePrincipalCertificate"
+	
+	//[Proposed Change] WorkloadIdentity represents  azure workload identity.
+	WorkloadIdentity IdentityType = "WorkloadIdentity"
+)
+
+```
+- Making the field `type` on AzureClusterIdentity immutable.
+
+### <a name='ProposedConfigurationChanges'></a>Proposed Deployment Configuration Changes
+
+- Service account token projected volume and volume mount config should be added in the CAPZ manager deployment config as described below: 
+
+```yaml
+          volumeMounts:
+            - mountPath: /var/run/secrets/azure/tokens
+              name: azure-identity-token
+              readOnly: true
+...
+      volumes:
+      - name: azure-identity-token
+        projected:
+          defaultMode: 420
+          sources:
+          - serviceAccountToken:
+              audience: api://AzureADTokenExchange
+              expirationSeconds: 3600
+              path: azure-identity-token
+```
+
+### <a name='ProposedControllerChanges'></a>Proposed Controller Changes
+
+The identity code workflow in capz should use `azidentity` module to exchange token from AAD as displayed in the next section.
+
+
+#### <a name='Identity'></a>Identity
+
+Azure client and tenant ID are injected as env variables by the azwi webhook. But for the azwi workflow, client ID and tenant ID will be fetched from AzureClusterIdentity first and will use the env variables as a fallback option. 
+
+Following is a sample code that should be made into capz identity workflow.
+
+```go
+
+			// see the next code section for details on this function
+			cred, err := newWorkloadIdentityCredential(tenantID, clientID, tokenFilePath, wiCredOptions)
+			if err != nil {
+				return nil, errors.Wrap(err, "failed to setup workload identity")
+			}
+
+			client := subscriptions.NewClient()
+
+			// setCredentialsForWorkloadIdentity just setups the 
+			// PublicCloud env URLs 
+			params.AzureClients.setCredentialsForWorkloadIdentity(ctx, params.AzureCluster.Spec.SubscriptionID, params.AzureCluster.Spec.AzureEnvironment)
+			client.Authorizer = azidext.NewTokenCredentialAdapter(cred, []string{"https://management.azure.com//.default"})
+			params.AzureClients.Authorizer = client.Authorizer
+
+```
+
+**NOTE:**
+`azidext.NewTokenCredentialAdapter` is used to get a authorizer in order to add to the existing code workflow to adapt an azcore.TokenCredential type to an autorest.Authorizer type.
+
+Also a go file e.g `workload_identity.go` in the `identity` package dealing with AZWI functionality.
+
+```go
+
+type workloadIdentityCredential struct {
+	assertion string
+	file      string
+	cred      *azidentity.ClientAssertionCredential
+	lastRead  time.Time
+}
+
+type workloadIdentityCredentialOptions struct {
+	azcore.ClientOptions
+}
+
+func newWorkloadIdentityCredential(tenantID, clientID, file string, options *workloadIdentityCredentialOptions) (*workloadIdentityCredential, error) {
+	w := &workloadIdentityCredential{file: file}
+	cred, err := azidentity.NewClientAssertionCredential(tenantID, clientID, w.getAssertion, &azidentity.ClientAssertionCredentialOptions{ClientOptions: options.ClientOptions})
+	if err != nil {
+		return nil, err
+	}
+	w.cred = cred
+	return w, nil
+}
+
+func (w *workloadIdentityCredential) GetToken(ctx context.Context, opts policy.TokenRequestOptions) (azcore.AccessToken, error) {
+	return w.cred.GetToken(ctx, opts)
+}
+
+func (w *workloadIdentityCredential) getAssertion(context.Context) (string, error) {
+	if now := time.Now(); w.lastRead.Add(5 * time.Minute).Before(now) {
+		content, err := os.ReadFile(w.file)
+		if err != nil {
+			return "", err
+		}
+		w.assertion = string(content)
+		w.lastRead = now
+	}
+	return w.assertion, nil
+}
+
+```
+
+### <a name='OpenQuestions'></a>Open Questions
+
+#### <a name='Howtomultitenancy'></a>1. How to achieve multi-tenancy?
+
+The identity is tied to the client ID which can be supplied via AzureClusterIdentity.
+
+#### <a name='Howtodistributekeys'></a>2. How to distribute key pair to management cluster?
+
+Keys can be distributed by using the CABPK feature by creating a secret with name `<cluster-name>-sa` encompassing the key pair. More details on it [here](https://cluster-api.sigs.k8s.io/tasks/certs/using-custom-certificates.html)
+
+#### <a name='UserExperience'></a>3. User Experience
+
+Though AZWI has a lot of advantages as compared to AAD pod identity, setting up AZWI involves couple of manual step for self managed clusters and it can impact the user experience. Though, for managed clusters the configurations steps are not required.
+
+### <a name='MigrationPlan'></a>Migration Plan
+
+Management clusters using AAD pod identity should have a seamless migration process which is well documented.
+
+For migrating an existing cluster to use AZWI following steps should be taken.
+
+**Migrating Greenfield Clusters**
+- The steps in this section applies to greenfield cluster creation that uses CAPI pivoting to use workload identity. As in this case the key pairs are already distributed to the control plane nodes.
+
+- Create a `UserAssignedIdentity` and create a new machine template to use user assigned identity. This step will become optional once cloud provider azure supports workload identity.
+
+- Patch the control plane to include the following flag on the api server.
+	- `service-account-issuer: <value-is-service-account-issuer-url>
+
+- Perform node rollout to propagate the changes by patching KCP and MachineDeployment objects to include the reference on the new AzureMachineTemplate created in the previous step.
+
+- Create a new `AzureClusterIdentity` and specify the client ID, tenant ID and `type` to `WorkloadIdentity`.
+
+- Patch AzureCluster to use the new `AzureClusterIdentity`.
+
+**Migrating Brownfield Clusters**
+
+- The steps in this section applies to existing clusters. 
+
+- Generate a key pair or use the same key pairs as present in `/etc/kubernetes/pki`
+
+- Perform the following pre-requistes as discussed earlier in the **Bootstrap Cluster** section of the [document](#proposal):
+	- Generate and upload the JWKS and Discovery Document.
+	- Install the AZWI mutating admission webhook controller.
+	- Establish federated credential for the identity and `<Issuer, service account namespace, service account name>`.
+
+- If you are not using the existing key pairs then perform a service account rotation using this guide. 
+https://azure.github.io/azure-workload-identity/docs/topics/self-managed-clusters/service-account-key-rotation.html
+
+- If you are using the existing key pairs or if you just configure workload identity using the keys present in /etc/kubernetes/pki then rotation is not required.
+ 	 
+- Create a `UserAssignedIdentity` and create a new machine template to use user assigned identity. This step will become optional once cloud provider azure supports workload identity.
+
+- This step is not required if you are using the existing key pairs. Patch the control plane to include the following flags on the api server.
+	- `service-account-key-file:<public-key-path-on-controlplane-node>`
+	- `service-account-signing-key-file:<private-key-path-on-controlplane-node>`
+	- **Note:** The key pairs are present in `/etc/kubernetes/pki` directory if using the `<cluster-name>-sa` key pair.
+
+- This step is not required if you are using the existing key pairs. Patch the control plane to include the following flags on the kube controller manager.
+	- `service-account-signing-key-file:<private-key-path-on-controlplane-node>`
+
+- Patch the control plane to include the following flag on the api server.
+	- `service-account-issuer: <value-is-service-account-issuer-url>`
+
+- Perform node rollout to propagate the changes by patching KCP and MachineDeployment objects to include the reference on the new AzureMachineTemplate created in the previous step.
+
+- Upgrade to the CAPZ version that supports AZWI. 
+
+- Create a new `AzureClusterIdentity` and specify the client ID, tenant ID and `type` to `WorkloadIdentity`.
+
+- Patch AzureCluster to use the new `AzureClusterIdentity`.
+
+**Note:** Migration to AZWI is a tedious manual process and poses risk of human error. It should be better done via an automation.
+
+### <a name='TestPlan'></a>Test Plan
+
+* Unit tests to validate newer workload identity functions and helper functions.  
+* Using AZWI for existing e2e tests for create, upgrade, scale down/up, and delete.
+
+## <a name='ImplementationHistory'></a>Implementation History
+
+- Enabling workload identity feature in CAPZ
+Ref Issue: https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/3588
+Ref PR: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3583
+
+- [ToDo]Integrating with cloud provider azure to use workload identity. 
+Ref Issue: https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/3589
+
+- [ToDo] Moving all the CI jobs to be using workload identity.
+Ref Issue: https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/3590
--- a/controllers/azureidentity_controller.go
+++ b/controllers/azureidentity_controller.go
@@ -185,7 +185,7 @@ func (r *AzureIdentityReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		}
 	}
 
-	// delete bindings and identites no longer used by a cluster
+	// delete bindings and identities no longer used by a cluster
 	for _, bindingToDelete := range bindingsToDelete {
 		binding := bindingToDelete
 		identityName := binding.Spec.AzureIdentity


docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,351 @@
...
+
+The first step for creating K8s cluster using capz is creating a bootstrap cluster which is a Kind cluster and then deploying capi, capz and other helper components. On a high level the workflow looks the following to be able to use AWI and create a management cluster.
+
+**Bootstrap Cluster**
+
+- The operator/admin should be able generate signing key pair or BYO key pair. 
+
+- The bootstrap cluster i.e the Kind K8s cluster should be configured with appropriate flags on kube-apiserver and kube-controller-manager and the key pairs should be mounted on the container path for control plane node. 
+
+- The operator/admin should set up OIDC issuer URL which means having access to azure blob storage where the following two documents will be stored in the blob storage container.
```
Ishani-Gupta, line 94: This assumes that storage account is already created, right? Also, along with container.

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,351 @@
...
+  - Generate and upload the Discovery Document.
```
Ishani-Gupta, line 95: fyi...Let me know if you need help in this part. I wrote go lang functions already for this for our internal need.

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,351 @@
...
+  - Generate and upload the JWKS Document.
+ 
+- The AWI mutating admission webhook should be installed on the bootstrap cluster.
+
+- A service account should be created in the same namespace as that of capz pod and tagged with the client ID of the Azure AD application or user addigned identity on the bootstrap cluster. The service account should also be associated as a Subject with the appropriate ClusterRoleBindings and RoleBindings for the usual RBAC. This is the service account that the capz pod should use.
+
+- The capz manager service account should be annotated with the Azure AD or User Assigned Client ID.
```
Ishani-Gupta, line 102: Is this service account different from the one mentioned above?

docs/proposals/20221611-workload-identity-integration.md
sonasingh46: Yes.

docs/proposals/20221611-workload-identity-integration.md
sonasingh46: That is the same. It's duplicate. I will remove

docs/proposals/20221611-workload-identity-integration.md
sonasingh46: Sure. But I think this should get included in the capz code.

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,351 @@
...
+
+
+#### <a name='Howtodistributekeys'></a>2. How to distribute key pair to management cluster?
+
+This an open question to discuss to find a better way to distribute keys to the control plane node for the management cluster.
+
+
+#### <a name='UserExperience'></a>3. User Experience
+
+Though AWI has a lot of advantages as compared to AWI, setting up AWI involves couple of manual step and it can impact the user experience.
```
jackfrancis, line 336: should it be

> Though AWI has a lot of advantages as compared to aad-pod-identity

?

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,351 @@
...
+
+- Integrate Azure Workload Identity into capz.
+
+### <a name='Non-GoalsFutureWork'></a>Non-Goals / Future Work
+
+- Migration to `azidentity` from `adal` module when using `go-autorest` library in capz as `adal` will be deprecated. 
+
+## <a name='Proposal'></a>Proposal
+
+In this model, Kubernetes cluster itself becomes the token issuer issuing tokens to Kubernetes service accounts. These service accounts can be configured to be trusted on Azure AD application or user assigned identity. Workloads/pods can use this service account token which is projected to it's volume and can exchange for an Azure AD token.
```
aramase, line 84:
```suggestion
In this model, Kubernetes cluster itself becomes the token issuer issuing tokens to Kubernetes service accounts. These service accounts can be configured to be trusted on Azure AD application or user-assigned managed identity. Workloads/pods can exchange the projected service account token for an Azure AD token.
```

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,351 @@
...
+
+- The bootstrap cluster i.e the Kind K8s cluster should be configured with appropriate flags on kube-apiserver and kube-controller-manager and the key pairs should be mounted on the container path for control plane node. 
+
+- The operator/admin should set up OIDC issuer URL which means having access to azure blob storage where the following two documents will be stored in the blob storage container.
+  - Generate and upload the Discovery Document.
+  - Generate and upload the JWKS Document.
+ 
+- The AWI mutating admission webhook should be installed on the bootstrap cluster.
+
+- A service account should be created in the same namespace as that of capz pod and tagged with the client ID of the Azure AD application or user addigned identity on the bootstrap cluster. The service account should also be associated as a Subject with the appropriate ClusterRoleBindings and RoleBindings for the usual RBAC. This is the service account that the capz pod should use.
```
aramase, line 100:
```suggestion
- A service account should be created in the same namespace as that of capz pod and tagged with the client ID of the Azure AD application or user-assigned managed identity on the bootstrap cluster. The service account should also be associated as a Subject with the appropriate ClusterRoleBindings and RoleBindings for the usual RBAC. This is the service account that the capz pod should use.
```

aramase: I assume CAPZ controller already uses a service account today that'll just be reused with workload identity?

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,351 @@
...
+
+- The capz manager service account should be annotated with the Azure AD or User Assigned Client ID. 
+
+- A fedreated credential should be created between the identity and the OIDC issuer URL and the service account(subject).
```
aramase, line 104:
```suggestion
- A federated credential should be created between the identity and the OIDC issuer URL and the service account(subject). 
```

Might be good to add a reference to https://azure.github.io/azure-workload-identity/docs/topics/federated-identity-credential.html

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,351 @@
...
+- The capz pod service account should be annotated with the client ID of the User Assigned Identity. The client ID can be passed as an argument to the `clusterctl init` command. 
+
+### <a name='ImplementationDetailsNotesConstraints'></a>Implementation Details/Notes/Constraints
+
+- AAD pod identity can co exist with AWI
+- Migration plan from AAD pod identity to AWI for existing cluster. Refer to the `Migration Plan` section at the bottom of the document.
+- For AWI to work the following prerequisites must be met.
+  - Key Generation
+  - OIDC URL Setup
+  - Set Service Account Signing Flags
```
aramase, line 123: These are only required if the management cluster is a kind cluster. For managed clusters, all these are set up as part of cluster bootstrap: https://azure.github.io/azure-workload-identity/docs/installation/managed-clusters.html

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,351 @@
...
+#### <a name='KeyGeneration'></a>Key Generation
+
+Admin should generate signing key pairs by using a tool such as openssl or bring their own public and private keys. 
+These keys will be mounted on a path on the containers running on the control plane node. These keys are required for signing the service account tokens that will be used by the capz pod. 
+
+#### <a name='OidcUrlSetup'></a>OIDC URL Setup
+
+Two documents i.e OIDC and JWKS json documents needs to be generated and published to a public URL. The OIDC document contains the metadata of the issuer. The JSON Web Key Sets (JWKS) document contains the public signing key(s) that allows AAD to verify the authenticity of the service account token.
+
+For more details on how to set up OIDC URL and more details on it, please g through this this [link](#https://azure.github.io/azure-workload-identity/docs/installation/self-managed-clusters/oidc-issuer.html)
```
aramase, line 134:
```suggestion
For more details on how to set up OIDC URL refer to [link](https://azure.github.io/azure-workload-identity/docs/installation/self-managed-clusters/oidc-issuer.html)
```

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,351 @@
...
+
+The steps on a high level to setup is the following
+- Create an azure blob storage account.
+- Create a storage container.
+- Generate the OIDC and JWKS document.
+- The doument should be accessible on the public accessible URL which will be used later.
```
aramase, line 140:
```suggestion
- The document should be accessible on the public accessible URL which will be used later.
```

aramase, line 140: Clarify these steps are only required when using a self-managed cluster as management cluster

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,351 @@
...
+    apiServer:
+      extraArgs:
+	    # the oidc url after it has been set up
+        service-account-issuer: ${SERVICE_ACCOUNT_ISSUER}
+        service-account-key-file: /etc/kubernetes/pki/sa.pub
+        service-account-signing-key-file: /etc/kubernetes/pki/sa.key
+    controllerManager:
+      extraArgs:
+        service-account-private-key-file: /etc/kubernetes/pki/sa.key
+```
```
aramase, line 170: Sample config we use for testing: https://github.com/Azure/azure-workload-identity/blob/main/scripts/create-kind-cluster.sh#L60-L91

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,351 @@
...
+
+#### <a name='FederatedCredential'></a>Federated Credential
+
+The service account that the capz pod will use should be annotated with the client ID of the AAD application or User Assigned Identity and a fedrated identity should be created by associating the ServiceAccont as a subject.
```
aramase, line 174:
```suggestion
The service account that the capz pod will use should be annotated with the client ID of the AAD application or user-sssigned Identity and a federated identity should be created by associating the Service Account as a subject.
```

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,351 @@
...
+
+### <a name='OpenQuestions'></a>Open Questions
+
+#### <a name='Howtomultitenancy'></a>1. How to achieve multi-tenancy?
+
+The identity is tied to service account via the client ID annotation present in the service account. Service account has one to one mapping with the pod. To have a many-to-one relationship between multiple identities and a service account, multiple federated identity credentials can be created that references the same service account.
+
+`azure.workload.identity/client-id` annotations contains the client ID of the identity that will be used.
+
+To use a different client ID, the client ID should be updated but the limitation is that at a time only one identity can be used.
```
aramase, line 326: This is not really a limitation. The annotated client-id is injected as an env var `AZURE_CLIENT_ID` that'll be automagically used by the SDKs when no client id is provided as part of the managed identity client initialization. If you wish to use a different identity, then you can explicitly pass the client id of the identity to the client.

xref: https://azure.github.io/azure-workload-identity/docs/faq.html#how-to-federate-multiple-identities-with-a-kubernetes-service-account

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,351 @@
...
+#### <a name='UserExperience'></a>3. User Experience
+
+Though AWI has a lot of advantages as compared to AWI, setting up AWI involves couple of manual step and it can impact the user experience. 
+
+### <a name='MigrationPlan'></a>Migration Plan
+
+Management clusters using AAD pod identity should have a seamless migration process which is well documented.
+
+As part of upgrade that supports AWI, the env variable can be set to true to use AWI. 
+But if capz uses AWI, it can use only one identity at a time and it can pose challenges to migration if the management cluster is using different identity simultaneously from the AAD pod identity.
```
aramase, line 343: You can use multiple identities with workload identity. The workload that's using the identity just needs to know which client id to use which can be obtained by some of kind of configuration provided by the users. The annotation on the service account is a value add which allows setting a fallback client id (`AZURE_CLIENT_ID`) when no client id is provided.

docs/proposals/20221611-workload-identity-integration.md
sonasingh46: For managed cluster -- you mean AKS, right? 

But I will add a note about this.

docs/proposals/20221611-workload-identity-integration.md
sonasingh46: Awesome. I will add that and will also test in the poc code that I have.

docs/proposals/20221611-workload-identity-integration.md
CecileRobertMichon: it's unfortunate that there isn't a way to do this without having to pass in some inline yaml, this degrades the simple UX of just doing ` kind create cluster`

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,351 @@
...
+
+Also, a federated identity should be created using the azure cli ( or via the azure portal).
+
+```bash
+az identity federated-credential create \
+  --name "kubernetes-federated-credential" \
+  --identity-name "${USER_ASSIGNED_IDENTITY_NAME}" \
+  --resource-group "${RESOURCE_GROUP}" \
+  --issuer "${SERVICE_ACCOUNT_ISSUER}" \
+  --subject "system:serviceaccount:${SERVICE_ACCOUNT_NAMESPACE}:${SERVICE_ACCOUNT_NAME}"
```
CecileRobertMichon, line 196: dumb question: how do we differentiate two service accounts in two different clusters that have the same name and namespace?

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,351 @@
...
+```
+
+This means there are two approaches to associate the correct service account to the capz deployment.
+
+- Deploy the capz with the usual configuration YAML and then patch the service account with the client ID.
+
+- Or create the service account prior to capz deployment which uses the same service account.
```
CecileRobertMichon, line 203: not sure I understand what this means, do you mean just add the annotation to the infrastructure-component.yaml with an env variable by changing https://github.com/kubernetes-sigs/cluster-api-provider-azure/tree/main/config?

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,351 @@
...
+
+
+#### <a name='DistributeKeys'></a>Distribute Keys To Management Cluster
+
+Setting up credentials for management cluster which is created from the bootstrap cluster is key activity. This requires storing the keys on control plane node. One approach to do so can be the following:
+
+- A reular `clusterctl init` is done.
```
CecileRobertMichon, line 210:
```suggestion
- A regular `clusterctl init` is done.
```

what do we mean by "regular" ?

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,351 @@
...
+- Once the cluster is up and running, the keys can be distributed on the control plane node via scp.
+- After distributing the key, the appropriate flags as it was set up for the bootstrap cluster can be setup.
+- The service account should be annotated with the client ID.
+- At this stage, `clusterctl move` can be executed.
+
+Above activities can be automated via clusterctl subcommands that is specific to azure infrastructre.
```
CecileRobertMichon, line 216: not sure I understand this part

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,351 @@
...
+
+### <a name='ProposedConfigurationChanges'></a>Proposed Deployment Configuration Changes
+
+- Azure Workload Identity mutating webhook should be deployed as part of capz deployment.
+
+- The capz deployment configuration should have a environment variable configuration that is used to enable/disable the use of AWI. By default it will be disabled.
```
CecileRobertMichon, line 222: this might be tricky as currently we have no way to do templating in provider templates with clusterctl init. The one thing we could do though is something like:

`azure.workload.identity/use: "${USE_AZURE_WORKLOAD_ID:=false}"` 

for the ServiceAccount labels. But I don't know if there is a way to conditionally install the mutating webhooks.

CecileRobertMichon: Also why do we want to disable it by default? Since AAD pod ID is deprecated I think we should aim to make AZWI the default

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,351 @@
...
+
+
+### <a name='ProposedControllerChanges'></a>Proposed Controller Changes
```
CecileRobertMichon, line 225: I don't see a section about proposed API changes. This is important since currently identity is defined in the API as `identityRef`  and ` AzureClusterIdentity`

docs/proposals/20221611-workload-identity-integration.md
CecileRobertMichon: Could we reuse the `AzureClusterIdentity`  CRD to specify which client ID to use per workload cluster?

docs/proposals/20221611-workload-identity-integration.md
CecileRobertMichon: @aramase is the annotation to set a fallback client id required? Or could we potentially omit that altogether and only provide pre-cluster client IDs?

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,351 @@
...
+Though AWI has a lot of advantages as compared to AWI, setting up AWI involves couple of manual step and it can impact the user experience. 
+
+### <a name='MigrationPlan'></a>Migration Plan
+
+Management clusters using AAD pod identity should have a seamless migration process which is well documented.
+
+As part of upgrade that supports AWI, the env variable can be set to true to use AWI. 
+But if capz uses AWI, it can use only one identity at a time and it can pose challenges to migration if the management cluster is using different identity simultaneously from the AAD pod identity. 
+
+### <a name='TestPlan'></a>Test Plan
```
CecileRobertMichon, line 345: I'm very interested to learn more about how we can set this up for our CI sub. @aramase do you have examples of CI using kind clusters to test AZWI?

CecileRobertMichon (commented):
Great research work @sonasingh46 and thank you @aramase for all the support. I'm excited to get this in!

docs/proposals/20221611-workload-identity-integration.md
aramase: > For managed cluster -- you mean AKS, right?

that's right! This [doc](https://azure.github.io/azure-workload-identity/docs/installation/managed-clusters.html) has the list of managed clusters and how to get the oidc URL for setting up federated identity credential.

aramase: > How does this work for CI when using kind as the management cluster? Can this setup be done once in the CI sub or will we need to do the steps every time? If every time, that requires CI to have creds to run these steps so we will need to authenticate via SP which defeats the purpose of no longer needing SPs to test CAPZ with kind

The discovery document and jwks url can be setup one time using the storage account. This is the one I setup for our `azwi` CI:
- https://azwi.blob.core.windows.net/oidc-test/.well-known/openid-configuration
- https://azwi.blob.core.windows.net/oidc-test/openid/v1/jwks

The service account keys are in keyvault and retrieved when bootstrapping the kind cluster for the CI. xref: https://github.com/Azure/azure-workload-identity/blob/main/scripts/create-kind-cluster.sh

The federated identity credential can be setup single time if the namespace, service account name, namespace and OIDC issuer url is fixed.

docs/proposals/20221611-workload-identity-integration.md
aramase: > @aramase is the annotation to set a fallback client id required? Or could we potentially omit that altogether and only provide pre-cluster client IDs?

The fallback client id is not required, i.e the client id annotation is optional. As long as the CAPZ controller knows which client id to use with the azure identity sdk clients we are good!

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,351 @@
...
+
+- CAPZ pod should use the `azidentity` module from the azure-sdk-for-go to exchange AD token in lieu of the projected service account token.
+
+**Management Cluster**
+
+- Once a K8s cluster is created from the bootstrap cluster, to convert it into a management cluster `clusterctl init` and `clusterctl move` commands are executed.
+
+- After `clusterctl init` command is executed, the control plane node should have the key pairs and then patched to include them in the container path and set the kube-apiserver and kube-controller-manager flags.
+
+- The capz pod service account should be annotated with the client ID of the User Assigned Identity. The client ID can be passed as an argument to the `clusterctl init` command.
```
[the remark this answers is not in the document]
aramase: The client ID annotation is **not required**. It's optional and can be omitted. In this scenario, CAPZ controller can use the clientID information from `AzureClusterIdentity`.

docs/proposals/20221611-workload-identity-integration.md
aramase: The issuer will be used to distinguish the federated identity credential. The uniqueness is based on the following tuple: `<issuer, service account name, service account namespace>`.

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,351 @@
...
+
+To learn more about AWI please visit this link https://azure.github.io/azure-workload-identity/docs/introduction.html
+
+### <a name='Goals'></a>Goals
+
+- Integrate Azure Workload Identity into capz.
+
+### <a name='Non-GoalsFutureWork'></a>Non-Goals / Future Work
+
+- Migration to `azidentity` from `adal` module when using `go-autorest` library in capz as `adal` will be deprecated.
```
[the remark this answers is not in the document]
sonasingh46: Yes. I will remove this section.

docs/proposals/20221611-workload-identity-integration.md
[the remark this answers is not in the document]
sonasingh46: Ok. So When we set up an azure blob storage to store files in a container we have a url like the following which can be publicly accessible. 
"https://${AZURE_STORAGE_ACCOUNT}.blob.core.windows.net/${AZURE_STORAGE_CONTAINER}"
https://azure.github.io/azure-workload-identity/docs/installation/self-managed-clusters/oidc-issuer.html

docs/proposals/20221611-workload-identity-integration.md
sonasingh46: Yes. It is correct. I need to remove this line. This was some previous thought which is not applicable.

docs/proposals/20221611-workload-identity-integration.md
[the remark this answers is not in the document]
sonasingh46: So what I meant here is the following
- Most oftenly, a Kind cluster is used for bootstrap cluster for simplicity ( though not a requirement ). Or at least I have seen this workflow to users/customers whom I have interacted with. 
- Now, when a bootstrap cluster is created cluster and all the pre-requisites are met including annotating the service account. If not annotating then in some way client ID should be provided.
- After this, a workload cluster is created on azure. This workload cluster now needs to be converted to a management cluster. This means it should also meet all the requirements including a way to provide client-id that should be used. 
- Now, passing client-id via clusterctl was just a naive thought from me to somehow pass the client ID to the management cluster. 
- But as @aramase said client ID is optional ( I misunderstood it as required ) 

So I think specifying which client ID to use should not be complex and I will remove this passing client ID via clusterctl part. 
Also on the other had, I think we should now refrain from using AzureClusterIdentity and may be introducing a client ID field itself in the AzureCluster.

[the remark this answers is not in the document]
sonasingh46: Though now I am thinking if we can use AzureClusterIdentity if it provides more value then just helping pass a information which is client ID. 
Let us discuss on this.

docs/proposals/20221611-workload-identity-integration.md
sonasingh46: Again in this case, the annotated part is optional. So I will remove this.
Infact, in the entire document, I am planning to remove this ''annotated part'' and mention it as a way to fallback to use client ID if it is no where supplied. 
Does this make sense?

abutcher: Where else could the client ID be specified? Would it be possible to provide the `client-id` and `federated-token-file` path as part of the existing credentials secret (while omitting the `client-secret` for example with AZWI) as an alternative to injecting environment variables via the mutating webhook? https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/config/default/credentials.yaml

docs/proposals/20221611-workload-identity-integration.md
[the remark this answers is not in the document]
CecileRobertMichon: I think for back compat reasons it might be easier to keep AzureClusterIdentity for now so existing clusters can migrate. Later we can think about deprecating AzureClusterIdentity and just having a clientID field on AzureCluster

docs/proposals/20221611-workload-identity-integration.md
CecileRobertMichon: Aah perfect

docs/proposals/20221611-workload-identity-integration.md
CecileRobertMichon: This is awesome, thank you

dtzar:
@sonasingh46 - thank you for this proposal and agreeing to help here.  This is a very important enhancement needed by many for CAPZ.  When do you think you'll have time to address feedback mentioned here and/or start on the coding work?

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,351 @@
...
+
+For more details about Azure AD Pod Identity please visit this [link](#https://github.com/Azure/aad-pod-identity)  
+
+## <a name='Motivation'></a>Motivation
+
+AWI gives the capability to federate the credentials with external identity providers in a Kubernetes native way. This approach overcomes several limittations of Azure AD Identity as mentioned below.
+- Removes scale and performance issues that existed for identity assignment.
+- Supports K8s clusters hosted in any cloud or on premise.
+- Supports both Linux and Windows workloads and removes the need for CRD and pods that intercept IMDS traffic.
```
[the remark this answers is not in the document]
sonasingh46: Done

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,351 @@
...
+
+**Bootstrap Cluster**
+
+- The operator/admin should be able generate signing key pair or BYO key pair. 
+
+- The bootstrap cluster i.e the Kind K8s cluster should be configured with appropriate flags on kube-apiserver and kube-controller-manager and the key pairs should be mounted on the container path for control plane node. 
+
+- The operator/admin should set up OIDC issuer URL which means having access to azure blob storage where the following two documents will be stored in the blob storage container.
+  - Generate and upload the Discovery Document.
+  - Generate and upload the JWKS Document.
```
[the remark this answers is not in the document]
sonasingh46: Done.

docs/proposals/20221611-workload-identity-integration.md
sonasingh46: Not applicable now. Ignore this. Sorry for the confusion.

docs/proposals/20221611-workload-identity-integration.md
sonasingh46: One of the usual workflows that I have seen from users is to use Bootstrap & Pivot : 
- Creating a kind cluster and using it as a bootstrap cluster to deploy a workload cluster on Azure. 
- Now that workload cluster is converted to a management cluster using the pivot process. 

More on this link
https://cluster-api.sigs.k8s.io/clusterctl/commands/move.html?highlight=pivot#bootstrap--pivot

Now with azwi, this means making the key pairs available non the control plane nodes and doing an extraMount as we are doing it for the bootstrap cluster.  

With that I was thinking on the lines for making a subcommand that can deal with such thing e.g. transfer the keys to all the control plane nodes via scp( as that looks a manual process that is now introduces to be able to use azwi.)

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,346 @@
...
+### <a name='OpenQuestions'></a>Open Questions
+
+#### <a name='Howtomultitenancy'></a>1. How to achieve multi-tenancy?
+
+The identity is tied to the client ID which can be supplied via AzureClusterIdentity. In case identity is not found in AzureClusterIdentity client ID from the service account annotation
+`azure.workload.identity/client-id` will be used.
+
+#### <a name='Howtodistributekeys'></a>2. How to distribute key pair to management cluster?
+
+This an open question to discuss to find a better way to distribute keys to the control plane node for the management cluster.
```
sonasingh46, line 326: cc @dtzar

docs/proposals/20221611-workload-identity-integration.md
sonasingh46: Though I do not have any strong opinions on keeping the support for both simultaneously. Just was thinking about it and mentioned.

docs/proposals/20221611-workload-identity-integration.md
sonasingh46: Though Cecile is already aware of this. Pasting this issue link that outlines this for more curious readers 
https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/2936

docs/proposals/20221611-workload-identity-integration.md
sonasingh46: cc @CecileRobertMichon

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,374 @@
...
+  --identity-name "${USER_ASSIGNED_IDENTITY_NAME}" \
+  --resource-group "${RESOURCE_GROUP}" \
+  --issuer "${SERVICE_ACCOUNT_ISSUER}" \
+  --subject "system:serviceaccount:${SERVICE_ACCOUNT_NAMESPACE}:${SERVICE_ACCOUNT_NAME}"
+```
+
+#### <a name='DistributeKeys'></a>Distribute Keys To Management Cluster
+
+Setting up credentials for management cluster which is created from the bootstrap cluster is a key activity. This requires storing the keys on control plane node and patching kube-apiserver and kube-controller-manager flags to include the service account signing flags similar to what is discussed in the above `Set Service Account Signing Flags` section.
+
```
sonasingh46, line 209: cc @CecileRobertMichon

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,374 @@
...
+| AD           | Active Directory        |
+| AAD          | Azure Active Directory  |
+| AZWI         | Azure Workload Identity |
+| OIDC         | OpenID Connect          |
+| JWKS         | JSON Web Key Sets       |
+
+## <a name='Summary'></a>Summary
+
+Workloads deployed in Kubernetes cluster may require Azure Active Directory application credential or managed identities to access azure protected resource e.g. Azure Key Vault, Virtual Machines etc. Azure AD Pod Identity helps access azure resources without the need of a secret management via Azure Managed Identities. Azure AD Pod Identity is now deprecated and Azure AD Workload Identity is the next iteration of the former. This design proposal aims to define the way for AZWI integration into capz for self managed clusters with keeping in mind other factor e.g. User Experience, Multi Tenancy, Backwards Compatibility etc.
+
```
sonasingh46, line 64: @CecileRobertMichon @aramase -- FWIW, 
I have intentionally written `AZWI integration into capz for self managed clusters` as if I understood correctly for a mangement/bootstrap cluster which is a managed cluster, there is no such special code handling that is required for azwi apart from  meeting the following pre-requisites: 
- Get the OIDC url from managed cluster and steps on how to do that is here https://azure.github.io/azure-workload-identity/docs/installation/managed-clusters.html
- Establish the federated credential using the OIDC url fetched from the above step.  

Yes, it is worth to get this info into the doc and I will put that after y'all see this comment and if it looks ok to you. Missed this part and it came to my mind when I was coding for azwi.

jackfrancis:
Let's prioritize starting a lazy consensus timer roughly a week before the planned 1.7.0 release date (10 Jan 2023)

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,374 @@
...
+- The operator/admin generates signing key pair or BYO key pair. 
+
+- The bootstrap cluster is configured with appropriate flags on kube-apiserver and kube-controller-manager and the key pairs are mounted on the container path for control plane node. See [this](#SetServiceAccountSigningFlags) section for more details on this.
+	- kube-apiserver flags
+		- --service-account-issuer
+		- --service-account-signing-key-file
+		- --service-account-key-file
+	- kube-controller-manager
+		- --service-account-private-key-file
```
fabriziopandini, line 101: Have you considered to provide a secret with clustername-sa before creating the cluster, so then KCP will automatically use it (see https://cluster-api.sigs.k8s.io/tasks/certs/using-custom-certificates.html)

This will have also the nice side effect that also move should take care of this secret without any additional operation

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,374 @@
...
+	ServicePrincipal IdentityType = "ServicePrincipal"
+
+	// ManualServicePrincipal represents a manual service principal.
+	ManualServicePrincipal IdentityType = "ManualServicePrincipal"
+
+	// ServicePrincipalCertificate represents a service principal using a certificate as secret.
+	ServicePrincipalCertificate IdentityType = "ServicePrincipalCertificate"
+	
+	//[Proposed Change] WorkloadIdentity represents  azure workload identity.
+	WorkloadIdentity IdentityType = "WorkloadIdentity"
```
fabriziopandini, line 247: Q: Are those fields immutable?
Just wondering what happens if the user changes settings in this struct on an existing cluster

slightly related

Q: What is the migration path for users currently using the deprecate azure pod identity?

docs/proposals/20221611-workload-identity-integration.md
yastij: +1 Ideally we should let KCP pick up pre-created service account key-pair

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,374 @@
...
+The AzureClusterIdentity spec has a `Type` field that can be used to define what type of Azure identity should be used.
+
+```go
+// AzureClusterIdentitySpec defines the parameters that are used to create an AzureIdentity.
+type AzureClusterIdentitySpec struct {
+	// Type is the type of Azure Identity used.
+	// ServicePrincipal, ServicePrincipalCertificate, UserAssignedMSI or ManualServicePrincipal.
+	Type IdentityType `json:"type"`
+
+	// ...
```
yastij, line 221: I feel like the API is still missing some semantics: If I have 100 clusters that will use workload identity, I still don't want to go ahead and change KCP to add the issuer URL and flags for key pair

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,374 @@
...
+### <a name='ProposedApiChanges'></a>Proposed API Changes
+
+The AzureClusterIdentity spec has a `Type` field that can be used to define what type of Azure identity should be used.
+
+```go
+// AzureClusterIdentitySpec defines the parameters that are used to create an AzureIdentity.
+type AzureClusterIdentitySpec struct {
+	// Type is the type of Azure Identity used.
+	// ServicePrincipal, ServicePrincipalCertificate, UserAssignedMSI or ManualServicePrincipal.
+	Type IdentityType `json:"type"`
```
yastij, line 219: 2 things here:

- this field seem to be required, If so what happens when upgrading clusters? what has been our default behaviour before?
- this field seems mutable for now, unless we figure out the upgrade that, I'd make this immutable for now. we can always open it up once we have a clear plan of transition

yastij (commented):
a few comments

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,418 @@
+---
+title: Workload Identity Integration
+authors:
+    - @sonasingh46
+reviewers:
+    - @aramase
+    - @CecileRobertMichon
```
CecileRobertMichon, line 7: let's add others who reviewed as well

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,418 @@
...
+To learn more about AZWI please visit this link https://azure.github.io/azure-workload-identity/docs/introduction.html
+
+### <a name='Goals'></a>Goals
+
+- Enable CAPZ pod to use AZWI on the management cluster to authenticate to Azure to create/update/delete resources as part of workload cluster lifecycle management
+
+### <a name='NonGoals'></a>Non Goals
+
+- Automation for migration from AAD pod identity to workload identity.
+- More flexible way so key distribution so that key pair other than the one present for the exiting cluster via `<cluster-name>-sa` secret can be used. To understand more on this please refer [MigrationPlan](#migration-plan)
```
CecileRobertMichon, line 89: can we also add a non-goal/future work to install workload identity on workload clusters created by CAPZ

CecileRobertMichon: and also one for using workload identity for cloud-provider auth once that's supported

docs/proposals/20221611-workload-identity-integration.md
CecileRobertMichon: @sonasingh46 thoughts on this?

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,418 @@
...
+
+For migrating an existing cluster to use AZWI following steps should be taken.
+
+- Generate a key pair or use the same key pairs as created via `<cluster-name>-sa` secret. 
+
+- Perform the following pre-requistes as discussed earlier in the **Bootstrap Cluster** section of the [document](#proposal):
+	- Generate and upload the JWKS and Discovery Document.
+	- Install the AZWI mutating admission webhook controller.
+	- Establish federated credential for the key pair. 	 
+- Create a `UserAssignedIdentity` and update the machine template to use user assigned identity.
```
CecileRobertMichon, line 390: just making sure: this is mutable right?

docs/proposals/20221611-workload-identity-integration.md
sonasingh46: done

docs/proposals/20221611-workload-identity-integration.md
sonasingh46: done

docs/proposals/20221611-workload-identity-integration.md
sonasingh46: Oh, right! I think AzureMahcineTemplate is immutable. If it is, then we need to create a a new AzureMahcineTemplate and perform node roll outs. 
Let me validate this and update. Thanks

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,422 @@
+---
+title: Workload Identity Integration
+authors:
+    - @sonasingh46
+reviewers:
+    - @aramase
+	- @CecileRobertMichon
```
CecileRobertMichon, line 7: indent is wrong here

CecileRobertMichon (commented):
@sonasingh46 what's the status on this? Is it ready for final review/merge? Are there folks you're waiting to hear back from?

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,422 @@
...
+	ServicePrincipal IdentityType = "ServicePrincipal"
+
+	// ManualServicePrincipal represents a manual service principal.
+	ManualServicePrincipal IdentityType = "ManualServicePrincipal"
+
+	// ServicePrincipalCertificate represents a service principal using a certificate as secret.
+	ServicePrincipalCertificate IdentityType = "ServicePrincipalCertificate"
+	
+	//[Proposed Change] WorkloadIdentity represents  azure workload identity.
+	WorkloadIdentity IdentityType = "WorkloadIdentity"
```
CecileRobertMichon, line 260: Added a comment on the naming of the identity type [here](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2924#discussion_r1102172216)

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,422 @@
...
+  - Set Service Account Signing Flags
+
+#### <a name='KeyGeneration'></a>Key Generation
+
+Admin should generate signing key pairs by using a tool such as openssl or bring their own public and private keys. 
+These keys will be mounted on a path on the containers running on the control plane node. These keys are required for signing the service account tokens that will be used by the capz pod. 
+
+#### <a name='OidcUrlSetup'></a>OIDC URL Setup
+
+Two documents i.e OIDC and JWKS json documents needs to be generated and published to a public URL. The OIDC document contains the metadata of the issuer. The JSON Web Key Sets (JWKS) document contains the public signing key(s) that allows AAD to verify the authenticity of the service account token.
```
aramase, line 149:
```suggestion
Two documents i.e OIDC and JWKS json documents needs to be generated and published to a public URL. The OIDC discovery document contains the metadata of the issuer. The JSON Web Key Sets (JWKS) document contains the public signing key(s) that allows AAD to verify the authenticity of the service account token.
```

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,422 @@
...
+
+For more details on how to set up OIDC URL and more details on it, please g through this this [link](https://azure.github.io/azure-workload-identity/docs/installation/self-managed-clusters/oidc-issuer.html)
```
aramase, line 151:
```suggestion
Refer to [link](https://azure.github.io/azure-workload-identity/docs/installation/self-managed-clusters/oidc-issuer.html) for steps to setup an OIDC issuer URL. 
```

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,422 @@
...
+The service account that the capz pod will use should be annotated with the client ID of the AAD application or User Assigned Identity and a federated identity should be created by associating the Service Account as a subject.
+
+```yaml
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  annotations:
+    azure.workload.identity/client-id: ${APPLICATION_CLIENT_ID:-$USER_ASSIGNED_IDENTITY_CLIENT_ID}
+  labels:
+    azure.workload.identity/use: "true"
```
aramase, line 200: This label in service account is no longer required starting from `v0.15.0` release. We've announced this as a breaking change in the `v1.0` alpha release: https://github.com/Azure/azure-workload-identity/releases/tag/v1.0.0-alpha.0. **tl;dr** This label is now required on the pod instead of the service account.

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,422 @@
...
+  --name "kubernetes-federated-credential" \
+  --identity-name "${USER_ASSIGNED_IDENTITY_NAME}" \
+  --resource-group "${RESOURCE_GROUP}" \
+  --issuer "${SERVICE_ACCOUNT_ISSUER}" \
+  --subject "system:serviceaccount:${SERVICE_ACCOUNT_NAMESPACE}:${SERVICE_ACCOUNT_NAME}"
+```
+
+#### <a name='DistributeKeys'></a>Distribute Keys To Management Cluster
+
+Setting up credentials for management cluster which is created from the bootstrap cluster is a key activity. This requires storing the keys on control plane node and patching kube-apiserver and kube-controller-manager flags to include the service account signing flags similar to what is discussed in the above `Set Service Account Signing Flags` section.
```
nawazkh, line 218:
```suggestion
Setting up credentials for management cluster which is created from the bootstrap cluster is a key activity. This requires storing the keys on control plane node and patching kube-apiserver and kube-controller-manager flags to include the service account signing flags similar to what is discussed in the above [Set Service Account Signing Flags](#set-service-account-signing-fl) section.
```

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,422 @@
...
+The identity is tied to the client ID which can be supplied via AzureClusterIdentity. In case identity is not found in AzureClusterIdentity client ID from the service account annotation
+`azure.workload.identity/client-id` will be used.
+
+#### <a name='Howtodistributekeys'></a>2. How to distribute key pair to management cluster?
+
+Keys can be distributed by using the CABPK feature by creating a secret with name `<cluster-name>-sa` encompassing the key pair. More details on it [here](https://cluster-api.sigs.k8s.io/tasks/certs/using-custom-certificates.html)
+
+#### <a name='UserExperience'></a>3. User Experience
+
+Though AZWI has a lot of advantages as compared to AZWI, setting up AZWI involves couple of manual step and it can impact the user experience.
```
nawazkh, line 375:
```suggestion
Though AZWI has a lot of advantages as compared to AAD-POD-IDENTITY, setting up AZWI involves couple of manual step and it can impact the user experience.
```
Did you mean to say the above?

nawazkh (commented):
I reviewed this mostly semantically and had some thought I've jotted down below. 
Thank you for putting this together @sonasingh46, this is so much effort! 🚀

docs/proposals/20221611-workload-identity-integration.md
sonasingh46: This is nice. Thanks for the heads up. Will include this in the doc.

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,422 @@
...
+
+## <a name='Summary'></a>Summary
+
+Workloads deployed in Kubernetes cluster may require Azure Active Directory application credential or managed identities to access azure protected resource e.g. Azure Key Vault, Virtual Machines etc. Azure AD Pod Identity helps access azure resources without the need of a secret management via Azure Managed Identities. Azure AD Pod Identity is now deprecated and Azure AD Workload Identity is the next iteration of the former. This design proposal aims to define the way for AZWI integration into capz for self managed clusters with keeping in mind other factor e.g. User Experience, Multi Tenancy, Backwards Compatibility etc.
+
+For more details about Azure AD Pod Identity please visit this [link](https://github.com/Azure/aad-pod-identity)  
+
+## <a name='Motivation'></a>Motivation
+
+AZWI gives the capability to federate the credentials with external identity providers in a Kubernetes native way. This approach overcomes several limitations of Azure AD Identity as mentioned below.
```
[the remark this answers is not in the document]
sonasingh46: Have added about the deprecation part below. Does that help?

docs/proposals/20221611-workload-identity-integration.md
sonasingh46: Yes. Have included this in the workflow.

sonasingh46: But for moving a existing cluster to use workload identity -- that will force to use the same key pairs encapsulated in the <cluster-name>-sa and we can think of removing this dependency going forward.

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,422 @@
...
+  extraMounts:
+      # path on node where the public key exists
+    - hostPath: ${SERVICE_ACCOUNT_KEY_FILE}
+      containerPath: /etc/kubernetes/pki/sa.pub
+      # path on node where the private key exists
+    - hostPath: ${SERVICE_ACCOUNT_SIGNING_KEY_FILE}
+      containerPath: /etc/kubernetes/pki/sa.key
+  kubeadmConfigPatches:
+  - |
+    kind: ClusterConfiguration
```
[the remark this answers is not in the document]
sonasingh46: That is intentional. 
https://azure.github.io/azure-workload-identity/docs/topics/self-managed-clusters/examples/kind.html#2-create-a-kind-cluster

docs/proposals/20221611-workload-identity-integration.md
sonasingh46: Have made the type immutable.

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,424 @@
...
+	- Install the AZWI mutating admission webhook controller.
+	- Establish federated credential for the key pair. 	 
+- Create a `UserAssignedIdentity` and update the machine template to use user assigned identity.
+
+- Patch the control plane to include the following flags on the api server.
+	- `service-account-issuer: <value-is-service-account-issuer-url>`
+	- `service-account-key-file:<public-key-path-on-controlplane-node>`
+	- `service-account-signing-key-file:<private-key-path-on-controlplane-node>`
+	- **Note:** The key pairs are present in `/etc/kubernetes/pki` directory if using the `<cluster-name>-sa` key pair.
+
```
randomvariable, line 403: Which of this is still required given kubeadm defaults most of this?

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,424 @@
...
+		- --service-account-signing-key-file
+		- --service-account-key-file
+	- kube-controller-manager
+		- --service-account-private-key-file
+
+- The operator/admin uploads the following two documents in a blob storage container. These documents is accessible publicy on a URL and this URL is commonly referred as issuer URL in this context. More about this [here](https://azure.github.io/azure-workload-identity/docs/installation/self-managed-clusters/oidc-issuer.html)  
+  - Generate and upload the Discovery Document.
+  - Generate and upload the JWKS Document.
+ 
+- The operator/admin installs the AZWI mutating admission webhook on the bootstrap cluster.
```
randomvariable, line 117: Workflow seems weird to me. I'm going to register a service account from my laptop as a federated identity provider in Azure?

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,424 @@
...
+
+Keys can be distributed by using the CABPK feature by creating a secret with name `<cluster-name>-sa` encompassing the key pair. More details on it [here](https://cluster-api.sigs.k8s.io/tasks/certs/using-custom-certificates.html)
+
+#### <a name='UserExperience'></a>3. User Experience
+
+Though AZWI has a lot of advantages as compared to AAD pod identity, setting up AZWI involves couple of manual step and it can impact the user experience.
+
+### <a name='Limitations'></a>Limitations
+
+- Cloud provider azure does not support workload identity yet and it also needs to authenticate to azure APIs.
```
CecileRobertMichon, line 381: https://github.com/kubernetes-sigs/cloud-provider-azure/commit/8d912fad87443dba2e2c3950a4d283fb4d718d31 merged but is not yet part of a release

Do we have a good idea of how it will be leveraged in CAPZ once it's supported? Will we need to install workload identity on the workload cluster before cloud-provider? if so how would that work since cloud-provider needs to be installed before the nodes can be "Ready" and Schedulable? 

@sonasingh46 @aramase

sonasingh46: If I am not wrong, only change that will be required is deploying cloud provider azure with workload identity label and rest will be taken care of. 
I will let @aramase also comment.

sonasingh46: >if so how would that work since cloud-provider needs to be installed before the nodes can be "Ready" and Schedulable?

Hmm.. This is a good point. Let me try experimenting with this. I can build a cloud provider image that has the workload identity support.

sonasingh46: Does that warrant workload identity to become a Daemonset? This is what I can think off while typing.

sonasingh46: Also when I think more about this. The defined approach for forming a management cluster is below. 
BootstrapCluster(On Kind, cloud provider not needed) --> ManagementCluster(cloud provider needed) 

As far as workload identity is considered, the defined/recommended approach will be : 
- Create a bootstrap cluster on kind using manual service principal.
- Use capi pivoting to create a management cluster on Azure. This still uses manual service principal. 
- Someone from the sec ops team migrates the management cluster to be using workload identity and the steps will be mentioned in the capz doc.

Above steps will not have the problem stated above.

sonasingh46: So if someone is using an existing cluster, the nodes will be ready, no? And then they can just perform the migration steps.
( Am I missing anything? )

aramase: > Do we have a good idea of how it will be leveraged in CAPZ once it's supported? Will we need to install workload identity on the workload cluster before cloud-provider? if so how would that work since cloud-provider needs to be installed before the nodes can be "Ready" and Schedulable?

This is a chicken and egg situation. The WI webhook is **not required** to use WIF. In this case, the cloud provider could just add the projected service account token volume as part of their pod spec by default and rely on their configuration to get the tenant id, client id and aad endpoint to use.

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,511 @@
...
+**Management Cluster on Azure Via Pivoting**
+
+Notes:
+- Cloud provider azure will be required to run on Kubernetes cluster in this case as the management cluster is created on Azure cloud.
+
+User Workflow:
+
+- All the steps are followed as described above in `Management Cluster on Kind` with a exception that a secret is created with name `<cluster-name>-sa` encompassing the key pairs that is generated in the previous step. This is done so that the key pairs gets distributed on the control plane node. More details on it [here](https://cluster-api.sigs.k8s.io/tasks/certs/using-custom-certificates.html)
+
+- A workload cluster is created using azure static credentials or aad pod identity.
```
CecileRobertMichon, line 160:
```suggestion
- A workload cluster is created using Azure static credentials or AAD Pod Identity. 
```

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,511 @@
...
+```
+
+#### <a name='DistributeKeys'></a>Distribute Keys
+
+Key pair can be distributed to a workload cluster by creating a secret with the name as `<cluster-name>-sa` encompassing the key pair. 
+Follow this [link](https://cluster-api.sigs.k8s.io/tasks/certs/using-custom-certificates.html) for more details.
+
+### <a name='CloudProviderAzureIntegration'></a>Cloud Provider Azure Integration
+
+- Cloud provider azure should be deployed with projected service account token config in the config YAML.
```
CecileRobertMichon, line 258: is this a setting on the Helm chart?

sonasingh46: It could be via helm or plain yaml.  However, cloud provider azure is deployed, the volume mount should be present on the config.

CecileRobertMichon:
Please squash commits and fix codespell errors

lgtm

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,511 @@
...
+  --identity-name "${USER_ASSIGNED_IDENTITY_NAME}" \
+  --resource-group "${RESOURCE_GROUP}" \
+  --issuer "${SERVICE_ACCOUNT_ISSUER}" \
+  --subject "system:serviceaccount:${SERVICE_ACCOUNT_NAMESPACE}:${SERVICE_ACCOUNT_NAME}"
+```
+
+#### <a name='DistributeKeys'></a>Distribute Keys
+
+Key pair can be distributed to a workload cluster by creating a secret with the name as `<cluster-name>-sa` encompassing the key pair. 
+Follow this [link](https://cluster-api.sigs.k8s.io/tasks/certs/using-custom-certificates.html) for more details.
```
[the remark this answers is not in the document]
sonasingh46: The signing keys that are used in kind cluster can be distributed to the workload cluster by creating a secret. Now, that same key pairs will be used by the apiserver and controller manager.

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,511 @@
...
+#### <a name='FederatedCredential'></a>Federated Credential
+
+The service account that the capz pod will use should be annotated with the client ID of the AAD application or User Assigned Identity and a federated identity should be created by associating the Service Account as a subject.
+
+```yaml
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  annotations:
+    azure.workload.identity/client-id: ${APPLICATION_CLIENT_ID:-$USER_ASSIGNED_IDENTITY_CLIENT_ID}
```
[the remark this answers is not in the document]
sonasingh46: No. Will remove this now.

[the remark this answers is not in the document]
sonasingh46: Infact, we do not need this whole service account section.

docs/proposals/20221611-workload-identity-integration.md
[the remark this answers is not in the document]
CecileRobertMichon: what's the need/use case to do that? is it optional or mandatory?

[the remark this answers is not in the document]
sonasingh46: This is optional. 

Use case is the following: 
- I set up a kind management cluster and have key pairs that can be used with workload identity. 
- I can easily just distribute these key pairs to the workload cluster that I created using this as I will be using workload identity on the workload cluster.

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,511 @@
...
+
+- Perform the following pre-requistes as discussed earlier in the **Bootstrap Cluster** section of the [document](#proposal):
+	- Generate and upload the JWKS and Discovery Document.
+	- Install the AZWI mutating admission webhook controller.
+	- Establish federated credential for the key pair.
+
+- If you are not using the existing key pairs then perform a service account rotation using this guide. 
+https://azure.github.io/azure-workload-identity/docs/topics/self-managed-clusters/service-account-key-rotation.html
+
+- If you are using the existing key pairs then rotation is not required.
```
sonasingh46, line 471: Or if you just set configure workload identity using the keys present in `/etc/kubernetes/pki`

docs/proposals/20221611-workload-identity-integration.md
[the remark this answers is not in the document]
sonasingh46: Or I will have to setup workload identity with keys that got generated for the workload cluster which resides in `/etc/kubernetes/pki` directory. 

Or I will have to rotate the keys with the keys with which the workload identity credential is configured. (As kubeadm generate key pairs by default if not specified by the secret)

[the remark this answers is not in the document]
aramase: > I can easily just distribute these key pairs to the workload cluster that I created using this as I will be using workload identity on the workload cluster.

Although this makes it easier for the user to setup WI in their workload cluster with less work in CAPZ, I wouldn't recommend doing this. As long as there is a 1:1 mapping between management cluster and workload cluster it's fine but if a single management cluster is used to create multiple workload clusters, allowing this would enable reusing same signing keys + issuer in different clusters which is insecure.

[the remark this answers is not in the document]
sonasingh46: >Although this makes it easier for the user to setup WI in their workload cluster with less work in CAPZ, I wouldn't recommend doing this. As long as there is a 1:1 mapping between management cluster and workload cluster it's fine but if a single management cluster is used to create multiple workload clusters, allowing this would enable reusing same signing keys + issuer in different clusters which is insecure.

Agreed. I think this does not come as recommendation. Though if this gives in impression of a recommendation like this, I will remove this section. 

If you the user story `Management Cluster Via Pivoting` -- this is where this becomes useful. 
https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2814/files/d62c690a3f2cb8a3fcb2e8de44f25e42b678eaed#diff-8f168651958e0a88d9402ac98e3b8764c9a4b54d29de8284def5e2c9070dd11bR158

So what happens in this case is -- 
1. The cloud admin creates a kind cluster 
2. The cloud admin create a workload cluster on azure ( 3 CP + 3 worker nodes )
3. The cloud admin wants to now convert this workload cluster to a mgmt cluster and user workload identity. 
4. The cloud admins nukes the kind cluster.

( This is the pivoting process followed by a lot of user to create a full fledged production management cluster ) 

So at step 2 if keys are not explicitly specified via the secret then kubeadm generates default ones and now there are some extra steps rotating those. So just to keep the life simple this is the approach.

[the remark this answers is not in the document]
sonasingh46: Also, this will be helpful in CI. 
We can pre create the secret using the same keys in the kind cluster in the CI. And now the cloud provider azure pods in workload cluster that got created will be capable of using workload identity.

CecileRobertMichon (commented):
/lgtm
/assign @aramase

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,496 @@
...
+
+- [S2] I am a cloud admin and I want to install CAP* on an existing Kubernetes cluster and make this a management cluster to create and manage workload clusters by using workload identity. For example, creating a management cluster on an already existing Kubernetes cluster on Azure.
+
+- [S3] As a cloud admin I want to migrate to using workload identity for my management cluster which is still using older AAD pod identity.   
+
+## <a name='Proposal'></a>Proposal
+
+In AZWI model, Kubernetes cluster itself becomes the token issuer issuing tokens to Kubernetes service accounts. These service accounts can be configured to be trusted on Azure AD application or user assigned managed identity. Workload pods can use this service account token which is projected to it's volume and can exchange the projected service account token for an Azure AD token.
```
aramase, line 115:
```suggestion
In AZWI model, Kubernetes cluster itself becomes the token issuer issuing tokens to Kubernetes service accounts. These service accounts can be configured to be trusted on Azure AD application or user-assigned managed identity. Workload pods can use this service account token which is projected to it's volume and can exchange the projected service account token for an Azure AD access token.
```

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,496 @@
...
+
+- The operator/admin uploads the following two documents in a blob storage container. These documents are accessible publicly on a URL and this URL is commonly referred as issuer URL in this context. More about this [here](https://azure.github.io/azure-workload-identity/docs/installation/self-managed-clusters/oidc-issuer.html)  
+  - Generate and upload the Discovery Document.
+  - Generate and upload the JWKS Document.
+
+- A federated identity credential should be created between the identity and <Issuer URL, service account namespace, service account name>. More on this [here](https://azure.github.io/azure-workload-identity/docs/topics/federated-identity-credential.html)
+
+- CAPI and CAPZ are deployed on the Kind cluster that supports workload identity.
+
+- CAPZ pod uses the client ID and tenant ID of the Azure AD or User Assigned Identity by passing it in AzureClusterIdentity CR. The AzureClusterIdentity also has a capability to specify type of identity as `WorkloadIdentity` on the field `type`.
```
aramase, line 146:
```suggestion
- CAPZ pod uses the client ID and tenant ID of the Azure AD or user-assigned Identity by passing it in AzureClusterIdentity CR. The AzureClusterIdentity also has a capability to specify type of identity as `WorkloadIdentity` on the field `type`. 
```

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,496 @@
...
+
+The identity is tied to the client ID which can be supplied via AzureClusterIdentity.
+
+#### <a name='Howtodistributekeys'></a>2. How to distribute key pair to management cluster?
+
+Keys can be distributed by using the CABPK feature by creating a secret with name `<cluster-name>-sa` encompassing the key pair. More details on it [here](https://cluster-api.sigs.k8s.io/tasks/certs/using-custom-certificates.html)
+
+#### <a name='UserExperience'></a>3. User Experience
+
+Though AZWI has a lot of advantages as compared to AAD pod identity, setting up AZWI involves couple of manual step and it can impact the user experience.
```
aramase, line 420: Explicitly call out the manual steps are required only for self-managed clusters used as management clusters. For managed clusters, the configuration steps are not required.

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,496 @@
...
+
+- Perform node rollout to propagate the changes by patching KCP and MachineDeployment objects to include the reference on the new AzureMachineTemplate created in the previous step.
+
+- Create a new `AzureClusterIdentity` and specify the client ID, tenant ID and `type` to `WorkloadIdentity`.
+
+- Patch AzureCluster to use the new `AzureClusterIdentity`.
+
+**Migrating Brownfield Clusters**
+
+- The steps in this section applies to existings clusters.
```
aramase, line 444:
```suggestion
- The steps in this section applies to existing clusters. 
```

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,496 @@
...
+**Migrating Brownfield Clusters**
+
+- The steps in this section applies to existings clusters. 
+
+- Generate a key pair or use the same key pairs as present in `/etc/kubernetes/pki`
+
+- Perform the following pre-requistes as discussed earlier in the **Bootstrap Cluster** section of the [document](#proposal):
+	- Generate and upload the JWKS and Discovery Document.
+	- Install the AZWI mutating admission webhook controller.
+	- Establish federated credential for the key pair.
```
aramase, line 451:
```suggestion
	- Establish federated credential for the identity and `<Issuer, service account namespace, service account name>`
```

docs/proposals/20221611-workload-identity-integration.md
```diff
@@ -0,0 +1,496 @@
...
+
+- Enable CAPZ pod to use AZWI on the management cluster to authenticate to Azure to create/update/delete resources as part of workload cluster lifecycle management
+
+### <a name='NonGoals'></a>Non Goals
+
+- Use workload identity for cloud provider azure once supported.
+
+- Migrate to using workload identity in CI pipelines.
+
+- Automation for migration from AAD pod identity to workload identity.
```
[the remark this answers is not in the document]
sonasingh46: done

aramase (commented):
/lgtm

CecileRobertMichon (commented):
/lgtm
/approve
/hold until EOD 6/15

CecileRobertMichon:
/hold cancel

Congrats @sonasingh46 🎉

Issue #2817, opened 2022-11-16
MachinePool reconciliation get stuck when resourceGroup name has any capital letter in it 


primeroz:
/kind bug

While testing machinepools i ended up with a stuck cluster due to the reconciliation of the machinepool machines never completing and getting stuck in **ScaleSetScalingUp** state,  with the machines never reaching a ready state even though they are up and running form the point of view of the workload cluster. 

After lots of testing this turns out to be related to the `resourceGroup` name , when the name has any capital letter in it the machines are not found during the reconcile loop and so their status cannot be set to ready. See steps below to reproduce 

**What steps did you take and what happened:**
Complete steps to reproduce the issue below

With a running kind cluster with clusterctl initialized and the cluster-identity-secret already created

---
***Working Example using upstream machinepool flavour***

* Create workload cluster
```
export AZURE_CLUSTER_IDENTITY_SECRET_NAME=cluster-identity-secret
export AZURE_CLUSTER_IDENTITY_SECRET_NAMESPACE=default
export AZURE_CONTROL_PLANE_MACHINE_TYPE=Standard_D2s_v3
export AZURE_LOCATION=westeurope
export AZURE_NODE_MACHINE_TYPE=Standard_D2s_v3 
export AZURE_SUBSCRIPTION_ID=XXX
export AZURE_CLIENT_ID=XXX
export AZURE_TENANT_ID=XXX
export CLUSTER_IDENTITY_NAME=cluster-identity
export WORKER_MACHINE_COUNT=2

clusterctl generate cluster upstream --kubernetes-version v1.25.3 --flavor machinepool > /dev/shm/upstream.yaml

kubectl apply -f /dev/shm/upstream.yaml 
```

* Wait for cluster control plane to be available and deploy calico
```
clusterctl get kubeconfig upstream > /dev/shm/upstream

wait for first CP node

kubectl --kubeconfig=/dev/shm/upstream apply -f https://raw.githubusercontent.com/kubernetes-sigs/cluster-api-provider-azure/main/templates/addons/calico.yaml
```

* Wait for machinepool to get provisioned and nodes to join the cluster
**The Cluster is successfully provisioned and all reconciliation is successfull 
```
$ kubectl --kubeconfig /dev/shm/upstream get node      
NAME                           STATUS   ROLES           AGE    VERSION
upstream-control-plane-8nlqm   Ready    control-plane   3m5s   v1.25.3
[...]

PR #2821, opened 2022-11-17
The providerID as returned from the VMSS instance needs to always be lower case to match the provideID defined in the kubernetes NODE Object

primeroz:
**What type of PR is this?**

/kind bug

There is a bug when using MachinePool with a resourceGroupName containing any uppercase letter. 

this is because in such a case the `providerID` saved in the `workload cluster Node Resource` and the one in the `azureMachinePoolMachine` do not match 

for example with a resourceGroup called `Tint`:
* NODE Resource providerID: azure:///subscriptions/XXXX/resourceGroups/`tilt`/providers/Microsoft.Compute/virtualMachineScaleSets/upstream-mp-0/virtualMachines/0
* AzureMachinePoolMachine Resource providerID: azure:///subscriptions/XXXX/resourceGroups/`Tilt`/providers/Microsoft.Compute/virtualMachineScaleSets/upstream-mp-0/virtualMachines/0

Fixes # https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/2817

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Convert the resourceGroup name in the AzureMachinePoolMachine object providerID to lowercase to match the providreID defined in the kubernetes Node Object on the workload cluster 
```

### DIFF
--- a/azure/scope/machinepoolmachine.go
+++ b/azure/scope/machinepoolmachine.go
@@ -32,6 +32,7 @@ import (
 	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
 	"sigs.k8s.io/cluster-api-provider-azure/azure"
 	infrav1exp "sigs.k8s.io/cluster-api-provider-azure/exp/api/v1beta1"
+	azureutil "sigs.k8s.io/cluster-api-provider-azure/util/azure"
 	"sigs.k8s.io/cluster-api-provider-azure/util/futures"
 	"sigs.k8s.io/cluster-api-provider-azure/util/tele"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
@@ -246,7 +247,13 @@ func (s *MachinePoolMachineScope) SetFailureReason(v capierrors.MachineStatusErr
 
 // ProviderID returns the AzureMachinePool ID by parsing Spec.FakeProviderID.
 func (s *MachinePoolMachineScope) ProviderID() string {
-	return s.AzureMachinePoolMachine.Spec.ProviderID
+	// Transform the VMSS instance resource representation to conform to the cloud-provider-azure representation
+	providerID, err := azureutil.ConvertResourceGroupNameToLower(s.AzureMachinePoolMachine.Spec.ProviderID)
+	if err != nil {
+		// return original providerID
+		return s.AzureMachinePoolMachine.Spec.ProviderID
+	}
+	return providerID
 }
 
 // PatchObject persists the MachinePoolMachine spec and status.
@@ -277,7 +284,7 @@ func (s *MachinePoolMachineScope) Close(ctx context.Context) error {
 // version of the VM instance if the node is found.
 // Note: This func should be called at the end of a reconcile request and after updating the scope with the most recent Azure data.
 func (s *MachinePoolMachineScope) UpdateNodeStatus(ctx context.Context) error {
-	ctx, _, done := tele.StartSpanWithLogger(
+	ctx, log, done := tele.StartSpanWithLogger(
 		ctx,
 		"scope.MachinePoolMachineScope.UpdateNodeStatus",
 	)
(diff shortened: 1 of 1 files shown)


primeroz:
@devigned thanks for pinging people. I am really just playing around here to understand the issue and familiarise myself with the codebase so by all means a PR by someone else would be probably much better :+1:

I really want to see if i can figure out how to save the providerID with the resourcegroupname in lower case in the azuremachinepoolmachine resource :eyes:

primeroz:
About saving the `providerID` in the right form it might not be the best idea. 

I can move the `lowerCase` conversion here https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/converters/vmss.go#L27 for the `ID` field which would save the providerID of the azuremachinepoolmachine correctly 

but this would not be backward compatible for already deployed machinepools 

`scope.MachinePoolScope.Close "msg"="failed to apply changes to the AzureMachinePoolMachines" "error"="failed creating AzureMachinePoolMachine: failed creating AzureMachinePoolMachine /subscriptions/a80dff0e-ab21-42a1-994a-f9285c4cfc92/resourceGroups/tilt/providers/Microsoft.Compute/virtualMachineScaleSets/upstream-mp-0/virtualMachines/0 in AzureMachinePool upstream-mp-0: azuremachinepoolmachines.infrastructure.cluster.x-k8s.io \"upstream-mp-0-0\" already exists"`

In _theory_ this would not cause problems to anyone already using lowercase resroucegroup names , and whoever is using uppercase resourcegroup names is not in a stable state at the moment ... but maybe is better to actually do the conversion when the property is read like i am doing now here ?

primeroz:
Hi, just wanted to check if there is anything you need me to do on this PR or is just waiting for review ? 

thanks :)

CecileRobertMichon:
@primeroz mostly needs review, this week is a bit slower due to US Thanksgiving holiday

Please add a release note and squash commits to help us get it ready for merge

/assign @mboersma

primeroz:
Release notes and Squash done , thanks for the support :pray:

mboersma (commented):
/lgtm

This seems like a reasonable approach, given the backward-compatibility requirement you pointed out with existing MPs.

azure/scope/machinepoolmachine.go
```diff
@@ -246,7 +247,13 @@ func (s *MachinePoolMachineScope) SetFailureReason(v capierrors.MachineStatusErr
 
 // ProviderID returns the AzureMachinePool ID by parsing Spec.FakeProviderID.
 func (s *MachinePoolMachineScope) ProviderID() string {
-	return s.AzureMachinePoolMachine.Spec.ProviderID
+	// Transform the VMSS instance resource representation to conform to the cloud-provider-azure representation
+	providerID, err := azureutil.ConvertResourceGroupNameToLower(s.AzureMachinePoolMachine.Spec.ProviderID)
+	if err != nil {
+		// return original providerID
+		return s.AzureMachinePoolMachine.Spec.ProviderID
```
CecileRobertMichon, line 254: for all the other ProviderIDs everywhere, we make sure to convert the RG name to lowercase before we even set it on the Object Spec. Where is this getting set for AzureMachinePoolMachine.Spec? 

This approach works to fix the CAPZ controller, but modifying the ID that gets set directly on the object would be slightly better IMO because then the Node and the CAPZ CRD provider IDs would actually match when the user looks at them. 

Side note: I opened #2892 as a follow-up because I was reminded that @jackfrancis did the work to make it exportable and we never followed up to use it.

CecileRobertMichon:
I think we'll want to cherry-pick this bug fix

/cherry-pick release-1.6
/cherry-pick release-1.5

azure/scope/machinepoolmachine.go
primeroz: While testing the change before opening this PR i did try to lowercase the provider ID for `AzureMachinePoolMachine` at write time. 

The reason why i decided to go for this is that if you had a cluster with the ongoing issue and updated the controller it would break even more since the controller would think that the 2 machines are different machines but it would fail to reconcile because the objects have the same name. 

With the change at read time updating the controller on a cluster experiencing the issue would cleanly fix and reconcile.

Maybe is not a real issue though since no one should be running a live production cluster with machinepools and a capitalized resourceGroup name ? 

I can definetely change it back to write time if you think that's the right way to do it

primeroz: @CecileRobertMichon i created a separate PR with the `fix on write` rather than the one on read 

https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2894

to keep things clean.

CecileRobertMichon: Thanks @primeroz. IMO #2894 is a better fix long term so we should go with that, especially given 

> no one should be running a live production cluster with machinepools and a capitalized resourceGroup name

If we wanted to we could technically do both (fix the root cause at write time then also merge this to fix existing MPs but IMO that's a bit overkill),

@mboersma @devigned any thoughts?

primeroz: > If we wanted to we could technically do both (fix the root cause at write time then also merge this to fix existing MPs but IMO that's a bit overkill),

I agree with you , and it might just become more confusing in code going forward for someone else looking at the code with no context and wondering why it happens in 2 different spots. 

but up to you :)

CecileRobertMichon:
/hold

we may not want to merge this

azure/scope/machinepoolmachine.go
mboersma: I think it's sufficient to fix it in one place, and #2894 is preferable.

CecileRobertMichon:
/close

in favor of #2894

PR #2822, opened 2022-11-17
test managed machine pool delete in e2e

nojnhuh:
**What type of PR is this?**
/kind cleanup

**What this PR does / why we need it**: This PR adds new e2e test coverage testing that a MachinePool can be successfully deleted from a managed cluster.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/aks.go
+++ b/test/e2e/aks.go
@@ -33,6 +33,7 @@ import (
 	"github.com/pkg/errors"
 	"golang.org/x/mod/semver"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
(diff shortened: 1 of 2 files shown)


willie-yao:
Is there a specific reason why we're only deleting the managed machine pool in the public ip test and not the other AKS tests? Otherwise, the deletion logic looks good to me!

nojnhuh:
> Is there a specific reason why we're only deleting the managed machine pool in the public ip test and not the other AKS tests? Otherwise, the deletion logic looks good to me!

There's no hard requirement for that decision, I don't think. Adding the delete here does make the test a bit more "symmetrical" though since that's where it's created. And I can see value in testing deleting machine pools both with and separately from the cluster.

Going forward, I think a good rule of thumb is to introduce minimal churn while keeping each spec from having any lingering side effects that bleed into other specs. By "minimal churn" I mean doing things like reusing a cluster or machine pool that already exists if possible to keep both costs down and velocity high by doing less work overall. That would justify having the manual machine pool scaling tests reuse the machine pools created with the cluster since we can scale them back to their initial state, but have the public IP prefix test create its own machine pool in the same cluster since those public IP fields on the machine pool are immutable and we'd otherwise need to tweak one of the original machine pools which is presumably used in many other tests.

jackfrancis:
@willie-yao @nojnhuh I love the idea of having these discrete behaviors be tested discretely, perhaps serially randomized per-test. Good thinking, let's aim for that in the future.

CecileRobertMichon:
> @willie-yao @nojnhuh I love the idea of having these discrete behaviors be tested discretely, perhaps serially randomized per-test. Good thinking, let's aim for that in the future.

I'm late to the party but... 100%. I would love us to consider how to do this for non-aks tests too, we should get to a point where it's not one feature == one test template but rather we have test specs we can use to add/remove node pools with a specific feature on any cluster "shape".

PR #2773, opened 2022-11-03
Uplift CAPI to v1.3.0 minor release

furkatgofurov7:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
This PR attempts to adopt and test CAPI **v1.3.0** release in CAPZ, more specifically:

By following https://github.com/kubernetes-sigs/cluster-api/blob/v1.3.0/docs/book/src/developer/providers/v1.2-to-v1.3.md, this PR uplifts dependencies:
- ~bumps Go version to 1.19.X~ (taken care by separate [PR](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2775)) ✅
- k8s.io/* to v0.25.0
- sigs.k8s.io/controller-tools to v0.10.0
- sigs.k8s.io/controller-runtime to v0.13.1
- Ginkgo to v2 (2.5.0) and Gomega to v1.24.1 and related changes
- cert-manager to v1.10.0
- kind version to v0.17.0
- github.com/joelanford/go-apidiff to 0.5.0
- KUBERNETES_VERSION to v1.24.6
- ETCD_VERSION_UPGRADE_TO 3.5.4-0
- COREDNS_VERSION_UPGRADE_TO v1.9.3
- KUBERNETES_VERSION_UPGRADE_TO 1.25
- KUBERNETES_VERSION_UPGRADE_FROM 1.24

Note: `make verify` & `make verify-conversions` and `make verify-modules` targets have been run and are passing locally. 

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #2769 

**Special notes for your reviewer**:
~Do we have to bump also images in https://github.com/kubernetes/test-infra/tree/master/config/jobs/kubernetes-sigs/cluster-api-provider-azure to use the `v1.25` image with Go `1.19`?~ (taken care by separate [PR](https://github.com/kubernetes/test-infra/pull/27894)) ✅ 

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**TODOs**:

- [ X] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

PR #2823, opened 2022-11-17
Add environment variables to run e2e test with an existing image

Jont828:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**: This PR adds environment variables to run e2e test with an existing image. This feature can be useful for users and developers that need to validate an existing image as currently the e2e test script will always build a new Docker image.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```


Jont828:
@CecileRobertMichon @jackfrancis I was hoping to get a bit of clarification about the `LOCAL_ONLY` variable. Is there any reason we set `REGISTRY=localhost:5000/ci-e2e` when we're not doing a docker push in the Make target?

CecileRobertMichon:
@Jont828 the `LOCAL_ONLY` variable is a way to run the tests without needing an external registry (eg. GCR, ACR) but instead uses the [kind local registry feature](https://kind.sigs.k8s.io/docs/user/local-registry/) to get the CAPZ/CAPI images to the management cluster. We build the image locally and load it directly, without needing to docker push/pull it externally. It doesn't run tests that require an external registry because they need to init a non-kind management cluster (eg. self-hosted test).

This is where we set up kind with a local registry: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/scripts/kind-with-registry.sh#L44

Jont828:
Okay gotcha, so in this case is `localhost:5000` just a placeholder (meaning it could be any url since we're not pushing)?

Makefile
```diff
@@ -662,6 +662,13 @@ test-e2e: ## Run "docker-build" and "docker-push" rules then run e2e tests.
 	$(MAKE) docker-build docker-push \
 	test-e2e-run
 
+.PHONY: test-e2e-existing
+test-e2e-existing:
```
CecileRobertMichon, line 666: it's not immediately clear what "existing"  means in this context, could be an existing cluster for example, consider naming this and the below target more explicitly

Makefile
```diff
@@ -662,6 +662,13 @@ test-e2e: ## Run "docker-build" and "docker-push" rules then run e2e tests.
...
+	$(MAKE) set-manifest-image MANIFEST_IMG=$(CONTROLLER_IMG)-$(ARCH) MANIFEST_TAG=$(TAG) TARGET_RESOURCE="./config/default/manager_image_patch.yaml" MANAGER_IMAGE=$(CONTROLLER_IMG)-$(ARCH):$(TAG)
+	$(MAKE) set-manifest-pull-policy TARGET_RESOURCE="./config/default/manager_pull_policy.yaml" PULL_POLICY=IfNotPresent
+	MANAGER_IMAGE=$(CONTROLLER_IMG)-$(ARCH):$(TAG) \
```
CecileRobertMichon, line 669: does MANAGER_IMAGE need to be set both here and above on line 667?

Makefile
```diff
@@ -671,6 +678,13 @@ test-e2e-local: ## Run "docker-build" rule then run e2e tests.
 	GINKGO_ARGS='$(LOCAL_GINKGO_ARGS)' \
 	test-e2e-run
 
+LOCAL_GINKGO_ARGS ?= -stream --progress
```
CecileRobertMichon, line 681: note: this is getting removed in #2773

scripts/ci-e2e.sh
```diff
@@ -42,6 +42,7 @@ source "${REPO_ROOT}/hack/util.sh"
 capz::util::ensure_azure_envs
 
 export LOCAL_ONLY=${LOCAL_ONLY:-"true"}
+export USE_EXISTING_IMAGE=${USE_EXISTING_IMAGE:-"false"}
```
CecileRobertMichon, line 45: Slight preference for the positive version of this so it's opt-out instead of opt-in:

```suggestion
export BUILD_MANAGER_IMAGE=${BUILD_MANAGER_IMAGE:-"true"}
```

Makefile
Jont828: Actually, I think we only need to set it on `make test-e2e-run` for an envsubst.

docs/book/src/developers/development.md
```diff
@@ -495,8 +495,13 @@ You can optionally set the following variables:
...
-| `LOCAL_ONLY`               | Use Kind local registry and run the subset of tests which don't require a remotely pushed controller image.                                                                                                                                                                                                                                                                | `true`                                                                                |
+| `LOCAL_ONLY`               | Use Kind local registry and run the subset of tests which don't require a remotely pushed controller image. If set, `REGISTRY` is also set to `localhost:5000/ci-e2e`.                                                                                                                                                                                                     | `true`                                                                                |
```
CecileRobertMichon, line 498: Thank you for adding docs 🙇

CecileRobertMichon (commented):
/assign @nojnhuh

Jont828:
@CecileRobertMichon Do we want to wait for #2773 to merge first?

Makefile
```diff
@@ -671,6 +671,13 @@ test-e2e-local: ## Run "docker-build" rule then run e2e tests.
 	GINKGO_ARGS='$(LOCAL_GINKGO_ARGS)' \
 	test-e2e-run
 
+.PHONY: test-e2e-skip-build-and-push
+test-e2e-skip-build-and-push:
```
nojnhuh, line 675: Is this basically equivalent to removing the `make docker-build` from the `test-e2e` target? If so, I think adding some sort of toggle controlling whether or not the `docker-build` gets invoked during `make test-e2e` might be a slightly lighter-weight and more flexible way of accomplishing the same goal vs. adding a separate target.

Jont828: Yep, it's removing the docker-build and docker-push steps. Do we want to include conditional logic in the Makefile? Another flag we have is `LOCAL_ONLY`, and we handled the conditional logic in `./scripts/ci-e2e.sh` so I followed that pattern.

nojnhuh: My rule of thumb would generally be to keep the conditional logic as close as possible to the behavior it's modifying. So in this case I'd tend to put a condition tweaking flags on the e2e test invocation in the Makefile vs. in ci-e2e.sh where it's kind of one level higher. I think that would be a bit easier to follow (Makefile syntax depending). If that would go against an established pattern though I think following that is likely more valuable.

nojnhuh:
This change from an open PR of mine should fix the e2e failure: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2861/files#diff-93ce65d81f5c1510135f6035d43f678bf9b5962b8e4ee45105e3345a1852efc6R583

I can't reproduce the timeout locally and it seems odd to me that the main branch jobs aren't also hitting this and that it's only come up recently, so I wonder if there's some weird Ginkgo stuff going on like https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2731#issuecomment-1298908474. I didn't see anything like that after a quick look though.

Jont828:
@nojnhuh I've rebased off of main today, looks like there were some failures on e2e and e2e-exp. Would your changes have overlapped with mine at all? I'm rerunning it to make sure it's not a flake.

nojnhuh:
> @nojnhuh I've rebased off of main today, looks like there were some failures on e2e and e2e-exp. Would your changes have overlapped with mine at all? I'm rerunning it to make sure it's not a flake.

@Jont828 It looks like you got the tests passing, but I can't think of any changes I've made recently that would conflict with these.

Jont828:
@nojnhuh @CecileRobertMichon Got the test working! Could I get a final pass of reviews for this?

nojnhuh:
@Jont828 Are we ok to lift the hold on this PR?

scripts/ci-e2e.sh
```diff
@@ -41,17 +41,21 @@ source "${REPO_ROOT}/hack/util.sh"
 # Verify the required Environment Variables are present.
 capz::util::ensure_azure_envs
 
-export LOCAL_ONLY=${LOCAL_ONLY:-"true"}
```
CecileRobertMichon, line 44: There are a bunch of jobs that currently set this variable and rely on it to not run certain tests in [our job definitions ](https://github.com/jackfrancis/test-infra/blob/c760cd560b643dda6377965fd481f7b2c9aaff30/config/jobs/kubernetes-sigs/cluster-api-provider-azure/cluster-api-provider-azure-presubmits-main.yaml) (I believe it's part of the anonymous-pull preset)

Merging this PR as-is would break all those jobs. We could update them before merging the PR to set both, or, and maybe that's safer, set the value of `USE_LOCAL_KIND_REGISTRY` to the value of `LOCAL_ONLY` if it's set for backwards compatibility purposes.

CecileRobertMichon: here where that variable is defined: https://github.com/kubernetes/test-infra/blob/166b008612e2d3319bfead89db593bc3bc52f127/config/prow/config.yaml#L950

Jont828: I think the latter option of maintaining backwards compatibility makes sense, I was just hoping to make the variable name a bit more descriptive.

Jont828:
Yes, I believe so!
/remove hold

scripts/ci-e2e.sh
Jont828: Sure thing, I'll update this in test-infra too. I wasn't planning on backporting it, but do you think it would make sense to do so?

CecileRobertMichon: not backporting is fine, but you'll have to keep the existing `LOCAL_ONLY` variable for a bit side by side for the jobs that run with the release branches

Jont828:
@CecileRobertMichon Are we good to merge this?

CecileRobertMichon:

to make sure the private cluster test still works with the LOCAL_ONLY var rename ^

CecileRobertMichon:
The test failed but the private cluster spec ran and passed which is what I was looking for

/retest
/lgtm
/approve

PR #2829, opened 2022-11-18
log ManagedCluster spec changes at higher verbosity

### LINKED ISSUE #2611: Show resource diffs for AKS clusters in the capz-controller-manager logs when making updates

nawazkh:
**What type of PR is this?**
/kind bug
/kind feature

- With this PR the user will be able to see AzureManagedCluster.spec updates in the CAPZ controller manager logs at a _verbosity level 4_ or higher. 
- This PR also updates the signature of `Parameters()` interface, adding `context.Context` as one of the inputs. 
- This PR also updates references to`Parameters()` in the codebase.

Fixes #2611 

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/azure/services/async/async_test.go
+++ b/azure/services/async/async_test.go
@@ -235,7 +235,7 @@ func TestCreateOrUpdateResource(t *testing.T) {
 				r.ResourceGroupName().Return("test-group")
 				s.GetLongRunningOperationState("test-resource", "test-service", infrav1.PutFuture).Return(nil)
 				c.Get(gomockinternal.AContext(), gomock.AssignableToTypeOf(&mock_azure.MockResourceSpecGetter{})).Return(&fakeExistingResource, nil)
-				r.Parameters(&fakeExistingResource).Return(&fakeResourceParameters, nil)
+				r.Parameters(gomockinternal.AContext(), &fakeExistingResource).Return(&fakeResourceParameters, nil)
 				c.CreateOrUpdateAsync(gomockinternal.AContext(), gomock.AssignableToTypeOf(&mock_azure.MockResourceSpecGetter{}), &fakeResourceParameters).Return("test-resource", nil, nil)
 			},
 		},
@@ -260,7 +260,7 @@ func TestCreateOrUpdateResource(t *testing.T) {
 				r.ResourceGroupName().Return("test-group")
 				s.GetLongRunningOperationState("test-resource", "test-service", infrav1.PutFuture).Return(nil)
 				c.Get(gomockinternal.AContext(), gomock.AssignableToTypeOf(&mock_azure.MockResourceSpecGetter{})).Return(nil, fakeNotFoundError)
-				r.Parameters(nil).Return(&fakeResourceParameters, nil)
+				r.Parameters(gomockinternal.AContext(), nil).Return(&fakeResourceParameters, nil)
 				c.CreateOrUpdateAsync(gomockinternal.AContext(), gomock.AssignableToTypeOf(&mock_azure.MockResourceSpecGetter{}), &fakeResourceParameters).Return(&fakeExistingResource, nil, nil)
 			},
 		},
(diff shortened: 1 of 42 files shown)


azure/services/managedclusters/managedclusters.go
```diff
@@ -79,6 +79,8 @@ func (s *Service) Reconcile(ctx context.Context) error {
 		return nil
 	}
 
+	// TODO: @nawazkh do you put the diff here?
+	// If yes, how do you compare the typeCasted struct b/w managedClusterSpec and containerservice.ManagedCluster ?
```
CecileRobertMichon, line 83: @nawazkh we are already calculating the diff in https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/services/managedclusters/spec.go#L330, the ask is to print it (and maybe also print a log message also when diff is nil) at high verbosity (to avoid making logs too noisy)

CecileRobertMichon: the only issue is that we're not currently passing in a context for the Parameters() func so this also requires changing the func signature and thus the `ResourceSpecGetter` interface to pass in a context

nawazkh: Thank you for the suggestion! I am able to puzzle the code flow together!
Testing with some local log changes before I make a change at interface level.

azure/services/async/async_test.go
```diff
@@ -260,7 +260,7 @@ func TestCreateOrUpdateResource(t *testing.T) {
 				r.ResourceGroupName().Return("test-group")
 				s.GetLongRunningOperationState("test-resource", "test-service", infrav1.PutFuture).Return(nil)
 				c.Get(gomockinternal.AContext(), gomock.AssignableToTypeOf(&mock_azure.MockResourceSpecGetter{})).Return(nil, fakeNotFoundError)
-				r.Parameters(nil).Return(&fakeResourceParameters, nil)
+				r.Parameters(context.TODO(), nil).Return(&fakeResourceParameters, nil)
```
CecileRobertMichon, line 263: for this (and all other unit test expect funcs) you can use `gomockinternal.AContext()` instead if `context.TODO()` so it works with any context that actually gets passed in, see the line above for the Get()

nawazkh: Thanks! I was going to test with context.Background(). But makes sense on using `gomockinternal.AContext()` too!

nawazkh: Addressed via ed9d6bba4c5dd98482b3e21a5a9f4cc2b23dfcf5

nawazkh: Waiting for tests..

azure/services/managedclusters/spec.go
```diff
@@ -329,8 +334,10 @@ func (s *ManagedClusterSpec) Parameters(existing interface{}) (params interface{
 
 		diff := computeDiffOfNormalizedClusters(managedCluster, existingMC)
 		if diff == "" {
+			log.V(4).Info("no changes found between user-updated spec and existing spec")
 			return nil, nil
 		}
+		log.V(4).Info("spec has been updated", "difference", diff)
```
CecileRobertMichon, line 340: nit: at that point nothing has been updated yet (diff could come from the Azure resource itself getting changed by the user outside of CAPZ). How about:

```suggestion
		log.V(4).Info("found a diff between the desired spec and the existing managed cluster", "difference", diff)
```

?

nawazkh: Good thought! adding it

nawazkh: Added via bb36e183ad2061a5f77073222e83ff0a54aa085d

nawazkh:
/retitle log ManagedCluster spec changes at higher verbosity

nawazkh:
@CecileRobertMichon ready for another look! Thanks for the suggestions earlier!

PR #2835, opened 2022-11-18
Fix image architecture bug in Docker build

### LINKED ISSUE #2834: Metadata for architecture on controller images is incorrect
/kind bug

**What steps did you take and what happened:** When you pull the arm64 version of the image as follows, it looks like the image is actually amd64.

```
$ docker pull registry.k8s.io/cluster-api-azure/cluster-api-azure-controller:v1.5.0 --platform linux/arm64
$ docker inspect registry.k8s.io/cluster-api-azure/cluster-api-azure-controller:v1.5.0 | grep Architecture

"Architecture": "amd64"
```

It looks like it's coming from the docker build commands since when you run `ARCH=arm64 make docker-build` the image metadata is still amd64.

**Environment:**

Jont828:
**What type of PR is this?**
/kind bug

Fixes #2834 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,6 +14,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Build architecture
+ARG ARCH
+
 # Build the manager binary
 FROM golang:1.19 as builder
 WORKDIR /workspace
@@ -52,7 +55,7 @@ RUN --mount=type=cache,target=/root/.cache/go-build \
     -o manager ${package}
 
 # Production image
-FROM gcr.io/distroless/static:nonroot
+FROM gcr.io/distroless/static:nonroot-${ARCH}
 WORKDIR /
 COPY --from=builder /workspace/manager .
 # Use uid of nonroot user (65532) because kubernetes expects numeric user when applying pod security policies


config/default/manager_image_patch.yaml
```diff
@@ -8,5 +8,5 @@ spec:
     spec:
       containers:
       # Change the value of image field below to your controller image URL
-      - image: gcr.io/k8s-staging-cluster-api-azure/cluster-api-azure-controller:latest
+      - image: docker.io/jont828/cluster-api-azure-controller-arm64:e2e
```
sonasingh46, line 11: Is it a miss -- The 'jont828' part of the image?

Jont828: Oops, let me fix that!

mboersma (commented):
Was the problem here that Docker was selecting using its default `ARCH` for the base image, so cross-platform builds failed?

zioproto:
Why you are not using `TARGETARCH` ?

https://docs.docker.com/desktop/extensions-sdk/extensions/multi-arch/#adding-multi-arch-binaries

zioproto:
So here is the root cause of the problem, it is in the way you call `docker build`:

https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/e552aab3a1a22378ea51aa81fe8d55cc915b5dd4/Makefile#L364-L367

Your patch will not fix #2834 because it does not fix the Docker image manifest.

Passing the  `--build-arg ARCH=$(ARCH)` is not the way to build Docker multiarch images. You are just passing `ARCH` as a parameter to the `go build`. To is just a env variable for your `go build` command. You are producing a golang binary for a specific architecture but you are not telling Docker that the image is meant for a specific architecture.

To notify the user that the image is for a specific architecture, there is an anti-pattern in place: you are ending up with multiple images for different architectures that differ by name, look in the tagging at the Docker push:

`-t $(CONTROLLER_IMG)-$(ARCH):$(TAG)`

But if you look at the image manifest, you will find the architecture of the machine that built that image.

The best practice is to have a single image name, and in the Docker image manifest state which platforms are supported.

Please review this doc:
https://docs.docker.com/desktop/extensions-sdk/extensions/multi-arch/

Your `docker build` in the Makefile should look like:

```
docker buildx build --push --platform=linux/amd64,linux/arm64 --tag=username/my-extension:0.0.1 .
```

and then in the `Dockerfile` you can use the built-in variables `BUILDARCH` and `TARGETARCH`. You should use `TARGETARCH` when calling `go build`.

nojnhuh:
@zioproto It looks to me like this will fix the issue because the `ARCH` build arg is used to construct the name of the final base image and selects the correct architecture. I can confirm this by building the image and checking its architecture with:

```console
$ REGISTRY=localhost:5000 ARCH=arm64 make docker-build && docker inspect localhost:5000/cluster-api-azure-controller-arm64:dev -f '{{.Architecture}}'
```

With these changes, the last line of that output will be `arm64` as expected. On `main`, the same command currently shows `amd64`.

Images for each architecture are built individually and a manifest pointing to each architecture is constructed more manually with the `docker manifest` commands, so the end goal of a single image reference for multiple architectures is still achieved. The buildx approach is another way to accomplish the same goal, but I'm guessing that wasn't as available or stable when these builds were first set up.

/lgtm

zioproto:
@nojnhuh I confirm you are right, I had totally missed this part of the Makefile:

https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/3c6d8b90b4dd7457bd1012e5fd740dc1e7e32464/Makefile#L393-L397

If I understand correctly the important line is:
`docker manifest annotate --arch $${arch} ${CONTROLLER_IMG}:${TAG} ${CONTROLLER_IMG}-$${arch}:${TAG}`

I checked it works like this, I clone the repo and then I do:

```
git fetch origin pull/2835/head:pr2835
git checkout pr2835
REGISTRY=zioproto make docker-build-all
for ARCH in s390x ppc64le arm64 arm amd64; do docker image inspect zioproto/cluster-api-azure-controller-${ARCH}:dev | jq ".[].Architecture" ; done
"s390x"
"ppc64le"
"arm64"
"arm"
"amd64"
```
LGTM

CecileRobertMichon (commented):
/lgtm
/approve

PR #2837, opened 2022-11-21
[cloud-provider-azure] Support specifying CCM cluster CIDR

lzhecheng:

 

 
**What type of PR is this?**
/kind cleanup

[cloud-provider-azure] Support specifying CCM cluster CIDR
Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
[cloud-provider-azure] Support specifying CCM cluster CIDR
```

### DIFF
--- a/scripts/ci-entrypoint.sh
+++ b/scripts/ci-entrypoint.sh
@@ -188,6 +188,15 @@ export ARTIFACTS="${ARTIFACTS:-${PWD}/_artifacts}"
 # create cluster
 create_cluster
 
+# Get CCM cluster CIDRs from Cluster object if not set
+if [[ -z "${CCM_CLUSTER_CIDR:-}" ]]; then
+    CCM_CLUSTER_CIDR=$(${KUBECTL} get cluster "${CLUSTER_NAME}" -o=jsonpath='{.spec.clusterNetwork.pods.cidrBlocks[0]}')
+    if CIDR1=$(${KUBECTL} get cluster "${CLUSTER_NAME}" -o=jsonpath='{.spec.clusterNetwork.pods.cidrBlocks[1]}' 2> /dev/null); then
+        CCM_CLUSTER_CIDR="${CCM_CLUSTER_CIDR:-}\,${CIDR1}"
+    fi
+fi
+echo "CCM cluster CIDR: ${CCM_CLUSTER_CIDR:-}"
+
 # export the target cluster KUBECONFIG if not already set
 export KUBECONFIG="${KUBECONFIG:-${PWD}/kubeconfig}"
 
@@ -225,7 +234,8 @@ if [[ -n "${TEST_CCM:-}" ]]; then
         --set cloudControllerManager.enableDynamicReloading="${ENABLE_DYNAMIC_RELOADING}" \
         --set cloudControllerManager.cloudConfig="${CLOUD_CONFIG}" \
         --set cloudControllerManager.cloudConfigSecretName="${CONFIG_SECRET_NAME}" \
-        --set cloudControllerManager.logVerbosity="${CCM_LOG_VERBOSITY}"
+        --set cloudControllerManager.logVerbosity="${CCM_LOG_VERBOSITY}" \
+        --set-string cloudControllerManager.clusterCIDR="${CCM_CLUSTER_CIDR}"
 
     echo "Waiting for all kube-system pods to be ready"
     "${KUBECTL}" wait --for=condition=Ready pod -n kube-system --all --timeout=10m


scripts/ci-entrypoint.sh
```diff
@@ -212,6 +212,7 @@ if [[ -n "${TEST_CCM:-}" ]]; then
     fi
 
     export CCM_LOG_VERBOSITY="${CCM_LOG_VERBOSITY:-4}"
+    export CCM_CLUSTER_CIDR="${CCM_CLUSTER_CIDR:-10.244.0.0/16}"
```
CecileRobertMichon, line 215: where is this default value coming from? Is there any way we can determine it programmatically instead of hardcoding a default?

CecileRobertMichon: Also what about dual stack clusters that have two cidrs?

lzhecheng: The default value is in the readme file and I don't find a good way to avoid hardcoding? https://github.com/kubernetes-sigs/cloud-provider-azure/blob/c8640b1bee6022318753bde033c8eb4d3089ebe7/helm/cloud-provider-azure/README.md#configurable-values 

If dualstack, then users should specify the cluster cidr. And this is the purpose of my change.

lzhecheng: @CecileRobertMichon sorry, I misunderstood your question. Yes, cluster config sounds reasonable and I think we should use the `cidrBlocks` values. Is the "not look at" a typo or you suggest some other fields?

`cloudControllerManager.clusterCIDR` of ccm can take 2 values, e.g. "10.244.0.0/16\,2001:1234:5678:9a40::/64"

scripts/ci-entrypoint.sh
```diff
@@ -212,6 +212,18 @@ if [[ -n "${TEST_CCM:-}" ]]; then
     fi
 
     export CCM_LOG_VERBOSITY="${CCM_LOG_VERBOSITY:-4}"
+    # Get CCM cluster CIDRs from cluster config template if not set
+    if [[ -z "${CCM_CLUSTER_CIDR:-}" ]]; then
+        curl -LO https://github.com/mikefarah/yq/releases/download/v4.30.3/yq_linux_amd64 && chmod +x yq_linux_amd64
```
CecileRobertMichon, line 217: yq is already part of ` hack/tools/bin` and there is a make target to install it, it can be installed like other binaries in this file https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2837/files#diff-2009f7be5363812e22862771f557363795759e67b3ca2e32c50d9fd5f8c5669bR31

We don't want to be hardcoding dependency versions in scripts

scripts/ci-entrypoint.sh
```diff
@@ -212,6 +212,18 @@ if [[ -n "${TEST_CCM:-}" ]]; then
...
+        while read -r CCM_CIDR; do
+            if [[ "${CCM_CLUSTER_CIDR:-}" == "" ]]; then
+                CCM_CLUSTER_CIDR="${CCM_CIDR}"
+            else
+                CCM_CLUSTER_CIDR="${CCM_CLUSTER_CIDR:-}\,${CCM_CIDR}"
+            fi
+        done <<< "$(./yq_linux_amd64 '.spec.clusterNetwork.pods.cidrBlocks' "${CLUSTER_TEMPLATE}" --no-doc|grep -v null | awk '{print $2}')"
```
CecileRobertMichon, line 224: Why do we need to grep the template here, can't we just use kubectl to get the value? At this point `Cluster`  is an object on the management cluster

CecileRobertMichon: You need to do this before line 191 so we're still pointing at the management cluster kubeconfig but then you should be able to do something like:

` "${KUBECTL}" get cluster "${CLUSTER_NAME}" -o=jsonpath='{.spec.clusterNetwork.pods.cidrBlocks}'`

lzhecheng: @CecileRobertMichon Thank you for the advice. PTAL.

scripts/ci-entrypoint.sh
```diff
@@ -188,6 +188,17 @@ export ARTIFACTS="${ARTIFACTS:-${PWD}/_artifacts}"
 # create cluster
 create_cluster
 
+# Get CCM cluster CIDRs from cluster config template if not set
```
CecileRobertMichon, line 191:
```suggestion
# Get CCM cluster CIDRs from Cluster object if not set
```

CecileRobertMichon:
@lzhecheng tests are failing, please check

scripts/ci-entrypoint.sh
lzhecheng: done

lzhecheng:
> @lzhecheng tests are failing, please check

PTAL

scripts/ci-entrypoint.sh
```diff
@@ -188,6 +188,16 @@ export ARTIFACTS="${ARTIFACTS:-${PWD}/_artifacts}"
 # create cluster
 create_cluster
 
+# Get CCM cluster CIDRs from Cluster object if not set
+if [[ -z "${CCM_CLUSTER_CIDR:-}" ]]; then
+    CIDR0=$(${KUBECTL} get cluster "${CLUSTER_NAME}" -o=jsonpath='{.spec.clusterNetwork.pods.cidrBlocks[0]}')
```
CecileRobertMichon, line 193: nit: do we need the extra `CIDR0` variable? I see we're just setting it below to `CCM_CLUSTER_CIDR` , could this just be:
```suggestion
    CCM_CLUSTER_CIDR=$(${KUBECTL} get cluster "${CLUSTER_NAME}" -o=jsonpath='{.spec.clusterNetwork.pods.cidrBlocks[0]}')
```

?

scripts/ci-entrypoint.sh
```diff
@@ -225,7 +235,8 @@ if [[ -n "${TEST_CCM:-}" ]]; then
         --set cloudControllerManager.enableDynamicReloading="${ENABLE_DYNAMIC_RELOADING}" \
         --set cloudControllerManager.cloudConfig="${CLOUD_CONFIG}" \
         --set cloudControllerManager.cloudConfigSecretName="${CONFIG_SECRET_NAME}" \
-        --set cloudControllerManager.logVerbosity="${CCM_LOG_VERBOSITY}"
+        --set cloudControllerManager.logVerbosity="${CCM_LOG_VERBOSITY}" \
+        --set-string cloudControllerManager.clusterCIDR="${CCM_CLUSTER_CIDR}"
```
CecileRobertMichon, line 239: @jackfrancis do we need to do this in https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/test/e2e/cloud-provider-azure.go#L44 as well?

@lzhecheng what's the impact if the clusterCIDR of CCM doesn't match the Cluster's CIDR(s)?

scripts/ci-entrypoint.sh
lzhecheng: I wanted to have `CIDR0` and `CIDR1` 2 variables for IPv4 and IPv6 but I have no strong preference. Updated.

scripts/ci-entrypoint.sh
lzhecheng: clusterCIDR is for route and node ipam controllers so I think these 2 components won't work properly.

CecileRobertMichon (commented):
/lgtm
/assign @jackfrancis

scripts/ci-entrypoint.sh
CecileRobertMichon: https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/2852

jackfrancis: Thanks for filing the issue @CecileRobertMichon, I agree we can add this CIDR configuration to our PR E2E scenarios as a follow-up step

jackfrancis:
/hold

sorr @lzhecheng but let's get https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2495 in first, which is a large change that's been waiting a long time in the queue :)

cc @CecileRobertMichon

CecileRobertMichon:
/hold

we have a rebase err at main, need to fix it first

jackfrancis:
/hold cancel

we can actually land this first

PR #2838, opened 2022-11-21
AzureManagedCluster: Full support for AKS Cluster Autoscaler

### LINKED ISSUE #2826: AzureManagedCluster E2E: Autoscaler
In order to take AzureManagedCluster out of Experimental we need to have full support for the Autoscaler. This means we need to have both the ability to set the Autoscaler configuration and have full E2E tests that validate that the Autoscaler works with AzureManagerCluster.

mtougeron:
**What type of PR is this?**

/kind feature

Adds full support for the AKS cluster-autoscaler for AzureManagerCluster

Fixes #2826

**TODOs**:

- [X] squashed commits
- [x] includes documentation
- [X] adds unit tests

**Release note**:

```release-note
AzureManagedCluster: Full support for AKS Cluster Autoscaler
```


mtougeron:
Still to do:

- [ ] Docs of how to configure the cluster autoscaler
- [ ] e2e test to validate that the cluster autoscaler was properly manipulated
- [ ] better unit tests on value validations

mtougeron:
Sorry for the noise, I'm having trouble getting the e2e tests to run locally so I'm pushing here temporarily until I figure that out so it runs while I troubleshoot my local setup.

test/e2e/aks.go
```diff
@@ -266,7 +267,7 @@ func GetWorkingAKSKubernetesVersion(ctx context.Context, subscriptionID, locatio
 	if err != nil {
 		return "", errors.Wrap(err, "failed to create an Authorizer")
 	}
-	containerServiceClient := containerservice.NewContainerServicesClient(subscriptionID)
+	containerServiceClient := containerservice2020.NewContainerServicesClient(subscriptionID)
 	containerServiceClient.Authorizer = authorizer
 	result, err := containerServiceClient.ListOrchestrators(ctx, location, ManagedClustersResourceType)
```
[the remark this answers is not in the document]
mtougeron: I should have linked to [AgentPoolsClient.GetAvailableAgentPoolVersions](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go@v63.4.0+incompatible/services/containerservice/mgmt/2022-03-01/containerservice#AgentPoolsClient.GetAvailableAgentPoolVersions)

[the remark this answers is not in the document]
CecileRobertMichon: The main difference I see with that function is that it requires an existing cluster which our existing function doesn't, since we're introspecting versions before creating the cluster

I've asked some folks on the AKS team to help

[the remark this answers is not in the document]
mtougeron: Actually, neither of those are going to work because they need a running cluster in order to lookup the possible versions for that cluster. In our scenario we need the version number in order to _create_ the cluster.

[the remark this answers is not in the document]
mtougeron: Well, I checked with the latest Python client and it downgrades the api version it calls to get the list of kubernetes versions available.

```
urllib3.connectionpool: https://management.azure.com:443 "GET /subscriptions/REDACTED/providers/Microsoft.ContainerService/locations/eastus/orchestrators?api-version=2019-04-01&resource-type=managedClusters HTTP/1.1" 200 None
```

```
$> az version                                   
{
  "azure-cli": "2.42.0",
  "azure-cli-core": "2.42.0",
  "azure-cli-telemetry": "1.0.8",
  "extensions": {
    "aks-preview": "0.5.117"
  }
}
```

@CecileRobertMichon I'm going to need some help determining if there's an updated equivalent of the Orchestrators API but so far I can't seem to find one to list the Kubernetes API versions.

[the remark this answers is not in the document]
CecileRobertMichon: @mtougeron I checked with AKS folks and currently there is no replacement :( They have plans to add a new API to list versions but for now looks like we're stuck using the old SDK.

exp/api/v1beta1/azuremanagedcontrolplane_types.go
```diff
@@ -247,6 +251,122 @@ type AzureManagedControlPlaneStatus struct {
 	LongRunningOperationStates infrav1.Futures `json:"longRunningOperationStates,omitempty"`
 }
 
+// AutoScalerProfile parameters to be applied to the cluster-autoscaler.
+// See the [FAQ](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-are-the-parameters-to-ca) for more details about each parameter.
+type AutoScalerProfile struct {
+	// BalanceSimilarNodeGroups - Valid values are 'true' and 'false'. The default is false.
+	// +kubebuilder:validation:Enum="true";"false"
+	// +optional
+	BalanceSimilarNodeGroups *BalanceSimilarNodeGroups `json:"balanceSimilarNodeGroups,omitempty"`
```
jackfrancis, line 260: @CecileRobertMichon @nojnhuh while reviewing this it occurs to me that this approach is different than the one I took in this PR:

- https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2781

An example:

- https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2781/files#diff-585942e07c6d3dd4ca6c2d35efdca0fa27c6a7bf6e3d6711648edfad330fb3adR46-R49

I think the approach here is safer, but perhaps requires more maintenance. I'm happy to adopt it in my PR, I'd love to have these type definitions be standardized.

CecileRobertMichon: I'm surprised it's different, I thought this approach above was exactly what we talked about doing for you PR as well when you were screen sharing the other day, it's also what we're doing already in the repo in most places AFAIK, eg. https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/api/v1beta1/types.go#L459

jackfrancis: The approach I’m talking about is wrapping the string type in a strongly typed “string” like thing, and then having a const for each allowed enum value.

I’m pretty sure we’re not doing this everywhere, but let’s do this from here on out. The only downside I can see to this is more boilerplate overhead; I can’t see a reason (other than dev laziness) not to do it this way as a rule. Curious if there are other reasons mot to bother creating our own types + consts.

mtougeron: Since it's been done both ways in the code I'm glad I chose wisely... While the additional boilerplate was kind of annoying, especially with the pointer aspect, it was pretty minor and a one-time thing. I do like it better than just calling it a string in the type.

test/e2e/aks.go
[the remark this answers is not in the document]
mtougeron: Thanks for checking, I appreciate it.

exp/api/v1beta1/azuremanagedcontrolplane_types.go
jackfrancis: I'll adapt my PR to follow this pattern as well.

exp/api/v1beta1/azuremanagedcontrolplane_webhook.go
```diff
@@ -486,3 +493,183 @@ func (m *AzureManagedControlPlane) validateName(_ client.Client) error {
...
+
+// validateAutoScalerProfile validates an AutoScalerProfile.
+func (m *AzureManagedControlPlane) validateAutoScalerProfile(_ client.Client) error {
+	var allErrs field.ErrorList
+
+	if m.Spec.AutoScalerProfile == nil {
+		return nil
+	}
+
+	if m.Spec.AutoScalerProfile.BalanceSimilarNodeGroups != nil {
```
jackfrancis, line 505: Because we're already enforcing this via the kubebuilder API spec, we don't need to add additional webhook enforcement here

azure/services/managedclusters/spec.go
```diff
@@ -417,6 +484,52 @@ func computeDiffOfNormalizedClusters(managedCluster containerservice.ManagedClus
 		}
 	}
 
+	// TODO: Not sure this comparison is correct since defaults are set at AKS level, do checks
```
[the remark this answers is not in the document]
nojnhuh: I'll keep looking at this tomorrow, but I think this might need to be split into checks for each field unfortunately. I tried creating a cluster without any autoscaler settings, changing one of them, and changing it back. The first change went as expected, but the second update seemed like it got the controller caught in a loop trying to update it but it was never taking effect. I'll dig in more tomorrow to figure out exactly what's going on in that scenario.

exp/api/v1beta1/azuremanagedcontrolplane_webhook.go
```diff
@@ -486,3 +493,183 @@ func (m *AzureManagedControlPlane) validateName(_ client.Client) error {
...
+
+// validateMaxNodeProvisionTime validates update to AutoscalerProfile.MaxNodeProvisionTime.
+func (m *AzureManagedControlPlane) validateMaxNodeProvisionTime() field.ErrorList {
+	var allErrs field.ErrorList
+	if to.String(m.Spec.AutoScalerProfile.MaxNodeProvisionTime) != "" {
+		_, err := time.ParseDuration(to.String(m.Spec.AutoScalerProfile.MaxNodeProvisionTime))
+		if err != nil {
+			allErrs = append(allErrs, field.Invalid(field.NewPath("Spec", "AutoscalerProfile", "MaxNodeProvisionTime"), m.Spec.AutoScalerProfile.MaxNodeProvisionTime, "invalid value"))
+		}
+		if !rMaxNodeProvisionTime.MatchString(to.String(m.Spec.AutoScalerProfile.MaxNodeProvisionTime)) {
```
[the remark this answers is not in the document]
mtougeron: I believe you are correct. I'll remove the `time.ParseDuration`

exp/api/v1beta1/azuremanagedcontrolplane_webhook.go
mtougeron: removed

azure/services/managedclusters/spec.go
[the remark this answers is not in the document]
mtougeron: yeah, tested setting the entire block of options and it syncs fine but with some missing it doesn't reconcile as I had hoped. It's looking like I'd need to set the default values for all settings if any value is set in the CR but I don't really like that :/ I'll think on it some more tomorrow.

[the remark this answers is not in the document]
mtougeron: I ended up setting default values for all the settings in the CR. It's not ideal imo, but it's the only way I could think of that handled the scenario when someone set a value but then removed it from the CR.

mtougeron:
@nojnhuh @jackfrancis @CecileRobertMichon I believe I've updated and adjusted this PR based on all the feedback I've gotten so far.

docs/book/src/topics/managedcluster.md
```diff
@@ -312,6 +312,34 @@ spec:
...
+    newPodScaleUpDelay: "0s"
+    okTotalUnreadyCount: "3"
+    scanInterval: "10s"
+    scaleDownDelayAfterAdd: "10m"
+    scaleDownDelayAfterDelete: "10s"
+    scaleDownDelayAfterFailure: "3m"
+    scaleDownUnneededTime: "10m"
+    scaleDownUnreadyTime: "20m"
+    scaleDownUtilizationThreshold: "0.5"
+    skipNodesWithLocalStorage: "true"
```
jackfrancis, line 339: this doc disagrees with the default applied below

mtougeron: fixed this doc so it matches what is actually applied by AKS rather than the defaults from the cluster-autoscaler docs

mtougeron:
weird docker test setup failure. retrying...

/retest

mtougeron:
`pull-cluster-api-provider-azure-e2e` failed due to github rate limiting. Will try again in an hour or so.

azure/services/managedclusters/spec.go
```diff
@@ -427,6 +535,57 @@ func computeDiffOfNormalizedClusters(managedCluster containerservice.ManagedClus
...
+			ScaleDownUnneededTime:         existingMC.AutoScalerProfile.ScaleDownUnneededTime,
+			ScaleDownUnreadyTime:          existingMC.AutoScalerProfile.ScaleDownUnreadyTime,
+			ScaleDownUtilizationThreshold: existingMC.AutoScalerProfile.ScaleDownUtilizationThreshold,
+			SkipNodesWithLocalStorage:     existingMC.AutoScalerProfile.SkipNodesWithLocalStorage,
+			SkipNodesWithSystemPods:       existingMC.AutoScalerProfile.SkipNodesWithSystemPods,
+		}
+	}
+
+	// Once the AKS autoscaler has been updated it will always return values so we need to
+	// respect those values even though the settings are now not being explicitly set by CAPZ.
```
nojnhuh, line 583: I gave a good shake the past couple days at seeing if we can obscure how AKS sets either all or none of the autoscaler settings but didn't quite get that totally working. I was hoping to get to where we don't need to explicitly set all the defaults in CAPZ, and nil values in the spec would passively assume the AKS default. But since it's impossible to know whether or not any existing value is user-defined or a default, we can't know whether or not a nil value in the spec should trigger an update (i.e. when a user-defined value should be reset) or not.

test/e2e/aks_azure_cluster_autoscaler.go
```diff
@@ -0,0 +1,83 @@
...
+	containerserviceClient.Authorizer = auth
+
+	amcp := &infrav1exp.AzureManagedControlPlane{}
+	err = mgmtClient.Get(ctx, types.NamespacedName{
+		Namespace: input.Cluster.Spec.ControlPlaneRef.Namespace,
+		Name:      input.Cluster.Spec.ControlPlaneRef.Name,
+	}, amcp)
+	Expect(err).NotTo(HaveOccurred())
+
+	// Verify that it's not set before making any changes.
```
nojnhuh, line 60: Lately we've been trying to structure new tests like this to be idempotent and not to assume anything about the initial state of the cluster to help make them easier to plug in to other tests down the road.

The latest test I added in #2917 establishes a decent pattern for this I think. Roughly:
1. Save the initial state
2. Delete the existing state (which might be a no-op)
3. Create
4. Update
5. Update back to the initial state (effectively a delete when the initial state was empty)

nojnhuh (commented):
Tried this out and it works well overall. Left one small comment inline about deleting all of `spec.autosclaingProfile`.

azure/services/managedclusters/spec.go
```diff
@@ -181,6 +222,69 @@ func (s *ManagedClusterSpec) CustomHeaders() map[string]string {
 	return s.Headers
 }
 
+// buildAutoScalerProfile builds the AutoScalerProfile for the ManagedClusterProperties.
+func buildAutoScalerProfile(autoScalerProfile *AutoScalerProfile) *containerservice.ManagedClusterPropertiesAutoScalerProfile {
+	if autoScalerProfile == nil {
+		return nil
```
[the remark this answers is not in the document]
mtougeron: If I don't set it to `nil` here, it will potentially impact anyone not managing those settings via CAPZ and create a scenario where CAPZ is continuously trying to reconcile the state. (see https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2838/files#r1061928877)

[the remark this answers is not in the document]
nojnhuh: I know it's possible to express replica count for machine pool as being managed outside of CAPI/CAPZ, but it seems like the predominant assumption is that CAPZ is authoritative, where changes made externally are expected to be overwritten to match CAPZ's view.

I think following that assumption that CAPZ will always manage the autoscaler settings would be most consistent with the rest of the API, but I could see possibly adding something like the [`cluster.x-k8s.io/replicas-managed-by`](https://github.com/kubernetes-sigs/cluster-api/blob/v1.3.1/api/v1beta1/common_types.go#L134-L138) annotation which would communicate that an empty `spec.autoscalerProfile` means "don't change the current settings" (nil) instead of "assume all default values" (empty, non-nil).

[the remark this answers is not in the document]
mtougeron: Do we want to takeover and potentially stomp on existing settings when this is released? Because of the way the settings come back from AKS I can't see a way to do this without _always_ setting all of the default values at first reconcile. That would mean it's a breaking change because we'd need to give users an opportunity to set that annotation ahead of the release, right? I would think it's more appropriate to say that if it isn't set in the manifest that it's a no-op. I might be missing something here though or over thinking this.

lmk if we should take this to a more real-time format to discuss the various scenarios I've run into with testing on this.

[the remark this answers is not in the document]
nojnhuh: @jackfrancis Have we ever run into this situation before, where the introduction of a new API field could cause CAPZ to override configuration set by the user by other means (since that was not yet exposed through CAPZ at the time)?

azure/services/managedclusters/spec.go
```diff
@@ -305,6 +409,8 @@ func (s *ManagedClusterSpec) Parameters(ctx context.Context, existing interface{
 		}
 	}
 
+	managedCluster.AutoScalerProfile = buildAutoScalerProfile(s.AutoScalerProfile)
```
jackfrancis, line 412: @nojnhuh @mtougeron to speak to the thread about forward-compatibility and how to deal with clusters that have no declared `AutoScalerProfile` details (because prior versions of CAPZ didn't expose this) it might just be easier to do this:

```go
if s.AutoScalerProfile != nil {
    managedCluster.AutoScalerProfile = buildAutoScalerProfile(s.AutoScalerProfile)
}

```

nojnhuh: At first glance that seems equivalent to the condition currently inside `buildAutoScalerProfile` which comes with the same limitation that once any value is set, none of them can ever be reset to back to AKS's implicit defaults:
https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/47d8731b86cb220ee4bfee7d833b4a7a90e7abf9/azure/services/managedclusters/spec.go#L227-L229

jackfrancis: Ah, sorry, I thought we were in the flow of reconciling existing AKS API data w/ local CAPZ resource data.

So, what I'm thinking is: can we passively apply AKS API's version of the state if the CAPZ resource data is empty, and only definitively update the AKS API if we actually have an opinion?

nojnhuh: I think the nuance here is that a possible opinion is "reset to the implicit defaults" which is indistinguishable from "use what's already defined in AKS" in the CAPZ API. Is there a way to construct the YAML such that the controller ultimately reads a non-nil, empty value from the API? If something like
```yaml
autoscalerProfile: {}
```
is non-nil and omitting it entirely (or `null`) is nil, then I think everything can work the way we want as-is. I vaguely remember trying something similar a long time ago and `{}` still came through as nil.

test/e2e/aks_azure_cluster_autoscaler.go
mtougeron: I made these changes. Can you take a look and let me know if it matches what you were expecting?

mtougeron:
I can't reproduce this error locally. It seems to build fine for me; trying again.

azure/services/managedclusters/spec.go
nojnhuh: I tried setting an explicit empty map like `autoscalerProfile: {}` in the YAML and that actually comes through as a non-nil, empty struct value the way I hoped. The defaulting in the webhook still populates all the fields in that case though, so that can't be used to reset all the fields back to the implicit defaults. And the defaulting is necessary to accurately detect when we need to trigger an update to the AKS instance.

I suppose there are workarounds we could experiment with like treating `{}` as a special value in the webhook or moving CAPZ's awareness of the defaults from the webhook further down closer to here, but those are hacky enough that they're not worth adding on a first pass and could maybe be introduced in a backwards-compatible manner later.

Sorry @mtougeron if I'm only now understanding things you figured out a while ago, but I think I at least have a more solid grasp on things now.

test/e2e/aks_azure_cluster_autoscaler.go
nojnhuh: Overall looks good! I made the above suggestion to model after that tags test before I understood there's a possibility we might not expose a "delete" so then it makes sense if this new test isn't a perfect 1:1 match with that one.

nojnhuh:
> I can't reproduce this error locally. It seems to build fine for me; trying again.

There's a known issue with AKS currently affecting Windows node pools using Calico CNI with Kubernetes 1.25 (https://github.com/Azure/AKS/issues/3394), so timeouts provisioning the cluster are expected to be exceeded depending on in which region the test happens to run. #2999 should work around the issue on our side.

azure/services/managedclusters/spec.go
jackfrancis: I think the tl;dr here for me is let's punt this out of this PR's scope for now, I think this is (unfortunately) one of many instances of this "property of properties" challenge that we have wrapping AKS config. Aspirationally we can come up with an approach that generalizes across all of these for CAPZ + AKS?

azure/services/managedclusters/spec.go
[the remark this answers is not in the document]
nojnhuh: SGTM

test/e2e/aks_azure_cluster_autoscaler.go
```diff
@@ -0,0 +1,120 @@
...
+	amcp.Spec.AutoScalerProfile = nil
+	Expect(mgmtClient.Update(ctx, amcp)).To(Succeed())
+	Eventually(func(g Gomega) {
+		amcp := &infrav1exp.AzureManagedControlPlane{}
+		err = mgmtClient.Get(ctx, types.NamespacedName{
+			Namespace: input.Cluster.Spec.ControlPlaneRef.Namespace,
+			Name:      input.Cluster.Spec.ControlPlaneRef.Name,
+		}, amcp)
+		Expect(err).NotTo(HaveOccurred())
+		g.Expect(amcp.Spec.AutoScalerProfile).To(BeNil())
```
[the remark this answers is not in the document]
jackfrancis: I actually think we'd want to test both ideally (did CAPZ resource update as expected? did AKS outcomes occur as expected?)

But the AKS outcome test as you point out is a little more complicated. I'm fine w/ punting the AKS outcome test. Wouldn't object to a // TODO comment after this test that reminds us of that.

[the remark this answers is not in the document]
jackfrancis: In lieu of a TODO: https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/3004

PR #2845, opened 2022-11-22
ci-build-kubernetes fixes

jackfrancis:
**What type of PR is this?**

/kind failing-test

This PR brings two additional changes to the `ci-build-kubernetes.sh` script that capz uses to accommodate building clusters from kubernetes source.

1) For increased macos portability, we now use `ggrep` to allow us to continue using the "perl regex" option for these existing patterns on macos. The version of `grep` that ships w/ macos doesn't have "perl regex" support, so this change will require macos devs to `brew install grep` (to be clear, this isn't a new requirement: macos doesn't work at all without these changes).
2) Due to the recent migration to `registry.k8s.io` in upstream Kubernetes, depending on the k/k branch that you build from,  built artifact data may have multiple images per source (one for `k8s.gcr.io` and one for `registry.k8s.io`). We can choose any one of the two, but we have to choose one. This change just uses the first image tag (via `head -n 1`).

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/scripts/ci-build-kubernetes.sh
+++ b/scripts/ci-build-kubernetes.sh
@@ -41,6 +41,11 @@ declare -a BINARIES=("kubeadm" "kubectl" "kubelet")
 declare -a WINDOWS_BINARIES=("kubeadm" "kubectl" "kubelet" "kube-proxy")
 declare -a IMAGES=("kube-apiserver" "kube-controller-manager" "kube-proxy" "kube-scheduler")
 
+GREP_BINARY="grep"
+if [[ "${OSTYPE}" == "darwin"* ]]; then
+  GREP_BINARY="ggrep"
+fi
+
 setup() {
     KUBE_ROOT="$(go env GOPATH)/src/k8s.io/kubernetes"
     export KUBE_ROOT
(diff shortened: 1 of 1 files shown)


scripts/ci-build-kubernetes.sh
```diff
@@ -92,7 +97,7 @@ main() {
 
         for IMAGE_NAME in "${IMAGES[@]}"; do
             # extract docker image URL form `docker load` output
-            OLD_IMAGE_URL="$(docker load --input "${KUBE_ROOT}/_output/release-images/amd64/${IMAGE_NAME}.tar" | grep -oP '(?<=Loaded image: )[^ ]*')"
+            OLD_IMAGE_URL="$(docker load --input "${KUBE_ROOT}/_output/release-images/amd64/${IMAGE_NAME}.tar" | ${GREP_BINARY} -oP '(?<=Loaded image: )[^ ]*' | head -n 1)"
```
CecileRobertMichon, line 100: in your comment here: https://github.com/kubernetes/kubernetes/pull/113395#issuecomment-1324029704

it seems that registry.k8s.io is the second image returned in the list not the first, does this mean we're going to still use gcr.io for those cases? If that's the case should we instead grep for registry.k8s.io so we select the right one? Not sure if it matters in practice

jackfrancis: It shouldn't matter so long as our `docker load` and `docker tag` commands refer to the same image.

Ultimately what we're doing here is re-tagging the image that's a part of the k/k build process, and uploading it to our own registry.

Of course we want to monitor upstream k8s evolution in case there are actually multiple, _different_ container images inside a single tar (definitely seems that should never happen).

`head -n 1` seems about the simplest way of doing this. In the case that there's a bunch of differently named images, it'll just pick one (the first, obviously), and in the case where there's only one image, it's essentially a no-op.

CecileRobertMichon (commented):
/lgtm
/approve

PR #2861, opened 2022-11-29
add e2e AKS upgrade test

nojnhuh:
**What type of PR is this?**
/kind cleanup

**What this PR does / why we need it**: This PR adds a new e2e test that performs a Kubernetes upgrade from 1.23 to 1.24 for an AKS cluster.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```


test/e2e/aks.go
```diff
@@ -223,8 +223,8 @@ func WaitForAKSSystemNodePoolMachinesToExist(ctx context.Context, input WaitForC
 }
 
 // GetAKSKubernetesVersion gets the kubernetes version for AKS clusters.
-func GetAKSKubernetesVersion(ctx context.Context, e2eConfig *clusterctl.E2EConfig) (string, error) {
-	e2eAKSVersion := e2eConfig.GetVariable(AKSKubernetesVersion)
+func GetAKSKubernetesVersion(ctx context.Context, e2eConfig *clusterctl.E2EConfig, varName string) (string, error) {
```
willie-yao, line 226: Would it be better to rename `varName` to something more descriptive like `version`, unless the variable needs to be more generic?

willie-yao (commented):
Just one minor nit from my end. Great work!

test/e2e/aks.go
nojnhuh: Yeah this could definitely be more clear. Since this does represent literally the name of a variable, would something like `versionVar` be a bit better? I'll update the doc comment here too.

willie-yao: Yup I think that would be perfect!

test/e2e/azure_test.go
```diff
@@ -617,6 +617,50 @@ var _ = Describe("Workload cluster creation", func() {
 				})
 			})
 		})
+
+		It("upgrades to a new Kubernetes version", func() {
```
jackfrancis, line 621: Is this an entirely new cluster test (build + upgrade)?

Are we not able to add the `By("Upgrading the Kubernetes version of the cluster", func() {` somewhere in the existing AKS E2E flow? Or are we simply restoring the original cluster state by re-applying the aks flavor spec?

nojnhuh: Yes, I made this a separate newly-built cluster to help ensure this doesn't affect the other tests and since the CAPI upgrade tests seem to be set up similarly. AFAICT AKS doesn't allow downgrades, even between patch versions, so I don't think this is something we can reverse easily. I suppose it wouldn't be disruptive to any of the existing tests as long as this is done last.

CecileRobertMichon: I do like the idea of this being a test spec we can run on any existing cluster configuration similar to what you did with the IP prefix spec, so we don't build one net new cluster for each feature we test, and we can run the same feature test across multiple cluster flavors/shapes. FWIW the self-managed cluster tests weren't structured this way but don't think they should be used as model, we should refactor them to do the same.

CecileRobertMichon: That being said it doesn't prevent us from skipping the upgrade "spec" depending on the job, for example if we find it too lengthy/costly to run on every PR test

jackfrancis: +1

The main concern we have to have here is that we can't move the "k8s version to build a cluster with" too quickly, to ensure we always have overhead to upgrade to.

And because of that we might consider generalizing any k8s functional tests we do against the clusters so that we can get some kind of "this basic stuff works w/ 1.23 *and* 1.24" signal. Not sure if any of our current tests are sensitive to the specific Kubernetes version, so maybe just a thought going forward.

nojnhuh: > The main concern we have to have here is that we can't move the "k8s version to build a cluster with" too quickly, to ensure we always have overhead to upgrade to.

Since we use "latest" for the Kubernetes version, this is already an issue if we want to run this test in the same cluster. Is it acceptable to move the test down to 1.23 from 1.24 to facilitate this change?

jackfrancis: (IMO) yes, that's O.K., if we move the upgrade flow to happen immediatly after that "latest -1" cluster is built

nojnhuh: Makes sense, just pushed that change.

test/e2e/config/azure-dev.yaml
```diff
@@ -148,6 +148,7 @@ providers:
 
 variables:
   AKS_KUBERNETES_VERSION: "latest"
+  AKS_KUBERNETES_VERSION_UPGRADE_FROM: "${AKS_KUBERNETES_VERSION_UPGRADE_FROM:-stable-1.23}"
```
jackfrancis, line 151: I think this is fine for now, but FYI we'll have to monitor when the mutable "latest" alias gets rev'd by AKS, as that will require us to rev our "from" version (I'm pretty sure AKS will not let you upgrade > 1 version ahead)

jackfrancis (approved):
/lgtm
/approve

test/e2e/config/azure-dev.yaml
CecileRobertMichon: I do have some concerns with this as it can allow our test to break due to AKS rolling out a new version (as versions become unsupported). This is especially an issue for maintaining tests on older cherry-pick branches where our tests will naturally break over time.

/hold

CecileRobertMichon: something more resilient (just off the top of my head) would be to maybe determine version to upgrade from as latest -1 (whether that's -1 minor or patch)

jackfrancis: That should work (subtract 1 from latest minor, that is). Using patch versions for testing upgrade can be tricky, as sometimes AKS will accelerate the deprecation of a patch version and leave a particular minor release (e.g., 1.25) with only one patch version available. (And there's also the case of a ".0" release, which will be the only patch available on that minor until the ".1" is released.)

CecileRobertMichon: For sure, I just meant take the latest available version at index len() -1 in the array of supported versions by AKS.

CecileRobertMichon:
/hold cancel
/assign @jackfrancis 

lgtm

CecileRobertMichon:
@nojnhuh please squash

nojnhuh:
> @nojnhuh please squash

Squashed!

PR #2862, opened 2022-11-29
WIP: Make resourceID required for AzureClusterIdentity when using UserAssignedMSI

### LINKED ISSUE #2832: Make resourceID required for AzureClusterIdentity when using UserAssignedMSI
/kind feature

When trying to authenticate a cluster with user managed identity, the resource ID field for the AzureClusterIdentity needs to be specified. Currently, the field is not required.

[Link to Slack thread about the issue](https://kubernetes.slack.com/archives/CEX9HENG7/p1668717810408669)

**Environment:**

willie-yao:
**What type of PR is this?**
/kind feature

When trying to authenticate a cluster with UserAssignedMSI, the resource ID field for the `AzureClusterIdentity` needs to be specified. This PR adds a validating webhook that requires the resourceID to be specified for UserAssignedMSI.

Fixes #2832 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Make resourceID required for AzureClusterIdentity when using UserAssignedMSI
```


api/v1beta1/azureclusteridentity_webhook.go
```diff
@@ -26,3 +32,30 @@ func (c *AzureClusterIdentity) SetupWebhookWithManager(mgr ctrl.Manager) error {
 		For(c).
 		Complete()
 }
+
+// +kubebuilder:webhook:verbs=create;update,path=/validate-infrastructure-cluster-x-k8s-io-v1beta1-azureclusteridentity,mutating=false,failurePolicy=fail,matchPolicy=Equivalent,groups=infrastructure.cluster.x-k8s.io,resources=azureclusteridentities,versions=v1beta1,name=validation.azureclusteridentity.infrastructure.cluster.x-k8s.io,sideEffects=None,admissionReviewVersions=v1;v1beta1
+
```
CecileRobertMichon, line 37: extra newline?

api/v1beta1/azureclusteridentity_webhook.go
```diff
@@ -26,3 +32,30 @@ func (c *AzureClusterIdentity) SetupWebhookWithManager(mgr ctrl.Manager) error {
...
+var _ webhook.Validator = &AzureClusterIdentity{}
+
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type.
+func (c *AzureClusterIdentity) ValidateCreate() error {
+	var allErrs field.ErrorList
+	if c.Spec.Type == UserAssignedMSI && c.Spec.ResourceID == "" {
+		err := errors.New("resourceID must be set when type is UserAssignedMSI")
+		allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "resourceID"), c.Spec.ResourceID, err.Error()))
```
CecileRobertMichon, line 45: Here instead of the generic `field.Invalid` error you can use `field.Required`

api/v1beta1/azureclusteridentity_webhook.go
```diff
@@ -26,3 +32,30 @@ func (c *AzureClusterIdentity) SetupWebhookWithManager(mgr ctrl.Manager) error {
...
 }
+
+// +kubebuilder:webhook:verbs=create;update,path=/validate-infrastructure-cluster-x-k8s-io-v1beta1-azureclusteridentity,mutating=false,failurePolicy=fail,matchPolicy=Equivalent,groups=infrastructure.cluster.x-k8s.io,resources=azureclusteridentities,versions=v1beta1,name=validation.azureclusteridentity.infrastructure.cluster.x-k8s.io,sideEffects=None,admissionReviewVersions=v1;v1beta1
+
+var _ webhook.Validator = &AzureClusterIdentity{}
+
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type.
+func (c *AzureClusterIdentity) ValidateCreate() error {
+	var allErrs field.ErrorList
+	if c.Spec.Type == UserAssignedMSI && c.Spec.ResourceID == "" {
```
CecileRobertMichon, line 43: should we also validate the inverse, ie. that resourceID is empty when the Type is not UserAssigned?

api/v1beta1/azureclusteridentity_webhook.go
willie-yao: I saw that other webhook files had the kubebuilder command separated in its own block like [this](https://github.com/willie-yao/cluster-api-provider-azure/blob/7db161723f1aaadb4d79f211c35e788139ece18f/api/v1beta1/azureclustertemplate_webhook.go#L40). Would this hold true for this command as well?

api/v1beta1/azureclusteridentity_webhook.go
willie-yao: Yes that would make sense! I didn't realize resourceID was unique to UserAssignedMSI

willie-yao: Would the error be a field.Invalid or something more specific? I was thinking field.NotSupported may fit, but didn't find anything specific to an extra field.

CecileRobertMichon: I think `Field.Forbidden`  might be most appropriate for that one:

``` 
// Forbidden returns a *Error indicating "forbidden".  This is used to
// report valid (as per formatting rules) values which would be accepted under
// some conditions, but which are not permitted by current conditions (e.g.
// security policy).
```

api/v1beta1/azureclusteridentity_webhook.go
CecileRobertMichon: true, pls ignore my comment

CecileRobertMichon (commented):
/lgtm
/assign @nawazkh

mboersma (approved):
/approve
/cherrypick release-1.6
/cherrypick release-1.5

mboersma:
Do others think this is a bug fix that should be cherry-picked?

This is categorized as a feature, but seemed to me like a bug fix. So I reflexively cherry-picked it into the supported branches.

jackfrancis:
@mboersma I agree

@willie-yao can your remove the `kind/feature` and replace it w/ `kind/bug`?

willie-yao:
@jackfrancis @mboersma Done! Yup this is definitely a bug, since the capz controller continuously crashes when the resourceID is not specified

api/v1beta1/azureclusteridentity_webhook.go
```diff
@@ -26,3 +30,31 @@ func (c *AzureClusterIdentity) SetupWebhookWithManager(mgr ctrl.Manager) error {
...
+}
+
+// ValidateDelete implements webhook.Validator so a webhook will be registered for the type.
+func (c *AzureClusterIdentity) ValidateDelete() error {
+	return nil
+}
+
+// ValidateUpdate implements webhook.Validator so a webhook will be registered for the type.
+func (c *AzureClusterIdentity) ValidateUpdate(oldRaw runtime.Object) error {
+	return nil
```
nawazkh, line 59: What happens when the user tries to update `AzureClusterIdentity`? 
Is it valid if the user updates `AzureClusterIdentity.Spec.ResourceID` to `""` after creating the resource? I.e. after creating a resource with `AzureClusterIdentity.Spec.ResourceID == UserAssignedMSI` and `AzureClusterIdentity.Spec.ResourceID !== ""`

api/v1beta1/azureclusteridentity_webhook.go
```diff
@@ -26,3 +30,31 @@ func (c *AzureClusterIdentity) SetupWebhookWithManager(mgr ctrl.Manager) error {
 		For(c).
 		Complete()
 }
+
+// +kubebuilder:webhook:verbs=create;update,path=/validate-infrastructure-cluster-x-k8s-io-v1beta1-azureclusteridentity,mutating=false,failurePolicy=fail,matchPolicy=Equivalent,groups=infrastructure.cluster.x-k8s.io,resources=azureclusteridentities,versions=v1beta1,name=validation.azureclusteridentity.infrastructure.cluster.x-k8s.io,sideEffects=None,admissionReviewVersions=v1;v1beta1
```
nawazkh, line 34: I am not sure but do we want to restrict the `webhook:verbs` to just `create` because we do not have any logic for `ValidateUpdate()`?

nawazkh: I am not sure if a Kubernetes API gives out an `UPDATE` call while creating the resource(i.e. at the time of object creation). Or to put it other way, I am not sure if Kubernetes API's UPDATE calls are really needed here for `ValidateCreate()` calls.

nawazkh (commented):
I am curious about some scenarios, sharing some questions!

api/v1beta1/azureclusteridentity_webhook.go
```diff
@@ -26,3 +30,31 @@ func (c *AzureClusterIdentity) SetupWebhookWithManager(mgr ctrl.Manager) error {
...
+
```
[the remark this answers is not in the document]
willie-yao: There were no existing default webhooks for the AzureClusterIdentity, so I thought they weren't necessary to implement here. I'm not exactly sure where the defaults are set. Will take a closer look now.

api/v1beta1/azureclusteridentity_webhook.go
willie-yao: The update webhook is there so the AzureClusterIdentity webhook can fully implement `webhook.Validator`. The update is a no-op right now because we were only concerned with blocking the creating of an AzureClusterIdentity. I'll try and observe the update behavior and make sure it isn't doing something unintended.

api/v1beta1/azureclusteridentity_webhook.go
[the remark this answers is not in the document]
CecileRobertMichon: None of the fields can be defaulted since they all pertain to users' credentials and there is no valid "default" value, they are unique for each user

api/v1beta1/azureclusteridentity_webhook.go
nawazkh: Oh sorry, I was pointing to the `update` verb in the `+kubebuilder:webhook:verbs=create;update` marker and not to the implementation of `ValidateUpdate()` function. 

I found the following upon investigation (referring from [Kubebuilder webhook page](https://book.kubebuilder.io/reference/markers/webhook.html))
```
verbs=<[]>string
specifies the Kubernetes API verbs that this webhook receives requests for.
Only modification-like verbs may be specified. May be “create“, “update“, “delete“, “connect“, or “*“ (for all).
```


I am not sure if the `verbs=update` is required here in the marker.

willie-yao: Ah good catch! That should be removed, I'll make a change now.

api/v1beta1/azureclusteridentity_webhook.go
CecileRobertMichon: Good question @nawazkh. @willie-yao the validate webhook runs every time an update run after initial create, could be for validating that the field doesn't change when it's immutable but it could also be for enforcing the same validity requirements. In the latter case, we usually share a common validate function between the two, like here https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/api/v1beta1/azurecluster_webhook.go#L135.

I think in this case we don't want the field to be immutable (user might want to rotate identities after cluster is created) but we do want to enforce the same rules that `ResourceID ` is not empty when type is `UserAssignedMSI `.

nawazkh (approved):
/lgtm 
🚀

PR #2879, opened 2022-11-30
fix bad generated template name on Mac

nojnhuh:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**: This PR fixes an issue where `make generate-flavors` on macOS would result in a new file named `templates/test/ci/cluster-template-{}.yaml` instead of a name like `templates/test/ci/cluster-template-prow-azurediskcsi-migration-off.yaml`.

For background, one side effect of #2523 was that it made `$REPO_ROOT` longer where it gets expanded in [this command in gen-flavors.sh](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/9e97aea0e2c21ad4d140afe31e4cd2d695a71112/hack/gen-flavors.sh#L35), adding an extra `../hack/`. This was taking the command past the 255-character limit set by the Mac version of `xargs` on the length of arguments containing the replacement string defined by `-I`, so the second `{}` wouldn't be replaced as expected. The [man page](https://ss64.com/osx/xargs.html) describes this behavior on the `-I` flag. My local man page also shows a `-S` flag to change this limit, but no such flag exists on Ubuntu since I don't think it has the limit to begin with.

This change removes the `$REPO_ROOT` definition from common-vars.sh since all scripts `source`-ing that already define it themselves and its value is more concise in that case.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/common-vars.sh
+++ b/hack/common-vars.sh
@@ -13,7 +13,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-REPO_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
+if [[ -z "$REPO_ROOT" ]]; then
+  echo >&2 "REPO_ROOT must be set"
+  exit 1
+fi
+
 # shellcheck disable=SC2034
 KUBECTL="${REPO_ROOT}/hack/tools/bin/kubectl"
 # shellcheck disable=SC2034
--- a/hack/gen-flavors.sh
+++ b/hack/gen-flavors.sh
@@ -28,6 +28,15 @@ flavors_dir="${REPO_ROOT}/templates/flavors/"
 ci_dir="${REPO_ROOT}/templates/test/ci/"
 dev_dir="${REPO_ROOT}/templates/test/dev/"
 
+# NOTE: On macOS, the kustomize commands run by xargs that use the -I
+# replacement string may only be 255 characters. Commands longer than that will
+# not expand the replacement string when that would go over that limit,
+# resulting in generated manifests like "templates/test/ci/cluster-template-{}".
+#
+# Currently, the longest command is *exactly* 255 characters, so making these
+# commands any longer or adding a new template with a longer name will run into
+# this problem.
+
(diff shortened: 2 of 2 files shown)


hack/common-vars.sh
```diff
@@ -13,7 +13,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-REPO_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
 # shellcheck disable=SC2034
 KUBECTL="${REPO_ROOT}/hack/tools/bin/kubectl"
```
mtougeron, line 17: What about these variables that assume `REPO_ROOT` is set?

nojnhuh: A check to verify that assumption seems worthwhile. The pattern right now seems to be setting `$REPO_ROOT` in each script and using that to source `common-vars.sh`, so it should always be set here.

mtougeron: nevermind, see the comment in the ticket description.

nojnhuh: Added a check since it's entirely possible future changes could source common-vars.sh without having already defined `$REPO_ROOT`.

hack/gen-flavors.sh
```diff
@@ -28,6 +28,15 @@ flavors_dir="${REPO_ROOT}/templates/flavors/"
 ci_dir="${REPO_ROOT}/templates/test/ci/"
 dev_dir="${REPO_ROOT}/templates/test/dev/"
 
+# NOTE: On macOS, the kustomize commands run by xargs that use the -I
```
nojnhuh, line 31: I added this comment mostly to be breadcrumbs for the next time we hit this since I think it might be difficult to track down this PR otherwise.

jackfrancis (approved):
/lgtm

CecileRobertMichon:
/lgtm

Great work @nojnhuh 🙇 

Do we need to cherry-pick this (ie. was the original commit that regressed this part of the last release?)

nawazkh:
Great find @nojnhuh ! Kudos!

nojnhuh:
> Do we need to cherry-pick this (ie. was the original commit that regressed this part of the last release?)

No, v1.6.0 was cut before this PR landed so that change isn't on the release branch.

PR #2889, opened 2022-12-01
replace xargs with for loop in gen-flavors.sh

nojnhuh:
**What type of PR is this?**
/kind cleanup

**What this PR does / why we need it**: As a follow-up to #2879, this PR removes the use of xargs for constructing and running the kustomize commands used to generate the flavor templates. This will let the script be more flexible without having to worry about the default 255-character limit when using replacement strings in xargs on macOS. The behavior with this change should stay the same.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/gen-flavors.sh
+++ b/hack/gen-flavors.sh
@@ -28,18 +28,16 @@ flavors_dir="${REPO_ROOT}/templates/flavors/"
 ci_dir="${REPO_ROOT}/templates/test/ci/"
 dev_dir="${REPO_ROOT}/templates/test/dev/"
 
-# NOTE: On macOS, the kustomize commands run by xargs that use the -I
-# replacement string may only be 255 characters. Commands longer than that will
-# not expand the replacement string when that would go over that limit,
-# resulting in generated manifests like "templates/test/ci/cluster-template-{}".
-#
-# Currently, the longest command is *exactly* 255 characters, so making these
-# commands any longer or adding a new template with a longer name will run into
-# this problem.
-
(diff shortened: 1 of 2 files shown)


hack/gen-flavors.sh
```diff
@@ -28,18 +28,16 @@ flavors_dir="${REPO_ROOT}/templates/flavors/"
...
-# replacement string may only be 255 characters. Commands longer than that will
-# not expand the replacement string when that would go over that limit,
-# resulting in generated manifests like "templates/test/ci/cluster-template-{}".
-#
-# Currently, the longest command is *exactly* 255 characters, so making these
-# commands any longer or adding a new template with a longer name will run into
-# this problem.
-
-find "${flavors_dir}"* -maxdepth 0 -type d -print0 | xargs -0 -I {} basename {} | grep -v base | xargs -I {} sh -c "${KUSTOMIZE} build --load-restrictor LoadRestrictionsNone --reorder none ${flavors_dir}{} > ${REPO_ROOT}/templates/cluster-template-{}.yaml"
+for name in $(find "${flavors_dir}"* -maxdepth 0 -type d -print0 | xargs -0 -I {} basename {} | grep -v base); do
```
jackfrancis, line 31: Would it be worth having a nested for loop?

1. enumerate through dirnames
2. enumerate through find output it each dirname

nojnhuh: I see that there's quite a bit of boilerplate here, but I think a loop would be slightly complicated considering the "flavors" templates are rendered to a different relative directory and use a different `grep -v` than the "ci" and "dev" templates. Certainly doable, but I'll keep an eye out for other ways of consolidating some of this too.

jackfrancis: Fair enough!

jackfrancis:
/lgtm
/approve

Thanks for solving this before we drop a 256 character filepath with our next test template addition! :)

Issue #629, opened 2020-05-18
Flatcar support for CAPZ


rbitia:
/kind feature

**Describe the solution you'd like**
Support users that have a dependency on CoreOS by supporting Flatcar containers in CAPZ.

detiber:
I suspect this will need additional support from the kubeadm bootstrap provider and image-builder support as well (https://github.com/kubernetes-sigs/cluster-api/issues/1582), since Flatcar is managed quite a bit differently than the currently supported OSs.

CecileRobertMichon:
https://github.com/kubernetes-sigs/image-builder/pull/248

CecileRobertMichon:
I'm going to close this one since it doesn't require any changes to the infra provider AFAIK and it's already being tackled from an image perspective in kubernetes-sigs/image-builder#248 and for bootstrapping in https://github.com/kubernetes-sigs/cluster-api/pull/3437

/close

vbatts:
@CecileRobertMichon I think it may require infra-provider updates, it just that the image-builder work has been a huge thing for review and feedback.

vbatts:
I guess I was mis-remembering. Looking through the code it doesn't seem to have particular things jump out as blockers. There is a good chance we need a change to incorporate the ignition-bootstrap-provider, rather than just kubeadm-bootstrap-provider.
The big this is for us to work through the providers and get testing on this.

CecileRobertMichon:
reopening to track testing / integration

/remove-lifecycle rotten
/reopen

vbatts:
@CecileRobertMichon nice. This is on our roadmap and I'm looking forward to making it happen.

vbatts:
cc @invidian

invidian:
/remove-lifecycle rotten

Small update. We've got image-builder to build CAPI Azure images, but there is still some thing which needs to be resolved before we open the PR.

Changes proposed in https://github.com/kubernetes-sigs/cluster-api/pull/4172 works out of the box with CAPZ, except new cluster configuration template needs to be submitted including some minor differences in node configuration. Mentioned here: https://github.com/kubernetes-sigs/cluster-api/pull/4172#issuecomment-802771490.

invidian:
CABPK PR kubernetes-sigs/cluster-api#4172 has been merged and will be part of v1.1 release, so we're getting closer getting this resolved :tada:

mboersma:
[...]

PR #2890, opened 2022-12-01
Add Flatcar Container Linux template

invidian:
**What type of PR is this?**
/kind feature

This PR adds a cluster template for Flatcar Container Linux based clusters.

Reopening #1729.

Related to https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/629

CC @jepio

For more context on why #1729 hasn't been merged for so long, we were waiting until https://github.com/flatcar/Flatcar/issues/734 gets resolved, so we have a Azure Community Image Gallery with testing reference Flatcar images which can be reliably used in CAPZ e2e tests. We have now created the gallery which is used in this PR (`flatcar4capi-742ef0cb-dcaa-4ecb-9cb0-bfd2e43dccc0`) and we will provide there at least images required for keeping the CI green here. Over time, we plan to automate the publishing process so we will provide images for Flatcar and Kubernetes releases.

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Flatcar Container Linux is now supported as a cluster flavor. Try `clusterctl generate cluster my-cluster --flavor flatcar`.
```


invidian:

Now the workaround for https://github.com/kubernetes-sigs/cluster-api/issues/7679 should work properly.

invidian:
Creating cluster with the template works fine for me locally, so I'm debugging e2e tests locally to figure out what's wrong.

templates/test/ci/prow-flatcar/cni-resource-set.yaml
```diff
@@ -0,0 +1,14 @@
+---
+apiVersion: addons.cluster.x-k8s.io/v1beta1
+kind: ClusterResourceSet
+metadata:
+  name: ${CLUSTER_NAME}-calico
```
CecileRobertMichon, line 5: we recently removed all ClusterResourceSets to install calico (see #2495). So here's what you need to do:

1. Rebase on top of latest main
2. Get rid of all the Calico stuff in the template/kustomization
3. Add 
``` 
				ControlPlaneWaiters: clusterctl.ControlPlaneWaiters{
					WaitForControlPlaneInitialized: EnsureControlPlaneInitialized,
				},
````

in azure_test.go in the `ApplyClusterTemplateAndWaitInput` (see https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2890/files#diff-93ce65d81f5c1510135f6035d43f678bf9b5962b8e4ee45105e3345a1852efc6L321 for an example)

**Comment by invidian** [comment]

Updated to not validate Calico stuff in e2e tests.

File: templates/test/ci/prow-flatcar/cni-resource-set.yaml (hunk 1 above)
Comment (invidian) [reply templates/test/ci/prow-flatcar/cni-resource-set.yaml]: Thanks again for the hint. I was suspecting something has changed but I did not investigate it further. Changes are done. Tests seems to be passing locally. Please have a look :)

**Comment by CecileRobertMichon** [comment]
both failures seem like flakes, can you please squash before we rerun?

**Comment by invidian** [comment]
Odd, Flatcar test now has failed with an I/O timeout... Restarted it. https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/2890/pull-cluster-api-provider-azure-e2e-optional/1598781490357866496

File: templates/cluster-template-flatcar.yaml [hunk 2]
```diff
@@ -0,0 +1,265 @@
...
+  template:
+    spec:
+      dataDisks:
+      - diskSizeGB: 256
+        lun: 0
+        nameSuffix: etcddisk
+      image:
+        computeGallery:
+          gallery: flatcar4capi-742ef0cb-dcaa-4ecb-9cb0-bfd2e43dccc0
+          name: flatcar-stable-amd64-capi-${KUBERNETES_VERSION}
```
CecileRobertMichon, line 160: will there be a new image published with each k8s version? if so, what is the expected delay betwen the k8s version being released and the image being available? Since our e2e test is wired to use the latest stable version available in the Azure VM image gallery, this could cause a race where the image is available for Ubuntu but not Flatcar yet and block PRs / fail periodic jobs.

templates/cluster-template-flatcar.yaml
```diff
@@ -0,0 +1,265 @@
...
+          version: 3374.2.0
```
CecileRobertMichon, line 161: why are we hardcoding a specific version? is the version guaranteed to be the same for every image / k8s version going forward?

templates/cluster-template-flatcar.yaml
```diff
@@ -0,0 +1,265 @@
...
+spec:
+  template:
+    spec:
+      dataDisks:
+      - diskSizeGB: 256
+        lun: 0
+        nameSuffix: etcddisk
+      image:
+        computeGallery:
+          gallery: flatcar4capi-742ef0cb-dcaa-4ecb-9cb0-bfd2e43dccc0
```
CecileRobertMichon, line 159: would be good to add some docs around this and specifically the image. Are users meant to use this in production or is it just provided as an example? Is the image patched for security vulnerabilities?

templates/cluster-template-flatcar.yaml
invidian: As mentioned in PR description, we plan to at least fulfill whatever is needed for CI here. Could you elaborate on the wiring? How are images published into Azure VM image gallery right now? Are they being published right after the new k8s release? I thought the k8s version used in e2e tests is periodically bumped by hand to be honest. CC @t-lo

templates/cluster-template-flatcar.yaml
invidian: So, for end users, we recommend explicitly selecting the version of Flatcar to use, as changing the version in machine deployment should trigger machines replacement and perform a cluster update.

For CI here, my assumption was that when KUBERNETES_VERSION used by CI changes, this will probably change as well, as we likely want to build latest K8s version images with latest Flatcar stable, but I _think_ we could patch it to just use `latest`, although we do plan to exclude all CAPI versions from latest so they must be selected explicitly.

templates/cluster-template-flatcar.yaml
invidian: Right. The images will have initially no guarantees as other reference CAPI images, they are only for testing/demo purposes. We might consider changing that over time though.

We will work on the docs for it as well. If I knew they would be needed for this PR, I'd wait with opening it until it's done :smile:

templates/cluster-template-flatcar.yaml
CecileRobertMichon: The E2E config specifies the k8s version as `latest-major.minor` https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/test/e2e/config/azure-dev.yaml#L151, which then gets resolve to the latest k8s version that has an available Windows _and_ Ubuntu image available: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/test/e2e/helpers.go#L713 in the Azure Marketplace.

I misspoke when I said VM image gallery earlier, I meant Azure Marketplace. We are planning on migrating the test image to a public Shared Image Gallery but currently, they are still in the Marketplace. The images are published by pipelines after each new k8s release but that process is still kicked off by hand (@mboersma can add more details on the publishing process if you have questions about that). 

We could change the logic to look for the latest Ubunut, Flatcar and Windows image available but that might slow down the rate at which we can test new versions if the Flatcar image publishing process is ad hoc / manual, which is why I asked.

templates/cluster-template-flatcar.yaml
CecileRobertMichon: As soon as we expose something to users (in this case via the new flavor that will get published as a release asset) there should be docs for it. Otherwise, users might use it in a way that's unintended. We do have a documentation box on the PR description TODOs and it's documented in https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/CONTRIBUTING.md#contributing-a-patch

> All new features should come with user documentation.

templates/cluster-template-flatcar.yaml
invidian: Probably @t-lo can add more to this and/or confirm, but if the requirement is to provide images for `latest-major.minor` K8s versions and building is still done by hand right now, I think Flatcar team can definitely commit to deliver those images. The building process is documented on our side already and we plan to automate it in the future.

Also, once all images used by CI are migrated to a public SIG, it shouldn't be a problem to build Flatcar images there as well then. We plan to upstream all required patches to image-builder and provide documentation on the building process.

templates/cluster-template-flatcar.yaml
invidian: Hmm, do you have some suggestion where it should be documented? I'm looking at https://capz.sigs.k8s.io/topics/custom-images.html and https://capz.sigs.k8s.io/introduction.html?highlight=flavor#flavors but I don't immediately see any good place to put such docs. I'd assume this images should be treated no different from other images used by CAPZ, except that they will be delivered by a different team for the time being, until CAPZ itself migrate to use SIG images.

CecileRobertMichon: We could start a new "Flatcar" topic page with an image section just like Windows https://capz.sigs.k8s.io/topics/windows.html#image-creation

invidian: Sounds good, thanks. I'll prepare some draft next week.

templates/cluster-template-flatcar.yaml
t-lo: Thanks for the ping. We're currently aiming to integrate gallery publishing into our release process, and use the latest Kubernetes version at the time of a release. We usually release new images every 2-3 weeks. We'll start doing this manually but plan to eventually have it 100% automated (in fact, we're preferring a public image gallery over marketplace offers because publishing to a gallery is easily automatable).

While we don't have full automation in place it's hard to commit to immediate image releases tough when a new Kubernetes version becomes available since that would put additional load on the Flatcar OS maintainers team. This is transient, and because of our 2-3 week release cycle we'd be catching up soon enough with a new Kubernetes release. Is there a way to bypass running Flatcar tests in the CI when a respective image is not yet available?

t-lo: (also, CC: our release engineer @sayanchowdhury for visibility)

CecileRobertMichon: > Is there a way to bypass running Flatcar tests in the CI when a respective image is not yet available

There is, but I'm not sure I agree that's something we should do since that could cause us to not have any signal on Flatcar for weeks without knowing it (test would show as "green" and I doubt people would check the output to see if the test ran). An alternative would be to add a separate k8s version for Flatcar and just have the Flatcar test run on its own latest available k8s version for that major.minor for Flatcar (we could add the func to discover what version that is in this PR).

@mboersma @jackfrancis @jsturtevant thoughts?

jackfrancis: 100% agree w/ shipping a test that uses a different version of k8s for Flatcar compared to Ubuntu. This sort of makes sense, as our k8s version velocity at present is intrinsically related to the progress of Ubuntu VHD images that we maintain. Not surprising that another image publication pipeline would have a different k8s version velocity.

invidian: Should I then implement a discovery for available K8s version for Flatcar images?

templates/cluster-template-flatcar.yaml
invidian: An alternative would be to introduce `FLATCAR_VERSION` and maybe `FLATCAR_CHANNEL` variables?

templates/cluster-template-flatcar.yaml
invidian: Added some initial docs so we can have something to comment on. Please have a look.

jackfrancis:
@invidian quick reminder that the 1.7 release is scheduled for next week, in case this is near ready and we want to include it in that release!

invidian:
@jackfrancis thanks for the ping, but from my side, this is ready to ship (since long time) :smile:

So please let me know how should I proceed with remaining unresolved conversations or what else is missing for this to be merged and I will work on it to get it done for next week. I didn't poke you or joined community meeting to mention this given Christmas break. And I was actually planning to do so next week :)

templates/cluster-template-flatcar.yaml
jackfrancis: I think that's right. You could define a new env var `KUBERNETES_VERSION_FLATCAR` with its own discovery implementation, and then just refer to that in the flatcar templates. That might be easier than overloading the existing `KUBERNETES_VERSION` in the existing discovery flow (we'd need to add a condition "is this for flatcar or no?", which might be more of a headache than simply discovering the latest flatcar-available version during each E2E run).

Thoughts @CecileRobertMichon @mboersma ?

invidian:
/retest

I missed canada central region for images replication.

templates/cluster-template-flatcar.yaml
invidian: I ended up introducing `FLATCAR_VERSION` variable, which in e2e tests is automatically discovered for available K8s version. I'm omitting `FLATCAR_CHANNEL` for the time being, as we plan to only publish `stable` images for the time being to avoid large number of images being published.

I'll mark it as resolved, as I think it's also related to version discovery discussion we have in another thread.

templates/cluster-template-flatcar.yaml
invidian: I've pushed initial implementation which I managed to get to work yesterday. I've tried to make code look similar to what's already there, although the structure seem rather arbitrary and I'm not sure where would I start to add tests for this code. But since this is an e2e test framework, I guess it's OK. Let me know what you think.

So now `FLATCAR_KUBERNETES_VERSION` is independent from `KUBERNETES_VERSION` in e2e tests, although templates still use `KUBERNETES_VERSION` as users would expect.
`FLATCAR_KUBERNETES_VERSION` supports `stable-1.24` format like `KUBERNETES_VERSION` and resolves to latest patch version by listing Flatcar community gallery images.
Once Kubernetes version is selected, we pick up latest Flatcar version for given Kubernetes version.

mboersma:
I used this branch with `make tilt-up` and clicked the "flatcar" flavor, but eventually it ended with an error:

```shell
./hack/gen-flavors.sh
make[2]: Nothing to be done for `kustomize'.
sh: command substitution: line 266: syntax error near unexpected token `|'
sh: command substitution: line 266: `      | jq -r .compute.name'
```

invidian:

>I used this branch with make tilt-up and clicked the "flatcar" flavor, but eventually it ended with an error:

Odd. Anyway, I removed the newline from kustomization patch files, as it does not affect resulting templates anyway. BTW, perhaps it would be nice to test Tilt stuff on CI somehow, if it may affect the mergability of the PR?

Did you have a chance to look into version discovery part? Does it look all right?

test/e2e/config/azure-dev.yaml
```diff
@@ -154,6 +154,8 @@ variables:
   AKS_KUBERNETES_VERSION: "latest"
   AKS_KUBERNETES_VERSION_UPGRADE_FROM: "latest-1"
   KUBERNETES_VERSION: "${KUBERNETES_VERSION:-stable-1.24}"
+  FLATCAR_KUBERNETES_VERSION: "${FLATCAR_KUBERNETES_VERSION:-stable-1.24}"
```
mboersma, line 157: Could this default to `${KUBERNETES_VERSION}` or does `${FLATCAR_KUBERNETES_VERSION}` need to be updated separately?

mboersma:
The tmp version discovery commit lgtm.

mboersma:
> it would be nice to test Tilt stuff on CI somehow

So true, especially since if we know something actually breaks Tilt we wouldn't want to merge the PR.

It might be possible to have a CI job that runs `make tilt-up` although it would require docker-in-docker type access, and once you run `tilt up` or `tilt ci` it looks like the only interaction you can have with the running instance is `tilt logs`. Maybe that's enough, or maybe there's some API? We're wary of adding CI costs, too, so we'd want it to be optional.

test/e2e/config/azure-dev.yaml
invidian: It could default to it though, just may need to be adjusted in case we lag with the image as for Flatcar. Whatever you think it's better.

mboersma:
Somehow even though I see you changed it to a one-liner in machine-deployment.yaml and kubeadm-controlplane.yaml, the main flatcar template and the CI one still have the newline. I've run `./hack/gen-flavors.sh` and full `make generate` locally and there's no diff.

invidian:
> The tmp version discovery commit lgtm.

Thanks, I'll squash next time I push then.

>It might be possible to have a CI job that runs make tilt-up although it would require docker-in-docker type access, and once you run tilt up or tilt ci it looks like the only interaction you can have with the running instance is tilt logs. Maybe that's enough, or maybe there's some API? We're wary of adding CI costs, too, so we'd want it to be optional.

It seem creating an issue for this would be good I think :)

> @invidian somehow even though I see you changed it to a one-liner in machine-deployment.yaml and kubeadm-controlplane.yaml, the main flatcar template and the CI one still have the newline. I've run ./hack/gen-flavors.sh and full make generate locally and there's no diff so tilt still stumbles.

Hmm, I was hoping it only complains about kustomize files... I'll have a deeper look then next week. Thanks for reviews!

invidian:
>Hmm, I was hoping it only complains about kustomize files... I'll have a deeper look then next week. Thanks for reviews!

Sigh:
https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/8e2e3beb3cfa86d49977503a0bb72d22d640be3a/Tiltfile#L342-L344

This replace to escape is not great. I'm actually concerned it may also expand `$(curl -s -H Metadata:true` part. The whole thing look _very_ fragile to me.

invidian:

>Perhaps whole logic of deploy_worker_templates should go to a script file?

Anyway, fixed with https://docs.tilt.dev/api.html#api.shlex.quote. Please have a look @mboersma. Tilt is still running locally for me though.

mboersma:
I tried to deploy Flatcar from `make tilt-up` but now I see this error about a `nil` cluster-name:

```shell
...
./hack/gen-flavors.sh
make[2]: Nothing to be done for `kustomize'.
error validating "STDIN": error validating data: unknown object type "nil" in KubeadmControlPlane.spec.kubeadmConfigSpec.clusterConfiguration.controllerManager.extraArgs.cluster-name; if you choose to ignore these errors, turn validation off with --validate=false
error when retrieving current configuration of:
Resource: "cluster.x-k8s.io/v1beta1, Resource=clusters", GroupVersionKind: "cluster.x-k8s.io/v1beta1, Kind=Cluster"
Name: "", Namespace: "default"
from server for: "STDIN": resource name may not be empty
error when retrieving current configuration of:
Resource: "infrastructure.cluster.x-k8s.io/v1beta1, Resource=azureclusters", GroupVersionKind: "infrastructure.cluster.x-k8s.io/v1beta1, Kind=AzureCluster"
Name: "", Namespace: "default"
from server for: "STDIN": resource name may not be empty
```

invidian:
>I tried to deploy Flatcar from make tilt-up but now I see this error about a nil cluster-name:

@mboersma sorry I didn't check it all works properly. Now I did, even though the last problem was not at all Flatcar related but it seemed broken since a while? In any case, I fixed it and now flatcar flavor target in Tilt works as expected.

However, given that Flatcar flavor requires some extra environment variables, I put them in temporary commit for the time being, as I'm not sure if this is the best place to put those. In general, it seem a bit confusing that Tilt has it's own separate K8s version configuration. Perhaps it could re-use the discovery mechanism implemented for e2e tests at some point.

mboersma:
@invidian Flatcar provisioning doesn't give me errors now in `make tilt-up`, but once the cluster starts provisioning I get this error in the CAPZ controller:

```
E0118 18:12:13.633735      11 controller.go:326]  "msg"="Reconciler error" "error"="failed to reconcile AzureMachine: failed to reconcile AzureMachine service virtualmachine: failed to create resource flatcar-28790/flatcar-28790-control-plane-gh74f (service: virtualmachine): compute.VirtualMachinesClient#CreateOrUpdate: Failure sending request: StatusCode=404 -- Original Error: Code=\"GalleryImageNotFound\" Message=\"\\\"The gallery image /CommunityGalleries/flatcar4capi-742ef0cb-dcaa-4ecb-9cb0-bfd2e43dccc0/Images/flatcar-stable-amd64-capi-v1.26.0/Versions/3374.2.1 is not available in eastus region. Please contact image owner to replicate to this region, or change your requested region.\\\"\" Target=\"imageReference\"" "AzureMachine"={"name":"flatcar-28790-control-plane-gh74f","namespace":"default"} "controller"="azuremachine" "controllerGroup"="infrastructure.cluster.x-k8s.io" "controllerKind"="AzureMachine" "name"="flatcar-28790-control-plane-gh74f" "namespace"="default" "reconcileID"="143b7f2f-eecc-45f7-b18a-e48a91b86d77"
I0118 18:12:16.254367      11 azurecluster_controller.go:203] controllers.AzureClusterReconciler.reconcileNormal "msg"="Reconciling AzureCluster" "AzureCluster"={"name":"flatcar-28790","namespace":"default"} "controller"="azurecluster" "controllerGroup"="infrastructure.cluster.x-k8s.io" "controllerKind"="AzureCluster" "name"="flatcar-28790" "namespace"="default" "reconcileID"="8222aafa-03c6-4a16-aad0-422172e88359" "x-ms-correlation-request-id"="7d62998d-03fc-4960-8933-c3da482af230"
```

I assume this is unexpected, or should I test in a region other than `eastus`?

invidian:
>I assume this is unexpected, or should I test in a region other than eastus?

Kind of unexpected. I only made sure images which are used by CI are replicated. I'll fix this now for remaining images. Default k8s/flatcar version specified in Tiltfile is also replicated. In general, it's best to look up what images are available before deploying, as we do not replicate to all regions yet.

invidian:
1.26.0 should be now replicated to other regions.

mboersma:
> 1.26.0 should be now replicated to other regions.

I went to lunch and came back to find my cluster had eventually found the images you replicated and came up all healthy. 👍🏻

templates/cluster-template-flatcar.yaml
invidian: This is now implemented and reviewed, so I'll mark as resolved.

Tiltfile
```diff
@@ -8,6 +8,8 @@ tools_bin = "./hack/tools/bin"
 
 #Add tools to path
 os.putenv("PATH", os.getenv("PATH") + ":" + tools_bin)
+os.putenv("EXP_KUBEADM_BOOTSTRAP_FORMAT_IGNITION", "true")
+os.putenv("FLATCAR_VERSION", "3374.2.1")
```
CecileRobertMichon, line 12: why is this here and not with the other defaulted env variables? Even if we default these we should still allow them to be overwritten

invidian: Thanks for feedback @CecileRobertMichon. Are you referring here only to `FLATCAR_VERSION` or also to `EXP_KUBEADM_BOOTSTRAP_FORMAT_IGNITION`? And what other env variables do you have in mind? As I said in the previous [comment](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2890#issuecomment-1386795155), I wasn't sure what's the best place to put those.

Perhaps `FLATCAR_VERSION` could be added to `substitutions` map or be deeply wired into Tilt settings.

`EXP_KUBEADM_BOOTSTRAP_FORMAT_IGNITION` may probably need to go in some different place, as it must be set when deploying the CAPI controlplane. Please advice me where should I put those so we can get this PR merged.

Thanks in advance.

(BTW, I didn't trigger the test or re-run them, as there is a temporary commit pushed which needs to be squashed and I know the tests are passing, as recently changed code only touches Tiltfile and I wanted to save some resources and notifications churn)

CecileRobertMichon: I was referring to both.

- `FLATCAR_VERSION` should go in https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/b438dbec0b0b5b26dc213a2d19ccbee1557fa57e/Tiltfile#L17. If we want it to be hardcoded (not user configuarable, follow `capi_version` as an example. If you do want it to be configurable, you can follow `kubernetes_version` as the example. This is what the dev experience for configuring it would look like: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/templates/flavors/README.md#defining-variable-overrides
- `EXP_KUBEADM_BOOTSTRAP_FORMAT_IGNITION ` goes in https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/Makefile#L709 with the other exp variables

invidian: Fixed, thank you for pointers.

mboersma (commented):
/lgtm

Tiltfile changes look good, and I re-tested Flatcar in `make tilt-up` with the latest commits.

CecileRobertMichon:
/lgtm
/approve
/hold

Hold to wait for optional job with Flatcar to pass

invidian:
Thanks for approving @CecileRobertMichon! Do you think we could backport this PR to v1.7.x branch, so users don't have to wait for v1.8.0?

CecileRobertMichon:
> Do you think we could backport this PR to v1.7.x branch, so users don't have to wait for v1.8.0?

This would not qualify as cherry-pick as we only backport bug fixes, test improvements, and CVE-related dependency bumps. However, it's only a matter of users using the template right? I don't think anything in this PR adds CPAZ support (aside from Tilt and tests), so I nothing prevents a user from starting to use this today?

invidian:
Yes, this is just a matter of adding the Flatcar template, so `clusterctl generate cluster --flavor flatcar` is working :) For this to work, the template must be included in the release artifacts. I could also create a separate PR to v1.7.x branch adding only the templates, if that helps.

mboersma:

Flatcar passed on the previous run but dual-stack failed with a network timeout.

CecileRobertMichon:
> Yes, this is just a matter of adding the Flatcar template, so clusterctl generate cluster --flavor flatcar is working :) For this to work, the template must be included in the release artifacts. I could also create a separate PR to v1.7.x branch adding only the templates, if that helps.

IMO we shouldn't backport any of it, even the template as it's adding brand-new functionality, not fixing an existing one. Not backporting it also gives us a month to run the new test so we know the template is stable before we release it as an official "flavor" asset.

If anyone really can't wait a month for our planned v1.8.0 release date ([March 7th](https://github.com/kubernetes-sigs/cluster-api-provider-azure/milestone/30)) to use this, they can always do `clusterctl generate cluster --from https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/templates/cluster-template-flatcar.yaml` once this merges.

invidian:
/unhold

Optional tests passed now.

>IMO we shouldn't backport any of it, even the template as it's adding brand-new functionality, not fixing an existing one. Not backporting it also gives us a month to run the new test so we know the template is stable before we release it as an official "flavor" asset.

Sure, make sense. I was just curious.

PR #2894, opened 2022-12-02
Convert resourceGroup name in ID for VMSS vms to lower case to keep it compatible with the rest of capz

primeroz:
**What type of PR is this?**

/kind bug

This PR is alternative to https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2821 and only one of them should be merged. 

In this PR the conversion of the ResourceGroup Name happens when the VM instance is returned from Azure SDK and the AzureMachinePoolMachine is created rather than at read time.

If for any reason the conversion fail the original string is used. 

Fixes # https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/2817

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
fix machinepool reconciliation by converting the resourceGroup name in the AzureMachinePoolMachine object providerID to lowercase to match the providerID defined in the kubernetes Node Object on the workload cluster.

This fix only apply to _new azuremachinepoolmachine_ objects so if any existing machinepool with resourceGroup name with capital letter already exist , the controller with this patch will not manage to fix the azuremachinepoolmachines in it and it will still fail to reconcile the pool until the pool is deleted. 
```

### DIFF
--- a/azure/converters/vmss.go
+++ b/azure/converters/vmss.go
@@ -21,6 +21,7 @@ import (
 	"github.com/Azure/go-autorest/autorest/to"
 	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
 	"sigs.k8s.io/cluster-api-provider-azure/azure"
+	azureutil "sigs.k8s.io/cluster-api-provider-azure/util/azure"
 )
 
 // SDKToVMSS converts an Azure SDK VirtualMachineScaleSet to the AzureMachinePool type.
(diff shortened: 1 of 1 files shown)


primeroz:
@CecileRobertMichon @mboersma 

this is the alternative implementation of the fix for the providerID resourceGroupName uppercase letter bugs

in this case the conversion happens when the object is written, the only drawback is that for `existing azuremachinepoolmachine` objects , that are already failing to reconcile because of the uppercase letter in the providerID, this fix does manage to to address the issue . this only works for new objects 

I don't think this is a big deal since those are already broken but anyway i wanted to point that out.

primeroz:
reference : https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2895/files

CecileRobertMichon:
Linter is failing

>azure/converters/vmss.go:66: unnecessary leading newline (whitespace)
func SDKToVMSSVM(sdkInstance compute.VirtualMachineScaleSetVM) *azure.VMSSVM {

azure/converters/vmss.go
```diff
@@ -63,8 +64,16 @@ func SDKToVMSS(sdkvmss compute.VirtualMachineScaleSet, sdkinstances []compute.Vi
 
 // SDKToVMSSVM converts an Azure SDK VirtualMachineScaleSetVM into an infrav1exp.VMSSVM.
 func SDKToVMSSVM(sdkInstance compute.VirtualMachineScaleSetVM) *azure.VMSSVM {
```
CecileRobertMichon, line 66: noticed we don't have unit test coverage for this so opened #2896

CecileRobertMichon (commented):
lgtm pending fixed linter

CecileRobertMichon:
Closes #2821

PR #2909, opened 2022-12-07
split out AKS E2E source files

jackfrancis:
**What type of PR is this?**

/kind cleanup

This PR splits up all of the various AKS E2E test foo in `test/e2e/aks.go` into a bunch of discrete files to maintain specific tests.

The practical impetus of this is to use a newer SDK version of the AKS API which has a more updated representation of configuration. The existing AKS "get me the latest version" foo, unfortunately, depends on an older version of the SDK which is not forward-compatible. Attempting to do that surgery at this time is more work than justifies (IMO).

This is also (IMO) a more maintainable path.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```


nawazkh:
Looks like timeout errors on the PR tests, thanks for putting this together!
```
[FAILED] Timed out after 1800.001s.
[1251](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/2909/pull-cluster-api-provider-azure-e2e/1600296720208498688#1:build-log.txt%3A1251)
  Timed out waiting for 3 control plane machines to exist
[1252](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/2909/pull-cluster-api-provider-azure-e2e/1600296720208498688#1:build-log.txt%3A1252)
  Expected
[1253](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/2909/pull-cluster-api-provider-azure-e2e/1600296720208498688#1:build-log.txt%3A1253)
      <int>: 2
[1254](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/2909/pull-cluster-api-provider-azure-e2e/1600296720208498688#1:build-log.txt%3A1254)
  to equal
[1255](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/2909/pull-cluster-api-provider-azure-e2e/1600296720208498688#1:build-log.txt%3A1255)
      <int>: 3
[1256](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/2909/pull-cluster-api-provider-azure-e2e/1600296720208498688#1:build-log.txt%3A1256)
  In [It] at: /home/prow/go/pkg/mod/sigs.k8s.io/cluster-api/test@v1.3.0/framework/controlplane_helpers.go:117 @ 12/07/22 01:58:59.972
```

test/e2e/aks_autoscaler.go
```diff
@@ -0,0 +1,125 @@
+//go:build e2e
+// +build e2e
+
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
```
CecileRobertMichon, line 8: instead of having a bunch of aks_ files, what are your thoughts on adding an `aks` sub folder?

jackfrancis: That's a slightly more advanced operation but if that has more long-term maintenance benefits, happy to do that.

@mboersma @nojnhuh thoughts here?

nojnhuh: I don't see any strong reasons to prefer one option over the other. I can see where a subdirectory would keep the file tree a bit tidier, but readability in azure_test.go might take a marginal hit if the AKS test specs need a package prefix. In that case if we had something like `AzureLBSpec` and `aks.AutoscaleSpec` used similarly, I might assume at first glance that they're more different conceptually than they actually are based on one being in a sub-package. That's quite a bit of "reading between the lines" though.

Overall I think I'm leaning slightly toward keeping the AKS files in this existing directory and not putting them in their own.

CecileRobertMichon: To clarify I don't have any strong arguments for adding a folder, just pointing it out since it jumped at me when seeing all the different files with the same prefix. Either option is fine with me.

PR #2915, opened 2022-12-08
Make e2e Helm install idempotent

CecileRobertMichon:
**What type of PR is this?**
/kind failing-test

**What this PR does / why we need it**: In order to fix CSI migration test errors (https://testgrid.k8s.io/sig-cluster-lifecycle-cluster-api-provider-azure#capi-periodic-upgrade-main-1-22-1-23) that started after #2495 merged, we need to make the Helm install idempotent for Calico and cloud provider. Also fixed the errors in the test by not installing CSI drivers before k8s 1.22 and not adding the HostProcess feature flag for windows when windows is not enabled on the template (and the right feature flags aren't in place).

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/cloud-provider-azure.go
+++ b/test/e2e/cloud-provider-azure.go
@@ -43,7 +43,7 @@ const (
 // and validates that expected pods exist and are Ready.
 func InstallCalicoAndCloudProviderAzureHelmChart(ctx context.Context, input clusterctl.ApplyClusterTemplateAndWaitInput, cidrBlocks []string, hasWindows bool) {
 	specName := "cloud-provider-azure-install"
-	By("Installing the correct version of cloud-provider-azure components via helm")
+	By("Installing cloud-provider-azure components via helm")
 	options := &helmVals.Options{
 		Values: []string{fmt.Sprintf("infra.clusterName=%s", input.ConfigCluster.ClusterName), fmt.Sprintf("cloudControllerManager.clusterCIDR=%s", strings.Join(cidrBlocks, `,`))},
 	}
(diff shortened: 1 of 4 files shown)


CecileRobertMichon:
All the CSI migration specific errors are gone, the last run (https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/2915/pull-cluster-api-provider-azure-e2e-csi-migration/1603196398872825856) fails with the upgrade CoreDNS image pull known issue (https://github.com/kubernetes-sigs/cluster-api/issues/7768) which is the same error we're seeing in https://testgrid.k8s.io/sig-cluster-lifecycle-cluster-api-provider-azure#capi-periodic-upgrade-main-1-23-1-24.

I vote we merge this as an incremental improvement even if it doesn't completely fix the test and work on fixing the CoreDNS image pull issue separately (that will likely require changes in CAPI).

/assign @marosset @jackfrancis

test/e2e/cloud-provider-azure.go
```diff
@@ -66,11 +66,15 @@ func InstallCalicoAndCloudProviderAzureHelmChart(ctx context.Context, input clus
...
+func InstallAzureDiskCSIDriverHelmChart(ctx context.Context, input clusterctl.ApplyClusterTemplateAndWaitInput, hasWindows bool) {
 	specName := "azuredisk-csi-drivers-install"
-	By("Installing the correct version of azure-disk CSI driver components via helm")
+	By("Installing azure-disk CSI driver components via helm")
 	options := &helmVals.Options{
-		Values: []string{"windows.useHostProcessContainers=true", "controller.replicas=1", "controller.runOnControlPlane=true"},
+		Values: []string{"controller.replicas=1", "controller.runOnControlPlane=true"},
+	}
+	// TODO: make this always true once HostProcessContainers are on for all supported k8s versions.
+	if hasWindows {
```
jackfrancis, line 76: Should we propose moving this logic into the official helm chart as well? (before injecting host process stuff into the actual resource spec you must enable the `windows.useHostProcessContainers` chart config *and* be installing onto a valid version of Kubernetes)

jackfrancis:
/lgtm

@CecileRobertMichon are these changes needed in release-1.6 and/or release-1.5 to fix broken tests therein?

CecileRobertMichon:
> are these changes needed in release-1.6 and/or release-1.5 to fix broken tests therein?

no, #2495 wasn't cherry-picked (rightly so) so the regressions this is fixing were never in 1.5/1.6

test/e2e/cloud-provider-azure.go
CecileRobertMichon: That's what the Helm chart is validating actually (as part of the daemonset) and why that option wasn't valid and failing the Helm chart install for templates that didn't enable to correct feature flag for k8s < 1.26

jackfrancis: Oh got it, thanks.

marosset: Do we support anything less than 1.22 now?

CecileRobertMichon: No and in fact I disabled this for anything < 1.23 (https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2915/files#diff-2c7fc9c9c12632c5e550abeb31614356abd17710d948a8664f9d2f2163e38b0dR266). This is to cover the case between 1.23 and 1.26 where HostProcess containers still require the feature flag  for this to work (otherwise the Helm install fails), so this makes sure we don't break e2e templates that don't have windows (and thus might not have the feature flag set, which is the case for some of the CSI migration test templates).

CecileRobertMichon:
/retest

known issue fixed by #2947

CecileRobertMichon (commented):
/approve

CecileRobertMichon:
Good news: https://testgrid.k8s.io/sig-cluster-lifecycle-cluster-api-provider-azure#capi-periodic-upgrade-main-1-22-1-23 is now passing since this merged

PR #2916, opened 2022-12-08
fix deleting all tags on AKS resources

nojnhuh:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**: When `spec.additionalTags` are defined on `AzureManagedControlPlane` or `AzureManagedMachinePool` resources, setting them to `null` or empty does not delete all tags because the resulting `PUT` request specifies `null` tags which seems to be interpreted by AKS as "do not change the tags." This change ensures that if the tags are specified in the CAPZ resources as `null`, then a non-nil, empty set of tags are sent to the AKS API to express "please delete all of the existing tags."

Fixes #

**Special notes for your reviewer**: I came across this issue while implementing an e2e test covering this scenario (currently drafted in #2917). Those tests verify this new behavior for me locally. Additional context in https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2745#discussion_r1016508287 and https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2802.

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Fixed a bug preventing `spec.additionalTags` from being deleted entirely on AzureManagedControlPlane
```
Note: `spec.additionalTags` is not yet implemented for AzureManagedMachinePools in any tagged release

### DIFF
--- a/azure/services/managedclusters/spec.go
+++ b/azure/services/managedclusters/spec.go
@@ -190,6 +190,7 @@ func (s *ManagedClusterSpec) Parameters(existing interface{}) (params interface{
 			Type: containerservice.ResourceIdentityTypeSystemAssigned,
 		},
 		Location: &s.Location,
+		Tags:     *to.StringMapPtr(s.Tags),
 		ManagedClusterProperties: &containerservice.ManagedClusterProperties{
 			NodeResourceGroup: &s.NodeResourceGroup,
 			EnableRBAC:        to.BoolPtr(true),
@@ -215,10 +216,6 @@ func (s *ManagedClusterSpec) Parameters(existing interface{}) (params interface{
 				NetworkPolicy:   containerservice.NetworkPolicy(s.NetworkPolicy),
 			},
 		},
-	}
-
-	if tags := *to.StringMapPtr(s.Tags); len(tags) != 0 {
-		managedCluster.Tags = tags
 	}
 
 	if s.PodCIDR != "" {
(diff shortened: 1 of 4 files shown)


azure/services/managedclusters/spec.go
```diff
@@ -327,6 +324,11 @@ func (s *ManagedClusterSpec) Parameters(existing interface{}) (params interface{
 		// AgentPool changes are managed through AMMP.
 		managedCluster.AgentPoolProfiles = existingMC.AgentPoolProfiles
 
+		// Do not trigger an update because of nil/empty discrepancies between the two sets of tags.
+		if len(existingMC.Tags) == 0 && len(managedCluster.Tags) == 0 {
+			existingMC.Tags = managedCluster.Tags
```
jackfrancis, line 329: would this be better?

```
existingMC.Tags, managedCluster.Tags = nil, nil
```

setting the "existing" to the "local" seems to suggest we're preferring the "local" value, when in fact what we're really trying to do is to ensure that they both have the same value so that a diff isn't computed between them

is that right?

nojnhuh: I suppose what I was trying to enforce here is that the resulting `managedCluster.Tags` always resolves to a non-nil, empty map when the `spec.additionalTags` are nil or empty. But I see now that it looks fine for `managedCluster.Tags` to be nil when `existingMC.Tags` is also empty since our desired state already matches the actual state in that case. Overall I think your suggestion is a bit more clear so I'll test that to make sure it works and make that change unless I run into any issues.

nawazkh:
Great find!
/lgtm

jackfrancis:
nevermind the cherry-picks, the `AdditionalTags` feature was only introduced in the current release cycle, it doesn't exist as a feature in 1.6 or below.

nojnhuh:
> nevermind the cherry-picks, the `AdditionalTags` feature was only introduced in the current release cycle, it doesn't exist as a feature in 1.6 or below.

The field is new for AzureManagedMachinePools, but this issue does affect AzureManagedControlPlane's `spec.additionalTags` on 1.6 and 1.5, so I'll work on manual cherry-picks for those.

PR #2917, opened 2022-12-08
add e2e test for AKS resource tag operations

### LINKED ISSUE #2866: AzureManagedCluster E2E: AdditionalTags CRUD tests
/kind feature
/area managedclusters

Create new AzureManagedCluster E2E: AdditionalTags CRUD tests

nojnhuh:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**: This change adds new e2e test coverage to validate creating, updating, and deleting tags on AzureManagedControlPlane and AzureManagedMachinePool resources.

Fixes #2866

**Special notes for your reviewer**: In its current state, this test will fail due a bug that will be fixed by #2916, so that PR should be merged first.

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/aks_tags.go
+++ b/test/e2e/aks_tags.go
@@ -1,0 +1,185 @@
+//go:build e2e
+// +build e2e
+
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package e2e
+
+import (
+	"context"
+	"sync"
+
+	"github.com/Azure/azure-sdk-for-go/services/containerservice/mgmt/2021-05-01/containerservice"
+	"github.com/Azure/go-autorest/autorest/azure/auth"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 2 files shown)


test/e2e/aks_tags.go
```diff
@@ -0,0 +1,172 @@
...
+			defer GinkgoRecover()
+			defer wg.Done()
+
+			Byf("Creating tags for machine pool %s", mp.Name)
+			ammp := &infrav1exp.AzureManagedMachinePool{}
+			Expect(mgmtClient.Get(ctx, types.NamespacedName{
+				Namespace: mp.Spec.Template.Spec.InfrastructureRef.Namespace,
+				Name:      mp.Spec.Template.Spec.InfrastructureRef.Name,
+			}, ammp)).To(Succeed())
+			Expect(ammp.Spec.AdditionalTags).To(BeNil())
```
jackfrancis, line 128: Is this base case durable enough? Should we clear out the tags first of all like we're doing for control plane so that this scenario will work w/ a cluster spec that ships w/ AdditionalTags in an AzureManagedMachinePool out of the gate?

nojnhuh: Yeah that seems like a good idea. I'll add that here on the front end in addition to the "delete all" later since we know for sure at that point that there are tags to delete. I might tweak the corresponding test against AzureManagedControlPlane too to be more symmetrical with this one too.

jackfrancis: Cool. Super excited to land these, this is awesome.

nojnhuh: Just pushed this change. Now the tests against AzureManagedControlPlane and AzureManagedMachinePool are doing the exact same sequence of operations, roughly:
1. Delete all the existing tags
2. Create some new tags
3. Update the tags
4. Restore the initial tags

Whether or not tags were initially present, this should cover the main scenarios. The only slightly unintuitive nuance is that when no tags are initially defined, the "delete all tags" step is basically a no-op and the "restore initial tags" step is really testing the delete.

jackfrancis: sgtm

I think the coverage of "no op delete all tags" is still an interesting scenario fwiw.

jackfrancis (approved):
/lgtm
/approve

PR #2920, opened 2022-12-09
remove --listen-client-urls injection from ipv6 template

jackfrancis:
**What type of PR is this?**

/kind failing-test

Since updating to Kubernetes 1.24 in https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2773, we are noticing failures in the ipv6 E2E scenario. It appears that removing a sed command that manually injects a ipv4 localhost configuration for `--listen-client-urls` into the etcd pod manifest via a postKubeadmCommand is no longer working reliably.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/templates/cluster-template-ipv6.yaml
+++ b/templates/cluster-template-ipv6.yaml
@@ -148,7 +148,6 @@ spec:
     - - LABEL=etcd_disk
       - /var/lib/etcddisk
     postKubeadmCommands:
-    - sed -i '\#--listen-client-urls#s#$#,https://127.0.0.1:2379#' /etc/kubernetes/manifests/etcd.yaml
     - echo "DNSStubListener=no" >> /etc/systemd/resolved.conf
     - mv /etc/resolv.conf /etc/resolv.conf.OLD && ln -s /run/systemd/resolve/resolv.conf
       /etc/resolv.conf
--- a/templates/flavors/ipv6/patches/kubeadm-controlplane.yaml
+++ b/templates/flavors/ipv6/patches/kubeadm-controlplane.yaml
@@ -5,7 +5,6 @@ metadata:
 spec:
   kubeadmConfigSpec:
     postKubeadmCommands:
-      - sed -i '\#--listen-client-urls#s#$#,https://127.0.0.1:2379#' /etc/kubernetes/manifests/etcd.yaml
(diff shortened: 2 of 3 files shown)


jackfrancis:
cc @aramase

mboersma:
So this `sed` command that's being removed is somehow causing side effects? We can remove it because it was optional or because it's not working?

jackfrancis:
@mboersma yes, it is causing side-effects (root cause not yet understood)

@aramase is investigating why we may or may not need what the sed command is trying to do any longer (set up a ipv4 localhost listener when cluster is ipv6-configured)

CecileRobertMichon:
Did a bit of digging around, there's a good chance this is the relevant etcd commit: https://github.com/etcd-io/etcd/commit/36351744ebc3da400c31e87c2a84151a5f0211f0 (@aramase to help confirm) - seems like it was just part of etcd release starting with `v3.6.0`

Now, acccording to https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/constants/constants.go#L480 both k8s 1.23 and k8s 1.24 use etcd v3.6.0. But looking at git blame that's a recent (12 days ago!) change https://github.com/kubernetes/kubernetes/pull/114093 AND it got cherry-picked... 

This could explain why our test suddenly started failing with the version bump but we can't repro with the previous version either (if they both got the cherry-pick).

jackfrancis:
@CecileRobertMichon good find, I agree that definitely looks the fix to the original etcd listener issue. So:

1) in terms of capz reference templates we probably need to still worry about users running older versions of etcd (the version of k8s is an env var)
2) why did the sed command suddenly start producing terminal etcd static pod issues at a roughly 10-15% per-kubeadm-init/join failure rate

Btw I haven’t had a chance to check today, are we still seeing this issue in the E2E tests?

CecileRobertMichon:
1. k8s 1.22 and above all use the same etcd version that has the fix: https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/constants/constants.go#L480 so I think we're good (1.21 is way out of support at this point)
2. 🤷‍♀️ 

Let's get another data point

PR #2925, opened 2022-12-12
add E2E tests for AzureManagedCluster webhooks

jackfrancis:
**What type of PR is this?**

/kind feature

This PR adds E2E tests for the full set of webhook update rules for AKS. We don't fully test every invalid permutation across all properties (or properties of properties, i.e., structs), but rather include a working "negative" E2E test case for every `AzureManagedCluster`, `AzureManagedControlPlane`, and `AzureManagedMachinePool` webhook.

This process uncovered an unnecessary webhook for the `AzureManagedControlPlane.Name` property — the `.Name` field is the unique ID for these k8s Object.Meta resources, and updates to `.Name` violate rules that prevent the webhooks from ever being reached (in practice I observed `409` conflict errors). That webhook (and its related UT) are removed in this PR.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/aks.go
+++ b/test/e2e/aks.go
@@ -25,6 +25,7 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"k8s.io/apimachinery/pkg/types"
+	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
 	infrav1exp "sigs.k8s.io/cluster-api-provider-azure/exp/api/v1beta1"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	expv1 "sigs.k8s.io/cluster-api/exp/api/v1beta1"
@@ -207,4 +208,69 @@ func WaitForAKSSystemNodePoolMachinesToExist(ctx context.Context, input WaitForC
 
 		return false
 	}, intervals...).Should(Equal(true), "System machine pools not detected")
+}
+
+func getAzureCluster(ctx context.Context, managementClusterClient client.Client, namespace, name string) (*infrav1.AzureCluster, error) {
+	key := client.ObjectKey{
+		Namespace: namespace,
+		Name:      name,
+	}
+
+	azCluster := &infrav1.AzureCluster{}
+	err := managementClusterClient.Get(ctx, key, azCluster)
+	return azCluster, err
+}
+
+func getAzureManagedControlPlane(ctx context.Context, managementClusterClient client.Client, namespace, name string) (*infrav1exp.AzureManagedControlPlane, error) {
+	key := client.ObjectKey{
+		Namespace: namespace,
+		Name:      name,
+	}
+
+	azManagedControlPlane := &infrav1exp.AzureManagedControlPlane{}
+	err := managementClusterClient.Get(ctx, key, azManagedControlPlane)
+	return azManagedControlPlane, err
+}
+
+func getAzureManagedCluster(ctx context.Context, managementClusterClient client.Client, namespace, name string) (*infrav1exp.AzureManagedCluster, error) {
+	key := client.ObjectKey{
+		Namespace: namespace,
+		Name:      name,
+	}
+	azManagedCluster := &infrav1exp.AzureManagedCluster{}
+	err := managementClusterClient.Get(ctx, key, azManagedCluster)
+	return azManagedCluster, err
+}
+
+func getAzureMachine(ctx context.Context, managementClusterClient client.Client, m *clusterv1.Machine) (*infrav1.AzureMachine, error) {
+	key := client.ObjectKey{
+		Namespace: m.Spec.InfrastructureRef.Namespace,
+		Name:      m.Spec.InfrastructureRef.Name,
+	}
+
+	azMachine := &infrav1.AzureMachine{}
+	err := managementClusterClient.Get(ctx, key, azMachine)
+	return azMachine, err
+}
+
+func getAzureMachinePool(ctx context.Context, managementClusterClient client.Client, mp *expv1.MachinePool) (*infrav1exp.AzureMachinePool, error) {
+	key := client.ObjectKey{
+		Namespace: mp.Spec.Template.Spec.InfrastructureRef.Namespace,
+		Name:      mp.Spec.Template.Spec.InfrastructureRef.Name,
+	}
+
+	azMachinePool := &infrav1exp.AzureMachinePool{}
+	err := managementClusterClient.Get(ctx, key, azMachinePool)
+	return azMachinePool, err
+}
+
+func getAzureManagedMachinePool(ctx context.Context, managementClusterClient client.Client, mp *expv1.MachinePool) (*infrav1exp.AzureManagedMachinePool, error) {
+	key := client.ObjectKey{
+		Namespace: mp.Spec.Template.Spec.InfrastructureRef.Namespace,
+		Name:      mp.Spec.Template.Spec.InfrastructureRef.Name,
+	}
+
+	azManagedMachinePool := &infrav1exp.AzureManagedMachinePool{}
+	err := managementClusterClient.Get(ctx, key, azManagedMachinePool)
+	return azManagedMachinePool, err
 }
--- a/test/e2e/aks_webhooks.go
+++ b/test/e2e/aks_webhooks.go
@@ -1,0 +1,384 @@
+//go:build e2e
+// +build e2e
+
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package e2e
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/Azure/azure-sdk-for-go/services/containerservice/mgmt/2021-05-01/containerservice"
+	"github.com/Azure/go-autorest/autorest/azure/auth"
+	"github.com/Azure/go-autorest/autorest/to"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/cluster-api-provider-azure/azure"
+	infrav1exp "sigs.k8s.io/cluster-api-provider-azure/exp/api/v1beta1"
+	azureutil "sigs.k8s.io/cluster-api-provider-azure/util/azure"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	expv1 "sigs.k8s.io/cluster-api/exp/api/v1beta1"
+)
+
+type AKSWebhooksSpecInput struct {
+	Cluster      *clusterv1.Cluster
+	MachinePools []*expv1.MachinePool
+}
+
+func AKSWebhooksSpec(ctx context.Context, inputGetter func() AKSWebhooksSpecInput) {
+	input := inputGetter()
+
+	settings, err := auth.GetSettingsFromEnvironment()
+	Expect(err).NotTo(HaveOccurred())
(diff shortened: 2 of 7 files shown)


jackfrancis:
@nojnhuh how does this initial E2E scenario look as a surface area for lots more webhook tests?

test/e2e/aks_webhooks.go
```diff
@@ -0,0 +1,74 @@
...
+	err = mgmtClient.Get(ctx, types.NamespacedName{
+		Namespace: input.Cluster.Namespace,
+		Name:      input.Cluster.Name,
+	}, amc)
+
+	By("Validating that AzureManagedCluster webhooks are working")
+	By("Validating that we can't manually add an annotation to AzureManagedCluster with a prefix of 'infrastructure.cluster.x-k8s.io/custom-header-'")
+	amc.ObjectMeta.Annotations[azure.CustomHeaderPrefix+"foo"] = "bar"
+	err = mgmtClient.Update(ctx, amc)
+	Expect(err).To(HaveOccurred())
```
nojnhuh, line 65: I think it would be worth doing some sort of minimal "this is an error from the webhook" validation here to be safe, like [`apierrors.IsForbidden`](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/errors#IsForbidden) (assuming that's correct). I've definitely made mistakes when creating tests that generate HTTP 409 conflicts or similarly unexpected errors so I'd personally appreciate a little bit of guardrail here.

nojnhuh (commented):
Left one small comment. Overall looks good!

test/e2e/azure_logcollector.go
```diff
@@ -177,61 +176,6 @@ func getHostname(m *clusterv1.Machine, isWindows bool) string {
 	return hostname
 }
 
-func getAzureCluster(ctx context.Context, managementClusterClient client.Client, namespace, name string) (*infrav1.AzureCluster, error) {
```
jackfrancis, line 180: I was able to re-use one of these existing funcs, but needed another one that wasn't there, so took the opportunity to add that (`getAzureManagedCluster`) in the more general-purpose `aks.go` source file rather than here. And because of the re-use of `getAzureManagedControlPlane` it seems like we can move all of these (reusable) funcs to `aks.go` from this log collection-specific source file.

test/e2e/aks_webhooks.go
```diff
@@ -0,0 +1,391 @@
...
+	amcp, err = getAzureManagedControlPlane(ctx, mgmtClient, input.Cluster.Namespace, input.Cluster.Name)
+	Expect(err).NotTo(HaveOccurred())
+	if to.String(amcp.Spec.NetworkPolicy) == "azure" {
+		amcp.Spec.NetworkPolicy = to.StringPtr("calico")
+	} else {
+		amcp.Spec.NetworkPolicy = to.StringPtr("azure")
+	}
+	err = mgmtClient.Update(ctx, amcp)
+	Expect(err).To(HaveOccurred())
+	Expect(apierrors.IsInvalid(err)).To(BeTrue())
```
nojnhuh, line 144: I wonder if a lot of the boilerplate could be removed with an extra layer of `BeforeEach`/`AfterEach`, if Ginkgo allows that here. I imagine that might be able to turn most of this into something like:

```go
	var amcp *infrav1exp.AzureManagedControlPlane

	BeforeEach(func() {
		amcp, err = getAzureManagedControlPlane(ctx, mgmtClient, input.Cluster.Namespace, input.Cluster.Name)
	})

	AfterEach(func() {
		err = mgmtClient.Update(ctx, amcp)
		Expect(err).To(HaveOccurred())
		Expect(apierrors.IsInvalid(err)).To(BeTrue())
	})

	It("fails when updating Spec.NetworkPlugin", func() {
		if to.String(amcp.Spec.NetworkPlugin) == "azure" {
			amcp.Spec.NetworkPlugin = to.StringPtr("kubenet")
		} else {
			amcp.Spec.NetworkPlugin = to.StringPtr("azure")
		}
	})
```

Certainly not necessary to address now, just an idea.

test/e2e/aks_webhooks.go
```diff
@@ -0,0 +1,391 @@
...
+		}
+		ammp.Spec.AvailabilityZones = append(ammp.Spec.AvailabilityZones, "99")
+		err = mgmtClient.Update(ctx, ammp)
+		Expect(err).To(HaveOccurred())
+		Expect(apierrors.IsInvalid(err)).To(BeTrue())
+
+		ammp, err = getAzureManagedMachinePool(ctx, mgmtClient, mp)
+		Expect(err).NotTo(HaveOccurred())
+		if len(ammp.Spec.AvailabilityZones) > 0 {
+			ammp.Spec.AvailabilityZones = ammp.Spec.AvailabilityZones[:len(ammp.Spec.AvailabilityZones)-1]
```
nojnhuh, line 291: If all we want to accomplish here is changing the slice, this would be a bit more concise:
```suggestion
			ammp.Spec.AvailabilityZones = ammp.Spec.AvailabilityZones[1:]
```

test/e2e/aks_webhooks.go
```diff
@@ -0,0 +1,391 @@
...
+		Expect(err).To(HaveOccurred())
+		Expect(apierrors.IsInvalid(err)).To(BeTrue())
+
+		By("Validating that we can't add or remove from AzureManagedMachinePool.Spec.AvailabilityZones")
+		ammp, err = getAzureManagedMachinePool(ctx, mgmtClient, mp)
+		Expect(err).NotTo(HaveOccurred())
+		if ammp.Spec.AvailabilityZones == nil {
+			ammp.Spec.AvailabilityZones = []string{}
+		}
+		ammp.Spec.AvailabilityZones = append(ammp.Spec.AvailabilityZones, "99")
```
nojnhuh, line 283: FWIW I think this nil check can be omitted since appending to a nil slice directly works the same way: https://go.dev/play/p/m5rP1TWFiDU

test/e2e/aks_webhooks.go
```diff
@@ -0,0 +1,391 @@
...
+		Expect(apierrors.IsInvalid(err)).To(BeTrue())
+
+		By("Validating that we can't update AzureManagedMachinePool.Spec.EnableUltraSSD")
+		ammp, err = getAzureManagedMachinePool(ctx, mgmtClient, mp)
+		Expect(err).NotTo(HaveOccurred())
+		if to.Bool(ammp.Spec.EnableUltraSSD) {
+			ammp.Spec.EnableUltraSSD = to.BoolPtr(false)
+		} else {
+			ammp.Spec.EnableUltraSSD = to.BoolPtr(true)
+		}
```
nojnhuh, line 336: Maybe a bit too clever, but:
```suggestion
		ammp.Spec.EnableUltraSSD = to.BoolPtr(!to.Bool(ammp.Spec.EnableUltraSSD))
```

test/e2e/aks_webhooks.go
```diff
@@ -0,0 +1,391 @@
...
+		} else {
+			ammp.Spec.KubeletConfig = &infrav1exp.KubeletConfig{
+				FailSwapOn: to.BoolPtr(true),
+			}
+		}
+		err = mgmtClient.Update(ctx, ammp)
+		Expect(err).To(HaveOccurred())
+		Expect(apierrors.IsInvalid(err)).To(BeTrue())
+
+		By(fmt.Sprintf("Changing AzureManagedMachinePool %s mode to User", ammp.Name))
```
nojnhuh, line 379: I can see where if the cluster template ever includes more than one System node pool, then this test would have a side effect of changing all but one of them to User node pools. I think the likelihood of the template changing that way and that causing any actual issues is low enough that I doubt it's worth addressing that now though.

test/e2e/azure_test.go
```diff
@@ -589,6 +589,16 @@ var _ = Describe("Workload cluster creation", func() {
...
+				AKSWebhooksSpec(ctx, func() AKSWebhooksSpecInput {
+					return AKSWebhooksSpecInput{
+						Cluster:       result.Cluster,
+						MachinePools:  result.MachinePools,
+						WaitIntervals: e2eConfig.GetIntervals(specName, "wait-machine-pool-nodes"),
+					}
+				})
+			})
+
 			By("Upgrading the Kubernetes version of the cluster", func() {
```
nojnhuh, line 602: Do we want to keep the upgrade test running first immediately after the cluster is created?

nojnhuh (commented):
Left a few small comments inline. This looked like a bit of a slog, so thanks for adding these!

As to whether we should test these e2e at all, I think the fact that you caught at least some unexpected behavior here is a decent indicator that unit tests aren't going to cover everything. If we can get the same benefits in narrower tests with something like envtest, then that might be worth looking into, but I don't think that should block this.

test/e2e/aks.go
```diff
@@ -208,3 +209,74 @@ func WaitForAKSSystemNodePoolMachinesToExist(ctx context.Context, input WaitForC
...
+	key := client.ObjectKey{
+		Namespace: namespace,
+		Name:      name,
+	}
+
+	azManagedControlPlane := &infrav1exp.AzureManagedControlPlane{}
+	Eventually(func() bool {
+		err := managementClusterClient.Get(ctx, key, azManagedControlPlane)
+		return err == nil
+	}, retryableOperationTimeout, retryableOperationSleepBetweenRetries).Should(BeTrue(), "AzureManagedControlPlane/%s in namespace %s unable to be retrieved from mgmt cluster", name, namespace)
```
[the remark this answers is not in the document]
jackfrancis: Didn't run into anything, just biased towards retrying anything that has .0001% I/O error outcomes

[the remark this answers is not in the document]
jackfrancis: I was actually going to submit a follow-up PR to add this retry tolerance to all these GET calls :) (including the newly added ones in AKS E2E tests)

test/e2e/aks.go
```diff
@@ -208,3 +209,74 @@ func WaitForAKSSystemNodePoolMachinesToExist(ctx context.Context, input WaitForC
...
+		Namespace: namespace,
+		Name:      name,
+	}
+
+	azCluster := &infrav1.AzureCluster{}
+	err := managementClusterClient.Get(ctx, key, azCluster)
+	return azCluster, err
+}
+
+func getAzureManagedControlPlane(ctx context.Context, managementClusterClient client.Client, namespace, name string) (*infrav1exp.AzureManagedControlPlane, error) {
```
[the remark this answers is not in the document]
jackfrancis: Indeed, I was just trying to not add a weird looking func (there are already a bunch w/ a well-established return pattern).

How about this, I'll remove the retry stuff from this PR, and submit a follow-up PR that turns all of these into retryable (and asserts on a terminal error, rather than returns an err).

test/e2e/aks_webhooks.go
jackfrancis: lol I had PTSD from the `map[string]string` above

test/e2e/aks_webhooks.go
jackfrancis: I think we probably actually do want to anticipate a test scenario where we have (or update to) > 1 system pool. Because changing back and forth is in fact allowed, we should be cool to do that a bunch of times, and this particular test which does that in an excessive/not allowed way supports that.

(So long as we don't do something weird like leave a million kube-system pods around after we do a multiple system pool scenario and then leave around a bunch of orphaned, non-schedulable pods when we return to only 1 system pool — although even that particular weird edge case should still allow tests tests to pass AFAIK)

test/e2e/azure_test.go
jackfrancis: I don't think I care, I put these here to get faster test cycles.

nojnhuh (approved):
/lgtm

test/e2e/aks.go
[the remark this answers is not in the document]
nojnhuh: IIRC client-go has some level of retry logic built in, so it might be worth looking to see if that's sufficient for us.

jackfrancis:
@CecileRobertMichon I haven't had cycles to look into envtest. I think if we don't feel confident that these tests are improving AKS stability let's just close this PR. Alternatively, if we do think these tests offer some improved protection, then let's add a work item to investigate replacing these w/ envtest in the future. The tests themselves will be easy to remove in the future if we land on a better solution.

My 2¢

CecileRobertMichon:
My vote is to close this and work on adding a less costly test via local apiserver. This adds costs in time and $$ to every PR run (last AKS run took 1.5 hours) which isn't worth the additional coverage this brings IMO.

jackfrancis:
@CecileRobertMichon fair enough!

Issue #2903, opened 2022-12-05
Fix Ginkgo v2 deprecation warnings in e2e runs 


furkatgofurov7:
/kind bug

**What steps did you take and what happened:**
https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2773 uplifts Ginkgo to v2 and CI logs are full of mostly 2 types of deprecation warnings:

1. *Fixed* ~~[Measure()](https://onsi.github.io/ginkgo/MIGRATING_TO_V2#removed-measure)~~:
```
 You're using deprecated Ginkgo functionality:
=============================================
  Measure is deprecated and will be removed in Ginkgo V2.  Please migrate to gomega/gmeasure.
  Learn more at: https://onsi.github.io/ginkgo/MIGRATING_TO_V2#removed-measure
    /home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/test/e2e/conformance_test.go:99 
 To silence deprecations that can be silenced set the following environment variable:
  ACK_GINKGO_DEPRECATIONS=2.5.0 

```
Suggested migration strategy: https://onsi.github.io/ginkgo/MIGRATING_TO_V2#migration-strategy-1

2. [CurrentGinkgoTestDescription](https://onsi.github.io/ginkgo/MIGRATING_TO_V2#changed-currentginkgotestdescription)():
```
 You're using deprecated Ginkgo functionality:
=============================================
    CurrentGinkgoTestDescription() is deprecated in Ginkgo V2.  Use CurrentSpecReport() instead.
  Learn more at: https://onsi.github.io/ginkgo/MIGRATING_TO_V2#changed-currentginkgotestdescription
    /home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/test/e2e/helpers.go:421
    /home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/test/e2e/common.go:270
    /home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/test/e2e/common.go:273
    /home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/test/e2e/helpers.go:421
To silence deprecations that can be silenced set the following environment variable:
  ACK_GINKGO_DEPRECATIONS=2.5.0 

```
Suggested migration strategy: https://onsi.github.io/ginkgo/MIGRATING_TO_V2#migration-strategy-4

**What did you expect to happen:**
[...]

Issue #2265, opened 2022-04-27
conformance_test.go exit code to 0 event the non last test suite has failures


bingbing8:
/kind bug

[Before submitting an issue, have you checked the [Troubleshooting Guide](https://capz.sigs.k8s.io/topics/troubleshooting.html)?]

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]
./conformance_test.go

**What did you expect to happen:**
Normally there are 2 test suites (junit.e2e_suite.1.xml is the last) run. If the first one has failure but the second test suite (junit.e2e_suite.1.xml) passed. The exit code of conformance_test.go is 0. I expect the script exit with error to indicate there are failures in any of the test suites.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- cluster-api-provider-azure version: 1.2
- Kubernetes version: (use `kubectl version`): 1.23
- OS (e.g. from `/etc/os-release`): ws2022

jackfrancis:
@bingbing8 I'd like to ensure we address this. In your original report you mention 2 test suites, do you mean the one from upstream kubernetes and the one from capz?

CecileRobertMichon:
@jackfrancis (when you're back in the new year): here is an example of this in practice: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/2941/pull-cluster-api-provider-azure-windows-containerd-upstream-with-ci-artifacts/1603522577655926784

The github status reports success, even though the conformance suite had 1 failed spec.

<img width="905" alt="image" src="https://user-images.githubusercontent.com/8650424/208962824-9767cfe8-a533-4455-88f8-86805b2d773d.png">

jackfrancis:
After an initial investigation, it appears that the cluster-api is itself wrapping kubetest inside a container execution lifecycle. And the above test output suggests that the underlying container, though its runtime reported 1 error, did not return a non-zero exit code.

Specifically, I'm not seeing evidence that this conditional path was reached:

https://github.com/kubernetes-sigs/cluster-api/blob/v1.3.2/test/infrastructure/container/docker.go#L500-L502

If that had been reached, then the capz test runner would get this back:

https://github.com/kubernetes-sigs/cluster-api/blob/v1.3.2/test/framework/kubetest/run.go#L194-L196

Which would trigger this:

[...]

PR #2941, opened 2022-12-15
Update conformance test for Ginkgo v2

mboersma:
**What type of PR is this?**

/kind failing-test

In updating to Ginkgo v2 in #2773, we missed the fact that [`Measure` is deprecated](https://onsi.github.io/ginkgo/MIGRATING_TO_V2#removed-measure) and *will not run*. 🤦🏻

This refactors the conformance tests to use the new [`stopwatch`](https://onsi.github.io/gomega/#measuring-durations-with-codestopwatchcode) approach so they should actually run.

Refs #2903

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/conformance_test.go
+++ b/test/e2e/conformance_test.go
@@ -30,6 +30,7 @@ import (
 	"github.com/blang/semver"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/gmeasure"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -96,7 +97,10 @@ var _ = Describe("Conformance Tests", func() {
 		Expect(os.Setenv(ClusterIdentitySecretNamespace, namespace.Name)).To(Succeed())
 	})
 
-	Measure(specName, func(b Benchmarker) {
+	It(specName, func() {
+		experiment := gmeasure.NewExperiment(specName)
+		AddReportEntry(experiment.Name, experiment)
+
 		var err error
 
 		kubernetesVersion := e2eConfig.GetVariable(capi_e2e.KubernetesVersion)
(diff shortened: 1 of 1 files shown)


test/e2e/conformance_test.go
```diff
@@ -159,31 +163,30 @@ var _ = Describe("Conformance Tests", func() {
...
+				WorkerMachineCount:       pointer.Int64Ptr(linuxWorkerMachineCount),
+			},
+			WaitForClusterIntervals:      e2eConfig.GetIntervals(specName, "wait-cluster"),
+			WaitForControlPlaneIntervals: e2eConfig.GetIntervals(specName, "wait-control-plane"),
+			WaitForMachineDeployments:    e2eConfig.GetIntervals(specName, "wait-worker-nodes"),
+			ControlPlaneWaiters: clusterctl.ControlPlaneWaiters{
+				WaitForControlPlaneInitialized: EnsureControlPlaneInitialized,
+			},
+		}, result)
+		stopwatch.Record("cluster creation")
```
CecileRobertMichon, line 188: I don't see anything in output about "cluster creation" or "conformance suite", is that supposed to appear as output somewhere?

https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/2941/pull-cluster-api-provider-azure-conformance/1603444304154464256/build-log.txt

mboersma:
/retest-required

This change appears to have worked: conformance is no longer a no-op. The windows-containerd failure does not appear to be related.

test/e2e/conformance_test.go
mboersma: Yes, it's *supposed* to by virtue of the `AddReportEntry` function call: https://onsi.github.io/ginkgo/#attaching-data-to-reports

mboersma: I too see neither of those strings in the logs. 😕

mboersma: By searching for `conformance-tests`, I see where `AddReportEntry` was called at around line 500 in the logs, but I don't see the contents.

```
  INFO: Creating namespace capz-conf-w0d1ke
  INFO: Creating event watcher for namespace "capz-conf-w0d1ke"
  conformance-tests - /home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/test/e2e/conformance_test.go:102 @ 12/15/22 20:46:08.684
    conformance-tests
    Name | N | Min | Median | Mean | StdDev | Max
  INFO: Creating the workload cluster with name "capz-conf-w0d1ke" using the "(default)" template (Kubernetes v1.24.9, 1 control-plane machines, 2 worker machines)
```

 Maybe an `experiment` doesn't [implement `fmt.Stringer`](https://onsi.github.io/ginkgo/#controlling-output) (which would be rude since it's recommended in code examples) or maybe we need to log the report entries manually, in `AfterEach`? The tests haven't run yet, so there's nothing really to log, but this is the recommended pattern... I'll dig around more.

CecileRobertMichon: Since this is fixing the original issue (the tests don't run at all anymore) which is a lot higher priority to fix than logging the duration in the output (IMO), how about we merge this as-is and open an issue to investigate the logging?

test/e2e/data/kubetest/conformance-fast.yaml
```diff
@@ -2,7 +2,7 @@ ginkgo.focus: \[Conformance\]
 ginkgo.skip: \[Serial\]
 disable-log-dump: true
 ginkgo.progress: true
-ginkgo.slowSpecThreshold: 120.0
+ginkgo.slow-spec-threshold: 120s
```
CecileRobertMichon, line 5: we have ginkgo v2 specific files for k8s versions that run with v2 https://github.com/kubernetes-sigs/cluster-api-provider-azure/tree/0ba78caed74d57737fd41434f819e73160f18ce5/test/e2e/data/kubetest, did you mean to fix those?

test/e2e/conformance_test.go
mboersma: I think that's reasonable. Let me pare this back to the original commit and open an issue.

sonasingh46: +1

test/e2e/data/kubetest/conformance-fast.yaml
mboersma: I saw a warning about `--slowSpecThreshold` had changed to `--slow-spec-threshold` and grepped about trying to find it, which implied this change. But this actually caused an error because `--gingko.slow-spec-threshold` isn't understood, so I reverted it and will follow up on the warning in a different PR.

test/e2e/conformance_test.go
mboersma: See #2951

CecileRobertMichon:
/lgtm
/approve

FYI I noticed https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/2941/pull-cluster-api-provider-azure-windows-containerd-upstream-with-ci-artifacts/1603522577655926784 ran into #2265 but that's its own issue so let's merge this for now

PR #2947, opened 2022-12-15
Add CSI proxy and containerd logger to machine pool test templates

CecileRobertMichon:
**What type of PR is this?**
/kind failing-test

**What this PR does / why we need it**: Fixes a flake in the MachinePool test where CSI drivers fail due to missing CSI proxy.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #2922 

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note

```

PR #2950, opened 2022-12-15
E2E: verify daemonset pods after machines

### LINKED ISSUE #2933: E2E tests aren't properly verifying the DaemonSets are running
/kind bug

When introducing the Calico HelmChart the Windows kube-proxy was moved into the calico namespace causing the upstream tests to fail. The fix for kube-proxy is in https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2931 but this issue is to track fixing the e2e framework to verify Daemonsets.

```
Warning  FailedCreate  8m49s (x18 over 19m)  daemonset-controller  Error creating: pods "kube-proxy-windows-" is forbidden: error looking up service account calico-system/kube-proxy: serviceaccount "kube-proxy" not found
```

This wasn't caught since the test doesn't really verify the number of pods running in the DS. This isn't isolated to Windows but also happens to Linux daemonsets too.

https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/c0f81b0b89e6640ccd6c9eee35aadac05ec36417/test/e2e/helpers.go#L276-L279

you can see this in the logs of the e2e tests, notice the log doesn't have the number of pods running

```
[1mSTEP[0m: waiting for daemonset calico-system/calico-node to be complete
Nov 29 21:42:44.456: INFO: waiting for daemonset calico-system/calico-node to be complete
Nov 29 21:42:44.489: INFO: daemonset calico-system/calico-node pods are running, took 33.187202ms
[1mSTEP[0m: Waiting for Ready calico windows pods
[1mSTEP[0m: waiting for daemonset calico-system/calico-node-windows to be complete
Nov 29 21:42:44.659: INFO: waiting for daemonset calico-system/calico-node-windows to be complete
Nov 29 21:42:44.691: INFO: daemonset calico-system/calico-node-windows pods are running, took 32.541702ms
[1mSTEP[0m: waiting for daemonset calico-system/kube-proxy-windows to be complete
Nov 29 21:42:44.831: INFO: waiting for daemonset calico-system/kube-proxy-windows to be complete
Nov 29 21:42:44.864: INFO: daemonset calico-system/kube-proxy-windows pods are running, took 32.690035ms
``` 

DaemonSets should be running before the e2e tests run.

Fixed the kubel

**Environment:**

jackfrancis:
**What type of PR is this?**

/kind failing-test

This PR moves the "validate expected daemonset pods" after "verify control plane is ready", because at that point in the cluster creation flow machines are not actually online, and so daemonset pods are never going to be scheduled.

Instead, we move daemonset pod validation after `ApplyClusterTemplateAndWait`, which waits for control plane *and* worker machines. At this point in the E2E flow we will have all of our expected number of machines verified as Ready, at which point we can check for the expected daemonset pods.

Fixes #2933

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/azure_test.go
+++ b/test/e2e/azure_test.go
@@ -184,6 +184,15 @@ var _ = Describe("Workload cluster creation", func() {
 					ControlPlaneWaiters: clusterctl.ControlPlaneWaiters{
 						WaitForControlPlaneInitialized: EnsureControlPlaneInitialized,
 					},
+					PostMachinesProvisioned: func() {
+						EnsureDaemonsets(ctx, func() DaemonsetsSpecInput {
+							return DaemonsetsSpecInput{
+								BootstrapClusterProxy: bootstrapClusterProxy,
+								Namespace:             namespace,
+								ClusterName:           clusterName,
+							}
+						})
+					},
 				}, result)
 
 				By("Creating a private cluster from the management cluster", func() {
@@ -235,6 +244,15 @@ var _ = Describe("Workload cluster creation", func() {
 				WaitForMachineDeployments:    e2eConfig.GetIntervals(specName, "wait-worker-nodes"),
 				ControlPlaneWaiters: clusterctl.ControlPlaneWaiters{
 					WaitForControlPlaneInitialized: EnsureControlPlaneInitialized,
+				},
+				PostMachinesProvisioned: func() {
+					EnsureDaemonsets(ctx, func() DaemonsetsSpecInput {
+						return DaemonsetsSpecInput{
+							BootstrapClusterProxy: bootstrapClusterProxy,
+							Namespace:             namespace,
+							ClusterName:           clusterName,
+						}
+					})
 				},
 			}, result)
 
@@ -320,6 +338,15 @@ var _ = Describe("Workload cluster creation", func() {
 				ControlPlaneWaiters: clusterctl.ControlPlaneWaiters{
 					WaitForControlPlaneInitialized: EnsureControlPlaneInitialized,
 				},
+				PostMachinesProvisioned: func() {
+					EnsureDaemonsets(ctx, func() DaemonsetsSpecInput {
+						return DaemonsetsSpecInput{
+							BootstrapClusterProxy: bootstrapClusterProxy,
+							Namespace:             namespace,
+							ClusterName:           clusterName,
+						}
+					})
+				},
 			}, result)
 
 			By("Verifying expected VM extensions are present on the node", func() {
@@ -378,6 +405,15 @@ var _ = Describe("Workload cluster creation", func() {
 				WaitForMachinePools:          e2eConfig.GetIntervals(specName, "wait-machine-pool-nodes"),
 				ControlPlaneWaiters: clusterctl.ControlPlaneWaiters{
 					WaitForControlPlaneInitialized: EnsureControlPlaneInitialized,
+				},
+				PostMachinesProvisioned: func() {
+					EnsureDaemonsets(ctx, func() DaemonsetsSpecInput {
+						return DaemonsetsSpecInput{
+							BootstrapClusterProxy: bootstrapClusterProxy,
+							Namespace:             namespace,
+							ClusterName:           clusterName,
+						}
+					})
 				},
 			}, result)
 
@@ -464,6 +500,15 @@ var _ = Describe("Workload cluster creation", func() {
 				// The workaround is to use server side apply by passing `--server-side` flag to kubectl apply.
 				// More on server side apply here: https://kubernetes.io/docs/reference/using-api/server-side-apply/
 				Args: []string{"--server-side"},
+				PostMachinesProvisioned: func() {
+					EnsureDaemonsets(ctx, func() DaemonsetsSpecInput {
+						return DaemonsetsSpecInput{
+							BootstrapClusterProxy: bootstrapClusterProxy,
+							Namespace:             namespace,
+							ClusterName:           clusterName,
+						}
+					})
+				},
 			}, result)
 
 			By("Verifying expected VM extensions are present on the node", func() {
@@ -581,6 +626,15 @@ var _ = Describe("Workload cluster creation", func() {
 				WaitForMachineDeployments:    e2eConfig.GetIntervals(specName, "wait-worker-nodes"),
 				ControlPlaneWaiters: clusterctl.ControlPlaneWaiters{
 					WaitForControlPlaneInitialized: EnsureControlPlaneInitialized,
+				},
+				PostMachinesProvisioned: func() {
+					EnsureDaemonsets(ctx, func() DaemonsetsSpecInput {
+						return DaemonsetsSpecInput{
+							BootstrapClusterProxy: bootstrapClusterProxy,
+							Namespace:             namespace,
+							ClusterName:           clusterName,
+						}
+					})
 				},
 			}, result)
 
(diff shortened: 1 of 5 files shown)


test/e2e/azure_test.go
```diff
@@ -466,6 +506,16 @@ var _ = Describe("Workload cluster creation", func() {
 				Args: []string{"--server-side"},
 			}, result)
 
+			By("Verifying that expected Daemonset Pods are running", func() {
+				EnsureDaemonsets(ctx, func() DaemonsetsSpecInput {
```
CecileRobertMichon, line 510: are there any hooks in the e2e framework flow that can allow us to do it as part of create after the worker machines come up? Just like EnsureControlPlaneInitialized allows us to install the Helm charts after the control plane machines are up

test/e2e/azure_test.go
```diff
@@ -186,6 +186,16 @@ var _ = Describe("Workload cluster creation", func() {
 					},
 				}, result)
 
+				By("Verifying that expected Daemonset Pods are running", func() {
```
marosset, line 189: Can we do this with less code duplication?

jackfrancis:
@CecileRobertMichon do you prefer this updated use of the `PostMachinesProvisioned` func?

@marosset this doesn't really address your valid observation. I'm not entirely certain why we wrap this test cases in a func and then invoke the input vars "just in time", rather than just passing them into the func as values. I assume there is some async value mutation that we need to properly track and that these func closures enable. Maybe @nojnhuh has more context?

test/e2e/daemonsets.go
```diff
@@ -0,0 +1,102 @@
...
+
+// DaemonsetsSpecInput is the input for EnsureDaemonsets.
+type DaemonsetsSpecInput struct {
+	BootstrapClusterProxy framework.ClusterProxy
+	Namespace             *corev1.Namespace
+	ClusterName           string
+}
+
+// EnsureDaemonsets implements a test that verifies expected Daemonset Pods are running.
+func EnsureDaemonsets(ctx context.Context, inputGetter func() DaemonsetsSpecInput) {
```
CecileRobertMichon, line 41: why does this need to take an inputGetter as a func and not simply a `DaemonsetsSpecInput` as param?

CecileRobertMichon:
> @CecileRobertMichon do you prefer this updated use of the PostMachinesProvisioned func?

yes 100%. Although not sure why we need to pass in a func as param, commented on that.

@marosset has a good point, there is a lot of duplication. Perhaps we could refactor our uses of the clusterctl `ApplyClusterTemplateAndWait` with a wrapper helper func that adds all the common stuff like this by default and just takes the stuff that changes as parameters. That's probably out of scope here but would be a nice code cleanup.

jackfrancis:
Now that we're actually testing for the presence of the calico-node daemonset, it's failing three of the E2E test scenarios 🤯

test/e2e/azure_test.go
marosset: > > @CecileRobertMichon do you prefer this updated use of the PostMachinesProvisioned func?
> 
> yes 100%. Although not sure why we need to pass in a func as param, commented on that.
> 
> @marosset has a good point, there is a lot of duplication. Perhaps we could refactor our uses of the clusterctl `ApplyClusterTemplateAndWait` with a wrapper helper func that adds all the common stuff like this by default and just takes the stuff that changes as parameters. That's probably out of scope here but would be a nice code cleanup.

Maybe just leave a TODO comment?
I'm OK doing the cleanup later

CecileRobertMichon: Let's open an issue (TODO comments tend to get lost)

jackfrancis:
@jsturtevant @marosset see @CecileRobertMichon's comment above, I'm attempting to address that.

tl;dr 🤦, not 🤯.

jackfrancis:
It definitely seems like our VMSS cluster template is failing to launch Windows csi-node-driver pods. See:

https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/2950/pull-cluster-api-provider-azure-e2e/1611109298186752000/artifacts/clusters/capz-e2e-jfwxpp-vmss/kube-system/csi-azuredisk-node-win-6cc4v/azuredisk.log

@marosset @jsturtevant any ideas?

CecileRobertMichon:
> It definitely seems like our VMSS cluster template is failing to launch Windows csi-node-driver pods. See:

That seems like the same error I was trying to fix in #2947

CecileRobertMichon:
in theory rebase shouldn't solve it because prow should always rebase before running tests... so if it's still failing chances are #2947 didn't actually fix the issue

jackfrancis:
@CecileRobertMichon It's still failing, and I'm noticing it on other PRs as well, is something environmental (different k8s version?) responsible?

CecileRobertMichon:
looked at logs, it seems the template is missing the actual ClusterResourceSet for csi-proxy so even though the configmap is there it's not getting installed https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/2950/pull-cluster-api-provider-azure-e2e/1611408924651556864/artifacts/clusters/bootstrap/capz-e2e-1ace1l-vmss-cluster-template.yaml

jackfrancis:
@CecileRobertMichon indeed: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2992

test/e2e/daemonsets.go
jackfrancis: To answer this question definitively: I implemented this E2E helper func following the pattern of existing helper funcs.

Do either @nojnhuh or @mboersma know why this pattern exists?

jackfrancis:
/retest

(cluster delete timeout flake)

jackfrancis:
@CecileRobertMichon @marosset this is passing tests and ready for another review round

CecileRobertMichon (commented):
This lgtm and is equivalent to previous functionality (but better because we're actually waiting for machines now). One thought: have you considered listing all daemonsets and waiting for all of them to to be available instead hardcoding a select few we can about?

If we want to make sure the daemonsets for calico exist we could potentially leave the existing code waiting for them to be available in the install Calico func and then do a general "wait for all daemonsets" post node provisioning

test/e2e/daemonsets.go
jackfrancis: Summarizing a discussion: we concluded that this pattern is carried over from CAPI E2E; cycles permitting, we can investigate whether or not this is needed in all the places we've copied it in CAPZ E2E.

jackfrancis:
@CecileRobertMichon this generalized approach appears to work fine, you can see the outcomes if you search for "Waiting for all DaemonSet Pods to be Running" in E2E output.

mboersma (commented):
This looks good. Squash?

jackfrancis:
@mboersma yeah just wanna get one more eyeball on this approach so I can revert if necessary

CecileRobertMichon (commented):
lgtm

CecileRobertMichon:
what's up with all the flakes 👀  anything related to this change?

jackfrancis:
- make lint timeout
- default/web-windows7jo8s4 flake during LB IIS test
- HA test delete cluster timeout
- AKS E2E flake during autoscale test (addressed here: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3058)
- AKS E2E flake during public IP test (also addressed here: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3058)

I think that's it!

jackfrancis:
@CecileRobertMichon sorry, not related to this change, I've seen every one of those flakes on other PRs :(

PR #2965, opened 2022-12-21
Allow for configurable scope in system assigned identities

### LINKED ISSUE #2923: Allow configuring role assignment scope for system assigned VM identity
/kind bug

I recently added the identity: SystemAssigned property to my control plane and machine pool nodes. The resulting reconciliation is erroring our when it fails to assign contributor roles to the newly created system assigned managed identity.

Here is the error: 
E1210 07:33:33.144651 1 controller.go:326] "msg"="Reconciler error" "error"="failed to reconcile AzureMachine: failed to reconcile AzureMachine service roleassignments: cannot assign role to VirtualMachine system assigned identity: failed to create resource cluster/not-a-guid (service: roleassignments): authorization.RoleAssignmentsClient#Create: Failure responding to request: StatusCode=403 -- Original Error: autorest/azure: Service returned an error. Status=403 Code=\"AuthorizationFailed\" Message=\"The client 'not-a-guid' with object id 'not-a-guid' does not have authorization to perform action 'Microsoft.Authorization/roleAssignments/write' over scope '/subscriptions/not-a-subscription-guid/providers/Microsoft.Authorization/roleAssignments/not-a-guid' or the scope is invalid. If access was recently granted, please refresh your credentials.\"" "azureMachine"={"name":"machine-name","namespace":"some-namespace"} "controller"="azuremachine" "controllerGroup"="infrastructure.cluster.x-k8s.io" "controllerKind"="AzureMachine" "name"="machine-name" "namespace"="some-namespace" "reconcileID"="not-a-guid"

System Assigned managed identity to be turned on with no roles being granted. If I wanted my control plane, worker nodes, or machine pool to have contributor roles, I would assign those to an Azure Identity and do a UserAssigned identity to help manage access and sprawl. In no production environment would you want all of your nodes to have contributor access to your subscription. 

**Environment:**

- cluster-api-provider-azure version: 1.6.0
- Kubernetes version: (use `kubectl version`):  1.23.5
- OS (e.g. from `/etc/os-release`): Ubuntu 20.04.4 LTS

willie-yao:
**What type of PR is this?**
/kind feature

Currently, `SystemAssigned` identities automatically assign the scope of the identity to the entire subscription, as well as the contributor role for the role assignment. This PR allows the scope and role to be configurable via the `AzureMachineTemplate` spec.

Fixes #2923 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Allow configurable scope and role assignment for SystemAssigned identities
```


azure/scope/machine.go
```diff
@@ -331,6 +331,22 @@ func (m *MachineScope) HasSystemAssignedIdentity() bool {
...
+	}
+	return azure.GenerateSubscriptionScope(m.SubscriptionID())
+}
+
+// GetRoleDefinitionID returns the specified role definition ID, and generates a contributor role definition ID if none is specified.
+func (m *MachineScope) GetRoleDefinitionID() string {
+	if m.AzureMachine.Spec.RoleDefinitionID != "" {
+		return m.AzureMachine.Spec.RoleDefinitionID
+	}
+	return azure.GenerateContributorRoleDefinitionID(m.SubscriptionID())
```
CecileRobertMichon, line 347: now that the field is exposed via the API this should get defaulted via the defaulting webhook so the default value is persisted and visible by users (same with scope)

willie-yao: I'm having some trouble setting the default inside the webhook, since the scope and role definition ID requires the machine scope's subscription ID. I'm not sure if the AzureMachine webhook would have access to the subscription ID.

CecileRobertMichon: Good point. The subscriptionID is part of the AzureCluster Spec, we'd have to fetch the owner AzureCluster object to be able to get to it. There is a way to do this by adding a Client to the webhook that can be used to fetch other objects in the k8s cluster (like https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/3f4543be33a1044a8fbb6746557baa4dd240e908/exp/api/v1beta1/azuremanagedmachinepool_webhook.go#L46) but it would require a bit more work to change the webhook for AzureMachine and AzureMachinePool since those aren't currently set up with a Client.

CecileRobertMichon:
@willie-yao fyi if you're going to implement the same webhook changes for MachinePool, @mboersma had to implement the same changes to add the Client in #999, you might be able to reuse his work

willie-yao:
@CecileRobertMichon is the PR getting merged soon? If so, I can wait for that and then rebase my branch on top of those changes!

CecileRobertMichon:
@willie-yao yes ETA end of day today

azure/scope/machine.go
willie-yao: After hacking away at this, I discovered that the machinepool is unable to retrieve the cluster name of the owner cluster when the default webhook is called. On this [line](https://github.com/kubernetes-sigs/cluster-api-provider-azure/commit/165a0d1653e3092b040c13f1939bfdb2534138c2#diff-435f85313e217cd2c7e0569bdace226dcb57ba2210dff964a2e511193010f809R291), the labels are empty. We were able to pin down the reason to the machine pool being declared and created directly from the spec, so the labels in [CAPI](https://github.com/willie-yao/cluster-api/blob/dff5c1a05a8aca1905b8a6bd8574f80f61f5f610/exp/internal/controllers/machinepool_controller_phases.go#L138) are not being set by the time the defaulting webhook is called. This is not the case for machines, as the machines are created by via a machine deployment and declared through a template.

Is there still a way to retrieve the clustername without using the labels field? Or is there a way to wait for the labels field to be populated. We were concerned about blocking or waiting in a webhook function though, so I'm not sure if there's a way around this. @CecileRobertMichon any thoughts?

CecileRobertMichon: MachinePool.Spec has a required `clusterName` field, are you able to use that?

willie-yao: I considered this option and I'm looking into how to get the MachinePool from the AzureMachinePool webhook. I don't have access to the machinepool scope or the name of the MachinePool from the webhook. I'm definitely just missing something though.

CecileRobertMichon: I think you might be able to use GetOwnerMachinePool (https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/92b41e932d17d416fc3c62831c8161337dd4d69c/controllers/helpers.go#L518) to get the MachinePool from the AzureMachinePool and then get the Cluster name from there

docs/book/src/topics/vm-identity.md
```diff
@@ -104,6 +104,23 @@ spec:
...
+kind: AzureMachineTemplate
+metadata:
+  name: ${CLUSTER_NAME}-md-0
+  namespace: default
+spec:
+  template:
+    spec:
+      identity: SystemAssigned
+      scope: /subscriptions/${AZURE_SUBSCRIPTION_ID}/resourceGroups/${RESOURCE_GROUP_NAME}
+      roleAssignmentID: $/subscriptions/${AZURE_SUBSCRIPTION_ID}/providers/Microsoft.Authorization/roleDefinitions/8e3af657-a8ff-443c-a75c-2fe8c4bcb635
```
primeroz, line 120: is this `roleDefinitionID` rhat than `roleAssignmentID` ?

exp/api/v1beta1/azuremachinepool_types.go
```diff
@@ -135,6 +135,16 @@ type (
 		// +optional
 		Identity infrav1.VMIdentity `json:"identity,omitempty"`
 
+		// Scope is the scope that the role assignment should be created at. If left empty, will default to the
+		// subscription scope.
+		// +optional
+		Scope string `json:"scope,omitempty"`
```
primeroz, line 141: Should this Field `Scope` and the other new one `RoleDefinitionID` have a name that includes somehow their relationship to the `SystemAssigned` identity  ? 

something like `SystemAssignedIdentityScope` and `SystemAssignedIdentityRoleDefinitionID` would probably be too verbose ... but something that helps understand what they refer to ?

docs/book/src/topics/vm-identity.md
willie-yao: This is definitely a mistake, thanks for catching this!

exp/api/v1beta1/azuremachinepool_types.go
willie-yao: I do agree that it would be too verbose. Maybe I can add a more descriptive comment. I am also planning on adding a check in the validating webhook to ensure the `identity` field is being set if `Scope` and/or `RoleDefinitionID` is specified.

@CecileRobertMichon any thoughts on how to name this better?

CecileRobertMichon: Needing to prefix a bunch of fields with the same value is usually a sign that the API isn't properly structured. It does seem like the right thing to do here would be to re-organize the API to something like:

```
spec
...
identity:
  type: SystemAssigned
  scope: ...
  role: ...
  roleAssignmentName: ...
``` 

instead of having a bunch of identity-related fields in top level spec. Unfortunately that would be a breaking API change so we would only be able to do this if we bump the API version. Would be good to open an issue in any case so we can flag it as something we'd want to do in vnext.

For now maybe at least combining these two new fields into a struct like this would be a good first step:

```
systemAssignedIdentityRole:
  definitionID: // RoleDefinitionID
  scope: // Scope
  name: // RoleAssignmentName (note: this would be breaking, could omit for now or make it backwards compatible)
```

api/v1beta1/azuremachine_default.go
```diff
@@ -79,11 +83,17 @@ func (s *AzureMachineSpec) SetDataDisksDefaults() {
...
 
 // SetIdentityDefaults sets the defaults for VM Identity.
-func (s *AzureMachineSpec) SetIdentityDefaults() {
+func (s *AzureMachineSpec) SetIdentityDefaults(subscriptionID string) {
 	if s.Identity == VMIdentitySystemAssigned {
 		if s.RoleAssignmentName == "" {
 			s.RoleAssignmentName = string(uuid.NewUUID())
 		}
+		if s.Scope == "" {
+			s.Scope = fmt.Sprintf("/subscriptions/%s/", subscriptionID)
```
primeroz, line 92: would it make sense to use the `azure.GenerateSubscriptionScope(<SubscriptionID>)` function here ( and in the azuremachinepool_default.go) and the `azure.GenerateContributorRoleDefinitionID(<SubscriptionID>)` for the RoleDefinition (from `azure/defaults.go` ) rather than redefining the same `printf` ? 

Alternatively I guess we can remove those functions from defaults since they are only used for this purpose ( at the moment only in the `azure/scope/machine.go` )

willie-yao: It would make sense! However, we can't import from the `azure` package from `api`, so I'll think about either moving the functions or removing them entirely

api/v1beta1/azuremachine_types.go
```diff
@@ -149,6 +153,24 @@ type SpotVMOptions struct {
...
+type SystemAssignedIdentityRole struct {
+	// DefinitionID is the ID of the role definition to create for a system assigned identity. It can be an Azure built-in role or a custom role.
+	// Refer to built-in roles: https://learn.microsoft.com/en-us/azure/role-based-access-control/built-in-roles
+	// +optional
+	DefinitionID string `json:"definitionID,omitempty"`
+
+	// Name is the name of the role assignment to create for a system assigned identity. It can be any valid GUID.
+	// If not specified, a random GUID will be generated.
+	// +optional
+	Name string `json:"name,omitempty"`
```
CecileRobertMichon, line 166: does this replace `RoleAssignmentName`? if so we should mark it as deprecated

azure/services/roleassignments/roleassignments.go
```diff
@@ -102,6 +103,7 @@ func (s *Service) Reconcile(ctx context.Context) error {
 	}
 
 	for _, roleAssignmentSpec := range s.Scope.RoleAssignmentSpecs(principalID) {
+		fmt.Printf("roleAssignmentSpec: %+v", roleAssignmentSpec)
```
CecileRobertMichon, line 106: is this a leftover debug statement?

exp/api/v1beta1/azuremachinepool_default.go
```diff
@@ -102,3 +139,35 @@ func (amp *AzureMachinePool) SetNetworkInterfacesDefaults() {
 		}
 	}
 }
+
+// findParentMachinePool finds the parent MachinePool for the AzureMachinePool.
+func (amp *AzureMachinePool) findParentMachinePool(cli client.Client) (*expv1.MachinePool, error) {
```
CecileRobertMichon, line 144: did you check if there were any existing helpers in CAPI or CAPZ that do this? If not, seems like a good bit of reusable functionality, might make more sense to place it in controller helpers or utils instead of unexported in the webhook file

exp/api/v1beta1/azuremachinepool_default.go
```diff
@@ -102,3 +139,35 @@ func (amp *AzureMachinePool) SetNetworkInterfacesDefaults() {
...
+	for _, mp := range machinePoolList.Items {
+		if mp.Spec.Template.Spec.InfrastructureRef.Name == amp.Name {
+			return &mp, nil
+		}
+	}
+	return nil, errors.Errorf("failed to get MachinePool %s/%s", amp.Namespace, amp.Name)
+}
+
+// getSubscriptionID returns the subscription ID for the given AzureMachinePool.
+func (amp *AzureMachinePool) getSubscriptionID(cli client.Client, clusterName string) (string, error) {
```
CecileRobertMichon, line 159: seems like this could also be shared/reused between AzureMachine and MachinePool (we'd just need to pass in the namespace too)

exp/api/v1beta1/azuremachinepool_types.go
```diff
@@ -135,6 +135,10 @@ type (
 		// +optional
 		Identity infrav1.VMIdentity `json:"identity,omitempty"`
 
+		// SystemAssignedIdentityRole defines the role and scope to assign to the system assigned identity.
+		// +optional
+		SystemAssignedIdentityRole *infrav1.SystemAssignedIdentityRole `json:"systemAssignedIdentityRole,omitempty"`
```
CecileRobertMichon, line 140: same comment here about marking `RoleAssignmentName` as deprecated

exp/api/v1beta1/azuremachinepool_webhook.go
```diff
@@ -203,6 +199,22 @@ func (amp *AzureMachinePool) ValidateSystemAssignedIdentity(old runtime.Object)
 	}
 }
 
+// ValidateSystemAssignedIdentityRole validates the scope and roleDefinitionID for the system-assigned identity.
+func (amp *AzureMachinePool) ValidateSystemAssignedIdentityRole() error {
```
CecileRobertMichon, line 203: same comments about validating name isn't set in both places here - consider how we might refactor this to share validation code with AzureMachine webhook

CecileRobertMichon:
This looks great @willie-yao, just some comments here and there around code reuse and handling deprecation of role assignment name. Please squash commits when you're ready!

api/v1beta1/azuremachine_types.go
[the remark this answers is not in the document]
willie-yao: Whoops, I didn't mean to keep this in my commit. I was playing around with the idea of adding `RoleAssignmentName` to the struct, but decided against it as it's a breaking change. Do you think we should include this field?

api/v1beta1/azuremachine_validation.go
```diff
@@ -120,8 +124,19 @@ func ValidateSystemAssignedIdentity(identityType VMIdentity, oldIdentity, newIde
 func ValidateUserAssignedIdentity(identityType VMIdentity, userAssignedIdenteties []UserAssignedIdentity, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 
-	if identityType == VMIdentityUserAssigned && len(userAssignedIdenteties) == 0 {
-		allErrs = append(allErrs, field.Required(fldPath, "must be specified for the 'UserAssigned' identity type"))
+	if identityType == VMIdentityUserAssigned {
+		if len(userAssignedIdenteties) == 0 {
```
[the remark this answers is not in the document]
willie-yao: Nope, not sure how this happened. Thanks for catching this!

api/v1beta1/azuremachine_default.go
```diff
@@ -143,15 +158,43 @@ func (s *AzureMachineSpec) SetNetworkInterfacesDefaults() {
...
-	s.SetDefaultCachingType()
-	s.SetDataDisksDefaults()
-	s.SetIdentityDefaults()
-	s.SetSpotEvictionPolicyDefaults()
-	s.SetDiagnosticsDefaults()
-	s.SetNetworkInterfacesDefaults()
+
+	subscriptionID, err := m.getSubscriptionID(client)
+	if err != nil {
+		ctrl.Log.WithName("SetDefault").Error(err, "getSubscriptionID failed")
```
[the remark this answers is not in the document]
willie-yao: I observed above this line that the default SSH Public key uses the same error handling pattern. The default webhook function that calls `SetDefaults` does not seem to return an error. Should I handle this one differently?

willie-yao:
@CecileRobertMichon Thanks for the review! I realized that I accidentally left the `Name` field in `SystemAssignedIdentityRole`. This was unintended and unused, and I meant to continue using the existing `RoleAssignmentName` field. Is this something I should keep, or should I refactor the Name to be used in `SystemAssignedIdentityRole`? I'm afraid of doing so since it's a breaking api change.

exp/api/v1beta1/azuremachinepool_default.go
willie-yao: Looks like there is actually an [existing helper](https://github.com/willie-yao/cluster-api-provider-azure/blob/a8c304ba616fc1c7306179e1e958bf7f6bb89656/controllers/helpers.go#L129) in the controller for getting the owner cluster name. However, I'm unable to import from the controller package from within exp v1beta1 (I'm not sure why). I'll see if there's another way for me to reuse the function. If not, I'll move the function to somewhere reusable!

CecileRobertMichon:
> Is this something I should keep, or should I refactor the Name to be used in SystemAssignedIdentityRole? I'm afraid of doing so since it's a breaking api change.

You have 2 ways forward: 

- include it and leave the existing field as "deprecated" (this means mark the field as deprecated and "moving" the existing value to the new field to handle migrating existing clusters, so it's not breaking). An example of us doing this previously in code is `AzureMachine.Spec.SubnetName` 
- or leave `RoleAssignmentName` as is without moving it to the new spec

2) is less code and less work but a worse UX as it results in the role-related fields being split up. The impact to UX is relatively low in this particular case though because we don't expect users to actually set `RoleAssignmentName`, it's mostly meant to be auto-generated by the webhook (it's just in spec because it needs to be persisted and can't change).

willie-yao:
Option 1 definitely seems like the way to go long-term. I'll work on implementing it!

api/v1beta1/azuremachine_default.go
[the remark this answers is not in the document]
CecileRobertMichon: I think returning an error is preferable whenever that's possible

api/v1beta1/azuremachine_default.go
```diff
@@ -79,10 +84,32 @@ func (s *AzureMachineSpec) SetDataDisksDefaults() {
...
+	}
 	if s.Identity == VMIdentitySystemAssigned {
-		if s.RoleAssignmentName == "" {
-			s.RoleAssignmentName = string(uuid.NewUUID())
+		if s.SystemAssignedIdentityRole == nil {
+			s.SystemAssignedIdentityRole = &SystemAssignedIdentityRole{}
+		}
+		if s.SystemAssignedIdentityRole.Name == "" {
+			// Default role name to the role assignment name.
+			s.SystemAssignedIdentityRole.Name = string(uuid.NewUUID())
```
CecileRobertMichon, line 100: super small nit

this could be 

``` 
if s.RoleAssignmentName != "" {
  // Move the existing value from the deprecated RoleAssignmentName field.
  s.SystemAssignedIdentityRole.Name = s.RoleAssignmentName
  s.RoleAssignmentName = ""
} else {
  // Default role name to the role assignment name.
  s.SystemAssignedIdentityRole.Name = string(uuid.NewUUID())
}
``` 

so it's clear that it's one or the other and we avoid running `uuid.NewUUID()` when not needed

api/v1beta1/azuremachine_default.go
```diff
@@ -79,11 +83,22 @@ func (s *AzureMachineSpec) SetDataDisksDefaults() {
...
+		if s.SystemAssignedIdentityRole == nil {
+			s.SystemAssignedIdentityRole = &SystemAssignedIdentityRole{}
+		}
+		if s.SystemAssignedIdentityRole.Scope == "" {
+			// Default scope to the subscription.
+			s.SystemAssignedIdentityRole.Scope = fmt.Sprintf("/subscriptions/%s/", subscriptionID)
+		}
+		if s.SystemAssignedIdentityRole.DefinitionID == "" {
+			// Default role definition ID to Contributor role.
+			s.SystemAssignedIdentityRole.DefinitionID = fmt.Sprintf("/subscriptions/%s/providers/Microsoft.Authorization/roleDefinitions/b24988ac-6180-42a0-ab88-20f7382dd24c", subscriptionID)
```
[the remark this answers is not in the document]
CecileRobertMichon: this one is still TODO

api/v1beta1/azuremachine_default.go
```diff
@@ -143,15 +170,44 @@ func (s *AzureMachineSpec) SetNetworkInterfacesDefaults() {
...
+func GetSubscriptionID(cli client.Client, clusterName string, namespace string) (string, error) {
+	ctx := context.Background()
+
+	ownerCluster := &AzureCluster{}
+	key := client.ObjectKey{
+		Namespace: namespace,
+		Name:      clusterName,
+	}
+
+	if err := cli.Get(ctx, key, ownerCluster); err != nil {
```
CecileRobertMichon, line 183: I wonder if we should add some retry mechanism here based on @mboersma's findings in #3080

api/v1beta1/azuremachine_validation.go
```diff
@@ -117,15 +121,27 @@ func ValidateSystemAssignedIdentity(identityType VMIdentity, oldIdentity, newIde
...
+	if identityType == VMIdentityUserAssigned && len(userAssignedIdentities) == 0 {
 		allErrs = append(allErrs, field.Required(fldPath, "must be specified for the 'UserAssigned' identity type"))
 	}
 	return allErrs
 }
 
+// ValidateSystemAssignedIdentityRole validates the system-assigned identity role.
+func ValidateSystemAssignedIdentityRole(identityType VMIdentity, roleAssignmentName string, role *SystemAssignedIdentityRole, fldPath *field.Path) field.ErrorList {
+	var allErrs field.ErrorList
+	if roleAssignmentName != "" && role.Name != "" {
```
CecileRobertMichon, line 136: do we have a unit test for this?

exp/api/v1beta1/azuremachinepool_default.go
```diff
@@ -18,13 +18,38 @@ package v1beta1
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+// SetDefaults sets the default values for an AzureMachinePool.
+func (amp *AzureMachinePool) SetDefaults(client client.Client) {
+	if err := amp.SetDefaultSSHPublicKey(); err != nil {
+		ctrl.Log.WithName("AzureMachinePoolLogger").Error(err, "SetDefaultSshPublicKey failed")
+	}
+
+	machinePool, err := azureutil.FindParentMachinePool(amp.Name, client)
```
CecileRobertMichon, line 38: same comment about retrying, although I think it might be easier to put the retry logic somewhere reusable https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3080/files#r1087264640

api/v1beta1/azuremachine_default.go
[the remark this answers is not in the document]
willie-yao: The parent `Default` function actually doesn't return an error. I refactored the `SetDefault` function to return an error. Where should I log the returned error after it is passed to the webhook `Default` function?

api/v1beta1/azuremachine_default.go
willie-yao: I tried this before which resulted in the machine controller's ownerRef not being set. I was very confused as to why, and I thought it might have to do with the defaulting webhook being called multiple times, and the Role Assignment Name is overwritten because we aren't checking if `s.SystemAssignedIdentityRole.Name == ""`. Would it make sense to use this structure, but with an extra check for that?

api/v1beta1/azuremachine_default.go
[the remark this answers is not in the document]
willie-yao: @CecileRobertMichon After making this change, it looks like the default function is returning before some of the other necessary defaults are set, particularly during unit tests that reference the default function. I did not observe any other Default webhook functions returning errors. Is there a standard for handling errors within webhooks? 

For example, the [`AzureMachineTemplate` webhook](https://github.com/willie-yao/cluster-api-provider-azure/blob/28eaa3755e32b6aa226a7e1a9d5baa99d0c399b3/api/v1beta1/azuremachinetemplate_webhook.go#L148) uses the `ctrl.log` pattern and doesn't return immediately if an error occurs.

api/v1beta1/azuremachine_default.go
CecileRobertMichon: yes the way you have it now looks good

api/v1beta1/azuremachine_default.go
```diff
@@ -143,15 +173,54 @@ func (s *AzureMachineSpec) SetNetworkInterfacesDefaults() {
...
+	key := client.ObjectKey{
+		Namespace: namespace,
+		Name:      clusterName,
+	}
+
+	for i := 1; ; i++ {
+		err := cli.Get(ctx, key, ownerCluster)
+		if err != nil {
+			if i > maxAttempts {
+				return "", errors.Wrap(err, "failed to find parent MachinePool")
```
CecileRobertMichon, line 190: wrong error wrap?

api/v1beta1/azuremachine_default.go
```diff
@@ -143,15 +173,54 @@ func (s *AzureMachineSpec) SetNetworkInterfacesDefaults() {
 	}
 }
 
+// GetSubscriptionID returns the subscription ID for the given cluster name and namespace.
+func GetSubscriptionID(cli client.Client, clusterName string, namespace string, maxAttempts int) (string, error) {
```
CecileRobertMichon, line 177: should we rename this to withRetry and move it to `util/azure` so it's consistent?

api/v1beta1/azuremachine_default.go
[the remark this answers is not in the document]
CecileRobertMichon: @willie-yao I think I missed this part before

> The default webhook function that calls SetDefaults does not seem to return an error

That's a tricky one because we don't want to prevent the other defaults from being set, in order to handle a potential failure we could best effort default (log error if error occurs but proceed with the rest of the func) then defer to the validation webhook to enforce that `getSubscriptionID` is set and fail if it isn't.

CecileRobertMichon (commented):
/assign @mboersma @sonasingh46 

for another pair of 👀

api/v1beta1/azuremachine_default.go
willie-yao: I tried moving this to `util/azure`, but that causes an import cycle due to the function needing to import `AzureCluster` from `v1beta1`. I'm not sure if there's a way around that with Go.

api/v1beta1/azuremachine_default.go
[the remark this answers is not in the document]
willie-yao: I think that's a better way to go about it. I noticed that this prevented other defaults from being set too, so I'll go ahead and defer to the validation webhook.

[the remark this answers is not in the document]
willie-yao: I made this [change](https://github.com/willie-yao/cluster-api-provider-azure/blob/7fd4f0ee12202517d5ec41a53083eba6fedfde66/api/v1beta1/azuremachine_validation.go#L140) which checks for a double slash `//` in the scope/definitionID, which would signify a subscription ID was not found. Is this a valid way to do this? I'm unable to call `GetSubscriptionID` in the validation webhook due to the absence of the client.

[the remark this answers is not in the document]
CecileRobertMichon: How about leaving role definitionID and scope empty in `SetIdentityDefaults` if `subscriptionID` is `""`, and then simply checking those are set in the validation webhook? Checking for `//` should catch this in theory but it could lead to edge cases since that field could also be set by a user. The other thing which is arguably better is we could do is add regex validation to make sure both scope and role definitionID have valid formats. For this you could reuse `azureautorest.ParseResourceID` (although it's getting deprecated #2974, @mboersma do you know if there is an alternative?), since both are in Azure resource ID format. That would also allow us to validate user input when a user passes in a custom role.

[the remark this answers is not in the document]
willie-yao: Hmm, looks like `azureautorest.ParseResourceID` actually returns an error for valid scope and roleDefinitionIDs, I'm trying to figure out what part of the regex doesn't match. In the meantime, I'll look for an alternative

CecileRobertMichon (commented):
/lgtm

CecileRobertMichon (commented):
@sonasingh46 @mboersma could one of you please review this one when you get a chance?

api/v1beta1/azuremachine_default.go
```diff
@@ -143,15 +174,52 @@ func (s *AzureMachineSpec) SetNetworkInterfacesDefaults() {
...
-	}
-	s.SetDefaultCachingType()
-	s.SetDataDisksDefaults()
-	s.SetIdentityDefaults()
-	s.SetSpotEvictionPolicyDefaults()
-	s.SetDiagnosticsDefaults()
-	s.SetNetworkInterfacesDefaults()
+func (m *AzureMachine) SetDefaults(client client.Client) {
+	if err := m.Spec.SetDefaultSSHPublicKey(); err != nil {
+		ctrl.Log.WithName("SetDefault").Error(err, "failed to set default SSH public key")
```
mboersma, line 205: Is it reasonable to have unit tests for `GetSubscriptionID` and `SetDefaults` on their own? They are probably invoked indirectly through `TestAzureMachineSpec_SetIdentityDefaults`, but I don't see test cases that hit these error return codepaths, for example.

api/v1beta1/azuremachine_default.go
```diff
@@ -79,10 +88,32 @@ func (s *AzureMachineSpec) SetDataDisksDefaults() {
...
-			s.RoleAssignmentName = string(uuid.NewUUID())
+		if s.SystemAssignedIdentityRole == nil {
+			s.SystemAssignedIdentityRole = &SystemAssignedIdentityRole{}
+		}
+		if s.RoleAssignmentName != "" {
+			// Move the existing value from the deprecated RoleAssignmentName field.
+			s.SystemAssignedIdentityRole.Name = s.RoleAssignmentName
+			s.RoleAssignmentName = ""
+		} else if s.SystemAssignedIdentityRole.Name == "" {
+			// Default role name to the role assignment name.
```
mboersma, line 107: I'm confused by this comment, because I guess it refers to an implied action: leaving `s.RoleAssignmentName` alone. But the next line is assigning a new UUID to a similar field... Maybe we can add something to the comment like `and assign a UUID`?

Or maybe I'm misunderstanding entirely. :-)

api/v1beta1/azuremachine_types.go
```diff
@@ -66,8 +66,11 @@ type AzureMachineSpec struct {
 	// +optional
 	UserAssignedIdentities []UserAssignedIdentity `json:"userAssignedIdentities,omitempty"`
 
-	// RoleAssignmentName is the name of the role assignment to create for a system assigned identity. It can be any valid GUID.
-	// If not specified, a random GUID will be generated.
+	// SystemAssignedIdentityRole defines the role and scope to assign to the system assigned identity.
```
mboersma, line 69: Nit: it should really be "system-assigned identity" with a hyphen, although Azure itself is not consistent. No need to update, I just can't refrain from pointing this out.

api/v1beta1/azuremachine_types.go
```diff
@@ -149,6 +152,24 @@ type SpotVMOptions struct {
 	EvictionPolicy *SpotEvictionPolicy `json:"evictionPolicy,omitempty"`
 }
 
+// SystemAssignedIdentityRole defines the role and scope to assign to the system assigned identity.
+type SystemAssignedIdentityRole struct {
+	// Name is the name of the role assignment to create for a system assigned identity. It can be any valid GUID.
```
mboersma, line 157: Should we say `UUID` rather than `GUID` here and on the next line? The two are really the same, but `GUID` is bit more Microsoft-specific, and the code that populates it calls `newUUID()`.

api/v1beta1/azuremachine_validation.go
```diff
@@ -117,15 +121,35 @@ func ValidateSystemAssignedIdentity(identityType VMIdentity, oldIdentity, newIde
...
 
+// ValidateSystemAssignedIdentityRole validates the system-assigned identity role.
+func ValidateSystemAssignedIdentityRole(identityType VMIdentity, roleAssignmentName string, role *SystemAssignedIdentityRole, fldPath *field.Path) field.ErrorList {
+	var allErrs field.ErrorList
+	if roleAssignmentName != "" && role != nil && role.Name != "" {
+		allErrs = append(allErrs, field.Invalid(fldPath, role.Name, "cannot set both roleAssignmentName and systemAssignedIdentityRole.name"))
+	}
+	if identityType == VMIdentitySystemAssigned {
+		if role.DefinitionID == "" {
+			allErrs = append(allErrs, field.Invalid(field.NewPath("Spec", "SystemAssignedIdentityRole", "DefinitionID"), role.DefinitionID, "the roleDefinitionID field cannot be empty"))
```
mboersma, line 141:
```suggestion
			allErrs = append(allErrs, field.Invalid(field.NewPath("Spec", "SystemAssignedIdentityRole", "DefinitionID"), role.DefinitionID, "the definitionID field cannot be empty"))
```

exp/api/v1beta1/azuremachinepool_default.go
```diff
@@ -18,13 +18,38 @@ package v1beta1
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+// SetDefaults sets the default values for an AzureMachinePool.
+func (amp *AzureMachinePool) SetDefaults(client client.Client) {
+	if err := amp.SetDefaultSSHPublicKey(); err != nil {
+		ctrl.Log.WithName("AzureMachinePoolLogger").Error(err, "SetDefaultSshPublicKey failed")
+	}
+
+	machinePool, err := azureutil.FindParentMachinePoolWithRetry(amp.Name, client, 5)
```
mboersma, line 38: (Thanks again for letting me hijack this utility code!)

mboersma (commented):
Looks good, I just had some minor questions and comments.

api/v1beta1/azuremachine_types.go
```diff
@@ -149,6 +152,24 @@ type SpotVMOptions struct {
...
+	// Refer to built-in roles: https://learn.microsoft.com/en-us/azure/role-based-access-control/built-in-roles
+	// +optional
+	DefinitionID string `json:"definitionID,omitempty"`
+
+	// Scope is the scope that the role assignment or definition applies to. The scope can be any REST resource instance.
+	// If not specified, the scope will be the subscription.
+	// +optional
+	Scope string `json:"scope,omitempty"`
+}
+
```
sonasingh46, line 172: Are these fields immutable?

sonasingh46: Got the answer on slack. These are immutable. 
https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2965/files#diff-172ee0e7e00b5887e6feb73c83eee87b8b9a1315e891ee454c4f1f4d9ee9f64fR73

api/v1beta1/azuremachine_default.go
```diff
@@ -79,10 +88,32 @@ func (s *AzureMachineSpec) SetDataDisksDefaults() {
...
 	if s.Identity == VMIdentitySystemAssigned {
-		if s.RoleAssignmentName == "" {
-			s.RoleAssignmentName = string(uuid.NewUUID())
+		if s.SystemAssignedIdentityRole == nil {
+			s.SystemAssignedIdentityRole = &SystemAssignedIdentityRole{}
+		}
+		if s.RoleAssignmentName != "" {
+			// Move the existing value from the deprecated RoleAssignmentName field.
+			s.SystemAssignedIdentityRole.Name = s.RoleAssignmentName
+			s.RoleAssignmentName = ""
```
sonasingh46, line 105: RoleAssignment name is immutable? Shall we still do this?

api/v1beta1/azuremachine_default.go
willie-yao: This comment is definitely misleading. Since we're deprecating `s.RoleAssignmentName`, we are copying the name over to the new `s.SystemAssignedIdentityRole` field. Otherwise if the name field is empty, we want to assign a UUID. I'll go ahead and fix the comment!

api/v1beta1/azuremachine_types.go
willie-yao: UUID is definitely the correct term here

sonasingh46 (approved):
/lgtm

api/v1beta1/azuremachine_default.go
willie-yao: Answered on slack. The role assignment name field is deprecated and is moved to the new field in `s.SystemAssignedIdentityRole` in the defaulting webhook.
https://kubernetes.slack.com/archives/CEX9HENG7/p1676312402769689?thread_ts=1676311400.223999&cid=CEX9HENG7

api/v1beta1/azuremachine_default.go
willie-yao: Yes, will be adding those unit tests now!

willie-yao: I think that a unit test for `GetSubscriptionID` will be sufficient. `SetDefaults` calls all of the other defaulting functions, and the only error left untested is fetching the clustername from the labels. Also, the error is logged and not returned which would make it tricky to test.

CecileRobertMichon (commented):
/lgtm
/approve

Issue #2972, opened 2022-12-29
Changing vmSize/kubernetesVersion on a MachinePool only refresh one node in the VMSS leaving the pool in an inconsistent state 


primeroz:
/kind bug

**What steps did you take and what happened:**
* create a Cluster with a single machine pool ( see below for the actual manifests ) with 3 nodes
* Once the pool is stable change the `vmSize` ( from `Standard_D2s_v5` to `Standard_D4s_v5` )
* the pool goes into `ScaleSetOufOfDate` state and one node gets replaced
  * The pool is grown to 4 nodes ( by the capz controller i guess ) 
  * when the new machine is ready the pool is shrunk to 3 nodes ( with a default strategy of deleting the oldest node ) 
* This only happens for one node instead of repeating for every node in the pool that is still running on the old instance type  

**What did you expect to happen:**
All nodes to be replaced using the same process that was used for the first node 

**Anything else you would like to add:**
The docs [suggests](https://capz.sigs.k8s.io/topics/machinepools.html#safe-rolling-upgrades-and-delete-policy) that when changing `vmSize` or `kubernetesVersion` a full refresh of nodes in the nodepool should be performed 

_Machine pool manifests_ ( pre-change )
```
---
apiVersion: cluster.x-k8s.io/v1beta1
kind: MachinePool
metadata:
  labels:
    cluster.x-k8s.io/cluster-name: "fctest1"
  name: fctest1-def00
  namespace: org-multi-project
spec:
  clusterName: fctest1
  replicas: 3
  template:
    spec:
      bootstrap:
        configRef:
          apiVersion: bootstrap.cluster.x-k8s.io/v1beta1
          kind: KubeadmConfig
          name: fctest1-def00
      clusterName: fctest1
      infrastructureRef:
        apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
        kind: AzureMachinePool
        name: fctest1-def00
      version: 1.24.8
---
apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
kind: AzureMachinePool
metadata:
  labels:
    cluster.x-k8s.io/cluster-name: "fctest1"
  name: fctest1-def00
  namespace: org-multi-project
spec:
  location: westeurope
  strategy:
    rollingUpdate:
      deletePolicy: Oldest
      maxSurge: 25%
      maxUnavailable: 1
    type: RollingUpdate
  template:
    osDisk:
      diskSizeGB: 50
[...]

PR #2975, opened 2023-01-02
Better Detect when Machines in MachinePool VMSS are not running on the latest model

primeroz:
**What type of PR is this?**

/kind feature

Uses the value for `latestModelApplied` as returned by Azure to determine if the Machine is up-to-date or not.

Capz will Delete and replace nodes that are not running the latest mode, fixing the issue of rolling out a change to the vmSize or Kubernetes Version to a set of nodes in a machinepool. 
I think this is what we want based on the documentation [here](https://capz.sigs.k8s.io/topics/machinepools.html#safe-rolling-upgrades-and-delete-policy) which at the moment is not true ( at least in my testing )

**Which issue(s) this PR fixes** :
Fixes # https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/2972

* I updated some tests to include the new field but i did not add any new test.
* I marked this as `feature` but maybe it should be a `bug` since at the moment the controller does not do what the docs says it should do when spec of machinepool change ? 

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Improve detecting when a `azureMachinePoolMachine` is not running the `latest version` of the spec to allow CAPZ to fully refresh the nodes in a `machinePool` when change is required.  
```

### DIFF
--- a/azure/converters/vmss.go
+++ b/azure/converters/vmss.go
@@ -99,6 +99,8 @@ func SDKToVMSSVM(sdkInstance compute.VirtualMachineScaleSetVM) *azure.VMSSVM {
 		instance.AvailabilityZone = to.StringSlice(sdkInstance.Zones)[0]
 	}
 
+	instance.LatestModelApplied = *sdkInstance.LatestModelApplied
+
 	return &instance
 }
 
--- a/azure/converters/vmss_test.go
+++ b/azure/converters/vmss_test.go
@@ -63,7 +63,8 @@ func Test_SDKToVMSS(t *testing.T) {
 							Name:       to.StringPtr("vm0"),
 							Zones:      to.StringSlicePtr([]string{"zone0"}),
 							VirtualMachineScaleSetVMProperties: &compute.VirtualMachineScaleSetVMProperties{
-								ProvisioningState: to.StringPtr(string(compute.ProvisioningState1Succeeded)),
+								ProvisioningState:  to.StringPtr(string(compute.ProvisioningState1Succeeded)),
+								LatestModelApplied: to.BoolPtr(true),
 								OsProfile: &compute.OSProfile{
 									ComputerName: to.StringPtr("instance-000000"),
 								},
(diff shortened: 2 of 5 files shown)


CecileRobertMichon (commented):
/ok-to-test
/assign @mboersma 

there is a good chance this conflicts with #999 so we might need to do some rebasing

primeroz:
> there is a good chance this conflicts with https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/999 so we might need to do some rebasing

Yes , i see. 

I will join the Office hours in case you want to talk about this ? 

On one side this is kind of a blocker for us to be able to use MachinePools but on the other side if such a big refactor of MachinePool codebase is coming through that other issue it might be worth waiting for it ?

mboersma:
I think it will be possible to rebase this around #2813, which is a priority for the next release and should merge soon. I'm happy to help with that if it's tricky @primeroz.

primeroz:
@mboersma it sounds good, i still want to do some testing anyway , when #2813 is merged i will either rebase or rebranch/rewrite ( which might just be quicker :) )

primeroz:
@mboersma while looking at #2813 i came up with a question about the approach in this PR 

This change i did is only making the detection of nodes that are not running the latestmodel more precise and relies on `deleting AzureMachinePoolMachine instances in the MachinePoolScope` to replace them one at the time . this does not take into account the `surge settings` of the `rollout strategy` 


while testing before writing this PR i can see that when a VMSS is patched there is a surge that only the number of nodes defined in the `rollout strategy` instead of replacing all the nodes in the pool ( which is the original issue i opened ) and it seems to be handled [here](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/services/scalesets/scalesets.go#L273)

To my understanding this code ( both now and with changes from #2813) does not change and will do exactly the same thing. 
Do you think it make more sense to change this `surge control` code to replace all nodes , using the rollout strategy, when a VMSS is patched rather than just replace the nodes that `surge` ?

primeroz:
note to self, during Office Hour it was brought up this could clash with the `join token` issue . I will bring this up in slack for discussion to understand what is the best way to proceed with this

primeroz:
I am closing this PR for now. 

I am not _so sure_ this is the right approach so i will write something up and post it into Slack for feedback before picking this up again.

PR #2978, opened 2023-01-03
[release-1.6] fix machine pool wait in AKS and OOT e2e test

jackfrancis:
Reverts kubernetes-sigs/cluster-api-provider-azure#2930

@nojnhuh I made a mistake approving this PR as the dual-stack scenario isn't using `MachinePool`, so when we removed the `MachinePool` input config we're now using the default (1 second) timeout. Ref:

- https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/periodic-cluster-api-provider-azure-e2e-v1beta1/1607069744546975744

### DIFF
--- a/test/e2e/azure_test.go
+++ b/test/e2e/azure_test.go
@@ -575,7 +575,7 @@ var _ = Describe("Workload cluster creation", func() {
 				},
 				WaitForClusterIntervals:      e2eConfig.GetIntervals(specName, "wait-cluster"),
 				WaitForControlPlaneIntervals: e2eConfig.GetIntervals(specName, "wait-control-plane"),
-				WaitForMachineDeployments:    e2eConfig.GetIntervals(specName, "wait-worker-nodes"),
+				WaitForMachinePools:          e2eConfig.GetIntervals(specName, "wait-worker-nodes"),
 				ControlPlaneWaiters: clusterctl.ControlPlaneWaiters{
 					WaitForControlPlaneInitialized:   WaitForAKSControlPlaneInitialized,
 					WaitForControlPlaneMachinesReady: WaitForAKSControlPlaneReady,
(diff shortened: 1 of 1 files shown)


jackfrancis:
@nojnhuh now I see what's going on, looks like the backport updated the wrong test spec

nojnhuh:
Yep, looks like I just made that change to the completely wrong test. Sorry! Can't even blame the bot for that one lol.
/lgtm

jackfrancis:
@assign @mboersma

CecileRobertMichon (commented):
/lgtm
/approve

CecileRobertMichon:
/hold

for tests

jackfrancis:
/hold cancel

the optional tests were mainly me getting pre-patch-release signal (taking advantage of this release-1.6 PR)

nojnhuh:
It looks like the test failure is a timeout waiting the right kube-proxy image to be applied after upgrade. I'm running the test locally right now and it's stuck at the same spot waiting for the k8s.gcr.io image but the image actually set is registry.k8s.io, and that is successfully being pulled. I haven't been following the registry migration stuff super closely, so I'm not exactly sure what the most correct fix is in this case, but I'm reasonably confident that's what's going on.

Here's a go playground for the snippet that selects which registry to expect:
https://go.dev/play/p/w_oc1gYQW_Y
https://github.com/kubernetes-sigs/cluster-api/blob/6f61373d9f32c32e62def4140bfc6becee46732b/test/framework/daemonset_helpers.go#L42-L52

jackfrancis:
@nojnhuh does this correlate w/ @sbueringer's matrix he presented in capi office hours?

https://github.com/kubernetes-sigs/cluster-api/issues/7833

nojnhuh:
Yes, that matches what I'm seeing, where the v1.24.9 image is being set to registry.k8s.io and being successfully pulled. And it looks like the check in CAPI has been updated since v1.2.6 to work around this particular issue at least, so a CAPI bump here to v1.2.8 should do the trick. https://github.com/kubernetes-sigs/cluster-api/blob/v1.2.8/test/framework/daemonset_helpers.go#L50

jackfrancis:
@CecileRobertMichon is it possible to land this PR in the `release-1.6` branch without that optional test blocking? If we rebase/force-push from `release-1.6` will that reset the test gate accounting?

CecileRobertMichon:
yes, it should, but why are we trying to skip the test if we have a good way to fix it? Is the plan to fix the test in a separate PR?

jackfrancis:
right, fixing that particular thing is unelated to this change

nojnhuh:
I took a swing at putting a CAPI bump PR together and am running into weird go.mod things inside dependencies but I'll keep working through those.

nojnhuh:
CAPI bump PR open: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2989

sbueringer:
I think the test failure happens because Cluster API v1.2.6 doesn't migrate to the new registry but our e2e test code expects the new registry (for >= v1.22.0)

I think an upgrade to v1.2.8 should fix that test.

This PR will also improve the validation code https://github.com/kubernetes-sigs/cluster-api/pull/7856 (but that shouldn't be necessary to get your test green as it''s only relevant in other cases)

PR #2983, opened 2023-01-04
fix ci lint job

sonasingh46:

The lint job was broken on the main branch and hence lint issues were creeping in to the main branch. 
This PR fixes it. 

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Fix CI lint job
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -646,7 +646,7 @@ go-test: $(SETUP_ENVTEST) ## Run go tests.
 
 .PHONY: test-cover
 test-cover: TEST_ARGS+= -coverprofile coverage.out
-test-cover: go-test-race ## Run tests with code coverage and generate reports.
+test-cover: test ## Run tests with code coverage and generate reports.
 	go tool cover -func=coverage.out -o coverage.txt
 	go tool cover -html=coverage.out -o coverage.html
 
--- a/azure/scope/clients.go
+++ b/azure/scope/clients.go
@@ -27,7 +27,6 @@ import (
 	"github.com/Azure/go-autorest/autorest"
 	"github.com/Azure/go-autorest/autorest/azure"
 	"github.com/Azure/go-autorest/autorest/azure/auth"
-
 	azureutil "sigs.k8s.io/cluster-api-provider-azure/util/azure"
 )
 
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -646,7 +646,7 @@ go-test: $(SETUP_ENVTEST) ## Run go tests.
 
 .PHONY: test-cover
 test-cover: TEST_ARGS+= -coverprofile coverage.out
-test-cover: go-test-race ## Run tests with code coverage and generate reports.
```
CecileRobertMichon, line 649: this stops running the race-detector on PRs (essentially reverting #2646), do we want to find a way to do both? maybe add lint to the existing target?

nawazkh: Sorry for commenting uninvited, I was looking at this PR the same time. 
With this change, looks like pre-requisite `test` for the target `test-cover` invokes `go-test-race` while running the target [`test`](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/64ac8815dacc6c086eb31af01b33d9db2be564ea/Makefile#L637). 
So, looks like we are good to go.

mboersma: This PR makes it seem as if target-specific variables aren't working in the CI pipeline. See [this comment](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2646#pullrequestreview-1209063664). Perhaps #2646 was where this broke?

sonasingh46: @CecileRobertMichon -- The `test` target has already `go-test-race` as dependency. 
Also -- I validated it by running locally that race detector is enabled. This is what I get in my log : 
```
KUBEBUILDER_ASSETS="/Users/kuashutosh/Library/Application Support/io.kubebuilder.envtest/k8s/1.20.2-darwin-amd64" go test ./... -coverprofile coverage.out -race
```
Let me know if I am still missing something.

sonasingh46: See line 636-641 on this file.

mboersma (commented):
/lgtm

I tested the relevant make targets locally and everything looks good. I see now how #2646 broke this silently: it's `make test-cover` that gets run in PRs, and that had removed `lint` as a dependant target.

Thanks for tracking this down @sonasingh46!

mboersma:
And to clarify: this leaves `make test` and `make test-cover` with the race detector enabled; run `make go-test` to skip the race detector and dependent make targets.

mboersma (approved):
/approve

PR #2996, opened 2023-01-09
prefer CAPI replicas-managed-by annotation for AKS autoscaler

### LINKED ISSUE #2993: Remove azure.ReplicasManagedByAutoscalerAnnotation in favor of CAPI ReplicasManagedByExternalAutoscaler()
Now that release-1.3 is used in CAPZ, we should remove the CAPZ annotation foo for externally managed replicas and replace it by the shared CAPI `ReplicasManagedByExternalAutoscaler()` helper function.

/assign @jackfrancis

jackfrancis:
**What type of PR is this?**

This PR changes the annotation that we add to AKS node pools to the CAPI-defined `"cluster.x-k8s.io/replicas-managed-by"` annotation.

See:

- https://github.com/kubernetes-sigs/cluster-api/pull/7107

This PR also deprecates the CAPZ-defined `"cluster.x-k8s.io/replicas-managed-by-autoscaler"` annotation. We include a fallback support for this during MachinePool reconciliation, to aid forward-compatibility for existing clusters using the previous annotation.

I've decided not to worry about cleaning up the legacy annotation: upon the first reconciliation of the AKS node pool, CAPZ will assign the new, CAPI annotation. This seems simpler than worry about checking for existence of the old annotation on each reconcile loop, and deleting it if it's affected. The additional fallback code that supports the legacy annotation is not a maintenance burden, IMO.

Fixes #2993

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
prefer CAPI replicas-managed-by annotation for AKS autoscaler
```

### DIFF
--- a/azure/const.go
+++ b/azure/const.go
@@ -28,8 +28,4 @@ const (
 	// See https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/
 	// for annotation formatting rules.
 	RGTagsLastAppliedAnnotation = "sigs.k8s.io/cluster-api-provider-azure-last-applied-tags-rg"
-
-	// ReplicasManagedByAutoscalerAnnotation is set to true in the corresponding capi machine pool
-	// when an external autoscaler manages the node count of the associated machine pool.
-	ReplicasManagedByAutoscalerAnnotation = "cluster.x-k8s.io/replicas-managed-by-autoscaler"
 )
--- a/azure/scope/machinepool.go
+++ b/azure/scope/machinepool.go
@@ -40,6 +40,7 @@ import (
 	"sigs.k8s.io/cluster-api/controllers/noderefutil"
 	capierrors "sigs.k8s.io/cluster-api/errors"
 	expv1 "sigs.k8s.io/cluster-api/exp/api/v1beta1"
+	"sigs.k8s.io/cluster-api/util/annotations"
 	"sigs.k8s.io/cluster-api/util/conditions"
 	"sigs.k8s.io/cluster-api/util/patch"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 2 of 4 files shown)


azure/scope/machinepool.go
```diff
@@ -735,7 +736,10 @@ func (m *MachinePoolScope) UpdateCAPIMachinePoolReplicas(ctx context.Context, re
 
 // HasReplicasExternallyManaged returns true if the externally managed annotation is set on the CAPI MachinePool resource.
 func (m *MachinePoolScope) HasReplicasExternallyManaged(ctx context.Context) bool {
-	return m.MachinePool.Annotations[azure.ReplicasManagedByAutoscalerAnnotation] == "true"
+	return annotations.ReplicasManagedByExternalAutoscaler(m.MachinePool) ||
+		// Support the legacy, CAPZ-only "cluster.x-k8s.io/replicas-managed-by-autoscaler" annotation
+		// introduced in CAPZ v1.4.2 and v1.5.0
+		m.MachinePool.Annotations[azure.ReplicasManagedByAutoscalerAnnotation] == "true"
```
CecileRobertMichon, line 742: I don't think we actually need to do this since this code was added in https://github.com/kubernetes-sigs/cluster-api-provider-azure/commit/66a3529e8a3f8fcfd5b17fbabf4cb44db12e378f which hasn't been part of a release (yet)

the annotation itself has been here since 1.5 but that was only used by AKS clusters and those don't use this function

jackfrancis: that makes sense!

azure/const.go
```diff
@@ -29,7 +29,7 @@ const (
 	// for annotation formatting rules.
 	RGTagsLastAppliedAnnotation = "sigs.k8s.io/cluster-api-provider-azure-last-applied-tags-rg"
 
-	// ReplicasManagedByAutoscalerAnnotation is set to true in the corresponding capi machine pool
-	// when an external autoscaler manages the node count of the associated machine pool.
+	// ReplicasManagedByAutoscalerAnnotation is deprecated, replaced by the CAPI annotatation
+	// "cluster.x-k8s.io/replicas-managed-by".
```
CecileRobertMichon, line 33: since this const is no longer used anywhere, should we just remove it?

jackfrancis: fair

CecileRobertMichon (commented):
/lgtm
/approve

PR #2999, opened 2023-01-09
don’t default to AKS networkPolicy=calico

jackfrancis:
**What type of PR is this?**

/kind flake
/kind feature

This PR removes an opinionated default for AKS clusters to use the calico NetworkPolicy feature.

I was unaware that this default configuration enforcement existed. It is not documented anywhere I can see, and I'm not aware of any justification to include this opinionated cluster configuration for CAPZ-originating AKS clusters: we should let the user decide whether or not to include the installation of either calico or azure NetworkPolicy, as either requires additional operational overhead and other considerations that aren't suitable for choosing one or the other as a default (IMO).

Feature support for calico and azure NetworkPolicy remain unchanged.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
don’t default to AKS networkPolicy=calico
```

### DIFF
--- a/azure/services/managedclusters/spec.go
+++ b/azure/services/managedclusters/spec.go
@@ -62,7 +62,7 @@ type ManagedClusterSpec struct {
 	// NetworkPlugin used for building Kubernetes network. Possible values include: 'azure', 'kubenet'. Defaults to azure.
 	NetworkPlugin string
 
-	// NetworkPolicy used for building Kubernetes network. Possible values include: 'calico', 'azure'. Defaults to azure.
+	// NetworkPolicy used for building Kubernetes network. Possible values include: 'calico', 'azure'.
 	NetworkPolicy string
 
 	// SSHPublicKey is a string literal containing an ssh public key. Will autogenerate and discard if not provided.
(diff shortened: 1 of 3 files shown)


jackfrancis:
@mtougeron @zmalik @dthorsen @primeroz @sonasingh46 @LochanRn @zioproto @austintackaberry @karthikbalasub @michalno1 @meixingdb Does anyone have any concerns about this change? I do have a medium-strong opinion that we CAPZ should not opinionate about calico for NetworkPolicy, but there is the fact that this is a behavioral change and user expectation may have acclimated to calico by default.

Please share any downsides to reverting this default, thank you!

dthorsen:
I agree. I think it makes sense to make this change now while AKS support is still in experimental. We should just ensure that the behavioral change is documented in the release notes.

mtougeron:
> I agree. I think it makes sense to make this change now while AKS support is still in experimental. We should just ensure that the behavioral change is documented in the release notes.

+1 to this

CecileRobertMichon (commented):
/lgtm

PR #3000, opened 2023-01-09
Set VMType=standard in azure.json if MachinePools aren't enabled

### LINKED ISSUE #2998: Don't set `VMType: vmss` in azure.json if MachinePools aren't enabled
/kind bug

CAPZ always sets `VMType: vmss` in the azure.json config, because it amounts to a feature flag that we want to enable in case the cluster contains MachinePools. But if MachinePools are disabled altogether, we should instead use `VMType: vm` for logical consistency and possibly to reduce unneeded API calls.

See [this code](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2813/files#diff-7d084ba6e9e15005850c1b44d72d457cef09f12badee416beda37f957a242364R203) for what inspired this issue.

mboersma:
**What type of PR is this?**

/kind bug
/priority important-longterm

Since MachinePool/VMSS support in cloud-provider-azure is essentially a feature flag, CAPZ always enables it by setting `VMType: vmss` in azure.json. But if the MachinePools feature is disabled, `VMType: standard` makes more sense and will skip some API calls.

Fixes #2998

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
NONE
```


mboersma:
/hold

There's just a linter error to fix, but let's wait until #2813 has merged before revisiting this.

controllers/helpers.go
```diff
@@ -253,7 +265,7 @@ func newCloudProviderConfig(d azure.ClusterScoper) (controlPlaneConfig *CloudPro
 			SecurityGroupName:            subnet.SecurityGroup.Name,
 			SecurityGroupResourceGroup:   d.Vnet().ResourceGroup,
 			Location:                     d.Location(),
-			VMType:                       "vmss",
+			VMType:                       "vm",
```
zmyzheng, line 268: Maybe I missed something, could you provide more info for this change? More specifically:
1. Will this cloudProviderConfig be applied to the azure.json file?
2. Besides Cloud Controller Manager, is there any other component using this cloud config?
3. I remember Cloud Contoller Manager expects the vmType to be set to "standard" rather than "vm" for non-vmss cases. Do you set it to "vm" for other purpose? Just want to double check in case it does not align with what Cloud Controller Mananger expects :)

mboersma: This change came about because we noticed CAPZ *always* sets `VMType: vmss`, but doesn't need to if MachinePools aren't even available.

1. Yes, this is the config that is the azure.json file.
2. I've been told there are other components that might read the config, but I'm not clear which--maybe cluster-autoscaler?
3. This is just a bug resulting from me not re-checking the provider docs, I intended to use `standard`, not `vm`. I'll update the PR soon.

mboersma:
/retitle Set VMType=standard in azure.json if MachinePools aren't enabled

mboersma:
/retest

Failure was on the ipv6 spec with this known error: `configmaps "kubeadm-config" not found`

controllers/helpers.go
mboersma: @zmyzheng I see now in the cloud-provider-azure docs that the `vmType` values have changed:

| Name | Description | Remark |
|---|---|---|
| vmType | The type of azure nodes. Candidate values are: `vmss`, `vmssflex` and `standard` | Optional, default to `standard` |
| enableVmssFlexNodes | Enable supporting for VMSS Flex virtual machines in vmss cluster. It should only be used when vmType is “vmss” | Since v1.26.0, default is false |

I'm confused about what this new `vmType: vmssflex` value is and how it interacts with `enableVmssFlexNodes`.

mboersma:
/hold

I have a question about the new `vmType: vmssflex` that appeared in cloud-provider-azure docs.

controllers/helpers.go
mboersma: This section of the docs explains it better: https://cloud-provider-azure.sigs.k8s.io/topics/node-types/

I actually think the main doc that I excerpted above is misleading and should be changed. I'll make a PR.

So based on CAPZ's usage patterns, we would never use `vmType: vmssflex` because we won't have a cluster that is 100% VMSS Flexible nodes. 

My only question is if this PR is worthwhile: if we know a cluster is 100% standalone or Availability Set nodes, is there any advantage to using `vmType: standard`?

mboersma:
/hold cancel

Based on [this comment](https://github.com/kubernetes-sigs/cloud-provider-azure/issues/3099#issuecomment-1383514575), this PR seems correct and worthwhile:

> when setting setting vmType: standard, no VMSS and VMSS Flex APIs would be invoked (No VMSS caches would be created as well), which would reduce number of ARM APIs on CRP.

nawazkh:
Would we like to make changes to https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/docs/book/src/topics/machinepools.md#example-machinepool-azuremachinepool-and-kubeadmconfig-resources as well, updating `capz-mp-0`'s `Spec.files[0].content.vmType = standard`?

mboersma:
> Would we like to make changes to...

That doc is in the MachinePools topic, so in that case azure.json should have `vmType: vmss` and doesn't need to change AFAICT. I grepped the codebase and didn't see `vmType` mentioned anywhere else but here--thanks for checking!

willie-yao (approved):
/lgtm

CecileRobertMichon:
My only question with this is, could this cause issues for existing clusters once we remove the EXP feature flag for MachinePools? We might end up in a situation where the azurejson existing secret is no longer correct. This optimization is only temporary as we'll no longer be able to use it once MachinePools are no longer experimental, so is it worth it?

mboersma:
> This optimization is only temporary as we'll no longer be able to use it once MachinePools are no longer experimental, so is it worth it?

D'oh, good point. I haven't benchmarked anything either, so the savings in terms of CRP API calls is theoretical.

I'll close this. Thanks everyone for the thoughtful comments and reviews! 😄

PR #3002, opened 2023-01-09
Handle ipv6 and dual stack CIDRs in ci-entrypoint.sh

CecileRobertMichon:
**What type of PR is this?**

**What this PR does / why we need it**: Needed for https://github.com/kubernetes-sigs/cloud-provider-azure/pull/2995. Cloud-provider-azure tests use custom CLUSTER_TEMPLATEs with ci-entrypoint.sh, some of which are ipv6 and dual-stack, so the Helm Calico install needs to be able to handle ipv6 and dual-stack CIRD ranges.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/scripts/ci-entrypoint.sh
+++ b/scripts/ci-entrypoint.sh
@@ -132,14 +132,9 @@ create_cluster() {
 }
 
 install_addons() {
-    # Get CCM cluster CIDRs from Cluster object if not set
-    if [[ -z "${CCM_CLUSTER_CIDR:-}" ]]; then
-        CCM_CLUSTER_CIDR=$(${KUBECTL} get cluster "${CLUSTER_NAME}" -o=jsonpath='{.spec.clusterNetwork.pods.cidrBlocks[0]}')
-        if CIDR1=$(${KUBECTL} get cluster "${CLUSTER_NAME}" -o=jsonpath='{.spec.clusterNetwork.pods.cidrBlocks[1]}' 2> /dev/null); then
-            CCM_CLUSTER_CIDR="${CCM_CLUSTER_CIDR:-}\,${CIDR1}"
-        fi
-    fi
-    echo "CCM cluster CIDR: ${CCM_CLUSTER_CIDR:-}"
+    # Get cluster CIDRs from Cluster object
+    CIDR0=$(${KUBECTL} get cluster "${CLUSTER_NAME}" -o=jsonpath='{.spec.clusterNetwork.pods.cidrBlocks[0]}')
+    CIDR1=$(${KUBECTL} get cluster "${CLUSTER_NAME}" -o=jsonpath='{.spec.clusterNetwork.pods.cidrBlocks[1]}' 2> /dev/null) || true
 
     # export the target cluster KUBECONFIG if not already set
     export KUBECONFIG="${KUBECONFIG:-${PWD}/kubeconfig}"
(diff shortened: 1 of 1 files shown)


scripts/ci-entrypoint.sh
```diff
@@ -132,14 +132,9 @@ create_cluster() {
...
-    if [[ -z "${CCM_CLUSTER_CIDR:-}" ]]; then
-        CCM_CLUSTER_CIDR=$(${KUBECTL} get cluster "${CLUSTER_NAME}" -o=jsonpath='{.spec.clusterNetwork.pods.cidrBlocks[0]}')
-        if CIDR1=$(${KUBECTL} get cluster "${CLUSTER_NAME}" -o=jsonpath='{.spec.clusterNetwork.pods.cidrBlocks[1]}' 2> /dev/null); then
-            CCM_CLUSTER_CIDR="${CCM_CLUSTER_CIDR:-}\,${CIDR1}"
-        fi
-    fi
-    echo "CCM cluster CIDR: ${CCM_CLUSTER_CIDR:-}"
+    # Get cluster CIDRs from Cluster object
+    CIDR0=$(${KUBECTL} get cluster "${CLUSTER_NAME}" -o=jsonpath='{.spec.clusterNetwork.pods.cidrBlocks[0]}')
+    CIDR1=$(${KUBECTL} get cluster "${CLUSTER_NAME}" -o=jsonpath='{.spec.clusterNetwork.pods.cidrBlocks[1]}' 2> /dev/null)
```
jackfrancis, line 137: This is a graceful command if there is only one CIDR in `.spec.clusterNetwork.pods.cidrBlocks`?

scripts/ci-entrypoint.sh
```diff
@@ -150,10 +145,24 @@ install_addons() {
...
+        CIDR_STRING_VALUES="installation.calicoNetwork.ipPools[0].cidr=${CIDR0},installation.calicoNetwork.ipPools[1].cidr=${CIDR1}"
+    else
+        echo "Cluster CIDR is IPv4"
+        CALICO_VALUES_FILE="${REPO_ROOT}/templates/addons/calico/values.yaml"
+        CIDR_STRING_VALUES="installation.calicoNetwork.ipPools[0].cidr=${CIDR0}"
+    fi
+
     "${HELM}" repo add projectcalico https://projectcalico.docs.tigera.io/charts
-    "${HELM}" install calico projectcalico/tigera-operator -f "${CALICO_VALUES:-${REPO_ROOT}/templates/addons/calico/values.yaml}" --namespace tigera-operator --create-namespace
+    "${HELM}" install calico projectcalico/tigera-operator -f "${CALICO_VALUES_FILE}" --set-string "${CIDR_STRING_VALUES}" --namespace tigera-operator --create-namespace
```
jackfrancis, line 165: Looks like this `installation.calicoNetwork.ipPools[0].cidr` config override is totally new w/ this PR? Does that suggest that even in the ipv4 scenario we've been doing this wrong previously?

CecileRobertMichon: not entirely. Before it was using the value from https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/templates/addons/calico/values.yaml#L8 which happens to be the default ipv4 CIDR we use in all test templates

scripts/ci-entrypoint.sh
CecileRobertMichon: correct

jackfrancis (approved):
/lgtm
/approve

scripts/ci-entrypoint.sh
CecileRobertMichon: actually let me double check that

/hold

scripts/ci-entrypoint.sh
```diff
@@ -174,6 +183,12 @@ install_addons() {
             copy_secret
         fi
 
+        CCM_CLUSTER_CIDR="${CIDR0}"
```
lzhecheng, line 186: Does it still allow the user to specify `CCM_CLUSTER_CIDR`?

CecileRobertMichon: No, is there any case where you'd like to be able to set a CCM CIDR that's different from the Cluster pod CIDR range?

scripts/ci-entrypoint.sh
lzhecheng: So without this change, is IPv6 test doing wrong?

scripts/ci-entrypoint.sh
jackfrancis: This occured to me as well: moving away from a random env var value for this seems like an improvement, this should always be the value in the cluster spec (I think?)

lzhecheng: Got it. No. I think this change is good.

scripts/ci-entrypoint.sh
CecileRobertMichon: not in release v1.6 and below (it would be wrong if we release v1.7 without this change though). The changes to install Calico with Helm are only in the main branch currently.

lzhecheng: Got it. Thanks.

lzhecheng:
LGTM

scripts/ci-entrypoint.sh
CecileRobertMichon: I got it wrong before, needed the `|| true` as well otherwise it fails the script (even if stderr is suppressed)

CecileRobertMichon: /hold cancel

PR #3003, opened 2023-01-10
Refactor repeated code in E2E test specs to helper functions

willie-yao:
**What type of PR is this?**
/kind cleanup

**What this PR does / why we need it**:
Currently, there is a lot of repeated code in our E2E test files, such as the multiple calls to the [applyClusterTemplateAndWait](https://github.com/willie-yao/cluster-api-provider-azure/blob/528b3706f6e04408184e121a20ffbbfc408a25e0/test/e2e/azure_test.go#L167) function. This PR fixes the issue by creating a helper function for repeated calls to `applyClusterTemplateAndWait`.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #2997

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

PR #3006, opened 2023-01-10
refactor agentpool test definitions

nojnhuh:
**What type of PR is this?**
/kind cleanup

This PR refactors the agentpools unit test definitions to be more readable and self-contained by removing the shared test fixtures and instead using flexible builder functions to construct objects for each individual test. With this change, new tests should be easier to add in the future without having to copy-paste a full fixture to change a couple of fields like in https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2751/files#diff-224617dc1cf5a3e1abbc71a49b79c42441871bef5a14e810921526267e6bde82.

AFAICT the actual tests are exactly the same, the only difference is the syntax to define them.

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

/area managedclusters

cc @willie-yao in case this is useful for #3003

### DIFF
--- a/azure/services/agentpools/agentpools_test.go
+++ b/azure/services/agentpools/agentpools_test.go
@@ -32,33 +32,7 @@ import (
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 )
 
-var (
-	fakeAgentPoolSpec = AgentPoolSpec{
-		Name:              "fake-agent-pool-name",
-		ResourceGroup:     "fake-rg",
-		Cluster:           "fake-cluster",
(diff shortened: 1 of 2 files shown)


jackfrancis (commented):
@nojnhuh any preference between landing this before or after the taints fix?

nojnhuh:
> @nojnhuh any preference between landing this before or after the taints fix?

Not necessarily, rebase shouldn't be bad either way.

jackfrancis (approved):
/lgtm
/approve

PR #3016, opened 2023-01-11
Add back commit-message.prefix as it is a dependency

nawazkh:
**What type of PR is this?**
/kind bug

- adds back `commit-message.prefix`. 
- `commit-message.prefix` is set as "dependabot"

Fixes #

- Thanks to @CecileRobertMichon for noticing the dependency in one of the post PR GitHub-actions run logs. 
Screenshot 2023-01-11 at 3 32 40 PM

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,6 +9,7 @@ updates:
   schedule:
     interval: "daily"
   commit-message:
+    prefix: "dependabot"
     include: scope
   labels:
     - "ok-to-test"
@@ -20,6 +21,7 @@ updates:
   schedule:
     interval: "daily"
   commit-message:
+    prefix: "dependabot"
     include: scope
   ignore:
   


nawazkh:
Can we skip running all the tests and merge this? 
/cc @CecileRobertMichon

CecileRobertMichon (commented):
/lgtm
/approve

.github/dependabot.yml
```diff
@@ -9,6 +9,7 @@ updates:
   schedule:
     interval: "daily"
   commit-message:
+    prefix: "dependabot:"
```
CecileRobertMichon, line 12: /lgtm cancel

CecileRobertMichon: if you were looking at `:seedling:` as reference it has semicolons on both sides because that's the syntax for emojis in markdown

.github/dependabot.yml
```diff
@@ -20,6 +21,7 @@ updates:
...
```
CecileRobertMichon, line 24: same here

.github/dependabot.yml
nawazkh: Yeah, I purposefully put `dependabot:` to get a commit message of type `dependabot: updated something`.

Do we want to use some other prefix?

nawazkh: or I can just remove the `:`

CecileRobertMichon: I'm assuming that dependabot will include the semicolon for us between prefix and commit message so `prefix: dependabot:` would result in a commit message like `dependabot:: updated dependency xyz` but maybe I'm wrong

nawazkh: Updated and removed the semicolon for now :)

CecileRobertMichon (commented):
/lgtm
/approve

Issue #2974, opened 2023-01-01
Azure/go-autorest will soon be unsupported


mboersma:
/kind deprecation
/priority important-soon

Related to the SDK v2 transition, the `Azure/go-autorest` library will go [out of support](https://github.com/Azure/go-autorest#note-the-modules-in-this-repo-will-go-out-of-support-by-march-31-2023--additional-information-can-be-found-here) on March 31, 2023. CAPZ should find alternatives to those modules–see [here](https://github.com/kubernetes-sigs/cloud-provider-azure/pull/2983) for an example of replacing `to.StringPtr()` and friends.

dtzar:
@mboersma - isn't this a duplicate of https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/2670 ?

mboersma:
@dtzar they're related but not the same: Azure/azure-sdk-for-go and Azure/go-autorest. CAPZ's direct usage of go-autorest can probably be addressed as a separate task, similarly to kubernetes-sigs/cloud-provider-azure#2983.

CecileRobertMichon:
@mboersma are you interested in working on this or should we mark it as help wanted?

mboersma:
/assign

I'm psyched to work on it but since it's also intertwined with our async framework it will be more work than just replacing our `to.StringPtr()`-type calls.

CecileRobertMichon:
Should we maybe start with the `to.*Ptr` type calls since we'll have to do that regardless of SDK track 2 vs. ASO stuff? that can be its own PR. Then we can evaluate moving async services separately.

dtzar:
@mboersma should this be attached to #3409 parent?

mboersma:
> @mboersma should this be attached to https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/3409 parent?

Yes, good call. Although really it will probably require all of ASO to be implemented as well before go-autorest is literally not used.

mboersma:
/unassign

I've done the up-front work that was possible here. Let's revisit this once we've switched all over to SDKv2 + ASO and see what remains from go-autorest.

mboersma:
/milestone next

I took a look around, and this will be a fair amount of work even after SDK v2 is fully integrated. go-autorest is woven into our opentelemetry tracing, we use autorest errors and objects as test data, and the go-autorest auth struct is everywhere.

mboersma:
[...]

PR #3026, opened 2023-01-12
Use k8s utils/pointer instead of Azure autorest/to

mboersma:
**What type of PR is this?**

/kind cleanup

Replaces usage of the `github.com/Azure/go-autorest/autorest/to` package with `k8s.io/utils/pointer` and a couple small helper funcs. This makes forward progress toward removing usage of Azure/go-autorest, which will soon be unsupported.

In general, the changes here consist of:

| autorest/to      | utils/pointer |
| ----------- | ----------- |
| to.StringPtr      | pointer.String       |
| to.Float64Ptr   | pointer.Float64        
| ...etc   |         |
| to.String  | pointer.StringDeref        |
| to.Int32   | pointer.Int32Deref       |
| ...etc  |         |
| to.StringSlicePtr   | &        |
| to.StringSlice   | (local) azure.StringSlice        |
| to.StringMapPtr   | (local) azure.StringMapPtr      |

See kubernetes-sigs/cloud-provider-azure#2983 for comparison.

Refs #2974

This changeset is huge; I'd be glad to break it up into multiple PRs if that helps with reviewing.

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```


mboersma:

AKS provisioning failure, unrelated to this code change.

mboersma:
/hold

Let's wait until #2959 merges first.

CecileRobertMichon (commented):
/assign @willie-yao @Jont828

api/v1beta1/azurecluster_default_test.go
```diff
@@ -139,7 +139,7 @@ func TestVnetDefaults(t *testing.T) {
 							},
 						},
 						NodeOutboundLB: &LoadBalancerSpec{
-							FrontendIPsCount: to.Int32Ptr(1),
+							FrontendIPsCount: pointer.Int32Ptr(1),
```
willie-yao, line 142: I see that there are some uses of `pointer.Int32Ptr` and `pointer.Int32Deref`. The PR description mentions that `pointer.Int32Deref` is being used to replace `to.Int32Ptr`. Will this be a problem? I noticed that `pointer.Int32Deref` is used for cases where the int might be nil, so I think it is the correct usage.

mboersma: > `pointer.Int32Deref` is being used to replace `to.Int32Ptr`

Actually it's like this:

`to.Int32Ptr` -> `pointer.Int32` (return a pointer / reference)
`to.Int32` -> `pointer.Int32Deref` (return an int32 or 0)

Those changes *should* preserve identical behavior.

mboersma:
Rebased after #3093 changed go.mod.

willie-yao (approved):
/lgtm

api/v1beta1/azurecluster_default_test.go
CecileRobertMichon:
```
// Int32Ptr is a function variable referring to Int32.
// Deprecated: Use Int32 instead.
var Int32Ptr = Int32 // for back-compat
```

seems we should be using `Int32` though? `Int32Ptr` is deprecated

mboersma: Oops, yes, that's a mistake that crept in through a rebase: `pointer.Int32` is preferred now. I'll update.

mboersma:
I had to rebase again for changes in `main`, and I replaced some pre-existing usages of `pointer.Int32Ptr` with `pointer.Int32` and the like.

mboersma:
This PR is old enough now that `prow` is holding it hostage because the `pull-cluster-api-provider-azure-e2e-exp` job didn't pass (and it never will because it's the `-aks` job now). So it may have to be admin-merged.

PR #3030, opened 2023-01-12
fix agentpool taints diff check

nojnhuh:
**What type of PR is this?**
/kind bug
/area managedclusters

**What this PR does / why we need it**: Currently the diff between the current and desired agent pool definitions to determine whether an update is necessary or not uses the same value for node taints, so an update to an AzureManagedMachinePool's `spec.taints` would not trigger an update.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Fixed a bug preventing changes to AzureManagedMachinePool's `spec.taints` from taking effect.
```

### DIFF
--- a/azure/services/agentpools/spec_test.go
+++ b/azure/services/agentpools/spec_test.go
@@ -27,6 +27,7 @@ import (
 	"github.com/google/go-cmp/cmp"
 	. "github.com/onsi/gomega"
 	"github.com/pkg/errors"
+	"k8s.io/utils/pointer"
 	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
 	"sigs.k8s.io/cluster-api-provider-azure/azure"
 )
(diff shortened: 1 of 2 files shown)


mboersma:
Is this something that could be unit tested in `TestParameters`?

nojnhuh:
> Is this something that could be unit tested in `TestParameters`?

Yes, good call. Just pushed those changes. That involved a bit of collateral damage so I'll put comments inline to explain.

azure/services/agentpools/spec_test.go
```diff
@@ -179,15 +183,16 @@ var (
 			EnableUltraSSD:      to.BoolPtr(true),
 			MaxCount:            to.Int32Ptr(5), // updates if changed
 			MaxPods:             to.Int32Ptr(10),
-			MinCount:            to.Int32Ptr(3),                                               // updates if changed
+			MinCount:            to.Int32Ptr(1),                                               // updates if changed
```
nojnhuh, line 186: Each of these fixtures should be changing one field from the "default" `fakeAgentPoolSpecWithAutoscaling` defined above, but some copy-paste resulted in multiple fields being changed. The changes around here to `MinCount` and `Tags` ensure only one field is being changed at a time.

This is necessary because the new test was passing even without the fix since `MinCount` and `Tags` were triggering a diff unexpectedly that `NodeTaints` should have, but didn't.

nawazkh:
Great find! lgtm!

jackfrancis (approved):
/lgtm
/approve

PR #3033, opened 2023-01-12
Wait for nodes after CCM install in ci-entrypoint

### LINKED ISSUE #2921: ci-entrypoint.sh fails intermittently with "error: no objects passed to create" due to missing kubeadm-config ConfigMap
/kind bug
/kind flake

The ci-entrypoint pre-submit CI job has been flaky lately and I've noticed a number of failing runs hit the same error here:
```
Error from server (NotFound): configmaps "kubeadm-config" not found
error: no objects passed to create
```
https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/e0d9257431826beb489db1db65204f55ca8f2155/scripts/ci-entrypoint.sh#L149-L151 

I suspect this is some sort of race between kubeadm creating that ConfigMap and reading it here? And waiting for that to exist would fix this error?

Runs:
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/2916/pull-cluster-api-provider-azure-ci-entrypoint/1601245720478748672
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/2752/pull-cluster-api-provider-azure-ci-entrypoint/1600913653387759616
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/2738/pull-cluster-api-provider-azure-ci-entrypoint/1599870510261342208
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/2411/pull-cluster-api-provider-azure-ci-entrypoint/1598784340660064256

The kubeadm-config ConfigMap exists before it's read so the test doesn't fail here.

**Environment:**

CecileRobertMichon:
**What type of PR is this?**

**What this PR does / why we need it**: This fixes an issue when testing external CCM (`TEST_CCM=true`) in ci-entrypoint.sh where we try to wait for nodes to be ready before CCM is installed (nodes won't be ready because cloud provider is not initialized). 

It also fails a race condition in the script which causes it to fail intermittently when kubeadm-config cm is not available right away.

Fixes #2921 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/scripts/ci-entrypoint.sh
+++ b/scripts/ci-entrypoint.sh
@@ -141,9 +141,14 @@ install_addons() {
 
     # Copy the kubeadm configmap to the calico-system namespace. This is a workaround needed for the calico-node-windows daemonset to be able to run in the calico-system namespace.
     "${KUBECTL}" create ns calico-system
+    until "${KUBECTL}" get configmap kubeadm-config --namespace=kube-system
+    do
+        # Wait for the kubeadm-config configmap to exist.
+        sleep 2
+    done
     "${KUBECTL}" get configmap kubeadm-config --namespace=kube-system -o yaml \
     | sed 's/namespace: kube-system/namespace: calico-system/' \
-    | ${KUBECTL} create -f -
+    | "${KUBECTL}" create -f -
 
     # install Calico CNI
     echo "Installing Calico CNI via helm"
@@ -163,9 +168,6 @@ install_addons() {
 
     "${HELM}" repo add projectcalico https://projectcalico.docs.tigera.io/charts
     "${HELM}" install calico projectcalico/tigera-operator -f "${CALICO_VALUES_FILE}" --set-string "${CIDR_STRING_VALUES}" --namespace tigera-operator --create-namespace
-
-    export -f wait_for_nodes
-    timeout --foreground 1800 bash -c wait_for_nodes
 
     # Add FeatureOverride for ChecksumOffloadBroken in FelixConfiguration.
     # This is the recommended workaround for https://github.com/projectcalico/calico/issues/3145.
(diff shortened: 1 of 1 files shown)


CecileRobertMichon:
/hold

need to check something

CecileRobertMichon:
/hold cancel

this fixes #2921 as well

scripts/ci-entrypoint.sh
```diff
@@ -207,6 +211,9 @@ install_addons() {
             --set-string cloudControllerManager.clusterCIDR="${CCM_CLUSTER_CIDR}"
     fi
 
+    export -f wait_for_nodes
+    timeout --foreground 1800 bash -c wait_for_nodes
```
marosset, line 215: Are we worried that waiting for the CCM install will make it more likely we'll hit this timeout?
Given that the timeout is 30 mins I would hope not but I wanted to ask...

CecileRobertMichon: I don't think I understand the question, could you please expand?

CecileRobertMichon:
> Pod got deleted unexpectedly 

maybe prow is having issues?

/retest

scripts/ci-entrypoint.sh
```diff
@@ -141,9 +141,16 @@ install_addons() {
 
     # Copy the kubeadm configmap to the calico-system namespace. This is a workaround needed for the calico-node-windows daemonset to be able to run in the calico-system namespace.
     "${KUBECTL}" create ns calico-system
+    for (( i=1; i<=30; i++ ))
+    do
+        if eval "${KUBECTL}" get configmap kubeadm-config --namespace=kube-system; then
```
jackfrancis, line 146: do we have to do eval here? does a negative lookup not produce a non-zero exit code (we should be able to just if/else based on the exit code of the command)

CecileRobertMichon: Good call, changed it to use until

scripts/ci-entrypoint.sh
CecileRobertMichon: I've verified this doesn't cause issues in local testing

jackfrancis (approved):
/lgtm
/approve

PR #2682, opened 2022-09-29
enable mutable AKS node taints

jackfrancis:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

This PR removes the "immutable" restriction for AKS node pool taints.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
enable mutable AKS node taints
```

PR #3038, opened 2023-01-13
add e2e test for AKS node pool taints

### LINKED ISSUE #2860: AzureManagedCluster E2E: node pool Taints CRUD tests
/kind feature
/area managedclusters

Create new AzureManagedCluster E2E: node pool Taints CRUD tests.

nojnhuh:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**: This change adds a new e2e test for AzureManagedMachinePools which updates `spec.taints` and ensures those changes are reflected in AKS.

Fixes #2860

Currently there's [no way to delete all taints from a node pool](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2682#issuecomment-1267701356), so this test has a side effect of leaving a `PreferNoSchedule` taint on each node pool that originally had no taints. I tried defining a taint with a bogus `effect` like `NoOp` but that gets rejected as invalid by AKS.

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/aks_node_taints.go
+++ b/test/e2e/aks_node_taints.go
@@ -1,0 +1,146 @@
+//go:build e2e
+// +build e2e
+
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package e2e
+
+import (
+	"context"
+	"fmt"
+	"sync"
+
+	"github.com/Azure/azure-sdk-for-go/services/containerservice/mgmt/2021-05-01/containerservice"
+	"github.com/Azure/go-autorest/autorest/azure/auth"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	expv1 "sigs.k8s.io/cluster-api/exp/api/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type AKSNodeTaintsSpecInput struct {
+	Cluster       *clusterv1.Cluster
+	MachinePools  []*expv1.MachinePool
+	WaitForUpdate []interface{}
+}
+
+func AKSNodeTaintsSpec(ctx context.Context, inputGetter func() AKSNodeTaintsSpecInput) {
+	input := inputGetter()
+
+	settings, err := auth.GetSettingsFromEnvironment()
+	Expect(err).NotTo(HaveOccurred())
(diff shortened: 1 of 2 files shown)


test/e2e/aks_node_taints.go
```diff
@@ -0,0 +1,143 @@
...
+				},
+				{
+					Effect: infrav1exp.TaintEffect(corev1.TaintEffectPreferNoSchedule),
+					Key:    "capz-e2e-2",
+					Value:  "test2",
+				},
+			}
+			Expect(mgmtClient.Get(ctx, client.ObjectKeyFromObject(ammp), ammp)).To(Succeed())
+			ammp.Spec.Taints = expectedTaints
+			Expect(mgmtClient.Update(ctx, ammp)).To(Succeed())
```
jackfrancis, line 121: pending thoughts on https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3058 let's wrap the Get and Update methods inside a common retry loop

nojnhuh: done

test/e2e/aks_node_taints.go
```diff
@@ -0,0 +1,149 @@
...
+func AKSNodeTaintsSpec(ctx context.Context, inputGetter func() AKSNodeTaintsSpecInput) {
+	input := inputGetter()
+
+	settings, err := auth.GetSettingsFromEnvironment()
+	Expect(err).NotTo(HaveOccurred())
+	subscriptionID := settings.GetSubscriptionID()
+	auth, err := settings.GetAuthorizer()
+	Expect(err).NotTo(HaveOccurred())
+
+	managedclustersClient := containerservice.NewManagedClustersClient(subscriptionID)
```
jackfrancis, line 54: Another one where it seems we're not actually doing anything w/ this client object.

Btw these test PRs are awesome!

test/e2e/aks_node_taints.go
```diff
@@ -0,0 +1,146 @@
...
+		go func(mp *expv1.MachinePool) {
+			defer GinkgoRecover()
+			defer wg.Done()
+
+			ammp := &infrav1.AzureManagedMachinePool{}
+			Expect(mgmtClient.Get(ctx, types.NamespacedName{
+				Namespace: mp.Spec.Template.Spec.InfrastructureRef.Namespace,
+				Name:      mp.Spec.Template.Spec.InfrastructureRef.Name,
+			}, ammp)).To(Succeed())
+			initialTaints := ammp.Spec.Taints
```
jackfrancis, line 80: So this setting (probably should have asked this on previous E2E tests that followed this pattern) will keep our reconcile code honest in terms of being able to handle the various combinatoric values that could be here:

- actual taints
- empty set
- zero value

I know we aren't going to have the cluster template coverage to test all of the above, but just wanna be clear that if we *do* mutate the templates going forward and in fact there is an edge case in terms of our reconciler not doing the right thing from `nil`--> some value or some other such thing that we've stubbed our toe on in the past, this E2E test will fail, right? (fail == what we want in this case :))

nojnhuh: If the template starts with a value, then this particular test will not do a nil -> non-nil update. The main hurdle with adding that step is that the from the initial non-nil to the nil state in CAPZ means "don't change the value in AKS." That might be a bit tricky to express robustly in a test since that would involve checking periodically for \*some TBD period of time\* that nothing happens https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2838#discussion_r1064898187.

Though perhaps we could still perform the update to nil, but skip checking the value in AKS normally following that. From there we can verify a nil -> non-nil update like we do now.

That still leaves the gap of verifying what happens after a non-nil -> nil update, but that's also not being tested currently and should be simple to slot in as necessary once we iron out that particular logic.

jackfrancis: ACK

PR #3049, opened 2023-01-17
fix/ diff for authorized ip ranges

bmiguel-teixeira:
**What type of PR is this?**
/kind bug

Fixes a reconcile loop bug for the Managed Control Plane in the AuthorizedIPRanges.

Current Behaviour:
Show below, this a diff bug that causes an unending reconcile loop, effectively putting the AKS cluster in a perpetual "updating" state.

image

New Behaviour:
Updated the spec `AuthorizedIPRanges` field to a pointer field (`*[]string`) much like other similar fields. This solves the diff issue, causing the cluster reconcile loop to stabilize properly.

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```release-note
fix AuthorizedIPRanges diff reconcile loop.
```

### DIFF
--- a/api/v1alpha4/azuremanagedcontrolplane_types.go
+++ b/api/v1alpha4/azuremanagedcontrolplane_types.go
@@ -157,7 +157,7 @@ type LoadBalancerProfile struct {
 type APIServerAccessProfile struct {
 	// AuthorizedIPRanges - Authorized IP Ranges to kubernetes API server.
 	// +optional
-	AuthorizedIPRanges []string `json:"authorizedIPRanges,omitempty"`
+	AuthorizedIPRanges *[]string `json:"authorizedIPRanges,omitempty"`
 	// EnablePrivateCluster - Whether to create the cluster as a private cluster or not.
 	// +optional
 	EnablePrivateCluster *bool `json:"enablePrivateCluster,omitempty"`
--- a/api/v1alpha3/conversion_test.go
+++ b/api/v1alpha3/conversion_test.go
@@ -70,9 +70,10 @@ func TestFuzzyConversion(t *testing.T) {
 	}))
 
 	t.Run("for AzureManagedControlPlane", utilconversion.FuzzTestFunc(utilconversion.FuzzTestFuncInput{
-		Scheme: scheme,
-		Hub:    &infrav1.AzureManagedControlPlane{},
-		Spoke:  &AzureManagedControlPlane{},
+		Scheme:      scheme,
+		Hub:         &infrav1.AzureManagedControlPlane{},
+		Spoke:       &AzureManagedControlPlane{},
+		FuzzerFuncs: []fuzzer.FuzzerFuncs{overrideAPIServerAccessProfileFunc},
 	}))
 
 	t.Run("for AzureManagedMachinePool", utilconversion.FuzzTestFunc(utilconversion.FuzzTestFuncInput{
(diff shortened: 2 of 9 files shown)


CecileRobertMichon:
@nojnhuh could this be related to #3009 ?

nojnhuh:
@CecileRobertMichon I suspect this is unrelated to that issue since the affected cluster there does not define `spec.apiServerAccessProfile`. I think I should've been able to reproduce that issue if this were the cause.

bmiguel-teixeira:
I only saw this issue with `spec.apiServerAccessProfile` defined (usually for private clusters). Not specifying the field did not cause this particular issues for me.

CecileRobertMichon:
@bmiguel-teixeira apologies for the conflict, we merged #2959 which changed a lot of this surface area, it might be easier to start from scratch on top of the latest changes at main.

bmiguel-teixeira:
Hi @CecileRobertMichon 

No worries. Rebased & fixed unit-tests.

CecileRobertMichon:
thanks!

/assign @nojnhuh @nawazkh

nojnhuh (approved):
/lgtm
Thanks @bmiguel-teixeira!

api/v1alpha4/azuremanagedcontrolplane_types.go
```diff
@@ -157,7 +157,7 @@ type LoadBalancerProfile struct {
 type APIServerAccessProfile struct {
 	// AuthorizedIPRanges - Authorized IP Ranges to kubernetes API server.
 	// +optional
-	AuthorizedIPRanges []string `json:"authorizedIPRanges,omitempty"`
+	AuthorizedIPRanges *[]string `json:"authorizedIPRanges,omitempty"`
```
CecileRobertMichon, line 160: /hold 

we shouldn't change the older API versions in the main branch, those are only there for conversion purposes. Youll need to remove this change and handle the conversion to pointer between v1alpha4 and v1beta1. Side node: is it really necessary to change the API? could you achieve similar results without making this field a pointer, given that in go slices have a built-in nil value? 

nawazkh (commented):
`v1beta1` and `azure/services/managedclusters/spec.go` changes look good to me, thanks for identifying this!

However, I agree that we should avoid making changes to older API versions. We should consider handling pointer conversion b/w `v1aplha4` and `v1beta1`

nawazkh:
Hey @bmiguel-teixeira , how is this fix coming along?

CecileRobertMichon:
@nawazkh @nojnhuh would one of you want to take this one over? 

It doesn't look like @bmiguel-teixeira is still actively working on it

nawazkh:
I can take this over since I was looking for something new to work on. What say @nojnhuh ?

nojnhuh:
Go for it @nawazkh! Thanks!

nawazkh:
> I only saw this issue with `spec.apiServerAccessProfile` defined (usually for private clusters). Not specifying the field did not cause this particular issues for me.

The second point in the `Limitations` sections of [api server authorized ip ranges](https://learn.microsoft.com/en-us/azure/aks/api-server-authorized-ip-ranges#limitations) states that _"The API server authorized IP ranges feature isn't supported on private clusters."_

I am wondering if this fix was intended for private clusters ? (which IMO, should be "fixed" in the first place since  authorized ip ranges is not supported for private clusters)

I tried reproducing this error on managed clusters(AKS) since this PR had a tag of managedClusters, but the AKS cluster did not run into any "updating" error as mentioned in this PR's description. However, I identified an area of improvement in here https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3553

nawazkh:
I think we can close this PR since it is being addressed by https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3639

PR #3050, opened 2023-01-17
fix unnecessary reconciles on AKS node taints

### LINKED ISSUE #3047: AKSPublicIPPrefixSpec test spec is buggy
/kind bug

@nojnhuh has already done a bit of high fidelity investigation, copied from public slack conversation:

> This is the test where we create a node pool with more nodes than available IPs in the prefix, and expect that to fail. I do see the “Ready” condition going to “Failed” like the test expects, but then the operation is immediately retried so it goes back to “Creating” and the polling in the test doesn’t happen to catch it when it’s “Failing.” In the test, it looks like we’re assuming it to fail once and not be retried. The failing runs show the expected error coming up multiple times in the controller-manager logs, but only once in a previous passing run, so that seems to agree with that assumption.
AFAICT, the error messages between the passing and failing runs are the same and being surfaced at the same point, so I’m still looking into why they’re being treated differently.
The passing run I was looking at: [https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluste[…]890/pull-cluster-api-provider-azure-e2e-exp/1613978841884135424](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/2890/pull-cluster-api-provider-azure-e2e-exp/1613978841884135424)
Failing: [https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluste[…]038/pull-cluster-api-provider-azure-e2e-exp/1614059198734143488](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/3038/pull-cluster-api-provider-azure-e2e-exp/1614059198734143488)

> It seems like both the passing and failing cases follow the path leading to [here](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/8e4ea53af9477ae79a2384d16ad561c8d5faeab8/exp/controllers/azuremanagedmachinepool_controller.go#L285), and since returning and error in Reconcile triggers a requeue, I’m not sure why we weren’t seeing this being retried before.

> And the diff between the last passing and first failing runs looks like a big nothing burger: https://github.com/kubernetes-sigs/cluster-api-provider-azure/compare/f2a45077f...8e4ea53af

**Environment:**

nojnhuh:
**What type of PR is this?**
/kind bug
/kind failing-test

**What this PR does / why we need it**: Fixes a bug where AKS node taints were constantly being reconciled when they were not defined on AzureManagedMachinePool because of a nil/empty discrepancy. More details: https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/3047#issuecomment-1386071049

Fixes #3047

**Special notes for your reviewer**: I will try to add a unit test for this after #3006.

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/azure/services/agentpools/spec.go
+++ b/azure/services/agentpools/spec.go
@@ -204,6 +204,9 @@ func (s *AgentPoolSpec) Parameters(ctx context.Context, existing interface{}) (p
 				Tags:                converters.TagsToMap(s.AdditionalTags),
 			},
 		}
+		if len(*normalizedProfile.NodeTaints) == 0 {
+			normalizedProfile.NodeTaints = nil
+		}
 
 		if s.KubeletConfig != nil {
 			normalizedProfile.KubeletConfig = &containerservice.KubeletConfig{


mboersma:
/retest

HA cluster provisioning failure in e2e job

PR #3080, opened 2023-01-24
Retry VMSS Flex validation if no parent MP is found

### LINKED ISSUE #3085: VMSS Flex clusters sometimes fail validation
/kind bug

In `make tilt-up`, deploy the "external-cloud-provider-machinepool" flavor. Some percentage of the time, the deployment will fail in webhook validation for the AzureMachinePool because the owning MachinePool isn't found.

The cluster should deploy, including the AzureMachinePool, so long as the Kubernetes version is v1.26.0 or later.

**Environment:**

mboersma:
**What type of PR is this?**

/kind bug

This retries VMSS Flex validation if an `AzureMachinePool` can't find its owner.

The webhook validation for VMSS Flex needs to involve both the `AzureMachinePool` and its owner `MachinePool`. (The first knows the orchestration mode, the second knows the Kubernetes version.)

Sometimes when applying a cluster, the AMP webhook fails to find its owner. This is returned as an error, preventing the AMP from deploying. When the owner is found, the webhook validates the Kubernetes version as expected. Retrying should make this more robust.

Fixes #3085

Is there a better way to make this lookup reliable? I'm assuming this is some sort of race condition where the YAML has been digested but objects aren't yet visible in Kubernetes, but maybe I misunderstand.

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Log VMSS Flex validation if no owner MP is found
```

### DIFF
--- a/exp/api/v1beta1/azuremachinepool_webhook.go
+++ b/exp/api/v1beta1/azuremachinepool_webhook.go
@@ -17,7 +17,6 @@ limitations under the License.
 package v1beta1
 
 import (
-	"context"
 	"fmt"
 	"reflect"
 
@@ -30,7 +29,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
 	"sigs.k8s.io/cluster-api-provider-azure/feature"
-	expv1 "sigs.k8s.io/cluster-api/exp/api/v1beta1"
+	"sigs.k8s.io/cluster-api-provider-azure/util/azure"
 	capifeature "sigs.k8s.io/cluster-api/feature"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -249,22 +248,15 @@ func (amp *AzureMachinePool) ValidateOrchestrationMode(c client.Client) func() e
 	return func() error {
 		// Only Flexible orchestration mode requires validation.
 		if amp.Spec.OrchestrationMode == infrav1.OrchestrationModeType(compute.OrchestrationModeFlexible) {
-			// Find the owner MachinePool
-			ownerMachinePool := &expv1.MachinePool{}
-			key := client.ObjectKey{
-				Namespace: amp.Namespace,
-				Name:      amp.Name,
+			parent, err := azure.FindParentMachinePoolWithRetry(amp.Name, c, 5)
+			if err != nil {
+				return errors.Wrap(err, "failed to find parent MachinePool")
 			}
-			ctx := context.Background()
-			if err := c.Get(ctx, key, ownerMachinePool); err != nil {
-				return errors.Wrap(err, "failed to get owner MachinePool")
-			}
-
 			// Kubernetes must be >= 1.26.0 for cloud-provider-azure Helm chart support.
-			if ownerMachinePool.Spec.Template.Spec.Version == nil {
+			if parent.Spec.Template.Spec.Version == nil {
 				return errors.New("could not find Kubernetes version in MachinePool")
 			}
-			k8sVersion, err := semver.ParseTolerant(*ownerMachinePool.Spec.Template.Spec.Version)
+			k8sVersion, err := semver.ParseTolerant(*parent.Spec.Template.Spec.Version)
 			if err != nil {
 				return errors.Wrap(err, "failed to parse Kubernetes version")
 			}
--- a/util/azure/azure.go
+++ b/util/azure/azure.go
@@ -17,8 +17,10 @@ limitations under the License.
 package azure
 
 import (
+	"context"
 	"os"
 	"strings"
+	"time"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud"
@@ -27,6 +29,9 @@ import (
 	azureautorest "github.com/Azure/go-autorest/autorest/azure"
 	"github.com/Azure/go-autorest/autorest/azure/auth"
 	"github.com/jongio/azidext/go/azidext"
+	"github.com/pkg/errors"
+	expv1 "sigs.k8s.io/cluster-api/exp/api/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // AzureSystemNodeLabelPrefix is a standard node label prefix for Azure features, e.g., kubernetes.azure.com/scalesetpriority.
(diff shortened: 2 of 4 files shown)


exp/api/v1beta1/azuremachinepool_webhook.go
```diff
@@ -257,7 +257,9 @@ func (amp *AzureMachinePool) ValidateOrchestrationMode(c client.Client) func() e
 			}
 			ctx := context.Background()
 			if err := c.Get(ctx, key, ownerMachinePool); err != nil {
```
jackfrancis, line 259: would retrying the get be another option to deal with this? are there terminal situations where the owner isn't found, or are they seemingly transient?

mboersma: My guess is they are transient situations, so perhaps a retry would help. I'll give it a go–it's easy to reproduce here.

jackfrancis: Maybe @CecileRobertMichon could weigh in from the capz side as to the more CAPI-idiomatic way to deal with a retryable error of this type.

1) return early and requeue a follow-up reconciliation
vs
2) retry in the inner loop of the active reconciliation

Is one of the above definitely more preferred in this scenario?

cc @fabriziopandini @sbueringer just curious about your thoughts

CecileRobertMichon:
Does this PR fix an issue? I see it's marked as a bug

exp/api/v1beta1/azuremachinepool_webhook.go
CecileRobertMichon: I agree retrying would be more resilient than simply ignoring the validation, which might result in invalid configurations being created. This is the webhook so there is no such thing as requeuing. 

Side question: Didn't catch this earlier in the PR that added this, but is the above assuming that the MachinePool name and AzureMachinePool name are the same? If so not sure that's always a valid assumption, there's nothing that prevents a user from naming them differently.

CecileRobertMichon: @willie-yao had to implement a helper function do find the parent MachinePool for his PR that you might be able to leverage https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2965/files#diff-88cb69b80a9047f0e0cd188951df71c621d843620b72bd52628899b14644209eR100

also the same issue here of the MachinePool not always being found might apply to your scenario @willie-yao FYI

jackfrancis: Ah, sorry, yes, we're in a webhook, so we're actually enforcing whether or not to initiate the series of events that will result in a potential reconciliation.

The only thing against a retry here is the UX expectation that a webhook would be responsive, so let's not retry for very long.

mboersma: > is the above assuming that the MachinePool name and AzureMachinePool name are the same?

It is. I'll take a look at @willie-yao's helper func.

Yeah since we're in a webhook AFAICT we can only write our own retry logic and do our best to keep it brief.

CecileRobertMichon: question is: if it fails once, will retrying help? because if objects are created serially then the MachinePool not being present will block the AzureMachinePool from creating which means it will create a deadlock

jackfrancis: Does this imply that Cluster<---<Provider>Cluster (and other similar CAPI resource<---Provider resource relationships) type validations cannot be implemented via webhooks because we assume that such tightly coupled resource specs are conventionally delivered simultaneously to the apiserver as part of an overall "cluster" template (e.g. multi-resource yaml file)?

mboersma:
/retitle Retry VMSS Flex validation if no parent MP is found

exp/api/v1beta1/azuremachinepool_webhook.go
mboersma: Retrying does seem to fix things at least in my local testing. Often the parent MP is found on the first try, but sometimes I see this before success:

```
I0125 18:10:52.499411     190 azuremachinepool_webhook.go:258] AzureMachinePoolLogger "msg"="finding parent MachinePool" "attempt"=1
I0125 18:10:53.500315     190 azuremachinepool_webhook.go:258] AzureMachinePoolLogger "msg"="finding parent MachinePool" "attempt"=2
I0125 18:10:54.501515     190 azuremachinepool_webhook.go:258] AzureMachinePoolLogger "msg"="finding parent MachinePool" "attempt"=3
```

mboersma: I still wonder if, after five failed tries, we should return an error or simply log it. Currently we return an error and fail the deployment, but given the loose linkage between the two objects, you could argue that this should be a "best effort" validation.

CecileRobertMichon: IMO we should return an error, a webhook validation is not meant to be "best effort", especially if we have evidence that a retry does fix things in some cases.

exp/api/v1beta1/azuremachinepool_webhook.go
```diff
@@ -249,22 +250,27 @@ func (amp *AzureMachinePool) ValidateOrchestrationMode(c client.Client) func() e
...
 		// Only Flexible orchestration mode requires validation.
 		if amp.Spec.OrchestrationMode == infrav1.OrchestrationModeType(compute.OrchestrationModeFlexible) {
-			// Find the owner MachinePool
-			ownerMachinePool := &expv1.MachinePool{}
-			key := client.ObjectKey{
-				Namespace: amp.Namespace,
-				Name:      amp.Name,
+			// Find the parent MachinePool
+			var parent *expv1.MachinePool
+			maxAttempts := 5
```
CecileRobertMichon, line 255: would it make sense to implement the retry inside the `FindParentMachinePool` helper function so that @willie-yao doesn't have to repeat this retry logic in #2965 ? Or if we don't want to retry in utils, add a wrapper helper func in webhooks that does the fetch + retry

mboersma: I don't know whether retry is appropriate for that PR, but I can move the retry logic into a wrapper function so its reusable.

exp/api/v1beta1/azuremachinepool_webhook.go
sbueringer: I think a short retry is probably fine. I have no idea how your webhook works but you're probably using a cached client. Which essentially means all get/list/requests could get their date from a cache with is slightly stale (usually catches up in <1s)

exp/api/v1beta1/azuremachinepool_webhook.go
```diff
@@ -249,22 +250,27 @@ func (amp *AzureMachinePool) ValidateOrchestrationMode(c client.Client) func() e
 	return func() error {
 		// Only Flexible orchestration mode requires validation.
```
sbueringer, line 251: Just fyi. A while back we implemented CustomDefaulter/CustomValidator in controller-runtime.

This should make large parts of the CAPZ webhook util redundant.

Just in case you want to follow-up on that in a separate PR.

Let me know if you need/want more information

mboersma: Thanks for noticing this! I opened #3103.

exp/api/v1beta1/azuremachinepool_webhook.go
willie-yao: I think we can implement this inside of the helper function, with an input for how many retries. For my own understanding: why is a retry necessary in this case? Is it because the parent machinepool may not be created at this point?

willie-yao: I implemented a [wrapper](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2965/commits/28eaa3755e32b6aa226a7e1a9d5baa99d0c399b3#diff-88cb69b80a9047f0e0cd188951df71c621d843620b72bd52628899b14644209eR116) since there are two places where the error can occur.

mboersma: > why is a retry necessary in this case?

My best guess is that it's what @sbueringer mentioned above: we're using a caching `client.Client` that needs a beat to find the `MachinePool` that was just created (just before the `AzureMachinePool` that triggered this webhook).

Either that, or the actual creation is somehow asynchronous and we're waiting for a controller to finish creating the `MachinePool` object. That seems less likely, because you can see the log line "machinepool.x.y.z yadda yadda created" printed out before the `AzureMachinePool` is processed and this webhook is called.

mboersma:
Updated to steal more of @willie-yao's helper code and add some basic unit tests for those funcs.

util/azure/azure.go
```diff
@@ -27,6 +29,9 @@ import (
 	azureautorest "github.com/Azure/go-autorest/autorest/azure"
 	"github.com/Azure/go-autorest/autorest/azure/auth"
 	"github.com/jongio/azidext/go/azidext"
+	"github.com/pkg/errors"
+	expv1 "sigs.k8s.io/cluster-api/exp/api/v1beta1"
```
CecileRobertMichon, line 33: how does this not create an import cycle? since exp/api/v1beta1 imports this to use the `FindParentMachinePoolWithRetry` func

willie-yao: This is because it's importing from CAPI's exp/api/v1beta1 since we only need to find the parent `MachinePool`. For `GetSubscriptionID`, we need access to the `AzureMachine` type from CAPZ.

mboersma:
I'm running into this bug often when trying to test other VMSS Flex scenarios. Could someone review this again when time permits?

willie-yao (approved):
/lgtm

CecileRobertMichon (commented):
/lgtm
/approve

mboersma:
/retest

AKS cluster provisioning timeout.

PR #3086, opened 2023-01-24
E2E: retries in CopyConfigMap func

jackfrancis:
**What type of PR is this?**

/kind flake

In response to an observed E2E flake:

https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/periodic-cluster-api-provider-azure-e2e-main/1617762815735500800

Let's retry the get/create flow that actually implements "copy" for us, as both calls are subject to transient errors during transmission/delivery.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/helpers.go
+++ b/test/e2e/helpers.go
@@ -986,14 +986,16 @@ func InstallHelmChart(ctx context.Context, input clusterctl.ApplyClusterTemplate
 	}
 }
 
-func CopyConfigMap(ctx context.Context, cl client.Client, cmName, fromNamespace, toNamespace string) {
+func CopyConfigMap(ctx context.Context, input clusterctl.ApplyClusterTemplateAndWaitInput, cl client.Client, cmName, fromNamespace, toNamespace string) {
 	cm := &corev1.ConfigMap{}
-	Expect(cl.Get(ctx, client.ObjectKey{Name: cmName, Namespace: fromNamespace}, cm)).To(Succeed())
-	cm.SetNamespace(toNamespace)
-	cm.SetResourceVersion("")
-	framework.EnsureNamespace(ctx, cl, toNamespace)
-	err := cl.Create(ctx, cm.DeepCopy())
-	if !apierrors.IsAlreadyExists(err) {
-		Expect(err).To(Succeed())
-	}
+	Eventually(func(g Gomega) {
+		g.Expect(cl.Get(ctx, client.ObjectKey{Name: cmName, Namespace: fromNamespace}, cm)).To(Succeed())
+		cm.SetNamespace(toNamespace)
+		cm.SetResourceVersion("")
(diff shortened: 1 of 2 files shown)


test/e2e/helpers.go
```diff
@@ -1019,14 +1019,16 @@ func createClusterWithControlPlaneWaiters(ctx context.Context, configCluster clu
...
 	cm := &corev1.ConfigMap{}
-	Expect(cl.Get(ctx, client.ObjectKey{Name: cmName, Namespace: fromNamespace}, cm)).To(Succeed())
-	cm.SetNamespace(toNamespace)
-	cm.SetResourceVersion("")
-	framework.EnsureNamespace(ctx, cl, toNamespace)
-	err := cl.Create(ctx, cm.DeepCopy())
-	if !apierrors.IsAlreadyExists(err) {
-		Expect(err).To(Succeed())
-	}
+	Eventually(func(g Gomega) {
```
CecileRobertMichon, line 1024: I see we have two patterns in the code:

1. The most common:

```
Eventually(func() error {
  err := doThing()
  return err
}, intervals).Should(Succeed())
```

2. What seems to have been introduced more recently in e2e retries, and this:

```
Eventually(func(g Gomega) {
  g.Expect(doThing()).To(Succeed())
}, intervals).Should(Succeed())
```

Are there any reasons for preferring one over the other?

jackfrancis: Thanks for bringing this up, I think that more important than preferring one over the other (IMO neither option is materially better or worse) is consistency.

That said, I prefer the 2nd as passing in the Gomega object allows us to continue to express our test flows in idiomatic ginkgo (though arguably not idiomatic go but I digress) and not go back and forth between ginkgo and vanilla go in a single func.

Would love to hear if any other maintainers have ideas, it would be trivial (and IMO beneficial) to do a scrub of our E2E code and standardize to one pattern or the other.

cc @mboersma @devigned

nojnhuh: I'm also +1 for preferring the `func (g Gomega)` flavor since I think that makes things a bit more internally consistent. I think the bar for breaking that rule should be pretty low though.

CecileRobertMichon: I don't think this conversation is blocking this PR, the inconsistency already exists before this change, so feel free to move it to an issue if that's easier

jackfrancis: Since it's easy work (Yes, Please!) I'll do a PR that sets this as the standard pattern in `test/e2e` after this merges.

mboersma (approved):
/lgtm
/approve

jackfrancis:
My vote is to not worry about manual cherry-picks as we don't see this underlying flake often enough to warrant that add'l effort.

PR #3091, opened 2023-01-25
Fix custom builds kube pending pods

### LINKED ISSUE #3082: k/k custom build presubmits pods pending
**Which jobs are failing:**

All k/k presubmit jobs using the custom builds template

**Which tests are failing:**

scripts/ci-entrypoint.sh

**Since when has it been failing:**

unknown. The actual job started failing in 1.7 because we added a check to wait for ready kube-system pods, but the same symptoms are reproducible with 1.6. 

**Testgrid link:**

https://testgrid.k8s.io/provider-azure-presubmit#pull-kubernetes-e2e-capz-azure-disk-1-7

kube-apiserver, kube-controller-manager and kube-scheduler pods appear pending. However, it's possible this is an etcd issue as the apiserver is reachable and other pods are getting scheduled.

```
NAMESPACE     NAME                                                      READY   STATUS    RESTARTS   AGE
kube-system   calico-kube-controllers-657b584867-dvjgq                  1/1     Running   0          11m
kube-system   calico-node-2qgr7                                         1/1     Running   0          10m
kube-system   calico-node-rw66d                                         1/1     Running   0          10m
kube-system   calico-node-z4gz4                                         1/1     Running   0          11m
kube-system   coredns-56f4c55bf9-b5zzw                                  1/1     Running   0          11m
kube-system   coredns-56f4c55bf9-plnvf                                  1/1     Running   0          11m
kube-system   etcd-capz-kuk63g-control-plane-n7qcq                      1/1     Running   0          11m
kube-system   kube-apiserver-capz-kuk63g-control-plane-n7qcq            0/1     Pending   0          11m
kube-system   kube-controller-manager-capz-kuk63g-control-plane-n7qcq   0/1     Pending   0          11m
kube-system   kube-proxy-54pz6                                          1/1     Running   0          11m
kube-system   kube-proxy-9mmdb                                          1/1     Running   0          10m
kube-system   kube-proxy-hb99g                                          1/1     Running   0          10m
kube-system   kube-scheduler-capz-kuk63g-control-plane-n7qcq            0/1     Pending   0          11m
kube-system   metrics-server-c9574f845-v7xfw                            1/1     Running   0          11m
``` 

/kind failing-test

CecileRobertMichon:
**What type of PR is this?**

/kind failing-test

**What this PR does / why we need it**: It appears that static pods are not getting properly restarted after their `/etc/kubernetes/manifests/` yaml file gets updated. This results in `Pending` kube-apiserver, kube-controller-manager, and kube-scheduler pods when running the ci-entrypoint.sh script with `TEST_K8s==true` (or in a k/k presubmit job). 

This applies two fixes to address the symptoms:

1) add a kubelet restart after changing the manifests on the control-plane nodes
2) add a wait for apiserver pod to be ready in the ci-entrypoint.sh script to avoid race conditions where kubelet gets restarted after we start interacting with the apiserver (installing addons).

Note that this doesn't cause the root cause. This may or may not be a kubelet bug.

Fixes #3082 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```


mboersma:
/retest

Linter timeout.

scripts/ci-entrypoint.sh
```diff
@@ -139,6 +139,10 @@ install_addons() {
     # export the target cluster KUBECONFIG if not already set
     export KUBECONFIG="${KUBECONFIG:-${PWD}/kubeconfig}"
 
+    # wait for the apiserver pod to be Ready.
+    APISERVER_POD=$(${KUBECTL} get pods -n kube-system -o name | grep apiserver)
+    "${KUBECTL}" wait --for=condition=Ready -n kube-system "${APISERVER_POD}" --timeout=5m
+
```
sonasingh46, line 145: Thinking if a kubectl describe of pod will help in useful logs after the timeout.

CecileRobertMichon: that's already part of the cleanup step https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3091/files#diff-2009f7be5363812e22862771f557363795759e67b3ca2e32c50d9fd5f8c5669bR260

scripts/ci-entrypoint.sh
```diff
@@ -139,6 +139,10 @@ install_addons() {
     # export the target cluster KUBECONFIG if not already set
     export KUBECONFIG="${KUBECONFIG:-${PWD}/kubeconfig}"
 
+    # wait for the apiserver pod to be Ready.
+    APISERVER_POD=$(${KUBECTL} get pods -n kube-system -o name | grep apiserver)
+    "${KUBECTL}" wait --for=condition=Ready -n kube-system "${APISERVER_POD}" --timeout=5m
```
jackfrancis, line 144: Are we confident that this wait for a single observation of `Ready` is sufficient to deal w/ observed apiserver weirdness after updating static pods? More specifically, are we confident that this wait check will trigger *after* the control plane static pod spec files have been modified?

CecileRobertMichon: Yes, the apiserver pod stays pending until it's ready and then it's ready for good

That's why I changed it to this instead of retrying actually because it's the best signal of "things are stable and ready for good now" from what I've observed. If we add retries instead we'd have to add retries to every single place that talks to the apiserver because there is no way of controlling the timing of when the apiserver will become temporarily unavailable before that pod `Ready` transition happens.

CecileRobertMichon: I can try that locally and see what happens

scripts/ci-entrypoint.sh
CecileRobertMichon: actually I lied, the logger collects events and pod logs, but trying to see if something like [this](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3097) could work to print a full kubectl describe pod in artifacts instead of just events

scripts/ci-entrypoint.sh
CecileRobertMichon: yeah that's equivalent, pushed the changes

scripts/ci-entrypoint.sh
CecileRobertMichon: yep that works, see #3097

jackfrancis (approved):
/lgtm
/approve

jackfrancis:
/retest

```
timeout --foreground 600 bash -c "while ! /home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/hack/tools/bin/kubectl-v1.22.4 --kubeconfig=./kubeconfig get nodes | grep control-plane; do sleep 1; done"
error: the server doesn't have a resource type "nodes"
```

ci-entrypoint.sh flake

Probably not worth updating the timeout value (time between when CAPI mints a workload cluster kubeconfig and when the infra has been created + kubeadm init succeeds and registers itself as a node w/ a functioning apiserver) unless we see this flake more regularly.

Issue #2719, opened 2022-10-11
async scalesets refactor


CecileRobertMichon:
We should refactor the scaleset service to reuse the `async.go` framework instead of using its own implementation.

This might need to wait until #999 is done first so it doesn't interfere.

PR #3096, opened 2023-01-25
Don't delete VMSS upon failure and add bootstrap status condition

### LINKED ISSUE #3087: VMSS deleted and recreated when AzureMachinePool ProvisioningState status field is Failed
/kind bug

There are likely multiple ways to enter into this condition, but this happened to us via the following path:
1. cluster-autoscaler scales up a VMSS, increases capacity field
2. the new VM(s) fail to bootstrap and join the cluster
3. The Azure VMSS treats that as a terminal failure, and sets its `provisioningState` field to `Failed`
4. Next `Reconcile()`, CAPZ reads that field and writes it into the `BootstrapSucceeded` condition
5. Then it hits [this case in the switch](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/9b2bf83f9077747306d233b647557a34fac9e87e/exp/controllers/azuremachinepool_controller.go#L317-L322) which triggers CAPZ to delete the VMSS and recreate it.

I would expect this state to be logged and for an event to be recorded

N/A

**Environment:**

- cluster-api-provider-azure version: `v1.7.0`
- Kubernetes version: (use `kubectl version`): `v1.21.14`

jamesgoodhouse:
**What type of PR is this?**
/kind bug

Changes logic during VMSS scale failure to log and record an event instead of deleting to VMSS. This will prevent healthy nodes in an existing VMSS from being deleted if a scale event fails. This also adds a new condition for the bootstrap VM extension.

Fixes #3087

N/A

**TODOs**:

- [x] squashed commits

**Release note**:

```release-note
NONE
```

### DIFF
--- a/azure/converters/vmss.go
+++ b/azure/converters/vmss.go
@@ -121,6 +121,16 @@ func SDKToVMSSVM(sdkInstance compute.VirtualMachineScaleSetVM) *azure.VMSSVM {
 		instance.Name = *sdkInstance.OsProfile.ComputerName
 	}
 
+	if sdkInstance.Resources != nil {
+		for _, r := range *sdkInstance.Resources {
+			if r.ProvisioningState != nil && r.Name != nil &&
+				(*r.Name == azure.BootstrappingExtensionLinux || *r.Name == azure.BootstrappingExtensionWindows) {
+				instance.BootstrappingState = infrav1.ProvisioningState(pointer.StringDeref(r.ProvisioningState, ""))
+				break
+			}
+		}
+	}
+
 	if sdkInstance.StorageProfile != nil && sdkInstance.StorageProfile.ImageReference != nil {
 		imageRef := sdkInstance.StorageProfile.ImageReference
 		instance.Image = SDKImageToImage(imageRef, sdkInstance.Plan != nil)
--- a/azure/defaults.go
+++ b/azure/defaults.go
@@ -52,6 +52,13 @@ const (
 )
 
 const (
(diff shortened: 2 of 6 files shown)


jamesgoodhouse:
@mboersma looks like it's not wanting to pass tests (2 failures now, but both different). I don't believe this has anything to do with my changes, but I'll defer to you on this.

jamesgoodhouse:
as per comments in #3087, i'm going to see about making some changes to the logic for deleting failed VMs

jamesgoodhouse:
hmm, seems to be stuck running the `pull-cluster-api-provider-azure-e2e-aks` job. tried triggering a /retest, but likely only accepts that command upon test completion. maybe just a super long-running test; will check back later.

jamesgoodhouse:
PR is good to go for review/feedback! Please and thank you :-)

CecileRobertMichon (commented):
/lgtm
/assign @mboersma @Jont828 

@Jont828 the bootstrap extension condition part of this PR may impact #2719 

Should we cherry-pick this?

mboersma (approved):
/lgtm
/approve

> should we cherry-pick this?

It's a bug fix, so I think so.

mboersma:
`azure/converters/vmss.go:128:61: undefined: to`

I think this needs to catch up with the changes in #3026. It's still using go-autorest/to.

jamesgoodhouse:
@CecileRobertMichon / @mboersma — i may have messed that up making the change locally and force pushing

CecileRobertMichon (commented):
/lgtm
/approve

thanks @jamesgoodhouse 🎉

PR #3111, opened 2023-01-27
Make scaleset reconcile/delete async

### LINKED ISSUE #2719: async scalesets refactor

Jont828:
**What type of PR is this?**

/kind feature
**What this PR does / why we need it**: Implementation of an async service for managed agent pools as part of an effort to make all services async. See #1610 and #1541.

Fixes #2719 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Make scalesets reconcile/delete async
```

### DIFF
--- a/azure/scope/machinepool.go
+++ b/azure/scope/machinepool.go
@@ -62,7 +62,7 @@ type (
 		MachinePool      *expv1.MachinePool
 		AzureMachinePool *infrav1exp.AzureMachinePool
 		ClusterScope     azure.ClusterScoper
-		Caches           *MachinePoolCache
+		Cache            *MachinePoolCache
 	}
 
 	// MachinePoolScope defines a scope defined around a machine pool and its cluster.
@@ -77,15 +77,19 @@ type (
 		cache                      *MachinePoolCache
 	}
 
-	// MachinePoolCache stores common machine pool information so we don't have to hit the API multiple times within the same reconcile loop.
-	MachinePoolCache struct {
-		VMSKU resourceskus.SKU
-	}
-
 	// NodeStatus represents the status of a Kubernetes node.
 	NodeStatus struct {
 		Ready   bool
 		Version string
+	}
+
+	// MachinePoolCache stores common machine pool information so we don't have to hit the API multiple times within the same reconcile loop.
+	MachinePoolCache struct {
+		BootstrapData           string
+		HasBootstrapDataChanges bool
+		VMImage                 *infrav1.Image
+		VMSKU                   resourceskus.SKU
+		MaxSurge                int
 	}
 )
 
@@ -133,6 +137,27 @@ func (m *MachinePoolScope) InitMachinePoolCache(ctx context.Context) error {
 		var err error
 		m.cache = &MachinePoolCache{}
 
+		m.cache.BootstrapData, err = m.GetBootstrapData(ctx)
+		if err != nil {
+			return err
+		}
+
+		m.cache.HasBootstrapDataChanges, err = m.HasBootstrapDataChanges(ctx)
+		if err != nil {
+			return err
+		}
+
+		m.cache.VMImage, err = m.GetVMImage(ctx)
+		if err != nil {
+			return err
+		}
+		m.SaveVMImageToStatus(m.cache.VMImage)
+
+		m.cache.MaxSurge, err = m.MaxSurge()
+		if err != nil {
+			return err
+		}
+
 		skuCache, err := resourceskus.GetCache(m, m.Location())
 		if err != nil {
 			return err
@@ -148,9 +173,19 @@ func (m *MachinePoolScope) InitMachinePoolCache(ctx context.Context) error {
 }
 
 // ScaleSetSpec returns the scale set spec.
-func (m *MachinePoolScope) ScaleSetSpec() azure.ScaleSetSpec {
-	return azure.ScaleSetSpec{
+func (m *MachinePoolScope) ScaleSetSpec(ctx context.Context) azure.ResourceSpecGetter {
+	ctx, log, done := tele.StartSpanWithLogger(ctx, "scope.MachinePoolScope.ScaleSetSpec")
+	defer done()
+
+	shouldPatchCustomData := false
+	if m.HasReplicasExternallyManaged(ctx) {
+		shouldPatchCustomData = m.cache.HasBootstrapDataChanges
+		log.V(4).Info("has bootstrap data changed?", "shouldPatchCustomData", shouldPatchCustomData)
+	}
+
+	return &scalesets.ScaleSetSpec{
 		Name:                         m.Name(),
+		ResourceGroup:                m.ResourceGroup(),
 		Size:                         m.AzureMachinePool.Spec.Template.VMSize,
 		Capacity:                     int64(ptr.Deref[int32](m.MachinePool.Spec.Replicas, 0)),
 		SSHKeyData:                   m.AzureMachinePool.Spec.Template.SSHPublicKey,
@@ -172,6 +207,16 @@ func (m *MachinePoolScope) ScaleSetSpec() azure.ScaleSetSpec {
 		NetworkInterfaces:            m.AzureMachinePool.Spec.Template.NetworkInterfaces,
 		IPv6Enabled:                  m.IsIPv6Enabled(),
 		OrchestrationMode:            m.AzureMachinePool.Spec.OrchestrationMode,
+		Location:                     m.AzureMachinePool.Spec.Location,
+		SubscriptionID:               m.SubscriptionID(),
+		VMSSExtensionSpecs:           m.VMSSExtensionSpecs(),
+		HasReplicasExternallyManaged: m.HasReplicasExternallyManaged(ctx),
+		ClusterName:                  m.ClusterName(),
+		AdditionalTags:               m.AzureMachinePool.Spec.AdditionalTags,
+		SKU:                          m.cache.VMSKU,
+		VMImage:                      m.cache.VMImage,
+		BootstrapData:                m.cache.BootstrapData,
+		ShouldPatchCustomData:        shouldPatchCustomData,
 	}
 }
 
@@ -614,11 +659,8 @@ func (m *MachinePoolScope) GetBootstrapData(ctx context.Context) (string, error)
 }
 
 // calculateBootstrapDataHash calculates the sha256 hash of the bootstrap data.
-func (m *MachinePoolScope) calculateBootstrapDataHash(ctx context.Context) (string, error) {
-	bootstrapData, err := m.GetBootstrapData(ctx)
-	if err != nil {
-		return "", err
-	}
+func (m *MachinePoolScope) calculateBootstrapDataHash(_ context.Context) (string, error) {
+	bootstrapData := m.cache.BootstrapData
 	h := sha256.New()
 	n, err := io.WriteString(h, bootstrapData)
 	if err != nil || n == 0 {
--- a/azure/services/roleassignments/roleassignments.go
+++ b/azure/services/roleassignments/roleassignments.go
@@ -47,7 +47,7 @@ type Service struct {
 	Scope                 RoleAssignmentScope
 	virtualMachinesGetter async.Getter
 	async.Reconciler
-	virtualMachineScaleSetClient scalesets.Client
+	virtualMachineScaleSetGetter async.Getter
 }
 
 // New creates a new service.
@@ -56,7 +56,7 @@ func New(scope RoleAssignmentScope) *Service {
 	return &Service{
 		Scope:                        scope,
 		virtualMachinesGetter:        virtualmachines.NewClient(scope),
-		virtualMachineScaleSetClient: scalesets.NewClient(scope),
+		virtualMachineScaleSetGetter: scalesets.NewClient(scope),
 		Reconciler:                   async.New(scope, client, client),
 	}
 }
@@ -141,10 +141,20 @@ func (s *Service) getVMSSPrincipalID(ctx context.Context) (*string, error) {
 	ctx, log, done := tele.StartSpanWithLogger(ctx, "roleassignments.Service.getVMPrincipalID")
 	defer done()
 	log.V(2).Info("fetching principal ID for VMSS")
-	resultVMSS, err := s.virtualMachineScaleSetClient.Get(ctx, s.Scope.ResourceGroup(), s.Scope.Name())
+	spec := &scalesets.ScaleSetSpec{
+		Name:          s.Scope.Name(),
+		ResourceGroup: s.Scope.ResourceGroup(),
+	}
+
+	resultVMSSIface, err := s.virtualMachineScaleSetGetter.Get(ctx, spec)
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to get principal ID for VMSS")
 	}
+	resultVMSS, ok := resultVMSSIface.(compute.VirtualMachineScaleSet)
+	if !ok {
+		return nil, errors.Errorf("%T is not a compute.VirtualMachineScaleSet", resultVMSSIface)
+	}
+
 	return resultVMSS.Identity.PrincipalID, nil
 }
 
--- a/azure/services/scalesets/scalesets.go
+++ b/azure/services/scalesets/scalesets.go
@@ -18,21 +18,18 @@ package scalesets
 
 import (
 	"context"
-	"encoding/base64"
 	"fmt"
-	"strconv"
-	"time"
 
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2021-11-01/compute"
 	"github.com/pkg/errors"
-	"k8s.io/utils/ptr"
 	azprovider "sigs.k8s.io/cloud-provider-azure/pkg/provider"
 	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
 	"sigs.k8s.io/cluster-api-provider-azure/azure"
 	"sigs.k8s.io/cluster-api-provider-azure/azure/converters"
+	"sigs.k8s.io/cluster-api-provider-azure/azure/services/async"
 	"sigs.k8s.io/cluster-api-provider-azure/azure/services/resourceskus"
 	azureutil "sigs.k8s.io/cluster-api-provider-azure/util/azure"
-	"sigs.k8s.io/cluster-api-provider-azure/util/generators"
+	"sigs.k8s.io/cluster-api-provider-azure/util/reconciler"
 	"sigs.k8s.io/cluster-api-provider-azure/util/slice"
 	"sigs.k8s.io/cluster-api-provider-azure/util/tele"
 )
@@ -44,18 +41,12 @@ type (
 	ScaleSetScope interface {
 		azure.ClusterDescriber
 		azure.AsyncStatusUpdater
-		GetBootstrapData(context.Context) (string, error)
-		GetVMImage(context.Context) (*infrav1.Image, error)
-		SaveVMImageToStatus(*infrav1.Image)
-		MaxSurge() (int, error)
-		ScaleSetSpec() azure.ScaleSetSpec
+		ScaleSetSpec(context.Context) azure.ResourceSpecGetter
 		VMSSExtensionSpecs() []azure.ResourceSpecGetter
 		SetAnnotation(string, string)
 		SetProviderID(string)
 		SetVMSSState(*azure.VMSS)
 		ReconcileReplicas(context.Context, *azure.VMSS) error
-		HasReplicasExternallyManaged(context.Context) bool
-		HasBootstrapDataChanges(context.Context) (bool, error)
 	}
 
 	// Service provides operations on Azure resources.
@@ -63,13 +54,16 @@ type (
 		Scope ScaleSetScope
 		Client
 		resourceSKUCache *resourceskus.Cache
+		async.Reconciler
 	}
 )
 
 // New creates a new service.
 func New(scope ScaleSetScope, skuCache *resourceskus.Cache) *Service {
+	client := NewClient(scope)
 	return &Service{
-		Client:           NewClient(scope),
+		Reconciler:       async.New(scope, client, client),
+		Client:           client,
 		Scope:            scope,
 		resourceSKUCache: skuCache,
 	}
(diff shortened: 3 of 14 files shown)


azure/scope/machinepool.go
```diff
@@ -117,9 +128,77 @@ func NewMachinePoolScope(params MachinePoolScopeParams) (*MachinePoolScope, erro
...
+
+	maxSurge, err := m.MaxSurge()
+	if err != nil {
+		log.Error(err, "failed to get max surge")
+		// TODO: do we return early here?
+	}
+
+	shouldPatchCustomData := false
+	if m.HasReplicasExternallyManaged(ctx) {
+		shouldPatchCustomData, err := m.HasBootstrapDataChanges(ctx)
```
mweibel, line 189: be aware here the assignment doesn't overwrite `shouldPatchCustomData` of the outer scope.

Jont828:
@CecileRobertMichon Could I get a review on this when you have the chance?

azure/services/scalesets/scalesets.go
```diff
@@ -84,32 +82,73 @@ func (s *Service) Reconcile(ctx context.Context) (retErr error) {
 	ctx, log, done := tele.StartSpanWithLogger(ctx, "scalesets.Service.Reconcile")
 	defer done()
 
+	// TODO: do we want this timeout?
```
CecileRobertMichon, line 85: This is a timeout for the overall service

azure/services/scalesets/spec.go
```diff
@@ -0,0 +1,525 @@
...
+			}
+
+			if disk.ManagedDisk.DiskEncryptionSet != nil {
+				dataDisks[i].ManagedDisk.DiskEncryptionSet = &compute.DiskEncryptionSetParameters{ID: pointer.String(disk.ManagedDisk.DiskEncryptionSet.ID)}
+			}
+		}
+	}
+	storageProfile.DataDisks = &dataDisks
+
+	// TODO: Find a way to save this to status, likely after CreateResource() is called.
```
CecileRobertMichon, line 423: this is still TODO

CecileRobertMichon (commented):
@Jont828 looking good so far but there are lots of TODOs and comments in code that make it seem like this is still in progress, can you please clean those up and only leave review comments for the things you still need to do ?

Jont828:
@CecileRobertMichon I've cleaned up the comments/TODOs, how does this look?

CecileRobertMichon:
> /cc @willie-yao @nawazkh

Thanks @Jont828, I requested review from reviewers above

controllers/nodelister_mock.go
```diff
@@ -16,6 +16,7 @@ limitations under the License.
 
 // Code generated by MockGen. DO NOT EDIT.
 // Source: ./azuremanagedmachinepool_reconciler.go
+// TODO: Add this to make generate target. So far it seems impossible to update this using generation.
```
willie-yao, line 19: Is this TODO intended to be left here? If so, it could help to create an issue so it isn't forgotten.

azure/services/scalesets/spec.go
```diff
@@ -0,0 +1,522 @@
...
+	BootstrapData                string
+	VMSSInstances                []compute.VirtualMachineScaleSetVM
+	MaxSurge                     int
+	ClusterName                  string
+	ShouldPatchCustomData        bool
+	HasReplicasExternallyManaged bool
+	AdditionalTags               infrav1.Tags
+}
+
+// ResourceName returns the name of the subnet.
```
willie-yao, line 77: Is this referring to the name of the subnet or scaleset?

azure/services/scalesets/spec.go
```diff
@@ -0,0 +1,522 @@
...
+func (s *ScaleSetSpec) ResourceName() string {
+	return s.Name
+}
+
+// ResourceGroupName returns the name of the resource group of the VNet that owns this subnet.
```
willie-yao, line 82: Same comment as above as well as line 87

azure/services/scalesets/spec.go
```diff
@@ -0,0 +1,522 @@
...
+	// Decreases in replica count is handled by deleting AzureMachinePoolMachine instances in the MachinePoolScope
+	if *vmss.Sku.Capacity <= existingInfraVMSS.Capacity && !hasModelChanges && !s.ShouldPatchCustomData {
+		// up to date, nothing to do
+		return nil, nil
+	}
+
+	return vmss, nil
+}
+
+// Parameters returns the parameters for the subnet.
```
willie-yao, line 135: Same comment as above

willie-yao (commented):
This looks good to me! Just a few minor comments. Great work @Jont828

azure/services/scalesets/spec.go
```diff
@@ -0,0 +1,522 @@
...
+	case compute.OrchestrationModeUniform:
+		vmss.VirtualMachineScaleSetProperties.Overprovision = pointer.Bool(false)
+		vmss.VirtualMachineScaleSetProperties.UpgradePolicy = &compute.UpgradePolicy{Mode: compute.UpgradeModeManual}
+	case compute.OrchestrationModeFlexible:
+		vmss.VirtualMachineScaleSetProperties.VirtualMachineProfile.NetworkProfile.NetworkAPIVersion =
+			compute.NetworkAPIVersionTwoZeroTwoZeroHyphenMinusOneOneHyphenMinusZeroOne
+		vmss.VirtualMachineScaleSetProperties.PlatformFaultDomainCount = pointer.Int32(1)
+		if len(s.FailureDomains) > 1 {
+			vmss.VirtualMachineScaleSetProperties.PlatformFaultDomainCount = pointer.Int32(int32(len(s.FailureDomains)))
+		}
```
nawazkh, line 217: Can we point to the source of these rules in a comment?

azure/services/scalesets/spec.go
```diff
@@ -0,0 +1,522 @@
...
+					ID: pointer.String(azure.AddressPoolID(s.SubscriptionID, s.ResourceGroup, s.PublicLBName, s.PublicLBAddressPoolName)),
+				})
+		}
+	}
+	nicConfigs := []compute.VirtualMachineScaleSetNetworkConfiguration{}
+	for i, n := range s.NetworkInterfaces {
+		nicConfig := compute.VirtualMachineScaleSetNetworkConfiguration{}
+		nicConfig.VirtualMachineScaleSetNetworkConfigurationProperties = &compute.VirtualMachineScaleSetNetworkConfigurationProperties{}
+		nicConfig.Name = pointer.String(s.Name + "-nic-" + strconv.Itoa(i))
+		nicConfig.EnableIPForwarding = pointer.Bool(true)
```
nawazkh, line 313: Do we want to leave it to what user specifies instead?

azure/scope/machinepool.go
```diff
@@ -116,9 +126,59 @@ func NewMachinePoolScope(params MachinePoolScopeParams) (*MachinePoolScope, erro
...
+
+		skuCache, err := resourceskus.GetCache(m, m.Location())
+		if err != nil {
+			return err
+		}
+
+		m.cache.VMSKU, err = skuCache.Get(ctx, m.AzureMachinePool.Spec.Template.VMSize, resourceskus.VirtualMachines)
+		if err != nil {
+			return errors.Wrapf(err, "failed to get VM SKU %s in compute api", m.AzureMachinePool.Spec.Template.VMSize)
+		}
```
nawazkh, line 156: What do you think about adding some debug level logs in here?

nawazkh (commented):
Looks good to me pending tests and some minor nits.

azure/services/scalesets/spec.go
mweibel: see https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/3224 / https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3243/files for a similar discussion - IIRC at this point it's not configurable and needs to be set to true otherwise breaks networking.

azure/services/scalesets/spec.go
Jont828: I'm not sure I understand. Do you want to link to this block in another section of the PR?

azure/scope/machinepool.go
Jont828: I think we're fine, I just followed the convention in [machine.go](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/9f211f052d74764cee3922481390b300da8954cc/azure/scope/machine.go#L110).

azure/services/scalesets/scalesets.go
```diff
@@ -85,32 +83,68 @@ func (s *Service) Reconcile(ctx context.Context) (retErr error) {
...
+	scaleSetSpec.VMSSInstances = vmssInstances
+	scaleSetSpec.VMImage, err = s.Scope.GetVMImage(ctx)
+	if err != nil {
+		resultErr = errors.Wrapf(err, "failed to get vm image")
+		s.Scope.UpdatePutStatus(infrav1.BootstrapSucceededCondition, serviceName, resultErr)
+		return resultErr
+	}
+	s.Scope.SaveVMImageToStatus(scaleSetSpec.VMImage)
+
+	scaleSetSpec.BootstrapData, err = s.Scope.GetBootstrapData(ctx)
```
CecileRobertMichon, line 119: it seems counter-intuitive here that we're getting `scaleSetSpec` and then modifying the object returned by calling more scope functions. Looks like you set up a cache like we do for `machines` but then never used it. We could use the cache to set all the values that need to make calls to Azure or k8s APIs like we do for machines instead of calling the functions twice (once here and once in InitMachinePoolCache) https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/scope/machine.go#L172C10-L172C10

azure/services/scalesets/scalesets.go
```diff
@@ -85,32 +83,68 @@ func (s *Service) Reconcile(ctx context.Context) (retErr error) {
...
+	s.Scope.UpdatePutStatus(infrav1.BootstrapSucceededCondition, serviceName, err)
+
+	if err == nil && result != nil {
+		vmss, ok := result.(compute.VirtualMachineScaleSet)
+		if !ok {
+			return errors.Errorf("%T is not a compute.VirtualMachineScaleSet", result)
 		}
 
+		fetchedVMSS := converters.SDKToVMSS(vmss, vmssInstances)
+		// Note: converters.SDKToVMSS() will never return nil but we should check anyway
```
CecileRobertMichon, line 142: maybe that func shouldn't be returning a pointer if it never returns nil?

azure/scope/machinepool.go
```diff
@@ -116,9 +126,59 @@ func NewMachinePoolScope(params MachinePoolScopeParams) (*MachinePoolScope, erro
...
+func (m *MachinePoolScope) ScaleSetSpec(ctx context.Context) azure.ResourceSpecGetter {
+	ctx, log, done := tele.StartSpanWithLogger(ctx, "scope.MachinePoolScope.ScaleSetSpec")
+	defer done()
+
+	shouldPatchCustomData := false
+	if m.HasReplicasExternallyManaged(ctx) {
+		var err error
+		shouldPatchCustomData, err = m.HasBootstrapDataChanges(ctx)
+		if err != nil {
+			log.Error(err, "failed to check for bootstrap data changes")
```
CecileRobertMichon, line 172: we shouldn't be dropping an error without returning it

Looking at this more closely, it seems that we could be making use of the cache in https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/scope/machinepool.go#L587 instead of getting it again to save us from another API call.

azure/services/scalesets/spec.go
nawazkh: Sorry for the confusion. I was suggesting if we could point the docs for these validations via comment. Something like `For rules, refer: azure-docs.com`. <- Something along the lines so that its easier to review the rules next time we come across this switch case

azure/services/scalesets/scalesets.go
Jont828: Good catch, I set it up in the cache but somehow forgot to use it. As for the client calls, would the convention be to add it to `InitMachinePoolCache()` or try to keep it in the Reconciler code? For example, for InboundNatRules, we're getting the spec and amending it after we list the existing rules.

https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/e52424eeb31cb5671a39416cc9e10be53a39c724/azure/services/inboundnatrules/inboundnatrules.go#L81-L109

azure/services/scalesets/scalesets.go
Jont828: I think that make sense. Do you think we should do that in this PR or a separate follow up?

azure/scope/machinepool.go
Jont828: Is it a convention for the Spec getter functions like `ScaleSetSpec()` to never return an error? I noticed that was the case in all of the other services, and we were calling the functions that might return an error in the reconciler and then setting the field in the spec.

CecileRobertMichon: > or just to change the nested call like this

The latter

The main reason for not ever returning an error from a spec getter function and pushing that error down to the reconcile function is that the spec getter is also called by the `Delete()` functions and we shouldn't fail delete base on some k8s client call failing that is only used for create (in this case it's true for bootstrap data) since that might cause a broken cluster/machine/machinepool to be stuck forever and not be able to delete.

Jont828: Sounds good. I just refactored those changes into `InitMachinePoolCache`.

azure/services/scalesets/spec.go
Jont828: @nawazkh Ah gotcha. I'm not sure where to find these rules from. @CecileRobertMichon any thoughts?

CecileRobertMichon: AFAIK these have to do with differences between flex and uniform VMSS. @mboersma might know more.

CecileRobertMichon (commented):
/cc @nawazkh @willie-yao

azure/scope/machinepool.go
```diff
@@ -484,6 +564,7 @@ func (m *MachinePoolScope) SetFailureMessage(v error) {
 }
 
 // SetFailureReason sets the AzureMachinePool status failure reason.
+// TODO: should this be capierrors.MachinePoolStatusFailure instead?
```
willie-yao, line 567: Is this TODO going to be left for a later PR or revision?

willie-yao (commented):
Looks great to me! Just one minor comment from my end.

azure/scope/machinepool.go
Jont828: It's more of a question I had. I can remove the TODO but just wanted to get some clarification on if I'm understanding correctly.

willie-yao (approved):
/lgtm just another minor comment! Thanks for your work on this!

azure/scope/machinepool.go
CecileRobertMichon: It won't change much in practice, they are both a string type, but yes IMO `capierrors.MachinePoolStatusFailure` would be more appropriate

exp/controllers/azuremachinepool_controller.go
```diff
@@ -300,14 +300,15 @@ func (ampr *AzureMachinePoolReconciler) reconcileNormal(ctx context.Context, mac
 	err := machinePoolScope.InitMachinePoolCache(ctx)
 	if err != nil {
 		if errors.As(err, &reconcileError) && reconcileError.IsTerminal() {
-			ampr.Recorder.Eventf(machinePoolScope.AzureMachinePool, corev1.EventTypeWarning, "SKUNotFound", errors.Wrap(err, "failed to initialize machine cache").Error())
-			log.Error(err, "Failed to initialize machine cache")
+			ampr.Recorder.Eventf(machinePoolScope.AzureMachinePool, corev1.EventTypeWarning, "SKUNotFound", errors.Wrap(err, "failed to initialize machinepool cache").Error())
+			log.Error(err, "Failed to initialize machinepool cache")
+			// TODO: Should this be InvalidConfigurationMachinePoolError instead?
```
CecileRobertMichon, line 305: Yes

azure/services/scalesets/scalesets.go
```diff
@@ -85,86 +79,57 @@ func (s *Service) Reconcile(ctx context.Context) (retErr error) {
...
 
-	// Try to get the VMSS to update status if we have created a long running operation. If the VMSS is still in a long
-	// running operation, getVirtualMachineScaleSetIfDone will return an azure.WithTransientError and requeue.
-	if future != nil {
-		fetchedVMSS, err = s.getVirtualMachineScaleSetIfDone(ctx, future)
+		// Transform the VMSS resource representation to conform to the cloud-provider-azure representation
+		providerID, err := azprovider.ConvertResourceGroupNameToLower(azureutil.ProviderIDPrefix + fetchedVMSS.ID)
 		if err != nil {
-			return errors.Wrapf(err, "failed to get VMSS %s after create or update", scaleSetSpec.Name)
+			log.Error(err, "failed to parse VMSS ID", "ID", fetchedVMSS.ID)
```
CecileRobertMichon, line 126: this should be returning the error

azure/services/scalesets/scalesets.go
```diff
@@ -173,13 +138,14 @@ func (s *Service) Delete(ctx context.Context) error {
 	ctx, log, done := tele.StartSpanWithLogger(ctx, "scalesets.Service.Delete")
 	defer done()
 
-	var err error
+	// TODO: do we want this timeout?
```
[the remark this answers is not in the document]
CecileRobertMichon: All the services should have a timeout, where are the services with missing timeouts? We should fix those to be consistent

CecileRobertMichon (commented):
/cc @mboersma

azure/services/scalesets/scalesets.go
```diff
@@ -85,86 +79,57 @@ func (s *Service) Reconcile(ctx context.Context) (retErr error) {
 	ctx, log, done := tele.StartSpanWithLogger(ctx, "scalesets.Service.Reconcile")
 	defer done()
 
+	ctx, cancel := context.WithTimeout(ctx, reconciler.DefaultAzureServiceReconcileTimeout)
+	defer cancel()
+
 	if err := s.validateSpec(ctx); err != nil {
```
CecileRobertMichon, line 85: would it make sense to move this into `Parameters` in spec.go? that's what we do for VMs already, and that way we wouldn't have to call `s.Scope.ScaleSetSpec(ctx)` twice

azure/services/scalesets/spec.go
CecileRobertMichon: this is still valid, should be

```suggestion
// Parameters returns the parameters for the Scale Set.
```

azure/services/scalesets/spec.go
```diff
@@ -0,0 +1,522 @@
...
+	HasReplicasExternallyManaged bool
+	AdditionalTags               infrav1.Tags
+}
+
+// ResourceName returns the name of the Scale Set.
+func (s *ScaleSetSpec) ResourceName() string {
+	return s.Name
+}
+
+// ResourceGroupName returns the name of the resource group of the VNet that owns this Scale Set.
```
CecileRobertMichon, line 82:
```suggestion
// ResourceGroupName returns the name of the resource group for this Scale Set.
```

exp/controllers/azuremachinepool_controller.go
Jont828: I think we would need to change the type of `AzureMachinePool.Status.FailureReason` in order to fix this. I can remove the TODO and open a follow up PR if that works for you.

azure/services/scalesets/scalesets.go
Jont828: I think the intention here was to do the validation before we call `client.Get()` and `client.ListInstances()`. The main thing for me is that `Parameters()` is already gigantic and pretty difficult to maintain/test, especially the different control flow possibilities. WDYT?

exp/controllers/azuremachinepool_controller.go
CecileRobertMichon: SGTM

Jont828:
@mboersma @willie-yao Could I get a review on this now that it's ready?

azure/services/scalesets/scalesets.go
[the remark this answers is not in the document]
CecileRobertMichon: @Jont828 can you please remove this and any other leftover TODOs? any unresolved TODOs should be tracked in issues

azure/scope/machinepool.go
mboersma: Should we change this to use `capierrors.MachinePoolStatusFailure` (or open an issue to refactor it afterward)?

azure/services/scalesets/scalesets.go
```diff
@@ -82,89 +76,60 @@ func (s *Service) Name() string {
...
 
-	// Try to get the VMSS to update status if we have created a long running operation. If the VMSS is still in a long
-	// running operation, getVirtualMachineScaleSetIfDone will return an azure.WithTransientError and requeue.
-	if future != nil {
-		fetchedVMSS, err = s.getVirtualMachineScaleSetIfDone(ctx, future)
+		// Transform the VMSS resource representation to conform to the cloud-provider-azure representation
+		providerID, err := azprovider.ConvertResourceGroupNameToLower(azureutil.ProviderIDPrefix + fetchedVMSS.ID)
 		if err != nil {
-			return errors.Wrapf(err, "failed to get VMSS %s after create or update", scaleSetSpec.Name)
+			return errors.Wrapf(err, "failed to parse VMSS ID %s", fetchedVMSS.ID)
```
mboersma, line 126: Thanks for noticing and fixing this–there's nothing worse than a misleading error message.

mboersma (commented):
This looks good to me. I didn't notice anything others haven't, but there are still a few outstanding `// TODO`s and comment changes to incorporate.

Jont828:
@mboersma I've opened #3825 and removed the TODO.

azure/services/scalesets/spec.go
mboersma: I honestly don't know where these values originate. I inherited them from @devigned's original work to support VMSS Flex and learned through experimentation that they were indeed necessary.
If we do want to try to link to a doc, I think https://learn.microsoft.com/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-orchestration-modes is good and may explain some of these parameters.

mboersma:
@Jont828 looks good! There are still a few outstanding comments AFAICT, hidden in the collapsed middle section of this epic PR:
https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3111#discussion_r1290681602
https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3111#discussion_r1290685846
https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3111#discussion_r1290364865

azure/services/scalesets/spec.go
Jont828: Cool I'll drop the link in a comment.

azure/services/scalesets/scalesets.go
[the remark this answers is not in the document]
Jont828: Sure, I thought I got them all but let me make a second pass through to double check.

azure/services/roleassignments/roleassignments.go
```diff
@@ -141,10 +141,20 @@ func (s *Service) getVMSSPrincipalID(ctx context.Context) (*string, error) {
...
+		ResourceGroup: s.Scope.ResourceGroup(),
+	}
+
+	resultVMSSIface, err := s.virtualMachineScaleSetGetter.Get(ctx, spec)
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to get principal ID for VMSS")
 	}
+	resultVMSS, ok := resultVMSSIface.(compute.VirtualMachineScaleSet)
+	if !ok {
+		return nil, errors.Errorf("%T is not a compute.VirtualMachine", resultVMSSIface)
```
nawazkh, line 155:
```suggestion
		return nil, errors.Errorf("%T is not a compute.VirtualMachineScaleSet", resultVMSSIface)
```

azure/services/roleassignments/roleassignments.go
```diff
@@ -141,10 +141,20 @@ func (s *Service) getVMSSPrincipalID(ctx context.Context) (*string, error) {
...
-	resultVMSS, err := s.virtualMachineScaleSetClient.Get(ctx, s.Scope.ResourceGroup(), s.Scope.Name())
+	spec := &scalesets.ScaleSetSpec{
+		Name:          s.Scope.Name(),
+		ResourceGroup: s.Scope.ResourceGroup(),
+	}
+
+	resultVMSSIface, err := s.virtualMachineScaleSetGetter.Get(ctx, spec)
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to get principal ID for VMSS")
 	}
```
nawazkh, line 152: Should this be updated to below?
```suggestion
	if err != nil {
		return nil, errors.Wrap(err, "failed to get resultVMSSIface")
	}
```

azure/services/scalesets/scalesets.go
```diff
@@ -82,89 +76,60 @@ func (s *Service) Name() string {
...
-				log.Error(err, "failed to parse VMSS ID", "ID", fetchedVMSS.ID)
-			}
-			s.Scope.SetProviderID(providerID)
-			s.Scope.SetVMSSState(fetchedVMSS)
+	_, err := s.Client.Get(ctx, spec)
+	if err == nil {
+		// We can only get the existing instances if the VMSS already exists
+		scaleSetSpec.VMSSInstances, err = s.Client.ListInstances(ctx, spec.ResourceGroupName(), spec.ResourceName())
+		if err != nil {
+			err = errors.Wrapf(err, "failed to get existing instances")
```
nawazkh, line 101:
```suggestion
			err = errors.Wrapf(err, "failed to get existing VMSS instances")
```

azure/services/scalesets/scalesets.go
```diff
@@ -82,89 +76,60 @@ func (s *Service) Name() string {
...
-		// HTTP(200)
-		// VMSS already exists and may have changes; update it with a PATCH
-		// we do this to avoid overwriting fields in networkProfile modified by cloud-provider
-		future, err = s.patchVMSSIfNeeded(ctx, fetchedVMSS)
-		if err != nil {
-			return errors.Wrap(err, "failed to start updating VMSS")
+
+		fetchedVMSS := converters.SDKToVMSS(vmss, scaleSetSpec.VMSSInstances)
+		if err := s.Scope.ReconcileReplicas(ctx, &fetchedVMSS); err != nil {
+			return errors.Wrap(err, "unable to reconcile replicas")
```
nawazkh, line 120:
```suggestion
			return errors.Wrap(err, "unable to reconcile VMSS replicas")
```

nawazkh (commented):
Great work @Jont828! Thanks for putting this together and working this out!
Also, I am sorry for the last-minute reviews. I had started the review earlier and then get diverted and had to shift my focus. 
I had some minor nits and have jotted them below.

nawazkh (changes requested):

nawazkh:
Just one nit 😬 
Everything else looks good to me pending tests 🌚

azure/services/roleassignments/roleassignments.go
Jont828: On second thought, I think we want to leave it as is. Before, we were giving the error when we failed to get the VM, which we're still doing here.

nawazkh (approved):
Great work @Jont828 🚀 
Thank you for going over the comments so patiently. 
/lgtm

azure/services/scalesets/spec.go
CecileRobertMichon: this is resolved

Jont828:
🚀 🚀 🚀

nawazkh:
🚀

devigned:
Awesome work!! 🚀

willie-yao:
🚀 💯

PR #3126, opened 2023-01-30
E2E: explicit kubeconfig in shell scripts

jackfrancis:
**What type of PR is this?**

/kind cleanup

While doing some local testing using `ci-entrypoint.sh`, I observed some environmental edge cases around exporting the common `KUBECONFIG` variable in various shell scripts and funcs. This PR removes that environment juggling in favor of explicit kubeconfig references, which makes the code easier to reason about (though arguably at a slightly higher cost due to increased args), especially as it's critical to understand "is this kubectl statement supposed to target the mgmt cluster or the workload cluster?".

Supporting the above, I've changed the way we persist a generic `./kubeconfig` file to instead persist a cluster-specific kubeconfig file, as the entire flow already depends upon a `CLUSTER_NAME` variable which contains the suffient uniquely identifying information that we use to distinguish one cluster from another.

Also while I was here I standardized the way we pass `-o` (output configuration) to `kubectl`, as I noticed there were three different flavors.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```


Makefile
```diff
@@ -315,10 +315,10 @@ create-workload-cluster: $(ENVSUBST) $(KUBECTL) ## Create a workload cluster.
 	# Wait for the kubeconfig to become available.
 	timeout --foreground 300 bash -c "while ! $(KUBECTL) get secrets | grep $(CLUSTER_NAME)-kubeconfig; do sleep 1; done"
 	# Get kubeconfig and store it locally.
-	$(KUBECTL) get secrets $(CLUSTER_NAME)-kubeconfig -o json | jq -r .data.value | base64 --decode > ./kubeconfig
-	timeout --foreground 600 bash -c "while ! $(KUBECTL) --kubeconfig=./kubeconfig get nodes | grep control-plane; do sleep 1; done"
+	$(KUBECTL) get secrets $(CLUSTER_NAME)-kubeconfig -o json | jq -r .data.value | base64 --decode > ./$(CLUSTER_NAME)-kubeconfig
+	timeout --foreground 600 bash -c "while ! $(KUBECTL) --kubeconfig=./$(CLUSTER_NAME)-kubeconfig get nodes | grep control-plane; do sleep 1; done"
 
-	@echo 'run "$(KUBECTL) --kubeconfig=./kubeconfig ..." to work with the new target cluster'
+	@echo 'run "$(KUBECTL) --kubeconfig=./$(CLUSTER_NAME)-kubeconfig ..." to work with the new target cluster'
```
CecileRobertMichon, line 321: should we do `$(CLUSTER_NAME).kubeconfig` to be consistent with what we already do in tilt? It would also work with our existing `clean-temporary` Makefile targets which deletes all kubeconfig files from tilt and with the existing .gitignore

https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/Tiltfile#L357

jackfrancis: if that's more consistent then definitely yes

jackfrancis: done

.gitignore
```diff
@@ -27,7 +27,6 @@ envfile
 # kubeconfigs
 kind.kubeconfig
 minikube.kubeconfig
-kubeconfig
```
CecileRobertMichon, line 30: why is this removed?

jackfrancis: After doing a further audit, I'm confident that this PR updates the entire set of tooling that we provide so that the vanilla `./kubeconfig` will no longer be persisted.

But, because we've been using that generic name for a long time, existing developers will likely still have those files around and so we should keep this entry in place to prevent accidentally including it in their next PR. :(

Also I checked and `make clean` (via `make clean-temporary` still includes a step to explicitly remove `./kubeconfig`, so eventually these will go away from folks' local cloned repos. But it doesn't hurt to keep this in `.gitignore` indefinitely.

scripts/ci-entrypoint.sh
```diff
@@ -133,26 +133,24 @@ create_cluster() {
...
-
-    # export the target cluster KUBECONFIG if not already set
-    export KUBECONFIG="${KUBECONFIG:-${PWD}/kubeconfig}"
+    CIDR0=$(${KUBECTL} get cluster "${CLUSTER_NAME}" -o jsonpath='{.spec.clusterNetwork.pods.cidrBlocks[0]}')
+    CIDR1=$(${KUBECTL} get cluster "${CLUSTER_NAME}" -o jsonpath='{.spec.clusterNetwork.pods.cidrBlocks[1]}' 2> /dev/null) || true
 
     # wait for the apiserver pod to be Ready.
-    APISERVER_POD=$("${KUBECTL}" get pods -n kube-system -o name | grep apiserver)
-    "${KUBECTL}" wait --for=condition=Ready -n kube-system "${APISERVER_POD}" --timeout=5m
+    timeout --foreground 600 bash -c "while ! ${KUBECTL} --kubeconfig=./${CLUSTER_NAME}.kubeconfig get pods -n kube-system -o name | grep apiserver; do sleep 5; done"
```
jackfrancis, line 140: In my testing I encountered further flakes that originate from grabbing the apiserver pod (L141 below) blindly, without first confirming that it exists. I think I'd like to address that in a separate PR.

My thoughts are: rather than doing this check for a particular grep'd apiserver pod (which I would expect to have unreliable results in a HA control plane scenario where the grep returns more than one line entry), why don't we just  put the below kubectl commands in appropriate retry blocks? That would seem more resilient than trying to identify every possible "apiserver pod isn't stable" symptom, and additionally a validation at one point in time that "apiserver pod is stable" isn't durable, and moments later a kubectl command against it may transiently fail (such as during a > 1 control plane node etcd leader election scenario during cluster bootstrapping).

scripts/ci-entrypoint.sh
```diff
@@ -213,32 +210,27 @@ install_addons() {
...
     # Ensure that all nodes are registered with the API server before checking for readiness
     local total_nodes="$((CONTROL_PLANE_MACHINE_COUNT + WORKER_MACHINE_COUNT + WINDOWS_WORKER_MACHINE_COUNT))"
-    while [[ $("${KUBECTL}" get nodes -ojson | jq '.items | length') -ne "${total_nodes}" ]]; do
+    while [[ $("${KUBECTL}" --kubeconfig "${PWD}/${CLUSTER_NAME}".kubeconfig get nodes -o json | jq '.items | length') -ne "${total_nodes}" ]]; do
         sleep 10
     done
 
-    "${KUBECTL}" wait --for=condition=Ready node --all --timeout=5m
-    "${KUBECTL}" get nodes -owide
+    "${KUBECTL}" --kubeconfig "$PWD"/"$CLUSTER_NAME".kubeconfig wait --for=condition=Ready node --all --timeout=5m
```
CecileRobertMichon, line 228:
```suggestion
    "${KUBECTL}" --kubeconfig "${PWD}/${CLUSTER_NAME}.kubeconfig" wait --for=condition=Ready node --all --timeout=5m
```

would this work?

jackfrancis: worth a try, updated

jackfrancis: looks good

scripts/ci-entrypoint.sh
jackfrancis: I've removed this change from this PR and scoped these changes just to the persisted workload cluster kubeconfig.

CecileRobertMichon (commented):
/lgtm

scripts/ci-entrypoint.sh
```diff
@@ -213,33 +210,32 @@ install_addons() {
...
-    "${KUBECTL}" wait --for=condition=Ready pod -n calico-system --all --timeout=10m
+    "${KUBECTL}" --kubeconfig "${PWD}/${CLUSTER_NAME}.kubeconfig" wait --for=condition=Ready pod -n calico-system --all --timeout=10m
 
     echo "Waiting for all kube-system pods to be ready"
-    "${KUBECTL}" wait --for=condition=Ready pod -n kube-system --all --timeout=10m
+    "${KUBECTL}" --kubeconfig "${PWD}/${CLUSTER_NAME}.kubeconfig" wait --for=condition=Ready pod -n kube-system --all --timeout=10m
+
+    # For convenience we export the KUBECONFIG environment variable
+    # so that future steps in a serialized workflow can target the workload cluster automatically
+    export KUBECONFIG="${PWD}/${CLUSTER_NAME}.kubeconfig"
```
jackfrancis, line 220: @CecileRobertMichon as I audited that many variations of k8s + Azure tests in the ecosystem that depend on this script I observed that we have essentially inherited a contract whereby we must export `KUBECONFIG` in this way to allow serialized CI jobs to target their work against the CAPZ-built Azure cluster.

Thanks @marosset for the extra eyes :)

CecileRobertMichon: I wonder if we should do this outside the install_addons func then, as a final step before returning control to the external test suite (i.e. on line 280 https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3126/files#diff-2009f7be5363812e22862771f557363795759e67b3ca2e32c50d9fd5f8c5669bR274)

willie-yao (approved):
/lgtm

jackfrancis:
/hold

needs more investigation to ensure that the `KUBECONFIG` env var value changes as we return are back-compat with existing job uses.

CecileRobertMichon:
```
  Normal   Pulling      20m (x4 over 21m)    kubelet            Pulling image "sigwindowstools/calico-install:v3.25.0-hostprocess"
  Warning  Failed       20m (x4 over 21m)    kubelet            Error: ErrImagePull
  Normal   BackOff      2m9s (x83 over 21m)  kubelet            Back-off pulling image "sigwindowstools/calico-install:v3.25.0-hostprocess"
  ```

retrying in case it was a timing fluke....

/retest

**Comment by jackfrancis** [comment]
After spending lots of hours testing and attempting to validate the many use-cases of ci-entrypoint.sh across the ecosystem, I am confident that the most important near-term effort is to reduce transient errors returning a functional, built cluster to the various CI jobs that rely upon this script. As such I think we should move forward with another ci-entrypoint.sh PR first:

https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3148

I still don't like the way we're handling kubeconfig and how we switch back and forth between the management and workload cluster at HEAD, so I'll keep this PR open, but until I'm able to definitely prove that these kubeconfig changes won't break various CI jobs that depend upon ci-entrypoint, let's keep this one in the queue and I'll comment when said proof has been collected.

File: scripts/ci-entrypoint.sh [hunk 6]
```diff
@@ -170,8 +167,8 @@ install_addons() {
         CIDR_STRING_VALUES="installation.calicoNetwork.ipPools[0].cidr=${CIDR0}"
     fi
 
-    "${HELM}" repo add projectcalico https://projectcalico.docs.tigera.io/charts
-    "${HELM}" install calico projectcalico/tigera-operator -f "${CALICO_VALUES_FILE}" --set-string "${CIDR_STRING_VALUES}" --namespace tigera-operator --create-namespace
+    "${HELM}" --kubeconfig "${PWD}/${CLUSTER_NAME}.kubeconfig" repo add projectcalico https://projectcalico.docs.tigera.io/charts
```
mboersma, line 170: These `helm` commands lead to these warnings:

```
Installing Calico CNI via helm
Cluster CIDR is IPv4
WARNING: Kubernetes configuration file is group-readable. This is insecure. Location: /home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/kubeconfig
WARNING: Kubernetes configuration file is world-readable. This is insecure. Location: /home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/kubeconfig
"projectcalico" has been added to your repositories
WARNING: Kubernetes configuration file is group-readable. This is insecure. Location: /home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/kubeconfig
WARNING: Kubernetes configuration file is world-readable. This is insecure. Location: /home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/kubeconfig
NAME: calico
```

Maybe we can fix that with a `chmod` somewhere?

nawazkh:
Just a thought, It is obvious to me that `${CLUSTER_NAME}` refers to the workload-cluster created using CAPZ, but would it be worth updating `${CLUSTER_NAME}` to something like `${WORKLOAD_CLUSTER_NAME}` to make it obvious to the reader that the command refers to the workload cluster ?

jackfrancis:
Looks like ci-entrypoint has stabilized a bit, so let's not rock the boat at this time by making any more changes!

PR #3128, opened 2023-01-30
log diffs for agentpools

nojnhuh:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**: This PR adds logging for diffs between the actual and expected AKS agent pool backing an AzureManagedMachinePool to determine why the controller is issuing a PUT against AKS during reconciliation. This will help debug cases where the controller is erroneously endlessly updating an agent pool like in #3009.

Related to #2829 which added this for managed cluster resources.

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/azure/services/agentpools/spec.go
+++ b/azure/services/agentpools/spec.go
@@ -29,6 +29,7 @@ import (
 	"sigs.k8s.io/cluster-api-provider-azure/azure"
 	"sigs.k8s.io/cluster-api-provider-azure/azure/converters"
 	azureutil "sigs.k8s.io/cluster-api-provider-azure/util/azure"
+	"sigs.k8s.io/cluster-api-provider-azure/util/tele"
 )
 
 // KubeletConfig defines the set of kubelet configurations for nodes in pools.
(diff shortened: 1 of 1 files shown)


azure/services/agentpools/spec.go
```diff
@@ -160,6 +161,9 @@ func (s *AgentPoolSpec) CustomHeaders() map[string]string {
 
 // Parameters returns the parameters for the agent pool.
 func (s *AgentPoolSpec) Parameters(ctx context.Context, existing interface{}) (params interface{}, err error) {
+	_, log, done := tele.StartSpanWithLogger(ctx, "agentpools.Service.Parameters")
```
nojnhuh, line 164: Before when this was `ctx, log, done := ...` this was throwing an "ineffectual assignment to ctx" linter error. If this function ever needs to pass on the `ctx` it receives, then we need to remember to set `ctx` here again so we pass on the wrapped one, not the one that gets passed in as-is. Is there a good way to remind ourselves to do that? A comment here could be easily missed if `ctx` gets used deep inside this function.

jackfrancis: indeed https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3130

PR #3129, opened 2023-01-30
Fix scale in for VMSS Flex MachinePools

### LINKED ISSUE #3077: CAPZ fails to scale in K8s VMSS Flex cluster
/kind bug

[Before submitting an issue, have you checked the [Troubleshooting Guide](https://capz.sigs.k8s.io/topics/troubleshooting.html)?] Yes

Step 1: Create a K8s cluster using the template [cluster-template-external-cloud-provider-machinepool.yaml](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/templates/cluster-template-external-cloud-provider-machinepool.yaml) to create a K8s cluster with VMSS Flex as MachinePool
Step 2: Scale in the MachinePool with command: kubectl scale --replicas=<a smaller replica count>  MachinePool/<MachinePoolName>
Step 3: After the above command is executed, all the nodes in the MachinePool will become "Ready,SchedulingDisabled" status but the nodes will never be deleted:
image

Some nodes should be deleted after scaling in.

image
image
image

### Root Cause Analysis:
I briefly went through the CAPZ scaling in logic for MachinePoolMachine, from my understanding, before kicking away the VM from the MachinePool, CAPZ does 2 things:
Thing 1:  run “[kubectl cordon](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/scope/machinepoolmachine.go#L458)” command to prevent new pods to be scheduled into this node 
Thing 2:  run “[kubectl drain](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/scope/machinepoolmachine.go#L463)” command to kill all existing pods in this node.

When I run the “kubectl describe node” command, I can see the taint “node.kubernetes.io/unschedulable:NoSchedule” is added to the node to prevent new pods to be scheduled to it:
 
image

This existence of this taints indicates the “cordon” command succeeds and the issue should be the “drain” command. In the above error message I captured, it looks like the CAPZ is complaining the corresponding node cannot be found. I took a look at the node name in the error message and found CAPZ is trying to use a self-defined node name to identify the node: <MachinePool name>-<VM name> (replace underscore to dash), and I guess this is where the issue is from: **The K8s cluster uses OS hostname (computer name) as the nodeName, but CAPZ defines a different name format so the cluster cannot find the corresponding node.** This partially explains the issue but one thing I don’t figure out yet is, why the issue only exists in the “drain” command while “cordon” command does not meet this issue. Maybe I missed something.

**Environment:**

- Kubernetes version: (use `kubectl version`): 1.26.0
- OS (e.g. from `/etc/os-release`): Linux

mboersma:
**What type of PR is this?**

/kind bug

Scaling in a MachinePool using `Flexible` orchestration mode fails, and a change in API response broke the previous mechanism for detecting Flex clusters.

Fixes #3077

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
 
```release-note
NONE
```

### DIFF
--- a/test/e2e/azure_machinepools.go
+++ b/test/e2e/azure_machinepools.go
@@ -117,7 +117,6 @@ func AzureMachinePoolsSpec(ctx context.Context, inputGetter func() AzureMachineP
 	}
 	wg.Wait()
 
-	/* TODO: uncomment with scale down fix for flexible mode
 	for _, mp := range machinepools {
 		goalReplicas := pointer.Int32Deref(mp.Spec.Replicas, 0) - 1
 		Byf("Scaling machine pool %s in from %d to %d", mp.Name, *mp.Spec.Replicas, goalReplicas)
@@ -135,7 +134,6 @@ func AzureMachinePoolsSpec(ctx context.Context, inputGetter func() AzureMachineP
 		}(mp)
 	}
 	wg.Wait()
-	*/
 
 	By("verifying that workload nodes are schedulable")
 	clientset := workloadClusterProxy.GetClientSet()
--- a/test/e2e/azure_test.go
+++ b/test/e2e/azure_test.go
@@ -535,7 +535,7 @@ var _ = Describe("Workload cluster creation", func() {
 				withControlPlaneInterval(specName, "wait-control-plane"),
 			), result)
 
-			By("Verifying machinepool can scale out", func() {
+			By("Verifying machinepool can scale out and in", func() {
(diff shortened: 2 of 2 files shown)


mboersma:
/retitle Fix orchestrator detection and scale in for VMSS Flex MachinePools

mboersma:
/retitle Fix scale in for VMSS Flex MachinePools

mboersma:
/close

See #3249

PR #3132, opened 2023-01-31
Add subnetName support to ammp

### LINKED ISSUE #1599: Virtual network in managed clusters should be moved from AMCP to AMMP
/kind bug

Current AzureManagedControlPlaneSpec.VirtualNetwork field is a "global" cluster field, it gets applied to all node pools.

In Azure, this part of the network spec belongs to [ManagedClusterAgentPoolProfile](https://docs.microsoft.com/en-us/rest/api/aks/managed-clusters/create-or-update#managedclusteragentpoolprofile), so it is configurable per node pool.

Virtual network is not even part of cluster-wide [ContainerServiceNetworkProfile](https://docs.microsoft.com/en-us/rest/api/aks/managed-clusters/create-or-update#containerservicenetworkprofile).

We must align CAPZ managed cluster network spec and adding vnetSubnetID and podSubnetID (from Azure's ManagedClusterAgentPoolProfile) to AzureManagedMachinePoolSpec as optional fields

See https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1590#issuecomment-896016819 

**Environment:**

LochanRn:
**What type of PR is this?**
/kind feature

Support to configure a different subnet for AzureManagedMachinePool. 

Fixes #1599

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Add subnetName to AzureManagedMachinePool.
```


api/v1beta1/azuremanagedmachinepool_types.go
```diff
@@ -203,6 +203,9 @@ type AzureManagedMachinePoolSpec struct {
 	// +kubebuilder:validation:Enum=OS;Temporary
 	// +optional
 	KubeletDiskType *KubeletDiskType `json:"kubeletDiskType,omitempty"`
+	// SubnetName selects the Subnet where the MachinePool will be placed
```
willie-yao, line 206:
```suggestion
	// SubnetName specifies the Subnet where the MachinePool will be placed
```

willie-yao (commented):
Overall this looks great! Just a few minor nits from me.

nawazkh:
I am wondering if a similar change as that of https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2411 is required for AzureManagedMachinePool ? 
cc: @CecileRobertMichon

CecileRobertMichon:
> I am wondering if a similar change as that of https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2411 is required for AzureManagedMachinePool ?

Good question. I don't believe AKS allows us to pass in custom network interfaces (does it?)

azure/scope/managedmachinepool_test.go
```diff
@@ -623,6 +623,145 @@ func TestManagedMachinePoolScope_OSDiskType(t *testing.T) {
...
+					InfraMachinePool: getAzureMachinePool("pool1", infrav1.NodePoolModeUser),
+				},
+			},
+			Expected: &agentpools.AgentPoolSpec{
+				Name:         "pool1",
+				SKU:          "Standard_D2s_v3",
+				Mode:         "User",
+				Cluster:      "cluster1",
+				Replicas:     1,
+				VnetSubnetID: "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups//providers/Microsoft.Network/virtualNetworks/my-vnet/subnets/my-vnet-subnet",
```
willie-yao, line 704:
```suggestion
				VnetSubnetID: "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/resource-group-name/providers/Microsoft.Network/virtualNetworks/my-vnet/subnets/my-vnet-subnet",
```

azure/scope/managedmachinepool_test.go
```diff
@@ -623,6 +623,145 @@ func TestManagedMachinePoolScope_OSDiskType(t *testing.T) {
...
+					InfraMachinePool: getAzureMachinePool("pool0", infrav1.NodePoolModeSystem),
+				},
+			},
+			Expected: &agentpools.AgentPoolSpec{
+				Name:         "pool0",
+				SKU:          "Standard_D2s_v3",
+				Replicas:     1,
+				Mode:         "System",
+				Cluster:      "cluster1",
+				VnetSubnetID: "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups//providers/Microsoft.Network/virtualNetworks//subnets/",
```
willie-yao, line 665:
```suggestion
				VnetSubnetID: "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/resource-group-name/providers/Microsoft.Network/virtualNetworks//subnets/",
```

LochanRn:
@CecileRobertMichon If the subnet does not exist should capz create it ??

We do that in the case of vnet and subnet configured in amcp.

azure/scope/managedmachinepool_test.go
LochanRn: The resourceGroup within the VirtualNetworkSpec can be empty and when empty it is replaced to the resourceGroup provided in amcp ((i.e. cluster resource group) by the defaulting webhook.

willie-yao (approved):
/lgtm

azure/scope/managedmachinepool.go
```diff
@@ -154,6 +154,11 @@ func buildAgentPoolSpec(managedControlPlane *infrav1.AzureManagedControlPlane,
 		replicas = *machinePool.Spec.Replicas
 	}
 
+	subnetName := managedControlPlane.Spec.VirtualNetwork.Subnet.Name
```
CecileRobertMichon, line 157: are we planning on deprecating this field at all? or is this a fallback/override situation where `managedControlPlane.Spec.VirtualNetwork.Subnet.Name` defined the default for all node pools that can be optionally overridden by each `managedMachinePool.Spec.SubnetName`? Assuming it's the latter but just want to double check

azure/scope/managedmachinepool.go
```diff
@@ -154,6 +154,11 @@ func buildAgentPoolSpec(managedControlPlane *infrav1.AzureManagedControlPlane,
...
+	if managedMachinePool.Spec.SubnetName != "" {
+		subnetName = managedMachinePool.Spec.SubnetName
```
CecileRobertMichon, line 159: have you considered defaulting the subnetName on the managedMachinePool object too so it's persisted in the API and visible by the user?

api/v1beta1/azuremanagedmachinepool_types.go
```diff
@@ -203,6 +203,9 @@ type AzureManagedMachinePoolSpec struct {
 	// +kubebuilder:validation:Enum=OS;Temporary
 	// +optional
 	KubeletDiskType *KubeletDiskType `json:"kubeletDiskType,omitempty"`
+	// SubnetName specifies the Subnet where the MachinePool will be placed
+	// +optional
+	SubnetName string `json:"subnetName,omitempty"`
```
CecileRobertMichon, line 208: I don't see any additions to validation webhooks, anything that needs to be validated for this new value?

CecileRobertMichon:
> If the subnet does not exist should capz create it ??
>
> We do that in the case of vnet and subnet configured in amcp.

That's a tricky one. In the case of AzureMachinePool (unmanaged) we don't create/reconcile subnets as part of the MachinePool controller. 

Here is the original PR that did this https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1411

azure/scope/managedmachinepool.go
LochanRn: yes you are right the managedMachinePool.Spec.SubnetName should be optionally overridden if the user wants to place a specific agentpool in a different subnet. If managedMachinePool.Spec.SubnetName is undefined capz will use managedControlPlane.Spec.VirtualNetwork.Subnet.Name to place the agent-pool.

azure/scope/managedmachinepool.go
LochanRn: ack

api/v1beta1/azuremanagedmachinepool_types.go
LochanRn: added validations :)

azure/scope/managedmachinepool.go
LochanRn: @CecileRobertMichon Thoughts about the above and let me know if i have to make the above changes.

CecileRobertMichon: +1 to defaulting

willie-yao: @LochanRn have you made the defaulting changes? If so, I'll go ahead and review the PR again!

LochanRn: @CecileRobertMichon I tried to implement the above, I was facing intermittent issues while testing. The issue is because after the above implementation AzureManagedMachinePool has a couple of dependencies.

- before AzureManagedMachinePool is created AzureManagedControlPlane has to be created.
- capi-controller should set label _cluster.x-k8s.io/cluster-name_ on AzureManagedMachinePool and AzureManagedControlPlane.

This is because we need to get the AzureManagedControlPlane object to get AzureManagedControlPlane.Spec.VirtualNetwork.SubnetName. If AzureManagedControlPlane is not present AzureManagedMachinePool will be created but AzureManagedMachinePool.Spec.SubnetName value will not get set until the next update.

If the label is not present we cannot get AzureManagedControlPlane.

LochanRn: The main issue is with the timing. When we create a cluster we apply all the manifests at once so at times the amcp used to get created first and even capi controller was also fast enough to set the clusterlabel before the machinepool was even created. Hence the ammp defaulting webhook was able to kick in at the right time and update the subnet value but for some cases it was not the same. ammp used to get created and it used to be stuck until the next update.  As the defaulting webhook would kick in only during creation or update.

LochanRn: The implementation with which i tested was

```// Default implements webhook.Defaulter so a webhook will be registered for the type.
func (m *AzureManagedMachinePool) Default(cli client.Client) {
	if m.Labels == nil {
		m.Labels = make(map[string]string)
	}
	m.Labels[LabelAgentPoolMode] = m.Spec.Mode

	if m.Spec.Name == nil || *m.Spec.Name == "" {
		m.Spec.Name = &m.Name
	}

	if m.Spec.OSType == nil {
		m.Spec.OSType = to.StringPtr(DefaultOSType)
	}
	m.defaultSubnetName(cli)
}

func (m *AzureManagedMachinePool) defaultSubnetName(cli client.Client) {

	clusterName, ok := m.Labels[clusterv1.ClusterLabelName]
	if !ok {
		log.Errorf("capi has not yet set label %s",clusterv1.ClusterLabelName)
		return nil
	}

	// Fetch the amcp.
	amcp := &AzureManagedControlPlane{}
	key := client.ObjectKey{
		Namespace: m.Namespace,
		Name:      clusterName,
	}

	ctx := context.Background()

	if err := cli.Get(ctx, key, amcp); err != nil {
		log.Error(err)
		return nil
	}

	if amcp.Spec.VirtualNetwork.Name != "" {
		if m.Spec.SubnetName == "" {
			if amcp.Spec.VirtualNetwork.Subnet.Name == "" {
				log.Error("a virtualnetwork is defined but a subnet was undefined")
				return nil
			}
			m.Spec.SubnetName = amcp.Spec.VirtualNetwork.Subnet.Name
		}
	}
	return nil
}```

Comment (LochanRn) [reply azure/scope/managedmachinepool.go:159]: okay !!

File: api/v1beta1/azuremanagedmachinepool_types.go (hunk 6 above)
Comment (nawazkh) [inline api/v1beta1/azuremanagedmachinepool_types.go:208]: I am inclined on changing SubnetName from type `string` to `*string`. It will be easier to validate this field and perform a nil check.

File: api/v1beta1/azuremanagedmachinepool_webhook.go [hunk 7]
```diff
@@ -343,6 +351,17 @@ func (m *AzureManagedMachinePool) validateEnableNodePublicIP() error {
 	return nil
 }
 
+func (m *AzureManagedMachinePool) validateSubnetName() error {
+	if m.Spec.SubnetName != "" {
+		subnetRegex := `^[-\w\._]+$`
```
nawazkh, line 356: I maybe wrong, but this regex seems to be validating `-_-_-_-_`. Do you think we could make this more strict on accepting well-formed strings? Or is `-_-_-_-_` an acceptable subnet name? 
What does Azure Docs say about the subnet name?

api/v1beta1/azuremanagedmachinepool_webhook_test.go
```diff
@@ -514,6 +514,34 @@ func TestAzureManagedMachinePoolUpdatingWebhook(t *testing.T) {
...
+					SubnetName: "my-subnet",
+				},
+			},
+			old: &AzureManagedMachinePool{
+				Spec: AzureManagedMachinePoolSpec{
+					SubnetName: "my-subnet",
+				},
+			},
+			wantErr: false,
+		},
```
nawazkh, line 544: Would love to see a couple more tests on subnet names variations!

nawazkh (commented):
Apologies if I am too late on sharing my thoughts on this. I went over the PR semantically and have shared some thoughts. 
Thanks for looking into this!

api/v1beta1/azuremanagedmachinepool_webhook.go
LochanRn: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/016209aa3b34393e8c23457f5afdf36db4fb5559/api/v1beta1/azurecluster_validation.go#L43

Copied it from here.

LochanRn: @nawazkh  good catch :)

LochanRn:
@CecileRobertMichon @nawazkh @willie-yao can you guys please take a look at this PR again :)

controllers/azuremanagedmachinepool_reconciler.go
```diff
@@ -95,6 +95,8 @@ func (s *azureManagedMachinePoolService) Reconcile(ctx context.Context) error {
 	ctx, log, done := tele.StartSpanWithLogger(ctx, "controllers.azureManagedMachinePoolService.Reconcile")
 	defer done()
 
+	s.scope.SetSubnetName()
```
nawazkh, line 98: Do we want to set SubnetName every time control enters this reconciles or 
can we have a check where we only set it if it is empty or if subnetName has changed ?

nawazkh (changes requested):
I have a couple of thoughts which I have noted down, especially using `pointer` package over `go-autorest/autorest/to` since we are trying to move away from it.
Apart from that, everything else looks great to me. Thank you for putting this together! Solid work!

api/v1beta1/azuremanagedmachinepool_webhook.go
```diff
@@ -356,6 +365,18 @@ func (m *AzureManagedMachinePool) validateEnableNodePublicIP() error {
 	return nil
 }
 
+func (m *AzureManagedMachinePool) validateSubnetName() error {
+	if m.Spec.SubnetName != nil {
+		subnetRegex := "^[a-zA-Z0-9][a-zA-Z0-9-]{0,78}[a-zA-Z0-9]$"
```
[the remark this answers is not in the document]
LochanRn: tbh i gave chatgpt the docs and asked it to generate a regex :). I then added the test-cases accordingly to check if the regex was correct turned out chatgpt did a solid work. The max character length for a name is 80 so on safe side i guess it stuck to 0-78 :)

azure/scope/managedmachinepool.go
```diff
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/Azure/go-autorest/autorest/to"
```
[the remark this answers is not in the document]
LochanRn: can i know why we want to move away from that  package ?? I think it's an amazing package (just my personal opinion).

controllers/azuremanagedmachinepool_reconciler.go
LochanRn: Why do you think adding the check is going to be advantageous ??

azure/scope/managedmachinepool.go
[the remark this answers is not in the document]
LochanRn: ack and updated the code accordingly :)

azure/scope/managedmachinepool.go
```diff
@@ -166,7 +182,7 @@ func buildAgentPoolSpec(managedControlPlane *infrav1.AzureManagedControlPlane,
 			managedControlPlane.Spec.SubscriptionID,
 			managedControlPlane.Spec.VirtualNetwork.ResourceGroup,
 			managedControlPlane.Spec.VirtualNetwork.Name,
-			managedControlPlane.Spec.VirtualNetwork.Subnet.Name,
+			to.String(getAgentPoolSubnet(managedControlPlane, managedMachinePool)),
```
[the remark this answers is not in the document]
LochanRn: ack

api/v1beta1/azuremanagedmachinepool_webhook.go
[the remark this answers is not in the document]
nawazkh: > chatgpt
/honk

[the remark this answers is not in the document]
nawazkh: > The max character length for a name is 80 so on safe side i guess it stuck to 0-78 :)

Let's make it 80 char length in that case! Gotta be consistent with the supported length to avoid edge cases, especially in CI scenarios.

controllers/azuremanagedmachinepool_reconciler.go
nawazkh: Check especially helps in scenarios when we are setting a nil pointer. (I think we ran into a problem earlier when setting `nil` struct to the _tags_. This `nil` struct updated the `resourceVersion` of the spec, and this triggered a  reconcile on every `resourceVersion` update.)

I am not completely sure if it is applicable in this scenario, but having a check shouldn't harm us.?
Let us see what others have to say for this scenario.
cc: @willie-yao @nojnhuh @CecileRobertMichon

azure/scope/managedmachinepool.go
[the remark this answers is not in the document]
CecileRobertMichon: default to `""` sgtm

azure/scope/managedmachinepool.go
```diff
@@ -134,11 +134,26 @@ func (s *ManagedMachinePoolScope) Name() string {
 	return s.InfraMachinePool.Name
 }
 
+// SetSubnetName updates AzureManagedMachinePool.SubnetName if AzureManagedMachinePool.SubnetName is empty with s.ControlPlane.Spec.VirtualNetwork.Subnet.Name.
+func (s *ManagedMachinePoolScope) SetSubnetName() {
+	s.InfraMachinePool.Spec.SubnetName = getAgentPoolSubnet(s.ControlPlane, s.InfraMachinePool)
```
CecileRobertMichon, line 139: `getAgentPoolSubnet` is just setting it to itself if `s.InfraMachinePool.Spec.SubnetName` is already set. It might be more readable to do:

```
func (s *ManagedMachinePoolScope) SetSubnetName() {
        if s.InfraMachinePool.Spec.SubnetName == nil {
	    s.InfraMachinePool.Spec.SubnetName = pointer.StringPtr(controlPlane.Spec.VirtualNetwork.Subnet.Name)
        }
}
```

And then always use `s.InfraMachinePool.Spec.SubnetName` as the subnet name value for the agent pool (get rid of the `getAgentPoolSubnet` func entirely).

controllers/azuremanagedmachinepool_reconciler.go
CecileRobertMichon: I agree it makes sense to only set the SubnetName if it's empty since we're trying to do defaulting here, not always ensure a value (the check should be part of the `SetSubnetName` function itself). It also improves code readability IMO: commented something like this below https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3132#discussion_r1144089256

api/v1beta1/azuremanagedmachinepool_webhook.go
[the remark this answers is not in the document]
LochanRn: ack

azure/scope/managedmachinepool.go
LochanRn: We cannot get rid of that function as it is used in buildAgentPoolSpec. The AMCP reconciler kicks in first and AMMP will not be triggered until AMCP is initialized. So while creating the AMCP we need that value.

api/v1beta1/azuremanagedmachinepool_webhook.go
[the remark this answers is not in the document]
LochanRn: @nawazkh the regex generated is correct and ensures the name length is restricted till 80 characters. I added a +ve and -ve test case for the same please check.

azure/scope/managedmachinepool.go
LochanRn: updated the functionality for better readability

LochanRn: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/79c47714bfd637d9d979494af08310e712c855cf/controllers/azuremanagedmachinepool_controller.go#L191

azuremanagedmachinepool will not even start reconciling until azuremanagedcontrolplane is initialized and amcp will not be initialized until it has a subnetName it would be a deadlock again.

azure/services/agentpools/agentpools.go
```diff
@@ -46,6 +46,7 @@ type AgentPoolScope interface {
 	SetCAPIMachinePoolReplicas(replicas *int32)
 	SetCAPIMachinePoolAnnotation(key, value string)
 	RemoveCAPIMachinePoolAnnotation(key string)
+	SetSubnetName()
```
CecileRobertMichon, line 49: where is this used by the scope interface? I don't see it used below

azure/scope/managedmachinepool.go
CecileRobertMichon: Ahh I didn't realize that `buildAgentPoolSpec` was also called by managedcontrolplane.go... In that case it would be a problem since the subnet name might not be set yet. This is a bit unusual and due to the fact that we can't create an AKS cluster with no node pools (I was just talking about this with @matthchr and @nojnhuh the other day) which causes us to not have a clean separation of concern between controllers and CRDs (the AzureManagedControlPlane controller reconciles the AzureManagedMachinePool objects too, on the first reconcile).

CecileRobertMichon: Not sure if there is a better/more elegant way to solve this for now, besides doing the same defaulting from the managedcontrolplane reconciler too, or switching back to the webhook approach to ensure that the SubnetName is always set on the AzureManagedMachinePool before the controllers run.

I know previously you ran into an issue with the webhook defaulting where order of creation is not always guaranteed when the objects are all created together, one potential fix for that is to retry getting the AzureManagedControlPlane a few times like https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/api/v1beta1/azuremachine_default.go#L187 is doing.

azure/services/agentpools/agentpools.go
LochanRn: https://github.com/spectrocloud/cluster-api-provider-azure/blob/8f5ec54637ee3fbb6dcba201a0c1f35e8aa8c11b/controllers/azuremanagedmachinepool_reconciler.go#L98

azure/scope/managedmachinepool.go
LochanRn: What is the issue with the current implementation ? Why do you think it is not good w.r.t code quality ? Why are we looking for a better approach ? Just trying to understand what is missing as we have been moving back and forth on this.

azure/services/agentpools/agentpools.go
CecileRobertMichon: This is using `ManagedMachinePoolScope`, not the `AgentPoolScope` scope interface. I bet if you remove this it will still work.

azure/scope/managedmachinepool.go
LochanRn: I am open to move the logic to the webhook in this PR. The idea is to add a wait in AMMP webhook until AMCP is created correct ?

CecileRobertMichon: Let's merge this as-is and explore moving to a webhook in a follow-up PR.

LochanRn: ack

PR #3134, opened 2023-01-31
Custom data model fix

BrennenMM7:
**What type of PR is this?**

When Bootstrap token refreshes the VMSS needs to have the latest model in order for autoscaling machinepools to take place succesfully. Stores a hash of the current custom data contents into AzureMachinePool.Status.CustomDataHash. This allows to update the VMSS data if the custom data changes.

Fixes #2803 

This is utilizing the same logic as created in this pull request [https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2803](url), the only changes are as follows:
- Added to the mock tests to past all unit testing, 
- Removed logic from node surge to prevent constant node rolling updates. 
- Added the watcher to the Kubeadmconfig resource as mentioned in the issue/PR linked. 
- API resource conversions for v1alpha3 and v1alpha4.

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Release note**:

```release-note
- AzureMachinePool Controller will no watch KubeadmConfig to ensure AzureMachinePool Bootstrap data is updated on the VMSS
- The patch loop for the AzureMachinePool will now compare CustomData field to ensure the VMSS instance is updated with the latest model 
```

### DIFF
--- a/azure/scope/machinepool.go
+++ b/azure/scope/machinepool.go
@@ -18,8 +18,10 @@ package scope
 
 import (
 	"context"
+	"crypto/sha256"
 	"encoding/base64"
 	"fmt"
+	"io"
 	"strings"
 
 	azureautorest "github.com/Azure/go-autorest/autorest/azure"
@@ -535,6 +537,12 @@ func (m *MachinePoolScope) Close(ctx context.Context) error {
 		if err := m.updateReplicasAndProviderIDs(ctx); err != nil {
 			return errors.Wrap(err, "failed to update replicas and providerIDs")
 		}
+		if m.HasReplicasExternallyManaged(ctx) {
+			if err := m.updateCustomDataHash(ctx); err != nil {
+				// ignore errors to calculating the custom data hash since it's not absolutely crucial.
+				log.V(4).Error(err, "unable to update custom data hash, ignoring.")
+			}
+		}
 	}
 
 	if err := m.PatchObject(ctx); err != nil {
@@ -566,6 +574,39 @@ func (m *MachinePoolScope) GetBootstrapData(ctx context.Context) (string, error)
 		return "", errors.New("error retrieving bootstrap data: secret value key is missing")
 	}
 	return base64.StdEncoding.EncodeToString(value), nil
+}
+
+// calculateBootstrapDataHash calculates the sha256 hash of the bootstrap data.
+func (m *MachinePoolScope) calculateBootstrapDataHash(ctx context.Context) (string, error) {
+	bootstrapData, err := m.GetBootstrapData(ctx)
+	if err != nil {
+		return "", err
+	}
+	h := sha256.New()
+	n, err := io.WriteString(h, bootstrapData)
+	if err != nil || n == 0 {
+		return "", fmt.Errorf("unable to write custom data (bytes written: %q): %w", n, err)
+	}
+	return fmt.Sprintf("%x", h.Sum(nil)), nil
+}
+
+// HasBootstrapDataChanges calculates the sha256 hash of the bootstrap data and compares it with the saved hash in AzureMachinePool.Status.
+func (m *MachinePoolScope) HasBootstrapDataChanges(ctx context.Context) (bool, error) {
+	newHash, err := m.calculateBootstrapDataHash(ctx)
+	if err != nil {
+		return false, err
+	}
+	return m.AzureMachinePool.GetAnnotations()[azure.CustomDataHashAnnotation] != newHash, nil
+}
+
+// updateCustomDataHash calculates the sha256 hash of the bootstrap data and saves it in AzureMachinePool.Status.
+func (m *MachinePoolScope) updateCustomDataHash(ctx context.Context) error {
+	newHash, err := m.calculateBootstrapDataHash(ctx)
+	if err != nil {
+		return err
+	}
+	m.SetAnnotation(azure.CustomDataHashAnnotation, newHash)
+	return nil
 }
 
 // GetVMImage picks an image from the AzureMachinePool configuration, or uses a default one.
--- a/azure/services/scalesets/scalesets.go
+++ b/azure/services/scalesets/scalesets.go
@@ -53,6 +53,8 @@ type (
 		SetProviderID(string)
 		SetVMSSState(*azure.VMSS)
 		ReconcileReplicas(context.Context, *azure.VMSS) error
+		HasReplicasExternallyManaged(context.Context) bool
+		HasBootstrapDataChanges(context.Context) (bool, error)
 	}
 
 	// Service provides operations on Azure resources.
@@ -276,6 +278,20 @@ func (s *Service) patchVMSSIfNeeded(ctx context.Context, infraVMSS *azure.VMSS) 
 		return nil, errors.Wrap(err, "failed to calculate maxSurge")
 	}
 
+	// If the VMSS is managed by an external autoscaler, we should patch the VMSS if customData has changed.
+	shouldPatchCustomData := false
+	if s.Scope.HasReplicasExternallyManaged(ctx) {
+		shouldPatchCustomData, err := s.Scope.HasBootstrapDataChanges(ctx)
+		if err != nil {
+			return nil, errors.Wrap(err, "unable to calculate custom data hash")
+		}
+		if shouldPatchCustomData {
+			log.V(4).Info("custom data changed")
+		} else {
+			log.V(4).Info("custom data unchanged")
+		}
+	}
+
 	hasModelChanges := hasModelModifyingDifferences(infraVMSS, vmss)
 	var isFlex bool
 	for _, instance := range infraVMSS.Instances {
@@ -295,10 +311,11 @@ func (s *Service) patchVMSSIfNeeded(ctx context.Context, infraVMSS *azure.VMSS) 
 		patch.Sku.Capacity = pointer.Int64(surge)
 	}
 
+	// If the VMSS is managed by an external autoscaler, we should patch the VMSS if customData has changed.
 	// If there are no model changes and no increase in the replica count, do not update the VMSS.
 	// Decreases in replica count is handled by deleting AzureMachinePoolMachine instances in the MachinePoolScope
-	if *patch.Sku.Capacity <= infraVMSS.Capacity && !hasModelChanges {
-		log.V(4).Info("nothing to update on vmss", "scale set", spec.Name, "newReplicas", *patch.Sku.Capacity, "oldReplicas", infraVMSS.Capacity, "hasChanges", hasModelChanges)
+	if *patch.Sku.Capacity <= infraVMSS.Capacity && !hasModelChanges && !shouldPatchCustomData {
+		log.V(4).Info("nothing to update on vmss", "scale set", spec.Name, "newReplicas", *patch.Sku.Capacity, "oldReplicas", infraVMSS.Capacity, "hasModelChanges", hasModelChanges, "shouldPatchCustomData", shouldPatchCustomData)
 		return nil, nil
 	}
 
--- a/exp/controllers/helpers.go
+++ b/exp/controllers/helpers.go
@@ -32,6 +32,7 @@ import (
 	infrav1exp "sigs.k8s.io/cluster-api-provider-azure/exp/api/v1beta1"
 	"sigs.k8s.io/cluster-api-provider-azure/util/reconciler"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	kubeadmv1 "sigs.k8s.io/cluster-api/bootstrap/kubeadm/api/v1beta1"
 	expv1 "sigs.k8s.io/cluster-api/exp/api/v1beta1"
 	"sigs.k8s.io/cluster-api/util"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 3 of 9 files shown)


CecileRobertMichon:
Should we consider only doing this when the MachinePool has the externally managed replicas annotation? I'd be worried about increasing the number of API calls for users that don't need this as this will now make a PUT API call per AzureMachinePool every 10 minutes or so (or whenever token is refreshed, I forget the exact number) which might not scale well for very large clusters with hundreds of MachinePools (granted this will be a problem for users with externally managed replicas too but if we can at least reduce the blast radius that would be an improvement).

CecileRobertMichon:
@BrennenMM7 build is failing, PTAL

BrennenMM7:
@CecileRobertMichon Fixed the build failures and squashed commits again, should be good to go. 👍

exp/api/v1beta1/azuremachinepool_types.go
```diff
@@ -306,6 +306,11 @@ type (
 		// next reconciliation loop.
 		// +optional
 		LongRunningOperationStates infrav1.Futures `json:"longRunningOperationStates,omitempty"`
+
+		// CustomDataHash saves the hash for the stored custom data to be able to check if we need to update the custom data
+		// in reconciliation loops.
+		// +optional
+		CustomDataHash *string `json:"customDataHash"`
```
CecileRobertMichon, line 313: Would an annotation be a better fit for this?

azure/services/scalesets/scalesets.go
```diff
@@ -295,11 +309,19 @@ func (s *Service) patchVMSSIfNeeded(ctx context.Context, infraVMSS *azure.VMSS)
 		patch.Sku.Capacity = pointer.Int64(surge)
 	}
 
+	// If the VMSS is managed by an external autoscaler, we should patch the VMSS if customData has changed.
 	// If there are no model changes and no increase in the replica count, do not update the VMSS.
 	// Decreases in replica count is handled by deleting AzureMachinePoolMachine instances in the MachinePoolScope
-	if *patch.Sku.Capacity <= infraVMSS.Capacity && !hasModelChanges {
-		log.V(4).Info("nothing to update on vmss", "scale set", spec.Name, "newReplicas", *patch.Sku.Capacity, "oldReplicas", infraVMSS.Capacity, "hasChanges", hasModelChanges)
-		return nil, nil
+	if replicaseManagedByAutoscaler {
```
CecileRobertMichon, line 315: this could be 

```
if *patch.Sku.Capacity <= infraVMSS.Capacity && !hasModelChanges && (!hasCustomDataChanges || !replicaseManagedByAutoscaler)  {
```

to avoid the duplication

azure/services/scalesets/scalesets.go
```diff
@@ -295,11 +309,19 @@ func (s *Service) patchVMSSIfNeeded(ctx context.Context, infraVMSS *azure.VMSS)
...
+		if *patch.Sku.Capacity <= infraVMSS.Capacity && !hasModelChanges && !hasCustomDataChanges {
+			log.V(4).Info("nothing to update on vmss", "scale set", spec.Name, "newReplicas", *patch.Sku.Capacity, "oldReplicas", infraVMSS.Capacity, "hasChanges", hasModelChanges)
```
CecileRobertMichon, line 317:
```suggestion
			log.V(4).Info("nothing to update on vmss", "scale set", spec.Name, "newReplicas", *patch.Sku.Capacity, "oldReplicas", infraVMSS.Capacity, "hasModelChanges", hasModelChanges, "hasCustomDataChanges", hasCustomDataChanges)
```

azure/services/scalesets/scalesets.go
```diff
@@ -276,6 +278,18 @@ func (s *Service) patchVMSSIfNeeded(ctx context.Context, infraVMSS *azure.VMSS)
...
+	if err != nil {
+		log.V(4).Error(err, "unable to calculate custom data hash, ignoring")
+	}
+	if hasCustomDataChanges {
+		log.V(4).Info("custom data changed")
+	} else {
+		log.V(4).Info("custom data unchanged")
+	}
+
 	hasModelChanges := hasModelModifyingDifferences(infraVMSS, vmss)
```
CecileRobertMichon, line 293: Have you checked if we get customData back from the existing VMSS? if so we might not even need to store the custom data hash since we would be able to just compare the custom data we have with the existing VMSS custom data

exp/api/v1beta1/azuremachinepool_types.go
CecileRobertMichon: this might be superseded by https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3134#discussion_r1107459190

azure/services/scalesets/scalesets.go
BrennenMM7: We do return the customData from the fetchedVMSS state under the getVirtualMachineScaleSet() function, however the SDKtoVMSS converters remove the OSProfile data in which its stored. If this is the preferred method I can update the converters and types to facilitate this comparison rather than storing a hash.

CecileRobertMichon: yes I think that would be a cleaner implementation and it can rely on the actual state of the world instead needing to store the hash in the API. We can modify the `SDKtoVMSS` converters helper function easily to do what we need since it's part of the CAPZ code.

exp/api/v1beta1/azuremachinepool_types.go
BrennenMM7: Resolved by using the SDKtoVMSS vs the API

azure/services/scalesets/scalesets.go
BrennenMM7: Consolidated if block

azure/services/scalesets/scalesets.go
```diff
@@ -276,6 +278,18 @@ func (s *Service) patchVMSSIfNeeded(ctx context.Context, infraVMSS *azure.VMSS)
 		return nil, errors.Wrap(err, "failed to calculate maxSurge")
 	}
 
+	replicaseManagedByAutoscaler := s.Scope.HasReplicasExternallyManaged(ctx)
```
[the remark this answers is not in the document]
BrennenMM7: I'm not sure how else we would be able to differentiate the reconcile loop based on the existence of the external autoscaler annotation. Open to suggestions.

azure/services/scalesets/scalesets.go
BrennenMM7: Ok I think I've gone through and solidified the logic to use the SDKtoVMSS data, just waiting for the tests to green light and I'll mark this as resolved.

azure/services/scalesets/scalesets.go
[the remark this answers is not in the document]
CecileRobertMichon: I just meant using `s.Scope.HasReplicasExternallyManaged(ctx)` directly instead of assigning it to a variable since it's only used once... not a huge deal though, definitely a nit

CecileRobertMichon (commented):
This looks good @BrennenMM7 thanks for reworking it to remove the need to change the API

/lgtm
/assign @mboersma

CecileRobertMichon:
/hold

Please add a release note in the PR description, this is definitely a user-facing change worth calling out

BrennenMM7:
Release note added.

mboersma (approved):
/lgtm
/approve
/cherry-pick release-1.7

BrennenMM7:
Actually, can we hold the PR before merging. I spoke with @dthorsen this morning and discovered that the API for VMSS in the CustomData field may always return a null field for security purposes and therefore would cause the machine pool to be patched every reconcile loop. 

The stored hash may be required as I suspected before. Need to verify.

BrennenMM7:
@CecileRobertMichon Just confirmed that the AzureSDK will always return a nil value for "Properties.VirtualMachineProfile.OSProfile.CustomData" even if its not true to protect the data stored within that var on Get requests. In order to actually have a working reconcile loop compare the data, the old method of storing a hash into the API before calling the patch is the only way to compare. 

Unless we have an alternative method of providing this token data to the VMSS model for consumption or Microsoft changes the rules on that portion of the API response lol. 

Let me know and I can roll back to the previous commits. 👍

dthorsen:
This is probably worth a discussion during the next office hours. There are some options to consider regarding how best to proceed with regard to API status fields, as well as with regard to documentation and some implications of this behavior.

CecileRobertMichon:
@dthorsen @BrennenMM7 great catch... Apologies for leading you on this incorrect path. I'm still not a fan of adding this into the API since it's not meant to be consumed by the user, maybe a good compromise is storing the hash in an annotation?

BrennenMM7:
@CecileRobertMichon Okay the logic has been reverted and is now stored inside an Annotation on the AMP object instead.

CecileRobertMichon:
build is failing PTAL @BrennenMM7

CecileRobertMichon:
looks like prow job was interrupted?

/retest

azure/scope/machinepool.go
```diff
@@ -535,6 +537,11 @@ func (m *MachinePoolScope) Close(ctx context.Context) error {
 		if err := m.updateReplicasAndProviderIDs(ctx); err != nil {
 			return errors.Wrap(err, "failed to update replicas and providerIDs")
 		}
+		if err := m.updateCustomDataHash(ctx); err != nil {
+			// ignore errors to calculating the custom data hash since it's not absolutely crucial.
+			log.V(4).Error(err, "unable to update custom data hash, ignoring.")
```
CecileRobertMichon, line 542:
```suggestion
			log.V(4).Error(err, "unable to update custom data hash, ignoring")
```

azure/scope/machinepool.go
```diff
@@ -568,6 +575,39 @@ func (m *MachinePoolScope) GetBootstrapData(ctx context.Context) (string, error)
...
+	return m.AzureMachinePool.GetAnnotations()["CustomDataHash"] != newHash, nil
+}
+
+// updateCustomDataHash calculates the sha256 hash of the bootstrap data and saves it in AzureMachinePool.Status.
+func (m *MachinePoolScope) updateCustomDataHash(ctx context.Context) error {
+	newHash, err := m.calculateBootstrapDataHash(ctx)
+	if err != nil {
+		return err
+	}
+	m.SetAnnotation("CustomDataHash", newHash)
```
CecileRobertMichon, line 607: let's make the annotation name a const

azure/services/scalesets/scalesets.go
CecileRobertMichon, line 281:
```suggestion
	replicasManagedByAutoscaler := s.Scope.HasReplicasExternallyManaged(ctx)
```

azure/services/scalesets/scalesets.go
```diff
@@ -276,6 +278,18 @@ func (s *Service) patchVMSSIfNeeded(ctx context.Context, infraVMSS *azure.VMSS)
...
+
+	hasCustomDataChanges, err := s.Scope.HasBootstrapDataChanges(ctx)
+	if err != nil {
+		log.V(4).Error(err, "unable to calculate custom data hash, ignoring")
```
CecileRobertMichon, line 285: This error seems important, at least when `replicasManagedByAutoscaler` is true

Maybe another way to approach this:

```
hasCustomDataChanges := false
if s.Scope.HasReplicasExternallyManaged(ctx) {
  hasCustomDataChanges, err := s.Scope.HasBootstrapDataChanges(ctx) { 
    return nil, errors.Wrap(err, "unable to calculate custom data hash")
  }
  if hasCustomDataChanges {
    log.V(4).Info("custom data changed")
  } else {
    log.V(4).Info("custom data unchanged")
  }
}

File: azure/scope/machinepool.go (hunk 6 above)
Comment (CecileRobertMichon) [reply azure/scope/machinepool.go:542]: > // ignore errors to calculating the custom data hash since it's not absolutely crucial.

not sure I agree, we should return all errors as much as possible, so the controller can requeue a new reconcile loop and retry 

this could lead to critical silent failure if the update is failing and causing a user to not be able to scale up a machine pool, or on the other hand throttling if the VMSS is trying to update on every reconcile because its hash annotation is out of date.

If you are concerned about impacting users who don't need this hash (when HasReplicasExternallyManaged is false), then maybe we can move this update inside an if statement to only make the update when it is needed (`if m.HasReplicasExternallyManaged(ctx)`) - this would make sense only if you apply this suggestion https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3134/files#r1114879941

File: azure/services/scalesets/scalesets.go (hunk 8 above)
Comment (BrennenMM7) [reply azure/services/scalesets/scalesets.go]: Added change

File: azure/services/scalesets/scalesets.go (hunk 5 above)
Comment (BrennenMM7) [reply azure/services/scalesets/scalesets.go]: Added change

File: azure/scope/machinepool.go (hunk 7 above)
Comment (BrennenMM7) [reply azure/scope/machinepool.go]: Added const

File: azure/scope/machinepool.go (hunk 6 above)
Comment (BrennenMM7) [reply azure/scope/machinepool.go]: Added if block to add a non impact to users not utilizing this feature.

File: azure/scope/machinepool.go [hunk 9]
```diff
@@ -51,7 +53,10 @@ import (
 // ScalesetsServiceName is the name of the scalesets service.
 // TODO: move this to scalesets.go once we remove the usage in this package,
 // added here to avoid a circular dependency.
-const ScalesetsServiceName = "scalesets"
+const (
+	ScalesetsServiceName = "scalesets"
+	CustomDataHash       = "CustomDataHash"
```
CecileRobertMichon, line 58: let's move this to https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/const.go where the other annotations are and use the same format and prefix, with a code comment that explains how it's used

```
sigs.k8s.io/cluster-api-provider-azure-vmss-custom-data-hash
```

CecileRobertMichon (commented):
lgtm aside from comment on annotation const and failing unit test job, thanks again and sorry for all the back and forth on this one

CecileRobertMichon:
/hold cancel

please squash!

BrennenMM7:
Can I get a tag on this PR please 😄

nawazkh (commented):
I don't fully understand the problem, will go back try reproducing the error. 
However, I had some doubts and questions jotted down below after reviewing the code.

azure/scope/machinepool.go
```diff
@@ -568,6 +576,39 @@ func (m *MachinePoolScope) GetBootstrapData(ctx context.Context) (string, error)
...
+	return m.AzureMachinePool.GetAnnotations()[azure.CustomDataHashAnnotation] != newHash, nil
+}
+
+// updateCustomDataHash calculates the sha256 hash of the bootstrap data and saves it in AzureMachinePool.Status.
+func (m *MachinePoolScope) updateCustomDataHash(ctx context.Context) error {
+	newHash, err := m.calculateBootstrapDataHash(ctx)
+	if err != nil {
+		return err
+	}
+	m.SetAnnotation(azure.CustomDataHashAnnotation, newHash)
```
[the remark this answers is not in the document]
CecileRobertMichon: the hash was moved to an annotation, see https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3134#issuecomment-1435315588

azure/services/scalesets/scalesets.go
```diff
@@ -276,6 +278,20 @@ func (s *Service) patchVMSSIfNeeded(ctx context.Context, infraVMSS *azure.VMSS)
...
+	shouldPatchCustomData := false
+	if s.Scope.HasReplicasExternallyManaged(ctx) {
+		shouldPatchCustomData, err := s.Scope.HasBootstrapDataChanges(ctx)
+		if err != nil {
+			return nil, errors.Wrap(err, "unable to calculate custom data hash")
+		}
+		if shouldPatchCustomData {
+			log.V(4).Info("custom data changed")
+		} else {
+			log.V(4).Info("custom data unchanged")
```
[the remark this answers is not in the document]
CecileRobertMichon: custom data changes will likely contain sensitive info such as the bootstrap token (which is why it's held in a secret) it should not be printed in logs

exp/controllers/helpers.go
```diff
@@ -317,3 +318,62 @@ func MachinePoolMachineHasStateOrVersionChange(logger logr.Logger) predicate.Fun
...
+		// fetch MachinePool to get reference
+		mp := &expv1.MachinePool{}
+		if err := c.Get(ctx, mpKey, mp); err != nil {
+			if !apierrors.IsNotFound(err) {
+				log.Error(err, "failed to fetch MachinePool for KubeadmConfig")
+			}
+			return []reconcile.Request{}
+		}
+
+		ref := mp.Spec.Template.Spec.Bootstrap.ConfigRef
```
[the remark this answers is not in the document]
CecileRobertMichon: `Bootstrap` is not a pointer so it can never be nil (ref https://github.com/kubernetes-sigs/cluster-api/blob/0d8d072c045572504f4eca32f9fa9cd63d4289ee/api/v1beta1/machine_types.go#L65)

CecileRobertMichon:
/lgtm
/hold

@nawazkh I think I've addressed all your questions/comments, let me know if you're good to remove the hold

nawazkh:
Thanks for the updates @CecileRobertMichon ! 
Kudos @BrennenMM7 🎉 
/hold-cancel
/lgtm

jackfrancis:
@BrennenMM7 did this PR fix #2683 as well?

Thank you!

cc @primeroz

BrennenMM7:
@jackfrancis Yes this should resolve #2683 as the watcher will ensure that the VMSS always has the latest bootstrap token inside the model of the VMSS. 👍

jackfrancis:
@BrennenMM7 thank you! I'll close #2683 and advise that the original reporter of that issue validate the fix with v1.8.0 after it is released (next week), because this PR will be included in that release.

Thanks again for the fix!

PR #3148, opened 2023-02-03
overcome transient errors in ci-entrypoint.sh

jackfrancis:
**What type of PR is this?**

/kind flake

This PR adds retries/idempotency into the `ci-entrypoint.sh` script that we maintain in order to build Kubernetes + Azure clusters on behalf of various k8s upstream projects.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -290,6 +290,9 @@ create-management-cluster: $(KUSTOMIZE) $(ENVSUBST) $(KUBECTL) $(KIND) ## Create
 	$(KUBECTL) wait --for=condition=Available --timeout=5m -n capi-kubeadm-bootstrap-system deployment -l cluster.x-k8s.io/provider=bootstrap-kubeadm
 	$(KUBECTL) wait --for=condition=Available --timeout=5m -n capi-kubeadm-control-plane-system deployment -l cluster.x-k8s.io/provider=control-plane-kubeadm
 
+	# Wait for the ClusterResourceSet CRD resource to be "installed" onto the mgmt cluster before installing CRS addons
+	timeout --foreground 300 bash -c "until $(KUBECTL) get clusterresourcesets -A; do sleep 3; done"
+
 	# install Windows Calico cluster resource set
 	$(KUBECTL) create configmap calico-windows-addon --from-file="$(ADDONS_DIR)/windows/calico" --dry-run=client -o yaml | kubectl apply -f -
 	$(KUBECTL) apply -f templates/addons/windows/calico-resource-set.yaml
@@ -298,7 +301,10 @@ create-management-cluster: $(KUSTOMIZE) $(ENVSUBST) $(KUBECTL) $(KIND) ## Create
 	$(KUBECTL) wait --for=condition=Available --timeout=5m -n capz-system deployment -l cluster.x-k8s.io/provider=infrastructure-azure
 
 	# required sleep for when creating management and workload cluster simultaneously
-	sleep 10
+	# Wait for the core CRD resources to be "installed" onto the mgmt cluster before returning control
+	timeout --foreground 300 bash -c "until $(KUBECTL) get clusters -A; do sleep 3; done"
+	timeout --foreground 300 bash -c "until $(KUBECTL) get azureclusters -A; do sleep 3; done"
+	timeout --foreground 300 bash -c "until $(KUBECTL) get kubeadmcontrolplanes -A; do sleep 3; done"
 	@echo 'Set kubectl context to the kind management cluster by running "$(KUBECTL) config set-context kind-capz"'
 
 .PHONY: create-workload-cluster
--- a/scripts/ci-entrypoint.sh
+++ b/scripts/ci-entrypoint.sh
@@ -131,36 +131,42 @@ create_cluster() {
     "${REPO_ROOT}/hack/create-dev-cluster.sh"
 }
 
-install_addons() {
+# get_cidrs derives the CIDR from the Cluster's '.spec.clusterNetwork.pods.cidrBlocks' metadata
+# any retryable operation in this function must return a non-zero exit code on failure so that we can
+# retry it using a `until get_cidrs; do sleep 5; done` pattern;
+# and any statement must be idempotent so that subsequent retry attempts can make forward progress.
+get_cidrs() {
     # Get cluster CIDRs from Cluster object
     CIDR0=$(${KUBECTL} get cluster "${CLUSTER_NAME}" -o=jsonpath='{.spec.clusterNetwork.pods.cidrBlocks[0]}')
-    CIDR1=$(${KUBECTL} get cluster "${CLUSTER_NAME}" -o=jsonpath='{.spec.clusterNetwork.pods.cidrBlocks[1]}' 2> /dev/null) || true
+    export CIDR0
+    CIDR_LENGTH=$(${KUBECTL} get cluster "${CLUSTER_NAME}" -o=jsonpath='{.spec.clusterNetwork.pods.cidrBlocks}' | jq '. | length')
+    if [[ "${CIDR_LENGTH}" == "2" ]]; then
+        CIDR1=$(${KUBECTL} get cluster "${CLUSTER_NAME}" -o=jsonpath='{.spec.clusterNetwork.pods.cidrBlocks[1]}')
+        export CIDR1
+    fi
+}
 
-    # export the target cluster KUBECONFIG if not already set
-    export KUBECONFIG="${KUBECONFIG:-${PWD}/kubeconfig}"
-
-    # wait for the apiserver pod to be Ready.
-    APISERVER_POD=$("${KUBECTL}" get pods -n kube-system -o name | grep apiserver)
-    "${KUBECTL}" wait --for=condition=Ready -n kube-system "${APISERVER_POD}" --timeout=5m
-
-    # Copy the kubeadm configmap to the calico-system namespace. This is a workaround needed for the calico-node-windows daemonset to be able to run in the calico-system namespace.
-    "${KUBECTL}" create ns calico-system
-    until "${KUBECTL}" get configmap kubeadm-config --namespace=kube-system
-    do
-        # Wait for the kubeadm-config configmap to exist.
-        sleep 2
-    done
-    "${KUBECTL}" get configmap kubeadm-config --namespace=kube-system -o yaml \
-    | sed 's/namespace: kube-system/namespace: calico-system/' \
-    | "${KUBECTL}" create -f -
-
+# install_calico installs Calico CNI componentry onto the Cluster
+# any retryable operation in this function must return a non-zero exit code on failure so that we can
+# retry it using a `until install_calico; do sleep 5; done` pattern;
+# and any statement must be idempotent so that subsequent retry attempts can make forward progress.
+install_calico() {
+    # Copy the kubeadm configmap to the calico-system namespace.
+    # This is a workaround needed for the calico-node-windows daemonset
+    # to be able to run in the calico-system namespace.
+    "${KUBECTL}" create namespace calico-system --dry-run=client -o yaml | kubectl apply -f -
+    if ! "${KUBECTL}" get configmap kubeadm-config --namespace=calico-system; then
+        "${KUBECTL}" get configmap kubeadm-config --namespace=kube-system -o yaml > kubeadm-config-kube-system
+        sed 's/namespace: kube-system/namespace: calico-system/' kubeadm-config-kube-system | "${KUBECTL}" apply -f -
+        rm kubeadm-config-kube-system
+    fi
     # install Calico CNI
     echo "Installing Calico CNI via helm"
-    if [[ "${CIDR0}" =~ .*:.* ]]; then
+    if [[ "${CIDR0:-}" =~ .*:.* ]]; then
         echo "Cluster CIDR is IPv6"
         CALICO_VALUES_FILE="${REPO_ROOT}/templates/addons/calico-ipv6/values.yaml"
         CIDR_STRING_VALUES="installation.calicoNetwork.ipPools[0].cidr=${CIDR0}"
-    elif [[ "${CIDR1}" =~ .*:.* ]]; then
+    elif [[ "${CIDR1:-}" =~ .*:.* ]]; then
         echo "Cluster CIDR is dual-stack"
         CALICO_VALUES_FILE="${REPO_ROOT}/templates/addons/calico-dual-stack/values.yaml"
         CIDR_STRING_VALUES="installation.calicoNetwork.ipPools[0].cidr=${CIDR0},installation.calicoNetwork.ipPools[1].cidr=${CIDR1}"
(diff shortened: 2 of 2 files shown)


scripts/ci-entrypoint.sh
```diff
@@ -132,91 +132,104 @@ create_cluster() {
...
+        if [[ $(${KUBECTL} get cluster "${CLUSTER_NAME}" -o=jsonpath='{.spec.clusterNetwork.pods.cidrBlocks}' | jq '. | length') == 2 ]]; then
+            CIDR1=$(${KUBECTL} get cluster "${CLUSTER_NAME}" -o=jsonpath='{.spec.clusterNetwork.pods.cidrBlocks[1]}') || continue
+        fi
+        CIDRS_RETRIEVED="true"
+    done
     # export the target cluster KUBECONFIG if not already set
     export KUBECONFIG="${KUBECONFIG:-${PWD}/kubeconfig}"
+    while [[ "${CLUSTER_READY}" != "true" ]]; do
+        # Copy the kubeadm configmap to the calico-system namespace. This is a workaround needed for the calico-node-windows daemonset to be able to run in the calico-system namespace.
+        cat <<EOF | "${KUBECTL}" apply -f - || continue
```
mboersma, line 153: Is this more robust somehow than `kubectl create ns calico-system`?

jackfrancis: it's idempotent so we don't have to worry about keeping track of its success separately and then not running it during subsequent retry cycles

Makefile
```diff
@@ -289,6 +289,7 @@ create-management-cluster: $(KUSTOMIZE) $(ENVSUBST) $(KUBECTL) $(KIND) ## Create
 	$(KUBECTL) wait --for=condition=Available --timeout=5m -n capi-system deployment -l cluster.x-k8s.io/provider=cluster-api
 	$(KUBECTL) wait --for=condition=Available --timeout=5m -n capi-kubeadm-bootstrap-system deployment -l cluster.x-k8s.io/provider=bootstrap-kubeadm
 	$(KUBECTL) wait --for=condition=Available --timeout=5m -n capi-kubeadm-control-plane-system deployment -l cluster.x-k8s.io/provider=control-plane-kubeadm
+	timeout --foreground 300 bash -c "until $(KUBECTL) get clusterresourcesets -A; do sleep 3; done"
```
mboersma, line 292: In the logs this printed:

```shell
timeout --foreground 300 bash -c "until /home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/hack/tools/bin/kubectl-v1.25.6 get clusterresourcesets -A; do sleep 3; done"
No resources found
# install Windows Calico cluster resource set
```

Shouldn't the script return an error if no CRSs are found?

scripts/ci-entrypoint.sh
```diff
@@ -132,91 +132,104 @@ create_cluster() {
...
+        else
+            echo "Cluster CIDR is IPv4"
+            CALICO_VALUES_FILE="${REPO_ROOT}/templates/addons/calico/values.yaml"
+            CIDR_STRING_VALUES="installation.calicoNetwork.ipPools[0].cidr=${CIDR0}"
         fi
 
-        CCM_CLUSTER_CIDR="${CIDR0}"
-        if [[ -n "${CIDR1}" ]]; then
-            CCM_CLUSTER_CIDR="${CIDR0}\,${CIDR1}"
+        "${HELM}" repo add projectcalico https://projectcalico.docs.tigera.io/charts || continue
```
mboersma, line 182: Not this PR's fault, but these `helm` commands lead to these warnings:

```
Installing Calico CNI via helm
Cluster CIDR is IPv4
WARNING: Kubernetes configuration file is group-readable. This is insecure. Location: /home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/kubeconfig
WARNING: Kubernetes configuration file is world-readable. This is insecure. Location: /home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/kubeconfig
"projectcalico" has been added to your repositories
WARNING: Kubernetes configuration file is group-readable. This is insecure. Location: /home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/kubeconfig
WARNING: Kubernetes configuration file is world-readable. This is insecure. Location: /home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/kubeconfig
NAME: calico
```

Presumably we can fix that with a `chmod`. @jackfrancis do you want to add that here or should I open an issue/PR?

Makefile
jackfrancis: Good question!

No, it should return an error if the ClusterResourceSet CRD hasn't yet been installed. While I was putting this together I noticed that flake (we install the calico CRS and the cluster complains about not knowing about that resource type).

scripts/ci-entrypoint.sh
jackfrancis: Actually can you add a note to https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3126

I think we should fix that there

jackfrancis:
/test pull-cluster-api-provider-azure-ci-entrypoint

Let's run this like 10 times to prove that this indeed is less flaky

CecileRobertMichon:
> Let's run this like 10 times to prove that this indeed is less flaky

don't you wish you could automate that

scripts/ci-entrypoint.sh
CecileRobertMichon: `kubectl create namespace <add-namespace-here> --dry-run=client -o yaml | kubectl apply -f -` is equivalent and slightly easier on the eyes

scripts/ci-entrypoint.sh
```diff
@@ -132,91 +132,104 @@ create_cluster() {
...
-            --set cloudControllerManager.cloudConfig="${CLOUD_CONFIG}" \
-            --set cloudControllerManager.cloudConfigSecretName="${CONFIG_SECRET_NAME}" \
-            --set cloudControllerManager.logVerbosity="${CCM_LOG_VERBOSITY}" \
-            --set-string cloudControllerManager.clusterCIDR="${CCM_CLUSTER_CIDR}"
-    fi
 
-    export -f wait_for_nodes
-    timeout --foreground 1800 bash -c wait_for_nodes
+        export -f wait_for_nodes
+        timeout --foreground 1800 bash -c wait_for_nodes || exit 1
```
CecileRobertMichon, line 224: do we need the explicit exit? won't the command already exit non-zero if it fails?

scripts/ci-entrypoint.sh
```diff
@@ -132,91 +132,104 @@ create_cluster() {
 }
 
 install_addons() {
```
CecileRobertMichon, line 134: this func is getting a little big, might be time to split up the each different operation in its own func?

CecileRobertMichon: that would also allow retrying each block independently instead of having to keep track of `CIDRS_RETRIEVED`, `CALICO_INSTALLED`, etc.

jackfrancis:
That's the 2nd run in a row that does this:

```
pod/calico-node-windows-l5q8d condition met
pod/calico-node-windows-xl4p9 condition met
```

when things look like this:

```
calico-system      calico-node-windows-l5q8d                                 1/2     NotReady   0          2m7s    10.1.0.6         capz-v0pj-zdtx7                   <none>           <none>
calico-system      calico-node-windows-xl4p9                                 1/2     NotReady   0          2m7s    10.1.0.5         capz-v0pj-tsb7l                   <none>           <none>
```

I'm working on a better way to block forward progress until all pods are verified as 100% operational.

scripts/ci-entrypoint.sh
jackfrancis: nice!

scripts/ci-entrypoint.sh
jackfrancis: ACK

scripts/ci-entrypoint.sh
jackfrancis: Depending on what set of kubectl ops we're making a "transaction" of we still will need to do this type of accounting, but if splitting up into multiple funcs is more maintainable we can easily do that.

scripts/ci-entrypoint.sh
jackfrancis: done

scripts/ci-entrypoint.sh
jackfrancis: I've tried to incorporate your suggestion, and I do think this makes things more readable and maintainable, PTAL, thanks!

scripts/ci-entrypoint.sh
```diff
@@ -32,6 +32,9 @@ make --directory="${REPO_ROOT}" "${KUBECTL##*/}" "${HELM##*/}" "${KIND##*/}" "${
 # export the variables so they are available in bash -c wait_for_nodes below
 export KUBECTL
 export HELM
+# initialize variables that are reused across functions
+CIDR0=""
```
CecileRobertMichon, line 36: not sure we need this? we can just use `{CIDR0:-}` where there are places where the variable might not be set yet, but for CIDR0 it seems that it's always set after occurrence and for CIDR1 we can just do `|| CIDR1=""` instead of `|| true` on waht used to be line 137

scripts/ci-entrypoint.sh
```diff
@@ -131,29 +134,33 @@ create_cluster() {
...
-    "${KUBECTL}" create ns calico-system
-    until "${KUBECTL}" get configmap kubeadm-config --namespace=kube-system
-    do
-        # Wait for the kubeadm-config configmap to exist.
-        sleep 2
-    done
-    "${KUBECTL}" get configmap kubeadm-config --namespace=kube-system -o yaml \
-    | sed 's/namespace: kube-system/namespace: calico-system/' \
-    | "${KUBECTL}" create -f -
+    export CIDR0
```
CecileRobertMichon, line 144: export and set on the same line ?

scripts/ci-entrypoint.sh
```diff
@@ -131,29 +134,33 @@ create_cluster() {
...
+    CIDR_LENGTH=$(${KUBECTL} get cluster "${CLUSTER_NAME}" -o=jsonpath='{.spec.clusterNetwork.pods.cidrBlocks}' | jq '. | length')
+    if [[ "${CIDR_LENGTH}" == "2" ]]; then
+        CIDR1=$(${KUBECTL} get cluster "${CLUSTER_NAME}" -o=jsonpath='{.spec.clusterNetwork.pods.cidrBlocks[1]}')
+        export CIDR1
```
CecileRobertMichon, line 148: I think this could be a one liner

```
export CIDR1=$(${KUBECTL} get cluster "${CLUSTER_NAME}" -o=jsonpath='{.spec.clusterNetwork.pods.cidrBlocks[1]}' 2> /dev/null) || export CIDR1=""
```

scripts/ci-entrypoint.sh
```diff
@@ -131,29 +134,35 @@ create_cluster() {
...
+# and any statement must be idempotent so that subsequent retry attempts can make forward progress.
+install_calico() {
+    # Copy the kubeadm configmap to the calico-system namespace.
+    # This is a workaround needed for the calico-node-windows daemonset
+    # to be able to run in the calico-system namespace.
+    "${KUBECTL}" create namespace calico-system --dry-run=client -o yaml | kubectl apply -f -
+    if ! "${KUBECTL}" get configmap kubeadm-config --namespace=calico-system; then
+        "${KUBECTL}" get configmap kubeadm-config --namespace=kube-system -o yaml > kubeadm-config-kube-system
+        sed 's/namespace: kube-system/namespace: calico-system/' kubeadm-config-kube-system | "${KUBECTL}" apply -f -
+        rm kubeadm-config-kube-system
```
CecileRobertMichon, line 164: what's the advantage of doing this in two steps instead of 1? I can think of at least one downside which is that by adding this new temporary file we could end up not cleaning up after ourselves if the script gets interrupted before the rm line

scripts/ci-entrypoint.sh
```diff
@@ -251,11 +296,6 @@ cleanup() {
 }
 
 on_exit() {
-    if [[ -n ${KUBECONFIG:-} ]]; then
-        "${KUBECTL}" get nodes -owide || echo "Unable to get nodes"
```
CecileRobertMichon, line 255: I think we're loosing information by removing this? Because the script may fail and be interrupted at any point, before reaching the `wait_for_nodes` and `wait_for_pods` functions getting called, whereas this always run (we're doing `trap on_exit EXIT` below)

scripts/ci-entrypoint.sh
jackfrancis: it was originally thus, and then shellcheck complained

scripts/ci-entrypoint.sh
jackfrancis: ditto above

scripts/ci-entrypoint.sh
jackfrancis: it's because a <i/o fetch> and then stream to sed in one line will swallow the error

For example:

```
Francisbook:cluster-api-provider-azure jackfrancis$ curl non-existent-url
curl: (6) Could not resolve host: non-existent-url
Francisbook:cluster-api-provider-azure jackfrancis$ echo $?
6
Francisbook:cluster-api-provider-azure jackfrancis$ curl non-existent-url | sed 's/foo/bar/'
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:--  0:00:04 --:--:--     0curl: (6) Could not resolve host: non-existent-url
Francisbook:cluster-api-provider-azure jackfrancis$ echo $?
0
```

We need to capture an error from `kubectl get` so that we know to retry.

scripts/ci-entrypoint.sh
CecileRobertMichon: got it, this one can still be

```
CIDR1=$(${KUBECTL} get cluster "${CLUSTER_NAME}" -o=jsonpath='{.spec.clusterNetwork.pods.cidrBlocks[1]}' 2> /dev/null) || CIDR1=""
export CIDR1
```

scripts/ci-entrypoint.sh
jackfrancis: Yeah, good question. I removed it because I wasn't sure of the value of printing the nodes/pods in the following scenarios that this trapped function cover:

- at the *very* end of a successful test run (like, long after capz has actually built a cluster and perhaps some other things have been done to it)
- at some failure point in the script, perhaps before a cluster has even been built?

Perhaps there is some contract between the ci-entrypoint user and this script that the user expects ci-entrypoint to log a terminal "this is what the cluster looks like at a glance, at the very end of the test run", but it's not clear why a ci-entrypoint-backed test wouldn't do that on its own.

Here's an example of a successful run using the current implementation of vanilla ci-entrypoint.sh run (kicked off via CAPZ PR CI) where there's some "kubectl get nodes stuttering" (it's presented twice).

It does seem like getting rid of this `k get pods -A` statement in the exit func removes that info from test stdout history. If we want that, I'd advocate we put that as a successful logging outcome of the new `wait_for_pods` func.

Basically, as far as ci-entrypoint.sh is concerned, it's only able to provide debug output for its own operations before we hit `on_exit`.

Anyways, those are my thoughts, open to other ideas!

scripts/ci-entrypoint.sh
jackfrancis: yes, I think we can remove these vars, good catch!

scripts/ci-entrypoint.sh
CecileRobertMichon: maybe the outputs of these should be moved to the logger as files that get saved in _artifacts? in the absence of that, it's definitely helped me debug a few failures "at a glance"

jackfrancis: ACK, I'll add it back, and add pods output to `wait_for_pods` as well, if folks find this info useful then it can't hurt to have a before/after display of nodes/pods

scripts/ci-entrypoint.sh
jackfrancis: I think we can skip the `||` empty string assignment (at least bash didn't complain about unbound variables, we'll see what the linter says)

scripts/ci-entrypoint.sh
jackfrancis: done

CecileRobertMichon (commented):
/lgtm
/assign @mboersma

mboersma (approved):
/lgtm
/approve

mboersma:
I did a manual cherry-pick in #3155.

PR #3166, opened 2023-02-10
Add clusterclass template for Tilt

willie-yao:
**What type of PR is this?**
/kind feature

This PR adds a default `clusterclass` template for Tilt. The only existing `clusterclass` template is generated for prow and doesn't provide a good example for testing/development purposes. This PR also generates the prow `clusterclass` template using the Tilt `clusterclass` template as a base

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```


sonasingh46:
+1 
Thanks for working on this.

templates/cluster-template-clusterclass.yaml
```diff
@@ -0,0 +1,280 @@
+apiVersion: cluster.x-k8s.io/v1beta1
```
CecileRobertMichon, line 1: Consider using kustomize to generate the template and reuse this base flavor for the prow test template as a lot of it is shared

CecileRobertMichon:

The clusterclass test is marked as optional

templates/cluster-template-clusterclass.yaml
willie-yao: @CecileRobertMichon Is there a way to patch a list using kustomize? I tried applying a patch to a list, but kustomize ends up replacing the entire list instead of merging it. [Example](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3166/commits/b3e516932f15c7affe90c15e3acc40bd203c4790#diff-c58e7ac2fdd1ff7852627da5e3b5e4cae9262265e522e4834aa7d978124427efL47)

CecileRobertMichon: not with patchesStrategicMerge :/ 

you'll need to "patch" with the entire desired list

CecileRobertMichon: or you can use `patchesJson6902`

https://stackoverflow.com/questions/71622419/adding-items-to-a-list-with-kubectl-kustomize

templates/test/ci/prow-clusterclass-ci-default/patches.yaml
```diff
@@ -1,8 +1,46 @@
 apiVersion: cluster.x-k8s.io/v1beta1
 kind: ClusterClass
 metadata:
-  name: ci-default
+  name: ${CLUSTER_NAME}
```
CecileRobertMichon, line 4: I think the idea is that a single ClusterClass can be reused by many clusters, so even though this technically works, I don't know that it really demonstrates how a ClusterClass should be used by tying it's name to the cluster's name, which implies there would be one ClusterClass per cluster.

willie-yao: That can definitely be misleading.. I'll refactor the name to not be dependent on the cluster name

CecileRobertMichon:
I'm now getting

> Error from server (Invalid): error when creating "STDIN": admission webhook "validation.cluster.cluster.x-k8s.io" denied the request: Cluster.cluster.x-k8s.io "clusterclass-2261" is invalid: spec.topology.workers.machineDeployments[0].class: Invalid value: "clusterclass-2261-worker": MachineDeploymentClass with name "clusterclass-2261-worker" does not exist in ClusterClass "default"

willie-yao:
Hmm that's weird, I don't run into this problem when I deploy with Tilt. Have you tried restarting Tilt before switching to this branch? @CecileRobertMichon

CecileRobertMichon:
@willie-yao you're right restarting Tilt helped. But now I'm running into a different error:

```
E0217 01:46:05.164105      42 controller.go:326]  "msg"="Reconciler error" "error"="failed to configure azure settings and credentials for Identity: error creating azure services. subscriptionID is not set in cluster or AZURE_SUBSCRIPTION_ID env var" "AzureMachine"={"name":"clusterclass-21350-md-0-infra-5lq6j-kbvhz","namespace":"default"} "controller"="azuremachine" "controllerGroup"="infrastructure.cluster.x-k8s.io" "controllerKind"="AzureMachine" "name"="clusterclass-21350-md-0-infra-5lq6j-kbvhz" "namespace"="default" "reconcileID"="973ec36c-ee7d-411b-9e82-30b10b52f57e"
```

templates/flavors/clusterclass/kustomization.yaml
```diff
@@ -0,0 +1,10 @@
+namespace: default
+resources:
+  - patches/clusterclass.yaml
+  - patches/cluster.yaml
```
CecileRobertMichon, line 4: I like how you split up each resource here, makes the templates a lot more modular. I think we should consider doing this for more of our templates. 

consider taking the base resources out of the "patches" directory since they aren't patches but base resources.

willie-yao:
> It probably works in the test because it relies on the manager environment variables but that's deprecated and we should be using AAD pod identity with per cluster identity (https://capz.sigs.k8s.io/topics/getting-started.html#warning-1)

Ah that makes sense! I think I had an environment variable set in my VM. Let me fix that and thanks for catching this!

CecileRobertMichon (commented):
I was able to create cluster locally with Tilt using the clusterclass flavor, looks good! Thanks @willie-yao 

/lgtm
/approve
/retest

PR #3177, opened 2023-02-14
Fix Windows Calico for Tilt

CecileRobertMichon:
**What type of PR is this?**

/kind cleanup

**What this PR does / why we need it**: Label in the template doesn't match so Calico was never installed for Windows: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/templates/cluster-template-windows.yaml#L5

Verified that this cluster resource set is not used anywhere else (e2e tests have their own in each template).

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/templates/addons/windows/calico-resource-set.yaml
+++ b/templates/addons/windows/calico-resource-set.yaml
@@ -7,7 +7,7 @@ spec:
   strategy: "ApplyOnce"
   clusterSelector:
     matchLabels:
-      cni: calico
+      cni-windows: calico
       windows: enabled
   resources:
     - name: calico-windows-addon


marosset (approved):
/lgtm

mboersma (approved):
/approve

mboersma:
/hold

Are there extra e2e tests applicable here? Also I assume this should be cherry-picked to the supported branches.

CecileRobertMichon:
@mboersma see PR description 

> Verified that this cluster resource set is not used anywhere else (e2e tests have their own in each template)

We can cherry-pick but since this only affects tilt local development could go either way...

mboersma:
/hold cancel

I guess since it's Tilt-specific it doesn't really count as a bug fix. My assumption is developers are basically always using `main`.

PR #3187, opened 2023-02-17
fix tag delete following resource create

### LINKED ISSUE #3185: Deleting from `spec.additionalTags` following create does not take effect
/kind bug

- Create an AzureCluster or AzureManagedCluster with values specified in `spec.additionalTags`
- After the cluster is provisioned, delete any or all values from `spec.additionalTags`
- The CAPZ resource is updated but the tags still exist on the Azure resource group after several minutes

- The tags deleted from `spec.additionalTags` should no longer exist on the Azure resource group

CAPZ stores extra state regarding additional tags in an annotation to determine exactly what changes to make based on the desired state in `spec.additionalTags`, e.g.
```
sigs.k8s.io/cluster-api-provider-azure-last-applied-tags-rg: '{"jon":"washere"}'
```

This annotation is only created after a change is noticed the first time:
https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/b1883331bf11d57f27e9a171f9cfa579d9739cd6/azure/services/tags/tags.go#L110

But deleting from `spec.additionalTags` results in a situation where all of the desired tags exist in the actual tags, so no change appears necessary.

For CAPZ to delete a tag, it must exist in the `last-applied-tags` annotation.

**Environment:**

nojnhuh:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**: Fixes an issue where deleting any values from an AzureCluster's or AzureManagedCluster's `spec.additionalTags` before making any other changes to that field never takes effect on the Azure resource group because the `last-applied-tags` annotation is not applied initially. This change ensures the annotation is set on each reconciliation loop, whether or not the tags need to be updated.

Fixes #3185

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/azure/services/tags/tags.go
+++ b/azure/services/tags/tags.go
@@ -105,12 +105,13 @@ func (s *Service) Reconcile(ctx context.Context) error {
 					return errors.Wrap(err, "cannot update tags")
 				}
 			}
-
-			// We also need to update the annotation if anything changed.
-			if err := s.Scope.UpdateAnnotationJSON(tagsSpec.Annotation, newAnnotation); err != nil {
-				return err
-			}
 			log.V(2).Info("successfully updated tags")
+		}
+
+		// We also need to update the annotation even if nothing changed to
+		// ensure it's set immediately following resource creation.
+		if err := s.Scope.UpdateAnnotationJSON(tagsSpec.Annotation, newAnnotation); err != nil {
+			return err
 		}
 	}
 	return nil
--- a/azure/services/tags/tags_test.go
+++ b/azure/services/tags/tags_test.go
@@ -218,6 +218,7 @@ func TestReconcileTags(t *testing.T) {
 					},
 				}}, nil)
 				s.AnnotationJSON("my-annotation").Return(map[string]interface{}{"key": "value"}, nil)
+				s.UpdateAnnotationJSON("my-annotation", map[string]interface{}{"key": "value"})
 			},
 		},
 	}


willie-yao (commented):
This looks great! Just one comment: Is there a way to add a test case in `TestReconcileTags` for ensuring the annotations are updated even when the tags aren't changed?

azure/services/tags/tags.go
```diff
@@ -105,13 +105,14 @@ func (s *Service) Reconcile(ctx context.Context) error {
...
-				return err
-			}
 			log.V(2).Info("successfully updated tags")
 		}
+
+		// We also need to update the annotation even if nothing changed to
+		// ensure it's set immediately following resource creation.
+		if err := s.Scope.UpdateAnnotationJSON(tagsSpec.Annotation, newAnnotation); err != nil {
+			return err
+		}
```
nawazkh, line 115: Will this change lead to a new reconciliation for every tag from the for loop up above (as the resource version gets updated on annotation update) ? 
If yes, could such a change create throttling situation?

nojnhuh:
> This looks great! Just one comment: Is there a way to add a test case in `TestReconcileTags` for ensuring the annotations are updated even when the tags aren't changed?

@willie-yao Yes, this does include an update to the "tags unchanged" test case.

azure/services/tags/tags.go
nojnhuh: This `UpdateAnnotationJSON` method doesn't make any API calls, it only sets the annotation on the corresponding CAPZ resource like this: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/7c920383f52a58547bd6ac9ae161b553ca564111/azure/scope/cluster.go#L980-L999

Then at the very end of the AzureCluster-level `Reconcile`, that resource is committed to the API with all of the changes made in that reconciliation, so different iterations of this for loop would be coalesced into a single Kubernetes API update. I think this change in particular will only introduce at most one additional reconciliation during the lifetime of the resource which adds the annotation the first time.

willie-yao (approved):
/lgtm

azure/services/tags/tags.go
```diff
@@ -105,13 +105,14 @@ func (s *Service) Reconcile(ctx context.Context) error {
...
-			// We also need to update the annotation if anything changed.
-			if err := s.Scope.UpdateAnnotationJSON(tagsSpec.Annotation, newAnnotation); err != nil {
-				return err
-			}
 			log.V(2).Info("successfully updated tags")
 		}
+
+		// We also need to update the annotation even if nothing changed to
+		// ensure it's set immediately following resource creation.
+		if err := s.Scope.UpdateAnnotationJSON(tagsSpec.Annotation, newAnnotation); err != nil {
```
CecileRobertMichon, line 113: instead of updating it every time, would it make sense (and be possible) to only set it the first time we go through here and the annotation doesn't exist? otherwise we're updating the annotation every time even if nothing changed

nojnhuh: When nothing changes, doesn't this basically turn into a no-op? A patch is still unconditionally applied at the end of the reconciliation loop whether or not anything in the AzureCluster, etc. changes, so I don't think this would generate any additional ongoing API overhead. That being said, it would be simple to wrap this in something like a `if tagsSpec.Annotation != newAnnotation` block.

azure/services/tags/tags.go
nawazkh: Thank you for pinning `cluster-api-provider-azure/azure/scope/cluster.go` lines above, helps to put things in perspective. 
Following the code flow from the perspective of generating additional reconciliations, I do not see any `Patch/Update/return reconcile:true` calls on updating annotations or in function `s.Scope.UpdateAnnotationJSON()`. 

So looks good to me!

azure/services/tags/tags.go
CecileRobertMichon: fair

PR #3194, opened 2023-02-21
MachinePools: Fix AzureMachinePool default for SystemAssignedIdentityRole

willie-yao:
**What type of PR is this?**
/kind bug

This PR fixes an issue with MachinePools where if the old deprecated `roleAssignmentName` field is set and the new `systemAssignedIdentityRole.Name` isn't, the defaulting webhook sets both fields, resulting in an error in the validation webhook. This is because only one of those fields should be set at a time.

This also fixes failures in the [CAPI upgrade tests](https://testgrid.k8s.io/sig-cluster-lifecycle-cluster-api-provider-azure#capi-periodic-upgrade-main-1-23-1-24).

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/exp/api/v1beta1/azuremachinepool_default.go
+++ b/exp/api/v1beta1/azuremachinepool_default.go
@@ -72,18 +72,14 @@ func (amp *AzureMachinePool) SetIdentityDefaults(subscriptionID string) {
 		return
 	}
 	if amp.Spec.Identity == infrav1.VMIdentitySystemAssigned {
-		if amp.Spec.RoleAssignmentName == "" {
-			amp.Spec.RoleAssignmentName = string(uuid.NewUUID())
-		}
 		if amp.Spec.SystemAssignedIdentityRole == nil {
 			amp.Spec.SystemAssignedIdentityRole = &infrav1.SystemAssignedIdentityRole{}
 		}
-		if amp.Spec.SystemAssignedIdentityRole.Name == "" {
+		if amp.Spec.RoleAssignmentName != "" {
+			amp.Spec.SystemAssignedIdentityRole.Name = amp.Spec.RoleAssignmentName
+			amp.Spec.RoleAssignmentName = ""
+		} else if amp.Spec.SystemAssignedIdentityRole.Name == "" {
 			amp.Spec.SystemAssignedIdentityRole.Name = string(uuid.NewUUID())
-			if amp.Spec.RoleAssignmentName != "" {
-				amp.Spec.SystemAssignedIdentityRole.Name = amp.Spec.RoleAssignmentName
-				amp.Spec.RoleAssignmentName = ""
-			}
 		}
 		if amp.Spec.SystemAssignedIdentityRole.Scope == "" {
 			// Default scope to the subscription.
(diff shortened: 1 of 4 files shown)


exp/api/v1beta1/azuremachinepool_default.go
```diff
@@ -72,18 +72,14 @@ func (amp *AzureMachinePool) SetIdentityDefaults(subscriptionID string) {
...
 	}
 	if amp.Spec.Identity == infrav1.VMIdentitySystemAssigned {
-		if amp.Spec.RoleAssignmentName == "" {
-			amp.Spec.RoleAssignmentName = string(uuid.NewUUID())
-		}
 		if amp.Spec.SystemAssignedIdentityRole == nil {
 			amp.Spec.SystemAssignedIdentityRole = &infrav1.SystemAssignedIdentityRole{}
 		}
-		if amp.Spec.SystemAssignedIdentityRole.Name == "" {
+		if amp.Spec.RoleAssignmentName != "" {
```
CecileRobertMichon, line 78: is there a unit test for this that can cover the broken scenario?

CecileRobertMichon:

runs the upgrade test ^

exp/api/v1beta1/azuremachinepool_default.go
```diff
@@ -72,18 +72,14 @@ func (amp *AzureMachinePool) SetIdentityDefaults(subscriptionID string) {
 		return
 	}
 	if amp.Spec.Identity == infrav1.VMIdentitySystemAssigned {
-		if amp.Spec.RoleAssignmentName == "" {
-			amp.Spec.RoleAssignmentName = string(uuid.NewUUID())
-		}
```
nawazkh, line 77: Should we be preserving this workflow where the `amp.Spec.RoleAssignmentName == ""` if empty will be assigned a new `string(uuid.NewUUID())` ?

nawazkh (commented):
Good find Willie! The change looks good to me, just curious if we should be preserving earlier functionality I mentioned below ?

exp/api/v1beta1/azuremachinepool_default.go
willie-yao: We used to set `amp.Spec.RoleAssignmentName` to a new UUID if it's empty so we can set `amp.Spec.SystemAssignedIdentityRole.Name` to its value. We never intend to use `amp.Spec.RoleAssignmentName` as it is a deprecated field. In the new code, we are just checking if the user did indeed set `amp.Spec.RoleAssignmentName`. If not, we set `amp.Spec.SystemAssignedIdentityRole.Name` to a new UUID, which is the new field that replaces the deprecated one.

CecileRobertMichon (commented):
/lgtm

mboersma (approved):
/lgtm
/approve

mboersma:
This is blocked on the "Header Rules" GitHub action, although if you click through it appears the Netlify docs preview actually succeeded.

@willie-yao I've run into this before, and AFAICT there's no way to manually restart the GH actions. Could you rebase this off master or make and squash a no-op commit, then force-push it? I'll re-review and approve.

willie-yao:
@mboersma Got it, thanks for letting me know!

mboersma (approved):
/lgtm
/approve

PR #2074, opened 2022-02-10
correct variable in SetupWithManager function in ammp controller

thiagompc:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #2058

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
None
```

PR #3195, opened 2023-02-22
Add unit tests for AzureManagedMachinePool Controller

### LINKED ISSUE #2078: Add unit tests for AzureManagedMachinePool controller
/kind feature

The file `exp/controllers/azuremanagedmachinepool_controller.go` does not have corresponding unit tests, but should. 

PR #2074 shows AMMPs are in use, and ideally that change would be covered by testing. I think it's too much to ask for that one-line change to implement an entire test module, so I opened this issue.

Its neighbor `exp/controllers/azuremachinepool_controller.go` has some unit tests in a couple different files. Those could perhaps be used as a model or starting point.

xiujuanx:
**What type of PR is this?**

Fixes #2078 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
NONE
```


xiujuanx:
Currently, my added unit tests can't pass because below issue:
```
=== RUN   TestAzureManagedMachinePoolReconcile
=== RUN   TestAzureManagedMachinePoolReconcile/should_successfully_reconcile
(string) (len=11) "==========="
    controller.go:269: missing call(s) to *mock_azure.MockReconciler.Reconcile(expected a context.Context, but got <nil>) /Users/xiujuanx/workspace/projects/cluster-api-provider-azure/controllers/azuremanagedmachinepool_controller_test.go:36
    controller.go:269: aborting test due to missing call(s)
=== RUN   TestAzureManagedMachinePoolReconcile/should_successfully_delete
(string) (len=11) "==========="
    controller.go:269: missing call(s) to *mock_azure.MockReconciler.Delete(expected a context.Context, but got <nil>) /Users/xiujuanx/workspace/projects/cluster-api-provider-azure/controllers/azuremanagedmachinepool_controller_test.go:50
    controller.go:269: aborting test due to missing call(s)
--- FAIL: TestAzureManagedMachinePoolReconcile (0.00s)
    --- FAIL: TestAzureManagedMachinePoolReconcile/should_successfully_reconcile (0.00s)
    --- FAIL: TestAzureManagedMachinePoolReconcile/should_successfully_delete (0.00s)
```
So kind reviewers, could you pls help me take a look my codes? Anything I am missing? Thanks a lot!

willie-yao:
@xiujuanx I tested this out locally and was able to reproduce your issue. I'm not sure why the error is happening.. basically, the test is waiting for calls to `Reconcile` and `Delete`, which should be covered by the call to `controller.Reconcile`. I will be looking into the source of the error. 

In the meantime, you can try testing this without the use of mock functions like in the [AzureMachinePool controller test](https://github.com/willie-yao/cluster-api-provider-azure/blob/main/exp/controllers/azuremachinepool_controller_test.go)

xiujuanx:
Thanks @willie-yao ! The root cause of my issue was I didn't call all necessary mock funcs during `Reconcile` and `Delete`. This causes some funcs return `nil` value. Then `nil` pointer call others funcs will cause this error.
Now I have fixed it. And locally run tests pass:
```
=== RUN   TestAzureManagedMachinePoolReconcile
=== RUN   TestAzureManagedMachinePoolReconcile/Reconcile_succeed
=== RUN   TestAzureManagedMachinePoolReconcile/Reconcile_delete
--- PASS: TestAzureManagedMachinePoolReconcile (0.01s)
    --- PASS: TestAzureManagedMachinePoolReconcile/Reconcile_succeed (0.01s)
    --- PASS: TestAzureManagedMachinePoolReconcile/Reconcile_delete (0.00s)
```
So could you pls help review my pr again? Thanks!

nawazkh:
I was spending some time of figuring this out too the other day, and came to conclusion that `azureManagedMachinePoolService.scope` had to be further mocked out, but did not go to the point of implementing the mocks. 
This is good work @xiujuanx ! 🚀  

Side thought, I wonder if there is anyway we can reduce the mocks created to get this test running? Can we import/reuse any existing mocks? 

Otherwise looks good to me!

xiujuanx:
Thanks @nawazkh for your kind review! 
>Side thought, I wonder if there is anyway we can reduce the mocks created to get this test running? 

As far as I know, if we want to do unit test `AzureManagedMachinePool` reconcile, then a mocked `AzureManagedMachinePoolReconciler` receiver is necessary for `Reconcile` method. So mock can't avoided I think.

>Can we import/reuse any existing mocks?

For `AzureManagedMachinePoolReconciler`, it contains below struct:
```
azureManagedMachinePoolService struct {
		scope         agentpools.AgentPoolScope
		agentPoolsSvc azure.Reconciler
		scaleSetsSvc  NodeLister
	}
```
So the `scope`, `agentPoolsSvc` and `scaleSetsSvc` are necessary to be mocked.  Mock these three variables, either we mock them via `mockgen` cli, or define a new struct  like `MockazureManagedMachinePoolService`. For the first method, mocked agentPoolScope and mocked reconciler have existed. We only need to mock the Nodelister interface via `mockgen`. 
However, for the second method, we must create some new funcs to implement the new mocked struct `MockazureManagedMachinePoolService`, it will make our test codes be more complicate and tedious I think.
That means all existing mocks are not enough for AzureManagedMachinePoolReconciler testing.

controllers/azuremanagedmachinepool_controller_test.go
```diff
@@ -0,0 +1,335 @@
...
+
+	for _, change := range changes {
+		change(&pool)
+	}
+
+	return pool
+}
+
+func fakeVirtualMachineScaleSetVM() []compute.VirtualMachineScaleSetVM {
+	virtualMachineScaleSetVm := []compute.VirtualMachineScaleSetVM{
```
willie-yao, line 293:
```suggestion
	virtualMachineScaleSetVM := []compute.VirtualMachineScaleSetVM{
```

controllers/azuremanagedmachinepool_controller_test.go
```diff
@@ -0,0 +1,335 @@
...
+			Zones:      &[]string{"zone0"},
+			VirtualMachineScaleSetVMProperties: &compute.VirtualMachineScaleSetVMProperties{
+				ProvisioningState: pointer.String(string(compute.ProvisioningState1Succeeded)),
+				OsProfile: &compute.OSProfile{
+					ComputerName: pointer.String("instance-000000"),
+				},
+			},
+		},
+	}
+	return virtualMachineScaleSetVm
```
willie-yao, line 307:
```suggestion
	return virtualMachineScaleSetVM
```

willie-yao (commented):
Great work! Just a few minor nits to satisfy the style requirements.

controllers/azuremanagedmachinepool_controller_test.go
xiujuanx: Done! Thanks @willie-yao !

jackfrancis:
/assign @mboersma to assess whether these new test cases cover the scenarios added in #2074, and thus should close #2078

I'm eager to merge these as-is as a net new sandbox for continuing to add coverage, thank you @xiujuanx!! (Just not sure if the cases would have assisted in the #2074 work spike, and thus maybe if not should we keep #2078 open?)

willie-yao (approved):
/lgtm

Great work on this! We have been needing this unit test for a while and this is a great help!

mboersma:
> whether these new test cases cover the scenarios added in #2074, and thus should close #2078

I don't think this test calls `ammpr.SetupWithManager`, so strictly speaking it doesn't cover the change in #2074.

But #2078 was more about noticing that the AMMP controller had no unit tests, and this looks like a huge improvement on that. IDK if it's worth holding #2078 open over the potential to cover that single line of code, but this PR on its own looks good to me.

willie-yao:
@xiujuanx Looks like something bugged out with the CLA authorization and you may need to re-sign it. When that's done this should be merged automatically!

PR #3198, opened 2023-02-23
Fix e2e Helm chart retry when install fails

### LINKED ISSUE #3104: Intermittent "cannot re-use a name that is still in use" in e2e when installing Azure Disk CSI driver Helm chart
**Which jobs are flaky:**
- periodic-cluster-api-provider-azure-capi-e2e-main
- periodic-cluster-api-provider-azure-e2e-main
- periodic-cluster-api-provider-azure-e2e-workload-upgrade-1-24-1-25-main

**Which tests are flaky:**
- Any that install the Azure Disk CSI driver Helm chart

**Testgrid link:**
- https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/periodic-cluster-api-provider-azure-capi-e2e-main/1618124950411939840
- https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/periodic-cluster-api-provider-azure-capi-e2e-main/1618306147695988736
- https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/periodic-cluster-api-provider-azure-e2e-main/1617218964859916288
- https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/periodic-cluster-api-provider-azure-e2e-workload-upgrade-1-24-1-25-main/1616553384792297472

/kind flake

CecileRobertMichon:
**What type of PR is this?**

/kind failing-test

Fixes #3104 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/azure_privatecluster.go
+++ b/test/e2e/azure_privatecluster.go
@@ -101,8 +101,8 @@ func AzurePrivateClusterSpec(ctx context.Context, inputGetter func() AzurePrivat
 
 	By("Ensure public API server is stable before creating private cluster")
 	Consistently(func() error {
-		kubeSystem := &corev1.Namespace{}
-		return publicClusterProxy.GetClient().Get(ctx, client.ObjectKey{Name: "kube-system"}, kubeSystem)
+		ns := &corev1.Namespace{}
+		return publicClusterProxy.GetClient().Get(ctx, client.ObjectKey{Name: kubesystem}, ns)
 	}, "5s", "100ms").Should(BeNil(), "Failed to assert public API server stability")
 
 	// **************
--- a/test/e2e/common.go
+++ b/test/e2e/common.go
@@ -47,7 +47,7 @@ import (
 	"sigs.k8s.io/cluster-api/test/framework"
 	"sigs.k8s.io/cluster-api/test/framework/clusterctl"
 	"sigs.k8s.io/cluster-api/util/kubeconfig"
-	crclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // Test suite constants for e2e config variables
@@ -225,7 +225,7 @@ func redactLogs() {
 }
 
 func createRestConfig(ctx context.Context, tmpdir, namespace, clusterName string) *rest.Config {
-	cluster := crclient.ObjectKey{
+	cluster := client.ObjectKey{
 		Namespace: namespace,
 		Name:      clusterName,
 	}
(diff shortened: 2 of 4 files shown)


CecileRobertMichon:
/hold

I need to validate that this actually works by faking an install error locally

CecileRobertMichon:
/hold cancel

This seems to work as expected with a broken install

test/e2e/helpers.go
```diff
@@ -75,7 +75,7 @@ const (
 	retryableOperationTimeout             = 30 * time.Second
 	retryableDeleteOperationTimeout       = 3 * time.Minute
 	retryableOperationSleepBetweenRetries = 3 * time.Second
-	helmInstallTimeout                    = 3 * time.Minute
+	helmInstallTimeout                    = 10 * time.Minute
```
jackfrancis, line 78: I assume we're increasing this timeout in response to recent flakes? Did you see this comment here?

- https://kubernetes.slack.com/archives/CEX9HENG7/p1677633345332789?thread_ts=1677618484.842959&cid=CEX9HENG7

I'm not sure we should be relying upon this much tolerance for helm install failures, maybe we're doing the helm installs too early?

CecileRobertMichon: /hold

I think you're right, let me try switching the order

CecileRobertMichon: /hold cancel

Switching the order won't work because the CAPI framework func checks for a node ref on the Machine which won't be there until cloud-provider is installed, but I added a check before installing the helm chart to make sure the api server is reachable and reverted the timeout increase.

test/e2e/azure_privatecluster.go
```diff
@@ -101,8 +101,8 @@ func AzurePrivateClusterSpec(ctx context.Context, inputGetter func() AzurePrivat
 
 	By("Ensure public API server is stable before creating private cluster")
 	Consistently(func() error {
-		kubeSystem := &corev1.Namespace{}
-		return publicClusterProxy.GetClient().Get(ctx, client.ObjectKey{Name: "kube-system"}, kubeSystem)
+		ns := &corev1.Namespace{}
```
jackfrancis, line 104: nice

test/e2e/common.go
```diff
@@ -257,9 +259,25 @@ func EnsureControlPlaneInitialized(ctx context.Context, input clusterctl.ApplyCl
 		Namespace: cluster.Spec.ControlPlaneRef.Namespace,
 		Name:      cluster.Spec.ControlPlaneRef.Name,
 	}
+
+	By("Ensuring KubeadmControlPlane is initialized")
+	Eventually(func() error {
+		if err := getter.Get(ctx, key, kubeadmControlPlane); err != nil {
+			return err
```
jackfrancis, line 266: I kinda think this is a candidate for passing in Gomega into the anonymous func inside `Eventually` so that we can do assertions against the errs and have more meaningful diagnostic stdout.

test/e2e/helpers.go
```diff
@@ -973,6 +973,16 @@ func InstallHelmChart(ctx context.Context, clusterProxy framework.ClusterProxy,
 			chartRequested, err := helmLoader.Load(cp)
 			g.Expect(err).NotTo(HaveOccurred())
 			release, err := i.RunWithContext(ctx, chartRequested, vals)
+			if err != nil {
+				Logf("Failed to install release %s, attempting to cleanup so we can retry", releaseName)
+				// Best effort attempt to delete the failed release so we can retry.
+				_, delErr := helmAction.NewUninstall(actionConfig).Run(releaseName)
+				if delErr != nil {
```
jackfrancis, line 980: ditto here about Gomega, I think a simple `g.Expect(delErr).NotTo(HaveOccurred())` will give us the equivalent amount of specific messaging that manually logging an error does, with additional data about what's in the `delErr`

PR #3105, opened 2023-01-27
[WIP] Switch flavor and test templates to external cloud-provider

CecileRobertMichon:
**What type of PR is this?**

**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note

```

PR #3213, opened 2023-03-02
more explicit delete in ci-entrypoint

jackfrancis:
**What type of PR is this?**

/kind failing-test

This PR adds more explicit handling mgmt (kind) and workload cluster operations in ci-entrypoint.sh so that we can more reliably clean up after ourselves following runs. Specifically:

- Persists a kubeconfig file in a well-known location so that downstream scripts can explicitly reference it, rather than rely upon their implicit kubeconfig context pointing at the right place.
- A more specific consumption of the `SKIP_CLEANUP` variable: we only skip the cleanup phase if it is explicitly set to the "true" value.
- A more consistent dispatching of the `KIND_CLUSTER_NAME` environment variable upstream and downstream of ci-entrypoint, so that can more explicitly track, target, and delete the management cluster.
- Remove the `timeout` wrapping around the `cluster delete` operation, and display a message if the delete operation fails
- Remove the `cleanup` func and instead include the two cleanup statements (delete workload cluster and delete management cluster) in the flow of the trapped `on_exit` func
  - this is due to observed edge cases in my macos environment and speculation that perhaps this is not a fully supported bash runtime composition pattern across all implementations of bash

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/scripts/ci-entrypoint.sh
+++ b/scripts/ci-entrypoint.sh
@@ -29,6 +29,8 @@ HELM="${REPO_ROOT}/hack/tools/bin/helm"
 KIND="${REPO_ROOT}/hack/tools/bin/kind"
 KUSTOMIZE="${REPO_ROOT}/hack/tools/bin/kustomize"
 make --directory="${REPO_ROOT}" "${KUBECTL##*/}" "${HELM##*/}" "${KIND##*/}" "${KUSTOMIZE##*/}"
+KIND_CLUSTER_NAME="${KIND_CLUSTER_NAME:-capz}"
+export KIND_CLUSTER_NAME
 # export the variables so they are available in bash -c wait_for_nodes below
 export KUBECTL
 export HELM
@@ -129,6 +131,10 @@ select_cluster_template() {
 
 create_cluster() {
     "${REPO_ROOT}/hack/create-dev-cluster.sh"
+    if [ ! -f "${REPO_ROOT}/${KIND_CLUSTER_NAME}.kubeconfig" ]; then
+        echo "Unable to find kubeconfig for kind mgmt cluster ${KIND_CLUSTER_NAME}"
+        exit 1
+    fi
 }
 
 # get_cidrs derives the CIDR from the Cluster's '.spec.clusterNetwork.pods.cidrBlocks' metadata
@@ -279,20 +285,12 @@ install_addons() {
 
 copy_secret() {
     # point at the management cluster
-    unset KUBECONFIG
-    "${KUBECTL}" get secret "${CLUSTER_NAME}-control-plane-azure-json" -o jsonpath='{.data.control-plane-azure\.json}' | base64 --decode >azure_json
+    "${KUBECTL}" --kubeconfig "${REPO_ROOT}/${KIND_CLUSTER_NAME}.kubeconfig" get secret "${CLUSTER_NAME}-control-plane-azure-json" -o jsonpath='{.data.control-plane-azure\.json}' | base64 --decode >azure_json
 
-    # set KUBECONFIG back to the workload cluster
-    export KUBECONFIG="${KUBECONFIG:-${PWD}/kubeconfig}"
+    # create the secret on the workload cluster
     "${KUBECTL}" create secret generic "${CONFIG_SECRET_NAME}" -n kube-system \
         --from-file=cloud-config=azure_json
     rm azure_json
-}
-
-# cleanup all resources we use
-cleanup() {
-    timeout 1800 "${KUBECTL}" delete cluster "${CLUSTER_NAME}" || true
-    make kind-reset || true
 }
 
 on_exit() {
@@ -305,9 +303,10 @@ on_exit() {
     unset KUBECONFIG
     go run -tags e2e "${REPO_ROOT}"/test/logger.go --name "${CLUSTER_NAME}" --namespace default
     "${REPO_ROOT}/hack/log/redact.sh" || true
-    # cleanup
-    if [[ -z "${SKIP_CLEANUP:-}" ]]; then
-        cleanup
+    # cleanup all resources we use
+    if [[ ! "${SKIP_CLEANUP:-}" == "true" ]]; then
+        timeout 1800 "${KUBECTL}" --kubeconfig "${REPO_ROOT}/${KIND_CLUSTER_NAME}.kubeconfig" delete cluster "${CLUSTER_NAME}" || echo "Unable to delete cluster ${CLUSTER_NAME}"
+        make --directory="${REPO_ROOT}" kind-reset || true
     fi
 }
 
--- a/scripts/kind-with-registry.sh
+++ b/scripts/kind-with-registry.sh
@@ -25,6 +25,7 @@ make --directory="${REPO_ROOT}" "${KUBECTL##*/}" "${KIND##*/}"
 
 # desired cluster name; default is "kind"
 KIND_CLUSTER_NAME="${KIND_CLUSTER_NAME:-capz}"
+export KIND_CLUSTER_NAME
 
 if [[ "$("${KIND}" get clusters)" =~ .*"${KIND_CLUSTER_NAME}".* ]]; then
   echo "cluster already exists, moving on"
@@ -54,9 +55,11 @@ EOF
 # (the network may already be connected)
 docker network connect "kind" "${reg_name}" || true
 
+"${KIND}" get kubeconfig -n "${KIND_CLUSTER_NAME}" > "${REPO_ROOT}/${KIND_CLUSTER_NAME}.kubeconfig"
+
 # Document the local registry
 # https://github.com/kubernetes/enhancements/tree/master/keps/sig-cluster-lifecycle/generic/1755-communicating-a-local-registry
-cat <<EOF | kubectl apply -f -
+cat <<EOF | "${KUBECTL}" --kubeconfig "${REPO_ROOT}/${KIND_CLUSTER_NAME}.kubeconfig" apply -f -
 apiVersion: v1
 kind: ConfigMap
 metadata:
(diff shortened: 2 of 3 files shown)


sonasingh46 (approved):
/lgtm 
PS: Have also verified by pulling in the changes.

mboersma (approved):
/lgtm
/approve

scripts/ci-entrypoint.sh
```diff
@@ -279,22 +285,14 @@ install_addons() {
 
 copy_secret() {
     # point at the management cluster
-    unset KUBECONFIG
-    "${KUBECTL}" get secret "${CLUSTER_NAME}-control-plane-azure-json" -o jsonpath='{.data.control-plane-azure\.json}' | base64 --decode >azure_json
+    "${KUBECTL}" --kubeconfig "${REPO_ROOT}/${KIND_CLUSTER_NAME}.kubeconfig" get secret "${CLUSTER_NAME}-control-plane-azure-json" -o jsonpath='{.data.control-plane-azure\.json}' | base64 --decode >azure_json
 
-    # set KUBECONFIG back to the workload cluster
-    export KUBECONFIG="${KUBECONFIG:-${PWD}/kubeconfig}"
```
CecileRobertMichon, line 286: shouldn't this be replaced by adding `--kubeconfig ${KUBECONFIG:-${PWD}/kubeconfig}" on all the kubectl commands below that expect to be pointing at the workload cluster?

scripts/ci-entrypoint.sh
```diff
@@ -305,9 +303,10 @@ on_exit() {
     unset KUBECONFIG
     go run -tags e2e "${REPO_ROOT}"/test/logger.go --name "${CLUSTER_NAME}" --namespace default
     "${REPO_ROOT}/hack/log/redact.sh" || true
-    # cleanup
-    if [[ -z "${SKIP_CLEANUP:-}" ]]; then
-        cleanup
+    # cleanup all resources we use
+    if [[ ! "${SKIP_CLEANUP:-}" == "true" ]]; then
```
CecileRobertMichon, line 307: why make the skip condition more strict? If the dev defined `SKIP_CLEANUP` I think it's safe to assume they want to skip, unless maybe it's false

scripts/ci-entrypoint.sh
```diff
@@ -305,9 +303,10 @@ on_exit() {
...
+        "${KUBECTL}" --kubeconfig "${REPO_ROOT}/${KIND_CLUSTER_NAME}.kubeconfig" delete cluster "${CLUSTER_NAME}" --wait=true || echo "Unable to delete cluster ${CLUSTER_NAME} nodes"
```
CecileRobertMichon, line 308: why remove the timeout here? I think that might cause the command to hang forever / until the job times out and then we wouldn't execute the rest nor print this error echo statement

scripts/kind-with-registry.sh
```diff
@@ -54,9 +55,11 @@ EOF
 # (the network may already be connected)
 docker network connect "kind" "${reg_name}" || true
 
+kind get kubeconfig -n "${KIND_CLUSTER_NAME}" > "${REPO_ROOT}/${KIND_CLUSTER_NAME}.kubeconfig"
```
CecileRobertMichon, line 58:
```suggestion
"${KIND}" get kubeconfig -n "${KIND_CLUSTER_NAME}" > "${REPO_ROOT}/${KIND_CLUSTER_NAME}.kubeconfig"
```

scripts/ci-entrypoint.sh
CecileRobertMichon: is `--wait true` actually doing anything? or is that just making it explicit

CecileRobertMichon, line 308:
```suggestion
        "${KUBECTL}" --kubeconfig "${REPO_ROOT}/${KIND_CLUSTER_NAME}.kubeconfig" delete cluster "${CLUSTER_NAME}" --wait=true || echo "Unable to delete cluster ${CLUSTER_NAME}"
```

we're deleting the entire cluster, not sure the word nodes is needed here

scripts/ci-entrypoint.sh
jackfrancis: So what's happening is that in the current implementation we're assuming that the default kubeconfig context is pointing at the kind mgmt cluster. Then we explicitly set `KUBECONFIG` to point to the workload cluster in `install_addons`. Later in that flow we want to point back to the mgmt cluster to grab the workload cluster's kubeconfig secret. Now that we have an explicit reference to the mgmt cluster (via `${REPO_ROOT}/${KIND_CLUSTER_NAME}.kubeconfig`) we can do that explicitly (see the changed L288 above), and we don't have to `unset` then `export` again.

scripts/ci-entrypoint.sh
jackfrancis: It's that latter thing. The obvious choices seem to be:

- Make `true` explicitly meaningful
- Make `false` and "" (empty string) and null meaningful.

The former is (IMO) easier to manage. If someone is passing in `SKIP_CLEANUP=foo` and assuming that's part of the contract, I think it's O.K. to regress that existing usage (because it doesn't make sense). On the other hand, for anyone passing in`SKIP_CLEANUP=false`, we may be finally doing the right thing.

Open to other thoughts here, but I think the `-z` is definitely the wrong thing here, as (IME) there is no obvious way to create a "boolean env var" in bash in the way that `-z` would make sense.

scripts/ci-entrypoint.sh
jackfrancis: Adding a timeout suggests that a successful operation lower than that threshold is an intrisic part of the test CI contract. A timeout combined with `|| true` is (IMO) the worst of both worlds, because we are enforcing a timeout but then ignoring its value.

The reason to remove the timeout would be to defer to the overall CI job timeout to enforce maximum long-running operation time.

Alternatively, we could pass `--wait=false`, and then add a sensible sleep statement (1 minute) to allow the capz controller to begin to set of delete operations w/ Azure before tearing down the mgmt cluster.

jackfrancis: IME `--wait=true` is the default, so yes just making this more explicit (and I could be wrong about that)

jackfrancis: ACK

scripts/ci-entrypoint.sh
CecileRobertMichon: fair, my opinion is loosely held

scripts/ci-entrypoint.sh
CecileRobertMichon: I am still skeptical on this one, IME the kubectl delete --wait true function will not "fail" unless the delete webhook rejects the delete (for example the cluster we're trying to delete doesn't exist), what will actually happen here is the delete command will just hang forever if the delete is unsuccessful and it will never reach the `||` command.

The way I understand the current code is "try to wait this amount of time for the cluster to be deleted but move on if that is unsuccessful". 

> The reason to remove the timeout would be to defer to the overall CI job timeout to enforce maximum long-running operation time.

I don't see this as beneficial because then we wouldn't give the job a chance to print out any useful information about why the delete failed and perform any follow-up steps like deleting the kind cluster, and the delete would take potentially up to 4 hours (the timeout we set for most conformance jobs) as opposed to the current 30 minutes. We expect a conformance test suite to take up to 4 hours, so it's an appropriate timeout, but a delete that takes more than 30 minutes is very likely to never succeed, so there's little use in hogging compute resources for those extra 3.5 hours.

jackfrancis: Is the delete command essentially waiting for a very specific status phase transition in the `Cluster` resource? If there was a terminal issue preventing the delete command from completing we would see that in the logs, no? And we should also get a failed test run if the overall timeout allotted is exceeded. I think that's the desired outcome.

Having a concrete timeout around the delete operation would also give us the desired non-zero test result (and thus encourage us to investigate what happened in the logs), do we want to carry over the 30 minute timeout (via `test/e2e/config/azure-dev.yaml`) that we currently require for a fully passing test outcome on CAPZ E2E for ci-entrypoint-dependent test jobs?

CecileRobertMichon: > Is the delete command essentially waiting for a very specific status phase transition in the Cluster resource?

not exactly, it's waiting for the object to no longer exist, after k8s has garbage collected the object (which happens once the CAPI controller removes the finalizer).

> And we should also get a failed test run if the overall timeout allotted is exceeded. I think that's the desired outcome.

If the desired outcome is to fail the run on delete, then we could just remove the `|| true` part. We don't need to wait 4 hours to determine whether the delete really failed. I think there was a good reason for not doing it this way at the time when this was implemented, which is that this script is _not_ meant to test CAPZ itself, but rather a tool for creating test clusters to run other tests such as cloud-provider e2e, so we didn't want to fail the test run if the thing we're testing (e.g. the cloud-provider test suite ran and all tests passed) was successful to not slow down/block development of other projects.

CecileRobertMichon: one argument for _not_ doing this right now: we aren't confident that delete is working and has ever worked for CCM jobs so if we merge this with the "fail on delete failure" change and it turns out delete is not working (and this doesn't fix it) we could cause all k8s / CCM jobs to be broken until we figure it out. Maybe a good compromise for now is to just log the failure, then once we're confident delete is working as expected, we switch to a strict fail.

jackfrancis: I think that makes sense. Any objections to increasing the timeout to 30m to match CAPZ E2E, or should we just stick to 20m to decrease the chances that we'll stumble onto something new and regress downstream tests further?

CecileRobertMichon: the existing timeout of 1800s is 30 minutes already no?

jackfrancis: Dog ate my calculator

jackfrancis: Restored previous implementation w/ echo statement replacing `true` assertion in the `||` path

CecileRobertMichon:
feel free to remove the hold when ready @jackfrancis

CecileRobertMichon:
:(

/retest

jackfrancis:
/hold

I prefer that https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3105 land first!

PR #3233, opened 2023-03-07
Build CCM when using k8s version latest in ci-entrypoint

CecileRobertMichon:
**What type of PR is this?**

/kind failing-test

**What this PR does / why we need it**: Fixes regression from #3105 where CCM is not getting built for CSI driver tests that use ci-entrypoint but don't test CCM: https://testgrid.k8s.io/provider-azure-master-signal#capz-azure-disk, resulting in an invalid Helm image version as the k8s version doesn't exist.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/util.sh
+++ b/hack/util.sh
@@ -42,10 +42,16 @@ capz::util::should_build_kubernetes() {
 }
 
 capz::util::should_build_ccm() {
+    # TEST_CCM is an environment variable set by a prow job to indicate that the CCM should be built and tested.
     if [[ -n "${TEST_CCM:-}" ]]; then
         echo "true" && return
     fi
+    # If conformance is being tested with CI artifacts, CCM should be built.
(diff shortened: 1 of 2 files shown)


CecileRobertMichon:
/hold

this needs a test-infra PR with it

hack/util.sh
```diff
@@ -42,12 +42,18 @@ capz::util::should_build_kubernetes() {
 }
 
 capz::util::should_build_ccm() {
+    # TEST_CCM is an environment variable set by a prow job to indicate that the CCM should be built and tested.
```
jackfrancis, line 45: Thanks for the comments!

CecileRobertMichon:
> this needs a test-infra PR with it

/hold cancel

I'll open that PR shortly but no need to wait for it since the jobs this is fixing are already broken....

Issue #3227, opened 2023-03-06
Windows pod error: failed to setup network for sandbox


jackfrancis:
/kind bug

[Before submitting an issue, have you checked the [Troubleshooting Guide](https://capz.sigs.k8s.io/topics/troubleshooting.html)?]

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]

Our VMSS tests include a step to validate a LoadBalancer that sits in front of a Windows pod. Those tests occasionally fail because that pod reports:

```
  Warning  FailedCreatePodSandBox  24m                  kubelet            Failed to create pod sandbox: rpc error: code = Unknown desc = failed to setup network for sandbox "6587e3b5ab2f01bbc40aef99de40539a4b7c77e9579efcdec323a404ce7baafc": plugin type="calico" failed (add): global strict affinity should not be false for Windows node
  Warning  FailedCreatePodSandBox  63s (x103 over 24m)  kubelet            (combined from similar events): Failed to create pod sandbox: rpc error: code = Unknown desc = failed to setup network for sandbox "1794612b2e974afa3866f41153f7e104c0fbc9bfd16951c00ac77fd8c703e9a2": plugin type="calico" failed (add): global strict affinity should not be false for Windows node
```

Node metadata for the Windows node that the pod was scheduled onto does not report anything that suggests root cause:

https://storage.googleapis.com/kubernetes-jenkins/logs/periodic-cluster-api-provider-azure-e2e-main/1631536507762249728/artifacts/clusters/capz-e2e-0wup7c-vmss/nodes/win-p-win000001/node-describe.txt

**What did you expect to happen:**

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- cluster-api-provider-azure version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

marosset:
@mikezappa87

MikeZappa87:
We should involve someone from Tigera to look at this.

jackfrancis:
Would calico logs from any particular pods be helpful?

jackfrancis:
Another repro: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/3213/pull-cluster-api-provider-azure-e2e/1632900919190360064

jsturtevant:
[...]

PR #3242, opened 2023-03-08
upgrade golangci-lint

mweibel:
**What type of PR is this?**
/kind cleanup

I couldn't run golangci-lint locally due to memory/cpu spikes. In version 1.51 they improved that a lot (by also building with go 1.20). This upgrades the linter.

I had to rename the build tag to `capztools` because when running with `tools` build tags there's an issue with another package:
```
$ go install -tags tools github.com/golangci/golangci-lint/cmd/golangci-lint@v1.51.2
../../../go/pkg/mod/github.com/ryancurrah/gomodguard@v1.3.0/tools.go:5:8: import "github.com/t-yuki/gocover-cobertura" is a program, not an importable package
```

I'm not 100% certain why since my Go module know-how is still not that great unfortunately. I suspect that module also uses `tools` build tag for something and is imported by golangci-lint.

Additionally, there are a couple of linting issues after upgrading:
```
azure/regional_baseuri.go:27:6: type `aliasAuth` is unused (unused)
type aliasAuth = Authorizer
     ^
azure/services/virtualnetworks/virtualnetworks_test.go:46:2: var `managedVnet` is unused (unused)
        managedVnet = network.VirtualNetwork{
        ^
azure/services/privatedns/privatedns_test.go:92:2: var `unmanagedTags` is unused (unused)
        unmanagedTags = resources.TagsResource{
        ^
```

I cleaned them up. `aliasAuth` is a false positive unfortunately.

**Release note**:

```release-note
NONE
```


mboersma (commented):
/lgtm
/assign @CecileRobertMichon @jackfrancis

mweibel:
Required `unknown` is a bit weird - is that not defined for `pull-cluster-api-provider-azure-e2e`?

mweibel:
hmm is `pull-cluster-api-provider-azure-e2e` flaky?

azure/regional_baseuri.go
```diff
@@ -24,7 +24,7 @@ import (
 	"github.com/pkg/errors"
 )
 
-type aliasAuth = Authorizer
+type aliasAuth = Authorizer //nolint:unused // This is a false positive since aliasAuth is used right below.
```
nawazkh, line 27: I am a little confused here, Why can't we use the below here?
```suggestion
type aliasAuth Authorizer
```

mweibel: The relevant check reports this as unused even though it’s used (false positive). I briefly searched for issues related to it in the staticcheck repo and found a few with the mention that they‘ll work on it.

CecileRobertMichon:
Failed with the kube-proxy image pull error again, that should be fixed now...

/retest

azure/regional_baseuri.go
nawazkh: Sorry, I did not frame the question right. I was wondering why is `type aliasAuth = Authorizer` and not `type aliasAuth Authorizer`. Looks like it doesn't really matter since `aliasAuth` also turns out to be an _interface_ type in both the scenarios.

mweibel:
FYI @mboersma @CecileRobertMichon tests passed here but LGTM has been removed since I rebased on top of latest main.

azure/regional_baseuri.go
mweibel: that's a good question. I tried adjusting the code according to your suggestion and lint doesn't complain anymore (of course also removed the nolint comment) and tests seem to pass too. 

I wonder what the original intention was of adding that type alias though? The [difference is explained quite well on this page](https://yourbasic.org/golang/type-alias/) but I wonder why it's needed at anyway? Why not just work with `Authorizer` itself?

CecileRobertMichon: should we try just removing it then? possibly it was needed at some point when it was added and that reason no longer exists.

mweibel: I tried to remove it and that led me to a way too long investigation why it doesn't work. 
Summary: the type alias is needed because type `Authorizer` also has a method named `Authorizer()` (same name) and the compiler & linter gets confused when type `baseURIAdapter` wants to embed the interface `Authorizer`

more or less minimum reproducing code: https://go.dev/play/p/nwjcUf9x4DI
how it would work without type alias (rename Authorizer method to Authorizor e.g.): https://go.dev/play/p/A5kCr0FrBPE

I think we can just settle to keep a type alias though. I'll use the other form @nawazkh proposed even though it has a slightly different semantic meaning.

mweibel: code is updated & commented. PTAL @CecileRobertMichon @nawazkh

mboersma (approved):
/lgtm
/approve

CecileRobertMichon:
ran into #3227 again...

/retest

CecileRobertMichon:
going to cherry-pick this because kpromo new version runs into the tools issue... if the bot lets me

/cherry-pick release-1.8 release-1.7

CecileRobertMichon:
ok bot

/cherry-pick release-1.8 
/cherry-pick release-1.7

PR #3188, opened 2023-02-17
Refactor scalesets NIC config

CecileRobertMichon:
**What type of PR is this?**

/kind cleanup

**What this PR does / why we need it**: When #2411 merged it added some duplicate/dead code in the VMSS spec for network interfaces. #3243 fixed some functional issues, this PR cleans up the code to remove dead code.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #3224 

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

PR #3243, opened 2023-03-08
fix: networkInterfaces IPForwarding and AcceleratedNetworking

### LINKED ISSUE #3224: MachinePools/VMSS NetworkInterfaces wrong configuration
/kind bug

We have a bunch of MachinePools with separate subnets. The corresponding code for setting up the NetworkInterfaces changed substantially in #2411. 
The code in question breaks CNI due to not setting enableIPForwarding if we have a networkInterface specified but can't use the default due to a separate subnetName:
```
kind: AzureMachinePool
metadata:
  name: REDACTED
  namespace: REDACTED
spec:
  template:
    networkInterfaces:
    - subnetName: REDACTED-subnet
```

Will result in the following networkInterface:
```
      "networkProfile": {"networkInterfaceConfigurations":[{"name":"REDACTED","properties":{"primary":true,"enableAcceleratedNetworking":false,"disableTcpStateTracking":false,"dnsSettings":{"dnsServers":[]},"enableIPForwarding":false,"ipConfigurations":[{"name":"private-ipConfig-0","properties":{"primary":true,"subnet":{"id":"/subscriptions/REDACTED/resourceGroups/REDACTED/providers/Microsoft.Network/virtualNetworks/REDACTED-vnet/subnets/REDACTED-subnet"},"privateIPAddressVersion":"IPv4"}}]}}]},
```

On a cluster built created with the old code, this is set to true. Also notice the `enableAcceleratedNetworking` is not set to true, even if the instance would support it and the docs state that it should be set to true if the VM type supports it. 

This probably breaks the CNI (we just built a fixed version of the controller to verify).

EnableIPForwarding set to true
AcceleratedNetworking set to true

The [code in question](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/services/scalesets/scalesets.go#L561) duplicates a lot of code from [code right above](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/services/scalesets/scalesets.go#L511) and misses setting a few things. 
I guess the code would need to be refactored to better match what is done for other machines and unified as to not have two code parts working on the same logic.

**Environment:**

mweibel:
**What type of PR is this?**
/kind bug

Enables IP Forwarding and AcceleratedNetworking when specifying networkInterfaces

Fixes #3224

I opened this as a quick fix if #3188 takes longer to complete.

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:
```release-note
fix networking when custom Network Interfaces are specified in MachinePools
```


azure/services/scalesets/scalesets.go
```diff
@@ -564,11 +564,15 @@ func (s *Service) buildVMSSFromSpec(ctx context.Context, vmssSpec azure.ScaleSet
 			nicConfig := compute.VirtualMachineScaleSetNetworkConfiguration{}
 			nicConfig.VirtualMachineScaleSetNetworkConfigurationProperties = &compute.VirtualMachineScaleSetNetworkConfigurationProperties{}
 			nicConfig.Name = pointer.String(vmssSpec.Name + "-" + strconv.Itoa(i))
-			if pointer.BoolDeref(n.AcceleratedNetworking, false) {
```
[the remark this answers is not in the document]
mweibel: thanks for pointing it out - a bit hard to read if you're not used to that output otherwise :)

fixed it and code is slightly easier to read anyway.

azure/services/scalesets/scalesets.go
```diff
@@ -564,11 +564,14 @@ func (s *Service) buildVMSSFromSpec(ctx context.Context, vmssSpec azure.ScaleSet
...
-				nicConfig.VirtualMachineScaleSetNetworkConfigurationProperties.EnableAcceleratedNetworking = pointer.Bool(true)
+			nicConfig.EnableIPForwarding = pointer.Bool(true)
+
+			if n.AcceleratedNetworking == nil {
+				nicConfig.VirtualMachineScaleSetNetworkConfigurationProperties.EnableAcceleratedNetworking = vmssSpec.AcceleratedNetworking
```
CecileRobertMichon, line 570: I'm not sure we should be defaulting the VMMS spec field in the controller since it's deprecated, and we have webhook logic in place to migrate the value of the old field to the new one https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/exp/api/v1beta1/azuremachinepool_default.go#L140 

maybe we're missing defaulting logic in the webhook?

azure/services/scalesets/scalesets.go
```diff
@@ -564,11 +564,14 @@ func (s *Service) buildVMSSFromSpec(ctx context.Context, vmssSpec azure.ScaleSet
 			nicConfig := compute.VirtualMachineScaleSetNetworkConfiguration{}
 			nicConfig.VirtualMachineScaleSetNetworkConfigurationProperties = &compute.VirtualMachineScaleSetNetworkConfigurationProperties{}
 			nicConfig.Name = pointer.String(vmssSpec.Name + "-" + strconv.Itoa(i))
-			if pointer.BoolDeref(n.AcceleratedNetworking, false) {
-				nicConfig.VirtualMachineScaleSetNetworkConfigurationProperties.EnableAcceleratedNetworking = pointer.Bool(true)
+			nicConfig.EnableIPForwarding = pointer.Bool(true)
```
CecileRobertMichon, line 567: why is this being harcoded to true?

azure/services/scalesets/scalesets.go
```diff
@@ -564,11 +564,14 @@ func (s *Service) buildVMSSFromSpec(ctx context.Context, vmssSpec azure.ScaleSet
...
+
+			if n.AcceleratedNetworking == nil {
+				nicConfig.VirtualMachineScaleSetNetworkConfigurationProperties.EnableAcceleratedNetworking = vmssSpec.AcceleratedNetworking
 			} else {
-				nicConfig.VirtualMachineScaleSetNetworkConfigurationProperties.EnableAcceleratedNetworking = pointer.Bool(false)
```
CecileRobertMichon, line 570: maybe the issue is that we're setting this to false if it's unset instead of setting it to true if supported by the VM size like we do for VM NICs https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/services/networkinterfaces/spec.go#L142

azure/services/scalesets/scalesets.go
mweibel: defaulting logic might help indeed but maybe a task for the refactoring?

setting it to `vmssSpec.AcceleratedNetworking` because that's being set anyway based on the SKU here:
https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3243/files#diff-a6779b0b37d0dcdbe7dcd6f00150a22e12ac067a96269038452d27ba39349a90L449

azure/services/scalesets/scalesets.go
mweibel: the default case when no `networkInterfaces` are configured, the forwarding is hardcoded:

https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3243/files#diff-a6779b0b37d0dcdbe7dcd6f00150a22e12ac067a96269038452d27ba39349a90R517

we could also set this just to the first interface, but I'm unsure about the implications. Possibly for Linux nodes it works anyway (though that might depend on the CNI) but for Windows nodes I'm not sure if it would be required for all interfaces or not. 
Also regarding the hardcoding: there's no property on the NetworkInterfaces to set this, and if we'd add it we'd probably need a validation webhook to ensure the first interface has set it to true.

mweibel:
Opened #3276 for the failing test since a couple of my PRs seem to fail because of the same reason.

mweibel:
/retest

let's try this once more

mweibel:
tests pass :tada: 
Sorry I didn't want to mess with the reviewers - wasn't aware it removes reviewers when I re-request a review from somebody. Anyway, PTAL again :)

CecileRobertMichon (commented):
Let's merge this as-is and I will follow up on #3188 for the refactor

/lgtm
/approve

mweibel:
@CecileRobertMichon can I help with merging this on top of release-1.7 somehow?

CecileRobertMichon:
@mweibel yes, if you'd like you can manually cherry-pick the commit in a branch off of release-1.7, resolve conflicts, then open a PR for it. The bot can't do cherry-picks when there's a merge conflict.

PR #3250, opened 2023-03-09
don't reduce replicas if ext. managed MachinePool

### LINKED ISSUE #3244: MachinePool reduces replicas when externally managed
/kind bug

Adjusting the replica count of a VMSS when having an externally managed MachinePool (by e.g. `cluster-autoscaler`) should only be done by the external scaler and not by the MachinePool controller. 
We discovered today it can happen that the MachinePool controller tries to match the number of AzureMachinePoolMachines with the replicas and tries to reduce the replica count.

https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/96fccea407457f0b0c382ed48d7d338ccd33a5fb/azure/scope/machinepool.go#L346-L364

This conflicts with the autoscaler since too many (or the wrong) instances get deleted. 

I'm investigating whether we need to skip doing reduce in this code block or something else is off. 

No adjustment of replicas by the MachinePool controller.

I'll be contributing a possible fix for this soon. 

**Environment:**

- cluster-api-provider-azure version: main (~1.8.0)
- Kubernetes version: (use `kubectl version`): 1.24.11
- OS (e.g. from `/etc/os-release`): windows/linux hybrid

mweibel:
**What type of PR is this?**
/kind bug

When MachinePool is externally managed, the AzureMachinePoolController reconciles after each reconciliation loop also the AzureMachinePoolMachines. At this point the `MachinePool.spec.Replicas` might not yet be up to date when the external scaler has right before, or during the reconcile loop, scaled down the VMSS by deleting a VMSSVM. At this point the controller notices an over-provision count > 0 and decides to delete another VMSSVM which often is not the same as the autoscaler is deleting. 

In general if an external autoscaler is in place for a MachinePool, we should avoid deleting any machines since the autoscaler is taking care of that. 

Fixes #3244

I considered permitting to delete failed machines by doing the externally managed replicas check later but decided it's better to leave this fully up to the autoscaler.
The added unit tests don't cover all of the Deletion possibilities so might want to be extended later on. I think it's good enough for now to have the two checks I added for the difference in behaviour we introduce with this change. 

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:
```release-note
fixes scaling down behavior when MachinePool is externally managed
```

### DIFF
--- a/azure/scope/machinepool.go
+++ b/azure/scope/machinepool.go
@@ -324,7 +324,7 @@ func (m *MachinePoolScope) applyAzureMachinePoolMachines(ctx context.Context) er
 			log.V(4).Info("deleting AzureMachinePoolMachine because it no longer exists in the VMSS", "providerID", key)
 			delete(existingMachinesByProviderID, key)
 			if err := m.client.Delete(ctx, &machine); err != nil {
-				return errors.Wrap(err, "failed deleting AzureMachinePoolMachine to reduce replica count")
+				return errors.Wrap(err, "failed deleting AzureMachinePoolMachine no longer existing in Azure")
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


CecileRobertMichon (commented):
/lgtm
/assign @mboersma

mweibel:
FYI @mboersma @CecileRobertMichon tests passed here but LGTM has been removed since I rebased on top of latest main.

jackfrancis:
@mweibel thank you so much for this fix! Let me know if you need help manually cherry-picking into `release-1.7`, I'd be happy to do so but didn't want to step on your toes.

PR #3249, opened 2023-03-09
[WIP] Fix delete for VMSS flex

CecileRobertMichon:
**What type of PR is this?**

**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Fix delete for VMSS flex
```

PR #3256, opened 2023-03-10
Fix delete for VMSS flex

### LINKED ISSUE #3077: CAPZ fails to scale in K8s VMSS Flex cluster

mboersma:
**What type of PR is this?**

/kind failing-test

Fixes a bug parsing a VMSS Flex resourceID that led to problems scaling in a MachinePool. Also adds unit test cases that will fail if the code reverts and enables scaling down in the relevant e2e test.

Fixes #3077

All credit goes to @CecileRobertMichon for tracking down the root cause of this issue. 🥇 

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Fix delete for VMSS flex
```

### DIFF
--- a/azure/services/scalesetvms/scalesetvms.go
+++ b/azure/services/scalesetvms/scalesetvms.go
@@ -151,11 +151,7 @@ func (s *Service) deleteVMSSFlexVM(ctx context.Context, resourceID string) error
 	if err != nil {
 		return errors.Wrap(err, fmt.Sprintf("failed to parse resource id %q", resourceID))
 	}
-
-	resourceGroup := parsed.ResourceGroup
-	resourceName := strings.TrimPrefix(s.Scope.ProviderID(), azure.ProviderIDPrefix)
-	resourceNameSplits := strings.Split(resourceName, "/")
-	resourceName = resourceNameSplits[len(resourceNameSplits)-3] + "_" + resourceNameSplits[len(resourceNameSplits)-1]
+	resourceGroup, resourceName := parsed.ResourceGroup, parsed.ResourceName
 
 	log.V(4).Info("entering delete")
 	future := s.Scope.GetLongRunningOperationState(resourceName, serviceName, infrav1.DeleteFuture)
--- a/azure/services/scalesetvms/scalesetvms_test.go
+++ b/azure/services/scalesetvms/scalesetvms_test.go
@@ -18,6 +18,7 @@ package scalesetvms
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 	"testing"
 	"time"
@@ -35,6 +36,7 @@ import (
 	"sigs.k8s.io/cluster-api-provider-azure/azure/converters"
 	"sigs.k8s.io/cluster-api-provider-azure/azure/scope"
 	"sigs.k8s.io/cluster-api-provider-azure/azure/services/scalesetvms/mock_scalesetvms"
+	"sigs.k8s.io/cluster-api-provider-azure/azure/services/virtualmachines/mock_virtualmachines"
 	infrav1exp "sigs.k8s.io/cluster-api-provider-azure/exp/api/v1beta1"
 	gomock2 "sigs.k8s.io/cluster-api-provider-azure/internal/test/matchers/gomock"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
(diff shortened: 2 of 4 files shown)


azure/services/scalesetvms/scalesetvms.go
```diff
@@ -151,11 +151,7 @@ func (s *Service) deleteVMSSFlexVM(ctx context.Context, resourceID string) error
 	if err != nil {
 		return errors.Wrap(err, fmt.Sprintf("failed to parse resource id %q", resourceID))
 	}
-
-	resourceGroup := parsed.ResourceGroup
-	resourceName := strings.TrimPrefix(s.Scope.ProviderID(), azure.ProviderIDPrefix)
-	resourceNameSplits := strings.Split(resourceName, "/")
-	resourceName = resourceNameSplits[len(resourceNameSplits)-3] + "_" + resourceNameSplits[len(resourceNameSplits)-1]
+	resourceGroup, resourceName := parsed.ResourceGroup, parsed.ResourceName
```
jackfrancis, line 154: So the response from the autorest library is smart enough to know about flex VM vs vanilla VM? I guess because you're passing in the resource ID we would expect it to be authoritative for parsing the various sub-components that represent the variety of resources across Azure.

CecileRobertMichon: I think it wasn't before but it now is since VMSS Flex IDs changed? @mboersma please correct me if I'm wrong

CecileRobertMichon:
/retest

the flex MP failed to provision

CecileRobertMichon:
/retest

e2e should be fixed now

mboersma:
> e2e should be fixed now

Sigh.

CecileRobertMichon:
/retest

this time for real

jackfrancis:
flex job failed (hasn't reported), but you can start investigating now @mboersma to see if this indeed didn't fix the issue:

- https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/3256/pull-cluster-api-provider-azure-e2e-optional/1635368788197117952

CecileRobertMichon:
it's weird because the flex e2e test passed on my WIP PR for the same fix https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3249

So either something changed since then or the autorest parse function isn't working as we expected

mboersma:
I've run this flex e2e spec locally, and tested scale in to 0 on Windows clusters manually through Tilt several times, for both Uniform and Flexible, and I don't see a problem.

I think the issue was this PR referenced k8s v1.26.2 which had some Windows cloud-init issues we're working to fix. I moved it back to v1.26.1.

CecileRobertMichon (commented):
/lgtm
/approve

CecileRobertMichon:
/retest

do we have an open flaky test issue for ClusterClass?

mboersma:
> do we have an open flaky test issue for ClusterClass?

I don't think so, but I'm repro'ing and investigating now and I'll add one.

CecileRobertMichon:
@mboersma I think I have a pretty decent idea of root cause (see comments in https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/3313) and there are exact repros of the same error in main (https://storage.googleapis.com/kubernetes-jenkins/logs/periodic-cluster-api-provider-azure-e2e-main/1636248867903115264/artifacts/clusters/bootstrap/resources/capz-e2e-851wjq/KubeadmControlPlane/capz-e2e-851wjq-cc-rtlzw.yaml) so I'm tempted to override the optional test if the next run fails again

/retest

PR #3265, opened 2023-03-10
Add support for confidential VMs

### LINKED ISSUE #3264: Add support for Confidential VMs and Trusted launch for VMs
/kind feature

I would like to be able to spawn [Azure Confidential VMs](https://learn.microsoft.com/en-us/azure/confidential-computing/confidential-vm-overview) and to enabled [Trusted launch for VMs](https://learn.microsoft.com/en-us/azure/virtual-machines/trusted-launch). 

Confidential VMs provide protection to data-in-use by encryption the VM memory with keys that Azure doesn't have access to. Those VMs can also be [attested](https://learn.microsoft.com/en-us/azure/attestation/overview) by the user. 

See [Azure Confidential Computing](https://learn.microsoft.com/en-us/azure/confidential-computing/overview) for more details. 

Trusted launch for VMs provide [SecureBoot](https://learn.microsoft.com/en-us/windows-hardware/design/device-experiences/oem-secure-boot) and [vTPM](https://learn.microsoft.com/en-us/azure/virtual-machines/trusted-launch#vtpm) attestation.

Confidential VMs require specific [OS images](https://learn.microsoft.com/en-us/azure/confidential-computing/confidential-vm-overview#os-support).
Trusted launch for VMs include [these limitations](https://learn.microsoft.com/en-us/azure/virtual-machines/trusted-launch#limitations).

**Environment:**

mresvanis:
**What type of PR is this?**

/kind feature

This change adds support for [Confidential VMs](https://learn.microsoft.com/en-us/azure/confidential-computing/confidential-vm-overview) and [Trusted launch](https://learn.microsoft.com/en-us/azure/virtual-machines/trusted-launch) for VMs.

Azure Confidential VMs are defined by their [SecurityProfile.SecurityType](https://learn.microsoft.com/en-us/rest/api/compute/virtual-machines/create-or-update?tabs=HTTP#securitytypes) `ConfidentialVM`, which should be defined along with the [OSDisk.ManagedDisk.SecurityProfile.SecurityEncryptionType](https://learn.microsoft.com/en-us/rest/api/compute/virtual-machines/create-or-update?tabs=HTTP#securityencryptiontypes) field.  Trusted launch for VMs is defined by the [SecurityProfile.SecurityType](https://learn.microsoft.com/en-us/rest/api/compute/virtual-machines/create-or-update?tabs=HTTP#securitytypes) `TrustedLaunch`, which should be defined along with the [SecurityProfile.UefiSettings](https://learn.microsoft.com/en-us/rest/api/compute/virtual-machines/create-or-update?tabs=HTTP#uefisettings) section, i.e. the `SecureBootEnabled` and `VTpmEnabled` fields.

Related image-builder [PR](https://github.com/kubernetes-sigs/image-builder/pull/1148).

Fixes #3264

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Add SecurityProfile SecurityType to API
Add OSDisk VMDiskSecurityProfile to API
Add SecurityProfile UefiSettings with SecureBootEnabled and VTpmEnabled to API
```


nawazkh:
I started looking into this PR, will take sometime in going over the Azure Docs and the new fields being added. 
Thanks for all the work!

CecileRobertMichon:
@nawazkh are you still reviewing this?

nawazkh:
Extremely sorry, It slipped my radar after I was half way done reviewing this PR. I will get started on it again! 
Sorry @mresvanis !

api/v1beta1/azuremachine_webhook_test.go
```diff
@@ -145,6 +145,21 @@ func TestAzureMachine_ValidateCreate(t *testing.T) {
...
+		{
+			name:    "azuremachine with valid security profile with confidential VM properties enabled",
+			machine: createMachineWithConfidentialCompute(false, SecurityEncryptionTypeVMGuestStateOnly, VirtualizedTrustedPlatformModulePolicyEnabled, SecureBootPolicyEnabled),
+			wantErr: false,
+		},
+		{
+			name:    "azuremachine with confidential compute properties and encryption at host enabled",
+			machine: createMachineWithConfidentialCompute(true, SecurityEncryptionTypeVMGuestStateOnly, "", ""),
+			wantErr: true,
+		},
```
willie-yao, line 162: Are you able to add some more test cases here that reflect the cases in `azuremachine_validation_test.go` such as with DiskWithVMGuestState encryption and vTPM disabled?

willie-yao (commented):
Thanks for all your hard work on this! I just had one comment and a general question: I think it'd be great to have an E2E test for this feature. This may be out of scope for this PR but this seems like a great feature that can be tested E2E.

api/v1beta1/azuremachine_webhook_test.go
mresvanis: Yes, it absolutely makes sense to be as comprehensive as possible here. I believe I added the test cases that were missing, but please let me know if you have thought about additional ones.

mresvanis:
> Thanks for all your hard work on this! I just had one comment and a general question: I think it'd be great to have an E2E test for this feature. This may be out of scope for this PR but this seems like a great feature that can be tested E2E.

Thank you for taking the time to review this change :) 

I agree that having an E2E test for this feature would be quite useful. In order to keep this PR at its current size, would it be an option to open an issue for the E2E test and then tackle it in a future PR? Or would you prefer to have the E2E test as part of this PR?

(I would be happy to propose a change for the E2E test either way, I would just need a bit more time)

willie-yao:
> In order to keep this PR at its current size, would it be an option to open an issue for the E2E test and then tackle it in a future PR?

Ideally, we would want to have an E2E test be a part of this PR in order to ensure the feature is fully working E2E. Since the next release is planned for May 2, it would be totally okay to take some time. The feature won't be available to users until then anyways. 

Would this be enough time for you to implement an E2E test? I apologize for the rush on this.. it's on us for not reviewing this PR sooner 😔

mresvanis:
> Would this be enough time for you to implement an E2E test? I apologize for the rush on this.. it's on us for not reviewing this PR sooner

Absolutely no problem, I'll get started as soon as I can and ping you here when I have something up for review. Thanks again for your time.

nawazkh (changes requested):
This is great effort @mresvanis , thanks for taking the time in putting this together!
Also, I apologize for sharing my review so late.

I would generally share all the review at once, but I am breaking it in chunks here to get the ball rolling. I plan on completing reviewing this PR in couple more hours.

api/v1beta1/types.go
```diff
@@ -580,6 +615,29 @@ type ManagedDiskParameters struct {
...
+	// VMGuest blob.
+	// +optional
+	DiskEncryptionSet *DiskEncryptionSetParameters `json:"diskEncryptionSet,omitempty"`
+	// SecurityEncryptionType specifies the encryption type of the managed disk.
+	// It is set to DiskWithVMGuestState to encrypt the managed disk along with the VMGuestState
+	// blob, and to VMGuestStateOnly to encrypt the VMGuestState blob only.
+	// When this field is defined, EncryptionAtHost should be disabled.
+	// When set to VMGuestStateOnly, the VirtualizedTrustedPlatformModule should be set to Enabled.
+	// When set to DiskWithVMGuestState, both SecureBoot and VirtualizedTrustedPlatformModule should be set to Enabled.
+	// It can be set only for Confidential VMs.
```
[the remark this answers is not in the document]
mresvanis: This is a great catch, I tried to follow the Azure Terraform Provider [docs](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/linux_virtual_machine#security_encryption_type) and [implementation](https://github.com/hashicorp/terraform-provider-azurerm/blob/7a3afb9d59a4cca6b1c410a3cd1a2d9db1d92d85/internal/services/compute/linux_virtual_machine_resource.go#L503). 

But I see now that the validation should be `When this field is set to DiskWithVMGuestState, EncryptionAtHost should be disabled`. I will update this PR with the respective changes.

azure/services/resourceskus/sku.go
```diff
@@ -68,6 +68,10 @@ const (
 	MaximumPlatformFaultDomainCount = "MaximumPlatformFaultDomainCount"
 	// UltraSSDAvailable identifies the capability for the support of UltraSSD data disks.
 	UltraSSDAvailable = "UltraSSDAvailable"
+	// TrustedLaunchDisabled identifies the absence of the trusted launch capability.
+	TrustedLaunchDisabled = "TrustedLaunchDisabled"
+	// ConfidentialComputingType identifies the capability for confidentical computing.
+	ConfidentialComputingType = "ConfidentialComputingType"
```
nawazkh, line 74: Jus curious about these values. `TrustedLaunchDisabled` and `ConfidentialComputingType`. Are these exposed by Azure SDK?

nawazkh (changes requested):
Part 2/2 of the review.
I have skipped reviewing _azuremachine_validation_(_.go_ and __test.go_), _azuremachine_validation_test.go_ since we are expecting for some refactor.

azure/services/resourceskus/sku.go
mresvanis: Great question, I should have mentioned something in the PR description or via a code comment. I found these values by querying the [Compute](https://learn.microsoft.com/en-us/rest/api/compute/resource-skus/list) Azure REST API.

If I understand correctly from this [discussion](https://github.com/Azure/azure-rest-api-specs/issues/10033#issuecomment-657701452), the resource SKUs list is dynamic because it's quite hard to maintain a hardcoded enum. So we don't/won't see them declared either in the Azure Go SDK or the Azure REST API specs.

Nevertheless I thought it's a useful validation to have in order to fail early enough. What do you think?

azure/services/virtualmachines/spec.go
```diff
@@ -277,18 +291,77 @@ func (s *VMSpec) generateOSProfile() (*compute.OSProfile, error) {
...
+		if s.SecurityProfile.VirtualizedTrustedPlatformModule != "" && s.SecurityProfile.VirtualizedTrustedPlatformModule == infrav1.VirtualizedTrustedPlatformModulePolicyDisabled {
+			return nil, azure.WithTerminalError(errors.New("vTPM should be enabled when SecurityEncrytionType is set"))
+		}
+
+		securityProfile.UefiSettings.VTpmEnabled = pointer.Bool(true)
+
+		securityProfile.SecurityType = compute.SecurityTypesConfidentialVM
+
+		return securityProfile, nil
+	}
```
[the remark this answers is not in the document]
mresvanis: > Would it be better if we enforce these validations at webhook level so that there are lesser chances for a user to run into Terminal Error ?

I absolutely agree, these validations are quite useful at the webhook level and we have them [here](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3265/files#diff-222a5b91649bbd8b784ec4d5476460390cc2f7e5839e63a968d719b3484ace68R395-R428). The latter is initially called through [here](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/api/v1beta1/azuremachine_webhook.go#L57).  Does this makes sense?

api/v1beta1/azuremachine_validation.go
```diff
@@ -234,6 +238,16 @@ func validateManagedDisk(m *ManagedDiskParameters, fieldPath *field.Path, isOSDi
 
 	if m != nil {
 		allErrs = append(allErrs, validateStorageAccountType(m.StorageAccountType, fieldPath.Child("StorageAccountType"), isOSDisk)...)
+
+		if isOSDisk && m.SecurityProfile != nil && m.SecurityProfile.DiskEncryptionSet != nil {
+			if m.SecurityProfile.SecurityEncryptionType == "" || m.SecurityProfile.SecurityEncryptionType != SecurityEncryptionTypeDiskWithVMGuestState {
```
alexeldeib, line 243: is the empty string check on SecurityEncryptionType necessary?

api/v1beta1/azuremachine_validation.go
```diff
@@ -378,3 +392,37 @@ func ValidateDiagnostics(diagnostics *Diagnostics, fieldPath *field.Path) field.
...
+
+	if profile != nil && securityEncryptionType != "" {
+		if profile.EncryptionAtHost != nil && *profile.EncryptionAtHost && securityEncryptionType == SecurityEncryptionTypeDiskWithVMGuestState {
+			allErrs = append(allErrs, field.Invalid(fieldPath.Child("EncryptionAtHost"), profile.EncryptionAtHost,
+				"EncryptionAtHost cannot be set to 'true' when securityEncryptionType is set to DiskWithVMGuestState"))
+		}
+
+		if profile.VirtualizedTrustedPlatformModule == "" || profile.VirtualizedTrustedPlatformModule == VirtualizedTrustedPlatformModulePolicyDisabled {
+			allErrs = append(allErrs, field.Invalid(fieldPath.Child("VirtualizedTrustedPlatformModule"), profile.VirtualizedTrustedPlatformModule,
+				fmt.Sprintf("VirtualizedTrustedPlatformModule should be set to '%s' when securityEncryptionType is defined",
```
alexeldeib, line 414: I believe this is not accurate, Azure allows you to deploy with vTPM disabled (kinda pointless though

api/v1beta1/azuremachine_validation.go
```diff
@@ -378,3 +392,37 @@ func ValidateDiagnostics(diagnostics *Diagnostics, fieldPath *field.Path) field.
...
+					VirtualizedTrustedPlatformModulePolicyEnabled)))
+		}
+
+		if securityEncryptionType == SecurityEncryptionTypeDiskWithVMGuestState {
+			if profile.SecureBoot == "" || profile.SecureBoot != SecureBootPolicyEnabled {
+				allErrs = append(allErrs, field.Invalid(fieldPath.Child("SecureBoot"), profile.SecureBoot,
+					fmt.Sprintf("SecureBoot should be set to '%s' when securityEncryptionType is set to '%s'",
```
alexeldeib, line 421: same comment as above, I'm sure about this one :)

you could argue CAPZ should enforce strong defaults but if you force secure boot and a user has custom unsigned modules, you can break workloads for no reason.

api/v1beta1/azuremachine_validation.go
```diff
@@ -378,3 +392,37 @@ func ValidateDiagnostics(diagnostics *Diagnostics, fieldPath *field.Path) field.
...
+	var securityEncryptionType SecurityEncryptionType
+
+	if managedDisk != nil && managedDisk.SecurityProfile != nil {
+		securityEncryptionType = managedDisk.SecurityProfile.SecurityEncryptionType
+	}
+
+	if profile != nil && securityEncryptionType != "" {
+		if profile.EncryptionAtHost != nil && *profile.EncryptionAtHost && securityEncryptionType == SecurityEncryptionTypeDiskWithVMGuestState {
+			allErrs = append(allErrs, field.Invalid(fieldPath.Child("EncryptionAtHost"), profile.EncryptionAtHost,
+				"EncryptionAtHost cannot be set to 'true' when securityEncryptionType is set to DiskWithVMGuestState"))
```
alexeldeib, line 409: cc @anujmaheshwari1 I wonder if we do this correctly for AKS...

api/v1beta1/azuremachine_validation.go
alexeldeib: ah, I might be mixing up CVM and TL. have you tested this out with Azure VMs?

api/v1beta1/azuremachine_validation.go
```diff
@@ -378,3 +392,37 @@ func ValidateDiagnostics(diagnostics *Diagnostics, fieldPath *field.Path) field.
...
+		}
+
+		if profile.VirtualizedTrustedPlatformModule == "" || profile.VirtualizedTrustedPlatformModule == VirtualizedTrustedPlatformModulePolicyDisabled {
+			allErrs = append(allErrs, field.Invalid(fieldPath.Child("VirtualizedTrustedPlatformModule"), profile.VirtualizedTrustedPlatformModule,
+				fmt.Sprintf("VirtualizedTrustedPlatformModule should be set to '%s' when securityEncryptionType is defined",
+					VirtualizedTrustedPlatformModulePolicyEnabled)))
+		}
+
+		if securityEncryptionType == SecurityEncryptionTypeDiskWithVMGuestState {
```
alexeldeib, line 418: are you sure DiskWithVMGuestState is what you want? I've only used CVM with VMGuestStateOnly

either way, VMGuestStateOnly should be a valid setting for CVM, I think, and works with secure boot.

api/v1beta1/azuremachine_validation.go
mresvanis: In that case, I incorrectly blindly followed the Azure Terraform Provider [docs](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/linux_virtual_machine#security_encryption_type), but I will definitely try out what you suggested and revert this change.

azure/services/virtualmachines/spec.go
```diff
@@ -277,18 +291,77 @@ func (s *VMSpec) generateOSProfile() (*compute.OSProfile, error) {
...
+		securityProfile.EncryptionAtHost = s.SecurityProfile.EncryptionAtHost
 	}
 
-	return &compute.SecurityProfile{
-		EncryptionAtHost: s.SecurityProfile.EncryptionAtHost,
-	}, nil
+	securityProfile.UefiSettings = &compute.UefiSettings{}
+
+	if s.SecurityProfile.SecureBoot != "" {
+		if s.SKU.HasCapability(resourceskus.TrustedLaunchDisabled) && s.SecurityProfile.SecureBoot == infrav1.SecureBootPolicyEnabled {
```
alexeldeib, line 342: need a gne1/gen2 check here as well. I believe you only get TrustedLaunchDisabled from SKU API on gen2 sizes which do _not_ support it, it's annoyingly inverted from most capabilities.

as-is I believe you allow a gen1 VM size with TL + secure boot which is invalid.

azure/services/virtualmachines/spec.go
```diff
@@ -277,18 +291,77 @@ func (s *VMSpec) generateOSProfile() (*compute.OSProfile, error) {
...
+		}
+
+		if s.SecurityProfile.SecureBoot == infrav1.SecureBootPolicyEnabled {
+			securityProfile.UefiSettings.SecureBootEnabled = pointer.Bool(true)
+			securityProfile.SecurityType = compute.SecurityTypesTrustedLaunch
+		}
+	}
+
+	if s.SecurityProfile.VirtualizedTrustedPlatformModule != "" {
+		if s.SKU.HasCapability(resourceskus.TrustedLaunchDisabled) &&
```
alexeldeib, line 353: nit -- calling this twice, and it's a long conditional. create a var? (other call site is the capability check earlier)

alexeldeib: or combine them?

api/v1beta1/azuremachine_validation.go
mresvanis: Great feedback, in the same incorrect sense, I blindly followed the Azure Terraform Provider [implementation](https://github.com/hashicorp/terraform-provider-azurerm/blob/7a3afb9d59a4cca6b1c410a3cd1a2d9db1d92d85/internal/services/compute/linux_virtual_machine_resource.go#L540-L546)... 

I will test it out (just for completeness sake) and revert this change as well.

alexeldeib:
apologies for drive-by review but hope I added some useful comments -- happen to have worked a little on CVM/TL from AKS perspective so was curious :)

azure/services/virtualmachines/spec.go
alexeldeib: could sanity check sku api to confirm -- 

az vm list-skus -r virtualMachines -s Something_Thats_only_gen1 (Standard_NC6?)

api/v1beta1/azuremachine_validation.go
alexeldeib: if we do confirm terraform is wrong, maybe we can open some issues :) if it was correct before, thanks for confirming and helping me learn something ;)

alexeldeib:
some other random high level thoughts
- sku api in webhooks: your intuition is correct I don't think we do that today due to requirements to fetch creds, call azure, etc. would be nice
- other early validation: you can technically regex a lot of stuff without the sku api. I think CVM _may_ work for that case, but you'd have to update the regex in the future. I got tired of manually updating sizes for AKS which is why I wrote the sku api clients to begin with. your mileage may vary :)

api/v1beta1/azuremachine_validation.go
mresvanis: Trying to create a `ConfidentialVM` with `SecurityTypeEncryption: VMGuestStateOnly | DiskWithVMGuestState` and `vTpmEnabled: false` fails with the following error:

```shell
> az vm create \
  --resource-group aResourceGroup \
  --name myConfidentialVM \
  --generate-ssh-keys \
  --size Standard_DC4as_v5 \
  --admin-username aUserame \
  --admin-password aPassword \
  --enable-vtpm false \
  --image "Canonical:0001-com-ubuntu-confidential-vm-focal:20_04-lts-cvm:latest" \
  --public-ip-sku Standard \
  --security-type ConfidentialVM \
  --os-disk-security-encryption-type VMGuestStateOnly \
  --enable-secure-boot false

ERROR: {"status":"Failed","error":{"code":"DeploymentFailed","message":"At least one resource deployment operation failed. Please list deployment operations for details. Please see https://aka.ms/arm-deployment-operations for usage details.","details":[{"code":"BadRequest","message":"{\r\n  \"error\": {\r\n    \"code\": \"BadRequest\",\r\n    \"message\": \"Confidential VM or Virtual Machine Scale Set VM with 'managedDisk.securityProfile.securityEncryptionType' set as VMGuestStateOnly should have securityProfile.uefiSettings.vtpmEnabled set to true.\"\r\n  }\r\n}"}]}}
```

```shell
> az vm create \
  --resource-group aResourceGroup \
  --name myConfidentialVM \
  --generate-ssh-keys \
  --size Standard_DC4as_v5 \
  --admin-username aUsername \
  --admin-password aPassword \
  --enable-vtpm false \
  --image "Canonical:0001-com-ubuntu-confidential-vm-focal:20_04-lts-cvm:latest" \
  --public-ip-sku Standard \
  --security-type ConfidentialVM \
  --os-disk-security-encryption-type DiskWithVMGuestState \
  --enable-secure-boot false \

ERROR: {"status":"Failed","error":{"code":"DeploymentFailed","message":"At least one resource deployment operation failed. Please list deployment operations for details. Please see https://aka.ms/arm-deployment-operations for usage details.","details":[{"code":"BadRequest","message":"{\r\n  \"error\": {\r\n    \"code\": \"BadRequest\",\r\n    \"message\": \"Confidential VM or Virtual Machine Scale Set VM with 'managedDisk.securityProfile.securityEncryptionType' set as DiskWithVMGuestState should have both values of securityProfile.uefiSettings.secureBootEnabled and securityProfile.uefiSettings.vtpmEnabled set to true.\"\r\n  }\r\n}"}]}}
```

api/v1beta1/azuremachine_validation.go
mresvanis: Trying to create a `ConfidentialVM` with `SecurityEncryptionType: DiskWithVMGuestState`, `vTpmEnabled: true` and `secureBootEnabled: false` fails with the following error:

```shell
> az vm create \
  --resource-group aResourceGroup \
  --name myConfidentialVM \
  --generate-ssh-keys \
  --size Standard_DC4as_v5 \
  --admin-username aUsername \
  --admin-password aPassword \
  --enable-vtpm true \
  --image "Canonical:0001-com-ubuntu-confidential-vm-focal:20_04-lts-cvm:latest" \
  --public-ip-sku Standard \
  --security-type ConfidentialVM \
  --os-disk-security-encryption-type DiskWithVMGuestState \
  --enable-secure-boot false \

ERROR: {"status":"Failed","error":{"code":"DeploymentFailed","message":"At least one resource deployment operation failed. Please list deployment operations for details. Please see https://aka.ms/arm-deployment-operations for usage details.","details":[{"code":"BadRequest","message":"{\r\n  \"error\": {\r\n    \"code\": \"BadRequest\",\r\n    \"message\": \"Confidential VM or Virtual Machine Scale Set VM with 'managedDisk.securityProfile.securityEncryptionType' set as DiskWithVMGuestState should have both values of securityProfile.uefiSettings.secureBootEnabled and securityProfile.uefiSettings.vtpmEnabled set to true.\"\r\n  }\r\n}"}]}}
```

api/v1beta1/azuremachine_validation.go
alexeldeib: yeah, I think you got this right for CVM. apologies for confusion :)

for clarity -- vTPM and SecureBoot I believe are both required for CVM and both optional for trusted launch.

it looks like this func is called only for securityEncryptionType != "", but is that exclusive to CVM? I haven't tried securityEncryptionType with TL myself I think?

either way I think at worst we "under validate" vs disallow valid settings, so we can always tweak later.

api/v1beta1/types.go
```diff
@@ -737,10 +795,24 @@ func (s SubnetSpec) IsNatGatewayEnabled() bool {
...
+	// +kubebuilder:validation:Enum=Enabled;Disabled
+	//+optional
+	SecureBoot SecureBootPolicy `json:"secureBoot,omitempty"`
+	// VirtualizedTrustedPlatformModule enable virtualized trusted platform module measurements to create a known good boot integrity policy baseline.
+	// The integrity policy baseline is used for comparison with measurements from subsequent VM boots to determine if anything has changed.
+	// This is required to be set to Enabled if SecurityEncryptionType is defined.
+	// If omitted, the platform chooses a default, which is subject to change over time, currently that default is Disabled.
+	// +kubebuilder:validation:Enum=Enabled;Disabled
+	// +optional
+	VirtualizedTrustedPlatformModule VirtualizedTrustedPlatformModulePolicy `json:"virtualizedTrustedPlatformModule,omitempty"`
```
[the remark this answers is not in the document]
nawazkh: I love this, thanks for such a detail explanation!
I am mostly in favor of Adhering to Azure REST API, but will also like to get opinion of @CecileRobertMichon. 

So it's a _yes_ from me to all of your questions
> 1. If there is no other objections, I will adjust the PR to the changes suggested. 
> 2. I will rename and convert these types to *bool

--- 

For 
> The Azure Go SDK defines this enum [here](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/resourcemanager/compute/armcompute/constants.go#L1685-L1698), so I would just add the same here, right?

You are right, a enum should work.

azure/services/resourceskus/sku.go
nawazkh: I am having trouble capturing the response of 
`GET https://management.azure.com/subscriptions/{subscriptionId}/providers/Microsoft.Compute/skus?api-version=2021-07-01`  as mentioned on [Compute reference docs](https://learn.microsoft.com/en-us/rest/api/compute/resource-skus/list). If you don't mind, can you please paste a response you see? I am interested in seeing `ConfidentialComputingType` in the response. 

However, I agree that these are useful validations.

azure/services/virtualmachines/spec.go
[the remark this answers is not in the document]
nawazkh: Makes sense, thanks! 

@CecileRobertMichon do you know why we perform validations in `generateSecurityProfile()` func? 
I am trying to understand the reasoning behind `generateSecurityProfile()` returning `azure.WithTerminalError()`. 
Can we not perform all the validations at webhook level ? Is it because we need `s` of type `*VMSpec` populated for these validations, and `s` can only be populated later in the reconciliation and not at the webhook level?

azure/services/resourceskus/sku.go
mresvanis: Of course, below you can find a single item with an `SNP` `ConfidentialComputingType` capability, by running `az vm list-skus --location eastus`:

```json
[
  {
    "apiVersions": null,
    "capabilities": [
      {
        "name": "MaxResourceVolumeMB",
        "value": "768000"
      },
      {
        "name": "OSVhdSizeMB",
        "value": "1047552"
      },
      {
        "name": "vCPUs",
        "value": "20"
      },
      {
        "name": "MemoryPreservingMaintenanceSupported",
        "value": "True"
      },
      {
        "name": "HyperVGenerations",
        "value": "V1,V2"
      },
      {
        "name": "MemoryGB",
        "value": "160"
      },
      {
        "name": "MaxDataDiskCount",
        "value": "32"
      },
      {
        "name": "CpuArchitectureType",
        "value": "x64"
      },
      {
        "name": "LowPriorityCapable",
        "value": "True"
      },
      {
        "name": "PremiumIO",
        "value": "True"
      },
      {
        "name": "VMDeploymentTypes",
        "value": "IaaS,PaaS"
      },
      {
        "name": "vCPUsAvailable",
        "value": "20"
      },
      {
        "name": "ACUs",
        "value": "230"
      },
      {
        "name": "vCPUsPerCore",
        "value": "2"
      },
      {
        "name": "CombinedTempDiskAndCachedIOPS",
        "value": "40000"
      },
      {
        "name": "CombinedTempDiskAndCachedReadBytesPerSecond",
        "value": "335544320"
      },
      {
        "name": "CombinedTempDiskAndCachedWriteBytesPerSecond",
        "value": "335544320"
      },
      {
        "name": "UncachedDiskIOPS",
        "value": "32000"
      },
      {
        "name": "UncachedDiskBytesPerSecond",
        "value": "503316480"
      },
      {
        "name": "EphemeralOSDiskSupported",
        "value": "True"
      },
      {
        "name": "EncryptionAtHostSupported",
        "value": "True"
      },
      {
        "name": "CapacityReservationSupported",
        "value": "True"
      },
      {
        "name": "ConfidentialComputingType",
        "value": "SNP"
      },
      {
        "name": "AcceleratedNetworkingEnabled",
        "value": "False"
      },
      {
        "name": "RdmaEnabled",
        "value": "False"
      },
      {
        "name": "MaxNetworkInterfaces",
        "value": "8"
      }
    ],
    "capacity": null,
    "costs": null,
    "family": "standardECADSv5Family",
    "kind": null,
    "locationInfo": [
      {
        "location": "eastus",
        "zoneDetails": [
          {
            "Name": [
              "2",
              "1"
            ],
            "capabilities": [
              {
                "name": "UltraSSDAvailable",
                "value": "True"
              }
            ],
            "name": null
          }
        ],
        "zones": [
          "2",
          "1"
        ]
      }
    ],
    "locations": [
      "eastus"
    ],
    "name": "Standard_EC20ads_v5",
    "resourceType": "virtualMachines",
    "restrictions": [],
    "size": "EC20ads_v5",
    "tier": "Standard"
  },
]
```

api/v1beta1/types.go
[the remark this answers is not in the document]
CecileRobertMichon: +1 on adhering to Azure API. 

For the record this isn't the first time we run into this (Terraform being different), for example, we had the same dilemma with VM images. In this case I'm leaning toward the Azure API's explicit implementation because the name of the user-facing feature is `Confidential VM` so it is IMO slightly more user friendly to have that explicitly defined.

And yes we can use an enum for validation purposes. See https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/api/v1beta1/types.go#L262 for prior art.

CecileRobertMichon (commented):
This looks great, thanks @mresvanis for all the effort and being patient with the review. My only comment aside from an e2e test which has already been raised is lack of documentation. Would be great to see this feature mentioned in https://github.com/kubernetes-sigs/cluster-api-provider-azure/tree/main/docs/book/src.

mresvanis:
> This looks great, thanks @mresvanis for all the effort and being patient with the review. My only comment aside from an e2e test which has already been raised is lack of documentation. Would be great to see this feature mentioned in https://github.com/kubernetes-sigs/cluster-api-provider-azure/tree/main/docs/book/src.

@CecileRobertMichon I have currently tested this feature e2e by creating my own images via [image-builder](https://github.com/kubernetes-sigs/image-builder/pull/1148). In order to be able to:

- add e2e unit tests
- add the respective `confidential VM` and `trusted launch for VMs` doc topics

we would need to publish additional VM CAPI images in the Azure Marketplace. The images needed would be based on:

- Ubuntu Server 22.04 (or 20.04) for CVMs - `Canonical:0001-com-ubuntu-confidential-vm-focal:20_04-lts-cvm:latest` 
- Windows Server 2022 (or 2019) for CVMs, e.g. `MicrosoftWindowsServer:windows-cvm:2022-datacenter-cvm:latest`
- Ubuntu Server gen2, e.g.  `Canonical:UbuntuServer:18_04-lts-gen2:latest`
- Windows Server gen2, e.g. `MicrosoftWindowsServer:windowsserver-gen2preview:2019-datacenter-gen2:latest`

Do you think that this would that be possible? (if I understand correctly this process is currently Azure/Microsoft internal)

api/v1beta1/types.go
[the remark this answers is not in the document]
mresvanis: @nawazkh thank you for your feedback regarding the refactoring of adhering closely to the Azure REST API. Please take a look when possible, I believe the suggested refactoring is now in place.

CecileRobertMichon:
@mresvanis while we may not be able to publish (and more importantly maintain long term) a new Marketplace image (especially given that we'd like to move away from publishing Azure Marketplace altogether, see #3370) a good compromise for now could be to write a doc that specifies all the steps needed for someone to use/test this like you did, including steps to build an image with Image-Builder.

api/v1beta1/azuremachine_validation.go
```diff
@@ -234,6 +238,16 @@ func validateManagedDisk(m *ManagedDiskParameters, fieldPath *field.Path, isOSDi
 
 	if m != nil {
 		allErrs = append(allErrs, validateStorageAccountType(m.StorageAccountType, fieldPath.Child("StorageAccountType"), isOSDisk)...)
+
+		if isOSDisk && m.SecurityProfile != nil && m.SecurityProfile.DiskEncryptionSet != nil {
+			if m.SecurityProfile.SecurityEncryptionType != SecurityEncryptionTypeDiskWithVMGuestState {
```
nawazkh, line 243: Do you think we could add a comment here explaining this validation logic for a quick review ? 
Would be great to add link to the validation rules too. 

For instance, I found referring to https://learn.microsoft.com/en-us/rest/api/compute/virtual-machines/create-or-update?tabs=HTTP#securityencryptiontypes to be helpful understanding this validation.

api/v1beta1/azuremachine_validation.go
```diff
@@ -378,3 +392,45 @@ func ValidateDiagnostics(diagnostics *Diagnostics, fieldPath *field.Path) field.
 
 	return allErrs
 }
+
+// ValidateConfidentialCompute validates the configuration options when the machine is a Confidential VM.
```
nawazkh, line 396: Do you think it would be worth adding a quick reference links for these validations? I was referring to the below ones, but still had some challenges in understanding the validations. 

- https://learn.microsoft.com/en-us/rest/api/compute/virtual-machines/create-or-update?tabs=HTTP#vmdisksecurityprofile
- https://learn.microsoft.com/en-us/rest/api/compute/virtual-machines/create-or-update?tabs=HTTP#securitytypes
- https://learn.microsoft.com/en-us/rest/api/compute/virtual-machines/create-or-update?tabs=HTTP#uefisettings

nawazkh: // TODO(@nawazkh) review these validations again for sanity check.

nawazkh (commented):
This is simply amazing @mresvanis ! Appreciate your meticulousness in having all the validations and supporting it with unit tests. Kudos! ✨ 

I had some thoughts on improving the code readability of some of the validations by adding more comments to it. Please share what you think about it? 
Moreover, I think test files do help in understanding the green paths. 

Also, thank you for rebasing, code looks much smaller to review! 

I echo Cecile's and Willie's thought on having an E2E tests (which I already think you are on it with the image-builder PR to get testable VM images) and having documentation explaining the confidential VM feature set.

nawazkh:
I haven't `lgtm`-ed it yet because I want to go over the validation and some of the defaulting logic again. It looks great on the first pass tho, holding it just for my sanity.

api/v1beta1/types.go
```diff
@@ -783,10 +829,36 @@ func (s SubnetSpec) IsIPv6Enabled() bool {
...
-	// or disabled for a virtual machine or virtual machine scale
-	// set. Default is disabled.
+	// or disabled for a virtual machine or virtual machine scale set.
+	// This should be disabled when SecurityEncryptionType is set to DiskWithVMGuestState.
+	// Default is disabled.
 	// +optional
 	EncryptionAtHost *bool `json:"encryptionAtHost,omitempty"`
+	// SecurityType specifies the SecurityType of the virtual machine. It has to be set to any specified value to
+	// enable UefiSettings. The default behavior is: UefiSettings will not be enabled unless this property is set.
+	// +optional
```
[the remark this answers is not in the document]
mresvanis: Absolutely, thanks for spotting this!

api/v1beta1/azuremachine_validation.go
mresvanis: I tried to add useful comments to the validations, but I'm afraid that for most of them there is no detailed Azure REST API doc link that we could use. The best way I managed to verify them was via trial and error, i.e. creating VMs with different configuration, after checking out the respective [Azure Terraform provider documentation](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/linux_virtual_machine#security_encryption_type).

api/v1beta1/azuremachine_validation.go
mresvanis: @nawazkh thank you for the sanity check! As mentioned in [this](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3265#discussion_r1188415831) comment, I tried to add comments on the validations. But I'm not quite sure how helpful they are. I absolutely agree though that having the docs linked is quite helpful. Please let me know if you have more suggestions on how to make it more readable and clearer.

nawazkh:
This PR looks good to me!

CecileRobertMichon:
I still don't see any documentation in the PR, @mresvanis are you still working on adding it?

mresvanis:
> I still don't see any documentation in the PR, @mresvanis are you still working on adding it?

@CecileRobertMichon yes, I was waiting for the image-builder [PR](https://github.com/kubernetes-sigs/image-builder/pull/1148) to get merged first, but I will add the documentation today so that you have time to review it.

mresvanis:
@CecileRobertMichon I added documentation for both Confidential VMs and Trusted Launch for VMs, by adding 2 new topics. Please feel free to review. Thank you.

nawazkh:
I looked into the newly added docs and they look good to me! 🚀 
I like the thought of waiting until we have a defined way of generating CVM images and then merging this doc with the steps.

/lgtm

mresvanis:
@mboersma thank you for your input in the respective image-builder [PR](https://github.com/kubernetes-sigs/image-builder/pull/1148)! Since that has been merged, your thoughts here would also be much appreciated.

api/v1beta1/types.go
```diff
@@ -31,6 +31,28 @@ const (
 	Bastion string = "bastion"
 )
 
+// SecurityEncryptionType represents the Encryption Type when the virtual machine is a
+// Confidential VM.
+type SecurityEncryptionType string
```
mboersma, line 36: This type and constants redefine some from [azure-sdk-for-go](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/resourcemanager/compute/armcompute/constants.go#L1722). Could we use those directly instead?

azure/services/virtualmachines/spec.go
```diff
@@ -196,6 +196,20 @@ func (s *VMSpec) generateStorageProfile() (*compute.StorageProfile, error) {
 		if s.OSDisk.ManagedDisk.DiskEncryptionSet != nil {
 			storageProfile.OsDisk.ManagedDisk.DiskEncryptionSet = &compute.DiskEncryptionSetParameters{ID: pointer.String(s.OSDisk.ManagedDisk.DiskEncryptionSet.ID)}
 		}
+		if s.OSDisk.ManagedDisk.SecurityProfile != nil {
+			if _, exists := s.SKU.GetCapability(resourceskus.ConfidentialComputingType); !exists {
+				return nil, azure.WithTerminalError(fmt.Errorf("vm size %s does not support confidential computing. select a different vm size or remove the security profile of the OS disk", s.Size))
```
mboersma, line 201:
```suggestion
				return nil, azure.WithTerminalError(fmt.Errorf("VM type %s does not support confidential computing. Select a different VM type or remove the security profile of the OS disk", s.Size))
```

We seem to be using "type" more consistently in the rest of this PR, and I slightly prefer that to "size," if we don't want to use "sku."

azure/services/virtualmachines/spec_test.go
```diff
@@ -417,6 +461,108 @@ func TestParameters(t *testing.T) {
...
+						VTpmEnabled:       pointer.Bool(true),
+					},
+				},
+				SKU: validSKUWithConfidentialComputingType,
+			},
+			existing: nil,
+			expect: func(g *WithT, result interface{}) {
+				g.Expect(result).To(BeNil())
+			},
+			expectedError: "reconcile error that cannot be recovered occurred: securityType should be set to ConfidentialVM when securityEncrytionType is set. Object will not be requeued",
```
mboersma, line 564:
```suggestion
			expectedError: "reconcile error that cannot be recovered occurred: securityType should be set to ConfidentialVM when securityEncryptionType is set. Object will not be requeued",
```

azure/services/virtualmachines/spec.go
```diff
@@ -277,18 +291,89 @@ func (s *VMSpec) generateOSProfile() (*compute.OSProfile, error) {
...
-	if !s.SKU.HasCapability(resourceskus.EncryptionAtHost) {
-		return nil, azure.WithTerminalError(errors.Errorf("encryption at host is not supported for VM type %s", s.Size))
+	securityProfile := &compute.SecurityProfile{}
+
+	if storageProfile.OsDisk.ManagedDisk != nil &&
+		storageProfile.OsDisk.ManagedDisk.SecurityProfile != nil &&
+		storageProfile.OsDisk.ManagedDisk.SecurityProfile.SecurityEncryptionType != "" {
+		if s.SecurityProfile.EncryptionAtHost != nil && *s.SecurityProfile.EncryptionAtHost &&
+			storageProfile.OsDisk.ManagedDisk.SecurityProfile.SecurityEncryptionType == compute.SecurityEncryptionTypesDiskWithVMGuestState {
+			return nil, azure.WithTerminalError(errors.Errorf("encryption at host is not supported when securityEncrytionType is set to %s", compute.SecurityEncryptionTypesDiskWithVMGuestState))
```
mboersma, line 306:
```suggestion
			return nil, azure.WithTerminalError(errors.Errorf("encryption at host is not supported when securityEncryptionType is set to %s", compute.SecurityEncryptionTypesDiskWithVMGuestState))
```

azure/services/virtualmachines/spec.go
```diff
@@ -277,18 +291,89 @@ func (s *VMSpec) generateOSProfile() (*compute.OSProfile, error) {
...
+		}
+
+		if s.SecurityProfile.SecurityType != infrav1.SecurityTypesConfidentialVM {
+			return nil, azure.WithTerminalError(errors.Errorf("securityType should be set to %s when securityEncrytionType is set", infrav1.SecurityTypesConfidentialVM))
```
mboersma, line 310:
```suggestion
			return nil, azure.WithTerminalError(errors.Errorf("securityType should be set to %s when securityEncryptionType is set", infrav1.SecurityTypesConfidentialVM))
```

azure/services/virtualmachines/spec.go
```diff
@@ -277,18 +291,89 @@ func (s *VMSpec) generateOSProfile() (*compute.OSProfile, error) {
...
+		}
+
+		if s.SecurityProfile.UefiSettings == nil {
+			return nil, azure.WithTerminalError(errors.New("vTpmEnabled should be true when securityEncrytionType is set"))
```
mboersma, line 314:
```suggestion
			return nil, azure.WithTerminalError(errors.New("vTpmEnabled should be true when securityEncryptionType is set"))
```

azure/services/virtualmachines/spec.go
```diff
@@ -277,18 +291,89 @@ func (s *VMSpec) generateOSProfile() (*compute.OSProfile, error) {
...
+		}
+
+		if storageProfile.OsDisk.ManagedDisk.SecurityProfile.SecurityEncryptionType == compute.SecurityEncryptionTypesDiskWithVMGuestState &&
+			!*s.SecurityProfile.UefiSettings.SecureBootEnabled {
+			return nil, azure.WithTerminalError(errors.Errorf("secureBootEnabled should be true when securityEncrytionType is set to %s", compute.SecurityEncryptionTypesDiskWithVMGuestState))
```
mboersma, line 319:
```suggestion
			return nil, azure.WithTerminalError(errors.Errorf("secureBootEnabled should be true when securityEncryptionType is set to %s", compute.SecurityEncryptionTypesDiskWithVMGuestState))
```

azure/services/virtualmachines/spec.go
```diff
@@ -277,18 +291,89 @@ func (s *VMSpec) generateOSProfile() (*compute.OSProfile, error) {
...
+		}
+
+		if s.SecurityProfile.UefiSettings.VTpmEnabled != nil && !*s.SecurityProfile.UefiSettings.VTpmEnabled {
+			return nil, azure.WithTerminalError(errors.New("vTpmEnabled should be true when securityEncrytionType is set"))
```
mboersma, line 323:
```suggestion
			return nil, azure.WithTerminalError(errors.New("vTpmEnabled should be true when securityEncryptionType is set"))
```

azure/services/virtualmachines/spec_test.go
```diff
@@ -437,6 +583,231 @@ func TestParameters(t *testing.T) {
...
+						VTpmEnabled: pointer.Bool(true),
+					},
+				},
+				SKU: validSKUWithConfidentialComputingType,
+			},
+			existing: nil,
+			expect: func(g *WithT, result interface{}) {
+				g.Expect(result).To(BeNil())
+			},
+			expectedError: "reconcile error that cannot be recovered occurred: encryption at host is not supported when securityEncrytionType is set to DiskWithVMGuestState. Object will not be requeued",
```
mboersma, line 703:
```suggestion
			expectedError: "reconcile error that cannot be recovered occurred: encryption at host is not supported when securityEncryptionType is set to DiskWithVMGuestState. Object will not be requeued",
```

azure/services/virtualmachines/spec_test.go
```diff
@@ -437,6 +583,231 @@ func TestParameters(t *testing.T) {
...
+						VTpmEnabled:       pointer.Bool(true),
+					},
+				},
+				SKU: validSKUWithConfidentialComputingType,
+			},
+			existing: nil,
+			expect: func(g *WithT, result interface{}) {
+				g.Expect(result).To(BeNil())
+			},
+			expectedError: "reconcile error that cannot be recovered occurred: secureBootEnabled should be true when securityEncrytionType is set to DiskWithVMGuestState. Object will not be requeued",
```
mboersma, line 739:
```suggestion
			expectedError: "reconcile error that cannot be recovered occurred: secureBootEnabled should be true when securityEncryptionType is set to DiskWithVMGuestState. Object will not be requeued",
```

azure/services/virtualmachines/spec_test.go
```diff
@@ -437,6 +583,231 @@ func TestParameters(t *testing.T) {
...
+						VTpmEnabled: pointer.Bool(false),
+					},
+				},
+				SKU: validSKUWithConfidentialComputingType,
+			},
+			existing: nil,
+			expect: func(g *WithT, result interface{}) {
+				g.Expect(result).To(BeNil())
+			},
+			expectedError: "reconcile error that cannot be recovered occurred: vTpmEnabled should be true when securityEncrytionType is set. Object will not be requeued",
```
mboersma, line 774:
```suggestion
			expectedError: "reconcile error that cannot be recovered occurred: vTpmEnabled should be true when securityEncryptionType is set. Object will not be requeued",
```

api/v1beta1/types.go
```diff
@@ -614,8 +636,32 @@ type VMExtension struct {
...
 	DiskEncryptionSet *DiskEncryptionSetParameters `json:"diskEncryptionSet,omitempty"`
+	// SecurityProfile specifies the security profile for the managed disk.
+	// +optional
+	SecurityProfile *VMDiskSecurityProfile `json:"securityProfile,omitempty"`
+}
+
+// VMDiskSecurityProfile specifies the security profile settings for the managed disk.
+// It can be set only for Confidential VMs.
+type VMDiskSecurityProfile struct {
+	// DiskEncryptionSet specifies the customer managed disk encryption set resource id for the
```
mboersma, line 650:
```suggestion
	// DiskEncryptionSet specifies the customer-managed disk encryption set resource id for the
```

api/v1beta1/types.go
```diff
@@ -614,8 +636,32 @@ type VMExtension struct {
 type ManagedDiskParameters struct {
 	// +optional
 	StorageAccountType string `json:"storageAccountType,omitempty"`
+	// DiskEncryptionSet specifies the customer managed disk encryption set resource id for the managed disk.
```
mboersma, line 639:
```suggestion
	// DiskEncryptionSet specifies the customer-managed disk encryption set resource id for the managed disk.
```

docs/book/src/topics/confidential-vms.md
```diff
@@ -0,0 +1,69 @@
+# Confidential VMs
+
+This document describes how to deploy a cluster with Azure [Confidential VM](https://learn.microsoft.com/en-us/azure/confidential-computing/confidential-vm-overview) nodes.
```
mboersma, line 3:
```suggestion
This document describes how to deploy a cluster with Azure [Confidential VM](https://learn.microsoft.com/azure/confidential-computing/confidential-vm-overview) nodes.
```

This is nitpicky, but I try to remove the `en-us/` section of URLs so the user can (possibly) land on a localized version of the document. Same for the other URLs in these new docs, but it's not crucial to fix. (We should probably make a comprehensive pass over the docs at some point for this.)

docs/book/src/topics/confidential-vms.md
```diff
@@ -0,0 +1,69 @@
...
+- [Limitations](https://learn.microsoft.com/en-us/azure/confidential-computing/confidential-vm-overview#limitations)
+
+## Confidential VM Images
+
+One of the limitations of Confidential VMs is that they support specific OS images, as they need to get [successfully attested](https://learn.microsoft.com/en-us/azure/confidential-computing/confidential-vm-overview#attestation-and-tpm) during boot.
+
+Confidential VM images are not included in the list of `capi` reference images. Before creating a cluster hosted on Azure Confidential VMs, you can create a [custom image](custom-images.md) based on a Confidential VM supported OS image using [image-builder](https://github.com/kubernetes-sigs/image-builder). For example, you can run the following to create such an image based on Ubuntu Server 22.04 LTS for CVMs:
+
+```bash
+$ make -C build-azure-sig-ubuntu-2204-cvm
```
mboersma, line 20:
```suggestion
$ make -C images/capi build-azure-sig-ubuntu-2204-cvm
```

docs/book/src/topics/trusted-launch-for-vms.md
```diff
@@ -0,0 +1,65 @@
...
+
+Before you begin, be aware of the following:
+
+- [Limitations](https://learn.microsoft.com/en-us/azure/virtual-machines/trusted-launch#limitations)
+- [SecureBoot](https://learn.microsoft.com/en-us/azure/virtual-machines/trusted-launch#secure-boot)
+- [vTPM](https://learn.microsoft.com/en-us/azure/virtual-machines/trusted-launch#vtpm)
+
+## Trusted Launch Images
+
+One of the limitations of trusted launch for VMs is that they are supported by [generation 2](https://learn.microsoft.com/en-us/azure/virtual-machines/generation-2) VMs.
```
mboersma, line 15:
```suggestion
One of the limitations of trusted launch for VMs is that they require [generation 2](https://learn.microsoft.com/en-us/azure/virtual-machines/generation-2) VMs.
```

docs/book/src/topics/trusted-launch-for-vms.md
```diff
@@ -0,0 +1,65 @@
...
+
+Trusted launch supported OS images are not included in the list of `capi` reference images. Before creating a cluster hosted on VMs with trusted launch features enabled, you can create a [custom image](custom-images.md) based on a one of the trusted launch supported OS images using [image-builder](https://github.com/kubernetes-sigs/image-builder). For example, you can run the following to create such an image based on Ubuntu Server 22.04 LTS:
+
+```bash
+$ make -C build-azure-sig-ubuntu-2204-gen2
```
mboersma, line 20:
```suggestion
$ make -C images/capi build-azure-sig-ubuntu-2204-gen2
```
(Or else drop the `-C`, assuming you already did `cd images/capi` in the image-builder project.)

mboersma (commented):
Thank you so much for your work and your patience. This is great!

I only found some typos and had some minor doc comments, feel free to ignore those you disagree with.

api/v1beta1/types.go
CecileRobertMichon: There are good reasons for not using types directly from the Azure Go SDK. We don't want to have a dependency on a particular SDK version in our API as it might result in breaking changes between API versions.

Redefining some of the SDK types and constants is a side effect of that, and what we've done for all other existing types. 
 
+1 for keeping it as-is

mresvanis: @CecileRobertMichon thank you for the clarification, it's helpful as to where we should be steering towards. Looking at the codebase, I think there are fields defined as generic `string`s, whereas they are in fact `enum`s (e.g. the [one](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/api/v1beta1/types.go#L616) mentioned above). Those get validated via the respective `azure-sdk-for-go` `enum`s. If I understand correctly, these validations will take place irrespective of whether we would define those `enum`s in the API or just leave them as generic `string`s. 

But IMHO, having the `enum`s defined here (even if they seem as duplicates of `azure-sdk-for-go`), communicates clearly the separation of our API from `azure-sdk-for-go`. Which potentially helps users easily identify what the current version of the API supports, regarding those particular types, irrespective of `azure-sdk-for-go`.

I'm happy to go with either implementation, as long as it adheres to where we should be heading (but I do have to admit that I would go with keeping it as-is as well :) ).

mboersma: Makes sense to me: let's leave it as is.

mboersma (commented):
/lgtm

CecileRobertMichon (commented):
/approve

PR #3268, opened 2023-03-10
.github/ISSUE_TEMPLATE/proposal.md: add comment about duplicated items

invidian:
**What type of PR is this?**
/kind cleanup

To avoid people trying to fix it with sequential numbering.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/.github/ISSUE_TEMPLATE/proposal.md
+++ b/.github/ISSUE_TEMPLATE/proposal.md
@@ -6,6 +6,7 @@ about: Propose larger efforts, breaking changes, or new features
 
 > ⚠️ Cluster API Azure maintainers can ask to turn an issue-proposal into a [CAEP](https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/proposals/YYYYMMDD-template.md) when necessary. This is to be expected for large changes that impact multiple components, breaking changes, or new large features.
 
+<!-- Repeat "1." so list items auto-increment when they are rendered. This is acceptable as this is a GitHub issue template, so this document is not intended to be read outside of GitHub. -->
 ***Goals***
 1. Goal 1
 1. Goal 2


nawazkh:
Thanks for spotting this!
/lgtm

mboersma:
I think this was done deliberately to exploit a feature of Markdown. An ordered list that uses `1.` to start each element will be rendered with incrementing numbers:

```
(plain text)
1. foo
1. bar
1. baz
```

Markdown transforms into
1. foo
1. bar
1. baz

This feature makes list editing more flexible, since the user can re-order things without having to do the tedious re-numbering.

Having said that, I always felt like this feature clashed with the overall goal that Markdown documents be easily readable as plain text. And especially in this case where the user is presented with the plain text in an editor; it's just confusing.

invidian:
Thanks for context @mboersma, TIL. I indeed immediately found it confusing, but with the context and given that we talk about GitHub issue template, I'd lean towards leaving it as it is, but perhaps adding a comment mentioning this behavior, so other people don't try to fix it? What do you think?

nawazkh:
> This feature makes list editing more flexible, since the user can re-order things without having to do the tedious re-numbering.

Thank you for pointing it out, I did not know about this.
Having learned this, I am inclined on leaving the template as is too, and probably adding a comment explaining this behavior.

/remove-lgtm

mboersma:
I think adding a short explanatory comment would help. Maybe:

```
<!-- Repeat "1." so Markdown auto-increments -->
***Goals***
1. Goal 1
1. Goal 2
```

invidian:
Added a comment and changed PR title and description as suggested.

mboersma (approved):
/lgtm
/approve

mboersma:
/retest

> Error response from daemon: Get "https://capzci.azurecr.io/v2/": unauthorized: authentication required, visit https://aka.ms/acr/authorization for more information.

PR #3272, opened 2023-03-10
ci-entrypoint points to correct kubeconfig

jackfrancis:
**What type of PR is this?**

/kind failing-test

This is a follow-up from https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3213

This fixes bugs in IPv6 and in-tree logic.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/scripts/ci-entrypoint.sh
+++ b/scripts/ci-entrypoint.sh
@@ -147,11 +147,11 @@ create_cluster() {
 # and any statement must be idempotent so that subsequent retry attempts can make forward progress.
 get_cidrs() {
     # Get cluster CIDRs from Cluster object
-    CIDR0=$(${KUBECTL} get cluster "${CLUSTER_NAME}" -o=jsonpath='{.spec.clusterNetwork.pods.cidrBlocks[0]}')
+    CIDR0=$(${KUBECTL} --kubeconfig "${REPO_ROOT}/${KIND_CLUSTER_NAME}.kubeconfig" get cluster "${CLUSTER_NAME}" -o=jsonpath='{.spec.clusterNetwork.pods.cidrBlocks[0]}')
     export CIDR0
-    CIDR_LENGTH=$(${KUBECTL} get cluster "${CLUSTER_NAME}" -o=jsonpath='{.spec.clusterNetwork.pods.cidrBlocks}' | jq '. | length')
+    CIDR_LENGTH=$(${KUBECTL} --kubeconfig "${REPO_ROOT}/${KIND_CLUSTER_NAME}.kubeconfig" get cluster "${CLUSTER_NAME}" -o=jsonpath='{.spec.clusterNetwork.pods.cidrBlocks}' | jq '. | length')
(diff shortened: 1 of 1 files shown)


scripts/ci-entrypoint.sh
```diff
@@ -223,14 +230,27 @@ install_cloud_provider_azure() {
...
+  enableDynamicReloading: "${ENABLE_DYNAMIC_RELOADING}"
+  cloudConfig: "${CLOUD_CONFIG}"
+  cloudConfigSecretName: "${CONFIG_SECRET_NAME}"
+  logVerbosity: "${CCM_LOG_VERBOSITY}"
+  clusterCIDR: "${CCM_CLUSTER_CIDR}"
+EOF
+    echo "Using the following cloud-provider-azure helm configuration:"
+    cat "${CLOUD_PROVIDER_HELM_VALUES}"
+    if [ -f "${CLOUD_PROVIDER_HELM_VALUES_CUSTOM_BUILD}" ]; then
+        cat "${CLOUD_PROVIDER_HELM_VALUES_CUSTOM_BUILD}"
```
CecileRobertMichon, line 248: instead of having to cat and rm a file on every run, would it be cleaner to persist this values file in the capz repo and simply use envsubst to sub in the env vars?

jackfrancis: Yes.

I tried to scope this change down so I could fix the missing `--kubeconfig` issue as quickly as possible w/ local repro data. (I was blocked doing so on my macos).

We could do one of two things:

1) We could remove this intermittent change entirely and trust that the exported "arrays" implementation seems to work via prow, and then implement your more durable solution as a follow-up.
2) Keep these intermittent changes temporarily and implement your more durable solution as a follow-up.

CecileRobertMichon: My vote is for 1) so we can merge this quickly and backport it with high confidence. I can confirm I can use the "arrays" implementation on Linux and I know it works in prow too, but we should definitely follow up and fix it for macos users.

jackfrancis:
/retest

Let's see if the E2E gods also agree that merging this quickly is a priority

scripts/ci-entrypoint.sh
jackfrancis: Done, filed https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/3292

CecileRobertMichon:
> /cherry-pick release-1.7

IIRC this doesn't need to go in 1.7 because the CCM changes only went into 1.8

jackfrancis:
@CecileRobertMichon I think you're right, can you cancel via chatbot or if not I'll just close the PR (but based on your comment the auto cherry-pick process will probably fail)

CecileRobertMichon:
I don't think you can cancel but the cherry-pick will definitely fail so we can just let it fail

PR #3284, opened 2023-03-13
CI: always build Linux and Windows images

### LINKED ISSUE #3270: Should build the CNM for both linux and Windows if image is not present
**Which jobs are flaky:**
We found that our Windows jobs weren't building the windows image for CCM: https://github.com/kubernetes/kubernetes/issues/116474

**Which tests are flaky:**

**Testgrid link:**

We were missing the ENV variable and have added this to fix the issue temporarily but we should be building the image for all architectures if missing.  

If we don't then we might create a tag for linux only and then later publich a multi-arch image under the same tag.

Remove this check:

https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/0f497f833a710ad2ad720124f2dd3003da9326e8/scripts/ci-build-azure-ccm.sh#L64C3-L68

- links to go.k8s.io/triage appreciated
- links to specific failures in spyglass appreciated

/kind flake

jackfrancis:
**What type of PR is this?**

This PR updates our build bits script so that we always include Windows images.

In practice this converts the `TEST_WINDOWS` environment variable into a "look for Windows nodes when validating". We can do further work to remove this manual `TEST_WINDOWS` environment variable in the near future.

Fixes #3270

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/scripts/ci-build-azure-ccm.sh
+++ b/scripts/ci-build-azure-ccm.sh
@@ -61,13 +61,8 @@ main() {
     if [[ "$(can_reuse_artifacts)" =~ "false" ]]; then
         echo "Build Linux Azure amd64 cloud controller manager"
         make -C "${AZURE_CLOUD_PROVIDER_ROOT}" build-ccm-image-amd64 push-ccm-image-amd64
-        if [[ -n "${TEST_WINDOWS:-}" ]]; then
-            echo "Building Linux amd64 and Windows ${WINDOWS_IMAGE_VERSION} amd64 cloud node managers"
+        echo "Building Linux amd64 and Windows ${WINDOWS_IMAGE_VERSION} amd64 cloud node managers"
             make -C "${AZURE_CLOUD_PROVIDER_ROOT}" build-node-image-linux-amd64 push-node-image-linux-amd64 push-node-image-windows-"${WINDOWS_IMAGE_VERSION}"-amd64 manifest-node-manager-image-windows-"${WINDOWS_IMAGE_VERSION}"-amd64
-        else
-            echo "Building Linux amd64 cloud node manager"
-            make -C "${AZURE_CLOUD_PROVIDER_ROOT}" build-node-image-linux-amd64 push-node-image-linux-push-name-amd64
-        fi
     fi
 }
 
@@ -79,13 +74,11 @@ can_reuse_artifacts() {
         fi
     done
 
-    if [[ -n "${TEST_WINDOWS:-}" ]]; then
-        FULL_VERSION=$(docker manifest inspect mcr.microsoft.com/windows/nanoserver:${WINDOWS_IMAGE_VERSION} | jq -r '.manifests[0].platform["os.version"]')
-        if ! docker manifest inspect "${REGISTRY}/${CNM_IMAGE_NAME}:${IMAGE_TAG}" | grep -q "\"os.version\": \"${FULL_VERSION}\""; then
-            echo "false" && return
-        fi
+    FULL_VERSION=$(docker manifest inspect mcr.microsoft.com/windows/nanoserver:${WINDOWS_IMAGE_VERSION} | jq -r '.manifests[0].platform["os.version"]')
+    if ! docker manifest inspect "${REGISTRY}/${CNM_IMAGE_NAME}:${IMAGE_TAG}" | grep -q "\"os.version\": \"${FULL_VERSION}\""; then
+        echo "false" && return
     fi
-    
+
     echo "true"
 }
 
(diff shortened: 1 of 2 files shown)


scripts/ci-build-azure-ccm.sh
```diff
@@ -61,13 +61,8 @@ main() {
     if [[ "$(can_reuse_artifacts)" =~ "false" ]]; then
         echo "Build Linux Azure amd64 cloud controller manager"
         make -C "${AZURE_CLOUD_PROVIDER_ROOT}" build-ccm-image-amd64 push-ccm-image-amd64
-        if [[ -n "${TEST_WINDOWS:-}" ]]; then
-            echo "Building Linux amd64 and Windows ${WINDOWS_IMAGE_VERSION} amd64 cloud node managers"
+        echo "Building Linux amd64 and Windows ${WINDOWS_IMAGE_VERSION} amd64 cloud node managers"
```
jsturtevant, line 64: should we build the image for both Windows versions right away? or only if we end up needing it?  Could go either way.

Not as worried about appending the additional image to the manifest list as I was when we were converting a single arch image to manifest list.

scripts/ci-entrypoint.sh
```diff
@@ -102,15 +102,16 @@ setup() {
     export CCM_COUNT="${CCM_COUNT:-1}"
     export WORKER_MACHINE_COUNT="${WORKER_MACHINE_COUNT:-2}"
     export EXP_CLUSTER_RESOURCE_SET="true"
-
     # this requires k8s 1.22+
+    if [[ -n "${K8S_FEATURE_GATES:-}" ]]; then
```
jsturtevant, line 106: do we need this for all clusters? I don't think it hurts anything

scripts/ci-build-azure-ccm.sh
jackfrancis: I think what I'm learning in this PR is that "if we end up needing it" is not something we can deterministically predict, and so building "everything that we may need" is a better approach

scripts/ci-entrypoint.sh
jackfrancis: IMO the answer is "yes" because "has a windows node" is not a function of a cluster, it is a potential cluster configuration state that may or may not exist at a given time. Explicitly disallowing your cluster to ever include key Windows functionality is a mistake (IMO).

jsturtevant: after 1.26 this flag is on by default anyways.

jackfrancis: cool, love me some non-controversy :)

scripts/ci-build-azure-ccm.sh
marosset: We normally don't set `WINDOWS_SERVER_VERSION` when we are testing WS2019. I think that would result in `WINDOWS_IMAGE_VERSION` not being set (based on lines 49-57 above)

jackfrancis: can we set an appropriate default for `WINDOWS_IMAGE_VERSION` if both it and `WINDOWS_SERVER_VERSION` are not set?

jackfrancis: I assume that would be something like `export WINDOWS_IMAGE_VERSION="1809"` given what we're doing above

marosset: Yea we can do that.
Can you please also refernce https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/8cc82a6cdbe13f38eba040249c0b51d9824e38c7/azure/defaults.go#L61-L65
So if/when we do switch the default from WS2019 to WS2022 we know all the places we are defaulting to 2019?

jackfrancis: actually I think L38 is already doing this for us

I can add a reference

marosset: Oh yea, I didn't scroll up enough in the file.
Let's add the ref and then this should be good to merge.

jackfrancis: Given the state of E2E and the importance of this if it passes current tests I think we should merge and then I'll submit a follow-up PR w/ code comments and such. Stay tuned...

marosset (approved):
/lgtm

jsturtevant:
/hold 
for https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/3270#issuecomment-1467110953

CecileRobertMichon:
Do we want to backport this to be able to revert https://github.com/kubernetes/test-infra/pull/28991 ?

marosset:
> Do we want to backport this to be able to revert [kubernetes/test-infra#28991](https://github.com/kubernetes/test-infra/pull/28991) ?

yes!

PR #3304, opened 2023-03-15
Add unit tests for azure/pointers.go

### LINKED ISSUE #3291: Add unit tests for azure/pointers.go
/kind cleanup

Add unit tests for the [`azure/pointers.go`](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/pointers.go) file. The goal is to reach at least 80% code coverage for `pointers.go`.

This is a part of a larger effort to reach 60% code coverage in CAPZ.

**Environment:**

PiotrLewandowski323:
**What type of PR is this?**
/kind cleanup

This PR adds missing unit tests for azure/pointers.go

Fixes #3291

I adjusted test names to function descriptions. For the StringMapPtr I'm not sure if the intended behavior is fully described, as this function returns a pointer to a map, where the values are pointers to the values from the original map.

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/azure/pointers.go
+++ b/azure/pointers.go
@@ -27,11 +27,14 @@ func StringSlice(s *[]string) []string {
 	return nil
 }
 
-// StringMapPtr returns a pointer to a given string map, or a pointer to an empty map if the map is nil.
-func StringMapPtr(m map[string]string) *map[string]*string { //nolint:gocritic
+// StringMapPtr converts a map[string]string into a map[string]*string. It returns nil if the map is nil.
+func StringMapPtr(m map[string]string) map[string]*string {
+	if m == nil {
+		return nil
+	}
 	msp := make(map[string]*string, len(m))
 	for k, v := range m {
 		msp[k] = pointer.String(v)
 	}
-	return &msp
+	return msp
 }
--- a/azure/pointers.go
+++ b/azure/pointers.go
@@ -27,7 +27,7 @@ func StringSlice(s *[]string) []string {
 	return nil
 }
 
-// StringMapPtr returns a pointer to a given string map, or nil if the map is nil.
+// StringMapPtr returns a pointer to a given string map, or a pointer to an empty map if the map is nil.
 func StringMapPtr(m map[string]string) *map[string]*string { //nolint:gocritic
 	msp := make(map[string]*string, len(m))
 	for k, v := range m {
--- a/azure/services/agentpools/spec.go
+++ b/azure/services/agentpools/spec.go
@@ -352,7 +352,7 @@ func (s *AgentPoolSpec) Parameters(ctx context.Context, existing interface{}) (p
 			VnetSubnetID:         vnetSubnetID,
 			EnableNodePublicIP:   s.EnableNodePublicIP,
 			NodePublicIPPrefixID: s.NodePublicIPPrefixID,
-			Tags:                 *azure.StringMapPtr(s.AdditionalTags),
+			Tags:                 azure.StringMapPtr(s.AdditionalTags),
 			LinuxOSConfig:        linuxOSConfig,
 		},
 	}
--- a/azure/services/agentpools/spec.go
+++ b/azure/services/agentpools/spec.go
@@ -352,7 +352,7 @@ func (s *AgentPoolSpec) Parameters(ctx context.Context, existing interface{}) (p
 			VnetSubnetID:         vnetSubnetID,
 			EnableNodePublicIP:   s.EnableNodePublicIP,
 			NodePublicIPPrefixID: s.NodePublicIPPrefixID,
-			Tags:                 azure.StringMapPtr(s.AdditionalTags),
+			Tags:                 converters.TagsToMap(s.AdditionalTags),
 			LinuxOSConfig:        linuxOSConfig,
 		},
 	}
--- a/azure/services/agentpools/spec.go
+++ b/azure/services/agentpools/spec.go
@@ -265,6 +265,11 @@ func (s *AgentPoolSpec) Parameters(ctx context.Context, existing interface{}) (p
 	if s.SKU != "" {
 		sku = &s.SKU
 	}
+	tags := converters.TagsToMap(s.AdditionalTags)
+	if tags == nil {
+		// Make sure we send a non-nil, empty map if AdditionalTags are nil as this tells AKS to delete any existing tags.
+		tags = make(map[string]*string, 0)
+	}
 	var vnetSubnetID *string
 	if s.VnetSubnetID != "" {
 		vnetSubnetID = &s.VnetSubnetID
@@ -352,7 +357,7 @@ func (s *AgentPoolSpec) Parameters(ctx context.Context, existing interface{}) (p
 			VnetSubnetID:         vnetSubnetID,
 			EnableNodePublicIP:   s.EnableNodePublicIP,
 			NodePublicIPPrefixID: s.NodePublicIPPrefixID,
-			Tags:                 converters.TagsToMap(s.AdditionalTags),
+			Tags:                 tags,
 			LinuxOSConfig:        linuxOSConfig,
 		},
 	}
--- a/azure/converters/tags.go
+++ b/azure/converters/tags.go
@@ -19,6 +19,7 @@ package converters
 import (
 	"k8s.io/utils/pointer"
 	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
+	"sigs.k8s.io/cluster-api-provider-azure/azure"
 )
 
 // MapToTags converts a map[string]*string into a infrav1.Tags.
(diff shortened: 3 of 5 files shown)


azure/pointers.go
```diff
@@ -27,7 +27,7 @@ func StringSlice(s *[]string) []string {
 	return nil
 }
 
-// StringMapPtr returns a pointer to a given string map, or nil if the map is nil.
+// StringMapPtr returns a pointer to a given string map, or a pointer to an empty map if the map is nil.
```
nawazkh, line 30: Good catch @PiotrLewandowski323!
Just wondering if we wanted `StringMapPtr` to return `nil` instead of _a pointer to empty map_ when input is an nil map ?
@mboersma @CecileRobertMichon What do you say?

nawazkh: Looks good so far. Just holding this PR for thoughts from others! 

/hold

willie-yao: @nawazkh This [line](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/services/agentpools/spec.go#L355) suggests that the intended behavior is to return a pointer to an empty map, otherwise there would be a nil pointer dereference here.

willie-yao (approved):
/lgtm

azure/pointers.go
CecileRobertMichon: Looking at this more closely... I actually think this is a bug, and we might not even need this function at all. For the line @willie-yao pointed to, we could be using https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/converters/tags.go#L40 which actually had a fix a very similar bug recently (https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2802)

Returning a pointer to a map in Go doesn't really make sense, since maps are already reference types https://go.dev/blog/maps (probably why  the `gocritic` linter was complaining in the first place?)

The only other reference I see of this func in the code is https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/services/managedclusters/spec.go#L350 and here again we're dereferencing the map right after `StringMapPtr`.

I would suggest:

1) Changing this function to 

```
func StringMapPtr(m map[string]string) map[string]*string {
        if m == nil {
          return nil
        }
	msp := make(map[string]*string, len(m))
	for k, v := range m {
		msp[k] = pointer.String(v)
	}
	return msp
}
```

2) Changing the Tags converter function so it's just a wrapper to this func:

```
// TagsToMap converts infrav1.Tags into a map[string]*string.
func TagsToMap(src infrav1.Tags) map[string]*string {
	return StringMapPtr(src)
}
```

3) Changing both occurrences in the code of `*StringMapPtr()` to either `StringMapPtr()` or `TagsToMap()` for the one that is handling tags.

I realize this is more than the original ask @PiotrLewandowski323 so if you would rather not do this work we can open a TODO issue for the above.

PiotrLewandowski323: @CecileRobertMichon Your suggestions seem clear to me, so I've made changes accordingly.

azure/services/agentpools/spec.go
```diff
@@ -352,7 +352,7 @@ func (s *AgentPoolSpec) Parameters(ctx context.Context, existing interface{}) (p
 			VnetSubnetID:         vnetSubnetID,
 			EnableNodePublicIP:   s.EnableNodePublicIP,
 			NodePublicIPPrefixID: s.NodePublicIPPrefixID,
-			Tags:                 *azure.StringMapPtr(s.AdditionalTags),
+			Tags:                 converters.TagsToMap(s.AdditionalTags),
```
CecileRobertMichon, line 355: It seems that this change is what is causing the aks e2e test to fail:

```
The function passed to Eventually failed at /home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/test/e2e/aks_tags.go:182 with:
Expected
    <v1beta1.Tags | len:2>: {"new": "value", "test": "updated"}
to be nil failed [FAILED] Timed out after 1800.000s.
The function passed to Eventually failed at /home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/test/e2e/aks_tags.go:182 with:
Expected
    <v1beta1.Tags | len:2>: {"new": "value", "test": "updated"}
to be nil
In [It] at: /home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/test/e2e/aks_tags.go:229 @ 03/17/23 15:25:09.421
```

CecileRobertMichon: @nojnhuh any ideas on this one? I think that's an e2e test you added recently, want to make sure we're not regressing things here

nojnhuh: I think in this case we wanted the behavior where a nil `s.AdditionalTags` should result in the `Tags` being set to a non-nil, empty map which tells AKS to delete all the tags which is what the test is trying to do. Sending nil to the API means "don't change the tags" which seems to be what the test is running into.

CecileRobertMichon: Got it, I wonder if we should just change this to 

```
		Tags: converters.TagsToMap(infrav1.Build(infrav1.BuildParams{
			Lifecycle:   infrav1.ResourceLifecycleOwned,
			ClusterName: s.ClusterName,
			Name:        pointer.String(s.Name),
			Role:        pointer.String(infrav1.CommonRole),
			Additional:  s. AdditionalTags,
		})),
```

then, just like you did in https://github.com/CecileRobertMichon/cluster-api-provider-azure/commit/ec154f70c636aca350990a81e980d9b4d7ebdb11#diff-d7496d9845d55dda5142aab2970a5a9aa1012fbb52d2c462cc5b220682ca520a so we're also tagging agentpools with the standard CAPZ owned tags?

nojnhuh: FWIW I didn't see tags on agent pools affected by the same constant reconciliation issue as the managed cluster resource from #3009, so it seems like tags on agent pools are classified differently somehow. That seems like a reasonable change to add those labels though since I don't see how that would hurt anything. I can put that change together quick and assuming that goes smoothly, this PR can maybe just ignore the whole issue.

nojnhuh: > assuming that goes smoothly

It did not :/ 

AKS doesn't seem to let you use the same "update the tags for this resource ID" API for agent pools. Everything looks like it works, but the tags are applied to the managed cluster, not the agent pool.

I thought I may have made a mistake wiring things up in CAPZ, but the CLI seems to work the same way:
```
% az tag update --resource-id /subscriptions/$AZURE_SUBSCRIPTION_ID/resourceGroups/aks-23760/providers/Microsoft.ContainerService/managedClusters/aks-23760 --operation merge --tags jontocluster=fromcli

% az tag update --resource-id /subscriptions/$AZURE_SUBSCRIPTION_ID/resourceGroups/aks-23760/providers/Microsoft.ContainerService/managedClusters/aks-23760/agentPools/pool0 --operation merge --tags jontopool=fromcli

% az tag list --resource-id /subscriptions/$AZURE_SUBSCRIPTION_ID/resourceGroups/aks-23760/providers/Microsoft.ContainerService/managedClusters/aks-23760/agentPools/pool0 -o yaml
id: /subscriptions/$AZURE_SUBSCRIPTION_ID/resourceGroups/aks-23760/providers/Microsoft.ContainerService/managedClusters/aks-23760/providers/Microsoft.Resources/tags/default
name: default
properties:
  tags:
    Name: aks-23760
    jontocluster: fromcli
    jontopool: fromcli
    sigs.k8s.io_cluster-api-provider-azure_cluster_aks-23760: owned
    sigs.k8s.io_cluster-api-provider-azure_role: common
resourceGroup: aks-23760
type: Microsoft.Resources/tags
```
^ The `id` from the output refers to the managed cluster.

I think getting this to work just like tags on managed cluster resources will be more complex than it might be worth, so I'm in favor of adding the minimal logic here to pass the e2e test at least for the short-term.

CecileRobertMichon: > so I'm in favor of adding the minimal logic here to pass the e2e test at least for the short-term
 
which minimal logic?

CecileRobertMichon: Makes sense to me. @PiotrLewandowski323 let us know if you need additional clarification on this. 

This would look something like this:

```
tags := converters.TagsToMap(s.AdditionalTags)
if tags == nil {
  // Make sure we send a non-nil, empty map if AdditionalTags are nil as this tells AKS to delete any existing tags.
  tags = make(map[string]*string, 0)
}

...

Tags:                 tags,

```

CecileRobertMichon: @nojnhuh one more question for my own understanding: if we set the agentpool tags to an empty map when AdditionalTags is nil, wouldn't that result in endless reconciliation if some other process is adding tags to the agentpools in the sub?

Specifically, this part: 

> I didn't see tags on agent pools affected by the same constant reconciliation issue as the managed cluster resource from https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/3009

nojnhuh: I think it would, but at least the "add a tag to all resources" Azure Policy I found to test that issue didn't seem to apply to agent pools, just the managed cluster. That makes me wonder how easy it would be to get into that scenario in the first place. It seems like it would have to be very deliberate.

CecileRobertMichon:
/retest

known flake #3227

nojnhuh (approved):
/lgtm

CecileRobertMichon (commented):
/approve

Thanks @PiotrLewandowski323!

PR #3322, opened 2023-03-16
Fetch AzureCluster name from OwnerCluster instead of assuming ClusterName = AzureCluster.Name

### LINKED ISSUE #3313: clusterclass test spec fails often
**Which jobs are failing:**

pull-cluster-api-provider-azure-e2e-optional
periodic-cluster-api-provider-azure-e2e-main

**Which tests are failing:**

"Creating clusters using clusterclass [OPTIONAL]"

**Since when has it been failing:**

Unclear, but at least for the last 2-3 days in PRs.

**Testgrid link:**

https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/3256/pull-cluster-api-provider-azure-e2e-optional/1636391586499661824
https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/periodic-cluster-api-provider-azure-e2e-main/1636248867903115264

"KubeadmControlPlane object capz-e2e-yadda-yadda was not initialized in time"

/kind failing-test

/area bootstrap

willie-yao:
**What type of PR is this?**
/kind bug

Recently, the clusterclass tests have been failing due to the `AzureCluster` name being different from the `ClusterName` whereas they were equal in other E2E tests. This exposed a bug in getting the subscription ID for a machine/machinepool where the code [erroneously assumed](https://github.com/willie-yao/cluster-api-provider-azure/blob/92f57d00e42f0004344cec079272d7a2e0830070/api/v1beta1/azuremachine_default.go#L184) that `Cluster.Name` is always equal to `AzureCluster.Name`. This PR fixes this issue by splitting the existing `GetSubscriptionID()` function into two steps:

1. Getting the owner cluster
2. Getting the subscription ID from the owner cluster's infrastructure ref, thus getting the correct `AzureCluster` name

Fixes #3313

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Fixed a bug with GetSubscriptionID where it fetched the AzureCluster object using the Cluster's name instead of the AzureCluster's name
```

### DIFF
--- a/api/v1beta1/azuremachine_default.go
+++ b/api/v1beta1/azuremachine_default.go
@@ -174,21 +174,20 @@ func (s *AzureMachineSpec) SetNetworkInterfacesDefaults() {
 	}
 }
 
-// GetSubscriptionID returns the subscription ID for the given cluster name and namespace.
-func GetSubscriptionID(cli client.Client, clusterName string, namespace string, maxAttempts int) (string, error) {
+// GetOwnerAzureClusterNameAndNamespace returns the owner azure cluster's name and namespace for the given cluster name and namespace.
+func GetOwnerAzureClusterNameAndNamespace(cli client.Client, clusterName string, namespace string, maxAttempts int) (azureClusterName string, azureClusterNamespace string, err error) {
 	ctx := context.Background()
 
-	ownerCluster := &AzureCluster{}
+	ownerCluster := &clusterv1.Cluster{}
 	key := client.ObjectKey{
 		Namespace: namespace,
 		Name:      clusterName,
 	}
 
 	for i := 1; ; i++ {
-		err := cli.Get(ctx, key, ownerCluster)
-		if err != nil {
+		if err := cli.Get(ctx, key, ownerCluster); err != nil {
 			if i > maxAttempts {
-				return "", errors.Wrapf(err, "failed to find owner cluster for %s/%s", namespace, clusterName)
+				return "", "", errors.Wrapf(err, "failed to find owner cluster for %s/%s", namespace, clusterName)
 			}
 			time.Sleep(1 * time.Second)
 			continue
@@ -196,7 +195,30 @@ func GetSubscriptionID(cli client.Client, clusterName string, namespace string, 
 		break
 	}
 
-	return ownerCluster.Spec.SubscriptionID, nil
+	return ownerCluster.Spec.InfrastructureRef.Name, ownerCluster.Spec.InfrastructureRef.Namespace, nil
+}
+
+// GetSubscriptionID returns the subscription ID for the AzureCluster given the cluster name and namespace.
+func GetSubscriptionID(cli client.Client, ownerAzureClusterName string, ownerAzureClusterNamespace string, maxAttempts int) (string, error) {
+	ctx := context.Background()
+
+	ownerAzureCluster := &AzureCluster{}
+	key := client.ObjectKey{
+		Namespace: ownerAzureClusterNamespace,
+		Name:      ownerAzureClusterName,
+	}
+	for i := 1; ; i++ {
+		if err := cli.Get(ctx, key, ownerAzureCluster); err != nil {
+			if i >= maxAttempts {
+				return "", errors.Wrapf(err, "failed to find AzureCluster for owner cluster %s/%s", ownerAzureClusterNamespace, ownerAzureClusterName)
+			}
+			time.Sleep(1 * time.Second)
+			continue
+		}
+		break
+	}
+
+	return ownerAzureCluster.Spec.SubscriptionID, nil
 }
 
 // SetDefaults sets to the defaults for the AzureMachineSpec.
@@ -212,7 +234,12 @@ func (m *AzureMachine) SetDefaults(client client.Client) error {
 		errs = append(errs, errors.Errorf("failed to fetch ClusterName for AzureMachine %s/%s", m.Namespace, m.Name))
 	}
 
-	subscriptionID, err := GetSubscriptionID(client, clusterName, m.Namespace, 5)
+	ownerAzureClusterName, ownerAzureClusterNamespace, err := GetOwnerAzureClusterNameAndNamespace(client, clusterName, m.Namespace, 5)
+	if err != nil {
+		errs = append(errs, errors.Wrapf(err, "failed to fetch owner cluster for AzureMachine %s/%s", m.Namespace, m.Name))
+	}
+
+	subscriptionID, err := GetSubscriptionID(client, ownerAzureClusterName, ownerAzureClusterNamespace, 5)
 	if err != nil {
 		errs = append(errs, errors.Wrapf(err, "failed to fetch subscription ID for AzureMachine %s/%s", m.Namespace, m.Name))
 	}
--- a/api/v1beta1/azuremachine_default_test.go
+++ b/api/v1beta1/azuremachine_default_test.go
@@ -26,6 +26,7 @@ import (
 	"github.com/google/uuid"
 	. "github.com/onsi/gomega"
 	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/utils/pointer"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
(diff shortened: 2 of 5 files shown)


api/v1beta1/azuremachine_default.go
```diff
@@ -174,11 +174,11 @@ func (s *AzureMachineSpec) SetNetworkInterfacesDefaults() {
 	}
 }
 
-// GetSubscriptionID returns the subscription ID for the given cluster name and namespace.
-func GetSubscriptionID(cli client.Client, clusterName string, namespace string, maxAttempts int) (string, error) {
+// GetOwnerCluster returns the owner cluster for the given cluster name and namespace.
+func GetOwnerCluster(cli client.Client, clusterName string, namespace string, maxAttempts int) (*clusterv1.Cluster, error) {
```
nawazkh, line 178: Would it be better to return `ownerCluster.Namespace` and `ownerCluster.Spec.InfrastructureRef.Name` instead of a pointer to the cluster object ? One more advantage(I can think of) of returning string(in this case) over a pointer would be that the variable `ownerCluster := &clusterv1.Cluster{}` would not "escape" `GetOwnerCluster()` func, adding to the function optimization.

willie-yao: That's a great suggestion! I initially left it as returning the whoel cluster in case this is useful elsewhere, but returning just the name and namespace makes more sense in this case. Will work on implementing this now!

willie-yao: Also, I realized I should be returning `ownerCluster.Spec.InfrastructureRef.Namespace` instead because I'm looking for the `AzureCluster` and not the `Cluster`. They should be the same but I don't think it's guaranteed.

api/v1beta1/azuremachine_default.go
```diff
@@ -188,15 +188,39 @@ func GetSubscriptionID(cli client.Client, clusterName string, namespace string,
...
+
+	ownerAzureCluster := &AzureCluster{}
+	key := client.ObjectKey{
+		Namespace: ownerAzureClusterNamespace,
+		Name:      ownerAzureClusterName,
+	}
+	for i := 1; ; i++ {
+		err := cli.Get(ctx, key, ownerAzureCluster)
+		if err != nil {
+			if i > maxAttempts {
```
nojnhuh, line 214: nit: I don't think it really matters and this is copypasta from above, but it looks like this will actually do `maxAttempts+1` attempts.
```suggestion
			if i >= maxAttempts {
```

api/v1beta1/azuremachine_default_test.go
```diff
@@ -401,39 +402,102 @@ func TestAzureMachineSpec_SetNetworkInterfacesDefaults(t *testing.T) {
...
+			maxAttempts:     5,
+			wantedName:      "test-cluster",
+			wantedNamespace: "default",
+			wantErr:         true,
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			client := mockClient{ReturnError: tc.wantErr}
```
nojnhuh, line 440: nit:
```suggestion
			g := NewWithT(t)
			client := mockClient{ReturnError: tc.wantErr}
```

api/v1beta1/azuremachine_default_test.go
```diff
@@ -453,9 +517,21 @@ func (m mockClient) Get(ctx context.Context, key client.ObjectKey, obj client.Ob
 	if m.ReturnError {
 		return errors.New("AzureCluster not found: failed to find owner cluster for test-cluster")
 	}
-	ac := &AzureCluster{}
-	ac.Spec.SubscriptionID = "test-subscription-id"
-	obj.(*AzureCluster).Spec.SubscriptionID = ac.Spec.SubscriptionID
+	// Check if we're calling Get on an AzureCluster or a Cluster
```
nojnhuh, line 520: This logic can probably be simplified with `switch obj.(type)` and `case *AzureCluster:`, `case *clusterv1.Cluster`, etc.

api/v1beta1/azuremachine_default.go
```diff
@@ -174,11 +174,11 @@ func (s *AzureMachineSpec) SetNetworkInterfacesDefaults() {
...
 }
 
-// GetSubscriptionID returns the subscription ID for the given cluster name and namespace.
-func GetSubscriptionID(cli client.Client, clusterName string, namespace string, maxAttempts int) (string, error) {
+// GetOwnerAzureClusterNameAndNamespace returns the owner azure cluster's name and namespace for the given cluster name and namespace.
+func GetOwnerAzureClusterNameAndNamespace(cli client.Client, clusterName string, namespace string, maxAttempts int) (azureClusterName string, azureClusterNamespace string, err error) {
 	ctx := context.Background()
 
-	ownerCluster := &AzureCluster{}
+	ownerCluster := &clusterv1.Cluster{}
```
nojnhuh, line 181: Since this "get CAPI Cluster" step is new, is the retry logic here still applicable the same way as for the "get CAPZ AzureCluster" step? Or is there an opportunity to make it more strict or even eliminate it entirely here?

api/v1beta1/azuremachine_default.go
```diff
@@ -188,15 +188,39 @@ func GetSubscriptionID(cli client.Client, clusterName string, namespace string,
...
+func GetSubscriptionID(cli client.Client, ownerAzureClusterName string, ownerAzureClusterNamespace string, maxAttempts int) (string, error) {
+	ctx := context.Background()
+
+	ownerAzureCluster := &AzureCluster{}
+	key := client.ObjectKey{
+		Namespace: ownerAzureClusterNamespace,
+		Name:      ownerAzureClusterName,
+	}
+	for i := 1; ; i++ {
+		err := cli.Get(ctx, key, ownerAzureCluster)
```
mboersma, line 212: Same nit here: we seem to prefer the single-line form when possible.

```go
if err := cli.Get(ctx, key, ownerAzureCluster); err != nil {
```

api/v1beta1/azuremachine_default.go
willie-yao: Good catch!

api/v1beta1/azuremachine_default_test.go
willie-yao: Is this the recommended pattern? I see that most other tests only declare `g := NewWithT(t)` at the start of the test function.

willie-yao: If so, I might be more in favor of keeping things consistent in this PR, and creating a separate PR to make this change across the package.

api/v1beta1/azuremachine_default_test.go
willie-yao: Ah I wasn't aware of this method!! Thanks for the tip

api/v1beta1/azuremachine_default_test.go
nojnhuh: From my point of view, that would be the recommended pattern. When the `g` is created inside the subtest, error messages will include the name of the exact subtest that failed. Otherwise it's not always obvious exactly which test case failed.

api/v1beta1/azuremachine_default.go
willie-yao: Looks like the retries are specific to machinepools from this comment. I will remove the retry logic for now from these util functions.

https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3080#discussion_r1089303163

exp/api/v1beta1/azuremachinepool_webhook_test.go
```diff
@@ -264,6 +264,18 @@ type mockDefaultClient struct {
 }
 
 func (m mockDefaultClient) Get(ctx context.Context, key client.ObjectKey, obj client.Object, opts ...client.GetOption) error {
+	_, ok := obj.(*infrav1.AzureCluster)
```
nojnhuh, line 267: It seems like the same `switch obj.(type)` could be used here too.

nojnhuh (approved):
/lgtm

CecileRobertMichon:
/hold

this is a bug fix, please add a release note @willie-yao

willie-yao:
@CecileRobertMichon done! This bug is present in 1.8 and was introduced by #2965

willie-yao:
> this is fixing more than a test no? this could repro in the wild? also isn't this fixing cluster class?

That's correct! I got that mixed up and I wrote a more descriptive release note. That's my bad :sweat:

CecileRobertMichon:
All good! Thank you for fixing it 

/hold cancel

jackfrancis:
Thank you @willie-yao!

CecileRobertMichon:
> CONFLICT (content): Merge conflict in api/v1beta1/azuremachine_default_test.go

looks like you won the manual cherry-pick lottery 😄

willie-yao:
oof for my first one too!! 😅

PR #3337, opened 2023-03-20
Replace some usage of Azure/go-autorest

mboersma:
**What type of PR is this?**

/kind cleanup

Replaces some usage of the soon-to-be-unsupported [Azure/go-autorest](https://github.com/Azure/go-autorest) package.

Refs #2974

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/azure/scope/machinepool.go
+++ b/azure/scope/machinepool.go
@@ -24,7 +24,7 @@ import (
 	"io"
 	"strings"
 
-	azureautorest "github.com/Azure/go-autorest/autorest/azure"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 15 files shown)


CecileRobertMichon (commented):
/lgtm

azure/scope/machinepool.go
```diff
@@ -377,11 +377,11 @@ func (m *MachinePoolScope) createMachine(ctx context.Context, machine azure.VMSS
 	ctx, _, done := tele.StartSpanWithLogger(ctx, "scope.MachinePoolScope.createMachine")
 	defer done()
 
-	parsed, err := azureautorest.ParseResourceID(machine.ID)
+	parsed, err := arm.ParseResourceID(machine.ID)
```
jackfrancis, line 380: Are we 100% confident these two `ParseResourceID` interfaces are identical?

mboersma: The implementations are very different (see [go-autorest](https://github.com/Azure/go-autorest/blob/553a90ae65a6a2b18306fa04d7b1625960c5decb/autorest/azure/azure.go#L213) and [azure-sdk-for-go](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/arm/internal/resource/resource_identifier.go#L63)), but the interface (API) and overall behavior appear to be the same–I tested with some resourceID strings. The SDKv2 `arm` version returns a different struct, but with the same information (although some field names changed).

jackfrancis: It seems like we don't have existing UT cases for these transformations, so I'll defer to your validations, thank you for doing that work!

CecileRobertMichon:
Flex passed but clusterclass failed...

/retest

CecileRobertMichon:
>       -- Original Error: Code="PublicIPCountLimitReached" Message="Cannot create more
      than 100 public IP addresses for this subscription in this region." Details=[]'

/retest

mboersma:
I don't think the `-optional` test will pass until #3322 merges, since it's the ClusterClass failure.

Issue #3303, opened 2023-03-15
Skip AzureManagedControlPlane.Spec.ControlPlaneEndpoint.Host and AzureManagedControlPlane.Spec.ControlPlaneEndpoint.Port validation while cluster is paused/moved


tapojit047:
/kind bug

**What steps did you take and what happened:**
- Create a local kind cluster.
- Set `EXP_MACHINE_POOL=true` and `EXP_AKS=true`.
- `clusterctl init` with `--infrastructure azure`.
- `clusterctl generate cluster` with `--flavor aks`.
- Init the target cluster (same as 3).
- `clusterctl move` from local `kind` to target AKS.
- Notice the error below:
```azure
clusterctl move --to-kubeconfig=$AZURE_KUBECONFIG
Performing move...
Discovering Cluster API objects
Moving Cluster API objects Clusters=1
Moving Cluster API objects ClusterClasses=0
Creating objects in the target cluster
Error: action failed after 10 attempts: error creating "infrastructure.cluster.x-k8s.io/v1beta1, Kind=AzureManagedControlPlane" default/new-cluster5: admission webhook "validation.azuremanagedcontrolplanes.infrastructure.cluster.x-k8s.io" denied the request: Spec.ControlPlaneEndpoint.Host: Forbidden: can not be set by the user, will be set automatically by AKS after the cluster is Ready
```

**What did you expect to happen:**

I expect `AzureManagedControlPlane.Spec.ControlPlaneEndpoint.Host` and  `AzureManagedControlPlane.Spec.ControlPlaneEndpoint.Port` can be assigned while performing `clusterctl move`. 

**Anything else you would like to add:**
- The error is raised [here](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/api/v1beta1/azuremanagedcontrolplane_webhook.go#L119).
- I expect that `clusterctl move` operation to be executed successfully.
- For that, maybe the webhook can check whether the `cluster` object of the `AzureManagedControlPlane` has the `spec.paused` set to `true` and `skip` the two aforementioned conditions of `AzureManagedControlPlane.Spec.ControlPlaneEndpoint.Host` and `AzureManagedControlPlane.Spec.ControlPlaneEndpoint.Port`.
- Or maybe `clusterctl move` should annotate the objects it is `moving` and then webhooks would honour that.

**Environment:**

- cluster-api-provider-azure version: `v1.8.0`
[...]

PR #3341, opened 2023-03-21
disable AKS control plane create webhook when paused

jackfrancis:
**What type of PR is this?**

/kind bug

This PR disables the create webhook for AKS control plane data when the cluster is paused, which is the cluster state that a "clusterctl move" operation yields.

Contributes to #3303

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
disable AKS control plane create webhook when paused
```

### DIFF
--- a/api/v1beta1/azuremanagedcontrolplane_webhook.go
+++ b/api/v1beta1/azuremanagedcontrolplane_webhook.go
@@ -116,17 +116,29 @@ func (mw *azureManagedControlPlaneWebhook) ValidateCreate(ctx context.Context, o
 		)
 	}
 
-	if m.Spec.ControlPlaneEndpoint.Host != "" {
-		return field.Forbidden(
-			field.NewPath("Spec", "ControlPlaneEndpoint", "Host"),
-			controlPlaneEndpointErrorMessage,
-		)
+	ownerCluster := &clusterv1.Cluster{}
+	key := client.ObjectKey{
+		Namespace: m.Namespace,
+		Name:      m.Name,
 	}
-	if m.Spec.ControlPlaneEndpoint.Port != 0 {
-		return field.Forbidden(
-			field.NewPath("Spec", "ControlPlaneEndpoint", "Port"),
-			controlPlaneEndpointErrorMessage,
-		)
+
+	if err := mw.Client.Get(ctx, key, ownerCluster); err != nil {
+		return err
+	}
+
+	if !ownerCluster.Spec.Paused {
+		if m.Spec.ControlPlaneEndpoint.Host != "" {
+			return field.Forbidden(
+				field.NewPath("Spec", "ControlPlaneEndpoint", "Host"),
+				controlPlaneEndpointErrorMessage,
+			)
+		}
+		if m.Spec.ControlPlaneEndpoint.Port != 0 {
+			return field.Forbidden(
+				field.NewPath("Spec", "ControlPlaneEndpoint", "Port"),
+				controlPlaneEndpointErrorMessage,
+			)
+		}
 	}
 
 	return m.Validate(mw.Client)
--- a/api/v1beta1/azuremanagedcontrolplane_webhook_test.go
+++ b/api/v1beta1/azuremanagedcontrolplane_webhook_test.go
@@ -27,7 +27,17 @@ import (
 	"sigs.k8s.io/cluster-api-provider-azure/feature"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	capifeature "sigs.k8s.io/cluster-api/feature"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
+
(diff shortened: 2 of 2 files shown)


nawazkh:
This is for my own understanding; 
- Can someone "pause" a cluster while it is spinning up to disable creation of webhooks ? I am guessing that the _non-existent cluster_ cannot be paused, but would like to confirm.. 

- Can disabling creation of webhooks, as done in this PR, be a loophole where a malign user feed the cluster with random values to try break it ?

jackfrancis:
> This is for my own understanding;
> 
> * Can someone "pause" a cluster while it is spinning up to disable creation of webhooks ? I am guessing that the _non-existent cluster_ cannot be paused, but would like to confirm..
> * Can disabling creation of webhooks, as done in this PR, be a loophole where a malign user feed the cluster with random values to try break it ?

I believe something like what you outline is possible. I'm not sure that a user creating and then breaking its own CAPI cluster in this way would qualify as an attack vector. The existing create webhook is meant as a user convenience of last resort, to remind the user if this data is pre-populated at cluster creation time, that that data configuration is invalid (the AKS service will be authoritative for populating the apiserver endpoint during its own creation lifecycle, not the user) and not allowed by a user.

api/v1beta1/azuremanagedcontrolplane_webhook.go
```diff
@@ -116,17 +116,29 @@ func (mw *azureManagedControlPlaneWebhook) ValidateCreate(ctx context.Context, o
...
 
-	if m.Spec.ControlPlaneEndpoint.Host != "" {
-		return field.Forbidden(
-			field.NewPath("Spec", "ControlPlaneEndpoint", "Host"),
-			controlPlaneEndpointErrorMessage,
-		)
+	ownerCluster := &clusterv1.Cluster{}
+	key := client.ObjectKey{
+		Namespace: m.Namespace,
+		Name:      m.Name,
```
nojnhuh, line 122: Is this making a similar assumption to the one that's being addressed in #3322? The owner references would probably be the easiest place to check for the Cluster's namespace/name, but that isn't set until after the AzureManagedControlPlane has been created. Otherwise it seems like we'd have to list all Clusters in all namespaces and see if any match this AzureManagedControlPlane and are paused.

jackfrancis: It's definitely true that the name of an AzureManagedControlPlane can be different from the name of the referencing cluster.

I think we can probably introspect each Cluster for Spec.ControlPlaneRef.Name == the name of the AzureManagedControlPlane; is that what you're thinking?

nojnhuh: Yeah, I was thinking of exactly that. Not sure there's any other way to correlate those right when the object is created.

nojnhuh: One other thought I had was to list with a field selector on the `spec.controlPlaneRef.*` fields but it looks like those aren't supported.

nojnhuh: And do we know for sure that `clusterctl move` will always create the `Cluster` before the `AzureManagedCluster`? The CAPI webhooks appear to allow either to be created first. If we have to account for when no matching Cluster exists yet here, I'm thinking we'd probably want to skip the `controlPlaneEndpoint` checks in that case to make sure we're not blocking a valid `clusterctl move`?

jackfrancis: Indeed maybe the most practical solution here is to remove these Create checks entirely. They were implemented defensively to protect the user from accidentally doing something wrong, but the outcome of "wrong" in this case is worst-case scenario a never-actually-manageable cluster, but best-case scenario the AKS controllers overwrite those user-provided values with the correct ones from the AKS API anyways, and so it literally has not effect. (We'd need to look into this to know for sure which outcome is the actual one. But either of those is probably not sufficiently catastrophic or likely to happen to justify all this additional webhook complexity.)

Thoughts?

cc @CecileRobertMichon @mtougeron

CecileRobertMichon: > Indeed maybe the most practical solution here is to remove these Create checks entirely.

I tend to agree

> the AKS controllers overwrite those user-provided values with the correct ones from the AKS API anyways

are we sure this is still the case? I seem to remember we changed it to only set them when empty

jackfrancis: Indeed UT tells the tale:

https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/v1.8.1/api/v1beta1/azuremanagedcontrolplane_webhook_test.go#L1228-L1278

I'll submit a PR that removes create webhooks for this and AzureCluster and we can discuss there.

PR #3347, opened 2023-03-23
use /12 for pod CIDR in reference and test templates

jackfrancis:
**What type of PR is this?**

/kind documentation

This PR updates our reference and test templates to use a `/12` CIDR for pod IP addresses, to accommodate larger clusters, especially those using the Windows-required `strictAffinity: true` calico configuration, which has the side-effect of allocating a relatively large, static set of IP addresses for pods ahead of time, per node (default is /26, or 64 IP addresses in the pod CIDR network). A `/16` network has 65k IP addresses of availability, but if each node is using 64, then the maximum node count is somewhere around 1024 before you reach IP exhaustion. That's a lot of nodes! But we know that it's definitely possible to create many more than that, and to anticipate folks using our published templates as a reference, allowing `/12` for pod IP addresses allows for ~1 million pod IP addresses, which will enable 16k or so nodes, which is a much more lenient upper boundary for very large cluster scenarios.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/templates/addons/calico/values.yaml
+++ b/templates/addons/calico/values.yaml
@@ -5,5 +5,5 @@ installation:
     bgp: Disabled
     mtu: 1350
     ipPools:
-    - cidr: 192.168.0.0/16
+    - cidr: 172.16.0.0/12
       encapsulation: VXLAN
--- a/docs/book/src/topics/addons.md
+++ b/docs/book/src/topics/addons.md
@@ -114,7 +114,7 @@ spec:
   clusterNetwork:
     pods:
       cidrBlocks:
-      - 192.168.0.0/16
+      - 172.16.0.0/12
 ```
 
 Download the file at `https://raw.githubusercontent.com/coreos/flannel/master/Documentation/kube-flannel.yml` and modify the `kube-flannel-cfg` ConfigMap.
@@ -134,7 +134,7 @@ metadata:
 data:
   net-conf.json: |
     {
-      "Network": "192.168.0.0/16",
+      "Network": "172.16.0.0/12",
       "Backend": {
         "Type": "vxlan"
       }
--- a/docs/book/src/topics/failure-domains.md
+++ b/docs/book/src/topics/failure-domains.md
@@ -202,7 +202,7 @@ spec:
   clusterNetwork:
     pods:
       cidrBlocks:
-      - 192.168.0.0/16
+      - 172.16.0.0/12
   controlPlaneRef:
     apiVersion: controlplane.cluster.x-k8s.io/v1alpha3
     kind: KubeadmControlPlane
--- a/docs/book/src/topics/managedcluster.md
+++ b/docs/book/src/topics/managedcluster.md
@@ -106,7 +106,7 @@ spec:
   clusterNetwork:
     services:
       cidrBlocks:
-      - 192.168.0.0/16
+      - 172.16.0.0/12
   controlPlaneRef:
     apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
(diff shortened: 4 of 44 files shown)


CecileRobertMichon (commented):
- What are the downsides of doing this? 
- This requires a change to the quickstart https://cluster-api.sigs.k8s.io/user/quick-start.html#install-a-cloud-provider
- All other CAPI providers are using this default value as example, are there any concerns with diverging?

templates/addons/calico/values.yaml
```diff
@@ -5,5 +5,5 @@ installation:
     bgp: Disabled
     mtu: 1350
     ipPools:
-    - cidr: 192.168.0.0/16
+    - cidr: 172.16.0.0/12
```
CecileRobertMichon, line 8: this change specifically is a bit concerning... Others might be using this values file at the master branch (eg. `az capi`) and changing this value will break them if they are still using the old cidr value in templates

jackfrancis: Are these reference templates part of our upgrade contract? If so, yes, this would be non-back-compat.

(But if our reference templates are part of a CAPZ upgrade contract, that's not really sustainable IMO.)

jackfrancis:
> What are the downsides of doing this?

There are no downsides AFAICT.

> This requires a change to the quickstart https://cluster-api.sigs.k8s.io/user/quick-start.html#install-a-cloud-provider

Makes sense

> All other CAPI providers are using this default value as example, are there any concerns with diverging?

I don't think so? The specific motivation for using a larger pod network is the realization that our reference CNI implementation (calico) requires per-node IP blocks in order to work with Windows, which in most configurations will result in IP's being wasted (we will pre-allocate 64 IPs by default per node, and in practice many nodes will run out of available resources before they accept that many pods for scheduling).

There's not really a great "solution" to this, but (IMO) we should be guiding folks to use as large of a CIDR for pod networking as possible as a default to anticipate cluster scaling configurations. A more resilient offering to users would be a detailed set of docs that outline various cluster configuration considerations that are critical when running clusters at large scale.

templates/addons/calico/values.yaml
CecileRobertMichon: @mboersma is `az capi` using the files at main?

mboersma: @CecileRobertMichon no, by default `az capi` uses a local yaml template (which needs to be updated, see Azure/azure-capi-cli-extension#188).

And the CAPZ ref image pipeline uses `az capi` but fetches templates from CAPZ 1.6 currently. I'm updating those to 1.7.3 and then adding MachinePools when time permits.

CecileRobertMichon: @mboersma that's not what the code is saying: https://github.com/Azure/azure-capi-cli-extension/blob/main/src/capi/azext_capi/custom.py#L653

this is for the helm install of Calico specifically. It would be an issue if the main branch values file changes but the templates are still using the old cidr.

/hold

mboersma: There is now a [v0.1.4 release](https://github.com/Azure/azure-capi-cli-extension/releases/tag/v0.1.4) of `az capi` that [pins those files](https://github.com/Azure/azure-capi-cli-extension/pull/191) to the release-1.7 branch instead. So it shouldn't be blocking this PR any longer.

jackfrancis: @mboersma do we really want to pin the CNI implementation to a static release of CAPZ indefinitely?

PR #3358, opened 2023-03-24
[e2e] Support more flexible env var for cloud-provider-azure

lzhecheng:
**What type of PR is this?**
/kind testing

[e2e] Support more flexible env var for cloud-provider-azure
Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/scripts/ci-build-azure-ccm.sh
+++ b/scripts/ci-build-azure-ccm.sh
@@ -48,7 +48,9 @@ setup() {
 
     # the azure-cloud-provider repo expects IMAGE_REGISTRY.
     export IMAGE_REGISTRY=${REGISTRY}
-    pushd "${AZURE_CLOUD_PROVIDER_ROOT}" && IMAGE_TAG=$(git rev-parse --short=7 HEAD) && export IMAGE_TAG && popd
+    pushd "${AZURE_CLOUD_PROVIDER_ROOT}" && IMAGE_TAG=$(git rev-parse --short=7 HEAD) &&
+      IMAGE_TAG_CCM="${IMAGE_TAG_CCM:-${IMAGE_TAG}}" && IMAGE_TAG_CNM="${IMAGE_TAG_CNM:-${IMAGE_TAG}}" &&
+      export IMAGE_TAG && export IMAGE_TAG_CCM && export IMAGE_TAG_CNM && popd
     echo "Image registry is ${REGISTRY}"
     echo "Image Tag is ${IMAGE_TAG}"
 
--- a/test/e2e/cloud-provider-azure.go
+++ b/test/e2e/cloud-provider-azure.go
@@ -54,8 +54,8 @@ func InstallCalicoAndCloudProviderAzureHelmChart(ctx context.Context, input clus
 		options.Values = append(options.Values, fmt.Sprintf("cloudNodeManager.imageName=%s", os.Getenv("CNM_IMAGE_NAME")))
 		options.Values = append(options.Values, fmt.Sprintf("cloudControllerManager.imageRepository=%s", os.Getenv("IMAGE_REGISTRY")))
 		options.Values = append(options.Values, fmt.Sprintf("cloudNodeManager.imageRepository=%s", os.Getenv("IMAGE_REGISTRY")))
-		options.StringValues = append(options.StringValues, fmt.Sprintf("cloudControllerManager.imageTag=%s", os.Getenv("IMAGE_TAG")))
-		options.StringValues = append(options.StringValues, fmt.Sprintf("cloudNodeManager.imageTag=%s", os.Getenv("IMAGE_TAG")))
+		options.StringValues = append(options.StringValues, fmt.Sprintf("cloudControllerManager.imageTag=%s", os.Getenv("IMAGE_TAG_CCM")))
+		options.StringValues = append(options.StringValues, fmt.Sprintf("cloudNodeManager.imageTag=%s", os.Getenv("IMAGE_TAG_CNM")))
 	}
 
 	clusterProxy := input.ClusterProxy.GetWorkloadCluster(ctx, input.ConfigCluster.Namespace, input.ConfigCluster.ClusterName)
--- a/scripts/ci-e2e.sh
+++ b/scripts/ci-e2e.sh
@@ -57,6 +57,13 @@ if [[ "${BUILD_MANAGER_IMAGE}" == "true" ]]; then
   export TAG="${defaultTag:-dev}"
 fi
 
+if [[ "$(capz::util::should_build_ccm)" == "true" ]]; then
+  # shellcheck source=scripts/ci-build-azure-ccm.sh
+  source "${REPO_ROOT}/scripts/ci-build-azure-ccm.sh"
+  echo "Will use the ${IMAGE_REGISTRY}/${CCM_IMAGE_NAME}:${IMAGE_TAG} cloud-controller-manager image for external cloud-provider-cluster"
+  echo "Will use the ${IMAGE_REGISTRY}/${CNM_IMAGE_NAME}:${IMAGE_TAG} cloud-node-manager image for external cloud-provider-azure cluster"
+fi
+
 export GINKGO_NODES=10
 
 export AZURE_LOCATION="${AZURE_LOCATION:-$(capz::util::get_random_region)}"


test/e2e/cloud-provider-azure.go
```diff
@@ -50,12 +52,24 @@ func InstallCalicoAndCloudProviderAzureHelmChart(ctx context.Context, input clus
...
-		options.Values = append(options.Values, fmt.Sprintf("cloudNodeManager.imageName=%s", os.Getenv("CNM_IMAGE_NAME")))
+		ccmImageName := os.Getenv("CCM_IMAGE_NAME")
+		if ccmImageName == "" {
+			ccmImageName = defaultCCMImageName
+		}
+		cnmImageName := os.Getenv("CNM_IMAGE_NAME")
+		if cnmImageName == "" {
+			cnmImageName = defaultCNMImageName
+		}
+		cnmImageTag := os.Getenv("IMAGE_TAG_CNM")
```
jackfrancis, line 63: Why aren't we introducing similar flexility for cloud-controller-manager? (e.g., `$IMAGE_TAG_CCM`)

lzhecheng: I think `IMAGE_TAG` is enough for CCM and we don't need 3 var for 2 options.

lzhecheng: Also, I don't want to change current behavior by replacing `IMAGE_TAG` with `IMAGE_TAG_CCM` and `IMAGE_TAG_CNM`.

CecileRobertMichon: For image tag, I don't fully understand how this is meant to be used since we are building CCM and CNM images from source in order to use them. I agree with Jack it's not consistent UX to have `IMAGE_TAG` for CCM but `IMAGE_TAG_CNM` for CNM. 

Is the goal to reuse an image that was already built?  Even the cloud-provider-azure Makefile uses `IMAGE_TAG`: https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/Makefile#L81

lzhecheng: @CecileRobertMichon my intention of this PR is to make the process of running `ci-e2e.sh` locally easier.

For image name, I think your suggested change in `ci-build-azure-ccm.sh` is good but not related to this problem. I met the situation that after I ran `ci-e2e.sh` and waited for the cluster was up, I found that CCM and CNM cannot pull images because I forgot to set `CCM_IMAGE_NAME` and `CNM_IMAGE_NAME` so these Pods are pulling `<acr>/:<IMGAE_TAG>`. Since for most of the time, these image name are fixed, I'd like to add a default value in e2e code.

For image tag, yes it is for images already built but I think `IMAGE_TAG_CNM` is like an "exception". Reason: Same `IMAGE_TAG` for CCM and CNM work for CAPZ CI process with multi arch CNM. However, if I run `ci-e2e.sh` locally, I only build linux amd64 CNM, which means the image has a suffix: `-linux-amd64`. Then CCM and CNM don't have the same `IMAGE_TAG`. So, I'd like to provide the flexibility that:
1. Users can reuse CCM and CNM (multi-arch) images with the same image tag: `<commit-id>`
2. Users can reuse CCM and CNM (linux/amd64) images with different image tags: `<commit-id>, `<commit-id-linux-amd64>`
The advantage with my change:
For case1, users set `IMAGE_TAG`, it works for CCM and CNM. They don't have to set two image tags with same value.
For case2, users set `IMAGE_TAG` and `IMAGE_TAG_CNM`.

Why other approaches are not good:
1. <IMAGE_TAG>, <IMAGE_TAG_CCM>, <IMAGE_TAG_CNM>: I think 3 variables for 2 options are too many.
2. <IMAGE_TAG_CCM>, <IMAGE_TAG_CNM>: For case1, users have to set 2 variables with same value

lzhecheng: Yes, <IMAGE_TAG_CNM> but no <IMAGE_TAG_CCM> don't look consistent but I think my change may be good from user side. I don't have strong opinion.

CecileRobertMichon: Got it. Thanks for the explanation. The e2e tests weren't adapted to run with custom versions of k8s (we use ci-entrypoint.sh or ci-conformance.sh for that usually) so we never added the option to build CI artifacts and CCM images in the ci-e2e.sh script. I think the right fix for this is to add the ` if [[ "$(capz::util::should_build_ccm)" == "true" ]]; then` block to ci-e2e.sh too, just like https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/scripts/ci-conformance.sh#L59.

That way, the image name will be set, and if the image already exists in the registry it won't build it again.

For

> Users can reuse CCM and CNM (linux/amd64) images with different image tags 

I prefer option 2 

> <IMAGE_TAG_CCM>, <IMAGE_TAG_CNM>: For case1, users have to set 2 variables with same value

Since this isn't user facing, this is only for devs running the tests and in most cases the TAG is set automatically by the `ci-build-azure-ccm.sh` script based on the local checked out commit anyways. IMO it's better to have consistent UX so it's not confusing which one is setting which. Then no need to default in here, you can just set the value directly to the var as it was doing before.

lzhecheng: Let's follow option 2. Since `ci-e2e.sh` focuses on e2e, I think we can let users build the image by themselves.

scripts/ci-build-azure-ccm.sh
```diff
@@ -48,7 +48,9 @@ setup() {
 
     # the azure-cloud-provider repo expects IMAGE_REGISTRY.
     export IMAGE_REGISTRY=${REGISTRY}
-    pushd "${AZURE_CLOUD_PROVIDER_ROOT}" && IMAGE_TAG=$(git rev-parse --short=7 HEAD) && export IMAGE_TAG && popd
+    pushd "${AZURE_CLOUD_PROVIDER_ROOT}" && IMAGE_TAG=$(git rev-parse --short=7 HEAD) &&
+      IMAGE_TAG_CCM="${IMAGE_TAG}" && IMAGE_TAG_CNM="${IMAGE_TAG}" &&
+      export IMAGE_TAG && export IMAGE_TAG_CCM && export IMAGE_TAG_CNM && popd
```
CecileRobertMichon, line 53: here we can simply do 

```
 IMAGE_TAG_CCM="${IMAGE_TAG_CCM:-IMAGE_TAG}" && IMAGE_TAG_CNM="${IMAGE_TAG_CNM:-IMAGE_TAG}" &&
      export IMAGE_TAG_CCM && export IMAGE_TAG_CNM && popd
```

This will ensure `IMAGE_TAG_CNM` and `IMAGE_TAG_CCM` get defaulted to the value of `IMAGE_TAG` if they are not set, but only if not already set.

hack/util.sh
```diff
@@ -41,13 +41,18 @@ capz::util::should_build_kubernetes() {
...
+# If false, no images are built. Then If CCM & CNM images are needed, tags should be IMAGE_TAG_CCM & IMAGE_TAG_CNM.
 capz::util::should_build_ccm() {
     # TEST_CCM is an environment variable set by a prow job to indicate that the CCM should be built and tested.
     if [[ -n "${TEST_CCM:-}" ]]; then
         echo "true" && return
     fi
-    # If conformance is being tested with CI artifacts, CCM should be built.
+    # If conformance is being tested with CI artifacts and IMAGE_TAG_CCM or IMAGE_TAG_CNM unset, CCM should be built.
     if [[ "${E2E_ARGS:-}" == "-kubetest.use-ci-artifacts" ]]; then
+        if [[ -n "${IMAGE_TAG_CCM}" && -n "${IMAGE_TAG_CNM}" ]]; then
```
CecileRobertMichon, line 53: I don't think we need to do this, if the images already exist the `can_reuse_artifacts` function will return true and no images will be built or pushed.

test/e2e/cloud-provider-azure.go
```diff
@@ -50,12 +52,20 @@ func InstallCalicoAndCloudProviderAzureHelmChart(ctx context.Context, input clus
 	}
 	// If testing a CI version of Kubernetes, use CCM and CNM images built from source.
 	if useCIArtifacts {
-		options.Values = append(options.Values, fmt.Sprintf("cloudControllerManager.imageName=%s", os.Getenv("CCM_IMAGE_NAME")))
-		options.Values = append(options.Values, fmt.Sprintf("cloudNodeManager.imageName=%s", os.Getenv("CNM_IMAGE_NAME")))
+		ccmImageName := os.Getenv("CCM_IMAGE_NAME")
```
CecileRobertMichon, line 55: this code is not needed: when `useCIArtifacts` is true (which is always true in this if block), we're calling `ci-build-azure-ccm.sh` in ci-e2e.sh. This means the `CCM_IMAGE_NAME` and `CNM_IMAGE_NAME` image vars will always be set by the time we reach this code.

scripts/ci-build-azure-ccm.sh
lzhecheng: Updated.

hack/util.sh
lzhecheng: Updated.

test/e2e/cloud-provider-azure.go
lzhecheng: Updated.

CecileRobertMichon (approved):
/lgtm
/approve

Thanks so much for reworking this

PR #3359, opened 2023-03-24
remove strict AKS create validations for spec.controlPlaneEndpoint

### LINKED ISSUE #3303: Skip AzureManagedControlPlane.Spec.ControlPlaneEndpoint.Host and AzureManagedControlPlane.Spec.ControlPlaneEndpoint.Port validation while cluster is paused/moved

jackfrancis:
**What type of PR is this?**

This PR removes the strict validations that forbid `spec.controlPlaneEndpoint` data during cluster create, as the `clusterctl move` operation uses this facility to migrate existing cluster resource data from one mgmt cluster to another.

Fixes #3303

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
fix clusterctl move for AKS clusters
```

### DIFF
--- a/api/v1beta1/azuremanagedcluster_webhook.go
+++ b/api/v1beta1/azuremanagedcluster_webhook.go
@@ -25,7 +25,6 @@ import (
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	"sigs.k8s.io/cluster-api-provider-azure/feature"
 	"sigs.k8s.io/cluster-api-provider-azure/util/maps"
-	webhookutils "sigs.k8s.io/cluster-api-provider-azure/util/webhook"
 	capifeature "sigs.k8s.io/cluster-api/feature"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
@@ -52,18 +51,6 @@ func (r *AzureManagedCluster) ValidateCreate() error {
 			"can be set only if the Cluster API 'MachinePool' feature flag is enabled",
 		)
 	}
-	if r.Spec.ControlPlaneEndpoint.Host != "" {
-		return field.Forbidden(
-			field.NewPath("Spec", "ControlPlaneEndpoint", "Host"),
-			controlPlaneEndpointErrorMessage,
-		)
-	}
-	if r.Spec.ControlPlaneEndpoint.Port != 0 {
-		return field.Forbidden(
-			field.NewPath("Spec", "ControlPlaneEndpoint", "Port"),
-			controlPlaneEndpointErrorMessage,
-		)
-	}
 	return nil
 }
 
@@ -81,24 +68,6 @@ func (r *AzureManagedCluster) ValidateUpdate(oldRaw runtime.Object) error {
 				field.NewPath("metadata", "annotations"),
 				r.ObjectMeta.Annotations,
 				fmt.Sprintf("annotations with '%s' prefix are immutable", CustomHeaderPrefix)))
-	}
-
-	if old.Spec.ControlPlaneEndpoint.Host != "" {
-		if err := webhookutils.ValidateImmutable(
-			field.NewPath("Spec", "ControlPlaneEndpoint", "Host"),
-			old.Spec.ControlPlaneEndpoint.Host,
-			r.Spec.ControlPlaneEndpoint.Host); err != nil {
-			allErrs = append(allErrs, err)
-		}
-	}
-
-	if old.Spec.ControlPlaneEndpoint.Port != 0 {
-		if err := webhookutils.ValidateImmutable(
-			field.NewPath("Spec", "ControlPlaneEndpoint", "Port"),
-			old.Spec.ControlPlaneEndpoint.Port,
-			r.Spec.ControlPlaneEndpoint.Port); err != nil {
-			allErrs = append(allErrs, err)
-		}
 	}
 
 	if len(allErrs) != 0 {
--- a/docs/book/src/topics/managedcluster.md
+++ b/docs/book/src/topics/managedcluster.md
@@ -194,6 +194,8 @@ spec:
   sku: Standard_D2s_v4
 ```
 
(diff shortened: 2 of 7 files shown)


api/v1beta1/azuremanagedcluster_webhook.go
```diff
@@ -52,18 +52,6 @@ func (r *AzureManagedCluster) ValidateCreate() error {
 			"can be set only if the Cluster API 'MachinePool' feature flag is enabled",
 		)
 	}
-	if r.Spec.ControlPlaneEndpoint.Host != "" {
```
CecileRobertMichon, line 55: if we're removing this I think we should also revert the changes in the controller from https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2711/files#diff-98a6a70399877f33a02b37dd04551528506730ad55c62e37980efa5e8100d570R168

jackfrancis: Honestly it seems like those changes in the controller are academic, as the update webhook will never allow this pathology, right?

So the idea is to just clean up the code and not leave the impression that that if condition will ever be fulfilled?

CecileRobertMichon: oh I didn't realize we were also enforcing immutability on the update webhook... Perhaps we should just let it be mutable and let the controller overwrite it everytime (basically revert #2711) so we can ensure the endpoint is always set to the correct value. Not the best UX but at least we don't run into weird edge cases. Arguably this is worse UX: `You _may not configure these data when authoring your templates for cluster creation_, otherwise CAPZ will not be able to modify it later with the correct data from the AKS API, and your cluster create operation will fail.`

jackfrancis: Cool, if we can remove that stern language I think that's a good thing.

docs/book/src/topics/managedcluster.md
```diff
@@ -194,6 +194,8 @@ spec:
   sku: Standard_D2s_v4
 ```
 
+Please note that we don't declare a configuration for the apiserver endpoint. This configuration data will be populated automatically based on the data returned from AKS API during cluster create as `.spec.controlPlaneEndpoint.Host` and `.spec.controlPlaneEndpoint.Port` in both the `AzureManagedCluster` and `AzureManagedControlPlane` resources. Any user-provided data will be ignored and overwritted by data returned from the AKS API.
```
nojnhuh, line 197:
```suggestion
Please note that we don't declare a configuration for the apiserver endpoint. This configuration data will be populated automatically based on the data returned from AKS API during cluster create as `.spec.controlPlaneEndpoint.Host` and `.spec.controlPlaneEndpoint.Port` in both the `AzureManagedCluster` and `AzureManagedControlPlane` resources. Any user-provided data will be ignored and overwritten by data returned from the AKS API.
```

docs/book/src/topics/managedcluster.md
```diff
@@ -717,7 +719,12 @@ Following is the list of immutable fields for managed clusters:
 
 | CRD                       | jsonPath                     | Comment                   |
 |---------------------------|------------------------------|---------------------------|
+| AzureManagedCluster       | .spec.controlPlaneEndpoint   | populated by the AKS API during cluster create |
+| AzureManagedCluster       | .name                        |                           |
```
nojnhuh, line 723: I think it's fair to omit the `metadata.*` fields here. If CAPZ had some additional requirements on top of what's common to all Kubernetes resources then I think that would be good to include here though. Otherwise there would be a lot more like `uid` and `creationTimestamp` if we wanted this list to be exhaustive and might seem like noise more than being helpful to me at least.

CecileRobertMichon: +1

docs/book/src/topics/managedcluster.md
jackfrancis: done

docs/book/src/topics/managedcluster.md
jackfrancis: thx, done

nojnhuh (approved):
/lgtm

CecileRobertMichon (commented):
/lgtm
/approve

PR #3221, opened 2023-03-02
Switch ipv6 template to external cloud-provider

CecileRobertMichon:
**What type of PR is this?**

/kind cleanup

**What this PR does / why we need it**: Follow up on #3105 to switch the ipv6 template to out of tree. It was previously blocked due to https://github.com/kubernetes-sigs/cloud-provider-azure/issues/3401.  Root cause can be found here: https://github.com/kubernetes-sigs/cloud-provider-azure/issues/3401#issuecomment-1480540832. The fix is to remove the `--node-ip` flag from kubelet.

> For IPv6, --node-ip should not be set either since it is actually a dual-stack cluster as for Nodes. When someday the cluster is truly IPv6 only, this option should be set.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Switch ipv6 template to external cloud-provider
```

PR #3361, opened 2023-03-24
Add ipv6 IP configs for VMSS

### LINKED ISSUE #3163: Cannot create a VMSS dual-stack cluster
/kind bug

I tried to create a VMSS dual-stack cluster with this [template](https://raw.githubusercontent.com/kubernetes-sigs/cloud-provider-azure/5084a35c0b1ddb77c578bb38fb3478f179adab15/tests/k8s-azure/manifest/cluster-api/linux-dualstack-vmss.yaml). What I got is a working dual-stack control-plane Node but the VMSS Nodes don't have IPv6 address.
```
zhechengli@devbox:~$ kdpo -n calico-system      calico-node-gl4jp
Name:                 calico-node-gl4jp
Namespace:            calico-system
Priority:             2000001000
Priority Class Name:  system-node-critical
Node:                 zhecheng-206-1-mp-0000000/10.1.0.4
Start Time:           Mon, 06 Feb 2023 08:28:23 +0000
Labels:               app.kubernetes.io/name=calico-node
                      controller-revision-hash=85dc7f8d55
                      k8s-app=calico-node
                      pod-template-generation=1
Annotations:          hash.operator.tigera.io/cni-config: 2d19559df12928367220b4d9c76b3883ce3e1755
                      hash.operator.tigera.io/tigera-ca-private: cfb4753a251d65748d10675458f7b80a5b261f09
Status:               Running
IP:                   10.1.0.4
IPs:
  IP:           10.1.0.4
...
    Environment:
      DATASTORE_TYPE:                      kubernetes
      WAIT_FOR_DATASTORE:                  true
      CLUSTER_TYPE:                        k8s,operator
      CALICO_DISABLE_FILE_LOGGING:         false
      FELIX_DEFAULTENDPOINTTOHOSTACTION:   ACCEPT
      FELIX_HEALTHENABLED:                 true
      FELIX_HEALTHPORT:                    9099
      NODENAME:                             (v1:spec.nodeName)
      NAMESPACE:                           calico-system (v1:metadata.namespace)
      FELIX_TYPHAK8SNAMESPACE:             calico-system
      FELIX_TYPHAK8SSERVICENAME:           calico-typha
      FELIX_TYPHACAFILE:                   /etc/pki/tls/certs/tigera-ca-bundle.crt
      FELIX_TYPHACERTFILE:                 /node-certs/tls.crt
      FELIX_TYPHAKEYFILE:                  /node-certs/tls.key
      FIPS_MODE_ENABLED:                   false
      FELIX_TYPHACN:                       typha-server
      CALICO_MANAGE_CNI:                   true
      CALICO_IPV4POOL_CIDR:                10.244.0.0/16
      CALICO_IPV4POOL_IPIP:                Never
      CALICO_IPV4POOL_BLOCK_SIZE:          26
      CALICO_IPV4POOL_NODE_SELECTOR:       all()
      CALICO_IPV4POOL_DISABLE_BGP_EXPORT:  false
      CALICO_IPV6POOL_CIDR:                2001:1234:5678:9a40::/58
```
```
root@zhecheng-206-1-mp-0000000:/# ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host
       valid_lft forever preferred_lft forever
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc prio state UP group default qlen 1000
    link/ether 00:0d:3a:8a:40:7b brd ff:ff:ff:ff:ff:ff
    inet 10.1.0.4/16 metric 100 brd 10.1.255.255 scope global eth0
       valid_lft forever preferred_lft forever
    inet6 fe80::20d:3aff:fe8a:407b/64 scope link
       valid_lft forever preferred_lft forever
```
From portal, the network interface of the VMSS VM only has IPv4
```
        "ipConfigurations": [
            {
                "name": "private-ipConfig-0",
                "id": "/subscriptions/xxxx/resourceGroups/zhecheng-208/providers/Microsoft.Compute/virtualMachineScaleSets/zhecheng-208-mp-0/virtualMachines/0/networkInterfaces/zhecheng-208-mp-0-0/ipConfigurations/private-ipConfig-0",
                "etag": "W/\"740ec70b-1861-4498-9c4c-8d139220b455\"",
                "properties": {
                    "provisioningState": "Succeeded",
                    "privateIPAddress": "10.1.0.4",
                    "privateIPAllocationMethod": "Dynamic",
                    "subnet": {
                        "id": "/subscriptions/xxx/resourceGroups/zhecheng-208/providers/Microsoft.Network/virtualNetworks/zhecheng-208-vnet/subnets/node-subnet"
                    },
                    "primary": true,
                    "privateIPAddressVersion": "IPv4",
                    "loadBalancerBackendAddressPools": [
                        {
                            "id": "/subscriptions/xxx/resourceGroups/zhecheng-208/providers/Microsoft.Network/loadBalancers/zhecheng-208/backendAddressPools/zhecheng-208-outboundBackendPool"
                        }
                    ]
                }
            }
        ],
```
More discussion in a PR: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2154#issuecomment-1418840357
A VMSS dual-stack cluster.

**Environment:**

- Kubernetes version: (use `kubectl version`): 1.25.3

CecileRobertMichon:
**What type of PR is this?**

/kind feature

Fixes #3163 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Add ability to provision IPv6 IPs for AzureMachinePools
```

### DIFF
--- a/azure/services/scalesets/scalesets.go
+++ b/azure/services/scalesets/scalesets.go
@@ -616,14 +616,24 @@ func (s *Service) getVirtualMachineScaleSetNetworkConfiguration(vmssSpec azure.S
 				},
 			}
 
-			ipconfig.Subnet = &compute.APIEntityReference{
-				ID: pointer.String(azure.SubnetID(s.Scope.SubscriptionID(), vmssSpec.VNetResourceGroup, vmssSpec.VNetName, n.SubnetName)),
-			}
 			if j == 0 {
 				// Always use the first IPConfig as the Primary
 				ipconfig.Primary = pointer.Bool(true)
 			}
 			ipconfigs = append(ipconfigs, ipconfig)
+		}
+		if vmssSpec.IPv6Enabled {
+			ipv6Config := compute.VirtualMachineScaleSetIPConfiguration{
+				Name: pointer.String("ipConfigv6"),
+				VirtualMachineScaleSetIPConfigurationProperties: &compute.VirtualMachineScaleSetIPConfigurationProperties{
+					PrivateIPAddressVersion: compute.IPVersionIPv6,
+					Primary:                 pointer.Bool(false),
+					Subnet: &compute.APIEntityReference{
+						ID: pointer.String(azure.SubnetID(s.Scope.SubscriptionID(), vmssSpec.VNetResourceGroup, vmssSpec.VNetName, n.SubnetName)),
+					},
+				},
+			}
+			ipconfigs = append(ipconfigs, ipv6Config)
 		}
 		if i == 0 {
 			ipconfigs[0].LoadBalancerBackendAddressPools = &backendAddressPools
--- a/templates/test/ci/cluster-template-prow-dual-stack.yaml
+++ b/templates/test/ci/cluster-template-prow-dual-stack.yaml
@@ -252,3 +252,75 @@ spec:
       - mv /etc/resolv.conf /etc/resolv.conf.OLD && ln -s /run/systemd/resolve/resolv.conf
         /etc/resolv.conf
       - systemctl restart systemd-resolved
+---
+apiVersion: cluster.x-k8s.io/v1beta1
+kind: MachinePool
+metadata:
+  name: ${CLUSTER_NAME}-mp-0
+  namespace: default
+spec:
+  clusterName: ${CLUSTER_NAME}
+  replicas: ${WORKER_MACHINE_COUNT}
+  template:
+    spec:
+      bootstrap:
+        configRef:
+          apiVersion: bootstrap.cluster.x-k8s.io/v1beta1
+          kind: KubeadmConfig
+          name: ${CLUSTER_NAME}-mp-0
+      clusterName: ${CLUSTER_NAME}
+      infrastructureRef:
+        apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
+        kind: AzureMachinePool
+        name: ${CLUSTER_NAME}-mp-0
+      version: ${KUBERNETES_VERSION}
+---
+apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
+kind: AzureMachinePool
+metadata:
+  name: ${CLUSTER_NAME}-mp-0
+  namespace: default
+spec:
+  location: ${AZURE_LOCATION}
+  strategy:
+    rollingUpdate:
+      deletePolicy: Oldest
+      maxSurge: 25%
+      maxUnavailable: 1
+    type: RollingUpdate
+  template:
+    osDisk:
+      diskSizeGB: 30
+      managedDisk:
+        storageAccountType: Premium_LRS
+      osType: Linux
+    sshPublicKey: ${AZURE_SSH_PUBLIC_KEY_B64:=""}
+    vmSize: ${AZURE_NODE_MACHINE_TYPE}
+---
+apiVersion: bootstrap.cluster.x-k8s.io/v1beta1
+kind: KubeadmConfig
+metadata:
+  name: ${CLUSTER_NAME}-mp-0
+  namespace: default
+spec:
+  files:
+  - contentFrom:
+      secret:
+        key: worker-node-azure.json
+        name: ${CLUSTER_NAME}-md-0-azure-json
+    owner: root:root
+    path: /etc/kubernetes/azure.json
+    permissions: "0644"
+  joinConfiguration:
+    nodeRegistration:
+      kubeletExtraArgs:
+        azure-container-registry-config: /etc/kubernetes/azure.json
+        cloud-config: /etc/kubernetes/azure.json
+        cloud-provider: azure
+        cluster-dns: '[fd00::10]'
+      name: '{{ ds.meta_data["local_hostname"] }}'
+  postKubeadmCommands:
+  - echo "DNSStubListener=no" >> /etc/systemd/resolved.conf
+  - mv /etc/resolv.conf /etc/resolv.conf.OLD && ln -s /run/systemd/resolve/resolv.conf
+    /etc/resolv.conf
+  - systemctl restart systemd-resolved
--- a/templates/test/ci/prow-dual-stack/kustomization.yaml
+++ b/templates/test/ci/prow-dual-stack/kustomization.yaml
@@ -3,6 +3,7 @@ kind: Kustomization
 namespace: default
 resources:
   - ../../../flavors/dual-stack
+  - machine-pool-dualstack.yaml
 patchesStrategicMerge:
   - ../patches/tags.yaml
   - ../patches/controller-manager.yaml
--- a/templates/test/ci/prow-dual-stack/machine-pool-dualstack.yaml
+++ b/templates/test/ci/prow-dual-stack/machine-pool-dualstack.yaml
@@ -1,0 +1,71 @@
+apiVersion: cluster.x-k8s.io/v1beta1
+kind: MachinePool
+metadata:
+  name: ${CLUSTER_NAME}-mp-0
+  namespace: default
+spec:
+  clusterName: ${CLUSTER_NAME}
+  replicas: ${WORKER_MACHINE_COUNT}
+  template:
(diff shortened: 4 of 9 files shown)


templates/test/ci/cluster-template-prow-dual-stack.yaml
```diff
@@ -252,3 +252,75 @@ spec:
...
       - systemctl restart systemd-resolved
+---
+apiVersion: cluster.x-k8s.io/v1beta1
+kind: MachinePool
+metadata:
+  name: ${CLUSTER_NAME}-mp-0
+  namespace: default
+spec:
+  clusterName: ${CLUSTER_NAME}
+  replicas: ${WORKER_MACHINE_COUNT}
```
lzhecheng, line 263: Maybe I have asked this question before: Should we use 2 variables for MachineDeployment and MachinePool in one template? If WORKER_MACHINE_COUNT == 2, we will have 4 Nodes in total.
It is good to have a template supporting both kinds of Machine but I think we can have a vmss only cluster, right?

CecileRobertMichon: Yes, that's a good point. The `WORKER_MACHINE_COUNT` really means how many replicas per MachineDeployment or per MachinePool. Because even if it's VMSS only, you could have two or more MachinePools, each with `WORKER_MACHINE_COUNT` nodes. When the e2e tests wait for nodes to be Ready, they list all Machines, MachineDeployments, and MachinePools in the cluster so this number isn't used in any context as "total number of nodes" in any way (except maybe in ci-entrypoint.sh, but we should fix that). 

Having one env var per node group is not scalable. What if we wanted to test scale with 10,20, 50 MachinePools in one test template? So I think maybe we should rename that var to something more explicit like `REPLICAS_PER_POOL`, wdyt? That would be out of scope for this specific PR but something we can do separately.

> we can have a vmss only cluster

the main reason I'm putting them together is to reduce costs since we run lots of tests. In the future I think we want to move more towards a model where a single cluster is used to test several "node" features by using different node groups instead of separate clusters.

CecileRobertMichon:
@lzhecheng now that I've rebased on top of #3221 and rerun the test, the LB test is still failing

templates/test/ci/cluster-template-prow-dual-stack.yaml
lzhecheng: > Having one env var per node group is not scalable. What if we wanted to test scale with 10,20, 50 MachinePools in one test template? So I think maybe we should rename that var to something more explicit like `REPLICAS_PER_POOL`, wdyt? That would be out of scope for this specific PR but something we can do separately.
Yes, renaming is a good idea. `REPLICAS_PER_POOL` seems to refer to MachinePool and how about `REPLICAS_PER_NODE_GROUP`?

> the main reason I'm putting them together is to reduce costs since we run lots of tests. In the future I think we want to move more towards a model where a single cluster is used to test several "node" features by using different node groups instead of separate clusters.
Yes, that's a good idea to reduce testing time.

lzhecheng:
The dualstack template has fields like `userAssignedIdentities` but IPv6 one doesn't but I thought they are similar. Is it expected?

CecileRobertMichon:
> The dualstack template has fields like userAssignedIdentities but IPv6 one doesn't but I thought they are similar. Is it expected?

I don't think that should make a difference, I believe someone reused the dual stack template to test userAssigned identities since it's an optional test. We should probably change it to a seperate node but shouldn't affect this PR.

lzhecheng:
@CecileRobertMichon could you please try again with CCM v1.25.12 instead of v1.25.11?

CecileRobertMichon:
@lzhecheng the v1.25.12 release is not part of the Helm chart, we need https://github.com/kubernetes-sigs/cloud-provider-azure/pull/3818 to merge first so we can test with it

lzhecheng:
> @lzhecheng the v1.25.12 release is not part of the Helm chart, we need [kubernetes-sigs/cloud-provider-azure#3818](https://github.com/kubernetes-sigs/cloud-provider-azure/pull/3818) to merge first so we can test with it

Got it. Thank you.

lzhecheng:
@CecileRobertMichon it works.

azure/services/scalesets/scalesets.go
```diff
@@ -616,15 +616,25 @@ func (s *Service) getVirtualMachineScaleSetNetworkConfiguration(vmssSpec azure.S
 				},
 			}
 
-			ipconfig.Subnet = &compute.APIEntityReference{
-				ID: pointer.String(azure.SubnetID(s.Scope.SubscriptionID(), vmssSpec.VNetResourceGroup, vmssSpec.VNetName, n.SubnetName)),
```
CecileRobertMichon, line 620: removed this because it was redundant (it's already being set above on line 614)

CecileRobertMichon (commented):
This is now ready (cloud-provider issue is fixed and part of the latest patch releases) 

/assign @jackfrancis @nawazkh

templates/test/ci/prow-dual-stack/machine-pool-dualstack.yaml
```diff
@@ -0,0 +1,71 @@
...
+        azure-container-registry-config: /etc/kubernetes/azure.json
+        cloud-config: /etc/kubernetes/azure.json
+        cloud-provider: azure
+        cluster-dns: '[fd00::10]'
+      name: '{{ ds.meta_data["local_hostname"] }}'
+  postKubeadmCommands:
+  - echo "DNSStubListener=no" >> /etc/systemd/resolved.conf
+  - mv /etc/resolv.conf /etc/resolv.conf.OLD && ln -s /run/systemd/resolve/resolv.conf
+    /etc/resolv.conf
+  - systemctl restart systemd-resolved
```
nawazkh, line 71: We may want to add a new line at the end of file?

templates/test/ci/prow-dual-stack/kustomization.yaml
```diff
@@ -3,6 +3,7 @@ kind: Kustomization
 namespace: default
 resources:
   - ../../../flavors/dual-stack
+  - machine-pool-dualstack.yaml
```
nawazkh, line 6: Wondering if we could maintain `machine-pool-dualstack.yaml` in `/templates/flavors/dual-stack/` so there will be one less file to maintain? Or is this file exclusive for prow ?

nawazkh (commented):
A couple of thoughts, else this looks good to me!

templates/test/ci/prow-dual-stack/kustomization.yaml
CecileRobertMichon: the machine pool dual stack combination is only used in prow for testing so we get test coverage for this specific combination of features (the dual stack flavor just has machinedeployments, since machinepools are experimental and behind a feature flag)

templates/test/ci/prow-ipv6/kustomization.yaml
```diff
@@ -3,6 +3,7 @@ kind: Kustomization
 namespace: default
 resources:
   - ../../../flavors/ipv6
+  - machine-pool-ipv6.yaml
```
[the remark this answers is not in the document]
CecileRobertMichon: same as above

templates/test/ci/prow-dual-stack/machine-pool-dualstack.yaml
CecileRobertMichon: added

mboersma (approved):
/lgtm
/approve

PR #3365, opened 2023-03-27
Default set NATGatway for outbound traffic if cluster is not using IPv6

### LINKED ISSUE #1532: Use only nat gateway for outbound connections

xiujuanx:
**What type of PR is this?**
/kind feature

This PR sets the default subnet to be NAT gateway if the cluster is not using IPv6. It both sets AzureCluster and AzureClusterTemplate.

Fixes #1532 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Default use the NAT gateway for the node outbound connection if cluster is not using IPv6
```


api/v1beta1/azurecluster_default.go
```diff
@@ -128,6 +130,12 @@ func (c *AzureCluster) setSubnetDefaults() {
 			if subnet.NatGateway.NatGatewayIP.Name == "" {
 				subnet.NatGateway.NatGatewayIP.Name = generateNatGatewayIPName(c.ObjectMeta.Name, subnet.Name)
 			}
+		} else {
+			if !subnet.IsIPv6Enabled() {
```
willie-yao, line 134: Should this if statement be outside of the `else` block? Looking through the issue details, it looks like the intended behavior is to set NAT Gateway by default if not using IPv6, regardless of whether a name has been provided or not.

api/v1beta1/azurecluster_default.go
```diff
@@ -146,6 +154,11 @@ func (c *AzureCluster) setSubnetDefaults() {
 			RouteTable: RouteTable{
 				Name: generateNodeRouteTableName(c.ObjectMeta.Name),
 			},
+			NatGateway: NatGateway{
+				NatGatewayClassSpec: NatGatewayClassSpec{
+					Name: fmt.Sprintf("%s-%s", c.ObjectMeta.Name, "node-natgw"),
```
willie-yao, line 159: Can you move the `Sprintf` to its own helper function `generateNatGatewayName` to match the pattern above?

api/v1beta1/types.go
```diff
@@ -733,6 +735,23 @@ func (s SubnetSpec) IsNatGatewayEnabled() bool {
 	return s.NatGateway.Name != ""
 }
 
+// IsIPv6Enabled returns whether or not IPv6 is enabled on the subnet.
+func (s SubnetSpec) IsIPv6Enabled() bool {
+	isIpv6Enabled := false
+	for _, cidr := range s.CIDRBlocks {
```
willie-yao, line 741: You can use `net.IsIPv6CIDRString(cidr)` instead of the current codeblock to avoid repeated code.

Example here: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/scope/cluster.go#L562

api/v1beta1/types_template.go
```diff
@@ -106,6 +110,23 @@ func (s SubnetTemplateSpec) IsNatGatewayEnabled() bool {
 	return s.NatGateway.Name != ""
 }
 
+// IsIPv6IsIPv6Enabled returns whether or not IPv6 is enabled on the subnet.
+func (s SubnetTemplateSpec) IsIPv6Enabled() bool {
+	isIpv6Enabled := false
+	for _, cidr := range s.CIDRBlocks {
```
willie-yao, line 116: Same as above https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3365/files#r1149676880

api/v1beta1/azurecluster_default.go
xiujuanx: Thanks @willie-yao for your review! I think put the statement be outside of the `else` block is ok. 
For a subnet, if the NAT gateway name is provided, this means NAT gateway has beed enabled, we don't need to set it again.  If the NAT gateway name doesn't be provided, then I will go to check if it's using IPv6, if not, then I will set NAT gateway by default, otherwise, I will keep the original subnet settings and do nothing.

api/v1beta1/azurecluster_default.go
xiujuanx: Sure, this is a good idea!

api/v1beta1/types.go
xiujuanx: Done as suggestion!

api/v1beta1/azurecluster_default.go
willie-yao: Ah that makes sense, thanks for the detailed explanation!

azure/scope/cluster_test.go
```diff
@@ -2005,7 +2005,7 @@ func TestOutboundLBName(t *testing.T) {
 			clusterName: "my-cluster",
 			name:        "public cluster node outbound lb",
 			role:        "node",
-			expected:    "my-cluster",
+			expected:    "",
```
willie-yao, line 2008: Just for my own understanding: Why is this change needed?

api/v1beta1/azureclustertemplate_default_test.go
```diff
@@ -354,6 +360,9 @@ func TestSubnetsTemplateDefaults(t *testing.T) {
 											Role:       SubnetNode,
 											CIDRBlocks: []string{DefaultNodeSubnetCIDR},
 										},
+										NatGateway: NatGatewayClassSpec{
+											Name: DefaultNATGateway,
+										},
```
willie-yao, line 365: Can you add a new test case for when both IPv6 and NatGateway is enabled? This way, you can cover the case of setting the NatGateway defaults with a Subnet

api/v1beta1/azurecluster_default_test.go
```diff
@@ -683,6 +710,14 @@ func TestSubnetDefaults(t *testing.T) {
...
 								SecurityGroup: SecurityGroup{Name: "cluster-test-node-nsg"},
 								RouteTable:    RouteTable{Name: "cluster-test-node-routetable"},
+								NatGateway: NatGateway{
+									NatGatewayIP: PublicIPSpec{
+										Name: "",
+									},
+									NatGatewayClassSpec: NatGatewayClassSpec{
+										Name: "cluster-test-node-natgw",
+									},
+								},
```
willie-yao, line 720: Can you add test case where the NatGateway Name is set but the subnet is using IPv6 and validate that the defaults are not being set? That way we can test the negative case of this [line](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3365/files#diff-016ca697ac4323c794d5dda09a16fd6152518acb4e3da6efef765fe5011a796bR134)

api/v1beta1/azureclustertemplate_default_test.go
xiujuanx: Thanks @willie-yao for bringing this up! NatGateway is not supported by IPv6 cluster, so I am worry if this is an valid case or not where NatGateway and IPv6 both are enabled.
BTW, I think below case: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/81f88115494b6c390621641e6074a760877b4074/api/v1beta1/azureclustertemplate_default_test.go#L381-L436 has covered the negative test for IPv6. This is a cluster enables IPv6, so NatGateway will not be set for AzureClusterTemplate.

api/v1beta1/azurecluster_default_test.go
xiujuanx: Same concern with above, and the below case: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/81f88115494b6c390621641e6074a760877b4074/api/v1beta1/azurecluster_default_test.go#L571-L636 has covered the negative test for IPv6.

azure/scope/cluster_test.go
xiujuanx: BTW, this also is the reason why the `loadBalancerName` is changed from `foo` to `""` for test cases in controllers/helpers_test.go file.

willie-yao:
/lgtm

Apologies for the delay. Thanks for getting back to me about the tests! This looks great and thanks for the great work on this @xiujuanx

xiujuanx:
@willie-yao Thanks for your kind review! Does this pr can be approved now?

willie-yao:
@xiujuanx Yes, pinging approvers for a review. @CecileRobertMichon @mboersma @jackfrancis

api/v1beta1/types_class.go
```diff
@@ -160,3 +160,8 @@ func (sgc *SecurityGroupClass) setDefaults() {
 		}
 	}
 }
+
+// setDefaults sets default values for NatGatewayClass.
+func (natgaw *NatGatewayClassSpec) setDefaults() {
+	natgaw.Name = DefaultNATGateway
```
CecileRobertMichon, line 166: what is the "scope" of a NAT gateway name in Azure? I'm wondering if it makes sense to default the name to a hardcoded value for all clusters in a clusterclass. Notice how the other resources (Subnet, VNet, etc) are not getting a default name value since they need to have unique names withing a resource group, in some cases even within a subscription.

api/v1beta1/azurecluster_default.go
```diff
@@ -128,6 +130,12 @@ func (c *AzureCluster) setSubnetDefaults() {
 			if subnet.NatGateway.NatGatewayIP.Name == "" {
 				subnet.NatGateway.NatGatewayIP.Name = generateNatGatewayIPName(c.ObjectMeta.Name, subnet.Name)
 			}
+		} else {
+			if !subnet.IsIPv6Enabled() {
+				// Keep LB support for IPv6 clusters, otherwise use NAT Gateway for outbound traffic
```
CecileRobertMichon, line 135: would be good to specify why in the comment, i.e. `NAT gateway supports the use of IPv4 public IP addresses for outbound connectivity `

api/v1beta1/azurecluster_default.go
CecileRobertMichon: instead of adding an else condition here with another nested if loop, have you considered directly modifying the conditions for which IsNatGatewayEnabled returns true? and setting the name in NAT gateway name in the block above when it's not already set. This would avoid duplicate code of setting the IP name and nested if/else statements.

CecileRobertMichon (commented):
Thanks for this PR @xiujuanx! Looks good overall, just some comments around default naming and logic.

With this change, we should update the [docs](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/docs/book/src/topics/node-outbound-lb.md) on outbound since the default is no longer outbound load balancer for outbound, and NAT Gateway is no longer opt-in. We should also document the ipv6 exception.

In addition, could you please update the example templates in the repo to no longer explicitly define a nat gateway? This will also allow us to test the default in our e2e tests. You can do so by updating the [base](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/templates/flavors/base/cluster-template.yaml#L34) template.

api/v1beta1/azurecluster_default.go
xiujuanx: @CecileRobertMichon Thanks for your review! Indeed, there has a duplicate code for setting NAT gateway IP name. But if I modify the condition of IsNatGatewayEnabled and set the name of NAT gateway name if it's not set yet in this func, then I need to call generateNatGatewayName func in type.go file. This will cause an inconsistency from current CAPZ codes in main branch. Although this does not cause any logical problems, it just looks less elegant formally. So I prefer to judge if this is a IPv6 cluster first, then go to judge if IsNatGatewayEnabled or not. This will also avoid duplication. Does this make sense?

api/v1beta1/types_class.go
xiujuanx: Sorry, I didn't realize the NAT gateway should be unique. Thanks for your kind reminder!
>what is the "scope" of a NAT gateway name in Azure?

This NAT gateway name is for `AzureClusterTemplate` object.
Now, I have removed the `DefaultNATGateway` variable and generate a unique name by default.

xiujuanx:
>With this change, we should update the [docs](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/docs/book/src/topics/node-outbound-lb.md) on outbound since the default is no longer outbound load balancer for outbound, and NAT Gateway is no longer opt-in. We should also document the ipv6 exception.

@CecileRobertMichon 
Because the NAT gateway will be used for node outbound connection instead of load balancer, I change the node outbound doc name to be node-outbound-connection.
And I go through related [NAT gateway](https://learn.microsoft.com/en-us/azure/virtual-network/nat-gateway/nat-overview) documents in azure website and codes in CAPZ, look like no matter a public cluster or private cluster supports NAT gateway excepts the IPv6 cluster. So I update this doc from the perspective of IPv6 and IPv4 respectively instead of public cluster and private cluster. 

>In addition, could you please update the example templates in the repo to no longer explicitly define a nat gateway? This will also allow us to test the default in our e2e tests. You can do so by updating the [base](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/templates/flavors/base/cluster-template.yaml#L34) template.

Updated.

Codes have been updated as your comments, so pls help review again, thanks!

api/v1beta1/azurecluster_default.go
```diff
@@ -124,10 +124,14 @@ func (c *AzureCluster) setSubnetDefaults() {
...
-			if subnet.NatGateway.NatGatewayIP.Name == "" {
-				subnet.NatGateway.NatGatewayIP.Name = generateNatGatewayIPName(c.ObjectMeta.Name, subnet.Name)
+
+		if !subnet.IsIPv6Enabled() {
+			// NAT gateway supports the use of IPv4 public IP addresses for outbound connectivity.
+			// So default use the NAT gateway for outbound traffic in IPv4 cluster instead of loadbalancer.
+			if !subnet.IsNatGatewayEnabled() {
+				subnet.NatGateway.Name = withIndex(generateNatGatewayName(c.ObjectMeta.Name), nodeSubnetCounter)
 			}
+			subnet.NatGateway.NatGatewayIP.Name = generateNatGatewayIPName(c.ObjectMeta.Name, subnet.Name)
```
CecileRobertMichon, line 134: this should still be inside an if block `if subnet.NatGateway.NatGatewayIP.Name == "" {` so we don't overwrite a user custom value

api/v1beta1/azurecluster_default.go
```diff
@@ -124,10 +124,14 @@ func (c *AzureCluster) setSubnetDefaults() {
...
 			subnet.RouteTable.Name = generateNodeRouteTableName(c.ObjectMeta.Name)
 		}
-		if subnet.IsNatGatewayEnabled() {
-			if subnet.NatGateway.NatGatewayIP.Name == "" {
-				subnet.NatGateway.NatGatewayIP.Name = generateNatGatewayIPName(c.ObjectMeta.Name, subnet.Name)
+
+		if !subnet.IsIPv6Enabled() {
+			// NAT gateway supports the use of IPv4 public IP addresses for outbound connectivity.
+			// So default use the NAT gateway for outbound traffic in IPv4 cluster instead of loadbalancer.
+			if !subnet.IsNatGatewayEnabled() {
```
CecileRobertMichon, line 131: this could just be 

`if subnet.NatGateway.Name == ""` 

the `IsNatGatewayEnabled` can probably be removed, or just changed to 

```
func (s SubnetSpec) IsNatGatewayEnabled() bool {
  return !s.IsIPv6Enabled()
}

File: api/v1beta1/types_class.go [hunk 12]
```diff
@@ -160,3 +160,8 @@ func (sgc *SecurityGroupClass) setDefaults() {
 		}
 	}
 }
+
+// setDefaults sets default values for NatGatewayClass.
+func (natgaw *NatGatewayClassSpec) setDefaults(clusterTemplateName string, nodeSubnetCounter int) {
+	natgaw.Name = withIndex(generateNatGatewayName(clusterTemplateName), nodeSubnetCounter)
```
CecileRobertMichon, line 166: `clusterTemplateName` is still not unique enough since it will be shared between clusters using the same template. That being said, I don't think we need to do the defaulting here in the ClusterClass spec. It seems that it's similar to Subnet Name in that Subnet.Name is a required field but it doesn't get defaulted in the template itself, it only gets defaulted per cluster in `azurecluster_default.go` using the cluster name as input to generate a subnet name. @willie-yao does this make sense to you?

willie-yao: Yes that makes sense. The name shouldn't be shared across clusters in the ClusterClass. The defaulting here shouldn't be needed as it is happening in [azurecluster_default](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3365/files#diff-016ca697ac4323c794d5dda09a16fd6152518acb4e3da6efef765fe5011a796bR132).

api/v1beta1/azurecluster_default.go
xiujuanx: @CecileRobertMichon I change it ti be `if subnet.NatGateway.Name == ""`.

api/v1beta1/azurecluster_default.go
xiujuanx: That's true, thanks!

api/v1beta1/types_class.go
xiujuanx: Thanks @CecileRobertMichon and @willie-yao ! Now, I have reverted default setting of NAT gateway for `AzureClusterTemplate` object. And during validation phase, use `IsIPv6Enabled` instead of `IsNatGatewayEnabled` to judge if we need the outboundLB or not. Please help review again, thanks again!

CecileRobertMichon (commented):
/retest 

#3403 should fix the broken test

/lgtm
/approve

Thanks @xiujuanx, great work!

sonasingh46:
Late to party on this PR. 
What happens in upgrade scenario when the cluster already have chosen a load balancer. Does it silently ignore it?

CecileRobertMichon:
/hold

to answer @sonasingh46's question above

CecileRobertMichon:
> What happens in upgrade scenario when the cluster already have chosen a load balancer. Does it silently ignore it?

This is only changing the webhook default so it shouldn't affect existing clusters AFAIK

willie-yao (approved):
/lgtm

Thanks again for all the great work on this PR! 🚀

sonasingh46:
@CecileRobertMichon -- Thanks. I am just thinking about the following scenario. 
- Prior to this change, a user can have an existing cluster that uses load balancer because of the defaults. If NAT gateway is not specified by default then load balancer is chosen.

- With this change, if NAT gateway is not specified and if it is not IPV6 then it uses NAT gateway by default now. 

- So as far as compatibility is concerned, nothing breaks and all works fine but I am thinking if we should have a way to migrate away from load balancer and choose NAT gateway. 

- Now, IIRC the migration was not possible earlier also so I think this discussion is out of scope of the PR.

CecileRobertMichon:
@sonasingh46 I think the migration would be possible but the user would have to opt in (i.e. edit the resource and specify a NAT), have not tested this though.

agree it's out of scope for this PR

/hold cancel

PR #3368, opened 2023-03-27
Run codecov job on every PR with threshold of 1%

### LINKED ISSUE #3312: Run codecov job for each PR and display message if code coverage is lower than before
/kind feature

As it was mentioned in this [issue comment](https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/3229#issuecomment-1470825135), it is good practice to use the previous code coverage % as a baseline. Currently, the code coverage job only runs on each new commit to `main`. It would be helpful to our code coverage practices to run the codecov job for each PR before it is merged, and displaying a reminder message if the code coverage is decreased. This way, we're not allowing regression of the quality of code for new contributions.

**Environment:**

willie-yao:
**What type of PR is this?**
/kind cleanup

This PR enables the codecov job to run on every PR. If the PR decreases the overall code coverage by more than 1%, the coverage report will not pass. The test will still be green even if the code coverage difference is above 1%. This is done to make it visible enough for the author to pay attention to how much code coverage is being decreased without blocking the PR.

This is part of a larger effort to improve code coverage in CAPZ.

Fixes #3312

This test PR on my fork of CAPZ shows an example code coverage report on a PR: https://github.com/willie-yao/cluster-api-provider-azure/pull/1

**Important**: This PR requires the GitHub codecov app to be added to CAPZ https://github.com/apps/codecov

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,0 +1,7 @@
+coverage:
+  status:
+    project:
+      default:
+        target: auto
+        threshold: 1%
+        informational: true
--- a/.github/workflows/cover.yaml
+++ b/.github/workflows/cover.yaml
@@ -2,6 +2,7 @@
 
 name: codecov
 on:
+  pull_request:
   push:
     branches:
     - main


CecileRobertMichon:
> If the PR decreases the overall code coverage by more than 1%, the job is marked as failed.

Do we really want to fail the job? I thought we were going to start with only displaying it (not enforcing it) based on the discussion in https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/3312

willie-yao:
@CecileRobertMichon That's correct. I misunderstood and thought that marking it as failed but making it non-blocking would be fine, but I think that's still too excessive and shows a check as failed. I just set `if_ci_failed: success` in `codecov.yml` so it'll show as passed.

.github/codecov.yml
```diff
@@ -0,0 +1,7 @@
+coverage:
+  status:
+    project:
+      default:
+        target: auto
+        threshold: 1%
+        if_ci_failed: success
```
nawazkh, line 7: Does `if_ci_failed: success` mean that the PR will be marked successful even if the CI Failed?
Also, by `success` will the PR be marked green? Or is the `success` status exclusive to `Codecov` ?

.github/codecov.yml
```diff
@@ -0,0 +1,7 @@
+coverage:
+  status:
+    project:
+      default:
+        target: auto
+        threshold: 1%
```
nawazkh, line 6: > `threshold: 1%`

Will there ever be scenarios where the PR has dropped the code coverage by more than `1%` and codecov marks that PR (or commit? sorry, I dont understand the `success` status) a failed?

.github/codecov.yml
nawazkh: Also, do you think we could configure codecov for [removed_code_behavior](https://docs.codecov.com/docs/commit-status#removed_code_behavior) ? Just bouncing ideas.

nawazkh (commented):
Some codecov related beginner thoughts..

.github/codecov.yml
willie-yao: Yes it means that the PR will be marked successful but only if codecov fails. This is a part of `codecov.yml` so it shouldn't affect other CI jobs.

willie-yao: > Also, do you think we could configure codecov for [removed_code_behavior](https://docs.codecov.com/docs/commit-status#removed_code_behavior) ? Just bouncing ideas.

Ohh yeah that's a good point! I think the default option of `fully_covered_patch` would be good in our case as it only considers the patch, and I don't think we should factor in removed code as increased test coverage.

.github/codecov.yml
willie-yao: Codecov will not mark the PR as failed, but the codecov report will show the threshold being exceeded. We don't want to mark a PR as failed or blocked due to code coverage, but make it visible enough for the author to pay attention to how much code coverage is being decreased.

nawazkh: >  make it visible enough for the author to pay attention to how much code coverage is being decreased

Understood, thanks!

.github/codecov.yml
nawazkh: > Yes it means that the PR will be marked successful but only if codecov fails.

I am not sure if I a getting this right, what are the scenarios where the code coverage test could fail? 
I am wondering if codecov's CI implies one of the CI providers as listed in here https://docs.codecov.com/docs/supported-ci-providers ?

nawazkh: Also, I am not sure what category does our CI fall into? Is it `prow` or `GitHub Actions`? cc: @CecileRobertMichon @nojnhuh

willie-yao:
/hold

For some reason, `if_ci_failed: success` does not do its intended purpose and CI still shows as failed. Will take a look into this. Example here: https://github.com/willie-yao/cluster-api-provider-azure/pull/2

.github/codecov.yml
willie-yao: In this case, we're using GitHub actions as the CI provider. The code coverage test fails if the coverage diff is greater than the threshold of 1% like so: https://github.com/willie-yao/cluster-api-provider-azure/pull/2

Right now, I'm working on a fix to not mark the action as failed, but still display the message that the threshold has been exceeded.

willie-yao:
> I assume it just updates the existing comment when you push changes?
@mboersma Yup it does! Thanks for the review ❤️

PR #3369, opened 2023-03-28
Use MCR registry for Calico images

### LINKED ISSUE #3323: Scaling up lots of nodes at once causes docker.io rate limits for calico pods
/kind bug

Create a cluster with 1 MachinePool containing 1 replica. When everything is ready, scale up to 300 replicas.

Observe calico-node pods are in ContainerCreating state and some of them go to ImagePullBackoff:

```
Warning  Failed     5m27s                  kubelet            Failed to pull image "docker.io/calico/pod2daemon-flexvol:v3.25.0": rpc error: code = Unknown desc = failed to pull and unpack image "docker.io/calico/pod2daemon-flexvol:v3.25.0": failed to copy: httpReadSeeker: failed open: unexpected status code https://registry-1.docker.io/v2/calico/pod2daemon-flexvol/blobs/sha256:5a110750980fbe31eda12a1dba9e9a38a04f805a0bfe8292f38d707ce9576267: 429 Too Many Requests
```

Some ideas:

1) we could use mcr images instead. It seems those are available (?) and would not cause the same rate limiting. We already use mcr images for cloud-provider-azure.

2) we could create a custom VM image that have the calico container images pre-pulled. The main issue with this is that it would tie the VM image to the CNI and CNI version so a new image would need to be built for each new Calico version per k8s version - this is good enough for your own testing but not great for a generic solution that works for all users.

3) we could spend some time revisiting adapting [jackfrancis/kamino: Kubernetes node clone factory (github.com)](https://github.com/jackfrancis/kamino) to CAPZ so we can dynamically take snapshots of VM images and reuse it for new nodes so we don't pull each container image 300 times when scaling up to 300. Probably more of a longer-term effort.

4) As suggested by @mtougeron in office hours, we could require (or recommend) users to add an ImagePull secret to authenticate with a docker account which would raise the limits. It seems the Calico Helm chart has an [option](https://github.com/projectcalico/calico/blob/master/charts/tigera-operator/values.yaml#L1) for it.

**Environment:**

CecileRobertMichon:
**What type of PR is this?**

/kind cleanup

**What this PR does / why we need it**: This PR changes the image registry used to install Calico in examples and test scripts to MCR to allow testing scale without running into docker.io throttling. It also pins the Calico version to v3.25.1 so we can bump the Calico after the sig-windows tools and the MCR images become available whenever there is a new release.

Fixes #3323 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Use MCR registry for Calico images and bump Calico to v3.25.1
```

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -390,7 +390,7 @@ def get_addons(flavor_name):
     else:
         calico_values = "./templates/addons/calico/values.yaml"
 
-    addon_cmd = "; " + helm_cmd + " --kubeconfig ./${CLUSTER_NAME}.kubeconfig install --repo https://docs.tigera.io/calico/charts calico tigera-operator -f " + calico_values + " --namespace tigera-operator --create-namespace"
+    addon_cmd = "; " + helm_cmd + " --kubeconfig ./${CLUSTER_NAME}.kubeconfig install --repo https://docs.tigera.io/calico/charts --version ${CALICO_VERSION} calico tigera-operator -f " + calico_values + " --namespace tigera-operator --create-namespace"
 
     if "intree-cloud-provider" not in flavor_name:
         addon_cmd += "; " + helm_cmd + " --kubeconfig ./${CLUSTER_NAME}.kubeconfig install --repo https://raw.githubusercontent.com/kubernetes-sigs/cloud-provider-azure/master/helm/repo cloud-provider-azure --generate-name --set infra.clusterName=${CLUSTER_NAME}"
--- a/templates/addons/calico-dual-stack/values.yaml
+++ b/templates/addons/calico-dual-stack/values.yaml
@@ -18,12 +18,9 @@ installation:
       natOutgoing: Enabled
       nodeSelector: all()
   registry: mcr.microsoft.com/oss
-  tag: v3.25.1
 # Image and registry configuration for the tigera/operator pod.
 tigeraOperator:
   image: tigera/operator
   registry: mcr.microsoft.com/oss
-  tag: v1.29.3
 calicoctl:
   image: mcr.microsoft.com/oss/calico/ctl
-  tag: v3.25.1
--- a/templates/addons/calico-dual-stack/values.yaml
+++ b/templates/addons/calico-dual-stack/values.yaml
@@ -17,3 +17,13 @@ installation:
       encapsulation: None
       natOutgoing: Enabled
       nodeSelector: all()
+  registry: mcr.microsoft.com/oss
+  tag: v3.25.1
+# Image and registry configuration for the tigera/operator pod.
+tigeraOperator:
+  image: tigera/operator
+  registry: mcr.microsoft.com/oss
+  tag: v1.29.3
+calicoctl:
+  image: mcr.microsoft.com/oss/calico/ctl
+  tag: v3.25.1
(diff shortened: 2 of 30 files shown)


templates/addons/calico-dual-stack/values.yaml
```diff
@@ -17,3 +17,10 @@ installation:
       encapsulation: None
       natOutgoing: Enabled
       nodeSelector: all()
+  registry: mcr.microsoft.com/oss
```
CecileRobertMichon, line 20: this seems to work. However, we may need to pin the version as we are now dependent on the image becoming available, which may not always be right when a new release is published

CecileRobertMichon (commented):
@jackfrancis thoughts on this PR? It solves #3323 however since we are no longer using the default image registry from the Helm chart we need to pin to a specific version so it doesn't start failing whenever a new Calico version is released if the image doesn't exist in MCR yet

jackfrancis:
@CecileRobertMichon I think pinning to a particular version of calico (and occasionally updating the version) is a fine way of demonstrating "known-working CNI", and for supporting E2E coverage. We did this for a long time and it worked well. I think reverting back to that pattern offers only a small loss of convenience.

templates/addons/calico-dual-stack/values.yaml
jackfrancis: It sure would be nice if we could include the chart version here so we could maintain it in one place....

Should we maybe define a Makefile env var that declares the version and then wire that reference in rather than having to update the same string in a bunch of places (and maybe missing some)?

CecileRobertMichon: I tried but Calico doesn't have the version as part of the values unfortunately :( 

> Should we maybe define a Makefile env var that declares the version and then wire that reference in rather than having to update the same string in a bunch of places (and maybe missing some)?

I agree that would be much better but not sure how feasible that it since all the places are different languages (md for docs, go for e2e test helpers, bash for test scripts, Tiltfile, etc.). I can try.

CecileRobertMichon:
@Jont828 is pinning to a specific version of the Helm chart something we'd be able to do with CAAPH (in #3571)?

Tiltfile
```diff
@@ -390,7 +390,7 @@ def get_addons(flavor_name):
     else:
         calico_values = "./templates/addons/calico/values.yaml"
 
-    addon_cmd = "; " + helm_cmd + " --kubeconfig ./${CLUSTER_NAME}.kubeconfig install --repo https://docs.tigera.io/calico/charts calico tigera-operator -f " + calico_values + " --namespace tigera-operator --create-namespace"
+    addon_cmd = "; CALICO_VERSION=$(make get-calico-version) " + helm_cmd + " --kubeconfig ./${CLUSTER_NAME}.kubeconfig install --repo https://docs.tigera.io/calico/charts --version ${CALICO_VERSION} calico tigera-operator -f " + calico_values + " --namespace tigera-operator --create-namespace"
```
jackfrancis, line 393: Is this a new pattern for storing env vars in one place? (Makefile)

CecileRobertMichon: Is this not what you meant by https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3369/files#r1199457600 ?

templates/addons/calico-dual-stack/values.yaml
CecileRobertMichon: ok was able to reuse the existing Makefile variable everywhere but the docs, PTAL

CecileRobertMichon:
Broken link job fixed by https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3598

Jont828:
@CecileRobertMichon Yes we should be able to pin to a specific chart version in CAAPH -- it only installs the latest version if it's omitted.

marosset (commented):
/lgtm

Jont828 (commented):
/lgtm

PR #3378, opened 2023-03-30
[DO_NOT_MERGE][DO_NOT_REVIEW] PR capi node labels fix test

ykakarap:
**What type of PR is this?**

This PR is to test a CAPI patch: https://github.com/kubernetes-sigs/cluster-api/pull/8427

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note

```


ykakarap:
The `pull-cluster-api-provider-azure-e2e-optional` job is stuck in pending for a long time. I will rebase and force push to see if that will restart the job.

ykakarap:
The machine should drop the taint from the node almost immediately after infra provider reports `.spec.providerID`. 
Could it be possible that the scheduler is going into some exponential back off if it cannot find a "good node" and the scheduler is not re-evaluating nodes soon enough after the taint is dropped?

> Maybe wait for nodes to be initialized before entering that part of the test suite?

Sounds like a good idea. In such cases the pods should be scheduled almost immediately as the taint should be dropped at this point.

CecileRobertMichon:
@ykakarap I think the issue might be that there's a delay in MachinePools because the MachinePool controller is not watching cluster nodes, unlike the Machine controller.

ykakarap:
@CecileRobertMichon I think you are right. There are not enough events that trigger the machinepool reconciler. It is only watching for changes on `Cluster` and `MachinePool` as opposed to Machine controller which is reconciled a lot more (as it also reconciles when there is a change on the `Node`). 


I was not able to force any delay in dropping the taints (tested locally) when using Machines, even when I delay the `.spec.providerID` form being set. So, I think it is not an exponential back-off problem. 

Can we move the "slow MachinePool reconciliation" into a separate issue and get the current fix for syncing node labels merged?

CecileRobertMichon:
> Can we move the "slow MachinePool reconciliation" into a separate issue and get the current fix for syncing node labels merged?

yes, I'm working on a fix

CecileRobertMichon:
pass 1/1

CecileRobertMichon:
pass 1/1

CecileRobertMichon:
pass 2/2

CecileRobertMichon:
pass 3/3

CecileRobertMichon:
pass 4/4

CecileRobertMichon:
5/5 passes 🎉

CecileRobertMichon:
Test with both node watcher + node patching fix

CecileRobertMichon:
pass 1/1

CecileRobertMichon:
pass 2/2

CecileRobertMichon:
pass 3/3

PR #3381, opened 2023-03-31
mark context canceled or exceeded as transient failures

### LINKED ISSUE #2976: Intermittent failures in capz with StatusCode=0 while creating azure resources
/kind bug

While trying to create a workload cluster, there has been intermittent failure when CAPZ tries to list, get or create an azure resource as part of reconciliation. 

There are two different category of errors that can be seen in CAPZ: 

1) StatusCode=0 -- Original Error: context deadline exceeded
2) StatusCode=0 -- Context canceled

For example: 

```
E1222 12:15:30.961325       1 controller.go:326]  "msg"="Reconciler error" "error"="failed to reconcile AzureMachine: 
failed to reconcile AzureMachine service inboundnatrules: failed to create resource sh-2646368-221222-023620-rg/tkg-upgrade-
9864azure-antrea-control-plane-v1-24-6-vmware-qbkbx (service: inboundnatrules):
network.InboundNatRulesClient#CreateOrUpdate: Failure sending request: StatusCode=0 -- Original Error: context deadline
exceeded" "azureMachine"={"name":"tkg-upgrade-9864azure-antrea-control-plane-v1-24-6-vmware-
qbkbx","namespace":"default"} "controller"="azuremachine" "controllerGroup"="infrastructure.cluster.x-k8s.io"
"controllerKind"="AzureMachine" "name"="tkg-upgrade-9864azure-antrea-control-plane-v1-24-6-vmware-qbkbx"
"namespace"="default" "reconcileID"="62fcdc95-5180-4d22-a464-5fbed9437f87" 
```

```
E1226 21:18:10.928108      44 controller.go:326]  "msg"="Reconciler error" "error"="client rate limiter Wait returned an error:
context canceled" "azureCluster"={"name":"default-20004","namespace":"default"} "controller"="azurecluster"
"controllerGroup"="infrastructure.cluster.x-k8s.io" "controllerKind"="AzureCluster" "name"="default-20004"
"namespace"="default" "reconcileID"="60095f02-adac-4885-b1b0-e8977db79c60"

E1226 21:18:10.931279      44 controller.go:326]  "msg"="Reconciler error" "error"="failed to create scope: failed to configure
azure settings and credentials for Identity: failed to create copied AzureIdentity default-20004-default-cluster-identity in capz-
system: client rate limiter Wait returned an error: context canceled" "azureCluster"={"name":"default-
20004","namespace":"default"} "controller"="azurecluster" "controllerGroup"="infrastructure.cluster.x-k8s.io"
"controllerKind"="AzureCluster" "name"="default-20004" "namespace"="default" "reconcileID"="63711da9-357a-43bb-9354-
53df525837d5"
```

As a side effect the condition with severity gets written to the AzureCluster which bubbles all the way up to Cluster object. Though this gets recovered in the next reconcile. 

Tried creating a workload cluster using capz v1.5.3, v1.6.0 and from main branch. All of them have above issues intermittently. 

The workload cluster creation should not have this intermittent issue which causes failures especially during upgrades. This issue is highly reproducible which does not seem could be due to network slowness. 

**Environment:**

sonasingh46:
**What type of PR is this?**
This PR adds code to mark the `context deadline exceeded or canceled` as a transient failure. 

Fixes #2976 

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Mark context canceled or exceeded as transient failures
```

### DIFF
--- a/azure/errors.go
+++ b/azure/errors.go
@@ -17,6 +17,7 @@ limitations under the License.
 package azure
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"time"
@@ -152,4 +153,13 @@ func IsOperationNotDoneError(target error) bool {
 		return IsOperationNotDoneError(reconcileErr.error)
 	}
 	return errors.As(target, &OperationNotDoneError{})
+}
+
+// IsContextDeadlineExceededOrCanceledError checks if it's a context deadline
+// exceeded or canceled error.
+func IsContextDeadlineExceededOrCanceledError(err error) bool {
+	if err == nil {
+		return false
+	}
+	return errors.Is(err, context.DeadlineExceeded) || errors.Is(err, context.Canceled)
 }
--- a/azure/errors_test.go
+++ b/azure/errors_test.go
@@ -1,0 +1,53 @@
+package azure
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/pkg/errors"
+)
+
+func TestIsContextDeadlineExceededOrCanceled(t *testing.T) {
+	tests := []struct {
+		name string
+		want bool
+		err  error
+	}{
+		{
+			name: "Context deadline exceeded error",
+			err: func() error {
+				ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(-7*time.Hour))
+				defer cancel()
+				return ctx.Err()
+			}(),
+			want: true,
+		},
+		{
+			name: "Context canceled exceeded error",
+			err: func() error {
+				ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(1*time.Hour))
+				cancel()
+				return ctx.Err()
+			}(),
+			want: true,
+		},
+		{
+			name: "Nil error",
+			err:  nil,
+			want: false,
+		},
+		{
+			name: "Error other than context deadline exceeded or canceled error",
+			err:  errors.New("dummy error"),
+			want: false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := IsContextDeadlineExceededOrCanceledError(tt.err); got != tt.want {
+				t.Errorf("IsContextDeadlineExceededOrCanceled() = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
--- a/azure/services/async/async.go
+++ b/azure/services/async/async.go
@@ -140,6 +140,7 @@ func (s *Service) CreateOrUpdateResource(ctx context.Context, spec azure.Resourc
 	log.V(2).Info(fmt.Sprintf("%sing resource", logMessageVerbPrefix), "service", serviceName, "resource", resourceName, "resourceGroup", rgName)
 	result, sdkFuture, err := s.Creator.CreateOrUpdateAsync(ctx, spec, parameters)
 	errWrapped := errors.Wrapf(err, fmt.Sprintf("failed to %se resource %s/%s (service: %s)", logMessageVerbPrefix, rgName, resourceName, serviceName))
+
 	if sdkFuture != nil {
 		future, err := converters.SDKToFuture(sdkFuture, infrav1.PutFuture, serviceName, resourceName, rgName)
 		if err != nil {
@@ -148,6 +149,11 @@ func (s *Service) CreateOrUpdateResource(ctx context.Context, spec azure.Resourc
 		s.Scope.SetLongRunningOperationState(future)
 		return nil, azure.WithTransientError(azure.NewOperationNotDoneError(future), getRequeueAfterFromFuture(sdkFuture))
 	} else if err != nil {
+		// If it is an intermittent failure with context deadline exceeded or canceled as the reconciler could not complete
+		// in the max amount of time, mark it as a transient error and return.
+		if azure.IsContextDeadlineExceededOrCanceledError(ctx.Err()) {
+			return nil, azure.WithTransientError(errWrapped, getRetryAfterFromError(err))
+		}
 		return nil, errWrapped
 	}
 
@@ -174,6 +180,7 @@ func (s *Service) DeleteResource(ctx context.Context, spec azure.ResourceSpecGet
 	// No long running operation is active, so delete the resource.
 	log.V(2).Info("deleting resource", "service", serviceName, "resource", resourceName, "resourceGroup", rgName)
 	sdkFuture, err := s.Deleter.DeleteAsync(ctx, spec)
+
 	if sdkFuture != nil {
 		future, err := converters.SDKToFuture(sdkFuture, infrav1.DeleteFuture, serviceName, resourceName, rgName)
 		if err != nil {
@@ -185,6 +192,11 @@ func (s *Service) DeleteResource(ctx context.Context, spec azure.ResourceSpecGet
 		if azure.ResourceNotFound(err) {
 			// already deleted
 			return nil
+		}
+		// If it is an intermittent failure with context deadline exceeded or canceled as the reconciler could not complete
+		// in the max amount of time, mark it as a transient error and return.
+		if azure.IsContextDeadlineExceededOrCanceledError(ctx.Err()) {
+			return azure.WithTransientError(err, getRetryAfterFromError(err))
 		}
 		return errors.Wrapf(err, "failed to delete resource %s/%s (service: %s)", rgName, resourceName, serviceName)
 	}


devigned (commented):
I haven't looked into it much, but is there a better way to determine the error type for context canceled or context exceeded? Feels weird to do a string contains on the message value.

sonasingh46:
@devigned -- Yes, right. Will look into the code base more. 
@CecileRobertMichon @mboersma -- Do you any insights/suggestions here?

nawazkh:
I am interested to know too! Do you think we could do something like line 18 and line 26 from my test https://go.dev/play/p/WmPqp7iLWor using the context itself?

azure/services/async/async.go
```diff
@@ -230,3 +244,19 @@ func getRetryAfterFromError(err error) time.Duration {
 	}
 	return ret
 }
+
+// ContextDeadlineExceededOrCanceled checks if it's an intermittent context deadline
+// exceeded or canceled error.
+func ContextDeadlineExceededOrCanceled(ctx context.Context, err error) bool {
```
sonasingh46, line 250: Just thinking if it is appropriate to keep this function here or move to azure/errors.go or somewhere else.
But then, I think it is better to keep it here until we find the need to use it at other place and at that time we will have better clarity to find the right home for this.

azure/services/async/async.go
```diff
@@ -230,3 +244,9 @@ func getRetryAfterFromError(err error) time.Duration {
 	}
 	return ret
 }
+
+// ContextDeadlineExceededOrCanceled checks if it's an intermittent context deadline
+// exceeded or canceled error.
+func ContextDeadlineExceededOrCanceled(ctx context.Context) bool {
+	return errors.Is(ctx.Err(), context.DeadlineExceeded) || errors.Is(ctx.Err(), context.Canceled)
```
nawazkh, line 251: I am wondering if `context.Canceled` qualifies as an _transient error_?
I understand `context.Canceled` as `abandon work`. So, shouldn't we be bubbling up the error or requeuing the request upon `context.Canceled`?

nawazkh: Sorry my bad. I looked it up, and `azure.WithTransientError()` bubbles up the error as transient and also requeues with default timeout unless other time is specified in the error. https://github.com/nawazkh/cluster-api-provider-azure/blob/f8ad961ef46fe856eda6d96056751e8b7e4a068f/azure/services/async/async.go#L208

nawazkh:
Looks good to me! :shipit: 
/lgtm

azure/services/async/async.go
CecileRobertMichon: can we move this helper func to https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/0837cd3019526060dc334b9b1e9cb82d94f44e64/azure/errors.go and add a small unit test for it?

azure/services/async/async.go
```diff
@@ -230,3 +244,9 @@ func getRetryAfterFromError(err error) time.Duration {
 	}
 	return ret
 }
+
+// ContextDeadlineExceededOrCanceled checks if it's an intermittent context deadline
+// exceeded or canceled error.
+func ContextDeadlineExceededOrCanceled(ctx context.Context) bool {
```
CecileRobertMichon, line 250:
```suggestion
func IsContextDeadlineExceededOrCanceledError(ctx context.Context) bool {
```
is a bit of a mouthful but more consistent with

```
func IsOperationNotDoneError(target error) bool 
```

azure/services/async/async.go
```diff
@@ -140,6 +140,13 @@ func (s *Service) CreateOrUpdateResource(ctx context.Context, spec azure.Resourc
 	log.V(2).Info(fmt.Sprintf("%sing resource", logMessageVerbPrefix), "service", serviceName, "resource", resourceName, "resourceGroup", rgName)
 	result, sdkFuture, err := s.Creator.CreateOrUpdateAsync(ctx, spec, parameters)
 	errWrapped := errors.Wrapf(err, fmt.Sprintf("failed to %se resource %s/%s (service: %s)", logMessageVerbPrefix, rgName, resourceName, serviceName))
+
+	// If it is an intermittent failure with context deadline exceeded or canceled as the reconciler could not complete
+	// in the max amount of time, mark it as a transient error and return.
+	if err != nil && ContextDeadlineExceededOrCanceled(ctx) {
```
CecileRobertMichon, line 146: in this case does err == ctx.Err() ? if so we could just pass err directly

azure/services/async/async.go
```diff
@@ -174,6 +181,13 @@ func (s *Service) DeleteResource(ctx context.Context, spec azure.ResourceSpecGet
 	// No long running operation is active, so delete the resource.
 	log.V(2).Info("deleting resource", "service", serviceName, "resource", resourceName, "resourceGroup", rgName)
 	sdkFuture, err := s.Deleter.DeleteAsync(ctx, spec)
+
+	// If it is an intermittent failure with context deadline exceeded or canceled as the reconciler could not complete
+	// in the max amount of time, mark it as a transient error and return.
+	if err != nil && ContextDeadlineExceededOrCanceled(ctx) {
```
nawazkh, line 187: Thinking a little bit more about this, do we want to handle `err != nil` and `ContextDeadlineExceededOrCanceled(ctx)` separately? 
Meaning, on an error, 
_if_ the `ContextDeadlineExceededOrCanceled(ctx)==true` and return with Transient error. 
_else_ you continue checking with `azure.ResourceNotFound(err)`. 

From current PR's line 183 till 207 could look like
```suggest
.
.
.
sdkFuture, err := s.Deleter.DeleteAsync(ctx, spec)
if err != nil { // check error first
    return nil // already deleted
  }

  if ContextDeadlineExceededOrCanceled(ctx) {
    return azure.WithTransientError(err, getRetryAfterFromError(err))
  }
  // none of the errors, bubble up this new error!
  return errors.Wrapf(err, "failed to delete resource %s/%s (service: %s)", rgName, resourceName, serviceName)
}

if sdkFuture != nil { // as its implemented already... 
  future, err := converters.SDKToFuture(sdkFuture, infrav1.DeleteFuture, serviceName, resourceName, rgName)
  if err != nil {
    return errors.Wrapf(err, "failed to delete resource %s/%s (service: %s)", rgName, resourceName, serviceName)
  }
  s.Scope.SetLongRunningOperationState(future)
  return azure.WithTransientError(azure.NewOperationNotDoneError(future), getRequeueAfterFromFuture(sdkFuture))
}

log.V(2).Info("successfully deleted resource", "service", serviceName, "resource", resourceName, "resourceGroup", rgName)
return nil
```

Same thoughts for the above. 
@CecileRobertMichon What do you think?

nawazkh: holding off for this thought. Sorry for early lgtm :| 
/lgtm cancel

CecileRobertMichon: hmm now that I'm also thinking about this more, doesn't this change short circuit `if sdkFuture != nil` ? We do a best effort `WaitForCompletion` then return the future + an error if the operation didn't complete within the timeout, for example: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/services/virtualmachines/client.go#L150

so this change would break our current async implementation because we would never store the future and would instead aggressively retry to create/delete, causing a higher number of API calls for each operation instead of doing the PUT/Create once and waiting for the operation to complete.

/hold

CecileRobertMichon: So modifying @nawazkh's recommendation above, I think preserve the current order of first checking the sdkFuture, then processing remaining errors

```
	sdkFuture, err := s.Deleter.DeleteAsync(ctx, spec)
	if sdkFuture != nil {
		future, err := converters.SDKToFuture(sdkFuture, infrav1.DeleteFuture, serviceName, resourceName, rgName)
		if err != nil {
			return errors.Wrapf(err, "failed to delete resource %s/%s (service: %s)", rgName, resourceName, serviceName)
		}
		s.Scope.SetLongRunningOperationState(future)
		return azure.WithTransientError(azure.NewOperationNotDoneError(future), getRequeueAfterFromFuture(sdkFuture))
	} else if err != nil {
			// already deleted
			return nil
		}

 		if ContextDeadlineExceededOrCanceled(ctx) { // or maybe this becomes IsContextExceededError(err) ? 
		    return azure.WithTransientError(err, getRetryAfterFromError(err))
 		}

		return errors.Wrapf(err, "failed to delete resource %s/%s (service: %s)", rgName, resourceName, serviceName)
	}

	log.V(2).Info("successfully deleted resource", "service", serviceName, "resource", resourceName, "resourceGroup", rgName)
	return nil
```

azure/services/async/async.go
sonasingh46: Passing err directly will not help to figure out if the context has exceeded or got canceled. We need the `ctx` object to figure that out. 
But I think, I can omit checking the error and just that function can check if ctx.Err() is nil or not. 
I will make that change.

azure/services/async/async.go
sonasingh46: Sure. I had added a review comment re this myself to keep it in azure/errors.go or not ( got outdated after I pushed the changes )
I decided to keep it here : 
- Pushing it to azure/errors.go will mean an import of ctx 
- I thought it is worth to keep it here until a situation arises when we need to use this function at other place which will give more info on what should be the right home for this. 

But also, I do not see a problem to push it to azure/errors.go and I will do that for now.

azure/services/async/async.go
sonasingh46: This is a good catch. Thank you.

azure/services/async/async.go
CecileRobertMichon: if we're not checking the error itself, it should be

```suggestion
func IsContextDeadlineExceededOrCanceled(ctx context.Context) bool {
```

azure/services/async/async.go
nawazkh: Thank you @CecileRobertMichon for adding context on `Futures`, current implementation with out this PR makes much more sense!

nawazkh:
Waaw, this looks much clean! Thanks @sonasingh46 
I have one more thought tho. Sorry, I should have brought it up much earlier..

- Will there be a scenario where the reconciler could get into transient error without actually be able to delete the resource? 
- Do you think we want to have a retry logic in here where we retry (read as tolerate) `x` amounts of transient failures before marking resource deletion as an error?

CecileRobertMichon:
Good questions.

> Will there be a scenario where the reconciler could get into transient error without actually be able to delete the resource?

No, that scenario wouldn't be covered by this new code, we would see the request fail to send, in which case `IsContextDeadlineExceededOrCanceled` will return `false` and we will return a normal error as before (which then gets retried by controller-runtime with exponential backoff), or the long-running operation itself would fail and the future would be marked as "done" with an error (also not context exceeded).

> Do you think we want to have a retry logic in here where we retry (read as tolerate) x amounts of transient failures before marking resource deletion as an error?

We shouldn't do this as it would violate the principle of re-entrancy (reconcile itself should have no state and be idempotent). Instead we should rely on controller-runtime to retry errors with exponential backoff, and only mark errors as + "transient" (i.e. add a requeueAfter time and not return an error) when we are confident the error is transient, which I believe is the case in this PR.

CecileRobertMichon (commented):
/hold cancel

azure/errors_test.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+func TestIsContextDeadlineExceededOrCanceled(t *testing.T) {
+	tests := []struct {
+		name string
+		want bool
+		err  error
+	}{
+		{
+			name: "Context deadline exceeded error",
```
CecileRobertMichon, line 16: can you please add a test case for nil err and one for another type error that returns false?

azure/errors.go
```diff
@@ -153,3 +154,12 @@ func IsOperationNotDoneError(target error) bool {
 	}
 	return errors.As(target, &OperationNotDoneError{})
 }
+
+// IsContextDeadlineExceededOrCanceled checks if it's a context deadline
+// exceeded or canceled error.
+func IsContextDeadlineExceededOrCanceled(err error) bool {
```
CecileRobertMichon, line 160: nit: for consistent naming, this should either be 

```
func IsContextDeadlineExceededOrCanceledError(err error) bool {
```

or 

```
func IsContextDeadlineExceededOrCanceled(ctx context.Context) bool {
```

CecileRobertMichon (commented):
lgtm

CecileRobertMichon (commented):
/lgtm
/approve

PR #3388, opened 2023-04-04
Allow feature gates to be passed through for Windows CI

jsturtevant:
**What type of PR is this?**
/kind feature (testing)

In upstream Windows we need ability to pass feature gates to test alpha features.  This adds a NODE_FEATURE_GATE which complements the K8S_FEATURE_GATE. They are different since some features need to be set at the node level or api level only.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```


test/e2e/conformance_test.go
```diff
@@ -148,8 +148,13 @@ var _ = Describe("Conformance Tests", func() {
 			// The templates use WORKER_MACHINE_COUNT for linux machines for backwards compatibility so clear it
 			linuxWorkerMachineCount = 0
 
-			// Windows e2e tests require HPAContainerMetrics featuregate
-			Expect(os.Setenv("K8S_FEATURE_GATES", "HPAContainerMetrics=true")).To(Succeed())
+			// Windows e2e tests require HPAContainerMetrics feature gate
+			preConfigured := os.Getenv("K8S_FEATURE_GATES")
+			hpaFeature := "HPAContainerMetrics=true"
```
jackfrancis, line 153: Is this feature-gate Windows-specific?

jsturtevant: no but we use it in the Windows upstream tests

jackfrancis: Got it. So I like what's happening here, seems like we are keeping the default `HPAContainerMetrics=true` feature-gate configuration for Windows, while allowing complementary/additional feature-gates to be passed in via `$K8S_FEATURE_GATES`.

I think L157 below is not doing what we want, as it's still alweays just setting `K8S_FEATURE_GATES` to `HPAContainerMetrics=true` no matter what.

Also, in order to make this as flexible and functional as possible, we probably want to do the following:

1. Check if the user has provided values via `$K8S_FEATURE_GATES`
  - If not, set `K8S_FEATURE_GATES` to `HPAContainerMetrics=true`
  - If those values already exist, do a substring search for `HPAContainerMetrics` inside `$K8S_FEATURE_GATES`
    - If `HPAContainerMetrics` is already in `$K8S_FEATURE_GATES` (either set to true or false), then do nothing and use `$K8S_FEATURE_GATES`
    - If `HPAContainerMetrics` is not present in `$K8S_FEATURE_GATES`, then append `HPAContainerMetrics=true` to the existing feature gate configuration in `$K8S_FEATURE_GATES`

Does that make sense?

jsturtevant: yup you're right :-) We could also just skip setting HPAContainerMetrics here and add it to the prow job config, which is more straightforward and reduces the logic here

mansikulkarni96:
@jsturtevant code changes LGTM, for my own understanding with this change we can use the NODE_FEATURE_GATE ENV var to set the feature gate like podAndContainerStatsFromCRI to true.

jsturtevant:
> for my own understanding with this change we can use the NODE_FEATURE_GATE ENV var to set the feature gate like podAndContainerStatsFromCRI to true.

Yes, I need to double check but I think this is only a node level, i don't think the API server needs it set.

test/e2e/conformance_test.go
jsturtevant: I've opened: https://github.com/kubernetes/test-infra/pull/29220

jsturtevant: There will be a couple other sig-windows specific tests but we can use that to test this PR.

test/e2e/conformance_test.go
```diff
@@ -147,9 +147,6 @@ var _ = Describe("Conformance Tests", func() {
 			// Conformance for windows doesn't require any linux worker machines.
 			// The templates use WORKER_MACHINE_COUNT for linux machines for backwards compatibility so clear it
 			linuxWorkerMachineCount = 0
-
-			// Windows e2e tests require HPAContainerMetrics featuregate
-			Expect(os.Setenv("K8S_FEATURE_GATES", "HPAContainerMetrics=true")).To(Succeed())
```
jackfrancis, line 152: @jsturtevant if you're willing to take on the add'l test config burden of this, I do think it makes more sense there than hard-coding it in CAPZ E2E

Lemme know if I can help review a test-infra PR!

marosset:
@jsturtevant - do we want to make these changes here or in https://github.com/kubernetes-sigs/windows-testing to the templates we use with az capi (or both places)?

marosset:
overall LGTM

jsturtevant:
> do we want to make these changes here or in https://github.com/kubernetes-sigs/windows-testing to the templates we use with az capi (or both places)?

eventually az capi, but for now we need to be able to run the PR job which uses the ci-entrypoint.sh.

jsturtevant:
/test pull-cluster-api-provider-azure-windows-containerd-upstream-with-ci-artifacts-serial-slow
note there might be a few flakes in the test that @marosset is looking into but it should demonstrate that the setting is applied from the env variable added in https://github.com/kubernetes/test-infra/pull/29220

CecileRobertMichon:

just using this PR to test an unrelated fix ^

jackfrancis:
@marosset were you able to confirm the changes in the most recent test ran by @jsturtevant 

should we retest until that test passes, or are we ready to merge this?

marosset:
> @marosset were you able to confirm the changes in the most recent test ran by @jsturtevant
> 
> should we retest until that test passes, or are we ready to merge this?

Yes, those failures are known flakes.
I'm validating some fixes for the HPA tests this week.
We can merge this PR

marosset (approved):

CecileRobertMichon (commented):
/lgtm
/approve

CecileRobertMichon:
@marosset looks like this one needs a manual cherry-pick

PR #3399, opened 2023-04-06
add explicit securityContexts to the controller and namespace label for privileged

chrischdi:
**What type of PR is this?**

/kind feature

This does not really change the configuration, it just makes it explicit and enforce the defaults, except for the seccompPolicy which changes from Unconfined to RuntimeDefault. Syscalls filtered by RuntimeDefault policy are 95% namespaced and require capabilities (which we drop) in the first place, so no practical change there either.

This allows the controller to be compatible to the restricted pod security admission profile.

However, the nmi daemonset requires the privileged profile which is why this also adds the corresponding label to the namespace, to allow cluster-wide enforcement of a policy lower than privileged, while preserving privileged for the capz-system namespace. If the nmi daemonset is replaced by Azure Workload Identity (https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/2205), and if  Azure Workload Identity does not require privileged, the namespace label could get removed again.

This is a recommendation in the [CAPI v1.3->1.4 upgrade guide](https://main.cluster-api.sigs.k8s.io/developer/providers/migrations/v1.3-to-v1.4.html#suggested-changes-for-providers).

Fixes #

Prior art:
* https://github.com/kubernetes-sigs/cluster-api/pull/7831
* https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4104

Note: this requires the v1.4 branch of CAPI for usage with tilt, because the Tiltfile then removes the securityContext for allowing it to succeed.

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Explicitly set securityContexts in the manifests to comply with the restricted pod security admission profile.
```

### DIFF
--- a/config/default/namespace.yaml
+++ b/config/default/namespace.yaml
@@ -2,3 +2,6 @@ apiVersion: v1
 kind: Namespace
 metadata:
   name: system
+  labels:
+    # this is required due to the nmi daemonset
+    pod-security.kubernetes.io/enforce: privileged
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -57,6 +57,17 @@ spec:
             valueFrom:
               fieldRef:
                 fieldPath: metadata.namespace
+          securityContext:
+            allowPrivilegeEscalation: false
+            capabilities:
+              drop:
+              - ALL
+            runAsUser: 65532
+            runAsGroup: 65532
+      securityContext:
+        runAsNonRoot: true
+        seccompProfile:
+          type: RuntimeDefault
       terminationGracePeriodSeconds: 10
       serviceAccountName: manager
       tolerations:


CecileRobertMichon:
Thank you @chrischdi 

> If the nmi daemonset is replaced by Azure Workload Identity (https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/2205), and if Azure Workload Identity does not require privileged, the namespace label could get removed again.

@sonasingh46 @aramase do you know if this is true?

Jont828:
This looks good to me and it seems to be passing the tests as well which gives me confidence that it works. I think if we get a second look from others we should be good to proceed.

aramase:
> Thank you @chrischdi
> 
> > If the nmi daemonset is replaced by Azure Workload Identity (#2205), and if Azure Workload Identity does not require privileged, the namespace label could get removed again.
> 
> @sonasingh46 @aramase do you know if this is true?

NMI does need `privileged: true` in order to add the iptable rules. Azure Workload Identity webhook doesn't require `privileged`, so this is correct.

config/default/namespace.yaml
```diff
@@ -2,3 +2,6 @@ apiVersion: v1
 kind: Namespace
 metadata:
   name: system
+  labels:
+    # this is required due to the nmi daemonset
```
CecileRobertMichon, line 6: @sonasingh46 can we make a note to remove this in #2924 ?

CecileRobertMichon (commented):
/lgtm
/assign @sonasingh46

jackfrancis:
cc @sonasingh46 if you're able to add another review to this in the 1.9 milestone cycle that would be great!

sonasingh46:
Sure. I will take a look at this tomorrow.

mboersma (commented):
/lgtm

PR #3411, opened 2023-04-10
return 1 in ci-entrypoint func’s when necessary

jackfrancis:
**What type of PR is this?**

/kind flake

This PR adds necessary per-command return statements inside `ci-entrypoint.sh` functions to fulfill the "retry until success contract" of those functions.

The following bash snippet demonstrates the way that not doing so can produce functions that do not retry after one of the commands fails, which in `ci-entrypoint.sh` will have the side-effect of leaving a functional gap in the set of required steps during cluster bootstrapping.

```bash
#!/bin/bash
set -o errexit
set -o nounset
set -o pipefail

pathology_one() {
    echo "entering pathology_one"
    if ! false; then
        ls does-not-exist
    fi
    echo "exiting pathology_one"
}

pathology_two() {
    echo "entering pathology_two"
    RET=$(ls does-not-exist)
    export RET
    echo "exiting pathology_two"
}

pathology_three() {
    echo "entering pathology_three"
    ls does-not-exist
    echo "exiting pathology_three"
}

until pathology_one; do
    echo "pathology_one failed"
    sleep 1
done


until pathology_two; do
    echo "pathology_two failed"
    sleep 1
done

until pathology_three; do
    echo "pathology_three failed"
    sleep 1
done
```

The above script reflects the current approach of `ci-entrypoint.sh`, and we would expect that the script should hang forever during the `until pathology_one` function call as the `ls does-not-exist` statement returns non-zero. Instead command proceeds immediately afterwards, and the same thing happens for the next two functions that wrap non-successful command statements:

```sh
$ ./test.sh 
entering pathology_one
ls: does-not-exist: No such file or directory
exiting pathology_one
entering pathology_two
ls: does-not-exist: No such file or directory
exiting pathology_two
entering pathology_three
ls: does-not-exist: No such file or directory
exiting pathology_three
$ echo $?
0
```

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note

```

### DIFF
--- a/scripts/ci-entrypoint.sh
+++ b/scripts/ci-entrypoint.sh
@@ -102,7 +102,7 @@ setup() {
     export CCM_COUNT="${CCM_COUNT:-1}"
     export WORKER_MACHINE_COUNT="${WORKER_MACHINE_COUNT:-2}"
     export EXP_CLUSTER_RESOURCE_SET="true"
-    
+
     # TODO figure out a better way to account for expected Windows node count
     if [[ -n "${TEST_WINDOWS:-}" ]]; then
         export WINDOWS_WORKER_MACHINE_COUNT="${WINDOWS_WORKER_MACHINE_COUNT:-2}"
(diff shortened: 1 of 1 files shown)


marosset:
LGTM

I think we should backport this to 1.7 and 1.8 too

jackfrancis:
/hold

for @andyzhangx to sign off

CecileRobertMichon (commented):
/lgtm

PR #3414, opened 2023-04-10
update SDKImageToImage func to handle different image types

### LINKED ISSUE #3413: `AzureMachinePool` Create/Destroy Loop When Non-Marketplace Image Used
/kind bug

When an `AzureMachinePool` resource is created using a VM image other than a marketplace one, the `AzureMachinePool` never becomes healthy, and replicas within the pool are created and deleted in a never-ending loop.

Create an `AzureMachinePool` resource with a custom VM image that is not a marketplace one.

The `AzureMachinePool` is successfully created with healthy and ready VMs.

This problem seems to be in the [SDKImageToImage](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/converters/vmss.go#L150-L164) function, as it does not account for images other than marketplace ones. Elsewhere in the code, in the [hasLatestModelApplied](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/scope/machinepoolmachine.go#L524-L547) function, `reflect.DeepEqual(s.instance.Image, *image)` is called, which results in the `hasLatestModelApplied` function returning false, resulting in the replicas being deleted and recreated.

**Environment:**
- cluster-api-provider-azure version: `main`/`v1.8.2`.
- Kubernetes version: (use `kubectl version`): `v1.24.5`
- OS (e.g. from `/etc/os-release`): MacOS 12.6.2

jamesgoodhouse:
**What type of PR is this?**
/kind bug

Modifies logic for converting from the SDK image to the infrav1 image to handle cases other than marketplace images.

Fixes #3413

**TODOs**:

- [x] squashed commits
- [x] adds unit tests

**Release note**:

```release-note
Updates to AzureMachinePool that take into account VM image types other than marketplace. Previous behavior resulted in a create/delete/create loop occurring when attempting to use an image other than a marketplace image.
```

### DIFF
--- a/azure/converters/vmss.go
+++ b/azure/converters/vmss.go
@@ -17,11 +17,20 @@ limitations under the License.
 package converters
 
 import (
+	"regexp"
+
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2021-11-01/compute"
 	"k8s.io/utils/pointer"
 	azprovider "sigs.k8s.io/cloud-provider-azure/pkg/provider"
 	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
 	"sigs.k8s.io/cluster-api-provider-azure/azure"
+)
+
+const (
+	// RegExpStrCommunityGalleryID is a regexp string used for matching community gallery IDs and capturing specific values.
+	RegExpStrCommunityGalleryID = `/CommunityGalleries/(?P<gallery>.*)/Images/(?P<name>.*)/Versions/(?P<version>.*)`
+	// RegExpStrComputeGalleryID is a regexp string used for matching compute gallery IDs and capturing specific values.
+	RegExpStrComputeGalleryID = `/subscriptions/(?P<subID>.*)/resourceGroups/(?P<rg>.*)/providers/Microsoft.Compute/galleries/(?P<gallery>.*)/images/(?P<name>.*)/versions/(?P<version>.*)`
 )
 
 // SDKToVMSS converts an Azure SDK VirtualMachineScaleSet to the AzureMachinePool type.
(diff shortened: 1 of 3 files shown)


CecileRobertMichon (commented):
/ok-to-test

jamesgoodhouse:
Gtg on this PR unless there are any changes requested

azure/converters/image.go
```diff
@@ -57,7 +57,7 @@ func computeImageToSDK(image *infrav1.Image) (*compute.ImageReference, error) {
 	idTemplate := "/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Compute/galleries/%s/images/%s/versions/%s"
 	if image.SharedGallery != nil {
 		return &compute.ImageReference{
-			ID: pointer.String(fmt.Sprintf(idTemplate,
+			SharedGalleryImageID: pointer.String(fmt.Sprintf(idTemplate,
```
Jont828, line 60: Just to clarify, would we still want to set the `ID` field for non-shared images? In other words, would we want to preserve the existing functionality?

Jont828 (commented):
Thanks for the contribution! From what I understand, it seems like you want to add support for additional cases on top of the existing ones. My first impression is to try to see if the behavior for existing cases is still the same, in other words, to see that the API call struct hasn't changed. I'm not sure I can easily follow your logic with the regex and `getParams()`, could you explain a bit on what you're trying to do?

azure/converters/vmss.go
```diff
@@ -149,8 +158,52 @@ func SDKToVMSSVM(sdkInstance compute.VirtualMachineScaleSetVM) *azure.VMSSVM {
 
 // SDKImageToImage converts a SDK image reference to infrav1.Image.
 func SDKImageToImage(sdkImageRef *compute.ImageReference, isThirdPartyImage bool) infrav1.Image {
+	if sdkImageRef.ID != nil {
```
willie-yao, line 161: Is this referring to a compute gallery image? If so, a comment here could help distinguish the `ID` from community and shared images.

azure/converters/vmss.go
```diff
@@ -163,10 +216,47 @@ func SDKImageToImage(sdkImageRef *compute.ImageReference, isThirdPartyImage bool
 	}
 }
 
-// GetOrchestrationMode returns the compute.OrchestrationMode for the given infrav1.OrchestrationModeType.
-func GetOrchestrationMode(modeType infrav1.OrchestrationModeType) compute.OrchestrationMode {
-	if modeType == infrav1.FlexibleOrchestrationMode {
-		return compute.OrchestrationModeFlexible
+func cgImageRefToImage(id string) infrav1.Image {
```
willie-yao, line 219: This naming may be confusing with both compute and community gallery abbreviating to `cg`. Would it be better to rename this function, or add a clarifying comment?

azure/converters/vmss_test.go
```diff
@@ -256,6 +252,49 @@ func Test_SDKImageToImage(t *testing.T) {
...
+			Image: infrav1.Image{
+				ComputeGallery: &infrav1.AzureComputeGalleryImage{
+					Gallery:        "gallery",
+					Name:           "image",
+					Version:        "version",
+					SubscriptionID: pointer.String("subscription"),
+					ResourceGroup:  pointer.String("rg"),
+				},
+			},
+		},
```
willie-yao, line 297: Would you be able to add some additional unit tests that test an image that doesn't pass the regex check in `getParams`?

willie-yao (commented):
Thanks so much for working on this! Just a few comments from my end

jamesgoodhouse:
> Thanks for the contribution! From what I understand, it seems like you want to add support for additional cases on top of the existing ones. My first impression is to try to see if the behavior for existing cases is still the same, in other words, to see that the API call struct hasn't changed. I'm not sure I can easily follow your logic with the regex and `getParams()`, could you explain a bit on what you're trying to do?

@Jont828 I'm using the regex and `getParams()` function because the Azure SDK is returning an ID back and we need to break it apart to fit the shape the data is for the `infrav1.Image` struct.

I'm unsure I follow what you're saying/asking about "try to see if the behavior for existing cases is still the same, in other words, to see that the API call struct hasn't changed".

azure/converters/image.go
jamesgoodhouse: I'm unsure on this one. I don't have a way that I know of to exercise this code to see that is returned from the Azure API, and since the `SharedGalleryImageID` field existed it seemed to make sense to me to use it, but maybe that was a bad idea since it originally used ID.

azure/converters/vmss_test.go
jamesgoodhouse: @willie-yao i added two additional test cases.

azure/converters/vmss.go
jamesgoodhouse: @willie-yao i added comments to each of the function definitions to further clarify what they're doing.

willie-yao: Thanks this looks great! For some reason, I'm unable to resolve these comments.

willie-yao (approved):
/lgtm

Thanks for all the great work on this! 🚀

willie-yao:
@jamesgoodhouse You might want to add a release note to the PR description. I think it would be helpful to include this in the release notes!

jamesgoodhouse:
@willie-yao good call! i just updated the release notes.

azure/converters/image.go
jamesgoodhouse: I switched back to ID.

azure/converters/vmss.go
```diff
@@ -149,8 +158,53 @@ func SDKToVMSSVM(sdkInstance compute.VirtualMachineScaleSetVM) *azure.VMSSVM {
...
+	// shared gallery image
+	if sdkImageRef.SharedGalleryImageID != nil {
+		return sgImageRefToImage(*sdkImageRef.SharedGalleryImageID)
+	}
+	// marketplace image
+	return mpImageRefToImage(sdkImageRef, isThirdPartyImage)
+}
+
+// GetOrchestrationMode returns the compute.OrchestrationMode for the given infrav1.OrchestrationModeType.
+func GetOrchestrationMode(modeType infrav1.OrchestrationModeType) compute.OrchestrationMode {
```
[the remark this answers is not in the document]
CecileRobertMichon: ah nvm looks like this is just an existing func that moved

CecileRobertMichon (commented):
/assign @mboersma

mboersma (approved):
/lgtm
/approve

PR #3425, opened 2023-04-13
feat: AKS user-assigned identity for control plane and kubelet

### LINKED ISSUE #3239: Support User Identity in CAPZ managed AKS Cluster
/kind feature

Currently, only system managed identity is supported on CAPZ managed AKS cluster for **control plane identity**. Based on your current use case, it would be great to have user managed identity support in AKS cluster.

Code path where we hard code identity to system managed identity: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/services/managedclusters/spec.go#L273

**Environment:**

- cluster-api-provider-azure version:  1.7.2

### LINKED ISSUE #3261: Support Kubelet User Managed Identity on AKS VMSS pool.
/kind feature

Currently, we reply on an auto-assigned **kubelet user identity** which is assigned via during AKS cluster creation. AKS provides capability to configure BYO user identity in the kubelet identity field. This enables user to define permissions before the cluster is created and assign the identity during cluster creation.

Link: https://learn.microsoft.com/en-us/azure/aks/use-managed-identity#use-a-pre-created-kubelet-managed-identity

Just to clarify, there are two types of identity supported in AKS. first is the control plane identity and other is the VMSS pool identity. The new feature ask is for the later one where we require pods running on AKS to have access to ARM resources.
Correspond to [this](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/resourcemanager/containerservice/armcontainerservice/models.go#L1530).)

**Environment:**

jackfrancis:
**What type of PR is this?**

/kind feature

This PR introduces the option to pass in a pre-existing user-assigned identity for AKS cluster identity. See the top two rows here:

- https://learn.microsoft.com/en-us/azure/aks/use-managed-identity#summary-of-managed-identities

Fixes #3239
Fixes #3261

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
AKS user-assigned identity for control plane and kubelet
```


api/v1beta1/azuremanagedcontrolplane_types.go
```diff
@@ -48,6 +48,20 @@ const (
 	ManagedControlPlaneOutboundTypeUserDefinedRouting ManagedControlPlaneOutboundType = "userDefinedRouting"
 )
 
+// ManagedControlPlaneIdentityType enumerates the values for managed control plane identity type.
+type ManagedControlPlaneIdentityType string
+
+const (
+	// ManagedControlPlaneIdentityTypeSystemAssigned Use an implicitly created system assigned managed identity to manage
+	// cluster resources. Master components in the control plane such as kube-controller-manager will use the
```
alexeldeib, line 56: language nit: I believe we prefer control plane these days

api/v1beta1/azuremanagedcontrolplane_webhook.go
```diff
@@ -669,3 +676,38 @@ func (m *AzureManagedControlPlane) validateIntegerStringGreaterThanZero(input *s
...
+// validateIdentityProfile validates an AutoScalerProfile.
+func (m *AzureManagedControlPlane) validateIdentityProfile(_ client.Client) error {
+	var allErrs field.ErrorList
+
+	if m.Spec.IdentityProfile != nil {
+		if m.Spec.IdentityProfile.Type == ManagedControlPlaneIdentityTypeUserAssigned {
+			if m.Spec.IdentityProfile.UserAssignedIdentityResourceID == "" {
+				allErrs = append(allErrs, field.Invalid(field.NewPath("Spec", "IdentityProfile", "UserAssignedIdentityResourceID"), m.Spec.IdentityProfile.UserAssignedIdentityResourceID, "cannot be empty if IdentityProfile.Type is UserAssigned"))
+			}
+			if m.Spec.IdentityProfile.UserAssignedIdentityPrincipalID == nil {
```
alexeldeib, line 689: have you tested this? I'm guessing you're following https://learn.microsoft.com/en-us/azure/aks/use-managed-identity ?

principal ID and client ID are _outputs_ there I thought, iirc you only need the UAMI ARM ID to pass to AKS. could be wrong.

api/v1beta1/azuremanagedcontrolplane_webhook.go
```diff
@@ -669,3 +676,38 @@ func (m *AzureManagedControlPlane) validateIntegerStringGreaterThanZero(input *s
...
+				allErrs = append(allErrs, field.Invalid(field.NewPath("Spec", "IdentityProfile", "UserAssignedIdentityPrincipalID"), m.Spec.IdentityProfile.UserAssignedIdentityPrincipalID, "cannot be empty if IdentityProfile.Type is UserAssigned"))
+			}
+			if m.Spec.IdentityProfile.UserAssignedIdentityClientID == nil {
+				allErrs = append(allErrs, field.Invalid(field.NewPath("Spec", "IdentityProfile", "UserAssignedIdentityClientID"), m.Spec.IdentityProfile.UserAssignedIdentityClientID, "cannot be empty if IdentityProfile.Type is UserAssigned"))
+			}
+		} else {
+			if m.Spec.IdentityProfile.UserAssignedIdentityResourceID != "" {
+				allErrs = append(allErrs, field.Invalid(field.NewPath("Spec", "IdentityProfile", "UserAssignedIdentityResourceID"), m.Spec.IdentityProfile.UserAssignedIdentityResourceID, "should be empty if IdentityProfile.Type is SystemAssigned"))
+			}
+			if m.Spec.IdentityProfile.UserAssignedIdentityPrincipalID != nil {
```
alexeldeib, line 699: I don't even think you should expose these fields except in status. but see above. 

if I'm mistaken, this makes sense to me.

alexeldeib (commented):
nice work :) not sure we actually need principal and client ID. I believe AKS will resolve these internally and emit them as outputs, and would ignore inputs. Worth double checking though.

I checked some of my dev HTTP requests in postman and they only use ARM ID, but that stuff circumvents ARM, so I'm not sure it's correct, especially for system identity (which is tricky in dev env)

CecileRobertMichon:
@jackfrancis are you still working on this one?

azure/services/managedclusters/spec.go
```diff
@@ -395,6 +398,21 @@ func (s *ManagedClusterSpec) Parameters(ctx context.Context, existing interface{
...
+	if s.IdentityProfile != nil {
+		managedCluster.Identity = &containerservice.ManagedClusterIdentity{
+			Type: containerservice.ResourceIdentityType(s.IdentityProfile.Type),
+		}
+		if s.IdentityProfile.Type == infrav1.ManagedControlPlaneIdentityTypeUserAssigned {
+			userAssignedIDValues := &containerservice.ManagedClusterIdentityUserAssignedIdentitiesValue{
+				PrincipalID: s.IdentityProfile.UserAssignedIdentityPrincipalID,
+				ClientID:    s.IdentityProfile.UserAssignedIdentityClientID,
+			}
+			managedCluster.Identity.UserAssignedIdentities = map[string]*containerservice.ManagedClusterIdentityUserAssignedIdentitiesValue{
```
jackfrancis, line 410: @alexeldeib thanks for the review (and sorry for the delayed response)

Are you suggesting that these are the wrong inputs to pass into the AKS API?

api/v1beta1/azuremanagedcontrolplane_types.go
```diff
@@ -48,6 +48,20 @@ const (
 	ManagedControlPlaneOutboundTypeUserDefinedRouting ManagedControlPlaneOutboundType = "userDefinedRouting"
 )
 
+// ManagedControlPlaneIdentityType enumerates the values for managed control plane identity type.
+type ManagedControlPlaneIdentityType string
```
luthermonson, line 52: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/api/v1beta1/types.go#L520-L531 can we just reuse these existing consts?

luthermonson:
> nice work :) not sure we actually need principal and client ID. I believe AKS will resolve these internally and emit them as outputs, and would ignore inputs. Worth double checking though.
> 
> I checked some of my dev HTTP requests in postman and they only use ARM ID, but that stuff circumvents ARM, so I'm not sure it's correct, especially for system identity (which is tricky in dev env)

Comparing to how the azurerm terraform provider solves this exact problem I believe you are correct and this feature is also designed to be receive multiple user assigned IDs. Please check out the following code:

https://github.com/hashicorp/terraform-provider-azurerm/blob/8bc0e5637346eeea808f5b63b97eed69835005b5/internal/services/containers/kubernetes_cluster_resource.go#L3265-L3283

I think a better API would be to reuse the idea in AzureMachinePools which is the (UserAssignedIdentities)[https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/exp/api/v1beta1/azuremachinepool_types.go#L144C6-L149] list and do a loop that looks like the one in the tf provider to add empty `UserAssignedIdentityDetails` with the ID as the map key.

api/v1beta1/azuremanagedcontrolplane_webhook.go
jackfrancis: It's a bit of a mystery!

E0619 20:02:48.596549       1 controller.go:329]  "msg"="Reconciler error" "error"="error creating AzureManagedControlPlane capz-e2e-rwuh71/capz-e2e-rwuh71-aks: failed to reconcile AzureManagedControlPlane service managedcluster: failed to create resource capz-e2e-rwuh71-aks/capz-e2e-rwuh71-aks (service: managedcluster): Code=\"InvalidIdentityValues\" Message=\"Invalid value for the identities '/subscriptions/<redacted>/resourcegroups/capz-ci/providers/Microsoft.ManagedIdentity/userAssignedIdentities/umd-test'. The 'UserAssignedIdentities' property keys should only be empty null during PATCH requests.\"" "AzureManagedControlPlane"={"name":"capz-e2e-rwuh71-aks","namespace":"capz-e2e-rwuh71"} "controller"="azuremanagedcontrolplane" "controllerGroup"="infrastructure.cluster.x-k8s.io" "controllerKind"="AzureManagedControlPlane" "name"="capz-e2e-rwuh71-aks" "namespace"="capz-e2e-rwuh71" "reconcileID"="b8fe5ded-9132-42cc-89ea-ae1621f4dbcf"

The above is in response to only passing in the URI as the key as one does here:

- https://github.com/Azure/azure-sdk-for-go/blob/v67.0.0/services/containerservice/mgmt/2022-03-01/containerservice/models.go#L1770-L1780

Your assertion that the principalID and clientID are output does check out with the type description here:

- https://github.com/Azure/azure-sdk-for-go/blob/v67.0.0/services/containerservice/mgmt/2022-03-01/containerservice/models.go#L1794-L1800

I'll try hacking around with a pointer to an empty `ManagedClusterIdentityUserAssignedIdentitiesValue` literal as the value instead of `nil` and see if it still complains about PATCH (or complains about anything), stand by....

jackfrancis: yep that worked!

api/v1beta1/azuremanagedcontrolplane_types.go
jackfrancis: I'd prefer to keep these strongly typed and distinct, but I have inherited the actual string const values from the vm consts

jackfrancis:
> > nice work :) not sure we actually need principal and client ID. I believe AKS will resolve these internally and emit them as outputs, and would ignore inputs. Worth double checking though.
> > I checked some of my dev HTTP requests in postman and they only use ARM ID, but that stuff circumvents ARM, so I'm not sure it's correct, especially for system identity (which is tricky in dev env)
> 
> Comparing to how the azurerm terraform provider solves this exact problem I believe you are correct and this feature is also designed to be receive multiple user assigned IDs. Please check out the following code:
> 
> https://github.com/hashicorp/terraform-provider-azurerm/blob/8bc0e5637346eeea808f5b63b97eed69835005b5/internal/services/containers/kubernetes_cluster_resource.go#L3265-L3283
> 
> I think a better API would be to reuse the idea in `AzureMachinePool` with the [UserAssignedIdentities](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/exp/api/v1beta1/azuremachinepool_types.go#L144C6-L149) list on `AzureManagedControlPlane` and do a loop that looks like the one in the tf provider to add empty `ManagedClusterIdentityUserAssignedIdentitiesValue` with the ID as the map key.

So it looks like AKS only accepts a single user-assigned ID (which makes sense):

E0620 15:58:58.430110       1 controller.go:329]  "msg"="Reconciler error" "error"="error creating AzureManagedControlPlane capz-e2e-n5ol1g/capz-e2e-n5ol1g-aks: failed to reconcile AzureManagedControlPlane service managedcluster: failed to create resource capz-e2e-n5ol1g-aks/capz-e2e-n5ol1g-aks (service: managedcluster): Code=\"BadRequest\" Message=\"The count of user-assigned identity specified in identity.userAssignedIdentities is 2, while expected value is 1.\"" "AzureManagedControlPlane"={"name":"capz-e2e-n5ol1g-aks","namespace":"capz-e2e-n5ol1g"} "controller"="azuremanagedcontrolplane" "controllerGroup"="infrastructure.cluster.x-k8s.io" "controllerKind"="AzureManagedControlPlane" "name"="capz-e2e-n5ol1g-aks" "namespace"="capz-e2e-n5ol1g" "reconcileID"="51a30f2d-bd9d-40ab-af2a-dec8b1ca4f9e"

I'll go back to using a single string interface in the CAPZ property definition so users aren't confused.

docs/book/src/topics/managedcluster.md
```diff
@@ -285,6 +285,27 @@ Current limitations
 - Only supports Azure Active Directory Managed by Azure.
   - We will not support Legacy Azure Active Directory
 
+### AKS Cluster Managed Identity
```
jackfrancis, line 288: @nojnhuh with your recent streamlining of AKS feature docs is there a better way to document this?

nojnhuh: I tried to establish some amount of a pattern in #3550 for what to include in the godoc comments for the type definitions. My initial sense is that what's included here overlaps with the godoc comments enough to justify not needing a separate section here. Definitely no hard rules though, so feel free to add anything here you think is valuable. I'd at least include the AKS doc link in the godoc in any case.

jackfrancis: godocs are generated after merge to main?

nojnhuh: The deploy preview for this PR has the generated version of these changes: https://deploy-preview-3425--kubernetes-sigs-cluster-api-provider-azure.netlify.app/reference/v1beta1-api.html#infrastructure.cluster.x-k8s.io/v1beta1.IdentityProfile

jackfrancis: Love it, will remove the add'l matter I added to the markdown.

azure/services/managedclusters/spec.go
```diff
@@ -581,3 +591,20 @@ func computeDiffOfNormalizedClusters(managedCluster containerservice.ManagedClus
 	diff := cmp.Diff(clusterNormalized, existingMCClusterNormalized)
 	return diff
 }
+
+func getIdentityProfile(identityProfile *infrav1.IdentityProfile) (managedClusterIdentity *containerservice.ManagedClusterIdentity, err error) {
+	if identityProfile.Type != "" {
+		managedClusterIdentity = &containerservice.ManagedClusterIdentity{
```
nojnhuh, line 597: nit: removing most of this level of indentation might make this a bit more readable:
```suggestion
	if identityProfile.Type == "" {
		return
	}
	managedClusterIdentity = &containerservice.ManagedClusterIdentity{
```

azure/services/managedclusters/spec.go
```diff
@@ -581,3 +591,20 @@ func computeDiffOfNormalizedClusters(managedCluster containerservice.ManagedClus
...
+			Type: containerservice.ResourceIdentityType(identityProfile.Type),
+		}
+		if managedClusterIdentity.Type == containerservice.ResourceIdentityTypeUserAssigned {
+			if identityProfile.UserAssignedIdentityResourceID == "" {
+				err = errors.Errorf("IdentityProfile is set to \"UserAssigned\" but no UserAssignedIdentityResourceID is present")
+				return
+			}
+			managedClusterIdentity.UserAssignedIdentities = make(map[string]*containerservice.ManagedClusterIdentityUserAssignedIdentitiesValue, 1)
+			managedClusterIdentity.UserAssignedIdentities[identityProfile.UserAssignedIdentityResourceID] = &containerservice.ManagedClusterIdentityUserAssignedIdentitiesValue{}
```
nojnhuh, line 606: nit: inlining this might be slightly less verbose? At the expense of some possible memory optimization Go does with `make(..., 1)` that doesn't happen with a literal.
```suggestion
			managedClusterIdentity.UserAssignedIdentities = map[string]*containerservice.ManagedClusterIdentityUserAssignedIdentitiesValue{
				identityProfile.UserAssignedIdentityResourceID: {},
			}
```

api/v1beta1/azuremanagedcontrolplane_webhook.go
```diff
@@ -676,3 +683,26 @@ func (m *AzureManagedControlPlane) validateIntegerStringGreaterThanZero(input *s
...
+	var allErrs field.ErrorList
+
+	if m.Spec.IdentityProfile != nil {
+		if m.Spec.IdentityProfile.Type == ManagedControlPlaneIdentityTypeUserAssigned {
+			if m.Spec.IdentityProfile.UserAssignedIdentityResourceID == "" {
+				allErrs = append(allErrs, field.Invalid(field.NewPath("Spec", "IdentityProfile", "UserAssignedIdentityResourceID"), m.Spec.IdentityProfile.UserAssignedIdentityResourceID, "cannot be empty if IdentityProfile.Type is UserAssigned"))
+			}
+		} else {
+			if m.Spec.IdentityProfile.UserAssignedIdentityResourceID != "" {
+				allErrs = append(allErrs, field.Invalid(field.NewPath("Spec", "IdentityProfile", "UserAssignedIdentityResourceID"), m.Spec.IdentityProfile.UserAssignedIdentityResourceID, "should be empty if IdentityProfile.Type is SystemAssigned"))
```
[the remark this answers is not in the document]
nawazkh: ^ looks like the implemented error is supposed to be shown to the user when `m.Spec.IdentityProfile.Type == ManagedControlPlaneIdentityTypeUserAssigned` and  `m.Spec.IdentityProfile.UserAssignedIdentityResourceID != ""` .
So "_`should`_ be empty if IdentityProfile.Type is SystemAssigned" is about right?

api/v1beta1/azuremanagedcontrolplane_webhook.go
```diff
@@ -676,3 +683,26 @@ func (m *AzureManagedControlPlane) validateIntegerStringGreaterThanZero(input *s
...
+			}
+		}
+	}
+
+	if len(allErrs) > 0 {
+		return kerrors.NewAggregate(allErrs.ToAggregate().Errors())
+	}
+
+	return nil
```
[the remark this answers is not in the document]
nawazkh: I agree with Jon that we could return `allErrs`, but is there a reason for aggregating the errors which I may be missing?

azure/services/managedclusters/spec.go
nawazkh: ^ in addition to the above change, we might want to add an error as well.

api/v1beta1/azuremanagedcontrolplane_webhook.go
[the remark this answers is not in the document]
nojnhuh: Ah yes, I can't read 🙃

api/v1beta1/azuremanagedcontrolplane_webhook.go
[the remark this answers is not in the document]
nojnhuh: If you mean "why do we return multiple errors" then in general, we want to return as much information as possible to the user for each request so they get one "spec.A is invalid, spec.B is invalid, ..." error on the first request vs. "spec.A is invalid." Fix spec.A. "spec.B is invalid." Fix spec.B...

If you mean "why are we calling `kerrors.NewAggregate` then that's because that's one way of turning a `[]error` into a single `error`.

api/v1beta1/azuremanagedcontrolplane_types.go
```diff
@@ -48,6 +48,20 @@ const (
...
+const (
+	// ManagedControlPlaneIdentityTypeSystemAssigned Use an implicitly created system assigned managed identity to manage
+	// cluster resources. Components in the control plane such as kube-controller-manager will use the
+	// system assigned managed identity to manipulate Azure resources.
+	ManagedControlPlaneIdentityTypeSystemAssigned ManagedControlPlaneIdentityType = ManagedControlPlaneIdentityType(VMIdentitySystemAssigned)
+	// ManagedControlPlaneIdentityTypeUserAssigned Use a user-specified identity to manage cluster resources.
+	// Components in the control plane such as kube-controller-manager will use the specified user assigned
+	// managed identity to manipulate Azure resources.
+	ManagedControlPlaneIdentityTypeUserAssigned ManagedControlPlaneIdentityType = ManagedControlPlaneIdentityType(VMIdentityUserAssigned)
+)
```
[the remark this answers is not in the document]
jackfrancis: Good question. The current implementation simply hard-codes the AKS cluster config to use system-assigned identity. The scope of this is to add user-assigned, so I hadn't considered including backfilling enabling service principal config.

api/v1beta1/azuremanagedcontrolplane_types.go
```diff
@@ -421,6 +439,18 @@ const (
 	ExpanderRandom Expander = "random"
 )
 
+// IdentityProfile represents the Identity configuration for an AKS control plane.
```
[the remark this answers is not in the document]
jackfrancis: love

api/v1beta1/azuremanagedcontrolplane_webhook.go
```diff
@@ -676,3 +683,26 @@ func (m *AzureManagedControlPlane) validateIntegerStringGreaterThanZero(input *s
 
 	return allErrs
 }
+
+// validateIdentityProfile validates an AutoScalerProfile.
+func (m *AzureManagedControlPlane) validateIdentityProfile(_ client.Client) error {
+	var allErrs field.ErrorList
+
+	if m.Spec.IdentityProfile != nil {
```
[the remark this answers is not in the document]
jackfrancis: I can test

api/v1beta1/azuremanagedcontrolplane_webhook.go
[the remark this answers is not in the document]
jackfrancis: fun with natural language inversions!

api/v1beta1/azuremanagedcontrolplane_webhook.go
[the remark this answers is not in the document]
jackfrancis: I think I'd have to refactor more to use the simpler approach, as the `Validate` method is doing the error wrapping by enumerating through a bunch of validator funcs (`validateIdentityProfile` is a new one I added here) and each of those funcs is returning an `err`.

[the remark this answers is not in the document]
nojnhuh: Ah I missed that, this is definitely fins as-is then IMO.

api/v1beta1/azuremanagedcontrolplane_types.go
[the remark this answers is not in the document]
nojnhuh: This would be easy and safe enough to add later if it's not a simple oversight.

azure/services/managedclusters/spec.go
```diff
@@ -581,3 +591,20 @@ func computeDiffOfNormalizedClusters(managedCluster containerservice.ManagedClus
...
 }
+
+func getIdentityProfile(identityProfile *infrav1.IdentityProfile) (managedClusterIdentity *containerservice.ManagedClusterIdentity, err error) {
+	if identityProfile.Type != "" {
+		managedClusterIdentity = &containerservice.ManagedClusterIdentity{
+			Type: containerservice.ResourceIdentityType(identityProfile.Type),
+		}
+		if managedClusterIdentity.Type == containerservice.ResourceIdentityTypeUserAssigned {
+			if identityProfile.UserAssignedIdentityResourceID == "" {
+				err = errors.Errorf("IdentityProfile is set to \"UserAssigned\" but no UserAssignedIdentityResourceID is present")
```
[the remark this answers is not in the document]
jackfrancis: I initially excluded the error check but then when I went to write the UT I thought "do we really want to rely upon the webhook forever to make this function sane?" and decided including an error response as a low-maintenance, high-defense outcome. So I think I still prefer it.

azure/services/managedclusters/spec.go
jackfrancis: we are returning an error it's just that this func is using named response variables so things are maybe a little cute (but using a named `*containerservice.ManagedClusterIdentity` response variable in particular does clean up the code a bit)

azure/services/managedclusters/spec.go
jackfrancis: deal

api/v1beta1/azuremanagedcontrolplane_types.go
```diff
@@ -161,6 +175,10 @@ type AzureManagedControlPlaneSpec struct {
 	// - USGovernmentCloud: "AzureUSGovernmentCloud"
 	// +optional
 	AzureEnvironment string `json:"azureEnvironment,omitempty"`
+
+	// IdentityProfile configuration used by the AKS control plane.
+	// +optional
+	IdentityProfile *IdentityProfile `json:"identityProfile,omitempty"`
```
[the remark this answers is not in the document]
jackfrancis: I don't love naming the struct just `identity` as that sort of suggests a shorthand property key for the type of identity.

api/v1beta1/azuremanagedcontrolplane_types.go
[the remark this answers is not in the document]
jackfrancis: Yeah I think it should be something we add in reponse to user demand (system-assigned identity is definitely the happy path IMO)

api/v1beta1/azuremanagedcontrolplane_types.go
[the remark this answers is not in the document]
nojnhuh: If we decide we need to expose `properties.identityProfile` is there a reasonable name we could use for it in the CAPZ API?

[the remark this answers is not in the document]
jackfrancis: Actually I was misunderstood your initial comment. You're right, this really is a sort of worst-case scenario in terms of API semantics. :(

From a test cluster built from this branch and configured for control plane user-assigned identity:

```
...
        "identityProfile": {
            "kubeletidentity": {
                "resourceId": "/subscriptions/<redacted>/resourcegroups/MC_capz-e2e-hikyno-aks_capz-e2e-hikyno-aks_eastus/providers/Microsoft.ManagedIdentity/userAssignedIdentities/capz-e2e-hikyno-aks-agentpool",
                "clientId": "<redacted>",
                "objectId": "<redacted>"
            }
        },
    "identity": {
        "type": "UserAssigned",
        "userAssignedIdentities": {
            "/subscriptions/<redacted>/resourcegroups/capz-ci/providers/Microsoft.ManagedIdentity/userAssignedIdentities/umd-test": {
                "clientId": "<redacted>",
                "principalId": "<redacted>"
            }
        }
    },
...
```

The latter is the one that we're using w/ the new CAPZ interface here.

Would it be easier to implement the kubelet `identityProfile` in this PR as well so we ensure we do them both right w/ respect to one another? (I don't think there's that much more work.)

In either case yeah we probably wanna rename to `identity` for the control plane identity interface.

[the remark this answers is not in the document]
nojnhuh: I'm not sure I fully understand the differences between those fields. Do those actually overlap at all and where it's CAPZ's responsibility to keep them in sync?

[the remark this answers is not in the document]
jackfrancis: They're the first two in the list here:

- https://learn.microsoft.com/en-us/azure/aks/use-managed-identity#summary-of-managed-identities

api/v1beta1/azuremanagedcontrolplane_webhook.go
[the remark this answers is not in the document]
jackfrancis: The AKS is accepting changes to the fields. Though the transition between, e.g., changing from one user-assigned identity resource to another, seemed very slow (cluster was in an Updating state for a long time).

For now I think we should keep these fields mutable (no webhook violations), but hold off on adding E2E tests that validate these updates in practice until we can get a better handle on how reliable these operations are in practice.

azure/services/managedclusters/spec.go
```diff
@@ -361,23 +367,7 @@ func (s *ManagedClusterSpec) Parameters(ctx context.Context, existing interface{
...
-			managedCluster.NetworkProfile.LoadBalancerProfile.OutboundIPPrefixes = &containerservice.ManagedClusterLoadBalancerProfileOutboundIPPrefixes{
-				PublicIPPrefixes: convertToResourceReferences(s.LoadBalancerProfile.OutboundIPPrefixes),
-			}
-		}
-		if len(s.LoadBalancerProfile.OutboundIPs) > 0 {
-			managedCluster.NetworkProfile.LoadBalancerProfile.OutboundIPs = &containerservice.ManagedClusterLoadBalancerProfileOutboundIPs{
-				PublicIPs: convertToResourceReferences(s.LoadBalancerProfile.OutboundIPs),
-			}
-		}
+		managedCluster.NetworkProfile.LoadBalancerProfile = s.GetLoadBalancerProfile()
```
jackfrancis, line 370: This was done to appease the cyclomatic complexity gods.

jackfrancis:
/assign @mboersma 

Matt, are you able to review and approve this so that it lands in the v1.10 milestone? This ticks off two features tracked by @dtzar in #3446

api/v1beta1/azuremanagedcontrolplane_types.go
```diff
@@ -48,6 +48,20 @@ const (
 	ManagedControlPlaneOutboundTypeUserDefinedRouting ManagedControlPlaneOutboundType = "userDefinedRouting"
 )
 
+// ManagedControlPlaneIdentityType enumerates the values for managed control plane identity type.
+type ManagedControlPlaneIdentityType string
+
+const (
+	// ManagedControlPlaneIdentityTypeSystemAssigned Use an implicitly created system assigned managed identity to manage
```
mboersma, line 55:
```suggestion
	// ManagedControlPlaneIdentityTypeSystemAssigned Use an implicitly created system-assigned managed identity to manage
```

api/v1beta1/azuremanagedcontrolplane_types.go
```diff
@@ -48,6 +48,20 @@ const (
...
+	// cluster resources. Components in the control plane such as kube-controller-manager will use the
+	// system assigned managed identity to manipulate Azure resources.
```
mboersma, line 57:
```suggestion
	// system-assigned managed identity to manipulate Azure resources.
```

api/v1beta1/azuremanagedcontrolplane_types.go
```diff
@@ -48,6 +48,20 @@ const (
...
+	ManagedControlPlaneIdentityTypeSystemAssigned ManagedControlPlaneIdentityType = ManagedControlPlaneIdentityType(VMIdentitySystemAssigned)
+	// ManagedControlPlaneIdentityTypeUserAssigned Use a user-specified identity to manage cluster resources.
+	// Components in the control plane such as kube-controller-manager will use the specified user assigned
```
mboersma, line 60:
```suggestion
	// Components in the control plane such as kube-controller-manager will use the specified user-assigned
```

mboersma (commented):
Looks good to moi, just had some minor documentation comments.

mboersma:
/test pull-cluster-api-provider-azure-apidiff

`Pod could not be created...` prow failure.

dtzar:
Assuming the provider azure test passes, are we good to merge this @mboersma ?

mboersma (approved):
/lgtm
/approve

maciaszczykm:
@jackfrancis This change seems to break reconcile process for me. Control plane sticks in updating status, after debug found out that diff (azure/services/managedclusters/spec.go:415) is not empty:

```diff
  &containerservice.ManagedCluster{
  	... // 2 identical fields
  	ExtendedLocation: nil,
  	Identity:         &{Type: "SystemAssigned"},
  	ManagedClusterProperties: &containerservice.ManagedClusterProperties{
  		... // 22 identical fields
  		APIServerAccessProfile: nil,
  		DiskEncryptionSetID:    nil,
- 		IdentityProfile:        nil,
+ 		IdentityProfile: map[string]*containerservice.UserAssignedIdentity{
+ 			"kubeletidentity": &{
+ 				ResourceID: &"/subscriptions/[REMOVED]/resourcegrou"...,
+ 			},
+ 		},
  		PrivateLinkResources: nil,
  		DisableLocalAccounts: nil,
  		... // 3 identical fields
  	},
  	Tags:     nil,
  	Location: nil,
  	... // 4 identical fields
  }
```

I did not specify this field. It looks like it was filled in on the AKS side but not in control plane resource.

dtzar:
@maciaszczykm can you please file an issue for this?

nojnhuh:
PR open to fix: #3682

PR #3428, opened 2023-04-14
[WIP] E2E: enforce timeout when creating SSH client connection

jackfrancis:
**What type of PR is this?**

Let's see what happens when we short-circuit during log collection ssh client connection foo...

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/helpers.go
+++ b/test/e2e/helpers.go
@@ -482,10 +482,22 @@ func getProxiedSSHClient(controlPlaneEndpoint, hostname, port string) (*ssh.Clie
 		return nil, errors.Wrapf(err, "dialing from control plane to target node at %s", hostname)
 	}
 
-	// Establish an authenticated SSH conn over the client -> control plane -> target transport
-	conn, chans, reqs, err := ssh.NewClientConn(c, hostname, config)
-	if err != nil {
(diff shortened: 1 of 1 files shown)


test/e2e/helpers.go
```diff
@@ -482,10 +482,22 @@ func getProxiedSSHClient(controlPlaneEndpoint, hostname, port string) (*ssh.Clie
 		return nil, errors.Wrapf(err, "dialing from control plane to target node at %s", hostname)
 	}
 
-	// Establish an authenticated SSH conn over the client -> control plane -> target transport
```
nojnhuh, line 485: I think what we're trying to do here is the same as adding the following:
```go
	err = c.SetDeadline(time.Now().Add(1 * time.Minute))
	if err != nil {
		return nil, err
	}
```

That timeout would be for the whole SSH command, so not sure what the best value for that would be. One random successful e2e run I looked at shows all of the Windows log collection commands together taking about 2.5m.

nojnhuh:
Closing in favor of #3444

PR #3450, opened 2023-04-18
add ASO install

### LINKED ISSUE #3520: Integrate ASO install with `clusterctl init`, `tilt up`

nojnhuh:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**: This PR adds the kustomize config necessary to install ASO. These changes deliberately omit integration with `tilt` or `clusterctl init` so ASO doesn't get installed before it can be used. This other branch on my fork includes the necessary foo to wire up the install if you'd like to try it yourself: https://github.com/nojnhuh/cluster-api-provider-azure/tree/aso-wired

Fixes #3520

- [ ] cherry-pick candidate

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```


config/aso/credentials.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: aso-controller-settings
+type: Opaque
+data:
+  AZURE_SUBSCRIPTION_ID: ${AZURE_SUBSCRIPTION_ID_B64:=""}
```
CecileRobertMichon, line 7: is this temporary until we can use workload identity?

CecileRobertMichon (approved):
lgtm

config/aso/credentials.yaml
nojnhuh: I was thinking auth for ASO would essentially mirror the AzureClusterIdentity for the Cluster, which could also be Service Principal. Or would it make sense to always configure ASO to use workload identity? I still don't have a clear idea of exactly how workload identity works so I couldn't quite figure out how to get that set up based on the docs: https://azure.github.io/azure-service-operator/guide/authentication/#azure-workload-identity

I was also thinking having a default set of credentials here would make it easy for users to get started using ASO directly even if CAPZ can work around relying on the credentials here.

config/aso/credentials.yaml
```diff
@@ -0,0 +1,10 @@
...
+  AZURE_TENANT_ID: ${AZURE_TENANT_ID_B64:=""}
```
Jont828, line 8: Just curious, is there a reason we want to add the `:=""` part to the variable references?

config/aso/kustomization.yaml
```diff
@@ -0,0 +1,53 @@
+apiVersion: kustomize.config.k8s.io/v1alpha1
+kind: Component
+namespace: capz-system
+resources:
+- https://github.com/Azure/azure-service-operator/releases/download/v2.0.0/azureserviceoperator_v2.0.0.yaml
+- https://github.com/Azure/azure-service-operator/releases/download/v2.0.0/azureserviceoperator_customresourcedefinitions_v2.0.0.yaml
+- credentials.yaml
+
+patches:
+  - patch: |- # default kustomization includes a namespace already
```
Jont828, line 10: Nit: do we want a line break at the end? Not sure if it's better to use `|-` or `|` in this case.

Jont828 (commented):
A couple small things but overall looks very solid!

config/aso/credentials.yaml
nojnhuh: I admit this was a shameless copy-paste from here and didn't really consider whether it's necessary or not: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/fec6314ebb00355ecc617270f40c6367d7d4d75f/config/default/credentials.yaml#L7-L8

It looks like ASO will get stuck in a crash loop either way when the vars aren't defined, but with `:=""` it doesn't crash until an ASO resource is created, whereas without `:=""` it'll crash as soon as it starts up even when no ASO resources exist.

CecileRobertMichon: For the CAPZ manager credentials, the idea is that you can optionally define global credentials but it's also possible to define per Cluster credentials via AzureClusterIdentity, so we don't want the credentials to be required when installing CAPZ as they aren't actually required until you create your first cluster (without the defaulting to "", which is what `:=""` does, clusterctl init would fail when the environment variables are not set).

nojnhuh: So I think with `:=""` is what we want then. Eventually, a set of ASO credentials will be generated for each individual cluster from an AzureClusterIdentity if one exists, so those users can still choose not to define the global credentials and ASO will work fine with CAPZ until the user tries to create an ASO resource themselves without specifying credentials and relying on the defaults.

nojnhuh:
I just noticed that CAPI Tilt assumes the root of the kustomize config is `config/default` (and can't be configured) where this changes wants it to be `config/release`. I haven't tried it yet, but 'm pretty sure that means spinning up CAPZ with CAPI Tilt won't deploy ASO. Should I modify the current `default` to be something like `capz` and then change `release` -> `default`?

Originally I didn't integrate ASO directly with `config/default` because some of the config applying to all resources (like `namePrefix`) would have messed up some of the ASO resources. I can try that again though to see if I can get that to work.

nojnhuh:
I managed to get this working from the existing `config/default`, but it feels like kind of a mess: https://github.com/nojnhuh/cluster-api-provider-azure/commit/39550967

Thoughts on this approach vs. renaming the current `config/default`?

CecileRobertMichon:
> CAPI Tilt assumes the root of the kustomize config is config/default (and can't be configured)

Have you considered making it configurable as an alternative?

nojnhuh:
> Have you considered making it configurable as an alternative?

Kind of, didn't take the time yet to figure out exactly how that might work though.
 
> another idea: what if we renamed `default` to something else, and `config/release` becomes `config/default`?

Yeah I think this will be the easiest option. I'll update this PR next week with that change.

nojnhuh:
/hold

I think I broke something here :/

```
Error from server (InternalError): error when creating "jon/cluster.yaml": Internal error occurred: failed calling webhook "default.azuremanagedcontrolplanes.infrastructure.cluster.x-k8s.io": failed to call webhook: Post "https://capz-webhook-service.capz-system.svc:443/mutate-infrastructure-cluster-x-k8s-io-v1beta1-azuremanagedcontrolplane?timeout=10s": x509: certificate is valid for azureserviceoperator-webhook-service.capz-system.svc, azureserviceoperator-webhook-service.capz-system.svc.cluster.local, not capz-webhook-service.capz-system.svc
```

nojnhuh:
> /hold
> 
> I think I broke something here :/
> 
> ```
> Error from server (InternalError): error when creating "jon/cluster.yaml": Internal error occurred: failed calling webhook "default.azuremanagedcontrolplanes.infrastructure.cluster.x-k8s.io": failed to call webhook: Post "https://capz-webhook-service.capz-system.svc:443/mutate-infrastructure-cluster-x-k8s-io-v1beta1-azuremanagedcontrolplane?timeout=10s": x509: certificate is valid for azureserviceoperator-webhook-service.capz-system.svc, azureserviceoperator-webhook-service.capz-system.svc.cluster.local, not capz-webhook-service.capz-system.svc
> ```

/hold cancel

Just pushed a fix for this. capz-webhook-service's selector was only selecting based on the `cluster.x-k8s.io/provider: "infrastructure-azure"` label which included the ASO controller manager so some webhook requests for CAPZ resources were being directed to ASO instead of CAPZ.

nojnhuh:
@Jont828 @CecileRobertMichon This is ready for another look whenever you have time.

PR #3454, opened 2023-04-19
Fix ci-entrypoint.sh array export for macOS

### LINKED ISSUE #3292: Custom CCM CI build flows don't work on macos
The `export CCM_IMG_ARGS` part of `ci-entrypoint.sh` doesn't work on macos. Let's decompose the helm args in a way that works all common bash environments.

One way would be to use local values.yaml files, e.g.:

```bash
cat <<EOF > "${CLOUD_PROVIDER_HELM_VALUES_CUSTOM_BUILD}"
 cloudControllerManager:
   imageRepository: "${IMAGE_REGISTRY}"
   imageName: "${CCM_IMAGE_NAME}"
   imageTag: "${IMAGE_TAG}"
 cloudNodeManager:
   imageRepository: "${IMAGE_REGISTRY}"
   imageName: "${CNM_IMAGE_NAME}"
   imageTag: "${IMAGE_TAG}"
 EOF
```

except do the above using kustomize, not inline in ci-entrypoint

There are lots of references in this stackoverflow thread:

- https://stackoverflow.com/questions/5564418/exporting-an-array-in-bash-script

/kind failing-test

CecileRobertMichon:
**What type of PR is this?**

/kind cleanup

Fixes #3292 

- [x] cherry-pick candidate

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/scripts/ci-entrypoint.sh
+++ b/scripts/ci-entrypoint.sh
@@ -58,7 +58,7 @@ setup() {
         echo "Will use the ${IMAGE_REGISTRY}/${CCM_IMAGE_NAME}:${IMAGE_TAG} cloud-controller-manager image for external cloud-provider-cluster"
         echo "Will use the ${IMAGE_REGISTRY}/${CNM_IMAGE_NAME}:${IMAGE_TAG} cloud-node-manager image for external cloud-provider-azure cluster"
 
-        export CCM_IMG_ARGS=(--set cloudControllerManager.imageRepository="${IMAGE_REGISTRY}"
+        CCM_IMG_ARGS=(--set cloudControllerManager.imageRepository="${IMAGE_REGISTRY}"
         --set cloudNodeManager.imageRepository="${IMAGE_REGISTRY}"
         --set cloudControllerManager.imageName="${CCM_IMAGE_NAME}"
         --set cloudNodeManager.imageName="${CNM_IMAGE_NAME}"


scripts/ci-entrypoint.sh
```diff
@@ -58,7 +58,7 @@ setup() {
         echo "Will use the ${IMAGE_REGISTRY}/${CCM_IMAGE_NAME}:${IMAGE_TAG} cloud-controller-manager image for external cloud-provider-cluster"
         echo "Will use the ${IMAGE_REGISTRY}/${CNM_IMAGE_NAME}:${IMAGE_TAG} cloud-node-manager image for external cloud-provider-azure cluster"
 
-        export CCM_IMG_ARGS=(--set cloudControllerManager.imageRepository="${IMAGE_REGISTRY}"
+        CCM_IMG_ARGS=(--set cloudControllerManager.imageRepository="${IMAGE_REGISTRY}"
```
CecileRobertMichon, line 61: verified this fixes the issue on macOS @jackfrancis

jackfrancis: This is a very simple fix, thanks!

Is the root misunderstanding here that you must export a variable in order to re-use it in another function scope? I see it is declared and assigned a value in the `setup` func, and then referenced in the `install_cloud_provider_azure` func.

CecileRobertMichon: Right, here this variable doesn't need to be exported since it's only used later in the same script. In most cases the `export` would be harmless but in this case because arrays + export doesn't work with older versions of bash we're better off not exporting.

jackfrancis: Nothing makes me feel better than being reminded that our world depends on code that uses variables with no function scope.

Jont828 (commented):
I can test this as well if we need but otherwise LGTM.

PR #3500, opened 2023-05-01
fix tilt live reload for capz-controller-manager

### LINKED ISSUE #3497: Tilt live reload for capz-controller-manager is broken
/kind bug

Tilt's live reloading no longer works for capz-controller-manager since https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3399. When a relevant file is changed, Tilt instead hangs at these logs trying to update capz-controller-manager while the old version still continues to run:
```
Will copy 1 file(s) to container: [capz-controller-manager-c97b7db77-cqqg9/manager]
- '/home/jon/github/capz/.tiltbuild/manager' --> '/manager'
tar: can't remove old file manager: Permission denied
```

Live reload works the way it used to.

I tried adding `--chmod=0777` to each of these but that didn't seem to help:
https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/7af4c118c1df5da6f4494cb542ad38c49af2a75c/Tiltfile#L122-L125

**Environment:**

nojnhuh:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**: This PR fixes an issue where Tilt was failing to perform live updates of capz-controller-manager due to insufficient permissions on the parent directory of the `manager` binary. This change moves the Tilt-specific files in the CAPZ container into a new `/tilt` directory with world-writeable permissions so the non-root user running the container newly specified in #3399 is able to delete the old `manager` binary.

Fixes #3497

- [ ] cherry-pick candidate (#3399 was not cherry picked so release branches are not affected)

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -118,7 +118,8 @@ RUN wget --output-document /restart.sh --quiet https://raw.githubusercontent.com
 
 tilt_dockerfile_header = """
 FROM gcr.io/distroless/base:debug as tilt
-WORKDIR /
+WORKDIR /tilt
+RUN ["/busybox/chmod", "0777", "."]
 COPY --from=tilt-helper /process.txt .
 COPY --from=tilt-helper /start.sh .
 COPY --from=tilt-helper /restart.sh .
@@ -219,7 +220,7 @@ def capz():
         tilt_dockerfile_header,
     ])
 
-    entrypoint = ["sh", "/start.sh", "/manager"]
+    entrypoint = ["sh", "/tilt/start.sh", "/tilt/manager"]
     extra_args = settings.get("extra_args")
     if extra_args:
         entrypoint.extend(extra_args)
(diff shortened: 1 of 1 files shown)


mboersma (commented):
/lgtm

Thank you for fixing this! I tested it locally and it seems to work.

I assume this was another victim of the `distroless` update? We had to do a similar permissions hack for `process.txt`.

nojnhuh:
> I assume this was another victim of the `distroless` update? We had to do a similar permissions hack for `process.txt`.

What changed in the distroless image? I think this issue was more related to the capz-controller-manager container now running as non-root. But since the Tilt-flavored image is built as root, then the tilt scripts and manager binary are all owned by root and not world-writeable which the live reload steps would need to run as non-root.

CecileRobertMichon (commented):
/approve

PR #3503, opened 2023-05-01
Enable working machinepool drain tests

### LINKED ISSUE #2159: Enable working machinepool drain tests
/kind bug

Recently we discovered the the machinpool drain tests run during capz E2E don't actually check for success/fail. While enabling that we observed that the original, serial validation implementation is flaky.

This issue is to remind us that we actually aren't functionally testing machinepool drain, and we do want to do that.

The deliverable is "implement working E2E test coverage that validates successful machinepoolmachine drain as a side-effect of the delete operation.

**Environment:**

willie-yao:
**What type of PR is this?**
/kind failing-test

This PR re-adds a working version of the machine pool drain test.

Fixes #2159

- [x] cherry-pick candidate

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Enable working machinepool drain tests
```

### DIFF
--- a/test/e2e/azure_machinepool_drain.go
+++ b/test/e2e/azure_machinepool_drain.go
@@ -41,13 +41,16 @@ import (
 	expv1 "sigs.k8s.io/cluster-api/exp/api/v1beta1"
 	"sigs.k8s.io/cluster-api/test/framework"
 	"sigs.k8s.io/cluster-api/util"
+	"sigs.k8s.io/cluster-api/util/conditions"
 	"sigs.k8s.io/cluster-api/util/patch"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
 const (
-	AzureMachinePoolDrainSpecName   = "azure-mp-drain"
-	waitforResourceOperationTimeout = 30 * time.Second
+	AzureMachinePoolDrainSpecName        = "azure-mp-drain"
+	waitforResourceOperationTimeout      = 30 * time.Second
+	azureMachinePoolMachineTestFinalizer = "azuremachinepoolmachine.infrastructure.cluster.x-k8s.io/test"
 )
 
 // AzureMachinePoolDrainSpecInput is the input for AzureMachinePoolDrainSpec.
@@ -145,10 +148,16 @@ func testMachinePoolCordonAndDrain(ctx context.Context, mgmtClusterProxy, worklo
 	Expect(err).NotTo(HaveOccurred())
 	labelNodesWithMachinePoolName(ctx, workloadClusterProxy.GetClient(), amp.Name, ampmls)
 
+	By("adding the AzureMachinePoolMachine finalizer to the machine pool machines")
+	for i := range ampmls {
+		controllerutil.AddFinalizer(&ampmls[i], azureMachinePoolMachineTestFinalizer)
+	}
+
 	By(fmt.Sprintf("deploying a publicly exposed HTTP service with pod anti-affinity on machine pool: %s/%s", amp.Namespace, amp.Name))
 	_, _, _, cleanup := deployHTTPService(ctx, clientset, isWindows, customizers...)
 	defer cleanup()
 
+	var decreasedReplicas int32
 	By(fmt.Sprintf("decreasing the replica count by 1 on the machine pool: %s/%s", amp.Namespace, amp.Name))
 	Eventually(func() error {
 		helper, err := patch.NewHelper(owningMachinePool, mgmtClusterProxy.GetClient())
@@ -157,13 +166,39 @@ func testMachinePoolCordonAndDrain(ctx context.Context, mgmtClusterProxy, worklo
 			return err
 		}
 
-		decreasedReplicas := *owningMachinePool.Spec.Replicas - int32(1)
+		decreasedReplicas = *owningMachinePool.Spec.Replicas - int32(1)
 		owningMachinePool.Spec.Replicas = &decreasedReplicas
 		return helper.Patch(ctx, owningMachinePool)
 	}, 3*time.Minute, 3*time.Second).Should(Succeed())
 
-	// TODO setup a watcher to validate expected 2nd order drain outcomes
-	// https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/2159
+	By(fmt.Sprintf("checking for a machine to start draining for machine pool: %s/%s", amp.Namespace, amp.Name))
+	Eventually(func() error {
+		ampmls, err := getAzureMachinePoolMachines(ctx, mgmtClusterProxy, workloadClusterProxy, amp)
+		if err != nil {
+			LogWarning(errors.Wrap(err, "failed to list the azure machine pool machines").Error())
+			return err
+		}
+
+		for i := range ampmls {
+			if conditions.IsTrue(&ampmls[i], clusterv1.DrainingSucceededCondition) {
+				controllerutil.RemoveFinalizer(&ampmls[i], azureMachinePoolMachineTestFinalizer)
+				return nil // started draining the node prior to delete
+			}
+		}
+
+		return errors.New("no machine has started to drain")
+	}, 10*time.Minute, 3*time.Second).Should(Succeed())
+
+	By(fmt.Sprintf("waiting for the machine pool to scale down to %d replicas: %s/%s", decreasedReplicas, amp.Namespace, amp.Name))
+	Eventually(func() (int32, error) {
+		mp, err := getOwnerMachinePool(ctx, mgmtClusterProxy.GetClient(), amp.ObjectMeta)
+		if err != nil {
+			LogWarning(err.Error())
+			return 0, err
+		}
+
+		return int32(len(mp.Spec.ProviderIDList)), nil
+	}, 10*time.Minute, 3*time.Second).Should(Equal(decreasedReplicas))
 }
 
 func labelNodesWithMachinePoolName(ctx context.Context, workloadClient client.Client, mpName string, ampms []infrav1exp.AzureMachinePoolMachine) {
--- a/test/e2e/azure_test.go
+++ b/test/e2e/azure_test.go
@@ -431,17 +431,16 @@ var _ = Describe("Workload cluster creation", func() {
 				})
 			})
 
-			// TODO: Implement more robust cordon and drain test
-			// By("Cordon and draining a node", func() {
-			// 	AzureMachinePoolDrainSpec(ctx, func() AzureMachinePoolDrainSpecInput {
-			// 		return AzureMachinePoolDrainSpecInput{
-			// 			BootstrapClusterProxy: bootstrapClusterProxy,
-			// 			Namespace:             namespace,
-			// 			ClusterName:           clusterName,
-			// 			SkipCleanup:           skipCleanup,
-			// 		}
-			// 	})
-			// })
+			By("Cordon and draining a node", func() {
+				AzureMachinePoolDrainSpec(ctx, func() AzureMachinePoolDrainSpecInput {
+					return AzureMachinePoolDrainSpecInput{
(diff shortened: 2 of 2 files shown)


willie-yao:
I added this [second check](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3503/files#diff-ef1035bcafba9959d0df4cb4d9544deafccfb9d9a9f7e89cbcf1cb8565a89880R191) for machine pool machines to equal the replica count due to the E2E test reporting an error trying to get logs from the removed machine:

```
Failed to get logs for MachinePool capz-e2e-d65x99-vmss-mp-win, Cluster capz-e2e-d65x99/capz-e2e-d65x99-vmss: [dialing from control plane to target node at win-p-win000001: ssh: rejected: connect failed (Temporary failure in name resolution), Unable to collect VMSS Boot Diagnostic logs: failed to get boot diagnostics data: compute.VirtualMachineScaleSetVMsClient#RetrieveBootDiagnosticsData: Failure responding to request: StatusCode=404 -- Original Error: autorest/azure: Service returned an error. Status=404 Code="NotFound" Message="The entity was not found in this Azure location.", [running command "ls 'c:\localdumps' -Recurse": Process exited with status 1, getting a new sftp client connection: ssh: subsystem request failed]]
```

It looks like this log is still showing up due to the providerID list not being up to date. Will look into this issue further.

test/e2e/azure_machinepool_drain.go
```diff
@@ -145,10 +147,16 @@ func testMachinePoolCordonAndDrain(ctx context.Context, mgmtClusterProxy, worklo
 	Expect(err).NotTo(HaveOccurred())
 	labelNodesWithMachinePoolName(ctx, workloadClusterProxy.GetClient(), amp.Name, ampmls)
 
+	By("adding the AzureMachinePoolMachine finalizer to the machine pool machines")
```
nojnhuh, line 150: Why is the test adding and removing the finalizer? That's something I'd normally expect the controller to do already.

test/e2e/azure_machinepool_drain.go
```diff
@@ -157,13 +165,39 @@ func testMachinePoolCordonAndDrain(ctx context.Context, mgmtClusterProxy, worklo
...
+	By(fmt.Sprintf("checking for a machine to start draining for machine pool: %s/%s", amp.Namespace, amp.Name))
+	Eventually(func() error {
+		ampmls, err := getAzureMachinePoolMachines(ctx, mgmtClusterProxy, workloadClusterProxy, amp)
+		if err != nil {
+			LogWarning(errors.Wrap(err, "failed to list the azure machine pool machines").Error())
+			return err
+		}
+
+		for i := range ampmls {
+			if conditions.Has(&ampmls[i], clusterv1.DrainingSucceededCondition) && conditions.IsFalse(&ampmls[i], clusterv1.DrainingSucceededCondition) {
```
nojnhuh, line 182: nit: I think `Has` is redundant here since `IsFalse` returns false if the condition doesn't exist:
```suggestion
			if conditions.IsFalse(&ampmls[i], clusterv1.DrainingSucceededCondition) {
```

test/e2e/azure_test.go
```diff
@@ -432,16 +432,16 @@ var _ = Describe("Workload cluster creation", func() {
 			})
 
 			// TODO: Implement more robust cordon and drain test
```
nojnhuh, line 434: Does this PR address this comment such that we can remove it?

test/e2e/azure_machinepool_drain.go
willie-yao: I think the test was flaky before because the machine will start to delete before the test has time to check for a successful drain. This was under Cecile's suggestion here: https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/2159#issuecomment-1075383407

nojnhuh: Ah, that makes sense.

nojnhuh (approved):
/lgtm

test/e2e/azure_machinepool_drain.go
CecileRobertMichon: GitHub wasn't letting me post a comment earlier... I think we should use a different "test machine pool drain" finalizer, because otherwise the test and the controller are racing to remove the same finalizer no?

willie-yao: @CecileRobertMichon Good catch! Just added a different finalizer specific for the test: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3503/files#diff-f20fb04862d5b7d8e948f7d513ecb40ec36bf15a4483835d6d7af5ee338c1a69R32

exp/api/v1beta1/azuremachinepoolmachine_types.go
```diff
@@ -27,6 +27,9 @@ import (
 const (
 	// AzureMachinePoolMachineFinalizer is used to ensure deletion of dependencies (nodes, infra).
 	AzureMachinePoolMachineFinalizer = "azuremachinepoolmachine.infrastructure.cluster.x-k8s.io"
+
+	// AzureMachinePoolMachineTestFinalizer is used to ensure deletion of dependencies during tests.
+	AzureMachinePoolMachineTestFinalizer = "azuremachinepoolmachine.infrastructure.cluster.x-k8s.io/test"
```
nojnhuh, line 32: Maybe it's worth defining this closer to the test itself? Could probably even be a local var right in the test.

test/e2e/azure_machinepool_drain.go
```diff
@@ -157,13 +166,39 @@ func testMachinePoolCordonAndDrain(ctx context.Context, mgmtClusterProxy, worklo
...
+	By(fmt.Sprintf("checking for a machine to start draining for machine pool: %s/%s", amp.Namespace, amp.Name))
+	Eventually(func() error {
+		ampmls, err := getAzureMachinePoolMachines(ctx, mgmtClusterProxy, workloadClusterProxy, amp)
+		if err != nil {
+			LogWarning(errors.Wrap(err, "failed to list the azure machine pool machines").Error())
+			return err
+		}
+
+		for i := range ampmls {
+			if conditions.IsFalse(&ampmls[i], clusterv1.DrainingSucceededCondition) {
```
CecileRobertMichon, line 183: why are we checking if the condition is False here?

willie-yao: I adapted this line from the previous iteration of the cordon and drain test, which checked if the condition is false.. However it does make a lot more sense to check if it is true. The blame here mentions the original version of this test checks to see if the condition is false: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blame/9629164862fc682181e579431d28a84186d6ceaa/test/e2e/azure_machinepool_drain.go#L179

However, in the original commit, the diff shows the condition being checked if it is true... I'm wondering what happened here and whether there was a reason it was changed to `isFalse` 🤔 
https://github.com/kubernetes-sigs/cluster-api-provider-azure/commit/d1c01511604f40360cfa7db9ffc61dabec3e702a#diff-ef1035bcafba9959d0df4cb4d9544deafccfb9d9a9f7e89cbcf1cb8565a89880R189

willie-yao: None of the history for this file showed that the `isTrue` was changed to `isFalse` prior to the code being removed. Is this even possible? https://github.com/jackfrancis/cluster-api-provider-azure/commits/c2781f00104ad84b22934c72cf56e22c50beabf0/test/e2e/azure_machinepool_drain.go

For now, I'll go ahead and set it to `isTrue` as it would make more sense to check if the condition is true.

test/e2e/azure_machinepool_drain.go
```diff
@@ -157,13 +166,39 @@ func testMachinePoolCordonAndDrain(ctx context.Context, mgmtClusterProxy, worklo
...
-	// https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/2159
+	By(fmt.Sprintf("checking for a machine to start draining for machine pool: %s/%s", amp.Namespace, amp.Name))
+	Eventually(func() error {
+		ampmls, err := getAzureMachinePoolMachines(ctx, mgmtClusterProxy, workloadClusterProxy, amp)
+		if err != nil {
+			LogWarning(errors.Wrap(err, "failed to list the azure machine pool machines").Error())
+			return err
+		}
+
+		for i := range ampmls {
```
jackfrancis, line 182: can we use the human-friendly var instead of the index int?

`for _, ampm := range ampls {`...

Also, I think we should be able to simply check for the `IsTrue` condition below, as it is inclusive of `conditions.Has`. So just:

`if conditions.IsTrue(&ampm, clusterv1.DrainingSucceededCondition) {`...

should work

test/e2e/azure_machinepool_drain.go
```diff
@@ -157,13 +166,39 @@ func testMachinePoolCordonAndDrain(ctx context.Context, mgmtClusterProxy, worklo
...
+			if conditions.Has(&ampmls[i], clusterv1.DrainingSucceededCondition) && conditions.IsTrue(&ampmls[i], clusterv1.DrainingSucceededCondition) {
+				controllerutil.RemoveFinalizer(&ampmls[i], azureMachinePoolMachineTestFinalizer)
+				return nil // started draining the node prior to delete
+			}
+		}
+
+		return errors.New("no machine has started to drain")
+	}, 3*time.Minute, 3*time.Second).Should(Succeed())
```
jackfrancis, line 190: I wonder if 3 minutes isn't long enough for the 99.9 percentile outcome of a successful k8s drain operation

test/e2e/azure_machinepool_drain.go
willie-yao: > can we use the human-friendly var instead of the index int?
> 
> `for _, ampm := range ampls {`...

Unfortunately, the linter complains about referencing a temporarily for loop variable if I do this.

test/e2e/azure_machinepool_drain.go
willie-yao: So far, I haven't run into a single failure yet. Do drain operations typically take a longer time? If so it could be safer to increase this to maybe 5 minutes.

jackfrancis: I think 10 mins is probably an appropriate timeout val

test/e2e/azure_machinepool_drain.go
```diff
@@ -157,13 +166,39 @@ func testMachinePoolCordonAndDrain(ctx context.Context, mgmtClusterProxy, worklo
...
+	By(fmt.Sprintf("waiting for the machine pool to scale down to %d replicas: %s/%s", decreasedReplicas, amp.Namespace, amp.Name))
+	Eventually(func() (int32, error) {
+		mp, err := getOwnerMachinePool(ctx, mgmtClusterProxy.GetClient(), amp.ObjectMeta)
+		if err != nil {
+			LogWarning(err.Error())
+			return 0, err
+		}
+
+		return int32(len(mp.Spec.ProviderIDList)), nil
+	}, 3*time.Minute, 3*time.Second).Should(Equal(decreasedReplicas))
```
jackfrancis, line 201: Prob 10 mins is good here as well

test/e2e/azure_machinepool_drain.go
jackfrancis: They are workload-dependent. If we're confident that the workloads we're running in our test environment can reliably shut down in a few seconds then we should be O.K.

test/e2e/azure_machinepool_drain.go
jackfrancis: This one I think we want to bump no matter what, as this can be Azure operations-dependent (deleting a VM)

test/e2e/azure_machinepool_drain.go
jackfrancis: I've learned to just do what the linter says

jackfrancis (approved):
/lgtm
/approve

PR #3509, opened 2023-05-02
Add AzureEnvironment to AzureManagedControlPlane

### LINKED ISSUE #3489: AzureEnvironment Support for Managed Clusters
***Goals***
1. Add a new string to managed cluster spec identical to this in [azureclusterspec](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/e1e1a1270b663a9b256f7d41783603f529454549/api/v1alpha4/azurecluster_types.go#L60-L67)
2. Pass through the AzureEnvironment for the managed cluster to the scope creation for example both [here](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/scope/managedcontrolplane.go#L76) and [here](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/scope/managedcontrolplane.go#L85). This is identical behavior to the self-managed cluster.

**User Story**
As a user I would like to create managed AKS clusters in Germany/China/USGovernment cloud environments.

**Detailed Description**
It looks like support for AzureEnvironment was done in self-managed clusters and hardcoded to the empty string defaults in the AKS code paths. The code changes would allow access to all the [environments](https://github.com/Azure/go-autorest/blob/main/autorest/azure/environments.go#L34-L41) offered by Azure.

[Describe contract changes between Cluster API Azure Provider controllers, if applicable.]

Will probably add an AzureEnvironment to azuremanagedcluster identical to the one in azurecluster

/kind proposal

luthermonson:
**What type of PR is this?**
/kind feature

Adds support for AzureManagedControlPlane.AzureEnvironment allowing a user to set a string value which translates into the configuration passed into the Azure Client. This functionality already exists in AzureCluster and this PR allows for Managed clusters to take advantage of the same feature.

Fixes #3489 

There was an [override](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1850) done to bypass some vmss caching issues for tests. This setup a region URL which doesn't work in govcloud and I refactored `newAzureManagedMachinePoolService` to accomodate additional ways to skip using regional authorizers based on config.

**Release note**:

```release-note
Add AzureEnvironment support to AzureManagedControlPlane
```

### DIFF
--- a/api/v1beta1/azuremanagedcontrolplane_types.go
+++ b/api/v1beta1/azuremanagedcontrolplane_types.go
@@ -138,6 +138,14 @@ type AzureManagedControlPlaneSpec struct {
 	// AutoscalerProfile is the parameters to be applied to the cluster-autoscaler when enabled
 	// +optional
 	AutoScalerProfile *AutoScalerProfile `json:"autoscalerProfile,omitempty"`
+
+	// AzureEnvironment is the name of the AzureCloud to be used.
+	// The default value that would be used by most users is "AzurePublicCloud", other values are:
+	// - ChinaCloud: "AzureChinaCloud"
+	// - PublicCloud: "AzurePublicCloud"
+	// - USGovernmentCloud: "AzureUSGovernmentCloud"
+	// +optional
+	AzureEnvironment string `json:"azureEnvironment,omitempty"`
 }
 
 // AADProfile - AAD integration managed by AKS.
--- a/api/v1beta1/azuremanagedcontrolplane_webhook.go
+++ b/api/v1beta1/azuremanagedcontrolplane_webhook.go
@@ -193,6 +193,13 @@ func (mw *azureManagedControlPlaneWebhook) ValidateUpdate(ctx context.Context, o
 		allErrs = append(allErrs, err)
 	}
 
+	if err := webhookutils.ValidateImmutable(
+		field.NewPath("Spec", "AzureEnvironment"),
+		old.Spec.AzureEnvironment,
+		m.Spec.AzureEnvironment); err != nil {
+		allErrs = append(allErrs, err)
+	}
+
 	if old.Spec.AADProfile != nil {
 		if m.Spec.AADProfile == nil {
 			allErrs = append(allErrs,
--- a/controllers/azuremanagedmachinepool_reconciler.go
+++ b/controllers/azuremanagedmachinepool_reconciler.go
@@ -74,20 +74,26 @@ func (a *AgentPoolVMSSNotFoundError) Is(target error) bool {
 
 // newAzureManagedMachinePoolService populates all the services based on input scope.
 func newAzureManagedMachinePoolService(scope *scope.ManagedMachinePoolScope) (*azureManagedMachinePoolService, error) {
-	var authorizer azure.Authorizer = scope
-	if scope.Location() != "" {
-		regionalAuthorizer, err := azure.WithRegionalBaseURI(scope, scope.Location())
-		if err != nil {
-			return nil, errors.Wrap(err, "failed to create a regional authorizer")
-		}
-		authorizer = regionalAuthorizer
+	scaleSetAuthorizer, err := scaleSetAuthorizer(scope)
+	if err != nil {
+		return nil, err
 	}
 
 	return &azureManagedMachinePoolService{
 		scope:         scope,
 		agentPoolsSvc: agentpools.New(scope),
-		scaleSetsSvc:  scalesets.NewClient(authorizer),
+		scaleSetsSvc:  scalesets.NewClient(scaleSetAuthorizer),
 	}, nil
+}
+
+// scaleSetAuthorizer takes a scope and determines if a regional authorizer is needed for scale sets
+// see https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1850 for context on region based authorizer.
+func scaleSetAuthorizer(scope *scope.ManagedMachinePoolScope) (azure.Authorizer, error) {
+	if scope.ControlPlane.Spec.AzureEnvironment == azure.PublicCloudName {
+		return azure.WithRegionalBaseURI(scope, scope.Location()) // public cloud supports regional end points
+	}
+
+	return scope, nil
 }
 
 // Reconcile reconciles all the services in a predetermined order.
--- a/api/v1beta1/types_class.go
+++ b/api/v1beta1/types_class.go
@@ -16,7 +16,9 @@ limitations under the License.
 
 package v1beta1
 
-import corev1 "k8s.io/api/core/v1"
+import (
+	corev1 "k8s.io/api/core/v1"
+)
 
 // AzureClusterClassSpec defines the AzureCluster properties that may be shared across several Azure clusters.
 type AzureClusterClassSpec struct {
--- a/azure/scope/managedcontrolplane.go
+++ b/azure/scope/managedcontrolplane.go
@@ -73,7 +73,7 @@ func NewManagedControlPlaneScope(ctx context.Context, params ManagedControlPlane
 	}
 
 	if params.ControlPlane.Spec.IdentityRef == nil {
-		if err := params.AzureClients.setCredentials(params.ControlPlane.Spec.SubscriptionID, ""); err != nil {
+		if err := params.AzureClients.setCredentials(params.ControlPlane.Spec.SubscriptionID, params.ControlPlane.Spec.AzureEnvironment); err != nil {
 			return nil, errors.Wrap(err, "failed to create Azure session")
 		}
 	} else {
@@ -82,7 +82,7 @@ func NewManagedControlPlaneScope(ctx context.Context, params ManagedControlPlane
 			return nil, errors.Wrap(err, "failed to init credentials provider")
 		}
 
-		if err := params.AzureClients.setCredentialsWithProvider(ctx, params.ControlPlane.Spec.SubscriptionID, "", credentialsProvider); err != nil {
+		if err := params.AzureClients.setCredentialsWithProvider(ctx, params.ControlPlane.Spec.SubscriptionID, params.ControlPlane.Spec.AzureEnvironment, credentialsProvider); err != nil {
 			return nil, errors.Wrap(err, "failed to configure azure settings and credentials for Identity")
 		}
 	}
--- a/config/crd/bases/infrastructure.cluster.x-k8s.io_azuremanagedcontrolplanes.yaml
+++ b/config/crd/bases/infrastructure.cluster.x-k8s.io_azuremanagedcontrolplanes.yaml
@@ -217,6 +217,12 @@ spec:
                     - "false"
                     type: string
                 type: object
+              azureEnvironment:
+                description: 'AzureEnvironment is the name of the AzureCloud to be
+                  used. The default value that would be used by most users is "AzurePublicCloud",
+                  other values are: - ChinaCloud: "AzureChinaCloud" - PublicCloud:
+                  "AzurePublicCloud" - USGovernmentCloud: "AzureUSGovernmentCloud"'
+                type: string
               controlPlaneEndpoint:
                 description: ControlPlaneEndpoint represents the endpoint used to
                   communicate with the control plane.


luthermonson:
@devigned for eyes on the rewrite of his previous PR functionality

devigned:
@mboersma this is the PR we chatted about earlier today.

luthermonson:
i just forgot about the conversions, this is probably done from my perspective when those get checked in

controllers/azuremanagedmachinepool_reconciler.go
```diff
@@ -74,22 +75,32 @@ func (a *AgentPoolVMSSNotFoundError) Is(target error) bool {
...
+func scaleSetAuthorizer(scope *scope.ManagedMachinePoolScope) (azure.Authorizer, error) {
+	if scope.Location() == "" {
+		return scope, nil // no location so use default
+	}
+
+	if scope.ControlPlane.Spec.AzureEnvironment == azureautorest.USGovernmentCloud.Name {
+		return scope, nil // no region support in usgovcloud
+	}
+
+	return azure.WithRegionalBaseURI(scope, scope.Location())
```
nojnhuh, line 101: I noticed `spec.azureEnvironment` for AzureCluster gets defaulted to AzurePublicCloud. It looks like this gracefully handles when it's not defined though. @CecileRobertMichon do you know if that defaulting would still be necessary for AzureManagedControlPlane?

controllers/azuremanagedmachinepool_reconciler.go
```diff
@@ -74,22 +75,32 @@ func (a *AgentPoolVMSSNotFoundError) Is(target error) bool {
...
 }
 
+// scaleSetAuthorizer takes a scope and determines if a regional authorizer is needed for scale sets
+// see https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1850 for context on region based authorizer.
+func scaleSetAuthorizer(scope *scope.ManagedMachinePoolScope) (azure.Authorizer, error) {
+	if scope.Location() == "" {
+		return scope, nil // no location so use default
+	}
+
+	if scope.ControlPlane.Spec.AzureEnvironment == azureautorest.USGovernmentCloud.Name {
```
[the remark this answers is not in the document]
luthermonson: unfortunately, i dont have access to any of those clouds. if there is anyone internal at msft who could confirm I would gladly add additional code to skip more clouds if we need too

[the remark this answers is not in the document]
luthermonson: yup, i was trying to create an AKS cluster in usgovarizona and it was failing DNS trying to contact the following:
```
https://usgovarizona.management.usgovcloudapi.net/subscriptions/5129b42f-3fc6-48e2-b674-eebe99ca6121/resourceGroups/SPC_MC_rnl-rg_luther-test_usgovarizona/providers/Microsoft.Compute/virtualMachineScaleSets?api-version=2021-11-01
```

and DNS fails for all the regions I checked for <region>.management.usgovcloudapi.net and after adding this code I retested and an AKS cluster was created successfully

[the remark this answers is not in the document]
luthermonson: i dont know if it hit the status. the dns errors were from capz logs, i didn't check the status tbh

luthermonson:
@nojnhuh any more thoughts on this?

controllers/azuremanagedmachinepool_reconciler.go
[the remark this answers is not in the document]
nojnhuh: I think if an error show up in at least one of those places that's reasonably covered.

api/v1beta1/azuremanagedcontrolplane_types.go
```diff
@@ -138,6 +138,15 @@ type AzureManagedControlPlaneSpec struct {
...
 	AutoScalerProfile *AutoScalerProfile `json:"autoscalerProfile,omitempty"`
+
+	// AzureEnvironment is the name of the AzureCloud to be used.
+	// The default value that would be used by most users is "AzurePublicCloud", other values are:
+	// - ChinaCloud: "AzureChinaCloud"
+	// - GermanCloud: "AzureGermanCloud"
+	// - PublicCloud: "AzurePublicCloud"
+	// - USGovernmentCloud: "AzureUSGovernmentCloud"
+	// +optional
+	AzureEnvironment string `json:"azureEnvironment,omitempty"`
```
nawazkh, line 149: Since we are expecting fixed number of Cloud options here, would it be worth adding a validation in the validating webhook for this field ?

nawazkh: Do we want users to be able to change this field once set? Or is this field to be set immutable? 
@CecileRobertMichon @nojnhuh We don't seem to be having a validation check for https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/2bffb4d6db3b04c517ae9f172a96e123f2165f4d/api/v1beta1/types_class.go#L48 in the AzureClusterClass webhooks. I am not sure if we want to really validate user entry for `spec.AzureEnvironment`?

nawazkh: I might have missed it, but since we are explicitly choosing a default value for azureManagedCluster (will look up where the default is set in the code), should we have a default function for AzureManagedCluster type in the defaulting webhook for `AzureManagedClusterSpec.AzureEnvironment` field?

nawazkh (commented):
Thanks for putting this together! 
I shared some questions around defaulting and validations.

nawazkh (commented):
Thanks for putting this together! 
I shared some questions around defaulting and validations.

api/v1beta1/azuremanagedcontrolplane_types.go
luthermonson: RE so i did in fact miss that the AzureCluster has a webhook to make this field immutable, I will add one but that work depends on the answer to the second half so I'll await a response if we should move the field...

RE AzureManagedCluster vs AzureManagedControlPlane... unfortunately, when making azuremanagedcontrolplane scope we never pull the AzureManagedCluster, while I agree that sort of makes spiritually the same as AzureCluster but the place we make clients only has the capi cluster. to accomplish this I'll need to pull and store the AzureManagedCluster in the AzureManagedControlPlane reconcile loop to use in the client creation. This is doable but a bit more work for the reconcile loop. In defense of not moving it, the AzureManagedCluster reconcile loop is pretty lame and all the real work is done in the AzureManagedControlPlane reconcile and even the [docs](https://capz.sigs.k8s.io/topics/managedcluster.html#specification) show the AzureManagedCluster as nothing more than a name and location/resourcegroup/subscriptionid/etc/etc/etc is in the Control Plane, it doesn't make sense to have the environment in the cluster and everything else in the control plane.

RE defaults... it was previously passing empty string to this same func which makes the azure client default to the public cloud. Since the new field is just a string the fallback defaults will be an empty string which is the same as before the change.

tl;dr let's keep azure env on the control plane and make the field immutable via a webhook and the default empty string should be fine

nojnhuh: +1 to keeping all configuration in AzureManagedControlPlane vs. AzureManagedCluster.

I think the argument in favor of defaulting a field is usually to protect against when the default changes in a future release to prevent those values from changing underneath the user. I think in this case that's less applicable since I imagine the default would never change from AzurePublicCloud. So I think it's acceptable to skip defaulting this here as long as the slight inconsistency with AzureCluster isn't an issue.

CecileRobertMichon: > tl;dr let's keep azure env on the control plane and make the field immutable via a webhook and the default empty string should be fine

+1 to keeping config in `AzureManagedControlPlane`, especially since `AzureManagedCluster` may get removed in the future with https://github.com/kubernetes-sigs/cluster-api/pull/8500

+1 to making immutable

For defaulting, I do think it'd be slightly better to default so the value is persisted in the spec and can be consumed by the user now that there's more than one possible value, even if Jon makes a good point that defaulting is less critical in this case. However, I don't see a reason not to default in the AzureManagedControlPlane webhook unless I'm missing something?

luthermonson: I have no problem making the default "AzurePublicCloud" but maybe do it like this https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/0ec7aa4f03f0437bf279141b8cbe2f77d9491fef/api/v1beta1/azurecluster_default.go#L70-L74 ?

luthermonson: never mind, i answered my own question.. that was for conversions and stuffs

luthermonson: added the webhook validation and switched the const to an already existing const the azure pkg

luthermonson: lol nevermind, moving the const back

controllers/azuremanagedmachinepool_reconciler.go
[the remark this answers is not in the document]
CecileRobertMichon: In this case I think it would be better to exclude all clouds by default and only opt-in when we know for sure that regional URIs are supported. The case where "we tried to use a regional base URI but it wasn't supported" is worse (total failure) than "we didn't use a regional base URI when we could have" (de-optimization which may cause transient failures)

```suggestion
	if scope.ControlPlane.Spec.AzureEnvironment != azure.PublicCloudName {
		return scope, nil // no region support in sovereign clouds
	}
```

Note that I intentionally used the AzurePublicCloudName const in https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/defaults.go#L34 to avoid adding anoter usage of `azureautorest` since it's deprecated.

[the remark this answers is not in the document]
luthermonson: @CecileRobertMichon so something more like this?

```
func scaleSetAuthorizer(scope *scope.ManagedMachinePoolScope) (azure.Authorizer, error) {
	if scope.Location() == "" {
		return scope, nil // no location so use default
	}

	if scope.ControlPlane.Spec.AzureEnvironment == azure.PublicCloudName {
		return azure.WithRegionalBaseURI(scope, scope.Location()) // no region support in usgovcloud
	}

	return scope, nil
}
```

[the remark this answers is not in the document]
luthermonson: i dont know why we do the location check in this method, so probably drop that too i guess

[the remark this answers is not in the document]
luthermonson: maybe this is enough...
```
func scaleSetAuthorizer(scope *scope.ManagedMachinePoolScope) (azure.Authorizer, error) {
	if scope.ControlPlane.Spec.AzureEnvironment == azure.PublicCloudName {
		return azure.WithRegionalBaseURI(scope, scope.Location()) // public cloud supports regional end points
	}

	return scope, nil
}
```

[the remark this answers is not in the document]
luthermonson: committed

api/v1beta1/azuremanagedcontrolplane_webhook.go
```diff
@@ -193,6 +193,18 @@ func (mw *azureManagedControlPlaneWebhook) ValidateUpdate(ctx context.Context, o
 		allErrs = append(allErrs, err)
 	}
 
+	if !reflect.DeepEqual(m.Spec.AzureEnvironment, old.Spec.AzureEnvironment) {
+		old.setAzureEnvironmentDefault()
```
CecileRobertMichon, line 197: since validation webhooks run after defaulting webhooks, you shouldn't need to do this here

Instead, you can use the same pattern that's used everywhere else in this file to validate immutability with `webhookutils.ValidateImmutable()` (see lines above for examples)

api/v1beta1/azuremanagedcontrolplane_types.go
```diff
@@ -138,6 +138,15 @@ type AzureManagedControlPlaneSpec struct {
 	// AutoscalerProfile is the parameters to be applied to the cluster-autoscaler when enabled
 	// +optional
 	AutoScalerProfile *AutoScalerProfile `json:"autoscalerProfile,omitempty"`
+
+	// AzureEnvironment is the name of the AzureCloud to be used.
+	// The default value that would be used by most users is "AzurePublicCloud", other values are:
+	// - ChinaCloud: "AzureChinaCloud"
+	// - GermanCloud: "AzureGermanCloud"
```
CecileRobertMichon, line 145: nit: GermanCloud is deprecated since 2018 and I don't think AKS was ever supported there so we should probably remove it from this list

https://learn.microsoft.com/en-us/azure/germany/

CecileRobertMichon (commented):
last few minor comments but otherwise lgtm! Thank you @luthermonson 

/assing @nojnhuh @nawazkh

nawazkh:
Looks good to me pending the nits mentioned by Cecile!
Thank you @luthermonson for putting it together! 🌱

Jont828 (commented):
Aisde from Cecile's comments, LGTM.

CecileRobertMichon (commented):
/lgtm
/approve

PR #3517, opened 2023-05-04
use default backend pool name in capz controller if empty

### LINKED ISSUE #3252: Issues when upgrading Controller to a version that include the `backendPool.name` feature
/kind bug

When upgrading controller to a version that include [backendPool.name](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2714) there is an issue with `AzureCluster` resources that were created before and did not include the required field

* there is a webhook that will add the `required fields backendPool.name` when the resource is Created / Updated 
* There is conversion code from old version of the api that will also handle this 
* when upgrading the `controller` any exisiting `Azurecluster` will not be updated by the webhook ( since no event happened ) and the requests to Azure will fail with 
```
message: 'loadbalancers failed to create or update. err: failed to update resource
      glippy/glippy-public-lb (service: loadbalancers): network.LoadBalancersClient#CreateOrUpdate:
      Failure sending request: StatusCode=400 -- Original Error: Code="InvalidResourceName"
      Message="Resource name  is invalid. The name can be up to 80 characters long.
      It must begin with a word character, and it must end with a word character or
      with ''_''. The name may contain word characters or ''.'', ''-'', ''_''." Details=[]'
```
* this is because the request _must contain_ a name for the backend pool that is actually `""` 

Adding the fields to the AzureClusters CR fixes the issue and everything is clean 

Upgrade would not require to modify the CRs , ideally the code should be backward compatible 

Discussion in slack https://kubernetes.slack.com/archives/CEX9HENG7/p1678370754712279 

**Environment:**

- cluster-api-provider-azure version: upgraded from `1.6.1` to `1.7.0` ( and 1.7.2 and 1.8.0 ) 
- Kubernetes version: (use `kubectl version`): 1.24.11

sonasingh46:
fixes #3252 

 

 
**What type of PR is this?**

- [x] cherry-pick candidate

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Use default backend pool name in capz controller.
```

### DIFF
--- a/api/v1beta1/azurecluster_default.go
+++ b/api/v1beta1/azurecluster_default.go
@@ -204,10 +204,7 @@ func (c *AzureCluster) setAPIServerLBDefaults() {
 			}
 		}
 	}
-
-	if lb.BackendPool.Name == "" {
-		lb.BackendPool.Name = generateBackendAddressPoolName(lb.Name)
-	}
+	c.SetAPIServerLBBackendPoolNameDefault()
 }
 
 // SetNodeOutboundLBDefaults sets the default values for the NodeOutboundLB.
@@ -247,10 +244,7 @@ func (c *AzureCluster) SetNodeOutboundLBDefaults() {
 	}
 
 	c.setOutboundLBFrontendIPs(lb, generateNodeOutboundIPName)
-
-	if lb.BackendPool.Name == "" {
-		lb.BackendPool.Name = generateOutboundBackendAddressPoolName(lb.Name)
-	}
+	c.SetNodeOutboundLBBackendPoolNameDefault()
 }
 
 // SetControlPlaneOutboundLBDefaults sets the default values for the control plane's outbound LB.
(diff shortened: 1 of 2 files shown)


azure/scope/cluster.go
```diff
@@ -258,18 +266,26 @@ func (s *ClusterScope) LBSpecs() []azure.ResourceSpecGetter {
...
+			ClusterName:       s.ClusterName(),
+			Location:          s.Location(),
+			ExtendedLocation:  s.ExtendedLocation(),
+			VNetName:          s.Vnet().Name,
+			VNetResourceGroup: s.Vnet().ResourceGroup,
+			FrontendIPConfigs: s.NodeOutboundLB().FrontendIPs,
+			Type:              s.NodeOutboundLB().Type,
+			SKU:               s.NodeOutboundLB().SKU,
+			// @sonasingh46: Ref issue #3253
+			// Change the code to be using only s.NodeOutboundLB().BackendPool.Name in the upcoming
```
CecileRobertMichon, line 281: I don't think we can do that as there is no guarantee that users are upgrading one CAPZ release at a time (and we don't enforce it). What do you think about moving this actually persist the defaulting in the spec, similar to `SetDNSName` https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/fce6cac52c5081ba642c08554e1f4596eae6228b/azure/scope/cluster.go#L945

azure/scope/cluster.go
```diff
@@ -230,6 +230,40 @@ func (s *ClusterScope) PublicIPSpecs() []azure.ResourceSpecGetter {
 	return publicIPSpecs
 }
 
+// SetAPIServerLBBackendPoolName sets the name of the backend pool for apiserver LB.
+func (s *ClusterScope) SetAPIServerLBBackendPoolName() {
```
CecileRobertMichon, line 234: this looks good, I think my only comment/question is whether it would be possible to move these functions to the default webhook file and use them in both the webhook and in `SetLBBackendPoolName()` below so we're sharing the default logic with the webhook in case anything changes later.

api/v1beta1/azurecluster_default.go
```diff
@@ -59,6 +59,7 @@ func (c *AzureCluster) setNetworkSpecDefaults() {
 	c.setAPIServerLBDefaults()
 	c.SetNodeOutboundLBDefaults()
 	c.SetControlPlaneOutboundLBDefaults()
+	c.SetBackendPoolNameDefault()
```
CecileRobertMichon, line 62: the unit tests for the LB default functions above are now failing because the defaulting is happening outside of the functions now... You can either 1) change the unit tests to not expect backend pool names after setting LB defaults (but then we lose unit test coverage, so you'd have to add a separate unit test), or 2) leave the `SetBackendPoolNameDefault` function in cluster.go and call `SetAPIServerLBBackendPoolNameDefault` from inside `setAPIServerLBDefaults` (and same with the other two). That way defaulting webhook code doesn't change  (it's an equivalent refactor, we're just extracting functions), and we can just use `SetBackendPoolNameDefault` in the reconciler.

willie-yao (approved):
/lgtm

PR #3519, opened 2023-05-04
allow unsetting the deprecated spec.AcceleratedNetworking field

### LINKED ISSUE #3518: Upgrade from version prior to v1.7.0 fails if accelerated networking is enabled.
/kind bug

- Create a workload cluster with accelerate networking enabled on versions prior to v1.7.0 e.g. v1.6.3
- Upgraded to version v1.7.0 or later e.g. v1.9.0.
- The upgrade will fail with below error:
```
E0504 12:44:39.534821       1 controller.go:329]  "msg"="Reconciler error" "error"="admission webhook \"validation.azuremachine.infrastructure.cluster.x-k8s.io\" denied the request: AzureMachine.infrastructure.cluster.x-k8s.io \"wl-antrea-1-control-plane-ntwzw-p9qv6\" is invalid: Spec.AcceleratedNetworking: Invalid value: \"null\": field is immutable, unable to set an empty value if it was already set" "AzureMachine"={"name":"wl-antrea-1-control-plane-ntwzw-p9qv6","namespace":"default"} "controller"="azuremachine" "controllerGroup"="infrastructure.cluster.x-k8s.io" "controllerKind"="AzureMachine" "name"="wl-antrea-1-control-plane-ntwzw-p9qv6" "namespace"="default" "reconcileID"="1ccf6fc6-4d2d-4eb0-adb8-068c8f9206d7"
```

The upgrade should pass successfully. 

The fix will be to remove the lines of code as put in the following link.
https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/api/v1beta1/azuremachine_webhook.go#L157-L162

**Environment:**

sonasingh46:
**What type of PR is this?**
Allow unsetting deprecated field `spec.AcceleratedNetworking` on azuremachine

Fixes #3518 

- [x] cherry-pick candidate

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
make deprecated field `acceleratedNetowrking` on azuremachine mutable
```

### DIFF
--- a/api/v1beta1/azuremachine_webhook.go
+++ b/api/v1beta1/azuremachine_webhook.go
@@ -154,7 +154,10 @@ func (mw *azureMachineWebhook) ValidateUpdate(ctx context.Context, oldObj, newOb
 		allErrs = append(allErrs, err)
 	}
 
-	if err := webhookutils.ValidateImmutable(
+	// Spec.AcceleratedNetworking can only be reset to nil and no other changes apart from that
+	// is accepted if the field is set.
+	// Ref issue #3518
+	if err := webhookutils.ValidateZeroTransition(
 		field.NewPath("Spec", "AcceleratedNetworking"),
 		old.Spec.AcceleratedNetworking,
 		m.Spec.AcceleratedNetworking); err != nil {
--- a/util/webhook/validator.go
+++ b/util/webhook/validator.go
@@ -52,6 +52,16 @@ func ValidateImmutable(path *field.Path, oldVal, newVal any) *field.Error {
 	return nil
 }
 
+// ValidateZeroTransition validates equality across two values, with only exception to allow
+// the value to transition of a zero value.
+func ValidateZeroTransition(path *field.Path, oldVal, newVal any) *field.Error {
+	if reflect.ValueOf(newVal).IsZero() {
+		// unsetting the field is allowed
+		return nil
+	}
+	return ValidateImmutable(path, oldVal, newVal)
+}
+
 // EnsureStringSlicesAreEquivalent returns if two string slices have equal lengths,
 // and that they have the exact same items; it does not enforce strict ordering of items.
 func EnsureStringSlicesAreEquivalent(a []string, b []string) bool {
--- a/api/v1beta1/azuremachine_webhook_test.go
+++ b/api/v1beta1/azuremachine_webhook_test.go
@@ -537,6 +537,34 @@ func TestAzureMachine_ValidateUpdate(t *testing.T) {
 			wantErr: false,
 		},
 		{
+			name: "validTest: azuremachine.spec.AcceleratedNetworking transition(from true) to nil is acceptable",
+			oldMachine: &AzureMachine{
+				Spec: AzureMachineSpec{
+					AcceleratedNetworking: pointer.Bool(true),
+				},
+			},
+			newMachine: &AzureMachine{
+				Spec: AzureMachineSpec{
+					AcceleratedNetworking: nil,
+				},
+			},
+			wantErr: false,
+		},
+		{
+			name: "validTest: azuremachine.spec.AcceleratedNetworking transition(from false) to nil is acceptable",
+			oldMachine: &AzureMachine{
+				Spec: AzureMachineSpec{
+					AcceleratedNetworking: pointer.Bool(false),
+				},
+			},
+			newMachine: &AzureMachine{
+				Spec: AzureMachineSpec{
+					AcceleratedNetworking: nil,
+				},
+			},
+			wantErr: false,
+		},
+		{
 			name: "invalidTest: azuremachine.spec.SpotVMOptions is immutable",
 			oldMachine: &AzureMachine{
 				Spec: AzureMachineSpec{
--- a/util/webhook/validator_test.go
+++ b/util/webhook/validator_test.go
@@ -299,3 +299,215 @@ func TestEnsureStringSlicesAreEquivalent(t *testing.T) {
 		})
 	}
 }
+
+func TestValidateZeroTransitionPtr(t *testing.T) {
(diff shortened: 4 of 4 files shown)


api/v1beta1/azuremachine_webhook.go
```diff
@@ -154,13 +154,6 @@ func (mw *azureMachineWebhook) ValidateUpdate(ctx context.Context, oldObj, newOb
 		allErrs = append(allErrs, err)
 	}
 
-	if err := webhookutils.ValidateImmutable(
```
CecileRobertMichon, line 157: should the new `AcceleratedNetworking` field be immutable? when it's not nil

nawazkh: I think this is a fair change. We are removing an immutability-validation on deprecated field. I think we should have caught this and removed validation on `m.Spec.AcceleratedNetworking` while merging multiple network interfaces PR.

CecileRobertMichon (commented):
/lgtm

nawazkh:
/retitle remove immutability check on the deprecated spec.AcceleratedNetworking field

CecileRobertMichon (commented):
/lgtm cancel

waiting for update

util/webhook/validator.go
```diff
@@ -52,6 +52,29 @@ func ValidateImmutable(path *field.Path, oldVal, newVal any) *field.Error {
 	return nil
 }
 
+// ValidateZeroTransition validates equality across two values, with only exception to allow
+// the value to transition of a zero value.
+func ValidateZeroTransition(path *field.Path, oldVal, newVal any) *field.Error {
```
jackfrancis, line 57: Rather than add a new func here that is so similar to `ValidateImmutable`, can we add a "tolerate reset to zero value" bool parameter to that existing function (and then update usage everywhere we call `ValidateImmutable`). Something like:

`func ValidateImmutable(path *field.Path, oldVal, newVal any, tolerateResetToZeroVal bool) *field.Error {`

Then we can add new UT cases for thes newly allowed outcomes to the set of UT in `validator_test.go`.

If we want to keep `ValidateImmutable` as-is and add a new, similar function like it, we do want to have UT coverage, and because these types of functions operate against any type (literally the `any` type) that means we're going to want to have UT across all concrete types that CAPZ is using this function for in practice. So a new func means lots of new UT cases.

Thoughts?

CecileRobertMichon: IMO that is a bit overkill for something that is deprecated (we would have to modify every single call of `ValidateImmutable` to add the new parameter, making this critical bug fix PR into a L/XL PR that would be more difficult to cherry-pick. Since we have to cherry-pick this quite far (the bug was introduced in v1.7), for practical reasons we should try to keep this code change as small and as targeted as possible.

That being said, I do agree with @jackfrancis adding a whole new function that isn't unit tested and has a lot of overlap with the validate immutable function is not ideal. How about this compromise: wrap the `ValidateImmutable` in this new function such that all it does is return early if the field is just being set to nil, else return `ValidateImmutable` ?

jackfrancis: Good compromise!

I'd still like to see the UT cases for the new func however, but this suggestion does guarantee that webhooks that need this back-compat tolerance will for the non-back-compat case (validate that the property didn't change) traverse the same codepaths as the rest of the webhook validation code.

nawazkh: In a scenario where we are going to deprecate this field (`spec.AcceleratedNetworking`), won't it just be better in removing its validation?  Provided we have validation (immutability and default check) for the newly created field?

sonasingh46: @CecileRobertMichon -- I did not get the wrap part. Can you please elaborate?

willie-yao: I agree with Nawaz. If `spec.AcceleratedNetworking` is set, the defaulting webhook should set the value to nil and transfer it over the `spec.NetworkInterfaces.AcceleratedNetworking`, as done here: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/api/v1beta1/azuremachine_default.go#L158

willie-yao: We did something similar when deprecating `roleAssignmentName` and moving it to `systemAssignedIdentityRole.Name`, and we omitted the deprecated field from the AzureMachine webhook now that it's a deprecated field. We only validate the new field once the defaulting webhook sets the value of the deprecated field to the new one.

https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/api/v1beta1/azuremachine_webhook.go#L64

CecileRobertMichon: I was thinking something like:

```
func ValidateZeroTransition(path *field.Path, oldVal, newVal any) *field.Error {
    if reflect.ValueOf(newVal).IsZero() {
        return nil
    }

     return ValidateImmutable(field.NewPath("Spec", "AcceleratedNetworking"), old.Spec.AcceleratedNetworking, m.Spec.AcceleratedNetworking)

CecileRobertMichon: @nawazkh @willie-yao that's fair. I guess the only value we get out of checking the old field can't be modified (aside from setting it to nil) is making sure the user doesn't think they can enable or disable accelerated networking when that isn't allowed on an existing cluster (I'm describing a situation where an existing cluster has acceleratedNetworking set to nil, both in the deprecated field and in the new field, and the user changes the deprecated field to "true" or "false". Granted, it's an edge-case scenario, but still a valid scenario.

willie-yao: I think that's valid and there's definitely some value in adding this check. Once this PR merges, I'll apply the same change to the roleAssignmentName so users aren't confused about the same scenario.

jackfrancis (approved):
/lgtm
/approve

Issue #3521, opened 2023-05-04
Implement reconciliation of ASO resources


nojnhuh:
- basic create/delete
- BYO Azure resource
- BYO ASO resource
- upgrade should adopt previously managed resources

This work is already mostly complete at https://github.com/nojnhuh/cluster-api-provider-azure/tree/aso. I'm waiting to open a PR with these specific changes until after #3450 merges.
/assign

PR #3548, opened 2023-05-09
add base ASO

nojnhuh:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**: This PR adds basic ASO resource reconciliation logic. It assumes all ASO and Azure resources are managed entirely by CAPZ, so it does not handle cases like BYO ASO resource or BYO Azure resource. Those changes are coming next.

Part of #3521

These changes aren't hooked up to anything yet, so this PR includes no functional changes.

- [ ] cherry-pick candidate

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Release note**:

```release-note
NONE
```


azure/services/aso/aso.go
```diff
@@ -0,0 +1,165 @@
...
+			return nil, azure.WithTransientError(errors.New("ready status unknown"), 20*time.Second)
+		}
+		var readyErr error
+		if cond := conds[i]; cond.Status != metav1.ConditionTrue {
+			switch {
+			case cond.Reason == conditions.ReasonReconciling.Name:
+				readyErr = newOperationNotDoneError(existing)
+			default:
+				readyErr = fmt.Errorf("resource is not Ready: %s", conds[i].Message)
+			}
```
nojnhuh, line 83: The only reason this isn't a plain `if` statement is because one of the next PRs will be adding another `case` branch here.

azure/services/aso/aso_test.go
```diff
@@ -0,0 +1,581 @@
...
+	err error
+}
+
+func (e ErroringDeleteClient) Delete(_ context.Context, _ client.Object, _ ...client.DeleteOption) error {
+	return e.err
+}
+
+// TestCreateOrUpdateResource tests the CreateOrUpdateResource function.
+func TestCreateOrUpdateResource(t *testing.T) {
+	t.Run("ready status unknown", func(t *testing.T) {
```
nojnhuh, line 69: I elected not to use table tests here because even though a lot of this looks like boilerplate, just about every test does something slightly different in a way I have a hunch would be really messy to encode in a table.

azure/services/aso/aso.go
jackfrancis: lol anticipating the case-switch haters

azure/services/aso/aso.go
```diff
@@ -0,0 +1,165 @@
...
+
+			if readyErr != nil {
+				if conds[i].Severity == conditions.ConditionSeverityError {
```
willie-yao, line 86: For my own understanding: if the condition has the severity Error, that means it's always a terminal error? If so, how are transient errors distinguished if there is not condition for it?

willie-yao (commented):
This looks great to me! Just a question from my end, otherwise lgtm.

azure/services/aso/aso.go
nojnhuh: My read of the [ASO docs](https://azure.github.io/azure-service-operator/guide/conditions/#severity) says `Error` should be terminal and any other severity is treated as transient. In either case, we will bubble up the ASO condition's `message` which might provide more context.

azure/services/aso/aso.go
```diff
@@ -0,0 +1,165 @@
...
+	log = log.WithValues("service", serviceName, "resource", resourceName, "namespace", resourceNamespace)
+
+	var existing genruntime.MetaObject
+	if err := s.Client.Get(ctx, client.ObjectKeyFromObject(resource), resource); err != nil {
+		if !apierrors.IsNotFound(err) {
+			return nil, errors.Wrapf(err, "failed to get existing resource %s/%s (service: %s)", resourceNamespace, resourceName, serviceName)
+		}
+	} else {
+		existing = resource
+		log.V(2).Info("successfully got existing resource")
```
jackfrancis, line 68: How do we know at this point that we're in a "got a GET 200" vs "got a 404"?

Related question: if we got a 404, won't `existing.GetConditions()` panic?

nojnhuh: This is the `err == nil` branch, and a 404 or other error would be reflected in `err` and captured in the previous branch.

azure/services/aso/aso.go
```diff
@@ -0,0 +1,165 @@
...
+	diff := cmp.Diff(existing, parameters)
+	if diff == "" {
+		log.V(2).Info("resource up to date")
+		return existing, nil
+	}
+
+	// Create or update the resource with the desired parameters.
+	logMessageVerbPrefix := "creat"
+	if existing != nil {
+		logMessageVerbPrefix = "updat"
```
jackfrancis, line 112: s/updat/update

azure/services/aso/aso.go
jackfrancis: ERR_2_MANY_BRANCHES

azure/services/aso/aso.go
```diff
@@ -0,0 +1,165 @@
...
+		}
+	} else {
+		existing = resource
+		log.V(2).Info("successfully got existing resource")
+
+		// Check if there is an ongoing long running operation.
+		conds := existing.GetConditions()
+		i, readyExists := conds.FindIndexByType(conditions.ConditionTypeReady)
+		if !readyExists {
+			return nil, azure.WithTransientError(errors.New("ready status unknown"), 20*time.Second)
```
CecileRobertMichon, line 74: can `20*time.Second` be a new or reuse an existing const?

azure/services/aso/aso.go
```diff
@@ -0,0 +1,165 @@
...
+				readyErr = newOperationNotDoneError(existing)
+			default:
+				readyErr = fmt.Errorf("resource is not Ready: %s", conds[i].Message)
+			}
+
+			if readyErr != nil {
+				if conds[i].Severity == conditions.ConditionSeverityError {
+					return nil, azure.WithTerminalError(readyErr)
+				}
+				return nil, azure.WithTransientError(readyErr, 20*time.Second)
```
CecileRobertMichon, line 89: same here with `20*time.Second`

azure/interfaces.go
```diff
@@ -127,3 +128,13 @@ type ResourceSpecGetterWithHeaders interface {
...
 	CustomHeaders() map[string]string
 }
+
+// ASOResourceSpecGetter is an interface for getting all the required information to create/update/delete an Azure resource.
+type ASOResourceSpecGetter interface {
+	// ResourceRef returns a concrete, named (and namespaced if applicable) ASO
+	// resource type to facilitate a strongly-typed GET.
+	ResourceRef() genruntime.MetaObject
+	// Parameters returns a modified object if it points to a non-nil resource.
+	// Otherwise it returns a new value.
```
CecileRobertMichon, line 138: can this comment also define the contract for "no updates needed"? i.e. return nil or return the same unmodified object

azure/services/aso/aso.go
```diff
@@ -0,0 +1,165 @@
...
+		}
+		return errors.Wrapf(err, "failed to delete resource %s/%s (service: %s)", resourceNamespace, resourceName, serviceName)
+	}
+
+	return azure.WithTransientError(newOperationNotDoneError(resource), 20*time.Second)
+}
+
+func newOperationNotDoneError(resource metav1.Object) error {
+	return azure.NewOperationNotDoneError(&infrav1.Future{
+		Type:          "ASOReconcile",
```
CecileRobertMichon, line 154: should this differentiate between Create/Update/Delete?

azure/services/aso/aso.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	// Create or update the resource with the desired parameters.
+	logMessageVerbPrefix := "creat"
+	if existing != nil {
+		logMessageVerbPrefix = "updat"
+	}
+	log.V(2).Info(logMessageVerbPrefix+"ing resource", "diff", diff)
+	if existing != nil {
+		err = s.Client.Update(ctx, parameters)
```
CecileRobertMichon, line 116: Since we're patching a k8s object here, have you looked into using CAPI's patch helper instead of directly calling Create/Update?

azure/services/aso/aso.go
nojnhuh: Yep, I'll pull this out.

azure/interfaces.go
nojnhuh: Yep I'll add that.

azure/services/aso/aso.go
nojnhuh: We could, at least between create/update and delete. I'll add that.

AFAICT this value is only logged but doesn't affect any behavior.

azure/services/aso/aso.go
nojnhuh: It looks like that works for update, but we'll still need to call create. I'll add that here.

CecileRobertMichon (commented):
/lgtm

Seems like it wouldn't be too hard to watch the ASO resources and requeue reconciliation when their conditions change and optimize requeues instead of requeuing every 20 seconds, is that something we're tracking as a near term improvement?

CecileRobertMichon (commented):
/lgtm

Seems like it wouldn't be too hard to watch the ASO resources and requeue reconciliation when their conditions change and optimize requeues instead of requeuing every 20 seconds, is that something we're tracking as a near term improvement?

nojnhuh:
> Seems like it wouldn't be too hard to watch the ASO resources and requeue reconciliation when their conditions change and optimize requeues instead of requeuing every 20 seconds, is that something we're tracking as a near term improvement?

I can try adding a watch on top of this and see if that speeds things up. I think that would be out of scope for this particular PR though since there would be some resource kind-specific details there.

nojnhuh:
Just tried adding a watch and it doesn't actually seem to make things any faster, I think because of this: 
https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/fe62bceb85444b5862d7265a76794213cf133e7a/main.go#L198-L202

With or without a watch on the ASO ResourceGroup, I was seeing about 10s between reconciliations. With or without a watch, cutting that interval to 1s showed about 1s between reconciliations.

CecileRobertMichon:
> Just tried adding a watch and it doesn't actually seem to make things any faster, I think because of this

@nojnhuh that makes sense, however it would still be an improvement as we would avoid unnecessary requeues and react faster when a resource does become ready, especially for resources that take longer to reconcile (eg. if a resource takes 4 minutes to reconcile we would queue the controller twice instead of 240s/20s = 12 times). This is especially important for scalability if we're reconciling hundreds of objects from one controller (which is also what that debouncing interval is attempting to help with).

In any case this doesn't have to be part of this PR.

/approve

PR #3552, opened 2023-05-10
adding conditional blocks around Add/Remove Finalizer

luthermonson:
**What type of PR is this?**
/kind cleanup

AddFinalizer/RemoveFinalizer returns a bool if there was actually a modification to the runtime object. This PR will add conditionals around their calls and if there were no modifications to the object to skip the Patch call. I believe how patch works is it will diff the before/after of the path and return nil if they're the same, but this eliminates the extra computation as we already know when we are adding/removing the finalizers.

Fixes #

The main point is to avoid the patch calls in the reconcile loops, I avoided new object creation and anything that looked like reconcile delete. My biggest logic concern is in `EnsureClusterIdentity` so please confirm you feel the code accomplishes the same task with the early return. I made a similar PR to CAPA which was well received and is now in the code base: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4095

**Release Note**
```release-note
Conditional blocks added to Add/Remove Finalizers to avoid unnecessary Patch calls
```

### DIFF
--- a/controllers/helpers.go
+++ b/controllers/helpers.go
@@ -649,20 +649,24 @@ func EnsureClusterIdentity(ctx context.Context, c client.Client, object conditio
 	if err != nil {
 		return err
 	}
+
 	if !scope.IsClusterNamespaceAllowed(ctx, c, identity.Spec.AllowedNamespaces, namespace) {
 		conditions.MarkFalse(object, infrav1.NetworkInfrastructureReadyCondition, infrav1.NamespaceNotAllowedByIdentity, clusterv1.ConditionSeverityError, "")
 		return errors.New("AzureClusterIdentity list of allowed namespaces doesn't include current cluster namespace")
 	}
-	identityHelper, err := patch.NewHelper(identity, c)
-	if err != nil {
-		return errors.Wrap(err, "failed to init patch helper")
+
+	// Remove deprecated finalizer if it exists, Register the finalizer immediately to avoid orphaning Azure resources on delete.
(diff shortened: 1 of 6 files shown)


controllers/helpers.go
```diff
@@ -649,6 +649,14 @@ func EnsureClusterIdentity(ctx context.Context, c client.Client, object conditio
 	if err != nil {
 		return err
 	}
+
+	// Remove deprecated finalizer if it exists, Register the finalizer immediately to avoid orphaning Azure resources on delete.
+	if controllerutil.RemoveFinalizer(identity, deprecatedClusterIdentityFinalizer(finalizerPrefix, namespace, name)) &&
+		controllerutil.AddFinalizer(identity, clusterIdentityFinalizer(finalizerPrefix, namespace, name)) {
+		// if both finalizers are set properly then nothing to do here
+		return nil
```
nojnhuh, line 657: Would it still be valuable to perform the namespace check below even in this case? I wonder if keeping these `Add/RemoveFinalizer` where they were before might be better.

So instead of
```
RemoveFinalizer(...)
AddFinalizer(...)
return Patch()
```
it might be
```
if RemoveFinalizer(...) or AddFinalizer(...) {
  return Patch()
}
return nil
```
like you had in at least one other spot in this PR.

luthermonson: ya we could do that. i wasn't 100% on what this code was trying to accomplish when i wrote it. i dont mind switching back to that

luthermonson:
@nojnhuh fixed up, lemme know if you want anything else changed

CecileRobertMichon (commented):
/lgtm
/approve

luthermonson:
@CecileRobertMichon done!

PR #3575, opened 2023-05-18
Support preexisting network interfaces for AzureMachines

### LINKED ISSUE #3255: Support bring your own Azure Network Interface
/kind feature

Add support for "bring your own" Azure Network Interface. In order to support applications that require IP addresses that never change, even across machine deletion/recreation, we would like to add support to create network interfaces outside of CAPZ and reference them for use by AzureMachines/AzureMachineTemplates. 

I propose that we add the following fields to the `NetworkInterface` [struct](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/0f497f833a710ad2ad720124f2dd3003da9326e8/api/v1beta1/types.go#L686-L701):
```yaml
name: "<name of existing azure network interface>"
resourceGroup: "<name of the resource group containing the network interface>"
```

The controller would implement logic similar to the "bring your own" Virtual Network implementation. The name and resource group are optional fields, and if specified would be used to check for an existing network interface.

If the network interface exists in Azure, the controller would check for the CAPZ ownership tag, and if it is not present, it would not manage the resource, but would only use it for attachment to a virtual machine, and adding to load balancer backend pools as necessary.

If the network interface exists in Azure, and the CAPZ ownership tag is present, CAPZ would manage the lifecycle of network interface.

If the network interface specified in the optional fields does not exist in Azure, then it would be created as it is today, but with the user-specified network interface name and resource group, then attached to the virtual machine.

Currently, the `NetworkInterface` type is also used within the AzureMachinePool spec, but this feature is not applicable there. It is not possible in the Azure VMSS API to create a VMSS with a network profile that specifies a specific network interface. The AzureMachinePool resource would be changed to use its own VMSSNetworkInterface type that does not contain these new fields.

dthorsen:
/kind feature

This PR adds support for using preexisting network interfaces for AzureMachines. This is useful for workloads that must always re-attach the same network interface when a Node is replaced.

Fixes #3255

- [ ] cherry-pick candidate

**TODOs**:
- [X] squashed commits
- [X] includes documentation
- [X] adds unit tests

**Release note**:
```release-note
Support preexisting network interfaces for AzureMachines
action required: all network interfaces created by CAPZ prior to version v1.5.0 
must be manually tagged with the resource ownership tag:
sigs.k8s.io_cluster-api-provider-azure_cluster_CLUSTERNAME: owned
```


dthorsen:
The failing apidiff test is only in the experimental AzureMachinePool resource. It is only a struct name change. The fields are identical to before so this isn't truly a breaking API change. The change was necessary because the bring your own NIC feature is not applicable to MachinePools. VMSS doesn't support such a thing and it doesn't really make sense anyway.

azure/services/networkinterfaces/networkinterfaces.go
```diff
@@ -116,7 +129,40 @@ func (s *Service) Delete(ctx context.Context) error {
...
+// meaning that the network interface's lifecycle is managed.
+func (s *Service) isManaged(ctx context.Context, spec azure.ResourceSpecGetter) (bool, error) {
+	_, _, done := tele.StartSpanWithLogger(ctx, "networkinterfaces.Service.IsManaged")
+	defer done()
+
+	if spec == nil {
+		return false, errors.New("cannot get network interface to check if it is managed: spec is nil")
+	}
+
+	scope := azure.NetworkInterfaceID(s.Scope.SubscriptionID(), spec.ResourceGroupName(), spec.ResourceName())
```
willie-yao, line 142: nit: Should we move this line below the nil checks for TagsGetter and ctx?

azure/services/networkinterfaces/networkinterfaces.go
```diff
@@ -116,7 +129,40 @@ func (s *Service) Delete(ctx context.Context) error {
...
+	}
+
+	tags := converters.MapToTags(tagsMap)
+	return tags.HasOwned(s.Scope.ClusterName()), nil
+}
+
+// IsManaged returns true if the network interface has an owned tag with the cluster name as value,
+// meaning that the network interface's lifecycle is managed.
 func (s *Service) IsManaged(ctx context.Context) (bool, error) {
+	// This method is unused, but it is required to satisfy the ServiceReconciler interface
```
willie-yao, line 166: I may be misunderstanding something here, but would it make sense to move the contents of the unexported `isManaged` here in case the method needs to be used elsewhere? Otherwise, I can see it being an issue in the future if someone needs to use the exported `IsManaged` and gets a true return value even if the NIC is not managed.

config/capz/manager_image_patch.yaml
```diff
@@ -8,5 +8,5 @@ spec:
     spec:
       containers:
       # Change the value of image field below to your controller image URL
-      - image: gcr.io/k8s-staging-cluster-api-azure/cluster-api-azure-controller:latest
+      - image: localhost:5000/ci-e2e/cluster-api-azure-controller-amd64:20230518170459
```
willie-yao, line 11: I think you may have mistakenly committed this after running an E2E test. The latest changes on main should have a fix to prevent this.

config/capz/manager_pull_policy.yaml
```diff
@@ -8,4 +8,4 @@ spec:
     spec:
       containers:
       - name: manager
-        imagePullPolicy: Always
+        imagePullPolicy: IfNotPresent
```
willie-yao, line 11: Same comment as above

willie-yao (commented):
Great work @dthorsen! I left a few comments below. Thanks so much for working on this feature 🚀

CecileRobertMichon:
> VMSS doesn't support such a thing and it doesn't really make sense anyway.

What about VMSS in Flexible Orchestration mode? https://learn.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-orchestration-modes#basic-setup

dthorsen:
I was unaware of Flexible Orchestration mode. Thank you for pointing that out @CecileRobertMichon. I will have to do some reading to better understand their usage.

azure/services/networkinterfaces/networkinterfaces.go
dthorsen: This is indeed confusing. The problem is that the interface this is fulfilling is external. It is called during reconciling delete of an AzureCluster. https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/4cfde3590d3f8b4db78031c64508c72c03b8e342/controllers/azurecluster_reconciler.go#L111
I'd argue that check isn't necessary. The check could be done in the services' `Delete()` implementations. And by stubbing out the call to always return true, the `Delete()` implementation in the networkinterfaces service can decide on a per-interface basis whether it should delete any network interfaces.

dthorsen: Actually, to clarify further, that call isn't even made for the networkinterfaces service, only the resource groups service. The stubbed method is unused, but must be present on all the services to ensure they fulfill the interface.

config/capz/manager_image_patch.yaml
dthorsen: Good catch! Sorry about that!

azure/services/networkinterfaces/networkinterfaces.go
CecileRobertMichon: The `IsManaged()` function was meant to be eventually used to determine if a resource is BYO (and we should skip reconcile) or if its lifecycle is managed by CAPZ. There are several services that implement it and use it (vnets, subnets, etc.), although not as part of the interface.

Since we're adding support for BYO NIC we should be able to use it to return `false` if a NIC is pre-existing. 

Not sure how much of this is impacted by https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3608 though, @nojnhuh PTAL

dthorsen: The problem is that this function can only be called per AzureMachine, not per network interface. Since an AzureMachine can have multiple interfaces, it is possible that some are BYO and others are not. Unless we choose to enforce that you must either bring all your own NIC's or none of them, this function won't be usable as it is.

nojnhuh: I don't think #3608 or ASO in general will fix this particular issue, I think it's something we have to account for in either case. I think the current approach of inlining individual NIC-specific `isManaged()` checks and letting the interface `IsManaged()` no-op seems reasonable for now, but I agree it's worth looking for ways to generalize this a bit more in the future.

nojnhuh: ^ Actually I think the current ASO framework will handle this case since it doesn't use the service-level `IsManaged()` and instead will decide whether or not to skip reconciling individual resources based on whether they're managed or not, just like how it's done inline here.

dthorsen:
As far as I can tell all comments on this PR are addressed. It should be ready for merging unless anyone has additional comments.

willie-yao (approved):
/lgtm

Apologies for the late follow up here @dthorsen. We'll do our best to review this and get it merged before the release!

azure/services/networkinterfaces/networkinterfaces.go
```diff
@@ -116,7 +129,42 @@ func (s *Service) Delete(ctx context.Context) error {
...
+	_, _, done := tele.StartSpanWithLogger(ctx, "networkinterfaces.Service.IsManaged")
+	defer done()
+
+	if spec == nil {
+		return false, errors.New("cannot get network interface to check if it is managed: spec is nil")
+	}
+
+	if s.TagsGetter == nil {
+		return false, fmt.Errorf("tagsgetter is nil")
+	}
```
nawazkh, line 144: Any specific reason on choosing `fmt.Errorf()` than `errors.New()` ?

azure/services/networkinterfaces/networkinterfaces.go
```diff
@@ -116,7 +129,42 @@ func (s *Service) Delete(ctx context.Context) error {
...
+	if ctx == nil {
+		return false, fmt.Errorf("ctx is nil")
+	}
```
nawazkh, line 147: Can we use errors.New instead?

nawazkh (commented):
Minor nits, over looks good to me!

azure/types.go
```diff
@@ -67,7 +67,7 @@ type ScaleSetSpec struct {
 	DiagnosticsProfile           *infrav1.Diagnostics
 	FailureDomains               []string
 	VMExtensions                 []infrav1.VMExtension
-	NetworkInterfaces            []infrav1.NetworkInterface
+	NetworkInterfaces            []infrav1.MachinePoolNetworkInterface
```
nawazkh, line 70: APIDiff test is complaining about this change.
Shall we be just calling it out in release notes about this change or is there more we can do about it? https://github.com/everyone.

dthorsen: This is an interesting one because it is a false-positive from the perspective of the JSON/YAML schema. I asked this during office hours and @mboersma mentioned those types of apidiff false-positives happen from time to time. Only the struct name changed, but no field names have changed. I was also told the apidiff test is not required for PRs for this reason.

azure/services/networkinterfaces/networkinterfaces.go
dthorsen: Just force of habit. Changed it now.

CecileRobertMichon:
> action required: all network interfaces created by CAPZ prior to version v1.5.0 
must be manually tagged with the resource ownership tag

@dthorsen this seems too disruptive of a breaking change for a minor release, what is the impact on existing users who do not wish to use this new feature?

azure/services/networkinterfaces/networkinterfaces.go
```diff
@@ -105,6 +111,12 @@ func (s *Service) Delete(ctx context.Context) error {
 	//  Order of precedence (highest -> lowest) is: error that is not an operationNotDoneError (i.e. error deleting) -> operationNotDoneError (i.e. deleting in progress) -> no error (i.e. deleted)
 	var result error
 	for _, nicSpec := range specs {
+		if managed, err := s.isManaged(ctx, nicSpec); err == nil && !managed {
+			log.V(4).Info("Skipping network interface deletion in custom network interface mode")
```
CecileRobertMichon, line 115: This might be an opportunity to make logging more visible and add a link to https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3575/files#diff-a109f59250df65bd0d5487b914d0bce7b1a07717583e9414cfe2c5e11f4fcf13R45 for users that may not have read the release notes or docs

azure/types.go
nawazkh: ^ thanks for the context, Dane! Makes sense!

dthorsen:
> > action required: all network interfaces created by CAPZ prior to version v1.5.0
> > must be manually tagged with the resource ownership tag
> 
> @dthorsen this seems too disruptive of a breaking change for a minor release, what is the impact on existing users who do not wish to use this new feature?
> 
> Edit: also this should say v1.10.0 (or v1.11.0 if it doesn't make it to 1.10)

@CecileRobertMichon Regarding how disruptive this change is, it is a little difficult to gauge impact. This is only an issue for users that have hopped from v1.5.0 directly to v1.10.0 / probably v1.11.0 now. There is also precedent for this change in the past. Example here: https://capz.sigs.k8s.io/topics/custom-dns.html?highlight=owned#manage-dns-via-capz-tool 
> Private DNS when created by CAPZ can be managed by CAPZ tool itself automatically. To give the flexibility to have BYO as well as managed DNS zone, an enhancement is made that causes all the managed zones created in the CAPZ version before the enhancement changes to be treated as unmanaged. The enhancement is captured in PR [1791](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1791)

> To manage the private DNS via CAPZ please tag it manually from azure portal.

No, capz resource ownership tagging was introduced by an earlier PR done by someone else, released in v1.5.0. Ultimately, the impact would simply be that some network interfaces could be left in existence after AzureMachine deletion if they were created prior to v1.5.0, but deleted in v1.11.0 or later, and that is only if they do not tag the older network interfaces prior to upgrading to v1.11.0.

If this change is deemed to disruptive for a minor release, then perhaps we could put it behind a feature flag in v1.11.0?

api/v1beta1/types.go
```diff
@@ -734,6 +734,36 @@ type NetworkInterface struct {
...
 	// +optional
 	AcceleratedNetworking *bool `json:"acceleratedNetworking,omitempty"`
+
+	// Name optionally allows you to override the Azure NetworkInterface name. If left unspecified, CAPZ will
+	// generate a NetworkInterface name based on the VM name.
+	// If specified, the specified networkInterface does not exist, CAPZ will create it when the VM is created and delete it when the VM is deleted.
+	// If you specify an existing Azure NetworkInterface, CAPZ will attach and detach the interface from the VM when the VM is created and deleted,
+	// but CAPZ will not delete the NetworkInterface if the VM is deleted.
+	// +optional
+	Name *string `json:"name,omitempty"`
```
Jont828, line 744: Does this need to be a string pointer instead of a string? I would assume we can tell it's unspecified by an empty string, meaning it can't be a network interface name.

api/v1beta1/types.go
```diff
@@ -734,6 +734,36 @@ type NetworkInterface struct {
...
+
+	// ResourceGroup optionally specifies the resource group for the NetworkInterface resource.
+	// +optional
+	ResourceGroup *string `json:"resourceGroup,omitempty"`
```
Jont828, line 748: Same as above

api/v1beta1/types.go
```diff
@@ -734,6 +734,36 @@ type NetworkInterface struct {
...
+}
+
+// MachinePoolNetworkInterface defines a VMSS network interface profile.
+type MachinePoolNetworkInterface struct {
+	// SubnetName specifies the subnet in which the new network interface will be placed.
+	SubnetName string `json:"subnetName,omitempty"`
```
Jont828, line 754: I think we could drop the `omitempty` tag if this is field not optional.

azure/scope/machine.go
```diff
@@ -237,9 +237,17 @@ func (m *MachineScope) NICSpecs() []azure.ResourceSpecGetter {
 	isMultiNIC := len(m.AzureMachine.Spec.NetworkInterfaces) > 1
 
 	for i := 0; i < len(m.AzureMachine.Spec.NetworkInterfaces); i++ {
+		nic := m.AzureMachine.Spec.NetworkInterfaces[i]
+		nicName := ""
 		isPrimary := i == 0
-		nicName := azure.GenerateNICName(m.Name(), isMultiNIC, i)
-		nicSpecs = append(nicSpecs, m.BuildNICSpec(nicName, m.AzureMachine.Spec.NetworkInterfaces[i], isPrimary))
+
+		if nic.Name != nil && *nic.Name != "" {
```
Jont828, line 244: Same as the earlier comment, I think we could probably make `nic.Name` a `string` instead of `*string` in this case.

azure/services/networkinterfaces/networkinterfaces.go
```diff
@@ -116,7 +128,42 @@ func (s *Service) Delete(ctx context.Context) error {
...
+// isManaged returns true if the network interface has an owned tag with the cluster name as value,
+// meaning that the network interface's lifecycle is managed.
+func (s *Service) isManaged(ctx context.Context, spec azure.ResourceSpecGetter) (bool, error) {
+	_, _, done := tele.StartSpanWithLogger(ctx, "networkinterfaces.Service.IsManaged")
+	defer done()
+
+	if spec == nil {
+		return false, errors.New("cannot get network interface to check if it is managed: spec is nil")
+	}
+
```
Jont828, line 140: Could we drop newline for spacing consistency? And same for the other `if` blocks below in this function.

docs/book/src/topics/custom-network-interfaces.md
```diff
@@ -0,0 +1,57 @@
...
+        cachingType: None
+        diskSizeGB: 30
+        osType: Linux
+      sshPublicKey: mykey
+      vmSize: Standard_B2s
+```
+
+The pre-existing network interface can be in the same resource group or a different resource group in the same
+subscription as the target cluster. When deleting the `AzureMachine`, the network interface will only be
+deleted if they are "managed" by capz, ie. they were created during `AzureMachine` deployment.  Pre-existing
```
Jont828, line 33:
```suggestion
deleted if they are "managed" by CAPZ, i.e. they were created during `AzureMachine` deployment. Pre-existing
```

docs/book/src/topics/custom-network-interfaces.md
```diff
@@ -0,0 +1,57 @@
...
+network interfaces will *not* be deleted. If a resource group is specified, it must already exist. CAPZ will
+*not* create or delete a resource group that is specific to a network interface. If the resource group is
+omitted, it will default to the resourceGroup of the `AzureCluster`.
```
Jont828, line 36:
```suggestion
omitted, it will default to the resource group of the `AzureCluster`.
```

docs/book/src/topics/custom-network-interfaces.md
```diff
@@ -0,0 +1,57 @@
+# Custom Network Interfaces for AzureMachines
+
+## Pre-existing Network Interfaces
```
Jont828, line 3: Newline for consistency with other docs in the repo
```suggestion
## Pre-existing Network Interfaces

```

docs/book/src/topics/custom-network-interfaces.md
```diff
@@ -0,0 +1,57 @@
...
+```
+
+The pre-existing network interface can be in the same resource group or a different resource group in the same
+subscription as the target cluster. When deleting the `AzureMachine`, the network interface will only be
+deleted if they are "managed" by capz, ie. they were created during `AzureMachine` deployment.  Pre-existing
+network interfaces will *not* be deleted. If a resource group is specified, it must already exist. CAPZ will
+*not* create or delete a resource group that is specific to a network interface. If the resource group is
+omitted, it will default to the resourceGroup of the `AzureCluster`.
+
+## Custom Network Interface
```
Jont828, line 38:
```suggestion
## Custom Network Interface

```

docs/book/src/topics/custom-network-interfaces.md
```diff
@@ -0,0 +1,57 @@
...
+Alternatively, if you specify a network interface name and optionally a resource group, but the network
```
Jont828, line 39:
```suggestion
Alternatively, if you specify a network interface name and optionally a resource group but the network
```

Jont828 (commented):
Nice work! Just had a some small things.

CecileRobertMichon:
/close 

Please reopen once this is ready for review

PR #3594, opened 2023-05-25
Prioritize workload ID creds for SDKv2 services

### LINKED ISSUE #3569: tenant and client ID are not used from AzureClusterIdentity object for virtualmachineimages client
/kind bug

The client at this code 
https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/services/virtualmachineimages/client.go#L52
does not use the tenant ID and client ID that is passed in AzureClusterIdentity object. 

When tenant ID and cluster ID is passed via AzureClusterIdentity, then the values present on AzureClusterIdentity should be used as a precedence over tenant ID and client ID env variables. Or
There should be a way to create client where we can explicitly pass tenant and client ID. 

**Environment:**

mboersma:
**What type of PR is this?**

/kind bug

Replaces the SDKv2 `DefaultAzureCredential` with a `ChainedTokenCredential` that prefers workload ID and managed ID to environment (or `az` CLI) auth.

Fixes #3569

- [ ] cherry-pick candidate

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/azure/defaults.go
+++ b/azure/defaults.go
@@ -20,9 +20,11 @@ import (
 	"fmt"
 	"net/http"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
 	"github.com/Azure/go-autorest/autorest"
 	"sigs.k8s.io/cluster-api-provider-azure/util/tele"
 	"sigs.k8s.io/cluster-api-provider-azure/version"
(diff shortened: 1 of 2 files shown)


azure/defaults.go
```diff
@@ -383,6 +385,54 @@ func (p userAgentPolicy) Do(req *policy.Request) (*http.Response, error) {
 	return req.Next()
 }
 
+// NewDefaultCredential creates a new credential that uses the following methods, in order of precedence:
+// 1. Workload Identity
```
CecileRobertMichon, line 389: should this be AAD Pod identity since we're not using workload identity anywhere in the main branch yet ?

azure/defaults.go
```diff
@@ -383,6 +385,54 @@ func (p userAgentPolicy) Do(req *policy.Request) (*http.Response, error) {
...
+// 2. Managed Identity
+// 3. Environment variables
+// 4. Azure CLI.
+func NewDefaultCredential(clientOptions *azcore.ClientOptions) (*azidentity.ChainedTokenCredential, error) {
```
sonasingh46, line 393: This function should read client ID and tenant ID from the AzureClusterIdentity or have a way to take this input preferably via a structure. 

I can think something like below : 

```
type CredentialOptions struct {
*azcore.ClientOptions
clientID string
tenantID string
}
```
func NewDefaultCredential(credOptions *CredentialOptions) (*azidentity.ChainedTokenCredential, error) {

}
```
Also at line no 400 we can set the tenantID and clientID in `azidentity.WorkloadIdentityCredentialOptions` options to whatever is set in `credOptions` which will fix the problem.

azure/defaults.go
sonasingh46: If we go via the chained identity mechanism of sdk, I think we should keep the order as it is specified in the sdk as of now i.e 
env > workload identity > managed identity > cli creds

mboersma: > So I think we can have wrapper over SDK that picks and choose what to use as what is specified in the AzureClusterIdentity.

I like this approach, because the SDK doesn't make it easy to reimplement the default cred chained logic properly. Several items we would want to reuse are not exported. What I did here is a "clean room" reimplementation because it's probably legally problematic to actually copy and modify the SDK sources here.

mboersma:
/close

I'll take a different approach with a new PR that adds virtualmachineimages back.

PR #3596, opened 2023-05-30
Improve api server LB probe

kahun:
**What type of PR is this?**
/kind feature

The apiserver can fail if a control plane is starting o regenerating and the service is up but not ready yet.

**TODOs**:
- [X] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Release note**:
```release-note
NONE
```

### DIFF
--- a/azure/services/loadbalancers/loadbalancers.go
+++ b/azure/services/loadbalancers/loadbalancers.go
@@ -27,10 +27,11 @@ import (
 )
 
 const (
-	serviceName = "loadbalancers"
-	tcpProbe    = "TCPProbe"
-	lbRuleHTTPS = "LBRuleHTTPS"
-	outboundNAT = "OutboundNATAllProtocols"
+	serviceName           = "loadbalancers"
+	httpsProbe            = "HTTPSProbe"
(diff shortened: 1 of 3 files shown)


CecileRobertMichon:
@jackfrancis do you have any thoughts on this one?

overall LGTM but I wonder if there could be any downsides

jackfrancis:
@kahun the purpose of this change is to prefer a non-Running apiserver Pod to transient non-2xx responses that occur because TCP 6443 connection is active but the web server is not yet bootstrapped and responding to the `/healthz` URL path?

kahun:
> @kahun the purpose of this change is to prefer a non-Running apiserver Pod to transient non-2xx responses that occur because TCP 6443 connection is active but the web server is not yet bootstrapped and responding to the `/healthz` URL path?

During a control plane failure or an upgrade process, when a new control plane is being initialized and the TCP 6443 port is already active but the web server is not ready yet, one out of three calls to the k8s API server will fail. This change ensures that the apiserver pod is made available in the load balancer only when it is completely ready.

PR #3608, opened 2023-06-06
add BYO ASO resource

nojnhuh:
**What type of PR is this?**
/kind feature

This PR makes it possible for users to bring-their-own ASO resources and have them participate in a CAPZ cluster. BYO ASO resources can be referred to by CAPZ objects but will not be modified or deleted by CAPZ, like how BYO Azure resources are handled currently.

Part of #3521

These changes are currently non-functional, but this branch includes all of the requisite changes to make this operable: https://github.com/kubernetes-sigs/cluster-api-provider-azure/compare/main...nojnhuh:cluster-api-provider-azure:aso-wired

- [ ] cherry-pick candidate

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/azure/services/aso/aso.go
+++ b/azure/services/aso/aso.go
@@ -29,12 +29,20 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
 	"sigs.k8s.io/cluster-api-provider-azure/azure"
+	"sigs.k8s.io/cluster-api-provider-azure/util/maps"
 	"sigs.k8s.io/cluster-api-provider-azure/util/tele"
 	"sigs.k8s.io/cluster-api/util/patch"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const (
+	// ReconcilePolicyAnnotation is the annotation key for ASO's reconcile policy mechanism.
+	ReconcilePolicyAnnotation = "serviceoperator.azure.com/reconcile-policy"
+	// ReconcilePolicySkip indicates changes will not be PUT or DELETEd in Azure from ASO.
+	ReconcilePolicySkip = "skip"
+	// ReconcilePolicyManage indicates changes will be PUT and DELETEd in Azure from ASO.
+	ReconcilePolicyManage = "manage"
+
 	requeueInterval = 20 * time.Second
 
 	createOrUpdateFutureType = "ASOCreateOrUpdate"
@@ -45,12 +53,15 @@ const (
 // and deletion of resources using ASO.
 type Service struct {
 	client.Client
+
+	clusterName string
 }
 
 // New creates a new ASO service.
-func New(ctrlClient client.Client) *Service {
+func New(ctrlClient client.Client, clusterName string) *Service {
 	return &Service{
-		Client: ctrlClient,
+		Client:      ctrlClient,
+		clusterName: clusterName,
 	}
 }
 
@@ -74,6 +85,11 @@ func (s *Service) CreateOrUpdateResource(ctx context.Context, spec azure.ASOReso
 	} else {
 		existing = resource
 		log.V(2).Info("successfully got existing resource")
+
+		if !ownedByCluster(existing.GetLabels(), s.clusterName) {
+			log.V(4).Info("skipping reconcile for unmanaged resource")
+			return existing, nil
+		}
 
 		// Check if there is an ongoing long running operation.
 		conds := existing.GetConditions()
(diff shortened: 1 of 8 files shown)


jackfrancis:
Can we ship documentation with this change? Definitely seems cool but also lots gotchas.

nojnhuh:
Should we omit docs until ASO is actually exposed? I can queue up some docs and link them here.

And are there specific gotchas you're thinking of that don't already exist with BYO resources without ASO?

jackfrancis:
@nojnhuh I guess I wasn't aware there was an existing BYO story where CAPZ will "adopt" pre-existing resources in your cluster, unless this PR is literally just the ASO equivalent of VNET, subnet, and the other things that are commonly integrated into a CAPZ cluster that are created by the user prior to the creation of the cluster.

nojnhuh:
Right, this PR enables that same resource group/vnet/subnet pre-creation as ASO resources. I haven't built any mechanism to restrict this behavior to only those types though, so this would allow users to pre-create any resource type in ASO. Is there a reason we only allow those specific types today?

Handling resources pre-created outside of ASO will come in a future change.

jackfrancis:
> Right, this PR enables that same resource group/vnet/subnet pre-creation as ASO resources. I haven't built any mechanism to restrict this behavior to only those types though, so this would allow users to pre-create any resource type in ASO. Is there a reason we only allow those specific types today?
> 
> Handling resources pre-created outside of ASO will come in a future change.

I don't think there's any reason we want to restrict the ASO resource types that can co-exist in the same cluster w/ CAPI mgmt cluster components, but I would imagine that there are a finite set of resources that CAPZ will be able to actually positively identify as serving a CAPZ cluster, and thus able to use instead of creating its own equivalent resource (via ASO) from scratch to serve the same purpose.

I think we're on the same page?

nojnhuh:
> I would imagine that there are a finite set of resources that CAPZ will be able to actually positively identify as serving a CAPZ cluster, and thus able to use instead of creating its own equivalent resource (via ASO) from scratch to serve the same purpose.

This works the same way I think the same scenario does today, where if CAPZ finds an existing resource with the expected identifier (resource group name/resource name for Azure or namespace/name for ASO), then CAPZ will use that instead of trying to create another resource. There isn't a way to explicitly check "is this ASO resource participating in a CAPZ cluster?" just by looking at the ASO resource.

azure/services/aso/aso.go
```diff
@@ -120,6 +135,21 @@ func (s *Service) CreateOrUpdateResource(ctx context.Context, spec azure.ASOReso
 	parameters.SetName(resourceName)
 	parameters.SetNamespace(resourceNamespace)
 
+	labels := make(map[string]string)
+	anns := make(map[string]string)
```
CecileRobertMichon, line 139: not a fan of the abbreviation, annotations is more clear/explicit to me (but feel free to disagree)

azure/services/aso/aso.go
```diff
@@ -188,3 +232,35 @@ func deepCopyOrNil(obj genruntime.MetaObject) genruntime.MetaObject {
...
+	}
+
+	return ownedByCluster(resource.GetLabels(), clusterName), nil
+}
+
+func ownedByCluster(labels map[string]string, clusterName string) bool {
+	return labels[infrav1.OwnedByClusterLabelKey] == clusterName
+}
+
+func mergeMaps[K comparable, V any](base map[K]V, overrides map[K]V) map[K]V {
```
CecileRobertMichon, line 255: Is there no existing util out there in go libraries? If not, should probably go in a util package since it's generic / not directly related to ASO

azure/services/aso/aso.go
```diff
@@ -45,12 +52,15 @@ const (
...
+	clusterName string
 }
 
 // New creates a new ASO service.
-func New(ctrlClient client.Client) *Service {
+func New(ctrlClient client.Client, clusterName string) *Service {
 	return &Service{
-		Client: ctrlClient,
+		Client:      ctrlClient,
+		clusterName: clusterName,
```
CecileRobertMichon, line 63: will this always be cluster name or would there be a use case for making it more generic (eg. `ownerName`) so a resource can be "owned" by an AzureMachine or AzureMachinePool for example (as opposed to the cluster itself)?

azure/services/asogroups/groups.go
```diff
@@ -90,39 +88,33 @@ func (s *Service) Delete(ctx context.Context) error {
...
-		}
-		return errors.Wrap(err, "could not get resource group management state")
-	}
-	if !managed {
-		log.V(2).Info("Skipping resource group deletion in unmanaged mode")
-		return nil
-	}
-
-	err = s.DeleteResource(ctx, groupSpec, ServiceName)
+	err := s.DeleteResource(ctx, groupSpec, ServiceName)
```
CecileRobertMichon, line 91: I think we're very close to completely getting rid of this code and having one common function for all services (food for thought)

azure/services/aso/aso.go
nojnhuh: It was honestly easier to write this than to try to find something that already exists that works the same way. I think this is small enough that it's not worth looking too deep for. `golang.org/x/exp/maps` is the only place off the top of my head I would think already has something like this but it doesn't, and I can't think of a similar util that exists in a module we already import. I'll move this to `util/maps`at least.

azure/services/asogroups/groups.go
nojnhuh: Yep, I've been trying to work in that direction where it seems reasonable for this first pass.

nojnhuh:
(reminder to squash)
/hold

CecileRobertMichon (commented):
/lgtm

CecileRobertMichon:
@nojnhuh can you please squash/rebase to fix the codespell error?

willie-yao (approved):
/lgtm

nojnhuh:
> (reminder to squash) /hold

Squashed!
/hold cancel

PR #3612, opened 2023-06-07
Update CAPI to v1.4.3

mboersma:
**What type of PR is this?**

/kind cleanup

Updates CAPI references to use the [v1.4.3 release](https://github.com/kubernetes-sigs/cluster-api/releases/tag/v1.4.3). Also bumps cert-manager to v1.12.1 to [match CAPI](https://github.com/kubernetes-sigs/cluster-api/pull/8748). Reimplements some trivial string processing methods that were deprecated in CAPI.

Closes #3613 

- [x] cherry-pick candidate

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Update CAPI to v1.4.3
```


azure/scope/machine.go
```diff
@@ -451,22 +451,14 @@ func (m *MachineScope) Role() string {
...
 }
 
 // ProviderID returns the AzureMachine providerID from the spec.
 func (m *MachineScope) ProviderID() string {
-	parsed, err := noderefutil.NewProviderID(pointer.StringDeref(m.AzureMachine.Spec.ProviderID, ""))
-	if err != nil {
-		return ""
-	}
-	return parsed.String()
+	return utilazure.ProviderID(pointer.StringDeref(m.AzureMachine.Spec.ProviderID, ""))
```
CecileRobertMichon, line 461: I wonder if we really need to do the parsing here or if we should just return the string as is. Since CAPI has gotten rid of Provider ID parsing altogether. And we're also not returning any error here if the parsing fails so this isn't adding any validation value.

azure/scope/machine.go
```diff
@@ -451,22 +451,14 @@ func (m *MachineScope) Role() string {
...
 
-// GetVMID returns the AzureMachine instance id by parsing Spec.FakeProviderID.
+// GetVMID returns the AzureMachine instance id by parsing the scope's providerID.
 func (m *MachineScope) GetVMID() string {
-	parsed, err := noderefutil.NewProviderID(m.ProviderID())
-	if err != nil {
-		return ""
-	}
-	return parsed.ID()
+	return utilazure.GetVMID(m.ProviderID())
```
CecileRobertMichon, line 456: same thing here

azure/scope/machine.go
mboersma: It passes it through regex validation, but doesn't actually return an error if it doesn't match (following the existing contract of these funcs). But you could argue the `ProviderID()` func should be a no-op. I'll try that.

azure/scope/machine.go
CecileRobertMichon: For this one we need to extract the VM ID so we do need to parse it. Could we use https://github.com/Azure/azure-sdk-for-go/blob/a688cfe08303253b05efb19a7645fbdb24d7375e/sdk/azcore/arm/resource_identifier.go#L20 instead of writing our own parsing function?

azure/scope/machine.go
mboersma: We do have a MachineScope test that expects its `ProviderID()` func to do validation here: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/scope/machine_test.go#L198

IDK if the unit test was just written to match the (unneeded) parsing functionality or not, but it's a simple change if we're ok removing that test case.

mboersma:
/hold

Let's rebase this after #3616 merges, so it doesn't duplicate code.

azure/scope/machinepool.go
```diff
@@ -153,13 +152,13 @@ func (m *MachinePoolScope) Name() string {
...
-// ProviderID returns the AzureMachinePool ID by parsing Spec.FakeProviderID.
+// ProviderID returns the AzureMachinePool ID by parsing Spec.ProviderID.
 func (m *MachinePoolScope) ProviderID() string {
-	parsed, err := noderefutil.NewProviderID(m.AzureMachinePool.Spec.ProviderID)
+	resourceID, err := azure.ParseResourceID(m.AzureMachinePool.Spec.ProviderID)
 	if err != nil {
 		return ""
 	}
-	return parsed.ID()
+	return resourceID.Name
```
CecileRobertMichon, line 161: is resourceID.Name() the equivalent of what we had before here? Seems that we're returning the instance ID (which itself seems strange since the function name implies it returns the provider ID of the resource.

Let's add a small unit test that passes before the change to validate before and after is equivalent?

mboersma: Comparing the code behind each, `resourceID.Name` should be equivalent, but I agree it deserves a unit test. Will do.

mboersma: I added a unit test for `MachinePoolScope.ProviderID()` against the `main` branch, then cherry-picked it into this branch to see that it continues to pass after the reimplementation. 👍🏻

azure/scope/machinepool_test.go
```diff
@@ -98,6 +98,58 @@ func TestMachinePoolScope_Name(t *testing.T) {
...
+		name             string
+		machinePoolScope MachinePoolScope
+		want             string
+	}{
+		{
+			name: "valid providerID",
+			machinePoolScope: MachinePoolScope{
+				AzureMachinePool: &infrav1exp.AzureMachinePool{
+					Spec: infrav1exp.AzureMachinePoolSpec{
+						ProviderID: "azure:///subscriptions/123/resourceGroups/rg/providers/Microsoft.Compute/virtualMachines/vm",
```
CecileRobertMichon, line 113: can we add a unit test with a VMSS provider ID? Would be good to have both coverage for both uniform and flex mode

jackfrancis (approved):
/lgtm

azure/scope/machinepool_test.go
mboersma: Sure, good idea.

mboersma: Updated with test cases specifically for VMSS Uniform and Flex strings.

PR #3616, opened 2023-06-08
Allow azure:// prefix when parsing resource IDs

### LINKED ISSUE #3597: userAssignedIdentities providerID value format unexpectedly changed in CAPZ v1.9.0
/kind bug

Tried to bootstrap a new cluster with an AzureMachineTemplate that worked in CAPZ v1.7.0. The azurejson_machinetemplate_controller failed to create a secret for the azure json. The following error was emitted in the logs:
```
E0530 20:42:02.758002       1 controller.go:329]  "msg"="Reconciler error" "error"="failed to get user-assigned identity ClientID: invalid resource ID: resource id 'azure:///subscriptions/<snip>/resourcegroups/<snip>/providers/Microsoft.ManagedIdentity/userAssignedIdentities/<snip>' must start with '/'" "AzureMachineTemplate"={"name":"<snip>","namespace":"<snip>"} "controller"="azuremachinetemplate" "controllerGroup"="infrastructure.cluster.x-k8s.io" "controllerKind"="AzureMachineTemplate" "name"="<snip>" "namespace"="<snip>" "reconcileID"="<snip>"
``` 

If I use an AzureMachineTemplate that starts with a `/subscriptions` instead of `azure:///subscriptions` the error goes away and things work as expected. However this is not how the field is documented to be used [here](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/19f7a4a54993b826b9386e1c2ea7d483e83553d0/api/v1beta1/types.go#L526). 

It looks like this bug was introduced because the behavior of the new `ParseResourceID` function from the new arm package is not fully compatible with the `ParseResourceID` function from the autorest package that was used previously.
Old function: https://github.com/Azure/go-autorest/blob/autorest/v0.11.29/autorest/azure/azure.go#L215-L217
New function: https://github.com/Azure/azure-sdk-for-go/blob/sdk/azcore/v1.6.0/sdk/azcore/arm/internal/resource/resource_identifier.go#L64C6-L71

Existing, working AzureMachineTemplate yaml should work when upgrading CAPZ from v1.7.0 to v1.9.2.

It looks like the breaking change was introduced in #3337.

**Environment:**

- cluster-api-provider-azure version: v1.9.2
- Kubernetes version: (use `kubectl version`):  v1.21.14
- OS (e.g. from `/etc/os-release`): Ubuntu 20.04

mboersma:
**What type of PR is this?**

/kind bug

Handles user-assigned identity resource identifiers starting with an `azure://` prefix, which the new `ParseResourceID()` func doesn't tolerate.

Fixes #3597

- [x] cherry-pick candidate

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Allow azure:// prefix when parsing resource IDs
```

### DIFF
--- a/azure/services/identities/client.go
+++ b/azure/services/identities/client.go
@@ -19,7 +19,6 @@ package identities
 import (
 	"context"
 
-	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
 	"github.com/Azure/azure-sdk-for-go/services/msi/mgmt/2018-11-30/msi"
 	"github.com/Azure/go-autorest/autorest"
 	"sigs.k8s.io/cluster-api-provider-azure/azure"
@@ -63,7 +62,7 @@ func (ac *AzureClient) GetClientID(ctx context.Context, providerID string) (stri
 	ctx, _, done := tele.StartSpanWithLogger(ctx, "identities.GetClientID")
 	defer done()
 
-	parsed, err := arm.ParseResourceID(providerID)
+	parsed, err := azure.ParseResourceID(providerID)
 	if err != nil {
 		return "", err
 	}
--- a/azure/defaults.go
+++ b/azure/defaults.go
@@ -19,7 +19,9 @@ package azure
 import (
 	"fmt"
 	"net/http"
+	"strings"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
 	"github.com/Azure/go-autorest/autorest"
 	"sigs.k8s.io/cluster-api-provider-azure/util/tele"
 	"sigs.k8s.io/cluster-api-provider-azure/version"
@@ -360,4 +362,9 @@ func msCorrelationIDSendDecorator(snd autorest.Sender) autorest.Sender {
 		}
 		return snd.Do(r)
 	})
+}
+
+// ParseResourceID parses a string to an *arm.ResourceID, first removing any "azure://" prefix.
+func ParseResourceID(id string) (*arm.ResourceID, error) {
+	return arm.ParseResourceID(strings.TrimPrefix(id, ProviderIDPrefix))
 }
--- a/azure/defaults_test.go
+++ b/azure/defaults_test.go
@@ -117,3 +117,55 @@ func TestMSCorrelationIDSendDecorator(t *testing.T) {
 		receivedReq.Header.Get(string(tele.CorrIDKeyVal)),
 	).To(Equal(string(corrID)))
 }
+
+func TestParseResourceID(t *testing.T) {
+	g := NewWithT(t)
+
+	tests := []struct {
+		name         string
+		id           string
+		expectedName string
+		errExpected  bool
+	}{
+		{
+			name:         "invalid",
+			id:           "invalid",
+			expectedName: "",
+			errExpected:  true,
+		},
+		{
+			name:         "invalid: must start with slash",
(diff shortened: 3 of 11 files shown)


mboersma:
This is a quick fix for that specific issue. Alternately, I could write a wrapper function around `ParseResourceID()` and have our codebase call that consistently.

controllers/azurejson_machine_controller.go
```diff
@@ -216,7 +217,7 @@ func (r *AzureJSONMachineReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		// TODO: remove this ClientID lookup code when the fixed cloud-provider-azure is default
 		idsClient := identities.NewClient(clusterScope)
 		userAssignedIdentityIfExists, err = idsClient.GetClientID(
-			ctx, azureMachine.Spec.UserAssignedIdentities[0].ProviderID)
+			ctx, strings.TrimPrefix(azureMachine.Spec.UserAssignedIdentities[0].ProviderID, "azure://"))
```
CecileRobertMichon, line 220: can the azure prefix be a const?

jackfrancis (commented):
you are good at doing work while being totally present in a meeting at the same time

azure/services/identities/client.go
```diff
@@ -63,7 +64,7 @@ func (ac *AzureClient) GetClientID(ctx context.Context, providerID string) (stri
 	ctx, _, done := tele.StartSpanWithLogger(ctx, "identities.GetClientID")
 	defer done()
 
-	parsed, err := arm.ParseResourceID(providerID)
+	parsed, err := arm.ParseResourceID(strings.TrimPrefix(providerID, "azure://"))
```
CecileRobertMichon, line 67: Did you remove the other changes because they were not needed? in any case this should probably be a const as well

Do we have webhook validation to ensure the provided ID is in this format?

mboersma: I refactored it so the change is in one place rather than three. I'll take a look at webhook validation and a unit test for this func.

controllers/azurejson_machine_controller.go
mboersma: It already is: `ProviderIDPrefix` in azure/defaults. Will update.

azure/services/identities/client.go
mboersma: I refactored this to be more general, so anywhere `arm.ParseResourceID()` was called, now we call a wrapper func that first removes the "azure://" prefix if it's present. I think this is a safer fix that prevents other similar problems.

mboersma:
> want to squash?

More changes incoming, including webhook validation and unit tests. Sorry for the thrashing, should have been a [WIP].

mboersma:
/retitle Allow azure:// prefix when parsing resource IDs

mboersma:
/override codecov/patch

This change is being penalized for three changed lines of code that were previously not unit tested. I looked at each one, and it's not trivial to write a test harness around those funcs, so I think it's out of scope for this PR to add that. I entered an issue to add unit test coverage: #3622.

mboersma:
I'm confused; I thought we had configured prow so codecov was informational, not blocking. Yet prow is saying it won't merge because codecov has not succeeded.

mboersma:
Can we do "`override/tide`"?

CecileRobertMichon (commented):
/lgtm
/approve

PR #3618, opened 2023-06-09
Validate providerID for user-assigned IDs in webhook

mboersma:
**What type of PR is this?**

/kind bug

Adds `ProviderID` to the validation of user-assigned identities called by the AzureMachine webhook. Also moves `ParseResourceID()` to `util/azure` to avoid circular import errors.

Inspired by [this comment](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3616#discussion_r1223374172).
Refs #3597

- [x] cherry-pick candidate

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Validate providerID for user-assigned IDs in webhook
```


api/v1beta1/azuremachine_validation.go
```diff
@@ -124,9 +126,19 @@ func ValidateSystemAssignedIdentity(identityType VMIdentity, oldIdentity, newIde
...
 
-	if identityType == VMIdentityUserAssigned && len(userAssignedIdentities) == 0 {
-		allErrs = append(allErrs, field.Required(fldPath, "must be specified for the 'UserAssigned' identity type"))
+	if identityType == VMIdentityUserAssigned {
+		if len(userAssignedIdentities) == 0 {
+			allErrs = append(allErrs, field.Required(fldPath, "must be specified for the 'UserAssigned' identity type"))
+		}
+		for _, identity := range userAssignedIdentities {
+			if identity.ProviderID != "" {
+				if _, err := arm.ParseResourceID(strings.TrimPrefix(identity.ProviderID, "azure://")); err != nil {
```
mboersma, line 135: We can't use the `azure.ParseResourceID()` func added in #3616 because it would cause a circular import error.

CecileRobertMichon: The azure.go file itself doesn't import the `api` package, I wonder if we should move all azure "utils" functions to its own package so we can use the same logic in both places

CecileRobertMichon: @mboersma thoughts on this?

mboersma:
/hold

I rebased this on #3612 so moving `ParseResourceID()` wasn't too confusing. I'll rebase again on `main` after that merges.

CecileRobertMichon (commented):
/lgtm

PR #3633, opened 2023-06-15
add BYO Azure resource with ASO

nojnhuh:
**What type of PR is this?**
/kind feature

This PR makes it possible from the ASO framework to pre-create Azure resources by any means and allow them to participate in a CAPI/CAPZ cluster. For resources managed outside of CAPZ's ASO, CAPZ will create an associated ASO proxy object created with the `serviceoperator.azure.com/reconcile-policy: skip` annotation so ASO does not reconcile the resource.

More context in the proposal: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/0d41b8fab42bad5e6fc7790854769da143280aa1/docs/proposals/20230123-azure-service-operator.md#byo-resources

Part of #3521

This branch includes this change and the necessary plumbing to make this functional if you'd like to test this yourself: https://github.com/kubernetes-sigs/cluster-api-provider-azure/compare/main...nojnhuh:cluster-api-provider-azure:aso-wired

- [ ] cherry-pick candidate

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/azure/services/aso/aso.go
+++ b/azure/services/aso/aso.go
@@ -77,6 +77,7 @@ func (s *Service) CreateOrUpdateResource(ctx context.Context, spec azure.ASOReso
 
 	log = log.WithValues("service", serviceName, "resource", resourceName, "namespace", resourceNamespace)
 
+	var adopt bool
 	var existing genruntime.MetaObject
 	if err := s.Client.Get(ctx, client.ObjectKeyFromObject(resource), resource); err != nil {
 		if !apierrors.IsNotFound(err) {
@@ -100,6 +101,19 @@ func (s *Service) CreateOrUpdateResource(ctx context.Context, spec azure.ASOReso
 		var readyErr error
 		if cond := conds[i]; cond.Status != metav1.ConditionTrue {
 			switch {
+			case cond.Reason == conditions.ReasonAzureResourceNotFound.Name &&
+				existing.GetAnnotations()[ReconcilePolicyAnnotation] == ReconcilePolicySkip:
+				// This resource was originally created by CAPZ and a
+				// corresponding Azure resource has been found not to exist, so
+				// CAPZ will tell ASO to adopt the resource by setting its
+				// reconcile policy to "manage". This extra step is necessary to
+				// handle user-managed resources that already exist in Azure and
+				// should not be reconciled by ASO while ensuring they're still
+				// represented in ASO.
+				log.V(2).Info("resource not found in Azure and \"skip\" reconcile-policy set, adopting")
+				// Don't set readyErr so the resource can be adopted with an
+				// update instead of returning early.
+				adopt = true
 			case cond.Reason == conditions.ReasonReconciling.Name:
 				readyErr = azure.NewOperationNotDoneError(&infrav1.Future{
 					Type:          createOrUpdateFutureType,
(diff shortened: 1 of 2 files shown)


azure/services/aso/aso.go
```diff
@@ -141,6 +148,9 @@ func (s *Service) CreateOrUpdateResource(ctx context.Context, spec azure.ASOReso
 
 	if existing == nil {
 		labels[infrav1.OwnedByClusterLabelKey] = s.clusterName
+		annotations[ReconcilePolicyAnnotation] = ReconcilePolicySkip
```
nojnhuh, line 151: One possible optimization that could be added is not to update ASO resources if `reconcile-policy: skip` is set since those changes aren't meant to propagate to the Azure resource anyway. I wonder if it might be valuable to do updates anyway though since that combined with the `status` would describe "what CAPZ wants the resource to look like" vs. "how it actually looks."

nojnhuh:
Also worth noting that this logic applies to all resource types, allowing users to BYO anything that CAPZ expects to exist. If there are reasons not to allow BYO for certain types, that enforcement would need to be added.

azure/services/aso/aso.go
```diff
@@ -100,6 +101,12 @@ func (s *Service) CreateOrUpdateResource(ctx context.Context, spec azure.ASOReso
 		var readyErr error
 		if cond := conds[i]; cond.Status != metav1.ConditionTrue {
 			switch {
+			case cond.Reason == conditions.ReasonAzureResourceNotFound.Name &&
```
jackfrancis, line 104: So this condition describes when we have a pre-existing ASO resource that doesn't actually have an Azure resource behind it?

nojnhuh: Yes, but only "pre-existing" in the sense that CAPZ created in the very near past. Pre-existing ASO resources not created by CAPZ should never reach here based on the `ownedByCluster` check above. In cases when the resource already exists in Azure but outside of CAPZ's ASO, those resources should never transition to this "AzureResourceNotFound" state.

azure/services/aso/aso.go
jackfrancis: So that I follow... in the nominal "new resource" case where we are creating a new ASO resource for the first time, we want to set the `"serviceoperator.azure.com/reconcile-policy"` annotation to `"skip"`? Why is that?

nojnhuh: Yep. This is to handle the "resource already exists in Azure but not managed by ASO" case, in which we assume the user is managing the resource themselves so ASO shouldn't make any changes to it. This is essentially to facilitate ASO doing its initial GET to Azure so we can see its status.

jackfrancis: Does this mean we are sort of "initializing" to skip and then will plan to reset in the next reconciliation when we realize that the resource doesn't in fact exist already in Azure?

nojnhuh: Yep, exactly. That happens here: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/45dc9456f266b920225b3682ac4ca0d2a20ed503/azure/services/aso/aso.go#L104-L109

jackfrancis: O.K. cool. Can we add code comments to describe what we're doing here and why? With these kinds of runloops the "init"/"base case" stuffs are often (IME) non-intuitive, especially when you're reading for the first time and sort of relying upon the literal execution loop and its conditional expressions for understanding "this because that". In fact it's more like "this right now because that later responds to this during the next loop" (sorry, I'm not making things clearer!)

nojnhuh: Just added a couple comments. LMK if that makes things clearer.

jackfrancis (approved):
/lgtm
/approve

PR #3653, opened 2023-06-20
fix AuthorizedIPRanges disable -> enable bug

nawazkh:
**What type of PR is this?**
/kind bug

- When user updated spec has `AuthorizedIPRanges:nil` (meaning, the user has removed the list of CIDRs), CAPZ will set an empty list, i.e. `AuthorizedIPRanges:&[]string`, at Azure SDK on the first reconcile to force Azure SDK in removing the list of CIDRs. 
- From the second and subsequent reconciles, CAPZ will set Azure SDK's `AuthorizedIPRanges:nil` matching the user updated spec

Fixes # https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/3654

- [x] cherry-pick candidate

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
authorized IP ranges will be set to []string then to nil upon deletion
```

### DIFF
--- a/azure/services/managedclusters/spec.go
+++ b/azure/services/managedclusters/spec.go
@@ -21,6 +21,7 @@ import (
 	"encoding/base64"
 	"fmt"
 	"net"
+	"reflect"
 	"time"
 
 	"github.com/Azure/azure-sdk-for-go/services/containerservice/mgmt/2022-03-01/containerservice"
@@ -387,7 +388,7 @@ func (s *ManagedClusterSpec) Parameters(ctx context.Context, existing interface{
 			EnablePrivateClusterPublicFQDN: s.APIServerAccessProfile.EnablePrivateClusterPublicFQDN,
 		}
 
-		if len(s.APIServerAccessProfile.AuthorizedIPRanges) > 0 {
+		if s.APIServerAccessProfile.AuthorizedIPRanges != nil {
 			managedCluster.APIServerAccessProfile.AuthorizedIPRanges = &s.APIServerAccessProfile.AuthorizedIPRanges
 		}
 	}
@@ -436,6 +437,16 @@ func (s *ManagedClusterSpec) Parameters(ctx context.Context, existing interface{
 		// Avoid changing agent pool profiles through AMCP and just use the existing agent pool profiles
 		// AgentPool changes are managed through AMMP.
 		managedCluster.AgentPoolProfiles = existingMC.AgentPoolProfiles
+
+		// if the AuthorizedIPRanges is nil in the user-updated spec, but not nil in the existing spec, then
+		// we need to set the AuthorizedIPRanges to empty array (&[]string{}) once so that the Azure API will
+		// update the existing authorized IP ranges to nil.
+		if !isAuthIPRangesNilOrEmpty(existingMC) && isAuthIPRangesNilOrEmpty(managedCluster) {
+			log.V(4).Info("managed cluster spec has nil AuthorizedIPRanges, updating existing authorized IP ranges to an empty list")
+			managedCluster.APIServerAccessProfile = &containerservice.ManagedClusterAPIServerAccessProfile{
+				AuthorizedIPRanges: &[]string{},
+			}
+		}
 
 		diff := computeDiffOfNormalizedClusters(managedCluster, existingMC)
 		if diff == "" {
@@ -671,4 +682,12 @@ func getIdentity(identity *infrav1.Identity) (managedClusterIdentity *containers
 		}
 	}
 	return
+}
+
+// isAuthIPRangesNilOrEmpty returns true if the managed cluster's APIServerAccessProfile or AuthorizedIPRanges is nil or if AuthorizedIPRanges is empty.
+func isAuthIPRangesNilOrEmpty(managedCluster containerservice.ManagedCluster) bool {
+	if managedCluster.APIServerAccessProfile == nil || managedCluster.APIServerAccessProfile.AuthorizedIPRanges == nil || reflect.DeepEqual(managedCluster.APIServerAccessProfile.AuthorizedIPRanges, &[]string{}) {
+		return true
+	}
+	return false
 }
--- a/azure/services/managedclusters/spec_test.go
+++ b/azure/services/managedclusters/spec_test.go
@@ -249,6 +249,67 @@ func TestParameters(t *testing.T) {
 				g.Expect(result).To(BeNil())
 			},
 		},
+		{
+			name:     "update authorized IP ranges with empty struct if spec does not have authorized IP ranges but existing cluster has authorized IP ranges",
+			existing: getExistingClusterWithAuthorizedIPRanges(),
+			spec: &ManagedClusterSpec{
+				Name:          "test-managedcluster",
+				ResourceGroup: "test-rg",
+				Location:      "test-location",
+				Tags: map[string]string{
+					"test-tag": "test-value",
+				},
+				Version:         "v1.22.0",
+				LoadBalancerSKU: "Standard",
+			},
+			expect: func(g *WithT, result interface{}) {
+				g.Expect(result).To(BeAssignableToTypeOf(containerservice.ManagedCluster{}))
+				g.Expect(result.(containerservice.ManagedCluster).APIServerAccessProfile).To(Not(BeNil()))
+				g.Expect(result.(containerservice.ManagedCluster).APIServerAccessProfile.AuthorizedIPRanges).To(Equal(&[]string{}))
+			},
+		},
+		{
+			name:     "update authorized IP ranges with authorized IPs spec has authorized IP ranges but existing cluster does not have authorized IP ranges",
+			existing: getExistingCluster(),
+			spec: &ManagedClusterSpec{
+				Name:          "test-managedcluster",
+				ResourceGroup: "test-rg",
+				Location:      "test-location",
+				Tags: map[string]string{
+					"test-tag": "test-value",
+				},
+				Version:         "v1.22.0",
+				LoadBalancerSKU: "Standard",
+				APIServerAccessProfile: &APIServerAccessProfile{
+					AuthorizedIPRanges: []string{"192.168.0.1/32, 192.168.0.2/32, 192.168.0.3/32"},
+				},
+			},
+			expect: func(g *WithT, result interface{}) {
+				g.Expect(result).To(BeAssignableToTypeOf(containerservice.ManagedCluster{}))
+				g.Expect(result.(containerservice.ManagedCluster).APIServerAccessProfile).To(Not(BeNil()))
+				g.Expect(result.(containerservice.ManagedCluster).APIServerAccessProfile.AuthorizedIPRanges).To(Equal(&[]string{"192.168.0.1/32, 192.168.0.2/32, 192.168.0.3/32"}))
+			},
+		},
+		{
+			name:     "no update needed when authorized IP ranges when both clusters have the same authorized IP ranges",
+			existing: getExistingClusterWithAuthorizedIPRanges(),
+			spec: &ManagedClusterSpec{
+				Name:          "test-managedcluster",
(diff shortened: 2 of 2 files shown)


nawazkh:
I tested this change locally and it works as expected. I could disable->enable _AuthorizedIPRanges_ by updating AzureManagedControlPlane's CRD.

nawazkh:
/retitle fix AuthorizedIPRanges disable -> enable bug

nawazkh:
My fix seems to be breaking this test
```
--- FAIL: TestParameters (0.00s)
    --- FAIL: TestParameters/managedcluster_does_not_exist (0.01s)
        spec_test.go:105: 
              containerservice.ManagedCluster{
              	... // 2 identical fields
              	ExtendedLocation: nil,
              	Identity:         &{Type: "SystemAssigned"},
              		... // 20 identical fields
              		AutoUpgradeProfile:     nil,
              		AutoScalerProfile:      nil,
            - 		APIServerAccessProfile: &containerservice.ManagedClusterAPIServerAccessProfile{AuthorizedIPRanges: &[]string{}},
            + 		APIServerAccessProfile: nil,
              		DiskEncryptionSetID:    nil,
              		IdentityProfile:        nil,
              		... // 5 identical fields
              	},
              	Tags:     {"Name": &"test-managedcluster", "sigs.k8s.io_cluster-api-provider-azure_cluster_test-cluster": &"owned", "sigs.k8s.io_cluster-api-provider-azure_role": &"common", "test-tag": &"test-value"},
              	Location: &"test-location",
              	... // 4 identical fields
              }
            
            Expected
                <bool>: false
            to be true
```
/hold

api/v1beta1/azuremanagedcontrolplane_types.go
```diff
@@ -266,7 +266,7 @@ type LoadBalancerProfile struct {
 type APIServerAccessProfile struct {
 	// AuthorizedIPRanges - Authorized IP Ranges to kubernetes API server.
 	// +optional
-	AuthorizedIPRanges []string `json:"authorizedIPRanges,omitempty"`
+	AuthorizedIPRanges *[]string `json:"authorizedIPRanges,omitempty"`
```
CecileRobertMichon, line 269: why use a pointer to a slice here? is an empty slice not enough info to tell us this should be empty?

CecileRobertMichon:
> Work in progress PR. Creating this as a reference for conversation.

@nawazkh please update the PR description

api/v1beta1/azuremanagedcontrolplane_types.go
nawazkh: While an empty slice gives us enough information about it being empty, changing `AuthorizedIPRanges` to a `*[]string` type makes it consistent with Azure SDK's `AuthorizedIPRanges` data type.

nawazkh: A minor improvement is that we no longer need to check the length of the array to determine if `AuthorizedIPRanges` is empty. But I don't think checking the length of `AuthorizedIPRanges` is really a big check.

azure/services/managedclusters/spec.go
```diff
@@ -165,7 +166,7 @@ type LoadBalancerProfile struct {
 // APIServerAccessProfile is the access profile for AKS API server.
 type APIServerAccessProfile struct {
 	// AuthorizedIPRanges are the authorized IP Ranges to kubernetes API server.
-	AuthorizedIPRanges []string
+	AuthorizedIPRanges *[]string
```
Jont828, line 169: Like Cecile said, could you explain why we're changing this into a `*[]string`?

azure/services/managedclusters/spec.go
```diff
@@ -437,6 +435,16 @@ func (s *ManagedClusterSpec) Parameters(ctx context.Context, existing interface{
...
 
+		// if the AuthorizedIPRanges is nil in the user-updated spec, but not nil in the existing spec, then
+		// we need to set the AuthorizedIPRanges to empty array (&[]string{}) once so that the Azure API will
+		// update the existing authorized IP ranges to nil.
+		if !apiServerAccessProfileOrAuthIPRangesIsNil(existingMC) && apiServerAccessProfileOrAuthIPRangesIsNil(managedCluster) {
+			log.V(4).Info("managed cluster spec has nil AuthorizedIPRanges, updating existing authorized IP ranges to empty")
+			managedCluster.APIServerAccessProfile = &containerservice.ManagedClusterAPIServerAccessProfile{
+				AuthorizedIPRanges: &[]string{},
+			}
+		}
```
Jont828, line 446: Heads up about test coverage, maybe add something in spec_test.go?

azure/services/managedclusters/spec.go
```diff
@@ -672,3 +680,11 @@ func getIdentity(identity *infrav1.Identity) (managedClusterIdentity *containers
 	}
 	return
 }
+
+// apiServerAccessProfileOrAuthIPRangesIsNil returns true if the managed cluster's APIServerAccessProfile or AuthorizedIPRanges is nil.
+func apiServerAccessProfileOrAuthIPRangesIsNil(managedCluster containerservice.ManagedCluster) bool {
+	if managedCluster.APIServerAccessProfile == nil || managedCluster.APIServerAccessProfile.AuthorizedIPRanges == nil || reflect.DeepEqual(managedCluster.APIServerAccessProfile.AuthorizedIPRanges, &[]string{}) {
+		return true
+	}
+	return false
```
Jont828, line 689: Do you think we should add a small unit test for this function so we can test the `return false` case?

Jont828 (commented):
Nice work! Just a couple small points

azure/services/managedclusters/spec.go
```diff
@@ -387,7 +388,7 @@ func (s *ManagedClusterSpec) Parameters(ctx context.Context, existing interface{
 			EnablePrivateClusterPublicFQDN: s.APIServerAccessProfile.EnablePrivateClusterPublicFQDN,
 		}
 
-		if len(s.APIServerAccessProfile.AuthorizedIPRanges) > 0 {
+		if s.APIServerAccessProfile.AuthorizedIPRanges != nil {
```
nawazkh, line 391: Not checking for length here since empty array can still be valid.

api/v1beta1/azuremanagedcontrolplane_types.go
nawazkh: Updated the data type to `[]string` instead of a pointer.

azure/services/managedclusters/spec.go
nawazkh: The original reason for this change was for 
1. consistency with Azure SDK
2. CAPZ can differentiate between unset field vs empty array. 

But since a slice also has a nil type, I have reverted this to `[]string` type instead of a pointer.

azure/services/managedclusters/spec.go
nawazkh: I have added more tests to spec_test.go covering scenarios testing this scenario. Wdyt?

azure/services/managedclusters/spec.go
nawazkh: It could be redundant in adding a test case for this specific function since I am already testing the response of this function in the scenarios I added in spec_test.go. 
What do you think ?

api/v1beta1/azuremanagedcontrolplane_webhook.go
```diff
@@ -369,13 +369,15 @@ func (m *AzureManagedControlPlane) validateLoadBalancerProfile(_ client.Client)
 func (m *AzureManagedControlPlane) validateAPIServerAccessProfile(_ client.Client) error {
 	if m.Spec.APIServerAccessProfile != nil {
 		var allErrs field.ErrorList
-		for _, ipRange := range m.Spec.APIServerAccessProfile.AuthorizedIPRanges {
-			if _, _, err := net.ParseCIDR(ipRange); err != nil {
-				allErrs = append(allErrs, field.Invalid(field.NewPath("Spec", "APIServerAccessProfile", "AuthorizedIPRanges"), ipRange, "invalid CIDR format"))
+		if m.Spec.APIServerAccessProfile.AuthorizedIPRanges != nil {
```
CecileRobertMichon, line 372: this is redundant, the whole for loop will already get skipped if the array is nil

https://play.golang.com/p/xycugWOnMXp

azure/services/managedclusters/spec.go
```diff
@@ -672,3 +683,11 @@ func getIdentity(identity *infrav1.Identity) (managedClusterIdentity *containers
 	}
 	return
 }
+
+// apiServerAccessProfileOrAuthIPRangesIsNil returns true if the managed cluster's APIServerAccessProfile or AuthorizedIPRanges is nil.
+func apiServerAccessProfileOrAuthIPRangesIsNil(managedCluster containerservice.ManagedCluster) bool {
+	if managedCluster.APIServerAccessProfile == nil || managedCluster.APIServerAccessProfile.AuthorizedIPRanges == nil || reflect.DeepEqual(managedCluster.APIServerAccessProfile.AuthorizedIPRanges, &[]string{}) {
```
CecileRobertMichon, line 689: Not sure I follow why `reflect.DeepEqual(managedCluster.APIServerAccessProfile.AuthorizedIPRanges, &[]string{}` is considered "nil"

api/v1beta1/azuremanagedcontrolplane_webhook.go
nawazkh: True that. Updating.

azure/services/managedclusters/spec.go
nawazkh: I could update the function name to `apiServerAccessProfileOrAuthIPRangesIsNilOrEmpty` instead. Would that be more apt?

api/v1beta1/azuremanagedcontrolplane_webhook.go
nawazkh: Updated.

azure/services/managedclusters/spec.go
nawazkh: Updated to `apiServerAccessProfileOrAuthIPRangesIsNilOrEmpty` for more clarity.

azure/services/managedclusters/spec.go
Jont828: Oh, I see. I just pointed it out b/c there was a note from codecov saying it wasn't tested, but if you've already covered it, then we're good.

azure/services/managedclusters/spec.go
Jont828: Looks good!

CecileRobertMichon:
Release note should be set since this is a bug fix

azure/services/managedclusters/spec.go
CecileRobertMichon: What do you think of renaming it to `AuthIPRangesIsNilOrEmpty` so it's not as lengthy? I think it's implied that if the `APIServerAccessProfile` is nil then `AuthorizedIPRanges` must be nil as well

nawazkh: Yep, that works too!

CecileRobertMichon (commented):
/lgtm

CecileRobertMichon:
/approve

should this be cherry picked @nawazkh ?

nawazkh:
> /approve
> 
> 
> 
> should this be cherry picked @nawazkh ?

Yes, this should be cherry picked.

nawazkh:
I checked ✅ the cherry-pick candidate option. Not sure it will cherry-pick automatically since the PR has been approved already. 
If it doesn't get cherry-picked, I will add the cherry-pick commands later.

nawazkh:
- Created https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3806 for release-1.9
- Created https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3805 for release-1.10

Issue #3070, opened 2023-01-21
SpotVMOptions with AzureManagedMachinePool Supported?


lmcdasm:
/kind bug

**Describe the solution you'd like**
in the SPOT options, you can provide spotVMOptions for AzureMachinePool for the VMs, however when we use AzureManagedMachinePool there doesnt seem to be support for that option (or perhaps im looking in the wrong place)?  - marked it below with ## THIS IS ## 

error when kubectl apply the manifest 
<img width="1780" alt="image" src="https://user-images.githubusercontent.com/9358276/213840123-46a8ac10-47a8-4d9b-b5ce-888924c80bae.png">

When creating an AKS (from say az cli or via the UI) you can use the Spot option for VMs backing the User Pools (not so for Control plane - for obvious reasons).  

Here is an example of a AKS User Pool that is backed by Spot instances  

<img width="1220" alt="image" src="https://user-images.githubusercontent.com/9358276/213839964-2971290c-5827-460e-883d-d09dd0c8c762.png">

here is the Manifest that im using (generated pretty much from the tutorial/docs provided) 

```
apiVersion: cluster.x-k8s.io/v1beta1
kind: Cluster
metadata:
  name: azure-managed-gen-ad3941-clus
  namespace: default
spec:
  clusterNetwork:
    services:
      cidrBlocks:
      - 192.168.0.0/16
  controlPlaneRef:
    apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
    kind: AzureManagedControlPlane
    name: azure-managed-gen-ad3941-clus
  infrastructureRef:
    apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
    kind: AzureManagedCluster
    name: azure-managed-gen-ad3941-clus
---
apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
kind: AzureManagedControlPlane
metadata:
  name: azure-managed-gen-ad3941-clus
  namespace: default
spec:
  identityRef:
    apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
    kind: AzureClusterIdentity
    name: cluster-identity
  location: eastus2
  resourceGroupName: azure-managed-gen-ad3941-clus
  sshPublicKey: ""
  subscriptionID: 4cebd231-affb-455b-9e26-dfd18bb64df3
  version: v1.24.6
---
apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
[...]

PR #3672, opened 2023-06-27
Allow setting ScaleDownMode and SpotMaxPrice for AzureManagedMachinePool

maciaszczykm:
**What type of PR is this?**
/kind feature

Related to https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/3070.

- [ ] cherry-pick candidate

**TODOs**:
- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:
```release-note
Allow setting ScaleDownMode and SpotMaxPrice for AzureManagedMachinePool
```


maciaszczykm:
Should I change `float` to `string`? What would be your preference?

maciaszczykm:
@CecileRobertMichon E2E failure might just be flaky test. I will rerun it a few times as I saw similar thing in my other PRs.

maciaszczykm:
cc @nojnhuh @CecileRobertMichon can you take a look on it?

api/v1beta1/azuremanagedmachinepool_types.go
```diff
@@ -520,6 +521,17 @@ type AzureManagedMachinePoolSpec struct {
...
+	// ScaleDownMode affects the cluster autoscaler behavior. Default to Delete. Possible values include: 'Deallocate', 'Delete'
+	// +kubebuilder:validation:Enum=Deallocate;Delete
+	// +kubebuilder:default=Delete
+	// +optional
+	ScaleDownMode *string `json:"scaleDownMode,omitempty"`
+
+	// SpotMaxPrice defines max price to pay for spot instance.
+	// Possible values are any decimal value greater than zero or -1 which indicates the willingness to pay any on-demand price.
+	// +optional
+	SpotMaxPrice *resource.Quantity `json:"spotMaxPrice,omitempty"`
```
nawazkh, line 533: Can we get away with using `*float64` instead of using `*resource.Quantity` ?
Just a thought, not that I am against using `*resource.Quantity`. My question is rather a self clarifying thought to learn more about this decision.

azure/services/agentpools/spec_test.go
```diff
@@ -215,6 +235,39 @@ func TestParameters(t *testing.T) {
...
+				withSpotMaxPrice("789.123"),
+			),
+			existing: sdkFakeAgentPool(
+				sdkWithProvisioningState("Succeeded"),
+			),
+			expected: sdkFakeAgentPool(
+				sdkWithSpotMaxPrice(789.123),
+			),
+			expectedError: nil,
+		},
```
nawazkh, line 270: Can we have some test cases supporting _five decimal places_ ?

nawazkh (commented):
This looks great, Thank you @maciaszczykm !!
I just have some questions, would like to get your thoughts on it.

api/v1beta1/azuremanagedmachinepool_types.go
CecileRobertMichon: API conventions recommend against using floats due to risk during conversion https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types

> Avoid floating-point values as much as possible, and never use them in spec. Floating-point values cannot be reliably round-tripped (encoded and re-decoded) without changing, and have varying precision and representations across languages and architectures.

azure/services/agentpools/spec.go
```diff
@@ -364,6 +379,10 @@ func (s *AgentPoolSpec) Parameters(ctx context.Context, existing interface{}) (p
 		},
 	}
 
+	if s.SpotMaxPrice != nil {
```
CecileRobertMichon, line 382: Let's move this above where the other optional configs get set (see `sku` for example). 

Also, should we do the same thing for `ScaleDownMode`? since they are both pointers

CecileRobertMichon (commented):
lgtm overall

I wonder if there's a way we could test this in e2e tests? I'm thinking we could add a spot pool to the aks test cluster https://github.com/kubernetes-sigs/cluster-api-provider-azure/tree/main/templates/test/ci/prow-aks/patches, although I could see good reasons for not doing that if the spot instances introduce flakiness in the tests (eg. they get deallocated during the test). Maybe we write a special test case that creates then deletes the spot pool? @nojnhuh @nawazkh thoughts?

api/v1beta1/azuremanagedmachinepool_types.go
nawazkh: Ah sweet, thanks for pinning it!

api/v1beta1/azuremanagedmachinepool_types.go
```diff
@@ -520,6 +521,17 @@ type AzureManagedMachinePoolSpec struct {
...
 
+	// ScaleDownMode affects the cluster autoscaler behavior. Default to Delete. Possible values include: 'Deallocate', 'Delete'
+	// +kubebuilder:validation:Enum=Deallocate;Delete
+	// +kubebuilder:default=Delete
+	// +optional
+	ScaleDownMode *string `json:"scaleDownMode,omitempty"`
+
+	// SpotMaxPrice defines max price to pay for spot instance.
+	// Possible values are any decimal value greater than zero or -1 which indicates the willingness to pay any on-demand price.
+	// +optional
```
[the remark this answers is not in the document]
maciaszczykm: I have updated it.

azure/services/agentpools/spec_test.go
maciaszczykm: Sure, done.

maciaszczykm:
@nawazkh @CecileRobertMichon Thank you for reviews. I have applied changes. PTAL.

I'm not really familiar with CAPZ E2E tests yet, but if you think that is required then I will try to add them. I'll wait for confirmation.

nawazkh:
> Maybe we write a special test case that creates then deletes the spot pool?

I agree. It is a good idea to add an e2e test to claim that this use case works.

IMO, I would probably make the new test "optional" and test the node allocation at a specific location (or a bunch of locations) to minimize flakiness. 

There are a bunch of `OPTIONAL` tests in [azure_test.go](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/test/e2e/azure_test.go) that can be used as an template to test ScaleDownMode and SpotMaxPrice scenario.

CecileRobertMichon:
> IMO, I would probably make the new test "optional" and test the node allocation at a specific location (or a bunch of locations) to minimize flakiness.

I don't think we should add whole new cluster test for this, it can be simply added to the AKS test itself so we don't have to build a separate cluster. I don't know whether a `By` spec can be marked as optional, if so we can do that.

@maciaszczykm you can use the `AKSPublicIPPrefixSpec` as an example of a test that tests a specific nodepool configuration with AKS https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/test/e2e/azure_test.go#L775C5-L775C26

test/e2e/aks_spot.go
```diff
@@ -0,0 +1,129 @@
+//go:build e2e
+// +build e2e
+
+/*
+Copyright 2022 The Kubernetes Authors.
```
CecileRobertMichon, line 5:
```suggestion
Copyright 2023 The Kubernetes Authors.
```

maciaszczykm:
PTAL

maciaszczykm:
@CecileRobertMichon Done, PTAL.

mboersma (approved):
/lgtm
/approve

PR #3676, opened 2023-06-29
Fix custom backendPool not being used

RadekManak:
**What type of PR is this?**
/kind bug

In some places the default backendPool name was being used instead of using the specified value. This caused machine creation to fail when attempting to use loadBalancer with custom backendPool name.

Fixes #

- [ ] cherry-pick candidate

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Fix default `backendPool` name being used when custom name is set.
```

### DIFF
--- a/azure/scope/cluster.go
+++ b/azure/scope/cluster.go
@@ -711,33 +711,37 @@ func (s *ClusterScope) GetPrivateDNSZoneName() string {
 }
 
 // APIServerLBPoolName returns the API Server LB backend pool name.
-func (s *ClusterScope) APIServerLBPoolName(loadBalancerName string) string {
-	return azure.GenerateBackendAddressPoolName(loadBalancerName)
+func (s *ClusterScope) APIServerLBPoolName() string {
+	return s.APIServerLB().BackendPool.Name
+}
+
+// OutboundLB returns the outbound LB.
+func (s *ClusterScope) outboundLB(role string) *infrav1.LoadBalancerSpec {
+	if role == infrav1.Node {
+		return s.NodeOutboundLB()
+	}
+	if s.IsAPIServerPrivate() {
+		return s.ControlPlaneOutboundLB()
+	}
+	return s.APIServerLB()
 }
 
 // OutboundLBName returns the name of the outbound LB.
 func (s *ClusterScope) OutboundLBName(role string) string {
-	if role == infrav1.Node {
-		if s.NodeOutboundLB() == nil {
-			return ""
-		}
-		return s.NodeOutboundLB().Name
+	lb := s.outboundLB(role)
+	if lb == nil {
+		return ""
 	}
-	if s.IsAPIServerPrivate() {
-		if s.ControlPlaneOutboundLB() == nil {
-			return ""
-		}
-		return s.ControlPlaneOutboundLB().Name
-	}
-	return s.APIServerLBName()
+	return lb.Name
 }
 
 // OutboundPoolName returns the outbound LB backend pool name.
-func (s *ClusterScope) OutboundPoolName(loadBalancerName string) string {
-	if loadBalancerName == "" {
+func (s *ClusterScope) OutboundPoolName(role string) string {
+	lb := s.outboundLB(role)
+	if lb == nil {
 		return ""
 	}
-	return azure.GenerateOutboundBackendAddressPoolName(loadBalancerName)
+	return lb.BackendPool.Name
 }
 
 // ResourceGroup returns the cluster resource group.
--- a/azure/scope/machine.go
+++ b/azure/scope/machine.go
@@ -277,13 +277,13 @@ func (m *MachineScope) BuildNICSpec(nicName string, infrav1NetworkInterface infr
 
 		if m.Role() == infrav1.ControlPlane {
 			spec.PublicLBName = m.OutboundLBName(m.Role())
-			spec.PublicLBAddressPoolName = m.OutboundPoolName(m.OutboundLBName(m.Role()))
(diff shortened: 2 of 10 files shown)


azure/scope/cluster.go
```diff
@@ -711,33 +711,37 @@ func (s *ClusterScope) GetPrivateDNSZoneName() string {
 }
 
 // APIServerLBPoolName returns the API Server LB backend pool name.
-func (s *ClusterScope) APIServerLBPoolName(loadBalancerName string) string {
-	return azure.GenerateBackendAddressPoolName(loadBalancerName)
+func (s *ClusterScope) APIServerLBPoolName() string {
+	return s.APIServerLB().BackendPool.Name
```
nawazkh, line 715: Is it possible that `s.APIServerLB()` to be `nil`? If so, can we run into _nil pointer dereference_ ?

nawazkh: Also, do you think it is valid to use `azure.GenerateBackendAddressPoolName(loadBalancerName)` as a backup name if `s.APIServerLB().BackendPool.Name` is nil?

azure/scope/cluster.go
```diff
@@ -711,33 +711,37 @@ func (s *ClusterScope) GetPrivateDNSZoneName() string {
...
 }
 
-// OutboundLBName returns the name of the outbound LB.
-func (s *ClusterScope) OutboundLBName(role string) string {
+// OutboundLB returns the outbound LB.
+func (s *ClusterScope) OutboundLB(role string) *infrav1.LoadBalancerSpec {
```
nawazkh, line 719: - Are we using `OutboundLB()` anywhere else outside this package? Any specific reason on exporting it out?

nawazkh: - If it's just the `lb.Name` we are interested in, can we just return the name instead of the whole spec?

azure/scope/machine.go
```diff
@@ -293,12 +293,7 @@ func (m *MachineScope) BuildNICSpec(nicName string, infrav1NetworkInterface infr
 		// If the NAT gateway is not enabled and node has no public IP, then the NIC needs to reference the LB to get outbound traffic.
 		if m.Role() == infrav1.Node && !m.Subnet().IsNatGatewayEnabled() && !m.AzureMachine.Spec.AllocatePublicIP {
 			spec.PublicLBName = m.OutboundLBName(m.Role())
-			spec.PublicLBAddressPoolName = m.OutboundPoolName(m.OutboundLBName(m.Role()))
-		}
-		// If the NAT gateway is not enabled and node has no public IP, then the NIC needs to reference the LB to get outbound traffic.
-		if m.Role() == infrav1.Node && !m.Subnet().IsNatGatewayEnabled() && !m.AzureMachine.Spec.AllocatePublicIP {
-			spec.PublicLBName = m.OutboundLBName(m.Role())
-			spec.PublicLBAddressPoolName = m.OutboundPoolName(m.OutboundLBName(m.Role()))
```
nawazkh, line 301: - Are we planning on putting this back ? Seems like an accidental delete?

nawazkh (commented):
I didn't review the changes on `*_mock.go`, putting it to later since I had some implementation questions.

azure/scope/cluster.go
RadekManak: > Is it possible that s.APIServerLB() to be nil? If so, can we run into nil pointer dereference ?

It cannot be nil. A function above is creating the pointer from a struct field.
// APIServerLB returns the cluster API Server load balancer.
func (s *ClusterScope) APIServerLB() *infrav1.LoadBalancerSpec {
	return &s.AzureCluster.Spec.NetworkSpec.APIServerLB
}

`azure.GenerateBackendAddressPoolName(loadBalancerName)` is a valid default name for the backendPool, but it is already being set by the defaulting webhook if it is empty.

azure/scope/cluster.go
RadekManak: You're right should be private. I have decomposed the load balancer selection logic into a function because it is needed in both OutboundLBName and OutboundPoolName.

azure/scope/machine.go
RadekManak: I am pretty sure that these 4 lines were duplicate and removing it once does not change the behavior.

CecileRobertMichon (commented):
/ok-to-test

azure/scope/machine.go
CecileRobertMichon: good catch

nawazkh:
_pull-cluster-api-provider-azure-apidiff_ test seems to be failing since we are changing the interface in here. Does it classify as a breaking change? Do we need to take care of any more stuff? Maybe call out in the Release Notes?

nawazkh:
Looks good to me. 
Just waiting on below.

> _pull-cluster-api-provider-azure-apidiff_ test seems to be failing since we are changing the interface in here. Does it classify as a breaking change? Do we need to take care of any more stuff? Maybe call out in the Release Notes?

/lgtm

CecileRobertMichon:
> pull-cluster-api-provider-azure-apidiff test seems to be failing since we are changing the interface in here. Does it classify as a breaking change? Do we need to take care of any more stuff? Maybe call out in the Release Notes?

thanks for calling that out. In this case I think we're fine since it's only touching the scope functions which we don't make guarantees on (breaking changes in the user-facing API is what we want to watch out for). That being said, we should really move code that isn't meant to be consumed as a library to an internal/ directory to avoid these false positives.

CecileRobertMichon:
/approve
/cherry-pick release-1.9

Let's cherry-pick this one since it fixes a bug ^

PR #3677, opened 2023-06-29
Allow skipping Linux profile on managed clusters

maciaszczykm:
**What type of PR is this?**
/kind feature

In order to be able to not set `LinuxProfile` in AKS I have changed `AzureManagedControlPlane` `SSHPublicKey` to be optional. If it is null then new key will not be autogenerated and `LinuxProfile` will not be set, if it is set to empty string then new key will be generated just like it was done until now.

See Slack thread: https://kubernetes.slack.com/archives/CEX9HENG7/p1687961825757119

- [ ] cherry-pick candidate

**TODOs**:
- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:
```release-note
Allow skipping Linux profile on managed clusters
```

### DIFF
--- a/api/v1beta1/azuremanagedcontrolplane_types.go
+++ b/api/v1beta1/azuremanagedcontrolplane_types.go
@@ -124,8 +124,10 @@ type AzureManagedControlPlaneSpec struct {
 	OutboundType *ManagedControlPlaneOutboundType `json:"outboundType,omitempty"`
 
 	// SSHPublicKey is a string literal containing an ssh public key base64 encoded.
+	// Use empty string to autogenerate new key. Use null value to not set key.
 	// Immutable.
-	SSHPublicKey string `json:"sshPublicKey"`
+	// +optional
+	SSHPublicKey *string `json:"sshPublicKey,omitempty"`
 
 	// DNSServiceIP is an IP address assigned to the Kubernetes DNS service.
 	// It must be within the Kubernetes service address range specified in serviceCidr.
--- a/azure/services/managedclusters/spec_test.go
+++ b/azure/services/managedclusters/spec_test.go
@@ -18,6 +18,7 @@ package managedclusters
 
 import (
 	"context"
+	"encoding/base64"
 	"testing"
 
 	"github.com/Azure/azure-sdk-for-go/services/containerservice/mgmt/2022-03-01/containerservice"
@@ -69,6 +70,7 @@ func TestParameters(t *testing.T) {
 				},
 				Version:         "v1.22.0",
 				LoadBalancerSKU: "Standard",
+				SSHPublicKey:    base64.StdEncoding.EncodeToString([]byte("test-ssh-key")),
 				GetAllAgentPools: func() ([]azure.ResourceSpecGetter, error) {
 					return []azure.ResourceSpecGetter{
 						&agentpools.AgentPoolSpec{
@@ -161,6 +163,67 @@ func TestParameters(t *testing.T) {
 				// Additional tags are handled by azure/services/tags, so a diff
 				// here shouldn't trigger an update on the managed cluster resource.
 				g.Expect(result).To(BeNil())
+			},
+		},
+		{
+			name:     "set Linux profile if SSH key is set",
+			existing: nil,
+			spec: &ManagedClusterSpec{
+				Name:            "test-managedcluster",
+				ResourceGroup:   "test-rg",
+				Location:        "test-location",
+				Tags:            nil,
+				Version:         "v1.22.0",
+				LoadBalancerSKU: "Standard",
+				SSHPublicKey:    base64.StdEncoding.EncodeToString([]byte("test-ssh-key")),
+				GetAllAgentPools: func() ([]azure.ResourceSpecGetter, error) {
+					return []azure.ResourceSpecGetter{
+						&agentpools.AgentPoolSpec{
+							Name:          "test-agentpool-0",
+							Mode:          string(infrav1.NodePoolModeSystem),
+							ResourceGroup: "test-rg",
+							Replicas:      int32(2),
+							AdditionalTags: map[string]string{
+								"test-tag": "test-value",
+							},
+						},
+					}, nil
+				},
+			},
+			expect: func(g *WithT, result interface{}) {
+				g.Expect(result).To(BeAssignableToTypeOf(containerservice.ManagedCluster{}))
+				g.Expect(result.(containerservice.ManagedCluster).LinuxProfile).To(Not(BeNil()))
+				g.Expect(*(*result.(containerservice.ManagedCluster).LinuxProfile.SSH.PublicKeys)[0].KeyData).To(Equal("test-ssh-key"))
+			},
+		},
+		{
+			name:     "skip Linux profile if SSH key is not set",
+			existing: nil,
+			spec: &ManagedClusterSpec{
+				Name:            "test-managedcluster",
+				ResourceGroup:   "test-rg",
+				Location:        "test-location",
+				Tags:            nil,
+				Version:         "v1.22.0",
+				LoadBalancerSKU: "Standard",
+				SSHPublicKey:    "",
+				GetAllAgentPools: func() ([]azure.ResourceSpecGetter, error) {
+					return []azure.ResourceSpecGetter{
+						&agentpools.AgentPoolSpec{
+							Name:          "test-agentpool-0",
+							Mode:          string(infrav1.NodePoolModeSystem),
+							ResourceGroup: "test-rg",
+							Replicas:      int32(2),
+							AdditionalTags: map[string]string{
+								"test-tag": "test-value",
+							},
+						},
+					}, nil
+				},
+			},
+			expect: func(g *WithT, result interface{}) {
+				g.Expect(result).To(BeAssignableToTypeOf(containerservice.ManagedCluster{}))
+				g.Expect(result.(containerservice.ManagedCluster).LinuxProfile).To(BeNil())
 			},
 		},
 		{
@@ -307,7 +370,7 @@ func getSampleManagedCluster() containerservice.ManagedCluster {
 				SSH: &containerservice.SSHConfiguration{
 					PublicKeys: &[]containerservice.SSHPublicKey{
 						{
-							KeyData: pointer.String(""),
+							KeyData: pointer.String("test-ssh-key"),
 						},
 					},
 				},
@@ -316,7 +379,7 @@ func getSampleManagedCluster() containerservice.ManagedCluster {
 			NodeResourceGroup:       pointer.String("test-node-rg"),
 			EnableRBAC:              pointer.Bool(true),
 			NetworkProfile: &containerservice.NetworkProfile{
-				LoadBalancerSku: containerservice.LoadBalancerSku("Standard"),
+				LoadBalancerSku: "Standard",
 			},
 		},
 		Identity: &containerservice.ManagedClusterIdentity{
--- a/api/v1beta1/azuremanagedcontrolplane_default.go
+++ b/api/v1beta1/azuremanagedcontrolplane_default.go
@@ -34,13 +34,13 @@ const (
 
 // setDefaultSSHPublicKey sets the default SSHPublicKey for an AzureManagedControlPlane.
 func (m *AzureManagedControlPlane) setDefaultSSHPublicKey() error {
-	if sshKeyData := m.Spec.SSHPublicKey; sshKeyData == "" {
+	if sshKey := m.Spec.SSHPublicKey; sshKey != nil && *sshKey == "" {
 		_, publicRsaKey, err := utilSSH.GenerateSSHKey()
 		if err != nil {
 			return err
 		}
 
-		m.Spec.SSHPublicKey = base64.StdEncoding.EncodeToString(ssh.MarshalAuthorizedKey(publicRsaKey))
+		m.Spec.SSHPublicKey = pointer.String(base64.StdEncoding.EncodeToString(ssh.MarshalAuthorizedKey(publicRsaKey)))
 	}
 
 	return nil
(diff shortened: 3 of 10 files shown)


maciaszczykm:
What do you think about this approach?

I was also considering adding another field like `AutogenerateSSHKey` that would be true by default, but once disabled then SSH key would not be autogenerated. Then we would not be changing `PublicSSHKey` type but adding new field instead.

cc @jackfrancis

api/v1beta1/azuremanagedcontrolplane_types.go
```diff
@@ -110,8 +110,10 @@ type AzureManagedControlPlaneSpec struct {
 	OutboundType *ManagedControlPlaneOutboundType `json:"outboundType,omitempty"`
 
 	// SSHPublicKey is a string literal containing an ssh public key base64 encoded.
+	// Use empty string to autogenerate new key. Use null value to not set key.
 	// Immutable.
-	SSHPublicKey string `json:"sshPublicKey"`
+	// +optional
```
jackfrancis, line 115: This is technically a breaking change, though I'll defer to @CecileRobertMichon and @nojnhuh and others to comment on the practical cost of evolving from "required" to "optional" technically this evolution won't break existing user templates.

api/v1beta1/azuremanagedcontrolplane_types.go
```diff
@@ -110,8 +110,10 @@ type AzureManagedControlPlaneSpec struct {
...
+	SSHPublicKey *string `json:"sshPublicKey"`
```
DavidSpek, line 116: Shouldn't this be `json:"sshPublicKey,omitempty"`?

maciaszczykm: Good catch, it should be.

.golangci.yml
```diff
@@ -99,6 +99,8 @@ linters-settings:
   gocritic:
     enabled-tags:
       - "experimental"
+  gocyclo:
+    min-complexity: 31
```
nojnhuh, line 103: Not relevant to this PR, but I'd be +1 to skip running gocyclo entirely since these kinds of lint errors are more annoying than helpful IMO if they only motivate us to keep raising the ceiling like this.

maciaszczykm: I agree, if you want I can disable it. I can also export code to function and keep it at 30 if you wish.

nojnhuh: I think this minimally-impacting change here is fine as-is for this PR. Disabling that linter entirely may warrant wider discussion that might fit best in a separate PR.

CecileRobertMichon: > "Function too complex" in general I think is subjective enough that human eyes seem like a better tool.

What specific function is causing the error here? I agree that linters are not always right, but that doesn't mean they're not helpful. Sometimes it's a useful nudge to reduce tech debt. For each case that the linter complains about we (PR author and reviewers) should carefully consider the function causing the issue and assess whether:
1) the linter is right, the function _is_ getting too complex: refactor it into smaller, testable functions  
2) the linter missed something, there is a good reason for the function length/complexity: add a `//nolint` [comment](https://golangci-lint.run/usage/false-positives/#nolint-directive) to the function that disables the linter for that specific function with a justification

I don't think raising the overall ceiling to "fix" the issue is the right approach. The fact that we're having to add this new configuration in `.golangci.yml` means we've never lifted the ceiling in the several years since the linter was introduced (that's a good thing). Introducing a specific configuration now would be a regression IMO. 

I'm also -1 on disabling the linter entirely in case that wasn't clear, but that should definitely be it's own discussion :)

maciaszczykm:
Is there anything that I can do to push it forward?

/retest

api/v1beta1/azuremanagedcontrolplane_types.go
CecileRobertMichon: I'm not too worried about the `required` -> `optional` change (although this is technically breaking for round-trip conversion) for the same reasons @nojnhuh mentions above, however, are we concerned about the type change (from `string` to `*string`)? Not sure if that would cause upgrade issues. @maciaszczykm have you been able to validate upgrade?

CecileRobertMichon:
> Is there anything that I can do to push it forward?

@maciaszczykm I think the remaining open items are:

1. the gocyclo linter bump discussion
2. the backwards compatibility of the pointer change
3. the aks e2e test is failing 

Once all of this is resolved, please squash commits

CecileRobertMichon:
Also going to start an apiversion upgrade test:

maciaszczykm:
@CecileRobertMichon E2E failure might just be flaky test. I will rerun it a few times as I saw similar thing in my other PRs.

api/v1beta1/azuremanagedcontrolplane_types.go
maciaszczykm: @CecileRobertMichon I haven't done image upgrade on running CAPZ, but the tests that you have started have passed.

.golangci.yml
maciaszczykm: > What specific function is causing the error here? 

It's `Parameters` function, where all fields are copied. The ones with pointers require additional checks that sum up.

If you want I can disable linting for this function or extract code to separate method (but then it's a bit inconsistent with the rest of whole function).

maciaszczykm:
@CecileRobertMichon E2E passed.

api/v1beta1/azuremanagedcontrolplane_types.go
nojnhuh: I think w.r.t. the JSON API, the `string` -> `*string` change is backwards compatible since specifying an empty ssh key behaves the same way and not specifying one at all wasn't possible before. This is a breaking change for consumers of the Go types though, but I'm not sure how strict we've been with that in the past.

.golangci.yml
nojnhuh: I'm +1 to ignoring gocyclo for this particular function or raising the ceiling in a more localized way if that's possible. Extracting code from there into separate functions hasn't reduced complexity or increased reasability/testability IMO.

The way this function is structured currently inherently requires lots of `if <field is nil> { <set field> }` blocks that count against gocyclo but don't seem to me like spaghetti code like I think the linter is trying to point out.

CecileRobertMichon: Let's disable gocyclo for this particular function 👍

api/v1beta1/azuremanagedcontrolplane_types.go
CecileRobertMichon: Sounds good

.golangci.yml
maciaszczykm: Done.

maciaszczykm:
I have disabled gocyclo for `Parameters` function only and reverted linter config to its original state. Commits are squashed as well. PTAL.

CecileRobertMichon (changes requested):
`azure/services/.DS_Store` looks like it was checked in by accident

maciaszczykm:
@CecileRobertMichon Fixed, PTAL.

api/v1beta1/azuremachine_validation_test.go
```diff
@@ -68,13 +68,13 @@ func TestAzureMachine_ValidateSSHKey(t *testing.T) {
...
-func generateSSHPublicKey(b64Enconded bool) string {
+func generateSSHPublicKey(b64Enconded bool) *string {
 	privateKey, _ := rsa.GenerateKey(rand.Reader, 2048)
 	publicRsaKey, _ := ssh.NewPublicKey(&privateKey.PublicKey)
 	if b64Enconded {
-		return base64.StdEncoding.EncodeToString(ssh.MarshalAuthorizedKey(publicRsaKey))
+		return pointer.String(base64.StdEncoding.EncodeToString(ssh.MarshalAuthorizedKey(publicRsaKey)))
 	}
-	return string(ssh.MarshalAuthorizedKey(publicRsaKey))
+	return pointer.String(string(ssh.MarshalAuthorizedKey(publicRsaKey)))
```
nawazkh, line 77: Can we instead return the string from the `generateSSHPublicKey` function so that the return variable doesn't _"escape"_ the function?
We can convert the returned string to a pointer as needed.

nawazkh (commented):
Minor nit overall looks good to me. Thank you @maciaszczykm for putting together this PR!

nawazkh:
One more question, the implemented UX of this PR looks a little cramped to me(although completely justified). Do you think we could add a new field `selfGenerateSSHKey bool` (or something analogous to it) to generate a new SSHKey instead of relying on different behaviors of `SSHPublicKey = nil` and `SSHPublicKey = ` ?

Its just a thought but I am completely ok with this current implementation as well.

api/v1beta1/azuremachine_validation_test.go
maciaszczykm: Sure, it should be done like you said.

maciaszczykm:
@nawazkh I considered adding new field in [my comment above](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3677#issuecomment-1612698386):

> I was also considering adding another field like AutogenerateSSHKey that would be true by default, but once disabled then SSH key would not be autogenerated. Then we would not be changing PublicSSHKey type but adding new field instead.

But after thinking longer about it, all fields that are part of control plane spec map somehow to AKS fields. This would be used only by CAPZ and it would be used only once during init, after that it would be redundant. I prefer current solution but if you would like then I can change it. Both ways should work.

api/v1beta1/azuremachine_validation_test.go
maciaszczykm: Done.

azure/services/managedclusters/spec_test.go
```diff
@@ -163,6 +165,67 @@ func TestParameters(t *testing.T) {
...
+								"test-tag": "test-value",
+							},
+						},
+					}, nil
+				},
+			},
+			expect: func(g *WithT, result interface{}) {
+				g.Expect(result).To(BeAssignableToTypeOf(containerservice.ManagedCluster{}))
+				g.Expect(result.(containerservice.ManagedCluster).LinuxProfile).To(Not(BeNil()))
+				g.Expect(result.(containerservice.ManagedCluster).LinuxProfile.SSH.PublicKeys).To(Not(BeNil()))
```
nawazkh, line 196: Can we update this check to a more fine-grained one?
```suggestion
				g.Expect(result.(containerservice.ManagedCluster).LinuxProfile.SSH.PublicKeys).To(Equal(base64.StdEncoding.EncodeToString([]byte("test-ssh-key"))))
```

nawazkh (commented):
Just one final suggestion from my end :) Everything else looks great to me! Thank you

nawazkh:
> @nawazkh I considered adding new field in [my comment above](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3677#issuecomment-1612698386):
> 
> > I was also considering adding another field like AutogenerateSSHKey that would be true by default, but once disabled then SSH key would not be autogenerated. Then we would not be changing PublicSSHKey type but adding new field instead.
> 
> But after thinking longer about it, all fields that are part of control plane spec map somehow to AKS fields. This would be used only by CAPZ and it would be used only once during init, after that it would be redundant. I prefer current solution but if you would like then I can change it. Both ways should work.


_"This would be used only by CAPZ and it would be used only once during init, after that it would be redundant."_ Makes sense, thanks for sharing the context again!

maciaszczykm:
@nawazkh Done, PTAL.

azure/services/managedclusters/spec_test.go
```diff
@@ -163,6 +165,67 @@ func TestParameters(t *testing.T) {
...
+								"test-tag": "test-value",
+							},
+						},
+					}, nil
+				},
+			},
+			expect: func(g *WithT, result interface{}) {
+				g.Expect(result).To(BeAssignableToTypeOf(containerservice.ManagedCluster{}))
+				g.Expect(result.(containerservice.ManagedCluster).LinuxProfile).To(Not(BeNil()))
+				g.Expect(result.(containerservice.ManagedCluster).LinuxProfile.SSH.PublicKeys).To(Equal(base64.StdEncoding.EncodeToString([]byte("test-ssh-key"))))
```
nawazkh, line 196: I apologize @maciaszczykm , my suggestion seems to have broken your test. 
I wanted to suggest testing the contents of `result.(containerservice.ManagedCluster).LinuxProfile.SSH.PublicKeys` rather than just checking _it not to be nil_ for higher confidence.
This is what the test complains about
```
Expected
                <*[]containerservice.SSHPublicKey | 0xc000012a98>: [
                    {
                        KeyData: "test-ssh-key",
                    },
                ]
            to equal
                <string>: dGVzdC1zc2gta2V5
```
We can either compare the value of `result.(containerservice.ManagedCluster).LinuxProfile.SSH.PublicKeys[0].KeyData` with `"test-ssh-key"` or revert the change, essentially removing the line `g.Expect(result.(containerservice.ManagedCluster).LinuxProfile.SSH.PublicKeys).To(Equal(base64.StdEncoding.EncodeToString([]byte("test-ssh-key"))))` .

Sorry to cause such a back-and-forth on this.

maciaszczykm: Sure, no worries. I will update it in a moment.

maciaszczykm: It's done.

maciaszczykm:
@nawazkh PTAL.

nawazkh:
Thank you for putting this together and working through the conversations. Let's get this merged! 🚀 
/lgtm

CecileRobertMichon (commented):
/lgtm
/approve

CecileRobertMichon:
@nawazkh IMO it's not eligible for cherry-pick because it's a feature 

see https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/docs/book/src/developers/releasing.md#release-support

nawazkh:
Makes sense, thank you!

Issue #3526, opened 2023-05-04
Implement all currently supported flavors of Azure authentication for ASO


nojnhuh:
CAPZ supports several different mechanisms for authenticating to the Azure API as documented at https://capz.sigs.k8s.io/topics/multitenancy.html. ASO supports the same options (AFAICT) as described here: https://azure.github.io/azure-service-operator/guide/authentication/.

Whichever option a user chooses to authenticate with CAPZ, that should be passed through to be used for ASO.

Some more context is described in the ASO proposal: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/docs/proposals/20230123-azure-service-operator.md#security-model

```[tasklist]
### Dependencies
- [ ] #3520
- [ ] https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/3521
```

nojnhuh:
Added a link in the description above to the relevant section from the ASO proposal.

Issue #3527, opened 2023-05-04
Enable ASO


nojnhuh:
This issue tracks "turning on" ASO from within CAPZ. Before then, ASO changes will be integrated in a way such that ASO is not usable so progress can be made iteratively while working toward feature parity with CAPZ's current feature set. The following list tracks reaching that parity. Once the prerequisites are met, one atomic change will be made to "turn on" ASO.

This depends on only at least one of the Azure service implementations being converted to ASO. Other services will still be using the Azure SDK when this issue is completed.

```[tasklist]
### Dependencies
- [ ] #3520
- [ ] #3521
- [ ] #3522
- [ ] #3524
- [ ] #3525
- [ ] #3526
- [ ] https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/3795
- [ ] https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/3732
- [ ] https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/3743
```

PR #3698, opened 2023-07-10
Add support for Azure authentication in ASO

adriananeci:
**What type of PR is this?**
/kind feature

More context is described in the ASO proposal: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/docs/proposals/20230123-azure-service-operator.md#security-model

Fixes  https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/3526

Because we can have multiple `AzureClusterIdentity` objects created inside the same namespace, we cannot make use of https://azure.github.io/azure-service-operator/guide/authentication/credential-scope/#namespace-scope so we'll have to reside only on global scope or resource scope

Haven't added any tests yet. I'll add those after the initial review since I'm looking for some early feedback to make sure I'm on the right path.
 
- [ ] cherry-pick candidate

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Add support for Azure authentication in ASO
```

### DIFF
--- a/azure/scope/identity.go
+++ b/azure/scope/identity.go
@@ -42,7 +42,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-const azureSecretKey = "clientSecret"
+// AzureSecretKey is the value for they client secret key.
+const AzureSecretKey = "clientSecret"
 
 // CredentialsProvider defines the behavior for azure identity based credential providers.
 type CredentialsProvider interface {
@@ -220,7 +221,7 @@ func (p *AzureCredentialsProvider) GetClientSecret(ctx context.Context) (string,
 		if err := p.Client.Get(ctx, key, secret); err != nil {
 			return "", errors.Wrap(err, "Unable to fetch ClientSecret")
 		}
-		return string(secret.Data[azureSecretKey]), nil
+		return string(secret.Data[AzureSecretKey]), nil
 	}
 	return "", nil
 }
--- a/azure/services/aso/aso.go
+++ b/azure/services/aso/aso.go
@@ -29,6 +29,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
 	"sigs.k8s.io/cluster-api-provider-azure/azure"
+	"sigs.k8s.io/cluster-api-provider-azure/util/aso"
 	"sigs.k8s.io/cluster-api-provider-azure/util/tele"
 	"sigs.k8s.io/cluster-api/util/patch"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -46,6 +47,9 @@ const (
 
 	createOrUpdateFutureType = "ASOCreateOrUpdate"
 	deleteFutureType         = "ASODelete"
+
+	// SecretNameAnnotation is the annotation key for ASO's credentials to use.
+	SecretNameAnnotation = "serviceoperator.azure.com/credential-from"
 )
 
 // Service is an implementation of the Reconciler interface. It handles creation
@@ -180,6 +184,10 @@ func (s *Service) CreateOrUpdateResource(ctx context.Context, spec azure.ASOReso
 	if adopt {
 		annotations[ReconcilePolicyAnnotation] = ReconcilePolicyManage
 	}
+
+	// Set the secret name annotation in order to leverage the ASO resource credential scope as defined in
+	// https://azure.github.io/azure-service-operator/guide/authentication/credential-scope/#resource-scope.
+	annotations[SecretNameAnnotation] = aso.GetASOSecretName(s.clusterName)
 
 	if len(labels) == 0 {
 		labels = nil
--- a/controllers/asosecret_controller.go
+++ b/controllers/asosecret_controller.go
@@ -1,0 +1,324 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/record"
+	"k8s.io/utils/ptr"
+	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
+	"sigs.k8s.io/cluster-api-provider-azure/azure/scope"
+	"sigs.k8s.io/cluster-api-provider-azure/util/aso"
+	"sigs.k8s.io/cluster-api-provider-azure/util/reconciler"
+	"sigs.k8s.io/cluster-api-provider-azure/util/tele"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"sigs.k8s.io/cluster-api/util"
+	"sigs.k8s.io/cluster-api/util/annotations"
+	"sigs.k8s.io/cluster-api/util/predicates"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+// ASOSecretReconciler reconciles ASO secrets associated with AzureCluster objects.
+type ASOSecretReconciler struct {
+	client.Client
+	Recorder         record.EventRecorder
+	ReconcileTimeout time.Duration
+	WatchFilterValue string
+}
+
+// SetupWithManager initializes this controller with a manager.
+func (asos *ASOSecretReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager, options controller.Options) error {
+	_, log, done := tele.StartSpanWithLogger(ctx,
+		"controllers.ASOSecretReconciler.SetupWithManager",
+		tele.KVP("controller", "ASOSecret"),
+	)
+	defer done()
+
+	c, err := ctrl.NewControllerManagedBy(mgr).
+		WithOptions(options).
+		For(&infrav1.AzureCluster{}).
+		WithEventFilter(predicates.ResourceNotPausedAndHasFilterLabel(log, asos.WatchFilterValue)).
+		WithEventFilter(predicates.ResourceIsNotExternallyManaged(log)).
+		Named("ASOSecret").
+		Owns(&corev1.Secret{}).
+		Build(asos)
+	if err != nil {
+		return errors.Wrap(err, "error creating controller")
+	}
+
+	// Add a watch on infrav1.AzureManagedControlPlane.
+	if err = c.Watch(
+		&source.Kind{Type: &infrav1.AzureManagedControlPlane{}},
+		&handler.EnqueueRequestForObject{},
+		predicates.ResourceNotPausedAndHasFilterLabel(log, asos.WatchFilterValue),
+	); err != nil {
+		return errors.Wrap(err, "failed adding a watch for ready AzureManagedControlPlanes")
+	}
+
+	// Add a watch on ASO secrets owned by an AzureManagedControlPlane
+	if err = c.Watch(
+		&source.Kind{Type: &corev1.Secret{}},
+		&handler.EnqueueRequestForOwner{
+			OwnerType:    &infrav1.AzureManagedControlPlane{},
+			IsController: true,
+		},
+	); err != nil {
+		return errors.Wrap(err, "failed adding a watch for secrets")
+	}
+
+	// Add a watch on clusterv1.Cluster object for unpause notifications.
+	if err = c.Watch(
+		&source.Kind{Type: &clusterv1.Cluster{}},
+		handler.EnqueueRequestsFromMapFunc(util.ClusterToInfrastructureMapFunc(ctx, infrav1.GroupVersion.WithKind("AzureCluster"), mgr.GetClient(), &infrav1.AzureCluster{})),
+		predicates.ClusterUnpaused(log),
+		predicates.ResourceNotPausedAndHasFilterLabel(log, asos.WatchFilterValue),
+	); err != nil {
+		return errors.Wrap(err, "failed adding a watch for ready clusters")
+	}
+
+	return nil
+}
+
+// Reconcile reconciles the ASO secrets associated with AzureCluster objects.
+func (asos *ASOSecretReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Result, reterr error) {
+	ctx, cancel := context.WithTimeout(ctx, reconciler.DefaultedLoopTimeout(asos.ReconcileTimeout))
+	defer cancel()
+
+	ctx, log, done := tele.StartSpanWithLogger(ctx, "controllers.ASOSecret.Reconcile",
+		tele.KVP("namespace", req.Namespace),
+		tele.KVP("name", req.Name),
+		tele.KVP("kind", "AzureCluster"),
+	)
+	defer done()
+
+	log = log.WithValues("namespace", req.Namespace, "AzureCluster", req.Name)
+
+	// asoSecretOwner is the resource that created the identity. This could be either an AzureCluster or AzureManagedControlPlane (if AKS is enabled).
+	// check for AzureCluster first and if it is not found, check for AzureManagedControlPlane.
+	var asoSecretOwner client.Object
+
+	azureCluster := &infrav1.AzureCluster{}
+	checkForManagedControlPlane := false
+	// Fetch the AzureCluster or AzureManagedControlPlane instance
+	asoSecretOwner = azureCluster
+	err := asos.Get(ctx, req.NamespacedName, azureCluster)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			checkForManagedControlPlane = true
+		} else {
+			return reconcile.Result{}, err
+		}
+	} else {
+		log = log.WithValues("AzureCluster", req.Name)
+	}
+
+	if checkForManagedControlPlane {
+		// Fetch the AzureManagedControlPlane instance instead
+		azureManagedControlPlane := &infrav1.AzureManagedControlPlane{}
+		asoSecretOwner = azureManagedControlPlane
+		err = asos.Get(ctx, req.NamespacedName, azureManagedControlPlane)
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				asos.Recorder.Eventf(azureCluster, corev1.EventTypeNormal, "AzureClusterObjectNotFound",
+					fmt.Sprintf("AzureCluster object %s/%s not found", req.Namespace, req.Name))
+				asos.Recorder.Eventf(azureManagedControlPlane, corev1.EventTypeNormal, "AzureManagedControlPlaneObjectNotFound",
+					fmt.Sprintf("AzureManagedControlPlane object %s/%s not found", req.Namespace, req.Name))
+				log.Info("object was not found")
+				return reconcile.Result{}, nil
+			} else {
+				return reconcile.Result{}, err
+			}
+		} else {
+			log = log.WithValues("AzureManagedControlPlane", req.Name)
+		}
+	}
+
+	var clusterIdentity *corev1.ObjectReference
+	var cluster *clusterv1.Cluster
+	var azureClient scope.AzureClients
+
+	switch ownerType := asoSecretOwner.(type) {
+	case *infrav1.AzureCluster:
+		clusterIdentity = ownerType.Spec.IdentityRef
+
+		// Fetch the Cluster.
+		cluster, err = util.GetOwnerCluster(ctx, asos.Client, ownerType.ObjectMeta)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+
+		// Create the scope.
+		clusterScope, err := scope.NewClusterScope(ctx, scope.ClusterScopeParams{
+			Client:       asos.Client,
+			Cluster:      cluster,
+			AzureCluster: ownerType,
+		})
+		if err != nil {
+			return reconcile.Result{}, errors.Wrap(err, "failed to create scope")
+		}
+
+		azureClient = clusterScope.AzureClients
+
+	case *infrav1.AzureManagedControlPlane:
+		clusterIdentity = ownerType.Spec.IdentityRef
+
+		// Fetch the Cluster.
+		cluster, err = util.GetOwnerCluster(ctx, asos.Client, ownerType.ObjectMeta)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+
+		// Create the scope.
+		clusterScope, err := scope.NewManagedControlPlaneScope(ctx, scope.ManagedControlPlaneScopeParams{
+			Client:       asos.Client,
+			Cluster:      cluster,
+			ControlPlane: ownerType,
+		})
+		if err != nil {
+			return reconcile.Result{}, errors.Wrap(err, "failed to create scope")
+		}
+
+		azureClient = clusterScope.AzureClients
+	}
+
+	if cluster == nil {
+		log.Info("Cluster Controller has not yet set OwnerRef")
+		asos.Recorder.Eventf(asoSecretOwner, corev1.EventTypeNormal, "OwnerRefNotFound",
+			fmt.Sprintf("Cluster Controller has not yet set OwnerRef for object %s/%s", req.Namespace, req.Name))
+		return reconcile.Result{}, nil
+	}
+
+	log = log.WithValues("cluster", cluster.Name)
+
+	// Return early if the ASO Secret Owner(AzureCluster or AzureManagedControlPlane) or Cluster is paused.
+	if annotations.IsPaused(cluster, asoSecretOwner) {
+		log.Info(fmt.Sprintf("%s or linked Cluster is marked as paused. Won't reconcile", asoSecretOwner.GetObjectKind()))
+		asos.Recorder.Eventf(asoSecretOwner, corev1.EventTypeNormal, "ClusterPaused",
+			fmt.Sprintf("%s or linked Cluster is marked as paused. Won't reconcile", asoSecretOwner.GetObjectKind().GroupVersionKind().Kind))
+		return ctrl.Result{}, nil
+	}
+
+	// Construct the ASO secret for this Cluster
+	newASOSecret, err := asos.createSecretFromClusterIdentity(ctx, clusterIdentity, cluster, azureClient)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	gvk := asoSecretOwner.GetObjectKind().GroupVersionKind()
+	owner := metav1.OwnerReference{
+		APIVersion: gvk.GroupVersion().String(),
+		Kind:       gvk.Kind,
+		Name:       asoSecretOwner.GetName(),
+		UID:        asoSecretOwner.GetUID(),
+		Controller: ptr.To(true),
+	}
+
+	newASOSecret.OwnerReferences = []metav1.OwnerReference{owner}
+
+	if err := reconcileAzureSecret(ctx, asos.Client, owner, newASOSecret, cluster.GetName()); err != nil {
+		asos.Recorder.Eventf(cluster, corev1.EventTypeWarning, "Error reconciling ASO secret", err.Error())
+		return ctrl.Result{}, errors.Wrap(err, "failed to reconcile ASO secret")
+	}
+
+	return ctrl.Result{}, nil
+}
+
+func (asos *ASOSecretReconciler) createSecretFromClusterIdentity(ctx context.Context, clusterIdentity *corev1.ObjectReference, cluster *clusterv1.Cluster, azureClient scope.AzureClients) (*corev1.Secret, error) {
+	newASOSecret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      aso.GetASOSecretName(cluster.GetName()),
+			Namespace: cluster.GetNamespace(),
+			Labels: map[string]string{
+				cluster.GetName(): string(infrav1.ResourceLifecycleOwned),
+			},
+		},
+		Data: map[string][]byte{
+			"AZURE_SUBSCRIPTION_ID": []byte(azureClient.SubscriptionID()),
+		},
+	}
+
+	if clusterIdentity != nil {
+		// if the namespace isn't specified then assume it's in the same namespace as the Cluster's one
+		namespace := clusterIdentity.Namespace
+		if namespace == "" {
+			namespace = cluster.GetNamespace()
+		}
+		identity := &infrav1.AzureClusterIdentity{}
+		key := client.ObjectKey{
+			Name:      clusterIdentity.Name,
+			Namespace: namespace,
+		}
+		if err := asos.Get(ctx, key, identity); err != nil {
+			return nil, errors.Wrap(err, "failed to retrieve AzureClusterIdentity")
+		}
+
+		newASOSecret.Data["AZURE_TENANT_ID"] = []byte(identity.Spec.TenantID)
+		newASOSecret.Data["AZURE_CLIENT_ID"] = []byte(identity.Spec.ClientID)
+
+		// Fetch identity secret, if it exists
+		key = types.NamespacedName{
+			Namespace: identity.Spec.ClientSecret.Namespace,
+			Name:      identity.Spec.ClientSecret.Name,
+		}
+		identitySecret := &corev1.Secret{}
+		err := asos.Get(ctx, key, identitySecret)
+		if err != nil && !apierrors.IsNotFound(err) {
+			return nil, errors.Wrap(err, "failed to fetch AzureClusterIdentity secret")
+		}
+
+		switch identity.Spec.Type {
+		case infrav1.ServicePrincipal, infrav1.ManualServicePrincipal:
+			newASOSecret.Data["AZURE_CLIENT_SECRET"] = identitySecret.Data[scope.AzureSecretKey]
+		case infrav1.ServicePrincipalCertificate:
+			newASOSecret.Data["AZURE_CLIENT_CERTIFICATE"] = identitySecret.Data["certificate"]
+			newASOSecret.Data["AZURE_CLIENT_CERTIFICATE_PASSWORD"] = identitySecret.Data["password"]
+		}
+	} else {
+		newASOSecret.Data["AZURE_TENANT_ID"] = []byte(azureClient.TenantID())
+		newASOSecret.Data["AZURE_CLIENT_ID"] = []byte(azureClient.ClientID())
+
+		// Populate ASO data in the following order:
+		// 1. Client credentials
+		// 2. Client certificate
+		if _, e := azureClient.GetClientCredentials(); e == nil {
+			newASOSecret.Data["AZURE_CLIENT_SECRET"] = []byte(azureClient.ClientSecret())
+		} else if clientCert, e := azureClient.GetClientCertificate(); e == nil {
+			cert, err := getCertificateFromFile(clientCert.CertificatePath)
+			if err != nil {
+				return nil, errors.Wrap(err, "failed to read client certificate")
+			}
+			newASOSecret.Data["AZURE_CLIENT_CERTIFICATE"] = cert
+			newASOSecret.Data["AZURE_CLIENT_CERTIFICATE_PASSWORD"] = []byte(clientCert.CertificatePassword)
+		} else {
+			return nil, errors.Wrap(e, "failed to configure an authentication method for ASO secret")
+		}
+	}
+	return newASOSecret, nil
+}
--- a/controllers/asosecret_controller_test.go
+++ b/controllers/asosecret_controller_test.go
@@ -1,0 +1,367 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import (
+	"context"
+	"os"
+	"testing"
+
+	aadpodv1 "github.com/Azure/aad-pod-identity/pkg/apis/aadpodidentity/v1"
+	"github.com/Azure/go-autorest/autorest/azure/auth"
+	. "github.com/onsi/gomega"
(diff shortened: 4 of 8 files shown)


azure/services/asogroups/spec.go
```diff
@@ -55,6 +55,9 @@ func (s *GroupSpec) Parameters(ctx context.Context, object genruntime.MetaObject
 	return &asoresourcesv1.ResourceGroup{
 		ObjectMeta: metav1.ObjectMeta{
 			OwnerReferences: []metav1.OwnerReference{s.Owner},
+			// Annotations:     map[string]string{
+			//   "serviceoperator.azure.com/credential-from": "my-resource-secret",
+			// },
```
rdtiu, line 60: why was this commented? can it be removed?

nojnhuh:
Adding WIP while we're not ready to merge this yet:
/retitle [WIP] Add support for Azure authentication in ASO

azure/services/asogroups/spec.go
```diff
@@ -55,6 +55,9 @@ func (s *GroupSpec) Parameters(ctx context.Context, object genruntime.MetaObject
 	return &asoresourcesv1.ResourceGroup{
 		ObjectMeta: metav1.ObjectMeta{
 			OwnerReferences: []metav1.OwnerReference{s.Owner},
+			// Annotations:     map[string]string{
+			//   "serviceoperator.azure.com/credential-from": "my-resource-secret",
```
nojnhuh, line 59: Is there a way we can more generically annotate ASO resources with the secret in azure/services/aso/aso.go without needing to explicitly specify it here?

nojnhuh (commented):
Overall approach looks good to me.

For the other ASO PRs, I've been splitting up those into non-functional changes (w.r.t. the current behavior) that get merged and functional changes (to be included in a PR resolving #3527) that I've been managing in my fork. The new controller and creation of ASO secrets here don't seem like they would break current behavior though so we may not need to split anything out here.

@CecileRobertMichon Do we have existing e2e coverage for all these different ways of authenticating to Azure?

azure/services/asogroups/spec.go
adriananeci: I'll remove it once the PR is ready to be merged. I left it here as an example to discuss around how do we want to pass the ASO `credential-from` label

azure/scope/identity.go
```diff
@@ -102,6 +105,10 @@ func NewAzureClusterCredentialsProvider(ctx context.Context, kubeClient client.C
 
 // GetAuthorizer returns an Azure authorizer based on the provided azure identity. It delegates to AzureCredentialsProvider with AzureCluster metadata.
 func (p *AzureClusterCredentialsProvider) GetAuthorizer(ctx context.Context, resourceManagerEndpoint, activeDirectoryEndpoint, tokenAudience string) (autorest.Authorizer, error) {
+	if err := reconcileASOSecret(ctx, p.AzureCredentialsProvider.Identity, p.Client, p.AzureCluster.Spec.SubscriptionID); err != nil {
```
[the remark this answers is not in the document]
adriananeci: Nope. The intention is to use the credentials ASO controller is configured with and avoid labelling ASO objects that are related to a cluster not specifying the `spec.identityRef`

azure/services/asogroups/spec.go
adriananeci: It might be, I was thinking the same. So my idea is to grab the `spec.identityRef` of the cluster((i.e. AzureClusters/AzureManagedControlPlanes) in `aso.go`, compute the secret name and label the ASO object with the  "serviceoperator.azure.com/credential-from": "<secret_name>"

adriananeci: Added the changes in https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3698/commits/34fdf794228d703d638522c2a258614e8f61e0f5

azure/scope/identity.go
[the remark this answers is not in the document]
nojnhuh: Since subscription ID is specified outside the AzureClusterIdentity, could two clusters to use the same set of default credentials but have different subscription IDs, or is that not possible given how auth generally works in Azure?

nojnhuh:
(reminder to squash once we get lgtms)

/hold

And @adriananeci feel free to remove the [WIP] from the title once you're finished and let me know so I can give this another look.

adriananeci:
@nojnhuh I think this is ready for another round reviews now

azure/scope/identity.go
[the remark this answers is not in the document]
adriananeci: Yup, it should possible to use the default set of credentials in different subscriptions, as long as the default credentials have access to those subscriptions.

azure/scope/cluster.go
```diff
@@ -755,6 +755,19 @@ func (s *ClusterScope) ClusterName() string {
...
+func (s *ClusterScope) AzureClusterIdentityName() string {
+	if s.AzureCluster.Spec.IdentityRef == nil {
+		return ""
+	}
+	return s.AzureCluster.Spec.IdentityRef.Name
+}
+
+// SubscriptionID returns the Azure client Subscription ID.
+func (s *ClusterScope) SubscriptionID() string {
+	return s.AzureClients.SubscriptionID()
```
nojnhuh, line 768: Since AzureClients is embedded in ClusterScope, I don't think this method definition is strictly required to fulfill the GroupScope interface. Unless there's another need for this I'm not seeing.

azure/scope/identity.go
```diff
@@ -351,3 +363,112 @@ func IsClusterNamespaceAllowed(ctx context.Context, k8sClient client.Client, all
...
+		Namespace: azureClusterIdentity.Spec.ClientSecret.Namespace,
+		Name:      azureClusterIdentity.Spec.ClientSecret.Name,
+	}
+	identitySecret := &corev1.Secret{}
+	err := kubeClient.Get(ctx, key, identitySecret)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return errors.Errorf("failed to fetch identity secret %s as defined in AzureClusterIdentity %q/%q: %v", identitySecret.Name, azureClusterIdentity.Namespace, azureClusterIdentity.Name, err)
+	}
+
+	// TODO add support for workload identity
```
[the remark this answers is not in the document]
nojnhuh: @adriananeci Could you please either unlink #3526 from this PR so that issue doesn't automatically close or open a new issue for Workload ID + ASO to make sure this is tracked?

azure/scope/identity.go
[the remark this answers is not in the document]
nojnhuh: Since subscription ID is part of the default ASO credentials, is it possible that we could run into an issue then if an AzureCluster's `spec.identityRef` is undefined and `spec.subscriptionID` is different from ASO's global default? Would we still need an ASO secret in that case to specify the cluster's subscription ID?

azure/scope/identity.go
```diff
@@ -361,3 +371,112 @@ func IsClusterNamespaceAllowed(ctx context.Context, k8sClient client.Client, all
...
+		if err := kubeClient.Create(ctx, asoSecret); err != nil && !apierrors.IsAlreadyExists(err) {
+			return errors.Wrap(err, "failed to create ASO secret")
+		}
+		return nil
+	}
+
+	// Otherwise, check ownership and data freshness. Update as necessary
+	hasOwner := false
+	for _, ownerRef := range oldSecret.OwnerReferences {
+		if referSameObject(ownerRef, owner) {
```
nojnhuh, line 442: Is there a case where only checking the UID wouldn't work here?

azure/services/asogroups/groups_test.go
```diff
@@ -67,13 +67,17 @@ func TestReconcileGroups(t *testing.T) {
 			expectedError: "",
 			expect: func(s *mock_asogroups.MockGroupScopeMockRecorder, _ *mock_aso.MockReconcilerMockRecorder) {
 				s.ASOGroupSpec().Return(nil)
+				s.SubscriptionID().Return("").AnyTimes()
+				s.AzureClusterIdentityName().Return("").AnyTimes()
```
nojnhuh, line 71: If every test needs these and we expect that to stay true, it might be worth moving these from each test case to the for loop with the rest of the test boilerplate.

util/identity/defaults.go
```diff
@@ -22,3 +22,11 @@ import "fmt"
 func GetAzureIdentityName(clusterName, clusterNamespace, identityName string) string {
 	return fmt.Sprintf("%s-%s-%s", clusterName, clusterNamespace, identityName)
 }
+
+// GetASOSecretName formats the name of the ASO Secret created by the capz controller.
+func GetASOSecretName(clusterIdentitySecret, subscriptionID string) string {
+	if subscriptionID == "" {
```
nojnhuh, line 28: For my own understanding, when might subscription ID be empty here? It looks to me like that's not possible given this:
https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/3562a2a29df8efbb9adb43442dde748c890882a1/azure/scope/clients.go#L116-L121

controllers/azureclusteridentity_controller.go
```diff
@@ -0,0 +1,121 @@
...
+	err := acir.Get(ctx, req.NamespacedName, azureClusterIdentity)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			log.Info("object was not found")
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, err
+	}
+
+	// Handle deleted azureClusterIdentity
```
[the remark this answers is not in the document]
adriananeci: That sounds reasonable to me. If we agree on relying solely on the Kubernetes garbage collection I can remove this controller.

azure/scope/identity.go
adriananeci: Maybe I'm misunderstanding the question, but we need to make sure the old secret has all the necessary stuffs in place(owner, data ,etc), so not sure how only checking the UID might work.

nojnhuh: I mean using only the UID to see if `ownerRef` and `owner` are equivalent instead of the combination of API version, kind, and name like `referSameObject` does.

util/identity/defaults.go
adriananeci: that's true, will remove the subscription check

azure/scope/identity.go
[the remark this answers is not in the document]
nojnhuh: Sorry for the back and forth @adriananeci but I synced with @dtzar and could you please re-link #3526 to this PR so it gets closed once this is merged? I've opened a separate issue for just workload ID (#3732).

azure/scope/cluster.go
adriananeci: I've added it initially since the `managedcontrolplane` has it too. Removed it for now.

adriananeci:
>Since subscription ID is part of the default ASO credentials, is it possible that we could run into an issue then if an AzureCluster's `spec.identityRef` is undefined and `spec.subscriptionID` is different from ASO's global default? Would we still need an ASO secret in that case to specify the cluster's subscription ID?

You are right, I thought initially that ASO resources support to specify the subscription ID inside the `spec` section, but that's not the case. I've extended the logic to also create an ASO secret if an AzureCluster's `spec.identityRef` is undefined and `spec.subscriptionID` is different from ASO's global default

azure/scope/clients.go
```diff
@@ -39,6 +49,15 @@ type AzureClients struct {
 	ResourceManagerVMDNSSuffix string
 }
 
+// ASOSecretClient contains all the necessary info for creating an ASO secret.
+type ASOSecretClient struct {
+	ClusterKind       string
+	ClusterObjectMeta metav1.ObjectMeta
```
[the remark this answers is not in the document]
adriananeci: Yup, should be cleaner, updated it in https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3698/commits/df11005c8dd1acff960ba497e53619109a374a92

azure/scope/clients.go
```diff
@@ -169,3 +197,109 @@ func setValue(settings auth.EnvironmentSettings, key string) {
 		settings.Values[key] = v
 	}
 }
+
+func (c *AzureClients) reconcileASOSecretWithoutClusterIdentity(ctx context.Context, asoSecret *ASOSecretClient) error {
```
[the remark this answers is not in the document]
adriananeci: Sure thing, did that in https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3698/commits/df11005c8dd1acff960ba497e53619109a374a92

azure/scope/clients.go
```diff
@@ -91,9 +112,16 @@ func (c *AzureClients) setCredentials(subscriptionID, environmentName string) er
 	c.ResourceManagerVMDNSSuffix = settings.Environment.ResourceManagerVMDNSSuffix
 	c.Values[auth.ClientID] = strings.TrimSuffix(c.Values[auth.ClientID], "\n")
 	c.Values[auth.ClientSecret] = strings.TrimSuffix(c.Values[auth.ClientSecret], "\n")
-	c.Values[auth.SubscriptionID] = strings.TrimSuffix(subscriptionID, "\n")
+	c.Values[auth.SubscriptionID] = strings.TrimSuffix(asoSecret.SubscriptionID, "\n")
 	c.Values[auth.TenantID] = strings.TrimSuffix(c.Values[auth.TenantID], "\n")
 
+	if createASOSecret {
+		err = c.reconcileASOSecretWithoutClusterIdentity(ctx, asoSecret)
```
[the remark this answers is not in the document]
adriananeci: Moved it in `setCredentialsWithProvider` in https://github.com/kubernetes-sigs/cluster-api-provider-azure/commit/df11005c8dd1acff960ba497e53619109a374a92

controllers/azureclusteridentity_controller.go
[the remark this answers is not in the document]
adriananeci: I went ahead and removed the azure cluster identity controller

azure/scope/clients.go
```diff
@@ -169,3 +205,151 @@ func setValue(settings auth.EnvironmentSettings, key string) {
 		settings.Values[key] = v
 	}
 }
+
+func (c *AzureClients) reconcileASOSecret(ctx context.Context, asoClientSecret *ASOSecretClient) error {
```
nojnhuh, line 209: This function seems big enough that I can see value in adding at least one unit test exercising the happy paths for each of `asoClientSecret.AzureClusterIdentity != nil` and the inverse.

nojnhuh (commented):
Looking really good! Just couple small things I saw.

CecileRobertMichon:
> Do we have existing e2e coverage for all these different ways of authenticating to Azure

We test both AAD pod identity and Workload identity 

For AAD pod identity we only test `ServicePrincipal` though, not `UserAssignedMSI` since testing managed identity requires the management cluster to be in Azure. We _could_ add one that created a cluster that has managed identity on the VMs and then pivot that cluster to be a management cluster and creates a cluster from there. We already have the private cluster scenario so I'm thinking it'd be pretty easy to extend it to do just that. 

We do test enabling user-assigned identity on the workload cluster VMs and using that for cloud-provider-azure.

azure/scope/clients.go
```diff
@@ -169,3 +205,151 @@ func setValue(settings auth.EnvironmentSettings, key string) {
...
+			return fmt.Errorf("MSI not available")
+		} else {
+			newASOSecret.Data["AZURE_IDENTITY_RESOURCE_ID"] = []byte(c.Values[auth.Resource])
+		}
+	}
+
+	return reconcileSecret(ctx, asoClientSecret.KubeClient, newASOSecret, owner)
+}
+
+func reconcileSecret(ctx context.Context, kubeClient client.Client, secret *corev1.Secret, owner metav1.OwnerReference) error {
```
CecileRobertMichon, line 298: this doesn't necessarily seem like the best place for this function to live, especially given that we're having to add all those k8s client imports in here. Would it make sense to add it to one of the controllers? Or even make it its own controller, just like the azurejson controllers? Is this something we expect to be shared between multiple clusters or is there one per AzureCluster?

util/aso/defaults.go
```diff
@@ -0,0 +1,24 @@
...
+limitations under the License.
+*/
+
+package aso
+
+import "fmt"
+
+// GetASOSecretName formats the name of the ASO Secret created by the capz controller.
+func GetASOSecretName(owner, subscriptionID string) string {
+	return fmt.Sprintf("%s-%s", owner, subscriptionID)
```
CecileRobertMichon, line 23: is subscriptionID unique enough? Do we expect to only ever have 1 secret per subscription + owner (I'm assuming owner is the cluster here)?

azure/services/aso/aso.go
```diff
@@ -47,21 +50,28 @@ const (
 
 	createOrUpdateFutureType = "ASOCreateOrUpdate"
 	deleteFutureType         = "ASODelete"
+
+	// SecretNameAnnotation is the annotation key for ASO's credentials to use.
+	SecretNameAnnotation = "serviceoperator.azure.com/credential-from"
```
CecileRobertMichon, line 55: is there a place this can be imported from in ASO?

azure/scope/clients.go
```diff
@@ -169,3 +205,151 @@ func setValue(settings auth.EnvironmentSettings, key string) {
...
+
+	var owner metav1.OwnerReference
+
+	if asoClientSecret.AzureClusterIdentity != nil {
+		newASOSecret.ObjectMeta = metav1.ObjectMeta{
+			Name:      aso.GetASOSecretName(asoClientSecret.AzureClusterIdentity.GetName(), asoClientSecret.SubscriptionID),
+			Namespace: asoClientSecret.AzureClusterIdentity.GetNamespace(),
+		}
+
+		newASOSecret.Data["AZURE_TENANT_ID"] = []byte(asoClientSecret.AzureClusterIdentity.Spec.TenantID)
```
CecileRobertMichon, line 224: please use constants for all the environment variable names (they are already available in the imported `auth` package), eg. `auth.TenantID`

azure/services/aso/aso.go
```diff
@@ -170,6 +180,17 @@ func (s *Service) CreateOrUpdateResource(ctx context.Context, spec azure.ASOReso
...
 
+	// Set the secret name annotation if any of the following conditions are met:
+	//- AzureClusterIdentity name is not empty
+	//- SubscriptionID is different from the one controller is running with
+	// This is needed in order to leverage the ASO resource credential scope as defined in
+	//https://azure.github.io/azure-service-operator/guide/authentication/credential-scope/#resource-scope.
+	if s.azureClusterIdentityName != "" {
+		annotations[SecretNameAnnotation] = aso.GetASOSecretName(s.azureClusterIdentityName, s.subscriptionID)
+	} else if s.subscriptionID != os.Getenv(auth.SubscriptionID) {
+		annotations[SecretNameAnnotation] = aso.GetASOSecretName("", s.subscriptionID)
```
CecileRobertMichon, line 191: this will result in a secret name starting with `-` if I'm reading the `GetASOSecretName` function right

nojnhuh:
> > Do we have existing e2e coverage for all these different ways of authenticating to Azure
> 
> We test both AAD pod identity and Workload identity
> 
> For AAD pod identity we only test `ServicePrincipal` though, not `UserAssignedMSI` since testing managed identity requires the management cluster to be in Azure. We _could_ add one that created a cluster that has managed identity on the VMs and then pivot that cluster to be a management cluster and creates a cluster from there. We already have the private cluster scenario so I'm thinking it'd be pretty easy to extend it to do just that.
> 
> We do test enabling user-assigned identity on the workload cluster VMs and using that for cloud-provider-azure.

Thanks, I think we can address any gaps outside of this PR if needed. It seems like the most sensitive changes in this PR are shared between all the (non-Workload ID) types, so unit test coverage to make sure parameters each auth method are set appropriately might be good enough for now.

azure/scope/clients.go
adriananeci: While I can re-use the existing constants, keep in mind that they are ASO related and some of the ASO secret data key names are slightly different compared with CAPZ ones(e.g `AZURE_CERTIFICATE_PASSWORD` in CAPZ vs `AZURE_CLIENT_CERTIFICATE_PASSWORD` in ASO or `AZURE_CERTIFICATE_PATH` in CAPZ and `AZURE_CLIENT_CERTIFICATE` in ASO).

I would prefer to use the ASO constants if they're getting exported(same as in https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3698#discussion_r1267394407)

azure/services/aso/aso.go
adriananeci: Ups, forgot to add the `s.clusterName`, will add it in the next commit

azure/scope/clients.go
```diff
@@ -169,3 +205,151 @@ func setValue(settings auth.EnvironmentSettings, key string) {
...
+			newASOSecret.Data["AZURE_CLIENT_SECRET"] = identitySecret.Data["clientSecret"]
+		case infrav1.ServicePrincipalCertificate:
+			newASOSecret.Data["AZURE_CLIENT_CERTIFICATE"] = identitySecret.Data["certificate"]
+			newASOSecret.Data["AZURE_CLIENT_CERTIFICATE_PASSWORD"] = identitySecret.Data["password"]
+		case infrav1.UserAssignedMSI:
+			newASOSecret.Data["AZURE_IDENTITY_RESOURCE_ID"] = []byte(asoClientSecret.AzureClusterIdentity.Spec.ResourceID)
+		}
+	} else {
+		newASOSecret.ObjectMeta = metav1.ObjectMeta{
+			Name:      aso.GetASOSecretName(asoClientSecret.ClusterObject.GetName(), asoClientSecret.SubscriptionID),
```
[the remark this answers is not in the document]
adriananeci: Yes, there is a chance to collide. I'll add an extra string while computing secret name based on the owner type(`cluster` or `cluster-identity`) in order to avoid collision in such case

util/aso/defaults.go
adriananeci: I would say it should be, with the caveat mentioned in https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3698#discussion_r1269407349 . Owner can be a cluster if it doesn't specify an AzureClusterIdentity ref or an AzureClusterIdentity if the cluster has defined one.

azure/scope/clients.go
adriananeci: I can try to push the ASO secrets reconciliation in a controller. There are few aspects to keep in mind here when creating/reconciling an ASO secret.
If a a new AzureCluster or ManagedControlPlane object is created we need to check few things:
- if it has an AzureClusterIdentity configured, create an ASO secret having this AzureClusterIdentity as the owner.
- if not, check if the subscriptionID is set. If subscriptionID not set, don't create any ASO secrets. If set and different than the one configured at the capz controller level, create an ASO secret having the cluster as the owner.

The new controller should watch for newly created  AzureCluster or ManagedControlPlane objects and react accordingly.

adriananeci: >Is this something we expect to be shared between multiple clusters or is there one per AzureCluster?

Yes, there might be cases when an ASO secret might be shared between multiple clusters like:
- when they don't have any AzureClusterIdentity configured and have the same subscription ID and it is different than the once configured at the ASO controller level. 
- when they are using the same subscriptionID and reference the same AzureClusterIdentity.

adriananeci:
Not sure why `codecov` and `pull-cluster-api-provider-azure-test` are failing. All tests passed when I ran them locally.

CecileRobertMichon:
this is the error I see in the test job

```
azure/services/aso/aso.go:1: : # sigs.k8s.io/cluster-api-provider-azure/azure/services/aso [sigs.k8s.io/cluster-api-provider-azure/azure/services/aso.test]
azure/services/aso/aso_test.go:698:15: not enough arguments in call to New
	have (client.WithWatch, string)
	want (client.Client, string, string, string)
azure/services/aso/aso_test.go:763:15: not enough arguments in call to New
	have (client.WithWatch, string)
	want (client.Client, string, string, string) (typecheck)
/*
```

adriananeci:
> this is the error I see in the test job
> 
> 
> 
> ```
> 
> azure/services/aso/aso.go:1: : # sigs.k8s.io/cluster-api-provider-azure/azure/services/aso [sigs.k8s.io/cluster-api-provider-azure/azure/services/aso.test]
> 
> azure/services/aso/aso_test.go:698:15: not enough arguments in call to New
> 
> 	have (client.WithWatch, string)
> 
> 	want (client.Client, string, string, string)
> 
> azure/services/aso/aso_test.go:763:15: not enough arguments in call to New
> 
> 	have (client.WithWatch, string)
> 
> 	want (client.Client, string, string, string) (typecheck)
> 
> /*
> 
> ```

I saw that, but it makes no sense to me🤔

nojnhuh:
@adriananeci Try merging the latest main into your branch? There was likely a PR merged since the base of this branch to add tests that haven't been modified in this set of changes and didn't cause a merge conflict.

azure/scope/clients.go
nojnhuh: > If a a new AzureCluster or ManagedControlPlane object is created we need to check few things:

Since these are things the current approach already does, I don't see a reason why doing the same things from a separate controller would necessarily be much different aside from the better architectural consistency that Cecile points out.

But if the ASO secrets are managed from a separate controller, is there a mechanism we can utilize to block the main controller from its normal flow until the ASO secret controller has done its job for a particular cluster? If not, I could see spurious errors coming up if there's a race between the secret controller creating the secret and the normal controller creating a resource referencing that secret.

One option I'm not sure has been said explicitly out loud yet is to never rely on the default ASO credentials and instead always create an ASO secret for each cluster. That would obviously mean creating more secrets than strictly necessary, but I wonder if that tradeoff is worth simplifying the logic from

> * if it has an AzureClusterIdentity configured, create an ASO secret having this AzureClusterIdentity as the owner.
> * if not, check if the subscriptionID is set. If subscriptionID not set, don't create any ASO secrets. If set and different than the one configured at the capz controller level, create an ASO secret having the cluster as the owner.

to something closer to

> * initialize a `ClusterScope` (which handles whether or not an AzureClusterIdentity exists)
> * create an ASO secret having this AzureCluster as the owner from the credentials stored in the `ClusterScope`.

CecileRobertMichon: > But if the ASO secrets are managed from a separate controller, is there a mechanism we can utilize to block the main controller from its normal flow until the ASO secret controller has done its job for a particular cluster

Yes, we could requeue the controller until the secret (or whatever outcome we're waiting for) is ready, something like:  https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/controllers/azuremachine_controller.go#L265. We'd also want to be watching the secrets so that the controller gets requeued automatically when a secret is created (and we could use a watch filter/predicate to only enqueue when it's specifically an ASO secret for that cluster).

That being said I like @nojnhuh's suggestion to simplify the logic. It does seem like the special case handling of "subscription ID is set and is different than the controller one" is a bit of an edge case and we're duplicating logic that's already in `ClusterScope` that handles getting the right credentials based on subscriptionID and identity ref

azure/scope/clients.go
```diff
@@ -169,3 +205,151 @@ func setValue(settings auth.EnvironmentSettings, key string) {
...
+		}
+
+		switch asoClientSecret.AzureClusterIdentity.Spec.Type {
+		case infrav1.ServicePrincipal, infrav1.ManualServicePrincipal:
+			newASOSecret.Data["AZURE_CLIENT_SECRET"] = identitySecret.Data[azureSecretKey]
+		case infrav1.ServicePrincipalCertificate:
+			newASOSecret.Data["AZURE_CLIENT_CERTIFICATE"] = identitySecret.Data["certificate"]
+			newASOSecret.Data["AZURE_CLIENT_CERTIFICATE_PASSWORD"] = identitySecret.Data["password"]
+		case infrav1.UserAssignedMSI:
+			newASOSecret.Data["AZURE_IDENTITY_RESOURCE_ID"] = []byte(asoClientSecret.AzureClusterIdentity.Spec.ResourceID)
```
nojnhuh, line 253: @adriananeci Could you share where you found to set `AZURE_IDENTITY_RESOURCE_ID` here? I can't find that string anywhere in the ASO docs or source code.

azure/services/aso/aso.go
```diff
@@ -176,6 +180,10 @@ func (s *Service) CreateOrUpdateResource(ctx context.Context, spec azure.ASOReso
 		annotations[ReconcilePolicyAnnotation] = ReconcilePolicyManage
 	}
 
+	// Set the secret name annotation in order to leverage the ASO resource credential scope as defined in
+	//https://azure.github.io/azure-service-operator/guide/authentication/credential-scope/#resource-scope.
```
Jont828, line 184:
```suggestion
	// https://azure.github.io/azure-service-operator/guide/authentication/credential-scope/#resource-scope.
```

controllers/asosecret_controller.go
```diff
@@ -0,0 +1,290 @@
...
+
+	// asoSecretOwner is the resource that created the identity. This could be either an AzureCluster or AzureManagedControlPlane (if AKS is enabled).
+	// check for AzureCluster first and if it is not found, check for AzureManagedControlPlane.
+	var asoSecretOwner client.Object
+
+	azureCluster := &infrav1.AzureCluster{}
+	// Fetch the AzureCluster or AzureManagedControlPlane instance
+	asoSecretOwner = azureCluster
+	err := asos.Get(ctx, req.NamespacedName, azureCluster)
+	if err != nil && apierrors.IsNotFound(err) {
```
Jont828, line 118: Nit: could probably reorganize the ifs for consistency with the other controllers as follows

```
if err != nil {
	if err apierrors.IsNotFound(err) {
		...
	}

}
```

controllers/asosecret_controller.go
```diff
@@ -0,0 +1,290 @@
...
+	}
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	var clusterIdentity *corev1.ObjectReference
+	var cluster *clusterv1.Cluster
+	var azureClient scope.AzureClients
+
+	switch ownerType := asoSecretOwner.(type) {
```
Jont828, line 140: TIL you could do that

Jont828 (commented):
Nice work! Overall, this seems reasonable to me. I'd have to defer to @nojnhuh on more specific feedback about the approach.

azure/scope/clients.go
adriananeci: You're right, there is no such thing. Managed identity is handled via Workload Identity or the deprecated aad-pod-identity, https://azure.github.io/azure-service-operator/guide/authentication/credential-format/#deprecated-managed-identity-aad-pod-identity (here it is mentioned about `IDENTITY_RESOURCE_ID`) .

controllers/asosecret_controller.go
adriananeci: Tried to reorganize them a bit, but they still look a bit heavy unfortunately :(

nojnhuh:
I'm not sure I'm seeing this work really. I tried merging my [aso-wired](https://github.com/nojnhuh/cluster-api-provider-azure/tree/aso-wired) branch into this branch and deploying an AzureManagedControlPlane and see this in the status:
```
- lastTransitionTime: "2023-07-28T17:48:00Z"
    message: 'group failed to create or update. err: resource is not Ready: default/jon-aks-aso-secret
      does not exist (credential secret not found: Secret "jon-aks-aso-secret" not
      found). Object will be requeued after 20s'
    reason: Failed
    severity: Error
    status: "False"
    type: ResourceGroupReady
```
I see something similar when creating an AzureCluster.

I verified that the secret doesn't exist and don't see any logs in the capz-controller-manager containing `controllers.ASOSecret`, so I wonder if there's something in `SetupWithManager` that isn't quite configured properly so no requests are coming through.

nojnhuh:
^ ah, I think it's just missing from main.go

adriananeci:
> ^ ah, I think it's just missing from main.go

Yes, sorry, just noticed that while checking what might have been blocking it.
Should be fixed in latest commit

azure/scope/clients.go
adriananeci: Thanks for the feedback. Based on the discussion from this thread, I went ahead and did an overhaul by simplifying and moving the logic inside a dedicated capz controller.

controllers/asosecret_controller.go
```diff
@@ -0,0 +1,290 @@
...
+	)
+	defer done()
+
+	c, err := ctrl.NewControllerManagedBy(mgr).
+		WithOptions(options).
+		For(&infrav1.AzureCluster{}).
+		WithEventFilter(predicates.ResourceNotPausedAndHasFilterLabel(log, asos.WatchFilterValue)).
+		WithEventFilter(predicates.ResourceIsNotExternallyManaged(log)).
+		Named("ASOSecret").
+		Build(asos)
```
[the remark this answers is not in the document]
adriananeci: I think it is a good idea, done.

controllers/asosecret_controller.go
```diff
@@ -0,0 +1,290 @@
...
+			return reconcile.Result{}, errors.Wrap(e, "failed to configure an authentication method for ASO secret")
+		}
+	}
+
+	gvk := asoSecretOwner.GetObjectKind().GroupVersionKind()
+	owner := metav1.OwnerReference{
+		APIVersion: gvk.GroupVersion().String(),
+		Kind:       gvk.Kind,
+		Name:       asoSecretOwner.GetName(),
+		UID:        asoSecretOwner.GetUID(),
```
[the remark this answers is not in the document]
CecileRobertMichon: Don't set which part?

controllers/asosecret_controller_test.go
```diff
@@ -0,0 +1,288 @@
...
+			})
+			if tc.event != "" {
+				g.Expect(reconciler.Recorder.(*record.FakeRecorder).Events).To(Receive(ContainSubstring(tc.event)))
+			}
+			if tc.fail {
+				g.Expect(err).To(MatchError(tc.err))
+			} else {
+				g.Expect(err).NotTo(HaveOccurred())
+			}
+		})
```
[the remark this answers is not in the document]
adriananeci: Makes sense, done.

controllers/asosecret_controller.go
```diff
@@ -0,0 +1,303 @@
...
+	// Return early if the AzureCluster or Cluster is paused.
+	if annotations.IsPaused(cluster, asoSecretOwner) {
+		log.Info(fmt.Sprintf("%s or linked Cluster is marked as paused. Won't reconcile", asoSecretOwner.GetObjectKind()))
+		return ctrl.Result{}, nil
+	}
+
+	// Construct ASO secret for this AzureCluster
+	newASOSecret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      aso.GetASOSecretName(asoSecretOwner.GetName()),
```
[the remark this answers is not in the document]
adriananeci: Makes sense, done.

controllers/asosecret_controller.go
```diff
@@ -0,0 +1,303 @@
...
+			Labels: map[string]string{
+				asoSecretOwner.GetName(): string(infrav1.ResourceLifecycleOwned),
+			},
+		},
+		Data: map[string][]byte{
+			"AZURE_SUBSCRIPTION_ID": []byte(azureClient.SubscriptionID()),
+		},
+	}
+
+	if clusterIdentity != nil {
```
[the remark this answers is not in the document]
adriananeci: Moved it into its own function. Testing is done already via the reconciliation testing

controllers/asosecret_controller_test.go
```diff
@@ -0,0 +1,288 @@
...
+		},
+		"should return if cluster does not exist": {
+			clusterName: defaultAzureCluster.Name,
+			objects: []runtime.Object{
+				defaultAzureCluster,
+			},
+			fail: true,
+			err:  "failed to get Cluster/my-cluster: clusters.cluster.x-k8s.io \"my-cluster\" not found",
+		},
+		"should return if cluster is paused": {
```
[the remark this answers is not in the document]
adriananeci: Added a new event if the cluster is paused and also enhanced the test to check about the event existence

controllers/asosecret_controller_test.go
```diff
@@ -0,0 +1,288 @@
...
+}
+
+func getASOAzureCluster(changes ...func(*infrav1.AzureCluster)) *infrav1.AzureCluster {
+	input := &infrav1.AzureCluster{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "my-azure-cluster",
+			Namespace: "default",
+			OwnerReferences: []metav1.OwnerReference{
+				{
+					APIVersion: "cluster.x-k8s.io/v1beta1",
```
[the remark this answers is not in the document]
adriananeci: done

controllers/asosecret_controller_test.go
```diff
@@ -0,0 +1,288 @@
...
+
+	defaultCluster := getASOCluster()
+	defaultAzureCluster := getASOAzureCluster()
+	defaultAzureManagedControlPlane := getASOAzureManagedControlPlane()
+
+	cases := map[string]struct {
+		clusterName string
+		objects     []runtime.Object
+		fail        bool
+		err         string
```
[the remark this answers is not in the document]
adriananeci: Makes sense, done.

controllers/asosecret_controller.go
```diff
@@ -206,19 +206,46 @@ func (asos *ASOSecretReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 
-	// Construct ASO secret for this AzureCluster
+	// Construct the ASO secret for this Cluster
+	newASOSecret, err := asos.createSecretFromClusterIdentity(ctx, clusterIdentity, cluster, azureClient)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	gvk := cluster.GetObjectKind().GroupVersionKind()
+	owner := metav1.OwnerReference{
```
nojnhuh, line 224: Would there be an issue keeping the AzureCluster/AzureManagedControlPlane as the owner of the Secret, but still name the Secret after the Cluster? I'm not sure it makes any practical difference whether the Secret's owner is the CAPZ resource or the Cluster, but having the AzureCluster/AzureManagedControlPlane as the owner seems slightly tidier to me.

adriananeci: I'm fine either way. I put the Cluster object as the owner initially because the secret label key for owned is also set based on the Cluster name, e.g. :
```
Labels: map[string]string{
    cluster.GetName(): string(infrav1.ResourceLifecycleOwned),
},
```

I've adjusted the secret owner to point to the AzureCluster/AzureManagedControlPlane

controllers/asosecret_controller.go
```diff
@@ -0,0 +1,313 @@
...
+// ASOSecretReconciler reconciles ASO secrets associated with AzureCluster objects.
+type ASOSecretReconciler struct {
+	client.Client
+	Recorder         record.EventRecorder
+	ReconcileTimeout time.Duration
+	WatchFilterValue string
+}
+
+// SetupWithManager initializes this controller with a manager.
+func (asos *ASOSecretReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager, options controller.Options) error {
```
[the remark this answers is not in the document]
CecileRobertMichon: Can we file an issue to follow up on this? @nojnhuh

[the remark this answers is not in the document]
nojnhuh: ^ Seems like the fix is simple enough, so I opened a PR: #3811 

I think I remember trying the same thing with these changes and that not completely fixing things, but I'll look again.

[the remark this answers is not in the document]
nojnhuh: I think I understand what's going on here now. The `Owns()` specifically looks for resources owned by the `For()` type ([code](https://github.com/kubernetes-sigs/controller-runtime/blob/ffb74e54a54bcb10a615c501416820e06051b9b4/pkg/builder/controller.go#L242-L245)), which matches what I was seeing where ASO secrets owned by an AzureCluster reconciled the way I expected but not AzureManagedControlPlane ones. Adding something like this I think will do the same thing for secrets owned by AzureManagedControlPlanes:
```go
	if err = c.Watch(
		&source.Kind{Type: &corev1.Secret{}},
		&handler.EnqueueRequestForOwner{
			OwnerType:    &infrav1.AzureManagedControlPlane{},
			IsController: true,
		},
	); err != nil {
		return errors.Wrap(err, "failed adding a watch for secrets")
	}
```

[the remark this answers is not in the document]
CecileRobertMichon: @nojnhuh is this something that should be added in this PR?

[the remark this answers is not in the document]
nojnhuh: @CecileRobertMichon I'd vote to include it.

/lgtm cancel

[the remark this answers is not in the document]
adriananeci: Done, just included it.(I've also squashed the commits btw)

nojnhuh (commented):
/lgtm

controllers/asosecret_controller.go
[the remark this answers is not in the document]
nojnhuh: > (I've also squashed the commits btw)

Thanks!
/hold cancel

CecileRobertMichon (commented):
/lgtm
/approve

Great work @adriananeci 🎉

PR #3710, opened 2023-07-12
allow AZWI opt-out for e2e tests

nojnhuh:
**What type of PR is this?**
/kind cleanup

**What this PR does / why we need it**:

This PR allows the `AZWI` environment variable to be set to `false` to disable Workload Identity when setting up a kind cluster for e2e tests. The default is still `true` for backwards compatibility.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

- [ ] cherry-pick candidate

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
e2e test invocations now respect the `AZWI=false` environment variable
```

PR #3717, opened 2023-07-13
fix `make kind-create` default for AZWI

nojnhuh:
**What type of PR is this?**
/kind bug

#3710 introduced a slight regression affecting dev workflows which changed the default assumption to Workload Identity being enabled, which breaks things like `make kind-create` and `make tilt-up` when workload ID is not set up. This change reverts to the old defaults, where `make kind-create` will default to disabling workload ID and `make kind-create-bootstrap` (invoked for e2e tests) defaults to enabling workload ID. `$AZWI` can still be set in the environment to explicitly enable or disable workload ID in either case.

Fixes #

- [X] cherry-pick candidate (only for release-1.10)

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -183,7 +183,6 @@ CLUSTER_TEMPLATE ?= cluster-template.yaml
 MANAGED_CLUSTER_TEMPLATE ?= cluster-template-aks.yaml
 
 export KIND_CLUSTER_NAME ?= capz
-export AZWI ?= true
 
 ## --------------------------------------
 ## Binaries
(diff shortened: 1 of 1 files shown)


nojnhuh:
/cherry-pick release-1.10

(this doesn't affect release-v1.9)

Makefile
```diff
@@ -670,7 +669,7 @@ test-cover: test ## Run tests with code coverage and generate reports.
 
 .PHONY: kind-create-bootstrap
 kind-create-bootstrap: $(KUBECTL) ## Create capz kind bootstrap cluster.
-	export KIND_CLUSTER_NAME=capz-e2e && ./scripts/kind-with-registry.sh
+	export AZWI=$${AZWI:-true} KIND_CLUSTER_NAME=capz-e2e && ./scripts/kind-with-registry.sh
```
CecileRobertMichon, line 672: do bash defaults work as expected in make targets? Asking because I don't see any other occurrences in the Makefile

nojnhuh: I tested every permutation of this locally (`AZWI unset/true/false` vs. `make kind-create, kind-create-bootstrap`) and saw what I expected to see each time, so I'm inclined to say yes.

CecileRobertMichon (commented):
/lgtm

Makefile
mboersma: > what is the double $$ doing?

`make` also uses `$` to denote variables, so if you want to pass a variable expression through to the shell for expansion, you have to double-dollar it.

mboersma (approved):
/lgtm
/approve

PR #3630, opened 2023-06-15
add makefile targets to install codespell and verify using codespell

nawazkh:
**What type of PR is this?**
/kind cleanup

**What this PR does / why we need it**:
- Adds two makefile `targets`
  - `make codespell` installs `codespell` binary in the _hack/tools/bin/codespell/_ dir
  - `make verify-codespell` runs `$(CODESPELL) $(ROOT_DIR) --ignore-words=$(ROOT_DIR)/.codespellignore --skip="*.git,*_artifacts,*.sum,$(ROOT_DIR)/hack"` at the root of the codebase. 

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
- `make verify` should help users in correcting codespell errors before raising a PR. 

**Special notes for your reviewer**:
- You may notice that codespell gets installed in `hack/tools/bin/codespell` directory. That is because installing codespell brings along other dependencies. I tried using `pip --no-deps install .. codespell` but could not get rid of other dependencies you see in the picture below.
    Screenshot 2023-06-21 at 2 59 20 PM

- [ ] cherry-pick candidate

**TODOs**:

- [ ] squashed commits
- [x] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

PR #3724, opened 2023-07-13
Run local verifying targets too before git commit

nawazkh:
**What type of PR is this?**
/kind cleanup

- This PR update `pre-commit-config.yaml` to run local Makefile targets as part of pre-commit workflow. 
  - verify-boilerplate
  - verify-modules
  - verify-shellcheck
  - verify-tiltfile
  - verify-codespell

Merge above PR before merging current PRs https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3706 and https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3630 before merging this PR.

- [ ] cherry-pick candidate

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Developer UX: Expand on the pre-commit workflow.
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,3 +20,38 @@ repos:
   rev: v2.17.2
   hooks:
   - id: pylint
+- repo: local
+  hooks:
+  - id: make-verify-boilerplate
+    name: Run make verify-boilerplate
+    description: Runs `make verify-boilerplate` to check for boilerplate headers.
+    entry: make verify-boilerplate
+    language: system
+- repo: local
+  hooks:
+  - id: make-verify-modules
+    name: Run make verify-modules
+    description: Runs `make verify-modules` to verify go.sum go.mod are the latest.
+    entry: make verify-modules
+    language: system
+- repo: local
+  hooks:
+  - id: make-verify-shellcheck
+    name: Run make verify-shellcheck
+    description: Runs `make verify-shellcheck` to verify shell files are passing lint.
+    entry: make verify-shellcheck
+    language: system
+- repo: local
+  hooks:
+  - id: make-verify-tiltfile
+    name: Run make verify-tiltfile
+    description: Runs `make verify-tiltfile` to verify Tiltfile format.
+    entry: make verify-tiltfile
+    language: system
+- repo: local
+  hooks:
+  - id: make-codespell
+    name: Run make verify-codespell
+    description: Runs `make verify-codespell` to verify codespell format.
+    entry: make verify-codespell
+    language: system
--- a/docs/book/src/developers/development.md
+++ b/docs/book/src/developers/development.md
@@ -70,9 +70,10 @@
    - `sudo apt install make` on Linux.
 9. Install [timeout][timeout]
    - `brew install coreutils` on macOS.
-9. Install [pre-commit framework](https://pre-commit.com/#installation)
+10. Install [pre-commit framework](https://pre-commit.com/#installation)
    - `brew install pre-commit` Or `pip install pre-commit`. Installs pre-commit globally.
    - run `pre-commit install` at the root of the project to install pre-commit hooks to read `.pre-commit-config.yaml`
+   - *Note*: use `git commit --no-verify` to skip running pre-commit workflow as and when needed.
 
 When developing on Windows, it is suggested to set up the project on Windows + WSL2 and the file should be checked out on as wsl file system for better results.
 


nawazkh:
We are waiting on two PRs to merge before merging this PR. 
I also need to add `make verify-codespell` to the local pre-commit config before merging this PR.
/hold

Jont828 (commented):
Interesting change! I had a couple things to clarify:

1. Would adding these targets slow down the git commit workflow? It seems like it could get clunky to have the log lines. 
2. Would it allow you to commit changes to your own branch if it fails the check? I know the shellcheck can be strict sometimes and I usually only clean up the `verify` issues when I'm ready for review.

nawazkh:
Good questions! TL;DR, yes it does slow down the workflow bit. 
However, IMO, it's worth the cost..? 
Although, the PR is open for vote.

> * Would adding these targets slow down the git commit workflow? It seems like it could get clunky to have the log lines.

Umm, it is a little subjective. How "slow" would be acceptable?
When everything is good, w.r.t the linters and checks, `git commit` takes around `26 seconds` to complete. 
Below is the output from my local
```
❯ git commit -m "update pre-commit file"
Detect hardcoded secrets.................................................Passed
golangci-lint........................................(no files to check)Skipped
Test shell scripts with shellcheck...................(no files to check)Skipped
fix end of files.........................................................Passed
trim trailing whitespace.................................................Passed
pylint...............................................(no files to check)Skipped
Run make verify-boilerplate..............................................Passed
Run make verify-modules..................................................Passed
Run make verify-shellcheck...............................................Passed
Run make verify-tiltfile.................................................Passed
Run make verify-codespell................................................Passed
[update_pre-commit-config 18988055] add codespell to pre commit config
 1 file changed, 7 insertions(+)
 ```

However, when any of the files is wrong(w.r.t linters) or is being run for the first time, `git commit` could get slow. But it is usually a one-off occurrence.
Output from my local: 
I had an error in the file being committed and the codespell bin had to be installed too since verify-codespell was being run the first time.
It took `49s` to complete.
```
❯ git commit -m "add codespell to pre commit config"
Detect hardcoded secrets.................................................Passed
golangci-lint........................................(no files to check)Skipped
Test shell scripts with shellcheck...................(no files to check)Skipped
fix end of files.........................................................Failed
- hook id: end-of-file-fixer
- exit code: 1
- files were modified by this hook

Fixing .pre-commit-config.yaml

trim trailing whitespace.................................................Passed
pylint...............................................(no files to check)Skipped
Run make verify-boilerplate..............................................Passed
Run make verify-modules..................................................Passed
Run make verify-shellcheck...............................................Passed
Run make verify-tiltfile.................................................Passed
Run make verify-codespell................................................Passed
```

**Comment by nawazkh** [comment]
> * Would it allow you to commit changes to your own branch if it fails the check? I know the shellcheck can be strict sometimes and I usually only clean up the verify issues when I'm ready for review.

That's understandable :) 
You could always use `--no-verify` with `git commit` command to skip running pre-commit configs.

**Comment by mboersma** [comment]
It takes a little while to finish, even with small diffs. It does make me change my usual work pattern to be more deliberate about what I commit, but overall I like the tradeoff because I'll forget to run some of these verify targets otherwise.

But it is a change: people should really try it out and see what they think.

**Comment by mboersma** [comment]
I like these checks overall, and I've gotten used to using `git commit --no-verify` already in the situations where it won't pass (sometimes I need to checkpoint but I know there are test or linter problems, for example).

**Comment by sonasingh46** [comment]
+1 to what @Jont828 says.
Having said that, this can be great way to have pre-commit settings configured into local machines for a lot of contributors but I would too prefer the `git` thingy a bit flexible and not enforce it. IMO, it should be left to individual contributors on how they want to set it up on their dev env.

**Comment by Jont828** [comment]
Maybe a compromise would be to move this to a `.dev/` folder where folks can easily copy it to their pre-commit config? I think I would really prefer this to be an opt-in change.

**Comment by nawazkh** [comment]
Happy to hear your thoughts! :) 

This is an opt-in change rather than an enforcing one. 
This PR merely adds to the already existing pre-commit config added via https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3703.

Contributors should be running `pre-commit install` to enable this workflow in their local dev. If they want to opt out they can run `pre-commit uninstall` and the pre-commit checks would not interfere anymore. 

If someone wants to skip checks on individual commits after installing the pre-commit framework, then one can run `git commit --no-verify` to skip the checks run on the commit.

**Comment by Jont828** [comment]
@nawazkh I think a make target is great idea! Could we add pre-commit changes from #3703 to the make target as well?

**Comment by nawazkh** [comment]
> @nawazkh I think a make target is great idea! Could we add pre-commit changes from #3703 to the make target as well?

That's a good thought! How about I follow it up with another PR? The ones in this PR already have make targets to them.

**Comment by Jont828** [comment]
Another PR sounds good to me. Also, I don't seem to see any Makefile changes in this PR, just the `pre-commit-config.yaml` changes.

File: .pre-commit-config.yaml [hunk 1]
```diff
@@ -20,3 +20,38 @@ repos:
   rev: v2.17.2
   hooks:
   - id: pylint
+- repo: local
```
CecileRobertMichon, line 23: Can we add docs on this in developer docs as well as clear instructions on how to opt out?

nawazkh: Added a note on skipping pre-commit workflow as and when needed. :)

mboersma (commented):
/lgtm

I've been using the pre-commit hook since it merged, and I've tried this expanded version out. It fits my workflow and doesn't add *too* much time to `git commit` IMHO. I use `--no-verify` whenever I need to commit work in progress that I know won't pass these checks.

Jont828:
/lgtm

@nawazkh Just to double check, if you don't run `pre-commit install`, we can opt out of this feature right?

CecileRobertMichon:
> @nawazkh Just to double check, if you don't run pre-commit install, we can opt out of this feature right?

@Jont828 correct, see https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3724#issuecomment-1661276053

/lgtm
/approve

PR #3727, opened 2023-07-14
Added Node Prototyping feature flag

bennycortese:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**: Allows us to develop the new Node Prototyping feature without having anything be on by default while in alpha

- [ ] cherry-pick candidate

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/feature/feature.go
+++ b/feature/feature.go
@@ -50,7 +50,7 @@ const (
 
 	// NodePrototyping is the feature gate for having prototyping mode on
 	// owner: @bennycortese
-	// alpha: v0.1
+	// alpha: v1.11
 	NodePrototyping featuregate.Feature = "NodePrototyping"
 )
 
--- a/feature/feature.go
+++ b/feature/feature.go
@@ -47,6 +47,11 @@ const (
 	// owner: @upxinxin
 	// alpha: v1.8
 	EdgeZone featuregate.Feature = "EdgeZone"
+
+	// NodePrototyping is the feature gate for having prototyping mode on
+	// owner: @bennycortese
+	// alpha: v0.1
+	NodePrototyping featuregate.Feature = "NodePrototyping"
 )
 
 func init() {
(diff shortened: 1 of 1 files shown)


feature/feature.go
```diff
@@ -47,6 +47,11 @@ const (
 	// owner: @upxinxin
 	// alpha: v1.8
 	EdgeZone featuregate.Feature = "EdgeZone"
+
+	// NodePrototyping is the feature gate for having prototyping mode on
+	// owner: @bennycortese
+	// alpha: v0.1
```
nojnhuh, line 53: This should be the CAPZ version where this was first introduced, which we expect to be the next minor release:
```suggestion
	// alpha: v1.11
```

nojnhuh:
For context, I suggested omitting the feature flag from the controller-manager YAML template for now to make sure no user can turn this on while we're working toward making this even minimally functional.

/ok-to-test

Jont828 (commented):
LGTM pending squash

feature/feature.go
```diff
@@ -47,6 +47,11 @@ const (
 	// owner: @upxinxin
 	// alpha: v1.8
 	EdgeZone featuregate.Feature = "EdgeZone"
+
+	// NodePrototyping is the feature gate for having prototyping mode on
+	// owner: @bennycortese
+	// alpha: v1.11
+	NodePrototyping featuregate.Feature = "NodePrototyping"
```
CecileRobertMichon, line 54: how does this get enabled by the user? we might be missing an addition in manager.yaml

nojnhuh: I mentioned my opinion above: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3727#issuecomment-1636217139

CecileRobertMichon: Got it. Would it make sense to still add the environment variable to the yaml template but hardcoding it to `false` for now to make it easier to quickly switch the value for development?

nojnhuh: Would changing the default value below also allow easily toggling it? Even if so, a placeholder for the feature flag hardcoded to false would be a better reminder that we should update it

Issue #3525, opened 2023-05-04
Set `reconcile-policy: skip` on ASO resources when Cluster is paused


nojnhuh:
Since a Cluster's `spec.paused` indicates whether reconciliation is disallowed, CAPZ should communicate that intent with ASO so all reconciliation stops when `spec.paused` is true. Likewise, when a Cluster is unpaused, reset the reconcile-policy annotation to its previous value (reconcile-policy may have already been `skip` even before the Cluster was paused if the corresponding Azure resource is managed by the user).

```[tasklist]
### Dependencies
-  [ ] https://github.com/kubernetes-sigs/cluster-api/issues/8473
```

Note: Since annotating all ASO resources is not an instantaneous operation, CAPZ may be able to perform a best-effort "pause reconciliation of as many ASO resources as possible" even without the above dependency, but it might turn into a bit of a minefield for flakes in tests involving `clusterctl move`.

nojnhuh:
> Note: Since annotating all ASO resources is not an instantaneous operation, CAPZ may be able to perform a best-effort "pause reconciliation of as many ASO resources as possible" even without the above dependency, but it might turn into a bit of a minefield for flakes in tests involving `clusterctl move`.

This is the strategy for now as a result of the tests done in #3795. #3839 will fill the gap down the road. Otherwise the work for this issue is complete.

/close

PR #3733, opened 2023-07-17
refactor ASO label/annotation updates

nojnhuh:
**What type of PR is this?**
/kind cleanup

This PR refactors how labels and annotations applied generically to ASO resources are constructed to enable `delete()`-ing existing labels or annotations which is planned for #3525. Otherwise this change should be functionally equivalent.

Fixes #

- [ ] cherry-pick candidate

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/azure/services/aso/aso.go
+++ b/azure/services/aso/aso.go
@@ -29,7 +29,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
 	"sigs.k8s.io/cluster-api-provider-azure/azure"
-	"sigs.k8s.io/cluster-api-provider-azure/util/maps"
 	"sigs.k8s.io/cluster-api-provider-azure/util/tele"
 	"sigs.k8s.io/cluster-api/util/patch"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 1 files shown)


azure/services/aso/aso.go
```diff
@@ -170,12 +175,14 @@ func (s *Service) CreateOrUpdateResource(ctx context.Context, spec azure.ASOReso
 		annotations[ReconcilePolicyAnnotation] = ReconcilePolicyManage
 	}
 
-	if len(labels) > 0 {
-		parameters.SetLabels(maps.Merge(parameters.GetLabels(), labels))
+	if len(labels) == 0 {
+		labels = nil
```
Jont828, line 179: Why do we want to set `labels` and `annotations` back to nil? Would it make sense instead to call `SetLabels()` and `SetAnnotations()` only when `existing == nil`?

nojnhuh: > Why do we want to set `labels` and `annotations` back to nil?

Because labels and annotations are both serialized with `omitempty`, the `existing` resource will never contain empty, non-nil values for either of those fields. This preemptively emulates `omitempty` for the values of `parameters` so when we calculate the diff just below these changes, we don't treat a non-nil and empty map as different from a nil one since they would both be serialized as nil.

> Would it make sense instead to call `SetLabels()` and `SetAnnotations()` only when `existing == nil`?

Any update (when `existing != nil`) may require labels and/or annotations to be updated (like when we notice a resource needs to be adopted), so I don't think that would work in all cases.

nojnhuh:
(squash reminder)
/hold

CecileRobertMichon (commented):
/assign @willie-yao @Jont828

willie-yao (approved):
/lgtm

nojnhuh:
> (squash reminder) /hold

done!
/hold cancel

CecileRobertMichon (commented):
/approve

PR #3735, opened 2023-07-17
add pause handling for AzureCluster controller

nojnhuh:
**What type of PR is this?**
/kind feature

This PR introduces the capability for CAPZ's Azure service interfaces (i.e. `azure/services/*`) to define logic to be run in reaction to a CAPZ resource being paused. The primary change is the introduction of a new `reconcilePause` at the same level of abstraction as `reconcileNormal` and `reconcileDelete`. 

The currently-identified use case is to set the [serviceoperator.azure.com/reconcile-policy annotation](serviceoperator.azure.com/reconcile-policy) to `skip` on ASO resources to communicate CAPI's notion of "paused" to ASO's controllers.

This change modifies the AzureCluster controller, but all other controllers using the same `azure.ServiceReconciler` pattern will need to modified similarly, at least once they start using ASO.

Part of #3525

This change only partially addresses #3525. Currently, `reconcilePause` will essentially no-op (aside from logging), so this PR is not expected to produce any meaningfully visible changes. Future PRs will implement the logic to add the annotation to ASO resources, handle the [impending block-move CAPI annotation](https://github.com/kubernetes-sigs/cluster-api/pull/8690), and implement the same pattern for other CAPZ controllers.

- [ ] cherry-pick candidate

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/controllers/azurecluster_controller_test.go
+++ b/controllers/azurecluster_controller_test.go
@@ -18,15 +18,20 @@ package controllers
 
 import (
 	"context"
+	"testing"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/record"
 	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
 	"sigs.k8s.io/cluster-api-provider-azure/internal/test"
 	"sigs.k8s.io/cluster-api-provider-azure/util/reconciler"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 )
 
 var _ = Describe("AzureClusterReconciler", func() {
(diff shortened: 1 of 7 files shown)


controllers/helpers.go
```diff
@@ -1023,3 +1026,33 @@ func MachinePoolToAzureManagedControlPlaneMapFunc(ctx context.Context, c client.
 		return nil
 	}
 }
+
+// ClusterUpdatePauseChange returns a predicate that returns true for an update event when a cluster's
+// Spec.Paused changes between any two distinct values.
+func ClusterUpdatePauseChange(logger logr.Logger) predicate.Funcs {
```
nojnhuh, line 1032: This is mostly copy-paste from CAPI's [`ClusterUpdateUnpaused`](https://github.com/kubernetes-sigs/cluster-api/blob/10d44b6ac7603f4aafd72dced33ae2c6915b7f04/util/predicates/cluster_predicates.go#L123).

controllers/azurecluster_reconciler.go
```diff
@@ -98,6 +98,24 @@ func (s *azureClusterService) Reconcile(ctx context.Context) error {
 	return nil
 }
 
+// Pause pauses all components making up the cluster.
+func (s *azureClusterService) Pause(ctx context.Context) error {
+	ctx, _, done := tele.StartSpanWithLogger(ctx, "controllers.azureClusterService.Pause")
+	defer done()
+
+	for _, service := range s.services {
```
nojnhuh, line 106: I don't think the order in which services are paused is significant, at least for ASO.

controllers/azurecluster_controller_test.go
```diff
@@ -51,3 +56,63 @@ var _ = Describe("AzureClusterReconciler", func() {
 		})
 	})
 })
+
+func TestAzureClusterReconcilePaused(t *testing.T) {
```
nojnhuh, line 60: I originally tried to specify this test like the one above with envtest, but the race detector was flagging something in the logging calls whenever two objects were being created, even serially and with a 10s sleep in between. It was easier to reqork this test to use a fake client than try to fix the race which seemed unrelated to the rest of these changes. The test itself hasn't lost any fidelity during that rewrite AFACT.

nojnhuh:
(squash reminder)

/hold

willie-yao (approved):
/lgtm

CecileRobertMichon (commented):
/lgtm
/approve

squash and remove hold?

nojnhuh:
squashed!
/hold cancel

PR #3744, opened 2023-07-20
Machine pool machine scope from ampm

bennycortese:
**What type of PR is this?**
/kind feature

This PR allows for us to quickly construct an MachinePoolMachineScope from our AzureMachinePool reconciler by just passing in an AzureMachinePoolMachine. This is important because functionality has already been written for cordoning and draining the node associated with this scope and getting the node associated with this scope, both of which we don't want to arbitrarily have to rewrite.

- [ ] cherry-pick candidate

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/exp/controllers/azuremachinepool_reconciler.go
+++ b/exp/controllers/azuremachinepool_reconciler.go
@@ -25,6 +25,7 @@ import (
 	"sigs.k8s.io/cluster-api-provider-azure/azure/services/resourceskus"
 	"sigs.k8s.io/cluster-api-provider-azure/azure/services/roleassignments"
 	"sigs.k8s.io/cluster-api-provider-azure/azure/services/scalesets"
+	infrav1exp "sigs.k8s.io/cluster-api-provider-azure/exp/api/v1beta1"
 	"sigs.k8s.io/cluster-api-provider-azure/util/tele"
 )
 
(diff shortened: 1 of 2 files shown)


exp/controllers/azuremachinepool_reconciler.go
```diff
@@ -87,3 +88,17 @@ func (s *azureMachinePoolService) Delete(ctx context.Context) error {
 
 	return nil
 }
+
+func (s *azureMachinePoolService) MachinePoolMachineScopeFromAmpm(ampm *infrav1exp.AzureMachinePoolMachine) *scope.MachinePoolMachineScope {
```
nojnhuh, line 92: Can you help illustrate where this method is going to be used? Either by adding some more (non-breaking) changes to this PR or linking to a WIP branch?

That might help us determine whether or not there might be a better way of doing things that might not involve a method like this.

willie-yao: Agreed. It could also help to add a comment to describe this function!

Jont828: Yeah, in any case we'd want to add a header comment for documenting this function as it's being exported.

@nojnhuh I paired with @bennycortese on this and explored alternatives, but this was the best solution IIRC. I believe the use case was that the AzureMachinePool controller looped through the AzureMachinePoolMachines and needed to call a function in the AMPM scope, which it didn't have access to.

exp/controllers/azuremachinepool_reconciler.go
```diff
@@ -87,3 +88,17 @@ func (s *azureMachinePoolService) Delete(ctx context.Context) error {
...
+	myscope, err := scope.NewMachinePoolMachineScope(scope.MachinePoolMachineScopeParams{
+		Client:                  s.scope.GetClient(),
+		MachinePool:             s.scope.MachinePool,
+		AzureMachinePool:        s.scope.AzureMachinePool,
+		AzureMachinePoolMachine: ampm,
+		ClusterScope:            s.scope.ClusterScoper,
+	})
+	if err != nil {
```
Jont828, line 100: We could probably just replace the whole return part with this since we're not doing anything special with the error.

```suggestion
return scope.NewMachinePoolMachineScope(scope.MachinePoolMachineScopeParams{
		Client:                  s.scope.GetClient(),
		MachinePool:             s.scope.MachinePool,
		AzureMachinePool:        s.scope.AzureMachinePool,
		AzureMachinePoolMachine: ampm,
		ClusterScope:            s.scope.ClusterScoper,
	})
```

PR #3707, opened 2023-07-11
Bump CAPI to v1.5.0

mboersma:
**What type of PR is this?**

/kind cleanup

**What this PR does / why we need it**:

Updates CAPI to [v1.5.0](https://github.com/kubernetes-sigs/cluster-api/releases/tag/v1.5.0) and all that entails.

**Which issue(s) this PR fixes**:

Fixes #3680
Closes #3751
Closes #3773

**Special notes for your reviewer**:

- [ ] cherry-pick candidate

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Bump CAPI to v1.5.0
```

PR #3756, opened 2023-07-25
Bump CAPI to v1.4.5

mboersma:
**What type of PR is this?**

/kind cleanup

Updates CAPI references to use the new [v1.4.5 release](https://github.com/kubernetes-sigs/cluster-api/releases/tag/v1.4.5).

- [x] cherry-pick candidate

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Update CAPI to v1.4.5
```


CecileRobertMichon:
/retest

looks like a flake...

go.mod
```diff
@@ -88,7 +88,7 @@ require (
 	github.com/dimchansky/utfbom v1.1.1 // indirect
 	github.com/docker/cli v20.10.21+incompatible // indirect
 	github.com/docker/distribution v2.8.2+incompatible // indirect
-	github.com/docker/docker v20.10.24+incompatible // indirect
+	github.com/docker/docker v24.0.5+incompatible // indirect
```
CecileRobertMichon, line 91: looks like this might be causing issues?

> /home/prow/go/pkg/mod/oras.land/oras-go@v1.2.2/pkg/auth/docker/login.go:86:39: cannot use remote (variable of type *"github.com/docker/docker/registry".Service) as "github.com/docker/docker/registry".Service value in argument to c.loginWithTLS
/home/prow/go/pkg/mod/oras.land/oras-go@v1.2.2/pkg/auth/docker/login_tls.go:181:30: assignment mismatch: 3 variables but registry.PingV2Registry returns 2 values

mboersma: Yes, there's an incompatibility between newer Docker packages and Helm. CAPI worked around it by not importing Helm since they were only using some constants from the package, but we don't have that option. I've been trying a variety of approaches, but no luck yet. This is also blocking v1.5.0 being integrated.

mboersma: I confirmed that removing our Helm imports would work around this. That's not a trivial amount of work, however: it involves rewriting `InstallHelmChart` to shell out to the `helm` CLI, which gives up a lot of control.

I'll keep investigating this path since I don't know of other options.

Jont828: Would switching the Helm install to CAAPH instead fix the issue? If we're installing it with clusterctl, then I believe we could bypass the need to import Helm directly.

CecileRobertMichon:
I believe we also need to update https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/scope/managedcontrolplane.go#L607 for AKS

This may also apply to the azurejson secrets https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/687e66f74de5099e82748ee73ad439189efed2d2/controllers/helpers.go#L179

see https://github.com/kubernetes-sigs/cluster-api/pull/9080/files#

mboersma:
@CecileRobertMichon that's just for CAPI v1.5, right, not v1.4.5? I'll make those changes in #3707.

CecileRobertMichon:
> @CecileRobertMichon that's just for CAPI v1.5, right, not v1.4.5? I'll make those changes in https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3707.

yep you're totally right, I commented on the wrong PR 🤦‍♀️

go.mod
mboersma: That's definitely the way to go, but since we need this change for CAPI v1.4.5 I think we should merge this first, then apply the CAAPH PR afterward.

mboersma:
/hold

I rebased this on #3757, which needs to merge first.

CecileRobertMichon (commented):
/lgtm
/assign @nojnhuh

nojnhuh (commented):
/lgtm
/approve

Issue #3736, opened 2023-07-17
Use CAAPH to install Helm charts in e2e tests


Jont828:
/kind feature

**Describe the solution you'd like**

Cluster API v1.5 brings support for installing add-on providers using clusterctl. This allows us to use CAAPH in the e2e tests to install cloud-provider-azure, calico, and azuredisk-csi-driver rather than a manual Helm install.

**Environment:**

- cluster-api-provider-azure version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

CecileRobertMichon:
re-opening since only Calico is done

PR #3757, opened 2023-07-25
Use Helm binary in e2e tests to avoid import

mboersma:
**What type of PR is this?**

/kind cleanup

Importing the Helm client in e2e tests conflicts with the new Docker package necessary to work around a Go v1.20.6 issue. This refactors CAPZ's use of Helm not to need those imports, but to use `exec.Command("helm")` instead.

Fixes #3721
Refs #3756, #3707

- [x] cherry-pick candidate

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```


mboersma:
3/4 of these CI failures should be fixed by kubernetes/test-infra#30201.

test/e2e/helm.go
```diff
@@ -0,0 +1,72 @@
...
+
+// HelmOptions handles arguments to a `helm install` command.
+type HelmOptions struct {
+	StringValues []string // --set-string
+	ValueFiles   []string // --values
+	Values       []string // --set
+}
+
+// InstallHelmChart takes a Helm repo URL, a chart name, and release name, and creates a Helm release on the E2E workload cluster.
+func InstallHelmChart(ctx context.Context, clusterProxy framework.ClusterProxy, namespace, repoURL, chartName, releaseName string, options *HelmOptions, version string) {
```
CecileRobertMichon, line 38: The previous implementation has a lot of built-in resiliency with retries and checking if the release already exists before installing which was added over time to address test flakes. Is `helm install` idempotent and somewhat equivalent? Are there any low-hanging fruits to make this new implementation closer to that (eg. skip install if the release already exists, uninstall the release if the install fails, etc.)

A lot of this should be temporary as it will soon be replaced by #3736 so we shouldn't put too much time/effort into this "quick fix" but if there's anything we can do in the meantime to make it not regress stability that might save us some retests.

CecileRobertMichon:

AFAIK these are some of the tests that failed before when trying to install Calico on a cluster that already has it (as part of the upgrade) ^

test/e2e/helm.go
mboersma: Good thinking: `helm upgrade --install` is the idempotent way to install a chart apparently, and `--atomic` should clean things up in case of failure. I'll make those changes.

We're already retrying here the same as before this change. 

The only real difference remaining is that we're not checking to see if the release exists before installing it, but I don't think that's needed.

CecileRobertMichon: > The only real difference remaining is that we're not checking to see if the release exists before installing it, but I don't think that's needed.

I think that is needed in cluster upgrade scenarios where the addon is already installed on the cluster, but `helm upgrade` should cover it?

mboersma: > `helm upgrade` should cover it?

I think so? CI will tell the tale.

mboersma:
After decoding the bytes:

> Error: release cloud-provider-azure-oot failed, and has been uninstalled due to atomic being set: context deadline exceeded

Oops. Let me remove `--atomic`, not sure why that made a difference but the previous implementation wasn't failing here.

CecileRobertMichon:
/retest

ipv6 test should be fixed now

mboersma:
Rebased again because e2e was failing on Kubernetes v1.25.12 (see #3762). Sorry for the testing churn.

mboersma:

🤞🏻

CecileRobertMichon (commented):
/lgtm

pending tests

test/e2e/helm.go
Jont828: I agree, I think it makes to not polish this a ton since it's going to be replaced soon.

Jont828 (commented):
This seems pretty straightforward and I don't see anything wrong. If it passes all the e2e tests, I'm happy to LGTM.

mboersma:
This passed all tests, but I changed it to use `cmd.CombinedOutput()` since we weren't seeing errors printed except as bytes in a stack trace.

test/e2e/helm.go
```diff
@@ -0,0 +1,75 @@
...
```
nojnhuh, line 38: I don't think this `version` parameter is being used at all, but should probably be used to set `--version`.

mboersma: Good catch–I'll update to add `--version` if it's set.

mboersma:
/retest

ILB deletion timeout and a VM creation flake.

CecileRobertMichon (commented):
/approve

sonasingh46 (approved):
/lgtm

PR #3783, opened 2023-07-31
add pause handling for AzureManagedControlPlane

nojnhuh:
**What type of PR is this?**
/kind feature

This PR adds the capability for the AzureManagedControlPlane controller to react when resources are paused using the same pattern as #3735. See that PR for more context.

Part of #3525

- [ ] cherry-pick candidate

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/controllers/azuremanagedcontrolplane_controller.go
+++ b/controllers/azuremanagedcontrolplane_controller.go
@@ -79,7 +79,7 @@ func (amcpr *AzureManagedControlPlaneReconciler) SetupWithManager(ctx context.Co
 	c, err := ctrl.NewControllerManagedBy(mgr).
 		WithOptions(options.Options).
 		For(azManagedControlPlane).
-		WithEventFilter(predicates.ResourceNotPausedAndHasFilterLabel(log, amcpr.WatchFilterValue)).
+		WithEventFilter(predicates.ResourceHasFilterLabel(log, amcpr.WatchFilterValue)).
 		// watch AzureManagedCluster resources
 		Watches(
 			&source.Kind{Type: &infrav1.AzureManagedCluster{}},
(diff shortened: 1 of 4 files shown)


nojnhuh:
(squash reminder)
/hold

controllers/azuremanagedcontrolplane_controller.go
```diff
@@ -209,6 +189,26 @@ func (amcpr *AzureManagedControlPlaneReconciler) Reconcile(ctx context.Context,
...
+	}
+
+	// check if the control plane's namespace is allowed for this identity and update owner references for the identity.
+	if azureControlPlane.Spec.IdentityRef != nil {
+		err := EnsureClusterIdentity(ctx, amcpr.Client, azureControlPlane, azureControlPlane.Spec.IdentityRef, infrav1.ManagedClusterFinalizer)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+	} else {
+		warningMessage := ("You're using deprecated functionality: ")
```
mboersma, line 205: This could be one string literal, which is maybe more readable:

```go
warningMessage := "You're using deprecated functionality: " +
        "Using Azure credentials from the manager environment is deprecated and will be removed in future releases. " +
        "Please specify an AzureClusterIdentity for the AzureManagedControlPlane instead, see: https://capz.sigs.k8s.io/topics/multitenancy.html"
```

mboersma (commented):
/lgtm

(My multiline string comment is just a style thing.) Looks good. I compared it to #3735.

controllers/azuremanagedcontrolplane_controller.go
```diff
@@ -95,12 +95,12 @@ func (amcpr *AzureManagedControlPlaneReconciler) SetupWithManager(ctx context.Co
...
 	}
 
-	// Add a watch on clusterv1.Cluster object for unpause & ready notifications.
+	// Add a watch on clusterv1.Cluster object for pause/unpause & ready notifications.
 	if err = c.Watch(
 		&source.Kind{Type: &clusterv1.Cluster{}},
 		handler.EnqueueRequestsFromMapFunc(amcpr.ClusterToAzureManagedControlPlane),
-		predicates.ClusterUnpausedAndInfrastructureReady(log),
-		predicates.ResourceNotPausedAndHasFilterLabel(log, amcpr.WatchFilterValue),
+		predicates.Any(log, predicates.ClusterCreateInfraReady(log), predicates.ClusterUpdateInfraReady(log), ClusterUpdatePauseChange(log)),
```
willie-yao, line 102: For `predicates.Any()`, doesn't it return true if any of the predicates inside return true? If so, in this case shouldn't we be only returning true if all three are true?

nojnhuh: This was adapted from `ClusterUnpausedAndInfrastructureReady` to essentially remove the `ClusterCreateNotPaused` condition and replace `ClusterUpdateUnpaused` with `ClusterUpdatePauseChange`. https://github.com/kubernetes-sigs/cluster-api/blob/54dfd423d80e42f79c33c0149f9719dba3863f11/util/predicates/cluster_predicates.go#L221

These update predicates only look at one particular change of a resource and not just its updated state, so `ClusterUpdateInfraReady` only passes if the old object's infra is not ready and the updated object's is, and that likely won't correspond to the same update when a Cluster is paused/unpaused, so we need `Any` to check if any of those things happened in a particular update and ignore all other ones.

mboersma (commented):
/lgtm

Pending squash of commits.

nojnhuh:
squashed!
/hold cancel

mboersma (approved):
/approve

PR #3788, opened 2023-08-01
add ASO pause logic

nojnhuh:
**What type of PR is this?**
/kind feature

This PR implements logic to pause ASO resources using the plumbing set up by #3735.

Specifically, the `serviceoperator.azure.com/reconcile-policy` annotation on ASO resources will be set to `skip` and the previous value will also be stored in a CAPZ-specific `sigs.k8s.io/cluster-api-provider-azure-pre-pause-reconcile-policy` annotation. Then when a Cluster is unpaused and reconciliation resumes normally, the `pre-pause-reconcile-policy` annotation will be used to restore the value as it was before the Cluster was paused. The reconcile policy cannot unconditionally be changed back to `manage` because ASO representations of BYO Azure resources (like ones created in the portal or CLI) will already be annotated with the `skip` reconcile policy.

Part of #3525

- [ ] cherry-pick candidate

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/azure/services/aso/aso.go
+++ b/azure/services/aso/aso.go
@@ -37,6 +37,9 @@ import (
 const (
 	// ReconcilePolicyAnnotation is the annotation key for ASO's reconcile policy mechanism.
 	ReconcilePolicyAnnotation = "serviceoperator.azure.com/reconcile-policy"
+	// PrePauseReconcilePolicyAnnotation is the annotation key for the value of
+	// ReconcilePolicyAnnotation that was set before pausing.
+	PrePauseReconcilePolicyAnnotation = "sigs.k8s.io/cluster-api-provider-azure-pre-pause-reconcile-policy"
 	// ReconcilePolicySkip indicates changes will not be PUT or DELETEd in Azure from ASO.
 	ReconcilePolicySkip = "skip"
 	// ReconcilePolicyManage indicates changes will be PUT and DELETEd in Azure from ASO.
@@ -67,7 +70,7 @@ func New(ctrlClient client.Client, clusterName string) *Service {
 // CreateOrUpdateResource implements the logic for creating a new or updating an
 // existing resource with ASO.
 func (s *Service) CreateOrUpdateResource(ctx context.Context, spec azure.ASOResourceSpecGetter, serviceName string) (result genruntime.MetaObject, err error) {
-	ctx, log, done := tele.StartSpanWithLogger(ctx, "aso.Service.CreateOrUpdateResource")
+	ctx, log, done := tele.StartSpanWithLogger(ctx, "services.aso.CreateOrUpdateResource")
 	defer done()
 
 	resource := spec.ResourceRef()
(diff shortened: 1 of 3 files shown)


azure/services/aso/aso.go
```diff
@@ -291,3 +298,46 @@ func IsManaged(ctx context.Context, ctrlClient client.Client, spec azure.ASOReso
 func ownedByCluster(labels map[string]string, clusterName string) bool {
 	return labels[infrav1.OwnedByClusterLabelKey] == clusterName
 }
+
+// PauseResource pauses an ASO resource by updating its `reconcile-policy` to `skip`.
+func PauseResource(ctx context.Context, ctrlClient client.Client, spec azure.ASOResourceSpecGetter, clusterName string, serviceName string) error {
+	ctx, log, done := tele.StartSpanWithLogger(ctx, "aso.Service.Pause")
```
mboersma, line 304:
```suggestion
	ctx, log, done := tele.StartSpanWithLogger(ctx, "services.aso.PauseResource")
```

azure/services/aso/aso.go
```diff
@@ -291,3 +298,46 @@ func IsManaged(ctx context.Context, ctrlClient client.Client, spec azure.ASOReso
...
+	defer done()
+
+	resource := spec.ResourceRef()
+	resourceName := resource.GetName()
+	resourceNamespace := resource.GetNamespace()
+
+	log = log.WithValues("service", serviceName, "resource", resourceName, "namespace", resourceNamespace)
```
mboersma, line 311: Should this also be a `log.V(4)`?

nojnhuh: This doesn't log anything by itself, it only says that future calls to `log.Info/Error` will include those as fields. Since it looks like all the actual log calls below already use `.V(4)`, I suppose we could include that here. Then we'd need to refactor as soon as we change this to log something at a different verbosity though, so I think it might be a bit better to specify the verbosity on each log line.

nojnhuh:
(squash reminder)
/hold

azure/services/aso/aso.go
mboersma: Ah nevermind, I read too fast and thought this was an actual logging call, not a setup func.

mboersma (commented):
/lgtm

nojnhuh:
Pushed one tiny `pointer` -> `ptr` change.

to squash:
/hold

azure/services/aso/aso.go
```diff
@@ -37,6 +37,9 @@ import (
 const (
 	// ReconcilePolicyAnnotation is the annotation key for ASO's reconcile policy mechanism.
 	ReconcilePolicyAnnotation = "serviceoperator.azure.com/reconcile-policy"
+	// PrePauseReconcilePolicyAnnotation is the annotation key for the value of
+	// ReconcilePolicyAnnotation that was set before pausing.
+	PrePauseReconcilePolicyAnnotation = "sigs.k8s.io/cluster-api-provider-azure-pre-pause-reconcile-policy"
```
sonasingh46, line 42: Why is the naming pattern of the annotation key different for `ReconcilePolicyAnnotation` and `PrePauseReconcilePolicyAnnotation` i.e `serviceoperator.azure.com` vs `sigs.k8s.io`?

sonasingh46 (approved):
/lgtm 
( Just one minor question )

mboersma (approved):
/lgtm
/approve

nojnhuh:
squashed!
/hold cancel

azure/services/aso/aso.go
sonasingh46: Makes sense. Thank you

PR #3789, opened 2023-08-01
Fix retrying in Helm e2e helper func

### LINKED ISSUE #3787: Helm chart e2e helper not retrying
/kind bug

#3757 merged but doesn't appear to be using the retry logic implied by Ginkgo's `Eventually` construct. It's been fairly reliable, but it appears when there's a single network hiccup on `helm install` the test fails.

The command should be retried every `retryableOperationSleepBetweenRetries` (3 seconds) for `helmInstallTimeout` (3 minutes).
 
**Environment:**

- cluster-api-provider-azure version: main

mboersma:
**What type of PR is this?**

/kind failing-test

Fixes a logic error in the refactoring of the `InstallHelmChart` helper function so it retries as intended if the command returns an error.

Fixes #3787

I created a demo Ginkgo test suite to reproduce this and isolate the problem, then ported the now-obvious fix back to CAPZ. I did a before-and-after test locally with `./scripts/ci-e2e.sh` and verified that we aren't currently retrying but that this change fixes that.

I guess the good news is that even without retry, the new implementation has only rarely failed. 😄 

- [x] cherry-pick candidate

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/helm.go
+++ b/test/e2e/helm.go
@@ -68,11 +68,11 @@ func InstallHelmChart(ctx context.Context, clusterProxy framework.ClusterProxy, 
 	}
 
 	// Install the chart and retry if needed
-	Eventually(func(g Gomega) {
+	Eventually(func() error {
 		cmd := exec.Command(helm, args...)
 		Logf("Helm command: %s", cmd.String())
 		output, err := cmd.CombinedOutput()
 		Logf("Helm install output: %s", string(output))
-		Expect(err).NotTo(HaveOccurred())
+		return err
 	}, helmInstallTimeout, retryableOperationSleepBetweenRetries).Should(Succeed())
 }


test/e2e/helm.go
```diff
@@ -68,11 +68,11 @@ func InstallHelmChart(ctx context.Context, clusterProxy framework.ClusterProxy,
 	}
 
 	// Install the chart and retry if needed
-	Eventually(func(g Gomega) {
+	Eventually(func() error {
 		cmd := exec.Command(helm, args...)
 		Logf("Helm command: %s", cmd.String())
 		output, err := cmd.CombinedOutput()
 		Logf("Helm install output: %s", string(output))
-		Expect(err).NotTo(HaveOccurred())
```
nojnhuh, line 76: FWIW, I think `g.Expect(...)` would have also fixed this issue.

nojnhuh (approved):
/lgtm
/approve

PR #3793, opened 2023-08-02
Require AzureClusterIdentity for auth

willie-yao:
**What type of PR is this?**
/kind cleanup

Fixes #

- [ ] cherry-pick candidate

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Require AzureClusterIdentity for auth
```


CecileRobertMichon:
@willie-yao thoughts on splitting this into two PRs:

- Require AzureClusterIdentity
- Update virtualmachineimages service to SDK v2
?

willie-yao:
@CecileRobertMichon That definitely makes more sense. I'll keep this one to just requiring AzureClusterIdentity once I'm done testing workload identity!

CecileRobertMichon:
@willie-yao sounds good

@mtougeron made a good point in office hours today that we should also audit the docs to make sure nothing is referring to the old way of setting creds (basically that sets the environment variables as base64 encoded), eg. https://capz.sigs.k8s.io/topics/getting-started#warning-1

willie-yao:
TODO: Add webhook validation for missing AzureClusterIdentities

sonasingh46:
Looks like the unit test is failing.

nojnhuh:
@willie-yao The new ASO secret controller could also be updated to assume the AzureClusterIdentity exists. I think that would only involve changing this function to replace the whole if-else with the contents of the `if`: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/a0dcdea1b8168babca3fc8935345a8026f89d568/controllers/asosecret_controller.go#L253

You may need to merge main into this branch or rebase to see that.

nojnhuh:
@willie-yao Also the release note looks like it could be updated with the change in scope of this PR.

api/v1beta1/azurecluster_validation.go
```diff
@@ -136,12 +141,24 @@ func validateBastionSpec(bastionSpec BastionSpec, fldPath *field.Path) *field.Er
 	return nil
 }
 
+// validateIdentityRef validates an IdentityRef.
+func validateIdentityRef(identityRef *corev1.ObjectReference, fldPath *field.Path) *field.Error {
+	if identityRef == nil {
+		return field.NotFound(fldPath, "identityRef is required")
```
CecileRobertMichon, line 147: this should use `field.Required`

api/v1beta1/azurecluster_validation.go
```diff
@@ -136,12 +141,24 @@ func validateBastionSpec(bastionSpec BastionSpec, fldPath *field.Path) *field.Er
...
+	}
+	if identityRef.Kind != "AzureClusterIdentity" {
+		return field.Invalid(fldPath.Child("name"), identityRef.Name,
```
CecileRobertMichon, line 150: nit: `field.NotSupported` is slightly more appropriate IMO

> // NotSupported returns a *Error indicating "unsupported value".
// This is used to report unknown values for enumerated fields (e.g. a list of
// valid values).

> // Invalid returns a *Error indicating "invalid value".  This is used
// to report malformed values (e.g. failed regex match, too long, out of bounds).

controllers/asosecret_controller.go
```diff
@@ -300,25 +300,6 @@ func (asos *ASOSecretReconciler) createSecretFromClusterIdentity(ctx context.Con
 			newASOSecret.Data["AZURE_CLIENT_CERTIFICATE"] = identitySecret.Data["certificate"]
 			newASOSecret.Data["AZURE_CLIENT_CERTIFICATE_PASSWORD"] = identitySecret.Data["password"]
 		}
-	} else {
```
CecileRobertMichon, line 303: should we remove the `if` altogether here?

CecileRobertMichon (commented):
Looks good, just a few minor comments from my side

controllers/asosecret_controller.go
willie-yao: Removing the `if` caused a nil pointer dereference with the asosecret controller reconcile test if IdentityRef is not configured. Although we always expect IdentityRef to be populated at this point due to the validating webhook, would it still be safer to include this `if`, or just removing it along with the unit tests?

Test: https://github.com/willie-yao/cluster-api-provider-azure/blob/require-azureclusteridentity/controllers/asosecret_controller_test.go#L65

CecileRobertMichon: I'm +1 for removing it since there is a validation in place to enforce it

If it's nil other things in the controller will fail later. We already assume other required things in this function aren't nil, like `cluster`

mboersma (approved):
/lgtm
/approve

PR #3799, opened 2023-08-04
Make scalesetvms delete async

### LINKED ISSUE #2720: async scalesetvms refactor

Jont828:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**: Implementation of an async service for managed agent pools as part of an effort to make all services async. See #1610 and #1541.

Fixes #2720

- [ ] cherry-pick candidate

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Make scalesetvms delete async
```

### DIFF
--- a/azure/scope/machinepoolmachine.go
+++ b/azure/scope/machinepoolmachine.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
@@ -32,7 +33,9 @@ import (
 	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
 	"sigs.k8s.io/cluster-api-provider-azure/azure"
 	"sigs.k8s.io/cluster-api-provider-azure/azure/converters"
+	"sigs.k8s.io/cluster-api-provider-azure/azure/services/scalesetvms"
 	infrav1exp "sigs.k8s.io/cluster-api-provider-azure/exp/api/v1beta1"
+	azureutil "sigs.k8s.io/cluster-api-provider-azure/util/azure"
 	"sigs.k8s.io/cluster-api-provider-azure/util/futures"
 	"sigs.k8s.io/cluster-api-provider-azure/util/tele"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
@@ -149,6 +152,24 @@ func NewMachinePoolMachineScope(params MachinePoolMachineScopeParams) (*MachineP
 	}, nil
 }
 
+// ScaleSetVMSpec returns the VMSS VM spec.
+func (s *MachinePoolMachineScope) ScaleSetVMSpec() azure.ResourceSpecGetter {
+	spec := &scalesetvms.ScaleSetVMSpec{
+		Name:          s.Name(),
+		InstanceID:    s.InstanceID(),
+		ResourceGroup: s.ResourceGroup(),
+		ScaleSetName:  s.ScaleSetName(),
+		ProviderID:    s.ProviderID(),
+		IsFlex:        s.OrchestrationMode() == infrav1.FlexibleOrchestrationMode,
+	}
+
+	if spec.IsFlex {
+		spec.ResourceID = strings.TrimPrefix(spec.ProviderID, azureutil.ProviderIDPrefix)
+	}
+
+	return spec
+}
+
 // Name is the name of the Machine Pool Machine.
 func (s *MachinePoolMachineScope) Name() string {
 	return s.AzureMachinePoolMachine.Name
@@ -224,6 +245,13 @@ func (s *MachinePoolMachineScope) UpdatePatchStatus(condition clusterv1.Conditio
 // SetVMSSVM update the scope with the current state of the VMSS VM.
 func (s *MachinePoolMachineScope) SetVMSSVM(instance *azure.VMSSVM) {
 	s.instance = instance
+}
+
+// SetVMSSVMState update the scope with the current provisioning state of the VMSS VM.
+func (s *MachinePoolMachineScope) SetVMSSVMState(state infrav1.ProvisioningState) {
+	if s.instance != nil {
+		s.instance.State = state
+	}
 }
 
 // ProvisioningState returns the AzureMachinePoolMachine provisioning state.
--- a/azure/services/scalesetvms/client.go
+++ b/azure/services/scalesetvms/client.go
@@ -18,42 +18,27 @@ package scalesetvms
 
 import (
 	"context"
-	"encoding/base64"
-	"encoding/json"
-	"time"
 
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2021-11-01/compute"
 	"github.com/Azure/go-autorest/autorest"
-	"github.com/pkg/errors"
+	azureautorest "github.com/Azure/go-autorest/autorest/azure"
 	"k8s.io/utils/ptr"
-	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
 	"sigs.k8s.io/cluster-api-provider-azure/azure"
-	"sigs.k8s.io/cluster-api-provider-azure/azure/converters"
+	"sigs.k8s.io/cluster-api-provider-azure/util/reconciler"
 	"sigs.k8s.io/cluster-api-provider-azure/util/tele"
 )
 
 // client wraps go-sdk.
 type client interface {
-	Get(context.Context, string, string, string) (compute.VirtualMachineScaleSetVM, error)
-	GetResultIfDone(ctx context.Context, future *infrav1.Future) (compute.VirtualMachineScaleSetVM, error)
-	DeleteAsync(context.Context, string, string, string) (*infrav1.Future, error)
+	Get(context.Context, azure.ResourceSpecGetter) (interface{}, error)
+	DeleteAsync(context.Context, azure.ResourceSpecGetter) (azureautorest.FutureAPI, error)
+	IsDone(ctx context.Context, future azureautorest.FutureAPI) (isDone bool, err error)
 }
 
-type (
-	// azureClient contains the Azure go-sdk Client.
-	azureClient struct {
-		scalesetvms compute.VirtualMachineScaleSetVMsClient
-	}
-
-	genericScaleSetVMFuture interface {
-		DoneWithContext(ctx context.Context, sender autorest.Sender) (done bool, err error)
-		Result(client compute.VirtualMachineScaleSetVMsClient) (vmss compute.VirtualMachineScaleSetVM, err error)
-	}
-
-	deleteFutureAdapter struct {
-		compute.VirtualMachineScaleSetVMsDeleteFuture
-	}
-)
+// azureClient contains the Azure go-sdk Client.
+type azureClient struct {
+	scalesetvms compute.VirtualMachineScaleSetVMsClient
+}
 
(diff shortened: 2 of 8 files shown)


Jont828:
@CecileRobertMichon @mboersma Should be ready for review, PTAL!

azure/services/scalesetvms/scalesetvms.go
```diff
@@ -73,211 +73,95 @@ func (s *Service) Reconcile(ctx context.Context) error {
...
-		return nil
+		reconciler = s.VMReconciler
+	} else {
+		log.V(4).Info("VMSS is uniform", "vmssName", scaleSetVMSpec.Name, "providerID", scaleSetVMSpec.ProviderID, "instanceID", scaleSetVMSpec.InstanceID)
 	}
 
-	log.V(4).Info("VMSS is uniform", "vmssName", vmssName, "providerID", providerID, "instanceID", instanceID)
-	// Using VMSS Uniform, so fetch by instance ID.
-	instance, err := s.Client.Get(ctx, resourceGroup, vmssName, instanceID)
+	// Because getter.Parameters() returns nil, this will only return the existing resource if it exists.
```
mboersma, line 98: I couldn't quite understand this comment. Is there another way to explain it?

azure/scope/machinepoolmachine.go
```diff
@@ -149,6 +152,24 @@ func NewMachinePoolMachineScope(params MachinePoolMachineScopeParams) (*MachineP
 	}, nil
 }
 
+// ScaleSetVMSpec returns the VMSS VM spec.
+func (s *MachinePoolMachineScope) ScaleSetVMSpec() azure.ResourceSpecGetter {
```
mboersma, line 156: Should we add a unit test for this and `SetVMSSVMState` below? They're far from complicated but it seems they don't have coverage.

azure/services/scalesetvms/spec.go
```diff
@@ -0,0 +1,78 @@
...
+	ScaleSetName  string
+	ProviderID    string
+	ResourceID    string
+	IsFlex        bool
+}
+
+// ResourceName returns the name of the VMSS VM.
+func (s *ScaleSetVMSpec) ResourceName() string {
+	return s.InstanceID
+}
```
mboersma, line 37: Should add some tests (or assertions to existing tests) for `ScaleSetVMSpec` and `VMSSFlexGetter`?

azure/services/scalesetvms/scalesetvms.go
Jont828: In `CreateOrUpdateResource()` in async.go, we attempt to get the existing resource and get the return value of `Parameters()`. If `Parameters()` returns `nil`, we return the existing resource if it exists, and if it is not `nil` we attempt to create or update it. By forcing `Parameters()` to return `nil`, we're never going to create or update and just using it as a getter. See [this line](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/a0dcdea1b8168babca3fc8935345a8026f89d568/azure/services/async/async.go#L129).

azure/services/scalesetvms/spec.go
```diff
@@ -0,0 +1,78 @@
...
+	ResourceGroup string
+	ScaleSetName  string
+	ProviderID    string
+	ResourceID    string
+	IsFlex        bool
+}
+
+// ResourceName returns the name of the VMSS VM.
+func (s *ScaleSetVMSpec) ResourceName() string {
+	return s.InstanceID
```
Jont828, line 36: @CecileRobertMichon I ran into a bit of an awkward situation trying to get this to work. Although we have a `Name` field, the uniform VMSS instances are identified by the `InstanceID`, not the `Name` field. WDYT about this work around?

willie-yao:
Hey @Jont828, is this PR still blocked? If not, I'll be happy to review it!

Jont828:
@willie-yao I think I got confused, I meant that this PR is blocked because I need reviews. Just changed the label to "needs review!"

azure/services/scalesetvms/client.go
```diff
@@ -74,79 +59,56 @@ func newVirtualMachineScaleSetVMsClient(subscriptionID string, baseURI string, a
...
-	return ac.scalesetvms.Get(ctx, resourceGroupName, vmssName, instanceID, "")
+	return ac.scalesetvms.Get(ctx, spec.ResourceGroupName(), spec.OwnerResourceName(), spec.ResourceName(), "")
 }
 
-// GetResultIfDone fetches the result of a long-running operation future if it is done.
-func (ac *azureClient) GetResultIfDone(ctx context.Context, future *infrav1.Future) (compute.VirtualMachineScaleSetVM, error) {
-	ctx, _, spanDone := tele.StartSpanWithLogger(ctx, "scalesetvms.azureClient.GetResultIfDone")
-	defer spanDone()
+// CreateOrUpdateAsync is a dummy implementation to fulfill the async.Reconciler interface.
+func (ac *azureClient) CreateOrUpdateAsync(ctx context.Context, spec azure.ResourceSpecGetter, parameters interface{}) (result interface{}, future azureautorest.FutureAPI, err error) {
```
willie-yao, line 70: Just for my own understanding: Is the Creating/Updating handled by the MachinePool service and that's why no implementation of CreateOrUpdateAsync is needed?

azure/services/scalesetvms/scalesetvms.go
```diff
@@ -73,211 +73,95 @@ func (s *Service) Reconcile(ctx context.Context) error {
...
+	var err error
 	// Fetch the latest instance or VM data. AzureMachinePoolReconciler handles model mutations.
-	if isFlex {
-		resourceID := strings.TrimPrefix(providerID, azureutil.ProviderIDPrefix)
-		log.V(4).Info("VMSS is flex", "vmssName", vmssName, "providerID", providerID, "resourceID", resourceID)
-		// Using VMSS Flex, so fetch by resource ID.
-		vm, err := s.VMClient.GetByID(ctx, resourceID)
+	if scaleSetVMSpec.IsFlex {
+		log.V(4).Info("VMSS is flex", "vmssName", scaleSetVMSpec.Name, "providerID", scaleSetVMSpec.ProviderID, "resourceID", scaleSetVMSpec.ResourceID)
+		getter, err = scaleSetVMSpecToVMSpec(*scaleSetVMSpec)
```
willie-yao, line 89: With the change to async, should we still be fetching the vm by resource ID if it's using Flex?

willie-yao (commented):
This looks great! Just some comments mostly for my own understanding

azure/services/scalesetvms/scalesetvms.go
willie-yao: Oh I see that it has been moved within the helper function here, got it! 
https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3799/files?diff=split&w=0#diff-afa3aa43d3541f80b10af246e174440ced390fc557b6a907525183d4d9dd5106R157

Jont828:
@willie-yao @CecileRobertMichon Can we LGTM or is there anything else I should change?

CecileRobertMichon:
@Jont828 I think we were mostly waiting on #3875 to be fixed (it now is) 

Aside from that, I see a few unresolved threads still, so we still need to resolve those. I'll try to give this another round of review later today.

azure/services/scalesetvms/client.go
Jont828: Yep!

azure/services/scalesetvms/client.go
```diff
@@ -74,79 +59,56 @@ func newVirtualMachineScaleSetVMsClient(subscriptionID string, baseURI string, a
...
 
-	return ac.scalesetvms.Get(ctx, resourceGroupName, vmssName, instanceID, "")
+	return ac.scalesetvms.Get(ctx, spec.ResourceGroupName(), spec.OwnerResourceName(), spec.ResourceName(), "")
 }
 
-// GetResultIfDone fetches the result of a long-running operation future if it is done.
-func (ac *azureClient) GetResultIfDone(ctx context.Context, future *infrav1.Future) (compute.VirtualMachineScaleSetVM, error) {
-	ctx, _, spanDone := tele.StartSpanWithLogger(ctx, "scalesetvms.azureClient.GetResultIfDone")
-	defer spanDone()
+// CreateOrUpdateAsync is a dummy implementation to fulfill the async.Reconciler interface.
```
CecileRobertMichon, line 69: you shouldn't need to do this, you should be able to do ` async.New(scope, nil, client),` when creating the service in scalesetvms.go so this client doesn't need to implement the Deleter interface

check out the disks client for a similar example

azure/services/scalesetvms/spec.go
CecileRobertMichon: I think this is fine, however can we add a code comment to explain?

azure/services/scalesetvms/scalesetvms.go
CecileRobertMichon: +1 on clarifying the comment here to make it clear that `CreateOrUpdateResource` never actually creates or updates the scale set vm

azure/services/scalesetvms/client.go
Jont828: If I understand correctly, in the disks service we don't attempt to call `CreateOrUpdateResource()` so we can pass in a nil to the Creator interface. In this service, we want to get the resource if it exists and so I'm trying to leverage the `CreateOrUpdateResource()` in the async interface to fetch the resource and handle the not found/transient errors. Alternatively, I could try to construct the client as well, but I felt that it would be clunky to declare a Reconciler and a client for each type of VM.

CecileRobertMichon: Now that I'm thinking about this more, I wonder if it might make more sense to add a `Getter` (https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/services/async/interfaces.go#L41)  in the Service and simply call `getter.Get()` in the `Reconcile` func instead of using `CreateOrUpdateResource` in a way it's not meant to be used.

azure/services/scalesetvms/spec.go
mboersma: Meh, taking a second look I don't think this level of unit test is productive.

azure/services/scalesetvms/client.go
Jont828: My thinking was to be able to take advantage of the error handling in [this block](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/c20099b2b691143f1034ef8b9309ed932c8e1b7e/azure/services/async/async.go#L116-L123) of `CreateOrUpdateResource()` so we don't have to duplicate that logic. Maybe we could add a `Get()` function to the `Reconciler` iface to give us the error handling?

CecileRobertMichon: @mboersma what do you think? I'm concerned this won't work the same way with the async poller framework when we try to migrade scaleset vms to sdk v2

mboersma: Actually, I think this will act the same way in `asyncpoller`; the `CreateOrUpdateResource` code isn't really changed. I think we can merge this as-is and it won't present problems for refactoring SDKv2 on top of it.

mboersma (approved):
/lgtm
/approve

I also wanted to see that VMSS Flex still works and it looks like `-e2e-optional` passed. 👍🏻

PR #3808, opened 2023-08-07
add pause handling for AzureMachine, AzureManagedMachinePool, AzureMachinePool

nojnhuh:
**What type of PR is this?**
/kind feature

This PR follows up from #3735 and #3783 to set up the remaining controllers to handle pausing ASO resources when the parent Cluster is paused. These changes aren't strictly necessary until more services are re-implemented to use ASO, but I figured it would be worth merging ahead of time so we don't have to remember to do that later.

Part of #3525

- [ ] cherry-pick candidate

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/controllers/azuremachine_reconciler.go
+++ b/controllers/azuremachine_reconciler.go
@@ -43,6 +43,7 @@ type azureMachineService struct {
 	services  []azure.ServiceReconciler
 	skuCache  *resourceskus.Cache
 	Reconcile func(context.Context) error
+	Pause     func(context.Context) error
 	Delete    func(context.Context) error
 }
 
@@ -68,6 +69,7 @@ func newAzureMachineService(machineScope *scope.MachineScope) (*azureMachineServ
 		skuCache: cache,
 	}
 	ams.Reconcile = ams.reconcile
+	ams.Pause = ams.pause
 	ams.Delete = ams.delete
 
 	return ams, nil
@@ -75,7 +77,7 @@ func newAzureMachineService(machineScope *scope.MachineScope) (*azureMachineServ
 
 // reconcile reconciles all the services in a predetermined order.
 func (s *azureMachineService) reconcile(ctx context.Context) error {
-	ctx, _, done := tele.StartSpanWithLogger(ctx, "controllers.azureMachineService.Reconcile")
+	ctx, _, done := tele.StartSpanWithLogger(ctx, "controllers.azureMachineService.reconcile")
 	defer done()
 
 	// Ensure that the deprecated networking field values have been migrated to the new NetworkInterfaces field.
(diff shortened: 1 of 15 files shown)


nojnhuh:
(squash reminder)
/hold

controllers/azuremachine_reconciler.go
```diff
@@ -94,6 +96,24 @@ func (s *azureMachineService) reconcile(ctx context.Context) error {
 	return nil
 }
 
+// pause pauses all components making up the machine.
+func (s *azureMachineService) pause(ctx context.Context) error {
+	ctx, _, done := tele.StartSpanWithLogger(ctx, "controllers.azureMachineService.Pause")
```
mboersma, line 101:
```suggestion
	ctx, _, done := tele.StartSpanWithLogger(ctx, "controllers.azureMachineService.pause")
```

controllers/helpers.go
```diff
@@ -1067,3 +1068,9 @@ func getCertificateFromFile(certificateFilePath string) ([]byte, error) {
 
 	return os.ReadFile(certificateFilePathTrimmed)
 }
+
+// ClusterPauseChangeAndInfrastructureReady is based on ClusterUnpausedAndInfrastructureReady, but
+// additionally accepts Cluster pause events.
+func ClusterPauseChangeAndInfrastructureReady(log logr.Logger) predicate.Funcs {
```
mboersma, line 1074: I realize this is a trivial function, but should it have a unit test anyway?

controllers/azuremachine_reconciler.go
nojnhuh: I noticed this was capitalized for both `reconcile` and `delete` as well. Should I update those to be lowercase too?

controllers/helpers.go
nojnhuh: I see CAPI doesn't really have many specific tests for these, but I'll see what I can come up with.

controllers/azuremachine_reconciler.go
mboersma: I think so, because aren't we trying to identify the literal name of the function so we can match it up when looking at telemetry?

It's a bit confusing here because you're actually assigning `pause` to `ams.Pause`, and the exported name feels like it should be more visible...or something. But I think the literal name is correct here.

controllers/helpers_test.go
```diff
@@ -1310,3 +1311,214 @@ func Test_ManagedMachinePoolToInfrastructureMapFunc(t *testing.T) {
 		})
 	}
 }
+
+func TestClusterPauseChangeAndInfrastructureReady(t *testing.T) {
```
mboersma, line 1315: 💚

mboersma (commented):
/lgtm

CecileRobertMichon (commented):
/lgtm
/approve

nojnhuh:
squashed!
/hold cancel

PR #3813, opened 2023-08-08
Add ARM client options func for SDKv2

mboersma:
**What type of PR is this?**

/kind feature

Adds the `ARMClientOptions()` factory func to standardize default options used in services that get converted to use Azure SDK for Go v2. This mimics the SDK v1 behavior in `SetAutoRestClientDefaults()`.

Refs #2974

- [ ] cherry-pick candidate

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/azure/defaults_test.go
+++ b/azure/defaults_test.go
@@ -24,10 +24,108 @@ import (
 	"sync"
 	"testing"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
 	"github.com/Azure/go-autorest/autorest"
 	. "github.com/onsi/gomega"
 	"sigs.k8s.io/cluster-api-provider-azure/util/tele"
 )
+
+// TestARMClientOptions tests the `ARMClientOptions()` factory function.
+func TestARMClientOptions(t *testing.T) {
+	tests := []struct {
+		name          string
+		cloudName     string
+		expectedCloud cloud.Configuration
+		expectError   bool
+	}{
+		{
+			name:          "should return default client options if cloudName is empty",
+			cloudName:     "",
+			expectedCloud: cloud.Configuration{},
+		},
+		{
+			name:          "should return Azure public cloud client options",
+			cloudName:     PublicCloudName,
+			expectedCloud: cloud.AzurePublic,
+		},
+		{
+			name:          "should return Azure China cloud client options",
+			cloudName:     ChinaCloudName,
+			expectedCloud: cloud.AzureChina,
+		},
+		{
+			name:          "should return Azure government cloud client options",
+			cloudName:     USGovernmentCloudName,
+			expectedCloud: cloud.AzureGovernment,
+		},
+		{
+			name:        "should return error if cloudName is unrecognized",
+			cloudName:   "AzureUnrecognizedCloud",
+			expectError: true,
+		},
+	}
+	for _, tc := range tests {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


azure/defaults_test.go
```diff
@@ -24,11 +24,99 @@ import (
...
+
+	corrID := "test-1234abcd-5678efgh"
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		g.Expect(r.Header.Get("User-Agent")).To(ContainSubstring("cluster-api-provider-azure/"))
+		g.Expect(r.Header.Get(string(tele.CorrIDKeyVal))).To(Equal(corrID))
+		fmt.Fprintf(w, "Hello, %s", r.Proto)
+	}))
+	defer server.Close()
+
+	opts, err := ARMClientOptions("")
```
Jont828, line 98: Could you break this up with some newlines and add comments explaining what we're trying to do here?

azure/defaults_test.go
```diff
@@ -24,11 +24,99 @@ import (
...
+	g.Expect(err).NotTo(HaveOccurred())
+	g.Expect(opts.PerCallPolicies).To(HaveLen(2))
+	ctx := context.WithValue(context.Background(), tele.CorrIDKeyVal, tele.CorrID(corrID))
```
Jont828, line 101: Just curious, why do we need a custom context here?

Jont828 (commented):
LGTM pending small nit.

azure/defaults_test.go
mboersma: This is actually how the `correlationIDPolicy` works: it looks for a Microsoft "correlation ID" in the context of the request object. 

So here we're setting up what ends up as a request "decorator" for the test that verifies the behavior of the `correlationIDPolicy`.

In SDKv1 it's done this way for comparison:
https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/a0dcdea1b8168babca3fc8935345a8026f89d568/azure/defaults.go#L336-L339

azure/defaults_test.go
mboersma: Done. Let me know if that's more readable, or if there's more I could do.

azure/defaults_test.go
```diff
@@ -24,11 +24,109 @@ import (
...
+		g.Expect(r.Header.Get(string(tele.CorrIDKeyVal))).To(Equal(corrID))
+		fmt.Fprintf(w, "Hello, %s", r.Proto)
+	}))
+	defer server.Close()
+
+	// Call the factory function and ensure it has both PerCallPolicies.
+	opts, err := ARMClientOptions("")
+	g.Expect(err).NotTo(HaveOccurred())
+	g.Expect(opts.PerCallPolicies).To(HaveLen(2))
+	g.Expect(opts.PerCallPolicies).To(ContainElement(BeAssignableToTypeOf(correlationIDPolicy{})))
```
mboersma, line 105: TIL you can put other matchers inside `ContainElement`. It's not as readable as most Ginkgo code, but it works as desired: does it contain an element of this type?

azure/defaults_test.go
```diff
@@ -24,11 +24,109 @@ import (
...
+		{
+			name:        "should return error if cloudName is unrecognized",
+			cloudName:   "AzureUnrecognizedCloud",
+			expectError: true,
+		},
+	}
+	for _, tc := range tests {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
```
nojnhuh, line 74: I'd vote to move the `NewWithT` here:
```suggestion
			t.Parallel()
			g := NewWithT(t)
```

CecileRobertMichon (commented):
/lgtm
/approve

PR #3843, opened 2023-08-15
Update virtualmachineimages service to SDK v2

### LINKED ISSUE #3473: Update VirtualMachineImages service to SDK v2

### LINKED ISSUE #3569: tenant and client ID are not used from AzureClusterIdentity object for virtualmachineimages client

willie-yao:
**What type of PR is this?**
/kind cleanup

Updates the virtualmachineimages service to use SDK v2.

Fixes #3569
Fixes #3473

- [ ] cherry-pick candidate

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Update virtualmachineimages service to SDK v2
```

### DIFF
--- a/azure/scope/clients.go
+++ b/azure/scope/clients.go
@@ -67,7 +67,7 @@ func (c *AzureClients) SubscriptionID() string {
 	return c.Values[auth.SubscriptionID]
 }
 
-// Token returns the Azure token credential of the cluster.
+// Token returns the Azure token credential of the cluster used for SDKv2 services.
 func (c *AzureClients) Token() azcore.TokenCredential {
 	return c.TokenCredential
 }
--- a/azure/scope/clients.go
+++ b/azure/scope/clients.go
@@ -24,6 +24,7 @@ import (
 	"os"
 	"strings"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/go-autorest/autorest"
 	azureautorest "github.com/Azure/go-autorest/autorest/azure"
 	"github.com/Azure/go-autorest/autorest/azure/auth"
@@ -35,6 +36,7 @@ type AzureClients struct {
 	auth.EnvironmentSettings
 
 	Authorizer                 autorest.Authorizer
+	TokenCredential            azcore.TokenCredential
 	ResourceManagerEndpoint    string
 	ResourceManagerVMDNSSuffix string
 }
@@ -63,6 +65,11 @@ func (c *AzureClients) ClientSecret() string {
 // either specified or from the environment.
 func (c *AzureClients) SubscriptionID() string {
 	return c.Values[auth.SubscriptionID]
+}
+
+// Token returns the Azure token credential of the cluster.
+func (c *AzureClients) Token() azcore.TokenCredential {
+	return c.TokenCredential
 }
 
 // HashKey returns a base64 url encoded sha256 hash for the Auth scope (Azure TenantID + CloudEnv + SubscriptionID +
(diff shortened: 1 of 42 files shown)


azure/services/virtualmachineimages/cache_test.go
```diff
@@ -34,22 +34,22 @@ func TestCacheGet(t *testing.T) {
...
-		have          compute.ListVirtualMachineImageResource
+		have          armcompute.VirtualMachineImagesClientListResponse
 		expectedError error
 	}{
 		"should find": {
 			location: "test", publisher: "foo", offer: "bar", sku: "baz",
-			have: compute.ListVirtualMachineImageResource{
-				Value: &[]compute.VirtualMachineImageResource{
+			have: armcompute.VirtualMachineImagesClientListResponse{
+				VirtualMachineImageResourceArray: []*armcompute.VirtualMachineImageResource{
```
Jont828, line 43: Just curious, is there a reason why this is a list of pointers instead of a normal list? I'm not sure how there would be a situation where you have nil list entries.

willie-yao: cc @mboersma Since he implemented this part. I'm not actually entirely sure

Jont828: Fair enough. I don't expect you to change anything especially if it's a type from the SDK, but I just wonder if we're overusing pointers sometimes.

mboersma: It's a type change they made basically *everywhere* in SDKv2: if a field used to be a pointer to a slice, now it's a slice of pointers. 🤷🏻

azure/scope/clients.go
```diff
@@ -133,7 +140,12 @@ func (c *AzureClients) setCredentialsWithProvider(ctx context.Context, subscript
 	}
 	c.Values[auth.ClientSecret] = strings.TrimSuffix(clientSecret, "\n")
 
-	c.Authorizer, err = credentialsProvider.GetAuthorizer(ctx, c.ResourceManagerEndpoint, c.Environment.ActiveDirectoryEndpoint, c.Environment.TokenAudience)
+	tokenCredential, err := credentialsProvider.GetTokenCredential(ctx, c.ResourceManagerEndpoint, c.Environment.ActiveDirectoryEndpoint, c.Environment.TokenAudience)
```
mboersma, line 143: Should we add unit tests to `clients_test.go` for the `HashKey()` and `Token()` funcs?

mboersma:
@willie-yao do you think it would be worth splitting this into two commits? One for the auth changes under `azure/scope`, another with everything else for the virtualmachineimages service. Just because it might make it easier to review now (and in the future, since several SDKv2 tasks might want to refer to this).

Just an idea, please ignore if it sounds like a hassle.

azure/scope/clients.go
willie-yao: I think both those functions would have trivial unit tests, and `HashKey()` was implemented in a previous PR. I do agree that this file needs unit tests and I think we can implement tests for the other functions in a separate issue/PR. wdyt?

willie-yao:
> do you think it would be worth splitting this into two commits?

@mboersma Done!!

mboersma (commented):
/lgtm
/assign @CecileRobertMichon

CecileRobertMichon (commented):
LGTM

Can we update code comments in the modified interfaces to account for the added `Token()` func? It would be also helpful to make the distinction that `Authorizer` is used by sdk v1 services while `Token` is used by sdk v2 services in code comments as the redundancy might be confusing to a future reader

CecileRobertMichon:
/lgtm
/approve

Great work @willie-yao, I like the simple and elegant solution you ended up with!

PR #3846, opened 2023-08-15
allow omitting clientSecret from identity in ASO secret controller

nojnhuh:
**What type of PR is this?**
/kind bug
/kind api-change

This change fixes an implicit assumption in the ASO Secret controller that an AzureClusterIdentity's `spec.clientSecret` is specified even though it's not necessary for types like WorkloadIdentity or UserAssignedMSI. Without this fix, the ASO controller does not create the Secret and logs the following error:

```
E0815 21:17:39.969029      42 controller.go:324]  "msg"="Reconciler error" "error"="failed to fetch AzureClusterIdentity secret: resource name may not be empty" "AzureCluster"={"name":"jon-aks","namespace":"default"} "controller"="ASOSecret" "controllerGroup"="infrastructure.cluster.x-k8s.io" "controllerKind"="AzureCluster" "name"="jon-aks" "namespace"="default" "reconcileID"="b9f13c36-003a-4fe2-8f02-aad486e0571b"
```

Fixes #

- [ ] cherry-pick candidate

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Release note**:

```release-note
The type of `AzureClusterIdentity.Spec.ClientSecret` has been changed from `corev1.SecretReference` to `*corev1.SecretReference`
```


controllers/asosecret_controller_test.go
```diff
@@ -115,6 +115,29 @@ func TestASOSecretReconcile(t *testing.T) {
 				}
 			}),
 		},
+		"should reconcile normally for AzureClusterIdentity without spec.clientSecret": {
```
nojnhuh, line 118: This test actually passes without the fix, presumably because the fake client doesn't return throw a `resource name may not be empty` error. It still seems like a worthwhile case to check though.

controllers/asosecret_controller.go
```diff
@@ -279,22 +279,25 @@ func (asos *ASOSecretReconciler) createSecretFromClusterIdentity(ctx context.Con
...
-		Namespace: identity.Spec.ClientSecret.Namespace,
-		Name:      identity.Spec.ClientSecret.Name,
-	}
-	identitySecret := &corev1.Secret{}
-	err := asos.Get(ctx, key, identitySecret)
-	if err != nil && !apierrors.IsNotFound(err) {
-		return nil, errors.Wrap(err, "failed to fetch AzureClusterIdentity secret")
-	}
+	if identity.Spec.ClientSecret != (corev1.SecretReference{}) {
+		key = types.NamespacedName{
```
nojnhuh, line 283: The only change here is wrapping this block in the `if`, which the "hide whitespace" checkbox should make clear.

CecileRobertMichon: Hmm I think `identity.Spec.ClientSecret` should actually be a pointer since it's marked as optional (`*corev1.SecretReference`) and then we'd be able to do `if identity.Spec.ClientSecret != nil`

CecileRobertMichon: For now a more readable check might be `if  identity.Spec.ClientSecret.Name != ""` but IMO we should just change it to be pointer (and make sure there are no other places in the code that assume it's set)

nojnhuh: Is the pointer change something we would backport? Or would it make sense to include as a separate commit in this PR?

CecileRobertMichon: Fair, if we don't want to backport the type change we can make it as a separate commit or even PR and only backport this one.

non-blocking, but I slightly prefer this for the cherry-pick, reads for easily as "not empty" to me, might be a personal preference though:

```suggestion
		if identity.Spec.ClientSecret.Name != "" {
```

controllers/asosecret_controller.go
```diff
@@ -279,22 +279,25 @@ func (asos *ASOSecretReconciler) createSecretFromClusterIdentity(ctx context.Con
...
-	key = types.NamespacedName{
-		Namespace: identity.Spec.ClientSecret.Namespace,
-		Name:      identity.Spec.ClientSecret.Name,
-	}
-	identitySecret := &corev1.Secret{}
-	err := asos.Get(ctx, key, identitySecret)
-	if err != nil && !apierrors.IsNotFound(err) {
-		return nil, errors.Wrap(err, "failed to fetch AzureClusterIdentity secret")
-	}
+	if identity.Spec.ClientSecret != (corev1.SecretReference{}) {
```
CecileRobertMichon, line 282: non-blocking, but I slightly prefer this for the cherry-pick, reads for easily as "not empty" to me, might be a personal preference though:

```suggestion
		if identity.Spec.ClientSecret.Name != "" {
```

nojnhuh: I'll do that.

controllers/asosecret_controller.go
nojnhuh: The fix part of this change probably doesn't make sense to cherry-pick since this entire controller only exists on the main branch. (added by #3698)

I'll make this PR 2 commits: one for the type change, then one for the fix.

CecileRobertMichon: So no need to cherry-pick any of it then?

nojnhuh: I think not.

CecileRobertMichon (commented):
/lgtm

mboersma (approved):
/lgtm
/approve

nojnhuh:
@CecileRobertMichon Is the pointer change here still valuable by itself? If so, I'm thinking of dropping the "fix" commit in this PR in favor of #3851. Otherwise I'll close this PR altogether.

azure/scope/identity.go
```diff
@@ -270,7 +271,7 @@ func createAzureIdentityWithBindings(ctx context.Context, azureIdentity *infrav1
 			Type:           azureIdentityType,
 			TenantID:       azureIdentity.Spec.TenantID,
 			ClientID:       azureIdentity.Spec.ClientID,
-			ClientPassword: azureIdentity.Spec.ClientSecret,
+			ClientPassword: ptr.Deref(azureIdentity.Spec.ClientSecret, corev1.SecretReference{}),
```
nawazkh, line 274: Just curious, sorry for the fly by review. 
Do you know if this could create a race scenario where we CAPZ controller is trying to set an empty `corev1.SecretReference{}` and the Azure resource in the cluster is accepting it as a "make it empty" and marks `ClientPassword: nil` ?

CecileRobertMichon:
@nojnhuh up to you but I'd probably lean toward closing this PR

azure/scope/identity.go
nojnhuh: This field doesn't map to an Azure API field on an Azure resource AFAIK. This change is purely to handle `azureIdentity.Spec.ClientSecret` being a pointer and the AzureIdentity field not. I don't think this involves any change in behavior.

nojnhuh:
> @nojnhuh up to you but I'd probably lean toward closing this PR

Easy enough!

/close

azure/scope/identity.go
nawazkh: Thanks!

PR #3847, opened 2023-08-15
Adding HTTP Proxy support for managed cluster

### LINKED ISSUE #3485: Support AKS cluster with HTTP Proxy mode
> ⚠️ Cluster API Azure maintainers can ask to turn an issue-proposal into a [CAEP](https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/proposals/YYYYMMDD-template.md) when necessary. This is to be expected for large changes that impact multiple components, breaking changes, or new large features.

***Goals***
1. Ability to create AKS clusters with http_proxy support https://learn.microsoft.com/en-us/azure/aks/http-proxy

***Non-Goals/Future Work***
1. Non-Goal 1
1. Non-Goal 2

**User Story**

As a [developer/user/operator] I would like to be able to create a Azure managed `Cluster` ( with `AzureManagedControlPlane`/`AzureManagedMachinePool` ) using the HTTP_PROXY mode.

**Detailed Description**

`AzureManagedControlPlane` can have a field `httpProxyConfig` akin to 
image
allowing us to create the cluster with proxy mode.

/kind proposal

mjnovice:
**What type of PR is this?**
/kind feature

Fixes #3485 

- [ ] cherry-pick candidate

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Adding HTTP proxy support for managed cluster
```


api/v1beta1/azuremanagedcontrolplane_types.go
```diff
@@ -186,6 +186,22 @@ type AzureManagedControlPlaneSpec struct {
...
 	// +optional
 	KubeletUserAssignedIdentity string `json:"kubeletUserAssignedIdentity,omitempty"`
+
+	// HTTPProxyConfig is the HTTP proxy configuration for the cluster.
+	// Immutable.
+	// +optional
+	HTTPProxyConfig *HTTPProxyConfig `json:"httpProxyConfig,omitempty"`
+}
+
+type HTTPProxyConfig struct {
```
CecileRobertMichon, line 196: should any/all of these be marked as optional?

mjnovice: Good point, however, since these are native types themselves (e.g. `*string`, `*[]string`) should we care to mark them explicitly optional/non-optional ?

Jont828: If it's non-optional, we don't need to do anything, but we do want to explicitly include the `// +optional` comment.

api/v1beta1/azuremanagedcontrolplane_types.go
```diff
@@ -186,6 +186,23 @@ type AzureManagedControlPlaneSpec struct {
...
+	// HTTPProxyConfig is the HTTP proxy configuration for the cluster.
+	// Immutable.
+	// +optional
+	HTTPProxyConfig *HTTPProxyConfig `json:"httpProxyConfig,omitempty"`
+}
+
+// HTTPProxyConfig is the HTTP proxy configuration for the cluster.
+type HTTPProxyConfig struct {
+	// HTTPProxy - The HTTP proxy server endpoint to use.
+	HTTPProxy *string `json:"httpProxy,omitempty"`
```
Jont828, line 199: Does this need to be a pointer, or would an empty string suffice to show that it's not set?

api/v1beta1/azuremanagedcontrolplane_types.go
```diff
@@ -186,6 +186,23 @@ type AzureManagedControlPlaneSpec struct {
...
+	// HTTPSProxy - The HTTPS proxy server endpoint to use.
+	HTTPSProxy *string `json:"httpsProxy,omitempty"`
+	// NoProxy - The endpoints that should not go through proxy.
+	NoProxy *[]string `json:"noProxy,omitempty"`
```
Jont828, line 203: I think we can change this to a `[]string` instead a slice pointer since an empty slice would indicate it's not set. That way, we could just do `len(NoProxy) > 0` instead of `NoProxy != nil && len(NoProxy) > 0'

api/v1beta1/azuremanagedcontrolplane_types.go
```diff
@@ -186,6 +186,23 @@ type AzureManagedControlPlaneSpec struct {
...
+	HTTPProxyConfig *HTTPProxyConfig `json:"httpProxyConfig,omitempty"`
+}
+
+// HTTPProxyConfig is the HTTP proxy configuration for the cluster.
+type HTTPProxyConfig struct {
+	// HTTPProxy - The HTTP proxy server endpoint to use.
+	HTTPProxy *string `json:"httpProxy,omitempty"`
+	// HTTPSProxy - The HTTPS proxy server endpoint to use.
+	HTTPSProxy *string `json:"httpsProxy,omitempty"`
+	// NoProxy - The endpoints that should not go through proxy.
```
Jont828, line 202:
```suggestion

	// NoProxy indicates that the endpoints that should not go through proxy.
```
Could you add a newline between all the fields similar to the other structs? And also reword slightly to avoid needing the `-`?

api/v1beta1/azuremanagedcontrolplane_types.go
```diff
@@ -186,6 +186,23 @@ type AzureManagedControlPlaneSpec struct {
...
+	NoProxy *[]string `json:"noProxy,omitempty"`
+	// TrustedCa - Alternative CA cert to use for connecting to proxy servers.
+	TrustedCa *string `json:"trustedCa,omitempty"`
```
Jont828, line 205:
```suggestion
	TrustedCA *string `json:"trustedCa,omitempty"`
```

azure/services/managedclusters/spec.go
```diff
@@ -113,6 +113,21 @@ type ManagedClusterSpec struct {
...
 	// KubeletUserAssignedIdentity is the user-assigned identity for kubelet to authenticate to ACR.
 	KubeletUserAssignedIdentity string
+
+	// HTTPProxyConfig is the HTTP proxy configuration for the cluster.
+	HTTPProxyConfig *HTTPProxyConfig
+}
+
+// HTTPProxyConfig is the HTTP proxy configuration for the cluster.
+type HTTPProxyConfig struct {
+	// HTTPProxy - The HTTP proxy server endpoint to use.
```
Jont828, line 123:
```suggestion

	// HTTPProxy is the HTTP proxy server endpoint to use.
```

Could we add newlines between the fields as well as reword it from `SomeField - Something` to `SomeField is something`?

Jont828:
Looks like you had a failure with the VMSS e2e test, rerunning in case it's a flake.

api/v1beta1/azuremanagedcontrolplane_types.go
mjnovice: I think it would make sense to have this as a pointer, as it will be in parallel to what the `containerservice` module expects ?

api/v1beta1/azuremanagedcontrolplane_types.go
mjnovice: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3847/files#r1296341246

api/v1beta1/azuremanagedcontrolplane_types.go
mjnovice: Ah I see. Makes sense.

api/v1beta1/azuremanagedcontrolplane_types.go
Jont828: That makes sense, I just wonder if we're using too many pointers tbh.

azure/services/managedclusters/spec.go
```diff
@@ -414,6 +432,15 @@ func (s *ManagedClusterSpec) Parameters(ctx context.Context, existing interface{
 		}
 	}
 
+	if s.HTTPProxyConfig != nil {
+		managedCluster.HTTPProxyConfig = &containerservice.ManagedClusterHTTPProxyConfig{
```
Jont828, line 436: Now that you mention it, I see how it would be awkward to use an empty string as the uninitialized value, since we'd have `HTTPProxy = ptr.To("")` instead of `nil.`

mjnovice: yep

api/v1beta1/azuremanagedcontrolplane_types.go
Jont828: Is there anyway we can change this type though? I searched through the CAPZ repo and we don't really have any mentions of a pointer to a slice in any of the types in CAPZ.

mjnovice: Yes, makes sense. This is one such instance where we make the conversion
```
	if s.APIServerAccessProfile != nil {
		managedCluster.APIServerAccessProfile = &containerservice.ManagedClusterAPIServerAccessProfile{
			EnablePrivateCluster:           s.APIServerAccessProfile.EnablePrivateCluster,
			PrivateDNSZone:                 s.APIServerAccessProfile.PrivateDNSZone,
			EnablePrivateClusterPublicFQDN: s.APIServerAccessProfile.EnablePrivateClusterPublicFQDN,
		}

		if s.APIServerAccessProfile.AuthorizedIPRanges != nil {
			managedCluster.APIServerAccessProfile.AuthorizedIPRanges = &s.APIServerAccessProfile.AuthorizedIPRanges
		}
	}
```
Let me make that change.

api/v1beta1/azuremanagedcontrolplane_types.go
nawazkh: ^ Jont828 asked a good question.

Questions for @mjnovice, do the empty fields, i.e. `""` mean anything significant for the fields of `HTTPProxyConfig` ? 
It makes sense to me that `HTTPProxyConfig *HTTPProxyConfig json:"httpProxyConfig,omitempty"` is a pointer. `httpProxyConfig` does not have a in-built `nil` type and we want to distinguish it between having it assigned vs leaving it empty. 

However, I believe the fields of `HTTPProxyConfig` can be defined as non-pointers unless we want to explicitly differentiate between `""` and it not being set.

This article was quite handy for me to understand if the field should be a pointer type vs a non-pointer type. https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md

nawazkh: I also ran a simple test to check if it could cause any problems. 
https://goplay.tools/snippet/IJM-wX1BDdR

Jont828:
@mjnovice The verify error seems to be a linting issue with the shellcheck script.

nawazkh:
[My comment](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3847#discussion_r1296455315) might have gotten buried into resolved comments, but I would love to hear your thoughts @mjnovice

api/v1beta1/azuremanagedcontrolplane_types.go
mjnovice: @nawazkh you both raise valid points, however I was just following a pattern which we already have in the file. For e.g https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/services/managedclusters/spec.go#L167 , didn't want to create an anti-pattern. 

I we want to move away from pointers, that could be a refactor PR ?

mjnovice:
> @mjnovice The verify error seems to be a linting issue with the shellcheck script.

@Jont828 seems to be a ratelimit issue ? https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/3847/pull-cluster-api-provider-azure-verify/1691920826074402816#1:build-log.txt%3A510

mjnovice:
> [My comment](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3847#discussion_r1296455315) might have gotten buried into resolved comments, but I would love to hear your thoughts @mjnovice

@nawazkh have replied to your comment. 🙂 let me know your thoughts.

api/v1beta1/azuremanagedcontrolplane_types.go
CecileRobertMichon: Here is the k8s API convention docs on optional vs required fields: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required

specifically:

```
Optional fields have the following properties:

- They have the +optional comment tag in Go.
- They are a pointer type in the Go definition (e.g. AwesomeFlag *SomeFlag) or have a built-in nil value (e.g. maps and slices).
- The API server should allow POSTing and PUTing a resource with this field unset.
```

so all the optional fields here should be pointer types and have the `+optional` comment

nawazkh:
Thanks for the explanation @mjnovice 
Looks good to me!
/lgtm

Jont828:
I've seen a Docker rate limiting issue when testing locally, but I haven't seen that in the CI jobs. If it comes up again we could open an issue.

api/v1beta1/azuremanagedcontrolplane_types.go
Jont828: @CecileRobertMichon except for the slice I believe

azure/services/managedclusters/spec.go
```diff
@@ -414,6 +432,18 @@ func (s *ManagedClusterSpec) Parameters(ctx context.Context, existing interface{
...
+			HTTPProxy:  s.HTTPProxyConfig.HTTPProxy,
+			HTTPSProxy: s.HTTPProxyConfig.HTTPSProxy,
+			TrustedCa:  s.HTTPProxyConfig.TrustedCA,
+		}
+
+		if s.HTTPProxyConfig.NoProxy != nil {
+			managedCluster.HTTPProxyConfig.NoProxy = &s.HTTPProxyConfig.NoProxy
+		}
```
Jont828, line 444: Could you add a couple unit tests to `spec_test.go` to cover these cases?

Jont828 (commented):
Just a quick thing, otherwise I think we're good to go!

api/v1beta1/azuremanagedcontrolplane_types.go
CecileRobertMichon: correct

azure/services/managedclusters/spec.go
mjnovice: sure.

Jont828 (approved):
/lgtm

Nice work!

Issue #3732, opened 2023-07-17
Allow authentication with Azure Workload Identity for ASO


nojnhuh:
Follow-up from #3526

AzureClusterIdentity resources specifying `type: WorkloadIdentity` should configure ASO to use the same Workload Identity parameters for authentication when interacting with resources for clusters associated with that AzureClusterIdentity.

nojnhuh:
Testing this out, this should actually already work without any "real" changes to CAPZ, or at least to the Secret that gets generated for ASO. The Workload Identity test on #3723 is only failing because ASO's ServiceAccount isn't authorized:
```
"AADSTS70021: No matching federated identity record found for presented assertion. Assertion Issuer: 'https://oidcissuercapzci.blob.core.windows.net/oidc-capzci/'. Assertion Subject: 'system:serviceaccount:capz-system:azureserviceoperator-default'. Assertion Audience: 'api://AzureADTokenExchange'. https://docs.microsoft.com/en-us/azure/active-directory/develop/workload-identity-federation\r\nTrace ID: 6003acdd-1a4f-4e22-8d81-fb5b46c04600\r\nCorrelation ID: 29d0cca0-c49b-43e9-b0b9-957d766ca2cb\r\nTimestamp: 2023-08-14 05:06:56Z"
```

As part of this issue, we should still document that extra step in the doc added by #3770 and update our own CI config.

adriananeci:
> Testing this out, this should actually already work without any "real" changes to CAPZ, or at least to the Secret that gets generated for ASO.

Yup, that's right because Workload Identity doesn't need any extra secret values compared with what we are currently adding into the ASO secret, see https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/controllers/asosecret_controller.go#L262-L283 for more context.

For readability, I'm looking to extend the cases statement while generating the ASO secret so that workload identity is explicitly treated in ASO secret generation logic.

> As part of this issue, we should still document that extra step in the doc added by https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3770 and update our own CI config.

I'll se how those settings can be added in CI config once the Workload Identity PR is merged.

PR #3770, opened 2023-07-27
add doc for workload identity

sonasingh46:
**What type of PR is this?**
This PR documents workload identity . 

**What this PR does / why we need it**:
This is a workload identity documentation PR.

- [ ] cherry-pick candidate

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
None
```

Issue #3743, opened 2023-07-19
Implement UserAssignedMSI auth for ASO


nojnhuh:
Follow-up from #3526

CAPZ currently allows users to create an AzureClusterIdentity with `spec.type=UserAssignedMSI` driven by AAD Pod Identity and does not require the manual setup steps of Workload Identity. The same credentials in this case should be propagated to CAPZ's ASO.

If AAD Pod Identity support is removed from CAPZ before #3527 closes, then this issue would no longer need to be implemented.

see also:
https://azure.github.io/azure-service-operator/guide/authentication/credential-format/#deprecated-managed-identity-aad-pod-identity

```[tasklist]
### Dependencies
- [ ] https://github.com/Azure/azure-service-operator/issues/3167
```

nojnhuh:
To recap yesterday's discussion with @CecileRobertMichon to help me understand this:
- CAPZ uses AAD Pod ID currently to enable the `ServicePrincipal` and `ServicePrincipalCertificate` AzureClusterIdentity types.
- ASO's [documented instructions](https://azure.github.io/azure-service-operator/guide/authentication/credential-format/#service-principal-using-a-client-secret) for Service Principal do not involve AAD Pod ID, though configuring a Service Principal with AAD Pod ID does work.
- If we have the option to either configure or not configure ASO to use AAD Pod ID, we'll opt not to use AAD Pod ID, though if some users report issues because they rely on auth going through AAD Pod Identity, we can add that capability as needed.

To follow up with some of the open questions:

> Does the NMI pod intercept IMDS requests from ASO when AAD Pod ID is not configured explicitly?

No AFAICT. I don't see any new logs in the NMI pod when creating an ASO resource with a Service Principal like this: https://azure.github.io/azure-service-operator/guide/authentication/credential-format/#service-principal-using-a-client-secret

> Does ASO support managed identities either with or without AAD Pod Identity (excluding Workload ID)?

Yes with AAD Pod ID, not without. I can get managed identity working with these instructions: https://azure.github.io/azure-service-operator/guide/authentication/credential-format/#deprecated-managed-identity-aad-pod-identity. Those steps modify the global `aso-controller-settings` secret though, and if I try to use the managed identity client ID in a secret referenced by a resource's `credential-from` annotation ASO hits an error that indicates it's trying to use Workload ID, so I'm not sure ASO supports multi-tenant managed identity setups with AAD Pod ID. Still looking into that.

PR #3851, opened 2023-08-16
Allow authentication with Azure Workload Identity for ASO

adriananeci:
**What type of PR is this?**
/kind feature

/kind feature

AzureClusterIdentity resources specifying `type: WorkloadIdentity` should configure ASO to use the same Workload Identity parameters for authentication when interacting with resources for clusters associated with that AzureClusterIdentity.

This PR explicitly handles the  AzureClusterIdentity of type `WorkloadIdentity` while generating the ASO secrets.

Fixes https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/3732

Workload ID setup might be needed for the CI setup

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/controllers/asosecret_controller.go
+++ b/controllers/asosecret_controller.go
@@ -278,6 +278,11 @@ func (asos *ASOSecretReconciler) createSecretFromClusterIdentity(ctx context.Con
 	newASOSecret.Data["AZURE_TENANT_ID"] = []byte(identity.Spec.TenantID)
 	newASOSecret.Data["AZURE_CLIENT_ID"] = []byte(identity.Spec.ClientID)
 
+	// If the identity type is WorkloadIdentity or UserAssignedMSI, then we don't need to fetch the secret so return early
+	if identity.Spec.Type == infrav1.WorkloadIdentity || identity.Spec.Type == infrav1.UserAssignedMSI {
+		return newASOSecret, nil
+	}
+
 	// Fetch identity secret, if it exists
 	key = types.NamespacedName{
 		Namespace: identity.Spec.ClientSecret.Namespace,
@@ -285,7 +290,7 @@ func (asos *ASOSecretReconciler) createSecretFromClusterIdentity(ctx context.Con
 	}
 	identitySecret := &corev1.Secret{}
 	err := asos.Get(ctx, key, identitySecret)
-	if err != nil && !apierrors.IsNotFound(err) {
+	if err != nil {
 		return nil, errors.Wrap(err, "failed to fetch AzureClusterIdentity secret")
 	}
 
--- a/controllers/asosecret_controller_test.go
+++ b/controllers/asosecret_controller_test.go
@@ -54,6 +54,7 @@ func TestASOSecretReconcile(t *testing.T) {
 	defaultAzureCluster := getASOAzureCluster()
 	defaultAzureManagedControlPlane := getASOAzureManagedControlPlane()
 	defaultASOSecret := getASOSecret(defaultAzureCluster)
+	defaultClusterIdentityType := infrav1.ServicePrincipal
 
 	cases := map[string]struct {
 		clusterName string
@@ -80,7 +81,13 @@ func TestASOSecretReconcile(t *testing.T) {
 						Namespace: "default",
 					}
 				}),
-				getASOAzureClusterIdentity(),
+				getASOAzureClusterIdentity(func(identity *infrav1.AzureClusterIdentity) {
+					identity.Spec.Type = defaultClusterIdentityType
+					identity.Spec.ClientSecret = corev1.SecretReference{
+						Name:      "fooSecret",
+						Namespace: "default",
+					}
+				}),
 				getASOAzureClusterIdentitySecret(),
 				defaultCluster,
 			},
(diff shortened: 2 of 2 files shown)


nojnhuh:
Sorry I may have stepped on your toes a bit with #3846. I put a hold on that PR so we can decide which route to take, but I like this approach a bit better since the WorkloadIdentity type is mentioned and doesn't do the extra API call even if `spec.clientSecret` is specified anyway (like it is currently in CI).

Let's also be sure to keep tabs on #3770 and add docs about authorizing the azureserviceoperator-default ServiceAccount after that PR lands, either in this PR or as a follow-up.

^ For the WorkloadIdentity test. Everything should already be set up in CI.

controllers/asosecret_controller.go
```diff
@@ -278,14 +278,19 @@ func (asos *ASOSecretReconciler) createSecretFromClusterIdentity(ctx context.Con
 	newASOSecret.Data["AZURE_TENANT_ID"] = []byte(identity.Spec.TenantID)
 	newASOSecret.Data["AZURE_CLIENT_ID"] = []byte(identity.Spec.ClientID)
 
+	// If the identity type is WorkloadIdentity, then we don't need to fetch the secret so return early
+	if identity.Spec.Type == infrav1.WorkloadIdentity {
```
CecileRobertMichon, line 282: what about `UserAssignedMSI` identity type? wouldn't that also not have a secret?

nojnhuh: I was wondering if it might be worth making that change alongside the other necessary changes for UserAssignedMSI like creating the AzureIdentity resources as part of #3743.

CecileRobertMichon (commented):
/lgtm

controllers/asosecret_controller_test.go
```diff
@@ -175,7 +252,7 @@ func TestASOSecretReconcile(t *testing.T) {
 				g.Expect(reconciler.Recorder.(*record.FakeRecorder).Events).To(Receive(ContainSubstring(tc.event)))
 			}
 			if tc.err != "" {
-				g.Expect(err).To(MatchError(tc.err))
+				g.Expect(err).To(MatchError(ContainSubstring(tc.err)))
```
Jont828, line 255: Is there a reason we're no longer trying to match the error message in this test?

controllers/asosecret_controller_test.go
```diff
@@ -80,7 +81,13 @@ func TestASOSecretReconcile(t *testing.T) {
 						Namespace: "default",
 					}
 				}),
-				getASOAzureClusterIdentity(),
+				getASOAzureClusterIdentity(func(identity *infrav1.AzureClusterIdentity) {
+					identity.Spec.Type = infrav1.IdentityType(defaultClusterClusterIdentityType)
```
nojnhuh, line 85: I wonder if it might be slightly more clear/readable to use the `infrav1.(WorkloadIdentity|ServicePrincipal|...)` constants directly in these tests.

Jont828: The benefit of introducing the variable is that it's clear that `ServicePrincipal` is the default type. I think it'd be fine to change it but maybe you add a comment or something.

controllers/asosecret_controller.go
```diff
@@ -278,14 +278,19 @@ func (asos *ASOSecretReconciler) createSecretFromClusterIdentity(ctx context.Con
...
+	}
+
 	// Fetch identity secret, if it exists
 	key = types.NamespacedName{
 		Namespace: identity.Spec.ClientSecret.Namespace,
 		Name:      identity.Spec.ClientSecret.Name,
 	}
 	identitySecret := &corev1.Secret{}
 	err := asos.Get(ctx, key, identitySecret)
-	if err != nil && !apierrors.IsNotFound(err) {
```
willie-yao, line 288: Are not found errors handled differently now that the secret is being returned early, or is this solving a different problem?

controllers/asosecret_controller_test.go
```diff
@@ -54,6 +54,7 @@ func TestASOSecretReconcile(t *testing.T) {
 	defaultAzureCluster := getASOAzureCluster()
 	defaultAzureManagedControlPlane := getASOAzureManagedControlPlane()
 	defaultASOSecret := getASOSecret(defaultAzureCluster)
+	defaultClusterClusterIdentityType := "ServicePrincipal"
```
willie-yao, line 57: nit
```suggestion
	defaultClusterClusterIdentityType := infrav1.ServicePrincipal
```

controllers/asosecret_controller_test.go
adriananeci: You're right. This was just a quick fix in order to make lint happy. Adjusted it based on your suggestion.

controllers/asosecret_controller_test.go
adriananeci: You're right. This was just a quick fix in order to make lint happy. Adjusted it based on your suggestion.

controllers/asosecret_controller.go
adriananeci: This is solving a different problem. If secret is not found, we need to return here too in order to avoid any issues few lines above were we are referencing identity secret data that won't be available if the secret doesn't exist, e.g. `identitySecret.Data`

controllers/asosecret_controller_test.go
adriananeci: Yes, some error messages are composed using different layers that are not necessarily ASO related and might be changed independent of the ASO controller, thus affecting the ASO secret tests. By introducing `ContainSubstring` we are checking only for the string we know will be thrown by the ASO controller.

CecileRobertMichon (commented):
/approve

PR #3887, opened 2023-08-24
Fix a natgateways return value

mboersma:
**What type of PR is this?**

/kind cleanup

Fixes the return value of `CreateOrUpdateAsync` and cleans up `newClient` to match other SDK v2 implementations.

This seems like a major bug, but in effect was just a deoptimization.

N/A

- [ ] cherry-pick candidate

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/azure/services/natgateways/client.go
+++ b/azure/services/natgateways/client.go
@@ -33,26 +33,17 @@ type azureClient struct {
 	natgateways *armnetwork.NatGatewaysClient
 }
 
-// newClient creates a new VM client from subscription ID.
+// newClient creates a new nat gateways client from an authorizer.
 func newClient(auth azure.Authorizer) (*azureClient, error) {
-	c, err := netNatGatewaysClient(auth)
-	if err != nil {
-		return nil, errors.Wrap(err, "failed to create natgateways client")
-	}
-	return &azureClient{c}, nil
-}
-
-// netNatGatewaysClient creates a new nat gateways client from subscription ID.
-func netNatGatewaysClient(auth azure.Authorizer) (*armnetwork.NatGatewaysClient, error) {
 	opts, err := azure.ARMClientOptions(auth.CloudEnvironment())
 	if err != nil {
-		return &armnetwork.NatGatewaysClient{}, errors.Wrap(err, "failed to create natgateways client options")
+		return nil, errors.Wrap(err, "failed to create natgateways client options")
 	}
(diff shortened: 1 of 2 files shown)


azure/services/natgateways/client.go
```diff
@@ -90,15 +81,15 @@ func (ac *azureClient) CreateOrUpdateAsync(ctx context.Context, spec azure.Resou
...
+	resp, err := poller.PollUntilDone(ctx, pollOpts)
 	if err != nil {
 		// if an error occurs, return the poller.
 		// this means the long-running operation didn't finish in the specified timeout.
 		return nil, poller, err
 	}
 
 	// if the operation completed, return a nil poller
-	return result, nil, err
+	return resp.NatGateway, nil, err
```
CecileRobertMichon, line 92: is there anything we can do to change the return type so it's not an `interface{}` to protect us from similar errors in the future?

mboersma: Each return value (both the `resp` and its `NatGateway` field) are individual structs without a common interface. https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork/v4@v4.0.0#NatGatewaysClientCreateOrUpdateResponse

Since this func (`CreateOrUpdateAsync`) is part of the `asyncpoller` interfaces, I'm not sure how we can avoid `interface{}` but I'm open to ideas.

mboersma: If all the returned resources (NatGateway, AgentPool, InboundNatRule, etc.) have some common field or fields, we can define an interface and return that. `ID`, `Name`, and `Type` are on the resources I've looked at.

mboersma: We could make an interface with the only two funcs that seem to be shared:

```go
// AzureResource is a basic interface for Azure resources.
type AzureResource interface {
	MarshalJSON() ([]byte, error)
	UnmarshalJSON(data []byte) error
}
```
Then we could return `AzureResource` instead of `interface{}`. It's more descriptive, and it would have prevented this particular bug.

But the `resp` part of `resp.NatGateway` also implements these funcs, so it wouldn't prevent accidentally returning `resp`. And we don't use the JSON funcs, so it feels a bit odd to use it just as a "marker" interface. WDYT?

CecileRobertMichon: I don't have any better suggestions for now so let's keep it as-is but I'll try to keep thinking about this to see if we can improve it

PR #3932, opened 2023-08-31
Implement UserAssignedMSI auth for ASO

### LINKED ISSUE #3743: Implement UserAssignedMSI auth for ASO

nojnhuh:
**What type of PR is this?**
/kind feature

This PR enables UserAssignedMSI auth with ASO by creating an AzureIdentityBinding bound to the existing AzureIdentity and referencing the same `AZURE_CLIENT_ID`, etc. variables in a generated Secret used by ASO.

Since this was the last reason to keep the legacy groups service implementation, that is also removed here.

Fixes #3743

This is ready for review but currently blocked on a new release of ASO which includes https://github.com/Azure/azure-service-operator/commit/9061839958230853ad6b7d9def8d35b4664c0766 and subsequent referencing of that release in CAPZ.
/hold

These changes are probably easiest to review one commit at a time. Otherwise the whole diff shows lots of changes that are really just renames.

- [ ] cherry-pick candidate

**TODOs**:

- [ ] squashed commits <-- These commits may be best to leave as-is in case we need them later. Squashing might muddy the history and make cherry-picking/reverting much more error-prone.
- [ ] includes documentation
- [X] adds unit tests

**Release note**:

```release-note
NONE
```


azure/scope/identity.go
```diff
@@ -299,28 +299,51 @@ func createAzureIdentityWithBindings(ctx context.Context, azureIdentity *infrav1
...
+				AzureIdentity: copiedIdentity.Name,
+				Selector:      infrav1.AzureIdentityBindingSelector, // should be same as selector added on controller
 			},
 		},
-		Spec: aadpodv1.AzureIdentityBindingSpec{
-			AzureIdentity: copiedIdentity.Name,
-			Selector:      infrav1.AzureIdentityBindingSelector, // should be same as selector added on controller
+		{
+			TypeMeta: metav1.TypeMeta{
+				Kind:       "AzureIdentityBinding",
```
nojnhuh, line 324: I tested again whether the NMI pod intercepts any requests that it doesn't need to and I only saw it intercept requests from ASO when UserAssignedMSI auth is in use, which is what I hoped to see.

controllers/asosecret_controller.go
```diff
@@ -287,7 +287,12 @@ func (asos *ASOSecretReconciler) createSecretFromClusterIdentity(ctx context.Con
 	newASOSecret.Data["AZURE_CLIENT_ID"] = []byte(identity.Spec.ClientID)
 
 	// If the identity type is WorkloadIdentity or UserAssignedMSI, then we don't need to fetch the secret so return early
-	if identity.Spec.Type == infrav1.WorkloadIdentity || identity.Spec.Type == infrav1.UserAssignedMSI {
+	if identity.Spec.Type == infrav1.WorkloadIdentity {
+		newASOSecret.Data["AUTH_MODE"] = []byte("workloadidentity")
```
nojnhuh, line 291: I didn't add new constants here for these since ASO will make those available to us in its new release as well.

nojnhuh:
/hold
to squash out the merge commit and any follow-ups.

nojnhuh:
For the private cluster test which uses UserAssignedMSI auth:

nojnhuh:
> For the private cluster test which uses UserAssignedMSI auth:

Actually that may not test this at all if the resource group already exists, but I ran a quick manual test and the resource group gets created successfully by ASO with UserAssignedIdentity auth.

controllers/asosecret_controller.go
nojnhuh: ^ I'm planning a follow-up to this that utilizes the constants.

nojnhuh:
AKS test failure is related to #3970

azure/scope/identity.go
```diff
@@ -299,28 +299,51 @@ func createAzureIdentityWithBindings(ctx context.Context, azureIdentity *infrav1
...
+				Namespace: copiedIdentity.Namespace,
+				Labels: map[string]string{
+					clusterv1.ClusterNameLabel:              clusterMeta.Name,
+					infrav1.ClusterLabelNamespace:           clusterMeta.Namespace,
+					clusterctl.ClusterctlMoveHierarchyLabel: "true",
+				},
+			},
+			Spec: aadpodv1.AzureIdentityBindingSpec{
+				AzureIdentity: copiedIdentity.Name,
+				Selector:      "aso-manager-binding", // should be same as selector added on controller
```
CecileRobertMichon, line 338: is there a constant available for this?

CecileRobertMichon (commented):
lgtm, just one tiny question

azure/scope/identity.go
nojnhuh: No, this is only specified in the ASO YAML manifests here: https://github.com/nojnhuh/azure-service-operator/blob/47739980423ffdd1fb281b6d2e0754ed8cd081c5/v2/config/manager/manager.yaml#L31

nojnhuh:
#3970 

/retest

CecileRobertMichon (commented):
/assign @willie-yao @mboersma

azure/services/groups/spec.go
```diff
@@ -19,52 +19,93 @@ package groups
...
-	}, nil
+	return resource.(*asoresourcesv1.ResourceGroup).Spec.Tags
+}
+
+// GetActualTags implements aso.TagsGetterSetter.
+func (s *GroupSpec) GetActualTags(resource genruntime.MetaObject) infrav1.Tags {
+	if resource == nil {
+		return nil
+	}
+	return resource.(*asoresourcesv1.ResourceGroup).Status.Tags
```
mboersma, line 102: Are we ok with this type assertion possibly `panic`king? Seems unlikely given the code that calls this, but I always ask when I see an unchecked type assertion.

azure/services/groups/spec.go
```diff
@@ -19,52 +19,93 @@ package groups
...
+}
+
+// SetTags implements aso.TagsGetterSetter.
+func (s *GroupSpec) SetTags(resource genruntime.MetaObject, tags infrav1.Tags) {
+	if resource == nil {
+		return
+	}
+	resource.(*asoresourcesv1.ResourceGroup).Spec.Tags = tags
```
mboersma, line 110: Same comment here WRT the type assertion.

mboersma (commented):
/lgtm

Just had a question about whether to check two type assertions. Nice work! Squash?

mboersma (commented):
Nice, lgtm pending squashing the commits.

(Wow, sorry--didn't realize that handling the type assertion error would make such a big splash.)

nojnhuh:
I squashed the merge commits out, but I think keeping the others separate will make things easier to audit later if need be.

/hold cancel

PR #3934, opened 2023-08-31
fix asyncpoller long running op state cleanup

nojnhuh:
**What type of PR is this?**
/kind bug

This PR fixes a bug where the asyncpoller framework would not properly clean up the `status.longRunningOperationStates` when the operations completed. The main change is to skip calling `Parameters()` when a resource already has a long-running operation in progress. This allows reconciliation of resources with in-progress operations to reach the `CreateOrUpdateAsync()` call instead of erroneously short-circuiting in the case where the result of `Parameters()` is nil and therefore never reaching `DeleteLongRunningOperationState()`. 

Fixes #

- [ ] cherry-pick candidate

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/azure/services/asyncpoller/asyncpoller.go
+++ b/azure/services/asyncpoller/asyncpoller.go
@@ -18,7 +18,6 @@ package asyncpoller
 
 import (
 	"context"
-	"fmt"
 	"net/http"
 	"strconv"
 	"time"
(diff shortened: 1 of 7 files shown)


azure/services/asyncpoller/asyncpoller.go
```diff
@@ -73,34 +73,40 @@ func (s *Service[C, D]) CreateOrUpdateResource(ctx context.Context, spec azure.R
...
-		return nil, azure.WithTransientError(errWrapped, getRetryAfterFromError(err))
-	} else if err == nil {
-		existingResource = existing
-		log.V(2).Info("successfully got existing resource", "service", serviceName, "resource", resourceName, "resourceGroup", rgName)
-	}
+	// Only when no long running operation is currently in progress do we need to get the parameters.
+	// The polling implemented by the SDK does not use parameters when a resume token exists.
+	var parameters interface{}
+	if resumeToken == "" {
+		// Get the resource if it already exists, and use it to construct the desired resource parameters.
```
nojnhuh, line 80: Most of these changes are purely indentation.

azure/services/asyncpoller/asyncpoller.go
```diff
@@ -73,34 +73,40 @@ func (s *Service[C, D]) CreateOrUpdateResource(ctx context.Context, spec azure.R
...
+		if existingResource != nil {
+			logMessageVerbPrefix = "updat"
+		}
+		log.V(2).Info(fmt.Sprintf("%sing resource", logMessageVerbPrefix), "service", serviceName, "resource", resourceName, "resourceGroup", rgName)
 	}
-	log.V(2).Info(fmt.Sprintf("%sing resource", logMessageVerbPrefix), "service", serviceName, "resource", resourceName, "resourceGroup", rgName)
+
 	result, poller, err := s.Creator.CreateOrUpdateAsync(ctx, spec, resumeToken, parameters)
-	errWrapped := errors.Wrapf(err, fmt.Sprintf("failed to %se resource %s/%s (service: %s)", logMessageVerbPrefix, rgName, resourceName, serviceName))
+	errWrapped := errors.Wrapf(err, "failed to create or update resource %s/%s (service: %s)", rgName, resourceName, serviceName)
```
nojnhuh, line 109: This wording change is not just me being lazy, but I don't think we can reliably know which of "create" or "update" is really happening here since we no longer always do a GET.

azure/services/availabilitysets/client.go
```diff
@@ -64,7 +64,7 @@ func (ac *AzureClient) CreateOrUpdateAsync(ctx context.Context, spec azure.Resou
 	defer done()
 
 	availabilitySet, ok := parameters.(armcompute.AvailabilitySet)
-	if !ok {
+	if !ok && parameters != nil {
```
nojnhuh, line 67: This change is necessary to each service since we first instantiate a `parameters` set to nil in the asyncpoller, only sometimes set it to the result of `Parameters` which has a different concrete type, then pass that value on to this method.

Reflection may be able to work around having to make these changes to the services, but I have a feeling that would end up pretty messy and brittle.

mboersma (commented):
/lgtm

azure/services/asyncpoller/asyncpoller.go
```diff
@@ -73,34 +73,40 @@ func (s *Service[C, D]) CreateOrUpdateResource(ctx context.Context, spec azure.R
...
+			log.V(2).Info("resource up to date", "service", serviceName, "resource", resourceName, "resourceGroup", rgName)
+			return existingResource, nil
+		}
 
-	// Create or update the resource with the desired parameters.
-	logMessageVerbPrefix := "creat"
-	if existingResource != nil {
-		logMessageVerbPrefix = "updat"
+		// Create or update the resource with the desired parameters.
+		logMessageVerbPrefix := "creat"
```
CecileRobertMichon, line 101: since we're no longer reusing this, can we simplify the logic and simply print two different log statements if existing resource is nil vs not nil?

CecileRobertMichon: this would allow us to remove https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/.codespellignore#L5

nojnhuh: Yep, makes sense.

CecileRobertMichon:
/lgtm
/approve
/hold

for squash

nojnhuh:
squashed!
/hold cancel

azure/services/asyncpoller/asyncpoller.go
nojnhuh: Or maybe we can't quite update codespell yet :/
```
/home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/azure/services/aso/aso.go:217: updat ==> update
/home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure/azure/services/async/async.go:138: updat ==> update
```

mboersma (approved):
/lgtm

PR #3938, opened 2023-09-01
convert securitygroups service to SDK v2

### LINKED ISSUE #3923: Update securitygroups service to SDK v2

nojnhuh:
**What type of PR is this?**
/kind cleanup

This PR updates the securitygroups service to use SDK v2 and the new asyncpoller framework.

Fixes #3923

This PR is split into two commits that I intend to squash before merging. The first commit is the "standard" conversion to SDK v2. The second commit adds back some special logic that existed to propagate the Etag from an existing resource to a subsequent PUT by creating a new SDK client that will add a HTTP header with the Etag. This isn't quite what the managedclusters or agentpools services does for custom headers because the Etag isn't known right at the beginning of reconciliation, only after we GET the existing resource.

- [ ] cherry-pick candidate

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/azure/defaults.go
+++ b/azure/defaults.go
@@ -388,17 +388,14 @@ func (p userAgentPolicy) Do(req *policy.Request) (*http.Response, error) {
 // CustomPutPatchHeaderPolicy adds custom headers to a PUT or PATCH request.
 // It implements the policy.Policy interface.
 type CustomPutPatchHeaderPolicy struct {
-	Getter ResourceSpecGetter
+	Headers map[string]string
 }
 
 // Do adds any custom headers to a PUT or PATCH request.
 func (p CustomPutPatchHeaderPolicy) Do(req *policy.Request) (*http.Response, error) {
 	if req.Raw().Method == http.MethodPut || req.Raw().Method == http.MethodPatch {
-		headerSpec, ok := p.Getter.(ResourceSpecGetterWithHeaders)
-		if ok {
-			for key, element := range headerSpec.CustomHeaders() {
-				req.Raw().Header.Set(key, element)
-			}
+		for key, element := range p.Headers {
+			req.Raw().Header.Set(key, element)
 		}
 	}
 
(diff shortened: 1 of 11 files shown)


willie-yao (approved):
/lgtm

CecileRobertMichon:
/hold 

for squash

azure/defaults.go
```diff
@@ -405,6 +405,21 @@ func (p CustomPutPatchHeaderPolicy) Do(req *policy.Request) (*http.Response, err
 	return req.Next()
 }
 
+// AddHeadersPolicy adds headers to a request.
+type AddHeadersPolicy struct {
+	KeyValues map[string][]string
+}
+
+// Do adds headers to the request.
+func (p AddHeadersPolicy) Do(req *policy.Request) (*http.Response, error) {
```
CecileRobertMichon, line 414: how is this different from the other `Do` func above to add custom headers? is there no reuse possible?

nojnhuh: The main difference is that this uses `Header.Add()` instead of `Header.Set()` to keep this working the same way it was before. I wasn't ambitious enough to test whether using `Set` here would be acceptable.

CecileRobertMichon: Ideally we'd use the same func for both. If we really can't, a comment to clarify the difference between the two would help

nojnhuh: Refactored `CustomPutPatchHeaderPolicy` to make it usable for this as well.

CecileRobertMichon (commented):
/lgtm

mboersma (commented):
/lgtm

Pending squash.

nojnhuh:
squashed!
/hold cancel

mboersma (approved):
/lgtm
/approve

PR #3936, opened 2023-09-01
Convert resourceskus service to SDKv2

mboersma:
**What type of PR is this?**

/kind cleanup

**What this PR does / why we need it**:

Converts the resource skus service to azure-sdk-for-go version 2.

**Which issue(s) this PR fixes**:

Fixes #3920

**Special notes for your reviewer**:

This service doesn't provide "Create" or "Delete" functionality, it just lists and gets existing SKUs.

- [ ] cherry-pick candidate

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
NONE
```

PR #3939, opened 2023-09-01
Convert tags service to SDKv2

### LINKED ISSUE #3927: Update tags service to SDK v2

mboersma:
**What type of PR is this?**

/kind cleanup

Converts the tags service to azure-sdk-for-go version 2 and the `ascynpoller` framework for long-running operations.

Fixes #3927

- [ ] cherry-pick candidate

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/azure/scope/managedcontrolplane_test.go
+++ b/azure/scope/managedcontrolplane_test.go
@@ -467,6 +467,12 @@ func TestManagedControlPlaneScope_OSType(t *testing.T) {
 	}
 }
 
+type fakeVnetDescriber struct{}
+
+func (f fakeVnetDescriber) IsManaged(ctx context.Context) (bool, error) {
+	return false, nil
+}
+
 func TestManagedControlPlaneScope_IsVnetManagedCache(t *testing.T) {
 	scheme := runtime.NewScheme()
 	_ = expv1.AddToScheme(scheme)
@@ -509,6 +515,7 @@ func TestManagedControlPlaneScope_IsVnetManagedCache(t *testing.T) {
 						InfraMachinePool: getLinuxAzureMachinePool("pool1"),
 					},
 				},
+				VnetDescriber: fakeVnetDescriber{},
 			},
 			Expected: false,
 		},
--- a/azure/services/tags/client.go
+++ b/azure/services/tags/client.go
@@ -19,51 +19,61 @@ package tags
 import (
 	"context"
 
-	"github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2019-10-01/resources"
-	"github.com/Azure/go-autorest/autorest"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armresources"
+	"github.com/pkg/errors"
 	"sigs.k8s.io/cluster-api-provider-azure/azure"
 	"sigs.k8s.io/cluster-api-provider-azure/util/tele"
 )
 
 // client wraps go-sdk.
 type client interface {
-	GetAtScope(context.Context, string) (resources.TagsResource, error)
-	UpdateAtScope(context.Context, string, resources.TagsPatchResource) (resources.TagsResource, error)
+	GetAtScope(context.Context, string) (armresources.TagsResource, error)
+	UpdateAtScope(context.Context, string, armresources.TagsPatchResource) (armresources.TagsResource, error)
 }
 
 // AzureClient contains the Azure go-sdk client.
 type AzureClient struct {
-	tags resources.TagsClient
+	tags *armresources.TagsClient
 }
 
 var _ client = (*AzureClient)(nil)
 
-// NewClient creates a new tags client from subscription ID.
-func NewClient(auth azure.Authorizer) *AzureClient {
-	c := newTagsClient(auth.SubscriptionID(), auth.BaseURI(), auth.Authorizer())
-	return &AzureClient{c}
-}
-
-// newTagsClient creates a new tags client from subscription ID.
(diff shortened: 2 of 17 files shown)


mboersma:
/retitle [WIP] Convert tags service to SDKv2

There's a panic that shows up in a unit test I still haven't figured out. Looks to be the same issue as #3936. 🤔

azure/services/tags/client.go
```diff
@@ -19,51 +19,68 @@ package tags
...
 // GetAtScope sends the get at scope request.
-func (ac *AzureClient) GetAtScope(ctx context.Context, scope string) (resources.TagsResource, error) {
+func (ac *AzureClient) GetAtScope(ctx context.Context, scope string) (tagsResource armresources.TagsResource, err error) {
 	ctx, _, done := tele.StartSpanWithLogger(ctx, "tags.AzureClient.GetAtScope")
 	defer done()
 
-	return ac.tags.GetAtScope(ctx, scope)
+	// Handle a panic from the SDK by returning an error. This only happens in a managedcontrolplane unit test
+	// where the tags client is not mocked.
+	defer func() {
```
mboersma, line 61: I debugged the path of the `TestManagedControlPlaneScope_IsVnetManagedCache` test case "no Cache value", and it eventually leads here. In SDKv1 that led to returning an error when network communication failed, but in SDKv2 it panics.

It doesn't look easy to inject a mock into this code path, and there don't seem to be any accessible `nil` fields that we could fill in to placate the test environment. I'm not proud of it, but after several hours in the debugger I think this might be the expedient solution.

nojnhuh: With ASO we won't need this tags service at all, so I think relatively dirty hacks here are acceptable.

controllers/azuremanagedcontrolplane_reconciler.go
```diff
@@ -78,8 +82,8 @@ func newAzureManagedControlPlaneReconciler(scope *scope.ManagedControlPlaneScope
 			subnetsSvc,
 			managedClustersSvc,
 			privateEndpointsSvc,
-			tags.New(scope),
 			resourceHealthSvc,
+			tagsSvc,
 		},
```
nojnhuh, line 87: Was flipping the order here intentional? In this order, if a cluster gets stuck in an unhealthy state as flagged by resource health, then its tags won't update until it becomes healthy again.

azure/services/tags/client.go
nojnhuh: ^ Seeing now that this is used more widely than I first thought, I shouldn't be too sure about that but I'm still optimistic.

controllers/azuremanagedcontrolplane_reconciler.go
mboersma: No, it was probably a mistake in rebasing. I'll restore the order.

nojnhuh (commented):
/lgtm

azure/services/tags/client.go
CecileRobertMichon: not sure I love this either... I would rather us remove the test or change the test to make it work with this new code rather than hack the code just to make the unit test pass

mboersma: It seems like a valid test, and I've tried pretty hard to find a way to rework it. I can look again.

mboersma: I created a `VnetDescriber` interface so we could mock `virtualNetworksSvc.IsManaged()`. This client creates that service just-in-time, unless we pass one into the scope.

(What does mocking virtualnetworks in a managedcluster unit test have to do with refactoring the tags service? Nothing directly, but this short-circuits the code path before it gets to the failing tags call.)

No need to panic. :-) This isn't the cleanest change but it's better. Let me know what you think.

CecileRobertMichon: Thank you, this is much better (and we're actually testing the thing the test intends to test).

Will let @nojnhuh chime in 

/lgtm

CecileRobertMichon:
/hold

in case you intend on squashing

mboersma:
/hold cancel

I squashed and made a small change to have the mock return an error, which more closely simulates the code path followed in the SDKv1 version of this test.

azure/scope/managedcontrolplane_test.go
```diff
@@ -467,6 +468,12 @@ func TestManagedControlPlaneScope_OSType(t *testing.T) {
 	}
 }
 
+type fakeVnetDescriber struct{}
+
+func (f fakeVnetDescriber) IsManaged(ctx context.Context) (bool, error) {
+	return false, errors.New("The subscription '00000000-0000-0000-0000-000000000000' could not be found.")
```
CecileRobertMichon, line 474: even though this is closer to what was happening before, I'm not sure this is actually what the test intends to test. The fact that we're swallowing the error is not great, but probably not in scope for this PR...

mboersma: Agreed, I think the SDKv1 version of this test wasn't intentionally following this path, just getting away with it. I can drop the error if you prefer.

CecileRobertMichon: I'm +1 for dropping the error

If we want to test the error scenario that should be a separate test case but not sure that's necessary since we're not handling the error, just logging it

mboersma: Done.

CecileRobertMichon (commented):
/lgtm
/assign @nojnhuh

azure/services/tags/client.go
nojnhuh: +1, this approach lgtm.

PR #3952, opened 2023-09-05
Convert virtualnetworks service to SDKv2

### LINKED ISSUE #3925: Update virtualnetworks service to SDK v2

mboersma:
**What type of PR is this?**

/kind cleanup

Converts the virtualnetworks service to azure-sdk-for-go version 2 and the `asyncpoller` framework.

Fixes #3925

- [ ] cherry-pick candidate

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/azure/converters/extendedlocation.go
+++ b/azure/converters/extendedlocation.go
@@ -17,11 +17,23 @@ limitations under the License.
 package converters
 
 import (
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork/v4"
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2021-11-01/compute"
 	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2021-08-01/network"
 	"k8s.io/utils/ptr"
 	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
 )
+
+// ExtendedLocationToNetworkSDKv2 converts an infrav1.ExtendedLocationSpec to an armnetwork.ExtendedLocation.
+func ExtendedLocationToNetworkSDKv2(src *infrav1.ExtendedLocationSpec) *armnetwork.ExtendedLocation {
+	if src == nil {
+		return nil
+	}
+	return &armnetwork.ExtendedLocation{
+		Name: ptr.To(src.Name),
+		Type: ptr.To(armnetwork.ExtendedLocationTypes(src.Type)),
+	}
+}
 
 // ExtendedLocationToNetworkSDK converts infrav1.ExtendedLocationSpec to network.ExtendedLocation.
 func ExtendedLocationToNetworkSDK(src *infrav1.ExtendedLocationSpec) *network.ExtendedLocation {
--- a/azure/converters/subnets.go
+++ b/azure/converters/subnets.go
@@ -35,13 +35,15 @@ func GetSubnetAddresses(subnet network.Subnet) []string {
 }
 
 // GetSubnetAddressesV2 returns the address prefixes contained in an SDK v2 subnet.
-func GetSubnetAddressesV2(subnet armnetwork.Subnet) []string {
+func GetSubnetAddressesV2(subnet *armnetwork.Subnet) []string {
 	var addresses []string
 	if subnet.Properties != nil && subnet.Properties.AddressPrefix != nil {
 		addresses = []string{ptr.Deref(subnet.Properties.AddressPrefix, "")}
 	} else if subnet.Properties != nil && subnet.Properties.AddressPrefixes != nil {
 		for _, address := range subnet.Properties.AddressPrefixes {
-			addresses = append(addresses, ptr.Deref(address, ""))
+			if address != nil {
+				addresses = append(addresses, *address)
+			}
 		}
 	}
 	return addresses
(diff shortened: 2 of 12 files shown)


azure/converters/extendedlocation.go
```diff
@@ -17,12 +17,27 @@ limitations under the License.
...
 )
 
+// ExtendedLocationToNetworkSDKv2 converts an infrav1.ExtendedLocationSpec to an armnetwork.ExtendedLocation.
+func ExtendedLocationToNetworkSDKv2(src *infrav1.ExtendedLocationSpec) *armnetwork.ExtendedLocation {
+	if src == nil {
+		return nil
+	}
+	if src.Type == "Edge" {
+		src.Type = "EdgeZone"
+	}
```
nojnhuh, line 34: Out of curiosity, when might we expect to reach this case? It looks like the CAPZ API has only ever allowed `EdgeZone` as a valid value here.

azure/converters/subnets.go
```diff
@@ -17,11 +17,27 @@ limitations under the License.
...
 )
 
+// GetSubnetAddressesV2 returns the address prefixes contained in a subnet.
+func GetSubnetAddressesV2(subnet *armnetwork.Subnet) []string {
+	var addresses []string
+	if subnet.Properties != nil && subnet.Properties.AddressPrefix != nil {
+		addresses = []string{ptr.Deref(subnet.Properties.AddressPrefix, "")}
+	} else if subnet.Properties != nil && subnet.Properties.AddressPrefixes != nil {
+		for _, prefix := range subnet.Properties.AddressPrefixes {
+			if ptr.Deref(prefix, "") != "" {
```
nojnhuh, line 33: nit: I wonder if it might match the previous behavior a tad more closely if we only check that the value here isn't nil:
```suggestion
			if prefix != nil {
```
I think the only case this affects is where `prefix` is a non-nil, empty string. I'm not sure if that scenario is even possible though.

azure/scope/managedcontrolplane.go
```diff
@@ -402,7 +402,11 @@ func (s *ManagedControlPlaneScope) IsVnetManaged() bool {
 	ctx := context.Background()
 	ctx, log, done := tele.StartSpanWithLogger(ctx, "scope.ManagedControlPlaneScope.IsVnetManaged")
 	defer done()
-	isManaged, err := virtualnetworks.New(s).IsManaged(ctx)
+	virtualNetworksSvc, err := virtualnetworks.New(s)
+	if err != nil {
+		log.Error(err, "failed to create virtualnetworks service")
+	}
+	isManaged, err := virtualNetworksSvc.IsManaged(ctx)
```
nojnhuh, line 409: `IsManaged` looks like it will panic when `virtualNetworksSvc` is nil, so I think we might need to return early if `virtualnetworks.New` returns an error. I'm not sure what the most reasonable bool to return would be though.

azure/services/virtualnetworks/virtualnetworks.go
```diff
@@ -84,26 +89,30 @@ func (s *Service) Reconcile(ctx context.Context) error {
...
 		vnet := s.Scope.Vnet()
 		vnet.ID = ptr.Deref(existingVnet.ID, "")
 		vnet.Tags = converters.MapToTags(existingVnet.Tags)
 
 		var prefixes []string
-		if existingVnet.VirtualNetworkPropertiesFormat != nil && existingVnet.VirtualNetworkPropertiesFormat.AddressSpace != nil {
-			prefixes = azure.StringSlice(existingVnet.VirtualNetworkPropertiesFormat.AddressSpace.AddressPrefixes)
+		if existingVnet.Properties != nil && existingVnet.Properties.AddressSpace != nil {
+			for _, prefix := range existingVnet.Properties.AddressSpace.AddressPrefixes {
+				if ptr.Deref(prefix, "") != "" {
```
nojnhuh, line 103: (same as above in `converters.GetSubnetAddressesV2`)

azure/converters/extendedlocation.go
mboersma: I think I was led astray by the existing (SDK v1) unit test that uses "Edge" as a test value. But it doesn't appear we'll see this in practice, so I removed this logic.

azure/scope/managedcontrolplane.go
mboersma: It seems like the effective default is `false`, so I changed it to log and return that.

It would be great to return an error here, but this func is called from a hundred different places.

azure/services/virtualnetworks/virtualnetworks.go
mboersma: Thank you so much for reviewing @nojnhuh! I think I've addressed all your comments.

controllers/azuremanagedcontrolplane_reconciler.go
```diff
@@ -65,12 +65,18 @@ func newAzureManagedControlPlaneReconciler(scope *scope.ManagedControlPlaneScope
...
 		kubeclient: scope.Client,
 		scope:      scope,
 		services: []azure.ServiceReconciler{
 			groupsService,
-			virtualnetworks.New(scope),
+			subnetsSvc,
+			managedClustersSvc,
+			virtualNetworksSvc,
 			subnetsSvc,
 			managedClustersSvc,
```
nojnhuh, line 81: Looks like some merge conflict resolution funk in here.

nojnhuh (commented):
/lgtm

CecileRobertMichon (commented):
/lgtm
/approve

nojnhuh (commented):
/lgtm

PR #3961, opened 2023-09-06
Add ability to specify multiple sources for a security rule

### LINKED ISSUE #3377: securityRules should support source address in plural
/kind feature

As a _User_ i want to group multiple network sources into one `securityRule`. At the moment i have to specify for each source one `securityRule`.

In theory it should be an easy one to extend the `securityRule` struct by a new field called `Sources` and do the conversion to [`SourceAddressPrefixes`](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/services/network/mgmt/2021-08-01/network#SecurityRulePropertiesFormat) in [here](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/79c47714bfd637d9d979494af08310e712c855cf/azure/converters/rules.go#L26-L38).

But maybe it's also worth the discussion to keep the existing `Source` field and convert it directly to `SourceAddressPrefixes` as at least in theory, on Azure side nothing will (should) change.

**Environment:**

- cluster-api-provider-azure version: `1.9.0`

nawazkh:
**What type of PR is this?**
/kind feature

- This PR will let users add multiple `Sources` to a single security rule.

Fixes #3377 

- [ ] cherry-pick candidate

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
add ability to specify multiple sources for a security rule
```


CecileRobertMichon:
@nawazkh is this still wip ?

nawazkh:
> @nawazkh is this still wip ?

Most of the code I was thinking of is in place, will give it a test and update its status.

CecileRobertMichon:
Sounds good, PR is still marked as draft so make sure you publish it if it's ready

nawazkh:
Tested it locally by adding multiple `sources` to a _dummy_ security rule allowing all the inbound connections to my control-plane NSG, Works as expected. 
Please take a look!

mboersma (commented):
/lgtm

I was going to ask whether we should deprecate `role.Source`, but I see there are both `SourceAddressPrefix` and -`Prefixes` in the API, so :shrug:.

api/v1beta1/types.go
```diff
@@ -305,6 +305,8 @@ type SecurityRule struct {
 	// Source specifies the CIDR or source IP range. Asterix '*' can also be used to match all source IPs. Default tags such as 'VirtualNetwork', 'AzureLoadBalancer' and 'Internet' can also be used. If this is an ingress rule, specifies where network traffic originates from.
 	// +optional
 	Source *string `json:"source,omitempty"`
+	// Sources specifies The CIDR or source IP ranges.
+	Sources []*string `json:"sources,omitempty"`
```
willie-yao, line 309: Just a clarifying question. Sources is a separate field from Source because the SDK separates it into two fields? I'm a bit confused why that is the case.

willie-yao (approved):
/lgtm

Just a clarifying comment from my end!

nawazkh:
> I was going to ask whether we should deprecate role.Source

I am +1 on deprecating `Source` as well. Makes the UX more unified with the one seen on the Azure Portal. 
Will bring it up in our CAPZ office hours as a callout on deprecating `Source`.

api/v1beta1/types.go
nawazkh: You are right. 
The Azure SDK provides two ways to specify the _Source_ for a security rule. One being `Source` and other being `Sources` for multiple sources for the same rule. 
We see that when converting CAPZ API to Azure SDK one as here https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3961/files#diff-f1bf1dc0cddcc772a47a442e0d748e6f41bf6f15a502cc138cad235de2bf2d74R32.

So as to why is that the case?

I would bet convenience? 
At Azure Portal, a user can specify a single Source in a rule or have a rule with multiple sources provided they are comma separated. 
So thinking about it, it may have been that Azure first released the API with Source and then added Sources later on to support comma separated Sources behavior.

mboersma (approved):
/lgtm
/approve

PR #3965, opened 2023-09-07
test: fix delete test subnet logic

jackfrancis:
**What type of PR is this?**

/kind failing-test

This PR updates the way we re-use and mutate test subnets in the test flow so that we can revert to the original properly to implement a cleanup at the end.

Fixes #

- [ ] cherry-pick candidate

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/azure_securitygroups.go
+++ b/test/e2e/azure_securitygroups.go
@@ -110,6 +110,7 @@ func AzureSecurityGroupsSpec(ctx context.Context, inputGetter func() AzureSecuri
 		Name:      input.Cluster.Spec.InfrastructureRef.Name,
 	}, azureCluster)
 	Expect(err).NotTo(HaveOccurred())
+	originalSubnets := azureCluster.Spec.NetworkSpec.Subnets
 
 	var expectedSubnets infrav1.Subnets
 	checkSubnets := func(g Gomega) {
@@ -146,8 +147,11 @@ func AzureSecurityGroupsSpec(ctx context.Context, inputGetter func() AzureSecuri
 			},
 		},
 	}
-	originalSubnets := azureCluster.Spec.NetworkSpec.Subnets
-	expectedSubnets = originalSubnets
+
+	expectedSubnets = infrav1.Subnets{}
+	for _, subnet := range originalSubnets {
+		expectedSubnets = append(expectedSubnets, subnet)
+	}
 	expectedSubnets = append(expectedSubnets, testSubnet)
 	Eventually(func(g Gomega) {
 		g.Expect(mgmtClient.Get(ctx, client.ObjectKeyFromObject(azureCluster), azureCluster)).To(Succeed())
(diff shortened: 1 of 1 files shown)


nojnhuh:
Just to help my understanding, is there a kind of before/after comparison that shows the effects of this change? It's not immediately obvious to me exactly what behavioral difference this makes.

And is this addressing this kind of test flake? https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/periodic-cluster-api-provider-azure-e2e-main/1694427233935953920

jackfrancis:
@nojnhuh I threw up a quick experiment yesterday EOD to see if in the flow of subnet create/delete we were unexpectedly following a single reference, instead of doing a copy.

What's really happening is the `"Deleting security rule from the subnet"` part of the flow isn't working (though the test is passing).

After we exit from the `AzureSecurityGroupsSpec` func the newly added "test-subnet" still exists in the `AzureCluster` resource, which has the side-effect of preventing `AzureMachines` in the cluster from scaling out, because the test template we're using for those machines doesn't explicitly name a subnet. In such a scenario you rely upon there only being one subnet for nodes, and the new azuremachines/nodes are added to that one subnet. If there's more than one subnet then the new machine/node doesn't know which subnet to use (again, because the `AzureMachineTemplate` doesn't explicitly declare a subnet) and the new replica azuremachines are never created.

nojnhuh:
Ok, that makes sense. I'm still not entirely convinced it's a pointer/reference thing though. What exactly was the experiment you ran and what were the results?

I wonder if the issue is that `checkSubnets` never checks that the subnet itself gets deleted, only the rules within the subnets. If more subnets exist than are expected, it looks like `checkSubnets` will skip over them entirely.

fyi @willie-yao in case you have other ideas.

nojnhuh:
I ran a quick test and see that deleting a subnet from AzureCluster's `spec.networkSpec.subnets` indeed doesn't delete the subnet in Azure since I think the subnets service logic is only "make sure the specified subnets are up to date" and deletion only kicks in when the whole AzureCluster gets deleted. Not sure if it would break existing users to change this behavior though if they're creating custom subnets.

One short-term hack to workaround this particular issue might be to set the test subnet's `role` to `bastion`.

jackfrancis:
This needs to be converted into an issue describing the bug. (This PR doesn't fix the underlying issue.)

nojnhuh:
> This needs to be converted into an issue describing the bug. (This PR doesn't fix the underlying issue.)

@jackfrancis Have you opened that issue? And is there value in keeping this PR open if so?

nojnhuh:
> This needs to be converted into an issue describing the bug. (This PR doesn't fix the underlying issue.)

@jackfrancis I just opened #4338.

Issue #3955, opened 2023-09-05
AKS e2e test flakes in the tags tests with a "not found" error


nojnhuh:
**Which jobs are flaky:**
- pull-cluster-api-provider-azure-e2e-aks

**Which tests are flaky:**
- Workload cluster creation Creating an AKS cluster [Managed Kubernetes] with a single control plane node and 1 node

**Testgrid link:**
- https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/3971/pull-cluster-api-provider-azure-e2e-aks/1704200991722704896
- https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/3975/pull-cluster-api-provider-azure-e2e-aks/1700198722740162560
- https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/3946/pull-cluster-api-provider-azure-e2e-aks/1699032232213090304
- https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/3935/pull-cluster-api-provider-azure-e2e-aks/1697661825484591104
- https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/4094/pull-cluster-api-provider-azure-e2e-aks/1710391381085655040

**Reason for failure (if possible):**

**Anything else we need to know:**

/kind flake

/area managedclusters

tapojit047:
Is this issue resolved? I am facing similar kind of issue in my PR (#3974)[https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3974]

Testgrid link:
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/3974/pull-cluster-api-provider-azure-e2e-aks/1700477335679012864

PR #3966, opened 2023-09-07
Convert scalesets service to SDKv2

### LINKED ISSUE #3942: Update scalesets service to SDK v2

mboersma:
**What type of PR is this?**

/kind cleanup

Converts the scalesets service to azure-sdk-for-go version 2 and the `asyncpoller` framework.

Fixes #3942

This PR is a real balancing act; I hope it works! 

To avoid converting the whole enchilada at once, this converts scalesets to SDK v2 but *not* scaleset VMs. So there are several places where SDK v1 and v2 are being used side-by-side. This will go away once subsequent PRs for scaleset VMs and virtual machines.

- [ ] cherry-pick candidate

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
NONE
```


azure/services/scalesets/client.go
```diff
@@ -97,166 +104,128 @@ func (ac *AzureClient) ListInstances(ctx context.Context, resourceGroupName stri
...
+	pollOpts := &runtime.PollUntilDoneOptions{Frequency: asyncpoller.DefaultPollerFrequency}
+	_, err = poller.PollUntilDone(ctx, pollOpts)
 	if err != nil {
+		// if an error occurs, return the poller.
+		// this means the long-running operation didn't finish in the specified timeout.
 		return err
 	}
-	_, err = future.Result(ac.scalesets)
+
+	// TODO: Retry this operation internally to this function so it appears synchronous
```
mboersma, line 196: /hold

Still need to figure this out. It's an async SDK function now, but the `asyncpoller` framework only handles CreateOrUpdate and Delete.

nojnhuh: Is this even being used at all? A plain grep for `UpdateInstances` on main only really shows this definition and the mock.

mboersma: Perhaps it isn't used: the relevant e2e tests passed even though the current implementation is probably broken. I'll see about removing this code.

mboersma: /hold cancel

Yep, `UpdateInstances` was unused and is now gone. Thanks @nojnhuh! I think this is ready for review now.

azure/converters/diagnostics.go
```diff
@@ -17,11 +17,41 @@ limitations under the License.
...
 
 import (
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute/v5"
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2021-11-01/compute"
 	"k8s.io/utils/ptr"
 	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
 )
 
+// GetDiagnosticsProfileV2 converts a CAPZ Diagnostics option to a Azure SDK Diagnostics Profile.
+func GetDiagnosticsProfileV2(diagnostics *infrav1.Diagnostics) *armcompute.DiagnosticsProfile {
```
mboersma, line 27: To point out one example of this "clone until everything is converted" strategy, `GetDiagnosticsProfile` is called by both the scalesets and virtualmachines services. We're converting scalesets, but don't want to convert virtualmachines yet.

So we clone the function, add a `V2` suffix, and change it to use `armcompute` types instead of old `compute` types. We also clone its unit tests.

Then when both scalesets and virtualmachines have converted to SDK v2, we get rid of the old SDK v1-based function that's no longer used, and we rename the -`V2` function back to the original name.

azure/converters/vmss.go
```diff
@@ -34,20 +35,20 @@ const (
 )
 
 // SDKToVMSS converts an Azure SDK VirtualMachineScaleSet to the AzureMachinePool type.
-func SDKToVMSS(sdkvmss compute.VirtualMachineScaleSet, sdkinstances []compute.VirtualMachineScaleSetVM) azure.VMSS {
+func SDKToVMSS(sdkvmss armcompute.VirtualMachineScaleSet, sdkinstances []compute.VirtualMachineScaleSetVM) azure.VMSS {
```
mboersma, line 38: This is weird: mixing SDKv2 and SDKv1 parameters. But it seems to work and allows us to keep the whole virtualmachines service conversion separate, for a future PR.

azure/converters/spotinstances_test.go
```diff
@@ -20,13 +20,110 @@ import (
...
+				},
+			},
+		},
+		{
+			name: "spot with ephemeral disk",
+			spot: &infrav1.SpotVMOptions{
+				MaxPrice: nil,
+			},
+			diffDiskSettings: &infrav1.DiffDiskSettings{
+				Option: string(compute.DiffDiskOptionsLocal),
```
nojnhuh, line 89:
```suggestion
				Option: string(armcompute.DiffDiskOptionsLocal),
```

azure/services/scalesets/client.go
```diff
@@ -18,47 +18,47 @@ package scalesets
...
 // AzureClient contains the Azure go-sdk Client.
 type AzureClient struct {
 	scalesetvms compute.VirtualMachineScaleSetVMsClient
-	scalesets   compute.VirtualMachineScaleSetsClient
+	scalesets   *armcompute.VirtualMachineScaleSetsClient
 }
 
 var _ Client = &AzureClient{}
 
 // NewClient creates a new VMSS client from subscription ID.
```
nojnhuh, line 52:
```suggestion
// NewClient creates a new VMSS client from an authorizer.
```

azure/services/scalesets/spec.go
```diff
@@ -206,39 +207,39 @@ func (s *ScaleSetSpec) Parameters(ctx context.Context, existing interface{}) (pa
 	// Set properties specific to VMSS orchestration mode
 	// See https://learn.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-orchestration-modes for more details
 	switch orchestrationMode {
-	case compute.OrchestrationModeUniform: // Uniform VMSS
-		vmss.VirtualMachineScaleSetProperties.Overprovision = ptr.To(false)
-		vmss.VirtualMachineScaleSetProperties.UpgradePolicy = &compute.UpgradePolicy{Mode: compute.UpgradeModeManual}
-	case compute.OrchestrationModeFlexible: // VMSS Flex, VMs are treated as individual virtual machines
-		vmss.VirtualMachineScaleSetProperties.VirtualMachineProfile.NetworkProfile.NetworkAPIVersion =
-			compute.NetworkAPIVersionTwoZeroTwoZeroHyphenMinusOneOneHyphenMinusZeroOne
```
nojnhuh, line 214: wow lol

mboersma:
/retest

#3955

azure/converters/spotinstances_test.go
nojnhuh: Actually I can see where we might technically want the old type here even though they're equivalent.

mboersma: Grrr... I think you're right again. 🤦🏻 

The unit test passes either way, however, so is it worth changing now? We will soon pave over all usage of `compute.` anyway...

nojnhuh: I'm pretty sure it's not gonna bite us either way, but I guess my gut is leaning slightly in favor of the `armcompute` variant.

nojnhuh (commented):
/lgtm

mboersma:
This failed on #3955, but passed twice previously.

@nojnhuh @CecileRobertMichon can we `/override` the AKS test?

nojnhuh:
+1 to overriding

mboersma:
/override pull-cluster-api-provider-azure-e2e-aks

Mea culpa, but there are a couple other PRs behind this that it would be great to update and run today.

Issue #3855, opened 2023-08-16
Remove backfilling to AzureCluster spec from vnet reconciliation


nojnhuh:
/kind api-change

**What needs cleanup:**

During reconciliation of virtual network resources, several fields under an AzureCluster's `spec.networkSpec.vnet` are backfilled with "actual" state gathered from the Azure API.

These fields include:
- `id`
- `tags`
- `cidrBlocks`

And under `spec.networkSpec.subnets`:
- `cidrBlocks`

Populating `spec` fields from a resource's own controller this way is considered an anti-pattern [citation needed, but IIRC mostly because of the not-entirely-philosophical "`spec` = desired state, `status` = actual state" distinction].

Relevant code: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/e7773fe017f3493be65c3368808d726a6343911e/azure/services/virtualnetworks/virtualnetworks.go#L91-L108

In addition to following Kubernetes best practices, part of the motivation for this change is to simplify the ASO implementation for reconciling vnets and remove a dependency on implementing ASO for subnets at the same time as for vnets.

**Describe the solution you'd like**

Instead of populating fields in `spec` representing "actual" state, the same info would fit better in an AzureCluster's `status` field. There, it would still be available from the API but not subject to the same validation which triggered the issue fixed by #2339 and would more clearly not collide with any user-specified values.

**Anything else you would like to add:**
This is to some degree a breaking change to the API, where users may currently be relying on the current behavior.

CecileRobertMichon:
related to #1685

Issue #3629, opened 2023-06-15
Automate AKS features available in ASO


dtzar:
/kind feature

**Describe the solution you'd like**
As a user of CAPZ I would like the ability to provision all features available in AKS as soon as they are made available.  

Even once we switch to the dependency of using ASO for AKS provisioning, all features present in ASO will take individual coding work to be able to be utilized in CAPZ.  Since the features in AKS rapidly update ([see comment](https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/2625#issuecomment-1557652425) on #2625), it would be ideal to have a mechanism for people to access all the available options in ASO via CAPZ via some coding mechanism which auto generates the code for it to work.

dtzar:
Prerequisites for this are #2625 and #3529

PR #3973, opened 2023-09-08
add oidcIssuerProfile to AzureManagedControlPlane

### LINKED ISSUE #2498: Add OIDCIssuerProfile configuration option
/kind feature

In order to enable Azure Workload Identity, the OIDC issuer needs to be enabled on new clusters. Add a field AzureManagedControlPlane for enabling OIDC issuer.

Basically just a configuration option that does the same as this cli command https://docs.microsoft.com/en-us/azure/aks/cluster-configuration#update-an-aks-cluster-with-oidc-issuer

**Environment:**

- cluster-api-provider-azure version: v1.4.0
- Kubernetes version: (use `kubectl version`): v1.22

nojnhuh:
**What type of PR is this?**
/kind feature
/area managedclusters

This PR exposes the AKS `oidcIssuerProfile` config from AzureManagedControlPlane at `spec.oidcIssuerProfile`. It also adds `status.oidcIssuerProfile` which includes the current actual state of `enabled` and the `issuerURL`.

Fixes #2498

- [ ] cherry-pick candidate

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Release note**:

```release-note
Added `spec.oidcIssuerProfile` to AzureManagedControlPlane
```


maciaszczykm (approved):
Really nice to have that! I hope it can be included in the incoming release.

CecileRobertMichon:
/milestone v1.11

@maciaszczykm yes that's the plan

api/v1beta1/azuremanagedcontrolplane_types.go
```diff
@@ -349,6 +353,21 @@ type AzureManagedControlPlaneStatus struct {
...
+	// OIDCIssuerProfile is the OIDC issuer profile of the Managed Cluster.
+	// +optional
+	OIDCIssuerProfile *OIDCIssuerProfileStatus `json:"oidcIssuerProfile,omitempty"`
+}
+
+// OIDCIssuerProfileStatus is the OIDC issuer profile of the Managed Cluster.
+type OIDCIssuerProfileStatus struct {
+	// Enabled is whether the OIDC issuer is enabled.
+	// +optional
+	Enabled *bool `json:"enabled,omitempty"`
```
CecileRobertMichon, line 366: this is interesting

It differs a bit from existing patterns we have for setting "read-only" fields (eg. kubelet identity). This is arguably better and more aligned with k8s pattern (spec is desired state and status is read-only actual state) but inconsistent with other fields we already have. I'm not sure about duplicating the enabled field in both spec and status, that strikes me as redundant.

What was the reasoning in including the status field?

nojnhuh: Overall this was done in the spirit of #3855. ASO's ManagedCluster type is also structured the same way. One concrete case that having `enabled` in the status helps with is the webhook validation to return an error when a user tries to disable the OIDC issuer, which is not allowed by AKS.

The field itself is passed through verbatim from CAPZ to the AKS API, where `nil` means "don't change" and non-nil `true` and `false` set it explicitly. That means when `enabled` is nil in the AzureManagedControlPlane spec, that by itself is not enough to know whether or not the OIDC issuer is actually enabled at that moment. Since AKS (always, AFAICT) populates `enabled` in its API when the user sends `nil`, propagating that to the status in AzureManagedControlPlane lets the webhook definitively know the actual state of `enabled` when validating an update.

CecileRobertMichon: What's the advantage of letting the user set the `spec` field to "nil"? Would the same enforcement be covered by a simpler immutable check (once it's set to true, cannot be unset or set to false)?

nojnhuh: Allowing `nil` means users that have already set this on their clusters via other means won't have this change out from under them when CAPZ upgrades. I believe @mtougeron expressed this kind of concern a while ago around the autoscaler settings.

mtougeron: agreed that the consistency would be nice. I can't remember though with the cluster-autoscaler if this was due to the way that Azure handled the nil values when passed or if it was because I didn't have a way to "reset" to the default values via an API when the user unset the field in CAPZ. Having this flag would make it more about CAPZ managing the resource/fields and that problem would pretty much go away.

nojnhuh: Opened #3990 to track this.

nojnhuh: And if `spec.oidcIssuerProfile` is null in the spec, should that always imply `spec.oidcIssuerProfile.enabled` is false?

CecileRobertMichon: @nojnhuh, no the danger with using a plain non-pointer is that the default could change and then we wouldn't know if a user is intentionally disabling it or not setting the field.

To summarize there are two options/patterns possible:

A) The spec field is an optional pointer bool and defaults the value to a sane default when nil. We don't support users setting things outside of the spec, the CAPZ spec is the only source of truth for the object. This is what we do for most fields if not all currently.
B) The spec field is an optional pointer bool, we don't default the value and let "nil" be a valid value to pass to the AKS API. We use a status field to track the "actual" value in Azure. This allows users to set this value outside of the CAPZ spec and not get overwritten, as long as AKS doesn't default/require a value. This is what this PR is doing currently, and what ASO does.

IMO both are valid. I tend to prefer consistency overall because that is less likely to lead users to make mistakes. This is not a hill I'm willing to die on (and I really want to see this PR get merged so I don't want to be standing in the way of that) but my personal preference would be to keep A) for now for consistency with the agreement that B) is what we're trying to evolve towards with a v2 of the API that wraps around ASO types (https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/3629).

nojnhuh: > my personal preference would be to keep A) for now for consistency with the agreement that B) is what we're trying to evolve towards with a v2 of the API that wraps around ASO types

Agree this seems like the best path for now. I'll update this.

nojnhuh:
Updated this so it should be more consistent with other fields.

/hold for squash

test/e2e/azure_test.go
```diff
@@ -839,15 +839,6 @@ var _ = Describe("Workload cluster creation", func() {
 					}
 				})
 			})
-
-			By("modifying OIDC issuer configuration", func() {
```
nojnhuh: If we don't keep the "actual" state in the status, then we have to wait to add this test back until we can use SDK v2 clients in the tests.

api/v1beta1/azuremanagedcontrolplane_types.go
nojnhuh: @CecileRobertMichon This should work the way you described now. Can we still get this in for the v1.11 release?

CecileRobertMichon (commented):
/lgtm
/approve

api/v1beta1/azuremanagedcontrolplane_default.go
```diff
@@ -144,3 +144,13 @@ func (m *AzureManagedControlPlane) setDefaultAutoScalerProfile() {
 		m.Spec.AutoScalerProfile.SkipNodesWithSystemPods = (*SkipNodesWithSystemPods)(ptr.To(string(SkipNodesWithSystemPodsTrue)))
 	}
 }
+
+func (m *AzureManagedControlPlane) setDefaultOIDCIssuerProfile() {
+	if m.Spec.OIDCIssuerProfile == nil {
+		m.Spec.OIDCIssuerProfile = &OIDCIssuerProfile{}
```
jackfrancis, line 150: I think this thread covers this: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3973#pullrequestreview-1618500917

But, just to clarify, the reason we're doing it this way instead of the way that autoscaler does it above (if user doesn't declare a profile config we omit it from the spec entirely) is because this more explicit "is disabled" implementation removes edge cases from upgrade scenarios?

CecileRobertMichon: @jackfrancis this comment summarizes it https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3973#discussion_r1324774308

nojnhuh: I think because autoscaler is a little different from other fields in that we want to send `nil` over the wire to AKS when not defined in the CAPZ spec. OIDC issuer is more like other fields where we always send a non-nil value.

nojnhuh:
squashed!
/hold cancel

PR #3974, opened 2023-09-08
Add proper validation for nodepool name

### LINKED ISSUE #3968: Set a unique valid default name for nodepool when spec.name in AzureManagedMachinePool is not specified
/kind bug

- Create an AKS Cluster using CAPZ
- Create a new `MachinePool` without specifying optional field `spec.name` in `AzureManagedMachinePool`
- Got the error below:
```
 message: 'agentpools failed to create or update. err: failed to create resource
      tapo-test/multiple-zone-dedicated (service: agentpools): failed to begin operation:
      Code="InvalidParameter" Message="The value of parameter agentPoolProfile.name
      is invalid. Please see https://aka.ms/aks-naming-rules for more details." Target="agentPoolProfile.name"'
    reason: Failed
```

If I do not mention the `spec.name` field in `AzureManagedMachinePool`, it should be able to generate a default unique `spec.name` such that the `MachinePool` is created.

- The error is raised [here](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/api/v1beta1/azuremanagedmachinepool_webhook.go#L74-L75C3)
- It is raised because of this naming restriction of `nodepool` in `AKS`:  [Limitations](https://learn.microsoft.com/en-us/azure/aks/create-node-pools#limitations)

**Environment:**

- cluster-api-provider-azure version: `v1.10.3`
- Kubernetes version: (use `kubectl version`): `v1.27.3`

tapojit047:
**What type of PR is this?**
/kind bug

1. According to [Azure Docs](https://learn.microsoft.com/en-us/azure/aks/create-node-pools#limitations), these are the nodepool naming restrictions:
    - The name of a node pool may only contain lowercase alphanumeric characters and must begin with a lowercase letter.
    - For Linux node pools, the length must be between 1-12 characters.

But the nodepool name is not validated properly (according to these restrictions). Proper validation following these restrictions in the [`azuremanagedmachinepool_webhook.go`](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/api/v1beta1/azuremanagedmachinepool_webhook.go#L365) is added.

2. If the name of the nodepool i.e. `spec.name` field is `not specified` in the YAML of `AzureManagedMachinePool`, then the default name is set to the `metadata.name`. Take a look [here](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/api/v1beta1/azuremanagedmachinepool_webhook.go#L74-L75C3). So `metadata.name` is validated when `spec.name` is not set. 

Fixes #3968 

- [ ] cherry-pick candidate

**TODOs**:

- [X] squashed commits
- [X] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Add proper validation for nodepool name
```

### DIFF
--- a/api/v1beta1/azuremanagedmachinepool_webhook.go
+++ b/api/v1beta1/azuremanagedmachinepool_webhook.go
@@ -23,6 +23,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"unicode"
 
 	"github.com/pkg/errors"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -363,14 +364,59 @@ func (m *AzureManagedMachinePool) validateOSType() error {
 }
 
 func (m *AzureManagedMachinePool) validateName() error {
-	if m.Spec.OSType != nil && *m.Spec.OSType == WindowsOS &&
-		m.Spec.Name != nil && len(*m.Spec.Name) > 6 {
-		return field.Invalid(
-			field.NewPath("Spec", "Name"),
-			m.Spec.Name,
-			"Windows agent pool name can not be longer than 6 characters.")
+	var name *string
+	var fieldNameMessage string
+	if m.Spec.Name == nil || *m.Spec.Name == "" {
+		name = &m.Name
+		fieldNameMessage = "when spec.name is empty, metadata.name"
+	} else {
+		name = m.Spec.Name
+		fieldNameMessage = "spec.name"
 	}
 
+	if err := validateNameLength(m.Spec.OSType, name, fieldNameMessage); err != nil {
+		return err
+	}
+	return validateNamePattern(name, fieldNameMessage)
+}
+
+func validateNameLength(osType *string, name *string, fieldNameMessage string) error {
+	if osType != nil && *osType == WindowsOS &&
+		name != nil && len(*name) > 6 {
+		return field.Invalid(
+			field.NewPath("Spec", "Name"),
+			name,
+			fmt.Sprintf("For OSType Windows, %s can not be longer than 6 characters.", fieldNameMessage))
+	} else if (osType == nil || *osType == LinuxOS) &&
+		(name != nil && len(*name) > 12) {
+		return field.Invalid(
+			field.NewPath("Spec", "Name"),
+			osType,
+			fmt.Sprintf("For OSType Linux, %s can not be longer than 12 characters.", fieldNameMessage))
+	}
+	return nil
+}
+
+func validateNamePattern(name *string, fieldNameMessage string) error {
+	if name == nil || *name == "" {
+		return nil
+	}
+
+	if !unicode.IsLower(rune((*name)[0])) {
+		return field.Invalid(
+			field.NewPath("Spec", "Name"),
+			name,
+			fmt.Sprintf("%s must begin with a lowercase letter.", fieldNameMessage))
+	}
+
+	for _, char := range *name {
+		if !(unicode.IsLower(char) || unicode.IsNumber(char)) {
+			return field.Invalid(
+				field.NewPath("Spec", "Name"),
+				name,
+				fmt.Sprintf("%s may only contain lowercase alphanumeric characters.", fieldNameMessage))
+		}
+	}
 	return nil
 }
 
--- a/api/v1beta1/azuremanagedmachinepool_webhook_test.go
+++ b/api/v1beta1/azuremanagedmachinepool_webhook_test.go
@@ -39,7 +39,7 @@ func TestAzureManagedMachinePoolDefaultingWebhook(t *testing.T) {
 	t.Logf("Testing ammp defaulting webhook with mode system")
 	ammp := &AzureManagedMachinePool{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: "fooName",
+			Name: "fooname",
 		},
 		Spec: AzureManagedMachinePoolSpec{
 			Mode:         "System",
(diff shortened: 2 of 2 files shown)


api/v1beta1/azuremanagedmachinepool_webhook.go
```diff
@@ -72,16 +74,45 @@ func (mw *azureManagedMachinePoolWebhook) Default(ctx context.Context, obj runti
...
-
 	if m.Spec.OSType == nil {
 		m.Spec.OSType = ptr.To(DefaultOSType)
 	}
 
 	return nil
 }
 
+// generates a valid unqiue nodepool name following limitations of Azure
+func generateAKSNodePoolName(name string) string {
```
tamalsaha, line 88:
```
func generateAKSNodePoolName(name string) string {
	name = strings.ToLower(name)
	name = strings.TrimLeftFunc(name, unicode.IsNumber)
	if len(name) > 6 {
		name = name[:6]
	}
	return fmt.Sprintf("%s%s", name, utilrand.String(6))
}
```

api/v1beta1/azuremanagedmachinepool_webhook.go
```diff
@@ -72,16 +74,34 @@ func (mw *azureManagedMachinePoolWebhook) Default(ctx context.Context, obj runti
 	m.Labels[LabelAgentPoolMode] = m.Spec.Mode
 
 	if m.Spec.Name == nil || *m.Spec.Name == "" {
-		m.Spec.Name = &m.Name
+		aksNodePoolName := generateAKSNodePoolName(m.Name)
+		m.Spec.Name = &aksNodePoolName
```
jackfrancis, line 78: We can simplify the above two lines to `m.Spec.Name = ptr.To(generateAKSNodePoolName(m.Name))`

tapojit047: Resolved @jackfrancis

api/v1beta1/azuremanagedmachinepool_webhook.go
```diff
@@ -72,16 +74,33 @@ func (mw *azureManagedMachinePoolWebhook) Default(ctx context.Context, obj runti
...
+func generateAKSNodePoolName(name string) string {
+	name = strings.ToLower(name)
+	name = strings.Map(func(r rune) rune {
+		if unicode.IsNumber(r) || unicode.IsLetter(r) {
+			return r
+		}
+		return -1
+	}, name)
+	name = strings.TrimLeftFunc(name, unicode.IsNumber)
+	if len(name) > 6 {
```
jackfrancis, line 96: Why do we want to truncate the name from `MachinePool` to 6 chars and then add 6 chars of random characters?

Wouldn't this be sufficient?:

- remove/replace invalid characters
- truncate at the maximum allowed character limit (12)

tamalsaha: We would like to avoid naming conflict created from different CAPZ cluster. So, we want to add a random part to the pool name that is passed to Azure.

jackfrancis: Thanks for the clarification.

So is the scenario that we might run multiple clusters w/ a common MachinePool naming convention. E.g.:

- cluster1
  - pool1
  - pool2
  - pool3
- cluster2
  - pool1
  - pool2
  - pool3

And you want to have CAPZ ensure that cluster1 and cluster2 don't produce identically named pool names?

tamalsaha: Yes. We want to avoid naming conflict.

Also, adding a random part to will not affect ux, since user don't really have to interact with this directly  unless they are doing something from the Azure portal.

jackfrancis: Thanks again.

So, my initial response to this is that is that this type of enforcement should not be the responsibility of CAPZ. There could be other environments with the exact opposite requirements (require identical pool names across common cluster ecosystems).

Could you enforce this pool naming requirement in your cluster creation CI so that you're fully in control of these requirements?

cc @nojnhuh @CecileRobertMichon

jackfrancis: In fact helm chars do support injecting random characters.

Ref: https://helm.sh/docs/chart_template_guide/function_list/#randalphanum-randalpha-randnumeric-and-randascii

tamalsaha: The problem with that is that `helm install` and `helm upgrade` does not keep the random parts same. So, it is not usable in this case.

CecileRobertMichon: How about validating the AzureManagedMachinePool metadata name according to the Azure pool name rules _if and only if_ the spec.Name field is not explicitly set? And if the spec.Name field is set, we should validate the spec.Name.

I agree with @jackfrancis that generating random default names should not be in scope for CAPZ. There is no Azure requirement that a pool name must be unique (it should only be unique within a cluster).

nojnhuh: > How about validating the AzureManagedMachinePool metadata name according to the Azure pool name rules _if and only if_ the spec.Name field is not explicitly set? And if the spec.Name field is set, we should validate the spec.Name.

Since we already default `spec.name` to `metadata.name` when not set, do we only need to validate `spec.name`?

CecileRobertMichon: > do we only need to validate spec.name

That would be equivalent, only reason I can think of for doing both is that it might be more user friendly to validate the thing the user is actually setting (otherwise I could see users getting confused with "why am I getting a validation error for something I didn't set"). That's an implementation detail though.

tamalsaha: > There is no Azure requirement that a pool name must be unique (it should only be unique within a cluster).

I did not realise that. Thanks for the info.

jackfrancis: Do we want to convert this into a new PR that validates the AKS 12 character limit, or simply close this PR?

tapojit047: Converted it into a validation for the nodepool name PR. @jackfrancis

CecileRobertMichon (commented):
/lgtm
/assign @jackfrancis @nojnhuh

nojnhuh (commented):
/lgtm
/approve

Issue #3970, opened 2023-09-07
asyncpoller fails when SDK `PollUntilDone` returns a Done poller


nojnhuh:
/kind bug

**What steps did you take and what happened:**

This AKS test run failed to delete an AzureManagedMachinePool: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/3932/pull-cluster-api-provider-azure-e2e-aks/1699797265842966528

The following appears in the logs several times:
```
E0907 16:04:55.072224       1 controller.go:324]  "msg"="Reconciler error" "error"="error deleting AzureManagedMachinePool capz-e2e-8c0xmy/poolspot: failed to delete machine pool poolspot: failed to convert poller to future: failed to get resume token: poller is in a terminal state" "AzureManagedMachinePool"={"name":"poolspot","namespace":"capz-e2e-8c0xmy"} "controller"="azuremanagedmachinepool" "controllerGroup"="infrastructure.cluster.x-k8s.io" "controllerKind"="AzureManagedMachinePool" "name"="poolspot" "namespace"="capz-e2e-8c0xmy" "reconcileID"="bde3f505-a842-41f2-b11a-9bd4058eddcc"
```

**What did you expect to happen:**

**Anything else you would like to add:**

It looks like the call to `DeleteAsync` in asyncpoller.go is returning a non-nil poller where `poller.Done() == true`. There are a few things I noticed that might help in coming up with a fix, but it's not obvious to me the best route to take:
- The `err` returned by the call to `DeleteAsync` in asyncpoller.go is not checked when `poller != nil`.
- The `poller` returned in the same call is not checked if it's `.Done()` before ultimately it's `.ResumeToken()` method is called, which fails when the poller is done. This directly correlates to the above error message form the CAPZ logs.
- If the poller is Done, might we find an error in its `.Result()`?

The same issue likely affects the CreateOrUpdate flow as well.

**Environment:**

- cluster-api-provider-azure version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`): 

fyi @mboersma

PR #3988, opened 2023-09-11
add spec.networkPluginMode to AzureManagedControlPlane

### LINKED ISSUE #3678: Add support for Azure CNI Overlay on AKS managed clusters
/kind feature

Add support for [Azure CNI Overlay](https://learn.microsoft.com/en-us/azure/aks/azure-cni-overlay) on AKS managed clusters.

Microsoft.ContainerService/managedClusters API implements this with:

[`networkProfile.networkPluginMode: overlay`](https://learn.microsoft.com/en-us/rest/api/aks/managed-clusters/get?tabs=HTTP#networkpluginmode)
[`networkProfile.podCidr: <cidr>`](https://learn.microsoft.com/en-us/rest/api/aks/managed-clusters/get?tabs=HTTP#containerservicenetworkprofile)

nojnhuh:
**What type of PR is this?**
/kind feature
/area managedclusters

**What this PR does / why we need it**: This PR exposes the `networkProfile.networkPluginMode` AKS API field from CAPZ at AzureManagedControlPlane's `spec.networkPluginMode`.

Fixes #3678

- [ ] cherry-pick candidate

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Release note**:

```release-note
Added `spec.networkPluginMode` to AzureManagedControlPlane
```

### DIFF
--- a/api/v1beta1/azuremanagedcontrolplane_default.go
+++ b/api/v1beta1/azuremanagedcontrolplane_default.go
@@ -30,6 +30,10 @@ const (
 	defaultAKSVnetCIDR = "10.0.0.0/8"
 	// defaultAKSNodeSubnetCIDR is the default Node Subnet CIDR.
 	defaultAKSNodeSubnetCIDR = "10.240.0.0/16"
+	// defaultAKSVnetCIDRForOverlay is the default Vnet CIDR when Azure CNI overlay is enabled.
+	defaultAKSVnetCIDRForOverlay = "10.224.0.0/12"
+	// defaultAKSNodeSubnetCIDRForOverlay is the default Node Subnet CIDR when Azure CNI overlay is enabled.
+	defaultAKSNodeSubnetCIDRForOverlay = "10.224.0.0/16"
 )
 
 // setDefaultSSHPublicKey sets the default SSHPublicKey for an AzureManagedControlPlane.
@@ -60,6 +64,9 @@ func (m *AzureManagedControlPlane) setDefaultVirtualNetwork() {
 	}
 	if m.Spec.VirtualNetwork.CIDRBlock == "" {
 		m.Spec.VirtualNetwork.CIDRBlock = defaultAKSVnetCIDR
+		if ptr.Deref(m.Spec.NetworkPluginMode, "") == NetworkPluginModeOverlay {
+			m.Spec.VirtualNetwork.CIDRBlock = defaultAKSVnetCIDRForOverlay
+		}
 	}
 	if m.Spec.VirtualNetwork.ResourceGroup == "" {
 		m.Spec.VirtualNetwork.ResourceGroup = m.Spec.ResourceGroupName
@@ -73,6 +80,9 @@ func (m *AzureManagedControlPlane) setDefaultSubnet() {
 	}
 	if m.Spec.VirtualNetwork.Subnet.CIDRBlock == "" {
 		m.Spec.VirtualNetwork.Subnet.CIDRBlock = defaultAKSNodeSubnetCIDR
+		if ptr.Deref(m.Spec.NetworkPluginMode, "") == NetworkPluginModeOverlay {
+			m.Spec.VirtualNetwork.Subnet.CIDRBlock = defaultAKSNodeSubnetCIDRForOverlay
+		}
 	}
 }
 
--- a/api/v1beta1/azuremanagedcontrolplane_types.go
+++ b/api/v1beta1/azuremanagedcontrolplane_types.go
@@ -62,6 +62,18 @@ const (
 	ManagedControlPlaneIdentityTypeUserAssigned ManagedControlPlaneIdentityType = ManagedControlPlaneIdentityType(VMIdentityUserAssigned)
 )
 
+// NetworkPluginMode is the mode the network plugin should use.
+type NetworkPluginMode string
+
+const (
+	// NetworkPluginModeOverlay is used with networkPlugin=azure, pods are given IPs from the PodCIDR address space but use Azure
+	// Routing Domains rather than Kubenet's method of route tables.
+	// See also [AKS doc].
+	//
+	// [AKS doc]: https://aka.ms/aks/azure-cni-overlay
+	NetworkPluginModeOverlay NetworkPluginMode = "overlay"
+)
+
 // AzureManagedControlPlaneSpec defines the desired state of AzureManagedControlPlane.
 type AzureManagedControlPlaneSpec struct {
 	// Version defines the desired Kubernetes version.
@@ -109,6 +121,12 @@ type AzureManagedControlPlaneSpec struct {
 	// +kubebuilder:validation:Enum=azure;kubenet
 	// +optional
 	NetworkPlugin *string `json:"networkPlugin,omitempty"`
+
+	// NetworkPluginMode is the mode the network plugin should use.
+	// Allowed value is "overlay".
+	// +kubebuilder:validation:Enum=overlay
+	// +optional
+	NetworkPluginMode *NetworkPluginMode `json:"networkPluginMode,omitempty"`
 
 	// NetworkPolicy used for building Kubernetes network.
 	// Allowed values are "azure", "calico".
--- a/api/v1beta1/azuremanagedcontrolplane_webhook.go
+++ b/api/v1beta1/azuremanagedcontrolplane_webhook.go
@@ -258,6 +258,10 @@ func (mw *azureManagedControlPlaneWebhook) ValidateUpdate(ctx context.Context, o
 		allErrs = append(allErrs, errs...)
 	}
 
+	if errs := m.validateNetworkPluginModeUpdate(old); len(errs) > 0 {
+		allErrs = append(allErrs, errs...)
+	}
+
 	if errs := m.validateOIDCIssuerProfileUpdate(old); len(errs) > 0 {
 		allErrs = append(allErrs, errs...)
 	}
@@ -285,6 +289,7 @@ func (m *AzureManagedControlPlane) Validate(cli client.Client) error {
 		m.validateManagedClusterNetwork,
 		m.validateAutoScalerProfile,
 		m.validateIdentity,
+		m.validateNetworkPluginMode,
 	}
 
 	var errs []error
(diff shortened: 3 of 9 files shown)


api/v1beta1/azuremanagedcontrolplane_types.go
```diff
@@ -110,6 +122,12 @@ type AzureManagedControlPlaneSpec struct {
 	// +optional
 	NetworkPlugin *string `json:"networkPlugin,omitempty"`
 
+	// NetworkPluginMode is the mode the network plugin should use.
+	// Allowed value is "overlay".
```
Jont828, line 126: I assume this means that there's only one possible value? In that case, what happens with the `// +optional` and `omitempty` tags?

nojnhuh: These allowed values are only enforced when the value is non-nil, so this is the behavior:
```
networkPluginMode: "overlay" # ok

networkPluginMode: null # ok

networkPluginMode: "" # error
```

azure/services/managedclusters/spec_test.go
```diff
@@ -379,6 +380,48 @@ func TestParameters(t *testing.T) {
 				g.Expect(result).To(BeNil())
 			},
 		},
+		{
+			name: "setting networkPluginMode from nil to \"overlay\" will update",
```
jackfrancis, line 384: Is this a valid use case? (i.e., does AKS allow us to change this value to an existing cluster?)

nojnhuh: Yes: https://learn.microsoft.com/en-us/azure/aks/azure-cni-overlay#upgrade-an-existing-cluster-to-cni-overlay

jackfrancis: Are there any CAPZ apiserver webhooks we can add to protect these potential failures?

- Doesn't use the dynamic pod IP allocation feature.
- Doesn't have network policies enabled.
- Doesn't use any Windows node pools with docker as the container runtime.

(i.e., does CAPZ support the above features that don't work w/ Azure CNI overlay)

nojnhuh: Where would we check for each of those things? If they're defined in CAPZ objects then we could check those. I'm not sure if we have plumbing set up in the webhooks to do Azure API calls though.

jackfrancis: Right, we would only do this effectively if there was a CAPZ representation that corrrelated to the violating AKS features. If we haven't yet implemented support for the above then we would probably not want to mess around with pre-validating against the AKS API mid-flight (we should defer to AKS on that and improve the UX by actually implementing CAPZ wrapping of those features, IMO).

We do seem to support NetworkPolicy in CAPZ, so maybe we could add a webhook to enforce that we don't ship that config w/ CNI overlay.

nojnhuh: I can add a check for network policy. I don't think we can (easily) add anything for those other items though.

api/v1beta1/azuremanagedcontrolplane_webhook.go
```diff
@@ -538,6 +543,17 @@ func (m *AzureManagedControlPlane) validateVirtualNetworkUpdate(old *AzureManage
 	return allErrs
 }
 
+// validateNetworkPluginModeUpdate validates update to NetworkPluginMode.
+func (m *AzureManagedControlPlane) validateNetworkPluginModeUpdate(old *AzureManagedControlPlane) field.ErrorList {
+	var allErrs field.ErrorList
+
+	if ptr.Deref(m.Spec.NetworkPluginMode, "") == NetworkPluginModeOverlay && old.Spec.NetworkPolicy != nil {
```
nojnhuh, line 550: I verified that AKS does allow this on create. It only errors on updates.

api/v1beta1/azuremanagedcontrolplane_types.go
Jont828: Gotcha, are there any other cases where we do validation like this instead of with the webhook? I'm fine to proceed as is, but figured it's worth noting that it might be confusing to have validation in more than one place.

Jont828:
LGTM pending squash

api/v1beta1/azuremanagedcontrolplane_types.go
nojnhuh: Not sure what you mean? The kubebuilder annotation here is the only thing defining this specific validation.

nojnhuh:
Just pushed a change to improve CIDR defaults in the webhook when overlay is enabled so they match AKS's defaults. Thank you @felfa01 for the feedback in https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/3678#issuecomment-1715680082!

api/v1beta1/azuremanagedcontrolplane_types.go
jackfrancis: Yeah, I think the general principle is

1. Prefer yaml schema validation (this is limited, but if your validation can be express in terms of kubebuilder pragmas then this is better)
2. Only do one or the other (both is worse)
3. When validation is trickier, use webhooks

api/v1beta1/azuremanagedcontrolplane_webhook.go
jackfrancis: Wut

We're saying:

1. You can create a cluster w/ CNI Overlay + NetworkPolicy when you create your cluster
2. You cannot update a cluster w/ NetworkPolicy configured from non-CNI Overlay to CNI Overlay
3. You can update a cluster to CNI Overlay if NetworkPolicy is not enabled

?

nojnhuh: Correct

nojnhuh:
AKS test failing due to #3970

DavidSpek:
Should this be part of the [v1.11](https://github.com/kubernetes-sigs/cluster-api-provider-azure/milestone/33) milestone?

nojnhuh:
@DavidSpek Yes, the plan is to get this in before the release.

/milestone v1.11

api/v1beta1/azuremanagedcontrolplane_webhook.go
jackfrancis: I wonder if scenario #1 actually works? Probably out-of-scope for this PR to definitely prove that, so we can take the API at its word and implement CAPZ accordingly.

Thanks for doing the extra validation here.

nojnhuh:
@jackfrancis @Jont828 Any other thoughts here? Looking to include this in the v1.11 release today if possible.

api/v1beta1/azuremanagedcontrolplane_webhook.go
```diff
@@ -713,3 +729,20 @@ func (m *AzureManagedControlPlane) validateIdentity(_ client.Client) error {
 
 	return nil
 }
+
+// validateNetworkPluginMode validates an Identity.
```
mboersma, line 733:
```suggestion
// validateNetworkPluginMode validates a NetworkPluginMode.
```

mboersma (commented):
lgtm, all I could find was a code comment discrepency.

api/v1beta1/azuremanagedcontrolplane_webhook.go
nojnhuh: caught red-handed lol. Fixed!

jackfrancis: @mboersma you win this round

api/v1beta1/azuremanagedcontrolplane_default.go
```diff
@@ -73,6 +80,9 @@ func (m *AzureManagedControlPlane) setDefaultSubnet() {
 	}
 	if m.Spec.VirtualNetwork.Subnet.CIDRBlock == "" {
 		m.Spec.VirtualNetwork.Subnet.CIDRBlock = defaultAKSNodeSubnetCIDR
+		if ptr.Deref(m.Spec.NetworkPluginMode, "") == NetworkPluginModeOverlay {
+			m.Spec.VirtualNetwork.Subnet.CIDRBlock = defaultAKSNodeSubnetCIDRForOverlay
```
jackfrancis, line 84: One final validation possibility here... so I see we're assigning an overlay-specific CIDR value for VNET and subnet if the user doesn't declare one. Is there any validation we can add here for the overlay CIDRs? Does AKS require that we use particular nets here?

We're already doing some basic hygiene in `validateSubnetCIDR` and `validateVnetCIDR` (`api/v1beta1/azurecluster_validation.go`) so we get that for free.

nojnhuh: Are you suggesting to call the same `validateVnetCIDR` and `validateSubnetCIDR` on these values?

> Does AKS require that we use particular nets here?

No AFAICT, these overlay-specific defaults match AKS's defaults when overlay is enabled per https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/3678#issuecomment-1715680082.

jackfrancis: I'm saying we're already calling those validate funcs, so that's good (as far as I can see we're reusing existing property CIDR fields w/ this implementation)

My question is what happens when a user *explicitly* passes in `"10.0.0.0/8"` (for example) as the VNET CIDR _and_ enables overlay CNI? Is AKS opinionated: "Hey, that CIDR block is reserved for non-overlay configurations!".

If we don't know all the answers to the above, I think it's fair to keep that out of scope, just curious.

nojnhuh: > My question is what happens when a user _explicitly_ passes in `"10.0.0.0/8"` (for example) as the VNET CIDR _and_ enables overlay CNI? Is AKS opinionated: "Hey, that CIDR block is reserved for non-overlay configurations!".

The comment I linked above shows the error with the non-overlay defaults, and it wasn't that specific:
`NetworkProfile CIDR 10.244.0.0/16 overlaps with reserved VNET CIDR 10.0.0.0/8`

I don't think a user explicitly passing in the value would affect that since it results in the same config sent to AKS either way.

jackfrancis: I think we can add more explicit CAPZ validations that mirror AKS validations over time if it has user value (in most cases failing the declarative spec is more user-friendly than looking for an API error in the CAPZ controller logs). Shouldn't block this change.

mboersma (approved):
/lgtm
/approve

nojnhuh:
Still need to resolve impending conflicts with #3973 after that merges, so I'll need one more lgtm after that @mboersma @jackfrancis. Then I'll sqaush and push such that the lgtm sticks.

nojnhuh:
@mboersma @CecileRobertMichon This should be ready for a final lgtm.

mboersma (commented):
/lgtm

nojnhuh:
squashed!
/hold cancel

PR #3997, opened 2023-09-13
use generics for ASO services

nojnhuh:
**What type of PR is this?**
/kind cleanup

This PR incorporates Go generics in the ASO services to get more compiler-enforced type checking and smooth out some code maintenance-related bumps. I mostly started this as an experiment and I think it came out well enough to see if others think this is an improvement. I'll take a review pass at this to add inline comments, but overall my biggest takeaways are:
- Type assertions at runtime never felt good, but now a lot of those are gone because the compiler checks for us!
- Each service implementation can deal exclusively with its native type directly without interfaces. That feels a lot more ergonomic.
- Type parameters are \*everywhere\*.
- Generics definitely do not make working with Gomega in tests any easier in tests since it already uses `any` so extensively.

Fixes #

Right after I open this I will take a review pass to add more comments inline.

- [ ] cherry-pick candidate

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Release note**:

```release-note
NONE
```


azure/interfaces.go
```diff
@@ -138,14 +138,14 @@ type ResourceSpecGetterWithHeaders interface {
 }
 
 // ASOResourceSpecGetter is an interface for getting all the required information to create/update/delete an Azure resource.
-type ASOResourceSpecGetter interface {
+type ASOResourceSpecGetter[T genruntime.MetaObject] interface {
```
nojnhuh, line 141: `T` here ultimately maps to an ASO root object type represented in Kubernetes like `ResourceGroup` or `VirtualMachine`.

The way I reason about this is that each `ASOResourceSpecGetter` is parameterized over a particular type, say `ResourceGroup`. That means a `ASOResourceSpecGetter[*ResourceGroup]` requires a method with the signature
```
ResourceRef() *ResourceGroup
```
Whereas before an implementation could return any type that implements the `genruntime.MetaObject` interface, where there's a risk that a non-ResourceGroup could be returned and lead to unexpected errors.

azure/services/aso/aso.go
```diff
@@ -47,25 +47,31 @@ const (
...
+// deepCopier is a genruntime.MetaObject with a typed DeepCopy method, usually generated by kubebuilder.
+type deepCopier[T any] interface {
+	genruntime.MetaObject
+	DeepCopy() T
+}
+
 // Service is an implementation of the Reconciler interface. It handles creation
 // and deletion of resources using ASO.
-type Service struct {
+type Service[T deepCopier[T]] struct {
```
nojnhuh, line 58: This looks funky and I admit I can't think of a good way to explain how this works.

azure/services/aso/aso.go
```diff
@@ -124,33 +133,25 @@ func (s *Service) CreateOrUpdateResource(ctx context.Context, spec azure.ASOReso
...
 	}
 
 	// Construct parameters using the resource spec and information from the existing resource, if there is one.
-	parameters, err := spec.Parameters(ctx, deepCopyOrNil(existing))
+	parameters, err := spec.Parameters(ctx, existing.DeepCopy())
 	if err != nil {
-		return nil, errors.Wrapf(err, "failed to get desired parameters for resource %s/%s (service: %s)", resourceNamespace, resourceName, serviceName)
-	}
-	// a nil result here is a special case for compatibility with the old
-	// SDK-driven service implementations.
```
nojnhuh, line 140: Because of the inability to compare `T` to `nil`, I had to walk back this workaround. Now implementations of `Parameters` must return the `existing` input object unmodified if no updates are required instead of returning `nil`.

azure/services/aso/interfaces.go
```diff
@@ -25,15 +25,15 @@ import (
...
 // TagsGetterSetter represents an object that supports tags.
-type TagsGetterSetter interface {
+type TagsGetterSetter[T genruntime.MetaObject] interface {
 	GetAdditionalTags() infrav1.Tags
-	GetDesiredTags(resource genruntime.MetaObject) (infrav1.Tags, error)
-	GetActualTags(resource genruntime.MetaObject) (infrav1.Tags, error)
-	SetTags(resource genruntime.MetaObject, tags infrav1.Tags) error
+	GetDesiredTags(resource T) infrav1.Tags
+	GetActualTags(resource T) infrav1.Tags
+	SetTags(resource T, tags infrav1.Tags)
```
nojnhuh, line 38: This is one of the biggest wins for generics here. Since the only way these methods could have errored before was due to unexpected types, those would now be compiler errors so we don't have to do any error handling at runtime to catch those same errors.

azure/services/aso/tags.go
```diff
@@ -32,40 +33,32 @@ import (
...
 			if err != nil {
 				return errors.Wrapf(err, "failed to unmarshal JSON from %s annotation", tagsLastAppliedAnnotation)
 			}
 		}
-	}
 
-	existingTags, err := t.GetActualTags(existing)
-	if err != nil {
-		return errors.Wrapf(err, "failed to get actual tags for %s %s/%s", existing.GetObjectKind().GroupVersionKind(), existing.GetNamespace(), existing.GetName())
+		existingTags = t.GetActualTags(existing)
```
nojnhuh, line 49: There is a little bit of light refactoring in this function not strictly-related to generics to workaround each `GetActualTags` implementation from needing to handle `nil` inputs.

azure/services/aso/aso_test.go
```diff
@@ -332,18 +331,17 @@ func TestCreateOrUpdateResource(t *testing.T) {
...
 				Name:      "name",
 				Namespace: "namespace",
 			},
 		})
-		specMock.EXPECT().Parameters(gomockinternal.AContext(), gomock.Not(gomock.Nil())).DoAndReturn(func(_ context.Context, object genruntime.MetaObject) (genruntime.MetaObject, error) {
-			group := object.DeepCopyObject().(*asoresourcesv1.ResourceGroup)
+		specMock.EXPECT().Parameters(gomockinternal.AContext(), gomock.Not(gomock.Nil())).DoAndReturn(func(_ context.Context, group *asoresourcesv1.ResourceGroup) (*asoresourcesv1.ResourceGroup, error) {
 			group.Spec.Location = ptr.To("location")
 			return group, nil
 		})
```
nojnhuh, line 347: This is representative of what an actual `Parameters` method implementation might look like for a service modifying an existing object.

(I think the DeepCopyObject that used to exist was redundant even before these changes.)

azure/services/aso/aso_test.go
```diff
@@ -629,7 +627,7 @@ func TestCreateOrUpdateResource(t *testing.T) {
 
 		g.Expect(result.GetName()).To(Equal("name"))
 		g.Expect(result.GetNamespace()).To(Equal("namespace"))
-		g.Expect(result.(*asoresourcesv1.ResourceGroup).Spec.Location).To(Equal(ptr.To("location")))
```
nojnhuh, line 632: ⚡ `result` is already a `*ResourceGroup`

azure/services/aso/aso.go
```diff
@@ -53,25 +53,25 @@ type deepCopier[T any] interface {
 	DeepCopy() T
 }
 
-// Service is an implementation of the Reconciler interface. It handles creation
+// CreatorDeleter is an implementation of the Reconciler interface. It handles creation
 // and deletion of resources using ASO.
-type Service[T deepCopier[T]] struct {
+type CreatorDeleter[T deepCopier[T]] struct {
```
nojnhuh, line 58: I don't love this name, but wanted to change it to free up `Service` for the new abstraction.

azure/services/aso/service.go
```diff
@@ -0,0 +1,123 @@
...
+	"context"
+
+	"github.com/pkg/errors"
+	"sigs.k8s.io/cluster-api-provider-azure/azure"
+	"sigs.k8s.io/cluster-api-provider-azure/util/reconciler"
+	"sigs.k8s.io/cluster-api-provider-azure/util/tele"
+)
+
+// Service provides operations on Azure resources.
+type Service[T deepCopier[T], S Scope] struct {
```
nojnhuh, line 28: `T` here is the same as it is elsewhere in this package. `S` represents a type of Scope defined by each service, like `GroupScope`.

azure/services/groups/groups.go
```diff
@@ -35,81 +33,42 @@ const ServiceName = "group"
...
-	GetClient() client.Client
-	ClusterName() string
 }
 
 // New creates a new service.
 func New(scope GroupScope) *Service {
+	svc := aso.NewService[ASOType](ServiceName, scope)
+	svc.Specs = []azure.ASOResourceSpecGetter[ASOType]{scope.GroupSpec()}
+	svc.PostReconcileHook = postReconcileHook
+	svc.PostDeleteHook = postDeleteHook
```
nojnhuh, line 51: Initializing a service is kind of clunky at the moment combining a constructor with setting individual fields.

azure/services/groups/groups.go
```diff
@@ -35,81 +33,42 @@ const ServiceName = "group"
...
-
-	groupSpec := s.Scope.GroupSpec()
-	if groupSpec == nil {
-		return nil
-	}
-
-	_, err := s.CreateOrUpdateResource(ctx, groupSpec, ServiceName)
-	s.Scope.UpdatePutStatus(infrav1.ResourceGroupReadyCondition, ServiceName, err)
+func postReconcileHook(scope GroupScope, err error) error {
+	scope.UpdatePutStatus(infrav1.ResourceGroupReadyCondition, ServiceName, err)
```
nojnhuh, line 59: This function is so trivial that it probably doesn't need a test, nonetheless I added one anyway to show how that might be done.

nojnhuh (commented):
These most recent changes add a generic service which implements the `Reconcile` and `Delete` methods generically, with hooks to inject logic for things like updating status.

This generic implementation should be usable by most existing services, but those that need more complex logic are still free to implement those methods themselves.

nojnhuh:
I'm reasonably confident this is overall a net-positive so I'm advocating that we merge some form of this PR. Marking ready for review.

azure/services/aso/service.go
```diff
@@ -0,0 +1,123 @@
...
+func (s *Service[T, S]) Reconcile(ctx context.Context) error {
+	ctx, _, done := tele.StartSpanWithLogger(ctx, "aso.Service.Reconcile")
+	defer done()
+
+	ctx, cancel := context.WithTimeout(ctx, reconciler.DefaultAzureServiceReconcileTimeout)
+	defer cancel()
+
+	// We go through the list of Specs to reconcile each one, independently of the result of the previous one.
+	// If multiple errors occur, we return the most pressing one.
+	//  Order of precedence (highest -> lowest) is: error that is not an operationNotDoneError (i.e. error creating) -> operationNotDoneError (i.e. creating in progress) -> no error (i.e. created)
```
mboersma, line 65: Maybe make this long comment a bulleted list?

azure/services/aso/service.go
```diff
@@ -0,0 +1,123 @@
...
+func (s *Service[T, S]) Delete(ctx context.Context) error {
+	ctx, _, done := tele.StartSpanWithLogger(ctx, "aso.Service.Delete")
+	defer done()
+
+	ctx, cancel := context.WithTimeout(ctx, reconciler.DefaultAzureServiceReconcileTimeout)
+	defer cancel()
+
+	// We go through the list of SubnetSpecs to delete each one, independently of the resultErr of the previous one.
+	// If multiple errors occur, we return the most pressing one.
+	//  Order of precedence (highest -> lowest) is: error that is not an operationNotDoneError (i.e. error deleting) -> operationNotDoneError (i.e. deleting in progress) -> no error (i.e. deleted)
```
mboersma, line 93: Same here: it would be easier to read in a bulleted list IMHO.

azure/services/aso/service.go
```diff
@@ -0,0 +1,123 @@
...
+
+	if s.PostDeleteHook != nil {
+		return s.PostDeleteHook(s.Scope, resultErr)
+	}
+	return resultErr
+}
+
+// Pause implements azure.Pauser.
+func (s *Service[T, S]) Pause(ctx context.Context) error {
+	var _ azure.Pauser = (*Service[T, S])(nil)
```
mboersma, line 110: I'm curious, why are we doing this type assertion thing here?

azure/services/aso/service_test.go
```diff
@@ -0,0 +1,386 @@
...
+			name:       serviceName,
+			PostDeleteHook: func(scopeParam *mock_aso.MockScope, err error) error {
+				g.Expect(scopeParam).To(BeIdenticalTo(scope))
+				g.Expect(err).To(MatchError("DeleteResource error"))
+				return errors.New("PostDeleteHook error")
+			},
+		}
+
+		err := s.Delete(context.Background())
+		g.Expect(err).To(MatchError("PostDeleteHook error"))
```
mboersma, line 320: These service unit tests are all well done and easy to follow. I like returning an error to validate that the hook funcs are called.

mboersma (commented):
This is great, it saves code in all the right places and sets services up for easier conversion to ASO. The new `Service` and interfaces make sense to me, although anything with generics takes a bit more brainpower to parse.

lgtm, just had a couple small comments & questions.

azure/services/aso/service.go
nojnhuh: +1, will do.

azure/services/aso/service.go
nojnhuh: Normally I'd put this outside the method, but the T and S type parameters aren't defined in that scope. I suppose we could use concrete types and move this outside the method. I can see where this has the marginal benefit of verifying the interface implementation for every set of types a Service is instantiated with, but I doubt that would catch anything new in this case.

Either way, I'm trusting that the compiler optimizes this out completely.

mboersma: If I'd seen this at the module level I wouldn't have blinked, but makes sense why the scope only exists here.

mboersma (commented):
/lgtm

Good stuff, let's see if we can get more eyes on this.

CecileRobertMichon (commented):
@nojnhuh thoughts about doing a PR walkthrough of this one at the next office hours? 

Doesn't have to block merge

nojnhuh:
> @nojnhuh thoughts about doing a PR walkthrough of this one at the next office hours?

Sure, I can do that!

azure/services/aso/aso.go
```diff
@@ -76,13 +82,16 @@ func (s *Service) CreateOrUpdateResource(ctx context.Context, spec azure.ASOReso
 	log = log.WithValues("service", serviceName, "resource", resourceName, "namespace", resourceNamespace)
 
 	var adopt bool
-	var existing genruntime.MetaObject
+	var existing T
+	var zero T // holds the zero value, to be returned with non-nil errors.
```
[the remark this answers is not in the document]
CecileRobertMichon: Are we using `result` anywhere? It seems like the aso implementation inherited the named `result` return parameter without actually using it. It was useful in the async framework when `result` was of type `interface` but now that we're able to pass in a type we might want to get rid of the named return or fix the function to actually use it.

Seems like there is a bit of redundancy in variables between zero, existing and result. Could we reuse existing and return it with a nil value for non-nil errors?  or use the result variable for both

azure/services/groups/groups.go
CecileRobertMichon: +1 on the concept of having a pre/post reconcile hook

That being said, for `UpdatePutStatus` specifically, we do it in all the services so it would be best to include that in the main Service reconcile so it's consistent for every service and we don't have to remember to add it to every single one.

azure/services/groups/groups.go
```diff
@@ -125,8 +85,7 @@ func (s *Service) ShouldDeleteIndividualResources(ctx context.Context) bool {
 	// For ASO, "managed" only tells us that we're allowed to delete the ASO
 	// resource. We also need to check that deleting the ASO resource will really
 	// delete the underlying resource group by checking the ASO reconcile-policy.
-	spec := s.Scope.GroupSpec()
-	group := spec.ResourceRef()
+	group := s.Specs[0].ResourceRef()
```
CecileRobertMichon, line 88: Even though we know there is a single group in the specs for groups, it might be best to check the length of s.Specs so we don't panic if the specs field is somehow empty

azure/services/groups/groups.go
```diff
@@ -34,81 +33,42 @@ const ServiceName = "group"
...
-var _ azure.Pauser = (*Service)(nil)
-
-// Pause implements azure.Pauser.
-func (s *Service) Pause(ctx context.Context) error {
-	groupSpec := s.Scope.GroupSpec()
-	if groupSpec == nil {
-		return nil
-	}
-	return aso.PauseResource(ctx, s.Scope.GetClient(), groupSpec, s.Scope.ClusterName(), ServiceName)
+	return aso.IsManaged(ctx, s.Scope.GetClient(), s.Specs[0], s.Scope.ClusterName())
```
CecileRobertMichon, line 71: same thing here

azure/services/groups/spec.go
```diff
@@ -18,16 +18,17 @@ package groups
...
 	asoresourcesv1 "github.com/Azure/azure-service-operator/v2/api/resources/v1api20200601"
-	"github.com/Azure/azure-service-operator/v2/pkg/genruntime"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/utils/ptr"
 	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
 	"sigs.k8s.io/cluster-api-provider-azure/azure/services/aso"
 )
 
+// ASOType aliases the ASO object type managed by the service.
+type ASOType = *asoresourcesv1.ResourceGroup
```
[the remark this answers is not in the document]
CecileRobertMichon: nice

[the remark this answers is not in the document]
CecileRobertMichon: although it looks like we're still using `asoresourcesv1.ResourceGroup` directly in some places, isn't it potentially more confusing if it's a mix of both?

azure/services/aso/aso.go
CecileRobertMichon: `ASOReconciler`? since it implements the `Reconciler` interface

azure/services/groups/groups.go
nojnhuh: +1, will update

azure/services/groups/groups.go
nojnhuh: Good callout. Instead though I think I'll refactor this to handle multiple specs a bit more natively so we don't need to index directly into this slice. This is something I was planning to do soon anyway to handle managing a vnet's resource group separately (if it's different).

azure/services/groups/spec.go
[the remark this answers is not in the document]
nojnhuh: I think the only place where it's necessary to use the regular name is in the `GroupScope` interface definition in groups.go. Using `ASOType` there results in an import cycle in the generated mocks.

I think I might slightly prefer the full name anyway since I think that's a bit more clear. I expect the compiler would catch any copy-paste mistakes.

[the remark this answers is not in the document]
CecileRobertMichon: I meant usages in this file specifically https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3997/files#diff-ba2c0ee6f840a091bd5c2362b53be5d53f384bcb347c7f11feff1c9c0ba9351bL57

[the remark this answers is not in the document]
nojnhuh: Do you prefer using the `ASOType` alias everywhere or the full `*asoresourcesv1.ResourceGroup` everywhere?

nojnhuh:
@nawazkh The test is hitting the same role assignments issue, so I wouldn't expect it to pass until we fix that:

`E0926 16:43:19.336660       1 controller.go:324]  "msg"="Reconciler error" "error"="error creating AzureManagedControlPlane capz-e2e-ojhbdg/capz-e2e-ojhbdg-aks: failed to reconcile AzureManagedControlPlane service managedcluster: failed to create or update resource capz-e2e-ojhbdg-aks/capz-e2e-ojhbdg-aks (service: managedcluster): GET https://management.azure.com/subscriptions/===REDACTED===/providers/Microsoft.ContainerService/locations/uksouth/operations/a4f59c54-3678-46c6-9817-52acc704a4aa\n--------------------------------------------------------------------------------\nRESPONSE 200: 200 OK\nERROR CODE: RoleAssignmentLimitExceeded\n--------------------------------------------------------------------------------\n{\n  \"name\": \"549cf5a4-7836-c646-9817-52acc704a4aa\",\n  \"status\": \"Failed\",\n  \"startTime\": \"2023-09-26T16:43:04.2041495Z\",\n  \"endTime\": \"2023-09-26T16:43:08.3739392Z\",\n  \"error\": {\n    \"code\": \"RoleAssignmentLimitExceeded\",\n    \"message\": \"Create role assignment failed. Scope: /subscriptions/===REDACTED===/resourceGroups/MC_capz-e2e-ojhbdg-aks_capz-e2e-ojhbdg-aks_uksouth; role assignment name: d667e4de-786e-47ae-a456-de1ac575f8dc\",\n    \"details\": [\n      {\n        \"code\": \"RoleAssignmentLimitExceeded\",\n        \"message\": \"No more role assignments can be created.\"\n      }\n    ]\n  }\n}\n--------------------------------------------------------------------------------\n" "AzureManagedControlPlane"={"name":"capz-e2e-ojhbdg-aks","namespace":"capz-e2e-ojhbdg"} "controller"="azuremanagedcontrolplane" "controllerGroup"="infrastructure.cluster.x-k8s.io" "controllerKind"="AzureManagedControlPlane" "name"="capz-e2e-ojhbdg-aks" "namespace"="capz-e2e-ojhbdg" "reconcileID"="6f9c1c5c-2820-4c03-a9b7-21bff5b707a2"`

nawazkh:
^Oops

azure/services/groups/groups.go
CecileRobertMichon: nice

azure/services/groups/spec.go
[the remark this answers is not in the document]
CecileRobertMichon: I don't have a strong preference, my only preference is consistency

CecileRobertMichon:
/lgtm

🤩

azure/interfaces.go
```diff
@@ -40,7 +40,6 @@ type Pauser interface {
 // ServiceReconciler is an Azure service reconciler which can reconcile an Azure service.
 type ServiceReconciler interface {
 	Name() string
-	IsManaged(ctx context.Context) (bool, error)
```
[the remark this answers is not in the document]
nojnhuh: This method is actually never invoked on a variable of type `ServiceReconciler`, only ever on structs local to each service, so it doesn't need to be a part of this interface. This PR changes the signature of the `IsManaged` implementation for the groups service such that the compiler was throwing an error when `IsManaged` was still defined this way on this interface.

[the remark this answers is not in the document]
jackfrancis: Someone left us a nice gift and you found it!

jackfrancis:
Nice work leveling up the value of the ASO work. :)

nojnhuh:
Just pushed one more change to swap the `reconciler` and `Reconciler` names. Now `Reconciler` defines the interface which can be embedded in service implementations outside the `aso` package and `reconciler` is a concrete implementation of `Reconciler` only needed within the `aso` package.

CecileRobertMichon (commented):
/lgtm

nojnhuh:
squashed!
/hold cancel

Issue #2328, opened 2022-05-25
Support autoscaling on self-managed MachinePool using MachinePool Machines


lzhecheng:
/kind feature

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]
Now autoscaling is supported only for managed clusters (aks). And regular clusters should be supported as well.
Issue for managed one: https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1669

I am add autoscaling CI for cloud-provider-azure with CAPZ.
The original job defined with aks-engine: https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-sigs/cloud-provider-azure/cloud-provider-azure-config.yaml#L422-L483
The current cluster template: https://raw.githubusercontent.com/kubernetes-sigs/cloud-provider-azure/master/tests/k8s-azure/manifest/linux-autoscaler.json

Config is:
```
"pools": [
    {
        "name": "agentpool1",
        "config": {
            "min-nodes": "1",
            "max-nodes": "1000"
        }
    }
],
"config": {
    "scale-down-delay-after-add": "5m",
    "scale-down-unneeded-time": "5m"
}
```

related: https://github.com/kubernetes-sigs/cloud-provider-azure/issues/919

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- cluster-api-provider-azure version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

CecileRobertMichon:
/remove-lifecycle rotten
/lifecycle active
/assign @Jont828 
/reopen

cc @dtzar

Jont828:
/retitle Support autoscaling on self-managed MachinePool using MachinePool Machines

Jont828:
This should be done since #3998 is merged, closing this issue out.

PR #3998, opened 2023-09-13
Add implementation for MachinePool Machines

Jont828:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**: #2328

Fixes #

- [ ] cherry-pick candidate

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Implements MachinePool Machines according to CAPI contract. 

Action required: The `NodeDrainTimeout` field is now removed from AzureMachinePools. The equivalent field can be set on the MachinePool spec.
```


Jont828:
@CecileRobertMichon @mboersma @devigned It's finally here!!!

azure/scope/machinepool.go
```diff
@@ -494,6 +502,27 @@ func (m *MachinePoolScope) createMachine(ctx context.Context, machine azure.VMSS
...
 
+func (m *MachinePoolScope) deleteMachine(ctx context.Context, ampm infrav1exp.AzureMachinePoolMachine) error {
+	ctx, _, done := tele.StartSpanWithLogger(ctx, "scope.MachinePoolScope.deleteMachine")
+	defer done()
+
+	machine, err := util.GetOwnerMachine(ctx, m.client, ampm.ObjectMeta)
+	if err != nil {
+		return errors.Wrapf(err, "error getting owner Machine for AzureMachinePoolMachine %s/%s", ampm.Namespace, ampm.Name)
+	}
+	if machine == nil {
```
Jont828, line 513: It's kind of up to us on how to handle this case. Do we want to try to delete the AMPM directly in the case that the MP is deleted, or just leave it to requeue?

exp/controllers/azuremachinepool_controller.go
```diff
@@ -390,6 +393,10 @@ func (ampr *AzureMachinePoolReconciler) reconcileDelete(ctx context.Context, mac
 		if err := amps.Delete(ctx); err != nil {
 			return reconcile.Result{}, errors.Wrapf(err, "error deleting AzureMachinePool %s/%s", clusterScope.Namespace(), machinePoolScope.Name())
 		}
+	} else {
```
Jont828, line 396: This part is a bit tricky since we need to keep the AzureMachinePool from being deleted until all its AMPMs are gone even when the resource group is being deleted. This is because otherwise we will be unable to construct the machinepool scope in the AMPM controller.

exp/controllers/azuremachinepoolmachine_controller.go
```diff
@@ -311,14 +327,10 @@ func (ampmr *AzureMachinePoolMachineController) reconcileDelete(ctx context.Cont
 	ctx, log, done := tele.StartSpanWithLogger(ctx, "controllers.AzureMachinePoolMachineController.reconcileDelete")
 	defer done()
 
+	// Note: If deletion is requested and either the cluster is being deleted or the AMP is missing, we short circuit earlier in the code before calling this function.
+	// This is because machineScope can't be constructed without the cluster or AMP.
 	log.Info("Handling deleted AzureMachinePoolMachine")
 
-	if machineScope.AzureMachinePool == nil || !machineScope.AzureMachinePool.ObjectMeta.DeletionTimestamp.IsZero() {
```
Jont828, line 316: This was the previous code to handle when the RG is being deleted. We can't progress with a nil AzureMachinePool in the current implementation so I've moved it to check the cluster scope instead.

Jont828 (commented):
@CecileRobertMichon @mboersma I saw the tests were green so I went ahead and squashed. Ready for review!

exp/controllers/azuremachinepoolmachine_controller.go
```diff
@@ -209,6 +187,44 @@ func (ampmr *AzureMachinePoolMachineController) Reconcile(ctx context.Context, r
 		return reconcile.Result{}, err
 	}
 
+	// Note: this is critical for deletion!
```
CecileRobertMichon, line 190: can you explain why?

test/e2e/azure_test.go
```diff
@@ -523,17 +523,6 @@ var _ = Describe("Workload cluster creation", func() {
 				})
 			})
 
-			By("Cordon and draining a node", func() {
```
CecileRobertMichon, line 526: ❤️

exp/controllers/azuremachinepoolmachine_controller.go
```diff
@@ -311,14 +327,10 @@ func (ampmr *AzureMachinePoolMachineController) reconcileDelete(ctx context.Cont
...
 	log.Info("Handling deleted AzureMachinePoolMachine")
 
-	if machineScope.AzureMachinePool == nil || !machineScope.AzureMachinePool.ObjectMeta.DeletionTimestamp.IsZero() {
-		// deleting the entire VMSS, so just remove finalizer and VMSS delete remove the underlying infrastructure.
-		controllerutil.RemoveFinalizer(machineScope.AzureMachinePoolMachine, infrav1exp.AzureMachinePoolMachineFinalizer)
-		return reconcile.Result{}, nil
-	}
-
 	// deleting a single machine
 	// 1) drain the node (TODO: @devigned)
```
CecileRobertMichon, line 335: can we remove this todo?

azure/scope/machinepool.go
```diff
@@ -454,6 +462,28 @@ func (m *MachinePoolScope) applyAzureMachinePoolMachines(ctx context.Context) er
 	return nil
 }
 
+// DeleteAllAzureMachinePoolMachines deletes all AzureMachinePoolMachines for the AzureMachinePool.
+func (m *MachinePoolScope) DeleteAllAzureMachinePoolMachines(ctx context.Context) error {
```
CecileRobertMichon, line 466: Are the AzureMachinePoolMachines owned by the Cluster or MachinePool? if so, could we leverage [k8s garbage collection](https://kubernetes.io/docs/tasks/administer-cluster/use-cascading-deletion/#set-orphan-deletion-policy) instead of deleting all the AMPM from here?

exp/api/v1beta1/azuremachinepool_types.go
```diff
@@ -157,12 +157,6 @@ type (
 		// +kubebuilder:default={type: "RollingUpdate", rollingUpdate: {maxSurge: 1, maxUnavailable: 0, deletePolicy: Oldest}}
 		Strategy AzureMachinePoolDeploymentStrategy `json:"strategy,omitempty"`
 
-		// NodeDrainTimeout is the total amount of time that the controller will spend on draining a node.
-		// The default value is 0, meaning that the node can be drained without any time limitations.
-		// NOTE: NodeDrainTimeout is different from `kubectl drain --timeout`
-		// +optional
-		NodeDrainTimeout *metav1.Duration `json:"nodeDrainTimeout,omitempty"`
```
CecileRobertMichon, line 164: note: this is a breaking change

it's okay since AzureMachinePool is still an experimental type but it needs to be clearly called out in the release notes and we might want to just mark it as deprecate for now instead of deleting it to be safe

exp/controllers/azuremachinepool_controller.go
CecileRobertMichon: can we not use a ownerReference/finalizer for this?

CecileRobertMichon (commented):
I have some questions about owner references and how the delete flow works that might be easier to talk through on a call

Glad to see all the drain code removed finally!

exp/controllers/azuremachinepoolmachine_controller.go
Jont828: This is because we need to check to see if the resource group is being deleted, and if so, we just go ahead and remove the finalizer since deletion can't finish until the finalizer is gone. I moved this up here as when we are deleting the RG, the AzureMachinePool is gone by the time we reach this code, and the AMPM reconciler gets stuck requeuing as it can't construct the scope when the AzureMachinePool is gone.

azure/scope/machinepool.go
Jont828: The AzureMachinePoolMachines are owned by the AzureMachinePool as well as a Machine. That Machine is owned by the MachinePool. What would you recommend?

exp/api/v1beta1/azuremachinepool_types.go
Jont828: If we leave it in, it will effectively do nothing. Does deprecating a field mean removing functionality in this way? I was under the impression that it meant it would no longer receive updates in favor of a new solution.

Jont828:
@CecileRobertMichon @mboersma I've added the delete priority behavior using the autoscaler annotation, and here is the cleaned up version of the diagrams documenting what this PR is meant to do. This should be ready for review once the tests pass!

Azure MPM Create
Azure MPM Delete

exp/controllers/azuremachinepool_controller.go
```diff
@@ -269,6 +269,9 @@ func (ampr *AzureMachinePoolReconciler) reconcileNormal(ctx context.Context, mac
 	defer done()
 
 	log.Info("Reconciling AzureMachinePool")
+
+	machinePoolScope.SetInfrastructureMachineKind()
```
CecileRobertMichon, line 273: Nit: it might be slightly better to call this in the machinepool scope `Close()` function since we don't actually patch the AzureMachinePool object until then, and that will ensure it always gets called before the function returns (regardless of what gets added in the future in `reconcileNormal`)

azure/scope/machinepool.go
```diff
@@ -477,18 +486,18 @@ func (m *MachinePoolScope) createMachine(ctx context.Context, machine azure.VMSS
...
+			Annotations: map[string]string{},
 		},
 		Spec: infrav1exp.AzureMachinePoolMachineSpec{
 			ProviderID: machine.ProviderID(),
 			InstanceID: machine.InstanceID,
 		},
 	}
 
+	labels := m.getMachinePoolMachineLabels()
+	labels[m.ClusterName()] = string(infrav1.ResourceLifecycleOwned)
```
CecileRobertMichon, line 498: why is this label not part of the `getMachinePoolMachineLabels` function?

azure/scope/machinepool.go
```diff
@@ -498,6 +507,28 @@ func (m *MachinePoolScope) createMachine(ctx context.Context, machine azure.VMSS
...
+func (m *MachinePoolScope) deleteMachine(ctx context.Context, ampm infrav1exp.AzureMachinePoolMachine) error {
+	ctx, log, done := tele.StartSpanWithLogger(ctx, "scope.MachinePoolScope.deleteMachine")
+	defer done()
+
+	machine, err := util.GetOwnerMachine(ctx, m.client, ampm.ObjectMeta)
+	if err != nil {
+		return errors.Wrapf(err, "error getting owner Machine for AzureMachinePoolMachine %s/%s", ampm.Namespace, ampm.Name)
+	}
+	if machine == nil {
+		// If Machine is nil, it has not been created. Wait for it to be created before deleting the AzureMachinePoolMachine if the MachinePool is not deleted.
```
CecileRobertMichon, line 519: this comment seems to contradict what we're actually doing, is it outdated?

azure/scope/machinepool.go
```diff
@@ -498,6 +507,28 @@ func (m *MachinePoolScope) createMachine(ctx context.Context, machine azure.VMSS
...
+		// If the MachinePool is deleted, the AzureMachinePoolMachine will be cleaned up by garbage collection.
+		log.Info("skipping deletion of AzureMachinePoolMachine because it has no owner", "ampm", ampm.Name)
+		return nil
+	}
+
+	if err := m.client.Delete(ctx, machine); err != nil {
+		return errors.Wrapf(err, "failed deleting Machine %s for AzureMachinePoolMachine %s in MachinePool %s", machine.Name, ampm.Name, m.MachinePool.Name)
```
CecileRobertMichon, line 526:
```suggestion
		return errors.Wrapf(err, "failed to delete Machine %s for AzureMachinePoolMachine %s in MachinePool %s", machine.Name, ampm.Name, m.MachinePool.Name)
```

exp/api/v1beta1/azuremachinepool_types.go
CecileRobertMichon: Ideally a non-breaking deprecation would first warn about the field getting removed, and then remove it in a future release. In this case since it will actually no longer work and the API is still experimental, removing it makes most sense but it should be called out as a breaking change in the PR release notes.

https://github.com/kubernetes/community/blob/master/contributors/guide/release-notes.md

azure/scope/strategies/machinepool_deployments/machinepool_deployment_strategy.go
```diff
@@ -296,6 +304,18 @@ func orderRandom(machines []infrav1exp.AzureMachinePoolMachine) []infrav1exp.Azu
 	return machines
 }
 
+// orderByDeleteMachineAnnotation will sort AzureMachinePoolMachines with the clusterv1.DeleteMachineAnnotation to the front of the list.
+// It will preserve the existing order of the list otherwise so that it respects the existing delete priority otherwise.
+func orderByDeleteMachineAnnotation(machines []infrav1exp.AzureMachinePoolMachine) []infrav1exp.AzureMachinePoolMachine {
```
CecileRobertMichon, line 309: can we add a unit test for this one?

exp/controllers/azuremachinepoolmachine_controller.go
```diff
@@ -110,6 +111,15 @@ func (ampmr *AzureMachinePoolMachineController) SetupWithManager(ctx context.Con
 		return errors.Wrapf(err, "failed adding a watch for AzureMachinePool model changes")
 	}
 
+	// Add a watch on CAPI Machines for MachinePool Machines
+	if err := c.Watch(
+		source.Kind(mgr.GetCache(), &clusterv1.Machine{}),
+		handler.EnqueueRequestsFromMapFunc(util.MachineToInfrastructureMapFunc(infrav1exp.GroupVersion.WithKind("AzureMachinePoolMachine"))),
+		predicates.ResourceNotPausedAndHasFilterLabel(log, ampmr.WatchFilterValue),
+	); err != nil {
+		return errors.Wrapf(err, "failed adding a watch for AzureMachinePool model changes")
```
CecileRobertMichon, line 120:
```suggestion
		return errors.Wrapf(err, "failed adding a watch for Machine model changes")
```

exp/controllers/azuremachinepoolmachine_controller.go
```diff
@@ -110,6 +111,15 @@ func (ampmr *AzureMachinePoolMachineController) SetupWithManager(ctx context.Con
 		return errors.Wrapf(err, "failed adding a watch for AzureMachinePool model changes")
 	}
 
+	// Add a watch on CAPI Machines for MachinePool Machines
+	if err := c.Watch(
+		source.Kind(mgr.GetCache(), &clusterv1.Machine{}),
+		handler.EnqueueRequestsFromMapFunc(util.MachineToInfrastructureMapFunc(infrav1exp.GroupVersion.WithKind("AzureMachinePoolMachine"))),
```
CecileRobertMichon, line 117: is this going to filter only Machines owned by an AzureMachinePool?

exp/controllers/azuremachinepoolmachine_controller.go
```diff
@@ -253,6 +292,8 @@ func (ampmr *AzureMachinePoolMachineController) reconcileNormal(ctx context.Cont
 		return reconcile.Result{}, nil
 	}
 
+	machineScope.UpdateDeleteMachineAnnotation()
```
CecileRobertMichon, line 295: Similar to above, not sure this is the best place to call this. In this case it would make sense to do this closer to where the annotation is being used, when selecting machines for delete.

exp/controllers/azuremachinepoolmachine_controller.go
CecileRobertMichon: > When the RG is being deleted, the AzureMachinePool gets deleted before the AzureMachinePoolMachines as the AMP removes the finalizer, bypassing the usual behavior of waiting for the children to be deleted

Are the AMPMs not owned by the AzureMachinePool or the MachinePool? if so, k8s garbage collection should take care of blocking the AMP deletion until all dependent AMPMs are deleted, regardless of CAPZ deleting its own controller finalizer...

https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/#owner-references-in-object-specifications

azure/scope/machinepool.go
Jont828: I meant for `getMachinePoolMachineLabels` to be a function for getting the labels we want to select AMPMs on. But maybe we could include that in the selector as well if it's present on every machine.

exp/controllers/azuremachinepoolmachine_controller.go
Jont828: Yes (implicitly) because it filters only for Machines that have an infraRef pointing to this specific AzureMachinePoolMachine.

exp/controllers/azuremachinepoolmachine_controller.go
Jont828: When we're selecting machines for delete, we're already in the AMP controller, which would require us to pull each of the owner Machines for each AMPM. I figured it would be easier to do so here as we already have access to the AMPM and its owner Machine. I think we could move it to `Close()` instead, though since that's when it gets patched.

exp/controllers/azuremachinepool_controller.go
Jont828: @CecileRobertMichon I'm running into an issue where when I move it into `Close()`, it fails the unit tests as it can't patch when the Cluster is paused. Is there a cleaner way to do this, since we don't have access to the Cluster in `Close()` to see if it's paused or not.

CecileRobertMichon: Are you saying it _should_ patch it when paused, or it _shouldn't_ patch it when paused? if it's the latter, should we just return before the defer if the machinepool is paused?

Can you point me to the unit test you're referring to ?

Jont828: To clarify, it should *not* patch when the Cluster is paused. Currently, [here](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/e2bd59253c859b151ac6f3222ad6c749e2a1ba75/exp/controllers/azuremachinepool_controller.go#L255) we have the defer `reconcilePause()` since the intention, I believe, is to always be closing the scope we create.

Jont828: [This](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/434c140926cbc4f88a66a6f1365ce0b4b54e9fc1/exp/controllers/azuremachinepool_controller_test.go#L62) is the test that's failing with an error saying we can't patch as the AzureMachinePool is not found. We can probably fix that error, but from a design POV I think it makes sense that we wouldn't want to silently update the status if the Cluster is paused.

CecileRobertMichon:
/retest

all the tests should be passing now

azure/scope/machinepool.go
```diff
@@ -55,6 +58,9 @@ import (
 // added here to avoid a circular dependency.
 const ScalesetsServiceName = "scalesets"
 
+// AzureMachinePoolMachineKind indicates the kind of an AzureMachinePoolMachine.
+const AzureMachinePoolMachineKind = "AzureMachinePoolMachine"
```
CecileRobertMichon, line 62: can we move this to the exp/API package? 

related to #4223

CecileRobertMichon:
Suggestion to update the release note to make the breaking change more explicit: `Removes the `NodeDrainTimeout` field from AzureMachinePools.` doesn't mark it as a breaking change (need to add action required: ) and doesn't give a clear action to take to the users

Suggest changing to: 

```
Implements MachinePool Machines according to CAPI contract. 

Action required: The `NodeDrainTimeout` field is now removed from AzureMachinePools. The equivalent field can be set on each Machine object.
```

Side question: is there a `NodeDrainTimeout` on MachinePools?

Jont828:
@CecileRobertMichon I believe it can be set on the Machine template in the MachinePool spec.

azure/scope/machinepool.go
```diff
@@ -234,6 +237,18 @@ func (m *MachinePoolScope) Name() string {
 	return m.AzureMachinePool.Name
 }
 
+// SetInfrastructureMachineKind sets the infrastructure machine kind in the status if it is not set already to support
```
mboersma, line 240:
```suggestion
// SetInfrastructureMachineKind sets the infrastructure machine kind in the status if it is not set already, returning
// `true` if the status was updated. This supports MachinePool Machines.
```

azure/scope/machinepool.go
```diff
@@ -234,6 +237,18 @@ func (m *MachinePoolScope) Name() string {
...
+// MachinePool Machines and returns a boolean indicating if the status was updated.
+func (m *MachinePoolScope) SetInfrastructureMachineKind() bool {
```
mboersma, line 242: I know this is a trivial func, but maybe it should have a unit test.

azure/scope/machinepool.go
```diff
@@ -436,16 +453,17 @@ func (m *MachinePoolScope) applyAzureMachinePoolMachines(ctx context.Context) er
...
 	}
 
-	for _, machine := range toDelete {
-		machine := machine
-		log.Info("deleting selected AzureMachinePoolMachine", "providerID", machine.Spec.ProviderID)
-		if err := m.client.Delete(ctx, &machine); err != nil {
+	// Delete MachinePool Machines as a part of scaling down
+	for i := range toDelete {
+		ampm := toDelete[i]
+		log.Info("deleting selected AzureMachinePoolMachine", "providerID", ampm.Spec.ProviderID)
```
mboersma, line 465:
```suggestion
		log.V(2).Info("deleting selected AzureMachinePoolMachine", "providerID", ampm.Spec.ProviderID)
```

Maybe? We seem to prefer `V(2)` in most other create/delete resource code.

azure/scope/machinepool.go
```diff
@@ -498,6 +515,34 @@ func (m *MachinePoolScope) createMachine(ctx context.Context, machine azure.VMSS
 	return nil
 }
 
+// DeleteMachine deletes an AzureMachinePoolMachine by fetching its owner Machine and deleting it to trigger node cordon/drain before deleting the infrastructure.
```
mboersma, line 518:
```suggestion
// DeleteMachine deletes an AzureMachinePoolMachine by fetching its owner Machine and deleting it. This ensures that the Node cordon/drain happens before deleting the infrastructure.
```

azure/scope/machinepool.go
```diff
@@ -498,6 +515,34 @@ func (m *MachinePoolScope) createMachine(ctx context.Context, machine azure.VMSS
...
+func (m *MachinePoolScope) DeleteMachine(ctx context.Context, ampm infrav1exp.AzureMachinePoolMachine) error {
+	ctx, log, done := tele.StartSpanWithLogger(ctx, "scope.MachinePoolScope.deleteMachine")
```
mboersma, line 520:
```suggestion
	ctx, log, done := tele.StartSpanWithLogger(ctx, "scope.MachinePoolScope.DeleteMachine")
```

azure/scope/machinepoolmachine.go
```diff
@@ -283,6 +285,19 @@ func (s *MachinePoolMachineScope) ProviderID() string {
 	return s.AzureMachinePoolMachine.Spec.ProviderID
 }
 
+// updateDeleteMachineAnnotation sets the clusterv1.DeleteMachineAnnotation on the AzureMachinePoolMachine if it exists on the owner Machine.
+func (s *MachinePoolMachineScope) updateDeleteMachineAnnotation() {
```
mboersma, line 289: Unit test?

azure/scope/machinepoolmachine.go
```diff
@@ -396,160 +412,6 @@ func (s *MachinePoolMachineScope) UpdateInstanceStatus(ctx context.Context) erro
 	return nil
 }
 
-// CordonAndDrain will cordon and drain the Kubernetes node associated with this AzureMachinePoolMachine.
```
mboersma, line 399: This gets deleted because we're reusing the Machine implementation of `CordonAndDrain`? 👏🏻

exp/controllers/azuremachinepoolmachine_controller.go
```diff
@@ -138,29 +149,55 @@ func (ampmr *AzureMachinePoolMachineController) Reconcile(ctx context.Context, r
...
+		logger.Info("AzureMachinePoolMachine or linked Cluster is marked as paused. Won't reconcile")
+		return ctrl.Result{}, nil
+	}
+
+	clusterScope, err := infracontroller.GetClusterScoper(ctx, logger, ampmr.Client, cluster)
+	if err != nil {
+		return reconcile.Result{}, errors.Wrapf(err, "failed to create cluster scope for cluster %s/%s", cluster.Namespace, cluster.Name)
+	}
 
+	logger.Info("Fetching AzureMachinePool with object meta", "meta", azureMachine.ObjectMeta)
```
mboersma, line 182:
```suggestion
	logger.V(2).Info("Fetching AzureMachinePool with object meta", "meta", azureMachine.ObjectMeta)
```

azure/scope/machinepool.go
Jont828: That log is from the existing code so I just kept it at the same level. IIRC log level 0 is for actions while level 2 is more for information.

azure/scope/machinepoolmachine.go
Jont828: Yep!

mboersma (approved):
/lgtm
/approve

Excellent work @Jont828! Thank you for being patient and persistent.

PR #4009, opened 2023-09-18
Support for custom dns prefix for aks clusters

LochanRn:
**What type of PR is this?**
/kind feature

Support for custom dns prefix for aks clusters

Fixes #

- [ ] cherry-pick candidate

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Support for custom dns prefix for aks clusters
```


api/v1beta1/azuremanagedcontrolplane_default.go
```diff
@@ -164,3 +164,9 @@ func (m *AzureManagedControlPlane) setDefaultOIDCIssuerProfile() {
 		m.Spec.OIDCIssuerProfile.Enabled = ptr.To(false)
 	}
 }
+
+func (m *AzureManagedControlPlane) setDefaultDNSPrefix() {
+	if m.Spec.DNSPrefix == nil {
+		m.Spec.DNSPrefix = ptr.To(m.Name)
```
Jont828, line 170:
```suggestion
		m.Spec.DNSPrefix = &m.Name
```
nit

api/v1beta1/azuremanagedcontrolplane_types.go
```diff
@@ -213,6 +213,10 @@ type AzureManagedControlPlaneSpec struct {
 	// OIDCIssuerProfile is the OIDC issuer profile of the Managed Cluster.
 	// +optional
 	OIDCIssuerProfile *OIDCIssuerProfile `json:"oidcIssuerProfile,omitempty"`
+
+	// This cannot be updated once the Managed Cluster has been created.
```
Jont828, line 217:
```suggestion
	// DNSPrefix is ... This cannot be updated once the Managed Cluster has been created.
```
Nit: add a doc comment explaining what the field is

api/v1beta1/azuremanagedcontrolplane_webhook.go
```diff
@@ -240,6 +241,14 @@ func (mw *azureManagedControlPlaneWebhook) ValidateUpdate(ctx context.Context, o
 		}
 	}
 
+	if ptr.Deref(m.Spec.DNSPrefix, "") != ptr.Deref(old.Spec.DNSPrefix, "") {
```
Jont828, line 244: I think we want to check if either value is nil first as nil is different from an empty string. Otherwise, the user could set the value from nil to an empty string and pass the check.

api/v1beta1/azuremanagedcontrolplane_webhook.go
```diff
@@ -302,6 +312,22 @@ func (m *AzureManagedControlPlane) Validate(cli client.Client) error {
...
+	if m.Spec.DNSPrefix == nil {
+		return nil
+	}
+
+	// Regex pattern for DNS prefix validation
+	// 1. Between 1 and 54 characters long: {1,54}
+	// 2. Alphanumerics and hyphens: [a-zA-Z0-9-]
+	// 3. Start and end with alphanumeric: ^[a-zA-Z0-9].*[a-zA-Z0-9]$
+	regex := regexp.MustCompile(`^[a-zA-Z0-9][a-zA-Z0-9-]{0,52}[a-zA-Z0-9]$`)
+	if regex.MatchString(ptr.Deref(m.Spec.DNSPrefix, "")) {
```
Jont828, line 325: If we know it's not nil we could probably dereference the pointer directly

api/v1beta1/azuremanagedcontrolplane_webhook_test.go
```diff
@@ -89,7 +91,7 @@ func TestDefaultingWebhook(t *testing.T) {
 	g.Expect(amcp.Spec.VirtualNetwork.Subnet.Name).To(Equal("fooSubnetName"))
 	g.Expect(amcp.Spec.SKU.Tier).To(Equal(PaidManagedControlPlaneTier))
 	g.Expect(*amcp.Spec.OIDCIssuerProfile.Enabled).To(BeTrue())
-
+	g.Expect(*amcp.Spec.DNSPrefix).To(Equal("test-prefix"))
```
Jont828, line 94:
```suggestion
	g.Expect(*amcp.Spec.DNSPrefix).To(Not(BeNil))
	g.Expect(*amcp.Spec.DNSPrefix).To(Equal("test-prefix"))
```

azure/services/managedclusters/spec.go
```diff
@@ -122,6 +122,9 @@ type ManagedClusterSpec struct {
 
 	// OIDCIssuerProfile is the OIDC issuer profile of the Managed Cluster.
 	OIDCIssuerProfile *OIDCIssuerProfile
+
+	// This cannot be updated once the Managed Cluster has been created.
```
Jont828, line 126:
```suggestion
	// DNSPrefix is ... This cannot be updated once the Managed Cluster has been created.
```
Start doc comment with name of field and what it does.

Jont828 (commented):
Thanks for the contribution! Seems reasonable to me, just a couple small nits

api/v1beta1/azuremanagedcontrolplane_default.go
LochanRn: Can I know why we should use the existing pointer of the m.Name and why it cannot be a new pointer ?

api/v1beta1/azuremanagedcontrolplane_types.go
LochanRn: Ack

api/v1beta1/azuremanagedcontrolplane_default.go
Jont828: I think it's more of a convention if anything, but elsewhere in the repo, we use the existing pointer whenever it's a variable and only use ptr.To for a constant or a function return result where we can't take the address with `&`.

nawazkh: This conversation reminded me of the subtle difference between `ptr.To` and `&`. Thanks for bringing it up.
TL;DR: `ptr.To` is safer and should be preferred over `&` unless we want other pointers to change our underlying value.

I have a demo here: https://goplay.tools/snippet/RRU2PypG0Z4
Check out how unexpected change to `ptr1` changes affects `ptr2` too. 
Whereas `ptr3` is safe, and its value is unaffected.

nawazkh: @LochanRn Is there a limit on the length of `m.Name`? Do you think it can violate the length constraint of 54 characters if `m.Spec.DNSPrefix` is unspecified and gets defaulted to `m.Name`? 
If not, can we add a logic that truncates the defaulted name to 54 characters and/or add a UID to the truncated name satisfying the `m.Spec.DNSPrefix` constraints?

azure/services/managedclusters/spec_test.go
```diff
@@ -589,7 +589,6 @@ func getSampleManagedCluster() armcontainerservice.ManagedCluster {
 	return armcontainerservice.ManagedCluster{
 		Properties: &armcontainerservice.ManagedClusterProperties{
 			KubernetesVersion: ptr.To("v1.22.0"),
-			DNSPrefix:         ptr.To("test-managedcluster"),
```
nawazkh, line 592: Was this deleted because `DNSPrefix` gets set to `s.DNSPrefix`?

api/v1beta1/azuremanagedcontrolplane_default.go
LochanRn: I have that constraint in the validating webhook. A user is using the custom dns prefix to set his own prefix name, I feel we should not tamper the same and returning the error will be better.

azure/services/managedclusters/spec_test.go
LochanRn: yes correct.

api/v1beta1/azuremanagedcontrolplane_webhook.go
LochanRn: Even if the user sets an empty string, that is invalid as it will not pass the regex so he will not be able to pass the check as m.Validate calls m.ValidateDNSPrefix which is going to throw an error.

api/v1beta1/azuremanagedcontrolplane_webhook.go
LochanRn: I think it is safer to deref regardless :)

api/v1beta1/azuremanagedcontrolplane_webhook.go
LochanRn: Updated the test cases PTAL.

api/v1beta1/azuremanagedcontrolplane_default.go
nawazkh: Sounds good, thanks.

Jont828: @nawazkh TIL, thanks for pointing that out. I think we can proceed with the ptr.To then

api/v1beta1/azuremanagedcontrolplane_webhook.go
Jont828: Maybe we can leave a comment explaining what you described? In general, having this line rely on validation elsewhere could cause problems for maintainability since someone changing it in the future could forget.

nawazkh:
@LochanRn you might want to run `make generate` and push the changes. 
From `pull-cluster-api-provider-azure-verify`
```
diff --git a/config/crd/bases/infrastructure.cluster.x-k8s.io_azuremanagedcontrolplanes.yaml b/config/crd/bases/infrastructure.cluster.x-k8s.io_azuremanagedcontrolplanes.yaml
index 91bc379e..3679424a 100644
--- a/config/crd/bases/infrastructure.cluster.x-k8s.io_azuremanagedcontrolplanes.yaml
+++ b/config/crd/bases/infrastructure.cluster.x-k8s.io_azuremanagedcontrolplanes.yaml
@@ -240,8 +240,8 @@ spec:
                 - port
                 type: object
               dnsPrefix:
-                description: This cannot be updated once the Managed Cluster has been
-                  created.
+                description: DNSPrefix allows the user to customize dns prefix. This
+                  cannot be updated once the Managed Cluster has been created.
                 type: string
               dnsServiceIP:
                 description: DNSServiceIP is an IP address assigned to the Kubernetes
generated files are out of date, run make generate
```

nawazkh:
Looks good to me :) 
/lgtm

api/v1beta1/azuremanagedcontrolplane_webhook.go
```diff
@@ -302,6 +312,22 @@ func (m *AzureManagedControlPlane) Validate(cli client.Client) error {
...
+		return nil
+	}
+	return errors.New("DNSPrefix is invalid")
```
nojnhuh, line 328: Including the regex in this message seems worthwhile so users have a bit more context as to why it's invalid.

api/v1beta1/azuremanagedcontrolplane_webhook.go
nojnhuh: Is there a reason we can't leverage the existing `webhookutils.ValidateImmutable` helper for this?

api/v1beta1/azuremanagedcontrolplane_types.go
```diff
@@ -213,6 +213,10 @@ type AzureManagedControlPlaneSpec struct {
 	// OIDCIssuerProfile is the OIDC issuer profile of the Managed Cluster.
 	// +optional
 	OIDCIssuerProfile *OIDCIssuerProfile `json:"oidcIssuerProfile,omitempty"`
+
+	// DNSPrefix allows the user to customize dns prefix. This cannot be updated once the Managed Cluster has been created.
```
nojnhuh, line 217: not to match formatting with other immutable fields:
```suggestion
	// DNSPrefix allows the user to customize dns prefix.
	// Immutable.
```

api/v1beta1/azuremanagedcontrolplane_webhook.go
nojnhuh: Or could this use [`// +kubebuilder:validation:Pattern`](https://book.kubebuilder.io/reference/markers/crd-validation.html)?

LochanRn: @nojnhuh updated changes based on your first comment.

api/v1beta1/azuremanagedcontrolplane_webhook.go
```diff
@@ -240,6 +241,14 @@ func (mw *azureManagedControlPlaneWebhook) ValidateUpdate(ctx context.Context, o
 		}
 	}
 
+	if err := webhookutils.ValidateImmutable(
+		field.NewPath("Spec", "OutboundType"),
```
nojnhuh, line 245:
```suggestion
		field.NewPath("Spec", "DNSPrefix"),
```

api/v1beta1/azuremanagedcontrolplane_webhook.go
```diff
@@ -302,6 +312,23 @@ func (m *AzureManagedControlPlane) Validate(cli client.Client) error {
...
+	// Regex pattern for DNS prefix validation
+	// 1. Between 1 and 54 characters long: {1,54}
+	// 2. Alphanumerics and hyphens: [a-zA-Z0-9-]
+	// 3. Start and end with alphanumeric: ^[a-zA-Z0-9].*[a-zA-Z0-9]$
+	pattern := `^[a-zA-Z0-9][a-zA-Z0-9-]{0,52}[a-zA-Z0-9]$`
+	regex := regexp.MustCompile(pattern)
+	if regex.MatchString(ptr.Deref(m.Spec.DNSPrefix, "")) {
+		return nil
+	}
+	return errors.New("DNSPrefix is invalid, does not match regex: " + pattern)
```
nojnhuh, line 329: Should we return a `field.Invalid()` error here like most of the other functions?

nawazkh (commented):
/lgtm
🚢

PR #4011, opened 2023-09-19
delete LRO state when operations fail

### LINKED ISSUE #3970: asyncpoller fails when SDK `PollUntilDone` returns a Done poller

nojnhuh:
**What type of PR is this?**
/kind bug
/kind flake

This PR fixes the CI flakes caused by #3970 by ensuring long-running operation states do not persist when the operation fails.

Fixes #3970

- [X] cherry-pick candidate
/cherry-pick release-1.11

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/azure/services/async/async.go
+++ b/azure/services/async/async.go
@@ -106,19 +106,22 @@ func (s *Service[C, D]) CreateOrUpdateResource(ctx context.Context, spec azure.R
 
 	result, poller, err := s.Creator.CreateOrUpdateAsync(ctx, spec, resumeToken, parameters)
 	errWrapped := errors.Wrapf(err, "failed to create or update resource %s/%s (service: %s)", rgName, resourceName, serviceName)
-	if poller != nil {
+	if poller != nil && azure.IsContextDeadlineExceededOrCanceledError(err) {
 		future, err := converters.PollerToFuture(poller, infrav1.PutFuture, serviceName, resourceName, rgName)
 		if err != nil {
 			return nil, errWrapped
 		}
 		s.Scope.SetLongRunningOperationState(future)
 		return nil, azure.WithTransientError(azure.NewOperationNotDoneError(future), requeueTime())
-	} else if err != nil {
-		return nil, errWrapped
 	}
 
-	// Once the operation is done, delete the long-running operation state.
+	// Once the operation is done, delete the long-running operation state. Even if the operation ended with
+	// an error, clear out any lingering state to try the operation again.
 	s.Scope.DeleteLongRunningOperationState(resourceName, serviceName, futureType)
+
+	if err != nil {
+		return nil, errWrapped
+	}
 
 	log.V(2).Info("successfully created or updated resource", "service", serviceName, "resource", resourceName, "resourceGroup", rgName)
 	return result, nil
@@ -147,19 +150,22 @@ func (s *Service[C, D]) DeleteResource(ctx context.Context, spec azure.ResourceS
 	// Delete the resource.
 	log.V(2).Info("deleting resource", "service", serviceName, "resource", resourceName, "resourceGroup", rgName)
 	poller, err := s.Deleter.DeleteAsync(ctx, spec, resumeToken)
-	if poller != nil {
+	if poller != nil && azure.IsContextDeadlineExceededOrCanceledError(err) {
 		future, err := converters.PollerToFuture(poller, infrav1.DeleteFuture, serviceName, resourceName, rgName)
 		if err != nil {
 			return errors.Wrap(err, "failed to convert poller to future")
 		}
 		s.Scope.SetLongRunningOperationState(future)
 		return azure.WithTransientError(azure.NewOperationNotDoneError(future), requeueTime())
-	} else if err != nil && !azure.ResourceNotFound(err) {
-		return errors.Wrapf(err, "failed to delete resource %s/%s (service: %s)", rgName, resourceName, serviceName)
 	}
 
-	// Once the operation is done, delete the long-running operation state.
+	// Once the operation is done, delete the long-running operation state. Even if the operation ended with
+	// an error, clear out any lingering state to try the operation again.
 	s.Scope.DeleteLongRunningOperationState(resourceName, serviceName, futureType)
+
+	if err != nil && !azure.ResourceNotFound(err) {
+		return errors.Wrapf(err, "failed to delete resource %s/%s (service: %s)", rgName, resourceName, serviceName)
+	}
 
 	log.V(2).Info("successfully deleted resource", "service", serviceName, "resource", resourceName, "resourceGroup", rgName)
 	return nil
--- a/azure/services/async/async_test.go
+++ b/azure/services/async/async_test.go
@@ -84,8 +84,22 @@ func TestServiceCreateOrUpdateResource(t *testing.T) {
 					r.ResourceName().Return(resourceName),
 					r.ResourceGroupName().Return(resourceGroupName),
 					s.GetLongRunningOperationState(resourceName, serviceName, infrav1.PutFuture).Return(validPutFuture),
-					c.CreateOrUpdateAsync(gomockinternal.AContext(), gomock.AssignableToTypeOf(azureResourceGetterType), resumeToken, gomock.Any()).Return(nil, fakePoller[MockCreator](g, http.StatusAccepted), nil),
+					c.CreateOrUpdateAsync(gomockinternal.AContext(), gomock.AssignableToTypeOf(azureResourceGetterType), resumeToken, gomock.Any()).Return(nil, fakePoller[MockCreator](g, http.StatusAccepted), context.DeadlineExceeded),
 					s.SetLongRunningOperationState(gomock.AssignableToTypeOf(&infrav1.Future{})),
+				)
+			},
+		},
+		{
+			name:          "operation failed",
+			serviceName:   serviceName,
+			expectedError: "failed to create or update resource mock-resourcegroup/mock-resource (service: mock-service): foo",
+			expect: func(g *WithT, s *mock_async.MockFutureScopeMockRecorder, c *mock_async.MockCreatorMockRecorder[MockCreator], r *mock_azure.MockResourceSpecGetterMockRecorder) {
+				gomock.InOrder(
+					r.ResourceName().Return(resourceName),
+					r.ResourceGroupName().Return(resourceGroupName),
+					s.GetLongRunningOperationState(resourceName, serviceName, infrav1.PutFuture).Return(validPutFuture),
+					c.CreateOrUpdateAsync(gomockinternal.AContext(), gomock.AssignableToTypeOf(azureResourceGetterType), resumeToken, gomock.Any()).Return(nil, fakePoller[MockCreator](g, http.StatusAccepted), errors.New("foo")),
+					s.DeleteLongRunningOperationState(resourceName, serviceName, infrav1.PutFuture),
 				)
 			},
 		},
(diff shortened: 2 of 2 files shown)


nojnhuh:
This time it didn't fail, but a few of these popped up:
```
I0919 21:38:16.442214       1 async.go:113] async.Service.CreateOrUpdateResource "msg"="CreateOrUpdateAsync returned poller in terminal state" "AzureManagedMachinePool"={"name":"capz-e2e-tycqsv-aks-pool1","namespace":"capz-e2e-tycqsv"} "controller"="azuremanagedmachinepool" "controllerGroup"="infrastructure.cluster.x-k8s.io" "controllerKind"="AzureManagedMachinePool" "error"="context deadline exceeded" "name"="capz-e2e-tycqsv-aks-pool1" "namespace"="capz-e2e-tycqsv" "reconcileID"="13f8d044-d520-41c4-9a00-763c3ae30812" "resultErr"=null "x-ms-correlation-request-id"="a200e06a-73f7-455f-9d8b-237ca0913740"
```
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/4011/pull-cluster-api-provider-azure-e2e-aks/1704240032052678656

I'd like to know if the ones causing the test to fail look the same though.

nojnhuh:
Finally got a repro and this is what the new log is writing:
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/4011/pull-cluster-api-provider-azure-e2e-aks/1704307759115145216
`I0920 02:35:39.306971       1 async.go:160] async.Service.DeleteResource "msg"="DeleteAsync returned poller in terminal state" "AzureManagedMachinePool"={"name":"pool3","namespace":"capz-e2e-6c5zgv"} "controller"="azuremanagedmachinepool" "controllerGroup"="infrastructure.cluster.x-k8s.io" "controllerKind"="AzureManagedMachinePool" "error"="GET https://management.azure.com/subscriptions/===REDACTED===/providers/Microsoft.ContainerService/locations/westeurope/operations/1cbd5af2-4080-45d9-8d28-5fef6ea455bd\n--------------------------------------------------------------------------------\nRESPONSE 200: 200 OK\nERROR CODE UNAVAILABLE\n--------------------------------------------------------------------------------\n{\n  \"name\": \"f25abd1c-8040-d945-8d28-5fef6ea455bd\",\n  \"status\": \"Canceled\",\n  \"startTime\": \"2023-09-20T02:35:29.2316449Z\",\n  \"endTime\": \"2023-09-20T02:35:37.4607486Z\",\n  \"error\": {\n    \"code\": \"\",\n    \"message\": \"operation discarded: Category: InternalError; Code: NotLatestOperation; SubCode: ; Message: Cannot proceed with the operation. Either the operation has been preempted by another one, or the information needed by the operation failed to be saved (or hasn't been saved yet).; InnerMessage: Operation ID does not match latest operation ID in goal state. Operation ID: 1cbd5af2-4080-45d9-8d28-5fef6ea455bd, goalstate latest operation ID: 966b6f7a-fa74-4968-ac3a-48aa70f7fc37. This could mean that after the operation is enqueued, the code that saves the goal state kept failing (HCP/database issue), or the operation has been preempted by a new DELETE operation.; Dependency: ; AKSTeam: ; OriginalError: current goalstate is not associated with the ongoing operation\"\n  }\n}\n--------------------------------------------------------------------------------\n" "name"="pool3" "namespace"="capz-e2e-6c5zgv" "reconcileID"="478e6211-4faf-4b5b-921e-17e5960b4927" "resultErr"="GET https://management.azure.com/subscriptions/===REDACTED===/providers/Microsoft.ContainerService/locations/westeurope/operations/1cbd5af2-4080-45d9-8d28-5fef6ea455bd\n--------------------------------------------------------------------------------\nRESPONSE 200: 200 OK\nERROR CODE UNAVAILABLE\n--------------------------------------------------------------------------------\n{\n  \"name\": \"f25abd1c-8040-d945-8d28-5fef6ea455bd\",\n  \"status\": \"Canceled\",\n  \"startTime\": \"2023-09-20T02:35:29.2316449Z\",\n  \"endTime\": \"2023-09-20T02:35:37.4607486Z\",\n  \"error\": {\n    \"code\": \"\",\n    \"message\": \"operation discarded: Category: InternalError; Code: NotLatestOperation; SubCode: ; Message: Cannot proceed with the operation. Either the operation has been preempted by another one, or the information needed by the operation failed to be saved (or hasn't been saved yet).; InnerMessage: Operation ID does not match latest operation ID in goal state. Operation ID: 1cbd5af2-4080-45d9-8d28-5fef6ea455bd, goalstate latest operation ID: 966b6f7a-fa74-4968-ac3a-48aa70f7fc37. This could mean that after the operation is enqueued, the code that saves the goal state kept failing (HCP/database issue), or the operation has been preempted by a new DELETE operation.; Dependency: ; AKSTeam: ; OriginalError: current goalstate is not associated with the ongoing operation\"\n  }\n}\n--------------------------------------------------------------------------------\n" "x-ms-correlation-request-id"="2e902dca-a495-4e9a-95d5-838fe833e34e"`

nojnhuh:
The root of the specific problem we're seeing in the e2e test seems to be that we weren't throttling transient errors in the AMMP controller. It looks like there's some hiccup when we end up sending a bunch of deletes in a short timeframe where AKS starts two different delete operations on the same agent pool. AKS lets the second delete supersede the first, but CAPZ doesn't notice the second one so keeps polling the first one, leading to the [above error](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/4011#issuecomment-1726883804).

I think adding transient error handling to the AMMP controller is an obvious-enough improvement that that's worth merging. That alone might fix the flakes, though it doesn't really target the scope of #3970. @mboersma @CecileRobertMichon Do you have thoughts on whether or not to keep the logging that this PR currently adds in case related issues pop up later?

CecileRobertMichon:
> It looks like there's some hiccup when we end up sending a bunch of deletes in a short timeframe where AKS starts two different delete operations on the same agent pool

Why are we sending a DELETE API request if there is already an ongoing long running DELETE operation? that seems like a bug that could also lead to API throttling

nojnhuh:
> > It looks like there's some hiccup when we end up sending a bunch of deletes in a short timeframe where AKS starts two different delete operations on the same agent pool
> 
> Why are we sending a DELETE API request if there is already an ongoing long running DELETE operation? that seems like a bug that could also lead to API throttling

I didn't get that far, but adding the RequeueAfter for transient errors seems to get around this AFAICT.

nojnhuh:
The only theory I can posit is that we add the `longRunningOperationStates` for the delete, immediately requeue, and the read cache is still stale and the object doesn't have that resume token yet, so we start a new operation. I'd hope that the client is smart enough to cache the result of the write though.

CecileRobertMichon:
> I didn't get that far, but adding the RequeueAfter for transient errors seems to get around this AFAICT.

It seems to me that "controller requeues too often for transient errors" and "async poller sends a DELETE call when there is already an ongoing one" are potentially two separate bugs, even if the first mitigates the symptoms of the second we should dig deeper to understand why the second occurs

nojnhuh:
I've opened a separate PR with the AMMP transient error handling and I'll leave this open to continue investigating.
#4039

nojnhuh:
So I think the root of the problem is that we set the `longRunningOperationStates` whenever an operation fails, for any reason, where we should be checking for sure that the operation failed because it's not done yet. If it fails for any other reason, we should delete the existing `longRunningOperationStates` so the next reconcile tries again from scratch.

With the changes I just pushed, I hit the same error locally (only once for the pool, as expected) and the test seemed to recover. Here are the full logs for that delete: [poolspot-delete.log](https://github.com/kubernetes-sigs/cluster-api-provider-azure/files/12731426/poolspot-delete.log)

azure/services/async/async.go
```diff
@@ -106,20 +106,23 @@ func (s *Service[C, D]) CreateOrUpdateResource(ctx context.Context, spec azure.R
 
 	result, poller, err := s.Creator.CreateOrUpdateAsync(ctx, spec, resumeToken, parameters)
 	errWrapped := errors.Wrapf(err, "failed to create or update resource %s/%s (service: %s)", rgName, resourceName, serviceName)
-	if poller != nil {
+	if poller != nil && azure.IsContextDeadlineExceededOrCanceledError(err) {
 		future, err := converters.PollerToFuture(poller, infrav1.PutFuture, serviceName, resourceName, rgName)
 		if err != nil {
 			return nil, errWrapped
 		}
 		s.Scope.SetLongRunningOperationState(future)
```
nojnhuh, line 114: The function change here is to only `SetLongRunningOperationState` when the error is something like "not done yet." In all other cases, successful or not, we `DeleteLongRunningOperationState`.

nojnhuh:
I'm fairly confident this fixes the issue now.

/retitle delete LRO state when operations fail
/hold for squash
/assign @mboersma @CecileRobertMichon

azure/services/async/async.go
```diff
@@ -106,20 +106,23 @@ func (s *Service[C, D]) CreateOrUpdateResource(ctx context.Context, spec azure.R
...
 		return nil, azure.WithTransientError(azure.NewOperationNotDoneError(future), requeueTime())
-	} else if err != nil {
-		return nil, errWrapped
 	}
 
-	// Once the operation is done, delete the long-running operation state.
+	// Once the operation is done, delete the long-running operation state. Even if the operation ended with
+	// an error, clear out any lingering state to try the operation again.
```
CecileRobertMichon, line 119: this is the equivalent of what we were doing before https://github.com/kubernetes-sigs/cluster-api-provider-azure/commit/a521b7b0ce8c200519ed779e951db019130cde7d#diff-0e657fbf13cf152e97cf8871a3baf550199d64f99f94316e7e1b9eeb5d6cc8e4L90

CecileRobertMichon:
Great find @nojnhuh

Do we still want the change to handle transient errors in the AMMP controller so we don't requeue too aggressively?

nojnhuh:
> Great find @nojnhuh
> 
> Do we still want the change to handle transient errors in the AMMP controller so we don't requeue too aggressively?

Thanks! And yes, I think that is also valuable still.

CecileRobertMichon:
I just saw you had a seperate PR open for that one already https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/4039 :)

/lgtm

mboersma (commented):
/lgtm

Needs squash. Nice work!

nojnhuh:
squashed!
/hold cancel

CecileRobertMichon (commented):
/approve

azure/services/async/async.go
nawazkh: Asking for clarity(I may be way off the workflow), what happens to _LongRunningRequests_ if they fail or if the higher abstraction resource cancels the context due to some failure ? Do we intend to re-queue this resource's request and not save its state in `Status.Condition`?

nojnhuh: Right. If an operation fails, then it's done, and we shouldn't keep polling that specific operation because it will remain failed. In that case we return the error which will requeue the resource and a new operation will start.

PR #4052, opened 2023-09-28
Allow joining AzureMachinePools to AKS clusters

### LINKED ISSUE #826: Integrate AzureMachine with AzureManagedControlPlane (BYO nodes on AKS)

CecileRobertMichon:
**What type of PR is this?**

**What this PR does / why we need it**: This allows joining a self-managed Azure VMSS nodepool to a CAPZ managed AKS cluster (AzureManagedControlPlane) for development and testing purposes.

Huge credits go to @alexeldeib for initial idea + POC.

Fixes #826 

- [ ] cherry-pick candidate

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Allow joining AzureMachinePools to AKS clusters
```

### DIFF
--- a/azure/interfaces.go
+++ b/azure/interfaces.go
@@ -21,8 +21,10 @@ import (
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-service-operator/v2/pkg/genruntime"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // Reconciler is a generic interface for a controller reconciler which has Reconcile and Delete methods.
@@ -79,6 +81,7 @@ type NetworkDescriber interface {
 type ClusterDescriber interface {
 	Authorizer
 	ResourceGroup() string
+	NodeResourceGroup() string
 	ClusterName() string
 	Location() string
 	ExtendedLocation() *infrav1.ExtendedLocationSpec
@@ -104,6 +107,9 @@ type AsyncStatusUpdater interface {
 type ClusterScoper interface {
 	ClusterDescriber
 	NetworkDescriber
+	AsyncStatusUpdater
+	GetClient() client.Client
+	GetDeletionTimestamp() *metav1.Time
 }
 
 // ManagedClusterScoper defines the interface for ManagedClusterScope.
--- a/azure/scope/machinepool.go
+++ b/azure/scope/machinepool.go
@@ -179,7 +179,7 @@ func (m *MachinePoolScope) ScaleSetSpec(ctx context.Context) azure.ResourceSpecG
 
 	spec := &scalesets.ScaleSetSpec{
 		Name:                         m.Name(),
-		ResourceGroup:                m.ResourceGroup(),
+		ResourceGroup:                m.NodeResourceGroup(),
 		Size:                         m.AzureMachinePool.Spec.Template.VMSize,
 		Capacity:                     int64(ptr.Deref[int32](m.MachinePool.Spec.Replicas, 0)),
 		SSHKeyData:                   m.AzureMachinePool.Spec.Template.SSHPublicKey,
@@ -741,7 +741,7 @@ func (m *MachinePoolScope) RoleAssignmentSpecs(principalID *string) []azure.Reso
 		roles[0] = &roleassignments.RoleAssignmentSpec{
 			Name:             m.SystemAssignedIdentityName(),
 			MachineName:      m.Name(),
-			ResourceGroup:    m.ResourceGroup(),
+			ResourceGroup:    m.NodeResourceGroup(),
 			ResourceType:     azure.VirtualMachineScaleSet,
 			Scope:            m.SystemAssignedIdentityScope(),
 			RoleDefinitionID: m.SystemAssignedIdentityDefinitionID(),
@@ -777,7 +777,7 @@ func (m *MachinePoolScope) VMSSExtensionSpecs() []azure.ResourceSpecGetter {
 				Settings:          extension.Settings,
 				ProtectedSettings: extension.ProtectedSettings,
 			},
-			ResourceGroup: m.ResourceGroup(),
+			ResourceGroup: m.NodeResourceGroup(),
 		})
 	}
 
@@ -787,7 +787,7 @@ func (m *MachinePoolScope) VMSSExtensionSpecs() []azure.ResourceSpecGetter {
 	if bootstrapExtensionSpec != nil {
 		extensionSpecs = append(extensionSpecs, &scalesets.VMSSExtensionSpec{
 			ExtensionSpec: *bootstrapExtensionSpec,
-			ResourceGroup: m.ResourceGroup(),
+			ResourceGroup: m.NodeResourceGroup(),
 		})
 	}
 
--- a/azure/scope/managedcontrolplane.go
+++ b/azure/scope/managedcontrolplane.go
@@ -19,14 +19,19 @@ package scope
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"strings"
 	"time"
 
 	asoresourcesv1 "github.com/Azure/azure-service-operator/v2/api/resources/v1api20200601"
 	"github.com/pkg/errors"
 	"golang.org/x/mod/semver"
+	"gopkg.in/yaml.v3"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+	bootstrapapi "k8s.io/cluster-bootstrap/token/api"
 	"k8s.io/utils/ptr"
 	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
 	"sigs.k8s.io/cluster-api-provider-azure/azure"
@@ -39,13 +44,19 @@ import (
 	"sigs.k8s.io/cluster-api-provider-azure/util/maps"
 	"sigs.k8s.io/cluster-api-provider-azure/util/tele"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"sigs.k8s.io/cluster-api/controllers/remote"
 	"sigs.k8s.io/cluster-api/util/conditions"
 	"sigs.k8s.io/cluster-api/util/patch"
 	"sigs.k8s.io/cluster-api/util/secret"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
-const resourceHealthWarningInitialGracePeriod = 1 * time.Hour
+const (
+	resourceHealthWarningInitialGracePeriod = 1 * time.Hour
+	// managedControlPlaneScopeName is the sourceName, or more specifically the UserAgent, of client used to store the Cluster Info configmap.
+	managedControlPlaneScopeName = "azuremanagedcontrolplane-scope"
+)
 
 // ManagedControlPlaneScopeParams defines the input parameters used to create a new managed
 // control plane.
@@ -137,6 +148,11 @@ type ManagedControlPlaneCache struct {
 // GetClient returns the controller-runtime client.
 func (s *ManagedControlPlaneScope) GetClient() client.Client {
 	return s.Client
+}
+
+// GetDeletionTimestamp returns the deletion timestamp of the cluster.
+func (s *ManagedControlPlaneScope) GetDeletionTimestamp() *metav1.Time {
+	return s.Cluster.DeletionTimestamp
 }
 
 // ResourceGroup returns the managed control plane's resource group.
@@ -418,7 +434,7 @@ func (s *ManagedControlPlaneScope) APIServerLBName() string {
 }
 
 // APIServerLBPoolName returns the API Server LB backend pool name.
-func (s *ManagedControlPlaneScope) APIServerLBPoolName(_ string) string {
+func (s *ManagedControlPlaneScope) APIServerLBPoolName() string {
 	return "" // does not apply for AKS
 }
 
@@ -680,6 +696,64 @@ func (s *ManagedControlPlaneScope) GetUserKubeconfigData() []byte {
 // SetUserKubeconfigData sets userKubeconfig data.
 func (s *ManagedControlPlaneScope) SetUserKubeconfigData(kubeConfigData []byte) {
 	s.userKubeConfigData = kubeConfigData
+}
+
+// MakeClusterCA returns a cluster CA Secret for the managed control plane.
+func (s *ManagedControlPlaneScope) MakeClusterCA() *corev1.Secret {
+	return &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      secret.Name(s.Cluster.Name, secret.ClusterCA),
+			Namespace: s.Cluster.Namespace,
+			OwnerReferences: []metav1.OwnerReference{
+				*metav1.NewControllerRef(s.ControlPlane, infrav1.GroupVersion.WithKind("AzureManagedControlPlane")),
+			},
+		},
+	}
+}
+
+// StoreClusterInfo stores the discovery cluster-info configmap in the kube-public namespace on the AKS cluster so kubeadm can access it to join nodes.
+func (s *ManagedControlPlaneScope) StoreClusterInfo(ctx context.Context, caData []byte) error {
+	remoteclient, err := remote.NewClusterClient(ctx, managedControlPlaneScopeName, s.Client, types.NamespacedName{
+		Namespace: s.Cluster.Namespace,
+		Name:      s.Cluster.Name,
+	})
+	if err != nil {
+		return errors.Wrap(err, "failed to create remote cluster kubeclient")
+	}
+
+	discoveryFile := clientcmdapi.NewConfig()
+	discoveryFile.Clusters[""] = &clientcmdapi.Cluster{
+		CertificateAuthorityData: caData,
+		Server: fmt.Sprintf(
+			"%s:%d",
+			s.ControlPlane.Spec.ControlPlaneEndpoint.Host,
+			s.ControlPlane.Spec.ControlPlaneEndpoint.Port,
+		),
+	}
+
+	data, err := yaml.Marshal(&discoveryFile)
+	if err != nil {
+		return errors.Wrap(err, "failed to serialize cluster-info to yaml")
+	}
+
+	clusterInfo := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      bootstrapapi.ConfigMapClusterInfo,
+			Namespace: metav1.NamespacePublic,
+		},
+		Data: map[string]string{
+			bootstrapapi.KubeConfigKey: string(data),
+		},
+	}
+
+	if _, err := controllerutil.CreateOrUpdate(ctx, remoteclient, clusterInfo, func() error {
+		clusterInfo.Data[bootstrapapi.KubeConfigKey] = string(data)
+		return nil
+	}); err != nil {
+		return errors.Wrapf(err, "failed to reconcile certificate authority data secret for cluster")
+	}
+
+	return nil
 }
 
 // SetKubeletIdentity sets the ID of the user-assigned identity for kubelet if not already set.
--- a/azure/services/groups/groups.go
+++ b/azure/services/groups/groups.go
@@ -53,23 +53,15 @@ func New(scope GroupScope) *Service {
 	}
 }
 
-// IsManaged returns true if the ASO ResourceGroup was created by CAPZ,
-// meaning that the resource group's lifecycle is managed.
-func (s *Service) IsManaged(ctx context.Context, spec azure.ASOResourceSpecGetter[*asoresourcesv1.ResourceGroup]) (bool, error) {
-	return aso.IsManaged(ctx, s.Scope.GetClient(), spec, s.Scope.ClusterName())
-}
-
-// ShouldDeleteIndividualResources returns false if the resource group is
+// IsManaged returns true if all resource groups are
 // managed and reconciled by ASO, meaning that we can rely on a single resource
 // group delete operation as opposed to deleting every individual resource.
-func (s *Service) ShouldDeleteIndividualResources(ctx context.Context) bool {
+func (s *Service) IsManaged(ctx context.Context) (bool, error) {
 	// Unless all resource groups are managed by CAPZ and reconciled by ASO, resources need to be deleted individually.
 	for _, spec := range s.Specs {
-		// Since this is a best effort attempt to speed up delete, we don't fail the delete if we can't get the RG status.
-		// Instead, take the long way and delete all resources one by one.
-		managed, err := s.IsManaged(ctx, spec)
+		managed, err := aso.IsManaged(ctx, s.Scope.GetClient(), spec, s.Scope.ClusterName())
 		if err != nil || !managed {
-			return true
+			return managed, err
 		}
 
 		// For ASO, "managed" only tells us that we're allowed to delete the ASO
@@ -78,8 +70,8 @@ func (s *Service) ShouldDeleteIndividualResources(ctx context.Context) bool {
 		group := spec.ResourceRef()
 		err = s.Scope.GetClient().Get(ctx, client.ObjectKeyFromObject(group), group)
 		if err != nil || group.GetAnnotations()[asoannotations.ReconcilePolicy] != string(asoannotations.ReconcilePolicyManage) {
-			return true
+			return false, err
 		}
 	}
-	return false
+	return true, nil
 }
--- a/controllers/azurecluster_reconciler.go
+++ b/controllers/azurecluster_reconciler.go
@@ -174,6 +174,7 @@ func (s *azureClusterService) Delete(ctx context.Context) error {
 		if err != nil {
 			return errors.Wrap(err, "failed to get group service")
 		}
+
 		// Delete the entire resource group directly.
 		if err := groupSvc.Delete(ctx); err != nil {
 			return errors.Wrap(err, "failed to delete resource group")
@@ -189,15 +190,6 @@ func (s *azureClusterService) Delete(ctx context.Context) error {
 	}
 
 	return nil
-}
-
-func (s *azureClusterService) getService(name string) (azure.ServiceReconciler, error) {
-	for _, service := range s.services {
-		if service.Name() == name {
-			return service, nil
-		}
-	}
-	return nil, errors.Errorf("service %s not found", name)
 }
 
 // setFailureDomainsForLocation sets the AzureCluster Status failure domains based on which Azure Availability Zones are available in the cluster location.
@@ -219,4 +211,13 @@ func (s *azureClusterService) setFailureDomainsForLocation(ctx context.Context) 
 	}
 
 	return nil
+}
+
+func (s *azureClusterService) getService(name string) (azure.ServiceReconciler, error) {
+	for _, service := range s.services {
+		if service.Name() == name {
+			return service, nil
+		}
+	}
+	return nil, errors.Errorf("service %s not found", name)
 }
--- a/controllers/azuremanagedcontrolplane_reconciler.go
+++ b/controllers/azuremanagedcontrolplane_reconciler.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 
 	"github.com/pkg/errors"
+	"k8s.io/client-go/tools/clientcmd"
 	"sigs.k8s.io/cluster-api-provider-azure/azure"
 	"sigs.k8s.io/cluster-api-provider-azure/azure/scope"
 	"sigs.k8s.io/cluster-api-provider-azure/azure/services/groups"
@@ -156,8 +157,31 @@ func (r *azureManagedControlPlaneService) reconcileKubeconfig(ctx context.Contex
 			}
 			return nil
 		}); err != nil {
-			return errors.Wrap(err, "failed to kubeconfig secret for cluster")
+			return errors.Wrap(err, "failed to reconcile kubeconfig secret for cluster")
 		}
+	}
+
+	// store cluster-info for the cluster with the admin kubeconfig.
+	kubeconfigFile, err := clientcmd.Load(kubeConfigs[0])
+	if err != nil {
+		return errors.Wrap(err, "failed to turn aks credentials into kubeconfig file struct")
+	}
+
+	cluster := kubeconfigFile.Contexts[kubeconfigFile.CurrentContext].Cluster
+	caData := kubeconfigFile.Clusters[cluster].CertificateAuthorityData
+	caSecret := r.scope.MakeClusterCA()
+	if _, err := controllerutil.CreateOrUpdate(ctx, r.kubeclient, caSecret, func() error {
+		caSecret.Data = map[string][]byte{
+			secret.TLSCrtDataName: caData,
+			secret.TLSKeyDataName: []byte("foo"),
+		}
+		return nil
+	}); err != nil {
+		return errors.Wrapf(err, "failed to reconcile certificate authority data secret for cluster")
+	}
+
+	if err := r.scope.StoreClusterInfo(ctx, caData); err != nil {
+		return errors.Wrap(err, "failed to construct cluster-info")
 	}
 
 	return nil
--- a/controllers/helpers.go
+++ b/controllers/helpers.go
@@ -76,6 +76,12 @@ type (
 		controller.Options
 		Cache *coalescing.ReconcileCache
 	}
+
+	// ClusterScoper is a interface used by AzureMachinePools that can be owned by either an AzureManagedCluster or AzureCluster.
+	ClusterScoper interface {
+		azure.ClusterScoper
+		groups.GroupScope
+	}
 )
 
 // AzureClusterToAzureMachinesMapper creates a mapping handler to transform AzureClusters into AzureMachines. The transform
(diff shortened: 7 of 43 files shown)


Tiltfile
```diff
@@ -380,8 +380,8 @@ def deploy_worker_templates(template, substitutions):
 
 def get_addons(flavor_name):
     # do not install calico and out of tree cloud provider for aks workload cluster
-    if "aks" in flavor_name:
-        return ""
+    # if "aks" in flavor_name:
```
CecileRobertMichon, line 383: need to revert this

azure/services/groups/groups.go
```diff
@@ -53,23 +53,21 @@ func New(scope GroupScope) *Service {
...
 	return aso.IsManaged(ctx, s.Scope.GetClient(), spec, s.Scope.ClusterName())
 }
 
-// ShouldDeleteIndividualResources returns false if the resource group is
-// managed and reconciled by ASO, meaning that we can rely on a single resource
+// IsManaged returns false if the resource group is
+// true and reconciled by ASO, meaning that we can rely on a single resource
 // group delete operation as opposed to deleting every individual resource.
-func (s *Service) ShouldDeleteIndividualResources(ctx context.Context) bool {
+func (s *Service) IsManaged(ctx context.Context) (bool, error) {
```
CecileRobertMichon, line 65: @nojnhuh curious what you think of this specifically... had to do a bit of refactor magic to avoid an import cycle and to avoid having to use `groups.Service` directly in the controller which made our existing tests that use mocks no longer work.

I ended up reverting https://github.com/kubernetes-sigs/cluster-api-provider-azure/commit/5c4827d3ee6d4ecbeef1288c0854ec2d05ad06fe#diff-72d15b0d5e2254b8eb66ae7e09bb8d11bb3f190b223f03972d72a4337af50480L43 so that https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/4052/files#diff-7d084ba6e9e15005850c1b44d72d457cef09f12badee416beda37f957a242364 can take a `azure.ServiceReconciler` instead of needing to create a new group service with a cluster scope on the fly. The main issues with that being:

clusterScope is now `ClusterScoper` in the AMP controller so that cluster infra can be either an `AzureCluster` or an `AzureManagedControlPlane`. In order to call  `ShouldDeleteIndividualResources` with `clusterScope` I would have needed to add `GroupSpecs()` to the `ClusterScoper` interface. That creates an import cycle between the azure package and the mock_azure package because the mock ClusterScoper now needs to reference `azure.ASOResourceSpecGetter`.

After trying different paths (changing packages around, moving interfaces, etc.) I don't see another way to make this work without making the controllers/reconcilers untestable (i.e. using a group.Service type directly). Curious to know your thoughts.

nojnhuh: Ideally I think `IsManaged` should live outside of `ServiceReconciler` if it's only the groups service that needs it outside of the service itself. Otherwise we end up with lots of dead `return true` copypasta in other services. And I think it's been brought up before that when `IsManaged` is defined for the whole service, that implies individual resources a service works with can't each be managed or unmanaged, even though e.g. the publicips service shows otherwise.

What if we split `IsManaged` into its own interface in azure/interfaces.go and then refer to that type in the definition of `ShouldDeleteIndividualResources`? I think it would look something like this: https://github.com/CecileRobertMichon/cluster-api-provider-azure/compare/byo-aks-node...nojnhuh:cluster-api-provider-azure:ismanageder

CecileRobertMichon: Thanks @nojnhuh I agree this is a lot better. I applied your suggestion (just renamed the interface to something I hope is a bit more descriptive). 

This should allow us to also get rid of the `IsManaged` functions in most services, I can do that cleanup separately. 

What do you think of the changes in groups.go specifically? I wasn't a fan of having two separate functions called `ShouldDeleteIndividualResources` that did different things and it doesn't seem like ASO needs the `IsManaged` function anywhere, so I think this simplifies things a bit (got rid of the separate `IsGroupManaged` function too since that's a one liner and not called anywhere else and calling that directly now).

nojnhuh: I don't feel strongly about the name of this function. The only thing I want to be conscious of is that the concept of "managed" is a little bit more ambiguous in the context of ASO since there's a difference between "managed ASO resource" and "managed Azure resource" and most of those details are tucked away inside the framework. For that reason I want to caution against establishing a common notion of "managed" among all the services since that's probably only useful in special cases like this where a very specific definition of "managed" is required.

> This should allow us to also get rid of the `IsManaged` functions in most services, I can do that cleanup separately.

Those will actually be useful to have for reference while converting services to use ASO and I've documented to remove it if it's no longer needed as part of this checklist: https://gist.github.com/nojnhuh/a1e06bb6d98b0eacb51134480ea4acea. I'd vote to keep the unused ones around in the meantime if we can bear it.

jsturtevant:
pull-cluster-api-provider-azure-windows-containerd-upstream-with-ci-artifacts  failed with 

```
E1017 06:02:22.380017    3828 remote_image.go:180] "PullImage from image service failed" err="rpc error: code = NotFound desc = failed to pull and unpack image \"capzci.azurecr.io/azure-cloud-node-manager:266c5e0\": no match for platform in manifest: not found" image="capzci.azurecr.io/azure-cloud-node-manager:266c5e0"
```

but it does exist:

```
regctl manifest get capzci.azurecr.io/azure-cloud-node-manager:266c5e0
Name:        capzci.azurecr.io/azure-cloud-node-manager:266c5e0
MediaType:   application/vnd.docker.distribution.manifest.list.v2+json
Digest:      sha256:677ab08c5061819b4e6866afa434a4d79cd7e8f3a5cea95d6b46fa7168d0a17f

Manifests:

  Name:      capzci.azurecr.io/azure-cloud-node-manager:266c5e0@sha256:d89b1fb20876f3804fa2f06577f4d071c7c5ee2d569ecf3495322daffbf4902e
  Digest:    sha256:d89b1fb20876f3804fa2f06577f4d071c7c5ee2d569ecf3495322daffbf4902e
  MediaType: application/vnd.docker.distribution.manifest.v2+json
  Platform:  linux/amd64

  Name:      capzci.azurecr.io/azure-cloud-node-manager:266c5e0@sha256:159d2aaf8fa076b4e65269d30119ed77de77d5d5d8988a58d314343e7d208f0b
  Digest:    sha256:159d2aaf8fa076b4e65269d30119ed77de77d5d5d8988a58d314343e7d208f0b
  MediaType: application/vnd.docker.distribution.manifest.v2+json
  Platform:  windows/amd64/10.0.17763.4974
  OSVersion: 10.0.17763.4974
```

as part of investigating #4144, retesting to see if get same error 
/test pull-cluster-api-provider-azure-windows-containerd-upstream-with-ci-artifacts

controllers/helpers.go
```diff
@@ -599,15 +598,18 @@ func GetAzureMachinePoolByName(ctx context.Context, c client.Client, namespace,
...
 	ctx, _, done := tele.StartSpanWithLogger(ctx, "controllers.ShouldDeleteIndividualResources")
 	defer done()
 
-	if clusterScope.Cluster.DeletionTimestamp.IsZero() {
+	if cluster.GetDeletionTimestamp().IsZero() {
 		return true
 	}
 
-	return groups.New(clusterScope).ShouldDeleteIndividualResources(ctx)
+	managed, err := grpSvc.IsManaged(ctx)
```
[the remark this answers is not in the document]
CecileRobertMichon: The `ShouldDeleteIndividualResources` was only defined for groups, I've refactored it to `IsManaged` and moved it to a new interface `BYOResourceReconciler` but the check should be equivalent.

Full discussion on this in this thread: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/4052#discussion_r1345027432

exp/api/v1beta1/azuremachinepool_default.go
```diff
@@ -73,14 +60,29 @@ func (amp *AzureMachinePool) SetDefaultSSHPublicKey() error {
...
 	if amp.Spec.RoleAssignmentName != "" && amp.Spec.SystemAssignedIdentityRole != nil && amp.Spec.SystemAssignedIdentityRole.Name != "" {
 		// Both the deprecated and the new fields are both set, return without changes
 		// and reject the request in the validating webhook which runs later.
-		return
+		return nil
 	}
 	if amp.Spec.Identity == infrav1.VMIdentitySystemAssigned {
+		machinePool, err := azureutil.FindParentMachinePoolWithRetry(amp.Name, client, 5)
+		if err != nil {
+			return errors.Wrap(err, "failed to find parent machine pool")
```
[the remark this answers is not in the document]
CecileRobertMichon: We're returning the error from the helper function but appending that error in `SetDefaults`. I think that's what we want because if this part of the function fails there's nothing else we can do to set identity defaults, but we do want to attempt to set other defaults in the AzureMachinePool spec.

exp/api/v1beta1/azuremachinepool_default_test.go
```diff
@@ -48,75 +53,138 @@ func TestAzureMachinePool_SetDefaultSSHPublicKey(t *testing.T) {
...
+	}
 
-	emptyTest.machinePool.SetIdentityDefaults(fakeSubscriptionID)
-	g.Expect(emptyTest.machinePool.Spec.SystemAssignedIdentityRole.Name).To(Not(BeEmpty()))
-	_, err := uuid.Parse(emptyTest.machinePool.Spec.SystemAssignedIdentityRole.Name)
-	g.Expect(err).To(Not(HaveOccurred()))
-	g.Expect(emptyTest.machinePool.Spec.SystemAssignedIdentityRole).To(Not(BeNil()))
-	g.Expect(emptyTest.machinePool.Spec.SystemAssignedIdentityRole.Scope).To(Equal(fmt.Sprintf("/subscriptions/%s/", fakeSubscriptionID)))
-	g.Expect(emptyTest.machinePool.Spec.SystemAssignedIdentityRole.DefinitionID).To(Equal(fmt.Sprintf("/subscriptions/%s/providers/Microsoft.Authorization/roleDefinitions/%s", fakeSubscriptionID, infrav1.ContributorRoleID)))
+	for _, tc := range tests {
```
nojnhuh, line 169: 💯 refactor

exp/api/v1beta1/azuremachinepool_default_test.go
```diff
@@ -48,75 +53,138 @@ func TestAzureMachinePool_SetDefaultSSHPublicKey(t *testing.T) {
...
 	fakeSubscriptionID := uuid.New().String()
 	fakeClusterName := "testcluster"
 	fakeRoleDefinitionID := "testroledefinitionid"
 	fakeScope := fmt.Sprintf("/subscriptions/%s/resourceGroups/%s", fakeSubscriptionID, fakeClusterName)
 	existingRoleAssignmentName := "42862306-e485-4319-9bf0-35dbc6f6fe9c"
-	roleAssignmentExistTest := test{machinePool: &AzureMachinePool{Spec: AzureMachinePoolSpec{
-		Identity: infrav1.VMIdentitySystemAssigned,
-		SystemAssignedIdentityRole: &infrav1.SystemAssignedIdentityRole{
-			Name: existingRoleAssignmentName,
+	machinePool := &expv1.MachinePool{
```
nojnhuh, line 61: I would vote to move these new objects inside the `t.Run` if they aren't referenced in the test cases just to make sure these don't get mutated in one test case such that it causes issues in another test case.

I realize though that here it's fairly obvious that the AzureMachinePool defaulting webhook won't be mutating these other kinds of objects.

exp/controllers/azuremachinepool_controller.go
```diff
@@ -210,25 +211,57 @@ func (ampr *AzureMachinePoolReconciler) Reconcile(ctx context.Context, req ctrl.
...
-	}
-	azureCluster := &infrav1.AzureCluster{}
-	if err := ampr.Client.Get(ctx, azureClusterName, azureCluster); err != nil {
-		logger.V(2).Info("AzureCluster is not available yet")
-		return reconcile.Result{}, nil
-	}
+	var clusterScope azure.ClusterScoper
+	var groupScope groups.GroupScope
+	switch cluster.Spec.InfrastructureRef.Kind {
+	case "AzureCluster":
```
nojnhuh, line 217: I imagine we have this string hardcoded in enough places already that it's probably worth making a constant for (and also AzureManagedControlPlane).

(Maybe this would be a good first issue?)

exp/controllers/azuremachinepool_controller.go
```diff
@@ -210,25 +211,57 @@ func (ampr *AzureMachinePoolReconciler) Reconcile(ctx context.Context, req ctrl.
...
-		Namespace: azMachinePool.Namespace,
-		Name:      cluster.Spec.InfrastructureRef.Name,
-	}
-	azureCluster := &infrav1.AzureCluster{}
-	if err := ampr.Client.Get(ctx, azureClusterName, azureCluster); err != nil {
-		logger.V(2).Info("AzureCluster is not available yet")
-		return reconcile.Result{}, nil
-	}
+	var clusterScope azure.ClusterScoper
+	var groupScope groups.GroupScope
```
nojnhuh, line 215: Offering this in case it makes anything easier if we know that these will always refer to the same thing:
```suggestion
	var clusterScope interface {
		azure.ClusterScoper
		groups.GroupScope
	}
```
Or perhaps a new typedef local to this package since it looks like that combo may be useful elsewhere.

exp/controllers/azuremachinepool_controller.go
```diff
@@ -210,25 +211,57 @@ func (ampr *AzureMachinePoolReconciler) Reconcile(ctx context.Context, req ctrl.
...
+	case "AzureManagedCluster":
+		logger = logger.WithValues("AzureManagedCluster", cluster.Spec.InfrastructureRef.Name)
+		azureManagedControlPlaneName := client.ObjectKey{
+			Namespace: azMachinePool.Namespace,
+			Name:      cluster.Spec.ControlPlaneRef.Name,
+		}
+		azureManagedControlPlane := &infrav1.AzureManagedControlPlane{}
+		if err := ampr.Client.Get(ctx, azureManagedControlPlaneName, azureManagedControlPlane); err != nil {
+			logger.V(2).Info("AzureManagedControlPlane is not available yet")
+			return reconcile.Result{}, nil
```
nojnhuh, line 249: Do we need to update the watches in `SetupWithManager` to include AzureManagedCluster/ControlPlane so we know this will eventually get requeued once the control plane becomes available?

exp/controllers/azuremachinepoolmachine_controller.go
```diff
@@ -186,27 +186,57 @@ func (ampmr *AzureMachinePoolMachineController) Reconcile(ctx context.Context, r
 		return ctrl.Result{}, nil
 	}
 
-	azureClusterName := client.ObjectKey{
-		Namespace: machine.Namespace,
-		Name:      cluster.Spec.InfrastructureRef.Name,
-	}
+	var clusterScope azure.ClusterScoper
+	switch cluster.Spec.InfrastructureRef.Kind {
```
nojnhuh, line 190: Is there a decent way to extract this "get cluster scoper" logic into a function that can be resued?

exp/controllers/azuremachinepoolmachine_controller.go
```diff
@@ -186,27 +186,57 @@ func (ampmr *AzureMachinePoolMachineController) Reconcile(ctx context.Context, r
...
+		clusterScope = scoper
+	case "AzureManagedCluster":
+		azureManagedControlPlaneName := client.ObjectKey{
+			Namespace: machine.Namespace,
+			Name:      cluster.Spec.ControlPlaneRef.Name,
+		}
+		azureManagedControlPlane := &infrav1.AzureManagedControlPlane{}
+		if err := ampmr.Client.Get(ctx, azureManagedControlPlaneName, azureManagedControlPlane); err != nil {
+			logger.Info("AzureManagedControlPlane is not available yet")
+			return reconcile.Result{}, nil
```
nojnhuh, line 223: ditto above re: watching AzureManagedControlPlane

exp/controllers/azuremachinepoolmachine_controller_test.go
```diff
@@ -132,13 +135,17 @@ func getAReadyMachinePoolMachineCluster() (*clusterv1.Cluster, *infrav1.AzureClu
 	}
 
 	cluster := &clusterv1.Cluster{
+		TypeMeta: metav1.TypeMeta{
+			Kind: "Cluster",
```
nojnhuh, line 139: Does the kind for these other objects need to be specified? I tend to lean toward not including them if they have no bearing on the result of the test.

azure/scope/managedcontrolplane.go
```diff
@@ -19,14 +19,18 @@ package scope
...
 	"context"
 	"encoding/json"
+	"fmt"
 	"strings"
 	"time"
 
 	asoresourcesv1 "github.com/Azure/azure-service-operator/v2/api/resources/v1api20200601"
 	"github.com/pkg/errors"
 	"golang.org/x/mod/semver"
+	"gopkg.in/yaml.v2"
```
nojnhuh, line 29: Could we import v3 here instead since it looks like we already use that through some dependency elsewhere? That seems to have a higher likelihood of getting updates down the line.

test/e2e/aks_byo_node.go
```diff
@@ -0,0 +1,190 @@
...
+	}, input.WaitIntervals...).Should(Succeed())
+
+	By("Verifying the MachinePool becomes ready")
+	Eventually(func(g Gomega) {
+		pool := &expv1.MachinePool{}
+		err := mgmtClient.Get(ctx, client.ObjectKeyFromObject(machinePool), pool)
+		g.Expect(err).NotTo(HaveOccurred())
+		g.Expect(conditions.IsTrue(pool, clusterv1.ReadyCondition)).To(BeTrue())
+	}, input.WaitIntervals...).Should(Succeed())
+}
```
nojnhuh, line 190: The other tests that create MachinePools have been deleting them at the end of the spec. Should we continue that pattern here?

nojnhuh (commented):
Leaving the comments I came up with for now. Will try this out tomorrow to see if I can get it working myself!

exp/controllers/azuremachinepoolmachine_controller_test.go
CecileRobertMichon: I'm pretty sure I added those because some of the tests that share this helper were failing because `AzureCluster` didn't have a Kind and we check the Kind value now in the controllers. I added them for all the objects returned by that function for consistency.

test/e2e/aks_byo_node.go
CecileRobertMichon: I did originally have the delete here. I can add it back. The reason I originally commented it out and then eventually removed it was that when we delete the machinePool before the end of the test we don't collect logs for it which turned out to be useful for debugging. On the plus side by doing the delete in the test we also test delete behavior.

Perhaps a good compromise is to not do the delete in a defer ("no matter what") but only delete the machine pool if everything else in the test passed? I do think it's helpful having logs of the machines and nodes if the test fails.

azure/scope/machinepool.go
```diff
@@ -179,7 +179,7 @@ func (m *MachinePoolScope) ScaleSetSpec(ctx context.Context) azure.ResourceSpecG
 
 	spec := &scalesets.ScaleSetSpec{
 		Name:                         m.Name(),
-		ResourceGroup:                m.ResourceGroup(),
+		ResourceGroup:                m.NodeResourceGroup(),
```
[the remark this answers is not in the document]
CecileRobertMichon: In practice it doesn't matter because AzureMachines don't work with anything else besides AzureCluster currently and the cluster scope just returns `ResourceGroup()` for `NodeResourceGroup()` but I changed everything in machine.go and machinepool.go to use `NodeResourceGroup()`.

azure/scope/managedcontrolplane.go
```diff
@@ -682,6 +697,67 @@ func (s *ManagedControlPlaneScope) SetUserKubeconfigData(kubeConfigData []byte)
...
+		return errors.Wrap(err, "failed to serialize cluster-info to yaml")
+	}
+
+	clusterInfo := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "cluster-info",
+			Namespace: "kube-public",
+		},
+		Data: map[string]string{
+			"kubeconfig": string(data),
```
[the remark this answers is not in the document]
CecileRobertMichon: done

azure/scope/managedcontrolplane.go
```diff
@@ -682,6 +697,67 @@ func (s *ManagedControlPlaneScope) SetUserKubeconfigData(kubeConfigData []byte)
...
+		},
+	}
+
+	// preserve before fetch into existing object
+	clusterInfoData := clusterInfo.Data
+
+	if _, err := controllerutil.CreateOrUpdate(ctx, remoteclient, clusterInfo, func() error {
+		clusterInfo.Data = clusterInfoData
```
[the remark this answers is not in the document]
CecileRobertMichon: changed

controllers/azurecluster_reconciler.go
```diff
@@ -159,21 +159,22 @@ func (s *azureClusterService) Delete(ctx context.Context) error {
 	ctx, _, done := tele.StartSpanWithLogger(ctx, "controllers.azureClusterService.Delete")
 	defer done()
 
-	if !ShouldDeleteIndividualResources(ctx, s.scope) {
+	groupSvc, err := GetService(groups.ServiceName, s.services)
```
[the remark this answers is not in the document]
CecileRobertMichon: I'm leaning toward keeping this as-is because it's not a huge gain but would require refactoring a bunch of unit tests and I already went down that road once when I was trying to have the function itself assert the type (ended up backing out) and that was not a good time. Also the function is used for vnetPeeringSvc too and I could see other usages in the future so it wouldn't make as much sense if it's more than 1-2 services being "special".

azure/interfaces.go
```diff
@@ -104,6 +112,9 @@ type AsyncStatusUpdater interface {
 type ClusterScoper interface {
 	ClusterDescriber
 	NetworkDescriber
+	AsyncStatusUpdater
+	GetClient() client.Client
+	GetDeletionTimestamp() *metav1.Time
```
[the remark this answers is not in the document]
CecileRobertMichon: I would prefer we add the Cluster itself if and when we do end up needing more fields, and keep it as-needed for now.

docs/book/src/topics/managedcluster.md
```diff
@@ -303,18 +300,6 @@ implementation simple. If you'd like to run managed AKS cluster with CAPZ
 and need an additional feature, please open a pull request or issue with
 details. We're happy to help!
 
-Current limitations
```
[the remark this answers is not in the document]
CecileRobertMichon: Yes, I checked all of them and none of these limitations apply anymore (yay)

exp/api/v1beta1/azuremachinepool_default_test.go
CecileRobertMichon: done

exp/controllers/azuremachinepool_controller.go
CecileRobertMichon: https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/4223

test/e2e/aks_byo_node.go
nojnhuh: For this one I can see it being valuable to keep around to collect logs more than the other ones. I think I remember trying to implement some kind of "do this if the test passed/failed" in Ginkgo and not seeing a good way to do that. I'm +1 to keeping this test as-is and not deleting the machine pool until the end with everything else.

controllers/azuremanagedcontrolplane_reconciler.go
```diff
@@ -156,9 +157,32 @@ func (r *azureManagedControlPlaneService) reconcileKubeconfig(ctx context.Contex
...
+	// store cluster-info for the cluster with the admin kubeconfig.
+	kubeconfigFile, err := clientcmd.Load(kubeConfigs[0])
+	if err != nil {
+		return errors.Wrap(err, "failed to turn aks credentials into kubeconfig file struct")
+	}
+
+	cluster := kubeconfigFile.Contexts[kubeconfigFile.CurrentContext].Cluster
+	caData := kubeconfigFile.Clusters[cluster].CertificateAuthorityData
+	caSecret := r.scope.MakeClusterCA()
+	if _, err := controllerutil.CreateOrUpdate(ctx, r.kubeclient, caSecret, func() error {
```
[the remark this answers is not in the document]
CecileRobertMichon: It's to make the KubeadmConfig controller happy when it generates the bootstrap data for the machinepool, it would normally be created when reconciling the KCP object but in this case there is none

it follows the convention defined here https://cluster-api.sigs.k8s.io/tasks/certs/using-custom-certificates

azure/scope/managedcontrolplane.go
```diff
@@ -738,18 +739,15 @@ func (s *ManagedControlPlaneScope) StoreClusterInfo(ctx context.Context, caData
 	clusterInfo := &corev1.ConfigMap{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "cluster-info",
```
nojnhuh, line 741: nit: I think I also saw there was a constant for this next to the `"kubeconfig"` one.

docs/book/src/topics/managedcluster.md
```diff
@@ -398,3 +383,130 @@ spec:
...
+kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/cluster-api-provider-azure/main/templates/addons/azure-cni-v1.yaml
+```
+
+### Notes
+
+Some notes about how this works under the hood:
+
+- CAPZ will fetch the kubeconfig for the AKS cluster and store it in a secret named `${CLUSTER_NAME}-kubeconfig` in the management cluster.
+- You can customize the `MachinePool`, `AzureMachinePool`, and `KubeadmConfig` resources to your liking. The example above is just a starting point.
+- The `KubeadmConfig` resource will be used to generate a `kubeadm join` command that will be executed on each node in the VMSS. It uses the cluster kubeconfig for discovery. The `kubeadm init phase upload-config all` is run as a preKubeadmCommand to ensure that the kubeadm and kubelet configurations are uploaded to a ConfigMap. This step would normally be done by the `kubeadm init` command, but since we're not running `kubeadm init` we need to do it manually.
```
nojnhuh, line 512: Other than `preKubeadmCommands`, are there any other non-standard aspects of this config that enable joining to AKS that users should know should \*not\* be changed from the example?

Or more generally, I wonder if it might be helpful for this doc to describe how to take a "default" machine pool config and transform it into something that works in AKS in addition to (or instead of) the reverse here ("here's a machine pool that works for AKS, customize it however you need").

docs/book/src/topics/managedcluster.md
```diff
@@ -398,3 +383,130 @@ spec:
...
+
+This is not an officially supported AKS scenario. It is meant to facilitate development and testing of alpha/beta Kubernetes features. Please use at your own risk.
+
+</aside>
+
+### Creating the MachinePool
+
+You can add a self-managed VMSS node pool to any CAPZ-managed AKS cluster by applying the following resources to the management cluster:
+
+```yaml
```
nojnhuh, line 401: Works On My Machine™ :tada:

exp/controllers/azuremachinepool_controller.go
CecileRobertMichon: Added a watch on `AzureManagedCluster`

exp/controllers/azuremachinepoolmachine_controller.go
CecileRobertMichon: This controller currently doesn't watch AzureClusters either, since it watches AzureMachinePools that should be enough though, I think?

docs/book/src/topics/managedcluster.md
CecileRobertMichon: Good question, everything that's in there is required (and defers from the otherwise example MachinePools we provide in other templates)

This part especially is important:

```
  joinConfiguration:
    discovery:
      file:
        kubeConfigPath: /etc/kubernetes/admin.conf
    nodeRegistration:
      kubeletExtraArgs:
        azure-container-registry-config: /etc/kubernetes/azure.json
        cloud-provider: external
```

along with:

```
files
  - contentFrom:
      secret:
        key: value
        name: ${CLUSTER_NAME}-kubeconfig
    owner: root:root
    path: /etc/kubernetes/admin.conf
    permissions: "0644"  
```

azure/scope/managedcontrolplane.go
CecileRobertMichon: nice! I looked for it and couldn't see it before 👁️

exp/controllers/azuremachinepoolmachine_controller.go
CecileRobertMichon: yes! thanks for the suggestion. Refactor incoming.

nojnhuh:
Tests should look better now?
/retest

exp/controllers/azuremachinepool_controller.go
CecileRobertMichon: This one was a pain... I tried to do this before but I ran into a tangle of import cycles between azure, mock_azure, groups, and aso packages 

Because of the refactor with the helper function it's now a lot cleaner to have a separate interface, I put it in the controllers helpers since it's used there and that way it avoids all the imprt cycles

docs/book/src/topics/managedcluster.md
CecileRobertMichon: Added a sentence to clarify

CecileRobertMichon:
Thanks for the review @nojnhuh, I believe I've addressed all your comments

exp/controllers/azuremachinepoolmachine_controller.go
CecileRobertMichon: The helper function now returns an error for this now so it should requeue right away (in practice I don't expect this error to happen outside of the initial yaml apply, unless the user forgot something in which case we want to alert them)

exp/controllers/azuremachinepoolmachine_controller.go
CecileRobertMichon: Done! this helped a lot

controllers/azurecluster_reconciler_test.go
```diff
@@ -180,15 +180,10 @@ func TestAzureClusterServiceDelete(t *testing.T) {
 	namespace := "ns"
 	resourceGroup := "rg"
 
-	type MockGroupsService struct {
-		*mock_azure.MockServiceReconciler
-		*mock_azure.MockBYOResourceReconciler
```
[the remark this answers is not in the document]
CecileRobertMichon: removed

nojnhuh (commented):
/lgtm

🪄🪄🪄

azure/scope/managedcontrolplane.go
```diff
@@ -418,7 +434,7 @@ func (s *ManagedControlPlaneScope) APIServerLBName() string {
 }
 
 // APIServerLBPoolName returns the API Server LB backend pool name.
-func (s *ManagedControlPlaneScope) APIServerLBPoolName(_ string) string {
```
jackfrancis, line 421: what was going on here earlier?

CecileRobertMichon: that was just a cleanup leftover from https://github.com/kubernetes-sigs/cluster-api-provider-azure/commit/cb9494062a4076368013ebcdfe85726eddf44a7a

azure/scope/managedcontrolplane.go
```diff
@@ -682,6 +698,64 @@ func (s *ManagedControlPlaneScope) SetUserKubeconfigData(kubeConfigData []byte)
...
+	remoteclient, err := remote.NewClusterClient(ctx, managedControlPlaneScopeName, s.Client, types.NamespacedName{
+		Namespace: s.Cluster.Namespace,
+		Name:      s.Cluster.Name,
+	})
+	if err != nil {
+		return errors.Wrap(err, "failed to create remote cluster kubeclient")
+	}
+
+	discoveryFile := clientcmdapi.NewConfig()
+	discoveryFile.Clusters[""] = &clientcmdapi.Cluster{
```
jackfrancis, line 725: What is the significance of using an empty string as the key that represents this `Cluster` object?

controllers/azurecluster_reconciler.go
```diff
@@ -162,18 +162,19 @@ func (s *azureClusterService) Delete(ctx context.Context) error {
 	if !ShouldDeleteIndividualResources(ctx, s.scope) {
 		// If the resource group is managed, delete it.
 		// We need to explicitly delete vnet peerings, as it is not part of the resource group.
-		vnetPeeringsSvc, err := s.getService(vnetpeerings.ServiceName)
+		vnetPeeringsSvc, err := getService(vnetpeerings.ServiceName, s.services)
```
jackfrancis, line 165: If we already have our set of services locally, why are we converting this from a private method into a func?

jackfrancis (approved):
added some comments, pending responses lgtm

great work @CecileRobertMichon!

controllers/azurecluster_reconciler.go
CecileRobertMichon: Good catch, at some point I refactored it to be able to use it across different reconcilers but that ended up not being necessary, will revert

azure/scope/managedcontrolplane.go
CecileRobertMichon: that's what kubeadm expects for discovery https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/util/kubeconfig/kubeconfig.go#L108

controllers/azurecluster_reconciler.go
CecileRobertMichon: Done!

CecileRobertMichon:
Thanks @jackfrancis! addressed your comments

PR #4059, opened 2023-09-30
migrate natgateways to use aso

### LINKED ISSUE #4040: ASO service: natgateways
Convert the natgateways service to use ASO.

```[tasklist]
### Dependencies
- [ ] #3527
```

nawazkh:
**What type of PR is this?**
/kind feature

- Migrate NAT Gateways to ASO framework

Fixes #4040 

- Commit d76bf75dca41dfed5465397569c2855eb0c6d564 updates API. Users will be able to configure NAT Gateway's namespace. 
  - [x] I need to check if exposing NAT Gateway's namespace is allowed and possible.
  - Will push for this(BYO Natgateway) once other Azure services have been migrated to ASO.
- Commit 7d67bd07ee4618b8a0b7e9b5442e1274c6c2edb9 generally updates NAT Gateways service to ASO. 

----

- [ ] cherry-pick candidate

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
NAT Gateway service will use the ASO framework
```


api/v1beta1/types.go
```diff
@@ -242,7 +242,8 @@ type NatGateway struct {
 
 // NatGatewayClassSpec defines a NAT gateway class specification.
 type NatGatewayClassSpec struct {
-	Name string `json:"name"`
+	Name      string `json:"name"`
+	Namespace string `json:"namespace"`
```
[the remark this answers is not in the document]
nawazkh: Good thought. 
I am happy to push it(BYO NAT Gateway) out to something we can do after we migrate all services to ASO, to minimize the churn. Reverting the commit.

azure/scope/cluster.go
```diff
@@ -326,17 +328,22 @@ func (s *ClusterScope) RouteTableSpecs() []azure.ResourceSpecGetter {
...
 				natGatewaySet[subnet.NatGateway.Name] = struct{}{} // empty struct to represent hash set
 				natGateways = append(natGateways, &natgateways.NatGatewaySpec{
-					Name:           subnet.NatGateway.Name,
+					Name:      subnet.NatGateway.Name,
+					Namespace: subnet.NatGateway.Namespace,
+					Owner: genruntime.KnownResourceReference{
+						// TODO: add ability to specify owner and thus bring your own NAT gateway
+						Name: s.ResourceGroup(),
+					},
 					ResourceGroup:  s.ResourceGroup(),
```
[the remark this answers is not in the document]
nawazkh: 👍🏼
Updating.

azure/scope/cluster_test.go
```diff
@@ -63,6 +65,25 @@ func specArrayToString(specs []azure.ResourceSpecGetter) string {
 	return sb.String()
 }
 
+func specToStringForAsoResource[T genruntime.MetaObject](spec azure.ASOResourceSpecGetter[T]) string {
```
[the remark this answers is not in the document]
nawazkh: 👍🏼 
Updating.

azure/services/aso/aso.go
```diff
@@ -117,7 +117,7 @@ func (r *reconciler[T]) CreateOrUpdateResource(ctx context.Context, spec azure.A
 				// handle user-managed resources that already exist in Azure and
 				// should not be reconciled by ASO while ensuring they're still
 				// represented in ASO.
-				log.V(2).Info("resource not found in Azure and \"skip\" reconcile-policy set, adopting")
+				log.V(2).Info("resource not found in Azure and \"skip\" reconcile-policy already set, adopting")
```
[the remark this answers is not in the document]
nawazkh: The idea behind adding `already` to the log was to indicate that the ASO resource _already_ has the reconcile policy set to `skip` and that CAPZ will be _adopting_ it going ahead.
In any case, I am happy to revert it.

azure/services/natgateways/natgateways_test.go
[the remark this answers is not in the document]
nawazkh: While implementing a test for `PostCreateOrUpdateResourceHook` I realized that I could instead have the function unit tested at the cluster level since 
- the function is defined at cluster level
- writing the same test at NAT Gateway scope looked too complex. I was trying to recreate the mocks already present at cluster scope.

I can give it another shot, but I have a test for PostCreateResourceHook's function written here
https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/4059/files#diff-bae329736fc788cebbe920f1680bfd39eb52e6f1bdaaa0a620c60037d3840462R1135-R1201, What do your thoughts about that?

azure/services/aso/aso.go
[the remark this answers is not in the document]
nojnhuh: I guess that's how I interpret the existing log message currently, but this isn't a hill I'm willing to die on.

azure/services/natgateways/spec.go
```diff
@@ -39,61 +40,66 @@ type NatGatewaySpec struct {
...
-func hasPublicIP(natGateway armnetwork.NatGateway, publicIPName string) bool {
-	for _, publicIP := range natGateway.Properties.PublicIPAddresses {
-		if publicIP != nil && publicIP.ID != nil {
-			resource, err := azureutil.ParseResourceID(*publicIP.ID)
+// WasManaged implements azure.ASOResourceSpecGetter.
+func (s *NatGatewaySpec) WasManaged(resource *asonetworkv1.NatGateway) bool {
+	// TODO: Earlier implementation of IsManaged(with ASO's framework is morphed into WasManaged) was checking for s.Scope.IsVnetManaged()
+	// we would want to update this to check for the same once we have the support for VnetManaged in ASO
+	// we return true for now so that CAPZ can automatically "manage" the nat gateway
+	return true
```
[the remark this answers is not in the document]
nojnhuh: We can do this without vnets being ASO yet. It might look something like:
1. Define a `IsVnetManaged` field on `NatGatewaySpec`
2. Set that field in scope/cluster.go based on the scope's `IsVnetManaged()` method
3. Return its value here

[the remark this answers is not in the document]
nawazkh: Updating.

api/v1beta1/types.go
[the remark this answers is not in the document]
nawazkh: Reverted via ddc33ab24403089e3062a1e6c4b2bcc4ff2e55c4

azure/scope/cluster.go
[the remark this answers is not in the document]
nawazkh: Updated via ea6139068fa648c7987aa4df2b32ce4320f54084

azure/scope/cluster_test.go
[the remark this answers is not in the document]
nawazkh: Updated via ea6139068fa648c7987aa4df2b32ce4320f54084

azure/services/aso/aso.go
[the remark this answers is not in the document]
nawazkh: Reverted via ddc33ab24403089e3062a1e6c4b2bcc4ff2e55c4 and 724fb514d3df0cbff831a775091325b6bfd2fcd8

azure/services/natgateways/spec.go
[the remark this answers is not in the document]
nawazkh: Updated via ea6139068fa648c7987aa4df2b32ce4320f54084

azure/services/natgateways/natgateways_test.go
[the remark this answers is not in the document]
nawazkh: Also added a unit test testing invoking `postCreateOrUpdateResourceHook()` func via 01632acedf662575242e1ba467ab5dc41c3ac121

nawazkh:
Squashing all and running a test locally.

azure/scope/cluster_test.go
```diff
@@ -63,6 +65,17 @@ func specArrayToString(specs []azure.ResourceSpecGetter) string {
 	return sb.String()
 }
 
+func specArrayToStringForAsoResource[T genruntime.MetaObject](specs []azure.ASOResourceSpecGetter[T]) string {
```
nojnhuh, line 68: Likewise, I don't think we need this function if we modify the existing `specArrayToString` to look like:
```suggestion
func specArrayToStringForAsoResource[T any](specs []T) string {
```

nawazkh: Updated and force pushed.

azure/services/natgateways/spec.go
```diff
@@ -19,81 +19,87 @@ package natgateways
...
+// WasManaged implements azure.ASOResourceSpecGetter.
+func (s *NatGatewaySpec) WasManaged(resource *asonetworkv1.NatGateway) bool {
+	return s.IsVnetManaged
+}
+
+// hasPublicIP returns true if the public IP Name is already attached to the NAT gateway.
+func hasPublicIP(natGateway *asonetworkv1.NatGateway, publicIPName string) bool {
+	for _, publicIP := range natGateway.Status.PublicIpAddresses {
+		if publicIP.Id != nil {
+			resource, err := azureutil.ParseResourceID(*publicIP.Id)
```
nojnhuh, line 102: I'm wondering if it might be preferable here to use `Spec` instead of `Status` (I know I said otherwise earlier). If the IP gets removed from the spec for some reason and ASO starts removing it, `Status` can be out of date and still have it, and this function would return `true`. Then we'd have to wait for ASO to converge, pick up that change, then notice that it's gone from the status to add it back to the spec. I think if we check the spec directly, we would basically skip waiting for ASO to converge on the removed IP before adding it back.

I'm not sure if that exact scenario is possible here, but I think for the sake of trying to establish some rules of thumb while we convert the other services it's worth thinking about.

```suggestion
	for _, publicIP := range natGateway.Spec.PublicIpAddresses {
		if publicIP.Reference != nil {
			resource, err := azureutil.ParseResourceID(publicIP.Reference.ARMID)
```

nawazkh: I think sticking to ASOResource’s `status` is safer than its `spec`. 

I agree that there will be a delay in propagating the change from `spec` to `status`; but it is _safer_, since CAPZ is relying on ASO to perform the background jobs of whatever it needs to do to reflect the changes in the the`spec` of the resource -> reconcile them -> and then show it in the resource’s `status`.  While grabbing _the detail(s)_ from the `spec` solves the delay issue, it can push the resource in weird scenarios where CAPZ tries to perform duplicate `PATCH` onto the K8s resource.

azure/scope/cluster.go
```diff
@@ -339,6 +341,9 @@ func (s *ClusterScope) NatGatewaySpecs() []azure.ResourceSpecGetter {
 						Name: subnet.NatGateway.NatGatewayIP.Name,
 					},
 					AdditionalTags: s.AdditionalTags(),
+					// This is a bit of a hack, but we need to know if the VNet is managed to know if
+					// this NAT Gateway was-managed or not.
+					IsVnetManaged: s.IsVnetManaged(),
```
willie-yao, line 346: For my own understanding: Why is this considered a hack? It seems like the `NatGatewaySpec` has a field for `IsVnetManaged` and follows the convention for the other fields above.

willie-yao (commented):
Great work @nawazkh! I just had a question and one suggestion. Would you be able to add a unit test for the new `Parameters()` function in `natgateways.go`? I think a basic unit test would help since there's some new logic there.

azure/scope/cluster.go
nojnhuh: I'd argue this is not really a hack.

azure/services/natgateways/spec.go
nojnhuh: I'm not sure I understand why using spec could be less safe here or produce more duplicate work than using status. If we check the status, then there's a chance that the spec is already up to date in which case we might try to issue an update even though the spec is already up to date.

azure/scope/cluster.go
nawazkh: I see that the word `hack` and the way I used it is creating controversy, happy to remove it. 
The context on why I sort of called it a hack is that we created a "new field" to save the answer if this NATGateway was managed. Now that I think of it, it is doing what it is supposed to.

The comment should mostly say 
`// we need to know if the VNet is managed to decide if this NAT Gateway was managed or not.`

nawazkh:
> Great work @nawazkh! I just had a question and one suggestion. Would you be able to add a unit test for the new `Parameters()` function in `natgateways.go`? I think a basic unit test would help since there's some new logic there.

Will do. :)

azure/services/natgateways/spec.go
nojnhuh: I think the conclusion we came to offline with @jackfrancis and @willie-yao here is that this check (and similar ones) shouldn't even be necessary at all in the context of ASO. Each `Parameters()` implementation should more-or-less blindly construct the `spec` of an ASO resource and leverage the higher-level `CreateOrUpdateResource()` defined in the `aso` package to skip sending updates to ASO if they aren't required.

Before with the SDK, that "skip update if not modified" logic had to be defined in each service because any non-nil value returned by `Parameters()` would always result in an Azure API call.

nawazkh: Thank you for summarizing @nojnhuh !

azure/scope/cluster.go
nawazkh: Updated via 8dc5cc3b4a4fee169b45efee1bc16fe34f91457e

azure/services/natgateways/spec.go
nawazkh: This should be now updated via c4dbab91d193e43e4d261c67f8414e3147f8cfa7

azure/services/natgateways/spec.go
```diff
@@ -19,88 +19,71 @@ package natgateways
...
+}
 
-	natGatewayToCreate := armnetwork.NatGateway{
-		Name:     ptr.To(s.Name),
+// Parameters returns the parameters for the NAT gateway. It implements azure.ASOResourceSpecGetter.
+func (s *NatGatewaySpec) Parameters(ctx context.Context, existingResource *asonetworkv1.NatGateway) (params *asonetworkv1.NatGateway, err error) {
+	// existingResource will be an empty asonetworkv1.NatGateway{} if the aso resource doesn't exist.
+	// if existingResource is not nil, we overwrite the spec with desired values, leaving the status and meta-data as is.
+	// Note that, an aso resource with matching name and namespace can exist but will not get to this point since it will
+	// not be owned by this azure cluster.
```
nawazkh, line 58: Let me know if this comment is appropriate or if I should reduce it. The aim is to clarify to the reader the reason we "overwrite" the spec.

nawazkh:
adding a unit test to Parameters func
/hold

azure/services/natgateways/spec.go
```diff
@@ -19,88 +19,87 @@ package natgateways
...
+		natGateway = existingResource
+	}
 
-// Parameters returns the parameters for the NAT gateway.
-func (s *NatGatewaySpec) Parameters(ctx context.Context, existing interface{}) (params interface{}, err error) {
-	if existing != nil {
-		existingNatGateway, ok := existing.(armnetwork.NatGateway)
-		if !ok {
-			return nil, errors.Errorf("%T is not an armnetwork.NatGateway", existing)
+	if natGateway.Spec.Owner == nil {
```
nawazkh, line 68: I can safely get rid of the `if` statements since 
- CAPZ, being the owner, is building the spec.
- BYO NAT Gateway is not yet available, so having a check for existing statements is redundant.

nawazkh (commented):
some thoughts before final review.
/hold

azure/services/natgateways/spec.go
nawazkh: Fixed.

nawazkh:
@nojnhuh Going over https://gist.github.com/nojnhuh/a1e06bb6d98b0eacb51134480ea4acea I noticed that CAPZ is _not_ setting `azureName` for NatGateways as of current implementation. 

We can either default to `asonetworkv1.NatGateway.Spec.azureName=NatGatewaySpec.Name` or have a field defined on the API level asking the user to provide one. 
What do you think?

nojnhuh:
> @nojnhuh Going over https://gist.github.com/nojnhuh/a1e06bb6d98b0eacb51134480ea4acea I noticed that CAPZ is _not_ setting `azureName` for NatGateways as of current implementation.
> 
> We can either default to `asonetworkv1.NatGateway.Spec.azureName=NatGatewaySpec.Name` or have a field defined on the API level asking the user to provide one. What do you think?

Using `NatGatewaySpec.Name` is how it was set before, I'm guessing.

nawazkh:
> > @nojnhuh Going over [gist.github.com/nojnhuh/a1e06bb6d98b0eacb51134480ea4acea](https://gist.github.com/nojnhuh/a1e06bb6d98b0eacb51134480ea4acea) I noticed that CAPZ is _not_ setting `azureName` for NatGateways as of current implementation.
> > We can either default to `asonetworkv1.NatGateway.Spec.azureName=NatGatewaySpec.Name` or have a field defined on the API level asking the user to provide one. What do you think?
> 
> Using `NatGatewaySpec.Name` is how it was set before, I'm guessing.

Yep, looks like it. Ref: `Name: ptr.To(s.Name),` from the diff. I will default `azureName` to `s.Name` too.

azure/services/natgateways/spec_test.go
```diff
@@ -20,104 +20,129 @@ import (
...
+			existingSpec: existingNatGateway,
+			expect: func(g *WithT, existing *asonetworkv1.NatGateway, parameters *asonetworkv1.NatGateway) {
+				diff := cmp.Diff(existing, parameters)
+				g.Expect(diff).To(BeEmpty())
 			},
-			expectedError: "",
 		},
 	}
-	for _, tc := range testCases {
+	format.MaxLength = 10000
```
nojnhuh, line 139: Does this affect all tests in all packages that run after this is called?

azure/services/natgateways/spec_test.go
```diff
@@ -20,104 +20,129 @@ import (
...
 
-			result, err := tc.spec.Parameters(context.TODO(), tc.existing)
-			if tc.expectedError != "" {
-				g.Expect(err).To(HaveOccurred())
-				g.Expect(err).To(MatchError(tc.expectedError))
-			} else {
-				g.Expect(err).NotTo(HaveOccurred())
-			}
-			tc.expect(g, result)
+			result, _ := tc.spec.Parameters(context.TODO(), tc.existingSpec)
```
nojnhuh, line 146: This should be passed either a deep copy to mimic how Parameters is actually called if `existing` is being asserted on afterwards:
```suggestion
			result, _ := tc.spec.Parameters(context.TODO(), tc.existingSpec.DeepCopy())
```

Without a deep copy, Parameters ends up making `existing` and `parameters` point to the same struct which makes it impossible for the second test case to fail.

azure/services/natgateways/spec_test.go
```diff
@@ -20,104 +20,129 @@ import (
...
-			expect: func(g *WithT, result interface{}) {
-				g.Expect(result).To(BeAssignableToTypeOf(armnetwork.NatGateway{}))
-				g.Expect(result.(armnetwork.NatGateway).Location).To(Equal(ptr.To[string](fakeNatGatewaySpec.Location)))
-				g.Expect(result.(armnetwork.NatGateway).Name).To(Equal(ptr.To[string](fakeNatGatewaySpec.ResourceName())))
+			name:         "create a new NAT Gateway spec when existing aso resource is nil",
+			spec:         fakeNatGatewaySpec,
+			existingSpec: nil,
+			expect: func(g *WithT, existing *asonetworkv1.NatGateway, parameters *asonetworkv1.NatGateway) {
+				g.Expect(parameters).NotTo(BeNil())
+				g.Expect(parameters).To(BeAssignableToTypeOf(&asonetworkv1.NatGateway{}))
```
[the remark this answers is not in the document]
nawazkh: Had it for my sanity, removing :)

nawazkh:
I think the users will soon reach out to us asking to expose other fields on NAT Gateways than one we allow them to specify in CAPZ. 
That is because any value a user adds/edits of a CAPZ-maintained ASO resource gets overwritten with CAPZ's default value. For example, the user cannot add `idleTimeoutInMinutes` field to the CAPZ managed ASO NAT Gateway(which makes sense).

We should either add all the exposed fields of ASO's NAT Gateway CRD to CAPZ or allow users to edit the editable fields.

nojnhuh:
> I think the users will soon reach out to us asking to expose other fields on NAT Gateways than one we allow them to specify in CAPZ. That is because any value a user adds/edits of a CAPZ-maintained ASO resource gets overwritten with CAPZ's default value. For example, the user cannot add `idleTimeoutInMinutes` field to the CAPZ managed ASO NAT Gateway(which makes sense).
> 
> We should either add all the exposed fields of ASO's NAT Gateway CRD to CAPZ or allow users to edit the editable fields.

I think there's an argument to be made that if users need to customize resources beyond what CAPZ allows, then they should be going the BYO route for those resources.

nojnhuh:
> > I think the users will soon reach out to us asking to expose other fields on NAT Gateways than one we allow them to specify in CAPZ. That is because any value a user adds/edits of a CAPZ-maintained ASO resource gets overwritten with CAPZ's default value. For example, the user cannot add `idleTimeoutInMinutes` field to the CAPZ managed ASO NAT Gateway(which makes sense).
> > We should either add all the exposed fields of ASO's NAT Gateway CRD to CAPZ or allow users to edit the editable fields.
> 
> I think there's an argument to be made that if users need to customize resources beyond what CAPZ allows, then they should be going the BYO route for those resources.

@CecileRobertMichon @jackfrancis Do you think it's a bug or a feature that fields never set by CAPZ on resources otherwise managed by CAPZ can be modified durably by users out-of-band with CAPZ?

nawazkh:
> > > I think the users will soon reach out to us asking to expose other fields on NAT Gateways than the one we allow them to specify in CAPZ. That is because any value a user adds/edits of a CAPZ-maintained ASO resource gets overwritten with CAPZ's default value. For example, the user cannot add `idleTimeoutInMinutes` field to the CAPZ managed ASO NAT Gateway(which makes sense).
> > > We should either add all the exposed fields of ASO's NAT Gateway CRD to CAPZ or allow users to edit the editable fields.
> > 
> > 
> > I think there's an argument to be made that if users need to customize resources beyond what CAPZ allows, then they should be going the BYO route for those resources.
> 
> @CecileRobertMichon @jackfrancis Do you think it's a bug or a feature that fields never set by CAPZ on resources otherwise managed by CAPZ can be modified durably by users out-of-band with CAPZ?

Adding my two cents. 
I can not classify it either, but in IMO, at least from NAT Gateways' point of view, I should be able to add the `idleTimeoutInMinutes` field to the ASO NAT Gateway resource created and managed by CAPZ. 
As a user, I know that CAPZ does not expose the `idleTimeoutInMinutes` field and I am not trying to edit an existing field, I am trying to add a new one that was not exposed by CAPZ. 

But allowing to *` add an unexposed field`* brings a new question, what if such an addition creates unwanted behavior and mutation within a service..?
Adding `idleTimeoutInMinutes` in NAT Gateway does not create any mutation per se, but it is a little early to comment about other services.

jackfrancis:
> > > I think the users will soon reach out to us asking to expose other fields on NAT Gateways than one we allow them to specify in CAPZ. That is because any value a user adds/edits of a CAPZ-maintained ASO resource gets overwritten with CAPZ's default value. For example, the user cannot add `idleTimeoutInMinutes` field to the CAPZ managed ASO NAT Gateway(which makes sense).
> > > We should either add all the exposed fields of ASO's NAT Gateway CRD to CAPZ or allow users to edit the editable fields.
> > 
> > 
> > I think there's an argument to be made that if users need to customize resources beyond what CAPZ allows, then they should be going the BYO route for those resources.
> 
> @CecileRobertMichon @jackfrancis Do you think it's a bug or a feature that fields never set by CAPZ on resources otherwise managed by CAPZ can be modified durably by users out-of-band with CAPZ?

I think it should read "can be modified durably by users out-of-band with kubectl", right? That's neither a feature nor a bug, but a practical architectural decision until we protect those ASO resources behind something like a serviceaccount.

nawazkh:
> Do you think it's a bug or a feature that fields never set by CAPZ on resources otherwise managed by CAPZ can be modified durably by users out-of-band with CAPZ?

In addition to what Jack said and Cross-posting the summary from a discussion we had async, below are some more observations made:
- It's not a bug per se that a user is able to set a field out-of-band but not a "supported" feature.
- Current SDK allows setting Azure resource fields out of band already.
- We should not make any user-facing behavior changes as part of migration to ASO.
- We should wait for actual user demand before exposing new features.

nojnhuh:
@jackfrancis 
> I think it should read "can be modified durably by users out-of-band with kubectl", right? That's neither a feature nor a bug, but a practical architectural decision until we protect those ASO resources behind something like a serviceaccount.

I mean even today by modifying the Azure resources with `az` or in the portal or whatever out-of-band with CAPZ and without ASO in the picture at all. This becomes more of an issue when we start using ASO though since ASO's continuous reconciliation will start to overwrite changes made by users outside of CAPZ/ASO that currently do not get overwritten.

I guess my main question here is in @nawazkh's scenario of a NAT gateway managed by CAPZ without `idleTimeoutInMinutes ` set, if a user edits the ASO NatGateway to define that themselves, should CAPZ overwrite that change? What if the user defined the field from the portal?

CecileRobertMichon:
> This becomes more of an issue when we start using ASO though since ASO's continuous reconciliation will start to overwrite changes made by users outside of CAPZ/ASO that currently do not get overwritten.

Could you please expand on this part specifically? Why is ASO overwriting changes made outside of CAPZ/Azure more than the CAPZ reconciler was before?

> if a user edits the ASO NatGateway to define that themselves, should CAPZ overwrite that change

Is CAPZ defaulting the field? Is it because the field required? 

> What if the user defined the field from the portal?

IMO behavior should be the same (if CAPZ is hardcoding and overwriting a field already it would overwrite portal changes today, ASO should not change that behavior). If CAPZ is not overwriting changes from the portal, then it should not overwrite changes from ASO either.

nojnhuh:
> > This becomes more of an issue when we start using ASO though since ASO's continuous reconciliation will start to overwrite changes made by users outside of CAPZ/ASO that currently do not get overwritten.
> 
> Could you please expand on this part specifically? Why is ASO overwriting changes made outside of CAPZ/Azure more than the CAPZ reconciler was before?

Several resources types created by CAPZ are never modified by CAPZ after they are created, e.g. bastionhosts: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/1a70b0108b4049f08e9918a56123233e6a586f91/azure/services/bastionhosts/spec.go#L67-L73

Currently, users are free to change any parameters with `az` or the portal after a bastion host is created and CAPZ will never overwrite those changes. If ASO were reconciling the bastion, then those changes would be overwritten if they overlap with CAPZ's initial opinion of what the resource should look like.

> > if a user edits the ASO NatGateway to define that themselves, should CAPZ overwrite that change
> 
> Is CAPZ defaulting the field? Is it because the field required?

I think CAPZ essentially has no knowledge of the field in this case. If CAPZ reconstructs the entire `spec` of the ASO NatGateway in each reconciliation loop though, implicitly zeroing out the value set by the user could overwrite the user's value.

> > What if the user defined the field from the portal?
> 
> IMO behavior should be the same (if CAPZ is hardcoding and overwriting a field already it would overwrite portal changes today, ASO should not change that behavior). If CAPZ is not overwriting changes from the portal, then it should not overwrite changes from ASO either.

That makes sense to me as long as we accept that changes from the portal that do not get overwritten today may then eventually get overwritten by ASO. I don't think there's a good way around that.

CecileRobertMichon:
> I think CAPZ essentially has no knowledge of the field in this case. If CAPZ reconstructs the entire spec of the ASO NatGateway in each reconciliation loop though, implicitly zeroing out the value set by the user could overwrite the user's value.

I thought the patch helper would handle this and only patch the changed values? CAPI patch helper uses the `MergeFrom` merge option (https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/client/patch.go#L150) and CAPZ passes the existing ASO object to the patch helper (here https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/services/aso/aso.go#L215).

nojnhuh:
> I thought the patch helper would handle this and only patch the changed values? CAPI patch helper uses the `MergeFrom` merge option (https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/client/patch.go#L150) and CAPZ passes the existing ASO object to the patch helper (here https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/services/aso/aso.go#L215).

It looks like it does not based on a quick test I ran.

nawazkh:
> If CAPZ reconstructs the entire spec of the ASO NatGateway in each reconciliation loop though, implicitly zeroing out the value set by the user could overwrite the user's value.

To add more clarity, there are two types of scenarios in discussion here. 
1. The spec fields which CAPZ has an opinion about (like `asonetworkv1.NatGateway.Spec.Location`). 
	- Ideally, we want the user to "try" editing the editable CAPZ opinionated fields by editing CAPZ CRD. 
	- And any changes made via Azure portal or _az cli_ would (and should) be overwritten by the CAPZ controller since CAPZ has a pre-defined opinion about them. (I am only talking about the spec fields here and not the annotations, labels)
2. The spec fields where CAPZ has no idea that they exist (like `asonetworkv1.NatGateway.Spec.IdleTimeoutInMinutes`).
	- Since CAPZ does not consider it to be important enough to have an opinion about it (read as having defaults or field in the CRD), the user should be allowed to edit them, and CAPZ should reflect those changes in its CRD.

The current implementation that I have put in for NAT Gateway spec supports scenario 2. Refer: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/4059/files?diff=split#diff-8592ed50d123a48d0e32068cf51a94a4b6d57588b0526d982c1c4f2723be085cR56-R59

willie-yao (approved):
/lgtm

Makefile
```diff
@@ -158,7 +158,7 @@ WEBHOOK_ROOT ?= $(MANIFEST_ROOT)/webhook
 RBAC_ROOT ?= $(MANIFEST_ROOT)/rbac
 ASO_CRDS_PATH := $(MANIFEST_ROOT)/aso/crds.yaml
 ASO_VERSION := v2.3.0
-ASO_CRDS := resourcegroups.resources.azure.com
+ASO_CRDS := resourcegroups.resources.azure.com natgateways.network.azure.com
```
CecileRobertMichon, line 161: @nojnhuh is this list going to keep growing everytime we add a new CRD? would it make sense to track it in some other way once the list ends up being more than a few?

CecileRobertMichon (commented):
/lgtm

great to see all that deleted code and added tests!

Makefile
nojnhuh: Yes, this will eventually get unwieldy here. Didn't get around to finding a better mechanism yet.

CecileRobertMichon: not blocking for this PR, we can circle back once the list gets big enough

nojnhuh (commented):
/lgtm
/approve

PR #4008, opened 2023-09-18
Disable local accounts for aks aad clusters

LochanRn:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
Support to disable local accounts for aks aad clusters

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #2129

**Special notes for your reviewer**:

As access for admin kubeconfig will be disabled, the user must ensure to add the service principle to the admin group configured for enabling AAD.

- [ ] cherry-pick candidate

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [X] adds unit tests

**Release note**:

```release-note
Support disabling of local accounts for AKS clusters with Azure Active Directory enabled. Disables the back door to get admin kubeconfig for AAD based clusters. The user must ensure to add the service principle to admin groups in azure AAD for capz to be able to deploy the target cluster..
```

Issue #2625, opened 2022-09-02
Expose AKS preview features


nojnhuh:
***Goals***
1. Determine a method for making AKS preview features consumable through CAPZ like GA AKS features

***Non-Goals/Future Work***
1. Implement every AKS preview feature
1. Guarantee the stability of AKS preview features in CAPZ beyond what AKS itself supports
1. Determine how to handle preview features of other APIs used for self-managed clusters

**User Story**

As a user I would like to configure AKS managed clusters with preview features to take advantage of the latest and greatest functionality available in AKS.

**Detailed Description**

Some AKS features are not available from Azure's stable API and can only be used from separate preview APIs. Since CAPZ currently only uses the stable AKS APIs, those preview features are not available and must be configured through some other means which introduces friction for users looking to use Cluster API for all their cluster lifecycle management. Exposing AKS preview features in CAPZ would eliminate that friction.

/kind proposal
/area managedclusters

mtougeron:
As a user I would love this functionality. Not being able to test preview features is a limitation that I'm bumping up against (e.g., BYO CNI).

Two high-level thoughts that I have on this:

1) How would we address api version compatibility? For example, if while a feature goes through the preview process there is a change to the interface, would we have to iterate on the apiVersion of the `AzureManagedControlPlane`? The contract we provide to the users of CAPZ would have to be very explicit and clear to avoid confusion.

2) With the lower SLA of a preview feature how would that impact the testing matrix that CAPZ would perform again the preview features? Would it significantly increase the load on the maintainers to keep the tests/functionality working if the preview functionality changes behind the scenes before an updated go sdk has been released.

mboersma:
> Since CAPZ currently only uses the stable AKS APIs

Probably stupid question, but is it not possible to use the preview AKS API only in the managed cluster code? Or can we not vendor in both at the same time?

nojnhuh:
The preview APIs exist side-by-side (at least in the same Go module) with the stable ones in the SDK from what I can see: https://github.com/Azure/azure-sdk-for-go/tree/ca4d22fbbacf9ddc32b42a14113b13ea3feb9ccd/services/preview. So it should be possible to do something pretty granular like:

```go
import (
    stable "github.com/Azure/azure-sdk-for-go/services/..."
    preview "github.com/Azure/azure-sdk-for-go/services/preview/..."
)

if r.previewFeatureDefined() {
    preview.CreateResource(r)
} else {
[...]

PR #4069, opened 2023-10-03
convert managedclusters and agentpools to ASO

### LINKED ISSUE #3529: ASO service: managedclusters and agentpools
Convert the virtualnetworks and subnets services to use ASO. I'm fairly confident these two will be co-dependent.

Very relevant ASO issue: https://github.com/Azure/azure-service-operator/issues/2791

```[tasklist]
### Dependencies
- [ ] #3527 
```

#3528 may be a dependency

nojnhuh:
**What type of PR is this?**
/kind cleanup

**What this PR does / why we need it**: This PR updates the managedclusters and agentpools services to use ASO.

Fixes #3529

- [ ] cherry-pick candidate

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [X] adds unit tests

/hold for squash, ASO v2.4 (#4248)

**Release note**:

```release-note
ASO now manages managed clusters and agent pools. Custom headers no longer have any effect.
```


api/v1beta1/azuremanagedcluster_webhook.go
```diff
@@ -57,24 +52,6 @@ func (r *AzureManagedCluster) ValidateCreate() (admission.Warnings, error) {
 
 // ValidateUpdate implements webhook.Validator so a webhook will be registered for the type.
 func (r *AzureManagedCluster) ValidateUpdate(oldRaw runtime.Object) (admission.Warnings, error) {
-	old := oldRaw.(*AzureManagedCluster)
-	var allErrs field.ErrorList
-
-	// custom headers are immutable
```
nojnhuh, line 63: ASO does not support adding custom headers to requests, so all of that is now gone.

azure/services/agentpools/spec.go
```diff
@@ -42,28 +41,34 @@ type KubeletConfig struct {
 	// CPUCfsQuotaPeriod - Sets CPU CFS quota period value.
 	CPUCfsQuotaPeriod *string
 	// ImageGcHighThreshold - The percent of disk usage after which image garbage collection is always run.
-	ImageGcHighThreshold *int32
+	ImageGcHighThreshold *int
```
nojnhuh, line 44: Lots of fields that were `*int32` in the SDK are now `*int` in ASO (or similar for other types). I elected not to change the API types but to change them everywhere else to match ASO.

azure/services/agentpools/spec.go
```diff
@@ -152,243 +154,228 @@ type AgentPoolSpec struct {
...
-		if !ok {
-			return nil, errors.Errorf("%T is not an armcontainerservice.AgentPool", existing)
-		}
-
 		// agent pool already exists
-		ps := *existingPool.Properties.ProvisioningState
-		if ps != string(infrav1.Canceled) && ps != string(infrav1.Failed) && ps != string(infrav1.Succeeded) {
-			msg := fmt.Sprintf("Unable to update existing agent pool in non terminal state. Agent pool must be in one of the following provisioning states: Canceled, Failed, or Succeeded. Actual state: %s", ps)
-			return nil, azure.WithTransientError(errors.New(msg), 20*time.Second)
-		}
```
nojnhuh, line 192: ASO already blocks based on provisioning state, so we don't need to block updating the ASO resource on the same thing.

azure/services/aso/aso.go
```diff
@@ -122,20 +122,18 @@ func (r *reconciler[T]) CreateOrUpdateResource(ctx context.Context, spec azure.A
...
 					ResourceGroup: existing.GetNamespace(),
 					Name:          existing.GetName(),
 				})
+				return zero, azure.WithTransientError(err, requeueInterval)
 			default:
-				readyErr = fmt.Errorf("resource is not Ready: %s", conds[i].Message)
-			}
-
-			if readyErr != nil {
+				err := fmt.Errorf("resource is not Ready: %s", conds[i].Message)
```
nojnhuh, line 132: The changes around here are because I noticed that before, if an ASO resource got stuck in a failed state, attempting to fix it by updating the CAPZ spec would never work because we were always returning an error here without calling `Parameters` and updating the ASO resource. These changes defer the equivalent check to after `Parameters` is called and generates no diff.

azure/services/aso/tags.go
```diff
@@ -47,6 +50,14 @@ func reconcileTags[T genruntime.MetaObject](t TagsGetterSetter[T], existing T, r
 		}
 
 		existingTags = t.GetActualTags(existing)
+		if !reflect.DeepEqual(t.GetDesiredTags(existing), existingTags) &&
```
nojnhuh, line 53: I had noticed a race condition while running the tags e2e test where if the tags have not converged, then CAPZ could lose track of some tags that have been deleted from `spec.additionalTags` and treat them like tags that existed outside of `spec.additionalTags` and keep them around without ever deleting them. Waiting for the tags to converge on the ASO resource fixes that issue AFAICT.

Overall, this makes tags reconciliation work more similarly to the more synchronous method used before when calling the tags APIs directly.

azure/services/managedclusters/managedclusters.go
```diff
@@ -37,113 +31,58 @@ const kubeletIdentityKey = "kubeletidentity"
...
-	defer cancel()
-
-	managedClusterSpec := s.Scope.ManagedClusterSpec()
-	if managedClusterSpec == nil {
-		return nil
+	// This field gets populated by AKS when not set by the user. Persist AKS's value so for future diffs,
+	// the "before" reflects the correct value.
+	if id := managedCluster.Status.IdentityProfile[kubeletIdentityKey]; id.ResourceId != nil {
+		scope.SetKubeletIdentity(*id.ResourceId)
 	}
```
nojnhuh, line 80: This may not be necessary for us to do anymore, but I will check.

test/e2e/aks_tags.go
```diff
@@ -136,14 +131,16 @@ func AKSAdditionalTagsSpec(ctx context.Context, inputGetter func() AKSAdditional
...
 
-		By("Restoring initial tags for control plane")
-		expectedTags = initialTags
-		Eventually(func(g Gomega) {
-			g.Expect(mgmtClient.Get(ctx, client.ObjectKeyFromObject(infraControlPlane), infraControlPlane)).To(Succeed())
-			infraControlPlane.Spec.AdditionalTags = expectedTags
-			g.Expect(mgmtClient.Update(ctx, infraControlPlane)).To(Succeed())
-		}, inputGetter().WaitForUpdate...).Should(Succeed())
-		Eventually(checkTags, input.WaitForUpdate...).Should(Succeed())
+		if initialTags != nil {
```
nojnhuh, line 134: ASO does not support deleting all tags: https://github.com/Azure/azure-service-operator/issues/3367

azure/services/agentpools/spec.go
Jont828: I think that's fair, changing the API type would be quite inconvenient as it would require adding new conversion functions for each `int32`.

nojnhuh: I asked in Slack if changing these types would be too disruptive and haven't gotten any feedback yet indicating they would be, so I'm still considering changing these to match ASO so CAPZ doesn't need to do any conversion.

api/v1beta1/azuremanagedcluster_webhook.go
dtzar: That is a bummer, we discussed this on the community call as well.  We will need to update [our documentation](https://capz.sigs.k8s.io/topics/managedcluster#enable-aks-features-with-custom-headers---aks-custom-headers) and put a release note to make sure people know it is no longer supported.

nojnhuh: Good call, I'll update the docs.

@CecileRobertMichon Do you know if there's a decent way to format release notes longer than one line? Or perhaps add multiple release notes for a single PR?

azure/converters/managedagentpool_test.go
```diff
@@ -19,70 +19,75 @@ package converters
 import (
 	"testing"
 
-	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/containerservice/armcontainerservice/v4"
+	asocontainerservicev1 "github.com/Azure/azure-service-operator/v2/api/containerservice/v1api20230201"
```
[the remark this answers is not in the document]
nojnhuh: I think this would be best addressed separately as part of #2625.

azure/services/managedclusters/managedclusters.go
```diff
@@ -62,106 +61,58 @@ type ManagedClusterScope interface {
...
 
-	ctx, cancel := context.WithTimeout(ctx, reconciler.DefaultAzureServiceReconcileTimeout)
-	defer cancel()
-
-	managedClusterSpec := s.Scope.ManagedClusterSpec()
-	if managedClusterSpec == nil {
-		return nil
+	// Update kubeconfig data
+	// Always fetch credentials in case of rotation
+	adminKubeConfigData, userKubeConfigData, err := reconcileKubeconfig(ctx, scope, managedCluster.Namespace)
```
nojnhuh, line 98: @LochanRn If you could please test this PR to make sure it doesn't break your changes from #4008 I would very much appreciate it!

I tested the scenario where there is no AADProfile and where there is a managed AADProfile without DisableLocalAccounts and both of those seemed to produce valid kubeconfigs.

For the case where there is an AADProfile with DisableLocalAccounts, the secrets looked fine but if you could confirm that they actually work that would be great!

LochanRn: Sure will test it and update here !

LochanRn: @nojnhuh I was able to test disableLocalAccounts functionality and it is not breaking the change.

Makefile
```diff
@@ -308,7 +308,7 @@ create-management-cluster: $(KUSTOMIZE) $(ENVSUBST) $(KUBECTL) $(KIND) ## Create
 
 	# Deploy CAPZ
 	$(KIND) load docker-image $(CONTROLLER_IMG)-$(ARCH):$(TAG) --name=$(KIND_CLUSTER_NAME)
-	$(KUSTOMIZE) build config/default | $(ENVSUBST) | $(KUBECTL) apply -f -
+	$(KUSTOMIZE) build config/default | $(ENVSUBST) | $(KUBECTL) apply -f - --server-side=true
```
willie-yao, line 311: What does this change accomplish? Mostly for my own understanding

api/v1beta1/azuremanagedcluster_webhook.go
willie-yao: Should we add a comment here as a TODO to add it back when it is supported?

api/v1beta1/azuremanagedcontrolplane_webhook.go
```diff
@@ -76,10 +78,20 @@ func (mw *azureManagedControlPlaneWebhook) Default(ctx context.Context, obj runt
...
 		m.Spec.NetworkPlugin = &networkPlugin
 	}
+
 	if m.Spec.LoadBalancerSKU == nil {
-		loadBalancerSKU := "Standard"
+		loadBalancerSKU := "standard"
 		m.Spec.LoadBalancerSKU = &loadBalancerSKU
 	}
+	lbSku := ptr.Deref(m.Spec.LoadBalancerSKU, "")
+	if lbSku == "Standard" {
```
willie-yao, line 87: Does this need to be uncapitalized? I noticed standard and basic is lowercase in other places, such as when `loadBalancerSKU` is set on line 83.

api/v1beta1/azuremanagedmachinepool_webhook.go
```diff
@@ -82,10 +78,6 @@ func (mw *azureManagedMachinePoolWebhook) Default(ctx context.Context, obj runti
 		m.Spec.OSType = ptr.To(DefaultOSType)
 	}
 
-	if ptr.Deref(m.Spec.ScaleSetPriority, "") == string(armcontainerservice.ScaleSetPrioritySpot) && m.Spec.SpotMaxPrice == nil {
-		m.Spec.SpotMaxPrice = ptr.To(resource.MustParse("-1"))
-	}
```
willie-yao, line 87: Do we no longer want to set the default max spot price?

azure/services/agentpools/spec.go
```diff
@@ -331,97 +254,37 @@ func (s *AgentPoolSpec) Parameters(ctx context.Context, existing interface{}) (p
...
-		if azureutil.IsAzureSystemNodeLabelKey(*aksNodeTaint) {
-			ret = append(ret, aksNodeTaint)
-		}
-	}
-	// Preserve nil-ness of capz
-	if capz == nil && len(ret) == 0 {
-		ret = nil
-	}
-	return ret
+// WasManaged implements azure.ASOResourceSpecGetter.
```
willie-yao, line 287: Should we clarify in the comment that it always returns true because we don't support byo agentpools?

willie-yao (commented):
Just a few comments on my end, mostly clarifying questions and nits! Great work on this 🚀

azure/services/managedclusters/spec.go
```diff
@@ -380,406 +386,226 @@ func (s *ManagedClusterSpec) Parameters(ctx context.Context, existing interface{
...
-		}
-	}
+func userKubeconfigSecretName(clusterName string) string {
+	return secret.Name(clusterName+"-user-aso", secret.Kubeconfig)
+}
 
-	if managedCluster.Properties.DisableLocalAccounts != nil {
-		clusterNormalized.Properties.DisableLocalAccounts = managedCluster.Properties.DisableLocalAccounts
-	}
+// WasManaged implements azure.ASOResourceSpecGetter.
```
willie-yao, line 586: Same comment as above regarding capz not supporting byo managed clusters

azure/services/agentpools/spec.go
```diff
@@ -331,110 +254,37 @@ func (s *AgentPoolSpec) Parameters(ctx context.Context, existing interface{}) (p
...
+		spec.Count = agentPool.Status.Count
 	}
 
 	return agentPool, nil
 }
 
-// mergeSystemNodeLabels appends any kubernetes.azure.com-prefixed labels from the AKS label set
-// into the local capz label set.
-func mergeSystemNodeLabels(capzLabels, aksLabels map[string]*string) map[string]*string {
-	// if both are nil, return nil for no change
```
nojnhuh, line 392: Until ASO v2.4 is available with https://github.com/Azure/azure-service-operator/pull/3407, the "set to nil to delete all labels" behavior enabled by #4122 will instead go back to the "setting to nil doesn't change" which it did before.

api/v1beta1/azuremanagedcluster_webhook.go
nojnhuh: It sounds like there's a good chance ASO will never support custom headers.

api/v1beta1/azuremanagedcontrolplane_webhook.go
nojnhuh: That's what this is accounting for, if I'm understanding your question right. Before the CAPZ API only accepted `Standard` but the ASO API only accepts `standard`. This change translates the incorrect `Standard` to `standard`.

api/v1beta1/azuremanagedmachinepool_webhook.go
nojnhuh: I don't think this provides any benefit since #4112 doesn't manifest without this anymore.

nojnhuh:
Blocking this PR on #4148, #4149, #4150 which are subsets of this PR.

nojnhuh:
> Blocking this PR on #4148, #4149, #4150 which are subsets of this PR.

These are all merged, so this PR is ready for review again.

azure/services/managedclusters/managedclusters.go
```diff
@@ -170,40 +121,58 @@ func (s *Service) IsManaged(ctx context.Context) (bool, error) {
...
 	}
+	adminKubeConfigData = asoSecret.Data[secret.KubeconfigDataName]
 	return adminKubeConfigData, userKubeConfigData, nil
 }
 
-// GetUserKubeconfigData gets user kubeconfig when aad is enabled for the aad clusters.
-func (s *Service) GetUserKubeconfigData(ctx context.Context, managedClusterSpec azure.ResourceSpecGetter) ([]byte, error) {
-	kubeConfigData, err := s.GetUserCredentials(ctx, managedClusterSpec.ResourceGroupName(), managedClusterSpec.ResourceName())
+// getUserKubeconfigData gets user kubeconfig when aad is enabled for the aad clusters.
+func getUserKubeconfigData(ctx context.Context, scope ManagedClusterScope, namespace string) ([]byte, error) {
```
nawazkh, line 156: Bringing to attention: `getUserKubeconfigData()` is no more exported. So is `getUserKubeConfigWithToken()`. 
I think this is localized to this package and hence not exported?
Curious why were they exported earlier?

nawazkh (approved):
Great work Jon! 🚢 
/lgtm

nojnhuh:
The e2e test is failing pretty consistently and I think I have a handle on why. It looks like the e2e tests are progressing past the `Upgrading the Kubernetes version of the cluster` step before the last agent pool is actually updated, which I think itself is a bug. Then when we update `spec.count` in the ASO ManagedClustersAgentPool while that agent pool is still upgrading, ASO doesn't seem to re-reconcile once the upgrade is finished, so the replica count stays stuck at 1 instead of increasing to 2. Eventually, the 30m timeout is hit and the test progresses to set the replica counts back to 1, the stuck agent pool already has one replica, so that succeeds quickly and the test keeps going and reports the timeout when it's finished.

I think the same bug in our tests may be related to #3955. Looking into that while checking to see if the ASO behavior is expected or not.

nojnhuh:
> The e2e test is failing pretty consistently and I think I have a handle on why. It looks like the e2e tests are progressing past the `Upgrading the Kubernetes version of the cluster` step before the last agent pool is actually updated, which I think itself is a bug. Then when we update `spec.count` in the ASO ManagedClustersAgentPool while that agent pool is still upgrading, ASO doesn't seem to re-reconcile once the upgrade is finished, so the replica count stays stuck at 1 instead of increasing to 2. Eventually, the 30m timeout is hit and the test progresses to set the replica counts back to 1, the stuck agent pool already has one replica, so that succeeds quickly and the test keeps going and reports the timeout when it's finished.

#4158 should workaround the ASO behavior.

azure/services/aso/aso.go
nojnhuh: Merged in #4148

azure/services/aso/tags.go
nojnhuh: Merged in #4149

azure/services/managedclusters/managedclusters.go
nojnhuh: It is not!

azure/services/managedclusters/managedclusters.go
nojnhuh: My guess was that these were only exported before because the other methods that were already here also were. I made them unexported here to help clarify that these are not and should not be used outside this package.

nojnhuh:
Editing to add the custom headers caveat. I'll refrain from documenting the "deleting all node labels/taints/tags" caveat for now since I think we have a fighting chance of resolving that before the v1.12 release with ASO v2.4.0 once that's released.

/release-note-edit
```release-note
ASO now manages managed clusters and agent pools. Custom headers no longer have any effect.
```

nojnhuh:
AKS test is passing again. @jackfrancis @nawazkh @willie-yao PTAL

api/v1beta1/azuremanagedcontrolplane_webhook.go
```diff
@@ -76,10 +78,20 @@ func (mw *azureManagedControlPlaneWebhook) Default(ctx context.Context, obj runt
 		networkPlugin := "azure"
 		m.Spec.NetworkPlugin = &networkPlugin
 	}
+
 	if m.Spec.LoadBalancerSKU == nil {
-		loadBalancerSKU := "Standard"
+		loadBalancerSKU := "standard"
```
jackfrancis, line 83: Seems like we have an opportunity to make a const out of these four LB strings.

azure/scope/managedcontrolplane.go
```diff
@@ -682,11 +677,6 @@ func (s *ManagedControlPlaneScope) SetUserKubeconfigData(kubeConfigData []byte)
 	s.userKubeConfigData = kubeConfigData
 }
 
-// SetKubeletIdentity sets the ID of the user-assigned identity for kubelet if not already set.
-func (s *ManagedControlPlaneScope) SetKubeletIdentity(id string) {
```
jackfrancis, line 686: Thanks, ASO!

controllers/azuremanagedcontrolplane_reconciler.go
```diff
@@ -45,10 +45,6 @@ type azureManagedControlPlaneService struct {
 
 // newAzureManagedControlPlaneReconciler populates all the services based on input scope.
 func newAzureManagedControlPlaneReconciler(scope *scope.ManagedControlPlaneScope) (*azureManagedControlPlaneService, error) {
-	managedClustersSvc, err := managedclusters.New(scope)
-	if err != nil {
-		return nil, err
-	}
```
jackfrancis, line 51: I'd kinda like the linter to send you some flare "nice job reducing a little bit of conditional complexity"

test/e2e/aks_tags.go
```diff
@@ -182,16 +182,11 @@ func AKSAdditionalTagsSpec(ctx context.Context, inputGetter func() AKSAdditional
 				}
 			}
 
-			Byf("Deleting all tags for machine pool %s", mp.Name)
```
jackfrancis, line 185: Remind me what we were doing before w/ respect to delete all tags via nil data, and why we can't do that anymore?

nojnhuh: Before, CAPZ had a bit more control with the SDK such that we could delete all tags when `nil` was specified to CAPZ. ASO currently doesn't have a way to distinguish between "I don't care what the tags are" which is how it currently interprets `nil` and "I want the tags to be empty." https://github.com/Azure/azure-service-operator/issues/3367 will enable us to specify `nil` in the ASO spec and it will delete any existing tags and there will no longer be a way to express "I don't care what the tags are" which I don't think we would ever need.

jackfrancis: Rather than deleting should we comment out and add a // TODO to re-enable when we consume a version of ASO that includes https://github.com/Azure/azure-service-operator/issues/3367 ?

nojnhuh: The linter doesn't like commented out code, but I'll add a comment.

azure/services/agentpools/spec_test.go
nojnhuh, line 1: I took the liberty of refactoring these tests since I'm confident this isn't introducing any regressions in the unit tests. And since we aren't doing any diffing in `Parameters` anymore, a lot of the existing tests felt clunky to include given how things work now. These new tests provide more coverage in 500+ fewer lines of code.

Makefile
CecileRobertMichon: How does this affect installation with clusterctl init or other mechanisms?

api/v1beta1/azuremanagedcontrolplane_types.go
```diff
@@ -155,7 +162,7 @@ type AzureManagedControlPlaneSpec struct {
 
 	// LoadBalancerSKU is the SKU of the loadBalancer to be provisioned.
 	// Immutable.
-	// +kubebuilder:validation:Enum=Basic;Standard
+	// +kubebuilder:validation:Enum=Basic;basic;Standard;standard
```
CecileRobertMichon, line 165: what's the reason for supporting both? Would it be easier to keep supporting one and do the conversion in CAPZ?

CecileRobertMichon (commented):
Love all the code removal

api/v1beta1/azuremanagedcontrolplane_webhook.go
nojnhuh: Fixed by deleting this altogether :)

api/v1beta1/azuremanagedcontrolplane_types.go
nojnhuh: Yeah, supporting only one set is likely easier. Just pushed that change.

Makefile
nojnhuh: clusterctl init doesn't do exactly the same thing as `kubectl apply` AFAIK where this would come into play. I'll try this again manually to make sure this isn't an issue there, but we would be seeing a similar error in e2e if it was, right?

nojnhuh: Just confirmed that `clusterctl init` completes successfully and all the CRDs get installed.

For any particular method of installing CAPZ, I'm not sure there's anything we can can realistically do here to prevent this specific issue since it essentially boils down to "resource too big" and we can't control how big the ASO CRDs are. If this could be proactively prevented, it would probably make the most sense to make that kind of change in ASO since it affects anyone installing the ManagedCluster CRD.

nojnhuh:
@willie-yao @nawazkh PTAL

azure/services/agentpools/spec.go
```diff
@@ -152,173 +151,97 @@ type AgentPoolSpec struct {
...
-	}
-	tags := converters.TagsToMap(s.AdditionalTags)
-	if tags == nil {
-		// Make sure we send a non-nil, empty map if AdditionalTags are nil as this tells AKS to delete any existing tags.
-		tags = make(map[string]*string, 0)
-	}
-	var vnetSubnetID *string
-	if s.VnetSubnetID != "" {
-		vnetSubnetID = &s.VnetSubnetID
+	spec := &agentPool.Spec
```
CecileRobertMichon, line 174: same comment here with the spec naming

azure/services/agentpools/spec.go
CecileRobertMichon: is that a regression? what's the eta for 2.4?

CecileRobertMichon (commented):
Did an in depth review of the spec files which I missed on my original review since they were collapsed 🙃  This is looking really good overall, great work @nojnhuh!

azure/services/managedclusters/spec.go
```diff
@@ -380,429 +386,209 @@ func (s *ManagedClusterSpec) Parameters(ctx context.Context, existing interface{
...
-		}
-
-		diff := computeDiffOfNormalizedClusters(managedCluster, existingMC)
-		if diff == "" {
-			log.V(4).Info("no changes found between user-updated spec and existing spec")
-			return nil, nil
-		}
-		log.V(4).Info("found a diff between the desired spec and the existing managed cluster", "difference", diff)
-	} else {
+	// Only include AgentPoolProfiles during initial cluster creation. Agent pools are managed solely by the
```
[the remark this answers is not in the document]
nojnhuh: Yes, this is still necessary for now since AKS requires them on create. And if we leave the agentPoolProfiles in the ManagedCluster spec after it's created, then later updates to the ManagedCluster may override the ManagedClustersAgentPools.

azure/services/managedclusters/spec.go
```diff
@@ -380,429 +386,209 @@ func (s *ManagedClusterSpec) Parameters(ctx context.Context, existing interface{
...
 	return
 }
 
-// isAuthIPRangesNilOrEmpty returns true if the managed cluster's APIServerAccessProfile or AuthorizedIPRanges is nil or if AuthorizedIPRanges is empty.
-func isAuthIPRangesNilOrEmpty(managedCluster armcontainerservice.ManagedCluster) bool {
-	return managedCluster.Properties.APIServerAccessProfile == nil ||
-		managedCluster.Properties.APIServerAccessProfile.AuthorizedIPRanges == nil ||
-		reflect.DeepEqual(managedCluster.Properties.APIServerAccessProfile.AuthorizedIPRanges, []*string{})
+func adminKubeconfigSecretName(clusterName string) string {
+	return secret.Name(clusterName+"-aso", secret.Kubeconfig)
```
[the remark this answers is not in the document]
nojnhuh: Yep, I'll add a comment.

azure/services/managedclusters/spec.go
```diff
@@ -304,53 +294,68 @@ func buildAutoScalerProfile(autoScalerProfile *AutoScalerProfile) *armcontainers
...
-	var decodedSSHPublicKey []byte
-	if s.SSHPublicKey != "" {
-		decodedSSHPublicKey, err = base64.StdEncoding.DecodeString(s.SSHPublicKey)
-		if err != nil {
-			return nil, errors.Wrap(err, "failed to decode SSHPublicKey")
+	managedCluster := existing
+	if managedCluster == nil {
+		managedCluster = &asocontainerservicev1.ManagedCluster{
+			Spec: asocontainerservicev1.ManagedCluster_Spec{
+				Tags: infrav1.Build(infrav1.BuildParams{
```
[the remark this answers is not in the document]
nojnhuh: We set them here which includes the `additionalTags` and all the other "standard" ones set on all resources, but then manage the `additionalTags` later with the tags service for the reason stated in that comment. This should match how they're managed today.

azure/services/managedclusters/spec.go
```diff
@@ -304,53 +294,68 @@ func buildAutoScalerProfile(autoScalerProfile *AutoScalerProfile) *armcontainers
...
+	spec.ServicePrincipalProfile = &asocontainerservicev1.ManagedClusterServicePrincipalProfile{
+		ClientId: ptr.To("msi"),
+	}
+	spec.NetworkProfile = &asocontainerservicev1.ContainerServiceNetworkProfile{
+		NetworkPlugin:   azure.AliasOrNil[asocontainerservicev1.ContainerServiceNetworkProfile_NetworkPlugin](&s.NetworkPlugin),
+		LoadBalancerSku: azure.AliasOrNil[asocontainerservicev1.ContainerServiceNetworkProfile_LoadBalancerSku](&s.LoadBalancerSKU),
+		NetworkPolicy:   azure.AliasOrNil[asocontainerservicev1.ContainerServiceNetworkProfile_NetworkPolicy](&s.NetworkPolicy),
+	}
+	spec.AutoScalerProfile = buildAutoScalerProfile(s.AutoScalerProfile)
+	spec.OperatorSpec = &asocontainerservicev1.ManagedClusterOperatorSpec{
```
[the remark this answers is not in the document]
nojnhuh: That seems to be an ASO convention for "configuration for the fancy Kubernetes-specific things that ASO does that are outside the scope of the API specs." In this case that's putting the kubeconfigs for the clusters in Secrets, and this defines what those Secrets should look like.

azure/services/managedclusters/spec.go
```diff
@@ -304,53 +294,68 @@ func buildAutoScalerProfile(autoScalerProfile *AutoScalerProfile) *armcontainers
...
-
-			ServicePrincipalProfile: &armcontainerservice.ManagedClusterServicePrincipalProfile{
-				ClientID: ptr.To("msi"),
-			},
-			AgentPoolProfiles: []*armcontainerservice.ManagedClusterAgentPoolProfile{},
-			NetworkProfile: &armcontainerservice.NetworkProfile{
-				NetworkPlugin:   azure.AliasOrNil[armcontainerservice.NetworkPlugin](&s.NetworkPlugin),
-				LoadBalancerSKU: azure.AliasOrNil[armcontainerservice.LoadBalancerSKU](&s.LoadBalancerSKU),
-				NetworkPolicy:   azure.AliasOrNil[armcontainerservice.NetworkPolicy](&s.NetworkPolicy),
+	spec := &managedCluster.Spec
```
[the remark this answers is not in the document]
nojnhuh: Would it be more clear to get rid of this `spec` variable and use its expanded form inline?

azure/services/managedclusters/spec.go
```diff
@@ -380,429 +386,209 @@ func (s *ManagedClusterSpec) Parameters(ctx context.Context, existing interface{
...
-	resourceReferences := make([]*armcontainerservice.ResourceReference, len(resources))
+func convertToResourceReferences(resources []string) []asocontainerservicev1.ResourceReference {
+	resourceReferences := make([]asocontainerservicev1.ResourceReference, len(resources))
 	for i := range resources {
-		resourceReferences[i] = &armcontainerservice.ResourceReference{ID: &resources[i]}
-	}
-	return resourceReferences
-}
-
-func computeDiffOfNormalizedClusters(managedCluster armcontainerservice.ManagedCluster, existingMC armcontainerservice.ManagedCluster) string {
```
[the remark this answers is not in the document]
nojnhuh: and the bugs!

azure/services/managedclusters/spec.go
```diff
@@ -380,429 +386,209 @@ func (s *ManagedClusterSpec) Parameters(ctx context.Context, existing interface{
...
+			AdminGroupObjectIDs: s.AADProfile.AdminGroupObjectIDs,
 		}
 		if s.DisableLocalAccounts != nil {
-			managedCluster.Properties.DisableLocalAccounts = s.DisableLocalAccounts
+			spec.DisableLocalAccounts = s.DisableLocalAccounts
+		}
+
+		if ptr.Deref(s.DisableLocalAccounts, false) {
+			// admin credentials cannot be fetched when local accounts are disabled
+			spec.OperatorSpec.Secrets.AdminCredentials = nil
```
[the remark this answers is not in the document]
nojnhuh: I'll see how I can improve this.

azure/services/managedclusters/spec.go
```diff
@@ -380,429 +386,209 @@ func (s *ManagedClusterSpec) Parameters(ctx context.Context, existing interface{
...
+		loadBalancerProfile.OutboundIPs = &asocontainerservicev1.ManagedClusterLoadBalancerProfile_OutboundIPs{
 			PublicIPs: convertToResourceReferences(s.LoadBalancerProfile.OutboundIPs),
 		}
 	}
 	return
 }
 
-func convertToResourceReferences(resources []string) []*armcontainerservice.ResourceReference {
-	resourceReferences := make([]*armcontainerservice.ResourceReference, len(resources))
+func convertToResourceReferences(resources []string) []asocontainerservicev1.ResourceReference {
```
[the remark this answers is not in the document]
nojnhuh: It looks like the types are getting in the way of being able to utilize this function there.

docs/book/src/topics/managedcluster.md
```diff
@@ -243,28 +243,6 @@ spec:
       name: test-subnet
 ```
 
-### Enable AKS features with custom headers (--aks-custom-headers)
```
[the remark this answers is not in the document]
nojnhuh: This thread seems to be the best I can find to describe the current stance. Do you know of a better link? https://github.com/Azure/azure-rest-api-specs/pull/18232

test/e2e/aks_tags.go
```diff
@@ -182,16 +182,13 @@ func AKSAdditionalTagsSpec(ctx context.Context, inputGetter func() AKSAdditional
...
-			expectedTags = nil
+			// TODO: restore the behavior to delete all tags like the above control plane test does once we
+			// include ASO v2.4.0: https://github.com/Azure/azure-service-operator/issues/3367
```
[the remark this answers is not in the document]
nojnhuh: No, I don't think we need to block this PR on ASO v2.4.0 just because of this. I'll open an issue and remove this TODO.

azure/services/agentpools/spec.go
nojnhuh: It is a regression, but if ASO v2.4.0 is included for CAPZ v1.12 then users shouldn't notice. @matthchr Do you think the next ASO minor release will be cut before mid-November?

azure/services/managedclusters/spec.go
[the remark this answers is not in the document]
nojnhuh: I'm finding that the inverse logic is quite hard to follow on its own, so I think I prefer this as-is unless I'm missing a better way to go about this.
```go
if s.AADProfile == nil || !ptr.Deref(s.DisableLocalAccounts, true) {
	// set the admin kubeconfig
}
```

test/e2e/aks_tags.go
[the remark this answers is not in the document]
nojnhuh: https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/4212

azure/services/managedclusters/managedclusters_test.go
```diff
@@ -21,234 +21,86 @@ import (
...
+		kclient := fakeclient.NewClientBuilder().
+			WithObjects(adminASOKubeconfig, userASOKubeconfig).
+			Build()
+		scope.EXPECT().GetClient().Return(kclient).AnyTimes()
+
+		scope.EXPECT().SetControlPlaneEndpoint(clusterv1.APIEndpoint{
+			Host: "fdqn",
+			Port: 443,
+		})
+		scope.EXPECT().ClusterName().Return(clusterName).AnyTimes()
```
willie-yao, line 81: Do we know how many times ClusterName() is expected to be called? AnyTimes() will be true even if ClusterName() isn't called, so maybe it's better to use MinTimes()

azure/services/managedclusters/spec.go
```diff
@@ -380,429 +389,209 @@ func (s *ManagedClusterSpec) Parameters(ctx context.Context, existing interface{
 			// https://golang.org/src/net/ip.go#L48
 			ip[15] = byte(10)
 			dnsIP := ip.String()
-			managedCluster.Properties.NetworkProfile.DNSServiceIP = &dnsIP
-		} else {
-			managedCluster.Properties.NetworkProfile.DNSServiceIP = s.DNSServiceIP
```
willie-yao, line 385: Do we no longer want to set the `DNSServiceIP` to the existing one? I'm assuming ASO will handle it but just making sure

willie-yao (commented):
Great work @nojnhuh! Just a few minor comments and questions from my end, otherwise LGTM

api/v1beta1/azuremanagedcontrolplane_webhook.go
```diff
@@ -76,10 +76,6 @@ func (mw *azureManagedControlPlaneWebhook) Default(ctx context.Context, obj runt
 		networkPlugin := "azure"
 		m.Spec.NetworkPlugin = &networkPlugin
 	}
```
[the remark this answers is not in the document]
nojnhuh: It can, but given this is the example I put in #4204 I think I'd rather leave this in to make getting started with that issue easier.

api/v1beta1/azuremanagedmachinepool_webhook_test.go
```diff
@@ -300,120 +286,39 @@ func TestAzureManagedMachinePoolUpdatingWebhook(t *testing.T) {
...
-			},
-			wantErr: false,
-		},
 		{
 			name: "Unchanged OSDiskType in an agentpool should not result in an error",
 			new: &AzureManagedMachinePool{
 				Spec: AzureManagedMachinePoolSpec{
 					Mode:         "System",
 					SKU:          "StandardD2S_V3",
 					OSDiskSizeGB: ptr.To[int32](512),
```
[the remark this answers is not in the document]
nojnhuh: I think I missed this one since it turns out to be a lot more subtle than the other ones. Thanks!

azure/services/managedclusters/managedclusters_test.go
nojnhuh: I don't think it's particularly important in this case that we record in the test exactly how many times this function should be called. Calls to this function I imagine will fluctuate over time more than most others as we add features and refactor, so keeping this a little vague seems okay to me.

azure/services/managedclusters/spec.go
nojnhuh: It's set above. This is a not-entirely-shameless code coverage hack.

azure/services/managedclusters/spec.go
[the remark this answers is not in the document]
CecileRobertMichon: yes potentially, or name it something more explicit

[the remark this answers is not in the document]
nojnhuh: Made this change. I think this makes things look a lot more similar to the SDK services so hopefully any potential bugs fail a sniff test pretty easily.

docs/book/src/topics/managedcluster.md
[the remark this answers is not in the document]
CecileRobertMichon: I don't :/

nojnhuh:
I've found that this PR hits a known AKS bug that our tests run into easily enough that I think we should block this PR on a fix. I haven't seen the same bug arise without ASO but I'll keep my eyes open.

nojnhuh:
> I've found that this PR hits a known AKS bug that our tests run into easily enough that I think we should block this PR on a fix. I haven't seen the same bug arise without ASO but I'll keep my eyes open.

The tests only really consistently hit this with tags, but not since I updated this PR back to using the separate tags APIs. Maybe we don't need to block this PR then if a fix is actively being worked on in AKS.

azure/services/managedclusters/spec.go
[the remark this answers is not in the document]
CecileRobertMichon: For this I meant specifically the fact that we're setting `spec.OperatorSpec.Secrets.AdminCredentials` way above on line 340 and then conditionally resetting it nil lower. Moving those two closer to each other might help, even if we don't inverse the logic?

https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/4069/files#diff-d7496d9845d55dda5142aab2970a5a9aa1012fbb52d2c462cc5b220682ca520aR340

CecileRobertMichon (commented):
LGTM aside from that last comment on the OperatorSpec admin kubeconfig in spec

azure/services/managedclusters/spec.go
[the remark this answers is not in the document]
nojnhuh: Agree that moving them closer together helps. Just pushed that change.

azure/services/agentpools/spec.go
nojnhuh: ASO v2.4 ETA is next week, which should be enough time for us.

CecileRobertMichon (commented):
/lgtm

We might still want to wait for ASO 2.4 to merge this for good measure, even if we're confident it will be available before the 1.12 release

azure/services/managedclusters/spec.go
```diff
@@ -512,12 +513,6 @@ func (s *ManagedClusterSpec) Parameters(ctx context.Context, existing *asocontai
 		}
 	}
 
-	if managedCluster.Status.Tags != nil {
-		// tags managed separately because updating tags concurrently with agent pools' can cause the cluster
-		// to get stuck in an "Updating" state forever.
```
nojnhuh: This fix is being rolled out in AKS and I checked that this workaround doesn't seem necessary anymore in westcentralus. Tests may fail depending on region before it's fully rolled out.

nojnhuh: northeurope seems to not have been updated yet.

/retest

CecileRobertMichon: @nojnhuh is this fixed now?

test/e2e/aks_tags.go
```diff
@@ -98,16 +98,11 @@ func AKSAdditionalTagsSpec(ctx context.Context, inputGetter func() AKSAdditional
 			}
 		}
 
-		By("Deleting all tags for control plane")
```
CecileRobertMichon, line 101: @nojnhuh does this need to be re-added when #4248 merges?

test/e2e/aks_tags.go
CecileRobertMichon: Does this need to be re-added now?

azure/services/managedclusters/spec.go
nojnhuh: Yes, in regions running v20231029.

https://releases.aks.azure.com/webpage/index.html

test/e2e/aks_tags.go
nojnhuh: It doesn't necessarily need to be, but it could. This should also be added to the node labels and node taints tests, so I was planning on doing that all in one follow-up PR (next week) unless we want that in sooner.

test/e2e/aks_tags.go
nojnhuh: Same as https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/4069#discussion_r1391756050

test/e2e/aks_tags.go
CecileRobertMichon: Do we have a follow-up issue for labels and taints? It wouldn't hurt to have that test signal when we merge this one IMO, if we're blocking it on v2.4 anyways but works either way.

nojnhuh: We do now: https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/4266.

jackfrancis (approved):

PR #3971, opened 2023-09-07
Install calico in e2e tests with CAAPH

Jont828:
**What type of PR is this?**

**What this PR does / why we need it**: Use CAAPH to install calico chart instead of Helm SDK in e2e tests. Replaces #3571 as the original PR's approach was outdated.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #3736

**Special notes for your reviewer**:

- [ ] cherry-pick candidate

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Install calico Helm chart using CAAPH
```

PR #4086, opened 2023-10-03
Add template to test ipv6 and dual stack with k8s CI versions

CecileRobertMichon:
**What type of PR is this?**

**What this PR does / why we need it**: Adds templates for ipv6 and dual stack using release-branch CI k8s versions so we can test various k8s release branches with CAPZ before releases to catch future regressions like https://github.com/kubernetes/kubernetes/issues/120999.

Fixes #

- [ ] cherry-pick candidate

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```


CecileRobertMichon:
@nojnhuh the broken link error is coming from https://github.com/nojnhuh/cluster-api-provider-azure/tree/aso/azure/services/asogroups which is referred to in the ASO proposal. I'm guessing that branch no longer exists... Is there a good replacement for it?

nojnhuh:
The link now magically works again. I'll think about what better link to keep in the doc.
sry

CecileRobertMichon:
This is ready for review

/assign @nawazkh @jackfrancis

CecileRobertMichon:
/hold

will merge #3971 first and rebase

templates/test/ci/prow-ci-version-ipv6/patches/machine-deployment.yaml
```diff
@@ -0,0 +1,38 @@
...
+            bind-address: "::"
+---
+apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
+kind: AzureMachineTemplate
+metadata:
+  name: "${CLUSTER_NAME}-md-0"
+spec:
+  template:
+    spec:
+      enableIPForwarding: true
```
nawazkh, line 38: Minor nit: EOF new line.

nawazkh (approved):
Just a minor nit of EOF which is not a blocker tho, can be fixed in another PR to avoid running all the tests again.
Looks good to me! 🚢

templates/test/ci/prow-ci-version-dual-stack/patches/machine-deployment.yaml
```diff
@@ -0,0 +1,23 @@
...
+  name: "${CLUSTER_NAME}-md-0"
+spec:
+  template:
+    spec:
+      postKubeadmCommands:
+        # This frees up :53 on the host for the coredns pods
+        - echo "DNSStubListener=no" >> /etc/systemd/resolved.conf
+        - mv /etc/resolv.conf /etc/resolv.conf.OLD && ln -s /run/systemd/resolve/resolv.conf /etc/resolv.conf
+        - systemctl restart systemd-resolved
+
```
willie-yao, line 23: nit: extra newline

test/e2e/conformance_test.go
```diff
@@ -118,6 +118,15 @@ var _ = Describe("Conformance Tests", func() {
 			}
 		}
 
+		// use the ipv6 flavor if ipv6 IP family is specified.
+		if e2eConfig.GetVariable(capi_e2e.IPFamily) == "IPv6" {
+			flavor += "-ipv6"
+			kubetestConfigFilePath = strings.Replace(kubetestConfigFilePath, ".yaml", "-ipv6.yaml", 1)
```
willie-yao, line 124: Quick clarifying question: Will this still find the right flavor if we use CI or PR artifacts? I don't see pre-existing templates for them even without the ip suffix so I'm wondering if this may cause problems if we're getting the templates from elsewhere.

The flavors being set: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/0151252433f9b0d284312fe0dc761173c4980d4a/test/e2e/conformance_test.go#L115

templates/test/ci/prow-ci-version-ipv6/patches/machine-deployment.yaml
CecileRobertMichon: fixed

templates/test/ci/prow-ci-version-dual-stack/patches/machine-deployment.yaml
CecileRobertMichon: fixed

test/e2e/conformance_test.go
CecileRobertMichon: Yes for CI, that's what this PR is enabling, the flavors are defined here: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/4086/files#diff-35f348a5d567ebeba39b1e40bbab0e1a7f64009cbadff142952f9b9691e94666R121

For PR artifacts no since those templates haven't been added. I could change the line so that it only changes the suffix for CI artifacts but then if you set "ipv6" and "pr artifacts" the test would silently fail (it would just use the regular template) instead of failing with a flavor not found error so I think it's arguably worse. This also is more future proof if we do add the equivalent ipv6/dual stack PR artifacts templates later.

CecileRobertMichon:
/hold

rebase is done

test/e2e/data/kubetest/conformance-ipv6-ginkgo-v2.yaml
```diff
@@ -0,0 +1,12 @@
+ginkgo.focus: \[Feature\:Networking-IPv6\]
+ginkgo.skip: \[Feature\:SCTPConnectivity\]|\[Experimental\]
```
jackfrancis, line 2: Why are we skipping `[Experimental]` in ipv6 but not dual-stack? Do `[Experimental]`-tagged test scenarios require ipv4, and our k8s dual-stack clusters are configured in such a way that those ipv4-dependent tests can be reliably tested?

templates/test/ci/cluster-template-prow-ci-version-dual-stack.yaml
```diff
@@ -0,0 +1,875 @@
...
+            PACKAGE_VERSION="$(apt-cache madison kubelet|grep $${VERSION_REGEX}- | head -n1 | cut -d '|' -f 2 | tr -d '[:space:]')"
+            for CI_PACKAGE in "$${PACKAGES_TO_TEST[@]}"; do
+              echo "* installing package: $$CI_PACKAGE $${PACKAGE_VERSION}"
+              DEBIAN_FRONTEND=noninteractive apt-get install -y $$CI_PACKAGE=$$PACKAGE_VERSION
+            done
+          else
+            CI_URL="https://storage.googleapis.com/k8s-release-dev/ci/$${CI_VERSION}/bin/linux/amd64"
+            for CI_PACKAGE in "$${PACKAGES_TO_TEST[@]}"; do
+              echo "* downloading binary: $$CI_URL/$$CI_PACKAGE"
+              wget --inet4-only "$$CI_URL/$$CI_PACKAGE" -nv -O "$$CI_DIR/$$CI_PACKAGE"
```
jackfrancis, line 159: I assume that this `--inet4-only` flag has nothing to do w/ ipv4 k8s clusters (sort of a funny coincidence that this change would land in this PR :)).

jackfrancis (approved):
/lgtm

unless any of the questions I posed are actionable

test/e2e/data/kubetest/conformance-ipv6-ginkgo-v2.yaml
CecileRobertMichon: Because there is no `[Experimental]` dual-stack test but there was one ipv6 one that was failing https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/4086/pull-cluster-api-provider-azure-conformance-dual-stack-with-ci-artifacts/1711771121797304320

I could add it to both but the dual-stack one would be a no-op, with ginkgo skip I prefer to only skip tests only as needed.

templates/test/ci/cluster-template-prow-ci-version-dual-stack.yaml
CecileRobertMichon: It does, this is to ensure we always connect to the ipv4 IP address of `https://storage.googleapis.com/k8s-release-dev`, IPv6 one was not reachable. It's a no-op for the ivp4 template since we always used the ipv4 adress but it ensure the ipv6/dual stack templates which are running on dual stack hosts use the ipv4 address.

jackfrancis: ACK, was a little strange at first to see this in the ipv6 template, but it makes sense that the actual OS is dual stack (ipv6-only config is just the k8s surface area)

CecileRobertMichon:
@jackfrancis can you please approve if all your comments were addressed?

PR #4094, opened 2023-10-06
Fix invalid boilerplate script issue

### LINKED ISSUE #4038: Fix invalid boilerplate
/kind cleanup

**What needs cleanup:**

`make verify-boilerplate` has been failing for some time because the script that invokes doesn't fail when the `python` command isn't available, which it's not in the prow environment:

https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/directory/pull-cluster-api-provider-azure-verify/1706387974339432448#1:build-log.txt%3A21

The script should be updated to invoke `python3` instead of `python` like CAPI: https://github.com/kubernetes-sigs/cluster-api/blob/4c6e89cf2ca962599856ff8e302894bdd4387bdf/hack/boilerplate/boilerplate.py#L1.

Once that change is made, `make verify-boilerplate` will report errors that should also be fixed.

wowsoundaryaa:
**What type of PR is this?**

/kind cleanup

This PR should fix boilerplate script failing when `python` version is not available.
Changes includes
- change command `python` to `python3`
- update dockerfile ref file with correct version
- Fix space related issues in copyright headers

Fixes #4038

### DIFF
--- a/azure/errors_test.go
+++ b/azure/errors_test.go
@@ -1,3 +1,19 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 package azure
 
 import (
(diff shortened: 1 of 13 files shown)


nojnhuh:
/lgtm
/release-note-none

Thank you @wowsoundaryaa! Could you please squash all commits into one and then force-push to this branch?
/hold

wowsoundaryaa:
@nojnhuh squashed the commits. Thanks

nojnhuh:
Thank you for squashing!
/hold cancel

azure/errors_test.go
```diff
@@ -1,3 +1,19 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
Jont828, line 2:
```suggestion
Copyright 2023 The Kubernetes Authors.
```

Jont828 (commented):
LGTM aside from a nit

azure/errors_test.go
nojnhuh: As best as I can tell, this file was originally added in 2019. Not sure exactly what it means that there wasn't an explicit copyright on it before then, but I think it's fine as-is.

Jont828: Gotcha, I just wasn't sure if the copyright header is allowed to be retroactive or not

PR #4097, opened 2023-10-06
fix webhook for CNI overlay

### LINKED ISSUE #4093: Can't create cluster with Azure CNI Overlay and NetworkPolicy
/kind bug

#3678 introduced support for [Azure CNI Overlay](https://learn.microsoft.com/en-us/azure/aks/azure-cni-overlay) on AKS managed clusters. However, there appears to be a bug when you create an AKS cluster with both `networkPolicy` and `networkPluginMode` enabled.
```yaml
networkPolicy: calico
networkPluginMode: overlay
```
With this configuration, the cluster creation does not happen and the capi-controller throws this error:
```
Reconciler error" err="[failed to retrieve Spec.ControlPlaneEndpoint from infrastructure provider for Cluster \"XXXXXXXX\" in namespace \"XXXXXXXXX\": field not found, admission webhook \"validation.azuremanagedcontrolplanes.infrastructure.cluster.x-k8s.io\" denied the request: AzureManagedControlPlane.infrastructure.cluster.x-k8s.io \"XXXXXXXXX\" is invalid: Spec.NetworkPluginMode: Forbidden: \"overlay\" NetworkPolicyMode cannot be enabled when NetworkPolicy is set]" controller="cluster" controllerGroup="cluster.x-k8s.io" controllerKind="Cluster" Cluster="XXXXXXX/XXXXXXX" namespace="XXXXXXX" name="XXXXXXX" reconcileID=b88e32a7-ea71-4769-a69f-ceb0196145e0
```
As can be seen in #3678 I actually tested this exact scenario with success at first. However, I did not do any follow up tests after the final commits happened on the PR after [this discussion.](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3988#discussion_r1323342321) This makes me think that the validation that was added introduces the bug, thus not fulfilling the criteria "You can create a cluster w/ CNI Overlay + NetworkPolicy when you create your cluster".

```yaml
#networkPolicy: calico
networkPluginMode: overlay
```
Commenting out `networkPolicy` works fine.

**Environment:**

- cluster-api-provider-azure version: 1.11.2
- Kubernetes version: (use `kubectl version`): 1.26.6

nojnhuh:
**What type of PR is this?**
/kind bug
/area managedclusters

This PR fixes a bug in the AzureManagedControlPlane update webhook when checking `spec.networkPluginMode`. The intent is to disallow changing the `networkPluginMode` to `overlay` when `spec.networkPolicy` is already enabled without overlay. The bug was disallowing updates to AzureManagedControlPlane whenever `overlay` is specified and `networkPolicy` was specified without verifying that the update modifies `networkPluginMode`. 

Fixes #4093

- [X] cherry-pick candidate

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Release note**:

```release-note
Fixed a bug disallowing certain valid updates to AzureManagedControlPlane with Azure CNI overlay enabled
```

### DIFF
--- a/api/v1beta1/azuremanagedcontrolplane_webhook_test.go
+++ b/api/v1beta1/azuremanagedcontrolplane_webhook_test.go
@@ -1556,6 +1556,7 @@ func TestAzureManagedControlPlane_ValidateUpdate(t *testing.T) {
 					Name: "test-cluster",
 				},
 				Spec: AzureManagedControlPlaneSpec{
+					NetworkPolicy:     ptr.To("anything"),
 					NetworkPluginMode: ptr.To(NetworkPluginModeOverlay),
 				},
 			},
@@ -1577,6 +1578,29 @@ func TestAzureManagedControlPlane_ValidateUpdate(t *testing.T) {
 					Name: "test-cluster",
 				},
 				Spec: AzureManagedControlPlaneSpec{
+					NetworkPluginMode: ptr.To(NetworkPluginModeOverlay),
+					Version:           "v0.0.0",
+				},
+			},
+			wantErr: false,
+		},
+		{
+			name: "NetworkPolicy is allowed when NetworkPluginMode is not changed",
+			oldAMCP: &AzureManagedControlPlane{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "test-cluster",
+				},
+				Spec: AzureManagedControlPlaneSpec{
+					NetworkPolicy:     ptr.To("anything"),
+					NetworkPluginMode: ptr.To(NetworkPluginModeOverlay),
+				},
+			},
+			amcp: &AzureManagedControlPlane{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "test-cluster",
+				},
+				Spec: AzureManagedControlPlaneSpec{
+					NetworkPolicy:     ptr.To("anything"),
 					NetworkPluginMode: ptr.To(NetworkPluginModeOverlay),
 					Version:           "v0.0.0",
 				},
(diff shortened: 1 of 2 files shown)


nojnhuh:
/cherry-pick release-1.11
(overlay cannot be enabled in v1.10)

api/v1beta1/azuremanagedcontrolplane_webhook_test.go
```diff
@@ -1583,6 +1584,29 @@ func TestAzureManagedControlPlane_ValidateUpdate(t *testing.T) {
 			},
 			wantErr: false,
 		},
+		{
+			name: "NetworkPolicy is allowed when NetworkPluginMode is not changed",
```
jackfrancis, line 1588: There's a permutation we're not covering in UT:

1. Old = some NetworkPolicy + overlay NetworkPlugin
2. New = some different NetworkPolicy + overlay NetworkPlugin

I assume that AKS would reject that update, do we have the right webhooks to optimize for CAPZ UX?

nojnhuh: We already have webhook and test coverage for the immutability of NetworkPlugin which isn't affected by the value of NetworkPluginMode. Or do you think it's worth checking to see if some updates might actually be allowed when overlay is enabled?

nojnhuh: s/NetworkPlugin/NetworkPolicy/

jackfrancis: The subtlety of this change just makes me want to cover every (potentially) artificial scenario.

If we're fixing a specific use case let's just ship this and then we can watch closely and fine-tune our webhooks as we observe edge cases. We're already in a place where the actual behaviors don't match AKS documentation so I think we're probably just gonna have to be reactive. :/

PR #4108, opened 2023-10-09
migrate private endpoints service to ASO

### LINKED ISSUE #3530: ASO service: privateendpoints

nawazkh:
**What type of PR is this?**
/kind feature

- This PR migrates private endpoints service to use ASO framework.

Fixes #3530 

- I tested it locally, and I was able to bring up a private-endpoint using the default template.

- [x] cherry-pick candidate

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
migrate Private Endpoints Service to use ASO Framework
```


controllers/azuremanagedcontrolplane_controller.go
```diff
@@ -113,6 +113,8 @@ func (amcpr *AzureManagedControlPlaneReconciler) SetupWithManager(ctx context.Co
 // +kubebuilder:rbac:groups=cluster.x-k8s.io,resources=clusters;clusters/status,verbs=get;list;watch
 // +kubebuilder:rbac:groups=resources.azure.com,resources=resourcegroups,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=resources.azure.com,resources=resourcegroups/status,verbs=get;list;watch
+// +kubebuilder:rbac:groups=network.azure.com,resources=privateendpoints,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=network.azure.com,resources=privateendpoints/status,verbs=get;list;watch
```
nawazkh, line 117: I can remove these annotations from here since adding these is duplicating it in `role.yaml`.
But keeping it here for discussion.

azure/services/privateendpoints/spec.go
```diff
@@ -54,209 +51,122 @@ type PrivateEndpointSpec struct {
...
 
-	privateEndpointProperties.CustomNetworkInterfaceName = ptr.To(s.CustomNetworkInterfaceName)
-
-	privateIPAddresses := make([]*armnetwork.PrivateEndpointIPConfiguration, 0, len(s.PrivateIPAddresses))
-	for _, address := range s.PrivateIPAddresses {
-		ipConfig := &armnetwork.PrivateEndpointIPConfigurationProperties{PrivateIPAddress: ptr.To(address)}
-
-		privateIPAddresses = append(privateIPAddresses, &armnetwork.PrivateEndpointIPConfiguration{
-			Properties: ipConfig,
+// Parameters implements azure.ASOResourceSpecGetter.
```
nawazkh, line 64: The `Parameters()` func fills up the fields of `asonetworkv1.PrivateEndpoint_Spec{}` only if they are present. 
This means that 
- _arrays_ with `length <= 0` do not get added.
- empty strings (`""` ) do not get added.

The reason to point this out specifically is that, in this implementation,
1. **Reason:**`Parameters()` func mimics the UX of a user creating an `asonetworkv1.PrivateEndpoint{}` type. Users will not care to populate nil values to the fields they are not interested in; they will only add the fields they are interested in. Therefore the untouched fields will default to their `nil types` and hence get omitted by the CRDs.
2. **Question:** Now that Parameters is mimicking UX of a user creating an `asonetworkv1.PrivateEndpoint{}` type , we may run into issues in the future if ever ASO decides to have different behavior for `nil types` and `empty types`. For instance, if ASO decides to have a different behavior for `privateEndpoint.Spec.Location = ptr.To("")` and `privateEndpoint.Spec.Location = nil`, CAPZ will run into errors. 
  - One way to solve this issue would be to update `s.Location` to be of type `*string` instead of `string`. But this could be a breaking change.

azure/scope/cluster_test.go
```diff
@@ -3438,3 +3439,246 @@ func TestVNetPeerings(t *testing.T) {
...
+					Spec: infrav1.AzureClusterSpec{
+						NetworkSpec: infrav1.NetworkSpec{
+							Subnets: infrav1.Subnets{},
+						},
+					},
+				},
+				cache: &ClusterCache{},
+			},
+			want: make([]azure.ASOResourceSpecGetter[*asonetworkv1.PrivateEndpoint], 0),
+		},
```
nawazkh, line 3462: Should the `PrivateEndpointSpecs()` func return a `nil` array or an `empty` array? 
cc: @nojnhuh and @willie-yao

nawazkh:
Updating unit tests. 
/hold

azure/services/privateendpoints/spec.go
nojnhuh: I'd default to sticking to whatever we were doing before with the SDK for each field to reduce the risk of introducing regressions. From what I've seen, the ASO and SDK types match pretty closely in terms of what is and isn't a pointer.

I think trying to align that behavior across all the fields for every resource is out of scope for these PRs for now.

azure/scope/managedcontrolplane_test.go
```diff
@@ -1038,3 +1041,171 @@ func TestAreLocalAccountsDisabled(t *testing.T) {
...
+		},
+	}
+	for _, c := range cases {
+		c := c
+		t.Run(c.Name, func(t *testing.T) {
+			g := NewWithT(t)
+			fakeClient := fake.NewClientBuilder().WithScheme(scheme).WithObjects(c.Input.ControlPlane).Build()
+			c.Input.Client = fakeClient
+			s, err := NewManagedControlPlaneScope(context.TODO(), c.Input)
+			g.Expect(err).To(Succeed())
```
nojnhuh, line 1205: Do we really need to build a client to test this function? Or could we get around having to call `NewManagedControlPlaneScope` by defining a `*ManagedControlPlaneScope` directly?

azure/services/privateendpoints/privateendpoints_test.go
```diff
@@ -21,114 +21,185 @@ import (
...
 	internalError = autorest.NewErrorWithResponse("", "", &http.Response{StatusCode: http.StatusInternalServerError}, "Internal Server Error")
 	notDoneError  = azure.NewOperationNotDoneError(&infrav1.Future{})
 )
 
+func init() {
+	_ = clusterv1.AddToScheme(scheme.Scheme)
+	_ = asonetworkv1.AddToScheme(scheme.Scheme)
+}
+
 func TestReconcilePrivateEndpoint(t *testing.T) {
```
nojnhuh, line 125: I still think it's not worth testing the generic `Reconcile` implementation in every service that uses it. If there are gaps in the tests for the generic service, then we should fill those there.

controllers/azuremanagedcontrolplane_controller.go
nojnhuh: I'm overall +1 to keeping this here since it serves as decent documentation even if it makes no tangible difference to the generated RBAC when it's already defined elsewhere.

azure/services/privateendpoints/spec.go
nawazkh: > I think trying to align that behavior across all the fields for every resource is out of scope for these PRs for now.

I agree for the sake of exploding this PR with a ton of changes to review and merge. 

> the ASO and SDK types match pretty closely in terms of what is and isn't a pointer.

To a certain extent, yes, but not everything. 
For instance, `asonetworkv1.privateendpoint.PrivateEndpoint_Spec.Location` is a type `*string` and `PrivateEndpointSpec.Location` (at CAPZ side) is of type `string`. Although `Location` as a field is innocuous in this scenario; meaning an `""` and a `nil` `asonetworkv1.privateendpoint.PrivateEndpoint_Spec.Location` mean the same for the ASO CRD and will get dropped(due to omit empty annotation), I am suggesting that we might have to be wary in carrying over other fields at other places. 

Happy to revert to direct copy over of values btw!

nawazkh: > But we're not changing the CAPZ type for Location here right?

Yes, we are not changing it. 

> So if we were translating a string from CAPZ into a *string for the SDK, can we do the same translation to *string for ASO?

I am not sure if we can do the same ? 

The way I understand it: 
- Working with SDK meant we created underlying Azure resources. 
- Working with ASO means we create a CR representing ASO resources's CRD, just like any user would manually create a CR. It is ASO's responsibility to create the Azure Resource for the requested CR.

So the fact that ASO chose to broadcast `asonetworkv1.privateendpoint.PrivateEndpoint_Spec.Location` as a type `*string` rather than a `string` suggests to me that they want to differentiate between a user initializing `asonetworkv1.privateendpoint.PrivateEndpoint_Spec.Location` and `""` value it gets assigned if left untouched.

(Although they list `Location` to be _omitted when empty_ (`Location *string 'json:"location,omitempty"'`) which means both `nil` and `""` would be treated the same way. 


I wonder if we will run into issues for fields of other data types like custom structs, etc.

azure/scope/managedcontrolplane_test.go
nawazkh: Good catch, we don't need to. Updating..

azure/services/privateendpoints/spec.go
nojnhuh: > The way I understand it:
> 
> * Working with SDK meant we created underlying Azure resources.
> * Working with ASO means we create a CR representing ASO resources's CRD, just like any user would manually create a CR. It is ASO's responsibility to create the Azure Resource for the requested CR.

These still seem like fundamentally the same thing to me though since ASO objects are basically just Kubernetes-shaped versions of the SDK objects.


> So the fact that ASO chose to broadcast `asonetworkv1.privateendpoint.PrivateEndpoint_Spec.Location` as a type `*string` rather than a `string` suggests to me that they want to differentiate between a user initializing `asonetworkv1.privateendpoint.PrivateEndpoint_Spec.Location` and `""` value it gets assigned if left untouched.

Right, and this was also the case with the SDK before, so I think we can keep the same behavior that we had previously.

> (Although they list `Location` to be _omitted when empty_ (`Location *string 'json:"location,omitempty"'`) which means both `nil` and `""` would be treated the same way.

With `omitempty` and any pointer type, only `nil` will get dropped from the serialized representation. The non-nil empty string is not omitted.

azure/services/privateendpoints/privateendpoints_test.go
nawazkh: I agree that testing the `Reconcile` functionality is redundant. 
But these are good tests to test the overall code flow; i.e. `specs` -> `reconcile` -> `createOrUpdate` -> `updateStatus` a service. 
It was through these tests I discovered that `PrivateEndpointSpecs` would never return `nil` `[]azure.ASOResourceSpecGetter[*asonetworkv1.PrivateEndpoint]`, instead it would return an empty list. 

(In our scenario, a `nil` list and an `empty` list doesn't really make a difference since we only `range` over them in _aso/service.go_

nojnhuh: privateendpoints.go isn't providing any logic itself outside of initializing a struct, and those struct fields are all generic enough that any tests here could be replicated in the generic tests. Especially if we keep replicating these tests across all the services, I think the maintenance cost will pretty quickly surpass how many new bugs we find. Before it was more necessary since each service had its own implementation for these functions, but these tests are executing literally the same code that's already tested.

azure/services/privateendpoints/privateendpoints.go
```diff
@@ -17,114 +17,34 @@ limitations under the License.
...
 type PrivateEndpointScope interface {
-	azure.Authorizer
-	azure.AsyncStatusUpdater
-	PrivateEndpointSpecs() []azure.ResourceSpecGetter
+	aso.Scope
+	PrivateEndpointSpecs() []azure.ASOResourceSpecGetter[*asonetworkv1.PrivateEndpoint]
 }
 
 // Service provides operations on Azure resources.
 type Service struct {
```
nojnhuh, line 36: I realized that these `Service` structs don't really seem necessary anymore. I removed them from my managedclusters PR: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/4069/commits/2d37c52093d7537425a506a066ce9d1f971a274f

azure/services/privateendpoints/spec.go
nawazkh: TIL: https://goplay.tools/snippet/y3p2j_hB24w

nawazkh: > > So the fact that ASO chose to broadcast `asonetworkv1.privateendpoint.PrivateEndpoint_Spec.Location` as a type `*string` rather than a `string` suggests to me that they want to differentiate between a user initializing `asonetworkv1.privateendpoint.PrivateEndpoint_Spec.Location` and `""` value it gets assigned if left untouched.
> 
> Right, and this was also the case with the SDK before, so I think we can keep the same behavior that we had previously.

👍🏼

azure/services/privateendpoints/privateendpoints.go
nawazkh: Good find!

azure/scope/managedcontrolplane_test.go
nawazkh: Added via 178d32f4039b9117767f49c599edcaf0bb1c202a

azure/services/privateendpoints/privateendpoints_test.go
nawazkh: Dropped the tests via 178d32f4039b9117767f49c599edcaf0bb1c202a

azure/services/privateendpoints/privateendpoints.go
nawazkh: Updated via 178d32f4039b9117767f49c599edcaf0bb1c202a

nawazkh:
@nojnhuh @willie-yao Please take a look when you get a chance. 
I tested this locally and was able to bring up a `privateendpoints.network.azure.com` resource.

azure/services/privateendpoints/privateendpoints.go
```diff
@@ -17,114 +17,25 @@ limitations under the License.
...
 type PrivateEndpointScope interface {
-	azure.Authorizer
-	azure.AsyncStatusUpdater
-	PrivateEndpointSpecs() []azure.ResourceSpecGetter
-}
-
-// Service provides operations on Azure resources.
-type Service struct {
-	Scope PrivateEndpointScope
-	async.Reconciler
```
willie-yao, line 43: Is there a reason we're removing the entire Service struct here? I see that for other ASO refactors, we keep the Service struct, and return a `*Service` instead of the specific `*aso.Service[*asonetworkv1.PrivateEndpoint, PrivateEndpointScope]` type. Example: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/4059/files#diff-633dd65a18aedbd1705ccb20c9455162123a91bde0b689d6248d97fca01f286eL44

azure/services/privateendpoints/spec_test.go
```diff
@@ -20,332 +20,228 @@ import (
...
-				PrivateLinkServiceConnections: []PrivateLinkServiceConnection{{
-					Name:                 privateEndpoint1.PrivateLinkServiceConnections[0].Name,
-					GroupIDs:             privateEndpoint1.PrivateLinkServiceConnections[0].GroupIDs,
-					PrivateLinkServiceID: privateEndpoint1.PrivateLinkServiceConnections[0].PrivateLinkServiceID,
-					RequestMessage:       privateEndpoint1.PrivateLinkServiceConnections[0].RequestMessage,
-				}},
-				SubnetID: "test-subnet",
+	}
+
+	fakeASOPrivateEndpointSpecWithExtendedLocation = asonetworkv1.PrivateEndpoint_Spec{
```
willie-yao, line 102: nit to reduce the amount of repeated code: it could help to use the `changes` pattern here since these PrivateEndpoint specs are so large, although it may not be worth the effort if we're only defining two private endpoints. An example is this: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/controllers/asosecret_controller_test.go#L320

willie-yao (commented):
Awesome job @nawazkh! Just minor comments from my end.

azure/services/privateendpoints/privateendpoints.go
nawazkh: Good catch!
Jon noticed this first and mentioned it in his comment at https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/4108#discussion_r1373847073

The `func NewService[T deepCopier[T], S Scope](name string, scope S) *Service[T, S]` returns us the `&service[T, S]` we are looking for when called at the cluster scope. Hence, the service struct defined at the `privateEndpoints.go` becomes redundant.

nawazkh: Thanks for pointing it out. I should update Natgateways service too and drop the Service struct.

azure/services/privateendpoints/spec_test.go
nawazkh: Good point. I think I can reduce at least a couple of lines. Making changes.
/hold

nawazkh: Updated. Please take a look. 
/unhold

azure/services/privateendpoints/privateendpoints.go
nawazkh: Created https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/4224 updating Natgateway.

nawazkh:
Tests are failing due to rebase churn. Please hold off from reviewing. I will unblock ones tests are stable.
/hold.

nawazkh:
[pull-cluster-api-provider-azure-test](https://prow.k8s.io/job-history/gs/kubernetes-jenkins/pr-logs/directory/pull-cluster-api-provider-azure-test) is complaining that it cannot load `/logs/artifacts/coverage.out` 🤔 , strange, probing.
```bash
make: *** [Makefile:561: lint] Error 1
make: Leaving directory '/home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure'
cp: cannot stat 'coverage.*': No such file or directory
go: downloading github.com/spf13/cobra v1.7.0
go: downloading golang.org/x/tools v0.8.0
Couldn't load /logs/artifacts/coverage.out: open /logs/artifacts/coverage.out: no such file or directory.+ EXIT_VALUE=1
+ set +o xtrace
```

nawazkh:
I was able to create a cluster enabled with private endpoint using this PR. 🎉 
```bash
cluster.cluster.x-k8s.io/default-11305 created
azurecluster.infrastructure.cluster.x-k8s.io/default-11305 created
kubeadmcontrolplane.controlplane.cluster.x-k8s.io/default-11305-control-plane created
azuremachinetemplate.infrastructure.cluster.x-k8s.io/default-11305-control-plane created
machinedeployment.cluster.x-k8s.io/default-11305-md-0 created
azuremachinetemplate.infrastructure.cluster.x-k8s.io/default-11305-md-0 created
kubeadmconfigtemplate.bootstrap.cluster.x-k8s.io/default-11305-md-0 created
azureclusteridentity.infrastructure.cluster.x-k8s.io/cluster-identity created
Cluster 'default-11305' created, don't forget to delete
NAME: cloud-provider-azure-1701293746
LAST DEPLOYED: Wed Nov 29 13:35:47 2023
NAMESPACE: default
STATUS: deployed
REVISION: 1
TEST SUITE: None
NAME: calico
LAST DEPLOYED: Wed Nov 29 13:35:57 2023
NAMESPACE: tigera-operator
STATUS: deployed
REVISION: 1
TEST SUITE: None
```
```bash
❯ kg privateendpoints.network.azure.com
NAME                             READY   SEVERITY   REASON      MESSAGE
default-11305-private-endpoint   True               Succeeded
```

azure/services/privateendpoints/spec_test.go
```diff
@@ -356,14 +229,14 @@ func TestParameters(t *testing.T) {
 			g := NewWithT(t)
 			t.Parallel()
 
-			result, err := tc.spec.Parameters(context.TODO(), tc.existing)
+			result, err := tc.spec.Parameters(context.TODO(), tc.existing.DeepCopy())
```
nojnhuh, line 232: FYI this particular `DeepCopy` isn't strictly necessary if `tc.existing` isn't being read after calling `Parameters`.

nojnhuh:
@nawazkh If there isn't one already, could you please open an issue to track adding e2e coverage for private endpoints?

azure/services/privateendpoints/spec.go
```diff
@@ -54,209 +51,118 @@ type PrivateEndpointSpec struct {
...
+		for _, applicationSecurityGroup := range s.ApplicationSecurityGroups {
+			applicationSecurityGroups = append(applicationSecurityGroups, asonetworkv1.ApplicationSecurityGroupSpec_PrivateEndpoint_SubResourceEmbedded{
+				Reference: &genruntime.ResourceReference{
+					ARMID: applicationSecurityGroup,
+				},
+			})
+		}
+		sort.SliceStable(applicationSecurityGroups, func(i, j int) bool {
+			return applicationSecurityGroups[i].Reference.ARMID < applicationSecurityGroups[j].Reference.ARMID
 		})
```
willie-yao, line 84: Why is this sort needed? I think a comment explaining this could be helpful.

willie-yao (commented):
Looks good pending one minor comment!

azure/services/privateendpoints/spec.go
nawazkh: This sort was carried over from the previous code flow. 
Earlier code seems to be preserving the order of both new and existing private endpoints. xref: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/4108/files#diff-c5a1236250878b000a50b1ce5a20dd4a7412176fd3e8bcb34f48e7dab6ac8a7cL163-L165 
But, yes, adding a comment is a good idea, will update.

azure/services/privateendpoints/spec.go
```diff
@@ -54,209 +51,118 @@ type PrivateEndpointSpec struct {
...
-			Name: manualPrivateLinkServiceConnection.Name,
-			Properties: &armnetwork.PrivateLinkServiceConnectionProperties{
-				PrivateLinkServiceID: manualPrivateLinkServiceConnection.Properties.PrivateLinkServiceID,
-				RequestMessage:       manualPrivateLinkServiceConnection.Properties.RequestMessage,
-				GroupIDs:             manualPrivateLinkServiceConnection.Properties.GroupIDs,
-			},
-		})
-	}
-	normalizedExistingPE.Properties.ManualPrivateLinkServiceConnections = existingManualPrivateLinkServiceConnections
+	// tags
```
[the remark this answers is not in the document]
nawazkh: I'm happy to remove them. 
I added these as markers when converting Parameter() foo to ASO.

azure/services/privateendpoints/spec.go
```diff
@@ -54,209 +51,118 @@ type PrivateEndpointSpec struct {
...
+func (s *PrivateEndpointSpec) Parameters(ctx context.Context, existingPrivateEndpoint *asonetworkv1.PrivateEndpoint) (*asonetworkv1.PrivateEndpoint, error) {
+	privateEndpoint := &asonetworkv1.PrivateEndpoint{}
+	privateEndpoint.Spec = asonetworkv1.PrivateEndpoint_Spec{}
+	if existingPrivateEndpoint != nil {
+		privateEndpoint = existingPrivateEndpoint
+	}
+
+	// ApplicationSecurityGroups
+	if len(s.ApplicationSecurityGroups) > 0 {
+		applicationSecurityGroups := make([]asonetworkv1.ApplicationSecurityGroupSpec_PrivateEndpoint_SubResourceEmbedded, 0, len(s.ApplicationSecurityGroups))
```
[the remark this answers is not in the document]
nawazkh: I agree. But if this minor over-engineering guardrails us from unforeseen scenarios and adds a bit of performance, I am happy to take it. 😃

azure/scope/managedcontrolplane_test.go
```diff
@@ -1027,3 +1031,180 @@ func TestAreLocalAccountsDisabled(t *testing.T) {
 		})
 	}
 }
+
+func TestManagedControlPlaneScope_PrivateEndpointSpecs(t *testing.T) {
+	scheme := runtime.NewScheme()
```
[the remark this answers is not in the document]
nawazkh: Good catch, we don't need this schema. Removing.

azure/services/privateendpoints/spec.go
```diff
@@ -54,209 +51,118 @@ type PrivateEndpointSpec struct {
...
-func (s *PrivateEndpointSpec) Parameters(ctx context.Context, existing interface{}) (interface{}, error) {
-	_, log, done := tele.StartSpanWithLogger(ctx, "privateendpoints.Service.Parameters")
-	defer done()
-
-	privateEndpointProperties := armnetwork.PrivateEndpointProperties{
-		Subnet: &armnetwork.Subnet{
-			ID: &s.SubnetID,
-			Properties: &armnetwork.SubnetPropertiesFormat{
-				PrivateEndpointNetworkPolicies:    ptr.To(armnetwork.VirtualNetworkPrivateEndpointNetworkPoliciesDisabled),
-				PrivateLinkServiceNetworkPolicies: ptr.To(armnetwork.VirtualNetworkPrivateLinkServiceNetworkPoliciesEnabled),
```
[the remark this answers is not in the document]
nawazkh: Good question @nojnhuh , Let me put a question on the ASO channel. 

My theory from probing the Azure SDK request body params (xref: https://learn.microsoft.com/en-us/rest/api/virtualnetwork/private-endpoints/create-or-update?view=rest-virtualnetwork-2023-05-01&tabs=HTTP#request-body) is that `PrivateEndpointNetworkPolicy: Disabled` and `PrivateLinkServiceNetworkPolicy: Enabled` appear to be the **default** values for the fields being explicitly specified. 

From the examples of creating PrivateEndpoints using Azure SDK, I see that privateEndpoints request payload only requires the SubnetIDs, I am _guessing_ ASO is also following the same. 

But, let me confirm.

azure/services/privateendpoints/spec.go
```diff
@@ -54,209 +51,118 @@ type PrivateEndpointSpec struct {
...
-	privateIPAddresses := make([]*armnetwork.PrivateEndpointIPConfiguration, 0, len(s.PrivateIPAddresses))
-	for _, address := range s.PrivateIPAddresses {
-		ipConfig := &armnetwork.PrivateEndpointIPConfigurationProperties{PrivateIPAddress: ptr.To(address)}
-
-		privateIPAddresses = append(privateIPAddresses, &armnetwork.PrivateEndpointIPConfiguration{
-			Properties: ipConfig,
+// Parameters implements azure.ASOResourceSpecGetter.
+func (s *PrivateEndpointSpec) Parameters(ctx context.Context, existingPrivateEndpoint *asonetworkv1.PrivateEndpoint) (*asonetworkv1.PrivateEndpoint, error) {
+	privateEndpoint := &asonetworkv1.PrivateEndpoint{}
+	privateEndpoint.Spec = asonetworkv1.PrivateEndpoint_Spec{}
```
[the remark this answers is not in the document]
nawazkh: Removing

nawazkh:
> @nawazkh If there isn't one already, could you please open an issue to track adding e2e coverage for private endpoints?

Created https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/4336

azure/services/privateendpoints/spec_test.go
```diff
@@ -20,332 +20,205 @@ import (
...
-		Name:                          "test-private-endpoint-manual",
-		ApplicationSecurityGroups:     []string{"asg1"},
-		PrivateLinkServiceConnections: []infrav1.PrivateLinkServiceConnection{{PrivateLinkServiceID: "testPl", RequestMessage: "Please approve my connection.", GroupIDs: []string{"aa", "bb"}}},
-		ManualApproval:                true,
-		CustomNetworkInterfaceName:    "test-if-name",
-		PrivateIPAddresses:            []string{"10.0.0.1", "10.0.0.2"},
-		Location:                      "test-location",
+	fakeASOPrivateEndpoint = asonetworkv1.PrivateEndpoint{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fakePrivateEndpoint.Name,
```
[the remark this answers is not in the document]
nawazkh: Good catch Jon, thanks. I will use `ptr.To()` to avoid unexpected mutations.

PR #4129, opened 2023-10-13
Support for auto upgrade channel

### LINKED ISSUE #4115: Support for AKS upgrade channels
/kind feature

Support for AKS automated upgrade channels

https://learn.microsoft.com/en-us/azure/aks/auto-upgrade-cluster

**Environment:**

LochanRn:
**What type of PR is this?**
/kind feature

Fixes #4115

- [ ] cherry-pick candidate

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Release note**:

```release-note
Support for Auto upgrade channels in AKS.
```


api/v1beta1/azuremanagedcontrolplane_types.go
```diff
@@ -34,6 +34,50 @@ const (
 	PrivateDNSZoneModeNone string = "None"
 )
 
+// UpgradeChannel - For more information see setting the AKS cluster auto-upgrade channel [https://docs.microsoft.com/azure/aks/upgrade-cluster#set-auto-upgrade-channel].
```
Jont828, line 37:
```suggestion
// UpgradeChannel is ...
// For more information see setting the AKS cluster auto-upgrade channel [https://docs.microsoft.com/azure/aks/upgrade-cluster#set-auto-upgrade-channel].
```
Could you add a brief summary of what it is and leave the link at the end?

api/v1beta1/azuremanagedcontrolplane_types.go
```diff
@@ -34,6 +34,50 @@ const (
...
+type UpgradeChannel string
+
+const (
+	// UpgradeChannelNodeImage - Automatically upgrade the node image to the latest version available. Consider using nodeOSUpgradeChannel
```
Jont828, line 41:
```suggestion
	// UpgradeChannelNodeImage automatically upgrades the node image to the latest version available. Consider using nodeOSUpgradeChannel
```

I think this came up in some of your other PRs but please reformat these doc comments to be a sentence without the dash like in this example. That's the doc convention we're using in the rest of the repo so it would be best to stick with it. It's fine if you want to change that at the end, though, once you've got everything else in this PR ready to go.

api/v1beta1/azuremanagedcontrolplane_types.go
```diff
@@ -34,6 +34,50 @@ const (
...
+	// instead as that allows you to configure node OS patching separate from Kubernetes version patching.
```
Jont828, line 42:
```suggestion
	// instead as that allows you to configure node OS patching separate from Kubernetes version patching.
	
```

Please add a newline in between the declaration as well for consistency

api/v1beta1/azuremanagedcontrolplane_types.go
```diff
@@ -221,6 +265,23 @@ type AzureManagedControlPlaneSpec struct {
...
+}
+
+// ManagedClusterAutoUpgradeProfile - Auto upgrade profile for a managed cluster.
+type ManagedClusterAutoUpgradeProfile struct {
+	// NodeOSUpgradeChannel is a manner in which the OS on your nodes is updated. The default is NodeImage. Possible values include: NodeImage,Unmanaged,None
+	// +kubebuilder:validation:Enum=NodeImage;Unmanaged;None
+	// +optional
+	NodeOSUpgradeChannel *NodeOSUpgradeChannel `json:"nodeOSUpgradeChannel,omitempty"`
+
+	// UpgradeChannel - upgrade channel for auto upgrade. Possible values include: "node-image","none","patch","rapid","stable"
```
Jont828, line 281:
```suggestion
	// UpgradeChannel - upgrade channel for auto upgrade. Possible values include: 'node-image','none','patch','rapid','stable'
```
Use single quotes for consistency with the rest of the repo

api/v1beta1/azuremanagedcontrolplane_types.go
```diff
@@ -364,6 +425,11 @@ type ManagedControlPlaneSubnet struct {
 
 // AzureManagedControlPlaneStatus defines the observed state of AzureManagedControlPlane.
 type AzureManagedControlPlaneStatus struct {
+	// AutoUpgradeVersion is the Kubernetes version populated after autoupgrade based on the upgrade channel.
+	// +kubebuilder:validation:MinLength:=2
```
Jont828, line 429:
```suggestion
	// +kubebuilder:validation:MinLength=2
```
I think it won't work with a semicolon btw

api/v1beta1/azuremanagedcontrolplane_webhook.go
```diff
@@ -492,6 +497,35 @@ func (m *AzureManagedControlPlane) validateManagedClusterNetwork(cli client.Clie
 	return nil
 }
 
+// validateAutoUpgradeProfile validates auto upgrade profile.
+func (m *AzureManagedControlPlane) validateAutoUpgradeProfile(old *AzureManagedControlPlane) field.ErrorList {
+	var allErrs field.ErrorList
+	if old.Spec.AutoUpgradeProfile != nil && m.Spec.AutoUpgradeProfile == nil {
+		// Prevent AutoUpgradeProfile to be set to nil.
+		// unsetting the field is not allowed
```
Jont828, line 505:
```suggestion
		// Unsetting the field is not allowed.
```

azure/scope/managedcontrolplane.go
```diff
@@ -478,6 +478,26 @@ func (s *ManagedControlPlaneScope) IsAADEnabled() bool {
...
+	s.ControlPlane.Status.AutoUpgradeVersion = version
+}
+
+// IsManagedVersionUpgrade checks if version is auto managed by AKS.
+func (s *ManagedControlPlaneScope) IsManagedVersionUpgrade() bool {
+	return isManagedVersionUpgrade(s.ControlPlane)
+}
+
+func isManagedVersionUpgrade(managedControlPlane *infrav1.AzureManagedControlPlane) bool {
+	if managedControlPlane.Spec.AutoUpgradeProfile != nil &&
```
Jont828, line 492: Instead of
```
if expr {
return true
}
return false
```
we could just do `return expr` instead as the `if` block is redundant.

azure/scope/managedmachinepool.go
```diff
@@ -341,3 +338,25 @@ func (s *ManagedMachinePoolScope) GetCAPIMachinePoolAnnotation(key string) (succ
...
 	return ok, val
 }
+
+// IsManagedAutoUpgrade checks if version is auto managed by AKS.
+func (s *ManagedMachinePoolScope) IsManagedAutoUpgrade() bool {
+	return isManagedVersionUpgrade(s.ControlPlane)
+}
+
+func getManagedMachinePoolVersion(managedControlPlane *infrav1.AzureManagedControlPlane,
+	machinePool *expv1.MachinePool) *string {
```
Jont828, line 348: Could keep that on the same line as above

azure/services/agentpools/spec.go
```diff
@@ -224,6 +247,13 @@ func (s *AgentPoolSpec) Parameters(ctx context.Context, existing interface{}) (p
 				Tags:                converters.TagsToMap(s.AdditionalTags),
 			},
 		}
+
+		if kubernetesVersion, err := s.GetManagedMachinePoolVersion(existing); err != nil {
+			return nil, err
+		} else {
```
Jont828, line 253: Nit: can drop the else clause and add it directly after `if` block as we return if we enter the `if`.

azure/services/managedclusters/managedclusters.go
```diff
@@ -164,11 +172,11 @@ func (s *Service) IsManaged(ctx context.Context) (bool, error) {
...
 /*
   Returns the admin kubeconfig and user kubeconfig
-  If aad is enabled a user kubeconfig will also get generated and stored in the secret <cluster-name>-kubeconfig-user
-  If we disable local accounts for aad clusters we do not have access to admin kubeconfig, hence we need to create
+  If AAD is enabled a user kubeconfig will also get generated and stored in the secret <cluster-name>-kubeconfig-user
+  If we disable local accounts for AAD clusters we do not have access to admin kubeconfig, hence we need to create
   the admin kubeconfig by authenticating with the user credentials and retrieving the token for kubeconfig.
   The token is used to create the admin kubeconfig.
-  The user needs to ensure to provide service principle with admin aad privileges.
+  The user needs to ensure to provide service principal with admin AAD privileges.
```
Jont828, line 179: Nice catch!

util/versions/version.go
```diff
@@ -0,0 +1,41 @@
...
+
+package versions
+
+import (
+	semverv4 "github.com/blang/semver"
+	"github.com/pkg/errors"
+)
+
+// GetHigherK8sVersion returns the higher k8s version out of a and b.
+func GetHigherK8sVersion(a, b string) (string, error) {
```
Jont828, line 25: Is the reason we need this function because one of `a` or `b` might be an empty string?

Jont828 (commented):
Thanks for the contribution! Here's some feedback so far.

api/v1beta1/azuremanagedcontrolplane_types.go
LochanRn: ack

api/v1beta1/azuremanagedcontrolplane_types.go
LochanRn: may be, not sure the same was used for spec.version.

api/v1beta1/azuremanagedcontrolplane_types.go
LochanRn: ack, will keep this in mind for all my upcoming PR's

azure/services/managedclusters/managedclusters.go
LochanRn: Thanks :)

util/versions/version.go
LochanRn: I did not quite follow could :(

azure/scope/managedmachinepool.go
LochanRn: ack

azure/scope/managedcontrolplane.go
LochanRn: ack

api/v1beta1/azuremanagedcontrolplane_webhook.go
LochanRn: ack

util/versions/version.go
Jont828: I'm wondering when we could be returning an error for `errv1` or `errv2` since I'm a bit confused on why we're handling the errors this way.

azure/scope/managedmachinepool.go
```diff
@@ -341,3 +338,24 @@ func (s *ManagedMachinePoolScope) GetCAPIMachinePoolAnnotation(key string) (succ
...
+
+func getManagedMachinePoolVersion(managedControlPlane *infrav1.AzureManagedControlPlane, machinePool *expv1.MachinePool) *string {
+	var v, av string
+	if machinePool != nil {
+		v = ptr.Deref(machinePool.Spec.Template.Spec.Version, "")
+	}
+	if managedControlPlane != nil {
+		av = managedControlPlane.Status.AutoUpgradeVersion
+	}
+	higherVersion, _ := versions.GetHigherK8sVersion(v, av)
```
Jont828, line 355: I don't think we necessarily want to be dropping the error here.

util/versions/version.go
LochanRn: The actual functionality what i wanted to implement is something like below. 
```
The actual functionality that is required is something like this. 
// GetHigherK8sVersion returns the higher k8s version out of a and b.
func GetHigherK8sVersion(a, b string) (string, error) {
	v1, errv1 := semverv4.ParseTolerant(a)
	v2, errv2 := semverv4.ParseTolerant(b)
	if errv1 != nil && errv2 != nil {
		return "", errors.Wrapf(errv1, "error parsing k8s version %s, %v error parsing k8s version %s", a, errv2, b)
	}
	if errv1 != nil {
		return b, nil
	}
	if errv2 != nil {
		return a, nil
	}
	if v1.GTE(v2) {
		return a, nil
	}
	return b, nil
}
```

There should be an error returned only if both the a and b are incorrect. 
If a is incorrect but b is fine return b
If b is incorrect but a is fine return a
If both the versions are correct then get the higher versions out of the two.

LochanRn: I am still thinking how to make this better, even i feel it's not very neat. 

@Jont828 open to suggestions actually.

Jont828: I think if either are incorrect, we should return `"", err` as I don't think we want to proceed if there was an error. That should probably simplify the logic in this function to something like:

```
// GetHigherK8sVersion returns the higher k8s version out of a and b.
func GetHigherK8sVersion(a, b string) (string, error) {
	v1, err := semverv4.ParseTolerant(a)
	if err != nil {
		return "", nil
	}
	v2, errv2 := semverv4.ParseTolerant(b)
		if err != nil {
		return "", nil
	}
	if v1.GTE(v2) {
		return a, nil
	}
	return b, nil
}
```

api/v1beta1/azuremanagedcontrolplane_webhook.go
```diff
@@ -492,6 +497,35 @@ func (m *AzureManagedControlPlane) validateManagedClusterNetwork(cli client.Clie
...
+	if old.Spec.AutoUpgradeProfile != nil && m.Spec.AutoUpgradeProfile == nil {
+		// Prevent AutoUpgradeProfile to be set to nil.
+		// Unsetting the field is not allowed.
+		allErrs = append(allErrs,
+			field.Invalid(
+				field.NewPath("Spec", "AutoUpgradeProfile"),
+				m.Spec.AutoUpgradeProfile,
+				"field cannot be set to nil, to disable auto upgrades set the channel to none."))
+	}
+
```
Jont828, line 512: @LochanRn Actually, I think a better approach would be to have `validateCreate()` above do a check on `m.Spec.Version` like this. Since the field is not optional, we shouldn't let the set the field to that value at all if won't parse.

```
_, err := semverv4.ParseTolerant(m.Spec.Version)
if err != nil {
	return errors.new("version is invalid", m.Spec.Version)
}
```

Then, in validateUpdate, we can just validate the new value `m.Spec.Version`, and since we know they're both valid, go ahead and use `GTE()` directly instead of needing a helper function.

```
newVersion, err := semverv4.ParseTolerant(m.Spec.Version)
if err != nil {
	return errors.new("version is invalid", m.Spec.Version)
}
oldVersion, err := semverv4.ParseTolerant(old.Spec.Version)
if err != nil {
	// Note: we don't expect this to fail as we already validated the version
	return errors.new("version is invalid", old.Spec.Version)
}
if newVersion.GTE(oldVersion) {
	...
}
```

util/versions/version.go
LochanRn: The function is being used in multiple places and not just in the webhook actually.

Jont828: Gotcha. WDYT about my other suggestion? I think it would be reasonable to have a validation check in the webhook that the version will parse so that in the code we don't need to do something funky with errors.

LochanRn: yup makes sense I can do that.

LochanRn: there is a validation that validates based on this regex 
`^v(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)([-0-9a-zA-Z_\.+]*)?$`
do you still think we need to validate it using semver parser ?

api/v1beta1/azuremanagedcontrolplane_webhook.go
LochanRn: I updated the helper funtcion to not return any error and in validate create i updated the code to parse the version for semver4.ParseTolerant.

util/versions/version_test.go
```diff
@@ -0,0 +1,100 @@
...
+			name:   "b is invalid",
+			a:      "1.18.1",
+			b:      "v1.17.8.",
+			output: "1.18.1",
+		},
+		{
+			name:   "b is invalid",
+			a:      "9.99.9999",
+			b:      "v1.17.8.",
+			output: "9.99.9999",
```
nawazkh, line 84: Just curious, although this unit test tests the function, in real life scenario, is `9.99.9999` a meaningful SemVer? 
Can user replicate this scenario?

api/v1beta1/azuremanagedcontrolplane_types.go
```diff
@@ -34,6 +34,50 @@ const (
 	PrivateDNSZoneModeNone string = "None"
 )
 
+// UpgradeChannel determines the type of upgrade channel for automatically upgrading the cluster.
+type UpgradeChannel string
+
+const (
+	// UpgradeChannelNodeImage automatically upgrades the node image to the latest version available.
+	// Consider using nodeOSUpgradeChannel instead as that allows you to configure node OS patching separate from Kubernetes version patching.
+	UpgradeChannelNodeImage UpgradeChannel = "node-image"
```
Jont828, line 43:
```suggestion
	UpgradeChannelNodeImage UpgradeChannel = "node-image"
	
```
Please add a newline between variable declarations!

nawazkh (commented):
Minor comments, overall looks great to me. 
Thank you @LochanRn 🚀

api/v1beta1/azuremanagedcontrolplane_webhook.go
```diff
@@ -336,6 +342,10 @@ func (m *AzureManagedControlPlane) validateVersion(_ client.Client) error {
 		return errors.New("must be a valid semantic version")
 	}
 
+	if _, err := semverv4.ParseTolerant(m.Spec.Version); err != nil {
```
Jont828, line 345: Have you thought about using [this package](https://pkg.go.dev/golang.org/x/mod/semver#IsValid) instead? It seems like it has a similar function for [compare](https://pkg.go.dev/golang.org/x/mod/semver#Compare) that you're looking for.

azure/services/managedclusters/spec.go
```diff
@@ -125,8 +127,20 @@ type ManagedClusterSpec struct {
...
+
+	// AutoUpgradeProfile - Profile of auto upgrade configuration.
+	AutoUpgradeProfile *ManagedClusterAutoUpgradeProfile
+}
+
+// ManagedClusterAutoUpgradeProfile auto upgrade profile for a managed cluster.
+type ManagedClusterAutoUpgradeProfile struct {
+	// NodeOSUpgradeChannel is a manner in which the OS on your nodes is updated. The default is NodeImage. Possible values include: NodeImage,Unmanaged,None
+	NodeOSUpgradeChannel *infrav1.NodeOSUpgradeChannel
+	// UpgradeChannel - upgrade channel for auto upgrade. Possible values include: 'UpgradeChannelRapid', 'UpgradeChannelStable', 'UpgradeChannelPatch', 'UpgradeChannelNodeImage', 'UpgradeChannelNone'
```
Jont828, line 142: Nit: comment format

util/versions/version.go
```diff
@@ -0,0 +1,40 @@
...
+*/
+
+package versions
+
+import (
+	semverv4 "github.com/blang/semver"
+)
+
+// GetHigherK8sVersion returns the higher k8s version out of a and b.
+func GetHigherK8sVersion(a, b string) string {
```
Jont828, line 24: WDYT about using this [compare function](https://pkg.go.dev/golang.org/x/mod/semver#Compare) instead? It seems like that would solve your issue of trying to handle error parsing.

azure/services/managedclusters/spec.go
```diff
@@ -125,8 +127,20 @@ type ManagedClusterSpec struct {
 
 	// DNSPrefix allows the user to customize dns prefix.
 	DNSPrefix *string
+
 	// DisableLocalAccounts disables getting static credentials for this cluster when set. Expected to only be used for AAD clusters.
 	DisableLocalAccounts *bool
+
+	// AutoUpgradeProfile - Profile of auto upgrade configuration.
```
Jont828, line 134: Nit: comment format

Jont828 (commented):
Just a couple small things!

util/versions/version_test.go
LochanRn: Not sure, was just trying out different test cases and checking if the function works fine that's is :slightly_smiling_face:

azure/services/managedclusters/spec.go
LochanRn: ack

azure/services/managedclusters/spec.go
LochanRn: ack

util/versions/version.go
LochanRn: ack and makes sense, thank you for pointing me out to that function, will use this function which will  reduce a lot of chaos :smile:

api/v1beta1/azuremanagedcontrolplane_types.go
LochanRn: ack

api/v1beta1/azuremanagedcontrolplane_webhook.go
LochanRn: ack

api/v1beta1/azuremanagedcontrolplane_types.go
```diff
@@ -221,6 +265,23 @@ type AzureManagedControlPlaneSpec struct {
...
 	DisableLocalAccounts *bool `json:"disableLocalAccounts,omitempty"`
+
+	// AutoUpgradeProfile - Profile of auto upgrade configuration.
+	// +optional
+	AutoUpgradeProfile *ManagedClusterAutoUpgradeProfile `json:"autoUpgradeProfile,omitempty"`
+}
+
+// ManagedClusterAutoUpgradeProfile - Auto upgrade profile for a managed cluster.
+type ManagedClusterAutoUpgradeProfile struct {
+	// NodeOSUpgradeChannel is a manner in which the OS on your nodes is updated. The default is NodeImage. Possible values include: NodeImage,Unmanaged,None
```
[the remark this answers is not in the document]
LochanRn: not defaulting it any where actually.

util/versions/version.go
LochanRn: yup changed it accordingly.

api/v1beta1/azuremanagedcontrolplane_types.go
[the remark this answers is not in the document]
LochanRn: Not sure if we have to default it, because i think if we don't set a value as well it would be NodeImage.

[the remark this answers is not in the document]
nawazkh: > Not sure if we have to default it, because i think if we don't set a value as well it would be NodeImage.

I see. You mean this gets set to `NodeImage` by the Azure API?

I wonder if we could add an annotation to default this field to `NodeImage`? I ask since the comment says `The default is NodeImage` It might puzzle the user if it doesn't reflect on the CAPZ spec when left uninitialized. What do you think?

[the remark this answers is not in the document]
LochanRn: I did not understand what you are trying to say, by adding an annotation how can annotation default the field ?

Jont828:
Seems like you're running into this error in the AKS test from the [controller logs](https://[storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/4129/pull-cluster-api-provider-azure-e2e-aks/1720143615415554048/artifacts/clusters/bootstrap/logs/capz-system/capz-controller-manager/capz-controller-manager-5b8ff65d59-xjhpg/manager.log](https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/4129/pull-cluster-api-provider-azure-e2e-aks/1720143615415554048/artifacts/clusters/bootstrap/logs/capz-system/capz-controller-manager/capz-controller-manager-5b8ff65d59-xjhpg/manager.log)). Could you try creating an AKS cluster with Tilt to see if this repros?

```
E1102 19:42:51.340297       1 controller.go:324]  "msg"="Reconciler error" "error"="error creating AzureManagedMachinePool capz-e2e-9rls4u/capz-e2e-9rls4u-aks-pool1: failed to reconcile machine pool pool1: failed to create or update resource capz-e2e-9rls4u-aks/pool1 (service: agentpools): PUT https://management.azure.com/subscriptions/===REDACTED===/resourceGroups/capz-e2e-9rls4u-aks/providers/Microsoft.ContainerService/managedClusters/capz-e2e-9rls4u-aks/agentPools/pool1\n--------------------------------------------------------------------------------\nRESPONSE 409: 409 Conflict\nERROR CODE: OperationNotAllowed\n--------------------------------------------------------------------------------\n{\n  \"code\": \"OperationNotAllowed\",\n  \"details\": null,\n  \"message\": \"Operation is not allowed: Another operation (Updating) is in progress, please wait for it to finish before starting a new operation. See https://aka.ms/aks-pending-operation for more details\",\n  \"subcode\": \"\"\n}\n--------------------------------------------------------------------------------\n" "AzureManagedMachinePool"={"name":"capz-e2e-9rls4u-aks-pool1","namespace":"capz-e2e-9rls4u"} "controller"="azuremanagedmachinepool" "controllerGroup"="infrastructure.cluster.x-k8s.io" "controllerKind"="AzureManagedMachinePool" "name"="capz-e2e-9rls4u-aks-pool1" "namespace"="capz-e2e-9rls4u" "reconcileID"="9794feef-dc25-4798-9850-5f73e97e041b"
```

api/v1beta1/azuremanagedcontrolplane_types.go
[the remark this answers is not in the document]
LochanRn: added defaulting for NodeOsImage

dtzar:
@LochanRn Thank you for your contribution.  We'd like to get this in the release coming out very soon.  Could you please just fix the merge conflict? Then I think we're good to merge.

LochanRn:
@dtzar done :)

azure/services/managedclusters/spec.go
```diff
@@ -125,8 +127,20 @@ type ManagedClusterSpec struct {
...
 	DisableLocalAccounts *bool
+
+	// AutoUpgradeProfile defines auto upgrade configuration.
+	AutoUpgradeProfile *ManagedClusterAutoUpgradeProfile
+}
+
+// ManagedClusterAutoUpgradeProfile auto upgrade profile for a managed cluster.
+type ManagedClusterAutoUpgradeProfile struct {
+	// NodeOSUpgradeChannel is a manner in which the OS on your nodes is updated. The default is NodeImage. Possible values include: NodeImage,Unmanaged,None
+	NodeOSUpgradeChannel *infrav1.NodeOSUpgradeChannel
```
Jont828, line 141: Nit: please fix spacing between declarations here

util/versions/version.go
```diff
@@ -0,0 +1,41 @@
...
+	v1 := a
+	if !strings.HasPrefix(a, "v") {
+		v1 = "v" + a
+	}
+	v2 := b
+	if !strings.HasPrefix(b, "v") {
+		v2 = "v" + b
+	}
+
+	comp := semver.Compare(v1, v2)
```
Jont828, line 36: Nit: 
```suggestion
	if comp := semver.Compare(v1, v2); comp < 0 {
	    return b
	}
```

Jont828 (commented):
LGTM pending small changes

azure/services/managedclusters/spec.go
LochanRn: ack

util/versions/version.go
LochanRn: @Jont828  What is the impact on updating the code the way it is expected here ?

Why should this be followed over the way I have implemented ??
I am only trying to understand the thought process behind this review comment.

Jont828: The difference is that in the latter case, the variable `comp` is scoped only within the if statement. It doesn't make a huge difference, but technically it's a bit cleaner since the variable is only declared in the scope it's used, i.e. within the if block rather than the whole function. That being said, it's fine if you want to leave it as is.

Jont828:
Nice work!

util/versions/version.go
LochanRn: Thanks a lot for the review and the explanation. I felt it does not make a difference but addressed it anyway.

LochanRn:
> Nice work!

Thank you :)

jackfrancis:
@LochanRn I have rebased this PR on top of latest main to help accelerate merging prior to the upcoming release. If you're willing to grant my github user access to push to your fork + branch, I can do so.

CecileRobertMichon:
> @LochanRn I have rebased this PR on top of latest main to help accelerate merging prior to the upcoming release. If you're willing to grant my github user access to push to your fork + branch, I can do so.

You can do this by just checking the box on the PR bottom right below participants:

Screenshot 2023-11-15 at 3 43 23 PM

azure/services/managedclusters/spec.go
```diff
@@ -125,8 +127,21 @@ type ManagedClusterSpec struct {
...
```
jackfrancis, line 141: Note that the azure-sdk-for-go exposes a `NodeOSUpgradeChannel` property in the upgrade profile:

- https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/resourcemanager/containerservice/armcontainerservice/models.go#L1148-L1154

The ASO version of this upgrade profile configuration does not seem to expose that:

-  https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/resourcemanager/containerservice/armcontainerservice/models.go#L1148-L1154

LochanRn:
> > @LochanRn I have rebased this PR on top of latest main to help accelerate merging prior to the upcoming release. If you're willing to grant my github user access to push to your fork + branch, I can do so.
> 
> You can do this by just checking the box on the PR bottom right below participants:
> 
> Screenshot 2023-11-15 at 3 43 23 PM

Not sure if I have the permission to do so, because the repo is managed by the org.

azure/services/managedclusters/spec.go
LochanRn: Ohhhh, so we are migrating from sdk-v4 to ASO api's ?

LochanRn:
> @LochanRn I have rebased this PR on top of latest main to help accelerate merging prior to the upcoming release. If you're willing to grant my github user access to push to your fork + branch, I can do so.

Let me try to rebase the PR :)

dtzar:
I believe @jackfrancis may have already done the rebase.  We didn't deem it very nice to have to rebase considering the significant code swap over to ASO from Go SDK.  Also, since ASO doesn't have this field yet (they plan to have it in 2.4.1), it isn't going to make it for 1.12 CAPZ.  Sorry!

LochanRn:
> I believe @jackfrancis may have already done the rebase. We didn't deem it very nice to have to rebase considering the significant code swap over to ASO from Go SDK. Also, since we discovered ASO doesn't have this field yet (they plan to have it in 2.4.1), it isn't going to make it for 1.12 CAPZ. Sorry!

That's okay... !

LochanRn:
> I believe @jackfrancis may have already done the rebase. We didn't deem it very nice to have to rebase considering the significant code swap over to ASO from Go SDK. Also, since we discovered ASO doesn't have this field yet (they plan to have it in 2.4.1), it isn't going to make it for 1.12 CAPZ. Sorry!

@dtzar @jackfrancis  The other concern I have is I have a few other PR's that I have opened and they are using azure-sdk and not the ASO flow, will they also be affected/impacted. Even those PR's will have to rebased ?

Also It would be nice for the reviewers to inform the developers about the impact. The current PR has been opened for almost a month and there was a good amount of review work that has gone into this PR. 

Time and effort can be saved if the developer is informed prior :)

LochanRn:
@jackfrancis I have rebased the PR.  

Should I remove the entire spec for NodeOSUpgradeChannel ??
Or can we leave it as it is and use the same code in the future when ASO provides support ?

Jont828:
@jackfrancis Thanks for rebasing this while I was out on PTO! @LochanRn It looks like you need to rebase again btw.

jackfrancis:
Hi @LochanRn, thanks for your patience here. Agree the process for accepting these PRs is not ideal. This should be smoother going forward as the ASO surface area is not largely stabilized.

@nojnhuh are you able to answer this question?

> Should I remove the entire spec for NodeOSUpgradeChannel ??
Or can we leave it as it is and use the same code in the future when ASO provides support ?

LochanRn:
@Jont828  Rebased the PR 
@jackfrancis @nojnhuh Removed NodeOSUpgrade property.

LochanRn:
@Jont828 PTAL :)

Jont828 (commented):
I think this is looking good! Just a nit on my end, and once you resolve @nojnhuh's comments I think we're good to go.

dtzar:
@LochanRn - the milestone is in 2 weeks, do you think you'll be able to finish addressing the comments?

azure/scope/managedmachinepool.go
```diff
@@ -343,3 +340,24 @@ func (s *ManagedMachinePoolScope) GetCAPIMachinePoolAnnotation(key string) (succ
...
+
+func getManagedMachinePoolVersion(managedControlPlane *infrav1.AzureManagedControlPlane, machinePool *expv1.MachinePool) *string {
+	var v, av string
+	if machinePool != nil {
+		v = ptr.Deref(machinePool.Spec.Template.Spec.Version, "")
+	}
+	if managedControlPlane != nil {
+		av = managedControlPlane.Status.AutoUpgradeVersion
+	}
+	higherVersion := versions.GetHigherK8sVersion(v, av)
```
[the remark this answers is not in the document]
LochanRn: Users are allowed to upgrade AKS versions to higher versions beyond the control plane version when auto upgrade is set. 

for example. if auto upgrade version is set to stable and the stable auto upgrade version is 1.26.x users are allowed to upgrade the AKS cluster to version 1.27.x

azure/services/agentpools/spec.go
```diff
@@ -201,6 +216,10 @@ func (s *AgentPoolSpec) Parameters(ctx context.Context, existing *asocontainerse
 	agentPool.Spec.EnableFIPS = s.EnableFIPS
 	agentPool.Spec.EnableEncryptionAtHost = s.EnableEncryptionAtHost
 
+	if kubernetesVersion := s.GetManagedMachinePoolVersion(existing); kubernetesVersion != nil {
+		agentPool.Spec.OrchestratorVersion = kubernetesVersion
+	}
```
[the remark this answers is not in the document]
LochanRn: The only reason why i kept that check is, AKS allows creation of kubernetes clusters without a version. If no version is specified it picks the stable k8s version. In future if we would want to support it we will have to make this check and it would not be redundant. But yes at this point of time it is redundant as we don't allow k8s version to be nil.

azure/services/managedclusters/managedclusters.go
```diff
@@ -105,17 +108,22 @@ func postCreateOrUpdateResourceHook(ctx context.Context, scope ManagedClusterSco
 		})
 	}
 
+	if scope.IsManagedVersionUpgrade() && managedCluster.Spec.KubernetesVersion != nil {
+		kubernetesVersion := fmt.Sprintf("v%s", *managedCluster.Spec.KubernetesVersion)
```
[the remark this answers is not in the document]
LochanRn: When auto upgrades are in place, we do not propagate the new version to the spec of the ASO resource. We should be looking this in the status.

azure/services/managedclusters/managedclusters.go
```diff
@@ -105,17 +108,22 @@ func postCreateOrUpdateResourceHook(ctx context.Context, scope ManagedClusterSco
...
+		scope.SetAutoUpgradeVersionStatus(kubernetesVersion)
+	}
```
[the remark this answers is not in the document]
LochanRn: A version field already exists in the status hence i am updating the version field as well everytime which could serve as a general purpose version.  Autoupgrade version is only updated when the autoupgrade channels are set indicating the cluster's version was auto upgraded by AKS. 

If you think the autoupgrade version is redundant let me know. Only reason i want to have it is to allow the user to know the cluster has been auto upgraded to the specific version.

LochanRn:
@nojnhuh addressed most of the comments a few of them I am open to suggestions :).

azure/services/managedclusters/managedclusters.go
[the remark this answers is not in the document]
nojnhuh: Should we update this then?

util/versions/version_test.go
```diff
@@ -0,0 +1,106 @@
...
+		{
+			name:   "a is greater than b",
+			a:      "v1.17.8",
+			b:      "v1.18.8",
+			output: "v1.18.8",
+		},
+		{
+			name:   "b is greater than a",
+			a:      "v1.18.9",
+			b:      "v1.18.8",
```
nojnhuh, line 41: Should these names be swapped?

api/v1beta1/azuremanagedcontrolplane_types.go
```diff
@@ -33,6 +33,38 @@ const (
...
+const (
+	// UpgradeChannelNodeImage automatically upgrades the node image to the latest version available.
+	// Consider using nodeOSUpgradeChannel instead as that allows you to configure node OS patching separate from Kubernetes version patching.
+	UpgradeChannelNodeImage UpgradeChannel = "node-image"
+
+	// UpgradeChannelNone disables auto-upgrades and keeps the cluster at its current version of Kubernetes.
+	UpgradeChannelNone UpgradeChannel = "none"
+
+	// UpgradeChannelPatch automatically upgrade the cluster to the latest supported patch version when it becomes available
+	// while keeping the minor version the same. For example, if a cluster is running version 1.17.7 and versions 1.17.9, 1.18.4,
```
Jont828, line 51:
```suggestion
	// while keeping the minor version the same. For example, if a cluster is running version 1.17.7 while versions 1.17.9, 1.18.4,
```

api/v1beta1/azuremanagedcontrolplane_types.go
```diff
@@ -33,6 +33,38 @@ const (
...
+	// UpgradeChannelRapid automatically upgrade the cluster to the latest supported patch release on the latest supported minor
+	// version. In cases where the cluster is at a version of Kubernetes that is at an N-2 minor version where N is the latest
+	// supported minor version, the cluster first upgrades to the latest supported patch version on N-1 minor version. For example,
+	// if a cluster is running version 1.17.7 and versions 1.17.9, 1.18.4, 1.18.6, and 1.19.1 are available, your cluster first
+	// is upgraded to 1.18.6, then is upgraded to 1.19.1.
+	UpgradeChannelRapid UpgradeChannel = "rapid"
+
+	// UpgradeChannelStable automatically upgrade the cluster to the latest supported patch release on minor version N-1, where
+	// N is the latest supported minor version. For example, if a cluster is running version 1.17.7 and versions 1.17.9, 1.18.4,
+	// 1.18.6, and 1.19.1 are available, your cluster is upgraded to 1.18.6.
```
Jont828, line 64:
```suggestion
	// 1.18.6, and 1.19.1 are available, the cluster will be upgraded to 1.18.6.
```

api/v1beta1/azuremanagedcontrolplane_webhook.go
```diff
@@ -499,6 +507,42 @@ func validateManagedClusterNetwork(cli client.Client, labels map[string]string,
...
+	if hv := versions.GetHigherK8sVersion(m.Spec.Version, old.Spec.Version); hv != m.Spec.Version {
+		allErrs = append(allErrs, field.Invalid(field.NewPath("Spec", "Version"),
+			m.Spec.Version, "field version cannot be downgraded"),
+		)
+	}
+
+	if old.Status.AutoUpgradeVersion != "" && m.Spec.Version != old.Spec.Version {
+		if hv := versions.GetHigherK8sVersion(m.Spec.Version, old.Status.AutoUpgradeVersion); hv != m.Spec.Version {
+			allErrs = append(allErrs, field.Invalid(field.NewPath("Spec", "Version"),
+				m.Spec.Version, "version is auto-upgraded to "+old.Status.AutoUpgradeVersion+",cannot be downgraded"),
```
Jont828, line 539:
```suggestion
				m.Spec.Version, "version is auto-upgraded to "+old.Status.AutoUpgradeVersion+", cannot be downgraded"),
```

api/v1beta1/types_class.go
```diff
@@ -211,6 +211,17 @@ type AzureManagedControlPlaneClassSpec struct {
 	// DisableLocalAccounts disables getting static credentials for this cluster when set. Expected to only be used for AAD clusters.
 	// +optional
 	DisableLocalAccounts *bool `json:"disableLocalAccounts,omitempty"`
+	// AutoUpgradeProfile defines the auto upgrade configuration.
+	// +optional
+	AutoUpgradeProfile *ManagedClusterAutoUpgradeProfile `json:"autoUpgradeProfile,omitempty"`
+}
+
+// ManagedClusterAutoUpgradeProfile defines Auto upgrade profile for a managed cluster.
```
Jont828, line 219:
```suggestion
// ManagedClusterAutoUpgradeProfile defines the auto upgrade profile for a managed cluster.
```

api/v1beta1/types_class.go
```diff
@@ -211,6 +211,17 @@ type AzureManagedControlPlaneClassSpec struct {
...
+type ManagedClusterAutoUpgradeProfile struct {
+	// UpgradeChannel upgrade channel for auto upgrade.
```
Jont828, line 221:
```suggestion
	// UpgradeChannel determines the type of upgrade channel for automatically upgrading the cluster.
```

azure/scope/managedcontrolplane_test.go
```diff
@@ -1453,3 +1453,95 @@ func TestManagedControlPlaneScope_PrivateEndpointSpecs(t *testing.T) {
 		})
 	}
 }
+
+func TestManagedControlPlaneScope_AutoUpgradeProfile(t *testing.T) {
+	cases := []struct {
+		Name     string
```
Jont828, line 1459:
```suggestion
		name     string
		input    ManagedControlPlaneScopeParams
		expected *managedclusters.ManagedClusterAutoUpgradeProfile
```

Nit: these don't need to be exported

azure/scope/managedmachinepool_test.go
```diff
@@ -915,6 +915,122 @@ func TestManagedMachinePoolScope_KubeletDiskType(t *testing.T) {
 	}
 }
 
+func Test_getManagedMachinePoolVersion(t *testing.T) {
+	cases := []struct {
+		Name                string
```
Jont828, line 920:
```suggestion
		name                string
```

azure/scope/managedmachinepool_test.go
```diff
@@ -915,6 +915,122 @@ func TestManagedMachinePoolScope_KubeletDiskType(t *testing.T) {
...
+		managedControlPlane *infrav1.AzureManagedControlPlane
+		machinePool         *expv1.MachinePool
+		Expected            *string
```
Jont828, line 923:
```suggestion
		expected            *string
```

azure/services/agentpools/spec.go
```diff
@@ -164,6 +165,21 @@ func (s *AgentPoolSpec) ResourceRef() *asocontainerservicev1.ManagedClustersAgen
 	}
 }
 
+// getManagedMachinePoolVersion gets the desired managed k8s version.
+// If autoupgrade channels is set to patch, stable or rapid, clusters can be upgraded to higher version by AKS.
```
Jont828, line 169:
```suggestion
// If the auto upgrade channel is set to patch, stable or rapid, clusters can be upgraded to a higher version by AKS.
```

azure/services/agentpools/spec.go
```diff
@@ -164,6 +165,21 @@ func (s *AgentPoolSpec) ResourceRef() *asocontainerservicev1.ManagedClustersAgen
...
+// If autoupgrade is triggered, existing kubernetes version will be higher than the user desired kubernetes version.
```
Jont828, line 170:
```suggestion
// If auto upgrade is triggered, the existing Kubernetes version will be higher than the user's desired Kubernetes version.
```

azure/services/agentpools/spec.go
```diff
@@ -164,6 +165,21 @@ func (s *AgentPoolSpec) ResourceRef() *asocontainerservicev1.ManagedClustersAgen
 	}
 }
 
+// getManagedMachinePoolVersion gets the desired managed k8s version.
```
Jont828, line 168:
```suggestion
// getManagedMachinePoolVersion gets the desired managed Kubernetes version.
```

azure/services/managedclusters/spec.go
```diff
@@ -292,6 +302,17 @@ func buildAutoScalerProfile(autoScalerProfile *AutoScalerProfile) *asocontainers
 	return mcAutoScalerProfile
 }
 
+// getManagedClusterVersion gets the desired managed k8s version.
+// If autoupgrade channels is set to patch, stable or rapid, clusters can be upgraded to higher version by AKS.
```
Jont828, line 306: Do you think we can reuse the same function in `agentpools/spec.go`?

azure/services/managedclusters/managedclusters.go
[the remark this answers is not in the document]
LochanRn: @nojnhuh good question. 

I think this PR has an issue and i am a bit confused about how to solve it. 
This PR was good prior to ASO getting merged and when CAPZ was making an api call to create the AKS resource.

When auto upgrade is set AKS manages the version. Upgrades ideally have to be completely managed by AKS.

But AKS also provides the users options to upgrade AKS to higher version than the existing one if required. 

So now there are 2 sources which had the control for updating AKS version.
1.) AKS itself based on the upgrade channel
2.) The user if he wants the version to be greater than the version chosen based on the upgrade channel. 

So what i did to handle both the cases is, to pick the higher version out of the 2 i.e version managed by AKS vs version managed by user. 

I was able to get the version managed by AKS from the object returned by the api call and always set it to the autupgradeversion in status. 

Now the problem is I don't know the version set by AKS. Or in other words how do i get the version managed by AKS ??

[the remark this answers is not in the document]
LochanRn: Also I don't know if allowing users to update a version after setting an upgrade channel is a bug in AKS itself. 

Because that kinda conflicts with the whole idea of having an upgrade channel.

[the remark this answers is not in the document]
nojnhuh: > Also I don't know if allowing users to update a version after setting an upgrade channel is a bug in AKS itself.
> 
> Because that kinda conflicts with the whole idea of having an upgrade channel.

I could be convinced that that should be allowed to enable users to have a more conservative auto upgrade policy while allowing larger upgrades to be done manually. If you find docs that imply users shouldn't touch the version at all when auto upgrades are enabled then that seems worth bringing up to AKS.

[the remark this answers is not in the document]
LochanRn: > In this case though it makes sense to me to populate the AzureManagedControlPlane's status based on the ASO ManagedCluster's status.

Right will make that update and test it, because with the existing code after i test it does not work as expected.

> If you find docs that imply users shouldn't touch the version at all when auto upgrades are enabled then that seems worth bringing up to AKS.

I don't see the anywhere in the docs that say we shouldn't be touching the version when an auto upgrade is configured for the AKS cluster. It was just a thought in my head, that we are doing two conflicting things at the same time. Like on 1 hand we say auto upgrade and on the other hand we say i would upgrade.

Jont828:
Looks like you got a test failure with the AKS e2e test as well.

azure/scope/managedmachinepool.go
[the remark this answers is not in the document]
LochanRn: We don't use it anywhere else in managed clusters.

azure/services/managedclusters/spec.go
LochanRn: Agentpools/spec.go requires asocontainerservicev1.ManagedClustersAgentPool as an input.

azure/services/managedclusters/managedclusters.go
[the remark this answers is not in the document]
LochanRn: @nojnhuh Created a cluster with 1.26.6 k8s version and upgrade channel was set to stable. The cluster got auto upgraded to 1.27.7. Digged through the logs of ASO 

We see the logs do print the version as 1.27.7

`I0115 19:17:09.559965       1 azure_generic_arm_reconciler_instance.go:579] "msg"="Got ARM status" "azureName"="rnl-capz-aks-auto-upgrade" "logger"="controllers.ManagedClusterController" "name"="rnl-capz-aks-auto-upgrade" "namespace"="default" "status"="{\"id\":\"/subscriptions/<<<<subscription-id>>>>/resourcegroups/lochan-rg/providers/Microsoft.ContainerService/managedClusters/rnl-capz-aks-auto-upgrade\",\"identity\":{\"delegatedResources\":null,\"principalId\":\"<<<<principal-id>>>>\",\"tenantId\":\"<<<<tenant-id>>>>\",\"type\":\"SystemAssigned\",\"userAssignedIdentities\":null},\"location\":\"eastus\",\"name\":\"rnl-capz-aks-auto-upgrade\",\"properties\":{\"addonProfiles\":null,\"agentPoolProfiles\":[{\"availabilityZones\":null,\"count\":3,\"currentOrchestratorVersion\":\"1.27.7\",\"enableAutoScaling\":false,\"enableFIPS\":false,\"kubeletDiskType\":\"OS\",\"maxPods\":30,\"mode\":\"User\",\"name\":\"pool1\",\"nodeImageVersion\":\"AKSUbuntu-2204gen2containerd-202312.06.0\",\"nodeLabels\":null,\"nodeTaints\":null,\"orchestratorVersion\":\"1.27.7\",\"osDiskSizeGB\":128,\"osDiskType\":\"Managed\",\"osSKU\":\"Ubuntu\",\"osType\":\"Linux\",\"powerState\":{\"code\":\"Running\"},\"provisioningState\":\"Succeeded\",\"scaleDownMode\":\"Delete\",\"tags\":null,\"type\":\"VirtualMachineScaleSets\",\"vmSize\":\"Standard_D2s_v3\",\"vnetSubnetID\":\"/subscriptions/<<<<subscription-id>>>>/resourceGroups/lochan-rg/providers/Microsoft.Network/virtualNetworks/rnl-capz-aks-auto-upgrade/subnets/rnl-capz-aks-auto-upgrade\"},{\"availabilityZones\":null,\"count\":3,\"currentOrchestratorVersion\":\"1.27.7\",\"enableAutoScaling\":false,\"enableFIPS\":false,\"kubeletDiskType\":\"OS\",\"maxPods\":30,\"mode\":\"System\",\"name\":\"pool0\",\"nodeImageVersion\":\"AKSUbuntu-2204gen2containerd-202312.06.0\",\"nodeLabels\":null,\"nodeTaints\":null,\"orchestratorVersion\":\"1.27.7\",\"osDiskSizeGB\":128,\"osDiskType\":\"Managed\",\"osSKU\":\"Ubuntu\",\"osType\":\"Linux\",\"powerState\":{\"code\":\"Running\"},\"provisioningState\":\"Succeeded\",\"scaleDownMode\":\"Delete\",\"tags\":null,\"type\":\"VirtualMachineScaleSets\",\"vmSize\":\"Standard_D2s_v3\",\"vnetSubnetID\":\"/subscriptions/<<<<subscription-id>>>>/resourceGroups/lochan-rg/providers/Microsoft.Network/virtualNetworks/rnl-capz-aks-auto-upgrade/subnets/rnl-capz-aks-auto-upgrade\"}],\"autoUpgradeProfile\":{\"nodeOSUpgradeChannel\":\"NodeImage\",\"upgradeChannel\":\"stable\"},\"azurePortalFQDN\":\"rnl-capz-aks-auto-upgrade-n55o6664.portal.hcp.eastus.azmk8s.io\",\"currentKubernetesVersion\":\"1.27.7\",\"dnsPrefix\":\"rnl-capz-aks-auto-upgrade\",\"enableRBAC\":true,\"fqdn\":\"rnl-capz-aks-auto-upgrade-n55o6664.hcp.eastus.azmk8s.io\",\"identityProfile\":{\"kubeletidentity\":{\"clientId\":\"<<<<client-id>>>>\",\"objectId\":\"<<<<object-id>>>>\",\"resourceId\":\"/subscriptions/<<<<subscription-id>>>>/resourcegroups/MC_lochan-rg_rnl-capz-aks-auto-upgrade_eastus/providers/Microsoft.ManagedIdentity/userAssignedIdentities/rnl-capz-aks-auto-upgrade-agentpool\"}},\"kubernetesVersion\":\"1.27.7\",\"linuxProfile\":{\"adminUsername\":\"azureuser\",\"ssh\":{\"publicKeys\":[{\"keyData\":\"<ssh-key>\"}]}},\"maxAgentPools\":100,\"networkProfile\":{\"dnsServiceIP\":\"192.168.0.10\",\"ipFamilies\":[\"IPv4\"],\"loadBalancerProfile\":{\"effectiveOutboundIPs\":[{\"id\":\"/subscriptions/<<<<subscription-id>>>>/resourceGroups/MC_lochan-rg_rnl-capz-aks-auto-upgrade_eastus/providers/Microsoft.Network/publicIPAddresses/<<<<publicIPId>>>>\"}],\"managedOutboundIPs\":{\"count\":1}},\"loadBalancerSku\":\"Standard\",\"networkDataplane\":\"azure\",\"networkPlugin\":\"azure\",\"outboundType\":\"loadBalancer\",\"podCidrs\":null,\"serviceCidr\":\"192.168.0.0/16\",\"serviceCidrs\":[\"192.168.0.0/16\"]},\"nodeResourceGroup\":\"MC_lochan-rg_rnl-capz-aks-auto-upgrade_eastus\",\"oidcIssuerProfile\":{\"enabled\":false},\"powerState\":{\"code\":\"Running\"},\"privateLinkResources\":null,\"provisioningState\":\"Succeeded\",\"resourceUID\":\"<<<<resouce-id>>>>\",\"securityProfile\":{},\"servicePrincipalProfile\":{\"clientId\":\"msi\"},\"storageProfile\":{\"diskCSIDriver\":{\"enabled\":true},\"fileCSIDriver\":{\"enabled\":true},\"snapshotController\":{\"enabled\":true}},\"supportPlan\":\"KubernetesOfficial\",\"windowsProfile\":{\"adminUsername\":\"azureuser\",\"enableCSIProxy\":true},\"workloadAutoScalerProfile\":{}},\"sku\":{\"name\":\"Base\",\"tier\":\"Free\"},\"tags\":{\"Name\":\"rnl-capz-aks-auto-upgrade\",\"sigs.k8s.io_cluster-api-provider-azure_cluster_rnl-capz-aks-auto-upgrade\":\"owned\",\"sigs.k8s.io_cluster-api-provider-azure_role\":\"common\"},\"type\":\"Microsoft.ContainerService/ManagedClusters\"}"`



But the ManagedCluster object is not getting set/updated with version 1.27.7 it is still at 1.26.6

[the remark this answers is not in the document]
LochanRn: Looks like some bug in the ASO.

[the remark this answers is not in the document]
nojnhuh: @LochanRn Hm, that's not I would expect either. Might be best to follow up in an ASO issue (https://github.com/azure/azure-service-operator) or on k8s Slack (#azure-service-operator).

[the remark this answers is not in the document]
nojnhuh: I started a slack thread: https://kubernetes.slack.com/archives/C046DEVLAQM/p1705423946331469

nojnhuh:
Since this was approved a long time ago and there have been lots of changes since then:
/approve cancel

nojnhuh:
> Since this was approved a long time ago and there have been lots of changes since then:
> /approve cancel

Or maybe this will work: 
/remove-approve

@LochanRn I don't mean to suggest this PR is being held up for any reason, I'm just trying to prevent mistakes in the review process to make sure we don't merge this too soon since we don't often deal with approved but un-LGTM'd PRs.

api/v1beta1/types_class.go
```diff
@@ -211,6 +211,17 @@ type AzureManagedControlPlaneClassSpec struct {
 	// DisableLocalAccounts disables getting static credentials for this cluster when set. Expected to only be used for AAD clusters.
 	// +optional
 	DisableLocalAccounts *bool `json:"disableLocalAccounts,omitempty"`
+	// AutoUpgradeProfile defines the auto upgrade configuration.
```
Jont828, line 214:
```suggestion

	// AutoUpgradeProfile defines the auto upgrade configuration.
```

Jont828 (commented):
Looks fine except for a spacing thing and a test failure. It seems like it's due to a linting issue in `azure/scope/managedcontrolplane_test.go`. Should be good to LGTM after that!

api/v1beta1/azuremanagedcontrolplane_webhook.go
```diff
@@ -256,6 +256,14 @@ func (mw *azureManagedControlPlaneWebhook) ValidateUpdate(ctx context.Context, o
 		allErrs = append(allErrs, errs...)
 	}
 
+	if errs := m.validateAutoUpgradeProfile(old); len(errs) > 0 {
+		allErrs = append(allErrs, errs...)
+	}
+
+	if errs := m.validateK8sVersionUpdate(old); len(errs) > 0 {
+		allErrs = append(allErrs, errs...)
+	}
```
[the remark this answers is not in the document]
LochanRn: Done

LochanRn:
@nojnhuh @Jont828 How do i fix the above errors ? 

I ran make test and make generate locally i don't see anymore changes reflecting !

nojnhuh:
> @nojnhuh @Jont828 How do i fix the above errors ?
> 
> I ran make test and make generate locally i don't see anymore changes reflecting !

Which errors do you mean? The only errors I see are Netlify and I think it's safe to assume these changes don't affect that.

azure/scope/managedcontrolplane_test.go
```diff
@@ -1533,3 +1534,95 @@ func TestManagedControlPlaneScope_AKSExtensionSpecs(t *testing.T) {
...
+	cases := []struct {
+		name     string
+		input    ManagedControlPlaneScopeParams
+		expected *managedclusters.ManagedClusterAutoUpgradeProfile
+	}{
+		{
+			name: "Without AutoUpgradeProfile",
+			input: ManagedControlPlaneScopeParams{
+				AzureClients: AzureClients{
+					Authorizer: autorest.NullAuthorizer{},
```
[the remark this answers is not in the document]
LochanRn: Done sir. 

I hope there are no more review comments and this PR is good to go :)

nojnhuh (commented):
/lgtm

I'll leave the hold to keep this from merging so @Jont828 has a chance to give this one more look.

Jont828:
Thank you for your patience on this - I know you've had to rebase this quite a few times. This PR has come a long way!

/lgtm

LochanRn:
> Thank you for your patience on this - I know you've had to rebase this quite a few times. This PR has come a long way!

Thank you for all the valuable review comments :)

PR #4139, opened 2023-10-16
Replace go-autorest GetSettingsFromEnvironment

### LINKED ISSUE #2974: Azure/go-autorest will soon be unsupported

mboersma:
**What type of PR is this?**

/kind cleanup

Removes one usage of the deprecated go-autorest package that can easily be replaced by a local function.

Refs #2974

- [ ] cherry-pick candidate

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/azure_logcollector.go
+++ b/test/e2e/azure_logcollector.go
@@ -30,7 +30,6 @@ import (
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute/v5"
-	"github.com/Azure/go-autorest/autorest/azure/auth"
 	"github.com/pkg/errors"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
@@ -416,16 +415,17 @@ func collectVMBootLog(ctx context.Context, am *infrav1.AzureMachine, outputPath 
 		return errors.Wrap(err, "failed to parse resource id")
 	}
 
-	settings, err := auth.GetSettingsFromEnvironment()
-	if err != nil {
-		return errors.Wrap(err, "failed to get settings from environment")
+	subscriptionID := os.Getenv("AZURE_SUBSCRIPTION_ID")
+	if subscriptionID == "" {
+		return errors.New("AZURE_SUBSCRIPTION_ID is not set")
 	}
 
 	cred, err := azidentity.NewDefaultAzureCredential(nil)
 	if err != nil {
 		return errors.Wrap(err, "failed to get default azure credential")
 	}
-	vmClient, err := armcompute.NewVirtualMachinesClient(settings.GetSubscriptionID(), cred, nil)
+
+	vmClient, err := armcompute.NewVirtualMachinesClient(subscriptionID, cred, nil)
 	if err != nil {
 		return errors.Wrap(err, "failed to create virtual machines client")
 	}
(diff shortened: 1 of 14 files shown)


test/e2e/azure_logcollector.go
```diff
@@ -416,16 +415,17 @@ func collectVMBootLog(ctx context.Context, am *infrav1.AzureMachine, outputPath
 		return errors.Wrap(err, "failed to parse resource id")
 	}
 
-	settings, err := auth.GetSettingsFromEnvironment()
-	if err != nil {
-		return errors.Wrap(err, "failed to get settings from environment")
+	subscriptionID := os.Getenv("AZURE_SUBSCRIPTION_ID")
+	if subscriptionID == "" {
+		return errors.New("AZURE_SUBSCRIPTION_ID is not set")
```
nojnhuh, line 420: I wonder if it might be worth having two overlapping flavors of the helper, where one looks like what happens here:
```go
func getSubAndError() (string, error) {...}
```

and a separate one that asserts the above err is nil:
```go
func assertSub() string {
	sub, err := getSubAndErr()
	return sub
}
```

(but with better names)

test/e2e/helpers.go
```diff
@@ -946,3 +936,9 @@ func CopyConfigMap(ctx context.Context, input clusterctl.ApplyCustomClusterTempl
 		}
 	}, input.WaitForControlPlaneIntervals...).Should(Succeed())
 }
+
+func getSubscriptionID() string {
+	subscriptionID := os.Getenv("AZURE_SUBSCRIPTION_ID")
+	Expect(subscriptionID).NotTo(BeEmpty())
```
nojnhuh, line 942: I don't think it's an issue, but if this were ever called in a context using a non-default `Gomega` handle then this might misbehave, like in an `Eventually` invoking a `func(g Gomega)`. It might be safer but more verbose then to pass a `Gomega` here as a parameter:
```suggestion
func getSubscriptionID(g Gomega) string {
	subscriptionID := os.Getenv("AZURE_SUBSCRIPTION_ID")
	g.Expect(subscriptionID).NotTo(BeEmpty())
```

I think the default used by plain `Expect` is called `Default`.

nojnhuh (commented):
Left a couple annoying comments, definitely not blocking.
/lgtm

CecileRobertMichon:
> Fixes #2974

is this the very last go-autorest usage?

mboersma:
> is this the very last go-autorest usage?

No, there's still more to do, I mean to say `Refs #2974`. Updated now.

test/e2e/azure_logcollector.go
mboersma: In all the cases where this is called, it seems to me it will be an error if `AZURE_SUBSCRIPTION_ID == ""`. And there are no other errors to pass on or to create. So I'm not sure why we need both versions; I just went for the equivalent of `assertSub()`.

test/e2e/helpers.go
mboersma: >  this might misbehave, like in an `Eventually` invoking a `func(g Gomega)`

As in `CopyConfigMap`, the function ten lines above this one. 😄

I agree that's safer–I'll make that change.

nojnhuh (commented):
/lgtm

CecileRobertMichon (commented):
/lgtm
/approve

PR #4143, opened 2023-10-16
Convert bastionhosts to ASO

### LINKED ISSUE #4085: ASO service: bastionhosts

willie-yao:
**What type of PR is this?**
/kind cleanup

This PR converts the BastionHosts service to ASO

Fixes #4085 

- [ ] cherry-pick candidate

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Convert bastionhosts to ASO
```


azure/scope/cluster.go
```diff
@@ -539,13 +539,14 @@ func (s *ClusterScope) AzureBastion() *infrav1.AzureBastion {
 }
 
 // AzureBastionSpec returns the bastion spec.
-func (s *ClusterScope) AzureBastionSpec() azure.ResourceSpecGetter {
+func (s *ClusterScope) AzureBastionSpec() azure.ASOResourceSpecGetter[*asonetworkv1.BastionHost] {
```
Jont828, line 542: Is there a reason we want to return a list of pointers rather than just a list of objects, i.e. `azure.ASOResourceSpecGetter[asonetworkv1.BastionHost]`?

Jont828 (commented):
Looks good so far! Just a question, otherwise LGTM

azure/scope/cluster.go
willie-yao: It's a convention for implementing ASO resources with Generics, although I'm not sure about the specific reason. @nojnhuh might know!

https://gist.github.com/nojnhuh/a1e06bb6d98b0eacb51134480ea4acea#option-1-use-azureserviceasos-service

willie-yao:
> Looks good so far! Just a question, otherwise LGTM

Thanks for the review! I'm still trying to fix up some test errors, so it'll be on hold before I can fix it.

azure/scope/cluster.go
nojnhuh: The `[]` syntax is a little confusing here in that it's not for a list but for type parameters. We use the pointer type instead of values because all of the handy methods common to Kubernetes resource types in ASO's [`genruntime.MetaObject`](https://pkg.go.dev/github.com/Azure/azure-service-operator/v2@v2.3.0/pkg/genruntime#MetaObject) are implemented on the pointer type, e.g. https://pkg.go.dev/github.com/Azure/azure-service-operator/v2@v2.3.0/api/resources/v1api20200601#ResourceGroup.

willie-yao:
> Thanks for the review! I'm still trying to fix up some test errors, so it'll be on hold before I can fix it.
Fixed! @nojnhuh @Jont828

azure/services/bastionhosts/bastionhosts.go
```diff
@@ -17,88 +17,37 @@ limitations under the License.
...
 // New creates a new service.
-func New(scope BastionScope) (*Service, error) {
-	client, err := newClient(scope)
-	if err != nil {
-		return nil, err
+func New(scope BastionScope) *Service {
+	svc := aso.NewService[*asonetworkv1.BastionHost, BastionScope](serviceName, scope)
+	spec := scope.AzureBastionSpec()
+	if spec != nil {
+		svc.Specs = []azure.ASOResourceSpecGetter[*asonetworkv1.BastionHost]{spec}
```
nawazkh, line 46:
```suggestion
	svc.Specs := scope.AzureBastionSpec()
```

Since svc.Specs is not set to empty `[]azure.ASOResourceSpecGetter[*asonetworkv1.BastionHost]`, we might as well set `svc.Specs` to the output of `scope.AzureBastionSpec()`

azure/services/bastionhosts/spec.go
```diff
@@ -40,69 +41,64 @@ type AzureBastionSpec struct {
...
-	if existing != nil {
-		if _, ok := existing.(armnetwork.BastionHost); !ok {
-			return nil, errors.Errorf("%T is not an armnetwork.BastionHost", existing)
-		}
-		// bastion host already exists
-		return nil, nil
+// Parameters implements azure.ASOResourceSpecGetter.
+func (s *AzureBastionSpec) Parameters(ctx context.Context, existingBastionHost *asonetworkv1.BastionHost) (parameters *asonetworkv1.BastionHost, err error) {
+	if existingBastionHost != nil {
+		return existingBastionHost, nil
```
nawazkh, line 57: Not sure why we want to return `existingBastionHost` early. 
Since we want CAPZ's opinions to supersede user inputs (only on the fields CAPZ has an opinion about), we should instead have something like below at line 62.
```suggestion
	if existingBastionHost != nil {
		bastionHost.Spec = existingBastionHost.Spec
	}
```

azure/services/bastionhosts/spec.go
```diff
@@ -40,69 +41,64 @@ type AzureBastionSpec struct {
...
-							ID: &s.SubnetID,
-						},
-						PublicIPAddress: &armnetwork.SubResource{
-							ID: &s.PublicIPID,
-						},
-						PrivateIPAllocationMethod: ptr.To(armnetwork.IPAllocationMethodDynamic),
-					},
+	bastionHostIPConfigName := fmt.Sprintf("%s-%s", s.Name, "bastionIP")
+	bastionHost.Spec.AzureName = s.Name
+	bastionHost.Spec.Location = ptr.To(s.Location)
```
nawazkh, line 65: I am a little divided in here. Should this be 
```suggestion
	if s.Location != "" {
		bastionHost.Spec.Location = ptr.To(s.Location)
	}
```

In this way, the field is only set when `s.Location` is not empty.?

controllers/azurecluster_controller.go
```diff
@@ -115,6 +115,8 @@ func (acr *AzureClusterReconciler) SetupWithManager(ctx context.Context, mgr ctr
 // +kubebuilder:rbac:groups=resources.azure.com,resources=resourcegroups/status,verbs=get;list;watch
 // +kubebuilder:rbac:groups=network.azure.com,resources=natgateways,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=network.azure.com,resources=natgateways/status,verbs=get;list;watch
+// +kubebuilder:rbac:groups=network.azure.com,resources=bastionhosts,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=network.azure.com,resources=bastionhosts/status,verbs=get;list;watch
```
nawazkh, line 119: You could also club these annotations with the natgateways annotations defined above. 
Shall we be discretely listing the kubebuilder annotations per resource type (like it has been done here) or should we be clubbing it in one like for the same group type? (as demonstrated below)

I prefer the former existing implementation, pointing it out for discussion. 
```suggestion
// +kubebuilder:rbac:groups=network.azure.com,resources=natgateways;bastionhosts,verbs=get;list;watch;create;update;patch;delete
// +kubebuilder:rbac:groups=network.azure.com,resources=natgateways/status;bastionhosts/status,verbs=get;list;watch
```

nawazkh (commented):
Some thoughts and comments from first review.
Thanks for putting it together @willie-yao !

azure/services/bastionhosts/bastionhosts.go
willie-yao: I kept it this way because we would still need to wrap scope.AzureBastionSpec() in a []azure.ASOResourceSpecGetter, and creating a list with a nil value causes some issues.

azure/services/bastionhosts/spec.go
willie-yao: Won't this actually cause unecessary fields to be populated like an IpConfiguration with empty ARMIDs? https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/4143/files#diff-62fea3ceadec31e79fa0bae79a9972f0c16cb5cc274a934828520986acda0a46R81

azure/services/bastionhosts/bastionhosts.go
nawazkh: I apologize, I overlooked the fact that `scope.AzureBastionSpec()` returns a _single_ spec instead of a list of specs. This makes sense.

azure/services/bastionhosts/spec.go
```diff
@@ -40,69 +41,65 @@ type AzureBastionSpec struct {
...
-			return nil, errors.Errorf("%T is not an armnetwork.BastionHost", existing)
-		}
-		// bastion host already exists
-		return nil, nil
+// Parameters implements azure.ASOResourceSpecGetter.
+func (s *AzureBastionSpec) Parameters(ctx context.Context, existingBastionHost *asonetworkv1.BastionHost) (parameters *asonetworkv1.BastionHost, err error) {
+	bastionHost := &asonetworkv1.BastionHost{}
+	bastionHost.Spec = asonetworkv1.BastionHost_Spec{}
+	if existingBastionHost != nil {
+		bastionHost.Spec = existingBastionHost.Spec
```
nawazkh, line 59: I think it is best to copy over the complete configuration including `ObjectMeta` (which includes Labels, Annotations etc), `TypeMeta`, `Spec`, and `Status`. 
And then later down the flow, CAPZ enforces fields it cares about. 

```suggestion
		bastionHost = existingBastionHost
```

I don't think the `Diff` func will raise a diff on copying over `ObjectMeta`, `TypeMeta`, and `Status`. Have you seen otherwise in your testing?

azure/services/bastionhosts/spec.go
```diff
@@ -40,69 +41,65 @@ type AzureBastionSpec struct {
...
-						},
-						PublicIPAddress: &armnetwork.SubResource{
-							ID: &s.PublicIPID,
-						},
-						PrivateIPAllocationMethod: ptr.To(armnetwork.IPAllocationMethodDynamic),
-					},
+	bastionHost.Spec.AzureName = s.Name
+	if s.Location != "" {
+		bastionHost.Spec.Location = ptr.To(s.Location)
+	}
```
nawazkh, line 66: I am sorry about asking you to change this assignment to the way it is now. 
We have an active conversation going on this at [this comment of privateendpoints to ASO PR](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/4108/files#r1371005240). 
I think we are inclining towards having `bastionHost.Spec.Location = ptr.To(s.Location)` without performing an empty `""` check. You might want to revert this assignment without an empty check as well. 
Sorry!

controllers/azurecluster_controller.go
```diff
@@ -113,8 +113,8 @@ func (acr *AzureClusterReconciler) SetupWithManager(ctx context.Context, mgr ctr
 // +kubebuilder:rbac:groups="",resources=namespaces,verbs=list;
 // +kubebuilder:rbac:groups=resources.azure.com,resources=resourcegroups,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=resources.azure.com,resources=resourcegroups/status,verbs=get;list;watch
-// +kubebuilder:rbac:groups=network.azure.com,resources=natgateways,verbs=get;list;watch;create;update;patch;delete
-// +kubebuilder:rbac:groups=network.azure.com,resources=natgateways/status,verbs=get;list;watch
+// +kubebuilder:rbac:groups=network.azure.com,resources=natgateways;bastionhosts,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=network.azure.com,resources=natgateways;bastionhosts/status,verbs=get;list;watch
```
nawazkh, line 117: Sorry again 😬
I think we want to have descriptive annotations rather than clubbing them together. Another discussion point at [this comment on PrivateEndpoint to ASO PR](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/4108#discussion_r1370788398)

nojnhuh: I think the point I was trying to make there was separate. I don't think it matters whether or not we have separate lines per-resource or per-group.

nawazkh: Nonetheless, I think we need to get `// +kubebuilder:rbac:groups=network.azure.com,resources=natgateways;bastionhosts/status,verbs=get;list;watch
` updated to `// +kubebuilder:rbac:groups=network.azure.com,resources=natgateways/status;bastionhosts/status,verbs=get;list;watch
`

azure/services/bastionhosts/spec.go
willie-yao: No problem, thanks for the heads up!

willie-yao:
@nawazkh Bumping this comment in case you missed it. The prior implementation returned before setting defaults. Wouldn't this cause some fields to be populated with empty data rather than just keeping them nil? https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/4143#discussion_r1372447990

azure/services/bastionhosts/spec.go
nawazkh: I don't fully understand your statement; _"unecessary fields to be populated like an IpConfiguration with empty ARMIDs"_.

Since 
- CAPZ is populating the spec of the ASO resource (and)
- CAPZ's opinion should be respected over user interventions(where user updates a CAPZ managed ASO resource), 

if the user initially created an IP configuration with Empty ARMIDs, then empty ARMIDs should be populated.

Jont828:
@willie-yao Looks like you got a unit test failure. Seems like it has to do with expecting the IP config and the tags to be nil in the result.

azure/services/bastionhosts/spec.go
nawazkh: Returning the object early if found (as it was implemented earlier) will let the users modify the CAPZ-managed resource, and CAPZ will not be able to enforce its values.

nawazkh:
> @nawazkh Bumping this comment in case you missed it. The prior implementation returned before setting defaults. Wouldn't this cause some fields to be populated with empty data rather than just keeping them nil? [#4143 (comment)](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/4143#discussion_r1372447990)

Replied on the thread @ https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/4143#discussion_r1378244855

azure/services/bastionhosts/spec.go
willie-yao: That makes sense, I think my confusion is because the previous implementation would get the result as nil if an existing [BastionHost is present with empty data](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/services/bastionhosts/spec_test.go#L80). But now, we would expect a BastionHost with default empty values like [so](https://github.com/willie-yao/cluster-api-provider-azure/blob/aso-bastionhosts/azure/services/bastionhosts/spec_test.go#L102). It just makes the unit test a bit awkward.

willie-yao:
> Seems like it has to do with expecting the IP config and the tags to be nil in the result.

The unit test is breaking due to changed behavior when passing in an empty Spec. Going with @nawazkh's suggestion here: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/4143#discussion_r1378244855

I have changed the unit test to reflect the new behavior when receiving an empty Spec. PTAL

azure/services/bastionhosts/spec.go
```diff
@@ -40,69 +41,63 @@ type AzureBastionSpec struct {
...
-					Properties: &armnetwork.BastionHostIPConfigurationPropertiesFormat{
-						Subnet: &armnetwork.SubResource{
-							ID: &s.SubnetID,
-						},
-						PublicIPAddress: &armnetwork.SubResource{
-							ID: &s.PublicIPID,
-						},
-						PrivateIPAllocationMethod: ptr.To(armnetwork.IPAllocationMethodDynamic),
-					},
+	bastionHost.Spec.AzureName = s.Name
```
nawazkh, line 63: Do we know if `s.Name` could ever be empty?
If yes, should we construct `bastionHostIPConfigName` only when `s.Name != ""` so that `bastionHostIPConfigName` doesn't end up being `-bastionIP` ?

azure/services/bastionhosts/spec.go
```diff
@@ -40,69 +41,63 @@ type AzureBastionSpec struct {
...
+				Reference: &genruntime.ResourceReference{
+					ARMID: s.SubnetID,
+				},
+			},
+			PublicIPAddress: &asonetworkv1.BastionHostSubResource{
+				Reference: &genruntime.ResourceReference{
+					ARMID: s.PublicIPID,
 				},
 			},
+			PrivateIPAllocationMethod: ptr.To(asonetworkv1.IPAllocationMethod_Dynamic),
```
nawazkh, line 92: Is it valid and possible to have a bastionHost with IPConfigurations such that the one of IPConfiguration just has `ptr.To(asonetworkv1.IPAllocationMethod_Dynamic)`?
I mean is the below configuration valid and possible?
```yaml
spec:
	.
	.
	ipConfigurations:
		- publicIPAddress: "Dynamic"
	.
```

nawazkh: The reason I ask is because the [spec_test/get result as BastionHost with empty values when existing BastionHost is present with empty data](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/4143/files#diff-0894c31cc81cb639d7b041d520ea154ffd418c6605a41cd8c444e17cc5d78b1cR102-R127) appears to be returning a spec which is not completely nil (it has a `ipConfigurations` with one element in it). I am wondering if it should be `nil`.

nawazkh (commented):
Welcome changes, thanks for pointing out to the unit tests. I see the weirdness you are pointing to @willie-yao. I have put in some more questions below.

azure/services/bastionhosts/spec.go
willie-yao: `s.Name` is never empty because of the defaulting webhook here: https://github.com/willie-yao/cluster-api-provider-azure/blob/aso-bastionhosts/api/v1beta1/azurecluster_default.go#L331-L332

azure/services/bastionhosts/spec.go
willie-yao: It's returning a spec that has the Subnet and PublicIPAddress specs populated. The only difference is that the ARMID field will have an empty string. I'm not sure if that's what we want but this goes back to not returning existingBastionHost early if it is not nil https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/4143#discussion_r1372234691. Any extra fields that aren't specified will have weird outputs such as this https://github.com/willie-yao/cluster-api-provider-azure/blob/aso-bastionhosts/azure/services/bastionhosts/spec_test.go#L80-L81

azure/services/bastionhosts/spec.go
```diff
@@ -40,69 +41,66 @@ type AzureBastionSpec struct {
...
+	bastionHost.Spec = asonetworkv1.BastionHost_Spec{}
+	if existingBastionHost != nil {
+		bastionHost = existingBastionHost
 	}
 
 	bastionHostIPConfigName := fmt.Sprintf("%s-%s", s.Name, "bastionIP")
+	bastionHost.Spec.AzureName = s.Name
+	bastionHost.Spec.Location = ptr.To(s.Location)
+	bastionHost.Spec.Owner = &genruntime.KnownResourceReference{
+		Name: s.ClusterName,
```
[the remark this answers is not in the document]
nojnhuh: Actually I wonder if this is causing the test to fail:
https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/4143/pull-cluster-api-provider-azure-e2e-optional/1722328437001031680/artifacts/clusters/capz-e2e-2qw9kx-public-custom-vnet/capz-system/capz-controller-manager-7666bbb944-9z6wn/manager.log
```
I1108 19:58:22.118902       1 azurecluster_controller.go:237] controllers.AzureClusterReconciler.reconcileNormal "msg"="transient failure to reconcile AzureCluster, retrying: resource is not Ready: Owner \"capz-e2e-87vhc9-private, Group/Kind: resources.azure.com/ResourceGroup\" cannot be found. Progress is blocked until the owner is created.. Object will be requeued after 20s" "AzureCluster"={"name":"capz-e2e-87vhc9-private","namespace":"capz-e2e-2qw9kx"} "controller"="azurecluster" "controllerGroup"="infrastructure.cluster.x-k8s.io" "controllerKind"="AzureCluster" "name"="capz-e2e-87vhc9-private" "namespace"="capz-e2e-2qw9kx" "reconcileID"="4801ea42-f974-469b-ae79-981166d368bf" "x-ms-correlation-request-id"="f66fb9da-43a3-4543-a54a-146d145f964c"
```

azure/services/bastionhosts/spec.go
```diff
@@ -40,69 +41,66 @@ type AzureBastionSpec struct {
...
-					Name: ptr.To(bastionHostIPConfigName),
-					Properties: &armnetwork.BastionHostIPConfigurationPropertiesFormat{
-						Subnet: &armnetwork.SubResource{
-							ID: &s.SubnetID,
-						},
-						PublicIPAddress: &armnetwork.SubResource{
-							ID: &s.PublicIPID,
-						},
-						PrivateIPAllocationMethod: ptr.To(armnetwork.IPAllocationMethodDynamic),
+	if s.SubnetID != "" && s.PublicIPID != "" {
```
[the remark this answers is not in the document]
willie-yao: Ah I think this goes back to my confusion about setting empty ARMIDs since the resulting BastionHostIPConfiguration would be invalid, as discussed here: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/4143#discussion_r1379362506

Would setting these IPConfigurations with empty ARMIDs cause problems?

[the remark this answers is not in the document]
nojnhuh: Does it cause problems now? I'd be surprised if it worked before and doesn't work now with ASO.

Either way, it seems like these will never be empty based on how we create the spec in azure/scope/cluster.go.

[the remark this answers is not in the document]
willie-yao: It didn't cause problems, I just noticed it provided weird outputs in the unit test. I wasn't aware they were never empty though so I don't think it's an issue. I'll revert the change

azure/services/bastionhosts/spec.go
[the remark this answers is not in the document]
willie-yao: Good catch! I think that could be it. I made the change and lets see if that fixes it.

Jont828:
@willie-yao Looks like you got some unit test failures, otherwise LGTM.

nojnhuh (commented):
/lgtm

Great work @willie-yao and thank you @nawazkh for the reviews!

CecileRobertMichon:
This one is ready for rebase @willie-yao

nojnhuh (commented):
/lgtm

willie-yao:
The failing test is a known issue with the Flatcar test

mboersma (approved):
/lgtm
/approve

nojnhuh:
Thank you @willie-yao for squashing!
/hold cancel

PR #4145, opened 2023-10-17
feat(cluster): add failureDomain spec label

handsomejack-42:
**What type of PR is this?**

/kind api-change

**Why we need it:**

We need zonal control plane nodes. We might have missed something, but that probably can't be achieved in current state of CAPZ, with our use-case.

image

1. in its [Status](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/30d08b2a6134420f404205f0d8232a7bf270f425/api/v1beta1/azurecluster_types.go#L63C15-L63C15) property, `AzureCluster` announces all discovered failure domains (FDs)
1. when rolling out control planes, `KubeadmControlPlane` [checks with the FDs reported by the cluster](https://github.com/kubernetes-sigs/cluster-api/blob/740db36269f104beae74f6085d4e5d898261178e/controlplane/kubeadm/internal/control_plane.go#L95)
   * _the default behavior is to [evenly distribute the control plane machines](https://github.com/kubernetes-sigs/cluster-api/blob/740db36269f104beae74f6085d4e5d898261178e/controlplane/kubeadm/internal/control_plane.go#L138) (nodes) between FDs ([doc](https://capz.sigs.k8s.io/topics/failure-domains#default-behaviour))_
1. there is currently no way for us to restrict FDs used by `KubeadmControlPlane` when rolling out nodes (no [failureDomain in spec](https://doc.crds.dev/github.com/kubernetes-sigs/cluster-api/controlplane.cluster.x-k8s.io/KubeadmControlPlane/v1alpha3@v0.3.13#spec))
1. CAPZ documentation only mentions that [explicit placement](https://capz.sigs.k8s.io/topics/failure-domains#explicit-placement) into FD is achieved by setting the label on `Machine(Deployment)`. In our use-case, however, there are no `Machine(Deployment)`s, the `KubeadmControlPlane` spins `AzureMachines` directly.

**What this PR does:**

The goal of this commit is to allow CAPZ users to specify which failure domains are eligible for control plane rollouts.

CAPI elders [suggested](https://github.com/kubernetes-sigs/cluster-api/issues/2800#issuecomment-606035365), that the implementation of this behavior lies within the individual providers (CAPZ in our case)
* _tl;dr the goal is that the FDs are not announced by the `AzureCluster` object status, so `KubeadmControlPlane` won't take it into consideration when rolling out control plane nodes_

There's a new label in `.spec.failureDomain` that is used to post-filter discovered failure domains.

The field is _optional_ - if it's missing, all discovered failure domains are eligible for control plane rollout.

- [ ] cherry-pick candidate

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Added AzureCluster.spec.failureDomain label, that can be used to prevent control plane node deployment to specified failure domains
```


CecileRobertMichon:
@handsomejack-42 please let us know once you've signed the CLA so we can trigger the tests

handsomejack-42:
> @handsomejack-42 please let us know once you've signed the CLA so we can trigger the tests

✅

azure/scope/cluster.go
```diff
@@ -880,7 +880,15 @@ func (s *ClusterScope) SetFailureDomain(id string, spec clusterv1.FailureDomainS
...
 		s.AzureCluster.Status.FailureDomains = make(clusterv1.FailureDomains)
 	}
-	s.AzureCluster.Status.FailureDomains[id] = spec
+
+	if len(s.AzureCluster.Spec.FailureDomains) == 0 {
+		s.AzureCluster.Status.FailureDomains[id] = spec
+		return
+	}
+
+	if fd, ok := s.AzureCluster.Spec.FailureDomains[id]; ok && fd.ControlPlane {
```
handsomejack-42, line 889: actually, on second thought, there is a change to current flow, albeit a bit exotic. but let's definitely discuss if this should be changed or not

previously, when FD would be discovered (and not suitable for control plane, aka `controlPlane: false`), it would get into the status field no matter what

in current code, imagine there is FD1 discovered with `controlPlane: false` and in `spec`, there is FD2 with an arbitrary value

the change is, FD1 will no longer be announced in the status.

thoughts?

handsomejack-42:
@mboersma 

> make: Leaving directory '/home/prow/go/src/sigs.k8s.io/cluster-api-provider-azure'
cp: cannot stat 'coverage.*': No such file or directory
go: downloading golang.org/x/tools v0.8.0
Couldn't load /logs/artifacts/coverage.out: open /logs/artifacts/coverage.out: no such file or directory.+ EXIT_VALUE=1

can you pls check this?

mboersma:
/retest

Seems like a flake in the container or with file I/O? Let's hope so.

CecileRobertMichon:
> We need zonal control plane nodes. We might have missed something, but that probably can't be achieved in current state of CAPZ, with our use-case.

@handsomejack-42 to clarify, it's possible to spread control plane nodes across availability zones, and in fact they do get spread automatically across all available zones. Can you please expand on the use case for selecting a subset of the failure domains to spread control planes? Is the goal to have nodes in specific Azure availability zones? or am I misunderstanding what the PR is doing?

azure/scope/cluster.go
CecileRobertMichon: > "FD is always propagated to status, only its controlPlane value can be overriden to false by spec"

This is more future-proof and safe, so +1

In practice, CAPZ sets every single Azure zone it discovers to `controlPlane: true`: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/controllers/azurecluster_reconciler.go#L217 so this shouldn't actually happen

CecileRobertMichon (commented):
Could you please also update the docs as part of this change https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/docs/book/src/topics/failure-domains.md?

handsomejack-42:
> > We need zonal control plane nodes. We might have missed something, but that probably can't be achieved in current state of CAPZ, with our use-case.
> 
> @handsomejack-42 to clarify, it's possible to spread control plane nodes across availability zones, and in fact they do get spread automatically across all available zones. Can you please expand on the use case for selecting a subset of the failure domains to spread control planes? Is the goal to have nodes in specific Azure availability zones? or am I misunderstanding what the PR is doing?

@CecileRobertMichon we want the opposite - we need to be able to control to which zone our control plane gets deployed to, possibly limiting it to just some of the zones that are available

mboersma:
/retest

Prow failed to start the container.

docs/book/src/topics/failure-domains.md
```diff
@@ -128,6 +128,21 @@ spec:
 
 ```
 
+If you can't use `Machine` (or `MachineDeployment`) to explicitly place your VMs (for example, there is a `KubeAdmControlPlane` does not accept those as an object reference but rather uses `AzureMachineTemplate` directly), then you can opt to restrict the announcement of discovered failure domains from the cluster's status itself.
```
CecileRobertMichon, line 131:
```suggestion
If you can't use `Machine` (or `MachineDeployment`) to explicitly place your VMs (for example, there is a `KubeadmControlPlane` does not accept those as an object reference but rather uses `AzureMachineTemplate` directly), then you can opt to restrict the announcement of discovered failure domains from the cluster's status itself.
```

docs/book/src/topics/failure-domains.md
```diff
@@ -128,6 +128,21 @@ spec:
...
+```yaml
+apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
+kind: AzureCluster
+metadata:
+  name: my-cluster
+  namespace: default
+spec:
+  location: eastus
+  failureDomains:
+    fd1:
```
CecileRobertMichon, line 142: is fd1 actually what an Azure fd id would look like?

CecileRobertMichon (commented):
lgtm aside from nits in doc

/assign @mboersma

api/v1beta1/types_class.go
```diff
@@ -56,6 +57,15 @@ type AzureClusterClassSpec struct {
 	// Note: All cloud provider config values can be customized by creating the secret beforehand. CloudProviderConfigOverrides is only used when the secret is managed by the Azure Provider.
 	// +optional
 	CloudProviderConfigOverrides *CloudProviderConfigOverrides `json:"cloudProviderConfigOverrides,omitempty"`
+
+	// FailureDomains specifies the list of unique failure domains for the location/region of the cluster.
```
mboersma, line 61: Maybe we could shorten this comment:

```
// FailureDomains is a list of failure domains in the cluster's region, used to restrict
// eligibility to host the control plane. A FailureDomain maps to an availability zone,
// which is a separated group of datacenters within a region.
// See: https://learn.microsoft.com/azure/reliability/availability-zones-overview
// +optional
```

azure/scope/cluster.go
```diff
@@ -875,11 +875,17 @@ func (s *ClusterScope) APIServerHost() string {
 	return s.APIServerPublicIP().DNSName
 }
 
-// SetFailureDomain will set the spec for a for a given key.
+// SetFailureDomain to cluster's status by given id. The value of provided control plane might
```
mboersma, line 878:
```suggestion
// SetFailureDomain sets a failure domain in a cluster's status by its id. The provided failure domain spec may
```

docs/book/src/topics/failure-domains.md
mboersma: > for example, there is a `KubeadmControlPlane` does not accept those as an object reference 

I think we're missing a word here; I couldn't quite parse this paragraph...

handsomejack-42: fixed, sorry

docs/book/src/topics/failure-domains.md
handsomejack-42: when we were testing it on dev cluster, the values reported in cluster status were `1`, `2` etc. but i wasn't sure if that's the standard or not and for some reason it felt more explicit adding the `fd`

changed to `1:`

api/v1beta1/types_class.go
handsomejack-42: applied, thx

azure/scope/cluster.go
handsomejack-42: i tried to avoid the stutter here "SetFailureDomain sets failure domain" but i didn't feel very confident about how my version read either

applied your suggestion

CecileRobertMichon (commented):
/lgtm

@handsomejack-42 can you please add a release note in the PR description?

handsomejack-42:
> /lgtm
> 
> @handsomejack-42 can you please add a release note in the PR description?

done:

> Added AzureCluster.spec.failureDomain label, that can be used to prevent control plane nodes deployment to specified failure domains

mboersma (approved):
/lgtm
/approve

PR #4148, opened 2023-10-17
don't wait for ASO resources to be ready for updates

nojnhuh:
**What type of PR is this?**
/kind bug

ASO will serialize updates to Azure as necessary, so CAPZ should not worry about not updating ASO resource specs while an Azure update is in progress.

From https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/4069/files#r1344629003:
> The changes around here are because I noticed that before, if an ASO resource got stuck in a failed state, attempting to fix it by updating the CAPZ spec would never work because we were always returning an error here without calling `Parameters` and updating the ASO resource. These changes defer the equivalent check to after `Parameters` is called and generates no diff.

Fixes #

- [X] cherry-pick candidate

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Release note**:

```release-note
Fixed a bug causing ASO resources to potentially get stuck in a failed state
```

### DIFF
--- a/azure/services/aso/aso.go
+++ b/azure/services/aso/aso.go
@@ -81,6 +81,7 @@ func (r *reconciler[T]) CreateOrUpdateResource(ctx context.Context, spec azure.A
 
 	log = log.WithValues("service", serviceName, "resource", resourceName, "namespace", resourceNamespace)
 
+	var readyErr error
 	var adopt bool
 	var existing T
 	var zero T // holds the zero value, to be returned with non-nil errors.
@@ -105,7 +106,6 @@ func (r *reconciler[T]) CreateOrUpdateResource(ctx context.Context, spec azure.A
 		if !readyExists {
 			return zero, azure.WithTransientError(errors.New("ready status unknown"), requeueInterval)
 		}
-		var readyErr error
 		if cond := conds[i]; cond.Status != metav1.ConditionTrue {
 			switch {
 			case cond.Reason == conditions.ReasonAzureResourceNotFound.Name &&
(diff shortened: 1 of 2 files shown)


mboersma (commented):
/lgtm

azure/services/aso/aso.go
```diff
@@ -200,6 +201,9 @@ func (r *reconciler[T]) CreateOrUpdateResource(ctx context.Context, spec azure.A
 
 	diff := cmp.Diff(existing, parameters)
 	if diff == "" {
+		if readyErr != nil {
+			return zero, readyErr
```
nawazkh, line 205: I am curious, why are we trying to return `zero-valued T` instead of `parameters` if the `diff == ""`
_Diff_ would be empty if the _desired spec matches existing spec_, shouldn't we return the spec instead of the `zero T`?

nawazkh (commented):
Good find. Thanks for raising this PR. 
A question in-line.

azure/services/aso/aso.go
nojnhuh: This is following what I think is a fairly standard pattern to only return either a non-zero value or a non-nil error, and not a non-zero value with a non-nil error. That's confusing to me when I see that. "Is the value safe to use?" "Did this really fail?"

If we find a valid use case for doing that anyway though I don't think it would be a huge deal to change.

nawazkh: Makes sense. 

A short comment explaining this change would greatly help anyone visiting this code. The point to be noted here is that "Parameters" func gets called at least once, updating the ASO resource and CAPZ waits until the changes get reconciled before sending another request.

willie-yao (approved):
/lgtm

mboersma (approved):
/lgtm
/approve

PR #4151, opened 2023-10-17
:sparkles: Add webhook-cert-dir flag

vincepri:
**What type of PR is this?**

/kind feature

Adds the --webhook-cert-dir flag to customize where the certificates are stored, and passes it to the controller-runtime webhook handler.

Fixes #

- [ ] cherry-pick candidate

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -109,6 +109,7 @@ var (
 	syncPeriod                         time.Duration
 	healthAddr                         string
 	webhookPort                        int
+	webhookCertDir                     string
 	reconcileTimeout                   time.Duration
 	enableTracing                      bool
 )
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -224,6 +225,9 @@ func InitFlags(fs *pflag.FlagSet) {
 		"Webhook Server port, disabled by default. When enabled, the manager will only work as webhook server, no reconcilers are installed.",
 	)
 
+	fs.StringVar(&webhookCertDir, "webhook-cert-dir", "",
```
mboersma, line 228: In CAPI, this flag has a default value. Does CAPZ need to do the same here?

https://github.com/kubernetes-sigs/cluster-api/blob/ef2a1df06e00934d82c3140445e9634201f7a9a1/main.go#L212-L213



> "only used when webhook-port is specified."

Does this mean the value is ignored unless `--webhook-port` was overridden? The code just passes both values through regardless.

vincepri: I copied the flag, but dropped the default because I noticed it was already set in the Deployment yaml; can add it back if we prefer.

The comment is a bit confusing, I agree.

mboersma: > it was already set in the Deployment yaml;

Ah, I missed that. Thanks for updating, I think this is more clear.

mboersma (commented):
/lgtm

CecileRobertMichon (commented):
/lgtm
/approve

PR #4152, opened 2023-10-17
[release-1.11] don't wait for ASO resources to be ready for updates

nojnhuh:
**What type of PR is this?**
/kind bug

Manual cherry-pick of #4148:

> ASO will serialize updates to Azure as necessary, so CAPZ should not worry about not updating ASO resource specs while an Azure update is in progress.
>
> From https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/4069/files#r1344629003:
> > The changes around here are because I noticed that before, if an ASO resource got stuck in a failed state, attempting to fix it by updating the CAPZ spec would never work because we were always returning an error here without calling `Parameters` and updating the ASO resource. These changes defer the equivalent check to after `Parameters` is called and generates no diff.

Fixes #

- [ ] cherry-pick candidate

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Release note**:

```release-note
Fixed a bug causing ASO resources to potentially get stuck in a failed state
```

### DIFF
--- a/azure/services/aso/aso.go
+++ b/azure/services/aso/aso.go
@@ -75,6 +75,7 @@ func (s *Service) CreateOrUpdateResource(ctx context.Context, spec azure.ASOReso
 
 	log = log.WithValues("service", serviceName, "resource", resourceName, "namespace", resourceNamespace)
 
+	var readyErr error
 	var adopt bool
 	var existing genruntime.MetaObject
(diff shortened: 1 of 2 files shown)


mboersma (approved):
/lgtm
/approve

PR #4154, opened 2023-10-18
[release-1.11] wait for ASO tags to converge before updating

nojnhuh:
**What type of PR is this?**
/kind bug

Manual cherry-pick of #4149:

> From https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/4069/files#r1344633228:
> 
> > I had noticed a race condition while running the tags e2e test where if the tags have not converged, then CAPZ could lose track of some tags that have been deleted from `spec.additionalTags` and treat them like tags that existed outside of `spec.additionalTags` and keep them around without ever deleting them. Waiting for the tags to converge on the ASO resource fixes that issue AFAICT.
> > Overall, this makes tags reconciliation work more similarly to the more synchronous method used before when calling the tags APIs directly.

Fixes #

- [ ] cherry-pick candidate

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Release note**:

```release-note
Fixed a bug causing some tags on ASO resources not to be deleted
```

### DIFF
--- a/azure/services/aso/tags.go
+++ b/azure/services/aso/tags.go
@@ -18,9 +18,13 @@ package aso
 
 import (
 	"encoding/json"
+	"reflect"
 
+	asoannotations "github.com/Azure/azure-service-operator/v2/pkg/common/annotations"
 	"github.com/Azure/azure-service-operator/v2/pkg/genruntime"
 	"github.com/pkg/errors"
+	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
+	"sigs.k8s.io/cluster-api-provider-azure/azure"
 	"sigs.k8s.io/cluster-api-provider-azure/azure/converters"
 	"sigs.k8s.io/cluster-api-provider-azure/azure/services/tags"
 	"sigs.k8s.io/cluster-api-provider-azure/util/maps"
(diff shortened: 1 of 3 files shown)


azure/services/aso/tags.go
```diff
@@ -42,14 +47,29 @@ func reconcileTags(t TagsGetterSetter, existing genruntime.MetaObject, parameter
 				return errors.Wrapf(err, "failed to unmarshal JSON from %s annotation", tagsLastAppliedAnnotation)
 			}
 		}
-	}
 
-	existingTags, err := t.GetActualTags(existing)
-	if err != nil {
-		return errors.Wrapf(err, "failed to get actual tags for %s %s/%s", existing.GetObjectKind().GroupVersionKind(), existing.GetNamespace(), existing.GetName())
+		var err error
```
mboersma, line 51: I see you had to add this code, but it ends up differing a fair amount from the source PR. Was there another PR we should have cherry-picked or something? Just curious; the changes look sane.

nojnhuh: Yeah I think the ASO generics PR conflicted with this a bit: #3997

mboersma (approved):
/lgtm
/approve

PR #4155, opened 2023-10-18
Add ClusterClass Support for Managed Clusters

### LINKED ISSUE #2684: Add support for ClusterClass and managed clusters
/kind feature

Add support for using ClusterClass when working with managed AKS clusters. Ideally, there would be a flavor for AKS that uses ClusterClass to make is easier for first time users to get started with ClusterClass and managed clusters.

A use case for this is the ability to use lifecycle hook runtime extensions, <https://cluster-api.sigs.k8s.io/tasks/experimental-features/runtime-sdk/implement-lifecycle-hooks.html>. Clusters need to be created using ClusterClass in order for the hooks to work.

**Environment:**

- cluster-api-provider-azure version: latest
- Kubernetes version: (use `kubectl version`): 1.25.2
- OS (e.g. from `/etc/os-release`): 20.04.3

willie-yao:
**What type of PR is this?**
/kind feature

This PR adds ClusterClass support for managed clusters.

Fixes #2684 

- [ ] cherry-pick candidate

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Add ClusterClass support for managed clusters
```


api/v1beta1/azuremanagedcontrolplane_webhook.go
```diff
@@ -327,12 +358,13 @@ func (m *AzureManagedControlPlane) validateDisableLocalAccounts(_ client.Client)
 }
 
 // validateVersion validates the Kubernetes version.
-func (m *AzureManagedControlPlane) validateVersion(_ client.Client) error {
-	if !kubeSemver.MatchString(m.Spec.Version) {
-		return errors.New("must be a valid semantic version")
+func validateVersion(version string, fldPath *field.Path) field.ErrorList {
```
willie-yao, line 361: Refactored these webhook validation functions to be usable by the template types

api/v1beta1/azuremanagedmachinepool_webhook.go
```diff
@@ -295,23 +316,23 @@ func (mw *azureManagedMachinePoolWebhook) ValidateDelete(ctx context.Context, ob
...
 	}
 
-	return nil, errors.Wrapf(m.validateLastSystemNodePool(mw.Client), "if the delete is triggered via owner MachinePool please refer to trouble shooting section in https://capz.sigs.k8s.io/topics/managedcluster.html")
+	return nil, errors.Wrapf(validateLastSystemNodePool(mw.Client, m.Labels, m.Namespace), "if the delete is triggered via owner MachinePool please refer to trouble shooting section in https://capz.sigs.k8s.io/topics/managedcluster.html")
 }
 
 // validateLastSystemNodePool is used to check if the existing system node pool is the last system node pool.
 // If it is a last system node pool it cannot be deleted or mutated to user node pool as AKS expects min 1 system node pool.
-func (m *AzureManagedMachinePool) validateLastSystemNodePool(cli client.Client) error {
+func validateLastSystemNodePool(cli client.Client, labels map[string]string, namespace string) error {
```
willie-yao, line 324: Refactored these functions for the same purpose as above ^^

test/e2e/config/azure-dev.yaml
```diff
@@ -3,11 +3,11 @@ managementClusterName: capz-e2e
 images:
   - name: ${MANAGER_IMAGE}
     loadBehavior: mustLoad
-  - name: registry.k8s.io/cluster-api/cluster-api-controller:v1.5.2
+  - name: gcr.io/k8s-staging-cluster-api/cluster-api-controller:nightly_main_20230929
```
willie-yao, line 6: We had to pin the test version of CAPI to the nightly build as MachinePool workers support is not in the latest release of CAPI (1.5)

api/v1beta1/azuremanagedclustertemplate_types.go
```diff
@@ -0,0 +1,56 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// AzureManagedClusterTemplateSpec defines the desired state of AzureManagedClusterTemplate.
+type AzureManagedClusterTemplateSpec struct {
+	Template AzureManagedClusterTemplateResource `json:"template"`
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:path=azuremanagedclustertemplates,scope=Namespaced,categories=cluster-api
```
Jont828, line 29:
```suggestion
// +kubebuilder:resource:path=azuremanagedclustertemplates,scope=Namespaced,categories=cluster-api,shortName=amct
```
Could probably add the shortform version for the new CRDs to make `kubectl get` a lot easier

api/v1beta1/azuremanagedclustertemplate_webhook.go
```diff
@@ -0,0 +1,84 @@
...
+var _ webhook.Validator = &AzureManagedClusterTemplate{}
+
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type.
+func (r *AzureManagedClusterTemplate) ValidateCreate() (admission.Warnings, error) {
+	// NOTE: AzureManagedClusterTemplate relies upon MachinePools, which is behind a feature gate flag.
+	// The webhook must prevent creating new objects in case the feature flag is disabled.
+	if !feature.Gates.Enabled(capifeature.MachinePool) {
+		return nil, field.Forbidden(
+			field.NewPath("spec"),
+			"can be set only if the Cluster API 'MachinePool' feature flag is enabled",
```
Jont828, line 52:
```suggestion
			"cannot be set if the Cluster API 'MachinePool' feature flag is not enabled",
```

api/v1beta1/azuremanagedclustertemplate_webhook.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+// ValidateUpdate implements webhook.Validator so a webhook will be registered for the type.
+func (r *AzureManagedClusterTemplate) ValidateUpdate(oldRaw runtime.Object) (admission.Warnings, error) {
+	old := oldRaw.(*AzureManagedClusterTemplate)
+	var allErrs field.ErrorList
+
+	// custom headers are immutable
+	oldCustomHeaders := maps.FilterByKeyPrefix(old.ObjectMeta.Annotations, CustomHeaderPrefix)
+	newCustomHeaders := maps.FilterByKeyPrefix(r.ObjectMeta.Annotations, CustomHeaderPrefix)
```
Jont828, line 65: I remember @nojnhuh had a PR about custom headers being removed. Would that affect this PR at all?

api/v1beta1/azuremanagedcontrolplane_types.go
```diff
@@ -368,6 +368,11 @@ type AzureManagedControlPlaneStatus struct {
 	// +optional
 	Ready bool `json:"ready,omitempty"`
 
+	// Total number of non-terminated machines targeted by this control plane
```
Jont828, line 371:
```suggestion
	// Replicas indicates the total number of non-terminated machines targeted by this control plane
```

api/v1beta1/azuremanagedcontrolplanetemplate_types.go
```diff
@@ -0,0 +1,56 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// AzureManagedControlPlaneTemplateSpec defines the desired state of AzureManagedControlPlaneTemplate.
+type AzureManagedControlPlaneTemplateSpec struct {
+	Template AzureManagedControlPlaneTemplateResource `json:"template"`
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:path=azuremanagedcontrolplanetemplates,scope=Namespaced,categories=cluster-api
```
Jont828, line 29:
```suggestion
// +kubebuilder:resource:path=azuremanagedcontrolplanetemplates,scope=Namespaced,categories=cluster-api,shortName=amcpt
```

api/v1beta1/azuremanagedmachinepooltemplate_types.go
```diff
@@ -0,0 +1,56 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// AzureManagedMachinePoolTemplateSpec defines the desired state of AzureManagedMachinePoolTemplate.
+type AzureManagedMachinePoolTemplateSpec struct {
+	Template AzureManagedMachinePoolTemplateResource `json:"template"`
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:path=azuremanagedmachinepooltemplates,scope=Namespaced,categories=cluster-api
```
Jont828, line 29:
```suggestion
// +kubebuilder:resource:path=azuremanagedmachinepooltemplates,scope=Namespaced,categories=cluster-api,shortName=ammpt
```

api/v1beta1/types_template.go
```diff
@@ -18,9 +18,286 @@ package v1beta1
...
 
+// AzureManagedControlPlaneTemplateResourceSpec specifies an Azure managed control plane template resource.
+type AzureManagedControlPlaneTemplateResourceSpec struct {
+	// MachineTemplate contains information about how machines
+	// should be shaped when creating or updating a control plane.
+	// +optional
+	MachineTemplate *AzureManagedControlPlaneTemplateMachineTemplate `json:"machineTemplate,omitempty"`
+
+	// Version defines the desired Kubernetes version.
+	// +kubebuilder:validation:MinLength:=2
```
Jont828, line 36:
```suggestion
	// +kubebuilder:validation:MinLength=2
```

Should there be a semicolon there? I searched through CAPZ and it seems like it's used both with and without the semicolon, so maybe try testing it out to see if the validation works.

api/v1beta1/types_template.go
```diff
@@ -18,9 +18,286 @@ package v1beta1
...
+	Version string `json:"version"`
+
+	// VirtualNetwork describes the vnet for the AKS cluster. Will be created if it does not exist.
```
Jont828, line 39:
```suggestion
	// VirtualNetwork describes the virtual network for the AKS cluster. It will be created if it does not already exist.
```

api/v1beta1/types_template.go
```diff
@@ -18,9 +18,286 @@ package v1beta1
...
+	// +optional
+	VirtualNetwork ManagedControlPlaneVirtualNetworkTemplate `json:"virtualNetwork,omitempty"`
+
+	// SubscriptionID is the GUID of the Azure subscription to hold this cluster.
```
Jont828, line 43:
```suggestion
	// SubscriptionID is the GUID of the Azure subscription that owns this cluster.
```

controllers/helpers.go
```diff
@@ -945,6 +945,10 @@ func MachinePoolToAzureManagedControlPlaneMapFunc(ctx context.Context, c client.
 
 		gk := gvk.GroupKind()
 		ref := cluster.Spec.ControlPlaneRef
+		if ref == nil || ref.Name == "" {
+			log.Error(err, "control plane ref is nil or empty")
```
Jont828, line 949:
```suggestion
			log.Error(errors.New("control plane ref is nil or empty"))
```

I don't think you want to use `err` here as it will always be nil if it makes it past line 941.

docs/book/src/topics/clusterclass.md
```diff
@@ -0,0 +1,88 @@
+# ClusterClass
+
+- **Feature status:** GA
+- **Feature gate:** MachinePool=true ClusterTopology=true
```
Jont828, line 4:
```suggestion
- **Feature gate:** `MachinePool=true`, `ClusterTopology=true`
```

docs/book/src/topics/clusterclass.md
```diff
@@ -0,0 +1,88 @@
...
+
+Each resource is a template for the corresponding CAPZ resource. For example, the AzureClusterTemplate is a template for the CAPZ AzureCluster resource. The template contains a set of parameters that are able to be shared across multiple clusters.
+
+## Deploying a Self-Managed Cluster with ClusterClass
+
+To deploy a self-managed cluster with ClusterClass, you must first create a ClusterClass resource. The ClusterClass resource defines the cluster topology, including the control plane and machine deployment templates. The ClusterClass resource also defines the parameters that can be used to customize the cluster topology. 
+
+Please refer to the Cluster API book for more information on how to write a ClusterClass topology: https://cluster-api.sigs.k8s.io/tasks/experimental-features/cluster-class/write-clusterclass.html
+
+For a self-managed cluster, the AzureClusterTemplate is used to define the Azure infrastructure for the cluster. The following example shows a basic AzureClusterTemplate resource:
```
Jont828, line 21:
```suggestion
For a self-managed cluster, the AzureClusterTemplate defines the Azure infrastructure for the cluster. The following example shows a basic AzureClusterTemplate resource:
```

templates/test/ci/prow-aks-clusterclass/patches/aks-clusterclass-pool1.yaml
```diff
@@ -0,0 +1,68 @@
...
+          netIpv4TCPTwReuse: false
+          netIpv4TCPkeepaliveIntvl: 75
+          netNetfilterNfConntrackBuckets: 65536
+          netNetfilterNfConntrackMax: 131072
+          vmMaxMapCount: 65530
+          vmSwappiness: 60
+          vmVfsCachePressure: 100
+        transparentHugePageDefrag: "madvise"
+        transparentHugePageEnabled: "always"
+---
```
Jont828, line 68: Nit: are we supposed to have the `---` at the bottom of the file?

test/e2e/azure_test.go
```diff
@@ -844,6 +844,135 @@ var _ = Describe("Workload cluster creation", func() {
 		})
 	})
 
+	Context("Creating an AKS cluster using ClusterClass [Managed Kubernetes]", func() {
+		It("with a single control plane node, one linux worker node, and one windows worker node", func() {
+			// use default as the clusterclass name so test infra can find the clusterclass template
+			os.Setenv("CLUSTER_CLASS_NAME", "default")
+
+			// use "cc" as spec name because natgw pip name exceeds limit.
```
Jont828, line 852:
```suggestion
			// Use "cc" as spec name because NAT gateway pip name exceeds limit.
```

What does pip mean?

test/e2e/azure_test.go
```diff
@@ -844,6 +844,135 @@ var _ = Describe("Workload cluster creation", func() {
 		})
 	})
 
+	Context("Creating an AKS cluster using ClusterClass [Managed Kubernetes]", func() {
+		It("with a single control plane node, one linux worker node, and one windows worker node", func() {
+			// use default as the clusterclass name so test infra can find the clusterclass template
```
Jont828, line 849:
```suggestion
			// Use default as the clusterclass name so test infra can find the clusterclass template
```

test/e2e/config/azure-dev.yaml
Jont828: Seems fine to me, maybe add a comment explaining the reason for that? And the condition for when we can go back to the normal build.

Jont828 (commented):
This is looking good! Here's some comments from a first pass. It seems like most of it is template stuff, so I think if we can get all the tests passing, we can consider the templates to be good to go.

api/v1beta1/azuremanagedclustertemplate_webhook.go
willie-yao: I think this might as he removed customHeaders in the managed clusters ASO refactor PR. Should I be good to remove this? @nojnhuh

api/v1beta1/types_template.go
willie-yao: Since the non-template version of this field uses the colon, I thought to keep it consistent with that https://github.com/willie-yao/cluster-api-provider-azure/blob/clusterclass-aks-final/api/v1beta1/azuremanagedcontrolplane_types.go#L81

test/e2e/azure_test.go
willie-yao: I think it stands for public ip but I'm not 100% sure. Can you confirm @CecileRobertMichon?

willie-yao:
> Not sure what's going on, but looks like the E2E test is having trouble applying the cluster template. It was working before but I made some updates to the webhoook validation since. Will be debugging this

Fixed!

nojnhuh (commented):
Still working through this, but wanted to share what I had so far.

api/v1beta1/azuremanagedcontrolplane_types.go
```diff
@@ -368,6 +368,11 @@ type AzureManagedControlPlaneStatus struct {
 	// +optional
 	Ready bool `json:"ready,omitempty"`
 
+	// Replicas indicates the total number of non-terminated machines targeted by this control plane
+	// (their labels match the selector).
+	// +optional
+	Replicas int32 `json:"replicas"`
```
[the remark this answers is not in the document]
willie-yao: It's not used, must've kept it by mistake. Thanks for catching this!

test/e2e/config/azure-dev.yaml
willie-yao: Yes we decided in CAPZ office hours that it would be okay to merge this change since we are only using the newer CAPI binaries for testing.

api/v1beta1/azuremanagedcontrolplane_webhook_test.go
```diff
@@ -113,6 +114,326 @@ func TestDefaultingWebhook(t *testing.T) {
 	g.Expect(amcp.Spec.VirtualNetwork.Subnet.CIDRBlock).To(Equal(defaultAKSNodeSubnetCIDRForOverlay))
 }
 
+func TestValidateDNSServiceIP(t *testing.T) {
+	g := NewWithT(t)
+	tests := []struct {
+		name      string
+		dnsIP     *string
+		expectErr bool
```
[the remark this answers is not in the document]
willie-yao: Done. I think DNSServiceIP only returns one error

api/v1beta1/azuremanagedcontrolplane_webhook_test.go
```diff
@@ -756,7 +1077,7 @@ func TestAzureManagedControlPlane_ValidateCreate(t *testing.T) {
 			name:     "invalid DNSServiceIP",
 			amcp:     createAzureManagedControlPlane("192.168.0.10.3", "v1.18.0", generateSSHPublicKey(true)),
 			wantErr:  true,
-			errorLen: 1,
+			errorLen: 2,
```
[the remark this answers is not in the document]
willie-yao: This happens inside of `validateManagedClusterNetwork` which is part of the amcp validateCreate: https://github.com/willie-yao/cluster-api-provider-azure/blob/clusterclass-aks-final/api/v1beta1/azuremanagedcontrolplane_webhook.go#L495

api/v1beta1/consts.go
```diff
@@ -166,3 +166,8 @@ const (
 	// value for the label is the CAPI Cluster Name.
 	OwnedByClusterLabelKey = NameAzureProviderPrefix + string(ResourceLifecycleOwned)
 )
+
+const (
+	// NetworkPluginName is the name of the network plugin.
+	NetworkPluginName = "azure"
```
nojnhuh, line 172: Should we also name this to distinguish "azure" from "kubenet" which is also a valid network plugin?

api/v1beta1/azuremanagedcontrolplane_webhook_test.go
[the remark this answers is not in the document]
nojnhuh: Right, but why are we also returning a `Cluster.Spec...` error when validating an AzureManagedControlPlane which doesn't have that field? AFAICT this is doing the exact same validation twice on the exact same data.

api/v1beta1/types_template.go
```diff
@@ -18,9 +18,286 @@ package v1beta1
...
+	// APIServerAccessProfile is the access profile for AKS API server.
+	// +optional
+	APIServerAccessProfile *APIServerAccessProfileTemplate `json:"apiServerAccessProfile,omitempty"`
+
+	// AutoscalerProfile is the parameters to be applied to the cluster-autoscaler when enabled
+	// +optional
+	AutoScalerProfile *AutoScalerProfile `json:"autoscalerProfile,omitempty"`
+}
+
+// AzureManagedControlPlaneTemplateMachineTemplate specifies an Azure managed control plane template.
```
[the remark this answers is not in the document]
nojnhuh: A note in a comment explaining that this is only to fulfill that contract and otherwise isn't used seems worthwhile then IMO.

nojnhuh:
@willie-yao Could you please not force-push to this branch while we iterate on this? It makes reviewing incrementally much more difficult. Let me know if you run into any scenario where you feel like you need to force push and I can show you how to get through that. I like to put a hold on PRs to remind myself to clean up the git history in case you're worried about accidentally not merging what you want.

docs/book/src/topics/clusterclass.md
```diff
@@ -0,0 +1,88 @@
+# ClusterClass
+
+- **Feature status:** GA
+- **Feature gate:** `MachinePool=true`, `ClusterTopology=true`
```
nojnhuh, line 4: MachinePool isn't required for all cluster class usage right? Just if you need machine pools?

docs/book/src/topics/clusterclass.md
```diff
@@ -0,0 +1,88 @@
...
+  subscriptionID: 00000000-0000-0000-0000-000000000000
+  version: 1.25.2
+```
+
+## Excluded Fields
+
+Since a ClusterClass is a template for a Cluster, there are some fields that are not allowed to be shared across multiple clusters. For each of the ClusterClass resources, the following fields are excluded:
+
+### AzureClusterTemplate
+- `spec.resourceGroup`
```
nojnhuh, line 71: Is there a better place we can refer to these fields? Or do we need to document them explicitly at all? Listing them in the docs like this is sure to go stale, especially has we continue to add features.

templates/cluster-template-aks-clusterclass.yaml
```diff
@@ -0,0 +1,155 @@
...
+  allowedNamespaces: {}
+  clientID: ${AZURE_CLIENT_ID}
+  clientSecret:
+    name: ${AZURE_CLUSTER_IDENTITY_SECRET_NAME}
+    namespace: ${AZURE_CLUSTER_IDENTITY_SECRET_NAMESPACE}
+  tenantID: ${AZURE_TENANT_ID}
+  type: ServicePrincipal
+---
+apiVersion: bootstrap.cluster.x-k8s.io/v1beta1
+kind: KubeadmConfigTemplate
```
nojnhuh, line 110: Do we need these kubeadm resources?

templates/cluster-template-clusterclass.yaml
```diff
@@ -79,30 +79,6 @@ spec:
             kind: AzureMachineTemplate
             name: ${CLUSTER_NAME}-md-0
 ---
-apiVersion: cluster.x-k8s.io/v1beta1
```
nojnhuh, line 82: Is a breaking change for users relying on `clusterctl generate`?

willie-yao:
@nojnhuh Sorry about that! I force-pushed to the 5 existing commits in case there are other reviewers who are reviewing each commit, so that they don't review outdated code in each commit. However it definitely makes an incremental review more difficult, so I'll avoid force-pushing from now on

test/e2e/aks_autoscaler.go
```diff
@@ -62,7 +63,14 @@ func AKSAutoscaleSpec(ctx context.Context, inputGetter func() AKSAutoscaleSpecIn
...
 	ammp := &infrav1.AzureManagedMachinePool{}
-	err = mgmtClient.Get(ctx, client.ObjectKeyFromObject(input.MachinePool), ammp)
+	if input.isClusterClass {
+		err = bootstrapClusterProxy.GetClient().Get(ctx, types.NamespacedName{
+			Namespace: input.MachinePool.Spec.Template.Spec.InfrastructureRef.Namespace,
+			Name:      input.MachinePool.Spec.Template.Spec.InfrastructureRef.Name,
+		}, ammp)
+	} else {
+		err = mgmtClient.Get(ctx, client.ObjectKeyFromObject(input.MachinePool), ammp)
+	}
```
nojnhuh, line 73: Would the `isClusterClass == true` branch also work in non-cluster class scenarios?

test/e2e/azure_test.go
```diff
@@ -844,6 +844,135 @@ var _ = Describe("Workload cluster creation", func() {
...
+				AKSNodeLabelsSpec(ctx, func() AKSNodeLabelsSpecInput {
+					return AKSNodeLabelsSpecInput{
+						Cluster:       result.Cluster,
+						MachinePools:  result.MachinePools,
+						WaitForUpdate: e2eConfig.GetIntervals(specName, "wait-machine-pool-nodes"),
+					}
+				})
+			})
+
+			By("modifying taints configuration", func() {
```
nojnhuh, line 964: Is it possible to share a good portion of this code with the existing AKS test?

api/v1beta1/azuremanagedcontrolplane_webhook.go
```diff
@@ -274,28 +275,58 @@ func (mw *azureManagedControlPlaneWebhook) ValidateDelete(ctx context.Context, o
...
 		m.validateNetworkPluginMode,
 		m.validateDNSPrefix,
 		m.validateDisableLocalAccounts,
 	}
-
-	var errs []error
 	for _, validator := range validators {
 		if err := validator(cli); err != nil {
-			errs = append(errs, err)
+			allErrs = append(allErrs, field.InternalError(field.NewPath("spec"), err))
```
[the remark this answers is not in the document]
willie-yao: Ah good catch. It would definitely cause an unecessary wrap around the existing errors.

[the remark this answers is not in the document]
willie-yao: Since now this is a `field.ErrorList`, I can't just append an error of type `error` to `allErrs`, so I'll be refactoring all of the validators to return a `field.ErrorList` instead.

api/v1beta1/azuremanagedcontrolplane_webhook_test.go
```diff
@@ -113,6 +114,345 @@ func TestDefaultingWebhook(t *testing.T) {
...
+	}
+}
+
+func TestValidateLoadBalancerProfile(t *testing.T) {
+	g := NewWithT(t)
+	tests := []struct {
+		name        string
+		profile     *LoadBalancerProfile
+		wantErr     bool
+		expectedErr field.Error
```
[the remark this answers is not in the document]
willie-yao: I saw this convention in other files but it's definitely unecessary. I'll go ahead and remove it
Example: https://github.com/willie-yao/cluster-api-provider-azure/blob/clusterclass-aks-final/api/v1beta1/azurecluster_validation_test.go#L1134

api/v1beta1/azuremanagedcontrolplane_webhook_test.go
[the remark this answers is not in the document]
willie-yao: It looks like this was added initially in the [async refactor for Managed Clusters](https://github.com/kubernetes-sigs/cluster-api-provider-azure/commit/2d9595066766693cb6f5d36c3afb7d05c1b47ecd#diff-794f776e55c518f77cccea935614365a8889a67d892918bd16d24f14fa0cbbc6R369). I'm not sure why this is being done in the managed control plane test and not the managed cluster test. @CecileRobertMichon Is there a specific reason this test is here?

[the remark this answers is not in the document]
willie-yao: Looks like the `DNSServiceIP` is being used in part of the test, so that might be why it's being tested here.

config/webhook/manifests.yaml
```diff
@@ -114,6 +114,28 @@ webhooks:
...
+  - v1
+  - v1beta1
+  clientConfig:
+    service:
+      name: webhook-service
+      namespace: system
+      path: /mutate-infrastructure-cluster-x-k8s-io-v1beta1-azuremanagedcontrolplanetemplate
+  failurePolicy: Fail
+  matchPolicy: Equivalent
+  name: default.azuremanagedcontrolplanetemplate.infrastructure.cluster.x-k8s.io
```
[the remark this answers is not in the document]
willie-yao: I didn't see any issues with it not being plural but it definitely should be plural. It was a mistake on my end when writing the kubebuilder command for azuremanagedcontrolplanetemplates

config/webhook/manifests.yaml
```diff
@@ -295,6 +339,26 @@ webhooks:
...
   sideEffects: None
+- admissionReviewVersions:
+  - v1
+  - v1beta1
+  clientConfig:
+    service:
+      name: webhook-service
+      namespace: system
+      path: /validate-infrastructure-cluster-x-k8s-io-v1beta1-azuremanagedclustertemplate
+  failurePolicy: Fail
```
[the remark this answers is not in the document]
willie-yao: According to the [kubebuilder book](https://book.kubebuilder.io/reference/markers/webhook.html), the default value of `matchPolicy` is `exact` meaning it will match only if it exactly matches the specified rule. `Equivalent` matchrd a request if it modifies a resource listed in rules, even via another API group or version. I think this means it's only necessary for types that exist in previous API versions? So I think it should be good to remove

controllers/helpers.go
```diff
@@ -945,6 +945,10 @@ func MachinePoolToAzureManagedControlPlaneMapFunc(ctx context.Context, c client.
 
 		gk := gvk.GroupKind()
 		ref := cluster.Spec.ControlPlaneRef
+		if ref == nil || ref.Name == "" {
+			log.Error(errors.New("control plane ref is nil or empty"), "failed to get the control plane ref")
```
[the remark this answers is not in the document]
willie-yao: Yes this did come up in testing and I agree that `Info` would be better as it resolves itself

templates/cluster-template-aks-clusterclass.yaml
willie-yao: Yes, these are necessary as the bootstrap ref for the workers https://github.com/willie-yao/cluster-api-provider-azure/blob/clusterclass-aks-final/templates/cluster-template-aks-clusterclass.yaml#L24

templates/cluster-template-clusterclass.yaml
willie-yao: I moved the cluster topology to its own file since it is not recommended to define a Cluster topology and its ClusterClass at the same time. For users using `clusterctl generate`, does it require a `Cluster` object in a template to apply it to your cluster? If so, users will have to `kubectl apply -f` the ClusterClass separately but yeah it is a breaking change. @CecileRobertMichon wdyt?

test/e2e/aks_autoscaler.go
willie-yao: Actually yes I think it does, I can remove this check then

api/v1beta1/types_template.go
```diff
@@ -18,9 +18,286 @@ package v1beta1
...
+	// +optional
+	NodeDeletionTimeout *metav1.Duration `json:"nodeDeletionTimeout,omitempty"`
+}
+
+// AzureManagedMachinePoolTemplateResourceSpec specifies an Azure managed control plane template resource.
+type AzureManagedMachinePoolTemplateResourceSpec struct {
+	// AdditionalTags is an optional set of tags to add to Azure resources managed by the
+	// Azure provider, in addition to the ones added by default.
+	// +optional
+	AdditionalTags Tags `json:"additionalTags,omitempty"`
```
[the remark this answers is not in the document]
willie-yao: Users can modify any field including AdditionalTags [via patches](https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/proposals/20210526-cluster-class-and-managed-topologies.md#create-a-new-clusterclass-with-patches), although I'm not sure if AdditionalTags is the best type to include in the template if it is something that is unique per resource.

test/e2e/azure_test.go
willie-yao: Added a common function in helpers.go: https://github.com/willie-yao/cluster-api-provider-azure/blob/clusterclass-aks-final/test/e2e/helpers.go#L946

templates/cluster-template-aks-clusterclass.yaml
nojnhuh: Ah, so this is another thing just to make CAPI happy? If so, are these as minimal as possible to reduce the risk of interfering with AKS?

templates/cluster-template-clusterclass.yaml
nojnhuh: I don't think there are any requirements as to how we structure our templates, I can see where this might be frustrating if the script I had been using before with this template no longer actually produces a cluster.

api/v1beta1/types_template.go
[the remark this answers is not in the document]
nojnhuh: Are the fields to be included in template types only things that \*definitely are not\* unique to each cluster or can they include things like this that \*could possibly be\* unique to each cluster if the user sets things up that way? If I want to tag all my clusters with `foo: bar`, it seems like I should be able to do that with a template.

templates/cluster-template-aks-clusterclass.yaml
willie-yao: I'm not entirely sure if they are necessary with a managed cluster so I'll try testing with the smallest template possible and get back to you!

api/v1beta1/types_template.go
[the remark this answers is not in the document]
willie-yao: I think the convention is that we only exclude types that would have to be unique to each cluster, rather than things that can be possibly unique. Otherwise, if a user wants to template a type that doesn't have to be unique and it's not included as a type, it would be much more difficult than patching a type that is included in the template.

api/v1beta1/azuremanagedcontrolplane_default.go
```diff
@@ -96,18 +98,51 @@ func (m *AzureManagedControlPlane) setDefaultSubnet() {
...
+func setDefaultSku(sku *AKSSku) {
+	if sku != nil && sku.Tier == PaidManagedControlPlaneTier {
+		sku.Tier = StandardManagedControlPlaneTier
+		ctrl.Log.WithName("AzureManagedControlPlaneWebHookLogger").Info("Paid SKU tier is deprecated and has been replaced by Standard")
+	}
+}
+
+func setDefaultNetworkPlugin(networkPlugin *string) {
+	if networkPlugin == nil {
+		networkPlugin = ptr.To(AzureNetworkPluginName)
```
nojnhuh, line 110: I think the linter is saying that for this to have the desired side effect, this would have to do `*networkPlugin = ...` instead of `networkPlugin = ...`. The reason the SKU one above doesn't throw the same error is because there only a field on a struct pointer is set, so the value of the input pointer isn't changed like it is here.

All in all I think it would be this:
```suggestion
		networkPlugin = new(string)
		*networkPlugin = AzureNetworkPluginName
```

Are there any tests that also catch this?

api/v1beta1/azuremanagedcontrolplane_default.go
```diff
@@ -96,18 +98,51 @@ func (m *AzureManagedControlPlane) setDefaultSubnet() {
...
+	}
+}
+
+func setDefaultLoadBalancerSKU(loadBalancerSKU *string) {
+	if loadBalancerSKU == nil {
+		loadBalancerSKU = ptr.To("Standard")
```
nojnhuh, line 116: For things this simple, is there a kubebuilder annotation we can add to the type instead of writing code to default the value?

api/v1beta1/azuremanagedcontrolplanetemplate_default.go
```diff
@@ -37,8 +37,8 @@ func (mcp *AzureManagedControlPlaneTemplate) setDefaults() {
 
 	mcp.setDefaultVirtualNetwork()
 	mcp.setDefaultSubnet()
-	mcp.Spec.Template.Spec.SKU = setDefaultSku(mcp.Spec.Template.Spec.SKU)
-	mcp.Spec.Template.Spec.AutoScalerProfile = setDefaultAutoScalerProfile(mcp.Spec.Template.Spec.AutoScalerProfile)
+	setDefaultSku(mcp.Spec.Template.Spec.SKU)
+	setDefaultAutoScalerProfile(mcp.Spec.Template.Spec.AutoScalerProfile)
```
nojnhuh, line 41: I have no self-control and can't stop playing with generics, so offering this in case it's useful:
```go
func setDefault[T comparable](field *T, value T) {
	if field == nil {
		// shouldn't happen with proper use
		return
	}
	var zero T
	if *field == zero {
		*field = value
	}
}

setDefault(&m.Spec.NetworkPlugin, ptr.To(AzureNetworkPluginName))
//         ^ where referencing a field (even if it's already a pointer) is "proper use"

func TestSetDefault(t *testing.T) {
	g := NewGomegaWithT(t)

	type Struct struct{ name string }

	var s *Struct
	setDefault(&s, &Struct{"hello"})
	g.Expect(s.name).To(Equal("hello"))
	setDefault(&s, &Struct{"world"})
	g.Expect(s.name).To(Equal("hello"))

	r := &Struct{}
	setDefault(&r, &Struct{"a name"})
	g.Expect(r.name).To(BeEmpty())
	setDefault(&r.name, "hello")
	g.Expect(r.name).To(Equal("hello"))
	setDefault(&r.name, "world")
	g.Expect(r.name).To(Equal("hello"))

	str := ""
	setDefault(&str, "a string")
	g.Expect(str).To(Equal("a string"))
	setDefault(&str, "another string")
	g.Expect(str).To(Equal("a string"))
}


```

templates/cluster-template-clusterclass.yaml
CecileRobertMichon: I do think this is a cleaner separation, and the likelihood that there is someone out there that gets broken by this change is pretty low. Hopefully, if you have scripts that rely on template files in a repo you are pinning the CAPZ version and testing the scripts before bumping the version...

so I'm leaning +1 on making this change

api/v1beta1/types_template.go
```diff
@@ -18,9 +18,289 @@ package v1beta1
...
 	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/utils/net"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 )
 
+// AzureManagedControlPlaneTemplateResourceSpec specifies an Azure managed control plane template resource.
+type AzureManagedControlPlaneTemplateResourceSpec struct {
```
CecileRobertMichon, line 29: It seems that this is diverging from the AzureCluster ClassClass implementation in that it duplicates all the fields instead of creating an embedded "class" type that embeds the shared fields between AzureManagedControlPlane Spec and AzureManagedControlPlaneTemplate Spec https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/api/v1beta1/azurecluster_types.go#L35 

History on that design decision: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1943

willie-yao:
Sorry for the force push. Needed to fix a merge conflict

nojnhuh:
> Sorry for the force push. Needed to fix a merge conflict

Doing `git merge` and pushing the merge commit will also fix conflicts without needing to force-push.

api/v1beta1/azuremanagedcontrolplane_default.go
willie-yao: This seems to not work with the field on a struct pointer such as `setDefaultSku`, even when using this method. I'm wondering if there is a better way to do this... Also I think we can just use kubebuilder the default here.

api/v1beta1/azuremanagedcontrolplanetemplate_webhook.go
```diff
@@ -0,0 +1,289 @@
...
+}
+
+// Default implements webhook.Defaulter so a webhook will be registered for the type.
+func (mcpw *azureManagedControlPlaneTemplateWebhook) Default(ctx context.Context, obj runtime.Object) error {
+	mcp, ok := obj.(*AzureManagedControlPlaneTemplate)
+	if !ok {
+		return apierrors.NewBadRequest("expected an AzureManagedControlPlaneTemplate")
+	}
+	mcp.setDefaults()
+	return nil
```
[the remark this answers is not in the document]
willie-yao: I added extensive webhook tests for both these files and there definitely seems to be room for improvement code-repetition wise. The only tricky thing is that some fields are not included in the template version of the type, and they already share any functions that test the contents of the field beyond immutability.

docs/book/src/topics/clusterclass.md
willie-yao: I think we can just leave it out, since users can refer to the CRDs for available fields

api/v1beta1/azuremanagedcontrolplanetemplate_default.go
willie-yao: Nice! I incorporated this into as many checks in `Default()` as I could. Some checks that didn't perform a nil check couldn't use this.

willie-yao:
Update for reviewers: I have addressed all the comments so far and something broke in the functionality of this feature through all the refactors. I will be working on addressing this ASAP as well as a few test failures and ping once it's ready for review again.

willie-yao:
@nojnhuh @Jont828 @CecileRobertMichon PR should be ready for another round of review. Turns out the certificates issue was due to the `KubeadmConfigTemplate` not being configured correctly. Since they only exist to satisfy CAPI, reducing them to the default fixed the issue.

willie-yao:
Still working on figuring out why the E2E test is broken and complaning about a missing `windowsProfile`. Besides that the PR is all good for review

Jont828:
Sounds good, I'll give this another pass tomorrow or on Thursday. Nice work moving this (huge) PR along!

willie-yao:
Looks like there are quite a few edge cases when running an AKS cluster built using ClusterClass against the same [AKSTestSuite](https://github.com/willie-yao/cluster-api-provider-azure/blob/clusterclass-aks-final/test/e2e/helpers.go#L946) that the existing AKS test uses. Instead using the same specs to test, would it be more helpful to write ClusterClass-specific E2E tests such as editing the template and seeing the change propagate to multiple clusters? Or for the sake of getting this merged before the release just doing a similar test where we just create a cluster like the existing [ClusterClass E2E](https://github.com/willie-yao/cluster-api-provider-azure/blob/clusterclass-aks-final/test/e2e/azure_test.go#L865-L866) test? 

cc @CecileRobertMichon @nojnhuh @Jont828

nojnhuh:
What are the particular issues you're seeing? If they have to do with the tests themselves being buggy, I think it's acceptable to ship with a smaller, stabler test as long as we document the gaps we want to fill in new issues. If the tests are behaving more-or-less as a user should and they're still failing, then depending on how severe those issues are we may want to make sure those are fixed before releasing this.

willie-yao:
I think a lot of these test scenarios are specific to AKS cluster that don't use ClusterClass, and some operations would be done differently in the context of ClusterClass. For example, if we want to upgrade the Kubernetes version of the cluster, we would want to do it through the cluster topology rather than changing the version of the control plane and then the machinepool as done in the [existing test spec](https://github.com/willie-yao/cluster-api-provider-azure/blob/clusterclass-aks-final/test/e2e/aks_upgrade.go#L56)

willie-yao:
@nojnhuh I just wrote up some quick ClusterClass-specific E2E tests that show how these operations are done differently https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/4155/commits/df2e8eee61081f8226a1b61a8f56d0bea8da88f2

nojnhuh:
> I think a lot of these test scenarios are specific to AKS cluster that don't use ClusterClass, and some operations would be done differently in the context of ClusterClass. For example, if we want to upgrade the Kubernetes version of the cluster, we would want to do it through the cluster topology rather than changing the version of the control plane and then the machinepool as done in the [existing test spec](https://github.com/willie-yao/cluster-api-provider-azure/blob/clusterclass-aks-final/test/e2e/aks_upgrade.go#L56)

I agree that having separate tests for cluster class probably makes the most sense then. I think the test you added looks good at first glance.

test/e2e/aks_public_ip_prefix.go
```diff
@@ -89,10 +89,12 @@ func AKSPublicIPPrefixSpec(ctx context.Context, inputGetter func() AKSPublicIPPr
...
 		Spec: infrav1.AzureManagedMachinePoolSpec{
-			Mode:                 "User",
-			SKU:                  os.Getenv("AZURE_NODE_MACHINE_TYPE"),
-			EnableNodePublicIP:   ptr.To(true),
-			NodePublicIPPrefixID: ptr.To("/subscriptions/" + subscriptionID + "/resourceGroups/" + resourceGroupName + "/providers/Microsoft.Network/publicipprefixes/" + *publicIPPrefix.Name),
+			AzureManagedMachinePoolClassSpec: infrav1.AzureManagedMachinePoolClassSpec{
+				Mode:                 "User",
+				SKU:                  os.Getenv("AZURE_NODE_MACHINE_TYPE"),
+				EnableNodePublicIP:   ptr.To(true),
+				NodePublicIPPrefixID: ptr.To("/subscriptions/" + subscriptionID + "/resourceGroups/" + resourceGroupName + "/providers/Microsoft.Network/publicipprefixes/" + *publicIPPrefix.Name),
```
Jont828, line 96: Is there a function to construct this ID, i.e. maybe in defaults.go?

test/e2e/azure_test.go
```diff
@@ -931,10 +961,10 @@ var _ = Describe("Workload cluster creation", func() {
 
 	Context("Creating clusters using clusterclass [OPTIONAL]", func() {
 		It("with a single control plane node, one linux worker node, and one windows worker node", func() {
-			// use ci-default as the clusterclass name so test infra can find the clusterclass template
+			// Use ci-default as the clusterclass name so test infra can find the clusterclass template
 			os.Setenv("CLUSTER_CLASS_NAME", "ci-default")
 
-			// use "cc" as spec name because natgw pip name exceeds limit.
+			// Use "cc" as spec name because NAT gateway pip name exceeds limit.
```
Jont828, line 982: Nice :)

test/e2e/config/azure-dev.yaml
```diff
@@ -226,4 +230,4 @@ intervals:
   csi-migration/wait-worker-nodes: ["60m", "10s"]
   csi-migration/wait-control-plane: ["60m", "10s"]
   csi-migration/wait-cluster: ["60m", "10s"]
-  node-drain/wait-machine-deleted: [ "10m", "10s" ]
+  node-drain/wait-machine-deleted: [ "10m", "10s" ]
```
Jont828, line 233: Is there supposed to be a change here?

api/v1beta1/azuremanagedcontrolplane_default.go
```diff
@@ -86,72 +98,87 @@ func (m *AzureManagedControlPlane) setDefaultSubnet() {
 	}
 }
 
-func (m *AzureManagedControlPlane) setDefaultSku() {
-	if m.Spec.SKU == nil {
-		m.Spec.SKU = &AKSSku{
-			Tier: FreeManagedControlPlaneTier,
-		}
+func setDefaultSku(sku **AKSSku) {
```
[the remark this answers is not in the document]
willie-yao: Yes I made these changes so these can be used for both `AzureManagedControlPlane` and `AzureManagedControlPlaneTemplate`. In order to change the parameter in-place, a double pointer must be used. I think @nojnhuh suggested this change but maybe there is a better way to do it than the current implementation?

api/v1beta1/azuremanagedcontrolplanetemplate_default.go
```diff
@@ -0,0 +1,69 @@
...
+	setDefault[*string](&mcp.Spec.Template.Spec.NetworkPlugin, ptr.To(AzureNetworkPluginName))
+	setDefault[*string](&mcp.Spec.Template.Spec.LoadBalancerSKU, ptr.To("Standard"))
+
+	if mcp.Spec.Template.Spec.Version != "" && !strings.HasPrefix(mcp.Spec.Template.Spec.Version, "v") {
+		setDefaultVersion(&mcp.Spec.Template.Spec.Version)
+	}
+
+	mcp.setDefaultVirtualNetwork()
+	mcp.setDefaultSubnet()
+	setDefaultSku(&mcp.Spec.Template.Spec.SKU)
```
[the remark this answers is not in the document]
willie-yao: This was how I had it originally but @nojnhuh suggested against this way of implementing it. More details here https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/4155#discussion_r1372010763

api/v1beta1/azuremanagedcontrolplanetemplate_default.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+	if mcp.Spec.Template.Spec.VirtualNetwork.Subnet.CIDRBlock == "" {
+		mcp.Spec.Template.Spec.VirtualNetwork.Subnet.CIDRBlock = defaultAKSNodeSubnetCIDR
+	}
+}
+
+// setDefault sets the default value for a pointer to a value for any comparable type.
+func setDefault[T comparable](field *T, value T) {
+	if field == nil {
+		// shouldn't happen with proper use
```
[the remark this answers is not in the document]
willie-yao: I think this is a part of trying to push for more use of generics. I'm not entirely sure what the disadvantage of not returning the value vs returning it. I think maybe Jon can weigh in a bit again, sorry for my lack of knowledge on this topic! The original suggestion: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/4155#discussion_r1373375563

api/v1beta1/azuremanagedcontrolplanetemplate_webhook.go
```diff
@@ -0,0 +1,293 @@
...
+
+	if old.Spec.Template.Spec.VirtualNetwork.Subnet.Name != mcp.Spec.Template.Spec.VirtualNetwork.Subnet.Name {
+		allErrs = append(allErrs,
+			field.Invalid(
+				field.NewPath("Spec", "Template", "Spec", "VirtualNetwork.Subnet.Name"),
+				mcp.Spec.Template.Spec.VirtualNetwork.Subnet.Name,
+				"Subnet Name is immutable"))
+	}
+
+	// NOTE: This only works because we force the user to set the CIDRBlock for both the
```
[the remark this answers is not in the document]
willie-yao: I copied this from the `AzureManagedControlPlane` webhook [here](https://github.com/willie-yao/cluster-api-provider-azure/blob/clusterclass-aks-final/api/v1beta1/azuremanagedcontrolplane_webhook.go#L581). I think it works the same way in the ClusterClass template as it does in  `AzureManagedControlPlane`.

api/v1beta1/types_class.go
```diff
@@ -66,6 +67,261 @@ type AzureClusterClassSpec struct {
...
+	// +optional
+	Subnet ManagedControlPlaneSubnet `json:"subnet,omitempty"`
+}
+
+// APIServerAccessProfileClassSpec defines the APIServerAccessProfile properties that may be shared across several API server access profiles.
+type APIServerAccessProfileClassSpec struct {
+	// EnablePrivateCluster - Whether to create the cluster as a private cluster or not.
+	// +optional
+	EnablePrivateCluster *bool `json:"enablePrivateCluster,omitempty"`
+	// PrivateDNSZone - Private dns zone mode for private cluster.
```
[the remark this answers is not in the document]
willie-yao: These comments are derived from the original types such as here: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/api/v1beta1/azuremanagedcontrolplane_types.go#L322

Should I change the comments there as well?

test/e2e/aks_public_ip_prefix.go
willie-yao: I created a function for it! Nice catch

api/v1beta1/types_class.go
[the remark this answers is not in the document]
willie-yao: I'm not sure if we should leave it as-is to match the original description or change both of them. Either way I think we need to keep it consistent

willie-yao:
All tests are passing!! 🎉

api/v1beta1/azuremanagedcontrolplane_default.go
```diff
@@ -50,6 +53,15 @@ func (m *AzureManagedControlPlane) setDefaultSSHPublicKey() error {
 	return nil
 }
 
+// setDefaultResourceGroupName sets the default ResourceGroupName for an AzureManagedControlPlane.
+func (m *AzureManagedControlPlane) setDefaultResourceGroupName() {
+	if m.Spec.ResourceGroupName == "" {
+		if clusterName, ok := m.Labels[clusterv1.ClusterNameLabel]; ok {
```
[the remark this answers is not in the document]
willie-yao: It's still required. I don't know why this is part of my diff and it must be left over from a rebase

api/v1beta1/azuremanagedcontrolplane_types.go
```diff
@@ -91,102 +88,22 @@ type AzureManagedControlPlaneSpec struct {
...
-	// Immutable.
-	// +kubebuilder:validation:Enum=azure;kubenet;none
-	// +optional
-	NetworkPlugin *string `json:"networkPlugin,omitempty"`
-
 	// NetworkPluginMode is the mode the network plugin should use.
 	// Allowed value is "overlay".
 	// +kubebuilder:validation:Enum=overlay
 	// +optional
 	NetworkPluginMode *NetworkPluginMode `json:"networkPluginMode,omitempty"`
```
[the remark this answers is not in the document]
willie-yao: Good catch. These were all added after the initial separation of types I did earlier this year. Will move these to the template type.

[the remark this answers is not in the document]
willie-yao: I think we might want to exclude `Identity` though. Is `UserAssignedIdentityResourceID` something we want to be sharing across clusters?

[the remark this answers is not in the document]
nojnhuh: Based on the AKS docs that doesn't seem like it has to be unique per-cluster: https://learn.microsoft.com/en-us/rest/api/aks/managed-clusters/create-or-update?view=rest-aks-2023-08-01&tabs=HTTP#managedclusteridentity

api/v1beta1/azuremanagedcontrolplane_types.go
```diff
@@ -385,6 +291,10 @@ type AzureManagedControlPlaneStatus struct {
 	// OIDCIssuerProfile is the OIDC issuer profile of the Managed Cluster.
 	// +optional
 	OIDCIssuerProfile *OIDCIssuerProfileStatus `json:"oidcIssuerProfile,omitempty"`
+
+	// Version defines the Kubernetes version for the control plane instance.
+	// +optional
+	Version string `json:"version"`
```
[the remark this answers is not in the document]
willie-yao: This field was added to fix an issue with updating the control plane via changing the version through the topology. This is because CAPI looks for the status.version field to be set to mark the amcp as ready 

Setting in CAPZ: https://github.com/willie-yao/cluster-api-provider-azure/blob/clusterclass-aks-final/controllers/azuremanagedcontrolplane_controller.go#L264
CAPI check: https://github.com/willie-yao/cluster-api/blob/fix-mp-cc-e2e/internal/contract/controlplane.go#L153

api/v1beta1/azuremanagedcontrolplane_webhook.go
```diff
@@ -72,41 +70,22 @@ func (mw *azureManagedControlPlaneWebhook) Default(ctx context.Context, obj runt
...
-	}
 
-	if m.Spec.Identity == nil {
-		m.Spec.Identity = &Identity{
-			Type: ManagedControlPlaneIdentityTypeSystemAssigned,
-		}
-	}
+	setDefault[*string](&m.Spec.NetworkPlugin, ptr.To(AzureNetworkPluginName))
+	setDefault[*string](&m.Spec.LoadBalancerSKU, ptr.To("Standard"))
+	m.Spec.Version = setDefaultVersion(m.Spec.Version)
```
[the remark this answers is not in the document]
willie-yao: Oof yeah it's pretty confusing to look at. It would be nice to get the generic `setDefault` to be compatible with non-pointers too. Some defaults also require more checks, so I think it's down to changing all of the type-specific defaults to passing in the field as a parameter.

nojnhuh (commented):
/lgtm

api/v1beta1/azuremanagedcontrolplane_types.go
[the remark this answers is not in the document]
CecileRobertMichon: Is this requirement documented in the CAPI provider contract?

api/v1beta1/azuremanagedcontrolplane_default.go
```diff
@@ -36,6 +39,15 @@ const (
 	defaultAKSNodeSubnetCIDRForOverlay = "10.224.0.0/16"
 )
 
+// setDefaultResourceGroupName sets the default ResourceGroupName for an AzureManagedControlPlane.
+func (m *AzureManagedControlPlane) setDefaultResourceGroupName() {
```
CecileRobertMichon, line 43: is this new defaulting?

api/v1beta1/azuremanagedcontrolplane_webhook.go
```diff
@@ -73,36 +71,27 @@ func (mw *azureManagedControlPlaneWebhook) Default(ctx context.Context, obj runt
 		return apierrors.NewBadRequest("expected an AzureManagedControlPlane")
 	}
 	if m.Spec.NetworkPlugin == nil {
-		networkPlugin := "azure"
+		networkPlugin := AzureNetworkPluginName
```
CecileRobertMichon, line 74: thank you :D

controllers/helpers.go
```diff
@@ -955,6 +955,10 @@ func MachinePoolToAzureManagedControlPlaneMapFunc(ctx context.Context, c client.
 
 		gk := gvk.GroupKind()
 		ref := cluster.Spec.ControlPlaneRef
+		if ref == nil || ref.Name == "" {
+			log.Info("control plane ref is nil or empty: failed to get the control plane ref")
```
CecileRobertMichon, line 959: an info log should contained the word "failed", so probably this either needs to be an error log or it should be reworded if it's purely informative

docs/book/src/topics/clusterclass.md
```diff
@@ -0,0 +1,79 @@
+# ClusterClass
+
+- **Feature status:** GA
```
CecileRobertMichon, line 3: ClusterClass isn't GA, it requires a feature gate?

docs/book/src/topics/clusterclass.md
```diff
@@ -0,0 +1,79 @@
...
+- **Feature gate:** `ClusterTopology=true`
+
+[ClusterClass](https://cluster-api.sigs.k8s.io/tasks/experimental-features/cluster-class/index.html) is a collection of templates that define a topology (control plane and machine deployments) to be used to continuously reconcile one or more Clusters. It is a new Cluster API feature that is built on top of the existing Cluster API resources and provides a set of tools and operations to streamline cluster lifecycle management while maintaining the same underlying API.
```
CecileRobertMichon, line 6:
```suggestion
[ClusterClass](https://cluster-api.sigs.k8s.io/tasks/experimental-features/cluster-class/index.html) is a collection of templates that define a topology (control plane and machine deployments) to be used to continuously reconcile one or more Clusters. It is built on top of the existing Cluster API resources and provides a set of tools and operations to streamline cluster lifecycle management while maintaining the same underlying API.
```

the new part will get old quickly

templates/flavors/aks-clusterclass/azure-managed-cluster-template.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
+kind: AzureManagedClusterTemplate
+metadata:
+  name: ${CLUSTER_NAME}
+  namespace: default
+spec:	
+  template:	
+    spec: {}
+
```
CecileRobertMichon, line 9: nit: extra new line

templates/flavors/aks-clusterclass/azure-managed-machinepool-template.yaml
```diff
@@ -0,0 +1,23 @@
...
+kind: AzureManagedMachinePoolTemplate
+metadata:
+  name: ${CLUSTER_NAME}-pool1
+  namespace: default
+spec:
+  template:
+    spec:
+      mode: User
+      name: pool1
+      sku: ${AZURE_NODE_MACHINE_TYPE}
```
CecileRobertMichon, line 23: nit: missing new line

test/e2e/aks_clusterclass.go
```diff
@@ -0,0 +1,126 @@
...
+		Name:      input.Cluster.Spec.ControlPlaneRef.Name,
+	}, amcp)
+	Expect(err).NotTo(HaveOccurred())
+
+	By("Editing the AzureManagedMachinePoolTemplate to change the scale down mode")
+	ammpt := &infrav1.AzureManagedMachinePoolTemplate{}
+
+	// TODO: We are hard-coding the ammpt name suffix since to get the name from
+	// the ClusterClass, we have to upgrade CAPI to v1.6.0. Fix this once CAPI
+	// v1.6.0 is released.
```
CecileRobertMichon, line 71: can you please add a comment on the v1.6.0 beta PR to remind us to do this?

test/e2e/azure_test.go
```diff
@@ -847,6 +847,52 @@ var _ = Describe("Workload cluster creation", func() {
 		})
 	})
 
+	Context("Creating an AKS cluster using ClusterClass [Managed Kubernetes]", func() {
+		It("with a single control plane node and 1 node", func() {
+			// Use default as the clusterclass name so test infra can find the clusterclass template
+			os.Setenv("CLUSTER_CLASS_NAME", "default")
+
+			// Use "cc" as spec name because NAT gateway pip name exceeds limit.
```
CecileRobertMichon, line 855: 😢

test/e2e/config/azure-dev.yaml
```diff
@@ -227,3 +231,4 @@ intervals:
   csi-migration/wait-control-plane: ["60m", "10s"]
   csi-migration/wait-cluster: ["60m", "10s"]
   node-drain/wait-machine-deleted: [ "10m", "10s" ]
+
```
CecileRobertMichon, line 234: nit: extra new line

CecileRobertMichon (commented):
LGTM overall, just some minor comments/questions

api/v1beta1/azuremanagedcontrolplane_default.go
willie-yao: Yes. I lost track of the exact part in the code this was needed, but the controlle expects there to be a resource group name set and caused the E2E test to fail otherwise. Will try and find the exact issue and update it here

docs/book/src/topics/clusterclass.md
willie-yao: oops!

test/e2e/config/azure-dev.yaml
jackfrancis: outdated

test/e2e/aks_clusterclass.go
CecileRobertMichon: sorry I meant add a comment to the bump PR, as-in in the PR thread

controllers/helpers.go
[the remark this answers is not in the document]
CecileRobertMichon: I made you change this back to error but it looks like there was a previous conversation about changing it to `Info`

According to the above Info is more appropriate, but we also need to change the message so it doesn't say "failed" or "error"

test/e2e/aks_clusterclass.go
CecileRobertMichon: done https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/4182#issuecomment-1814929259

CecileRobertMichon (commented):
/lgtm
/approve

api/v1beta1/consts.go
```diff
@@ -170,4 +170,6 @@ const (
 	AzureManagedControlPlaneKind = "AzureManagedControlPlane"
 	// AzureClusterIdentityKind indicates the kind of an AzureClusterIdentity.
 	AzureClusterIdentityKind = "AzureClusterIdentity"
+	// AzureNetworkPluginName is the name of the Azure network plugin.
+	AzureNetworkPluginName = "azure"
```
CecileRobertMichon, line 174: nit: can we move this out of that const block, it's not a Kind so doesn't belong with the others

api/v1beta1/azuremanagedcontrolplanetemplate_webhook.go
```diff
@@ -0,0 +1,289 @@
...
+
+	if errs := mcp.validateAPIServerAccessProfileTemplateUpdate(old); len(errs) > 0 {
+		allErrs = append(allErrs, errs...)
+	}
+
+	if len(allErrs) == 0 {
+		return nil, mcp.validateManagedControlPlaneTemplate(mcpw.Client)
+	}
+
+	return nil, apierrors.NewInvalid(GroupVersion.WithKind("AzureManagedControlPlaneTemplate").GroupKind(), mcp.Name, allErrs)
```
CecileRobertMichon, line 179: now that we have consts for kinds, would be good to also use constants for these new kinds 

Can be a follow up since that other PR just merged

api/v1beta1/consts.go
CecileRobertMichon: can do in follow up

willie-yao:
> let's merge this to get it into the release. Can you please open a follow-up PR to include the new kinds as constants following https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/4233 ?

Will do, thanks!

CecileRobertMichon:
Bypassing test rerun to speed up release as the last main commit is a metadata.yaml only update (#4284) and we have green signal on the current commit already from the previous run: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-azure/4155/pull-cluster-api-provider-azure-e2e-aks/1725228064436129792

PR #4158, opened 2023-10-20
workaround ASO swallowing updates into ongoing operations

nojnhuh:
**What type of PR is this?**
/kind bug

This PR works around a bug in ASO (https://github.com/Azure/azure-service-operator/issues/3451) where updating the spec of an ASO resource while another update is already in progress swallows the requeue request from the new update and the new update will not take effect until after the next periodic ASO resync which could be up to an hour later by default.

Fixes tests on #4069

- [X] cherry-pick candidate

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Release note**:

```release-note
Fixed a bug causing some updates to take effect significantly later than before
```

### DIFF
--- a/azure/services/aso/aso.go
+++ b/azure/services/aso/aso.go
@@ -122,11 +122,15 @@ func (r *reconciler[T]) CreateOrUpdateResource(ctx context.Context, spec azure.A
 				// update instead of returning early.
 				adopt = true
 			case cond.Reason == conditions.ReasonReconciling.Name:
-				readyErr = azure.NewOperationNotDoneError(&infrav1.Future{
+				// Updating the spec of an ASO resource that is already being updated will swallow this update
+				// and ignore it until the next periodic ASO resync.
+				// ref: https://github.com/Azure/azure-service-operator/issues/3451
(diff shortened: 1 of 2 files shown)


nawazkh (approved):
/lgtm

nojnhuh:
/hold
I'm running this alongside #4069 and seeing the tests fail further down the line. Want to check and see if it's related to this change.

nojnhuh:
> /hold I'm running this alongside #4069 and seeing the tests fail further down the line. Want to check and see if it's related to this change.

I'm reasonably confident this failure is unrelated to the rest of these changes. I think it's a weird concurrency thing trying to update managed clusters and agent pools at the same time like we do in the tags test. The real bug is likely in AKS and/or ASO, but I think we can work around that by using the plain tags API like we were doing before instead of using ASO to manage those. I'll update #4069 with that change but I think this change is still valuable.

/hold cancel
/assign @CecileRobertMichon

CecileRobertMichon (commented):
/lgtm
/approve

PR #4177, opened 2023-10-25
fix default azurecluster subnet webhook

LochanRn:
**What type of PR is this?**
/kind bug

Fixes the bug in azure_cluster_default and azure_cluster_template_default webhhoks.

Default security groups for node subnet are not being set. 

Fixes #

- [ ] cherry-pick candidate

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/api/v1beta1/azurecluster_default.go
+++ b/api/v1beta1/azurecluster_default.go
@@ -119,7 +119,7 @@ func (c *AzureCluster) setSubnetDefaults() {
 		if subnet.SecurityGroup.Name == "" {
 			subnet.SecurityGroup.Name = generateNodeSecurityGroupName(c.ObjectMeta.Name)
 		}
-		cpSubnet.SecurityGroup.SecurityGroupClass.setDefaults()
+		subnet.SecurityGroup.SecurityGroupClass.setDefaults()
 
 		if subnet.RouteTable.Name == "" {
 			subnet.RouteTable.Name = generateNodeRouteTableName(c.ObjectMeta.Name)
--- a/api/v1beta1/azurecluster_default_test.go
+++ b/api/v1beta1/azurecluster_default_test.go
@@ -703,6 +703,30 @@ func TestSubnetDefaults(t *testing.T) {
 									Name: "my-custom-sg",
 								},
 							},
+							{
+								SubnetClassSpec: SubnetClassSpec{
+									Role: "node",
+									Name: "cluster-test-node-subnet",
+								},
+								SecurityGroup: SecurityGroup{
+									SecurityGroupClass: SecurityGroupClass{
+										SecurityRules: []SecurityRule{
+											{
+												Name:             "allow_port_50000",
+												Description:      "allow port 50000",
+												Protocol:         "*",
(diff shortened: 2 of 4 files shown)


sonasingh46:
@LochanRn -- Do we have an issue describing what bug it fixes. If yes, please link it to the PR. 
If no, can you please put a little description so that we know what bug is it.

LochanRn:
> @LochanRn -- Do we have an issue describing what bug it fixes. If yes, please link it to the PR. If no, can you please put a little description so that we know what bug is it.

@sonasingh46 The issue is default security groups for node subnet are not being set. 
Updated the description.

mboersma (commented):
/lgtm

Thanks for finding and fixing this @LochanRn!

api/v1beta1/azurecluster_default.go
```diff
@@ -119,7 +119,7 @@ func (c *AzureCluster) setSubnetDefaults() {
 		if subnet.SecurityGroup.Name == "" {
 			subnet.SecurityGroup.Name = generateNodeSecurityGroupName(c.ObjectMeta.Name)
 		}
-		cpSubnet.SecurityGroup.SecurityGroupClass.setDefaults()
+		subnet.SecurityGroup.SecurityGroupClass.setDefaults()
```
CecileRobertMichon, line 122: I'm not sure that is a bug or by design, do the defaults apply to node subnets as well

https://capz.sigs.k8s.io/topics/custom-vnet#custom-security-rules for reference

CecileRobertMichon: Also, is this a recent regression or it's always been like this?

LochanRn: @CecileRobertMichon It looks like a regression when we transitioned from release-1.1 to release-1.2. Quite an old one I guess 

release-1.1: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/4a53842ec20f3a9fcdfcbb6605d5a32c4737ea49/api/v1beta1/azurecluster_default.go#L121

release-1.2: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/357e0b7a97a03f2481cf85bf34cfb8ea13747913/api/v1beta1/azurecluster_default.go#L120

I was actually working on providing support for a single subnet for both cp and node, hence was going though the code and found it weird that we are updating the cpSubnet security group rules in the for loop when it was already updated over [here](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/6987809212f86a583c7ab0b4fc47a7a0be8a5c3e/api/v1beta1/azurecluster_default.go#L97)

Hence felt it was a bug, confirmed it with @mboersma on slack and then raised a PR :smile:

CecileRobertMichon (commented):
/lgtm
/approve

I wonder if we should backport this? On one hand, it's a bug so should be backported, but on the other hand it's quite a big behavior change and the regression is so old I wouldn't be surprised if there are users out there relying on this bug...

Issue #3528, opened 2023-05-04
ASO service: virtualnetworks and subnets


nojnhuh:
Convert the virtualnetworks and subnets services to use ASO. These two are co-dependent, where ASO subnets must refer to an owning ASO vnet, and I don't remember exactly why vnets depend on subnets 🙃.

```[tasklist]
### Dependencies
- [ ] #3527
```

PR #4191, opened 2023-10-26
fix vnet in separate resource group

nojnhuh:
**What type of PR is this?**
/kind cleanup

This PR updates the custom vnet e2e test to use different resource groups for the vnet and the rest of the cluster resources. I was trying to test this same scenario for #3528 and was running into issues that didn't seem related, so I'd like to make sure this scenario works on main first.

Fixes #

- [ ] cherry-pick candidate

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Fixed a bug where AzureCluster's spec.networkSpec.vnet.resourceGroup was not always properly honored
```

### DIFF
--- a/azure/scope/cluster.go
+++ b/azure/scope/cluster.go
@@ -310,7 +310,7 @@ func (s *ClusterScope) RouteTableSpecs() []azure.ResourceSpecGetter {
 			specs = append(specs, &routetables.RouteTableSpec{
 				Name:           subnet.RouteTable.Name,
 				Location:       s.Location(),
-				ResourceGroup:  s.ResourceGroup(),
+				ResourceGroup:  s.Vnet().ResourceGroup,
 				ClusterName:    s.ClusterName(),
 				AdditionalTags: s.AdditionalTags(),
 			})
@@ -358,7 +358,7 @@ func (s *ClusterScope) NSGSpecs() []azure.ResourceSpecGetter {
 		nsgspecs[i] = &securitygroups.NSGSpec{
 			Name:                     subnet.SecurityGroup.Name,
 			SecurityRules:            subnet.SecurityGroup.SecurityRules,
-			ResourceGroup:            s.ResourceGroup(),
+			ResourceGroup:            s.Vnet().ResourceGroup,
 			Location:                 s.Location(),
 			ClusterName:              s.ClusterName(),
 			AdditionalTags:           s.AdditionalTags(),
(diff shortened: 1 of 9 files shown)


azure/scope/cluster.go
```diff
@@ -310,7 +310,7 @@ func (s *ClusterScope) RouteTableSpecs() []azure.ResourceSpecGetter {
 			specs = append(specs, &routetables.RouteTableSpec{
 				Name:           subnet.RouteTable.Name,
 				Location:       s.Location(),
-				ResourceGroup:  s.ResourceGroup(),
+				ResourceGroup:  s.Vnet().ResourceGroup,
```
nojnhuh, line 313: The first discrepancy I noticed was that the e2e test sets up the vnet with the route table and NSGs in the vnet resource group, but here we create them in the main resource group. Which one is correct?

cc @CecileRobertMichon

nojnhuh: *The first and only it seems.

nojnhuh:
This should be ready for review!

/retitle use different rg for vnet in custom vnet e2e test

azure/scope/cluster.go
CecileRobertMichon: Is this potentially a breaking change for existing clusters? what if the route table was previously created in the cluster RG?

nojnhuh: Vnet resource group gets defaulted to the cluster RG in the webhooks, so I don't think would result in any changes to existing clusters.

CecileRobertMichon: NSG Spec is also using `s.Vnet().ResourceGroup` so this does look like a bug

test/e2e/azure_test.go
```diff
@@ -155,6 +156,7 @@ var _ = Describe("Workload cluster creation", func() {
 				clusterName = getClusterName(clusterNamePrefix, "public-custom-vnet")
 				By("Creating a custom virtual network", func() {
 					Expect(os.Setenv(AzureCustomVNetName, "custom-vnet")).To(Succeed())
+					Expect(os.Setenv(AzureCustomVnetResourceGroup, clusterName+"-vnetrg")).To(Succeed())
```
CecileRobertMichon, line 159: is everything getting cleaned up as it should now that we're using a separate RG?

nojnhuh: Not sure, but I'll check

nojnhuh: Yep, I don't see any extra resource groups around after running the test locally.

CecileRobertMichon (commented):
/lgtm
/approve

nojnhuh:
/retitle fix vnet in separate resource group
/release-note-edit
```release-note
Fixed a bug where AzureCluster's spec.networkSpec.vnet.resourceGroup was not always properly honored
```
/remove-kind cleanup
/kind bug

nawazkh:
lgtm++. Nice fix!

PR #4201, opened 2023-10-30
Make identityRef required on AzureManagedControlPlane

CecileRobertMichon:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it**: IdentityRef is required since https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3793, fixing the CRD to reflect that.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

- [x] cherry-pick candidate

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
 identityRef is required on AzureManagedControlPlane Spec
```

PR #4208, opened 2023-10-31
Remove unused function setCredentials

willie-yao:
**What type of PR is this?**
/kind cleanup

`setCredentials` is unused since identityRef is now required as of #4201 

Fixes #

- [ ] cherry-pick candidate

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```


CecileRobertMichon:
@willie-yao unit tests are failing PTAL

willie-yao:
@CecileRobertMichon Turns out a ton of unit tests need to be refactored (basically all the controller tests) due to them needing to expect AzureClusterIdentity for all cases, so it may take me a bit to get all that done. Will mark the PR as WIP

azure/scope/identity.go
```diff
@@ -239,7 +239,8 @@ func (p *AzureCredentialsProvider) GetClientSecret(ctx context.Context) (string,
 		if err := p.Client.Get(ctx, key, secret); err != nil {
 			return "", errors.Wrap(err, "Unable to fetch ClientSecret")
 		}
-		return string(secret.Data[AzureSecretKey]), nil
+
+		return string(secret.Data["AZURE_CLIENT_SECRET"]), nil
```
willie-yao, line 243: I think I may have found a bug here. The secret is stored under `"AZURE_CLIENT_SECRET"`, but `AzureSecretKey` is set to `"clientSecret"`, which returns an empty secret name. @nojnhuh can you confirm if this is a bug or is this intentional?

nojnhuh: I believe this should be `clientSecret`, at least to stay in sync with this:
https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/9d2157f31d73e3cacda925df242c8720c262df90/hack/create-identity-secret.sh#L32

I'd be surprised if this were an existing bug given the state of the tests seems green overall.

CecileRobertMichon: `clientSecret` is correct according to https://capz.sigs.k8s.io/topics/multitenancy#aad-pod-identity-using-service-principal-with-client-password-deprecated

nojnhuh:
> controllers/azurecluster_controller_test.go:277:15: string `default` has 3 occurrences, make it a constant (goconst)
>	namespace := "default"

Ugh, #4326

willie-yao:
There is quite a lot of repeated code that had to be included to fix the unit tests. I found that this pattern is pre-existing and I think a separate PR to consolidate all shared test variables and cleanup the test files would be helpful.

CecileRobertMichon:
I will wait for this one to merge before starting on #4249 to avoid too many conflicts

azure/scope/cluster_test.go
```diff
@@ -77,6 +83,9 @@ func TestAPIServerHost(t *testing.T) {
 				Spec: infrav1.AzureClusterSpec{
 					AzureClusterClassSpec: infrav1.AzureClusterClassSpec{
 						SubscriptionID: fakeSubscriptionID,
+						IdentityRef: &corev1.ObjectReference{
```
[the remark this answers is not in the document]
willie-yao: I agree it is inconsistent as some identity refs include a name/namespace while others don't. This just ends up working with an empty secret object and is a pattern I found in other tests.

controllers/azuremachine_controller_test.go
```diff
@@ -714,29 +758,45 @@ func TestConditions(t *testing.T) {
...
+			g.Expect(fakeClient.Get(context.TODO(), key, resultIdentity))
 			recorder := record.NewFakeRecorder(10)
 
-			reconciler := NewAzureMachineReconciler(client, recorder, reconciler.DefaultLoopTimeout, "")
+			reconciler := NewAzureMachineReconciler(fakeClient, recorder, reconciler.DefaultLoopTimeout, "")
 
 			clusterScope, err := scope.NewClusterScope(context.TODO(), scope.ClusterScopeParams{
-				Client:       client,
+				AzureClients: scope.AzureClients{
+					Authorizer: autorest.NullAuthorizer{},
```
mboersma, line 790: Is this line required? I ask because `azure/go-autorest` is unsupported, and in trying to fix #2974 I found the use of `NullAuthorizer` generally to be dead code that I could remove and still have the tests pass.

controllers/azuremanagedcontrolplane_controller_test.go
```diff
@@ -105,6 +106,8 @@ func TestAzureManagedControlPlaneReconcilePaused(t *testing.T) {
 		infrav1.AddToScheme,
 		asoresourcesv1.AddToScheme,
 		asocontainerservicev1.AddToScheme,
+		corev1.AddToScheme,
+		aadpodv1.AddToScheme,
```
mboersma, line 110: I assume this is needed for the test to pass? But ideally we wouldn't be adding usage of AAD Pod ID; see #4249.

azure/scope/cluster_test.go
```diff
@@ -169,8 +186,20 @@ func TestAPIServerHost(t *testing.T) {
...
+		}
+		fakeSecret := &corev1.Secret{}
+
+		initObjects := []runtime.Object{cluster, &tc.azureCluster, fakeIdentity, fakeSecret}
 		fakeClient := fake.NewClientBuilder().WithScheme(scheme).WithRuntimeObjects(initObjects...).Build()
+		resultIdentity := &infrav1.AzureClusterIdentity{}
+		key := client.ObjectKey{Name: fakeIdentity.Name, Namespace: fakeIdentity.Namespace}
+		g.Expect(fakeClient.Get(context.TODO(), key, resultIdentity)).To(Succeed())
 
 		clusterScope, err := NewClusterScope(context.TODO(), ClusterScopeParams{
```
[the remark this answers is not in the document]
nojnhuh: Whatever's easier for you, but I suspect that might be a separate PR at this point.

controllers/azuremachine_controller_test.go
willie-yao: Looks like removing it had no effect! Nice catch.

controllers/azuremanagedcontrolplane_controller_test.go
willie-yao: Removing it causes this error:
```
failed to create scope: failed to configure azure settings and credentials for Identity: failed to create copied AzureIdentity paused-kqvhtglxrf-default-fake-identity in capz-system: no kind is registered for the type v1.AzureIdentity in scheme "pkg/runtime/scheme.go:100"
```
This error happens on this [line](https://github.com/willie-yao/cluster-api-provider-azure/blob/remove-setcredentials/azure/scope/identity.go#L297). Is this something you'd remove in your PR?

mboersma: Yes, I can address that when fixing #4249.

mboersma (commented):
/lgtm
/assign @nojnhuh

nojnhuh (commented):
/lgtm
/approve

@willie-yao Could you please open an issue describing any follow-up refactoring for tests?

nojnhuh:
@willie-yao Was the dependabot PR in the commit history here intentional?

willie-yao:
>@willie-yao Was the dependabot PR in the commit history here intentional?

Whoops lemme fix that

nojnhuh (commented):
/lgtm

mboersma (approved):
/lgtm
/approve

willie-yao:
Note: I'm holding off on making a cherry-pick to 1.11 for now since the ClusterClass for Managed Clusters PR was not merged for that release which causes massive merge conflicts. @CecileRobertMichon Would we want to do this cherry-pick even though there isn't a functional difference?

CecileRobertMichon:
If this is just a cleanup (no bug fix) + refactor then we don't believe we want to cherry-pick this (even in 1.12)

PR #4209, opened 2023-10-31
Fix containerd and systemd-resolved race condition in ipv6 clusters

CecileRobertMichon:
**What type of PR is this?**

Fixes #

- [x] cherry-pick candidate

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Fix containerd and systemd-resolved race condition in ipv6 templates
```

### DIFF
--- a/templates/test/ci/prow-ipv6/machine-pool-ipv6.yaml
+++ b/templates/test/ci/prow-ipv6/machine-pool-ipv6.yaml
@@ -78,4 +78,4 @@ spec:
   - echo "DNSStubListener=no" >> /etc/systemd/resolved.conf
   - mv /etc/resolv.conf /etc/resolv.conf.OLD && ln -s /run/systemd/resolve/resolv.conf
     /etc/resolv.conf
-  - systemctl restart systemd-resolved
+  - systemctl restart systemd-resolved containerd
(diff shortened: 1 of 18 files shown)


templates/test/ci/prow-ipv6/machine-pool-ipv6.yaml
```diff
@@ -74,8 +74,3 @@ spec:
         cloud-provider: azure
         cluster-dns: '[fd00::10]'
       name: '{{ ds.meta_data["local_hostname"] }}'
-  postKubeadmCommands:
```
CecileRobertMichon, line 77: so this only works with new versions of containerd (1.6.24+ AFAICT, this is because "Newer versions of Go will no longer use the system DNS resolver by default" and those are the versions that were built with new go versions). The latest VM images have containerd 1.7.6 and no longer need this workaround (in fact it causes a race condition if it's still in place). However, any VM image with containerd v1.6.23 (which is what the previous images built by image-builder had) will fail with these changes. 

So the alternatives are:
1) we keep the workaround for now and add the containerd restart to mitigate the race condition (that's the first commit in this PR)
2) we remove the workaround and assume that anyone using these new templates will use a newer version of containerd (we can't document this)

Thoughts? @mboersma

mboersma: @CecileRobertMichon thank you for pinning this down!

As much as I would like to move forward and get rid of the workaround, I think option `1.` is safer.

CecileRobertMichon: I'm afraid you're right. I dropped the last commit and opened a follow up issue for myself to drop the workaround later https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/4210

nojnhuh (commented):
/lgtm
/approve

Issue #3649, opened 2023-06-19
Improve unit test coverage in the controllers package


willie-yao:
/kind cleanup

**What needs cleanup:**
This mega issue is meant to track individual issues for improving code coverage within the `controllers` package.

**Describe the solution you'd like**

```[tasklist]
### Controllers Needing Unit Tests
- [ ] https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/3360
- [ ] https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/4114
- [ ] https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/4227
- [ ] https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/4267
- [ ] azurejson_machine_controller.go
- [ ] azurejson_machinepool_controller.go
- [ ] azurejson_machinetemplate_controller.go
- [ ] azuremanagedcluster_controller.go
- [ ] azuremanagedcontrolplane_controller.go
- [ ] azuremanagedcontrolplane_reconciler.go
- [ ] azuremanagedmachinepool_controller.go
- [ ] azuremanagedmachinepool_reconciler.go
- [ ] helpers.go
```

Older parent issue: https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/3229

aniruddha2000:
Hi @willie-yao I can take this one by one, but I need little more context, can youhelp?

willie-yao:
Awesome @aniruddha2000, thanks for working on this! Basically, most controllers and reconcilers within the `controllers` package do not have unit tests, or have very basic ones. I have created more detailed unit tests for the AzureMachine controller and reconciler as an example, and you can implement tests for other controllers similar to it: #3360

aniruddha2000:
@willie-yao can you create the issue for next azurecluster_reconciler.go as well? I will start this week if possible

aniruddha2000:
the AzureClusterReconciler is done as we discuss, can you create the issue for the next one @willie-yao ?

willie-yao:
@aniruddha2000 Done, feel free to assign yourself when you're ready to work on it!

troy0820:
@willie-yao I was going to create an issue for the azuremanagedcluster_controller unit test and assign it to myself.  I can update this issue when that is done.

troy0820:
The issue is linked above but here is the link to edit the intial description: 
[...]

PR #4211, opened 2023-11-01
eliminate short-circuiting in logic to add/remove finalizers

nojnhuh:
**What type of PR is this?**
/kind bug

Go will short-circuit compound boolean expressions like `A || B` such that if `A` is true, then `B` will not be evaluated. The `Add/RemoveFinalizer` helpers evaluate to booleans but also perform side-effects, so if both parts of `RemoveFinalizer() || AddFinalizer()` are true, a finalizer will only be removed and one will not be added, where both changes should occur.

This is a toy example to demonstrate this behavior: https://go.dev/play/p/LCRHzzBzrYE

Fixes #

I don't know if this is possible to reproduce in a real-world scenario, but seems worth the change anyway to make sure finalizers get applied the way we expect. Neither of these cases are covered by unit tests currently.

- [X] cherry-pick candidate

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Fixed a bug potentially causing finalizers not to be applied in some cases
```

### DIFF
--- a/controllers/azuremanagedcontrolplane_controller.go
+++ b/controllers/azuremanagedcontrolplane_controller.go
@@ -226,8 +226,9 @@ func (amcpr *AzureManagedControlPlaneReconciler) reconcileNormal(ctx context.Con
 	log.Info("Reconciling AzureManagedControlPlane")
 
 	// Remove deprecated Cluster finalizer if it exists, if the AzureManagedControlPlane doesn't have our finalizer, add it.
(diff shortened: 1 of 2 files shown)


nojnhuh:
> Should we create an issue to add tests?

I think https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/3649 might cover this already.

CecileRobertMichon (commented):
/lgtm
/approve

PR #4225, opened 2023-11-02
Set correct fqdn on private cluster

mihaiandreiratoiu:
**What type of PR is this?**

/kind bug

On private cluster with the managedcontrolplane
- EnablePrivateCluster: true
- EnablePrivateClusterPublicFQDN: false
The fqdn property is not available. The cluster defaults to the `privateFQDN`
The Cluster resource remains in provisioning state due to the capz error 
``` 
spec.controlPlaneEndpoint.host: Required value" "AzureManagedCluster"
```

**TODOs**:

- [X] squashed commits
- [X] includes documentation
- [X] adds unit tests

**Release note**:

```release-note
 Fix: Set correct host on private cluster with public fqdn disabled
```

### DIFF
--- a/azure/services/managedclusters/managedclusters.go
+++ b/azure/services/managedclusters/managedclusters.go
@@ -109,6 +109,14 @@ func (s *Service) Reconcile(ctx context.Context) error {
 			Host: ptr.Deref(managedCluster.Properties.Fqdn, ""),
 			Port: 443,
 		}
+		if managedCluster.Properties.APIServerAccessProfile != nil &&
+			ptr.Deref(managedCluster.Properties.APIServerAccessProfile.EnablePrivateCluster, false) &&
+			!ptr.Deref(managedCluster.Properties.APIServerAccessProfile.EnablePrivateClusterPublicFQDN, false) {
+			endpoint = clusterv1.APIEndpoint{
+				Host: ptr.Deref(managedCluster.Properties.PrivateFQDN, ""),
+				Port: 443,
+			}
+		}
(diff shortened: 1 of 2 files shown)


mihaiandreiratoiu:
Tested with develop branch and somehow I get this error. `is invalid: Spec.DNSPrefix: Invalid value: "null": field is immutable, unable to set an empty value if it was already set` I'm thinking this is unrelated because when testing from the release branch it doesn't give this error

ionutleca:
Any chance the error from the main branch is caused by how the defaulter sets the spec.DNSPrefix after this [commit](https://github.com/kubernetes-sigs/cluster-api-provider-azure/commit/24c7651cec6d6a7268f50ebef30c4886259c7a57)?

CecileRobertMichon (commented):
/lgtm
/assign @nojnhuh

CecileRobertMichon:
@mihaiandreiratoiu thanks for the fix. Should this be cherry-picked to the release branches?

mihaiandreiratoiu:
It's not the same code unfortunately, the ptr was updated from pointer

mboersma (approved):
/lgtm
/approve

mboersma:
/hold

So @nojnhuh can take a look as well.

nojnhuh (commented):
/lgtm
/hold cancel

> @mihaiandreiratoiu thanks for the fix. Should this be cherry-picked to the release branches?
>
> It's not the same code unfortunately, the ptr was updated from pointer

@mihaiandreiratoiu Could you please open up PRs manually to the release branches (release-1.10, release-1.11) with this fix that incorporate that change?

PR #4239, opened 2023-11-06
Remove use of Azure/go-autorest/tracing package

mboersma:
**What type of PR is this?**

/kind cleanup

Removes usage of the `github.com/Azure/go-autorest/tracing` package. That project is unsupported, and since CAPZ uses Azure SDKv2 and ASO now, this code has effectively been unreachable.

Refs #2974

I tested manually with `make tilt-up` and tracing seems to be working as before.

- [ ] cherry-pick candidate

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```


go.mod
```diff
@@ -71,6 +69,7 @@ require (
 	github.com/Azure/go-autorest/autorest/to v0.4.0 // indirect
 	github.com/Azure/go-autorest/autorest/validation v0.3.1 // indirect
 	github.com/Azure/go-autorest/logger v0.2.1 // indirect
+	github.com/Azure/go-autorest/tracing v0.6.0 // indirect
```
willie-yao, line 72: This might be due to my lack of understanding of `go.mod` files, but shouldn't any mention of this package be removed? I see that it is moved here from above.

mboersma: I thought that was confusing as well, but it persists after `make modules` / `go mod tidy`. I assume it's a dependency of some other package we do import.

nojnhuh (commented):
/lgtm
/approve

Issue #4277, opened 2023-11-15
Add unit tests to `azurejson_machinepool_controller.go`


troy0820:
/kind cleanup
/assign @troy0820 

**What needs cleanup:**
- [x] Add unit test to controller azurejson_machinepool_controller.go

Part of a larger issue https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/3649

PR #4278, opened 2023-11-15
Create test to cover path of azureMachinePool.Spec.UserAssignedIdenti…

troy0820:
**What type of PR is this?**
/kind cleanup 

Adds testing to the `controllers` package 

- [x] Add unit test for controller azurejson
- [x] Create func to monkey patch during testing
- [x] Export interface instead of concrete type for testing 

Fixes #https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/4277
Partially fixes: https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/3649

_I had to change the return value to an interface _(which the return type implements)_ to allow the patch to ensure mocking can take place.  The mocking will use the generated mocks and the EXPECT() to allow to hit that code path while testing._ 

- [ ] cherry-pick candidate

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:
```release-note
NONE
```

### DIFF
--- a/azure/services/identities/client.go
+++ b/azure/services/identities/client.go
@@ -39,7 +39,7 @@ type AzureClient struct {
 }
 
 // NewClient creates a new MSI client from an authorizer.
-func NewClient(auth azure.Authorizer) (*AzureClient, error) {
+func NewClient(auth azure.Authorizer) (Client, error) {
 	opts, err := azure.ARMClientOptions(auth.CloudEnvironment())
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to create identities client options")
--- a/controllers/azurejson_machinepool_controller.go
+++ b/controllers/azurejson_machinepool_controller.go
@@ -149,7 +149,7 @@ func (r *AzureJSONMachinePoolReconciler) Reconcile(ctx context.Context, req ctrl
 	// Construct secret for this machine
 	userAssignedIdentityIfExists := ""
 	if len(azureMachinePool.Spec.UserAssignedIdentities) > 0 {
-		idsClient, err := identities.NewClient(clusterScope)
+		idsClient, err := getClient(clusterScope)
 		if err != nil {
 			return reconcile.Result{}, errors.Wrap(err, "failed to create identities client")
 		}
@@ -194,3 +194,5 @@ func (r *AzureJSONMachinePoolReconciler) Reconcile(ctx context.Context, req ctrl
 
 	return ctrl.Result{}, nil
 }
+
+var getClient = identities.NewClient
--- a/controllers/azurejson_machinepool_controller_test.go
+++ b/controllers/azurejson_machinepool_controller_test.go
@@ -18,18 +18,26 @@ package controllers
 
 import (
 	"context"
-	"os"
+	"fmt"
 	"testing"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
 	"github.com/Azure/go-autorest/autorest/azure/auth"
 	"github.com/google/go-cmp/cmp"
+	. "github.com/onsi/gomega"
+	"go.uber.org/mock/gomock"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
+	"k8s.io/utils/ptr"
 	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
+	"sigs.k8s.io/cluster-api-provider-azure/azure"
+	"sigs.k8s.io/cluster-api-provider-azure/azure/services/identities"
+	"sigs.k8s.io/cluster-api-provider-azure/azure/services/identities/mock_identities"
 	infrav1exp "sigs.k8s.io/cluster-api-provider-azure/exp/api/v1beta1"
+	"sigs.k8s.io/cluster-api-provider-azure/util/reconciler"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	expv1 "sigs.k8s.io/cluster-api/exp/api/v1beta1"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -178,9 +186,9 @@ func TestAzureJSONPoolReconciler(t *testing.T) {
 		},
 	}
 
-	os.Setenv(auth.ClientID, "fooClient")
-	os.Setenv(auth.ClientSecret, "fooSecret")
-	os.Setenv(auth.TenantID, "fooTenant")
+	t.Setenv(auth.ClientID, "fooClient")
+	t.Setenv(auth.ClientSecret, "fooSecret")
+	t.Setenv(auth.TenantID, "fooTenant")
 
 	for name, tc := range cases {
 		t.Run(name, func(t *testing.T) {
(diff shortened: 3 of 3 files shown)


azure/services/identities/client.go
```diff
@@ -39,7 +39,7 @@ type AzureClient struct {
 }
 
 // NewClient creates a new MSI client from an authorizer.
-func NewClient(auth azure.Authorizer) (*AzureClient, error) {
+func NewClient(auth azure.Authorizer) (Client, error) {
```
troy0820, line 42: This doesn't change anything but allow this interface to be mocked as a return value for testing.  The concrete type is still in use here as the return value for that is the concrete type which implements this interface.

controllers/azurejson_machinepool_controller.go
```diff
@@ -194,3 +194,7 @@ func (r *AzureJSONMachinePoolReconciler) Reconcile(ctx context.Context, req ctrl
 
 	return ctrl.Result{}, nil
 }
+
+var getClient = func(auth azure.Authorizer) (identities.Client, error) {
```
troy0820, line 198: This "patch" allows for this to be overridden during testing but not in production.

controllers/azurejson_machinepool_controller_test.go
```diff
@@ -18,17 +18,24 @@ package controllers
 
 import (
 	"context"
-	"os"
```
troy0820, line 21: Removed to use `t.Setenv`

controllers/azurejson_machinepool_controller.go
```diff
@@ -114,7 +114,6 @@ func (r *AzureJSONMachinePoolReconciler) Reconcile(ctx context.Context, req ctrl
 		}
 		return reconcile.Result{}, err
 	}
-
```
willie-yao, line 117: nit: You probably didn't mean to make this change

controllers/azurejson_machinepool_controller_test.go
```diff
@@ -210,3 +217,140 @@ func TestAzureJSONPoolReconciler(t *testing.T) {
...
 	}
 }
+
+func TestAzureJSONPoolReconcilerUserAssignedIdentities(t *testing.T) {
+	g := NewWithT(t)
+	ctrlr := gomock.NewController(t)
+	defer ctrlr.Finish()
+	req := ctrl.Request{NamespacedName: types.NamespacedName{Name: "fake-machine-pool", Namespace: "fake-ns"}}
+	ctx := context.Background()
+	scheme := runtime.NewScheme()
```
willie-yao, line 227: Similar to the other PR, I think you can use this existing function: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/controllers/azurejson_machine_controller_test.go#L240

controllers/azurejson_machinepool_controller_test.go
```diff
@@ -210,3 +217,140 @@ func TestAzureJSONPoolReconciler(t *testing.T) {
...
+		Spec: infrav1exp.AzureMachinePoolSpec{
+			UserAssignedIdentities: []infrav1.UserAssignedIdentity{
+				{
+					ProviderID: "fake-id",
+				},
+			},
+		},
+	}
+
+	clusterA := &clusterv1.Cluster{
```
willie-yao, line 269: nit: I think naming this just `cluster` would be better since the "A" implies there's another cluster (or at least that's how I see it)

controllers/azurejson_machinepool_controller_test.go
```diff
@@ -210,3 +217,140 @@ func TestAzureJSONPoolReconciler(t *testing.T) {
...
+	azureMP := &infrav1exp.AzureMachinePool{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "fake-machine-pool",
+			Namespace: "fake-ns",
+			Labels: map[string]string{
+				clusterv1.ClusterNameLabel: "fake-cluster",
+			},
+			OwnerReferences: []metav1.OwnerReference{
+				{
+					APIVersion: expv1.GroupVersion.Group + "/" + expv1.GroupVersion.Version,
```
willie-yao, line 253: nit: I think the convention in tests is to use fmt.Sprintf for concats. Example: https://github.com/willie-yao/cluster-api-provider-azure/blob/aks-fleet/api/v1beta1/azuremachine_validation_test.go#L201

controllers/azurejson_machinepool_controller_test.go
```diff
@@ -210,3 +217,140 @@ func TestAzureJSONPoolReconciler(t *testing.T) {
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "fake-other-machine-pool",
+			Namespace: "fake-ns",
+			Labels: map[string]string{
+				clusterv1.ClusterNameLabel: "fake-cluster",
+			},
+		},
+	}
+
+	azureCluster := &infrav1.AzureCluster{
```
willie-yao, line 293: I think you can use this `getFakeAzureCluster` function that exists here: https://github.com/willie-yao/cluster-api-provider-azure/blob/aks-fleet/controllers/azuremachine_controller_test.go#L529

However, I think there could be a case to be made to create a new file that contains all of these functions for tests, since it's a bit confusing they exist in specific controller tests but can be used across the package. I think this could be done in a follow-up PR since it'll be a bit out of scope for this issue.

controllers/azurejson_machinepool_controller_test.go
```diff
@@ -210,3 +217,140 @@ func TestAzureJSONPoolReconciler(t *testing.T) {
...
+	}
+
+	client := fake.NewClientBuilder().WithScheme(scheme).WithObjects(azureMP, ownerMP, clusterA, azureCluster, sec).Build()
+	rec := AzureJSONMachinePoolReconciler{
+		Client:           client,
+		Recorder:         record.NewFakeRecorder(42),
+		ReconcileTimeout: 20 * time.Second,
+	}
+	id := "fake-id"
+	getClient = func(auth azure.Authorizer) (identities.Client, error) {
```
willie-yao, line 348: Where is this function used?

controllers/azurejson_machinepool_controller_test.go
```diff
@@ -210,3 +217,140 @@ func TestAzureJSONPoolReconciler(t *testing.T) {
...
+				},
+			},
+		},
+	}
+
+	client := fake.NewClientBuilder().WithScheme(scheme).WithObjects(azureMP, ownerMP, clusterA, azureCluster, sec).Build()
+	rec := AzureJSONMachinePoolReconciler{
+		Client:           client,
+		Recorder:         record.NewFakeRecorder(42),
+		ReconcileTimeout: 20 * time.Second,
```
willie-yao, line 345: I think we can use`reconciler.DefaultLoopTimeout` here rather than hard-coding the value

controllers/azurejson_machinepool_controller_test.go
troy0820: I didn't want to shadow any variable but I can change this.

controllers/azurejson_machinepool_controller_test.go
troy0820: `getClient` equals `identities.NewClient` 

If we don't patch that we have to pass an authorizer into the `identities.NewClient` func and try to auth to Azure.  Even if we pass an mocked authorizer, it will still try to go out and make the call to Aure to provide the "identity" client.  

This allows us to override this method during testing but not during production.  

In the test we do this 
```go
 getClient = func(auth azure.Authorizer) (identities.Client, error) {
       mockClient := mock_identities.NewMockClient(ctrlr)
      mockClient.EXPECT().GetClientID(gomock.Any(), gomock.Any()).Return(id, nil)
      return mockClient, nil
     }
```
So when we call `Reconcile, it doesn't try to go out and make a call to Azure but use our mock_identities caller in this method.`

So basically "patching" the method so we can test or else we get blocked here from a unit perspective.

controllers/azurejson_machinepool_controller_test.go
troy0820: It doesn't shadow, so I'll change this.

controllers/azurejson_machinepool_controller_test.go
troy0820: I would definitely do this but to migrate from this to that may be additional effort that may pollute the addition of this test.  

Also, setting up fixtures to settle all cases where that may be useful may also pollute the repo where that may need to be tested to ensure code coverage (I know that metric isn't bound to quality, but I digress).  

I can create an issue and follow up to call that out as something we would like to have.

controllers/azurejson_machinepool_controller_test.go
troy0820: Updated on latest push, resolving.

controllers/azurejson_machinepool_controller_test.go
troy0820: Utilized `fmt.Sprintf()` convention as suggested.  Resolving.

controllers/azurejson_machinepool_controller_test.go
troy0820: Changed to cluster as suggested, resolving.

controllers/azurejson_machinepool_controller_test.go
troy0820: Utilized the `newScheme` test method to grab all the schemes as suggested.  Resolving.

controllers/azurejson_machinepool_controller.go
troy0820: Updated this to include the space unintended for this change.  Resolving.

controllers/azurejson_machinepool_controller_test.go
willie-yao: Sounds good! I think it would be good to leave as-is for the current PR

willie-yao (approved):
/lgtm

Thanks for your awesome work on this!

CecileRobertMichon (commented):
/approve 

Welcome @troy0820 🥳

PR #4122, opened 2023-10-12
fix NodeTaints and NodeLabels return type for AzureManagedMachinePools

nawazkh:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:
- CAPZ will send out the below return types to AKS API so that the AKS API clears-out `NodeLabels` and `NodeTaints` respectively.
  - an empty Map when the `NodeLabels` are deleted.
  - an empty pointer array when the `NodeTaints` are deleted.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #4116 

**Special notes for your reviewer**:

- [x] cherry-pick candidate

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
fix NodeTaints and NodeLabels return type for AzureManagedMachinePools
```

PR #4295, opened 2023-11-20
test deleting all tags, node labels, and node taints in AKS e2e

### LINKED ISSUE #4266: Test deleting all tags, node labels, and node taints in AKS e2e
/kind cleanup
/area managedclusters

**What needs cleanup:**

Add back the logic to test deleting all tags in the AKS e2e tests:
- https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/4069/files#r1391753077
- https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/4069/files#r1372225084

And add similar logic in the tests for node labels and node taints.

nojnhuh:
**What type of PR is this?**
/kind cleanup

This PR adds tests that remove all defined tags, node labels, and node taints on AKS resources. Deleting all tags had been previously tested but was removed in #4069 due to an ASO limitation at the time that has since been lifted. Deleting all node labels and taints was first made possible in #4122 and has never been tested in e2e before.

Fixes #4266

- [X] cherry-pick candidate

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/aks_tags.go
+++ b/test/e2e/aks_tags.go
@@ -27,6 +27,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/containerservice/armcontainerservice/v4"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	"golang.org/x/exp/maps"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/utils/ptr"
 	infrav1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
@@ -107,10 +108,10 @@ func AKSAdditionalTagsSpec(ctx context.Context, inputGetter func() AKSAdditional
 		}, inputGetter().WaitForUpdate...).Should(Succeed())
 
 		By("Creating tags for control plane")
-		expectedTags = infrav1.Tags{
-			"test":    "tag",
-			"another": "value",
-		}
+		// Preserve "creationTimestamp" so the RG cleanup doesn't fire on this cluster during this test.
+		expectedTags = maps.Clone(initialTags)
+		expectedTags["test"] = "tag"
+		expectedTags["another"] = "value"
 		Eventually(func(g Gomega) {
 			g.Expect(mgmtClient.Get(ctx, client.ObjectKeyFromObject(infraControlPlane), infraControlPlane)).To(Succeed())
 			infraControlPlane.Spec.AdditionalTags = expectedTags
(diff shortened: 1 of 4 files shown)


CecileRobertMichon:
@nojnhuh fyi we found an issue with @jackfrancis last week where the test deletes the `creationTimestamp` tag on the AzureManagedControlPlane which is synced to the MC_ rg by AKS, so the tag gets deleted from the RG. This causes the test sub RG cleanup job to delete the node resource group mid-test when it happens to be running at the same time, causing flakes (rg cleanup uses tags to know if a rg is old enough to delete and deleted everything without a tag https://[github.com](https://github.com/Azure/rg-cleanup)/Azure/rg-cleanup).

So we might not be able to remove all tags as-is, we might need to preserve some of the "essential" tags, at least on the RGs.

nojnhuh:
@CecileRobertMichon Ah, that makes sense and sounds like the real root cause of #3955. Maybe we can skip the "delete all tags" step for the AzureManagedControlPlane and keep it for the AzureManagedMachinePools. I think the mechanism in ASO is similar enough that that would still give us enough confidence that that still works for ASO ManagedClusters if we only test it for ManagedClustersAgentPools.

CecileRobertMichon (commented):
/lgtm
/assign @jackfrancis

nojnhuh:
@jackfrancis Could you PTAL?

test/e2e/aks_tags.go
```diff
@@ -129,16 +130,14 @@ func AKSAdditionalTagsSpec(ctx context.Context, inputGetter func() AKSAdditional
...
-			Eventually(checkTags, input.WaitForUpdate...).Should(Succeed())
-		}
+		By("Restoring initial tags for control plane")
+		expectedTags = initialTags
+		Eventually(func(g Gomega) {
+			g.Expect(mgmtClient.Get(ctx, client.ObjectKeyFromObject(infraControlPlane), infraControlPlane)).To(Succeed())
+			infraControlPlane.Spec.AdditionalTags = expectedTags
+			g.Expect(mgmtClient.Update(ctx, infraControlPlane)).To(Succeed())
+		}, inputGetter().WaitForUpdate...).Should(Succeed())
+		Eventually(checkTags, input.WaitForUpdate...).Should(Succeed())
```
jackfrancis, line 140: Looks like we're skipping the "delete all tags from the AzureManagedControlPlane resource" test case, because we now know if we do that we introduce test flakes?

jackfrancis:
/lgtm

(added one clarifying question, I don't expect it will result in any changes)

PR #340, opened 2019-12-10
✨Add support for BYO vnet

CecileRobertMichon:
**What this PR does / why we need it**: Allow user to create a cluster using a pre-existing Azure virtual network. The vnet can either be in the target cluster RG or in a different RG (in the same subscription for now). When the vnet is provided, we expect the the subnets for control plane and node to also be created. The route table and NSGs are optional but creation will be skipped if the vnet is not managed by capz.

Example AzureCluster spec for a vnet in a different resource group:

```
apiVersion: infrastructure.cluster.x-k8s.io/v1alpha2
kind: AzureCluster
metadata:
  name: cluster-byo-vnet
  namespace: default
spec:
  location: southcentralus
  networkSpec:
    vnet:
      resourceGroup: custom-vnet
      name: my-vnet
    subnets:
      - name: control-plane-subnet
        role: control-plane
      - name: node-subnet
        role: node
  resourceGroup: cluster-byo-vnet
```

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #314 

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**Release note**:

```release-note
Add support for custom and existing vnet
```

PR #4300, opened 2023-11-20
convert subnets and virtualnetworks to ASO

### LINKED ISSUE #3528: ASO service: virtualnetworks and subnets

nojnhuh:
**What type of PR is this?**
/kind cleanup

**What this PR does / why we need it**: This PR updates the subnets and virtualnetworks services to use ASO.

Fixes #3528

This is split into two commits with one for each service. I kept these in the same PR though because converting subnets to ASO before virtualnetworks would have required some gymnastics to get the unit tests in ./controllers to pass that I didn't think were worth ironing out. I intend to squash this into one commit before merging since the unit tests don't pass for the first commit that only updates subnets.

/hold for squash

- [ ] cherry-pick candidate

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Release note**:

```release-note
subnets and virtualnetworks are now managed with ASO
```


azure/services/subnets/spec.go
```diff
@@ -43,107 +45,71 @@ type SubnetSpec struct {
...
+// Parameters implements azure.ASOResourceSpecGetter.
+func (s *SubnetSpec) Parameters(ctx context.Context, existing *asonetworkv1.VirtualNetworksSubnet) (parameters *asonetworkv1.VirtualNetworksSubnet, err error) {
+	subnet := existing
+	if subnet == nil {
+		subnet = &asonetworkv1.VirtualNetworksSubnet{}
 	}
 
-	if !s.IsVNetManaged {
-		// TODO: change this to terminal error once we add support for handling them
-		return nil, errors.Errorf("custom vnet was provided but subnet %s is missing", s.Name)
```
nojnhuh, line 76: This check is currently being removed and I'll edit this comment to explain why later.

azure/scope/cluster.go
```diff
@@ -25,7 +25,8 @@ import (
 	"strconv"
 	"strings"
 
-	asonetworkv1 "github.com/Azure/azure-service-operator/v2/api/network/v1api20220701"
+	asonetworkv1api20201101 "github.com/Azure/azure-service-operator/v2/api/network/v1api20201101"
+	asonetworkv1api20220701 "github.com/Azure/azure-service-operator/v2/api/network/v1api20220701"
```
CecileRobertMichon, line 29: why do we need 2 different network API versions?

azure/services/subnets/spec.go
CecileRobertMichon: I think that was actually a requested user feature in the past (there might even be an issue for it somewhere)

How do we handle delete for those vnets/subnets?

azure/scope/cluster.go
nojnhuh: ASO only has subnet/vnet resources with 20201101 and nat gateways and bastion hosts with 20220701: https://azure.github.io/azure-service-operator/reference/network/.

Compared to AKS resources which have multiple API versions available: https://azure.github.io/azure-service-operator/reference/containerservice/

azure/services/subnets/spec.go
nojnhuh: My archaeology skills led me here, where it seems this was added as a part of a general effort to support BYO vnet: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/340/files#diff-6d4bea1a244164bf3d16b025868b878dffcf0c78963b59bad9c189bcee04d35eR87-R90

CAPZ would take full responsibility for any subnet it creates, whether it's part of a managed or unmanaged vnet. So a managed subnet in an unmanaged vnet can be deleted by CAPZ.

nojnhuh: ^ But in that same scenario, the unmanaged vnet would not be deleted.

nojnhuh:
@willie-yao @nawazkh PTAL

nawazkh:
Started reviewing this, gonna take a bit more time.

willie-yao:
Same, taking a look now

azure/scope/managedcontrolplane.go
```diff
@@ -406,20 +400,14 @@ func (s *ManagedControlPlaneScope) IsVnetManaged() bool {
...
-	if vnetDescriber == nil {
-		virtualNetworksSvc, err := virtualnetworks.New(s)
-		if err != nil {
-			log.Error(err, "failed to create virtualnetworks service")
-			return false
-		}
-		vnetDescriber = virtualNetworksSvc
-	}
-	isManaged, err := vnetDescriber.IsManaged(ctx)
+	vnet := s.VNetSpec().ResourceRef()
```
Jont828, line 403: Much cleaner :)

azure/services/subnets/spec.go
```diff
@@ -38,112 +40,77 @@ type SubnetSpec struct {
...
-// OwnerResourceName returns the name of the VNet that owns this subnet.
-func (s *SubnetSpec) OwnerResourceName() string {
-	return s.VNetName
+// ResourceRef implements azure.ASOResourceSpecGetter.
+func (s *SubnetSpec) ResourceRef() *asonetworkv1.VirtualNetworksSubnet {
+	return &asonetworkv1.VirtualNetworksSubnet{
+		ObjectMeta: metav1.ObjectMeta{
+			// s.Name isn't unique per-cluster, so combine with vnet name to avoid collisions.
+			// ToLower makes the name compatible with standard Kubernetes name requirements.
+			Name:      s.VNetName + "-" + strings.ToLower(s.Name),
```
Jont828, line 53: Just to clarify, does this name have any correlation to the name of the subnet in Azure, or is it up to us on how to implement it?

azure/services/subnets/spec_test.go
```diff
@@ -20,291 +20,143 @@ import (
...
-			fields: fields{
-				Name:           "my-subnet",
-				ResourceGroup:  "my-rg",
-				SubscriptionID: "123",
-				IsVNetManaged:  false,
-			},
-			args: args{
-				existingSubnet: armnetwork.Subnet{
-					Name: ptr.To("my-subnet"),
+			name: "no existing subnet",
```
Jont828, line 39: It seems like we're just testing the case where there is an existing subnet and one where there isn't. Do you think it's worth preserving any of the other test cases from before, or would they not really be checking any new control flow?

azure/services/subnets/subnets.go
```diff
@@ -19,133 +19,41 @@ package subnets
...
 
-	s.Scope.UpdateDeleteStatus(infrav1.SubnetsReadyCondition, serviceName, result)
-	return result
-}
-
-// IsManaged returns true if the route tables' lifecycles are managed.
-func (s *Service) IsManaged(ctx context.Context) (bool, error) {
-	_, _, done := tele.StartSpanWithLogger(ctx, "subnets.Service.IsManaged")
-	defer done()
+	name := subnet.AzureName()
```
Jont828, line 54: Great cleanup!!!

azure/services/subnets/subnets_test.go
```diff
@@ -18,423 +18,40 @@ package subnets
...
-			err := s.Delete(context.TODO())
-			if tc.expectedError != "" {
-				g.Expect(err).To(HaveOccurred())
-				g.Expect(err).To(MatchError(tc.expectedError))
-			} else {
-				g.Expect(err).NotTo(HaveOccurred())
-			}
-		})
-	}
+func TestPostCreateOrUpdateResourceHook(t *testing.T) {
```
Jont828, line 31: Glad to see this test file getting simplified

azure/services/virtualnetworks/spec.go
```diff
@@ -19,59 +19,67 @@ package virtualnetworks
...
-		Location:         ptr.To(s.Location),
-		ExtendedLocation: converters.ExtendedLocationToNetworkSDK(s.ExtendedLocation),
-		Properties: &armnetwork.VirtualNetworkPropertiesFormat{
-			AddressSpace: &armnetwork.AddressSpace{
-				AddressPrefixes: azure.PtrSlice(&s.CIDRs),
-			},
-		},
-	}, nil
+// WasManaged implements azure.ASOResourceSpecGetter.
+func (s *VNetSpec) WasManaged(resource *asonetworkv1.VirtualNetwork) bool {
```
Jont828, line 83: Q: is there a reason this is `WasManaged` instead of `IsManaged`?

azure/services/virtualnetworks/virtualnetworks.go
```diff
@@ -19,169 +19,65 @@ package virtualnetworks
...
 	if err != nil {
-		return false, err
+		return errors.Wrap(err, "failed to list subnets")
 	}
-
-	tagsMap := make(map[string]*string)
-	if result.Properties != nil && result.Properties.Tags != nil {
-		tagsMap = result.Properties.Tags
+	for _, subnet := range subnets.Items {
+		scope.UpdateSubnetCIDRs(subnet.AzureName(), converters.GetSubnetAddresses(subnet))
```
Jont828, line 73: I saw we also call `scope.UpdateSubnetCIDRs` in `subnets.go`. Is there a reason for doing this in two separate places?

azure/services/virtualnetworks/virtualnetworks_test.go
```diff
@@ -18,322 +18,94 @@ package virtualnetworks
...
-		AdditionalTags: map[string]string{"foo": "bar"},
-	}
-
-	managedTags = armresources.TagsResource{
-		Properties: &armresources.Tags{
-			Tags: map[string]*string{
-				"foo": ptr.To("bar"),
-				"sigs.k8s.io_cluster-api-provider-azure_cluster_test-cluster": ptr.To("owned"),
+func TestPostCreateOrUpdateResourceHook(t *testing.T) {
+	t.Run("failed to create or update", func(t *testing.T) {
```
Jont828, line 39: Q: Do you think it's worth it to set up a `testcases` struct like before here? I think it's probably fine now since there's only two test cases, but in the future if we have further logic after creating or updating, we might need to refactor.

Jont828 (commented):
Nice work, glad to see a lot of old code getting cleaned up with this change. I'm not too familiar with ASO so I did have a question: what part of the code is responsible for making sure ASO CRDs get created?

azure/scope/cluster.go
```diff
@@ -25,7 +25,8 @@ import (
 	"strconv"
 	"strings"
 
-	asonetworkv1 "github.com/Azure/azure-service-operator/v2/api/network/v1api20220701"
+	asonetworkv1api20201101 "github.com/Azure/azure-service-operator/v2/api/network/v1api20201101"
```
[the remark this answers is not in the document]
nojnhuh: There are multiple 2020 versions for the network API group that we'll end up using eventually, so I'd vote to keep these as-is just so we don't have to get too creative to avoid collisions. 

https://azure.github.io/azure-service-operator/reference/network/

azure/services/subnets/spec.go
nojnhuh: Yes, this becomes the name of the Azure resource by default. To override it, every ASO resource has a `spec.azureName` field, which is defined for the subnet below in `Parameters`. Overall this should be creating subnets that look exactly the same as before.

azure/services/subnets/spec_test.go
nojnhuh: `Parameters` ultimately has fewer responsibilities now than it had before (doesn't error for unmanaged vnet, doesn't diff with existing subnet) and it seemed like some of the old cases were redundant, so I think these two cases provide adequate coverage.

azure/services/virtualnetworks/spec.go
nojnhuh: `WasManaged` corresponds to the current `IsManaged` for SDK-driven services. It tells CAPZ whether or not this resource should be adopted if no ASO resource already exists so we can onboard existing workload clusters to ASO.

azure/services/virtualnetworks/virtualnetworks.go
nojnhuh: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2339

azure/services/virtualnetworks/virtualnetworks_test.go
nojnhuh: I'd vote to cross that bridge when we get there.

azure/converters/subnets.go
```diff
@@ -17,21 +17,17 @@ limitations under the License.
...
-	} else if subnet.Properties != nil && subnet.Properties.AddressPrefixes != nil {
-		for _, address := range subnet.Properties.AddressPrefixes {
-			if address != nil {
-				addresses = append(addresses, *address)
-			}
-		}
+	if subnet.Status.AddressPrefix != nil {
+		addresses = []string{ptr.Deref(subnet.Status.AddressPrefix, "")}
+	} else if subnet.Status.AddressPrefixes != nil {
+		addresses = subnet.Status.AddressPrefixes
```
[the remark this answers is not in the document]
nojnhuh: These values are no longer pointers, just plain strings.

azure/services/subnets/spec.go
```diff
@@ -38,112 +40,77 @@ type SubnetSpec struct {
...
+	//nolint:prealloc // pre-allocating this slice isn't going to make any meaningful performance difference
+	// and makes it harder to keep this value nil when s.ServiceEndpoints is empty as is necessary.
+	var serviceEndpoints []asonetworkv1.ServiceEndpointPropertiesFormat
 	for _, se := range s.ServiceEndpoints {
-		se := se
-		serviceEndpoints = append(serviceEndpoints, armnetwork.ServiceEndpointPropertiesFormat{Service: ptr.To(se.Service), Locations: azure.PtrSlice(&se.Locations)})
+		serviceEndpoints = append(serviceEndpoints, asonetworkv1.ServiceEndpointPropertiesFormat{Service: ptr.To(se.Service), Locations: se.Locations})
 	}
-	subnetProperties.ServiceEndpoints = azure.PtrSlice(&serviceEndpoints)
+	subnet.Spec.ServiceEndpoints = serviceEndpoints
```
nawazkh, line 108: > makes it harder to keep this value nil when s.ServiceEndpoints is empty as is necessary

We could have something like
```suggestion
if len(s.ServiceEndpoints) != 0 {
	serviceEndpoints := make([]asonetworkv1.ServiceEndpointPropertiesFormat, 0, len(s.ServiceEndpoints))
	for...
	.
	.
	subnet.Spec.ServiceEndpoints = serviceEndpoints
}
```

But happy with current implementation as `serviceEndpoints`'s scope is limited to this function.

CecileRobertMichon (commented):
/lgtm

nawazkh (approved):
Great work @nojnhuh!
Mostly some minor thoughts, nothing blocking.
/lgtm

azure/converters/subnets.go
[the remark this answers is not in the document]
nojnhuh: > 1. If yes, shall we be considering either one of it? Is there a reason to choose `subnet.Status.AddressPrefix` over `subnet.Status.AddressPrefixes` ?

I'm not entirely sure, but I was trying to keep this behavior the same in this PR so I think changing it should be done separately.

> 2. Can we append both if `subnet.Status.AddressPrefix` and `subnet.Status.AddressPrefixes` are different ?

I don't think this would ever happen. My guess is that both exist only because it started out as the singular field and then the API evolved to support multiple but kept the old field for compatibility.

azure/scope/cluster.go
```diff
@@ -423,16 +424,28 @@ func (s *ClusterScope) SubnetSpecs() []azure.ResourceSpecGetter {
...
+	if s.Vnet().ResourceGroup != s.ResourceGroup() {
+		specs = append(specs, &groups.GroupSpec{
+			Name:           s.Vnet().ResourceGroup,
+			Namespace:      s.Namespace(),
+			Location:       s.Location(),
+			ClusterName:    s.ClusterName(),
+			AdditionalTags: s.AdditionalTags(),
+			Owner:          owner,
+		})
+	}
```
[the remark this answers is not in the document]
nojnhuh: There are not as far as I'm aware. We can evaluate each service as we convert them though.

azure/services/subnets/spec.go
nojnhuh: Right, but that additional necessary syntax doesn't outweigh the potential performance improvement IMO.

azure/converters/subnets.go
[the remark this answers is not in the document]
CecileRobertMichon: > My guess is that both exist only because it started out as the singular field and then the API evolved

That's correct

AddressPrefixes was added to support dual stack clusters (one ipv4 cidr and one ipv6). Originally only single stack ipv4 was supported. In the Azure SDK the same two fields exist and only one can be set.

azure/scope/managedcontrolplane.go
```diff
@@ -504,7 +492,7 @@ func (s *ManagedControlPlaneScope) ManagedClusterSpec() azure.ASOResourceSpecGet
 		DNSServiceIP:      s.ControlPlane.Spec.DNSServiceIP,
 		VnetSubnetID: azure.SubnetID(
 			s.ControlPlane.Spec.SubscriptionID,
-			s.VNetSpec().ResourceGroupName(),
+			s.Vnet().ResourceGroup,
 			s.ControlPlane.Spec.VirtualNetwork.Name,
 			s.ControlPlane.Spec.VirtualNetwork.Subnet.Name,
```
willie-yao, line 497: Should these be also using `s.Vnet()`? Just to keep things consistent with the line above.

azure/scope/managedcontrolplane.go
```diff
@@ -861,20 +849,20 @@ func (s *ManagedControlPlaneScope) AvailabilityStatusFilter(cond *clusterv1.Cond
...
+			ResourceGroup:              s.Vnet().ResourceGroup,
 			Location:                   privateEndpoint.Location,
 			CustomNetworkInterfaceName: privateEndpoint.CustomNetworkInterfaceName,
 			PrivateIPAddresses:         privateEndpoint.PrivateIPAddresses,
 			SubnetID: azure.SubnetID(
 				s.ControlPlane.Spec.SubscriptionID,
-				s.VNetSpec().ResourceGroupName(),
+				s.Vnet().ResourceGroup,
 				s.ControlPlane.Spec.VirtualNetwork.Name,
 				s.ControlPlane.Spec.VirtualNetwork.Subnet.Name,
```
willie-yao, line 867: Same comment as above ^^

willie-yao (commented):
Looks good pending a minor comment

azure/scope/managedcontrolplane.go
nojnhuh: `s.Vnet()` doesn't define the `Subnet`, and all the other places using the vnet resource group refer to it like this, so it seems like there's some inconsistency necessary here without some wider refactoring.

nojnhuh:
@CecileRobertMichon This should be ready for a final review.

nojnhuh:
squashed!
/hold cancel

nojnhuh:
@willie-yao  @CecileRobertMichon Resolved conflicts, PTAL.

/hold for squash

CecileRobertMichon (commented):
/lgtm

nojnhuh:
squashed!
/hold cancel

PR #4274, opened 2023-11-15
Fix "log.SetLogger(...) was never called" error

willie-yao:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #3999

**Special notes for your reviewer**:

- [ ] cherry-pick candidate

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

PR #4301, opened 2023-11-21
fix log flags

nojnhuh:
**What type of PR is this?**
/kind bug

This PR fixes a regression by #4274 where the command line flag configuration from main.go was no longer matching the logging implementation being set up, so flags like `-v` were no longer having any effect. This PR removes the `textlogger.Config` from main.go and instead sets up command line flags with `klog.InitFlags`.

Fixes #

- [ ] cherry-pick candidate <-- this bug doesn't exist on any release branch

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```
^ this bug has never affected any users of published releases.

### DIFF
--- a/main.go
+++ b/main.go
@@ -41,7 +41,6 @@ import (
 	cgrecord "k8s.io/client-go/tools/record"
 	"k8s.io/component-base/logs"
 	"k8s.io/klog/v2"
-	"k8s.io/klog/v2/textlogger"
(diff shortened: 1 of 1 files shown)


CecileRobertMichon (commented):
/lgtm
/approve

good catch!

